PR #1914, opened 2020-09-25
Feature: Store all transactions as they are processed for debugging

kantai:
This adds a compile-enabled feature for storing to a table a copy of the transaction ID and transaction data whenever a transaction is processed. This lets you figure out whether or not a particular TXID was processed, and if so, what the result of it was.

```
$  cargo build --workspace --features tx_log
$ ./target/debug/stacks-node krypton
```

And then to find a transaction...

```
$ echo 'SELECT index_block_hash, result FROM transactions WHERE txid = "9e4ef900f521dd7c07c1c22e4b58b33df656880c8035bdb5e2b0c44ba8356af6"' | sqlite3 -separator ", " /tmp/stacks-testnet-f7c1618951418340/chainstate/chain-00000080-testnet/vm/index
87cbab207957c48bcb5b56943b26ef78c30ae4d6110dbdc7af21b9d93ca3946d, (ok true)
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,6 +91,7 @@ criterion = "0.3"
 developer-mode = []
 default = ["developer-mode"]
 monitoring_prom = ["prometheus"]
+tx_log = []
 
 [target.'cfg(all(target_arch = "x86_64", not(target_env = "msvc")))'.dependencies]
 sha2-asm = "0.5.3"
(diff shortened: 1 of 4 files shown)


jcnelson (approved):
This LGTM, but do you maybe instead want to store all the txs in a separate database file altogether?  Then, this could be a config option, and you could clear out unwanted state if you disable it later.

lgalabru (approved):
Looks good to me!

kantai:
> This LGTM, but do you maybe instead want to store all the txs in a separate database file altogether? Then, this could be a config option, and you could clear out unwanted state if you disable it later.

I think that's a good idea if we want this to be something that developers would frequently turn on and off, but for now, I think that'd increase the footprint of this change a lot (it would require obtaining a new database transaction in `append_block`, which there's not really an elegant way to do).

PR #1915, opened 2020-09-25
Log broadcasted messages

jcnelson:


kantai (approved):
This looks fine to me, though I think this could be implemented as the Display trait for the messages.

PR #1916, opened 2020-09-26
Fix event ID resource leak for connecting sockets that time out

jcnelson:
This should cause the follower node to _not_ freeze up after a few hours.

### DIFF
--- a/src/net/p2p.rs
+++ b/src/net/p2p.rs
@@ -1411,7 +1411,7 @@ impl PeerNetwork {
     }
 
     /// Remove unresponsive peers
-    fn disconnect_unresponsive(&mut self) -> () {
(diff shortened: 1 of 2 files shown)


psq (approved):
looks good, and works on my follower node, which before was not able to catch up with master node.
does not seem to be related to miner disagreeing on pox consensus

kantai (commented):
This looks good to me, but can you add a regression test that makes sure a socket that is deregistered in the connecting state is removed from the event ID vec? (preferably a test that doesn't need to be #ignored)

jcnelson:
> This looks good to me, but can you add a regression test that makes sure a socket that is deregistered in the connecting state is removed from the event ID vec? (preferably a test that doesn't need to be #ignored)

Done

PR #1917, opened 2020-09-27
Fix/remove downloader index hash cache

jcnelson:
This removes the index hash cache from the block downloader.  This had been used to prevent blocks from being re-downloaded, but this functionality is already checked by simply looking at the staging DB and seeing if a block or microblock stream is already stored.  The presence of this cache seems to be interfering with the relayer discarding blocks in the event the PoX bit vector gets invalidated -- the act of discarding the downloaded block does _not_ clear this cache.  This, in turn, seems to be causing nodes unlucky enough to experience a PoX bit vector invalidation to download blocks and never re-download them if they were truly necessary for the chainstate.

### DIFF
--- a/src/net/download.rs
+++ b/src/net/download.rs
@@ -191,11 +191,6 @@ pub struct BlockDownloader {
 
     /// how often to download
     download_interval: u64,
-
-    /// set of blocks and microblocks we have successfully downloaded (even if they haven't been
-    /// stored yet)
-    blocks_downloaded: HashSet<StacksBlockId>,
-    microblocks_downloaded: HashSet<StacksBlockId>
 }
 
 impl BlockDownloader {
@@ -234,9 +229,6 @@ impl BlockDownloader {
             broken_neighbors: vec![],
 
             download_interval: download_interval,
-
-            blocks_downloaded: HashSet::new(),
-            microblocks_downloaded: HashSet::new(),
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/net/download.rs
```diff
@@ -725,7 +717,7 @@ impl PeerNetwork {
 
     /// Create block request keys for a range of blocks that are available but that we don't have in a given range of
     /// sortitions.  The same keys can be used to fetch confirmed microblock streams.
-    fn make_requests(&mut self, sortdb: &SortitionDB, chainstate: &StacksChainState, downloader: &BlockDownloader, start_sortition_height: u64, microblocks: bool) -> Result<HashMap<u64, VecDeque<BlockRequestKey>>, net_error> {
+    fn make_requests(&mut self, sortdb: &SortitionDB, chainstate: &StacksChainState, _downloader: &BlockDownloader, start_sortition_height: u64, microblocks: bool) -> Result<HashMap<u64, VecDeque<BlockRequestKey>>, net_error> {
```
psq, line 720: if the param is no longer needed, shouldn't you remove it?  this will change a few functions, but all contained in `donwload.rs`

psq:
First run from scratch was successful :rocket:  Will run a few more...

psq:
restarted a miner from scratch 3 more times, no issues

psq (approved):

kantai (approved):
This looks good to me, though I agree with @psq that we should remove the BlockDownloader parameter to `make_requests` if it isn't being used.

PR #1929, opened 2020-09-30
Ability to build a chain anchored to Bitcoin Testnet

lgalabru:
This PR is addressing a few issues that were blocking me from running a node on a Bitcoin Testnet network.
Proof of a functional PoX setup available here: https://tbtc.bitaps.com/mxVFsFW5N4mu1HPkxPttorvocvzeZ7KZyk

### DIFF
--- a/src/burnchains/bitcoin/spv.rs
+++ b/src/burnchains/bitcoin/spv.rs
@@ -21,6 +21,7 @@ use std::cmp;
 use std::fs;
 use std::io::{Read, Seek, SeekFrom, Write};
 use std::ops::Deref;
+use std::collections::VecDeque;
 
 use deps::bitcoin::blockdata::block::{BlockHeader, LoneBlockHeader};
 use deps::bitcoin::blockdata::constants::genesis_block;
@@ -291,41 +292,41 @@ impl SpvClient {
         }
 
         for i in interval_start..interval_end {
-            let (bits, difficulty) = match self.get_target(i)? {
-                Some(x) => x,
-                None => {
-                    // out of headers
-                    return Ok(());
-                }
-            };
-
+            let mut headers = VecDeque::new();
             for block_height in
                 (i * BLOCK_DIFFICULTY_CHUNK_SIZE)..((i + 1) * BLOCK_DIFFICULTY_CHUNK_SIZE)
             {
-                match self.read_block_header(block_height)? {
+                let header_i = match self.read_block_header(block_height)? {
+                    None => return Ok(()),
+                    Some(res) => res.header,
+                };
+
+                let (bits, difficulty) = match self.get_target(block_height, &header_i, &headers, i)? {
+                    Some(x) => x,
                     None => {
                         // out of headers
                         return Ok(());
                     }
-                    Some(header_i) => {
-                        if header_i.header.bits != bits {
-                            error!("bits mismatch at block {} of {} (offset {} interval {} of {}-{}): {:08x} != {:08x}",
-                                   block_height, self.headers_path, block_height % BLOCK_DIFFICULTY_CHUNK_SIZE, i, interval_start, interval_end, header_i.header.bits, bits);
-                            return Err(btc_error::InvalidPoW);
-                        }
-                        let header_hash = header_i.header.bitcoin_hash().into_le();
-                        if difficulty <= header_hash {
-                            error!(
-                                "block {} hash {} has less work than difficulty {} in {}",
-                                block_height,
-                                header_i.header.bitcoin_hash(),
-                                difficulty,
-                                self.headers_path
-                            );
-                            return Err(btc_error::InvalidPoW);
-                        }
-                    }
                 };
+        
+                if header_i.bits != bits {
+                    error!("bits mismatch at block {} of {} (offset {} interval {} of {}-{}): {:08x} != {:08x}",
+                            block_height, self.headers_path, block_height % BLOCK_DIFFICULTY_CHUNK_SIZE, i, interval_start, interval_end, header_i.bits, bits);
+                    return Err(btc_error::InvalidPoW);
+                }
+                let header_hash = header_i.bitcoin_hash().into_le();
+                if difficulty <= header_hash {
+                    error!(
+                        "block {} hash {} has less work than difficulty {} in {}",
+                        block_height,
+                        header_i.bitcoin_hash(),
+                        difficulty,
+                        self.headers_path
+                    );
+                    return Err(btc_error::InvalidPoW);
+                }
+
+                headers.push_front(header_i);
             }
         }
         return Ok(());
@@ -717,10 +718,17 @@ impl SpvClient {
     /// Determine the target difficult over a given difficulty adjustment interval
     /// the `interval` parameter is the difficulty interval -- a 2016-block interval.
     /// Returns (new bits, new target)
-    pub fn get_target(&self, interval: u64) -> Result<Option<(u32, Uint256)>, btc_error> {
-        let max_target = if self.network_id == BitcoinNetworkType::Regtest {
-            // in regtest mode there's no difficulty adjustment active.
-            // it uses the highest possible difficulty -- represents nBits = 0x207fffff
+    pub fn get_target(&self, current_header_height: u64, current_header: &BlockHeader, headers_in_range: &VecDeque<BlockHeader>, interval: u64) -> Result<Option<(u32, Uint256)>, btc_error> {
+        if interval == 0 {
+            panic!(
+                "Invalid argument: interval must be positive (got {})",
+                interval
+            );
+        }
+
+        // In Regtest mode there's no difficulty adjustment active.
+        // it uses the highest possible difficulty -- represents nBits = 0x207fffff
+        if self.network_id == BitcoinNetworkType::Regtest {
             return Ok(Some((
                 0x207fffff,
                 Uint256([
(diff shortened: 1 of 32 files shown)


testnet/stacks-node/conf/xenon-miner-conf.toml
```diff
@@ -9,9 +9,12 @@ miner = true
 [burnchain]
 chain = "bitcoin"
 mode = "xenon"
-peer_host = "xenon.blockstack.org"
+peer_host = "127.0.0.1"
```
jcnelson, line 12: Are we setting up some public testnet nodes?

src/burnchains/bitcoin/spv.rs
```diff
@@ -725,56 +740,73 @@ impl SpvClient {
...
             };
 
         // find actual timespan as being clamped between +/- 4x of the target timespan
-        let measured_timespan = (last_header.header.time - first_header.header.time) as u64;
+        let mut actual_timespan = (last_header.time - first_header.time) as u64;
         let target_timespan = BLOCK_DIFFICULTY_INTERVAL as u64;
-        let timespan_highpass = cmp::max(measured_timespan, target_timespan / 4);
-        let filtered_timespan = cmp::min(timespan_highpass, target_timespan * 4);
+        if actual_timespan < (target_timespan / 4) {
+            actual_timespan = target_timespan / 4;
```
jcnelson, line 796: Please revert this change between lines 793 and 809.  The original code better represents what's actually going on inside bitcoind.

src/burnchains/mod.rs
```diff
@@ -77,7 +74,7 @@ impl_array_hexstring_fmt!(Txid);
 impl_byte_array_newtype!(Txid, u8, 32);
 pub const TXID_ENCODED_SIZE: u32 = 32;
 
-#[derive(Serialize, Deserialize)]
+#[derive(Default, Serialize, Deserialize)]
```
jcnelson, line 77: Please remove `Default`.  It's better to have an explicit constructor for a default value, because then it will stay the same even if Rust's behavior changes.

testnet/stacks-node/conf/xenon-miner-conf.toml
lgalabru: I'm not sure, but in my mind the main point of xenon is to make it so community members run their own Bitcoin node.

src/burnchains/mod.rs
```diff
@@ -102,8 +99,6 @@ pub struct BurnchainParameters {
     chain_name: String,
     network_name: String,
     network_id: u32,
-    first_block_height: u64,
```
jcnelson, line 105: Please revert this.  The first block height and first header hash are very much burnchain _parameters_.  You should instead update the constructors for `BurnchainParameters` to take a first block height and first block hash.

src/burnchains/bitcoin/spv.rs
lgalabru: Can you share with me a link of the code in bitcoin that looks like the previous version? It probably has been updated since then, I've actually been looking at the code of Bitcoin, and this version is, factually, easier to compare to the original version:

```
unsigned int CalculateNextWorkRequired(const CBlockIndex* pindexLast, int64_t nFirstBlockTime, const Consensus::Params& params)
{
    if (params.fPowNoRetargeting)
        return pindexLast->nBits;


    // Limit adjustment step
    int64_t nActualTimespan = pindexLast->GetBlockTime() - nFirstBlockTime;
    if (nActualTimespan < params.nPowTargetTimespan/4)
        nActualTimespan = params.nPowTargetTimespan/4;
    if (nActualTimespan > params.nPowTargetTimespan*4)
        nActualTimespan = params.nPowTargetTimespan*4;


    // Retarget
    const arith_uint256 bnPowLimit = UintToArith256(params.powLimit);
    arith_uint256 bnNew;
    bnNew.SetCompact(pindexLast->nBits);
    bnNew *= nActualTimespan;
    bnNew /= params.nPowTargetTimespan;
   ...
```

(https://github.com/bitcoin/bitcoin/blob/78dae8caccd82cfbfd76557f1fb7d7557c7b5edb/src/pow.cpp#L49-L70)

src/burnchains/mod.rs
jcnelson: Some background: a `BurnchainParameters` instance is meant to be a set of immutable constants that uniquely identify one of the ways the Stacks node can attach to a burnchain.  Once we release mainnet, the first burnchain block height and first burnchain block hash on mainnet Bitcoin will become protocol constants for mainnet Stacks.

src/net/mod.rs
```diff
@@ -2195,41 +2198,78 @@ pub mod test {
...
-                            let boot_code_address = StacksAddress::from_string(&STACKS_BOOT_CODE_CONTRACT_ADDRESS.to_string()).unwrap();
-                            let boot_code_account = StacksAccount {
-                                principal: PrincipalData::Standard(StandardPrincipalData::from(boot_code_address.clone())),
-                                nonce: 0,
-                                stx_balance: STXBalance::zero(),
-                            };
-                            let boot_code_auth = TransactionAuth::Standard(TransactionSpendingCondition::Singlesig(SinglesigSpendingCondition {
+            let conf = config.clone();
+            let post_flight_callback = move |clarity_tx: &mut ClarityTx| {
+                if conf.setup_code.len() > 0 {
```
jcnelson, line 2203: The boot code is always going to be instantiated, regardless of `setup_code` (also, why is this no longer `init_code`?).  This conditional should be removed.

src/burnchains/bitcoin/spv.rs
jcnelson: I stand corrected.  If that's what's currently in bitcoind, then that's what we should use :+1:

src/burnchains/mod.rs
jcnelson: Just updating this comment from the architecture meeting.  In general, we'd like to move away from using `first_block_hash` altogether since (1) the SPV client will have all headers available anyway, and (2) the code shouldn't depend on knowing `first_block_hash` in the first place.  But, `first_block_height` is definitely an immutable protocol parameter, and should live in `BurnchainParameters`.

src/burnchains/mod.rs
lgalabru: Address with latest

src/burnchains/mod.rs
lgalabru: Address with latest, however we had to keep the `first_block_hash`, required for initializing the Chainstate and the SortitionDB

src/chainstate/stacks/db/mod.rs
```diff
@@ -517,6 +537,35 @@ pub const MINER_FEE_MINIMUM_BLOCK_USAGE: u64 = 80; // miner must share the first
...
+pub struct ChainStateBootData {
+    pub first_burnchain_block_hash: BurnchainHeaderHash,
+    pub first_burnchain_block_height: u32,
+    pub first_burnchain_block_timestamp: u32,
+    pub initial_balances: Vec<(PrincipalData, u64)>,
+    pub post_flight_callback: Option<Box<dyn FnOnce(&mut ClarityTx) -> ()>>,
+}
+
+impl ChainStateBootData {
+    pub fn preprend_post_flight_callback(
```
kantai, line 549: Typo: `preprend` -> `prepend`

kantai: Can you also add a rustdoc here that explains when the callback will be invoked, e.g., after initial balances are inserted, and after boot contracts are initialized, but before any previously inserted post flight callbacks?

kantai (approved):
Apart from a few small comments, this LGTM!

src/burnchains/mod.rs
```diff
@@ -124,10 +132,14 @@ impl BurnchainParameters {
             chain_name: "bitcoin".to_string(),
             network_name: BITCOIN_TESTNET_NAME.to_string(),
             network_id: BITCOIN_NETWORK_ID_TESTNET,
-            first_block_height: BITCOIN_FIRST_BLOCK_TESTNET,
-            first_block_hash: FIRST_BURNCHAIN_BLOCK_HASH_TESTNET.clone(),
             stable_confirmations: 7,
             consensus_hash_lifetime: 24,
+            first_block_height: 1891063,
```
jcnelson, line 137: Can you make these `const` values instead of hard-coded values?  This applies to `first_block_height`, `first_block_hash`, and `first_block_timestamp`.

src/chainstate/coordinator/mod.rs
```diff
@@ -257,12 +254,25 @@ impl<'a, T: BlockEventDispatcher, U: RewardSetProvider> ChainsCoordinator<'a, T,
     ) -> ChainsCoordinator<'a, T, (), U> {
         let burnchain = burnchain.clone();
 
+        let mut boot_data = ChainStateBootData {
```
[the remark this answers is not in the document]
jcnelson: Bump

src/chainstate/stacks/db/mod.rs
jcnelson: Honestly, this code scares me.  It makes it difficult to figure out what boot code gets executed, and in what order it gets executed.  This, in turn, makes auditing the boot block more difficult than it needs to be.  I'd much rather keep the code that instantiates the boot block in the same call site as the code that instantiates the chainstate (without any chaining), since it makes it much clearer as to what's supposed to happen when the chainstate is created.

src/burnchains/mod.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/1929/commits/6225b4ab7526b9a8454f1efcd759cd345fb6ce22

src/chainstate/coordinator/mod.rs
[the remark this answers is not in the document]
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/1929/commits/6225b4ab7526b9a8454f1efcd759cd345fb6ce22

src/chainstate/stacks/db/mod.rs
lgalabru: I understand the concern, and when I'm also not a fan of this nested callbacks, this construction / pattern is not something that this PR is introducing, it's already present in the `next` branch (https://github.com/blockstack/stacks-blockchain/blob/c0364bcf01a23fb2dfeb31721856e1b23285423e/src/chainstate/coordinator/mod.rs#L256-L288), the only thing that this branch is doing, is encapsulating all the elements in one struct, the ChainStateBootData.
I'm happy to take on this refactoring and see if and how sites can be merged, but I think that it should be addressed in a separate PR, so that we can unblock Xenon deployments now.

lgalabru: Comment added in https://github.com/blockstack/stacks-blockchain/pull/1929/commits/6225b4ab7526b9a8454f1efcd759cd345fb6ce22

lgalabru: Issue created https://github.com/blockstack/stacks-blockchain/issues/2066

jcnelson: Thanks!

jcnelson (approved):
LGTM!  Thanks @lgalabru!

PR #1937, opened 2020-10-02
PoX Delegation / Pooling

kantai:
This PR adds delegation and pooling support to the pox contract. It does so by adding four public methods:

* `delegate-stx` - This method creates a delegation relationship between the sender and the designated principal. This _does not_ lock the corresponding STX. Rather, it allows the designated principal to perform the lock, subject to the constraints of the sender:
     * a maximum amount the delegator is allowed to lock and stack
     * optional reward address that must be used
     * optional expiration burn block height
* `revoke-delegate-stx` - This method allows the sender to revoke their designated delegator
* `delegator-stack-stx` - This method is called by a delegator to _lock_ a single stacker's STX, _but does not_ add the pox-addr to the reward cycles. Rather, when invoked, it adds the amount of locked STX to a map of "partial" rewards. Then, delegators will call the `stack-aggregation-commit` method to finalize these partial rewards. This separation of the locking and commit operations allows delegators to invoke this method multiple times with different stackers in order to build up enough STX to clear the minimum reward threshold (i.e. pooling).
* `stack-aggregation-commit` - This method checks and adds a pox-addr to a reward cycle if the corresponding partially stacked amounts clear the threshold.

In addition, this PR makes a few changes to the PoX contract's behavior:

1. Reward addresses may be used multiple times. As long as each entry clears the minimum threshold, it is fine to include a reward address multiple times. There's a few use cases where users might want to do this, e.g., a user holds STX across multiple addresses, but would like to receive Stacking rewards at a single BTC address, and I don't think there's any performance consequences to allowing this (ultimately, any cases where a reward address would be used multiple times, would instead just be a case where the same number of different addresses were used).

2. `tx-sender` must be equal to `contract-caller`, unless a contract has received an allowance. Post conditions do not protect stacking, so it's dangerous to just use `tx-sender` without the check.

3.  The minimum threshold was previously always `liquid_supply / 20000`, which isn't the case in testnet, so this was updated to use the defined constant.

### DIFF
--- a/src/chainstate/stacks/boot/pox.clar
+++ b/src/chainstate/stacks/boot/pox.clar
@@ -14,6 +14,7 @@
 (define-constant ERR_STACKING_ALREADY_REJECTED 17)
 (define-constant ERR_STACKING_INVALID_AMOUNT 18)
 (define-constant ERR_NOT_ALLOWED 19)
+(define-constant ERR_STACKING_ALREADY_DELEGATED 20)
 
 ;; PoX disabling threshold (a percent)
 (define-constant POX_REJECTION_FRACTION u25)
@@ -64,6 +65,22 @@
     )
 )
 
+;; Delegation relationships
+(define-map delegation-state
+    ((stacker principal))
+    ((amount-ustx uint)              ;; how many uSTX delegated?
+     (delegated-to principal)        ;; who are we delegating?
+     (until-burn-ht (optional uint)) ;; how long does the delegation last?
+     ;; does the delegator _need_ to use a specific
+     ;;   pox recipient address?
+     (pox-addr (optional { version: (buff 1),
+                           hashbytes: (buff 20) }))))
+
+;; Delegator's allowed contract-callers
+(define-map delegator-contract-callers
+    ((delegator principal) (contract-caller principal))
+    ((until-burn-ht (optional uint))))
+
 ;; How many uSTX are stacked in a given reward cycle.
 ;; Updated when a new PoX address is registered, or when more STX are granted
 ;; to it.
@@ -87,16 +104,15 @@
     ((len uint))
 )
 
-;; When is a PoX address active?
-;; Used to check of a PoX address is registered or not in a given
-;; reward cycle.
-(define-map pox-addr-reward-cycles
-    ((pox-addr (tuple (version (buff 1)) (hashbytes (buff 20)))))
-    (
-        (first-reward-cycle uint)
-        (num-cycles uint)
-    )
-)
+;; how much has been locked up for this address before
+;;   committing?
+;; this map allows stackers to stack amounts < minimum
+;;   by paying the cost of aggregation during the commit
+(define-map partial-stacked-by-cycle
+  ((pox-addr (tuple (version (buff 1)) (hashbytes (buff 20))))
+   (reward-cycle uint)
+   (sender principal))
+  ((stacked-amount uint)))
 
 ;; Amount of uSTX that reject PoX, by reward cycle
 (define-map stacking-rejection
@@ -153,8 +169,33 @@
             )
         ;; no state at all
         none
-    )
-)
+    ))
+
+(define-private (check-delegation-caller-allowed)
+    (or (is-eq tx-sender contract-caller)
+        (let ((delegation-caller-allowed 
+                 ;; if not in the delegation/caller map, return false
+                 (unwrap! (map-get? delegator-contract-callers
+                                    { delegator: tx-sender, contract-caller: contract-caller })
+                          false)))
+          ;; is the caller allowance expired?
+          (if (< burn-block-height (unwrap! (get until-burn-ht delegation-caller-allowed) true))
+              (begin (map-delete delegator-contract-callers
+                                 { delegator: tx-sender, contract-caller: contract-caller })
+                     false)
+              true))))
+
+(define-private (get-check-delegation (stacker principal))
+    (let ((delegation-info (try! (map-get? delegation-state { stacker: stacker }))))
+      ;; did the existing delegation expire?
+      (if (match (get until-burn-ht delegation-info)
+                 until-burn-ht (> burn-block-height until-burn-ht)
+                 false)
+          ;; it expired, delete the entry and return none
+          (begin (map-delete delegation-state { stacker: stacker })
+                 none)
+          ;; delegation is active
+          (some delegation-info))))
 
 ;; Get the size of the reward set for a reward cycle.
 ;; Note that this does _not_ return duplicate PoX addresses.
@@ -166,26 +207,6 @@
     (default-to
         u0
         (get len (map-get? reward-cycle-pox-address-list-len { reward-cycle: reward-cycle }))))
-
-;; Is a PoX address registered anywhere in a given range of reward cycles?
-;; Checkes the integer range [reward-cycle-start, reward-cycle-start + num-cycles)
-;; Returns true if it's registered in at least one reward cycle in the given range.
-;; Returns false if it's not registered in any reward cycle in the given range.
-(define-private (is-pox-addr-registered (pox-addr (tuple (version (buff 1)) (hashbytes (buff 20))))
-                                        (reward-cycle-start uint)
-                                        (num-cycles uint))
-    (let (
-        (pox-addr-range-opt (map-get? pox-addr-reward-cycles { pox-addr: pox-addr }))
-    )
-    (match pox-addr-range-opt
-        ;; some range
-        pox-addr-range
-            (and (>= (+ reward-cycle-start num-cycles) (get first-reward-cycle pox-addr-range))
-                    (< reward-cycle-start (+ (get first-reward-cycle pox-addr-range) (get num-cycles pox-addr-range))))
-        ;; none
-        false
-    ))
-)
 
 ;; Add a single PoX address to a single reward cycle.
 ;; Used to build up a set of per-reward-cycle PoX addresses.
@@ -230,13 +251,14 @@
                                                             (amount-ustx uint)
                                                             (i uint))))
     (let ((reward-cycle (+ (get first-reward-cycle params) (get i params)))
+          (num-cycles (get num-cycles params))
           (i (get i params)))
     {
         pox-addr: (get pox-addr params),
         first-reward-cycle: (get first-reward-cycle params),
-        num-cycles: (get num-cycles params),
+        num-cycles: num-cycles,
         amount-ustx: (get amount-ustx params),
-        i: (if (< i (get num-cycles params))
+        i: (if (< i num-cycles)
             (let ((total-ustx (get-total-ustx-stacked reward-cycle)))
               ;; record how many uSTX this pox-addr will stack for in the given reward cycle
               (append-reward-cycle-pox-addr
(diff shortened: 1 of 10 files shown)


src/vm/functions/special.rs
```diff
@@ -48,38 +48,25 @@ fn parse_pox_stacking_result(
...
-                    stacker, sender,
-                    "BUG: tx-sender is not contract-call origin!"
-                );
-
                 // if this fails, then there's a bug in the contract (since it already does
                 // the necessary checks)
                 match StacksChainState::pox_lock(
                     &mut global_context.database,
-                    &sender,
+                    &stacker,
```
jcnelson, line 69: Just to help my own understanding -- this is the `stacker` now, not the `sender`, because the `stacker` will now reflect either the principal that owns the STX (i.e. `tx-sender`), or the specific delegated principal that the STX-owner set.  Is my understanding correct?

src/chainstate/stacks/boot/pox.clar
```diff
@@ -153,8 +169,33 @@
...
+(define-private (check-caller-allowed)
+    (or (is-eq tx-sender contract-caller)
+        (let ((caller-allowed 
+                 ;; if not in the caller map, return false
+                 (unwrap! (map-get? allowance-contract-callers
+                                    { sender: tx-sender, contract-caller: contract-caller })
+                          false)))
+          ;; is the caller allowance expired?
+          (if (< burn-block-height (unwrap! (get until-burn-ht caller-allowed) true))
+              (begin (map-delete allowance-contract-callers
```
jcnelson, line 183: Why delete?

src/chainstate/stacks/boot/pox.clar
```diff
@@ -153,8 +169,33 @@
             )
         ;; no state at all
         none
-    )
-)
+    ))
+
+(define-private (check-caller-allowed)
+    (or (is-eq tx-sender contract-caller)
```
jcnelson, line 175: This first branch `(is-eq tx-sender contract-caller)` is only ever true if this method is being invoked by a contract-call transaction, right?

src/chainstate/stacks/boot/pox.clar
```diff
@@ -153,8 +169,33 @@
...
+              true))))
+
+(define-private (get-check-delegation (stacker principal))
+    (let ((delegation-info (try! (map-get? delegation-state { stacker: stacker }))))
+      ;; did the existing delegation expire?
+      (if (match (get until-burn-ht delegation-info)
+                 until-burn-ht (> burn-block-height until-burn-ht)
+                 false)
+          ;; it expired, delete the entry and return none
+          (begin (map-delete delegation-state { stacker: stacker })
```
jcnelson, line 195: Same question here -- why should we ever delete?  Also, it seems wrong to me to be mutating state on a "getter" method.

src/chainstate/stacks/boot/pox.clar
```diff
@@ -257,36 +279,63 @@
...
+    (fold add-pox-partial-stacked-to-ith-cycle cycle-indexes 
+          { pox-addr: pox-addr, reward-cycle: first-reward-cycle, num-cycles: num-cycles, amount-ustx: amount-ustx })
+    true))
 
 ;; What is the minimum number of uSTX to be stacked in the given reward cycle?
 ;; Used internally by the Stacks node, and visible publicly.
 (define-read-only (get-stacking-minimum)
-    (/ stx-liquid-supply u20000)
-)
+    (/ stx-liquid-supply STACKING_THRESHOLD_25))
```
jcnelson, line 338: So, the threshold won't be dynamic in this PR?  Can you open an issue to make it dynamic?

src/chainstate/stacks/boot/pox.clar
```diff
@@ -380,6 +461,140 @@
...
+        { stacker: tx-sender }
+        { amount-ustx: amount-ustx,
+          delegated-to: delegator,
+          until-burn-ht: until-burn-ht,
+          pox-addr: pox-addr })
+
+      (ok true)))
+
+;; Commit partially stacked STX.
+;;   this allows a stacker/delegator to lock fewer STX than the minimal threshold in multiple transactions,
```
jcnelson, line 507: Can you add punctuation and capitalization to this?  The code comments here will be set in stone forever once the system launches, so they should be "documentation-quality" ;)

src/chainstate/stacks/boot/pox.clar
```diff
@@ -380,6 +461,140 @@
...
+      (add-pox-addr-to-ith-reward-cycle
+       u0
+       { pox-addr: pox-addr,
+         first-reward-cycle: reward-cycle,
+         num-cycles: u1,
+         amount-ustx: amount-ustx,
+         i: u0 })
+      ;; don't update the stacking-state map,
+      ;;  because it _already has_ this stacker's state
+      ;; don't lock the STX, because the STX is already locked
```
jcnelson, line 532: Can you quickly unpack how these two invariants are preserved in this public method?  I.e. how do we know that the stacking-state map already has the stacker's state, and how do we know the STX are already locked?

src/vm/functions/special.rs
kantai: Yep

src/chainstate/stacks/boot/pox.clar
kantai: Because the entries are expired, so the entries can be deleted. Ultimately, it wouldn't make much difference if we let expired entries remain here.

src/chainstate/stacks/boot/pox.clar
kantai: Yep -- or is invoked by a contract in an `(as-contract ...)` closure.

src/chainstate/stacks/boot/pox.clar
kantai: Same as above -- these are expired entries, and it keeps the contract from having to continuously re-check expired entries. Ultimately, there's not really a performance benefit to removing these, because the MARF lookup is so much more expensive than parsing the returned tuple and checking the block height. If you'd prefer, we can just leave expired entries in the map.

src/chainstate/stacks/boot/pox.clar
kantai: That's correct, opened #1947

src/chainstate/stacks/boot/pox.clar
kantai: This is preserved because the `partial-stacked-by-cycle` map is only ever updated along side a lock + stacking state.

src/chainstate/stacks/boot/pox.clar
kantai: Added in c833b65ec

src/chainstate/stacks/boot/pox.clar
jcnelson: Unless we need to check for the absence of entries in `allowance-contract-callers`, I think we should avoid deleting.  It would make the contract cheaper to execute without affecting correctness.

src/chainstate/stacks/boot/pox.clar
jcnelson: I think if we can avoid doing work, we should -- it makes the contract cheaper without sacrificing correctness.

Also, avoiding mutations in both this and the other getter will make it possible to invoke these methods in a read-only manner.

src/chainstate/stacks/boot/pox.clar
kantai: Okay, removed in 49605a7a39b61dd6a70d439d62039cc902be4e84

src/chainstate/stacks/boot/pox.clar
kantai: Okay, removed in 49605a7

src/chainstate/stacks/boot/pox.clar
```diff
@@ -380,6 +458,141 @@
...
+      ;;
+      ;; clear the partial-stacked state
+      (map-delete partial-stacked-by-cycle { pox-addr: pox-addr, sender: tx-sender, reward-cycle: reward-cycle })
+      (ok true))))
+
+;; As a delegator, stack the given principal's STX using partial-stacked-by-cycle
+;; Once the delegator has stacked > minimum, the delegator should call stack-aggregation-commit
```
jcnelson, line 537: I don't think "delegator" is the right word here -- the principal who is yielding control of the STX lockup to the delegate would be the "delegator," whereas the principal who carries out the STX lockup on behalf of one or more other principals is merely the "delegate."  This method is called by the delegate, not the delegator ;)

jcnelson (approved):
This LGTM.  Just want to clarify the right word "delegator" vs. "delegate" first.

src/chainstate/stacks/boot/pox.clar
kantai: Yep, that works for me -- updated.

src/chainstate/stacks/boot/pox.clar
```diff
@@ -380,6 +458,141 @@
...
+               ;; if pox-addr is set, must be equal to pox-addr
+               (match (get pox-addr delegation-info)
+                      specified-pox-addr (is-eq pox-addr specified-pox-addr)
+                      true)
+               ;; delegation must not expire before lock period
+               (match (get until-burn-ht delegation-info)
+                      until-burn-ht (>= until-burn-ht
+                                        unlock-burn-height)
+                      true)))
+        (err ERR_STACKING_PERMISSION_DENIED))
```
lgalabru, line 566: 1 assert, 1 test, 1 error would probably increase readability, but also the UX for the people interacting with that contract.

src/chainstate/stacks/boot/pox.clar
```diff
@@ -380,6 +458,141 @@
...
+      ;;  because it _already has_ this stacker's state
+      ;; don't lock the STX, because the STX is already locked
+      ;;
+      ;; clear the partial-stacked state
+      (map-delete partial-stacked-by-cycle { pox-addr: pox-addr, sender: tx-sender, reward-cycle: reward-cycle })
+      (ok true))))
+
+;; As a delegate, stack the given principal's STX using partial-stacked-by-cycle
+;; Once the delegate has stacked > minimum, the delegate should call stack-aggregation-commit
+(define-public (delegate-stack-stx (stacker principal)
```
lgalabru, line 538: I know that we already had the debate a few months ago, but the fact that we're implementing this mechanism in one of our very first contracts could be a signal that we should re-evaluate our position.
Basically wondering if we should revisit the design of our tokens, and bake an approver / spender scheme in it.
I don't think it's a blocker for mainnet.

lgalabru (approved):

src/chainstate/stacks/boot/pox.clar
kantai: Yeah, good idea -- will add a ERR_STACKING_PERMISSION_EXPIRED

src/chainstate/stacks/boot/pox.clar
kantai: Yep -- I was thinking about this too. I think if we wanted some kind of built-in support for it, it wouldn't be tied to tokens, but rather to a contract.

src/chainstate/stacks/boot/pox.clar
kantai: Updated in 4745049f2, added:
```
(define-constant ERR_DELEGATION_EXPIRES_DURING_LOCK 21)
(define-constant ERR_DELEGATION_TOO_MUCH_LOCKED 22)
(define-constant ERR_DELEGATION_POX_ADDR_REQUIRED 23)
```

PR #1944, opened 2020-10-02
Discussion: Structured logging via `slog`

kantai:
This PR does an initial port of our logging macros to [slog](https://github.com/slog-rs/slog). This allows us to add key-value pairs to our logs which could be parsed immediately by tools like kibana. Eventually, we could also support scoped/contextual logging, which would allow us to filter logs from different subsystems (e.g., logs from the net module could have a different logging backend), and out-of-thread logging which could improve performance when logging lots of data. This would also allow us to plug into syslog if we wanted to do that.

The only potential downside here is that this does involve including a few crates from `slog-rs`, and there's a fair bit of code there. This could _perhaps_ be remedied by something like a feature gate, but that would require that we implement a compatible macro for the k-v pairs.

To see the key-value pairs in action, this PR changed one of the log lines in `src/chainstate/stacks/index/marf.rs`:

```rust
...
    fn inner_get_extension_height(
        &mut self,
        chain_tip: &T,
        next_chain_tip: &T,
    ) -> Result<u32, Error> {
        // current chain tip must exist if it's not the "sentinel"
        let is_parent_sentinel = chain_tip == &T::sentinel();
        if !is_parent_sentinel {
            debug!("Extending off of existing node {}", chain_tip);
        } else {
            info!("First-ever block {}", next_chain_tip; "block" => %next_chain_tip);
        }
...
```

And you can see it in JSON logs:

```
$ BLOCKSTACK_LOG_JSON=1 BITCOIND_TEST=1 cargo test --workspace --bin=stacks-node  -- --nocapture --test-threads 1 --ignored

...

{"msg":"First-ever block 55c9861be5cff984a20ce6d99d4aa65941412889bdc665094136429b84f8c2ee","level":"INFO","ts":"2020-10-02T16:51:37.157992-05:00","block":"55c9861be5cff984a20ce6d99d4aa65941412889bdc665094136429b84f8c2ee"}
{"msg":"ST2VHM28V9E5QCRD6C73215KAPSBKQGPWTEE5CMQT credited (genesis): 10000 uSTX","level":"INFO","ts":"2020-10-02T16:51:40.487250-05:00"}
{"msg":"First-ever block 55c9861be5cff984a20ce6d99d4aa65941412889bdc665094136429b84f8c2ee","level":"INFO","ts":"2020-10-02T16:51:40.494040-05:00","block":"55c9861be5cff984a20ce6d99d4aa65941412889bdc665094136429b84f8c2ee"}
```

Now, this PR would need some changes to include all the same information in the log records that we had previously (e.g., thread id), but I wanted to open this first to see if this is a direction we want to go before making any more changes.

Issue: https://github.com/blockstack/stacks-blockchain/issues/1796

### DIFF
--- a/src/util/log.rs
+++ b/src/util/log.rs
@@ -17,8 +17,8 @@
  along with Blockstack. If not, see <http://www.gnu.org/licenses/>.
 */
 
-use std::env;
 use slog::{Drain, Logger};
+use std::env;
 use std::sync::Mutex;
 
 lazy_static! {
--- a/src/util/log.rs
+++ b/src/util/log.rs
@@ -17,153 +17,76 @@
  along with Blockstack. If not, see <http://www.gnu.org/licenses/>.
 */
 
-use std::cell::RefCell;
 use std::env;
+use slog::{Drain, Logger};
+use std::sync::Mutex;
 
-// Message Priorities/Levels
-// Apache Conventions defined here: https://commons.apache.org/proper/commons-logging/guide.html#Message_PrioritiesLevels
-//
-// Severe errors that cause premature termination.
-// Expect these to be immediately visible on a status console.
-pub const LOG_FATAL: u8 = 6;
-// Other runtime errors or unexpected conditions.
-// Expect these to be immediately visible on a status console.
-pub const LOG_ERROR: u8 = 5;
-// Use of deprecated APIs, poor use of API, 'almost' errors, other runtime situations that are undesirable or unexpected, but not necessarily "wrong".
-// Expect these to be immediately visible on a status console.
-pub const LOG_WARN: u8 = 4;
-// Interesting runtime events (startup/shutdown).
-// Expect these to be immediately visible on a console, so be conservative and keep to a minimum
-pub const LOG_INFO: u8 = 3;
-// Detailed information on the flow through the system.
-// Expect these to be written to logs only.
-pub const LOG_DEBUG: u8 = 2;
-// More detailed information.
-// Expect these to be written to logs only.
-pub const LOG_TRACE: u8 = 1;
-
-// per-thread log level and log format
-thread_local!(static loglevel: RefCell<u8> = RefCell::new(LOG_INFO));
-
-pub fn set_loglevel(ll: u8) -> Result<(), String> {
-    loglevel.with(move |level| match ll {
-        LOG_TRACE..=LOG_FATAL => {
-            *level.borrow_mut() = ll;
-            Ok(())
-        }
-        _ => Err("Invalid log level".to_string()),
-    })
+lazy_static! {
+    pub static ref LOGGER: Logger = make_logger();
 }
 
-pub fn get_loglevel() -> u8 {
-    let mut res = 0;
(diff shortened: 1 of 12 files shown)


lgalabru (commented):

src/util/log.rs
```diff
@@ -17,153 +17,76 @@
...
-        if ::util::log::get_loglevel() <= ::util::log::LOG_DEBUG {
-            use std::time::SystemTime;
-            use std::thread;
-            let (ts_sec, ts_msec) = match SystemTime::now().duration_since(SystemTime::UNIX_EPOCH) {
-                Ok(n) => (n.as_secs(), n.subsec_nanos() / 1_000_000),
-                Err(_) => (0, 0)
-            };
-            eprintln!("DEBUG [{}.{:03}] [{}:{}] [{:?}] {}", ts_sec, ts_msec, file!(), line!(), thread::current().id(), format!($($arg)*));
-        }
+        slog_error!($crate::util::log::LOGGER, $($arg)*)
```
jcnelson, line 62: Instead of deleting this code (and the monitoring code that ties into it), can you preserve it, and make it so that the `slog` path only executes if `BLOCKSTACK_LOG_JSON` is set to `1`?  I'd prefer to have `slog` as a compile-time feature flag if possible.

kantai: I think it's hard to make it a compile-time flag, because the macro has a different footprint: the slog macros accept key-value pairs (that's how you make the structured entries).

jcnelson: Ah, right.

jcnelson:
I took a look at the four new crates you're adding.

* I'm fine with `slog`.  It's pretty minimal and self-contained, and has no `unsafe` code.
* I'm not fine with the others.  Namely:
   * Why do we need `slog-json`?  Can't we just have the logging macros encapsulate arguments into JSON key/value pairs?  Why do we have to bring in all of `chrono` and its dependencies just to get JSON-formatted logging?
   * Why do we need `slog-term`?  It brings in `term`, `chrono`, and `atty`.  I'm annoyed by the presence of, but fundamentally okay with `atty`, since it's basically a cross-platform `isatty(2)` and nothing more.  However, `term` really shouldn't be necessary -- we're only ever going to be `eprintln!`-ing the data, not color-coding it or dealing with `terminfo`-related stuff (also, I _suspect_ that the only reason this package needs `atty` is because it wants to turn on/off color-coding text based on whether or not `stdout` or `stderr` is a terminal).
   * Why do we need `slog-async`?  It marries us to `crossbeam-channel` (which itself brings in a giant hairball of dependencies that will be used _nowhere else_ in the code) and `take_mut` (which is a crate with a single method that is `unsafe` and probably shouldn't even exist).  All `slog-async` is trying to do is send data to another thread -- why can't it use `std::sync::mpsc` channels, instead of `crossbeam-channel` channels?

kantai:
@jcnelson -- this has been updated to make slog_json a compile-time option, and `slog_async` has been removed.

jcnelson (approved):
LGTM

Issue #1947, opened 2020-10-06
PoX: Implement dynamic threshold adjustment


kantai:
Per SIP-007, PoX participation thresholds should adjust (in increments of 10k STX) for participation levels.

kantai:
This is closed by #1954, and has been deployed to Krypton.

Issue #1924, opened 2020-09-30
PoX: Stacker reward does not depend on amount stacked


psq:
I registered 2 addresses and registered them with `stack-stx`, one with 2_000_000_000_000, for 5 cycles, and one with 6_100_000_000_000, expecting the second address would get 3 times the reward as the minimum for stacking was about 2_000_000_000_000 and a bit (I forgot to account for the increasing supply with new STX minted each block on the first address, and was not expecting to get rewards, but did anyway (maybe that's the 25% case kicking in, as no one else was stacking).

Here's the data from calling the pox contract (`get-reward-set-size`, `get-reward-set-pox-address`), both addresses are mine, and no one else was stacking, so that makes it easier to track.

```
Cycle 0: 0 0-119 active
Cycle 1: 0 120-239 active
Cycle 2: 0 240-359 active
Cycle 3: 0 360-479 active
Cycle 4: 0 480-599 active
Cycle 5: 1 600-719 active
==> 5 0 ST1TWA18TSWGDAFZT377THRQQ451D1MSEM69C761 mfrQxG6v9DVhUVmSQiV9Ac7NcQiXS4BUGH 2000000000000
Cycle 6: 1 720-839 active
==> 6 0 ST1TWA18TSWGDAFZT377THRQQ451D1MSEM69C761 mfrQxG6v9DVhUVmSQiV9Ac7NcQiXS4BUGH 2000000000000
Cycle 7: 2 840-959 active
==> 7 0 ST1TWA18TSWGDAFZT377THRQQ451D1MSEM69C761 mfrQxG6v9DVhUVmSQiV9Ac7NcQiXS4BUGH 2000000000000
==> 7 1 ST30X92ZMEVF8W5FCPVX3CXFCH230496K229ZCC4X myBqL5eC3tSZ7HBUYAmi8773NXAuST3PBG 6100000000000
Cycle 8: 2 960-1079 active
==> 8 0 ST1TWA18TSWGDAFZT377THRQQ451D1MSEM69C761 mfrQxG6v9DVhUVmSQiV9Ac7NcQiXS4BUGH 2000000000000
==> 8 1 ST30X92ZMEVF8W5FCPVX3CXFCH230496K229ZCC4X myBqL5eC3tSZ7HBUYAmi8773NXAuST3PBG 6100000000000
Cycle 9: 2 1080-1199 active
[...]

PR #1954, opened 2020-10-07
PoX: Threshold Adjustment and Reward Address Repetition

kantai:
This PR implements a few different things:

1. It implements the threshold adjustment described in SIP-007 (#1947)
2. When a Stacker reward address uses _more_ than the threshold, the address is repeated in the reward set (#1924)
3. The chains coordinator, when booting the node, executes `set-burnchain-parameters` on the PoX contract.
4. To speed up the integration tests in the `stacks-node`, this parameterizes the wait time in the syncctl module.

The implementations of 1 and 2 are tested via unit tests, as well as some integration paths in the `pox` test in `neon_integrations` (which also tests 3).

### DIFF
--- a/src/chainstate/coordinator/mod.rs
+++ b/src/chainstate/coordinator/mod.rs
@@ -15,12 +15,13 @@ use chainstate::burn::{
 };
 use chainstate::stacks::{
     db::{ClarityTx, StacksChainState, StacksHeaderInfo},
+    boot::STACKS_BOOT_CODE_CONTRACT_ADDRESS,
     events::StacksTransactionReceipt,
     Error as ChainstateError, StacksAddress, StacksBlock, StacksBlockHeader, StacksBlockId,
 };
 use monitoring::increment_stx_blocks_processed_counter;
 use util::db::Error as DBError;
-use vm::{costs::ExecutionCost, types::PrincipalData};
+use vm::{costs::ExecutionCost, Value, types::{PrincipalData, QualifiedContractIdentifier}};
 
 pub mod comm;
 use chainstate::stacks::index::MarfTrieId;
@@ -217,7 +218,26 @@ impl<'a, T: BlockEventDispatcher>
             stacks_chain_id,
             chain_state_path,
             initial_balances,
-            boot_block_exec,
+            |clarity_tx| {
+                let burnchain = burnchain.clone();
+                let contract = QualifiedContractIdentifier::parse(&format!("{}.pox", STACKS_BOOT_CODE_CONTRACT_ADDRESS))
+                    .expect("Failed to construct boot code contract address");
+                let sender = PrincipalData::from(contract.clone());
+                
+                clarity_tx.connection().as_transaction(|conn| {
+                    conn.run_contract_call(
+                        &sender,
+                        &contract,
+                        "set-burnchain-parameters",
+                        &[Value::UInt(burnchain.first_block_height as u128),
+                          Value::UInt(burnchain.pox_constants.prepare_length as u128),
+                          Value::UInt(burnchain.pox_constants.reward_cycle_length as u128),
+                          Value::UInt(burnchain.pox_constants.pox_rejection_fraction as u128)],
+                        |_,_| false)
+                        .expect("Failed to set burnchain parameters in PoX contract");
+                });
+                boot_block_exec(clarity_tx)
+            },
             block_limit,
         )
         .unwrap();
--- a/src/chainstate/coordinator/mod.rs
+++ b/src/chainstate/coordinator/mod.rs
@@ -1,7 +1,5 @@
 use std::collections::VecDeque;
-use std::convert::{
-    TryInto, TryFrom
-};
+use std::convert::{TryFrom, TryInto};
 use std::time::Duration;
 
 use burnchains::{
@@ -14,14 +12,18 @@ use chainstate::burn::{
     BlockHeaderHash, BlockSnapshot, ConsensusHash,
 };
 use chainstate::stacks::{
-    db::{ClarityTx, StacksChainState, StacksHeaderInfo},
     boot::STACKS_BOOT_CODE_CONTRACT_ADDRESS,
+    db::{ClarityTx, StacksChainState, StacksHeaderInfo},
     events::StacksTransactionReceipt,
     Error as ChainstateError, StacksAddress, StacksBlock, StacksBlockHeader, StacksBlockId,
 };
 use monitoring::increment_stx_blocks_processed_counter;
 use util::db::Error as DBError;
-use vm::{costs::ExecutionCost, Value, types::{PrincipalData, QualifiedContractIdentifier}};
+use vm::{
+    costs::ExecutionCost,
+    types::{PrincipalData, QualifiedContractIdentifier},
+    Value,
+};
 
 pub mod comm;
 use chainstate::stacks::index::MarfTrieId;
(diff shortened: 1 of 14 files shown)


src/chainstate/coordinator/mod.rs
```diff
@@ -196,7 +217,32 @@ impl<'a, T: BlockEventDispatcher>
...
+            |clarity_tx| {
+                let burnchain = burnchain.clone();
+                let contract = QualifiedContractIdentifier::parse(&format!(
+                    "{}.pox",
+                    STACKS_BOOT_CODE_CONTRACT_ADDRESS
+                ))
+                .expect("Failed to construct boot code contract address");
+                let sender = PrincipalData::from(contract.clone());
+
+                clarity_tx.connection().as_transaction(|conn| {
```
jcnelson, line 229: This is going to fail when run on mainnet, because the `set-burnchain-parameters` method is wired to only run if the node is on regtest/testnet.  This is because this method has this line in it:

```
        (asserts! (and is-in-regtest (not (var-get configured))) (err ERR_NOT_ALLOWED))
```

kantai: That check was removed in this PR.

src/chainstate/stacks/boot/mod.rs
```diff
@@ -183,6 +185,72 @@ impl StacksChainState {
...
+
+        assert!(
+            participation <= liquid_ustx,
+            "CORRUPTION: More stacking participation than liquid STX"
+        );
+
+        let scale_by = cmp::max(participation, liquid_ustx / POX_MAXIMAL_SCALING as u128);
+
+        let reward_slots = pox_settings.reward_slots() as u128;
+        let threshold_precise = scale_by / reward_slots;
```
jcnelson, line 240: Suggestion: remove `scale_by` in order to remove the double integer division.  The use of `cmp::max` may cause `threshold_precise` to be calculated as `(liquid_ustx / POX_MAXIMAL_SCALING) / reward_slots`, which risks losing precision.

src/chainstate/coordinator/mod.rs
jcnelson: Nevermind -- I see that you fixed that already.

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -33,6 +33,16 @@ pub struct RunLoop {
...
 
+#[cfg(not(test))]
+const BURNCHAIN_POLL_TIME: u64 = 30; // TODO: this is testnet-specific
+#[cfg(test)]
+const BURNCHAIN_POLL_TIME: u64 = 1; // TODO: this is testnet-specific
+
+#[cfg(not(test))]
+const POX_SYNC_WAIT_MS: u64 = 1000;
+#[cfg(test)]
+const POX_SYNC_WAIT_MS: u64 = 0;
```
jcnelson, line 44: If this value is anything other than 1000, then the block download and processing time estimations won't work correctly -- the number of samples taken is supposed to be exactly equal to the number of seconds over which the samples were taken.

src/chainstate/stacks/boot/mod.rs
```diff
@@ -183,6 +185,72 @@ impl StacksChainState {
...
+        let participation = addresses
+            .iter()
+            .fold(0, |agg, (_, stacked_amt)| agg + stacked_amt);
+
+        assert!(
+            participation <= liquid_ustx,
+            "CORRUPTION: More stacking participation than liquid STX"
+        );
+
+        let scale_by = cmp::max(participation, liquid_ustx / POX_MAXIMAL_SCALING as u128);
```
jcnelson, line 237: It took me a couple reads to understand that `scale_by` is what's ensuring that the minimum participation threshold is 25% of the total liquid supply.  Maybe add a comment here, and in the definition of `POX_MAXIMAL_SCALING`?

src/chainstate/stacks/boot/mod.rs
```diff
@@ -293,6 +360,107 @@ pub mod test {
...
+        // with zero participation, should scale to 25% of liquid
+        assert_eq!(
+            StacksChainState::get_reward_threshold(&PoxConstants::new(1000, 1, 1, 1), &[], liquid),
+            50_000 * MICROSTACKS_PER_STACKS as u128
+        );
+        // should be the same at 25% participation
+        assert_eq!(
+            StacksChainState::get_reward_threshold(
+                &PoxConstants::new(1000, 1, 1, 1),
+                &[(rand_addr(), liquid / 4)],
```
jcnelson, line 418: should this be `liquid / POX_MAXIMAL_SCALING`?

src/chainstate/stacks/boot/mod.rs
```diff
@@ -739,7 +907,12 @@ pub mod test {
         block_id: &StacksBlockId,
     ) -> Result<Vec<(StacksAddress, u128)>, Error> {
         let burn_block_height = get_par_burn_block_height(state, block_id);
-        state.get_reward_addresses(burnchain, sortdb, burn_block_height, block_id)
+        state
+            .get_reward_addresses(burnchain, sortdb, burn_block_height, block_id)
+            .and_then(|mut addrs| {
+                addrs.sort_by_key(|k| k.0.bytes.0);
```
jcnelson, line 913: This means that addresses with more leading 0's are prioritized for PoX payouts.  Is that what we want?  Do we instead want to sort and then permute using the reward cycle block's VRF seed or something, to avoid creating an incentive for address grinding?

kantai: This only sets the indexes used for the subsequent random choosing (implemented by the sortition db: each “next” address is chosen using the VRF seed). It _is_ important for consensus that this ordering is the same for all nodes, but it doesn’t impact the order of payouts in a predictable way.

jcnelson: Ah, right, I'd forgotten that this was handled elsewhere.  Closing...

testnet/stacks-node/src/run_loop/neon.rs
kantai: Left at 1000ms, the integration tests take about 2x as long to run -- is there a better way to disable the stalling behavior during tests?

src/chainstate/stacks/boot/mod.rs
kantai: I wanted this test to be hardcoded, so that if the POX_MAXIMAL_SCALING is adjusted, this test would need to be adjusted as well.

src/chainstate/stacks/boot/mod.rs
kantai: Added in 6772aab41

src/chainstate/stacks/boot/pox.clar
```diff
@@ -34,7 +34,7 @@
 ;; This function can only be called once, when it boots up
 (define-public (set-burnchain-parameters (first-burn-height uint) (prepare-cycle-length uint) (reward-cycle-length uint) (rejection-fraction uint))
     (begin
-        (asserts! (and is-in-regtest (not (var-get configured))) (err ERR_NOT_ALLOWED))
+        (asserts! (not (var-get configured)) (err ERR_NOT_ALLOWED))
```
lgalabru, line 37: If we're getting rid of `is-in-regtest`, should we kill it completely?

kantai: I think we should keep it for the time being -- it may be useful elsewhere.

lgalabru (approved):

testnet/stacks-node/src/run_loop/neon.rs
jcnelson: You can set a smaller value for `ConnectionOptions::timeout`.  This is what gets used to determine how many samples the PoX sync watchdog takes.  The units are in seconds -- you can set this to something like 3 instead of 30 for a local test.

jcnelson (approved):
This LGTM. Thanks @kantai!

testnet/stacks-node/src/run_loop/neon.rs
kantai: Made this configurable via 33ec31b39

PR #1956, opened 2020-10-07
Fix/1955

### LINKED ISSUE #1955: P2P: accept BlocksData from both inbound and outbound neighbors if the block is not already present
Right now, the code path in `handle_unsolicited_BlocksData()` only considers blocks sent from outbound neighbors.  Instead, it needs to consider blocks from both inbound and outbound neighbors.  Moreover, it should drop messages for blocks it already has.

jcnelson:
This fixes #1955, and hopefully the strange miner behavior we're seeing on the Krypton testnet.  It makes it so a node can accept `BlocksData` messages from inbound authenticated neighbors, as well as outbound.


kantai (approved):
This looks good to me, but it looks like it needs a cargo fmt run.

PR #1957, opened 2020-10-08
Auto-generation of PoX contract docs

kantai:
This PR adds `vm::docs::contracts`, which supports (somewhat) auto-generating JSON documentation for boot code contracts. Currently, this is just documentation for the `pox.clar` contract.

It does this by:

* generating a signature, input, and output types from the type checker for public and read-only functions
* evaluating any error code constants to generate error code documentation
* requiring a "description" for each published public and read-only function defined

This also adds the new JSON documentation to the PR robot (as well as changing the PR robot's git email address so that it can sign the CLA).

### DIFF
--- a/src/vm/docs/contracts.rs
+++ b/src/vm/docs/contracts.rs
@@ -1,0 +1,237 @@
+use chainstate::stacks::boot::STACKS_BOOT_CODE_MAINNET;
+use vm::analysis::{mem_type_check, ContractAnalysis};
+use vm::docs::{get_input_type_string, get_output_type_string, get_signature};
+use vm::types::{FunctionType, Value};
+
+use vm::execute;
+
+use std::collections::{BTreeMap, HashMap, HashSet};
+use std::iter::FromIterator;
+
+#[derive(Serialize)]
+struct ContractRef {
+    public_functions: Vec<FunctionRef>,
+    read_only_functions: Vec<FunctionRef>,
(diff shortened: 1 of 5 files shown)


kantai:
Built docs: https://gist.github.com/kantai/17fe06eacdf1864c7df490bb17c771da

lgalabru (approved):
Looks good to me!

src/vm/docs/contracts.rs
```diff
@@ -0,0 +1,238 @@
...
+    value_type: String,
+    value: String,
+}
+
+struct ContractSupportDocs {
+    descriptions: HashMap<&'static str, &'static str>,
+    skip_func_display: HashSet<&'static str>,
+}
+
+fn make_contract_support_docs() -> HashMap<&'static str, ContractSupportDocs> {
```
jcnelson, line 40: This is fine for now, but do we want to extract the comments directly from the PoX contract?  Maybe there could be a docstring annotation -- e.g. `;;;` indicates a docstring in Clarity?

jcnelson (approved):
LGTM!

src/vm/docs/contracts.rs
kantai: Yeah, that would be a nice feature, but would require some reworking of the parser (it currently doesn't parse comments at all). Will open an issue for that.

Issue #1897, opened 2020-09-22
PoX Proposal: Use PoX lockups to enforce recency in anchor selection


kantai:
In the current anchor block design, there's no recency limit for a PoX anchor block selection. However, Stacker's messages can be used to enforce this recency by requiring some threshold of active participant locks. To do this, we would need to add a burn height parameter to the `stacks-stx` function, such that Stackers signal the following information:

* The amount they are locking up
* The number of reward cycles they want to participate for (capped at 3 months)
* The Bitcoin block height of the first reward cycle they wish to lock up for (this is the new piece of information)

Then, when choosing a PoX anchor block, the network requires that some threshold % of liquid STX issued those messages (I think a low threshold is the right thing here, maybe 5%). This constraint, combined with the Bitcoin block height information forces PoX anchor blocks to be relatively recent, because otherwise, there would be no valid lock-ups: an old Bitcoin block height + 3 reward cycles would have expired. This, in turn, biases the miners towards forward progress because ensuring that anchor blocks are chosen makes mining slightly easier (because of the incentives of PoX vs. PoB).

jcnelson:
I guess I'm not quite following what you mean by "recency limit".  Are you saying that, because miners could spend their prepare phase mining a non-canonical fork branched off of any arbitrarily-far-back block during reward cycle _N - 1_, they can require PoX to occur on a minority fork in reward cycle _N_?  If so, I can see why we'd want to discourage miners from doing this (and am open to modifying Stacking to get Stackers to help encourage it), but I'm curious why you don't think that miners are sufficiently incentivized by the fact that they can only cash out their STX if their blocks are confirmed by the canonical Stacks chain tip?

kantai:
Miners are already definitely incentivized to not do this (mine on a non-canonical fork), but this allows us to use Stacking to provide some further incentivization, which also extracts some more usefulness from the process of Stacking.

jcnelson:
This is implemented in #1937, right?

kantai:
No, it isn't, I'm implementing most of these PoX changes separately so that the PRs will be a bit more bite sized.

jcnelson:
Removing the "proposal" tag.  I think the consensus is that this is going forward.

kantai:
This is closed by #1961, and has been deployed to Krypton.

PR #1961, opened 2020-10-09
PoX: Enforce anchor block recency with burn-block-height parameter

kantai:
This PR implements #1897, by adding `burn-block-ht` parameters to `stack-stx` and `delegate-stack-stx`, which must correspond to the next reward cycle (this parameter ensures that the transaction could not be replayed in a Stacks fork occurring at a much later burn block). Further, this PR requires that PoX participation cross a threshold before activation.

Closing #1897

### DIFF
--- a/src/chainstate/stacks/boot/pox.clar
+++ b/src/chainstate/stacks/boot/pox.clar
@@ -18,6 +18,7 @@
 (define-constant ERR_DELEGATION_EXPIRES_DURING_LOCK 21)
 (define-constant ERR_DELEGATION_TOO_MUCH_LOCKED 22)
 (define-constant ERR_DELEGATION_POX_ADDR_REQUIRED 23)
+(define-constant ERR_INVALID_START_BURN_HEIGHT 24)
 
 ;; PoX disabling threshold (a percent)
 (define-constant POX_REJECTION_FRACTION u25)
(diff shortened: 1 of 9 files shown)


src/chainstate/stacks/boot/pox.clar
```diff
@@ -425,13 +426,22 @@
...
+                          (start-burn-ht uint)
                           (lock-period uint))
     ;; this stacker's first reward cycle is the _next_ reward cycle
-    (let ((first-reward-cycle (+ u1 (current-pox-reward-cycle))))
+    (let ((first-reward-cycle (+ u1 (current-pox-reward-cycle)))
+          (specified-reward-cycle (+ u1 (burn-height-to-reward-cycle start-burn-ht))))
+      ;; the start-burn-ht must result in the next reward cycle, do not allow stackers
+      ;;  to "post-date" their `stack-stx` transaction
+      (asserts! (is-eq first-reward-cycle specified-reward-cycle)
+                (err ERR_INVALID_START_BURN_HEIGHT))
```
lgalabru, line 444: What's your take on moving this check to the function `can-stack-stx`?

src/chainstate/stacks/boot/pox.clar
```diff
@@ -547,10 +557,16 @@
 (define-public (delegate-stack-stx (stacker principal)
                                    (amount-ustx uint)
                                    (pox-addr { version: (buff 1), hashbytes: (buff 20) })
+                                   (start-burn-ht uint)
```
lgalabru, line 560: Note: I think that this changes in the 2 signatures (+ comments) will outdate https://github.com/blockstack/stacks-blockchain/pull/1957

src/chainstate/stacks/boot/pox.clar
kantai: I don't think there'd be much benefit to move it into `can-stack-stx` -- that method is also used to check the delegate "aggregation commit", which doesn't need to supply a starting burn height, so by moving the check, we'd eliminate one piece of duplicated code, but need to introduce some other code path for the aggregation commit to pass the check.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -759,7 +759,9 @@ fn pox_integration_test() {
     let http_origin = format!("http://{}", &conf.node.rpc_bind);
 
     let mut burnchain_config = btc_regtest_controller.get_burnchain();
-    burnchain_config.pox_constants = PoxConstants::new(10, 5, 4, 5);
+    let mut pox_constants = PoxConstants::new(10, 5, 4, 5);
+    pox_constants.pox_participation_threshold_pct = 15;
```
jcnelson, line 763: Maybe the `PoxConstants` constructor needs to include a percentage?

jcnelson (approved):
This LGTM.  Thanks @kantai!  Just had one minor comment.

lgalabru (approved):
LGTM!

PR #1962, opened 2020-10-13
Fix/1960

### LINKED ISSUE #1960: P2P: consider a message's relayer list before relaying
Right now, the P2P relay logic does _not_ look at a message's relayer vector to prune the sets of inbound and outbound next-hop neighbors that should receive the message being forwarded along.  This is a problem because if a node sees itself in the relayer vector, it will close the connection to the peer that sent it.  This, in turn, will cause a node to eventually lose all of its outbound neighbors, leading to a stall.

jcnelson:
This PR fixes #1960.  It does so in the following ways:

* When sending a block or microblock, relay logic carries out one neighbor sampling _per block_, based on which peers have the block already, and then combine the samples into the set of peers to relay the message to.  This can sometimes cause the node to send the message to someone who has already seen it, leading to a peer disconnect.  This patch fixes this by verifying that the message to be relayed is not sent to anyone in the relayer set _at the time of relay_ instead of at the time of sampling.

* When relaying a message, the relay logic will now coalesce the set of recipients such that a _host_ will receive a message at most once.  Before, it was possible that a message could be sent twice -- once on the remote peer's inbound connection, and once on its outbound connection.

* When signing a message for relay, the signing logic will prevent the signer from signing if the message already has the maximum number of relayers.  Before, the logic would remove the oldest relayer, and add the current signer as the newest relayer.  This change makes it so the relayers vector acts as a TTL on the message, in addition to providing a route trace.

Don't be intimidated by the code length -- most of it is just tests.


src/net/relay.rs
```diff
@@ -421,7 +423,11 @@ impl RelayerStats {
 
                 w += rankings_vec[i].1;
                 if w >= target {
-                    ret.push(rankings_vec[i].0.clone());
+                    if !ret_set.contains(&rankings_vec[i].0) {
+                        ret.push(rankings_vec[i].0.clone());
```
kantai, line 427: Rather than maintaining `ret` separately, can this function just invoke `ret_set.into_iter().collect()` to convert the hashset into a vec for returning?

src/net/relay.rs
```diff
@@ -2189,6 +2210,33 @@ mod test {
...
+        block: StacksBlock,
+    ) -> bool {
+        test_debug!(
+            "{:?}: Broadcast block {}/{}",
+            peer.to_neighbor().addr,
+            &consensus_hash,
+            block.block_hash(),
+        );
+
+        let sn = SortitionDB::get_block_snapshot_consensus(
```
kantai, line 2226: Why invoke `get_block_snapshot_consensus` when this function already is given a consensus_hash?

kantai (approved):
This looks good to me -- just a couple comments.

src/net/relay.rs
jcnelson: To force the test to panic if the consensus hash doesn't exist.

src/net/relay.rs
jcnelson: Fixed

Issue #1967, opened 2020-10-14
P2P: walk timeout and walk state-transition timeouts should be enforced


jcnelson:
A neighbor walk shouldn't spend too long in a single state, which currently can happen if the remote neighbors are slow to respond.  Also, a neighbor walk shouldn't last more than a configurable amount of time before being re-instantiated with a peer sampled from the peer database.  Both of these things are related to #1963

Issue #1966, opened 2020-10-14
P2P: trying to connect to the node's bind address during a neighbor walk should cause the walk to fail


jcnelson:
Right now, a node can get caught in a neighbor walk loop where it constantly tries to handshake with its own bind address.  This does _not_ cause the walk to fail, but it should.  This, in turn, can prevent the node from re-connecting with other peers in the event of a disconnect, leading to network partitions.

This is also the source of the `Denied` errors people have been seeing.

psq:
fixes this is going to make a lot of errors disappear, and make a lot of people happy (even if they realize they will be ...)

diwakergupta:
Fixed with #1980

Issue #1963, opened 2020-10-14
P2P: neighbor walks do not resample the peer DB frequently enough, causing network partitions


jcnelson:
When performing a random walk, it is important to periodically _restart_ the random walk so that if a node's walk gets stuck in a graph cycle, it can recover.  The logic for doing so exists, but is flawed: it only restarts a walk if the current walk ever _fails_.  This leads to absurdly long peer walks.  This, in turn, can cause the network to partition:  if a node encounters a remote peer who reports no neighbors besides the requester, then the node will eventually (through churn) evict all other neighbors from its routing table.  We've seen this in Krypton multiple times -- nodes disconnect from the seed node periodically, but never re-connect with it because they get stuck talking to a long-running, publicly-routable, but stale node that is out-of-sync with the network.

Thank you to whoever is running long-running, publicly-routable, but stale nodes.  You've helped uncover a pretty nasty routing bug!

PR #1968, opened 2020-10-14
Fix/neighbor walk timeouts

jcnelson:
This PR forces the neighbor walk logic to re-sample the peer DB every so often, no matter how "error-free" the current walk is (#1963).  This is important, because without this timeout, a neighbor walk can get caught in a loop talking to a neighbor that only has the requester as its neighbor.  By forcing the node to re-instantiate a neighbor walk from a sampled node in the peer DB, the node can get itself out of this situation (#1967).

In addition, this PR fixes a bug whereby a node can get stuck in a loop trying to continuously re-connect to its own bind address (#1966).


psq:
very enlightening code, it looks like the changes achieve the intended behavior, but not familiar enough with the existing code to assess whether this would break any existing behavior that is still desirable (but it doesn't seem so)

kantai (approved):
LGTM!

PR #1972, opened 2020-10-14
update failed to check connection message

### LINKED ISSUE #1971: More noisy neighbors

Following up on #1912, neighbors.rs is still flooding the screen with messages, this time while trying to connect to itself. A few others in Discord are reporting the same problem.

From @jcnelson in Discord:
> that's fine -- it just means that the node tried to connect to itself (and got the net::Error::Denied error)
> maybe it shouldn't be info-level
> (generally speaking, the node won't try to connect to 0.0.0.0)

I would like to change the output so this message only shows at debug level, not info level. A PR will follow.

Same with the last issue, the goal is to create a clean and easier to read log at the info level, while still showing this information if debug is enabled.

An alternative would be to check if the error happens when connecting to itself versus connecting to a different address, and only displaying the info-level message if the address is not `0.0.0.0` or `pub=None`. At this point, simple seems better?

**Additional context**

Error text:
`INFO [1602251990.602] [src/net/neighbors.rs:2003] [ThreadId(8)] local.80000000://(bind=V4(0.0.0.0:20444)) (pub=None): Failed to check connection to facade01+80000000://V4(0.0.0.0:20444): Denied`

Screenshot:
VirtualBox_bks_automine_server1_14_10_2020_05_21_36

whoabuddy:
This PR fixes #1971 by changing the message from info to debug.

Smallest. PR. Ever. 

:upside_down_face:

### DIFF
--- a/src/net/neighbors.rs
+++ b/src/net/neighbors.rs
@@ -2000,7 +2000,7 @@ impl PeerNetwork {
                                 .and_then(|handle| Ok(Some(handle)))?
                         }
                         Err(e) => {
-                            info!(
+                            debug!(
                                 "{:?}: Failed to check connection to {:?}: {:?}",
                                 &network.local_peer, &my_addr, &e
                             );


kantai:
LGTM! Merging.

PR #1974, opened 2020-10-15
[Stacks v1 → v2 Upgrade] Export liquid and vesting balances from fast-dump

zone117x:
CLI feature for the Stacks 1.0 -> 2.0 migration.

This PR adds a new command arg that takes takes a fast-dump file and it's consensus hash, then outputs a JSON file containing all account STX balances. Both vesting and liquid STX are included. 

The fast-dump file is validated against the provided consensus hash. And a sha256 of the JSON data is output for Stacks 2.0 to validate. 

```
usage: blockstack-core export_migration_json [-h] [--verify]
                                             [--working-dir WORKING_DIR]
                                             [path] [block_height]
                                             [consensus_hash] [output_path]

positional arguments:
  path                  the path to the fastdump file
  block_height          the block height to export from
  consensus_hash        the known-good consensus hash string or file path
  output_path           the migration JSON output file path
```

A follow-up PR will add BNS data to the migration JSON.

## Testing

First, mimic the threshold-triggered fast-sync dump by running this command:
```
$ python blockstack-core fast_sync_snapshot 0 ./snapshot-test-data --working-dir /tmp/.blockstack-server
Fast-sync-snapshot at block 653759 with consensus hash 17e4af7e088cbd043625843b642e2b2d
Copying backups...
Copying namespace keychains...
Compressing chainstate...
Fast-sync-snapshot completed at block 653759 with consensus hash 17e4af7e088cbd043625843b642e2b2d
```

Note: the fast sync dump code was optimized, it now takes around 10 minutes rather than 40+ minutes:
```
real	9m30.709s
user	2m33.243s
sys 4m18.080s
```

Next, run the new `export_migration_json` command:
```
$ python blockstack-core export_migration_json ./snapshot-test-data 653759 17e4af7e088cbd043625843b642e2b2d /tmp/migration-data.json --working_dir /tmp/blockstack-migration-test
Saving configuration to /tmp/blockstack-migration-test/blockstack-server.ini
Importing fast-sync dump ./snapshot-test-data into /tmp/blockstack-migration-test
Restored to /tmp/blockstack-migration-test
Running setup recovery...
Querying account addresses...
Querying account balances...
Querying account vesting addresses...
Writing migration data to /tmp/migration-data.json
Migration data sha256 hash: f66e8c26d187bfddb0773dae495a18f125529883c1b23774845b6560eabc4386
Migration data sha256 hash wrote to file /tmp/migration-data.sha256
Migration data export complete
```

Note: this command, when ran without `--verify`, takes around 10 minutes:
```
real	11m0.472s
user	1m26.869s
sys	8m48.100s
```


blockstack/blockstackd.py
```diff
@@ -3173,6 +3173,23 @@ def run_blockstackd():
         'version',
         help='Print version and exit')
 
+    # -------------------------------------
+    parser = subparsers.add_parser(
+        'export_migration_json',
+        help='validate a fastdump file and export as JSON for Stacks 2.0 migration')
```
jcnelson, line 3179: "fastsync file"

blockstack/blockstackd.py
```diff
@@ -3482,6 +3499,82 @@ def run_blockstackd():
...
+        if not rc:
+           print 'fast_sync failed'
+           sys.exit(1)
+
+        # treat this as a recovery
+        setup_recovery(working_dir)
+
+        # allow consensus_hash to be provided as a string or within a file
+        consensus_hash = args.consensus_hash
+        if os.path.exists(consensus_hash):
```
jcnelson, line 3514: I'm curious why the consensus hash would need to be in a file?  It's a 40-character string that could easily be passed through `argv`.

blockstack/blockstackd.py
```diff
@@ -3482,6 +3499,82 @@ def run_blockstackd():
...
+        ch = db.get_current_consensus()
+        if str(ch) != consensus_hash:
+            print "Fast-sync import does not match consensus hash!"
+            sys.exit(1)
+        block = db.get_current_block()
+
+        tmpdir = tempfile.mkdtemp('blockstack-verify-chainstate-XXXXXX')
+        db_verified = False
+        try:
+            db_verified = verify_database(consensus_hash, block, working_dir, tmpdir)
```
jcnelson, line 3528: Just so you know, on mainnet, this method could take _days_ to run.

blockstack/blockstackd.py
```diff
@@ -3482,6 +3499,82 @@ def run_blockstackd():
...
+        except:
+            pass
+        if db_verified:
+           print "Database is consistent with %s" % args.consensus_hash
+        else:
+           print "Database is NOT CONSISTENT"
+           # TODO: can't get this passing using regtest fastdumps
```
jcnelson, line 3535: Do you have a log snippet?  You should see from the log that the `verify_database()` method re-generates the same sequence of consensus hashes that got produced when the database itself was produced.  You would also see the history of consensus hashes diverge if something was wrong with the snapshot.

blockstack/lib/fast_sync.py
```diff
@@ -597,11 +597,19 @@ def logerr(s):
         logerr("No such directory {}".format(working_dir))
         return False
 
-    # go get it 
-    import_path = fast_sync_fetch(working_dir, import_url)
+    import_path = None
```
jcnelson, line 600: Please don't do this.  If you have to download the snapshot file, you should do so separately.  This method shouldn't fail because the computer is offline.

blockstack/lib/fast_sync.py
```diff
@@ -612,44 +620,44 @@ def logerr(s):
...
-
-    with open(import_path, 'r') as f:
-        info = fast_sync_inspect( f )
-        if 'error' in info:
-            logerr("Failed to inspect snapshot {}: {}".format(import_path, info['error']))
-            return False
-
-        signatures = info['signatures']
-        ptr = info['payload_size']
+    if num_required > 0:
```
jcnelson, line 623: This check shouldn't be necessary -- the code will work (and emit helpful log output) even if `num_required` is zero.

blockstack/lib/nameset/virtualchain_hooks.py
```diff
@@ -585,9 +585,10 @@ def db_save( block_height, consensus_hash, ops_hash, accepted_ops, virtualchain_
                 # check if the required number of blocks have been mined since threshold crossed
                 if (block_height - threshold_block_id == v2_upgrade_signal_import_threshold):
                     # emit parsable log entry and export datafile
-                    log.warn('[v2-upgrade] import threshold reached at block: {}'.format(block_height))
-                    db_state.set_v2_import_block_reached(block_height)
-                    db_state.perform_v2_upgrade_datafile_export(consensus_hash)
+                    export_block_id = db_state.get_current_block()
+                    log.warn('[v2-upgrade] import threshold reached at block: {}'.format(export_block_id))
+                    db_state.set_v2_import_block_reached(export_block_id)
+                    db_state.perform_v2_upgrade_datafile_export(export_block_id)
```
jcnelson, line 591: Do you know how long this method will run in production?  Because if it takes more than 10 minutes (a likely possibility, given how many rows there are to process), then all nodes with this patch can fall behind the Bitcoin chain tip.

blockstack/lib/nameset/namedb.py
```diff
@@ -955,12 +955,12 @@ def get_account_history(self, address, offset=None, count=None):
...
 
 
-    def get_all_account_addresses(self):
+    def get_all_account_addresses(self, force_allow=False):
         """
         TESTING ONLY
         Get all account addresses
         """
-        assert BLOCKSTACK_TEST, 'BUG: this method can only be accessed in test mode'
+        assert BLOCKSTACK_TEST or force_allow, 'BUG: this method can only be accessed in test mode'
```
jcnelson, line 963: The reason this method won't run in production is because it can trigger OOMs and stall the node.  But, I see that you need it for the migration CLI option.  Instead of adding `force_allow`, can you add something more descriptive here, like `from_cli` or something, so the reader knows that this is permitted from the CLI but not from, say, the RPC codepath?

jcnelson:
> I've ran into issues trying to get integration tests for this feature working and could use some pointers. Currently, I run the initial integration test to create a fast-dump, then test manually with that file. This seems to also cause issues with getting the consensus hash validated -- looks like the problem is a discrepancy between bitcoin's regtest and mainnet block heights, not sure.

Can you show a dump of the integration test run, and a dump of what happens when you try to have the integration test runner validate the snapshot?  The integration test runner sets a few different environment variables that change the behavior of the node in a few ways in order to make some tests go faster (see `blockstack/lib/config.py`).  This may be what's tripping it up.

blockstack/blockstackd.py
zone117x: If a file is provided then it's read, if not the 40-char string is used. The previous fast dump step exports the consensus hash in a file. Should this be restricted to only using only a string?

blockstack/lib/fast_sync.py
zone117x: This change allows the method to _work_ when offline and passed a file. Previously it required a url and always attempted download.

blockstack/lib/nameset/virtualchain_hooks.py
zone117x: This discussion should probably be in https://github.com/blockstack/stacks-blockchain/pull/1910 where the logic was implemented?

blockstack/blockstackd.py
zone117x: Suggestions on alternatives? It seems like this is the only way to validate

jcnelson: It's the only sure-fire way, unfortunately :/  I'm bringing it up so as to set expectations when we get this running -- we might not be able to do a full verification, depending on how much time we'll have between when the node creates the final v1 export and when v2 needs to boot up (or, we'll need to structure the upgrade logic so that we have enough time).  We'll probably want to test this with Xenon first, for example.

blockstack/lib/fast_sync.py
jcnelson: Oh, my bad.  Then this is definitely an improvement :+1:

jcnelson:
To test on a production dataset:
* grab a fast-sync snapshot 
* decompress on a filesystem with lots of free inodes (shoot for at least 100,000).

blockstack/lib/nameset/namedb.py
zone117x: Fixed 👍

blockstack/lib/nameset/virtualchain_hooks.py
zone117x: Note: running this on my local machine against mainnet originally took 40+ minutes. @wileyj said it takes a similar amount of time on our prod infrastructure. 

I've added a few optimizations that brings the time down to less than 10 minutes on my machine.

blockstack/blockstackd.py
zone117x: Fixed 👍

blockstack/blockstackd.py
zone117x: We solved this by making the verification only run when passed a `--verify` argument. This allows miners to generate their own exports in 10-20 minutes rather than waiting days for this to complete.

blockstack/blockstackd.py
zone117x: We will now be testing this on mainnet data on a secondary test node in PBC infra.

blockstack/blockstackd.py
jcnelson: Cool!

blockstack/lib/nameset/virtualchain_hooks.py
jcnelson: Good good

jcnelson (approved):
LGTM!  Thanks @zone117x

kantai (approved):
LGTM!

PR #1975, opened 2020-10-15
PoX: Two reward addresses per block

kantai:
This PR changes `OUTPUTS_PER_COMMIT` to 2.

This requires that all commitments, regardless of whether or not they were burn commitments, have `OUTPUTS_PER_COMMIT` outputs (i.e., if intending to burn, then the tx would require 2 burn outputs).

### DIFF
--- a/sip/sip-007-stacking-consensus.md
+++ b/sip/sip-007-stacking-consensus.md
@@ -218,7 +218,7 @@ addresses for a reward cycle, then each miner commitment would have
 ## Adjusting Reward Threshold Based on Participation
 
 Each reward cycle may transfer miner funds to up to 4000 Bitcoin
-addresses (2 addresses in a 1000 burn block cycle). To ensure that
+addresses (2 addresses in a 2000 burn block cycle). To ensure that
 this number of addresses is sufficient to cover the pool of
 participants (given 100% participation of liquid STX), the threshold
 for participation must be 0.025% (1/4000th) of the liquid supply of
--- a/sip/sip-007-stacking-consensus.md
+++ b/sip/sip-007-stacking-consensus.md
@@ -191,10 +191,10 @@ Address validity is determined according to two different rules:
    descendant of the anchor block*, all of the miner's commitment
    funds must be burnt.
 2. If a miner is building off a descendant of the anchor block, the
-   miner must send commitment funds to 5 addresses from the reward
+   miner must send commitment funds to 2 addresses from the reward
    set, chosen as follows:
     * Use the verifiable random function (also used by sortition) to
-      choose 5 addresses from the reward set. These 5 addresses are
+      choose 2 addresses from the reward set. These 2 addresses are
       the reward addresses for this block.
     * Once addresses have been chosen for a block, these addresses are
       removed from the reward set, so that future blocks in the reward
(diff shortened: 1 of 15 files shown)


kantai:
Update: We've decided to go with this approach:

```
If we instead required that all commitments, regardless of whether or not they were burn commitments, had OUTPUTS_PER_COMMIT outputs (i.e., if intending to burn, then the tx would require 2 burn outputs), then we wouldn't need to use these variant forms.
```

kantai:
Updated to eliminate the need for two block commit variants.

sip/sip-007-stacking-consensus.md
```diff
@@ -191,10 +191,10 @@ Address validity is determined according to two different rules:
    descendant of the anchor block*, all of the miner's commitment
    funds must be burnt.
 2. If a miner is building off a descendant of the anchor block, the
-   miner must send commitment funds to 5 addresses from the reward
+   miner must send commitment funds to 2 addresses from the reward
```
jcnelson, line 194: Thank you for updating SIP-007!

src/chainstate/burn/db/sortdb.rs
```diff
@@ -986,20 +986,26 @@ impl<'a> SortitionHandleTx<'a> {
                     return Ok(None);
                 }
 
-                let chosen_recipient = reward_set_vrf_seed.choose(
+                if OUTPUTS_PER_COMMIT != 2 {
```
jcnelson, line 989: Is this necessary?  `OUTPUTS_PER_COMMIT` is a static constant, right?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -3230,9 +3240,18 @@ impl<'a> SortitionHandleTx<'a> {
...
-                        let (addr, ix) = recipient_info.unwrap().recipient.clone();
-                        assert_eq!(&reward_set.remove(ix as usize), &addr,
-                            "BUG: Attempted to remove used address from reward set, but failed to do so safely");
+                        let mut recipients_to_remove: Vec<_> = recipient_info
+                            .unwrap()
+                            .recipients
+                            .iter()
+                            .map(|(addr, ix)| (addr.clone(), *ix))
+                            .collect();
+                        recipients_to_remove.sort_unstable_by(|(_, a), (_, b)| b.cmp(a));
```
jcnelson, line 3249: Sorry -- I think you've answered this question before, but I've forgotten the answer -- why is this an unstable sort?  Is this the point in the code that removes addresses from the reward set, and thus must be deterministic?

src/chainstate/burn/mod.rs
```diff
@@ -179,12 +180,22 @@ impl SortitionHash {
         SortitionHash(ret)
     }
 
-    /// Choose 1 index from the range [0, max).
-    pub fn choose(&self, max: u32) -> u32 {
+    /// Choose two indices (without replacement) from the range [0, max).
+    pub fn choose_two(&self, max: u32) -> Vec<u32> {
```
jcnelson, line 184: Not required for this PR, but just in case the parameter choice somehow changes _again_, do you think you could make this method more general (e.g. `choose_n(&self, max: u32, n: u32) -> Vec<u32>`)?

src/chainstate/coordinator/tests.rs
```diff
@@ -960,6 +969,229 @@ fn test_sortition_with_reward_set() {
     }
 }
 
+#[test]
+fn test_sortition_with_burner_reward_set() {
```
jcnelson, line 973: Can this test be updated so there's an odd number of non-burn addresses in the recipient set?  Then, we can see that it works for the "last" recipient (whose block-commit will include a burn address).

src/chainstate/burn/db/sortdb.rs
kantai: It's not strictly necessary, but because the implementation of the address choosing depends on `OUTPUTS_PER_COMMIT` to be equal to 2, I want this code to panic if we ever change that constant.

src/chainstate/burn/db/sortdb.rs
kantai: It uses an unstable sort because it can: the indexes it is sorting are unique anyways, so it doesn't matter if the sort preserves the original ordering, and unstable sorts have faster implementations. Though it hardly matters here, the input size is really just 2: this step is sorting the indexes that are getting removed, so that it can remove them without invalidating the next index.

src/chainstate/burn/mod.rs
kantai: It's possible, but I'd prefer we didn't until we wanted to change the parameter choice. The reason for that is that the way we implement choosing `n` from `max` is consensus bearing, and a generalized approach to do this increases the complexity here when we probably don't want it to (we'd need to do a general swap-and-shrink on the possible indexes).

src/chainstate/coordinator/tests.rs
kantai: Sure -- updated in 00ec45dc0

src/chainstate/burn/db/sortdb.rs
jcnelson: Thanks for refreshing my memory!

lgalabru (approved):

jcnelson (approved):
LGTM!

PR #1982, opened 2020-10-20
Add Dockerfile that can run net-test scripts

kantai:
This adds a Dockerfile that can run the net-test scripts, as well as adding more 500 logging to the faucet, and a fallbackfee option to bitcoind (this was necessary to run for me: bitcoind was consistently failing to estimate fees otherwise).


jcnelson (approved):
LGTM!

Issue #1950, opened 2020-10-06
Include btc block & tx hashes in sidecar event


zone117x:
The burnchain block hash and tx hash should be included in the event data sent to the sidecar. 

The sidecar will expose these to the explorer, which will link them to an external Bitcoin explorer that others can use for inspection and verification purposes. 

Reference: https://github.com/blockstack/explorer/issues/225

diwakergupta:
@kantai @lgalabru this should be a pretty small change, right? Some explorer work is blocked on this. Could one of you, or maybe @reedrosenbluth can take this on?

kantai:
Sure, I can add this pretty quickly.

diwakergupta:
Awesome, thanks @kantai ! @aulneau FYI

PR #1983, opened 2020-10-20
#1950 - add burn block info to events

kantai:
This adds:

`burn_block_hash`
`burn_block_height`
`miner_txid`

To the block events sent out via the event_dispatcher.


jcnelson (approved):
LGTM!

PR #1985, opened 2020-10-21
Fix 1984

jcnelson:
`Value::expect_buff(x)` should work or buffers with fewer than `x` bytes.


psq (approved):
will no longer panic on a bad hash param

src/vm/types/mod.rs
```diff
@@ -726,47 +726,56 @@ impl Value {
...
         }
     }
 
     pub fn expect_buff(self, sz: usize) -> Vec<u8> {
-        if let Value::Sequence(SequenceData::Buffer(buffdata)) = self {
-            if buffdata.data.len() == sz {
+        if let Value::Sequence(SequenceData::Buffer(mut buffdata)) = self {
+            if buffdata.data.len() <= sz {
+                for _ in buffdata.data.len()..sz {
+                    buffdata.data.push(0u8)
```
kantai, line 747: Since these `expect_` functions are basically destructuring routines, it makes more sense for data padding to happen in the caller rather than in this function (a destructuring routine shouldn't mutate the returned data).

jcnelson: Fair enough.  Will fix.

kantai (approved):
LGTM! Before performing a chain reset, can you also merge `next` to `master`?

PR #1987, opened 2020-10-21
PoX Sunsetting

kantai:
This PR implements the sunset phase of PoX, per the description in SIP-007.

In particular:

* [Before Sunset](https://en.wikipedia.org/wiki/Before_Sunset) - everything here behaves the same as before -- there are 2 commit outputs required for every block commit, selected using the PoX anchor block logic.
* After sunset begin, Before sunset end - an additional output is required in each block commit. This output is the "sunset burn", and it must be greater than the required sunset ratio (described in SIP-007: for each reward cycle during the sunset phase, the sunset burn ratio must be > the progress percentage through the sunset phase).
* After sunset end - the first output of the block commit must be a burn output, and it is the only output counted in the block commit

This is tested with parsing tests in leader_block_commit, some integration tests in coordinator::tests (that test a sunset end occurring before the end of a reward cycle), and an extension of the pox test in neon_integrations.

### DIFF
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -17,6 +17,7 @@
 use deps;
 use deps::bitcoin::util::hash::Sha256dHash as BitcoinSha256dHash;
 
+use std::convert::TryFrom;
 use std::fs;
 use std::path::PathBuf;
 use std::sync::mpsc::sync_channel;
@@ -56,8 +57,7 @@ use burnchains::bitcoin::{BitcoinInputType, BitcoinTxInput, BitcoinTxOutput};
 use chainstate::burn::db::sortdb::{PoxId, SortitionDB, SortitionHandleConn, SortitionHandleTx};
 use chainstate::burn::distribution::BurnSamplePoint;
 use chainstate::burn::operations::{
-    BlockstackOperationType, LeaderBlockCommitOp, LeaderKeyRegisterOp,
-    UserBurnSupportOp,
+    BlockstackOperationType, LeaderBlockCommitOp, LeaderKeyRegisterOp, UserBurnSupportOp,
 };
 use chainstate::burn::{BlockSnapshot, Opcodes};
 
@@ -365,6 +365,35 @@ impl Burnchain {
         })
     }
 
+    /// the expected sunset burn is:
+    ///   total_commit * (progress through sunset phase) / (sunset phase duration)
+    pub fn expected_sunset_burn(&self, burn_height: u64, total_commit: u64) -> u64 {
+        if burn_height < self.pox_constants.sunset_start
+            || burn_height >= self.pox_constants.sunset_end
+        {
+            return 0;
+        }
+
+        let reward_cycle_height = self.reward_cycle_to_block_height(
+            self.block_height_to_reward_cycle(burn_height)
+                .expect("BUG: Sunset start is less than first_block_height"),
+        );
+
+        if reward_cycle_height <= self.pox_constants.sunset_start {
+            return 0;
+        }
+
+        let sunset_duration =
+            (self.pox_constants.sunset_end - self.pox_constants.sunset_start) as u128;
+        let sunset_progress = (reward_cycle_height - self.pox_constants.sunset_start) as u128;
+
+        // use u128 to avoid any possibilities of overflowing in the calculation here.
+        let expected_u128 = (total_commit as u128) * (sunset_progress) / sunset_duration;
+        u64::try_from(expected_u128)
+            // should never be possible, because sunset_burn is <= total_commit, which is a u64
+            .expect("Overflowed u64 in calculating expected sunset_burn")
+    }
+
     pub fn is_reward_cycle_start(&self, block_height: u64) -> bool {
         if block_height <= (self.first_block_height + 1) {
             // not a reward cycle start if we're the first block after genesis.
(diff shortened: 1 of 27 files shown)


src/burnchains/burnchain.rs
```diff
@@ -365,6 +365,35 @@ impl Burnchain {
         })
     }
 
+    /// the expected sunset burn is:
+    ///   total_commit * (progress through sunset phase) / (sunset phase duration)
+    pub fn expected_sunset_burn(&self, burn_height: u64, total_commit: u64) -> u64 {
+        if burn_height < self.pox_constants.sunset_start
```
jcnelson, line 371: Is `.sunset_start` in units of reward cycles or block heights?  Because, I see it get used for both here.

src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -234,41 +252,63 @@ impl LeaderBlockCommitOp {
...
-            // all pox outputs must have the same fee
-            if let Some(pox_fee) = pox_fee {
-                if output.amount != pox_fee {
-                    warn!("Invalid commit tx: different output amounts for different PoX reward addresses");
-                    return Err(op_error::ParseError);
+            let BurnchainRecipient { address, amount } = outputs.remove(0);
+            (vec![address], 0, amount)
+        } else {
+            // check if this transaction provided a sunset burn
+            let sunset_burn = if let Some(sunset_burn_output) = outputs.get(OUTPUTS_PER_COMMIT) {
```
jcnelson, line 264: I take it we couldn't burn BTC by sending it to the `OP_RETURN`?

kantai: Oh, shoot, I completely forgot about that 😞  Can update.

kantai: Updated in b9af529

src/burnchains/burnchain.rs
```diff
@@ -365,6 +365,35 @@ impl Burnchain {
...
+            || burn_height >= self.pox_constants.sunset_end
+        {
+            return 0;
```
jcnelson, line 374: Shouldn't this be `total_commit` if `burn_height >= self.pox_constants.sunset_end`?

kantai: No -- the sunset_burn isn't ever used to determine the total commitment amount -- just the `burn_fee` field.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -3052,14 +3064,15 @@ impl<'a> SortitionHandleTx<'a> {
...
+            &block_commit.burn_fee.to_string(),
             &tx_input_str,
             sort_id,
             &serde_json::to_value(&block_commit.commit_outs).unwrap(),
+            &block_commit.sunset_burn.to_string(),
         ];
 
-        self.execute("INSERT INTO block_commits (txid, vtxindex, block_height, burn_header_hash, block_header_hash, new_seed, parent_block_ptr, parent_vtxindex, key_block_ptr, key_vtxindex, memo, burn_fee, input, sortition_id, commit_outs) \
-                      VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10, ?11, ?12, ?13, ?14, ?15)", args)?;
+        self.execute("INSERT INTO block_commits (txid, vtxindex, block_height, burn_header_hash, block_header_hash, new_seed, parent_block_ptr, parent_vtxindex, key_block_ptr, key_vtxindex, memo, burn_fee, input, sortition_id, commit_outs, sunset_burn) \
```
jcnelson, line 3074: Not saying it's wrong, but is there a particular reason why `sunset_burn` is its own column, instead of just being added to the `burn_fee`?

src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -477,6 +503,46 @@ impl LeaderBlockCommitOp {
                 return Err(op_error::BlockCommitBadOutputs);
             }
         };
+        Ok(())
+    }
+
+    fn check_after_pox_sunset(&self) -> Result<(), op_error> {
+        if self.commit_outs.len() != 1 {
```
jcnelson, line 510: `commit_outs` only contains either PoX transfers or burns, right?  i.e. this check works even if there are change outputs (or even multiple change outputs)?

jcnelson: Ah, I see you test this below.  Nevermind :)

src/chainstate/burn/db/sortdb.rs
kantai: Yes -- it's because we apply the sunset ratio checks to a LeaderBlockCommitOp after construction (through the check method), so that struct needs to separate the burn_fee from the sunset_burn. While this check has already been performed by the point that it is being inserted here, these tables serialize the whole struct, so it seemed appropriate to keep that field in the table.

jcnelson (approved):
LGTM!

src/burnchains/burnchain.rs
```diff
@@ -365,6 +365,35 @@ impl Burnchain {
         })
     }
 
+    /// the expected sunset burn is:
+    ///   total_commit * (progress through sunset phase) / (sunset phase duration)
+    pub fn expected_sunset_burn(&self, burn_height: u64, total_commit: u64) -> u64 {
```
lgalabru, line 370: `burn_height` is a stacks block height, not a burnchain block height, correct?

kantai: It's a burnchain block height -- PoX sunset timelines and reward cycles (and locks/unlocks) are governed by the burnchain block height, not the stacks height.

lgalabru (approved):

PR #1988, opened 2020-10-21
Clarity defined Cost Functions

reedrosenbluth:
## Changes
- New boot contract containing cost functions for every native Clarity function.
- Refactored `CostTracker` to include `compute_cost` function, which can compute costs via supplied clarity contracts.
- Refactored `LimitedCostTracker` impl to keep track of a mapping relating cost function names to a `QualifiedId`, and a cache of cost contracts. 
- Replace `runtime_cost!` macro with `runtime_cost` function which uses updated `CostTracker`

## Considerations
- Currently the boot cost contract is parsed and loaded into `LimitedCostTrackers` every time they are constructed with `new...()`, unless using `new_free()`. This is unnecessary. The constructors should be refactored to require the relevant contract data be passed in.

Review by commit


jcnelson:
Ready for a first glance (and possibly a PR review!)

src/vm/ast/definition_sorter/mod.rs
```diff
@@ -53,6 +54,14 @@ impl<'a> DefinitionSorter {
         Ok(())
     }
 
+    pub fn run_pass_free(
```
kantai, line 57: Can `run_pass_free` be implemented by passing a "free" LimitedCostTracker to `pass.run()` instead of creating the new `run_free()` function?

kantai (approved):
This looks good to me! The net tests look like they're failing because of the Clarity slowdowns, but I'm fairly convinced that most of that can be addressed later (most of the slowdown is coming from areas where we can probably skip cost calculations anyways).

src/vm/costs/mod.rs
```diff
@@ -18,28 +18,38 @@ pub mod constants;
...
+pub fn runtime_cost<T: TryInto<u64>, C: CostTracker>(
+    cost_function: ClarityCostFunction,
+    tracker: &mut C,
+    input: T) -> Result<()> {
+
+    // let start = Instant::now();
+    let size: u64 = input.try_into().map_err(|_| CostErrors::CostOverflow)?;
+    let cost = tracker.compute_cost(cost_function, size)?;
+    // let duration = start.elapsed();
+    // dbg!(cost_function, duration);
```
jcnelson, line 50: Linting -- please remove the commented-out code if it's not going to be used.

src/vm/costs/mod.rs
```diff
@@ -93,37 +112,95 @@ impl CostTracker for () {
...
 #[derive(Debug, PartialEq, Eq)]
 pub enum CostErrors {
+    CostComputationFailed(String),
     CostOverflow,
     CostBalanceExceeded(ExecutionCost, ExecutionCost),
     MemoryBalanceExceeded(u64, u64),
 }
 
 impl LimitedCostTracker {
+    //TODO: require cost functions/contracts to be included
```
jcnelson, line 149: If possible, can you address this before merging?  We want to make invalid states unrepresentable.

src/vm/costs/mod.rs
```diff
@@ -134,6 +211,73 @@ impl LimitedCostTracker {
...
+        Ok(Some(_)) => Err(CostErrors::CostComputationFailed("Clarity cost function returned something other than a Cost tuple".to_string())),
+        Ok(None) => Err(CostErrors::CostComputationFailed("Clarity cost function returned nothing".to_string())),
+        Err(e) => Err(CostErrors::CostComputationFailed(format!("Error evaluating result of cost function {}: {}", cost_function.get_name(), e))),
+    }
+}
+
+fn compute_cost(
+    cost_tracker: &mut LimitedCostTracker,
+    cost_function: ClarityCostFunction,
+    input_size: u64) -> Result<ExecutionCost> {
```
jcnelson, line 251: I think this method does a lot of unnecessary duplicate computation -- it re-parses and re-evals the same cost functions over and over again.  I think we talked about making it so the node parses the cost functions once per block, and caches them over the course of evaluating the block so it doesn't have to do this?  You could put this cache into a `ClarityBlockConnection`, for example, and have it be instantiated as part of its constructor.

src/chainstate/stacks/boot/costs.clar
```diff
@@ -0,0 +1,495 @@
...
+        write_count: u0,
+        read_count: u0,
+        read_length: u0,
+    })
+
+;; Linear cost-assessment function
+(define-private (linear (n uint) (a uint) (b uint))
+    (+ (* a n) b))
+
+;; TODO: fix this once log is available
```
jcnelson, line 19: Is there an issue open somewhere for implementing integer log-2 as a Clarity function?

Actually, you could implement integer log-2 _in_ Clarity if you wanted -- all you'd need to do is find out the index of the highest bit set in a given uint (i.e. find the value for _i_ for a given _n_ where _(1 << (i + 1)) <= _n_ && _n_ <= (1 << i_)).

src/chainstate/stacks/boot/costs.clar
```diff
@@ -0,0 +1,495 @@
...
+(define-read-only (cost_not (n uint))
+    (runtime u1))
+
+(define-read-only (cost_eq (n uint))
+    (runtime (linear n u1 u1)))
+
+(define-read-only (cost_begin (n uint))
+    (runtime u1))
+
+(define-read-only (cost_hash160 (n uint))
```
jcnelson, line 227: Should the hash functions be linear in _n_, since hashing _n_ bytes is _Omega(n)_?

src/chainstate/stacks/boot/costs.clar
```diff
@@ -0,0 +1,495 @@
...
+(define-read-only (cost_sha512 (n uint))
+    (runtime u1))
+
+(define-read-only (cost_sha512t256 (n uint))
+    (runtime u1))
+
+(define-read-only (cost_keccak256 (n uint))
+    (runtime u1))
+
+(define-read-only (cost_secp256k1recover (n uint))
```
jcnelson, line 242: Secp256k1 verification and recovery is _O(1)_, but it's probably a "bigger" _O(1)_ than hashing.

src/chainstate/stacks/boot/costs.clar
```diff
@@ -0,0 +1,495 @@
...
+(define-read-only (cost_append (n uint))
+    (runtime (linear n u1 u1)))
+
+(define-read-only (cost_concat (n uint))
+    (runtime (linear n u1 u1)))
+
+(define-read-only (cost_as_max_len (n uint))
+    (runtime u1))
+
+(define-read-only (cost_contract_call (n uint))
```
jcnelson, line 308: A `contract-call?` also has a storage read cost.

src/chainstate/stacks/boot/costs.clar
```diff
@@ -0,0 +1,495 @@
...
+(define-read-only (cost_let (n uint))
+    (runtime (linear n u1 u1)))
+
+(define-read-only (cost_if (n uint))
+    (runtime u1))
+
+(define-read-only (cost_asserts (n uint))
+    (runtime u1))
+
+(define-read-only (cost_map (n uint))
```
jcnelson, line 152: What's `_n_` here -- is it the length of the list being mapped?  If so, then wouldn't this be linear?  Same question for `filter` and `fold` below.

src/vm/costs/cost_functions.rs
```diff
@@ -1,319 +1,109 @@
...
+    Let("cost_let"),
+    If("cost_if"),
+    Asserts("cost_asserts"),
+    Map("cost_map"),
+    Filter("cost_filter"),
+    Len("cost_len"),
+    Fold("cost_fold"),
+    ListCons("cost_list_cons"),
+    TypeParseStep("cost_type_parse_step"),
+    DataHashCost("cost_data_hash_cost"),
```
jcnelson, line 40: Where's `DataHashCost` used?

src/chainstate/stacks/boot/costs.clar
```diff
@@ -0,0 +1,495 @@
+;; the .costs contract
+
+;; Helper Functions
```
jcnelson, line 3: Overall, I think for this contract it would be helpful to add a comment to each cost function to briefly explain what the instance characteristic variables represent (i.e. what's _n_ represent?).

src/vm/ast/mod.rs
```diff
@@ -68,6 +69,21 @@ pub fn build_ast<T: CostTracker>(
     Ok(contract_ast)
 }
 
+pub fn build_ast_free(
```
jcnelson, line 72: Where is this function used?

src/vm/costs/mod.rs
```diff
@@ -93,37 +112,95 @@ impl CostTracker for () {
...
+            limit: ExecutionCost::max_value(),
+            total: ExecutionCost::zero(),
+            memory: 0,
+            memory_limit: CLARITY_MEMORY_LIMIT,
+            free: false,
+        };
+        cost_tracker.load_boot_costs();
+        cost_tracker
+    }
+    pub fn new_free() -> LimitedCostTracker {
```
jcnelson, line 178: I've been thinking more about the implications of having a "free" cost tracker.  Since the only case where we would use a "free" cost-tracker is to parse and evaluate a cost function (risky, since it's a DDoS vector), might it make more sense to have a protocol-defined constant maximum cost for processing a cost function?

kantai: We do free tracking in other places as well, like determining the PoX reward set, for example.

src/vm/costs/mod.rs
kantai: This method uses pre-parsed ASTs to avoid the overhead of re-parsing, and the LimitedCostTracker has the same lifetime as a block. That combined with the fact that the cost tracker is also what is invoking this function, I think it makes sense to keep the cache in the tracker itself (as is currently done). I do think there's still a lot of opportunity for speedups here, but I would note that this PR is opened against `next-cost` with the intention being that there are going to be a lot of subsequent PRs both improving the efficiency of the basic implementation here, as well as implementing the remaining items to support cost voting.

src/vm/costs/mod.rs
jcnelson: Some context.  We already know that the cost tracker's cost functions are going to come from a set of voted-upon public methods in various contracts, which will be tracked on a per-chaintip basis.  This means that the Clarity VM will already have all the knowledge it needs to load up and instantiate the cost functions when it begins processing the next block -- it will already need to look them up from the current chain tip, for example.  It's entirely reasonable to me to simply have this method take the set of cost functions as an argument.

src/vm/costs/mod.rs
```diff
@@ -134,6 +211,73 @@ impl LimitedCostTracker {
...
+        Err(e) => Err(CostErrors::CostComputationFailed(format!("Error evaluating result of cost function {}: {}", cost_function.get_name(), e))),
+    }
+}
+
+fn compute_cost(
+    cost_tracker: &mut LimitedCostTracker,
+    cost_function: ClarityCostFunction,
+    input_size: u64) -> Result<ExecutionCost> {
+
+    let mut null_store = NullBackingStore::new();
```
jcnelson, line 253: The code doesn't seem to be parsing through the cost function code to verify that no cost function does any I/O.  This is necessary because the use of `NullBackingStore` will cause the node to panic if it uses a cost function that attempts any Clarify function that does I/O under the hood.

Can you update the code that parses the cost functions into an AST to return an error if the AST contains any nodes that do I/O (or do fold, map, or filter)?

src/vm/costs/mod.rs
kantai: Note also that the importance of performing free calculations isn't that there should be a maximum cost limit for processing a cost function, but instead the fact that otherwise we have a recursive problem: if we use a cost-tracking Clarity VM to compute costs, do we then use another cost-tracking Clarity VM to track cost of that VM? By making the actual computation of costs free, we escape that recursive problem. Of course, we want to be sure that these cost functions aren't going to be DDOS vectors, but we'd want to do that by limiting the space of those functions themselves to arithmetic (which is the plan).

src/vm/costs/mod.rs
```diff
@@ -134,6 +211,73 @@ impl LimitedCostTracker {
...
+    let conn = null_store.as_clarity_db();
+    let mut global_context = GlobalContext::new(conn, LimitedCostTracker::new_free());
+
+    let cost_function_reference = cost_tracker.cost_function_references
```
jcnelson, line 257: Is this actually a reference?  Looks like you're making an owned copy of the function here.

src/vm/costs/mod.rs
kantai: I think that that is true, but I don't think it should be implemented in this PR -- implementation of the chain state passing in cost functions is planned for a separate PR.

src/vm/costs/mod.rs
jcnelson, line 251: The use of a single scalar `input_size` is probably fine for now, but we'll want it to be possible to have cost functions that have multiple instance characteristics.  You don't have to address it in this PR, but please open an issue for it if you don't.

src/vm/database/marf.rs
```diff
@@ -594,3 +594,59 @@ impl ClarityBackingStore for MemoryBackingStore {
...
+    pub fn as_clarity_db<'a>(&'a mut self) -> ClarityDatabase<'a> {
+        ClarityDatabase::new(self, &NULL_HEADER_DB, &NULL_BURN_STATE_DB)
+    }
+
+    pub fn as_analysis_db<'a>(&'a mut self) -> AnalysisDatabase<'a> {
+        AnalysisDatabase::new(self)
+    }
+}
+
+impl ClarityBackingStore for NullBackingStore {
```
jcnelson, line 612: Should all of these methods just panic?  I can't think of an instance where a cost function will do anything that accesses the backing store.

src/vm/costs/mod.rs
jcnelson: Right, understood about the need to bootstrap cost-tracking without recursion.  That's why I was wondering if we should consider having a tighter bound on "free" cost-tracking than the maximum possible value.

> but we'd want to do that by limiting the space of those functions themselves to arithmetic (which is the plan).

I asked about that (indirectly) below, regarding the use of `NullBackingStore`.  We'll want to limit ourselves to arithmetic for cost-tracking.

There's not a similar concern about PoX, right?  The PoX functions can't be removed or changed through a miner vote, so there's no need to worry about constraining the set of operations the PoX functions can do?

kantai: > There's not a similar concern about PoX, right? The PoX functions can't be removed or changed through a miner vote, so there's no need to worry about constraining the set of operations the PoX functions can do?

Yep, that's right -- no need to constrain the set of operations for those functions, but we would want to execute them without executing the cost-tracking functions for efficiency.

src/chainstate/stacks/boot/costs.clar
njordhov: Feel free to take inspiration from the `log2` Clarity implementation in ClarityTools by running the _Generate Clarity_ menu command on https://clarity.tools/code?KGxvZzIgOCk=

src/vm/costs/cost_functions.rs
reedrosenbluth: It doesn't appear to be used anywhere (I copied this over from the previous list of cost functions), however it is mentioned in SIP-006: https://github.com/blockstack/stacks-blockchain/blob/473ae8ac7c9760928ab273ccdb45e7344c1241a5/sip/sip-006-runtime-cost-assessment.md#data-lookup-costs

Is this supposed to be being used somewhere?

src/vm/ast/mod.rs
reedrosenbluth: Good catch, meant to get rid of this!

src/vm/costs/mod.rs
reedrosenbluth: This isn't actually a reference (in the Rust sense). Perhaps this is just poor naming, but I was using the word reference to mean contract ID + function name. Will reconsider this naming...

src/vm/costs/mod.rs
reedrosenbluth: https://github.com/blockstack/stacks-blockchain/issues/2047

src/vm/costs/mod.rs
jcnelson: Well, before renaming, is there a way you can reduce the number of `.clone()` calls here?  It might be fine to literally take the reference.

kantai: I don't think it's likely to be able to convert these to non-owned references. It may be possible in this PR, but looking ahead, these references to contracts are going to be controlled by the voting contract, meaning that state will come from the Stacks chain state. So if a reference was held to that data, Stacks chain state processing could not occur as long as that reference is held (because no mutable references would be usable).

kantai: Also -- regarding the naming, these are clarity function references: they refer to clarity functions, they are not themselves the data of clarity functions.

lgalabru (approved):
Looks good to me, thank you @reedrosenbluth!

src/chainstate/stacks/boot/costs.clar
reedrosenbluth: https://github.com/blockstack/stacks-blockchain/issues/1862

src/vm/analysis/type_checker/natives/sequences.rs
```diff
@@ -312,7 +313,7 @@ pub fn check_special_len(
     check_argument_count(1, args)?;
 
     let collection_type = checker.type_check(&args[0], context)?;
-    runtime_cost!(cost_functions::ANALYSIS_ITERABLE_FUNC, checker, 1)?;
+    runtime_cost(ClarityCostFunction::AnalysisIterableFunc, checker, 0)?;
```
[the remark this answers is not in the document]
reedrosenbluth: This cost function doesn't take any input, so it doesn't matter what this number is. Not sure why it used to be 1, I think it was a typo. We are using 0 to represent a lack of input. The reason we are using this "ignored" parameter instead of an Optional is that it makes it easier to vote in new cost functions that will take args.

src/vm/costs/mod.rs
kantai: This is captured in #2041

src/vm/costs/cost_functions.rs
jcnelson: I'm not sure.  Let's ask @kantai -- should `DataHashCost` exist, or should we remove it from the code and from SIP-006?

src/vm/costs/mod.rs
jcnelson: So, we're not worried about the free cost tracker being maximal?  I see that the rest of my concern in this issue is recorded in #2041.

src/vm/costs/cost_functions.rs
kantai: Ah, yes, this can be removed -- it was originally going to be used for variable name lookups, but variable names are bounded, so this didn't need to be used.

src/vm/costs/mod.rs
kantai: Nope-- not worried about the free cost tracker being maximal. Anywhere the free cost tracker is invoked should be getting locked down through other means.

src/chainstate/stacks/boot/costs.clar
kantai: Yes, that's the size of the list. However, the "linear" cost of the map is charged during the execution of the function. This cost corresponds only to the constant overhead imposed by invoking the function through `map` (e.g., the function lookup cost).

src/chainstate/stacks/boot/costs.clar
kantai: Yes, I think that's true -- I started an issue to make sure that these get addressed before merging `next-costs` into `next`: https://github.com/blockstack/stacks-blockchain/issues/2058

src/chainstate/stacks/boot/costs.clar
kantai: Yes, I think that is true, but we would be updating these constants through benchmarking anyways, so I don't think it needs to be addressed here.

src/chainstate/stacks/boot/costs.clar
kantai: Yep -- added this to the issue: https://github.com/blockstack/stacks-blockchain/issues/2058

kantai:
@jcnelson -- I think the outstanding comments are covered by other PRs and open issues. Once @reedrosenbluth updates NullBackingStore to panic in all methods, and gets `cargo fmt --all` to pass, is it okay to merge this PR to `next-cost`?

jcnelson:
@kantai In addition to that, can you or he also address #2058?  It doesn't seem to be a big enough change to warrant a separate PR -- isn't it just changing the definition of the cost function (i.e. less than 10 lines)?

src/chainstate/stacks/boot/costs.clar
reedrosenbluth: https://github.com/blockstack/stacks-blockchain/issues/2061

PR #1990, opened 2020-10-22
Fix: set default burn cap for multiple commits

kantai:
With the number of PoX outputs increasing, the default burn cap has to increase to clear the dust UTXO limit.


diwakergupta (approved):
LGTM, thanks!

Issue #1981, opened 2020-10-17
contract-call? can not be used in define-read-only


psq:
The following code
```
(define-read-only (symbol)
  (contract-call? 'ST3J2GVMMM2R07ZFBJDWTYEYAR8FZH5WKDTFJ9AHA.swapr symbol 'ST3J2GVMMM2R07ZFBJDWTYEYAR8FZH5WKDTFJ9AHA.plaid-token 'ST3J2GVMMM2R07ZFBJDWTYEYAR8FZH5WKDTFJ9AHA.stx-token)
)
```
generates an error `expecting read-only statements, detected a writing operation.` when deploying, even if `contract-call?` is calling an other `define-read-only` function.

This should be allowed.

Fix may be related to what is needed for https://github.com/blockstack/stacks-blockchain/issues/1641, and the 2 issues prevent using the RPC read-only call to call any function using `contract-call`.

psq:
This is actually a bit more subtle and specific to traits.  Traits definition do not specify whether a function is read-only or not, so read-only detection can only be done either at run time, or call time, not deploy time.  If calling a read-only function from an other contract, this works fine.

More specifically, here's a full example that fails with the same error:
```
(define-trait fun-trait ( (fun1 () (response uint uint)) ))
```

```
(impl-trait .trait.fun-trait) (define-read-only (fun1) (ok u1))
```

```
(use-trait fun-trait .trait.fun-trait) (define-read-only (fun2 (fun <fun-trait>)) (contract-call? fun fun1))
```

updated the title to reflect the actual issue.

One possibility is to relax the implementation of `type_check`, and then check the contract passed at runtime uses a `read-only` function before accepting the transaction.

An other possibility would be to change the trait definition to be able to specify some functions should only be defined as `read-only`.

I tend to favor the first option.

cc @kantai @lgalabru @jcnelson

friedger:
A third option would be to introduce `contract-call-read-only?` or so, and then check at runtime. I think this makes it more obvious for developers (and users) to understand that the state does not change.

jcnelson:
[...]

Issue #1641, opened 2020-05-28
RPC: allow clients to call any Clarity function, not just read-only functions


jcnelson:
Make it so that the RPC endpoint used for calling read-only functions can be used to call any functions.  Make it so that any mutations are discarded.

jcnelson:
Supersedes #1630 and #1629

PR #1992, opened 2020-10-23
Implement #1641 rpc call public functions as read-only

psq:
In starting to look into implementing #1981, I stumbled upon the implementation of `call-read` and thought maybe there was a way to call the same function in a read only fashion if it was made `public` instead of `read-only`.

And turns out, that the implementation was quite simple by changing a parameter and checking whether the code made any writes, and fail if it did.

#1981 still forthcoming.

Not consensus altering, so can be merged to master directly.

### DIFF
--- a/src/net/inv.rs
+++ b/src/net/inv.rs
@@ -1358,7 +1358,7 @@ impl PeerNetwork {
         };
 
         debug!(
-            "{:?}: Send GetPoxInv to {:?} for {} rewad cycles starting at {} ({})",
+            "{:?}: Send GetPoxInv to {:?} for {} reward cycles starting at {} ({})",
             &self.local_peer,
             nk,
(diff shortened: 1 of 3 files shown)


psq:
couldn't link to  #1641, but this will fix

jcnelson:
We will want to make sure that this works with multiple concurrent write operations.  If this creates savepoints, and unwinds savepoints, we'll need to make sure that one RPC call terminating does not (a) error out or (b) cause other in-flight RPCs to error out due to savepoint unwinding.  Consider using a HTTP load tester to simulate lots of concurrent requests.  We should also add a test to make sure that the order in which savepoints are unwound does not cause other savepoints to be clobbered.

psq:
closing for now, will reopen when ready

PR #2011, opened 2020-10-28
fix rbf calculation - fixes #2010

### LINKED ISSUE #2010: [TESTNET BUG] RBF fee calculation spends too much
While running a miner during the competition, the 0.5 btc ran out much faster than anticipated (after about 1/3 of the blocks it was supposed to last for.  Here's why:

With the burn fee set to 22_000, when attempt = 1:
total_spent: 22_000
tx_fee: 5_500

but when attempt  = 2
total_spent: 22_000
tx_fee: 726_396

The code is adding last_tx_len * min_relay_fee to the original fee

It seems we have 2 problems:
- it is multiplying the calculated fee by `attempt`, while it should be using `attempt - 1` (i.e. add the fee once if that's the second attempt, the fee for the first attempt is the original 5_500)
- the relay fee (as returned by my bitcoind node) is 0.00001000 (or 1000 sats, the code uses 1024), but that's the fee per kb, not per byte

so the current calculation is off by 2048, and instead of 726_396 should be 5500 + 352 = 5852, which seems more reasonable.

psq:
Fixes #2010.  Current calculation overpays by a factor of 2048 (should use `attempt-1` as multiplier, and fee is per kb, not per byte

### DIFF
--- a/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
+++ b/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
@@ -664,7 +664,7 @@ impl BitcoinRegtestController {
 
         // RBF
         let tx_fee = self.config.burnchain.burnchain_op_tx_fee
-            + (attempt * self.last_tx_len * self.min_relay_fee);
+            + ((attempt - 1) * self.last_tx_len * self.min_relay_fee / 1024);
 
         let public_key = signer.get_public_key();
         let mut total_consumed = 0;
@@ -691,6 +691,7 @@ impl BitcoinRegtestController {
             return None;
         }
         let value = total_consumed - total_spent - tx_fee;
+        debug!("Payments value: {:?}, total_consumed: {:?}, total_spent: {:?}, tx_fee: {:?}, attempt: {:?}", value, total_consumed, total_spent, tx_fee, attempt);
         if value >= DUST_UTXO_LIMIT {
             let change_output = BitcoinAddress::to_p2pkh_tx_out(&change_address_hash, value);
             tx.output.push(change_output);
--- a/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
+++ b/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
@@ -664,7 +664,7 @@ impl BitcoinRegtestController {
 
         // RBF
         let tx_fee = self.config.burnchain.burnchain_op_tx_fee
-            + ((attempt - 1) * self.last_tx_len * self.min_relay_fee / 1024);
+            + ((attempt.saturating_sub(1) * self.last_tx_len * self.min_relay_fee) / 1024);
 
         let public_key = signer.get_public_key();
         let mut total_consumed = 0;


testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -664,7 +664,7 @@ impl BitcoinRegtestController {
 
         // RBF
         let tx_fee = self.config.burnchain.burnchain_op_tx_fee
-            + (attempt * self.last_tx_len * self.min_relay_fee);
+            + ((attempt - 1) * self.last_tx_len * self.min_relay_fee / 1024);
```
jcnelson, line 667: For good measure, can you change `attempt - 1` to `attempt.saturating_sub(1)` so it doesn't ever underflow?

jcnelson: Also, can you group the numerator explicitly?  I.e. `((attempt.saturating_sub(1) * self.last_tx_len * self.min_relay_fee) / 1024);`

jcnelson:
Thanks for catching this @psq!  Are you able to mine blocks on the testnet with these settings?

Once my comment is addressed this will LGTM.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
psq: good idea.  addressed in 00d7880

psq:
> Are you able to mine blocks on the testnet with these settings

yes, the miner is mining, but I'm no sure it is the second transaction is the one accepted when there is a second attempt based on how the UTXO seem to get updated, so I'll dig deeper tomorrow.

psq:
seems to work fine, I changed the code to send 2 sats more per new attempt, and the replaced transaction is the one that gets committed, not the original one.  So all good.

```
bitcoin-cli -regtest -conf=/Users/psq/src/perso/stacks-blockchain/bitcoin.conf -rpcport=19443 gettransaction f39adee67e803ee2faeec0d6733c8f6b905f34881e2d8ebdca7aeac8dc90d598 true
{
  "amount": -0.00022002,
  "fee": -0.00005852,
  "confirmations": 446,
  "blockhash": "291bf835aafebd56eb8e6529f821c82b1815fc61e16a407243520df9cb9c8e36",
  "blockheight": 3015,
  "blockindex": 7,
  "blocktime": 1604047618,
  "txid": "f39adee67e803ee2faeec0d6733c8f6b905f34881e2d8ebdca7aeac8dc90d598",
  "walletconflicts": [
  ],
  "time": 1604047618,
  "timereceived": 1604100263,
  "bip125-replaceable": "no",
  "details": [
    {
      "involvesWatchonly": true,
      "category": "send",
      "amount": 0.00000000,
      "vout": 0,
      "fee": -0.00005852,
      "abandoned": false
    },
    {
      "involvesWatchonly": true,
      "address": "mfWxJ45yp2SFn7UciZyNpvDKrzbhyfKrY8",
      "category": "send",
      "amount": -0.00011001,
      "vout": 1,
      "fee": -0.00005852,
      "abandoned": false
    },
    {
      "involvesWatchonly": true,
      "address": "mfWxJ45yp2SFn7UciZyNpvDKrzbhyfKrY8",
      "category": "send",
      "amount": -0.00011001,
      "vout": 2,
      "fee": -0.00005852,
      "abandoned": false
    },
    {
      "involvesWatchonly": true,
      "address": "mxtGS9DHV2TafucTmcDJhNipRZ7svFoi7C",
      "category": "send",
      "amount": -0.40695581,
      "label": "",
      "vout": 3,
      "fee": -0.00005852,
      "abandoned": false
    },
    {
      "involvesWatchonly": true,
      "address": "mxtGS9DHV2TafucTmcDJhNipRZ7svFoi7C",
      "category": "receive",
      "amount": 0.40695581,
      "label": "",
      "vout": 3
    }
  ],
  "hex": "01000000016740ed391ace48d0025eb8f67cbb9c0cb44d08067b074c3cd245733830a784f1030000006a47304402201894cce7b5c9d30d961891751f3beb1d2f65aa8c7ccf80de5b2a8d57032c6e6f0220610e266761a28c2fd5391661f074000882f8ec28ccb2f5e17356f8b71d17f1b0012103c74baf92e45c41c5d16022165753f015b032b0bce30dc848d2a3495128800ccafdffffff040000000000000000536a4c5069645b26fae0a3d035f2f2a891f1fcdd667ac9755cd5709c305bf9e3e173d0db065bac81488024f279e4e99f36fff6c7cc62a4853222e1ce66cec8b925986aa051cd6b00000bc5001300000bc6000d00f92a0000000000001976a914000000000000000000000000000000000000000088acf92a0000000000001976a914000000000000000000000000000000000000000088ac1df76c02000000001976a914be8203fccbca453e0b0dd4edc919451c7895764088ac00000000"
}
```
you can see 11_001 per address instead of 11_000, and the first transaction is not found.

jcnelson (approved):
LGTM.  Thanks @psq!

kantai (approved):
LGTM!

Issue #2000, opened 2020-10-26
P2P: off-by-one error in BlocksAvailable message handling leading to slow download requests.


jcnelson:
For a variety of reasons, the block downloader state machine treats the 0th block as the Stacks block mined, whereas the block inventory state machine treats the 0th block as the first sortition.  This however has produced an off-by-one error when handling a `BlocksAvailable` message -- the P2P state-machine needs to convert the sortition height of the affected inventory bit into the Stacks block height for the downloader.

While this does not affect the system's correctness -- a node that receives a `BlocksAvailable` will eventually download the associated block -- it does slow it down, because the node will do a full pass on the sortition history before wrapping around to the block referred to by the `BlocksAvailable` message.  The test described in #1999 will confirm that this is sufficiently addressed.

jcnelson:
Fixed in #2014

Issue #1999, opened 2020-10-26
P2P: network integration test: verify that multiple NAT'ed miners can build on the same chain if there are no public miners


jcnelson:
There should be an integration test that spawns a master public node that does _not_ mine, and spawns a set of NAT'ed miners.  The test should verify that all nodes build on the same fork almost all the time.  This currently does not happen in practice -- NAT'ed miners can accidentally omit blocks on the chain they're working on (will be fixed by #1927).  We need to confirm that the public node's P2P state machine is capable of sending out `BlocksAvailable` messages to all NAT'ed miners in a timely manner once it receives a new block, and that the NAT'ed miners will all fetch that block and build on it in a timely manner -- i.e. all within the burnchain block interval.

jcnelson:
Fixed in #2014

Issue #1998, opened 2020-10-26
P2P: there's a bug in the way that the block downloader tracks in-flight blocks and microblock requests that leads to starvation


jcnelson:
There's a bug in the downloader state-machine that will cause the node to hang on fetching confirmed microblock streams until the node believes that the associated _block_ is still in-flight (which can be several minutes).  This should be fixed -- the code is meant to treat in-flight microblock streams separately from in-flight blocks.

jcnelson:
Fixed in #2014

Issue #1997, opened 2020-10-26
P2P: the download state-machine should be active for a minimum amount of time


jcnelson:
Right now, the block-download state machine will do a pass over the sequence of sortitions until it has `max_inflight` requests.  Once this pass is complete, the main state-machine advances through the other states.  In certain cases, this can lead to download starvation -- for example, if inventory synchronization takes longer than the inventory throttle interval, then the node can be forced to do a full inventory synchronization for each batch of `max_inflight` downloads.  Instead, the P2P state machine should keep the node in the `BlockDownloads` state for a minimum amount of time before advancing to the next work-state (if the download state-machine does a full pass over all sortitions, then the P2P state machine can transition to the next state early).

jcnelson:
Fixed in #2014

PR #2014, opened 2020-10-29
Improve block relay logic to accommodate NAT'ed miners

### LINKED ISSUE #2006: P2P: buffer blocks, microblocks, blocksavailable, and microblocksavailable messages if they are close to the target sortition
Instead of rejecting these messages outright, buffer them up for a time (e.g. 2x the burnchain block time) and try to re-process them each time the burnchain tip advances.  This ought to help nodes speed up propagation of blocks and microblocks even if not all nodes advance their view of the burnchain tip at the same time (or, if a public node advances its view more slowly than a NAT'ed miner).

jcnelson:
This PR addresses several open problems in the block-relay and mining logic that make it difficult for NAT'ed miners to build on the longest chain tip.  In particular, this PR fixes #2006, #2000, #1999, #1998, #1997, #1996, #1927, and #1680.  As a result of applying this PR, NAT'ed miners should have a much easier time detecting the latest chain tip and propagating blocks.

## What this PR does

This PR makes the following changes to the behavior of the P2P state-machine:

* It adds an `Antientropy` state in-between `DownloadBlocks` and `Prune`, in which the node scans its local block inventory in reverse reward cycle order (i.e. newest reward cycle to oldest) and examines its neighbors' inventories to see if there exists a local copy of a block or confirmed microblock stream that the remote neighbor does _not_ have.  If so, then the node queues a `BlocksData` or `MicroblocksData` message for broadcasting to the remote neighbors that do not report having them.  This state is only active if the node has no inbound peers (i.e. it's NAT'ed); if there's at least one inbound peer, then this state is skipped.

   This addresses a long-standing problem that we see with NAT'ed miners -- if a NAT'ed miner mines a block, but currently has no neighbors, then no one will receive it once it gets mined.  The NAT'ed miner does not (currently) try to rectify this, and will keep building on the block it mined as long as it believes it can out-pace the current chain tip (which is often the case).  As a result, other nodes receive a slew of blocks they cannot process -- they're missing an unrelayed ancestor block from this miner, but they receive the subsequent blocks.  As a result, the entire chain slows down processing (the NAT'ed miner wins sortition and propagates a block, but no one else can process it).  By adding this antientropy step, the NAT'ed miner will quickly discover that no one has its block and propagate it before attempting to build on it.

   This behavior is tested by the `net-test/tests/test_2_nat_miners.sh` integration test -- the test spawns a non-mining master node with two NAT'ed nodes that cannot talk directly to each other nor discover each other, and verifies the pair of nodes collectively build a high-quality chain of at least 100 blocks (i.e. over 90% of Stacks blocks are on the same fork and over 90% of burnchain blocks have a sortition).

* It adds a relay message buffer for nodes that will cause them to hold onto `BlocksAvailable`, `MicroblocksAvailable`, `BlocksData`, and `MicroblocksData` messages for at least one sortition if they correspond to sortition state that is currently unknown to the node (e.g. they contain an unrecognized consensus hash).  Once a new sortition happens, any buffered messages are re-tried to see if they now correspond to known sortition state (and if so, they're processed as if they had just arrived).

   This addresses a problem I saw last Friday with @diwakergupta's NAT'ed miner.  It forwarded the block it mined to a bunch of public nodes, but not the master node.  All of these public nodes were behind his miner by one sortition, so they all dropped the block, and his block (and all of its children) were never incorporated into the other nodes' chain states.  By buffering up block data that depends on having up-to-date sortition state, publicly-routable nodes can handle cases like this where they receive data that is imminently processable once their view of the burnchain advances by one block (this appears to be a common phenomenon in practice -- just look at public nodes' `/v2/info` results).

   This new behavior is tested by a unit test in `src/net/relay.rs`, and by an end-to-end integration test in `net-test/tests/test_2_miners_1_follower.sh`.  The integration test spawns a public non-mining follower, a NAT'ed master, and two NAT'ed miners who can only send their blocks to the follower.  The follower buffers the blocks up if they don't yet correspond to fresh burnchain state, processes them once they do, and then proceeds to forward `BlocksAvailable` messages to the master and other NAT'ed miner.  The test verifies that the two NAT'ed miners are able to produce a high-quality chain (at least 90% of Stacks blocks are on the same fork, and at least 90% of Bitcoin blocks have a sortition once mining begins).

* It fixes a couple of performance bugs.  If block (or confirmed microblock stream) _N_ is advertised as available, the code now correctly updates the _N_th bit in the neighbor's inventory.  The PR also makes it so the downloader state machine is hinted to begin fetching blocks starting at height _N_, instead of _N+1_ (as had been the case).

* It fixes a couple of PoX synchronization issues that interfered with a node's ability to boot up and mine correctly.  In particular:
   * If the node's view of the burnchain is within the same reward cycle as the burnchain tip, the node waits for a fixed timeout before trying to download new burnchain blocks.
   * If the node's view of the burnchain is in a prior reward cycle than the reward cycle containing the burnchain's tip, then the node waits for the inventory state machine to do a full pass to fetch the next reward cycle's block inventories before beginning to sample block downloads.  This ensures that sampling begins only once downloading begins (which begins only once the inventory state machine has had a chance to report new data to it).
   * The download state-machine will now perform a full pass over all sortitions before moving onto the next P2P work state (now `Antientropy`; previously `Prune`).  This helps ensure that during the initial block download, Stacks blocks are aggressively fetched as quickly as possible, instead of having the P2P state machine cycle through the other work states after fetching only 5 Stacks blocks (as is the case today).  It doesn't affect steady-state behavior.

* The end-to-end integration test framework has been significantly expanded (in `net-test/`).  In particular:
   * The `faucet.sh` program now reports miner power, blocks mined by miner, and mempool information.
   * The `faucet.sh` program now supports a `report` mode that lets you dump the result of an API call directly to stdout, for use in doing things like assessing chain quality
   * The `start.sh` program now serves mainly as a shell library for integration tests (mostly concerning itself with spinning up nodes and auxiliary programs, and acting as a process supervisor for them), and depends on variables set in an associated `config.sh` file.
   * The test framework's `config.sh` library allows tests to be expressed in a DSL-like manner, instead of nitty-gritty shell commands.  For example, the following snippet will spin up a miner node that won't listen to incoming handshakes (simulating being NAT'ed), and won't talk to a node at 127.0.0.1:20444:

        ```
        source ./config.sh
        PROCESS_EXIT_AT_BLOCK_HEIGHT=450
        
        miner_config "miner-1-$NAME" "21443" "21444"
            ban_peer "127.0.0.1" "22444"
            set_nat "true"
            start_node &
            MINER_1_PID=$!

        wait_pids $MINER_1_PID
        ```
   I intend to add more integration tests that make use of this feature.

* In the service of the above, the testnet config file now supports a few more directives that will let you block peers from connecting, as well as force the node to disconnect from all its neighbors periodically.  A few of the fault-injection flags in `src/net/connection.rs` are now available outside of unit-tests in order to run integration tests.

## Testing

* Issues #2006, and #1927 have unit tests added in `src/net/relay.rs`.
* Issues #2000 and #1998 are not directly tested since they do not affect correctness, but you can see the speed difference in download tests (which now run more quickly).
* Issues #1999 and #1996 are tested by two new end-to-end integration tests in `net-test/tests/`.  I haven't wired them up to a Github action yet (they both take about 45 minutes to run), but you can run them locally and observe that they verify that a set of Stacks nodes will collectively build a high-quality chain.

## Consensus

This PR does not affect consensus.  Nodes with this PR can be run on the current testnet.


kantai:
Can you add a github action that runs the new tests automatically?

jcnelson:
Added.  They take a while to run in GH -- about 2.5 hours :grimacing:.  This is considerably slower than on my laptop, which only takes about an hour to run them both.

src/net/p2p.rs
```diff
@@ -2448,7 +2895,7 @@ impl PeerNetwork {
                                 network_result,
                             )? {
                                 // advance work state
-                                self.work_state = PeerNetworkWorkState::Prune;
+                                self.work_state = PeerNetworkWorkState::Antientropy;
```
kantai, line 2898: Can Antientropy be AntiEntropy?

kantai (approved):
Other than the naming issue with `AntiEntropy` (which when written as Antientropy, I can't help but pronounce like anshunropy), this looks good to me.

src/net/p2p.rs
jcnelson: haha sure.  I was just being lazy

Issue #1892, opened 2020-09-21
PoX: on-Bitcoin Stacking operations


jcnelson:
Per SIP-007, the Stacking operation needs to be able to occur on the Bitcoin chain in over to overpower Stacks miners who may be disinclined to mine them.  In particular, we need at least two operations to be on-chain:

* Locking up STX for Stacking
* Rejecting the PoX reward cycle

These need to have fall-back implementations that work on the Bitcoin chain because their execution has monetary repercussions for Stacks miners who participate in Stacking.  Making it so that they are represented on the Bitcoin makes it impossible for correct Stacks miners to ignore them.

diwakergupta:
@kantai this is a superset of #1856 or you want to track them as independent issues?

kantai:
I think they are independent issues, #1856 is about providing a spec for all of our Bitcoin transactions.

Issue #1856, opened 2020-09-04
SIP for Bitcoin operation wire formats


kantai:
We currently do not have a SIP describing the wire formats used by the Bitcoin burnchain implementation. This SIP would need to describe the wire formats for:

1. Leader block commits
2. Leader key registers
3. User burn supports

This would also describe the eventual bitcoin formats for PoX operations like lockups and "nay" anchor votes.

jcnelson:
I think the leader key registers, user burn supports, and PoB block commits should be described in SIP-001, and the PoX block commits, on-chain PoX lockups, and on-chain PoX delegations should be described in SIP-007.

kantai:
That's ultimately fine with me. The reason I am advocating for a new SIP is that this description will necessarily be burnchain-dependent: these wire formats are specific to Bitcoin-as-the-Burnchain, whereas the other SIPs (SIP-001 in particular), try to describe things in a burnchain-agnostic way.

jcnelson:
Ah, that makes sense.

I think in general that SIPs should have a section that describes implementation-specific concerns.  If we did this, then we'd have a section in SIP-001 and SIP-007 that described how you would do this in Bitcoin, as well as how you would do this in another burnchain if it came to it (I think we will ultimately describe how to do this on Stacks as well).

Issue #1615, opened 2020-05-21
Burnchain: Allow VRF public key reuse


kantai:
Block commit operations are currently checked by matching the input address of the block commit to the output bytes of the VRF key register operation. _However_, this is not a sufficient check, because Bitcoin outputs could allow matching scriptsigs to validate for _other_ kinds of outputs (e.g., an output that _always_ validate).

There are (at least) two potential options:

1. Require block commits to spend the exact UTXO generated by the VRF key register.
2. Lookup the block commit's spent UTXO to obtain the output "scriptpubkey" and validate using that.

(1) would require miners to very specifically link UTXOs between VRF key registers and block commits.

(2) would require running a Bitcoin node with txindex=1

A radical alternative would be to revisit the necessity of one-time use VRF public keys. If my memory is correct, the rationale for one-time use public keys is that it prevents miners from assembling a large set of public keys, and using insights about the seed to chose a key. However, the VRF seed is hidden from everyone but the burnchain miner (and even, to a certain extent from them -- changing the public key they commit to will lead to a different vrf seed, because the proof-of-work hash will change).

jcnelson:
To be clear, (2) would also require that the burnchain indexer make multiple (very slow and expensive) RPC calls to bitcoind to resolve each consider transactions' input's consumed output.  Stacks 1.0 does this, and I consider it to be a design flaw because it not only slows down chain synchronization, but also makes it much more expensive for people to run their own bitcoin nodes to use Stacks.  I'd like to avoid option (2) if at all possible.

I'm not sure that was the rationale.  We can't stop a miner from assembling a large set of VRF public keys even if they're one-time-use, since they can just register a bunch of them under different addresses.  Also, the VRF seed is public knowledge for the duration of the epoch right before the miner's tenure begins -- this is necessary in order to ensure that all miners have a chance to make their VRF proofs in order to submit their block-commits.  There is a window of time during which the next epoch's seed is known only to the burnchain miner (and that may give them a chance to grind on VRF keys), but this advantage can be severely mitigated by requiring that the VRF key have a minimum number of confirmations before it can be paired with a block-commit (the code right now implicitly requires that the VRF key has at least 1 confirmation -- i.e. it must already be in the burn DB -- but this can be pushed back).

[...]

PR #2015, opened 2020-10-29
PoX: Stacking Operations via Bitcoin transaction

### LINKED ISSUE #1527: PoX: Stacking wire format for burnchain/bitcoin operations
Stacking messages need to deserializable and serializable -- this will involve a new burnchain operation, and a contract-call for a "bootcode" smart contract (for issuing stacking messages on the Stacks chain rather than the burnchain)

kantai:
This PR implements #1892 and by documenting our bitcoin wire formats in SIPs 1 and 7, closes #1527 and #1856 

This PR introduces two new Bitcoin ops, PreStackStx and StackStx. PreStackStx is required to ensure that the signer of StackStx is using a valid UTXO (as opposed to using a fake UTXO, see the discussion in #1615)

StackStxOps, once validated and included in a sortition, are applied in _every_ subsequent Stacks fork. This does not include the Stacks block chosen in that sortition, but rather all subsequent blocks (this is because the miner could not have known about the StackStxOps in that sortition before broadcasting their commitment).

When applied, StackStxOps execute the `stack-stx` function of the PoX contract.


src/chainstate/burn/db/sortdb.rs
```diff
@@ -1038,6 +1109,37 @@ impl<'a> SortitionHandleTx<'a> {
...
+    }
+
+    ///
+    /// Returns the number of blocks in this Sortition fork with Stacking-related
+    ///   bitcoin operations.
+    ///  the `next` parameter controls whether or not to return the value used for the
+    ///   block chosen in the current sortition or the value that should be used for blocks
+    ///   in the subsequent sortition
+    ///
+    pub fn get_blocks_with_stacks_txs(
```
jcnelson, line 1124: Not a bug, but more of a lint/readability concern.  Can this method be called `get_num_blocks_with_stacks_txs` instead, so as to clarify that it is returning a count instead of the actual blocks?  Also, might I suggest creating two methods to remove the need for `next` (whose usage isn't obvious) -- `get_num_blocks_with_stacks_txs` and `get_next_num_blocks_with_stacks_txs`?

jcnelson: I'm also unclear why `next` is needed?  Maybe I need to re-read this code a few times.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -1988,6 +2090,54 @@ impl<'a> SortitionDBConn<'a> {
...
+                    chain_tip,
+                    &db_keys::burn_header_hash_for_stacking_tx_block(block_number),
+                )?
+                .expect("CORRUPTION: expected to find burnchain header entry, but none found"),
+        )
+        .expect("CORRUPTION: bad BurnchainHeaderHash written to sortition db");
+
+        query_rows(
+            self.conn(),
+            "SELECT * FROM stack_stx WHERE burn_header_hash = ?",
```
jcnelson, line 2136: Won't this return stacking information in multiple PoX forks?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -3022,6 +3232,25 @@ impl<'a> SortitionHandleTx<'a> {
         Ok(())
     }
 
+    /// Insert a pre-stack-stx op
+    fn insert_stack_stx(&mut self, op: &StackStxOp, sort_id: &SortitionId) -> Result<(), db_error> {
```
jcnelson, line 3236: Should this be `insert_pre_stack_stx`, per the comment?

src/chainstate/burn/operations/stack_stx.rs
```diff
@@ -0,0 +1,495 @@
...
+    ) -> Result<PreStackStxOp, op_error> {
+        PreStackStxOp::parse_from_tx(
+            block_header.block_height,
+            &block_header.block_hash,
+            tx,
+            pox_sunset_ht,
+        )
+    }
+
+    /// parse a LeaderBlockCommitOp
```
jcnelson, line 85: comment typo?

src/chainstate/burn/operations/stack_stx.rs
```diff
@@ -0,0 +1,495 @@
...
+            warn!(
+                "StacksStxOp payload is malformed ({} bytes, expected {})",
+                data.len(),
+                17
+            );
+            return None;
+        }
+
+        let stacked_ustx = parse_u128_from_be(&data[0..16]).unwrap();
+        let num_cycles = data[16].clone();
```
jcnelson, line 185: Is the `.clone()` necessary?  Doesn't `u8` implement `Copy`?

src/chainstate/burn/operations/stack_stx.rs
```diff
@@ -0,0 +1,495 @@
...
+            block_header.block_height,
+            &block_header.block_hash,
+            tx,
+            sender,
+            pox_sunset_ht,
+        )
+    }
+
+    /// parse a LeaderBlockCommitOp
+    /// `pox_sunset_ht` is the height at which PoX *disables*
```
jcnelson, line 228: Another comment typo?

src/chainstate/burn/operations/stack_stx.rs
```diff
@@ -0,0 +1,495 @@
...
+
+        assert_eq!(&op.sender, &sender);
+        assert_eq!(
+            &op.reward_addr,
+            &StacksAddress::from_bitcoin_address(&tx.outputs[0].address)
+        );
+        assert_eq!(op.stacked_ustx, u128::from_be_bytes([1; 16]));
+        assert_eq!(op.num_cycles, 1);
+    }
+}
```
jcnelson, line 495: Can you expand these tests to cover error paths in parsing?

src/chainstate/burn/operations/stack_stx.rs
kantai: Ah yeah, this clone() can be removed.

src/chainstate/burn/db/sortdb.rs
kantai: Yeah, that's a comment typo

src/chainstate/burn/db/sortdb.rs
kantai: Next is necessary because StackStxOps are validated in the burnchain block that they are included in, but not _processed_ until subsequent blocks.

src/chainstate/burn/db/sortdb.rs
kantai: Yes -- this table should just place a uniqueness constraint on `txid`, because the operations do not depend on the PoX fork for validity.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -3662,10 +3661,152 @@ impl StacksChainState {
...
+
+        let parent_tip = StacksBlockId::new(parent_consensus_hash, parent_block_hash);
+        let mut connection =
+            clarity_db.read_only_connection(&parent_tip, &NULL_HEADER_DB, &NULL_BURN_STATE_DB);
+        let processed_blocks =
+            connection.with_clarity_db_readonly(|db| db.get_stx_btc_ops_processed());
+        connection.done();
+
+        let mut stacking_ops = vec![];
+        // process the interval (processed_block, cur_blocks]
```
jcnelson, line 3686: How big can this range be?  Is there a way to cap it -- i.e. perhaps we can say that this range is no bigger than the maximal number of reward cycles, times the length of a reward cycle?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -3662,10 +3661,152 @@ impl StacksChainState {
...
+                        };
+
+                        all_receipts.push(receipt);
+                    } else {
+                        unreachable!(
+                            "BUG: Non-response value returned by Stacking STX burnchain op"
+                        )
+                    }
+                }
+                Err(e) => {
```
jcnelson, line 3784: Are there cases where this is a non-panicking error?

kantai: Yeah -- if the block cost budget is exceeded.

src/chainstate/stacks/db/blocks.rs
kantai: Yes, I think this can be capped at (maximum number of cycles that can be locked for * cycle length)

src/chainstate/stacks/db/blocks.rs
```diff
@@ -3870,13 +4019,34 @@ impl StacksChainState {
...
+                   "microblock_parent_count" => %microblocks.len());
+
+            // process stacking operations from bitcoin ops
+            let mut receipts = match StacksChainState::process_stacking_ops(
+                &mut clarity_tx,
+                stacking_burn_ops,
+                processed_total,
+            ) {
+                Err(e) => {
+                    let msg = format!("Invalid Stacks block {}: {:?}", block.block_hash(), &e);
```
jcnelson, line 4038: How can this error path be reached?  I'm not sure we want to punish a Stacks miner (i.e. invalidate the block) if someone sent a Stacking Bitcoin transaction that wouldn't have been a valid Stacks transaction, had it been sent to the Stacks chain.  The Stacks miner can't help what transactions get included in the Bitcoin chain.

kantai: It unreachable code it turns out -- I'll patch `process_stacking_ops` to just return a receipt Vec

src/chainstate/burn/db/sortdb.rs
kantai: Updated in `74da104fb`

src/chainstate/burn/db/sortdb.rs
kantai: Updated in 74da104fb

src/chainstate/burn/operations/stack_stx.rs
kantai: Updated in 74da104fb

src/chainstate/burn/operations/stack_stx.rs
kantai: Updated in 74da104fb

src/chainstate/burn/operations/stack_stx.rs
kantai: Updated in 74da104fb

src/chainstate/stacks/db/blocks.rs
kantai: Updated in 74da104fb

src/chainstate/stacks/db/blocks.rs
kantai: Yes -- this makes deep forks more expensive, because they must apply the stacking operations that occur between the parent burn block and the current burn block. _Not_ counting that towards the block's compute budget is much more consequential for the network (instead of possibly making certain deep forks impossible, it is a huge DDOS vector). There's things we could do to limit this -- checking the account balance "for free" before invoking the contract and placing a high threshold on BTC stacking ops could place both a theoretical and practical limit on the number of operations that would need to be processed.

jcnelson: > There's things we could do to limit this -- checking the account balance "for free" before invoking the contract and placing a high threshold on BTC stacking ops could place both a theoretical and practical limit on the number of operations that would need to be processed.

Yeah -- I think we'll want to impose some rules on the BTC-based Stacking operations to bound the total number of Stacking operations that will be considered, no matter how deep the Stacks fork is.  In particular, there's two cost concerns here:

* Bounding the number of MARF reads required to load up the history of Stacking operations (this can be achieved by setting `earliest-possible-burn-block` appropriately).
* Bounding the execution time of all Stacking operations such that they, collectively, do not exceed a single Stacks block's budget.

I think there's a pretty wide array of strategies we can implement to achieve the latter.  Some ideas come to mind:

* Prioritize the top-N Stacking operations by BTC tx fee, if there are too many to fit into a Stacks block (i.e. if there are more than N of them, where N is a derived protocol constant equal to the maximal number of calls to `stack-stx` that can fit into a block).
* Prioritize the top-N Stacking operations by amount Stacked
* Prioritize the top-N Stacking operations by STX transaction fee paid
* Take the last N Stacking operations that occur in the Bitcoin chain
* Sample N Stacking operations without replacement, using the VRF seed of `burnchain-tip` as the randomness seed
* Require that all on-Bitcoin Stacking operations stack for only 1 reward cycle, assuming that a Bitcoin block will always contain less than or equal to N Stacking transactions

kantai:
Per discussion in the weekly blockchain meeting, this PR should update to only process operations in the parent bitcoin block of a new Stacks block.

src/burnchains/db.rs
```diff
@@ -271,9 +272,25 @@ impl BurnchainDB {
         })
     }
 
+    pub fn get_burnchain_op(&self, txid: &Txid) -> Option<BlockstackOperationType> {
```
jcnelson, line 275: This could be problematic if the burnchain itself forks, and the transaction gets mined on a nephew fork.  Is there code somewhere that either drops nephew forks, or only considers transactions if they're on the canonical burnchain fork?

kantai: For how this is used, it won't matter -- this is only used to lookup an input's corresponding spent output. If the spending tx is in the current fork, then the corresponding txid would be as well (otherwise, the tx would never have been accepted by Bitcoin).

src/chainstate/burn/operations/mod.rs
```diff
@@ -136,6 +145,34 @@ impl From<db_error> for Error {
...
+    /// the output address
+    pub output: StacksAddress,
+
+    // common to all transactions
+    pub txid: Txid,                            // transaction ID
+    pub vtxindex: u32,                         // index in the block where this tx occurs
+    pub block_height: u64,                     // block height at which this tx occurs
+    pub burn_header_hash: BurnchainHeaderHash, // hash of the burn chain block header
+}
+
```
jcnelson, line 175: I'm now noticing that neither of these operations contain a consensus hash, even though they have space for it.  Should they contain a consensus hash, so a client can guarantee that these operations will only apply in a node that follows the consensus rules and is on the right PoX fork?

src/chainstate/stacks/db/blocks.rs
jcnelson: Okay, I see you addressed this below by making it so that only the penultimate burnchain block's stacking operations will be applied.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -3836,6 +3912,18 @@ impl StacksChainState {
...
+            let parent_burn_hash = SortitionDB::get_block_snapshot_consensus(
+                &burn_dbconn.tx(),
+                &chain_tip_consensus_hash,
+            )?
+            .expect(
+                "BUG: Failed to load snapshot for block snapshot during Stacks block processing",
+            )
+            .parent_burn_header_hash;
+            let stacking_burn_ops =
+                SortitionDB::get_stack_stx_ops(&burn_dbconn.tx(), &parent_burn_hash)?;
```
jcnelson, line 3925: Do we have an upper bound on how many operations this can be?  Or do we still need to figure that out?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -935,6 +935,8 @@ impl BurnchainController for BitcoinRegtestController {
             BlockstackOperationType::UserBurnSupport(payload) => {
                 self.build_user_burn_support_tx(payload, op_signer, attempt)
             }
+            BlockstackOperationType::PreStackStx(_payload) => unimplemented!(),
+            BlockstackOperationType::StackStx(_payload) => unimplemented!(),
```
jcnelson, line 939: Maybe this is a separate PR, but is there any miner support planned for on-BTC stacking operations?  If not, can you open an issue so we can test out stacking on the testnet?

src/chainstate/burn/operations/mod.rs
kantai: I'm not sure what it would add to the user experience. I'd almost certainly prefer that my transaction was picked up regardless of the consensus hash (definitely in the case of the PreStackStx transaction, and probably in the StackStx transaction as well).

src/chainstate/stacks/db/blocks.rs
kantai: There's two bounds -- one is the bitcoin block size, and the other is the compute limit. The bitcoin block size will limit this far more than the compute limit.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
kantai: I don't plan on implementing it, and I don't see it as a vital element of this repo -- it seems like something that is strictly within the domain of client tooling.

src/chainstate/stacks/db/blocks.rs
jcnelson: :+1: Will still want to measure it when implementing the block-fill logic, but this is still a good outcome.

src/burnchains/db.rs
jcnelson: This is true.

src/chainstate/burn/operations/mod.rs
jcnelson: Fair enough.  The consensus hash's original purpose was to enable the network to survive hard forks -- a consensus hash represents one set of nodes.  But there are other ways to do this, such as by simply changing the magic bytes in each Bitcoin transaction.

jcnelson (approved):
LGTM!  Thanks @kantai!

kantai:
@lgalabru -- does this PR look ready to merge to you?

lgalabru:
@kantai apologies for the latency, I had a first glance at this PR last week, wanted to think more about it, and forgot to get back to it. The PR looks great, but I think but I have 2 philosophical issues that you probably have an opinion on.
1) I think that the concept of using the Bitcoin chain as a transport layer for interacting with Stacks smart contracts is very powerful (vision, feature-set, etc). Our current thinking is to introduce Stacks as `Smart contract for Bitcoin`, and with a generalization of your approach (I understand that we have some constraints, but even a limited, generalized approach) would bring us closer of that statement. 
2) If my understanding is correct (and I could be missing something!), if a user is using the "Bitcoinized flow", then the user is paying a fee in satoshis for the transport on the Bitcoin chain, however he's not paying for the operation on the Stacks chain, and that feels a bit unfair to me.
What do you think?

lgalabru (approved):

kantai:
> I think that the concept of using the Bitcoin chain as a transport layer for interacting with Stacks smart contracts is very powerful (vision, feature-set, etc). Our current thinking is to introduce Stacks as Smart contract for Bitcoin, and with a generalization of your approach (I understand that we have some constraints, but even a limited, generalized approach) would bring us closer to that statement. As in, with a very well crafted contract, Bitcoin users would be able to use smart contracts, from the bitcoin chain! I think this is huge.

I agree that this would be generally useful. However, your next question informs what the difficulty here would be. Because these transactions are broadcasted on Bitcoin _and must_ be included in the next block, the normal economic pricing scheme for transactions wouldn't work: for general transactions, we rely on miners to figure out the fee-maximizing block packing, which informs what a tx price should be. We could possibly come up with a pricing scheme for general transactions broadcasted on Bitcoin, but I think any such scheme would have some downsides. That's partially why this needs to be special-cased: it ensures that there's a limit to the "unfairness" to miners.

Issue #1930, opened 2020-10-01
Emitting stacking events for unlocks and rewards


agraebe:
I'm working on a sample integration for Stacking and I would like to be able to obtain 3 types of stacking events:
- [x] lockups - already implemented in the node as `STXLockEvent` but I believe not yet exposed the Stacks Blockchain API
- ~~unlocks~~
- [x] rewards

The rewards paid out to the BTC address could be obtained through the bitcoin chain, but it would require me to know who transferred coins when

cc @lgalabru and @zone117x

jcnelson:
Rewards can be done in a straightforward manner -- we'd just emit an event for the particular BTC address of the winning block-commit.

Unlocks are less straightforward.  In particular, token unlocks don't happen eagerly; they happen only once you _spend_ your tokens at some indeterminate time _after_ your lock-up period expires.  But, you probably don't need an event for unlocks (unless you specifically need an event for when a spend happens to remove the lock state from the DB, which I'm guessing is _not_ what you're interested in, since this happens lazily?), because the relevant information for determining when locked tokens become spendable again is captured in the `STXLockEvent` struct already.

agraebe:
Right, unlocking is less important and I can  get that in other ways. Rewards are the important ones.

agraebe:
fyi: @hstove and @kyranjamie

kantai:
Getting reward events should be doable, but it would require a bit of effort for the node to provide the information (because the total burns of a block aren't all sent to the PoX addresses -- some will be burnt)

zone117x:
@kantai any ETA on this feature? We have this pre-codefreeze issue blocked on it https://github.com/blockstack/stacks-blockchain-api/issues/268

kantai:
I can slate this after wrapping up the Bitcoin Stacking operations. So either late this week or early next week.

kantai:
Closing via #2021

PR #2021, opened 2020-10-30
Event dispatcher updates

kantai:
This PR implements #1930 by adding a new endpoint for the event dispatcher:

`POST /new_burn_block`

This payload includes information about burn blocks as their sortitions are processed.
In the event of PoX forks, a `new_burn_block` event may be triggered for a burn block
previously processed.

Example:

```json
{
  "burn_block_hash": "0x4eaabcd105865e471f697eff5dd5bd85d47ecb5a26a3379d74fae0ae87c40904",
  "reward_recipients": [
    {
      "recipient": "1C56LYirKa3PFXFsvhSESgDy2acEHVAEt6",
      "amount": 5000
    }
  ],
  "burn_amount": 12000
}
```

This also implements the `stacks-blockchain` side of things for this issue:

https://github.com/blockstack/stacks-blockchain-api/issues/314

By including a new field in the `new_block` endpoint:

```json
{
   "matured_miner_rewards": [
    {
      "recipient": "ST31DA6FTSJX2WGTZ69SFY11BH51NZMB0ZZ239N96",
      "coinbase_amount": "1000",
      "tx_fees_anchored_shared": "800",
      "tx_fees_anchored_exclusive": "0",
      "tx_fees_streamed_confirmed": "0"
    }
   ]
}
```

For any miner rewards which have matured during the block.


docs/event-dispatcher.md
```diff
@@ -67,10 +67,39 @@ Example:
       "tx_index": 1,
       "txid": "0x738e4d44636023efa08374033428e44eca490582bd39a6e61f3b6cf749b4214c"
     }
-  ]
+   ],
+   "matured_miner_rewards": [
```
zone117x, line 71: Should this include the associated coinbase txid? Thinking about this will be implemented in the API db

kantai: Hmmm -- these matured rewards don't have an exact correspondence with coinbase transactions -- a miner may be receiving matured rewards for microblocks that are being confirmed by a subsequent miner, microblocks they confirmed, etc. This field could add the reward's block hash, which would let the API correlate the rewards with the mined block that generated those rewards.

zone117x: > This field could add the reward's block hash

Makes sense. Should the reward's block hash be used in the API's re-org handling?

kantai: I don't think so. I'll add two fields:

```
from_index_block_hash
from_stacks_block_hash
```

The index block hash will let you pair any rewards to the re-org'ed block.

zone117x (approved):

src/chainstate/stacks/db/accounts.rs
```diff
@@ -57,6 +57,8 @@ pub struct MinerReward {
     pub tx_fees_streamed_produced: u128,
     pub tx_fees_streamed_confirmed: u128,
     pub vtxindex: u32, // will be 0 for the reward to the miner, and >0 for user burn supports
+    pub from_block_consensus_hash: ConsensusHash,
+    pub from_stacks_block_hash: BlockHeaderHash,
```
jcnelson, line 61: This isn't the right place for this information.  The `MinerReward` struct is meant only to encode the breakdown of payments to a particular address, for use in the block-processing loop and miner code to know who to credit STX to and how much.  The lack of a block pointer in this struct is intentional -- a particular address may receive different amount of STX at different blocks (for example, Muneeb asked me to consider making it so that transaction fees mature at a different height than the coinbase -- as much as a _month_ later).  Can you put this information somewhere else, like directly into the epoch receipt?

src/chainstate/stacks/db/mod.rs
```diff
@@ -148,6 +148,7 @@ pub struct StacksHeaderInfo {
 pub struct StacksEpochReceipt {
     pub header: StacksHeaderInfo,
     pub tx_receipts: Vec<StacksTransactionReceipt>,
+    pub matured_rewards: Vec<MinerReward>,
```
jcnelson, line 151: All of the `MinerReward`s in this vector will be matured rewards from the same block, so maybe it makes more sense to put the matured block's block hash and consensus hash in this struct directly, instead of duplicating it _n_ times?

kantai: Sure

jcnelson (approved):
LGTM!  Thanks @kantai

PR #2023, opened 2020-11-02
fix: correct tuple signature

friedger:
## Description

As a developer, I want to know how to create a tuple in clarity. The signature spec and the given example contradict each other.

For details refer to issue https://github.com/blockstack/docs/issues/496

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

No

Rebuild docs once merged.

## Testing information

No

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated
- [x] Tag 1 of @agraebe

### DIFF
--- a/src/vm/docs/mod.rs
+++ b/src/vm/docs/mod.rs
@@ -621,13 +621,8 @@ const TUPLE_CONS_API: SpecialAPI = SpecialAPI {
     description: "The `tuple` special form constructs a typed tuple from the supplied key and expression pairs.
 A `get` function can use typed tuples as input to select specific values from a given tuple.
 Key names may not appear multiple times in the same tuple definition. Supplied expressions are evaluated and
-associated with the expressions' paired key name.
-
-As a shorthand the name `tuple` can be removed, or curly brackets can be used as shown in the examples.",
-    example: "(tuple (name \"blockstack\") (id 1337)) ;; using tuple
-((name \"blockstack\") (id 1337)) ;; using shorthand
-{name: \"blockstack\", id: 1337};; using curly brackets
-",
+associated with the expressions' paired key name.",
+    example: "(tuple (name \"blockstack\") (id 1337))",
 };
 
 const TUPLE_GET_API: SpecialAPI = SpecialAPI {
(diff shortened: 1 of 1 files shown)


agraebe (approved):
lgtm

src/vm/docs/mod.rs
```diff
@@ -617,12 +617,17 @@ If a value did not exist for this key in the data map, the function returns `fal
...
+    description: "The `tuple` special form constructs a typed tuple from the supplied key and expression pairs.
 A `get` function can use typed tuples as input to select specific values from a given tuple.
 Key names may not appear multiple times in the same tuple definition. Supplied expressions are evaluated and
-associated with the expressions' paired key name.",
-    example: "(tuple (name \"blockstack\") (id 1337))"
+associated with the expressions' paired key name.
+
+As a shorthand the name `tuple` can be removed, or curly brackets can be used as shown in the examples.",
+    example: "(tuple (name \"blockstack\") (id 1337)) ;; using tuple
+((name \"blockstack\") (id 1337)) ;; using shorthand
```
kantai, line 628: The `((key value) (key value))` short-hand form isn't allowed except in pretty specific contexts (e.g., `define-map`), so it probably shouldn't be included here. This line is also what's causing the tests to fail, because the line isn't valid Clarity code.

friedger: I have removed the shorthand and also the curly brackets as they are not allowed everywhere neither. Where should the docs for these shorthands go?

kantai: The curly brackets are allowed everywhere that the `(tuple ...)` function is, so I still think it would be appropriate to leave that example in there.

As for the `((key value) (key value))` shorthand, the places that it is allowed already provide examples for its usage (e.g., `define-map`), so I wouldn't worry about adding it anywhere. It's also a shorthand form that is on the possible chopping block (https://github.com/blockstack/stacks-blockchain/issues/1337)

friedger: ok, added the curly brackets shorthand.

kantai (approved):
LGTM!

PR #2024, opened 2020-11-02
Add public helm chart

CharlieC3:
## Description

To help make the stacks blockchain as easy as possible to run in a variety of ways, this PR adds a public Helm chart for the community to easily deploy a stacks node to a local or remote Kubernetes cluster.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] Other

## Test this locally with Minikube
```bash
# If you don't use brew or OSX, https://minikube.sigs.k8s.io/docs/start/
brew install minikube helm
minikube start
helm repo add blockstack https://charts.blockstack.xyz
helm repo update
# Deploys a follower node with basic defaults
helm install stacks-blockchain blockstack/stacks-blockchain
```

No

Not required, but they will be added afterwards (likely this week as it's docs week)
- [X] Link to documentation updates: https://github.com/blockstack/docs/issues/848

### DIFF
--- a/deployment/helm/stacks-blockchain/.helmignore
+++ b/deployment/helm/stacks-blockchain/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/deployment/helm/stacks-blockchain/Chart.yaml
+++ b/deployment/helm/stacks-blockchain/Chart.yaml
@@ -1,0 +1,6 @@
+apiVersion: v2
+name: stacks-blockchain
+description: A Helm chart for the Stacks Blockchain
+type: application
+version: 1.0.0
+appVersion: v23.0.0.10
--- a/deployment/helm/stacks-blockchain/README.md
+++ b/deployment/helm/stacks-blockchain/README.md
@@ -1,0 +1,136 @@
+# Stacks Blockchain
+
+[stacks-blockchain](https://github.com/blockstack/stacks-blockchain) is a layer-1 blockchain that connects to Bitcoin for security and enables decentralized apps and predictable smart contracts.
+
+## TL;DR
+
+```bash
+$ helm repo add blockstack https://charts.blockstack.xyz
+$ helm install my-release blockstack/stacks-blockchain
+```
+
+## Introduction
+
+This chart bootstraps a [stacks-blockchain](https://github.com/blockstack/stacks-blockchain) deployment on a [Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
+
+## Prerequisites
+
+- Kubernetes 1.12+
+- Helm 2.12+ or Helm 3.0-beta3+
+
+## Installing the Chart
+
+To install the chart with the release name `my-release` and run the node as a **follower**:
+
+```bash
+$ helm repo add blockstack https://charts.blockstack.xyz
+$ helm install my-release blockstack/stacks-blockchain
+```
+
+To install the chart with the release name `my-release` and run the node as a **miner** using your private key [generated from the instructions on this page](https://docs.blockstack.org/mining):
+
+```bash
+$ helm repo add blockstack https://charts.blockstack.xyz
+$ helm install my-release blockstack/stacks-blockchain --set config.node.miner=true --set config.node.seed="REPLACE-WITH-YOUR-PRIVATE-KEY"
+```
+
+These commands deploy the Stacks Blockchain node on a Kubernetes cluster in the default configuration.
+
+> **Tip**: List all releases using `helm list`
+
+## Uninstalling the Chart
+
+To uninstall/delete the `my-release` deployment:
+
+```bash
+$ helm delete my-release
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Parameters
+
+The following tables lists the configurable parameters of the stacks-blockchain chart and their default values.
+
+| Parameter | Description | Default |
+|-|-|-|
+| nameOverride | String to partially override stacks-blockchain.fullname template (will maintain the release name) | nil |
+| fullnameOverride | String to fully override stacks-blockchain.fullname template | nil |
+| node.image.repository | stacks-blockchain container image repository | blockstack/stacks-blockchain |
+| node.image.pullPolicy | stacks-blockchain container image pull policy | IfNotPresent |
+| node.image.tag | stacks-blockchain container image tag | latest |
+| node.imagePullSecrets | Docker registry secret names as an array | [] |
+| node.replicaCount | Number of stacks-blockchain nodes to run | 1 |
+| node.rpcPort | Port used for RPC connections | 20443 |
+| node.p2pPort | Port used for P2P connections | 20444 |
+| node.labels | Labels to be added to the node Deployment | {} |
+| node.annotations | Annotations to be added to the node Deployment | {} |
+| node.podAnnotations | Annotations to be added to the node pod(s) | {} |
+| node.podAnnotationConfigChecksum | Add Config Checksum to pod Annotations.<br>This will trigger Rolling Updates on configuration changes. | false |
+| node.podSecurityContext | Security Context policies for the node pod(s) | {} |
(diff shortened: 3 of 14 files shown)


helm/stacks-blockchain/README.md
```diff
@@ -0,0 +1,136 @@
+# Stacks Blockchain
+
+[stacks-blockchain](https://github.com/blockstack/stacks-blockchain) is an layer-1 blockchain that connects to Bitcoin for security and enables decentralized apps and predictable smart contracts.
```
wileyj, line 3: `is an layer-1 blockchain` -> `is a layer-1 blockchain`

helm/stacks-blockchain/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+The following tables lists the configurable parameters of the stacks-blockchain chart and their default values.
+
+| Parameter | Description | Default |
+|-|-|-|
+| nameOverride | String to partially override stacks-blockchain.fullname template (will maintain the release name) | nil |
+| fullnameOverride | String to fully override stacks-blockchain.fullname template | nil |
+| node.image.repository | stacks-blockchain container image repository | blockstack/stacks-blockchain |
+| node.image.pullPolicy | stacks-blockchain container image pull policy | IfNotPresent |
+| node.image.tag | stacks-blockchain container image tag | v23.0.0.10-krypton |
```
wileyj, line 61: For public release, i propose we use the `latest` tag, else we'll have to keep updating this for every new release until mainnet. post-mainnet, we can revert to a stable versioned tag.

helm/stacks-blockchain/README.md
```diff
@@ -0,0 +1,136 @@
...
+| node.p2pPort | Port used for P2P connections | 20444 |
+| node.labels | Labels to be added to the node Deployment | {} |
+| node.annotations | Annotations to be added to the node Deployment | {} |
+| node.podAnnotations | Annotations to be added to the node pod(s) | {} |
+| node.podAnnotationConfigChecksum | Add Config Checksum to pod Annotations.<br>This will trigger Rolling Updates on configuration changes. | false |
+| node.podSecurityContext | Security Context policies for the node pod(s) | {} |
+| node.securityContext | Security Context policies for the node container | {} |
+| node.debug | Set to true for verbose logging. Set to false for normal verbosity | false |
+| node.jsonLogging | Set to true to enable json logging. Set to false for normal logging | false |
+| node.command | Overriding this is typically not necessary, unless you wish to run the node in a different way | ["/bin/stacks-node"] |
```
wileyj, line 74: Is this present in the case of someone using their own image with the chart? Otherwise, i don't see the need to keep this (or at least expose it as an option to override).

helm/stacks-blockchain/README.md
```diff
@@ -0,0 +1,136 @@
...
+| node.args | Overriding this is typically not necessary, unless you wish to run the node in a different way | ["start", "--config", "/src/stacks-node/config.toml"] |
```
wileyj, line 75: Could this be modified so we default to `["start", "--config", "$configFile"]` ? 
I think in the args option, the only thing we'd want to support changing is the config file location

helm/stacks-blockchain/README.md
```diff
@@ -0,0 +1,136 @@
...
+| node.terminationGracePeriodSeconds | Total time it takes for the Container to stop normally | 60 |
+| node.volumes | Additional volumes for the node | [] |
+| node.volumeMounts | Additional volumeMounts for the node | [] |
+| node.extraContainers | Additional containers to run alongside the node. Useful if adding a sidecar | [] |
+| node.initContainers | Containers which are run before the node container is started | [] |
+| config | More configs can be added than what's shown below.All children fields under the node, burnchain, and mstx_balance fields will be converted from YAML to valid TOML format in the Configmap.<br><br>For info on more available config fields, please reference to our [example config files located here](https://github.com/blockstack/stacks-blockchain/tree/master/testnet/stacks-node/conf). |  |
+| config.node.rpc_bind |  | 0.0.0.0:20443 |
+| config.node.p2p_bind |  | 0.0.0.0:20444 |
+| config.node.seed | Replace with your private key if deploying a miner node | nil |
+| config.node.miner | Set this to true if deploying a miner node.<br>Set this to true if deploying a follower node. | false |
```
wileyj, line 93: typo - one of these bools should not be `true`. i suspect the follower mode should be `false`?

helm/stacks-blockchain/README.md
CharlieC3: We've had to change the "command" field on a couple occasions. It's provided as a means for us to support PBC operations but could be useful for others as well.

Tangentially, I'd like the docker image to have an ENTRYPOINT of `/bin/stacks-node` so the k8s deployment wouldn't need to specify a `command` field. Wasn't planning on looping that change in with this PR though.

helm/stacks-blockchain/README.md
CharlieC3: We _could_ allow the location of the config file to be modified, but I don't see the reason for doing that. In the project's current state it'll always need a toml config file, and the volumeMount for the config file is hardcoded to the same location:
https://github.com/blockstack/stacks-blockchain/pull/2024/files#diff-405c70b73ffe7336e4a8cd2e3d655cf1209d535e9f09ffcfc1d37753eca894ecR69-R71

I could calculate the location in `_helpers.tpl` so it's defined in one place, but I don't think we need to add this as a changeable field in `values.yaml` unless there's a reason to do so.

wileyj: resolving this comment - for now it's fine, we should look again once we get closer to mainnet if it still makes sense to keep this as-is (along with any other potential changes for mainnet)

CharlieC3:
@kantai @lgalabru Any thoughts on this PR? I'd like to get it merged into `next`, and then `master` soon if possible, as the PR to add docs was merged a bit early, and it'll have some dead links until these changes are in `master`: https://github.com/blockstack/docs/pull/858

kantai:
This looks good to me, just two points:

1. This is fine to be opened against `master`, because it doesn't have any breaking changes for the stacks-node itself.
2. Can the `helm` directory be moved to a subdirectory of `deployment`?

lgalabru (approved):
Looks good to me, + 1 for moving the `helm` directory!

helm/stacks-blockchain/values.yaml
```diff
@@ -0,0 +1,360 @@
...
+  #   persistentVolumeClaim:
+  #     claimName: stacks-node-chain
+
+  ## Additional volumeMounts for the node
+  ##
+  volumeMounts: []
+  # - name: stacks-node-chain
+  #   mountPath: /root/stacks-node
+
+  ## Additional containers to run alongside the node. Useful if adding a sidecar
```
wileyj, line 182: I think here we shouldn't use the `sidecar` term, since it's named "follower" in the docs. 
To us, we understand a sidecar to mean a follower, but sidecar is also a term used in k8s so it could cause some confusion

helm/stacks-blockchain/values.yaml
```diff
@@ -0,0 +1,360 @@
...
+  #       cpu: "100m"
+
+  ## Containers which are run before the node container is started
+  ##
+  initContainers: []
+  # - name: init-myservice
+  #   image: busybox
+  #   command: ['sh', '-c', 'until nslookup myservice; do echo waiting for myservice; sleep 2; done;']
+
+################################
```
wileyj, line 203: I think we should add some links to our docs in here, specifically in regard to the configfile, but possibly in other areas as well to add some more context for the values a user can set

helm/stacks-blockchain/values.yaml
```diff
@@ -0,0 +1,360 @@
...
+    #     - get
+    #     - list
+    #     - watch
+    #   resources:
+    #     - pods
+
+################################
+####### Metrics Settings #######
+################################
+metrics:
```
wileyj, line 349: I would be more clear here that these metrics are for prometheus (with a link to the prom chart maybe, or prom main site), rather than assume a user knows this is for prometheus

helm/stacks-blockchain/values.yaml
CharlieC3: Because this is a public chart, the term here is meant in the form of the latter, where it's simply a container running on the side of the main container, meant to help in some way. It's not intended to be thought of as the internal PBC sense of the word (which we're moving away from using anyways). I figured it would be appropriate here since this section is for extra containers to run on the side.

CharlieC3:
Will rebase against the `master` branch instead of `next`.

helm/stacks-blockchain/values.yaml
CharlieC3: Added more info and a reference to prometheus docs

helm/stacks-blockchain/values.yaml
CharlieC3: Added a link to our docs and a link to some example config files

CharlieC3:
@kantai @wileyj Comments addressed!

PR #2025, opened 2020-11-03
Fix/2010

jcnelson:
This fixes a regression introduced by #2010 that didn't quite calculate the new RBF relay fee correctly.

It also makes the anti-entropy p2p work step a little less aggressive, only allowing it to run once per burnchain block instead of once per p2p cycle.


kantai (approved):
LGTM!

One thing I've found helpful in avoiding waiting until opening the PR to discover I need to run `cargo fmt` is adding a `cargo fmt -q --all --check` call to my local pre-commit hook.

PR #2026, opened 2020-11-03
bump MAXIMUM_MEMPOOL_TX_CHAINING to 25, make it configurable, resolves #2009

### LINKED ISSUE #2009: Raise value of MAXIMUM_MEMPOOL_TX_CHAINING
The current value is set to 5 (https://github.com/blockstack/stacks-blockchain/blob/master/src/core/mempool.rs#L60), which seems low when considering 10 minutes between blocks.

This currently limits the faucet, and the current version of `swapr` requires deploying 10 contracts, so this would require to split into 2 sets (maybe microblocks would help, a bit, but these would still end up within a block).

I suggest increasing the value to 25, or better yet, make this a node configuration parameter so miners can decide whether they'd want to allow larger values.

psq:
## Description
Addresses #2009, including making MAXIMUM_MEMPOOL_TX_CHAINING by setting node.max_mempool_tx_chaining in the config file, new default is 25 (from 5)

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

- [x] Link to documentation updates: many configurations options are not documented in a consistent manner, yet, so consider documenting all possible configuration options in one location

## Checklist
- [x] Tag  @jcnelson

### DIFF
--- a/net-test/bin/txload.sh
+++ b/net-test/bin/txload.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 FEE_RATE=300
-MAX_CHAINING=5
+MAX_CHAINING=25
 CONFIRMATIONS=1
 
 exit_error() {
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -57,7 +57,7 @@ use rusqlite::Error as SqliteError;
 
 // maximum number of confirmations a transaction can have before it's garbage-collected
 pub const MEMPOOL_MAX_TRANSACTION_AGE: u64 = 256;
-pub const MAXIMUM_MEMPOOL_TX_CHAINING: u64 = 5;
+pub const MAXIMUM_MEMPOOL_TX_CHAINING: u64 = 25;
 
 pub struct MemPoolAdmitter {
     // mempool admission should have its own chain state view.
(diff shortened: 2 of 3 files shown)


src/chainstate/coordinator/tests.rs
```diff
@@ -293,8 +295,13 @@ pub fn get_chainstate_path(path: &str) -> String {
 }
 
 pub fn get_chainstate(path: &str) -> StacksChainState {
-    let (chainstate, _) =
-        StacksChainState::open(false, 0x80000000, &get_chainstate_path(path)).unwrap();
+    let (chainstate, _) = StacksChainState::open(
+        false,
+        0x80000000,
+        &get_chainstate_path(path),
+        MAXIMUM_MEMPOOL_TX_CHAINING,
```
lgalabru, line 302: If we are dealing with a constant, why do we need to pass it in constructors? I'd be in favor of keeping this constructor as lean as possible.

lgalabru: Oh I see we can override using the settings. Isn't something that @jcnelson pushed back on?

jcnelson: Yes, I don't think this should be a setting.  I think that correct nodes should always permit a _minimum_ degree of transaction-chaining that is equal to the constant `MAXIMUM_MEMPOOL_TX_CHAINING`.  If node operators have a particular reason to permit longer chains, they can always re-compile with a bigger value.

psq: will remove

lgalabru (approved):
LGTM!

jcnelson (approved):
LGTM!

PR #2029, opened 2020-11-04
[Do-not-merge] [Stacks v1 → v2 Upgrade] Perform export step on block interval

zone117x:
Ref https://github.com/blockstackpbc/devops/issues/478

Added a new option to trigger the migration export on a block interval (as opposed to name registration threshold). 

Instructions: set the following env vars:
`V2_MIGRATION_EXPORT=1` (enables exporting)
`V2_MIGRATION_EXPORT_TEST_INTERVAL=10` (the block interval to export at)

The file `<datadir>/v2_migration_data.consensus_hash.txt` will be written once the export is complete.

### DIFF
--- a/blockstack/lib/nameset/virtualchain_hooks.py
+++ b/blockstack/lib/nameset/virtualchain_hooks.py
@@ -565,8 +565,22 @@ def db_save( block_height, consensus_hash, ops_hash, accepted_ops, virtualchain_
             log.error("FATAL: failed to update subdomains db at {}".format(block_height))
             os.abort()
 
+        def perform_export():
+            # emit parsable log entry and export datafile
+            log.warn('[v2-upgrade] import threshold reached at block: {}'.format(block_height))
+            db_state.perform_v2_upgrade_datafile_export(block_height, consensus_hash)
+            log.warn('[v2-upgrade] migration datafile export successful')
 
-        # TODO: create a new RPC endpoint to expose the state changes below, for testing when namespace signal threshold reached, and when import block is reached
+        # if the v2_migration_export_test_interval flag is set, perform a datafile export at a block interval
+        test_interval = db_state.v2_migration_export_test_interval
+        if test_interval is not None and test_interval > 0 and block_height % test_interval == 0:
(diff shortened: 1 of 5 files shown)


blockstack/lib/nameset/virtualchain_hooks.py
```diff
@@ -565,8 +565,22 @@ def db_save( block_height, consensus_hash, ops_hash, accepted_ops, virtualchain_
...
+        # if the v2_migration_export_test_interval flag is set, perform a datafile export at a block interval
+        test_interval = db_state.v2_migration_export_test_interval
+        if test_interval is not None and test_interval > 0 and block_height % test_interval == 0:
+            log.warn('[v2-upgrade] triggering export test at block: {}'.format(block_height))
+            try:
+                perform_export()
+            except Exception as e:
+                log.exception(e)
+                log.error("FATAL: failed to export at block {}".format(block_height))
+                log.error("\n".join(traceback.format_stack()))
```
jcnelson, line 583: Do you think this should maybe cause a fatal crash if it fails?

jcnelson:
Thanks @zone117x!  Do you think you can add (or modify) an integration test to cover this new behavior?

jcnelson:
Also, one other thing -- can you make sure that the export step happens a _minimum_ of X blocks after the threshold is reached?  X must be at least 7 (preferably 10).  The reason is that we'll want to give the network a chance to replicate all zone files in the event that the threshold is reached concurrently with a `NAME_UPDATE` or the like.

zone117x:
> Also, one other thing -- can you make sure that the export step happens a _minimum_ of X blocks after the threshold is reached? X must be at least 7 (preferably 10). The reason is that we'll want to give the network a chance to replicate all zone files in the event that the threshold is reached concurrently with a `NAME_UPDATE` or the like.

Good idea. I think this should be implemented as a part of the next PR addressing https://github.com/blockstack/stacks-blockchain/issues/2035. 

The purpose of this PR is only for providing additional testing features required by devops.

PR #2043, opened 2020-11-09
chore: git clone efficiency in readme

zone117x:
Add `--depth=1` to the `git clone...` instructions. 
Reduces disk usage and network download time by around 80%.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ _For building on Windows, follow the rustup installer instructions at https://ru
 From there, you can clone this repository:
 
 ```bash
-git clone https://github.com/blockstack/stacks-blockchain.git
+git clone --depth=1 https://github.com/blockstack/stacks-blockchain.git
 
 cd stacks-blockchain
 ```


jcnelson:
Why?

zone117x:
> Why?

My internet was slow and git clone was taking a surprising amount of time. _Why not_?

jcnelson:
> My internet was slow and git clone was taking a surprising amount of time. Why not?

Why do we need to provide instructions for how to use `git clone` in our README?  Isn't that `git`'s job?  Isn't it up to the person running `git clone` to know how to use it in the appropriate manner?  Also, neither this PR nor my pushback are addressing the underlying problem -- the repo is too big.  How might we address that instead?

zone117x:
`depth=1` is a common instruction for quick start guides. We can blame it on git, delete 4+ years of history from this repo, or leave it be. Seems like a simple decision to me 🤷‍♀️

zone117x:
> Why do we need to provide instructions for how to use `git clone` in our README

We already provide the git command, should we instead change `git clone...` to `acquire this repo`?

Issue #1895, opened 2020-09-21
Update cost functions based on voting contract


reedrosenbluth:
Implement a mechanism for updating a cost-assessment function based on the results of the voting contract

lgalabru:
@reedrosenbluth can this issue be marked as closed?

kantai:
Yep -- it can be closed, was merged in #2143

Issue #2041, opened 2020-11-09
Cost-tracking: Limit cost-functions to arithmetic operations


kantai:
Clarity-defined cost functions must be limited to arithmetic operations, because evaluating these functions is not charged against the block limit, so we must ensure that these functions are simple and may be evaluated efficiently.

jcnelson:
Does this include preventing access to Clarity globals, like `tx-sender`, `contract-caller`, `total-liquid-ustx`, etc.?

kantai:
Additionally, the cost-function verification should ensure that any contract-call targets that are being intercepted (see #2042) are `define-read-only`.

lgalabru:
@kantai can this issue be marked as closed?

kantai:
Yep!

PR #2051, opened 2020-11-11
Load dynamic costs from chainstate

kantai:
## Description

This PR implements #1895 (or, in the absence of the voting contract for now, it implements a sort of stubbed interface), by reading the required cost contract contexts from the chain state at the beginning of a new block.

By virtue of reading the contract context from the chain state, rather than re-parsing in each block, this PR also improves the performance of the LimitedCostTracker initialization a bit. There's a couple other performance improvements here too:

1. The guarded strings in vm/representations use lazy_static to avoid re-initializing the regexes continuously.
2. The AST parsing is not as efficient as it probably could be, but in testing it appeared much much less efficient, because the regex library's opt-level=0 build is painfully slow. This just builds that at opt-level=2 even in dev builds.

Note: because the genesis block _does not_ have the boot cost contract initialized yet, the genesis block must be evaluated "free" (i.e., with cost-tracking disabled).


src/chainstate/stacks/db/mod.rs
```diff
@@ -1072,6 +1054,61 @@ impl StacksChainState {
...
+        // mix burn header hash and stacks block header hash together, since the stacks block hash
+        // it not guaranteed to be globally unique (but the burn header hash _is_).
+        let parent_index_block =
+            StacksChainState::get_parent_index_block(parent_consensus_hash, parent_block);
+
+        let new_index_block =
+            StacksBlockHeader::make_index_block_hash(new_consensus_hash, new_block);
+
+        test_debug!(
+            "Begin processing Stacks block off of {}/{}",
```
jcnelson, line 1081: Can you update the `test_debug!` macros to indicate that this is the instantiation of the genesis block, and not just any old block?

src/net/rpc.rs
```diff
@@ -209,11 +211,11 @@ impl RPCPoxInfoData {
         sortdb: &SortitionDB,
         chainstate: &mut StacksChainState,
         tip: &StacksBlockId,
-        options: &ConnectionOptions,
+        _options: &ConnectionOptions,
     ) -> Result<RPCPoxInfoData, net_error> {
         let contract_identifier = boot::boot_code_id("pox");
         let function = "get-pox-info";
-        let cost_track = LimitedCostTracker::new(options.read_only_call_limit.clone());
+        let cost_track = LimitedCostTracker::new_free();
```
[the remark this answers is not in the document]
kantai: This path only ever calls `get-pox-info`, the read-only call endpoint still sets a limit.

src/vm/costs/mod.rs
```diff
@@ -220,57 +270,70 @@ fn parse_cost(cost_function: ClarityCostFunction, eval_result: InterpreterResult
...
+    let cost_function_reference = cost_tracker
+        .cost_function_references
         .get(&cost_function)
-        .ok_or(CostErrors::CostComputationFailed("CostFunction not defined".to_string()))?.clone();
-
-    let cost_contract = cost_tracker.cost_contracts
-        .get(&cost_function_reference.contract_id)
-        .ok_or(CostErrors::CostComputationFailed("Cost Contract not cached".to_string()))?;
+        .ok_or(CostErrors::CostComputationFailed(
+            "CostFunction not defined".to_string(),
```
jcnelson, line 318: Do you think you could identify the function name and contract address for which there is no cost function in this error message?  Otherwise, debugging this error path is going to be somewhat more difficult.

src/vm/costs/mod.rs
```diff
@@ -220,57 +270,70 @@ fn parse_cost(cost_function: ClarityCostFunction, eval_result: InterpreterResult
...
+        ))?
+        .clone();
 
-    let mut contract_context = ContractContext::new(cost_function_reference.contract_id.clone());
+    let cost_contract = cost_tracker
+        .cost_contracts
+        .get_mut(&cost_function_reference.contract_id)
+        .ok_or(CostErrors::CostComputationFailed(
```
jcnelson, line 325: Should this be a panicking failure?  In what world will the contract be defined but its context not defined?

src/vm/costs/mod.rs
```diff
@@ -220,57 +270,70 @@ fn parse_cost(cost_function: ClarityCostFunction, eval_result: InterpreterResult
...
+    ];
 
-    let function_invocation = SymbolicExpression::list(program.into_boxed_slice());
-    expressions.append(&mut vec![function_invocation]);
+    let function_invocation = [SymbolicExpression::list(program.into_boxed_slice())];
 
-    let eval_result = global_context.execute(|g| {
-        eval_all(&expressions, &mut contract_context, g)
-    });
+    let eval_result = eval_all(&function_invocation, cost_contract, &mut global_context);
```
jcnelson, line 336: Something occurred to me when I was reading this line:  a cost function can assign different costs for a function to different principals, based on who the `sender` and `caller` are.  Is this something we want to support, or something we should prohibit?  The case for prohibiting is that it could lead to a world where miners vote themselves (or their paying clients) discounts for all functions, and get to run smart contracts for as cheap as possible.

kantai: I'd say we definitely want to prevent that -- I'd recommend adding this comment to #2041, as this is another kind of constraint we want to apply to these functions.

src/vm/costs/mod.rs
kantai: The voting contract isn't implemented yet, so it's hard to say where exactly all the checks will be in the interface between this cost tracker initialization and the voting contract, but in theory, a cost function reference could be created that doesn't refer to a loaded contract. Once that check is added to the cost contract validation, yes, this should be panicking.

src/chainstate/stacks/db/mod.rs
kantai: Yep, addressed in 983796523

src/vm/costs/mod.rs
kantai: Addressed in 983796523

src/vm/costs/mod.rs
```diff
@@ -189,26 +218,38 @@ impl LimitedCostTracker {
             free: true,
         }
     }
-    pub fn load_boot_costs(&mut self) {
-        let boot_costs_id = boot_code_id("boot_costs");
+    pub fn load_boot_costs(&mut self, clarity_db: &mut ClarityDatabase) -> Result<()> {
+        let boot_costs_id = boot_code_id("costs");
```
reedrosenbluth, line 222: should `STACKS_BOOT_COST_CONTRACT` be used here instead?

reedrosenbluth (approved):

src/vm/costs/mod.rs
kantai: Yep, updated in 4242ee1

jcnelson (approved):
LGTM!

PR #2052, opened 2020-11-11
Fix 1934: mix burn block height into microblock pubkey

### LINKED ISSUE #1934: Miner: don't re-use microblock public key hashes when starting up for the 2nd time with the same seed
Right now, if you bootstrap miner node a second time with the same seed, it will cycle through the same microblock public key hashes it used when it mined previous blocks.  This causes the miner to mine invalid blocks.  We can fix this by having the miner software generate its microblock keypairs similar to how it does so for VRF keypairs -- by using the block height as input to the key generator.

kantai:
## Description

Fix #1934 by mixing the burn block height into the microblock private key rotation. When attempting RBF, don't rotate the private key (this is just to make testing more deterministic).

## Type of Change

Bug fix

No

No

## Testing information

These changes are covered in existing tests, and required updating two tests' "find_microblock_key" logic.

### DIFF
--- a/src/vm/docs/mod.rs
+++ b/src/vm/docs/mod.rs
@@ -545,7 +545,7 @@ const PRINT_API: SpecialAPI = SpecialAPI {
     input_type: "A",
     output_type: "A",
     signature: "(print expr)",
-    description: "The `print` function evaluates and returns its input expression. On Blockstack Core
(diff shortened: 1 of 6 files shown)


jcnelson (approved):
LGTM.  Can you test it on the testnet before merging?

kantai:
Yep -- tested this on testnet, and my miner successfully mined a handful of blocks

PR #2054, opened 2020-11-12
Log2 native clarity function

reedrosenbluth:
This PR adds an integer binary logarithm function: `(log2 n)`. It also updates the costs boot contract to use said function.

Hoping to merge this into my `feat/dynamic-costs branch`, and then merge #1988 into `next-costs`


src/chainstate/stacks/boot/costs.clar
```diff
@@ -212,6 +210,9 @@
 (define-read-only (cost_sqrti (n uint))
     (runtime u1))
 
+(define-read-only (cost_log2 (n uint))
+    (runtime u1))
```
reedrosenbluth, line 214: should this be constant or linear based on the number of bits?

src/vm/docs/mod.rs
```diff
@@ -233,6 +233,17 @@ const SQRTI_API: SimpleFunctionAPI = SimpleFunctionAPI {
 "
 };
 
+const LOG2_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: None,
+    signature: "(log2 n)",
+    description: "Returns the power to which the number 2 must be raised to to obtain the value `n`, rounded down to the nearest integer.  Fails on a negative numbers.",
```
njordhov, line 239: Also known as the _characteristic_ of the logarithm.

src/vm/functions/arithmetic.rs
```diff
@@ -226,10 +226,21 @@ macro_rules! make_arithmetic_ops {
                     }
                 }
             }
+            fn log2(n: $type) -> InterpreterResult<Value> {
+                if n < 0 {
```
njordhov, line 230: Should be `n < 1` as the logarithm of zero is undefined.

src/vm/functions/arithmetic.rs
```diff
@@ -226,10 +226,21 @@ macro_rules! make_arithmetic_ops {
...
+                    return Err(RuntimeErrorType::Arithmetic(
+                        "log2 must be passed a positive number".to_string()
```
njordhov, line 232: Should be _positive integer_.

src/chainstate/stacks/boot/costs.clar
kantai: This should be constant -- there's a size limit for our integers.

src/vm/tests/simple_apply_eval.rs
```diff
@@ -477,6 +477,12 @@ fn test_simple_arithmetic_functions() {
         // from https://en.wikipedia.org/wiki/128-bit_computing
         "(sqrti 170141183460469231731687303715884105727)", // max i128
         "(sqrti u340282366920938463463374607431768211455)", // max u128
+        "(log2 u8)",
+        "(log2 u9)",
+        "(log2 8)",
+        "(log2 9)",
+        "(log2 170141183460469231731687303715884105727)", // max i128
+        "(log2 u340282366920938463463374607431768211455)", // max u128
```
kantai, line 485: Can you add error cases to the `test_simple_arithmetic_errors`?

jcnelson (approved):
I have nothing more to add on top of @njordhov's and @kantai's feedback.  Please address them before merging, and this will LGTM.  Thanks for the quick turn-around!

src/vm/docs/mod.rs
njordhov: Should be _fails on a non-positive integer_ as the logarithm of zero is undefined.

Issue #1315, opened 2020-03-10
Sortition: use mean of min and median as weights, smoothed over 6 bitcoin blocks


jcnelson:
Right now, cryptographic sortition probability weights are assigned according to each miner's relative PoX transfers.  In order to (1) decrease the profit margin a miner can obtain by both Stacking and mining (which leads to consolidation over long enough timescales), and (2) ensure that the most profitable course of action for a miner is to mine at a steady pace at all times (instead of trying to play games with their PoX rates when their Stacking addresses come up), the sortition probability weights need to be changed.  Modeling this out, it seems like taking the average of a miner's minimum transfer amount in the last 6 blocks, and the median transfer amount in the last 6 blocks, achieves this.

muneeb-ali:
Makes sense to do this over 6 blocks. This gives a STX miner roughly 1 hour to change strategy.

diwakergupta:
@blockstack/blockchain-team this hasn't been implemented yet, right?

jcnelson:
Nope, still a TODO

kantai:
One snag here is that we need a way to identify a miner across multiple leader_block_commits (even if we don't ever see the associated Stacks blocks). We cannot use VRF keys for this, because those aren't "authenticated" until the Stacks block is revealed.

I think we'd need to do something like require that each block commit spend an output from the prior commit (being careful to ensure that no two leader block commits could claim the same prior commit). That's certainly implementable (it would make use of the tx_ref field already added in #2015), but I want to confirm that we should go this route beforehand.

jcnelson:
I think doing this through UTXO-chaining would be sufficient.  A simpler and potentially more-robust approach would be to just add a dedicated "miner ID" BTC output to the block-commit, which must contain a scriptPubKey that we can parse.  In most cases, this "miner ID" output would also be the miner's change output.  This removes the need for miners to keep the UTXO chain going (i.e. they can spend their BTC intermittently between block-commits, or use a different funding key), and it should achieve the same purpose (i.e. identify miners' block-commits).

Regardless of how we do it, though, we'll want to store each miner's previous burn/xfer amounts in the sortition DB MARF, not just the burnchain DB (since the burnchain can fork).

kantai:
[...]

PR #2063, opened 2020-11-16
#1315: Min/Median Block Mining

kantai:
## Description

This implements mean of min and median block mining #1315.

There's a couple things happening in this PR:

1. Identification of miners is done through UTXO chaining.
2. User burn operations may identify multiple block commits (the prior implementation relied on VRF key validation, which cannot be done before Stacks block reveal)
3. The burn distribution is calculated using a mean of min and median burns over a 6-block window.

## Type of Change
- New feature

Yes -- this is a consensus breaking change. Miners _do not_ need to change their behavior, though doing so would probably be advantageous.

## Testing information

There's unit test coverage for the new implementation of `BurnSamplePoint::make_min_median_distribution`. Integration with the sortdb / burnchain processing code is covered by existing integration tests (the integration itself is checked using panicking assertions in the new `BurnSamplePoint::make_min_median_distribution` function).

### DIFF
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -2114,6 +2114,7 @@ pub mod tests {
                 block_122_snapshot.consensus_hash.clone(),
                 ConsensusHash::from_hex("0000000000000000000000000000000000000000").unwrap(),
             ];
+
             let burn_total = block_ops_124.iter().fold(0u64, |mut acc, op| {
                 let bf = match op {
                     BlockstackOperationType::LeaderBlockCommit(ref op) => op.burn_fee,
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -1753,6 +1753,14 @@ pub mod tests {
 
             burn_fee: 12345,
             input: (Txid([0; 32]), 0),
+            apparent_sender: BurnchainSigner {
+                public_keys: vec![StacksPublicKey::from_hex(
+                    "02d8015134d9db8178ac93acbc43170a2f20febba5087a5b0437058765ad5133d0",
+                )
+                .unwrap()],
+                num_sigs: 1,
+                hash_mode: AddressHashMode::SerializeP2PKH,
+            },
 
             txid: Txid::from_bytes(
                 &hex_bytes("3c07a0a93360bc85047bbaadd49e30c8af770f73a37e10fec400174d2e5f27cf")
@@ -1785,6 +1793,14 @@ pub mod tests {
 
             burn_fee: 12345,
             input: (Txid([0; 32]), 0),
+            apparent_sender: BurnchainSigner {
+                public_keys: vec![StacksPublicKey::from_hex(
+                    "02d8015134d9db8178ac93acbc43170a2f20febba5087a5b0437058765ad5133d0",
+                )
+                .unwrap()],
+                num_sigs: 1,
+                hash_mode: AddressHashMode::SerializeP2PKH,
+            },
 
             txid: Txid::from_bytes(
                 &hex_bytes("3c07a0a93360bc85047bbaadd49e30c8af770f73a37e10fec400174d2e5f27d0")
@@ -1817,6 +1833,14 @@ pub mod tests {
 
             burn_fee: 23456,
             input: (Txid([0; 32]), 0),
+            apparent_sender: BurnchainSigner {
+                public_keys: vec![StacksPublicKey::from_hex(
+                    "0283d603abdd2392646dbdd0dc80beb39c25bfab96a8a921ea5e7517ce533f8cd5",
+                )
+                .unwrap()],
+                num_sigs: 1,
+                hash_mode: AddressHashMode::SerializeP2PKH,
+            },
 
             txid: Txid::from_bytes(
                 &hex_bytes("301dc687a9f06a1ae87a013f27133e9cec0843c2983567be73e185827c7c13de")
@@ -2251,6 +2275,14 @@ pub mod tests {
 
                 burn_fee: (i + 1) as u64,
                 input: (Txid([0; 32]), 0),
+                apparent_sender: BurnchainSigner {
+                    public_keys: vec![StacksPublicKey::from_hex(
+                        "02113c274c05ed0b7f9d08f41ca674b22e42188408caaff82a350b024442de353c",
+                    )
+                    .unwrap()],
+                    num_sigs: 1,
+                    hash_mode: AddressHashMode::SerializeP2PKH,
+                },
 
                 txid: Txid::from_bytes(&vec![
                     i, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
@@ -2277,6 +2309,14 @@ pub mod tests {
 
             burn_fee: 256,
             input: (Txid([0; 32]), 0),
+            apparent_sender: BurnchainSigner {
+                public_keys: vec![StacksPublicKey::from_hex(
+                    "02113c274c05ed0b7f9d08f41ca674b22e42188408caaff82a350b024442de353c",
+                )
+                .unwrap()],
+                num_sigs: 1,
+                hash_mode: AddressHashMode::SerializeP2PKH,
+            },
 
             txid: Txid([0xdd; 32]),
             vtxindex: 1,
@@ -2493,6 +2533,14 @@ pub mod tests {
 
                     burn_fee: i as u64,
                     input: (Txid([0; 32]), 0),
+                    apparent_sender: BurnchainSigner {
+                        public_keys: vec![StacksPublicKey::from_hex(
+                            &leader_bitcoin_public_keys[(i - 1) as usize].clone(),
+                        )
+                        .unwrap()],
+                        num_sigs: 1,
+                        hash_mode: AddressHashMode::SerializeP2PKH,
+                    },
 
                     txid: Txid::from_bytes(&vec![
                         i, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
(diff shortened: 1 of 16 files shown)


src/burnchains/burnchain.rs
```diff
@@ -157,14 +160,55 @@ impl BurnchainStateTransition {
...
+
+        for blocks_back in 0..(MINING_COMMITMENT_WINDOW - 1) {
+            if parent_snapshot.block_height < (blocks_back as u64) {
+                debug!("Mining commitment window shortened because block height is less than window size";
+                       "block_height" => %parent_snapshot.block_height,
+                       "window_size" => %MINING_COMMITMENT_WINDOW);
+                break;
+            }
+            let block_height = parent_snapshot.block_height - (blocks_back as u64);
+            let sortition_id = match sort_tx.get_block_snapshot_by_height(block_height)? {
```
jcnelson, line 175: It might be more efficient to just read the parent snapshot using the current snapshot's parent snapshot pointer (this would save you a MARF read per block height).

src/burnchains/burnchain.rs
```diff
@@ -157,14 +160,55 @@ impl BurnchainStateTransition {
...
+                &sortition_id,
+            )?);
+        }
+
+        // reverse vecs so that windows are in ascending block height order
+        windowed_block_commits.reverse();
+        windowed_user_burns.reverse();
+
+        // figure out if the PoX sunset finished during the window
+        let window_end_height = parent_snapshot.block_height + 1;
```
jcnelson, line 194: Does the window get disabled after PoX sunset?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -1424,14 +1430,47 @@ impl<'a> SortitionHandleConn<'a> {
...
+        ];
+        let shared_miners = query_count(self, qry, &args)? as u64;
 
-        query_rows(self, qry, &args)
+        assert!(
+            shared_miners >= 1,
+            "BUG: Should be at least 1 matching miner for the winning block commit"
+        );
+
+        for winning_user_burn in winning_user_burns.iter_mut() {
```
jcnelson, line 1469: I would prefer it if this adjustment happened within the sortition distribution logic, instead of in the sortition DB.

src/burnchains/burnchain.rs
kantai: The snapshot only stores a burnchain header hash, which would need a MARF lookup to resolve to a SortitionId. We could extend the BlockSnapshot structure to store a parent pointer as SortitionId, though.

src/burnchains/burnchain.rs
kantai: No -- it just changes which output it tries to match in the BlockCommitOp for UTXO chaining (because after the sunset completes, the BlockCommitOp has 1 fewer output).

src/chainstate/burn/distribution.rs
```diff
@@ -340,6 +442,217 @@ mod tests {
...
+        assert_eq!(result[0].candidate.txid, commits[5][0].txid);
+        assert_eq!(result[1].candidate.txid, commits[5][1].txid);
+
+        assert_eq!(result[0].user_burns.len(), 1);
+        assert_eq!(result[1].user_burns.len(), 1);
+
+        assert_eq!(result[1].user_burns[0].txid, user_burns[5][0].txid);
+        assert_eq!(result[0].user_burns[0].txid, user_burns[5][0].txid);
+    }
+
```
jcnelson, line 655: Can you add a test for the edge case where the PoX sunset takes effect?

src/chainstate/burn/distribution.rs
```diff
@@ -38,195 +39,301 @@ use util::log;
...
-    pub fn make_distribution(
-        all_block_candidates: Vec<LeaderBlockCommitOp>,
-        consumed_leader_keys: Vec<LeaderKeyRegisterOp>,
-        user_burns: Vec<UserBurnSupportOp>,
+    ///
+    /// * `block_commits`: this is a mapping from relative block_height to the block
+    ///     commits that occurred at that height. These relative block heights start
+    ///     at 0 and increment towards the present. When the mining window is 6, the
+    ///     "current" sortition's block commits would be in index 5.
+    /// * `block_commits`: this is a mapping from relative block_height to the user
```
jcnelson, line 84: Typo -- `user_burns`?

src/chainstate/burn/distribution.rs
```diff
@@ -38,195 +39,301 @@ use util::log;
...
-
-        // sanity check -- each block commit must refer to a consumed key
-        for i in 0..all_block_candidates.len() {
-            let bc = &all_block_candidates[i];
-            match key_index.get(&(bc.key_block_ptr as u64, bc.key_vtxindex as u32)) {
-                None => {
-                    panic!(
-                        "No leader key for block commitment {} at ({},{}) -- points to ({},{})",
-                        &bc.txid, bc.block_height, bc.vtxindex, bc.key_block_ptr, bc.key_vtxindex
+        // sanity check
```
jcnelson, line 95: For clarity's sake, it might be worth putting the sanity check into its own private method.

src/chainstate/burn/db/sortdb.rs
kantai: This method is only ever invoked by the anchor block processing -- I could update this to return the original block commit ops and the number of block commits that were identified by the same user burns. Then the caller (`preprocess_anchor_block`) would be responsible for applying the adjustment before storing the staging user burns. Otherwise, I think this would entail a lot of refactoring of how the miner rewards are calculated.

src/chainstate/burn/distribution.rs
```diff
@@ -38,195 +39,301 @@ use util::log;
...
+
+                let user_burn_target_key = UserBurnIdentifier {
+                    rel_block_height: rel_block_height as u8,
+                    key_vtxindex: key_vtxindex,
+                    key_block_ptr: key_block_ptr,
+                    block_hash: block_header_hash_160,
+                };
+
+                if let Some(user_burn_recipients) = user_burn_targets.get(&user_burn_target_key) {
+                    let per_recipient = burn_fee / (user_burn_recipients.len() as u64);
```
jcnelson, line 235: Wait, didn't we already do this in the sortition DB?

src/chainstate/burn/distribution.rs
```diff
@@ -38,195 +39,301 @@ use util::log;
...
+            .map(|mut linked_commits| {
+                let mut all_burns: Vec<_> = linked_commits
+                    .iter()
+                    .map(|commit| {
+                        if let Some(commit) = commit {
+                            (commit.op.burn_fee as u128) + (commit.user_burns as u128)
+                        } else {
+                            // use 1 as the linked commit min. this gives a miner a _small_
+                            //  chance of winning a block even if they haven't performed chained utxos yet
+                            1
```
jcnelson, line 289: What happens if this is 0?

src/chainstate/burn/distribution.rs
kantai: No -- the function in the sortition db is only ever invoked by the block preprocessing logic. The primary difference between what's happening in that code and what's happening here is that in the other function, the caller wants specifically only the user burns that supported a particular known-to-have-won block. That data is used in matured rewards calculations.

Here, we need to evaluate all the possible user burns over the window.

src/chainstate/burn/distribution.rs
```diff
@@ -38,195 +39,301 @@ use util::log;
...
-                        candidate: bc.clone(),
-                        key: consumed_leader_keys[*i].clone(),
-                        user_burns: vec![],
+        let window_size = block_commits.len() as u8;
+
+        // first, let's link all of the current block commits to the priors
+        let mut commits_with_priors: Vec<_> =
+            // start with the most recent
+            block_commits
+            .remove((window_size - 1) as usize)
```
jcnelson, line 132: I'm a little confused by this -- are we dropping the current sortition's block commit here?

src/chainstate/burn/distribution.rs
kantai: It leads to a corner case where if there's no mining for some period of time, no miner will get pick for 3-5 blocks once mining resumes.

src/chainstate/burn/distribution.rs
kantai: No -- it's pulling it out of the input array to begin building an array with all the linked commits attached.

src/chainstate/burn/distribution.rs
jcnelson: ah, gotcha

src/burnchains/burnchain.rs
jcnelson: I thought the BlockSnapshot already stored the sortition ID?  If not, then don't worry about it in this PR.

src/chainstate/burn/distribution.rs
kantai: Yep, will add

src/chainstate/burn/db/sortdb.rs
jcnelson: Ah, I see.

jcnelson (approved):
This LGTM, but maybe just add one more test to cover the case where PoX sunset occurs during the calculation.

src/chainstate/burn/distribution.rs
kantai: Updated in 5c80770d4

src/chainstate/burn/distribution.rs
kantai: Updated in 5c80770d4

src/chainstate/burn/distribution.rs
kantai: Added in 5c80770d4

lgalabru (approved):
I keep wondering if the rigidity of our expectations, when diving into the outputs of a BTC transaction, is a good thing or not (the order, the total count, etc).
I understand that it simplifies the retrieval, but there are 2 downsides that I can think about:
- Combining Stacks mining / support with BTC transactions: transaction's price are rising, BTC fee costs of Stacks mining could be amortized if the transaction could be used for something else.   
- Combining / Multiplexing Stacks protocols with other protocols: can we / what kind of meta-protocols can be built by combining some Bitcoin protocols (Stacks, Lightning, wBTC, etc).

kantai:
Opened an issue to discuss the rigidity of our bitcoin wire formats https://github.com/blockstack/stacks-blockchain/issues/2094

PR #1443, opened 2020-04-19
Eliminate implicit tuples

njordhov:
3L1m1N473 1MPL1c17 7upl32 R3s0Lv1n' 155uE #1337 ;-)

Simplifies the codebase by eliminating hardcoded handling of the legacy implicit tuple syntax. Updates clarity test cases, examples and code snippets to consistently use tuple literals.

PR #2064, opened 2020-11-16
Fix: do not bypass read only checks for writing operations

lgalabru:
Address https://github.com/blockstack/stacks-blockchain/issues/1852.
Note: this PR is breaking compatibility with tuple's implicit syntax, so we should first sunset this deprecated syntax, before merging this PR.

### DIFF
--- a/.github/actions/bitcoin-int-tests/Dockerfile.rustfmt
+++ b/.github/actions/bitcoin-int-tests/Dockerfile.rustfmt
@@ -9,4 +9,4 @@ RUN rustup component add rustfmt
 
 COPY . .
 
-RUN cargo fmt --all -q -- --check
+RUN cargo fmt --all -- --check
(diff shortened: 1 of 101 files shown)


kantai (changes requested):
I think if this PR is going to deprecate the implicit tuple syntax, it should also remove the implicit tuple processing code from the other analysis passes (and the runtime). I think this also be opened against `next`.

njordhov:
> if this PR is going to deprecate the implicit tuple syntax, it should also remove the implicit tuple processing code from the other analysis passes (and the runtime).

PR #1443 eliminated implicit tuple processing code but was never merged thus withdrawn when the code went stale. Consider checking whether parts can be used so you don't have to repeat the work.

lgalabru:
Apologies if the opening comment was confusing. I initially wrote a patch for the bug that this PR is addressing, then realized that the tuple's implicit syntax would make things harder. 

I was thinking about creating another PR, sunsetting the implicit syntax and then enquire peer reviews on this one.

kantai (approved):

PR #2065, opened 2020-11-17
Add zonefile_hash to chainstate.json

zone117x:
Adds the property `zonefile_hash` to the names exported in `chainstate.json`.

Note that `blockstack.id` and `bitcoinbook.id` do not have zonefiles (or zonefile_hashes), so we can just use zeros/null for these. 

Here's an updated export data file dump: https://mega.nz/file/ERdmlZBb#ghlwCwEk5AY3JqJkFAtdLGe_Jnb6HthccBe9H249W_M


lgalabru (approved):
LGTM, thank you @zone117x!

jcnelson (approved):
LGTM!

kantai (approved):

Issue #2042, opened 2020-11-09
Cost-tracking: Allow defined costs to interpose on contract-call


kantai:
Voted-in cost contracts should be allowed to define cost functions for existing read-only functions.

kantai:
Closed via #2072

PR #2072, opened 2020-11-18
Cost-tracking: #2042, intercepting contract-call costs

kantai:
## Description

This PR implements #2042 -- it allows a CostTracker to intercept the cost calculation of contract-call invocations when specific targets are used.

This will be used in the cost voting system to allow STX holders and miners to introduce new "native" Clarity functions by specifying a new cost for those functions.

## Type of Change
- New feature

No, this is not a breaking change.

No.

## Testing information

This PR includes unit tests covering the behavior of the LimitedCostTracker with contract-call interceptions.

### DIFF
--- a/src/vm/contexts.rs
+++ b/src/vm/contexts.rs
@@ -60,7 +60,7 @@ pub struct Environment<'a, 'b> {
 }
 
 pub struct OwnedEnvironment<'a> {
-    pub context: GlobalContext<'a>,
+    context: GlobalContext<'a>,
     default_contract: ContractContext,
     call_stack: CallStack,
 }
--- a/src/vm/contexts.rs
+++ b/src/vm/contexts.rs
@@ -17,6 +17,7 @@
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::convert::TryInto;
 use std::fmt;
+use std::mem::replace;
 
 use vm::ast;
 use vm::ast::ContractAST;
@@ -59,7 +60,7 @@ pub struct Environment<'a, 'b> {
 }
 
 pub struct OwnedEnvironment<'a> {
-    context: GlobalContext<'a>,
+    pub context: GlobalContext<'a>,
     default_contract: ContractContext,
     call_stack: CallStack,
 }
@@ -644,6 +645,16 @@ impl CostTracker for Environment<'_, '_> {
     fn reset_memory(&mut self) {
         self.global_context.cost_track.reset_memory()
     }
+    fn short_circuit_contract_call(
+        &mut self,
+        contract: &QualifiedContractIdentifier,
+        function: &ClarityName,
+        input: &[u64],
+    ) -> std::result::Result<bool, CostErrors> {
+        self.global_context
+            .cost_track
+            .short_circuit_contract_call(contract, function, input)
+    }
 }
 
 impl CostTracker for GlobalContext<'_> {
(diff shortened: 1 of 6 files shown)


src/vm/contexts.rs
```diff
@@ -779,6 +799,23 @@ impl<'a, 'b> Environment<'a, 'b> {
...
+    pub fn run_free<F, A>(&mut self, to_run: F) -> A
+    where
+        F: FnOnce(&mut Environment) -> A,
+    {
+        let original_tracker = replace(
+            &mut self.global_context.cost_track,
+            LimitedCostTracker::new_free(),
+        );
+        // note: it is important that this method not return until original_tracker has been
+        //  restored. DO NOT use the try syntax (?).
```
jcnelson, line 813: Not sure what this comment is saying -- isn't that already guaranteed here by the way the code is written?  Or, is this a comment on the implementation of `to_run`?

jcnelson:
So, issue #2042 said that this should be limited to read-only contract-calls?  I don't see any read-only checks.  Is this okay?

src/vm/costs/mod.rs
```diff
@@ -409,6 +457,26 @@ impl CostTracker for LimitedCostTracker {
...
+        input: &[u64],
+    ) -> Result<bool> {
+        if self.free {
+            // if we're already free, no need to worry about short circuiting contract-calls
+            return Ok(false);
+        }
+        // grr, if HashMap::get didn't require Borrow, we wouldn't need this cloning.
+        let lookup_key = (contract.clone(), function.clone());
+        if let Some(cost_function) = self.contract_call_circuits.get(&lookup_key).cloned() {
+            let input_size = input.iter().fold(0, |agg, cur| agg + cur);
```
jcnelson, line 473: I guess I don't understand what `input` is here.  Is input just a list of instance characteristics?  If so, then why is _summing_ them the right thing to do?  Consider matrix multiplication as a counterexample -- the runtime would be the _product_ of the instance characteristics.  Or, am I misunderstanding what's going on here?

src/vm/contexts.rs
kantai: Yeah, it's a note to future modifiers of that code.

src/vm/costs/mod.rs
kantai: That's the size of each input argument. Ultimately, this is going to change to just be multiple arguments, and then the defined cost function can choose how to combine them.

jcnelson: Gotcha.  Is there an issue open for this?  If not, can you open one?

kantai:
> So, issue #2042 said that this should be limited to read-only contract-calls? I don't see any read-only checks. Is this okay?

Ah, yes, this should check that the target is read-only.

src/vm/costs/mod.rs
kantai: Opened: https://github.com/blockstack/stacks-blockchain/issues/2075

kantai:
>> So, issue #2042 said that this should be limited to read-only contract-calls? I don't see any read-only checks. Is this okay?
>
> Ah, yes, this should check that the target is read-only.

Actually, looking more into this, I think we'll want to enforce this at the same time we enforce #2041. To enforce it at time-of-invocation, we'd need to look up the target function in the MARF to check whether or not it is read-only, which is already much of the cost of invocation.

src/vm/costs/mod.rs
reedrosenbluth: We've actually already got this issue: https://github.com/blockstack/stacks-blockchain/issues/2047

lgalabru (approved):

lgalabru:
Just asked for a round of review - merging this branch would unblock progress on https://github.com/blockstack/stacks-blockchain/pull/2109

reedrosenbluth (approved):

jcnelson (approved):
LGTM!

Issue #2066, opened 2020-11-17
Simplify the closure applications on chainstate initialization  


lgalabru:
Context https://github.com/blockstack/stacks-blockchain/pull/1929#discussion_r523016975

TLDR: we are currently nesting a closure in a closure for properly setting up the chainstate. We should revisit this structure and simplify int, to increase the legibility of the boot sequence

PR #2073, opened 2020-11-18
Xenon adjustments

lgalabru:
This PR is fixing an off by one error, previously patched in the previous Xenon PR, mistakenly discarded during one of the multiple rebase / conflict resolution sessions.

It's also addressing a race condition that I've been experiencing during my tests, where the main StacksChainState in charge of building the genesis block, is being initialized after the one initialized in the PoxSyncWatchdog.
By doing so, it's also addressing #2066.


testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -149,45 +155,66 @@ impl RunLoop {
...
 
-        thread::spawn(move || {
-            let mut boot_data =
-                ChainStateBootData::new(&coordinator_burnchain_config, initial_balances, None);
+        let first_block_height = burnchain_config.first_block_height as u128;
+        let pox_prepare_length = burnchain_config.pox_constants.prepare_length as u128;
+        let pox_reward_cycle_length = burnchain_config.pox_constants.reward_cycle_length as u128;
+        let pox_rejection_fraction = burnchain_config.pox_constants.pox_rejection_fraction as u128;
+
+        let boot_block = Box::new(move |clarity_tx: &mut ClarityTx| {
```
kantai, line 166: It doesn't seem correct to me for this code to be outside the `chainstate` module -- I think it'd be better to do something like introduce an `initialize` method to coordinator, which we can run synchronously, to ensure that the coordinator is the first method to call `open`.

This should probably also change the watchdog to `connect`, rather than `open`

lgalabru: How would you address https://github.com/blockstack/stacks-blockchain/issues/2066 with this approach?

kantai: I think one of two things:

1. Not address it all. I'm not too worried about this being difficult to understand-- I think it's ultimately pretty easy to find where the boot block code is getting initialized, since there are relatively few calls to open_and_exec. We could also always add a debug! log to the ChainStateBootData object to log every time a closure is added if it makes debugging easier.
2. Move the code from the chain state coordinator into the StacksChainState code that inserts the boot contracts, so that it is executed always, this eliminates one of the closures.

jcnelson: I'd be in favor of (2).  Part of the rationale for #2066 is that external parties may need to audit the genesis block -- i.e. for legal and tax compliance purposes (this has happened before in Stacks 1.0).  Making it straightforward to show auditors that the genesis block was correctly instantiated will significantly reduce our collective anxiety when assisting with these audits.

kantai: Either way, I'm also worried about the race on `open()` -- open() should only be called once, and everywhere else that is currently calling `open` should switch to `connect` and panic if the database hasn't already been created.

lgalabru: A similar panic behavior was already implemented. It's not at upfront at the method you're describing, but essentially, if the first call to open in not providing a `ChainStateBootData`, then you'll get a panic. That's how I was able to experience this bug on Bitcoin Testnet.

testnet/stacks-node/src/neon_node.rs
```diff
@@ -600,8 +601,11 @@ fn spawn_miner_relayer(
                     }
                 }
                 RelayerDirective::RegisterKey(ref last_burn_block) => {
-                    debug!("Relayer: Register key");
-                    rotate_vrf_and_register(
+                    if did_register_key {
```
kantai, line 604: The miner wouldn't know that the register key operation was accepted or not when `did_register_key` was set -- I think that would just cause the miner to never successfully mine, and could be remedied by restarting. Was this necessary because `RegisterKey` directives are being issued multiple times on the first block that the miner activates for (because the miner gets "woken up" multiple times per block now, unlike when the neon_node was first written)?

lgalabru: >  Was this necessary because RegisterKey directives are being issued multiple times on the first block

Correct

lgalabru: I can reset this flag, when the a directive is occurring on a different block height, but it still feels like a patch, from what I saw at runtime PoxWatchdog looks very "nervous" and could receive an upgrade.

lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2073/commits/c1798bc165053d18c1c2ccf4275041237fa0b9b0

src/burnchains/bitcoin/indexer.rs
```diff
@@ -851,6 +851,21 @@ impl BurnchainIndexer for BitcoinIndexer {
             .map_err(burnchain_error::Bitcoin)
     }
 
+    fn get_highest_header_height(&self) -> Result<u64, burnchain_error> {
```
jcnelson, line 854: What was wrong with `get_headers_height()`?

testnet/stacks-node/src/neon_node.rs
```diff
@@ -1026,11 +1035,25 @@ impl InitializedNeonNode {
...
-                error!(
-                    "Failed to generate proof with {:?}",
-                    &registered_key.vrf_public_key
-                );
-                return None;
+                // Try to recover a key registered in a former session.
+                // registered_key.block_height gives us a pointer to the height of the block
+                // holding the key register op, but the VRF was derived using the height of one
+                // of the parents blocks.
+                let _ = keychain.rotate_vrf_keypair(registered_key.block_height - 1);
```
jcnelson, line 1042: I'm not sure what's going on here?

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -236,10 +263,12 @@ impl RunLoop {
             });
         }
 
-        let mut burnchain_height = 1;
+        let mut block_height = burnchain_tip.block_snapshot.block_height;
```
jcnelson, line 266: If the node is booting up, then shouldn't the first target burnchain height be equal to the first burnchain block height at which the node starts indexing?

src/burnchains/bitcoin/indexer.rs
lgalabru: Using `get_headers_height()` was fine (except in one place - burnchain.rs line 508), but introducing `get_highest_header_height` increase legibility: there are many places in the codebase where we're calling `get_header_height()` and subtracting 1 + adding a comment explaining that it's 1-indexed.

jcnelson: That's because `get_headers_height()` reports the number of headers the node knows about (i.e. on the longest bitcoin fork).  If you want to convert the number of headers (which is 1-indexed, because there may be 0 headers) into a block height (which is 0-indexed, because block 0 is the first block), then maybe instead of duplicating the `get_headers_height()` implementation, you could add a helper method that converts the number of headers into a block height?  Something like `num_headers_to_block_height(num_headers: u64) -> Option<u64> { num_headers.checked_sub(1) }`.  That would more accurately describe what's going on here.

testnet/stacks-node/src/run_loop/neon.rs
lgalabru: Not sure, I think this is what we want, however there might be an issue line 271, where `target_burnchain_block_height` should probably be:

```
        let blocks_before_reward_cycle = (burnchain_height - burnchain_config.first_block_height) % pox_constants.reward_cycle_length as u64;
        target_burnchain_block_height = burnchain_height + blocks_before_reward_cycle;
```
No? (cc @kantai)

src/burnchains/bitcoin/indexer.rs
lgalabru: > which is 1-indexed, because there may be 0 headers

This statement and introducing this count as 1-indexed is very confusing, we **always** have at least 1 header in this DB - the genesis header (inserted when initializing these DBs).

lgalabru: > Something like num_headers_to_block_height(num_headers: u64) -> Option<u64> { num_headers.checked_sub(1) }. That would more accurately describe what's going on here.

I don't see how 
```
let destination = (source + 1) - 1
```
is more accurately describing
```
let destination = source
```

testnet/stacks-node/src/run_loop/neon.rs
kantai: I think there's a little confusion here with line 216 -- `burnchain.wait_for_sortitions()`, which will try to wait until the sortition db has caught up with the burnchain tip, and then the activity of the pox sync watchdog, which also does some initial synchronization between the burnchain tip and the sortition db.

I'm not sure what the best way to clean that up is -- the pox sync watchdog should probably be in charge of handling this, but I think if we just blindly removed the `burnchain.wait_for_sortitions()` call, the neon_node would probably receive _way_ more events during start up than it currently does (the `block_height` variable is used to ignore sending "mine block" directives before the first tip).

src/burnchains/bitcoin/indexer.rs
jcnelson: My main concern is that you have two method implementations that do almost the exact same thing.  I think you should either fix `get_headers_height()` so it does what you want (and remove `get_highest_header_height()`), or you should make `get_highest_header_height()` call `get_headers_height()` internally.  If it means changing `get_headers_height()` to be 0-indexed, then that's fine by me.

testnet/stacks-node/src/neon_node.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2073/commits/f0baf47900640401ab6f6dee385b6b7d7e25e89f

src/burnchains/bitcoin/indexer.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2073/commits/249aa38a1e3a630f1b5ad6748e25f8f5846a2bc5

testnet/stacks-node/src/run_loop/neon.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2073/commits/8285cb26fce46c9d21bba5786d9fa399ba615a09

testnet/stacks-node/src/run_loop/neon.rs
lgalabru: Stashing some code for now and will address in https://github.com/blockstack/stacks-blockchain/issues/2095 so we can make some progress on Xenon deployments

src/burnchains/bitcoin/spv.rs
```diff
@@ -332,12 +332,18 @@ impl SpvClient {
 
     /// Report how many block headers we have downloaded to the given path.
```
jcnelson, line 333: Can you add a comment to explain how `get_headers_height()` differs from `get_highest_header_height()`?

testnet/stacks-node/src/run_loop/neon.rs
jcnelson: I think the code will need to do the following:

* Fetch _all_ the bitcoin headers, starting from the genesis header (Stacks 1.0 does this, for example)
* Once all bitcoin headers have been downloaded, proceed to download and process bitcoin blocks starting at `BITCOIN_TESTNET_FIRST_BLOCK_HEIGHT` (or similar -- whatever is in the `BurnchainParameters`), in batches of one reward cycle.  The PoX watchdog should be used to determine when to advance to the next reward cycle.
* Once all reward cycles have been processed, proceed to attempt to mine blocks.

To do this, I think the runloop's setup code needs to have an extra bootstrapping step that will fetch, process, and insert the bitcoin block at `first_block_height` into burnchain DB (note that `first_block_hash` won't be needed, since the node will already have all bitcoin headers).  Then, `BurnchainDB::get_canonical_chain_tip()` will return the bitcoin block at this height as the canonical chain tip, and the `wait_for_sortitions()` method can be instructed to wait until all the block at height at `BITCOIN_TESTNET_FIRST_BLOCK_HEIGHT + burnchain.pox_constants.reward_cycle_length` has been fetched.

I don't think we need to process any prior bitcoin blocks.

src/burnchains/bitcoin/spv.rs
lgalabru: addressed with https://github.com/blockstack/stacks-blockchain/pull/2073/commits/854b21e5639c32065e4d1ed150e4df556da62201

testnet/stacks-node/src/run_loop/neon.rs
lgalabru: The current `next` branch is broken, and when Xenon is probably incomplete without this last step, it's a good step forward and unblocking progress on many front.
@jcnelson do you mind if we merge this as is. and address your feedback in a follow-up PR?

kantai (approved):

PR #2074, opened 2020-11-18
Fix: #2053, apply should not add functions to stack while evaling args

### LINKED ISSUE #2053: Circular reference error on nested function invocation
**Describe the bug**

In Clarity, nested calls to custom function fails with an unexpected _circular reference_ error.

**Steps To Reproduce**

Evaluate in the [Clarity REPL](https://lgalabru.github.io/clarity-repl/):

```clojure
(define-read-only (identity (n int)) n)
(begin (identity (identity 1)))
```
=>
`Runtime Error: Unchecked(CircularReference(["S1G2081040G2081040G2081040G208105NK8PE5.snippet-0:identity"]))`

**Expected behavior**

Should evaluate to the original value.

[Clarity REPL](https://lgalabru.github.io/clarity-repl/):

**Additional context**

Possibly due to inconsistent invocation of custom functions.

kantai:
## Description

Fixes #2053, by waiting until after eval'ing function arguments to put an applied function on the stack. To prevent excessive apply/eval recursion, the call stack needs to track apply-depth in addition to stack depth.

## Type of Change
- Bug fix

Yes, this is a consensus breaking change.

## Testing information

This includes a test that witnesses the original issue #2053, which now passes. The apply-depth change is covered by the existing tests:

```
    vm::tests::contracts::test_arg_stack_depth
    vm::tests::contracts::test_cc_stack_depth
    vm::tests::contracts::test_cc_trait_stack_depth
```

### DIFF
--- a/src/vm/tests/simple_apply_eval.rs
+++ b/src/vm/tests/simple_apply_eval.rs
@@ -912,6 +912,18 @@ fn test_lets() {
 }
 
 #[test]
+fn test_2053_stacked_user_funcs() {
+    let test = "
+(define-read-only (identity (n int)) n)
+(begin (identity (identity 1)))
+";
+
+    let expectation = Value::Int(1);
+
+    assert_eq!(expectation, execute(test));
+}
+
+#[test]
 fn test_asserts() {
     let tests = [
         "(begin (asserts! (is-eq 1 1) (err 0)) (ok 1))",
--- a/src/vm/contexts.rs
+++ b/src/vm/contexts.rs
@@ -123,6 +123,7 @@ pub struct LocalContext<'a> {
 pub struct CallStack {
     stack: Vec<FunctionIdentifier>,
     set: HashSet<FunctionIdentifier>,
+    apply_depth: usize,
 }
 
 pub type StackTrace = Vec<FunctionIdentifier>;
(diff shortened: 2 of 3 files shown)


src/vm/tests/simple_apply_eval.rs
```diff
@@ -911,6 +911,18 @@ fn test_lets() {
...
+fn test_2053_stacked_user_funcs() {
+    let test = "
+(define-read-only (identity (n int)) n)
+(begin (identity (identity 1)))
+";
+
+    let expectation = Value::Int(1);
+
+    assert_eq!(expectation, execute(test));
+}
```
jcnelson, line 924: With the changes to how we measure stack depth, the VM would rightfully prevent us from doing something like `(begin (identity (identity (identity (identity ( ... (identity 1))))))) ... )`, right?

kantai: Yep -- it prevented this before this PR, but the way this was enforced was what was causing this problem. The tests:

```
    vm::tests::contracts::test_arg_stack_depth
    vm::tests::contracts::test_cc_stack_depth
```

demonstrate that prevention.

kantai: In order to preserve that behavior, we needed to add the `apply_depth` value.

jcnelson (approved):
LGTM!

lgalabru (approved):

Issue #1886, opened 2020-09-21
Fees: Convert transaction cost to a dimensionless fill value


jcnelson:
We need to come up with a way to convert a transaction's measured cost into a single dimensionless fill value _F_, between 0.0 and 1.0, that represents the portion of the block's total compute budget consumed.  Once we can do this, then we can assess transaction fees as `fee_rate * F`.

Now, I _think_ that the following properties hold.  Please correct me if I'm wrong:
* The unit costs for `read_count` and `write_count` are _O(1)_ -- doing a single MARF read or write is _O(1)_
* The unit costs for `read_length` and `write_length` are _O(1)_ -- reading or writing a single byte is _O(1)_ [1]
* The unit cost for `runtime` is _O(1)_ -- each atomic unit of work in the Clarity VM is _O(1)_

If so, then we can represent the cost conversion as simply taking the inner product between the transaction cost (a 5-element vector _C_) and the constant conversion vector (a 5-element vector _P_) to get the transaction's total fill _F).  The VM would continue to enforce the total cost of each block in terms of the maximum amount of each resource dimension a block can occupy (represented by `ExecutionCost`), but transaction fee assessment would be calculated with respect to _F_.

[1] Reading/writing data in Sqlite3 is technically _O(n log n)_ for _n_ rows in a table, but this is an implementation artifact that comes with using Sqlite3.  We treat Sqlite3 like an on-disk hash table, so we should assess costs on reading/writing to it as such.

jcnelson:
Estimated time to write and test: 3 days
Estimated time to review the PR: 1 day

lgalabru:
I think this was addressed, @kantai can we close?

kantai:
I don't think we can close -- this was tabled for Stacks 2.0 release, but may be tabled for 2.1.

jcnelson:
I'm not sure we're going to do this after all.

Issue #1695, opened 2020-06-23
Transaction wire format: Fee rate versus fee


kantai:
Currently, the wire format implementation describes a "fee rate", and there's segments of code that treat the supplied fee as a rate, e.g., in `core::mempool`, `estimated_fee = len * fee_rate`, but elsewhere (in the transaction processing and in the mempool _acceptance_ code), the code expects to be supplied a _fee_. We should figure out which of these alternatives we actually want to implement.

jcnelson:
It needs to be `fee_rate`.  We incorrectly use it as a fee right now, but once the block-fill calculation is finished, this field will be multiplied against the runtime-assessed fraction of the block's budget that gets filled.

kantai:
Okay -- that works for me -- I'll leave this issue open until we've implemented the changes.

kantai:
From discussion on #1190 -- we are going to use fee amounts rather than fee rates, so we should update any documentation and references in the code to `fee_amt` rather than `fee_rate`.

lgalabru:
I think this issue was addressed, can we close @kantai ?

kantai:
Yep!

Issue #1190, opened 2019-12-10
Fees: Implement Kth-highest Transaction Fees


jcnelson:
~The Vickrey auction-based block reward implementation from SP1 is broken, and will need to be fixed for SP2 (once we figure out how to ensure the static analysis code remains _outside_ of the consensus code).~

Slight correction -- technically, this isn't a Vickrey auction because bids aren't sealed.  But, we do need to make it so the Kth-highest transaction fee rate sets the transaction fee rate for all transaction in an epoch, subject to the rules below:

jcnelson:
The rules are as follows:

* The leader shares the coinbase with any/all users that burn in support of it
* The leader gets _all_ of the anchored tx fees for the block it produces
* The leader shares the microblock fees 40/60 with the next leader
* The cost per compute unit will be calculated once the full epoch has been evaluated, and the K+1'th highest rate will be used to bill accounts that sent the K-highest transactions (the remaining accounts will not be billed).  **This means that the static analysis framework must _always_ over-estimate the transaction fee**.

diwakergupta:
Not necessarily a Neon blocker, we can push something simple now and add more nuance later.

jcnelson:
* I believe this must be done for Stacks 2.0; failure to do so will create a DDoS vector.

kantai:
My understanding of the original thinking is roughly:

1. Transactions should be charged for their consumed scalar execution units.
1. The rate at which transactions are charged is the Kth highest fee rate in the block (I'm not sure what K is exactly here -- is it the lowest fee rate in the block?)

The above can be hard to implement directly, because accounts must be charged _before_ transaction execution. Other blockchains that implement similar pricing schemes typically do an overcharge and refund. We can support that using fee amount, rather than fee rate:

1. Transactions broadcast a fee _amount_, not a rate. This, however, is effectively a rate, because nodes in the network measure the scalar "execution cost" of each transaction.
1. After executing the transaction, nodes can measure the rate. Miners interested in doing an efficient block fill can use static analysis to aid in this process (to avoid an exponential bin-packing problem in which they need to re-execute the transactions on each attempt).
1. After executing the block, the fee rate is calculated based on the measured fee rates of the transactions in the block.
1. Refunds are issued to accounts with higher fee rates than the block rate.

The only thing unclear to me is how to compute the "block rate".

kantai:
[...]

PR #2076, opened 2020-11-18
Calculate an execution cost scalar

kantai:
This implements #1886, by providing a method to collapse an execution cost into a scalar value, given a block limit. It does this by calculating the dot product of the following vectors:

```
<runtime, read_count, write_count, read_length, write_length> ·
 <1/runtime_limit, 1/read_count_limit, 1/write_count_limit, 1/read_length_limit, 1/write_length_limit>
```

By itself, this PR doesn't make a ton of progress on addressing the outstanding block fee calculation and rewards issues, but I wanted to open this to first get a sense of what would be required for wrapping up the fee calculation and rewards (relevant issues are: #1190 and #1695).

### DIFF
--- a/src/vm/costs/mod.rs
+++ b/src/vm/costs/mod.rs
@@ -317,7 +317,7 @@ impl ExecutionCost {
                     || limit.write_count == 0),
                 "Limit must not have a zero dimension"
             );
-            let dimensional_max = u64::max_value() as u128 / 5;
+            let dimensional_max = u64::max_value() as u128;
             // use u128 math to avoid overflows
             let mut total_score = 0u128;
             for (self_dimension, lim_dimension) in &[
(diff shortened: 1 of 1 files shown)


src/vm/costs/mod.rs
```diff
@@ -302,6 +302,42 @@ impl ExecutionCost {
...
+                (self.write_length, limit.write_length),
+            ] {
+                let dimension_score =
+                    (*self_dimension as u128) * dimensional_max / (*lim_dimension as u128);
+                total_score += dimension_score;
+            }
+            // divide by the number of dimensions.
+            total_score /= 5;
+
+            Some(u64::try_from(total_score).expect("BUG: overflowed u64 calculating fill_scalar"))
```
jcnelson, line 337: This will _never_ fail, right?

kantai: That's right -- one of the unit tests supplies a maximum "stress" on this calculation.

jcnelson (approved):
LGTM, but as you point out, this may be moot if we're not going to implement #1190

lgalabru (approved):
Looks good to me!

kantai:
Closing this PR, as this block fill evaluation is no longer required -- we're using a "first-price auction", meaning that in practice, tx fees are assessed according to the tx_fee on the transaction (rather than using block fill to calculate the second-price).

PR #2082, opened 2020-11-19
Fix: replace loses with wins

friedger:
## Description

User support burns cause payouts if the supported tip wins, not loses.

This PR
* changes this in the section 5. Fair mining pools. The detailed description is correct.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

## Checklist
- [x] Tag 1 of @diwakergupta

### DIFF
--- a/sip/sip-001-burn-election.md
+++ b/sip/sip-001-burn-election.md
@@ -111,7 +111,7 @@ the risk of operating in traditional mining pools.
 In addition to helping lots of small burners mine blocks, fair mining pools
 are also used to give different block leaders a way to hedge their bets on their
 chain tips:  they can burn some cryptocurrency to competing chain tips and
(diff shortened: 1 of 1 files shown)


CharlieC3:
I pushed up a fix for the `rustfmt` GH Action step to `master`. This branch should be **rebased** to pull in the fix via:
```bash
git pull --rebase origin master
# fix any conflicts, then
git push -f
```

friedger:
done

jcnelson (changes requested):
The original wording is correct.  The text is describing how a miner can send user burn supports for _competing_ block-commits, so that they'll get some STX even if the block-commit they sent loses.

friedger:
Thanks for the clarification. The original wording was not clear to me.

Issue #2044, opened 2020-11-09
DB: merge Stacks chainstate DB with headers DB


jcnelson:
We should really store the chainstate and headers into the same database, since it would improve crash consistency, and since we already process blocks by opening simultaneous transactions on both databases.  No need to keep them separate, now that a MARF'ed database is a sqlite database.

jcnelson:
Just an update on this -- it turned out that this was actually necessary to do in order to get poison-microblock handling to work correctly (since the transaction-processing logic needs access to the chainstate to see if a miner's microblock public key has been registered).  Fortunately, it's a relatively painless change.

Issue #1939, opened 2020-10-02
Chainstate: node crash due to Sqlite database lock


jcnelson:
Somewhere we're not setting a busy handler in chainstate processing:

```
DEBUG [1601641208.564] [src/chainstate/stacks/db/transactions.rs:849] [ThreadId(6)] Process transaction 6a5a60313a37e2970e8f512cd9191c9e972ca54b452e4d88dd0166c5d7c9a554 (Coinbase)
ERROR [1601641208.568] [src/vm/database/sqlite.rs:30] [ThreadId(6)] Failed to insert/replace (f94747b71a877b0d2f8e1e9b437c5ca012db4d242ec8921d9a39c7306d5b9a0c0000000000000000,0000000000000000000002bbc5c28700000000000000000000000000000000000000000000000000): SqliteFailure(Error { code: DatabaseBusy, extended_code: 5 },
 Some("database is locked"))                                                                                                                                   
Process abort due to thread panic
   0: stacks_node::main::{{closure}}                                                                                                                                                                                                                                                                                           
             at src/main.rs:51                                                                                                                                                                                                                                                                                                 
   1: std::panicking::rust_panic_with_hook                                                                                                                                                                                                                                                                                     
   2: std::panicking::begin_panic                                                                                                                                                                                                                                                                                              
             at /home/buildozer/aports/community/rust/src/rustc-1.44.0-src/src/libstd/panicking.rs:438                                                                                                                                                                                                                         
   3: blockstack_lib::vm::database::sqlite::sqlite_put                                                                                                                                                                                                                                                                         
             at /home/jude/Desktop/research/git/blockstack-ng.feat-1754/src/vm/database/sqlite.rs:31                                                                                                                                                                                                                           
   4: blockstack_lib::vm::database::sqlite::SqliteConnection::put                                                                                                                                                                                                                                                              
             at /home/jude/Desktop/research/git/blockstack-ng.feat-1754/src/vm/database/sqlite.rs:62                                                                                                                                                                                                                           
   5: <blockstack_lib::vm::database::marf::MarfedKV as blockstack_lib::vm::database::marf::ClarityBackingStore>::put_all                                                                                                                                                                                                       
             at /home/jude/Desktop/research/git/blockstack-ng.feat-1754/src/vm/database/marf.rs:509                                                                                                                                                                                                                            
   6: blockstack_lib::vm::database::key_value_wrapper::RollbackWrapper::commit                                                                                 
[...]

Issue #1935, opened 2020-10-01
SIP: remove tunable proofs from SIP-001 and SIP-005


jcnelson:
This is no longer on the development roadmap, now that we have PoX.

Issue #1878, opened 2020-09-15
VM: mutation methods in STXBalance considered harmful


jcnelson:
Per https://github.com/blockstack/stacks-blockchain/pull/1850#discussion_r485831244, I'm of the opinion that `STXBalance` should not permit mutations.  This is because the only reasonable case where we would want to mutate a `STXBalance` instance is when we want to store the mutations.  It would be a less error-prone design to ensure that both `STXBalance` mutations and actual DB mutations occur in the same method call, thereby rendering it impossible to accidentally mutate a `STXBalance` without committing it to storage.

jcnelson:
So, I got bitten by this exact problem when working on microblock fork-handling.  I'm going to go ahead and submit a fix for it as part of the microblock fork PR.

PR #2085, opened 2020-11-21
Microblock fork-handling and basic mining

jcnelson:
## Description

This PR implements microblock fork-handling and basic microblock mining -- a deceptively simple problem that had a lot of sub-problems that were not initially obvious.  At a high-level, this PR implements the following set of inter-related changes:

* It refactors the chainstate storage and indexing logic for microblocks, so that microblock streams are indexed by the _tail hash_.  That is, the last hash of a stream of microblocks uniquely identifies a confirmed microblock stream.  Before, the code had used the _head hash_, which was incorrect since it could not handle the cases where either (a) the microblock stream had been forked by the miner and each fork had been confirmed by a separate anchored block, or (b) the microblock stream had multiple anchored block descendants.  This change was an expected change.

* It updates the block inventory and download logic to use the new microblock indexing system.  This change was an expected change, and is tightly coupled to the above.

* It implements the `PoisonMicroblock` transaction variant.  If anyone detects that a miner has produced a microblock fork, they can sign a transaction that contains both signed headers at the same sequence number.  If this transaction is (a) confirmed before the offending miner's coinbase has matured, and (b) refers to the _earliest_ such fork in the stream by the time the reward would have matured, then the miner's coinbase is burnt and a 5% commission fee is instead granted to the address of the account that reported the transaction.  This change was an expected change, and was the main goal of this PR.

* It implements a rudimentary microblock miner in the testnet node, for purposes of implementing end-to-end testing and for testing on the public testnet.  The miner is implemented as an extension to a `StacksChainState`'s unconfirmed state, which represents the state of all unconfirmed microblocks applied.  The act of mining a microblock is the act of carrying out the following steps:
   1. refreshing the unconfirmed state trie
   2. finding unrepresented but minable transactions in the mempool
   3. packaging them into a microblock
   4. applying the microblock to the unconfirmed state trie
   5. saving the unconfirmed state trie
   6. relaying the microblock

   This was an expected change, and was necessary in order to adequately test microblock streams in the wild -- something we will want to have up-and-running as soon as possible.

* It modifies the anchored block mining logic to detect microblock forks in the microblock stream it confirms, and to automatically submit a `PoisonMicroblock` transaction to the local mempool to claim the commission for discovering the fork.  This was an expected change -- we already had code that could detect this, but it needed to be refactored due to the new way of indexing microblock streams.

* In order to correctly process a `PoisonMicroblock` transaction, this PR merges the Stacks block and headers databases into a single database (#2044), so that a single (database) transaction can both validate the `PoisonMicroblock` against the chainstate, insert/update an internal "report" structure to keep track of all reported forks for this particular stream, and pay the commission to the right reporter.  I did not anticipate needing to do this, but it was necessary in order to carry out the record-keeping within the context of the Clarity transaction connection.

* In order to get the miner to produce a valid unconfirmed microblock stream, the p2p thread in the testnet node has been refactored to assume the additional responsibilities of (a) validating transactions for mempool admission against both the confirmed and unconfirmed chainstate, (b) recalculating the unconfirmed state view, and (c) mining the next microblock.  The reason for these changes is that the p2p thread is the _only_ thread that can maintain a coherent view of the unconfirmed state trie -- it already needs to do so for servicing queries against it.  Other threads can neither read nor write the unconfirmed state trie directly -- namely, the `cur_block_id` for the unconfirmed trie changes each time the trie is saved, meaning that the thread that saves the trie would invalidate the `cur_block_id` view for all other threads trying to view it.  Therefore, any code paths that needs access to the unconfirmed state trie must either run in the same thread as the p2p thread, or must implement some kind of IPC synchronization to query it.  I chose to do the former, since the design is much simpler (happy to revisit it though).  This was also an unexpected change -- I was hoping this problem wouldn't be an issue, and yet it was blocking me from making progress (and whose fix would only make logical sense within the context of this refactoring).

* Related to threads sharing state leading to problems, this PR implements a partial fix for #1939, whereby a node can encounter an intermittent and hard-to-reproduce runtime panic due to the Clarity DB's sidestore database being locked.  I will admit up front:  this is a **partial, kludgy, temporary fix that I am not proud of, and will need to be reverted once a more satisfactory solution is available**.  Issue #1939 should remain open.  The only reason I attempted a partial fix is because I was able to reliably trigger the panic while testing the microblock miner with the existing integration tests: if the unconfirmed state-processing logic (or microblock miner, which runs in the _p2p thread_) encounters an expensive transaction that consumes the whole block budget (something we test for), it can easily lock the Clarity side-store database while the _chain coordinator thread_ races ahead to process new Stacks blocks.  One thread will hold the lock, and the other thread will panic.  My fix implements a mutex guard that enables the p2p thread to block the chains coordinator thread from processing blocks, and enables the chains coordinator thread to prevent the p2p thread from considering a new microblock.  The chains coordinator will wait for the p2p thread to finish up, and the p2p thread will simply skip mining a microblock if it's blocked.  Getting this temporary fix to work was an unexpected blocker -- our integration tests _would not pass_ without this, and because triggering the race condition that leads to the panic happens naturally as part of producing microblocks, I had to ship a fix immediately.

* In the process of implementing and testing this code, I had been bitten by the problem described in #1878.  Because it is relatively straight-forward to do, this PR includes a harm-reducing change to the way in which STX balances are represented.  Namely, the code now guarantees that at most one balance mutation may be in progress within the context of a Clarity DB instance (the `STXBalance` structure is no longer directly mutated), and the code hides the context required to carry out balance mutations (i.e. the burn block height is hidden from the developer when the mutation begins).  This was an expected change that would have had to have been done eventually, and the opportunity presented itself here (since the fix made it so I would at least stop shooting myself in the foot while testing).

* The PR updates the SIPs to describe the refactoring change, and it removes all mention of "tunable proofs" in the process (#1935).  SIP-005 already describes the poison-microblock behavior above.

* The PR removes the miner payment cache, and removes coinbase sharing and smoothing.  This was a target of opportunity because simply addressing this now was _faster_ and required _less code_ than trying to keep it around during the headers/blocks database merger, only to be deleted later.

* This PR implements #2126 at @agraebe's request, which will be imminently necessary for the explorer and wallets to track transactions as they get confirmed by a microblock stream.

## Type of Change
- [X] New feature
- [X] Bug fix
- [X] API reference/documentation update
- [X] Other

Yes:
* This is a consensus-breaking change -- support for poison-microblock transactions has been added.
* The block inventory messages now reference microblock streams _confirmed_ by blocks, not produced by them.
* The block-download logic serves confirmed microblocks in _reverse_ order, because it now works by walking the microblock stream fork from its tail to its head.

Included.

## Testing information

I still have to implement an adequate `net-test` for microblocks, and possibly extend the `net-test` tooling to be able to query microblocks.  I won't merge until I have this ready.

I don't know yet if it will be necessary to implement an unconfirmed microblock anti-entropy protocol as part of adequately testing this out.  It may be worth waiting on this PR until this is included; otherwise, mining unconfirmed microblocks may prove needlessly difficult on the public testnet.


src/chainstate/stacks/boot/mod.rs
```diff
@@ -1398,7 +1398,7 @@ pub mod test {
 
                 if cur_reward_cycle >= alice_reward_cycle {
                     // this will grow as more miner rewards are unlocked, so be wary
-                    if tenure_id >= (MINER_REWARD_MATURITY + MINER_REWARD_WINDOW + 1) as usize {
+                    if tenure_id >= (MINER_REWARD_MATURITY + 1) as usize {
```
lgalabru, line 1401: `if tenure_id > MINER_REWARD_MATURITY as usize` would be more readable

src/chainstate/stacks/boot/mod.rs
```diff
@@ -1596,7 +1596,7 @@ pub mod test {
                     // alice's tokens are locked for only one reward cycle
                     if cur_reward_cycle == alice_reward_cycle {
                         // this will grow as more miner rewards are unlocked, so be wary
-                        if tenure_id >= (MINER_REWARD_MATURITY + MINER_REWARD_WINDOW + 1) as usize {
+                        if tenure_id >= (MINER_REWARD_MATURITY + 1) as usize {
```
lgalabru, line 1599: Same here

src/chainstate/stacks/boot/mod.rs
```diff
@@ -1863,7 +1863,7 @@ pub mod test {
 
                 if cur_reward_cycle >= first_reward_cycle {
                     // this will grow as more miner rewards are unlocked, so be wary
-                    if tenure_id >= (MINER_REWARD_MATURITY + MINER_REWARD_WINDOW + 1) as usize {
+                    if tenure_id >= (MINER_REWARD_MATURITY + 1) as usize {
```
lgalabru, line 1866: Same here

src/chainstate/stacks/boot/mod.rs
```diff
@@ -2261,7 +2261,7 @@ pub mod test {
 
                 if cur_reward_cycle >= alice_reward_cycle {
                     // this will grow as more miner rewards are unlocked, so be wary
-                    if tenure_id >= (MINER_REWARD_MATURITY + MINER_REWARD_WINDOW + 1) as usize {
+                    if tenure_id >= (MINER_REWARD_MATURITY + 1) as usize {
```
lgalabru, line 2264: Same here

src/chainstate/stacks/boot/mod.rs
```diff
@@ -2552,7 +2552,7 @@ pub mod test {
             eprintln!("\ntenure: {}\nreward cycle: {}\nmin-uSTX: {}\naddrs: {:?}\ntotal_liquid_ustx: {}\ntotal-stacked: {}\n", tenure_id, cur_reward_cycle, min_ustx, &reward_addrs, total_liquid_ustx, total_stacked);
 
             // this will grow as more miner rewards are unlocked, so be wary
-            if tenure_id >= (MINER_REWARD_MATURITY + MINER_REWARD_WINDOW + 1) as usize {
+            if tenure_id >= (MINER_REWARD_MATURITY + 1) as usize {
```
lgalabru, line 2555: Same here

src/chainstate/stacks/boot/mod.rs
```diff
@@ -3460,7 +3460,7 @@ pub mod test {
 
                 if cur_reward_cycle >= alice_reward_cycle {
                     // this will grow as more miner rewards are unlocked, so be wary
-                    if tenure_id >= (MINER_REWARD_MATURITY + MINER_REWARD_WINDOW + 1) as usize {
+                    if tenure_id >= (MINER_REWARD_MATURITY + 1) as usize {
```
lgalabru, line 3463: Same here

src/chainstate/stacks/db/accounts.rs
```diff
@@ -407,8 +380,23 @@ impl StacksChainState {
...
     }
 
-    /// Get the scheduled miner rewards in a particular Stacks fork at a particular height
-    pub fn get_scheduled_block_rewards_in_fork<'a>(
+    /// Find the reported poison-microblock data for this block
+    /// Returns None if there are no forks.
+    pub fn get_poison_microblock_report<T: ClarityConnection>(
+        clarity_tx: &mut T,
+        height: u64,
+    ) -> Result<Option<(StacksAddress, u16)>, Error> {
```
lgalabru, line 388: Is the Poison Microblock report subject to be augmented? Should we introduce a dedicated struct?

src/chainstate/stacks/db/accounts.rs
```diff
@@ -436,17 +424,17 @@ impl StacksChainState {
...
+        tx: &mut StacksDBTx<'a>,
+        tip: &StacksHeaderInfo,
     ) -> Result<Vec<MinerPaymentSchedule>, Error> {
-        let qry =
-            "SELECT * FROM payments WHERE index_block_hash = ?1 ORDER BY vtxindex ASC".to_string();
-        let args: &[&dyn ToSql] = &[index_block_hash];
-        let rows =
-            query_rows::<MinerPaymentSchedule, _>(conn, &qry, args).map_err(Error::DBError)?;
-        Ok(rows)
+        if tip.block_height < MINER_REWARD_MATURITY {
```
lgalabru, line 432: `<=` ? Not sure, but I don't think that Block 0 is a rewarding block

src/chainstate/stacks/db/accounts.rs
```diff
@@ -555,178 +521,143 @@ impl StacksChainState {
...
-            miner
-                .coinbase
-                .checked_mul(this_burn_total as u128)
-                .expect("FATAL: STX coinbase reward overflow")
-                / (burn_total as u128)
+            // no poison microblock reported
+            (participant.address, coinbase_reward)
         };
 
         // TODO: missing transaction fee calculation
```
lgalabru, line 568: We might be able to mark this todo as resolved

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -35,7 +35,7 @@ fn neon_integration_test_conf() -> (Config, StacksAddress) {
     let keychain = Keychain::default(conf.node.seed.clone());
 
     conf.node.miner = true;
-    conf.node.wait_time_for_microblocks = 500;
+    conf.node.wait_time_for_microblocks = 6000;
```
kantai, line 38: Is this only needed for the microblock integration test? If so, can you just set this wait time in that test?

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -447,8 +447,7 @@ impl EventDispatcher {
                         json!({
                             "recipient": reward.address.to_string(),
                             "coinbase_amount": reward.coinbase.to_string(),
-                            "tx_fees_anchored_shared": reward.tx_fees_anchored_shared.to_string(),
-                            "tx_fees_anchored_exclusive": reward.tx_fees_anchored_exclusive.to_string(),
+                            "tx_fees_anchored": reward.tx_fees_anchored.to_string(),
```
kantai, line 450: @zone117x -- this is a change to one of the event types (a simplification because there's never any "shared" anchored tx fees.

@jcnelson -- can you update the example in the `docs/event-dispatcher.md` file?

src/vm/functions/special.rs
```diff
@@ -131,3 +139,173 @@ pub fn handle_contract_call_special_cases(
...
+
+/// Process a poison-microblock transaction.  This isn't something that can happen from a Clarity
+/// program, but it's "special" because it does a non-trivial amount of VM work that needs to be
+/// accounted for in the encapsulating Environment.
+/// Returns a Value that represents the miner slashed:
+/// * contains the block height of the block with the slashed microblock public key hash
+/// * contains the microblock public key hash
+/// * contains the sender that reported the poison-microblock
+/// * contains the sequence number at which the fork occured
+pub fn handle_poison_microblock(
```
kantai, line 182: I'm not sure if this file/module is the best place for this function. The `vm::functions` module is meant to just be where the Clarity native functions are defined -- this probably makes more sense somewhere in the `chainstate::stacks` module, even though it operates on a lot of Clarity structures.

src/net/p2p.rs
```diff
@@ -3986,6 +4029,88 @@ impl PeerNetwork {
...
+        tx: StacksTransaction,
+    ) -> bool {
+        let txid = tx.txid();
+        if mempool.has_tx(&txid) {
+            debug!("Already have tx {}", txid);
+            return false;
+        }
+
+        if let Err(e) = mempool.submit(chainstate, consensus_hash, block_hash, tx) {
+            info!("Reject transaction {}: {:?}", txid, &e);
```
kantai, line 4069: Can you use the key-value format for `info` logs? It'll make ingesting logs a lot easier for different frameworks (not just kibana, but our testground harness also will process JSON logs).

src/vm/errors.rs
```diff
@@ -60,6 +60,7 @@ pub enum InterpreterError {
     FailureConstructingTupleWithType,
     FailureConstructingListWithType,
     InsufficientBalance,
+    InvalidPoisonMicroblockTransaction(String),
```
kantai, line 63: Related to my suggestion to move the poison microblock evaluation out of the `vm::functions` module, it seems like this shouldn't be a possible interpreter error.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -1069,6 +982,19 @@ impl StacksChainState {
...
+    /// Load up a preprocessed block from the staging DB, regardless of its processed status.
+    /// Do not load the associated block.
+    pub fn load_staging_block_info(
+        block_conn: &DBConn,
+        index_block_hash: &StacksBlockId,
+    ) -> Result<Option<StagingBlock>, Error> {
+        let sql = "SELECT * FROM staging_blocks WHERE index_block_hash = ?1 AND orphaned = 0";
+        let args: &[&dyn ToSql] = &[&index_block_hash];
+        let mut rows =
+            query_rows::<StagingBlock, _>(block_conn, sql, args).map_err(Error::DBError)?;
```
kantai, line 994: This should use one of the `query_row` functions -- it avoids allocating an uneccessary `Vec`

src/chainstate/stacks/db/blocks.rs
```diff
@@ -1130,169 +1056,337 @@ impl StacksChainState {
...
+                    consensus_hash,
+                    block_hash,
+                )? {
+                    Some(block_header) => block_header,
+                    None => {
+                        // parent isn't available
+                        return Ok(None);
+                    }
+                };
+                header.microblock_pubkey_hash.clone()
```
kantai, line 1085: Is this clone necessary?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -1130,169 +1056,337 @@ impl StacksChainState {
...
+        let sql = "SELECT * FROM staging_microblocks WHERE index_block_hash = ?1 AND microblock_hash = ?2 AND orphaned = 0 LIMIT 1".to_string();
+        let args: &[&dyn ToSql] = &[&parent_index_block_hash, &microblock_hash];
         let mut rows =
             query_rows::<StagingMicroblock, _>(blocks_conn, &sql, args).map_err(Error::DBError)?;
-        let len = rows.len();
-        match len {
-            0 => Ok(None),
-            1 => {
+
+        Ok(rows.pop())
```
kantai, line 1105: Same here -- the `query_row` function has the same semantics as just calling `pop()` on the SQL query result, but it avoids the vec.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -347,41 +350,83 @@ impl StagingMicroblock {
...
     pub fn new_block(index_block_hash: StacksBlockId) -> BlockStreamData {
         BlockStreamData {
-            block_hash: index_block_hash,
+            index_block_hash: index_block_hash,
             rowid: None,
             offset: 0,
             total_bytes: 0,
+
             is_microblock: false,
+            microblock_hash: BlockHeaderHash([0u8; 32]),
```
lgalabru, line 359: Should we introduce a `MicroblockTailHash`?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -1130,169 +1056,337 @@ impl StacksChainState {
...
+                &parent_consensus_hash,
+                &parent_anchored_block_hash,
+                &microblock.block_hash(),
+                &microblock.header.prev_block,
+                last_seq.saturating_sub(1)
+            );
+
+            if last_seq < u16::MAX
+                && microblock.header.sequence < u16::MAX
+                && microblock.header.sequence > 0
```
kantai, line 1200: Shouldn't this be true when `microblock.header.sequence == 0` as well?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4130,28 +4262,131 @@ impl StacksChainState {
...
-            match StacksChainState::find_next_staging_block(
-                &mut chainstate_tx.blocks_tx,
-                &blocks_path,
-                &chainstate_tx.headers_tx,
-                sort_tx,
+    /// Get the parent header info for a block we're processing, if it's known.
+    /// The header info will be pulled from the headers DB, so this method only succeeds if the
+    /// parent block has been processed.
+    /// If it's not known, return None.
+    fn get_parent_header_info<'a>(
```
kantai, line 4269: Functions where there's only one lifetime parameter and that lifetime parameter isn't in the return type don't need to specify the lifetime. So this function (and a lot of others in this file) could be specified like:

`fn get_parent_header_info(chainstate_tx: &mut ChainstateTx, next_staging_block: ...`

testnet/stacks-node/src/neon_node.rs
```diff
@@ -287,33 +493,66 @@ fn spawn_peer(
                 );
                 100
             } else {
-                poll_timeout
+                cmp::min(poll_timeout, config.node.microblock_frequency)
+            };
+
+            // Mine microblocks.
+            // NOTE: this has to go *here* because control over who can refresh the unconfirmed
```
kantai, line 500: Once the Clarity database refactor is finished, can the microblock mining occur in the miner/relayer thread? If so, can you add a TODO to this note to update this path?

src/chainstate/stacks/miner.rs
```diff
@@ -77,12 +99,17 @@ impl<'a> StacksMicroblockBuilder<'a> {
         anchor_block_consensus_hash: ConsensusHash,
         chainstate: &'a mut StacksChainState,
         burn_dbconn: &'a dyn BurnStateDB,
-        initial_cost: ExecutionCost,
-        bytes_so_far: u64,
     ) -> Result<StacksMicroblockBuilder<'a>, Error> {
+        let runtime = if let Some(ref unconfirmed_state) = chainstate.unconfirmed_state.as_ref() {
+            MicroblockMinerRuntime::from_unconfirmed_state(unconfirmed_state)
```
kantai, line 104: Style nit: one of the reasons for the `From` trait is to avoid type stuttering, e.g.,

```rust
MicroblockMinerRuntime::from(unconfirmed_state)
```
vs.
```rust
MicroblockMinerRuntime::from_unconfirmed_state(unconfirmed_state)
```

kantai (commented):
This looks great to me! At a high level, this all makes sense and looks good to me (and thanks for the adding the borrow-checked mutations to the StxBalance struct). Apart from some slight tweaks, I think my only major comment would be about moving the poison microblock checking function out of the `vm` module.

src/vm/functions/mod.rs
```diff
@@ -42,8 +42,8 @@ use vm::{eval, Environment, LocalContext};
 
 use address::AddressHashMode;
 use chainstate::stacks::{StacksAddress, C32_ADDRESS_VERSION_TESTNET_SINGLESIG};
-pub use vm::functions::assets::{get_stx_balance_snapshot, stx_transfer_consolidated};
-pub use vm::functions::special::handle_contract_call_special_cases;
+pub use vm::functions::assets::stx_transfer_consolidated;
+pub use vm::functions::special::{handle_contract_call_special_cases, handle_poison_microblock};
```
lgalabru, line 46: Not necessarily the right place for this feedback / question.
From a smart contract point of view, I think it would be interesting to expose the microblock height / sequence number in `get-block-info?`. Thought?

src/chainstate/stacks/db/accounts.rs
jcnelson: No, I don't think so.  This code is used to determine whether or not a coinbase recipient should be slashed in favor of some other address that sent a valid `PoisonMicroblock`.  The production code only ever looks at the `StacksAddress` -- the `u16` sequence number is only ever examined by the transaction test code in order to verify that multiple `PoisonMicroblock` reports in the same stream correctly reward the address who reported the earliest such poison.

jcnelson: I think the code should continue to report the `u16` sequence number, though, because if the miner wants to include a `PoisonMicroblock` transaction to claim the coinbase from an earlier miner, and someone else has already reported the poison, then this miner will need to see if the poison it detected is earlier in the stream (otherwise there's no point to mining the `PoisonMicroblock` transaction).

src/chainstate/stacks/db/accounts.rs
jcnelson: It's `<` because of the code below:

```
let block_height = tip.block_height - MINER_REWARD_MATURITY;
```

`block_height` needs to be at least 1 -- there's no block reward at height 0, since that's the boot block.

src/chainstate/stacks/db/accounts.rs
jcnelson: It's not resolved -- we're not yet including the transaction fees, nor are we calculating the 40/60 microblock stream split (which is waiting on a transaction fee calculation model).

testnet/stacks-node/src/tests/neon_integrations.rs
jcnelson: It's also used in the size-check test, but I can move this change to just these two tests.

testnet/stacks-node/src/event_dispatcher.rs
jcnelson: Sure thing!

src/vm/errors.rs
jcnelson: Sure, but because `handle_poison_microblock()` is needs to return `InterpreterResult` (a consequence of having to run within `OwnedEnvironment::execute_in_env()`), what error type would you recommend?  I could either introduce this error type, or add other more-specific error types that will only get used in this code-path, or try and shoe-horn errors from `vm::errors` into the relevant error paths.  In particular, I need error types to describe somewhat-specific errors:

* Invalid signature validation (for when the poison-microblock payload is corrupt)
* The public key hash hasn't been seen in this fork
* The reward that this poison-microblock transaction is targeting has matured

src/vm/functions/special.rs
jcnelson: I'll go ahead and put it where I had it originally -- in `chainstate::stacks::db::transactions`.

src/vm/errors.rs
jcnelson: I could just use a catch-all `InterpreterError::InterpreterError(String)` if you prefer.

src/chainstate/stacks/db/blocks.rs
jcnelson: nope; will remove

src/chainstate/stacks/db/blocks.rs
jcnelson: can do!

src/chainstate/stacks/db/blocks.rs
jcnelson: That line could be removed, since `sequence` is a `u16`

src/vm/errors.rs
kantai: I'd prefer updating the signature of `execute_in_env` to like:

```rust
    pub fn execute_in_env<F, A, E>(
        &mut self,
        sender: Value,
        f: F,
    ) -> std::result::Result<(A, AssetMap, Vec<StacksTransactionEvent>), E>
    where
        F: FnOnce(&mut Environment) -> std::result::Result<A, E>,
```

That way, `handle_poison_microblock` can return a `chainstate::stacks::Error` (or any other error type) without forcing it to be a `InterpreterError`.

src/chainstate/stacks/db/blocks.rs
jcnelson: Maybe in a separate PR.  This would be a pretty disruptive change, since it would significantly affect the chainstate and the networking code.

testnet/stacks-node/src/neon_node.rs
jcnelson: Yes, I think it can be moved to the relayer once it's safe to have multiple threads writing to the DB.

src/chainstate/stacks/miner.rs
jcnelson: This isn't tractable -- `MicroblockMinerRuntime::from()` would take a `UnconfirmedState` instance, instead of a reference.  In doing so, we'd need to implement `Clone` for a `ClarityInstance`, as well as a `MarfedKV`, a `MARF`, a `SqliteConnection`, etc.  So, I added `from_unconfirmed_state()` which takes a `&UnconfirmedState` in order to avoid this.  Is that alright with you?

src/vm/functions/mod.rs
jcnelson: I'm open to it, but it would be a separate PR.

testnet/stacks-node/src/tests/neon_integrations.rs
jcnelson: Addressed in ee12847

testnet/stacks-node/src/event_dispatcher.rs
jcnelson: Addressed in ee12847

src/vm/functions/special.rs
jcnelson: Addressed in ee12847

src/vm/errors.rs
jcnelson: Addressed in ee12847

src/chainstate/stacks/db/blocks.rs
jcnelson: Addressed in ee12847

src/chainstate/stacks/db/blocks.rs
jcnelson: Addressed in ee12847

src/chainstate/stacks/db/blocks.rs
jcnelson: Addressed in ee12847

src/chainstate/stacks/db/blocks.rs
jcnelson: Addressed in ee12847

src/chainstate/stacks/db/blocks.rs
jcnelson: Addressed in ee12847

testnet/stacks-node/src/neon_node.rs
jcnelson: Addressed in ee12847

src/net/p2p.rs
jcnelson: Addressed in c73b3f5

src/chainstate/stacks/boot/mod.rs
jcnelson: `MINER_REWARD_MATURITY` is a `u64` because it is used with `u64` block-height values in non-test code.  If avoiding `as X` in non-test code requires that unit tests have to use `as X`, then that's a trade-off I'm willing to make.  Also, making a protocol constant a `usize` is risky, because `usize` does not have a guaranteed number of bits in its representation (whereas `u64` does).

src/chainstate/stacks/miner.rs
kantai: While leaving it as is would be fine, you can define a trait `From<&UnconfirmedState>` which would have the same borrow semantics as this function.

jcnelson: :man_facepalming: ah, right, forgot you could do that.  Will use that instead.

src/chainstate/stacks/miner.rs
```diff
@@ -100,39 +127,86 @@ impl<'a> StacksMicroblockBuilder<'a> {
...
+                    header_info.consensus_hash,
+                    header_info.anchored_header.block_hash(),
+                    header_info.block_height,
+                )
+            } else {
+                // unconfirmed state needs to be initialized
+                return Err(Error::NoSuchBlockError)?;
+            };
+
+        let mut clarity_tx = chainstate
```
jcnelson, line 176: Okay, there's an issue here that's blocking merge.

The `Drop` handler for `StacksMicroblockBuilder` does a _block_ rollback of the inner Clarity transaction.  First, I don't think this should happen, since a "block" rollback destroys all of the MARF state for the microblock stream.  The rollback should only drop the transaction state for the last microblock mined.  Second, the block-level rollback invalidates the Clarity instance inside the `UnconfirmedState`, which in turn causes a subsequent query against it via a Clarity TX to panic (since the inner MARF no longer exists).  Both of these issues need to be fixed before microblock mining is at all possible.

jcnelson: Marking this as a draft until this is fixed.

jcnelson:
Okay, this is now ready for review :crossed_fingers:

src/chainstate/stacks/boot/mod.rs
jcnelson: I'm not to worried about this in tests.  `MINER_REWARD_MATURITY` should be whatever type it gets used with in production code (which is `u64`).

src/chainstate/stacks/miner.rs
jcnelson: Addressed

src/chainstate/stacks/miner.rs
jcnelson: This is now fixed.

src/chainstate/stacks/db/unconfirmed.rs
```diff
@@ -755,13 +825,9 @@ mod test {
...
                         .map(|tx| {
-                            MemPoolTxInfo::from_tx(
-                                tx,
-                                0,
-                                consensus_hash.clone(),
-                                stacks_block.block_hash(),
-                                tenure_id as u64,
-                            )
+                            let mut bytes = vec![];
+                            tx.consensus_serialize(&mut bytes).unwrap();
```
kantai, line 829: The `StacksMessageCodec` trait provides a handy `serialize_to_vec` method as a convenience to avoid needing to write this boiler plate.

kantai (approved):
LGTM!

lgalabru (approved):
Looks good to me!

Issue #1805, opened 2020-08-12
Absent PoX anchors should be "forkable"


kantai:
There are a few cases where Stacks blocks should be processable in multiple PoX forks. For example, if a reward cycle begins where nobody Stacked, then all of the block commits of that reward cycle would be _burns_, regardless of whether or not the anchor block has already been processed by the node. However, the node cannot _know_ that this is the case until it has processed that anchor block. This would result in a set of sortitions that would select the same sortition winners in two different PoX forks. It is _vitally_ important that those sortition winners be valid in either PoX fork. If they are not valid in either PoX fork, that would mean that _any_ missing PoX anchor block essentially stalls the blockchain (because the discovery of that anchor block would always invalidate all subsequent blocks).

There are a handful of implementation details and Clarity features that prevent blocks from being valid in multiple PoX forks:

1. The `write_consensus_bytes` implementation for back pointers uses the referred-to MARFTrieID as the consensus bytes for the back pointer. This could instead use the root hash of the referred-to trie. See: https://github.com/blockstack/stacks-blockchain/blob/cb75c8f9a555aff85eafc068909d9d573941775d/src/chainstate/stacks/index/node.rs#L311

2. The `BLOCK_HEIGHT_TO_HASH_MAPPING_KEY` and `BLOCK_HASH_TO_HEIGHT_MAPPING_KEY` key/values stored in each MARF trie. These store the parent "MarfTrieId" in each trie. Because the MarfTrieId is a function of the consensus hash and the Stacks block hash, this mixes the PoX fork identifier into the root hash. These key/value pairs are necessary for a bunch of implementation reasons-- they are how we evaluate `(at-block ...)`, compute the ancestor hash skip list, among other things. However, these probably do _not_ need to be committed to. If we could find a way to track these without storing them in the Stacks block commitment MARF, that would be very helpful.

3. Right now: exposing the `index-block-hash` in Clarity. Eventually: exposing other PoX reward cycle information.

jcnelson:
Regarding (1), I think that, by itself, this change should be fine.  But, please see (2) and (3) below, since they may influence (1).

Regarding (2), could we do the following?
* Change the mapping to store the parent trie root hash instead of the MarfTrieId?  Then, we can still make use of the `BLOCK_HEIGHT_TO_HASH_MAPPING_KEY` and `BLOCK_HASH_TO_HEIGHT_MAPPING_KEY` key/value pairs; it's just that now they would not be dependent on any PoX state.
[...]

PR #2087, opened 2020-11-23
Cost Voting Contract

reedrosenbluth:
Boot contract for voting on cost functions

Issue: https://github.com/blockstack/stacks-blockchain/issues/1894


src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,106 @@
...
+(define-fungible-token cost-vote-token)
+
+;; Proposal counter
+(define-data-var proposal-counter uint 0)
+
+;; Rollback proposal counter
+(define-data-var rollback-proposal-counter uint 0)
+
+;; cost-function proposals
+(define-map proposals
```
kantai, line 16: Can you add another `define-map` for "confirmed" proposals and maps for tracking entries in it?

```scheme
(define-map confirmed-cost-functions
   ((cost-function-id uint))
   ((confirmed-function 
      (optional  
        {  function-contract: principal,
           function-name: (string-ascii 50),
           cost-function-contract: principal,
           cost-function-name: (string-ascii 50),
           confirmed-height: uint })))
(define-map functions-to-confirmed-ids
   ((function-contract principal) (function-name (string-ascii 50))
   ((cost-function-id uint)))
(define-data-var cost-function-max-id uint)
```

These data structures would allow us to iterate over the currently defined cost functions by reading from the map from `(0, cost-function-max-id)`. The reason the "optional" is required is to be able to delete a cost-function in O(1) by storing a `none`. The reason we need the `functions-to-confirmed-ids` map is that we'll want to check whether or not a definition already exists when confirming a proposal, and if so, overwrite rather than add a new entry.

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,106 @@
...
+
+)
+
+;; Withdraw rollback votes
+(define-public (withdraw-rollback-vote (rollback-id uint) (amount uint))
+
+)
+
+;; Withdraw STX after rollback vote is over
+(define-public (withdraw-after-rollback-vote (rollback-id uint))
```
kantai, line 99: As currently written, these method stubs won't pass validation -- can you update the stubs to return like `(err 1)` or something along those lines? That way, we can work on integrating the rest of the cost updating code in the stacks-node with this API while the voting contract itself is being implemented.

reedrosenbluth: yup!

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,208 @@
...
+
+;; cost vote token
+(define-fungible-token cost-vote-token)
+
+;; proposal counters
+(define-data-var proposal-count uint u0)
+(define-data-var confirmed-proposal-count uint u0)
+
+;; max confirmed cost-funciton ID
+(define-data-var cost-function-max-id uint u1000)
```
kantai, line 14: Is `cost-function-max-id` used for anything? If not, can you remove the data-var definition?

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,208 @@
...
+    (if (> burn-block-height expiration-block-height)
+        (begin
+            (unwrap! (as-contract (stx-transfer? cur-principal-votes tx-sender sender)) (err 1))
+            (unwrap! (as-contract (ft-transfer? cost-vote-token cur-principal-votes sender tx-sender)) (err 1))
+            (ok true))
+        (err 1)))
+)
+
+;; Miner veto
+(define-public (veto (proposal-id uint))
```
kantai, line 155: I think you need to add a data-map like `(define-data-map excercised-vetos ((proposol-id uint) (veto-height uint)) ((vetoed bool)))` which you can use to ensure that at any given stacks block height, at most one veto has been issued per proposal.

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,208 @@
...
+            (unwrap! (as-contract (ft-transfer? cost-vote-token amount sender tx-sender)) (err 1))
+            (map-set proposal-votes { proposal-id: proposal-id } { votes: (- cur-votes amount) })
+            (map-set principal-proposal-votes { address: tx-sender, proposal-id: proposal-id }
+                                              { votes: (- cur-principal-votes amount) })
+            (ok true))
+        (err 1)))
+)
+
+;; Withdraw STX after vote is over
+(define-public (withdraw-after-votes (proposal-id uint))
```
kantai, line 137: Is `withdraw-after-votes` necessary? Could a voter just use `withdraw-votes` if we removed the `expiration-block-height` check?

src/chainstate/stacks/boot/cost-voting.clar
reedrosenbluth: I don't think it's currently being used. I think I may have pasted it in from your branch. Is there a reason we might want to have a max-id?

kantai: I don't think it's necessary.

src/chainstate/stacks/boot/cost-voting.clar
reedrosenbluth: I don't think it's necessary. Will make this change. I think we'll want `withdraw-votes` to behave slightly differently if the proposal is expired or not though. If the proposal is expired, we can keep the vote count the same (to preserve a record of the vote)

kantai (commented):
This looks good to me -- I think that there are two changes that should be made before this PR can be approved:

1. Make it so that vetos can't be applied by a miner multiple times in the same block.
2. Get rid of `withdraw-after-votes` and just allow voters to use `withdraw-votes`

kantai (approved):
LGTM!

I also think that you should change the base of the PR to `next` -- this PR won't break any behavior in `next`, so I don't think the `next-costs` branch is really required anymore.

I think we may want to tweak the veto behavior (so that there's something like a "veto period" after the vote is confirmed, i.e., a proposal receives the required number of STX votes, gets "stacks holder confirmed", and then there's a period of ~100 blocks or so, when a veto could be applied by miners. But I think addressing that can be done in another PR -- I'll open an issue for this.

src/chainstate/stacks/boot/contract_tests.rs
```diff
@@ -846,3 +865,283 @@ fn delegation_tests() {
...
+            )
+            .unwrap()
+            .0,
+            Value::Response(ResponseData {
+                committed: true,
+                data: Value::Bool(true).into()
+            })
+        );
+
+        // TODO: add test asserting proposal was confirmed
```
jcnelson, line 1145: Want to address this before merge?

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,205 @@
...
+(define-fungible-token cost-vote-token)
+
+;; proposal counters
+(define-data-var proposal-count uint u0)
+(define-data-var confirmed-proposal-count uint u0)
+
+;; cost-function proposals
+(define-map proposals
+    ((proposal-id uint))
+    ((cost-function-contract principal)
```
jcnelson, line 28: Not a bug, but can you use consistent tuple notation?  I see you use `{ }` below.

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,205 @@
...
+        {  function-contract: principal,
+           function-name: (string-ascii 128),
+           cost-function-contract: principal,
+           cost-function-name: (string-ascii 128),
+           confirmed-height: uint })))
+
+(define-map proposal-confirmed-id ((proposal-id uint)) ((confirmed-id uint)))
+
+(define-map functions-to-confirmed-ids
+   ((function-contract principal) (function-name (string-ascii 50)))
```
jcnelson, line 47: Wait, aren't function names 128 bytes long in the proposal maps above?

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,205 @@
...
+   ((proposal-id uint)))
+
+;; cost-function proposal votes
+(define-map proposal-votes ((proposal-id uint)) ((votes uint)))
+
+;; cost-function proposal vetos
+(define-map proposal-vetos ((proposal-id uint)) ((vetos uint)))
```
jcnelson, line 54: Duplicate map definition.

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,205 @@
...
+        (ok (- (var-get proposal-count) u1))))
+
+;; Vote on a proposal
+(define-public (vote-proposal (proposal-id uint) (amount uint))
+    (let (
+        (cur-votes (default-to u0 (get votes (map-get? proposal-votes { proposal-id: proposal-id }))))
+        (cur-principal-votes (default-to u0 (get votes (map-get? principal-proposal-votes {
+            address: tx-sender,
+            proposal-id: proposal-id }))))
+        (expiration-block-height (get expiration-block-height (unwrap! (map-get? proposals {
```
jcnelson, line 105: Slight optimization: you can load `expiration-block-height` first, and error out with `ERR_NO_SUCH_PROPOSAL` without having to load `cur-votes` and `cur-principal-votes`.

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,205 @@
...
+            proposal-id: proposal-id }) (err ERR_NO_SUCH_PROPOSAL))))
+    )
+
+    (asserts! (> amount u0) (err ERR_AMOUNT_NOT_POSITIVE))
+    (asserts! (< burn-block-height expiration-block-height) (err ERR_PROPOSAL_EXPIRED))
+    (asserts! (is-none (map-get? proposal-confirmed-id { proposal-id: proposal-id }))
+        (err ERR_PROPOSAL_CONFIRMED))
+
+    (unwrap! (stx-transfer? amount tx-sender (as-contract tx-sender)) (err ERR_INSUFFICIENT_FUNDS))
+    (unwrap! (ft-mint? cost-vote-token amount tx-sender) (err ERR_FT_MINT))
```
jcnelson, line 115: The `err` case should be unreachable, because `amount > u0`.

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,205 @@
...
+;; Withdraw votes
+(define-public (withdraw-votes (proposal-id uint) (amount uint))
+    (let (
+        (cur-votes (default-to u0 (get votes (map-get? proposal-votes { proposal-id: proposal-id }))))
+        (cur-principal-votes (default-to u0 (get votes (map-get? principal-proposal-votes {
+            address: tx-sender,
+            proposal-id: proposal-id }))))
+        (expiration-block-height (get expiration-block-height (unwrap! (map-get? proposals {
+            proposal-id: proposal-id }) (err ERR_NO_SUCH_PROPOSAL))))
+        (sender tx-sender)
```
jcnelson, line 132: Slight optimization -- you can remove `(sender tx-sender)` and just use `tx-sender`.

src/chainstate/stacks/boot/cost-voting.clar
kantai: I think that until #2115 is merged, this syntax needs to be used in `define-map`

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,205 @@
...
+;; Miner veto
+(define-public (veto (proposal-id uint))
+    (let (
+        (cur-vetos (default-to u0 (get vetos (map-get? proposal-vetos { proposal-id: proposal-id }))))
+        (expiration-block-height (get expiration-block-height (unwrap! (map-get? proposals {
+            proposal-id: proposal-id }) (err ERR_NO_SUCH_PROPOSAL))))
+        (vetoed (default-to false (get vetoed (map-get? exercised-veto { proposal-id: proposal-id,
+                                                                         veto-height: block-height }))))
+        (last-miner (unwrap! (get-block-info? miner-address (- block-height u1))
+            (err ERR_FETCHING_BLOCK_INFO)))
```
jcnelson, line 157: @kantai I remember a conversation we had in the past about certain PoX-derived state, like the consensus hash, being unsafe to expose to the Clarity VM because it could cause the same block to have two different interpretations (https://github.com/blockstack/stacks-blockchain/issues/1805).  Does this concern still apply here, since a block's miner is derived from PoX state (i.e. which block wins sortition is determined in part from the PoX bitvector)?  I think my fix to #1805 removes this concern, but I just want to make sure.

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,205 @@
...
+    )
+    
+    (asserts! (not vetoed) (err ERR_ALREADY_VETOED))
+    (asserts! (is-eq contract-caller last-miner) (err ERR_NOT_LAST_MINER))
+    (asserts! (> burn-block-height expiration-block-height) (err ERR_PROPOSAL_EXPIRED))
```
jcnelson, line 162: Shouldn't this be `>=`?  You have `(asserts! (< burn-block-height expiration-block-height) (err ERR_PROPOSAL_EXPIRED))` earlier.

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,205 @@
...
+    (asserts! (is-none (map-get? proposal-confirmed-id { proposal-id: proposal-id }))
+        (err ERR_PROPOSAL_CONFIRMED))
+
+    (map-set proposal-vetos { proposal-id: proposal-id } { vetos: (+ u1 cur-vetos) })
+    (map-set exercised-veto { proposal-id: proposal-id, veto-height: block-height }
```
jcnelson, line 167: I'm a little confused why `veto-height` is the _Stacks_ block height, and not the burn block height (which seems to be what you're using to track when votes happen).  Can you briefly explain the rationale?

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,205 @@
...
+        (votes (default-to u0 (get votes (map-get? proposal-votes { proposal-id: proposal-id }))))
+        (vetos (default-to u0 (get vetos (map-get? proposal-vetos { proposal-id: proposal-id }))))
+        (proposal (unwrap! (map-get? proposals { proposal-id: proposal-id }) (err ERR_NO_SUCH_PROPOSAL)))
+        (confirmed-count (var-get confirmed-proposal-count))
+    )
+    (let (
+        (expiration-block-height (get expiration-block-height proposal))
+    )
+    (let (
+        (blocks-since-proposal (- block-height (- expiration-block-height u2016)))
```
jcnelson, line 184: `block-height` is the Stacks height, not the Bitcoin block height.  Is that what you mean here?

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,205 @@
...
+                                (function-name (string-ascii 128))
+                                (cost-function-contract principal)
+                                (cost-function-name (string-ascii 128)))
+    (begin
+        (map-insert proposals { proposal-id: (var-get proposal-count) }
+                              { cost-function-contract: cost-function-contract,
+                                cost-function-name: cost-function-name,
+                                function-contract: function-contract,
+                                function-name: function-name,
+                                expiration-block-height: (+ burn-block-height u2016) })
```
jcnelson, line 93: I'm curious why `u2016` is a hard-coded value?  Do we want to allow proposals to have variable voting periods (IIRC Bitcoin has allowed voting periods for up to a whole year, for example)?  If not, then `u2016` should be a constant in the code, since you use it in multiple places.

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,205 @@
...
+    )
+    (let (
+        (expiration-block-height (get expiration-block-height proposal))
+    )
+    (let (
+        (blocks-since-proposal (- block-height (- expiration-block-height u2016)))
+    )
+
+    (asserts! (< burn-block-height expiration-block-height) (err ERR_PROPOSAL_EXPIRED))
+    (asserts! (>= (/ (* votes u100) stx-liquid-supply) u20) (err ERR_INSUFFICIENT_VOTES))
```
jcnelson, line 188: You might want to make these constants, and document them.  Is this documentation correct?

* "A proposal must voted for by at least 20% of the total liquid STX"
* "A proposal must be vetoed by no more than 20% of the mined blocks since the proposal was created"

src/chainstate/stacks/boot/cost-voting.clar
kantai: This map is tracking whether or not this particular miner has exercised their veto in this block -- so the miner here is identified using the stacks block height. This map is to prevent the same miner from exercising a veto twice in the same block.

src/chainstate/stacks/boot/cost-voting.clar
kantai: I don't think that concern applies to the winning block -- if a block is ever being evaluated, that means that block is the winner -- so data from that block is always safe to expose (e.g., who the miner is and how much they committed).

src/chainstate/stacks/boot/contract_tests.rs
reedrosenbluth: Yup working on this now!

src/chainstate/stacks/boot/cost-voting.clar
reedrosenbluth: Right, I wasn't quite sure how to handle this. Is the best practice to have an error const called `ERR_UNREACHABLE` and use that?

src/chainstate/stacks/boot/cost-voting.clar
lgalabru: Done!

src/chainstate/stacks/boot/cost-voting.clar
jcnelson: That's what we've been doing -- it indicates a pretty severe bug (and may even be considered grounds for triggering a node crash if it occurs in the boot code).

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -149,17 +162,30 @@
...
+    ;; a miner can only veto once per block
     (asserts! (not vetoed) (err ERR_ALREADY_VETOED))
+
+    ;; vetoes can only be cast after the voting period has ended
+    (asserts! (>= burn-block-height begin-block-height) (err ERR_VETO_PERIOD_NOT_STARTED))
+
+    ;; vetoes must be case within the veto period
+    (asserts! (< burn-block-height expiration-block-height) (err ERR_VETO_PERIOD_OVER))
+
+    ;; a miner can only veto if they mined the previous block, proving their hash power
```
jcnelson, line 185: Typo -- miners don't have hash power in PoX/PoB.

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,256 @@
...
+        (err ERR_INSUFFICIENT_VOTES))
+
+    (map-insert vote-confirmed-proposals { proposal-id: proposal-id }
+        { expiration-block-height: (+ VETO_LENGTH expiration-block-height) })
+
+    (ok true)))
+)
+
+;; Confirm proposal hasn't been vetoed
+(define-public (confirm-miners (proposal-id uint))
```
jcnelson, line 223: I'm a little confused about these two methods -- `confirm-votes` and `confirm-miners`.  It seems that somebody (can be anybody) must call both `confirm-votes` and, if it success, proceed to then call `confirm-miners`.  `confirm-votes` can be called as soon as there are enough tokens voted, but no later than 2016 blocks after the proposal was created, right?  Once the 2016 blocks have passed, miners then have a 1008-block window in which they can veto a proposal.  If they do not do so, then at any time after the 2016 + 1008 block window, someone (anyone) can call `confirm-miners` to activate the proposal.  Is my understanding here correct?

src/chainstate/stacks/boot/cost-voting.clar
kantai: You can also use `unwrap-panic` which will cause the whole transaction to abort.

src/chainstate/stacks/boot/cost-voting.clar
kantai: Yeah -- that's the idea. A proposal goes through two phases:

Phase 1: STX holder voting. If the proposal isn't confirmed before expiration block, then the proposal fails. Confirmation requires crossing a threshold of the liquid stx.
Phase 2: Miner vetoing. Once a proposal is stx-voter-confirmed, then miners are able to veto the proposal. Vetos must be applied before the veto period ends. Once the veto period ends, a proposal can be confirmed if fewer than a threshold of vetos were applied.

Splitting this into phases makes it clear to miners when a proposal has gained enough support that it needs to be considered for vetoing. It also limits the amount of time that a miner needs to be issuing veto votes.

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,264 @@
...
+
+    ;; confirmation fails if invoked after proposal has expired
+    (asserts! (< burn-block-height expiration-block-height) (err ERR_PROPOSAL_EXPIRED))
+
+    ;; confirmation fails if the required threshold of votes is not met
+    (asserts! (>= (/ (* votes u100) stx-liquid-supply) REQUIRED_PERCENT_STX_VOTE)
+        (err ERR_INSUFFICIENT_VOTES))
+
+    (map-insert vote-confirmed-proposals { proposal-id: proposal-id }
+        { expiration-block-height: (+ VETO_LENGTH expiration-block-height) })
```
kantai, line 227: I think the expiration-block-height here should be `(+ VETO_LENGTH burn-block-height)`, such that the veto period begins immediately once the proposal is "vote-confirmed".

src/chainstate/stacks/boot/contract_tests.rs
```diff
@@ -846,3 +865,468 @@ fn delegation_tests() {
...
+                .0,
+                Value::Response(ResponseData {
+                    committed: false,
+                    data: Value::Int(9).into()
+                })
+            );
+        })
+    }
+
+    for _ in 0..100 {
```
jcnelson, line 1199: Can you make this bound tighter?  Miners have exactly 1008 blocks to veto.

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,263 @@
...
+;; getter for cost-function proposal votes, for specific principal
+(define-read-only (get-principal-votes (address principal) (proposal-id uint))
+    (get votes (map-get? principal-proposal-votes { address: address, proposal-id: proposal-id })))
+
+;; Propose cost-functions
+(define-public (submit-proposal (function-contract principal)
+                                (function-name (string-ascii 128))
+                                (cost-function-contract principal)
+                                (cost-function-name (string-ascii 128)))
+    (begin
```
jcnelson, line 111: Just want to confirm -- if somehow the chain manages to confirm a cost function that doesn't exist or is malformed (e.g. it does I/O), the cost-assessment logic will fall back to the last valid cost-function for the given function here?

Also, what happens if `function-contract` and `function-name` don't correspond to a real function?  Will the cost function still apply to it even if the function is created later?

src/chainstate/stacks/boot/cost-voting.clar
```diff
@@ -0,0 +1,263 @@
...
+
+    ;; the vote must occur before the expiration
+    (asserts! (< burn-block-height expiration-block-height) (err ERR_PROPOSAL_EXPIRED))
+
+    ;; the proposal must not already be voter confirmed
+    (asserts! (is-none (map-get? vote-confirmed-proposals { proposal-id: proposal-id }))
+        (err ERR_VOTE_ENDED))
+
+    (unwrap! (stx-transfer? amount tx-sender (as-contract tx-sender)) (err ERR_INSUFFICIENT_FUNDS))
+    (unwrap! (ft-mint? cost-vote-token amount tx-sender) (err ERR_UNREACHABLE))
```
jcnelson, line 144: Can you briefly explain why there's a fungible voting token?  I can understand the need to ensure that STX deposits and withdrawals are necessary, but why is a data map insufficient for that?

Also, `unwrap-panic` can be used in place of having a `(err ERR_UNREACHABLE)` if you want to have the whole transaction abort.  A good reason to use an `ERR_UNREACHABLE`-like error is if you want to trigger some kind of error (or panic) in the VM itself, whereby the VM intercepts the result and checks the error message.  Is that going to happen here?

src/chainstate/stacks/boot/cost-voting.clar
kantai: Those questions are answered in #2143 -- invalid proposals are simply never "applied" to the cost tracker. A warning is emitted, and then they aren't considered again.

src/chainstate/stacks/boot/cost-voting.clar
kantai: The FT allows users to apply post-conditions to prevent inadvertent withdrawals.

jcnelson: Also, having a dedicated voting token allows you to prevent surreptitious withdrawals by using the "deny" mode in post-conditions -- unless you intended the voting token to move, any attempt to withdraw your STX from the vote will cause the transaction to abort.

src/chainstate/stacks/boot/cost-voting.clar
jcnelson: Thanks!

jcnelson (approved):
LGTM!  Just make sure the tests are all passing before merge.

kantai (approved):
Great work! This is good to merge once the tests pass!

PR #2089, opened 2020-11-23
Update reward cycle length to 2000.

### LINKED ISSUE #2088: Update reward cycle length to 2000
Currently set to 1000, but should be 2000:

* https://github.com/blockstack/stacks-blockchain/blob/master/src/chainstate/stacks/boot/pox-mainnet.clar#L10
* https://github.com/blockstack/stacks-blockchain/blob/master/src/core/mod.rs#L64

diwakergupta:
Fixes #2088

### DIFF
--- a/src/core/mod.rs
+++ b/src/core/mod.rs
@@ -61,7 +61,7 @@ pub const CHAINSTATE_VERSION: &'static str = "23.0.0.0";
 pub const MICROSTACKS_PER_STACKS: u32 = 1_000_000;
 
 pub const POX_PREPARE_WINDOW_LENGTH: u32 = 240;
-pub const POX_REWARD_CYCLE_LENGTH: u32 = 1000;
+pub const POX_REWARD_CYCLE_LENGTH: u32 = 2000;
 /// The maximum amount that PoX rewards can be scaled by.
 ///  That is, if participation is very low, rewards are:
 ///      POX_MAXIMAL_SCALING x (rewards with 100% participation)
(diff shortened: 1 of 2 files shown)


diwakergupta:
Results from `cargo test` locally -- both failures seem unrelated to the commit:

```
failures:

---- net::poll::test::test_register_deregister stdout ----
thread 'net::poll::test::test_register_deregister' panicked at 'called `Result::unwrap()` on an `Err` value: SocketError', src/net/poll.rs:520:44

---- net::poll::test::test_register_too_many_peers stdout ----
thread 'net::poll::test::test_register_too_many_peers' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }', src/net/poll.rs:349:45


failures:
    net::poll::test::test_register_deregister
    net::poll::test::test_register_too_many_peers

test result: FAILED. 1147 passed; 2 failed; 87 ignored; 0 measured; 0 filtered out
```

kantai (approved):
Thanks for fixing this, LGTM!

jcnelson (approved):
LGTM!

src/core/mod.rs
```diff
@@ -61,7 +61,7 @@ pub const CHAINSTATE_VERSION: &'static str = "23.0.0.0";
 pub const MICROSTACKS_PER_STACKS: u32 = 1_000_000;
 
 pub const POX_PREPARE_WINDOW_LENGTH: u32 = 240;
```
diwakergupta, line 63: @kantai shouldn't this match `PREPARE_CYCLE_LENGTH` in `pox-mainnet.clar` above? Happy to send a separate PR if so.

Issue #2004, opened 2020-10-26
Mining: pay forward coinbases in the absence of a sortition


jcnelson:
If a Bitcoin block has no sortition (at height _N_), then any Stacks block mined in a subsequent sortition that builds off of any Stacks chain tip that existed at the penultimate sortition (at height _N-1_) may claim its coinbase.  This encourages miners to keep mining even if Bitcoin fees are high.

kantai:
The initial mining bonus is a special case of this, with some additional features (the reward isn't given to one block, but spread out):

```
The STX block rewards that are “accumulated” prior to the actual launch of
Stacks 2.0 (the launch block) will be added to some initial number of blocks
(e.g., 10,000 blocks or roughly 10 weeks), in order to ensure added incentive
for the start of mining initially.
```

muneeb-ali:
Linking the forum post here where the supply + bonus was discussed: https://forum.stacks.org/t/pox-consensus-and-stx-future-supply/11232?u=muneeb

kantai:
Closed by #2101

Issue #1888, opened 2020-09-21
Block reward: alternative PoX-optimized coinbase schedule?


jcnelson:
Right now, the coinbase schedule is set at 500 STX for the first 5 years, 400 STX for the next 5 years, and 300 in perpetuity afterwards.  These values were chosen arbitrarily, and not to optimize for a particular expected use-case (in fact, they were chosen before PoX was conceived).

An alternative coinbase schedule suggested by @muneeb-ali could be to do the following:

* 1000 STX for years 0-4
* 500 STX for years 4-8
* 250 STX for years 8-12
* 125 STX in perpetuity 

This would result in more initial liquidity for Stacking, but with fewer overall liquid STX by 2050 than the original proposal.

I'm opening this issue to get community feedback as well as to track its implementation (or non-implementation).  I'm mainly curious to see if it would _hurt_ the chain -- i.e. by giving miners more liquidity with which to discount-mine.

jcnelson:
Estimated time to write and test: 1 day
Estimated time to review the PR: 1 day

diwakergupta:
Corresponding forum post: https://forum.blockstack.org/t/pox-consensus-and-stx-future-supply/11232

kantai:
Closed by #2101

PR #2101, opened 2020-11-30
Coinbase schedule and paying-forward

kantai:
## Description

This PR implements the proposed coinbase schedule in #1888 and the pay-forward of missed sortitions scheme proposed in #2004. This includes the initial block mining bonus behavior, where any burnchain blocks between the first burn block height and the first sortition winner pay their coinbases out to an initial block window (constant set in `core/mod.rs`).

## Type of Change
- [x] New feature

Yes.

No -- documentation on coinbase rewards already reflect this behavior.

## Testing information

An integration test is included in the coordinator::tests module that tests both the initial block mining bonus and the paying forward of missing sortitions.

Many of the existing integration tests needed to be updated with the new coinbase reward schedule, but especially the initial mining bonus.


src/chainstate/stacks/db/blocks.rs
```diff
@@ -4170,6 +4183,22 @@ impl StacksChainState {
...
+            )?
+            .expect("CORRUPTION: failed to load snapshot that elected processed block")
+            .accumulated_coinbase_ustx;
+
+            let coinbase_at_block = StacksChainState::get_coinbase_reward(
+                chain_tip_burn_header_height as u64,
+                burn_dbconn.context.first_block_height,
+            );
+
+            let total_coinbase = coinbase_at_block + accumulated_rewards;
```
jcnelson, line 4200: Maybe do a saturating add here?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -3382,6 +3438,31 @@ impl<'a> SortitionHandleTx<'a> {
             values.push(snapshot.sortition_id.to_hex());
         }
 
+        if let Some(initialize_bonus) = initialize_bonus {
+            // first sortition with a winner, set the initial mining bonus fields
+            keys.push(db_keys::initial_mining_bonus_per_block().into());
+            values.push(initialize_bonus.per_block.to_string());
+
+            let total_reward_remaining = initialize_bonus.total_reward - initialize_bonus.per_block;
```
jcnelson, line 3446: Is there a case where this could be negative?  Should we use a saturating sub to be safe?

jcnelson (approved):
This LGTM.  Thanks @kantai!  Just please address the saturating add/sub comments before merging.

src/core/mod.rs
```diff
@@ -39,6 +39,14 @@ pub const NETWORK_P2P_PORT: u16 = 6265;
 
 pub const MINING_COMMITMENT_WINDOW: u8 = 6;
 
+/// The number of blocks which will share the block bonus
+///   from burn blocks that occurred without a sortition.
+///   (See: https://forum.stacks.org/t/pox-consensus-and-stx-future-supply)
+#[cfg(test)]
+pub const INITIAL_MINING_BONUS_WINDOW: u16 = 10;
+#[cfg(not(test))]
+pub const INITIAL_MINING_BONUS_WINDOW: u16 = 10_000;
```
diwakergupta, line 48: Does it make sense to make this a multiple of the PoX reward cycle length? That way we can ensure that the mining bonus couldn't potentially interfere with Stacking mid-cycle

kantai: Yeah, I think that's probably a good idea -- a related change would need to be made to the way we calculate the PoX reward cycle. Right now, we simply take `(burn_block_height - first_burn_block) % reward_cycle_length`. But the initial mining bonus window starts at `first_burn_block_with_active_mining` (the difference between `first_burn_block` and `first_burn_block_with_active_mining` is what builds up that initial bonus).

src/chainstate/burn/db/sortdb.rs
kantai: This couldn't be negative, as `per_block` is set as just `total_reward / INITIAL_MINING_WINDOW`, but I updated to saturating sub just to be on the safe side in ce026c8a1.

src/chainstate/stacks/db/blocks.rs
kantai: Updated in ce026c8a1

lgalabru (approved):
Looks good to me, thank you @kantai !

src/core/mod.rs
kantai: Opened a new issue for tracking this discussion: #2140.

PR #2105, opened 2020-11-30
[Stacks v1 → v2 Upgrade] Import STX balances & vesting into Stacks 2.0

zone117x:
Closes https://github.com/blockstack/stacks-blockchain/issues/2086

Uses a rust [build script](https://doc.rust-lang.org/cargo/reference/build-scripts.html) to compress the STX balance and vesting contents from `chainstate.txt` into deflate streams. These are embedded in the binary using `include_bytes!(...)`. The binary is ~8MB larger. The compression and embedding is contained in a separate lib to make local dev less painful (rebuilds are not slowed down when making changes to the main src). 

This file contains the all STX balances and vesting entries from Stacks 1.0.
(Note: it also contains on-chain name metadata, but name related data is not handled in this PR).

Combined, there are nearly 400k entries from a Stacks 1.0 export. An `iterator<T>` is used to alleviate boot memory usage issues. This PR includes a sample `chainstate.txt` file with only 10s of entries, which loads instantaneously. When the final Stacks 1.0 migration PR is opened, it will commit the full ~20MB `chainstate.txt` file. 

The initial token balances are emitted as STX mint events in the pseudo-genesis-block. Emitting the token vesting entry calls is still WIP -- may need to be done after PR https://github.com/blockstack/stacks-blockchain/pull/2109.


testnet/stacks-node/src/genesis.rs
```diff
@@ -0,0 +1,98 @@
+use stacks::chainstate::stacks::StacksAddress;
+use stacks::{
+    burnchains::bitcoin::address::BitcoinAddress,
+    chainstate::stacks::db::{AccountBalance, VestingSchedule},
+    util::hash::Sha256Sum,
+};
+
+pub const GENESIS_DATA_BYTES: &str = include_str!("../chainstate.txt");
+pub const GENESIS_DATA_SHA_BYTES: &str = include_str!("../chainstate.txt.sha256");
```
zone117x, line 9: Should this hash be exposed in the `v2/info` endpoint?

src/chainstate/coordinator/mod.rs
```diff
@@ -306,7 +306,15 @@ impl<'a, T: BlockEventDispatcher, U: RewardSetProvider> ChainsCoordinator<'a, T,
     ) -> ChainsCoordinator<'a, T, (), U> {
         let burnchain = burnchain.clone();
 
-        let mut boot_data = ChainStateBootData::new(&burnchain, vec![], None);
```
kantai, line 309: Can you add a constructor like `::new_without_vesting` that could be used in all the places where `ChainStateBootData::new(` used to be invoked?

src/chainstate/stacks/db/mod.rs
```diff
@@ -760,6 +785,92 @@ impl StacksChainState {
...
+
+                                Ok(())
+                            })
+                            .unwrap();
+                    });
+                }
+                println!(
+                    "Done initializing chain with {} vesting schedules",
+                    vesting_schedule_count
+                );
```
lgalabru, line 871: I think we need an extra step here: you should be summing all the vested tokens, and use this value for setting the balance of the lockup contract.

lgalabru: We should also make sure that this amount of tokens is not being minted twice (from the balances + from this routine).

zone117x: Note for myself: the total sum of all eventual vested STX must be allocated to the principal of the lockup contract, because this is where the STX payouts will be transferred from (as opposed to vesting minting STX).

src/chainstate/coordinator/mod.rs
zone117x: Done, restored original usage of the `::new` constructor.

src/chainstate/stacks/db/mod.rs
zone117x: Done!

zone117x:
@kantai do you have any idea about this pox related test failure? Locally, `cargo test --workspace --bin=stacks-node -- --test-threads 1 --ignored` passes every time for me, but fails every time in CI. I'm not sure how the changes could cause this.

kantai:
The bitcoind tests additionally require setting `BITCOIND_TEST=1` to test. It looks like the failure may be due to the burn height calculation changing during genesis initialization -- the "unlock" height for the stack-stx operation appears to have changed.

kantai (approved):
This LGTM! Though I think this may need to be opened against `next`, rather than `master`, because this would change the change consensus (the genesis block is different, so this would cause a chain split).

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -259,7 +251,7 @@ pub struct EventDispatcher {
     mempool_observers_lookup: HashSet<u16>,
     stx_observers_lookup: HashSet<u16>,
     any_event_observers_lookup: HashSet<u16>,
-    boot_receipts: Vec<StacksTransactionReceipt>,
+    boot_receipts: RefCell<Option<Vec<StacksTransactionReceipt>>>,
```
kantai, line 254: Why use a RefCell here?

zone117x: Only using it for mutability, to be able to drop the (very large) array from memory once consumed. Is there a better way to handle this?

kantai: I think you'd get the same behavior from just calling `.take()` on an Option.

kantai: Oh, I see -- you're using the RefCell so that the event dispatcher doesn't need to be invoked mutably. Hmm, I think that's fine then.

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -365,13 +357,15 @@ impl EventDispatcher {
...
-        let boot_receipts = if chain_tip.metadata.block_height == 1 {
-            Some(&self.boot_receipts)
-        } else {
-            None
-        };
+        let mut boot_receipts = vec![];
+        if chain_tip.metadata.block_height == 1 {
+            let mut boot_receipts_result = self.boot_receipts.borrow_mut();
+            if let Some(mut val) = boot_receipts_result.take() {
+                boot_receipts.append(&mut val);
```
kantai, line 364: Append will be an expensive operation for a large genesis block -- because it will be iterating over every entry in the vec.

I think you should be able to do something like this:

```suggestion
 let mut boot_receipts = 
        if chain_tip.metadata.block_height == 1 {
            let mut boot_receipts_result = self.boot_receipts.borrow_mut();
            if let Some(val) = boot_receipts_result.take() {
                val
            } else {
                vec![]
            }
        } else {
           vec![]
        };
```

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -365,13 +357,15 @@ impl EventDispatcher {
...
+            }
+        }
```
kantai, line 366: Append will be an expensive operation for a large genesis block -- because it will be iterating over every entry in the vec.

I think you should be able to do something like this:

```suggestion
 let mut boot_receipts = 
        if chain_tip.metadata.block_height == 1 {
            let mut boot_receipts_result = self.boot_receipts.borrow_mut();
            if let Some(val) = boot_receipts_result.take() {
                val
            } else {
                vec![]
            }
        } else {
           vec![]
        };
```

zone117x: Good suggestion, thanks!

kantai (approved):

src/chainstate/stacks/db/mod.rs
```diff
@@ -752,14 +773,152 @@ impl StacksChainState {
...
+                            .checked_add(balance.amount as u128)
+                            .expect("FATAL: liquid STX overflow");
+                        let mint_event = StacksTransactionEvent::STXEvent(
+                            STXEventType::STXMintEvent(STXMintEventData {
+                                recipient: balance.address.into(),
+                                amount: balance.amount.into(),
+                            }),
+                        );
+                        allocation_events.push(mint_event);
+                    }
```
jcnelson, line 819: Any initial reading on what peak memory usage looks like with all these events loaded?  As written, these all live on the stack, not the heap -- I'd recommend testing this on a synthetic load of however many hundreds of thousands of entries we're likely to encounter (because we _don't_ want a surprising stack overflow come launch day).

src/chainstate/stacks/db/mod.rs
```diff
@@ -752,14 +773,152 @@ impl StacksChainState {
...
+                                        "stx-height".into(),
+                                        Value::UInt(schedule.block_height.into()),
+                                    ),
+                                    ("index".into(), Value::UInt(index.into())),
+                                ])
+                                .unwrap();
+
+                                let value = TupleData::from_data(vec![
+                                    ("owner".into(), Value::Principal(schedule.address.into())),
+                                    ("metadata".into(), Value::buff_from_byte(0)),
```
jcnelson, line 871: If we're not going to pass on the `metadata` from Stacks 1.0, we can drop it from the `lockup` contract altogether.

src/chainstate/stacks/db/mod.rs
jcnelson: @lgalabru this may not be desirable -- this means that `total-liquid-ustx` will reflect all of the still-locked tokens, which in turn makes it more expensive to participate in PoX (even though these tokens aren't liquid).

jcnelson: I think we should be minting the locked tokens on-the-fly, by intercepting calls to the `lockup` contract and doing the minting there (much like how we have special-case logic for PoX token locks/unlocks).

stx-genesis/build.rs
```diff
@@ -0,0 +1,102 @@
...
+        let balances_file_path = Path::new(&out_dir).join("account_balances.gz");
+        let balances_file = File::create(balances_file_path)?;
+        let mut balances_encoder = deflate::Encoder::new(balances_file);
+
+        for line in reader
+            .lines()
+            .map(|line| line.unwrap())
+            .skip_while(|line| !line.eq(&"-----BEGIN STX BALANCES-----"))
+            // skip table header line "address,balance"
+            .skip(2)
```
jcnelson, line 39: Just making sure -- `.skip(2)` will skip `-----BEGIN STX BALANCES-----` and `address,balance`, not `address,balance` and the first actual address and balance, right?

stx-genesis/chainstate.txt
```diff
@@ -0,0 +1,32 @@
...
+32E6hsvSf5TP2TWSBKkccHcDtF5Uk9Mnki,347222222225
+32FCnf9V7CmuP58BNKW6taybVDjpPh9vLx,347222222225
+SM37EFPD9ZVR3YRJE7673MJ3W0T350JM1HVZVCDC3,180555557
+SM28P04DXXDEY3WY02VQJD4TS7VZS00Z6SX54XH5J,356850000
+SM260QHD6ZM2KKPBKZB8PFE5XWP0MHSKTD1B7BHYR,2708333337
+SM1TP7CNY63KQY7DVRDVPNE1X73ND3PT8JX46971V,166667000000
+SM3TJ7J6DR9KFQXQGK9SY7M7AEJT33C8NV639J99C,42311000000
+-----END STX BALANCES-----
+-----BEGIN STX VESTING-----
+address,value,blocks
```
jcnelson, line 17: For the test data, can you add some small-valued heights so there can be unit tests for token unlocking that don't require generating thousands of empty burnchain blocks?

testnet/stacks-node/src/config.rs
```diff
@@ -25,6 +25,7 @@ pub struct ConfigFile {
     pub burnchain: Option<BurnchainConfigFile>,
     pub node: Option<NodeConfigFile>,
     pub mstx_balance: Option<Vec<InitialBalanceFile>>,
+    pub mstx_vesting_schedule: Option<Vec<InitialVestingScheduleFile>>,
```
jcnelson, line 28: Might want to make these `ustx_balance` and `ustx_vesting_schedule` -- these are microSTX, not milliSTX

testnet/stacks-node/src/event_dispatcher.rs
jcnelson: We're not worried about concurrent access here, are we?  Otherwise it might be prudent to use a `Arc` instead.

jcnelson (approved):
LGTM!

jcnelson (changes requested):
I think we need to discuss whether or not all of the STX are minted in the genesis block and dolled out by the `lockup` contract (i.e. what's written now), versus having STX minted on-the-fly when the unlock height comes.  I think it will have a painful impact on PoX if we do the former.

testnet/stacks-node/src/event_dispatcher.rs
zone117x: Looking through the call graphs I don't think this would be an issue, not sure.. @kantai?

src/chainstate/stacks/db/mod.rs
zone117x: This `allocation_events` vector ends up using ~90MB with a production chainstate.txt. 

However, the memory usage is less than ideal after building up the above `ClarityTx` with those repeated `StacksChainState::account_genesis_credit(...)` called. Once the `clarity_tx.connection().as_transaction(...)` callback completes, around ~1.3GB of memory is used. Can something like `clarity_tx.commit_unconfirmed()` be periodically called to "flush" the in-memory state to disk?

src/chainstate/stacks/db/mod.rs
zone117x: 👍 
@jcnelson just to be clear, this `metadata` was originally to match the metadata column in the Stacks 1.0 `accounts` table with addresses that had a `lock_id` in the future. However, these have since all unlocked (since block 657212). The remaining vesting entries are from the `account_vesting` table, which does not contain the equivalent `metadata`. 
Does that make sense?

stx-genesis/build.rs
zone117x: Correct! I'd remove the table header, but I like the readability of a "csv" enclosed in PEM-like headers.

testnet/stacks-node/src/genesis.rs
jcnelson: Works for me

testnet/stacks-node/src/event_dispatcher.rs
jcnelson: Assuming it behaves the same as a `RefCell` (it should), why not make it a `Arc` to be safe?  I don't think access to this data is performance-critical, so we don't need to worry about the tiny extra cost of acquiring/releasing the internal mutex.

src/chainstate/stacks/db/mod.rs
jcnelson: > Can something like clarity_tx.commit_unconfirmed() be periodically called to "flush" the in-memory state to disk?

Not without significantly re-designing the MARF implementation :/  The MARF was designed under the assumption that each block would take a known upper-bound amount of memory, but the fact that the genesis block isn't subject to the normal block RAM usage restrictions kinda blows this assumption up.

It's probably fine if it takes 1.3 GB to boot up -- it's not like we're going to be running this on hardware with less than 2 GB of RAM.  But, we should probably note this in https://github.com/blockstack/stacks-blockchain/issues/1959

src/chainstate/stacks/db/mod.rs
jcnelson: Yup, makes sense.  It's fine to leave as-is as well -- up to you.

zone117x: Looks like @lgalabru has taken care of this in follow-up PR https://github.com/blockstack/stacks-blockchain/pull/2109
(we should try to get this one merged to unblock that one)

stx-genesis/chainstate.txt
zone117x: Related, should the stx-genesis lib use a debug/release (or unit test) conditional to toggle between including a small test file like this, vs the full production file?

Also, @kantai and I discussed including the (as-of-today) production chainstate.txt in this PR, because it does introduce some pain points that seem helpful to get accustomed to. @jcnelson thoughts?

testnet/stacks-node/src/event_dispatcher.rs
zone117x: I gave this a try and can't get the same mutability provided by `RefCell.borrow_mut()`. Any suggestions?

src/chainstate/stacks/db/mod.rs
lgalabru: @jcnelson can we let @zone117x merge this PR as is and let me address your feedback in https://github.com/blockstack/stacks-blockchain/pull/2109, already revisiting the lockup contract?

testnet/stacks-node/src/event_dispatcher.rs
zone117x: Got this to work with an `Arc<Mutex<...>>` (thx @kantai for the guidance). @jcnelson does this look good to you?

stx-genesis/chainstate.txt
jcnelson: Let's go with @kantai's suggestion.

src/chainstate/stacks/db/mod.rs
jcnelson: @lgalabru Can you open an issue to describe the problem first?

testnet/stacks-node/src/event_dispatcher.rs
jcnelson: Almost -- can you replace the `.unwrap()` with an `.expect()` with an error message to indicate that the mutex got corrupted?  I think the panicking crash is the right thing to do either way, but I'd prefer a descriptive error message.

jcnelson (approved):
Almost there -- please change the `.unwrap()` to `.expect()` and make sure there's an open an issue describing how eagerly minting all the locked-up STX at boot-time effectively breaks PoX.  Once these two tasks are done, then it LGTM.

testnet/stacks-node/src/event_dispatcher.rs
zone117x: Done!

src/chainstate/stacks/db/mod.rs
zone117x: Issue created https://github.com/blockstack/stacks-blockchain/issues/2139

stx-genesis/chainstate.txt
zone117x: Issue created https://github.com/blockstack/stacks-blockchain/issues/2137
Will do in a follow up PR

testnet/stacks-node/src/genesis.rs
zone117x: Issue created https://github.com/blockstack/stacks-blockchain/issues/2138
Doing this in a follow up PR because right now it requires a small refactor to remove the cyclical dependency between the blockstack core lib and the lib that contains this hash.

src/chainstate/stacks/db/mod.rs
lgalabru: @zone117x beat me to it with https://github.com/blockstack/stacks-blockchain/issues/2139, thanks!

PR #2109, opened 2020-11-30
Auto-unlock vesting schedules

lgalabru:
This PR is addressing https://github.com/blockstack/stacks-blockchain/issues/1062 and https://github.com/blockstack/stacks-blockchain/issues/2139
I spent some time drafting an implementation with the lockup contract initially in place, but after looking into the v1 chainstate, it turns out pursuing would have lead to an inefficient approach, since a block can perform up to 4426 unlock operations (-> 4426 MARF reads).


lgalabru:
Screen Shot 2020-12-08 at 12 05 43 PM

src/chainstate/stacks/boot/lockup.clar
```diff
@@ -1,145 +1,6 @@
-;; The .lockup boot contract
-;; Error codes
-(define-constant ERR_UNLOCK_UNREACHABLE 255)
+(define-map vesting-schedules uint (list 4426 { recipient: principal, amount: uint }))
```
jcnelson, line 1: We need to avoid using the word "vesting" when describing these locked tokens, since it has legal implications.  The tokens are not being vested -- they are already owned (and owed by us) to their respective address owners.  This code is meant to unlock them.

It was unfortunate that we called it "vesting" in Stacks 1.0, since this was inaccurate.

src/chainstate/stacks/db/mod.rs
```diff
@@ -894,87 +904,51 @@ impl StacksChainState {
...
+                        TupleData::from_data(vec![
+                            (
+                                "recipient".into(),
+                                Value::Principal(schedule.address.into()),
+                            ),
+                            ("amount".into(), Value::UInt(schedule.amount.into())),
+                        ])
+                        .unwrap(),
+                    );
+                    match vesting_schedules_per_block.entry(schedule.block_height) {
```
jcnelson, line 922: Minor style nit -- if you're inserting into a HashMap, you can do it more economically like so:

```
if let Some(ref mut schedules) = vesting_schedules_per_block.get_mut(&schedule.block_height) {
   schedules.push(value);
}
else {
   vesting_schedules_per_block.insert(schedule.block_height, vec![value]);
}
```

This is what we do elsewhere in the codebase.

src/chainstate/stacks/db/mod.rs
```diff
@@ -894,87 +904,51 @@ impl StacksChainState {
...
-                                ])
-                                .unwrap();
-
-                                let value = TupleData::from_data(vec![
-                                    ("owner".into(), Value::Principal(schedule.address.into())),
-                                    ("metadata".into(), Value::buff_from_byte(0)),
-                                    ("unlock-ustx".into(), Value::UInt(schedule.amount.into())),
-                                ])
-                                .unwrap();
+                    clarity
```
jcnelson, line 934: I feel weird about this, now that it's finally clear how `chainstate.txt` is getting used to instantiate the initial balances and token unlock schedules.

We're going through a lot of trouble to generate a `chainstate.txt` file, only to load it up via `get_schedules()` and import the lines into the Clarity VM as entries in a data map.  If the `lockup` data map is the final destination for this data, then why not just generate the file `lockup.clar` so that it has a `(begin ..)` code block that contains all of the Clarity code needed to populate the data map?  Then, there's no need for most of this code -- all this code would need to do is load up the code from the `lockup.clar` text file and instantiate it as a contract.

src/chainstate/stacks/boot/lockup.clar
jcnelson: This applies to Clarity and to the Rust code.

jcnelson (changes requested):
I wish this was clearer sooner, but it strikes me as odd that we're generating a `chainstate.txt` file, shipping it with the node, and spending all this effort to load up and parse it only to use it to manually instantiate a data-map in a stub smart contract.

Instead of having _two_ representations for the token balances and lock state, and shipping what basically amounts to a transpiler in the boot-up code to translate the `chainstate.txt` file into a sequence of map inserts, why not just have the `stx-genesis` program generate the `lockup.clar` code such that, when evaluated by the VM, it populates the locked-up token data map for you?

Also, before merging this, we should not use the word "vesting" to describe locked tokens.

src/chainstate/stacks/db/mod.rs
lgalabru: I like the elegancy of your proposal, but I'm sceptic on the overhead that it would create - the vm would have to be able to ingest a 20 mb contract. It sounds like a bet we can take when everything else is secured?

src/chainstate/stacks/boot/lockup.clar
lgalabru: Will fix!

src/chainstate/stacks/db/mod.rs
jcnelson: Yes, runtime is a major consideration.  Loading the 35kb PoX contract takes ~5 seconds on my laptop.  If this scales up, then a 20mb contract might take about 47 _minutes_ (and who knows how much memory).

Regardless, once we choose how we're going to ingest the data, we can't change it later -- it's consensus-critical since it affects the final root hash of the genesis block.  So let's strive to be happy with whatever decision we make -- we're stuck with it.

Do you have any performance numbers on how long it takes to load a production `chainstate.txt` into the data map?

zone117x: When I run this against a prod chainstate.txt, this vesting code takes ~2 seconds to complete on my machine. There are ~60k lockup entries. 

Loading the 300k liquid balances takes much longer, around 70 seconds (that code path is unrelated to this PR).

jcnelson (approved):
Per standup today, it's probably fine if we continue using `chainstate.txt` as-is.  My main concern was correctness, so if we can just make sure that we can audit the resulting data-map and verify that every entry in `chainstate.txt` is present, we should be good to go :+1: 

Feel free to merge once you switch `vesting` to something more legally appropriate, like `lockup`.

kantai (approved):
This LGTM!

src/chainstate/stacks/boot/lockup.clar
lgalabru: addressed with https://github.com/blockstack/stacks-blockchain/pull/2109/commits/52fc632b8e594b00a579daf9c6a7e2f31c76f6bc (thank you @zone117x)

PR #2112, opened 2020-12-01
Fix: #2097, event dispatcher should not unpack response results

### LINKED ISSUE #2097: [TESTNET BUG] block_txs_payload unwraps Clarity response object
## Describe the bug

The following contract defines a fungible token. It was deployed and a transfer call failed because the user didn't own the asset. However, the transaction result is shown as `(err u1)` instead of `(ok (err u1))`
```
(define-constant owner 'ST3X2W2SH9XQZRHHYJ21KWGTT1N6WX3D48K1NSTPE)
(define-fungible-token connect-token)
(begin (ft-mint? connect-token u100000000 owner))
(define-public (transfer (recipient principal) (amount uint))
  (ok (ft-transfer? connect-token amount tx-sender recipient)))
```

## Steps To Reproduce
1. setup mocknet and stacks-node-api
1. deploy contract to mocknet
1. send transfer tx with post condition
1. use tx id to find tx on stacks-node-api
1. see tx_result is equal to `(err u1)`

## Expected behavior

The tx_result should be `(ok (err u1))`

## Example (25 Nov 2020)
contract:
https://stacks-node-api.blockstack.org/extended/v1/tx/0x312cb2ba8c91183a57e02a2928b325ed18753d0043c50da7dd4ad2aa0675f1b6

transfer tx:
https://stacks-node-api.blockstack.org/extended/v1/tx/0x731b5ba5af6b07bea60a0f5d8552555794955dc41350e47c46d745264b22959f

----

If you think this is eligible for a [bug bounty](https://testnet.blockstack.org/bounties), please check the relevant boxes below:

### Critical, Launch Blocking Bugs
**Consensus critical bugs**
- [ ] Can cause a chain split
- [ ] Can cause an invalid transaction to get mined
- [ ] Can cause an invalid block to get accepted
- [ ] Can cause a node to stall

**State corruption**
- [ ] Can modify a smart contract’s data maps and data vars without a `contract-call?

**Stolen funds**
- [ ] Any address losing STX without a corresponding transfer
- [ ] Modify token balances and NFT ownership in other contracts without a `contract-call?`

**Take control and/or bring network to a halt**
- [ ] Take control and/or bring network to a halt

### Major, Launch Blocking Bugs
**Major bugs**
- [ ] Performance or correctness bugs that don’t rise to P0 level
- [ ] Stress test or DoS attacks that slow things down enough
- [ ] Resource exhaustion
- [ ] Expected functionality doesn’t work in obvious ways (important to be super specific with this wording)

### Minor, Non-launch blocking bugs
**Minor bugs**
- [x] Bugs in non-critical software (CLI, UI, etc) that doesn’t impact critical functionality

kantai:
This fixes #2097 by not "unwrapping" transaction response results. A unit test was added for the specific behavior of #2097 in the `blockstack_lib` (where there wasn't an issue -- the transaction receipt contained the correct result), and an assertion was added to an existing neon_integrations test that ensures that the returned result is actually the whole result, not an unpacked value.

### DIFF
--- a/src/chainstate/stacks/db/transactions.rs
+++ b/src/chainstate/stacks/db/transactions.rs
@@ -3555,6 +3555,114 @@ pub mod test {
         conn.commit_block();
     }
 
+    #[test]
+    fn process_post_conditions_tokens_deny_2097() {
+        let privk_origin = StacksPrivateKey::from_hex(
+            "027682d2f7b05c3801fe4467883ab4cff0568b5e36412b5289e83ea5b519de8a01",
+        )
+        .unwrap();
+        let privk_recipient = StacksPrivateKey::from_hex(
+            "7e3af4db6af6b3c67e2c6c6d7d5983b519f4d9b3a6e00580ae96dcace3bde8bc01",
+        )
+        .unwrap();
+        let auth_origin = TransactionAuth::from_p2pkh(&privk_origin).unwrap();
+        let auth_recv = TransactionAuth::from_p2pkh(&privk_recipient).unwrap();
+        let addr_publisher = auth_origin.origin().address_testnet();
+        let addr_principal = addr_publisher.to_account_principal();
+
+        let contract = "
+(define-constant owner 'ST3X2W2SH9XQZRHHYJ21KWGTT1N6WX3D48K1NSTPE)
+(define-fungible-token connect-token)
+(begin (ft-mint? connect-token u100000000 owner))
+(define-public (transfer (recipient principal) (amount uint))
+  (ok (ft-transfer? connect-token amount tx-sender recipient)))
+"
+        .to_string();
+
+        let contract_name = ContractName::try_from("hello-world").unwrap();
+
+        let recv_addr = StacksAddress::from_public_keys(
+            C32_ADDRESS_VERSION_TESTNET_SINGLESIG,
+            &AddressHashMode::SerializeP2PKH,
+            1,
+            &vec![StacksPublicKey::from_private(&privk_recipient)],
+        )
+        .unwrap();
+        let recv_principal = recv_addr.to_account_principal();
+        let contract_id = QualifiedContractIdentifier::new(
+            StandardPrincipalData::from(addr_publisher.clone()),
+            contract_name.clone(),
+        );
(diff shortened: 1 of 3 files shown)


lgalabru (approved):

src/chainstate/stacks/db/transactions.rs
```diff
@@ -3555,6 +3555,114 @@ pub mod test {
...
+            FungibleConditionCode::SentEq,
+            10,
+        ));
+
+        let mut signer = StacksTransactionSigner::new(&tx_contract_call);
+        signer.sign_origin(&privk_origin).unwrap();
+        let contract_call_tx = signer.get_tx().unwrap();
+
+        let mut chainstate =
+            instantiate_chainstate(false, 0x80000000, "process-post-conditions-tokens-deny");
```
jcnelson, line 3647: Might need to make this test name `process-post-conditions-tokens-deny-2097` so it doesn't conflict with any other on-disk chainstate.

src/chainstate/stacks/db/transactions.rs
```diff
@@ -3555,6 +3555,114 @@ pub mod test {
...
+        );
+
+        // publish contract
+        let _ =
+            StacksChainState::process_transaction(&mut conn, &signed_contract_tx, false).unwrap();
+
+        let (_fee, _) =
+            StacksChainState::process_transaction(&mut conn, &contract_call_tx, false).unwrap();
+
+        conn.commit_block();
```
jcnelson, line 3663: I'm not sure what's being tested here?  Is it just testing that the transaction can be mined?  Or does it also need to test that the contract-call returns the appropriate `Value::Response` structure?

jcnelson (approved):
This LGTM once the tests all pass.  Thanks @kantai!

src/chainstate/stacks/db/transactions.rs
kantai: Resolved in ff936105070b3091fb599533b45206b3e88cd357

src/chainstate/stacks/db/transactions.rs
kantai: Yep -- there were some assertions here previously, added back in ff936105070b3091fb599533b45206b3e88cd357

PR #2113, opened 2020-12-01
Merge `next-costs` into `next`

kantai:
This merges `next-costs` into `next`. The two PRs currently in `next-costs` were previously reviewed -- they introduce:

1. Calculation of the Clarity cost functions using Clarity code.
2. Loading of the Clarity cost functions from the chain state.

I think we should merge this into `next`, because these two PRs work as-is, and the "free" cost-tracker (plus the interception in #2072) will be useful to other work -- in particular @lgalabru's work on the auto-unlock of vesting accounts.


lgalabru (approved):
🚀

jcnelson (approved):
:boom:

PR #2115, opened 2020-12-01
Clarity: enhancing tuples - volume 1

lgalabru:
This PR is addressing:
- https://github.com/blockstack/stacks-blockchain/issues/1337 Remove implicit tuple syntax
- https://github.com/blockstack/stacks-blockchain/issues/1338 Support non-tuple key-values in data maps 

I migrated almost every constructed tuple to the new recommended curly brace notation, it ended up bumping our test coverage for this syntax, and helped uncover some nasty things that could have been done with this syntax: cycles and storing trait references.


jcnelson (approved):
I love it when we delete code!  LGTM, thanks @lgalabru!

kantai (commented):
Awesome! This looks great to me. My only request is that this PR should add some test cases to `type_checker::tests` and `vm::tests` that cover map operations with non-tuple keys and values (e.g., `(define-map kv-store int (list 10 int))`)

jcnelson:
To be clear, before merging, you're going to send a PR to this branch that updates the cost functions for variadic map, tuple-set, etc., right?

lgalabru:
@jcnelson We have a bunch of PR relying on this PR, so I'd rather first merge this PR in `next`, and update separately the other PRs so that they use the new `next` as a target branch.

kantai (approved):
Thanks for adding the additional tests!

README.md
```diff
@@ -99,7 +99,7 @@ cargo run --bin blockstack-cli generate-sk --testnet
 We will interact with the following simple contract `kv-store`. In our examples, we will assume this contract is saved to `./kv-store.clar`:
 
 ```scheme
-(define-map store ((key (string-ascii 32))) ((value (string-ascii 32))))
+(define-map store { key: (string-ascii 32) } { value: (string-ascii 32))
```
reedrosenbluth, line 102: Are you missing a closing curly brace here?

PR #2117, opened 2020-12-02
Clarity: introduce merge function

lgalabru:
This PR is addressing https://github.com/blockstack/stacks-blockchain/issues/1733 by introducing a `(set base-tuple tuple)`.
The initial issue was suggesting `merge`, it could also be `combine`, or `set` (i personally like the symmetry with `get`) - please chime in on the naming :).

Todo left:
- [x] Set cost functions


njordhov:
The verb `set` is used in the mutating functions [map-set](https://docs.blockstack.org/references/language-functions#map-set) and [var-set](https://docs.blockstack.org/references/language-functions#var-set) while `merge` should be a pure function returning a new record.

src/vm/functions/mod.rs
```diff
@@ -232,6 +233,7 @@ pub fn lookup_reserved_functions(name: &str) -> Option<CallableType> {
             DeleteEntry => SpecialFunction("special_delete-entry", &database::special_delete_entry),
             TupleCons => SpecialFunction("special_tuple", &tuples::tuple_cons),
             TupleGet => SpecialFunction("special_get-tuple", &tuples::tuple_get),
+            TupleSet => SpecialFunction("special_set-tuple", &tuples::tuple_set),
```
njordhov, line 236: Both arguments are evaluated, so this should be a regular native function rather than a special.

kantai: Yep -- this can be a regular native function.

src/vm/ast/definition_sorter/mod.rs
```diff
@@ -254,6 +254,14 @@ impl<'a> DefinitionSorter {
                                     }
                                     return Ok(());
                                 }
+                                NativeFunctions::TupleSet => {
```
kantai, line 257: Does this need to be handled with a special case in this match? Or could TupleSet just be handled like any other native function (i.e., fall through to line 270, which just iterates over the expressions)?

kantai (approved):
This looks good to me -- just a couple small comments. One is that I agree with @njordhov that the function can (and should) be implemented as a normal native function. I also agree that it makes more sense to call this `merge` -- I think `set` connotes that it is a mutating method, which it isn't.

src/vm/docs/mod.rs
```diff
@@ -643,6 +643,17 @@ the tuple. If the supplied option is a `(none)` option, get returns `(none)`.",
...
+const TUPLE_SET_API: SpecialAPI = SpecialAPI {
+    input_type: "tuple, tuple",
+    output_type: "tuple",
+    signature: "(set tuple { key1: val1 })",
+    description: "The `set` function returns a new tuple with the combined fields, without mutating the supplied tuples.",
+    example: "(define-map users { id: int } { name: (string-ascii 12), address: (optional principal) })
+(map-insert users { id: 1337 } { name: \"john\", address: none }) ;; Returns true
+(let ((user (unwrap-panic (map-get? users { id: 1337 }))))
+    (set user { address: (some 'SPAXYA5XS51713FDTQ8H94EJ4V579CXMTRNBZKSF) })) ;; Returns (tuple (address (some SPAXYA5XS51713FDTQ8H94EJ4V579CXMTRNBZKSF)) (name \"john\"))"
+};
```
njordhov, line 655: Can this example be simplified to focus more on the documented function rather than map-functionality?

src/vm/ast/definition_sorter/mod.rs
njordhov: #2123

src/vm/types/mod.rs
```diff
@@ -1139,6 +1139,35 @@ impl TupleData {
...
         })
     }
+
+    pub fn deep_merge(base: TupleData, updates: TupleData) -> Result<TupleData> {
+        let mut base = base;
+        for (name, value) in updates.data_map.into_iter() {
+            let (_, existing_value) =
+                base.data_map
+                    .remove_entry(&name)
+                    .ok_or(CheckErrors::NoSuchTupleField(
```
jcnelson, line 1149: This wasn't clear from the documentation:  `(set a b)` only works if all keys in `b` exist in `a`, right?  For example, this should fail to type-check: `(set { a: 1 } { b: 2 })`?

src/vm/types/mod.rs
```diff
@@ -1139,6 +1139,35 @@ impl TupleData {
...
+                        name.to_string(),
+                        base.type_signature.clone(),
+                    ))?;
+
+            let new_value = match existing_value {
```
jcnelson, line 1154: Just to make sure I understand the implications of this code:

* `(set { a: { c: 1 }, b: 2 } { a: 1 })` will fail at runtime, because when processing `a` in `base`, `existing_value` would be a `Value::Tuple` but `value` would be a `Value::Uint(_)`, triggering the `CheckErrors::ExpectedTuple` error.
* `(set { a: 1, b: 2 } { a: { c: 1} })` would succeed, because when processing `a` in `base`, `existing_value` would be a `Value::Uint(_)` and would just fall through this case.  The resulting tuple would be `{ a: { c: 1 }, b: 2 }`.
* `(set { a: { c: 1 }, b: 2 } { a: { d: 2 } })` would succeed, and the tuple values for `a` in both `base` and `updates` would be merged through this inner call to `TupleData::deep_merge()`.  The resulting tuple would be `{ a: { c: 1, d: 2 }, b: 2}`.

Is my understanding of the above correct?  If so, then I think it's kind of surprising.  As a programmer, I would have expected the following behaviors:

* `(set { a: { c: 1 }, b: 2 } { a: 1 })` should succeed, but produce the tuple `{ a: 1, b: 2 }`.
* `(set { a: 1, b: 2 } {a: {c: 1 } })` should behave as described above, and produce `{ a: { c: 1 }, b: 2 }`.
* `(set { a: { c: 1 }, b: 2 } { a: {d: 2} })` should succeed, but produce the tuple `{ a: { d: 2 }, b: 2 }`.

What do you all think?  Also cc @kantai

jcnelson (changes requested):
I need some clarity (no pun intended) in the intended behavior of `set`, as outlined in my comments.

src/vm/types/mod.rs
kantai: Hmm -- yes, I think I agree with this. I think it makes more sense for this function to produce a shallow "union" tuple, rather than trying to do a deep merge.

I'm not so worried about the cost-tracking of a deep merge, which could be implemented as a different function (though, I'm not sure how useful it would be -- in our boot contracts and from the contracts people have mostly produced, tuples tend not to nest deeply, and when they do, it's rare that only a field of a child tuple would be updated) -- the cost of a deep merge is still linear in the size of the two inputs, and the depth of the merge is bounded by the type depth limit.

lgalabru: Sure. I designed this feature based on what I've been experiencing while writing contracts. I think in 100% of cases, I would have use this function for updating a tuple coming from storage.
If you look at this use case, then you never want to update the "base" type signature, and be as terse as possible.

src/vm/types/mod.rs
njordhov: A `merge` function should return the merged properties, resulting in `{ a: 1, b: 2}`.

src/vm/types/mod.rs
jcnelson: > Sure. I designed this feature based on what I've been experiencing while writing contracts. I think in 100% of cases, I would have use this function for updating a tuple coming from storage.
> If you look at this use case, then you never want to update the "base" type signature, and be as terse as possible.

This use-case doesn't invalidate the argument for creating a "union" tuple via `(set ..)`.  A subsequent call to `var-set`, `map-set` or `map-insert` would still need to pass type-check.  Assuming we did something like what I suggest above, then this code snippet would _not_ pass analysis, for example:

```
(define-map foo { a: uint, b : { c : uint } } { d: uint, e: { f: uint } })
(let (
   (test (unwrap-panic (map-get { a: 1, b: { c: 2 } })))
) (
   ;; will fail to type-check, since `(set ..)` has type `{ d: uint, e: { f: uint, g: uint } }`,
   ;; which is incompatible with the value type definition for `foo`
   map-set { a: 1, b: { c: 2 } } (set test { d: 3, e: { g: 4 } }))
)
```

So, having the "union" behavior shouldn't prevent you from safely loading, updating, and storing structured data.

njordhov: @lgalabru A common use case for `merge` in Clarity is to merge the state in a reducer with updated values. See for example the `add-pox-addr-to-ith-reward-cycle` reducer function the the pox contract, which could use a `(merge params ...)` expression to replace the explicit passing of properties  in line 264-267:

https://github.com/blockstack/stacks-blockchain/blob/1ad60a216e53473d6fb103e2e171cfe1003b9883/src/chainstate/stacks/boot/pox.clar#L254-L284

lgalabru: Yes this is exactly my point. You won't be able to use this merge function on nested structures, which I think is limiting.
So in the case of something like this:
```
(define-map users { user-id: uint } { profile: { address: principal, name: (string 32) }, balances: { stx: uint, btc: uint, usd: uint } })
```
If you just want to update the path `balances.stx` then you'll end up enumerating all the siblings with some `get` on the previous version of the structure, which was the main problem we were going after (vs creating some union that could be solved by introducing a very explicit `(union ...)`).

lgalabru: @njordhov I'm not saying that the behavior expected by @jcnelson does not make sense.
I'm just saying that it's not exactly the issue that this issue was supposed to solve - ie **improving the `map-get` / `map-set` experience**.
Yes, I'm sure that there are plenty of use cases for `union`, `intersect`, `difference`, etc, but that should be a different effort.

lgalabru: My proposal would be the following: if we can reach consensus on the fact that a deep merge is what is needed for improving the `map-get`/`map-set` experience, then:
- rename this function to `deep-merge`
- backlog `union` and `merge` functions (which could be a bit different) as non mainnet blockers
Thoughts?

jcnelson: The `(set ..)` behavior in this code is surprising, though.  I can't think of a single programming language where updating structured data happens this way.

It seems like the only benefit of the current implementation is that it makes modifying nested tuples slightly less verbose, at the expense of making it so that constructing new tuples with new types from existing tuples is much more verbose.  But, it does not have any real impact on type safety, because the moment you try to do something with the tuple you create through `(set ..)`, the analysis logic will ensure that the tuple has the appropriate type.  It is not always the case that you want the `(set ..)`-created tuple to have the same type as the `base` tuple.

As for your example, what's wrong with this?

```
;; update balance.stx in user u123
(let (
   (my-user (unwrap-panic map-get? users u123))
   (my-balances (unwrap-panic map-get? my-user balances))
)
(let (
   (my-new-user (set my-user { balances: (set my-balances { stx: u100000 }) }))
   ;; do something with my-new-user
)))
```

While it is true that in your version, you can reduce this by one line to this:

```
;; update balance.stx in user u123
(let (
   (my-user (unwrap-panic map-get? u123))
)
(let (
   (my-new-user (set my-user { balances: { stx: u100000 } }))
   ;; do something with my-new-user
)))
```

It is worth bearing in mind that by having a "union"-like `(set ..)` method, you don't need to implement a separate `union` keyword.  It is also worth bearing in mind that if you wanted to do something that required to build a new type of tuple from an existing tuple, you can't use `(set ..)` to help you -- you'll have to explicitly do so.  Imagine that you wanted to migrate your users to a new map structure.  Having a "union"-like `(set ..)` makes this much more straightforward.  For example, consider adding a new balance item:

```
(define-map users-v2
   { user-id: uint }
   { profile:
      { address: principal, name: (string 32) },
     balances: { stx: uint, btc: uint, usd: uint, eth: uint } })

;; migrate user u123 from v1 to v2
(let (
  (my-user (unwrap-panic map-get? u123))
  (my-balances (unwrap-panic map-get? my-user balances))
)
(let (
  (my-new-user (set my-user { balances: (set my-balances { eth: u0 }) }))
)))
```

Whereas with your implementation of `(set ..)`, you would need to explicitly construct the `users-v2` entry since `(set ..)` won't let you construct a tuple of a different type from `base`:

```
(define-map users-v2
   { user-id: uint }
   { profile:
      { address: principal, name: (string 32) },
     balances: { stx: uint, btc: uint, usd: uint, eth: uint } })

;; migrate user u123 from v1 to v2
(let (
  (my-user (unwrap-panic map-get? u123))
  (my-user-balances (unwrap-panic map-get? my-user balances))
)
(let (
  (my-new-user {
    profile: (get profile my-user)
    balances: {
       stx: (get my-user-balances stx)
       btc: (get my-user-balances btc)
       usd: (get my-user-balances usd)
       eth: u0
    }
  })
)))
```

Does that change your perspective?

njordhov: > It seems like the only benefit of the current implementation is that it makes modifying nested tuples slightly less verbose

While I'd expect `merge` to return a shallow union of properties, one issue to consider for Clarity is that due to the lack of recursion, a deep merge is cumbersome to implement, so it could make sense to down the road add a `merge-deep` function.

kantai: I think that a shallow merge does a pretty good job handling the use case you're thinking about @lgalabru: looking at the boot contracts, for example, the only case of nested tuples in the defined maps are the address tuples, which you wouldn't want to update through a deep merge anyways.

lgalabru: ok, I'll fold :trollface:

lgalabru: I'm working on re-implementing this feature, but one problem that we would now be facing, is how do we handle optionals without turning this`merge` feature into a footgun:

1) `(merge { a: (some { x: 0, y: 1 }), b: 2, c: 3 } { a: none })`
2) `(merge { a: 1, b: 2, c: 3 } { a: none })`
3) `(merge { b: 2, c: 3 } { a: none })`

Happy to get a new round of thoughts on this.

njordhov: > how do we handle optionals

@lgalabru Option properties should be handled like any other type. Consider that it is is just an enum with two variants (Some and None) just like a Response is an enum with two variants (Ok and Err). Despite the clarity type name, the property is not _optional_ (https://github.com/clarity-lang/reference/issues/22).

kantai: In all cases, I think the output type would be the same:

```
    (merge { a: (some { x: 0, y: 1 }), b: 2, c: 3 } { a: none }) -> { a: (optional NoType), b: int, c: int }
    (merge { a: 1, b: 2, c: 3 } { a: none }) -> { a: (optional NoType), b: int, c: int }
    (merge { b: 2, c: 3 } { a: none }) -> { a: (optional NoType), b: int, c: int }
```

This should be the natural output from a type checker that just set the output tuple type to be the "shallow" union of the two input tuples. The `NoType` here won't present difficulties to the type checker unless someone tried to `unwrap!` the `optional`, which is a type error.

lgalabru: Gotcha, I was not familiar with the type admission rules of `NoType` / `UnknownType`, and it looks like this works as expected: 

```
        (define-map users uint { address: principal, name: (optional (string-ascii 32)) })
        (let
            ((user (unwrap-panic (map-get? users u0))))
            (map-set users u0 (merge user { name: none })))
```

Approach revisited in https://github.com/blockstack/stacks-blockchain/pull/2117/commits/8b5c1f07cf34fa5ecd8d5f9843a7d9323679f55d.

src/vm/functions/mod.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2117/commits/6f976eb11225e8f4eb4bb4ed4ba96b7c20c3983f

src/vm/ast/definition_sorter/mod.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2117/commits/a1a30823ad531d20310d347a63f1c710ce190742

jcnelson (approved):
LGTM!  Thank you for addressing my feedback!

src/vm/types/signatures.rs
```diff
@@ -64,7 +64,7 @@ impl AssetIdentifier {
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct TupleTypeSignature {
-    type_map: BTreeMap<ClarityName, TypeSignature>,
+    pub type_map: BTreeMap<ClarityName, TypeSignature>,
```
kantai, line 67: Does this need to be `pub`?

kantai (approved):
This LGTM! Only comment is about the visibility of `TupleTypeSignature::type_map`, I think that field should remain private.

src/vm/types/signatures.rs
lgalabru: Yeah when doing the merge operation, I'm draining, in order to avoid cloning. Do you prefer a method returning a mutable ref? or cloning?

kantai: Sorry for the slow response -- it looks like the draining/mutation is all happening in `shallow_merge`,  which is defined in this same module, so the `pub` shouldn't be necessary.

lgalabru: Indeed, got confused by `pub` behavior, thanks. 
Addressed with https://github.com/blockstack/stacks-blockchain/pull/2117/commits/65706f366960b80874843ec3bf07e8dd5a94ba91

njordhov: Perhaps a reference counted pointer could be of use here:
https://doc.rust-lang.org/book/ch15-04-rc.html

PR #2119, opened 2020-12-02
Fix/2108

### LINKED ISSUE #2108: ChainsCoordinator/Burnchain integration breaks during Bitcoin forks
Bitcoin forks can lead the integration between the `ChainsCoordinator` and the `Burnchain::sync_with_indexer` to break -- in particular:

1. The ChainsCoordinator attempts to process a new burnchain fork by finding the first ancestor in the current sortition fork that is in common (see: https://github.com/blockstack/stacks-blockchain/blob/master/src/chainstate/coordinator/mod.rs#L481)
2. During the processing of the burn block, the chains coordinator updates the current canonical sortition fork _only if_ the newly processing burn block's height is greater than the current canonical sortition.
3. The `Burnchain::sync_with_indexer` polls the coordinator to process new burn blocks _as they are processed_.

This, together, means that if for example, the following fork occurred:

```
A -> B -> C -> D
         \__> C' -> D' -> E'
```

With block arrivals like `A, B, C, D, C', D', E'`, then the chains coordinator would process A-> D, setting the canonical chain tip to `D`, then `C'` would be correctly processed, but when the coordinator is polled to process `D'`, the canonical chain tip is _still_ `D`, so the coordinator will attempt to process `C'` _again_. This causes the chains coordinator to abort (because it will eventually fail a SQLite uniqueness constraint when trying to insert `C'` a second time).

I believe the best way to address this is by making the ChainsCoordinator check the block height of the "canonical" burnchain tip before processing sortitions, and ensuring that the height is greater than its view of the canonical tip.

kantai:
## Description

Fixes #2108 through 2 changes in the ChainsCoordinator:

1. `is_sortition_processed` searches all sortitions for a _still_ `pox_valid` sortition at the given burnchain header hash. Previously, it only looked in the current "sortition fork" -- that's what led to the chains coordinator incorrectly concluding that a block hadn't been processed yet.
2. `canonical_sortition_tip` is _always_ updated based on the canonical burnchain tip. 

This additionally required a slight re-working of the RBF logic in the neon_node. The previous RBF logic would keep all the mined blocks in the same vec regardless of the burnchain tip at which they mined -- if a burnchain fork occurred, the "attempts" should be reset. This PR does this by keeping a map of the mined blocks at each burnchain tip. When a burnchain tip is processed, the vec for that tip is removed from the map.

## Type of Change
- [x] Bug fix

No -- this corrects a previously (thread) panic inducing bug, which would not create a fork, but rather just halt chain processing.

No

## Testing information

This PR adds a new integration test to `neon_integrations` which uses the `invalidateblock` command to trigger a reorg in the spawned regtest bitcoind node. Without the changes in this PR, this test would cause the same behavior witnessed by the Xenon miner in #2108.

### DIFF
--- a/testnet/stacks-node/src/neon_node.rs
+++ b/testnet/stacks-node/src/neon_node.rs
@@ -1,6 +1,7 @@
 use super::{BurnchainController, BurnchainTip, Config, EventDispatcher, Keychain};
 use crate::config::HELIUM_BLOCK_LIMIT;
 use crate::run_loop::RegisteredKey;
+use std::collections::HashMap;
 
 use std::cmp;
 use std::collections::VecDeque;
@@ -401,7 +402,8 @@ fn spawn_miner_relayer(
     let mut mem_pool = MemPoolDB::open(false, TESTNET_CHAIN_ID, &stacks_chainstate_path)
         .map_err(NetError::DBError)?;
 
-    let mut last_mined_blocks = vec![];
+    let mut last_mined_blocks: HashMap<BurnchainHeaderHash, Vec<AssembledAnchorBlock>> =
+        HashMap::new();
     let burn_fee_cap = config.burnchain.burn_fee_cap;
     let mine_microblocks = config.node.mine_microblocks;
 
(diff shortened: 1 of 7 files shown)


testnet/stacks-node/src/neon_node.rs
```diff
@@ -401,7 +402,8 @@ fn spawn_miner_relayer(
     let mut mem_pool = MemPoolDB::open(false, TESTNET_CHAIN_ID, &stacks_chainstate_path)
         .map_err(NetError::DBError)?;
 
-    let mut last_mined_blocks = vec![];
+    let mut last_mined_blocks: HashMap<BurnchainHeaderHash, Vec<AssembledAnchorBlock>> =
```
jcnelson, line 405: Trying to understand this change.  We need to keep `last_mined_blocks` grouped by burnchain header hash because the burnchain can fork, and the miner can be working on two or more burnchain forks at once?

jcnelson (approved):
LGTM!  Thanks @kantai!

testnet/stacks-node/src/neon_node.rs
kantai: Yes -- this deals with the following situation (which occurs in the regtest forking integration test, but would probably be rare in practice):

A -> B -> C -> D
          \_> C' -> D' -> E'

1. Miner mines on burnchain block "C", so the block commit would be included in "D".
2. Before D is mined, C'->D'->E' is discovered by the miner, so the current chain tip is E'
3. `last_mined_blocks` would contain block commits on a different chain tip -- as previously implemented, the miner would deduce that an RBF should be attempted, and try to re-spend a UTXO that was forked out of existence.

With a burn header hash-keyed map, this problem goes away, because the miner would only see their prior attempts at the same burnchain tip. If the C->D chain regained power and became part of the canonical chain again, the miner would have their corresponding block commit data that could be broadcasted if they won the sortition.

lgalabru (approved):
Looks good to me, nice hunt @kantai :)

PR #2120, opened 2020-12-02
Clarity: variadic map function 

lgalabru:
This PR is addressing https://github.com/clarity-lang/reference/issues/27

Todo:
- [x] update cost functions


src/vm/analysis/type_checker/natives/sequences.rs
```diff
@@ -65,31 +66,47 @@ pub fn check_special_map(
...
-            let mapped_type =
-                function_type.check_args(checker, &[TypeSignature::min_string_utf8()])?;
-            (mapped_type, utf8_data.into())
-        }
-        _ => return Err(CheckErrors::ExpectedSequence(argument_type).into()),
-    };
-    TypeSignature::list_of(mapped_type, len)
+    let mut func_args = vec![];
+    let mut min_args = u32::MAX;
+    let mut max_args = u32::MIN;
```
jcnelson, line 72: Isn't this just `0`?

src/vm/docs/mod.rs
```diff
@@ -448,12 +448,14 @@ inputted value.",
 };
 
 const MAP_API: SpecialAPI = SpecialAPI {
-    input_type: "Function(A) -> B, (list A)",
-    output_type: "(list B)",
-    signature: "(map func list)",
+    input_type: "Function(A) -> B, (list A), (list B), ...",
```
jcnelson, line 451: Shouldn't this be:
```
input_type: "Function(A1, A2, ..., An) -> B, (list A1), (list A2), ..., (list An)",
output_type: "(list B)",
signature: "(map func list-A1 list-A2, ..., list-An)",
```

The function doesn't have to return type `B` if the third argument is a `(list B)`, does it?

njordhov: Shouldn' this refer to a _sequence_ rather than a list?
Also: https://github.com/clarity-lang/reference/issues/7

src/vm/analysis/type_checker/natives/sequences.rs
```diff
@@ -65,31 +66,47 @@ pub fn check_special_map(
...
+                    ListType(list_data) => list_data.destruct(),
+                    BufferType(buffer_data) => (TypeSignature::min_buffer(), buffer_data.into()),
+                    StringType(ASCII(ascii_data)) => {
+                        (TypeSignature::min_string_ascii(), ascii_data.into())
+                    }
+                    StringType(UTF8(utf8_data)) => {
+                        (TypeSignature::min_string_utf8(), utf8_data.into())
+                    }
+                };
+                min_args = min_args.min(len);
```
kantai, line 87: I'm not sure `min_args` and `max_args` are operating correctly here. I think the behavior we want is that the type checker estimates the maximum possible "output" length, and then at runtime we can do one of two things:

1. "silent exit" -- if the input sequences are different length, just iterate until the shortest sequence is complete
2. "abort" -- if the input sequences are different length, abort the transaction

I'm inclined to think we want (1), but either way, I think the type checker should be a little more optimistic, for example:

```scm
(define-private (plus a b) (+ a b))
(define-private (foo (first (list 15 int)) (second (list 10 int)))
    (if (is-eq (len first) (len second))
         (some (map plus first second))
          none))
```

Should be able to type-check, because while we know the maximum _possible_ length of the input lists, we don't know their actual lengths.

So the type checker should, I think, just compute the min length across the inputs, and use that as the output length.

njordhov: https://github.com/clarity-lang/reference/issues/27 proposes another alternative for resolving different sequence lengths:

3. Extend shorter input sequences to the length of the longest by repeating their last item.

This can be used to pass on arguments to the applied function, without requiring an explicit expression such as repeat to fill the shorter sequences.

lgalabru: Solution 1 makes sense to me, will revisit the approach.

lgalabru: Actually, what about solution 4, and iterate until the longest sequence is complete?
If the application is a variadic, I think I would expect the following behavior:
```
(map + (list 1 1 1 1 1) (list 1 1 1)  (list 1)) -> (list 3 2 2 1 1)
```

lgalabru: It probably has to be a mix - Solution 1 for fixed-len parameters applications, and solution 4 for variadic applications.

jcnelson: Also in favor of solution 1.  AFAIK user-defined Clarity functions can't be variadic, so solution 4 would be a potentially-confusing special case that applies only to native functions like `+`.

lgalabru: My current thinking: all the solutions will lead to trade-off / limitations so we should pick the one making the more sense for blockchain applications, and optimize for predictability / consistency.

As of today, one of the main use cases of blockchains are manipulation of values, so I think that we should ease manipulating list of numbers + mathematical operators (which includes variadics).

The fact that the type signature of the result could depend on the nature of the application (variadic or not) is a good signal for keeping the current behavior in the type checker.

https://github.com/blockstack/stacks-blockchain/pull/2120/commits/08b3f1c48572082b763663f6158532cd90e210d2 is revisiting the behavior of the execution.
@kantai Thoughts?

src/vm/analysis/type_checker/natives/sequences.rs
lgalabru: Yep

src/vm/docs/mod.rs
lgalabru: Is
```
    input_type: "Function(A, B, ..., N) -> X, (list A1 A2 ... Am), (list B1 B2 ... Bm), ..., (list N1 N2 ... Nm)",
    output_type: "(list X)",
    signature: "(map func list-A list-B ... list-N)",
```
maybe more clear?

njordhov: > maybe more clear?

The `map` function can also take other sequence types (buffers and strings), so shouldn't the signature be more like:

```clojure
(map func seq-A seq-B ... seq-N) 
```

src/vm/analysis/type_checker/natives/sequences.rs
kantai: Definitely prefer solution 1 or 2 -- whether or not the applied function is variadic itself should not alter the behavior of map -- that seems very surprising to me. For how this is handled in most languages see (https://en.wikipedia.org/wiki/Map_(higher-order_function)#Language_comparison) -- Common Lisp, Haskell, C#, F#, Python 3 (but not 2!), Rust, Scala, and Swift all stop at the shortest list. Scheme and a few other sane languages throw an error if the lists are not the same length. PHP and Javascript do some craziness with undefineds/nulls that we couldn't support in our type system.

> The fact that the type signature of the result could depend on the nature of the application (variadic or not) is a good signal for keeping the current behavior in the type checker.

I don't think so. The type signature of the result wouldn't depend on whether or not it was variadic -- the type signature of the result is always the return type of the applied function with maximum length set to be the minimum of the maximum length of its inputs.

If we wanted to go with solution 2 (i.e., abort if the lists aren't the same length), then we _could_ keep this as a type error -- it would basically require the previous example to implemented using `as-max-len`:

```
(define-private (plus a b) (+ a b))
(define-private (foo (first (list 15 int)) (second (list 10 int)))
    (match (as-max-len? first 10) first-limited
         (some (map plus first-limited second))
          none))
```

So that the maximum length of the lists is the same. However, that doesn't solve the issue entirely, because the lists could still be different lengths (our list types specify the maximum length, not the actual length), so to avoid an abort, it would need to be written as:

```
(define-private (plus a b) (+ a b))
(define-private (foo (first (list 15 int)) (second (list 10 int)))
    (match (as-max-len? first 10) first-limited
         (if (is-eq? (len first-limited) (len second))
             (some (map plus first-limited second))
             none)
          none))
```

Which doesn't seem to be better -- the type checker isn't really protecting the contract from anything.

lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2120/commits/bd6686aa2794726c153f5ac35305b3de2b75c3c9

njordhov: Other languages have a good reason to stop at the shortest list in variadic map (option 1): If called with a lazy/infinite list, the loop will go on forever. This applies to for example the [map in clojure](https://clojuredocs.org/clojure.core/map) where it is common practice to map a finite list with a [lazy infinite](https://clojuredocs.org/clojure.core/repeat) list. But this design rationale for  obviously does not apply to Clarity.

kantai:
```

kantai (approved):
LGTM!

jcnelson (approved):
LGTM!

PR #2122, opened 2020-12-02
Clarity: Static failures for unchecked results

lgalabru:
This PR is addressing https://github.com/clarity-lang/reference/issues/29.

### DIFF
--- a/src/chainstate/stacks/boot/mod.rs
+++ b/src/chainstate/stacks/boot/mod.rs
@@ -860,7 +860,8 @@ pub mod test {
             )
             (begin
                 ;; take the stx from the tx-sender
-                (stx-transfer? amount-ustx tx-sender this-contract)
+                
+                (unwrap-panic (stx-transfer? amount-ustx tx-sender this-contract))
 
                 ;; this contract stacks the stx given to it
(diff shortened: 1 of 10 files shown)


kantai (approved):
This LGTM! Do you think we should have an `(ignore ...)` method to just silence the response type?

jcnelson (approved):
LGTM!

lgalabru:
@kantai

> method to just silence the response type?

Yeah that's a great nice to have idea, however it feels like going with `ignore` in terms of semantic could be a a bit confusing?
`pursue`? `continue`? post December 15th? :)

kantai:
> @kantai
> 
> > method to just silence the response type?
> 
> Yeah that's a great nice to have idea, however it feels like going with `ignore` in terms of semantic could be a a bit confusing?
> `pursue`? `continue`? post December 15th? :)

Yeah, `ignore` could be confusing, as would the others I think. Let's just push this off for now

lgalabru:
> Yeah, ignore could be confusing, as would the others I think. Let's just push this off for now

Sounds good. Meanwhile, I think that `(default-to ... none)` might be a legit workaround.

PR #2124, opened 2020-12-02
feat: emit STX mint events for initial STX allocations

zone117x:
Emit STX mint events for genesis/bootcode allocations. A "stub" STX transaction is created to contain the events, similar to the contract-deploy transactions created for the bootcode contracts.


zone117x:
Note: merging this into the parent branch `feat/v1-to-v2-import`, will be reviewed in the existing PR https://github.com/blockstack/stacks-blockchain/pull/2105

Issue #2118, opened 2020-12-02
Support STX transfer via Bitcoin operations


kantai:
Similar to our support for invoking `stack-stx` via on-Bitcoin operations, we should support STX transfers. This would allow users to submit STX transfers via broadcasting Bitcoin operations (similar to how Stacks 1.0 transfers work). These operations would be constrained in the same manner as the on-Bitcoin `stack-stx` operation:

* The operation must consume a "known good" UTXO -- we can reuse the PreStackStx op for this, but will rename to something like PreStxOp
* Operations are included in _only_ the Stacks block elected following the burn block that includes the STX transfer. This means that the operations _are not_ "resistant" to forks in the Stacks blockchain: if the block that includes the STX transfer is "forked" away from, the operation isn't re-included in "nephew" blocks.

friedger:
With the property of not being fork resistant what is the use case?

kantai:
The most immediate benefits are:

1. It is more resistant to miner transaction censorship -- miners would need to be able to successfully initiate a fork in order to censor the transaction.
1. Wallets that only support Bitcoin operations would still be able to submit transfers (this maintains a kind of feature parity for transfers with Stacks 1.0).

kantai:
Closed via #2129

Issue #2093, opened 2020-11-24
Disable User Burn Support Ops


kantai:
For Stacks 2.0 mainnet, we should disable the user burn support operation.

kantai:
The most straight-forward way to do this is to simply ignore the ops on parsing.

kantai:
Closed via #2129

Issue #2094, opened 2020-11-24
Bitcoin wire format rigidity


kantai:
From @lgalabru: 

> I keep wondering if the rigidity of our expectations, when diving into the outputs of a BTC transaction, is a good thing or not (the order, the total count, etc).
> I understand that it simplifies the retrieval, but there are 2 downsides that I can think about:
>
> *    Combining Stacks mining / support with BTC transactions: transaction's costs are rising, BTC fee costs of Stacks mining could be amortized if the transaction could be used for something else.
> *    Combining / Multiplexing Stacks protocols with other protocols: can we / what kind of meta-protocols can be built by combining some Bitcoin protocols (Stacks, Lightning, wBTC, etc).

I agree with the above, and think there's some ways we can explore to reduce this rigidity. I think for things like the total count of outputs and the order of them, there's a good reason for doing this beyond simplicity (we want all transactions to cost the same whether they are burns or PoX transfers: we shouldn't favor one over the other).

Another option for the UTXO chaining requirement is performing Bitcoin signature validation in the stacks-node itself, rather than relying on bitcoind's own validation. That's probably easier said than done, but it's definitely _possible_.

jcnelson:
> Another option for the UTXO chaining requirement is performing Bitcoin signature validation in the stacks-node itself, rather than relying on bitcoind's own validation. That's probably easier said than done, but it's definitely possible.

Can you elaborate on this idea?

kantai:
Yeah -- the thing we would need to do is identify that the scriptSig is in a format that we understand, for example, we could see a script that looks like:

```
<signature> <pubkey>
```

And _assume_ that the spent output is the normal p2pkh script (we'd compute the hash160 of the pubkey ourselves). Then, we'd need to actually perform the Bitcoin signature validation on the transaction as if this were the case. If it passes the signature validation, then we can associate the transaction with that pubkey.

We could perform similar validation for other standardized script types.

jcnelson:
We may need to bump this to stacks 2.1

jcnelson:
A really simple way to remove the pain from this is to increase the window of consideration for a `PreStxOp` and similar SIP-007 on-Bitcoin operations to 5 blocks, instead of 1 block.

PR #2129, opened 2020-12-04
Transfer STX via On-Bitcoin Operations

kantai:
## Description

This PR implements #2118: adding a StxTransferOp to our Bitcoin wire formats. This operation behaves the same as the StackStxOp -- an operation broadcasted on a burnchain block will be included in the Stacks block that is elected by the following burnchain block.

This PR also fixes one bug and one bad behavior from the previous implementation of the PreStxOp (renamed from PreStackStxOp, because the op is used for both StxTransferOp and StackStxOp):

1. Bug: The burnchain::bitcoin::bits parser update that included the referenced txid encoded the txid in the reverse order that the rest of the code base uses.
2. Bad behavior: Previously, the PreStxOp and corresponding transfer/stacking op could not be in the same block.

This PR fixes both of the above. 

This PR also implements #2093: disabling the user burn support by always failing to parse the OP.

## Type of Change
- [x] New feature

Yes, this change affects consensus.

Yes -- this PR updates the description of the Bitcoin wire formats in SIP-007.

## Testing information

This PR includes three kinds of tests:

1. Parse tests for the StxTransferOp
2. Integration tests in coordinator::tests
3. End-to-end tests in a new `neon_integrations` test. These tests perform two transfers, one with the pre-op and op split across two blocks and one with the pre-op and op in the same block.


sip/sip-007-stacking-consensus.md
```diff
@@ -534,21 +534,26 @@ Where `total_block_commit_amount` is equal to the sum of outputs [1, M+1].
...
 
-As described above, PoX allows stackers to submit `stack-stx` operations on Bitcoin as well as on
-the Stacks blockchain. Any such operations must be evaluated by miners before beginning execution of
-any anchor block following those operations.
+As described above, PoX allows stackers to submit `stack-stx`
+operations on Bitcoin as well as on the Stacks blockchain. The Stacks
+chain also allows addresses to submit STX transfers on the Bitcoin
+chain. Such operations are only evaluated by the miner of an anchor block
+elected in the burn block that immediately follows the burn block that included the
+operations. For example, if a TransferStxOp occurs in burnchain block 100, then the
```
jcnelson, line 544: Can you put `TransferStxOp` in code-quotes?

sip/sip-007-stacking-consensus.md
```diff
@@ -579,6 +584,26 @@ This operation includes an `OP_RETURN` output for the first Bitcoin output:
...
+            0      2  3                             19        80
+            |------|--|-----------------------------|---------|
+             magic  op     uSTX to transfer (u128)     memo (up to 61 bytes)
+```
+
+Where `op = $` (ascii encoded).
+
+Where the unsigned integer is big-endian encoded.
+
+The second Bitcoin output is the B58-encoding of the recipient Stacks address.
```
jcnelson, line 609: Bitcoin outputs do not contain addresses.  Should this say something about the `scriptPubKey` needing to be either a `p2pkh` or `p2sh` output, so that a Stacks address can be derived from the internal 20-byte hash?

src/burnchains/bitcoin/bits.rs
```diff
@@ -520,10 +520,9 @@ impl BitcoinTxInput {
 }
 
 fn to_txid(txin: &BtcTxIn) -> (Txid, u32) {
-    (
-        Txid(txin.previous_output.txid.0.clone()),
-        txin.previous_output.vout,
-    )
+    let mut bits = txin.previous_output.txid.0.clone();
```
jcnelson, line 523: Might make a comment here about how the `bitcoin-rs` library works in a different endian-ness than we're used to.

src/burnchains/burnchain.rs
```diff
@@ -637,11 +640,14 @@ impl Burnchain {
...
 
     /// Try to parse a burnchain transaction into a Blockstack operation
+    /// `pre_stx_op_map` should contain any valid PreStxOps that occurred before
+    ///   the currently-being-evaluated tx.
     pub fn classify_transaction(
         burnchain_db: &BurnchainDB,
         block_header: &BurnchainBlockHeader,
         burn_tx: &BurnchainTransaction,
         sunset_end_ht: u64,
+        pre_stx_op_map: &HashMap<Txid, PreStxOp>,
```
jcnelson, line 650: These are only the ones that occur in the same block though, right?

kantai: Yep -- I'll be more specific in the rustdoc description.

src/chainstate/burn/operations/transfer_stx.rs
```diff
@@ -0,0 +1,324 @@
...
+        TransferStxOp::parse_from_tx(
+            block_header.block_height,
+            &block_header.block_hash,
+            tx,
+            sender,
+        )
+    }
+
+    /// parse a StackStxOp
+    /// `pox_sunset_ht` is the height at which PoX *disables*
```
jcnelson, line 149: Copypasta comment? ;)

src/chainstate/stacks/db/blocks.rs
```diff
@@ -3738,6 +3739,54 @@ impl StacksChainState {
...
     }
 
+    /// Process any STX transfer bitcoin operations
+    ///  that haven't been processed in this Stacks fork yet.
+    pub fn process_transfer_ops(
+        clarity_tx: &mut ClarityTx,
+        operations: Vec<TransferStxOp>,
+    ) -> Vec<StacksTransactionReceipt> {
+        let mut all_receipts = vec![];
+        // NOTE: do _not_ return from this function without replacing this original tracker
```
jcnelson, line 3749: Could this be enforced by the compiler?  Maybe you could encapsulate the "body" of this method in a closure and run it, so any internal returns would be captured by the call?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -3973,6 +4024,11 @@ impl StacksChainState {
             let mut receipts =
                 StacksChainState::process_stacking_ops(&mut clarity_tx, stacking_burn_ops);
 
+            receipts.extend(StacksChainState::process_transfer_ops(
```
jcnelson, line 4027: Should we process on-Bitcoin operations in the order in which they appear on-Bitcoin?  I don't think correctness is a problem either way, but it may be surprising to users that on-Bitcoin operations aren't processed in the same order as they appear in the block (all Stacking operations happen in block-order, and then all STX-transfer operations happen in block-order).

jcnelson (approved):
This LGTM.  Thanks @kantai!  Only minor comments.

sip/sip-007-stacking-consensus.md
```diff
@@ -534,21 +534,26 @@ Where `total_block_commit_amount` is equal to the sum of outputs [1, M+1].
...
-As described above, PoX allows stackers to submit `stack-stx` operations on Bitcoin as well as on
-the Stacks blockchain. Any such operations must be evaluated by miners before beginning execution of
-any anchor block following those operations.
+As described above, PoX allows stackers to submit `stack-stx`
+operations on Bitcoin as well as on the Stacks blockchain. The Stacks
+chain also allows addresses to submit STX transfers on the Bitcoin
+chain. Such operations are only evaluated by the miner of an anchor block
+elected in the burn block that immediately follows the burn block that included the
+operations. For example, if a `TransferStxOp` occurs in burnchain block 100, then the
+Stacks block elected by burnchain block 101 will process that transfer.
```
psq, line 545: Is this as accurate as we can make it?  Aren't miners that mine on a different fork off block 100 also responsible to include the transfer in their block?  Seems like the code is doing this, but this should also be reflected here.

jcnelson: There's no "fair" way to do this.  The fundamental problem is that there can be an unbound number of Bitcoin blocks, and thus an unbound number of on-Bitcoin operations to process, in-between any pair of Stacks blocks.  We can't process them all; otherwise, the network would be vulnerable to a DDoS attack if a miner simply mines a Stacks block, waits for a long time (e.g. 10,000 Bitcoin blocks), and then mines a child off of that Stacks block.  Without constraining which Bitcoin blocks' on-chain operations will be considered, each node on the network network would be obliged to consider _all_ on-Bitcoin operations in those 10,000 Bitcoin blocks (even the act of iterating through them would be time-consuming).

The act of only considering on-chain Bitcoin operations from the last Bitcoin block is one of many possible ways to deal with this problem, and was chosen because it's the easiest to communicate to users.  Since miners will build off of the longest Stacks chain most of the time, and will try to mine blocks in each Bitcoin block most of the time, we can explain to users that if they send an on-Bitcoin STX-transfer, it'll get picked up by the Stacks miner who mines in the next Bitcoin block (with the caveats that there may be no block mined, or that the block mined might get orphaned; in both cases, the remedy is to try again).  Most of the time, this would work -- the user experience would be equivalent to sending a stx-transfer transaction and having it get mined in the next Stacks block.

psq: given the cost of a btc transaction, I don't think we have to worry much about having too many transactions used as a DDoS vector.  I'm actually a lot more worried of people having to spend several dollars to move only a few hundred STX (when a lot of them have been told to wait for Stacks 2.0 where the fees will be cheaper), although I'm not clear how many people will be affected as I'm not sure whether that's anyone with STX prior to 2.0, or only some using a certain wallet.

My understanding of the code (https://github.com/blockstack/stacks-blockchain/pull/2129/files#diff-36b251011e9d6834be636547f23dd28aa9b24b555331f637dd40cfd2a1552467R3790) is that it will try to apply all stx transfers from the last pox anchor block if it has not been already included in the current fork, and that behavior should be reflected in the SIP.  Same logic used to consider stacking via BTC.  But maybe I misunderstood the code, some times it is hard to read just from diffs....

src/chainstate/stacks/db/blocks.rs
kantai: Yep -- addressed in b4c7b8878ad442a81a5fe47b6e94de287de49305 by removing the `set_cost_tracker` function and adding a `with_temporary_cost_tracker` function.

src/chainstate/burn/operations/transfer_stx.rs
kantai: Addressed in https://github.com/blockstack/stacks-blockchain/pull/2129/commits/b4c7b8878ad442a81a5fe47b6e94de287de49305

src/burnchains/burnchain.rs
kantai: Addressed in https://github.com/blockstack/stacks-blockchain/pull/2129/commits/b4c7b8878ad442a81a5fe47b6e94de287de49305

src/burnchains/bitcoin/bits.rs
kantai: Added in https://github.com/blockstack/stacks-blockchain/pull/2129/commits/b4c7b8878ad442a81a5fe47b6e94de287de49305

sip/sip-007-stacking-consensus.md
kantai: Added in https://github.com/blockstack/stacks-blockchain/pull/2129/commits/b4c7b8878ad442a81a5fe47b6e94de287de49305

sip/sip-007-stacking-consensus.md
kantai: Yep -- https://github.com/blockstack/stacks-blockchain/pull/2129/commits/b4c7b8878ad442a81a5fe47b6e94de287de49305

src/chainstate/stacks/db/blocks.rs
kantai: Yep, switched to an ordered processing in https://github.com/blockstack/stacks-blockchain/pull/2129/commits/3320c16534e986ae33a118a7964db39722df7ff0

sip/sip-007-stacking-consensus.md
jcnelson: > given the cost of a btc transaction, I don't think we have to worry much about having too many transactions used as a DDoS vector.

We _absolutely_ have to worry about this.  Per my above comment, the number of on-BTC operations between any parent/child Stacks block is unbound -- it can be arbitrarily high, since the number of BTC blocks in-between any parent/child Stacks block can be arbitrarily high (since I can mine a Stacks block off of any existing Stacks block).  It may not be a problem right now, but that does not mean it won't ever be a problem.  In fact, the "size" of the problem increases with the length of the Bitcoin chain -- even if there are no on-BTC operations in-between a parent/child Stacks block pair, the processing logic still needs to iterate through all of the Bitcoin blocks in-between these two Stacks blocks' sortitions (which grows unbound with the length of the Bitcoin chain).

> My understanding of the code is that it will try to apply all stx transfers from the last pox anchor block if it has not been already included in the current fork, and that behavior should be reflected in the SIP

It applies all STX transfers from the last Bitcoin block, not the last PoX anchor block.

kantai: Yep -- the place where the stacks transfers to be processed are selected is roughly here: https://github.com/blockstack/stacks-blockchain/pull/2129/files#diff-36b251011e9d6834be636547f23dd28aa9b24b555331f637dd40cfd2a1552467R4014 -- the block processor fetches all the valid transfer operations from the parent burn block, and those are the ones later passed into the function you linked. So that behavior is pretty much as described in the SIP. The only way a Stacks transfer broadcasted on Bitcoin may be processed in two different forks is if there's a Bitcoin fork or a PoX fork.

> given the cost of a btc transaction, I don't think we have to worry much about having too many transactions used as a DDoS vector. I'm actually a lot more worried of people having to spend several dollars to move only a few hundred STX (when a lot of them have been told to wait for Stacks 2.0 where the fees will be cheaper), although I'm not clear how many people will be affected as I'm not sure whether that's anyone with STX prior to 2.0, or only some using a certain wallet.

The reason that it becomes a DDoS vector is that it's somewhat "unbound" how much processing would need to be done if the Stacks node processed all the on-Bitcoin operations between it's Stacks block parent and the current block. Imagine this scenario:

```
Bitcoin blocks:    1 -> 2 -> 3 -> 4 -> ... -> 10,000
```

And those Bitcoin blocks "elect" the following Stacks blocks:

```
Stacks blocks:     a -> b -> c -> d -> ...
                    \---------------------> b'
```
where `b'` is elected in Bitcoin block 10,000. Now, if there's 5 STX transfers in every Bitcoin block from 2 -> 10,000, then each Stacks node needs to process 50,000 STX transfers right before processing `b'`. On top of that, this DDoS vector never goes away -- a subsequent miner could mine `c'` as a child of `b` and nodes would need to perform that processing again.

psq: thank you for the explanation @kantai 

I just can't get over the fact the finality of the transaction is not guaranteed.  There has to be a way to make this work like the mempool works for regular Stacks transactions and guarantee all Stacks forks get the transaction.

kantai: Yeah -- it could work that way if we implemented some of the items discussed in #2094. In particular, if the `stacks-node` could validate Bitcoin signatures itself without relying on bitcoind (this would be something like a parallel implementation of bitcoin signature verification), then these transactions could be considered "legal" stacks transactions in their own right.

lgalabru (approved):
Looks good to me!

PR #2133, opened 2020-12-07
Fix/marf 1294

jcnelson:
This fixes the bug that NCC group discovered in the MARF proof verifier back in February.  The actual fix is just one line, but I also added in a modified version of the test case they wrote that demonstrated the flaw.


kantai (approved):

lgalabru (approved):
LGTM

PR #2143, opened 2020-12-09
Cost Voting Contract Integration

kantai:
This PR integrates the cost voting contract with the boot code and LimitedCostTracker. At the start of each block, the LimitedCostTracker checks if any proposals have been confirmed since the last block, and if so, checks the proposal for validity, and applies the proposals. The current state of the cost tracker is materialized in the Clarity DB as a metadata entry.

This implements #1895.


src/vm/analysis/arithmetic_checker/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+        use vm::functions::define::DefineFunctionsParsed::*;
+        if let Some(define_type) = DefineFunctionsParsed::try_parse(expr)
+            .map_err(|_| Error::UnexpectedContractStructure)?
+        {
+            match define_type {
+                // The _arguments_ to Constant, PersistedVariable, FT defines must be checked to ensure that
+                //   any _evaluated arguments_ supplied to them are valid with respect to read-only requirements.
+                Constant { value, .. } => self.check_expression(value),
+                PrivateFunction { signature, body } => self.check_define_function(signature, body),
+                ReadOnlyFunction { signature, body } => self.check_define_function(signature, body),
```
lgalabru, line 101: Why are we allowing private and read-only, but not public functions?

src/vm/analysis/arithmetic_checker/tests.rs
```diff
@@ -0,0 +1,250 @@
...
+#[test]
+fn test_bad_defines() {
+    let tests = [
+        ("(define-public (foo) (ok 1))", DefineTypeForbidden(DefineFunctions::PublicFunction)),
+        ("(define-map foo-map ((a uint)) ((b uint))) (define-private (foo) (map-get? foo-map {a: u1}))", DefineTypeForbidden(DefineFunctions::Map)),
+        ("(define-data-var foo-var uint u1) (define-private (foo) (var-get foo-var))", DefineTypeForbidden(DefineFunctions::PersistedVariable)),
+        ("(define-fungible-token tokaroos u500)", DefineTypeForbidden(DefineFunctions::FungibleToken)),
+        ("(define-fungible-token tokaroos)", DefineTypeForbidden(DefineFunctions::FungibleToken)),
+        ("(define-non-fungible-token tokaroos uint)", DefineTypeForbidden(DefineFunctions::NonFungibleToken)),
+        ("(define-trait foo-trait ((foo (uint)) (response uint uint)))", DefineTypeForbidden(DefineFunctions::Trait)),
```
lgalabru, line 62: What are our expectations for top level `let` / `begin` expressions?

src/vm/analysis/arithmetic_checker/mod.rs
lgalabru: nvm

src/vm/analysis/arithmetic_checker/tests.rs
kantai: As written now, they would be allowed, but still must be "arithmetic only". I can't think of a good use case for why a contract would want to do this, but also don't think there's much of a downside for the allowing it.

src/vm/analysis/arithmetic_checker/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+
+#[cfg(test)]
+mod tests;
+
+///
+/// A static-analysis pass that checks whether or not
+///  a proposed cost-function defining contract is allowable.
+/// Cost-function defining contracts may not use contract-call,
+///  any database operations, or iterating operations (e.g., list
+///  operations)
```
jcnelson, line 43: I think you forgot to mention in this comment that it cannot have any trait definitions or trait references, going by the error variant below.

src/vm/analysis/arithmetic_checker/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+        self.check_expression(body)
+    }
+
+    fn check_top_levels(&self, expr: &SymbolicExpression) -> Result<(), Error> {
+        use vm::functions::define::DefineFunctionsParsed::*;
+        if let Some(define_type) = DefineFunctionsParsed::try_parse(expr)
+            .map_err(|_| Error::UnexpectedContractStructure)?
+        {
+            match define_type {
+                // The _arguments_ to Constant, PersistedVariable, FT defines must be checked to ensure that
```
jcnelson, line 97: Comment copypasta?  It doesn't look like you're doing anything with `PersistedVariable` and `FT`.

src/vm/costs/mod.rs
```diff
@@ -194,6 +260,235 @@ pub enum CostErrors {
...
+    clarity_db: &mut ClarityDatabase,
+    to_store: &CostStateSummary,
+) -> Result<()> {
+    let block_height = clarity_db.get_current_block_height();
+
+    clarity_db.put(
+        "vm-costs::last-processed-at-height",
+        &Value::UInt(block_height as u128),
+    );
+    let serialized_summary =
```
jcnelson, line 296: Pretty sure this is the case, but just confirming -- metadata stored for a contract is _not_ consensus-critical, right?  It's just a (block-hash, key, value) triple in the side-store, with no direct link to the MARF, right?  Otherwise we'll want to avoid using JSON here.

src/vm/costs/mod.rs
```diff
@@ -194,6 +260,235 @@ pub enum CostErrors {
...
+        serde_json::to_string(&SerializedCostStateSummary::from(to_store.clone()))
+            .expect("BUG: failure to serialize cost state summary struct");
+    clarity_db.set_metadata(
+        &STACKS_BOOT_COST_VOTE_CONTRACT,
+        "state_summary",
```
jcnelson, line 301: This key can never conflict with anything the cost voting contract will ever insert, right?  Would it make sense to make this value an illegal Clarity string, or something, to make sure?

src/vm/costs/mod.rs
```diff
@@ -194,6 +260,235 @@ pub enum CostErrors {
...
+                    .unwrap(),
+                ),
+            )
+            .expect("BUG: Failed querying confirmed-proposals")
+            .expect_optional()
+            .expect("BUG: confirmed-proposal-count exceeds stored proposals")
+            .expect_tuple();
+        let target_contract = match entry
+            .get("function-contract")
+            .unwrap()
```
jcnelson, line 351: Not a bug, but maybe use an `expect()` instead of an `unwrap()`?

src/vm/costs/mod.rs
```diff
@@ -194,6 +260,235 @@ pub enum CostErrors {
...
+            .clone()
+            .expect_principal()
+        {
+            PrincipalData::Contract(contract_id) => contract_id,
+            _ => {
+                warn!("Confirmed cost proposal invalid: function-contract is not a contract principal";
```
jcnelson, line 357: Is this reachable with the way the cost voting contract is written?

jcnelson: Ah, yup, it is -- the Clarity contract only requires this to be a `principal`, and doesn't check to see if it's a contract principal.  Is there a way the Clarity contract can do that check, so no one can vote on a proposal with an invalid principal?

src/vm/costs/mod.rs
```diff
@@ -194,6 +260,235 @@ pub enum CostErrors {
...
+        let cost_func_ref = match clarity_db.load_contract_analysis(&cost_contract) {
+            Some(c) => {
+                if !c.is_cost_contract_eligible {
+                    warn!("Confirmed cost proposal invalid: cost-function-contract uses non-arithmetic or otherwise illegal operations";
+                          "confirmed_proposal_id" => confirmed_proposal,
+                          "contract_name" => %cost_contract,
+                    );
+                    continue;
+                }
+                if !c.public_function_types.contains_key(&cost_function)
```
jcnelson, line 416: We don't care if a cost function is `define-public`, `define-read-only`, or `define-private`?  Would it help things any if we constrained cost functions to all be `define-read-only`?  I'm not sure a cost function could be `define-public`, for example, if the cost function wasn't already required to evaluate to a `(response A B)`.

jcnelson: More specifically, `cost_contract` is the contract that contains the cost function, right?  If so, I think you may have confused this `if`-condition with the `if`-condition below for `target_contract` (which I think is the contract who is about to get the new cost function?)

src/vm/costs/mod.rs
```diff
@@ -296,7 +599,7 @@ impl LimitedCostTracker {
         self.cost_function_references = m;
         self.cost_contracts = cost_contracts;
 
-        clarity_db.roll_back();
+        clarity_db.commit();
```
jcnelson, line 602: Won't this cause a new MARF trie to be stored?

src/vm/costs/mod.rs
```diff
@@ -194,6 +260,235 @@ pub enum CostErrors {
...
+    let confirmed_proposals_count = clarity_db
+        .lookup_variable(&cost_voting_contract, "confirmed-proposal-count")
+        .map_err(|e| CostErrors::CostComputationFailed(e.to_string()))?
+        .expect_u128();
+    debug!("Check cost voting contract";
+           "confirmed_proposal_count" => confirmed_proposals_count,
+           "last_processed_count" => last_processed_count);
+
+    // we need to process any confirmed proposals in the range [fetch-start, fetch-end)
+    let (fetch_start, fetch_end) = (last_processed_count, confirmed_proposals_count);
```
jcnelson, line 329: I'm not sure this is getting adequate stress-testing.  What happens if, for example, 10 different cost-functions all get approved in the same block?  What happens if all cost-functions are for the same target contract -- which one gets applied for interception?  Can we confirm that cost functions take effect only on the next _block_ (but not on the next _transaction_)?

src/vm/analysis/arithmetic_checker/mod.rs
kantai: Ah, I think this comment may have been left-over from a first implementation of this pass to make it more generic, will clean up the comments.

src/vm/costs/mod.rs
kantai: Yep, metadata isn't consensus critical.

src/vm/costs/mod.rs
kantai: Yeah, it couldn't conflict with anything that the contract would insert (metadata keys aren't contract chosen), but I can add some more illegal characters to be sure.

src/vm/costs/mod.rs
kantai: Not currently -- there's no `is-contract` method in Clarity. That should perhaps be added, but not necessarily for this case. Getting a proposal accepted is an expensive proposition, people voting on those proposals will be incentivized to confirm that the proposal they are voting on is valid.

src/vm/costs/mod.rs
kantai: No, these `if` statements are correct, the`target` should always be a read-only function. The cost function can only be a read-only or private (though this `if` statement shouldn't check public functions, which wouldn't be allowed in any case by the cost-contract eligibility analysis).

src/vm/costs/mod.rs
kantai: Yeah, this is just creating and releasing `SAVEPOINT`s, it won't commit the trie (the `begin` and `commit` methods in the clarity.rs and clarity_db/marfed_kv modules are not very descriptively named).

src/vm/costs/mod.rs
kantai: > What happens if, for example, 10 different cost-functions all get approved in the same block?

They would all be applied. I think we should probably add something like a minimum STX holding period for confirmed proposals, which would make confirming many cost-functions in the same block very difficult (and impose a limit on the number approved in a single block: the limit would be ~5 if the vote threshold is 20% of liquid stx).

> What happens if all cost-functions are for the same target contract -- which one gets applied for interception?

These proposals are applied in order, and proposals can override prior proposals.

> Can we confirm that cost functions take effect only on the next block (but not on the next transaction)?

The integration tests for this do test this behavior, but there isn't an explicit assertion in the code for this -- it's a consequence of the fact that the LimitedCostTracker is only constructed at the beginning of the block. The LimitedCostTracker could add an assertion that the block has not yet started.

src/vm/analysis/arithmetic_checker/mod.rs
kantai: Cleaned up in https://github.com/blockstack/stacks-blockchain/pull/2143/commits/c03277838d1e045d9f6c72c2e13e861b90819369

src/vm/analysis/arithmetic_checker/mod.rs
kantai: Added in https://github.com/blockstack/stacks-blockchain/pull/2143/commits/c03277838d1e045d9f6c72c2e13e861b90819369

src/vm/costs/mod.rs
kantai: The `public` option has been removed in https://github.com/blockstack/stacks-blockchain/pull/2143/commits/c03277838d1e045d9f6c72c2e13e861b90819369

src/vm/costs/mod.rs
jcnelson: TODO: add a `panic!()` if the `LimitedCostTracker` is instantiated in the middle of a block

src/vm/analysis/arithmetic_checker/mod.rs
```diff
@@ -0,0 +1,272 @@
...
+                    .ok_or(Error::UnexpectedContractStructure)?;
+
+                for pair in binding_list.iter() {
+                    let pair_expression = pair
+                        .match_list()
+                        .ok_or(Error::UnexpectedContractStructure)?;
+                    if pair_expression.len() != 2 {
+                        return Err(Error::UnexpectedContractStructure);
+                    }
+
```
jcnelson, line 224: Do we need to check that `pair_expression[0]` is an atom?

src/vm/costs/mod.rs
jcnelson: > They would all be applied. I think we should probably add something like a minimum STX holding period for confirmed proposals, which would make confirming many cost-functions in the same block very difficult (and impose a limit on the number approved in a single block: the limit would be ~5 if the vote threshold is 20% of liquid stx).

I think this is a good idea, and something we should test to make sure a bunch of (expensive) proposals don't all hit the network at once.

src/vm/tests/costs.rs
```diff
@@ -427,3 +436,383 @@ fn test_cost_contract_short_circuits() {
...
+                assert_eq!(
+                    &referenced_function.function_name,
+                    target.get_name_str(),
+                    "Cost function should still point to the boot costs"
+                );
+            }
+        }
+    };
+
+    marf_kv.test_commit();
```
jcnelson, line 817: Does it make any sense to check the `ExecutionCost` after evaluating each contract-call with the newly-applied cost function, in order to verify that the cost function is getting invoked?

jcnelson (approved):
So, this LGTM, but there are just a couple of minor questions and comments that need to be addressed first (plus all tests need to pass).  Once that's done, feel free to merge :+1:

src/vm/analysis/arithmetic_checker/mod.rs
kantai: No, we shouldn't need to here, because this check is invoked after the type checker (and that check is performed there).

src/vm/tests/costs.rs
kantai: It's tested elsewhere as well, but I added it here as well in 926383028

src/vm/costs/mod.rs
kantai: Addressed in ee998d20dc043c9cac68669aa8f03a7a8a02f407

src/vm/costs/mod.rs
kantai: Added in ee998d20dc043c9cac68669aa8f03a7a8a02f407

src/vm/costs/mod.rs
kantai: Added a panicking assertion in 196c893 that checks to make sure that there's no pending writes on the block.

kantai: And, in 950d829, I added a check / error type to the contract that prevents more than 10 proposals being confirmed in the same block (and a test case for that limit). We can revisit that constant or how we're limiting it later on.

jcnelson: Thanks!

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1649,7 +1649,8 @@ enum Signal {
 #[test]
 #[ignore]
 fn atlas_integration_test() {
-    if env::var("BITCOIND_TEST") != Ok("1".into()) {
+    // to run this test, BITCOIND_TEST=1 and ATLAS_TEST=1 must be set
+    if env::var("BITCOIND_TEST") != Ok("1".into()) || env::var("ATLAS_TEST") != Ok("1".into()) {
```
lgalabru, line 1653: Ha, do you mind if we go with
https://github.com/blockstack/stacks-blockchain/pull/2173/files#diff-e782414dee532a3fcdf6a65f99f5d6455c8773d7511827ed3011082d0440976b
as a patch? I'm sleeping thanks to the test :)

lgalabru (approved):
Can we just remove the gate for the ATLAS TEST? Bumping the timeout look fine on my other PR. If it keeps bothering us, I'll split this test in two.

testnet/stacks-node/src/tests/neon_integrations.rs
kantai: Sure -- will update.

PR #2145, opened 2020-12-09
Clarity Conveniences: `let*`, `element-at`, `contains`

kantai:
## Description

This implements three pretty nice conveniences for Clarity:

* `let*` -- This is an iterative let, where binding definitions may use prior binding values in the same let. This would be useful in all of the boot contracts (each boot contract uses nested lets to achieve the same thing, or does repetitive evaluation of the same expression to avoid nesting).
* `element-at` -- Get the *i*th element from a sequence. Again, this is something that would be useful in the boot contracts (particularly BNS), and probably broadly useful elsewhere.
* `contains` -- Does the given sequence contain an entry that `is-eq` to the given item? Useful in PoX contract.

Each of these is fairly simple to implement, and I think will end up being used in a lot of contracts.

## Type of Change
- [x] New feature

Yes -- new Clarity functions are consensus breaking.

`vm::docs` has been updated to auto-generate docs and examples for these new methods.

## Testing information

Test coverage is provided by tests that cover the new type checking code and the interpreter's implementation of these functions.


njordhov:
> let* -- This is an iterative let, where binding definitions may use prior binding values in the same let.

As Clarity doesn't allow rebinding, this version can just as well replace the parallel binding `let` with no loss in functionality. It is unnecessary to keep both. The existing one is just cumbersome.

src/vm/analysis/read_only_checker/mod.rs
```diff
@@ -200,7 +199,7 @@ impl<'a, 'b> ReadOnlyChecker<'a, 'b> {
                 self.check_all_read_only(args)?;
                 Ok(false)
             }
-            Let => {
+            Let | LetStar => {
```
lgalabru, line 202: Why not just replacing `let` with `let*`?

njordhov:
> contains -- Does the given sequence contain an entry that is-eq to the given item?

Shouldn't this function be typed to return a `bool` rather than an `optional`?

What about making the function determine whether a sequence contains a subsequence? Testing for a single entry is just a special case of testing for a subsequence.

kantai:
> > contains -- Does the given sequence contain an entry that is-eq to the given item?
> 
> Shouldn't this function be typed to return a `bool` rather than an `optional`?
> 

No -- the function returns the index of the first occurrence in the `some` case. This is strictly more useful: obtaining a bool is as simple as calling `is-some`.

> What about making the function determine whether a sequence contains a subsequence? Testing for a single entry is just a special case of testing for a subsequence.

While subsequence search may also be useful, these are almost always treated as different methods in language libraries. One reason for this is that doing subsequence search with a single element would require an unneccessary structure construction (`e.g., (contains foo (list elem))`), which would also be a concern in Clarity (contract authors would pay for the construction of the single-element list.

src/vm/analysis/read_only_checker/mod.rs
kantai: Yeah -- I suppose we could just replace `let` with `let*`.

njordhov:
> > Shouldn't this function be typed to return a `bool` rather than an `optional`?
> 
> No -- the function returns the index of the first occurrence in the `some` case. This is strictly more useful: obtaining a bool is as simple as calling `is-some`.

If Clarity wasn't strictly typed, I'd agree with you on `contains` returning an index. But the intent of the code becomes harder to discern when it has to be wrapped in `is-some`:

```clojure
(if (is-some (contains "blockstack" ch)) "yes" "no")
```

In contrast, someone familiar with the syntax can easily guess what this will do:
 
```clojure
(if (contains "blockstack" ch) "yes" "no")
 ```
Nor does `contains` make much sense when the intent is to get the index of the item:

```clojure
(match (contains "blockstack" ch)
  ix (element-at "blockstack" ix)
  "")
```

Perhaps another name would be a better fit, like `index-of`?

```clojure
(match (index-of "blockstack" ch)
  ix (element-at "blockstack" ix)
  "")
```

src/chainstate/stacks/boot/bns.clar
```diff
@@ -106,18 +106,11 @@
...
-  (if (is-eq (get cursor acc) (get limit acc))
-    { limit: (get limit acc), cursor: (+ u1 (get cursor acc)), value: i }
-    { limit: (get limit acc), cursor: (+ u1 (get cursor acc)), value: (get value acc) }))
-  
+    (ok (unwrap-panic
+      (element-at NAMESPACE_PRICE_TIERS (min u7 (- namespace-len u1)))))))
+
 (define-private (get-exp-at-index (buckets (list 16 uint)) (index uint))
-  (get value (fold element-at buckets { limit: index, cursor: u0, value: u0 })))
+  (unwrap-panic (element-at buckets index)))
```
kantai, line 113: @lgalabru -- do these updates to `bns.clar` look correct to you?

I think there was an off-by-one error in the indexing of the namespace price tiers (index 0 in the tiers should be the price of a length 1 namespace, not a length 0 namespace). I'd guess this affects the tests in `bns-test`, but they all fail because they don't run against the git repo's version of the clarity-bin:

```
  ExecutionError: Launch contract failed with bad exit code 1: Contract initialization error: 
    UnknownFunction("element-at")
    Near:
    [SymbolicExpression { expr: Atom(ClarityName("let")), id: 421, span: Span { start_line: 105, start_column: 4, end_line: 105, end_column: 6 } }, SymbolicExpression { expr: List([SymbolicExpression { expr: List([SymbolicExpression { expr: Atom(ClarityName("namespace-len")), id: 424, span: Span { start_line: 105, start_column: 10, end_line: 105, end_column: 22 } }, SymbolicExpression { expr: List([SymbolicExpression { expr: Atom(ClarityName("len")), id: 426, span: Span { start_line: 105, start_column: 25, end_line: 105, end_column: 27 } }, SymbolicExpression { expr: Atom(ClarityName("namespace")), id: 427, span: Span { start_line: 105, start_column: 29, end_line: 105, end_column: 37 } }]), id: 425, span: Span { start_line: 105, start_column: 24, end_line: 105, end_column: 38 } }]), id: 423, span: Span { start_line: 105, start_column: 9, end_line: 105, end_column: 39 } }]), id: 422, span: Span { start_line: 105, start_column: 8, end_line: 105, end_column: 40 } }, SymbolicExpression { expr: List([SymbolicExpression { expr: Atom(ClarityName("asserts!")), id: 429, span: Span { start_line: 106, start_column: 6, end_line: 106, end_column: 13 } }, SymbolicExpression { expr: List([SymbolicExpression { expr: Atom(ClarityName(">")), id: 431, span: Span { start_line: 107, start_column: 8, end_line: 107, end_column: 8 } }, SymbolicExpression { expr: Atom(ClarityName("namespace-len")), id: 432, span: Span { start_line: 107, start_column: 10, end_line: 107, end_column: 22 } }, SymbolicExpression { expr: LiteralValue(UInt(0)), id: 433, span: Span { start_line: 107, start_column: 24, end_line: 107, end_column: 24 } }]), id: 430, span: Span { start_line: 107, start_column: 7, end_line: 107, end_column: 26 } }, SymbolicExpression { expr: List([SymbolicExpression { expr: Atom(ClarityName("err")), id: 435, span: Span { start_line: 108, start_column: 8, end_line: 108, end_column: 10 } }, SymbolicExpression { expr: Atom(ClarityName("ERR_NAMESPACE_BLANK")), id: 436, span: Span { start_line: 108, start_column: 12, end_line: 108, end_column: 30 } }]), id: 434, span: Span { start_line: 108, start_column: 7, end_line: 108, end_column: 31 } }]), id: 428, span: Span { start_line: 106, start_column: 5, end_line: 108, end_column: 32 } }, SymbolicExpression { expr: List([SymbolicExpression { expr: Atom(ClarityName("ok")), id: 438, span: Span { start_line: 109, start_column: 6, end_line: 109, end_column: 7 } }, SymbolicExpression { expr: List([SymbolicExpression { expr: Atom(ClarityName("unwrap-panic")), id: 440, span: Span { start_line: 109, start_column: 10, end_line: 109, end_column: 21 } }, SymbolicExpression { expr: List([SymbolicExpression { expr: Atom(ClarityName("element-at")), id: 442, span: Span { start_line: 110, start_column: 8, end_line: 110, end_column: 17 } }, SymbolicExpression { expr: Atom(ClarityName("NAMESPACE_PRICE_TIERS")), id: 443, span: Span { start_line: 110, start_column: 19, end_line: 110, end_column: 39 } }, SymbolicExpression { expr: List([SymbolicExpression { expr: Atom(ClarityName("min")), id: 445, span: Span { start_line: 110, start_column: 42, end_line: 110, end_column: 44 } }, SymbolicExpression { expr: LiteralValue(UInt(7)), id: 446, span: Span { start_line: 110, start_column: 46, end_line: 110, end_column: 46 } }, SymbolicExpression { expr: List([SymbolicExpression { expr: Atom(ClarityName("-")), id: 448, span: Span { start_line: 110, start_column: 50, end_line: 110, end_column: 50 } }, SymbolicExpression { expr: Atom(ClarityName("namespace-len")), id: 449, span: Span { start_line: 110, start_column: 52, end_line: 110, end_column: 64 } }, SymbolicExpression { expr: LiteralValue(UInt(1)), id: 450, span: Span { start_line: 110, start_column: 66, end_line: 110, end_column: 66 } }]), id: 447, span: Span { start_line: 110, start_column: 49, end_line: 110, end_column: 68 } }]), id: 444, span: Span { start_line: 110, start_column: 41, end_line: 110, end_column: 69 } }]), id: 441, span: Span { start_line: 110, start_column: 7, end_line: 110, end_column: 70 } }]), id: 439, span: Span { start_line: 109, start_column: 9, end_line: 110, end_column: 71 } }]), id: 437, span: Span { start_line: 109, start_column: 5, end_line: 110, end_column: 72 } }]

      68 |     provider = await NativeClarityBinProvider.create(allocations, dbFileName, binFile);
```


Or should I just rename `element-at` to my like `bns-element-at` in this PR to avoid this issue for now?

lgalabru: Yeah working with BNS became particularly tedious yesterday, after merging the new clarity features, breaking backward compatibility. 
We might have a solution for this soon with https://github.com/blockstack/clarity-js-sdk/pull/83 thanks to @friedger.

lgalabru: I'm taking note of the off by one error and will investigate thank you!

src/vm/types/mod.rs
```diff
@@ -253,6 +253,96 @@ impl SequenceData {
...
+
+        Some(result)
+    }
+
+    pub fn contains(&self, to_find: Value) -> Result<Option<usize>> {
+        match self {
+            SequenceData::Buffer(ref data) => {
+                if let Value::Sequence(SequenceData::Buffer(to_find_vec)) = to_find {
+                    if to_find_vec.data.len() != 1 {
+                        Ok(None)
```
jcnelson, line 282: Is this reachable?  Doesn't the type-checker already ensure that `to_find` has length 1?

jcnelson: Same question for the other cases below.

jcnelson (approved):
LGTM!  Great turn-around on this @kantai!

src/chainstate/stacks/boot/costs.clar
```diff
@@ -159,6 +159,12 @@
 (define-read-only (cost_len (n uint))
     (runtime u1))
 
+(define-read-only (cost_element_at (n uint))
+    (runtime (linear n u1 u1)))
```
jcnelson, line 163: Wait, isn't this _O(1)_?

src/vm/functions/mod.rs
```diff
@@ -542,7 +553,7 @@ fn special_let(
                     return Err(CheckErrors::NameAlreadyUsed(binding_name.clone().into()).into())
                 }
 
-            let binding_value = eval(var_sexp, env, context)?;
+            let binding_value = eval(var_sexp, env, &inner_context)?;
```
jcnelson, line 556: Will this let me re-bind the same variable, like so?

```
(let (
   (x 0)
   (x (+ x 1))
   (x (+ x 2))
) (
   (print x) ;; should print 3
))
```

It's fine by me if it doesn't; I'm just curious.

src/vm/types/mod.rs
kantai: Yes, it is -- the type system only operates on the maximum possible length of the list, so a type signature of `(list 1 uint)` would admit an empty list (this is true of our buffer and string types as well).

src/vm/functions/mod.rs
kantai: No, it won't -- that's a type of variable shadowing, which we don't allow in Clarity.

src/chainstate/stacks/boot/costs.clar
kantai: Ah -- yes it is.

src/vm/docs/mod.rs
```diff
@@ -430,10 +430,13 @@ const LET_API: SpecialAPI = SpecialAPI {
     output_type: "A",
     signature: "(let ((name1 expr1) (name2 expr2) ...) expr-body1 expr-body2 ... expr-body-last)",
     description: "The `let` function accepts a list of `variable name` and `expression` pairs,
-evaluating each expression and _binding_ it to the corresponding variable name. The _context_
-created by this set of bindings is used for evaluating its body expressions. The let expression returns the value of the last such body expression.
+evaluating each expression and _binding_ it to the corresponding variable name.
+`let` bindings are iterative: when a `let` binding is evaluated, it may refer to prior binding.
```
njordhov, line 434: The more accurate terminology here is that bindings are performed **sequentially** (rather than in parallel as in the original implementation).

src/chainstate/stacks/boot/bns.clar
```diff
@@ -106,18 +106,11 @@
...
 
 (define-private (is-digit (char (buff 1)))
```
njordhov, line 115: Some 50 LOC for `is-digit` and related functions can now be simplified with `index-of`.

Issue #1926, opened 2020-09-30
Stacks chain differentiation on Bitcoin testnet


kantai:
In preparation for the Xenon testnet, we will need a strategy for differentiating between Stacks chains on the Bitcoin testnet. While the network would continue to function without this differentiation, it would be pretty degraded because of various nodes seeing block commits as valid vs. invalid, or trying and failing to communicate over the p2p system.

Possible options for differentiation:

1. first_burn_block_height
1. chain_id
1. Magic bytes in the Bitcoin wireformat

In all likelihood, I think we'd want to use all of the above.

diwakergupta:
@kantai @lgalabru don't we still need this?

kantai:
Yeah -- I think we do.

PR #2146, opened 2020-12-09
Xenon Magic Bytes

kantai:
This PR makes `next`'s Xenon mode use `Xe` as the magic bytes for Bitcoin operations.

The PR implements this by doing two things:

1. Actually parsing the `magic_bytes` field in the burnchain config from the node config file.
2. Checking if the burnchain mode is set to xenon, and if so, using the default Xenon magic bytes (this makes it so that people don't need to update their config files if they aren't running just the default xenon config).

Related: #1926

### DIFF
--- a/src/chainstate/stacks/block.rs
+++ b/src/chainstate/stacks/block.rs
@@ -947,7 +947,6 @@ mod test {
     use burnchains::BurnchainBlockHeader;
     use burnchains::BurnchainSigner;
     use burnchains::Txid;
-    use burnchains::BLOCKSTACK_MAGIC_MAINNET;
 
     use burnchains::bitcoin::BitcoinNetworkType;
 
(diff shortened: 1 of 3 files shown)


lgalabru (approved):
Can you also please update the xenon templates config (testnet/stacks-node/conf/xenon-*.toml)?

kantai:
>  Can you also please update the xenon templates config (testnet/stacks-node/conf/xenon-*.toml)?

With their current configs, these will end up going down the "default xenon" path for the MagicBytes, so they'll get `Xe`. If we leave it this way, if we want to do future testnet "resets" with the MagicBytes (e.g., switch to `X2` or something), then those configs will just update to the new MagicBytes.

jcnelson (approved):
LGTM!

PR #2147, opened 2020-12-10
chainstate.txt genesis data hash in `v2/info` endpoint

zone117x:
Closes https://github.com/blockstack/stacks-blockchain/issues/2138

Also updates the address import code to use the current network version.

```jsonc
{
    "peer_version": 385875968,
    "pox_consensus": "8e0561978fc5506b68a589c402dad97e862edb59",
    "burn_block_height": 15,
    "stable_pox_consensus": "6d37b8792f6bb6d3ee301082f483517274ffd66f",
    "stable_burn_block_height": 14,
    "server_version": "blockstack-core 0.0.1 => 23.0.0.0 (feat/genesis-hash-v2-info:c41889bb7+, debug build, macos [x86_64])",
    "network_id": 2147483648,
    "parent_network_id": 3669344250,
    "stacks_tip_height": 14,
    "stacks_tip": "cdc0320166bdadbe76daff27e6ff469315208004eb95db6a6e8aaab622dcd070",
    "stacks_tip_consensus_hash": "8e0561978fc5506b68a589c402dad97e862edb59",
    "genesis_chainstate_hash": "4f57293c77cec345c3ea89eeced83f03234c9e2b3edc3b37f2a131358fcc1de0",  // NEW
    "unanchored_tip": "0000000000000000000000000000000000000000000000000000000000000000",
    "unanchored_seq": 0,
    "exit_at_block_height": null
}
```


src/chainstate/stacks/db/mod.rs
```diff
@@ -748,6 +751,21 @@ impl StacksChainState {
         Ok(path_str)
     }
 
+    fn parse_genesis_address(addr: &str, address_version: u8) -> PrincipalData {
```
kantai, line 754: I think this function, rather than accepting `address_version`, should take the `mainnet` bool.

Then, the address version needs to differentiate between single sig and multi sig. As currently written, it just forces all addresses to single sig, which will break any of the multi sig addresses.

src/net/rpc.rs
```diff
@@ -103,6 +103,7 @@ pub const STREAM_CHUNK_SIZE: u64 = 4096;
 #[derive(Default)]
 pub struct RPCHandlerArgs<'a> {
     pub exit_at_block_height: Option<&'a u64>,
+    pub genesis_chainstate_hash: String,
```
kantai, line 106: This should be a `util::hash::Sha256Sum` struct, which can be constructed with `Sha256Sum::from_hex(&str)`

src/net/rpc.rs
```diff
@@ -163,6 +164,7 @@ impl RPCPeerInfoData {
         chainstate: &StacksChainState,
         peerdb: &PeerDB,
         exit_at_block_height: &Option<&u64>,
+        genesis_chainstate_hash: &String,
```
kantai, line 167: While this arg should probably be `&Sha256Sum` anyways, using `&String` rather than `&str` is usually discouraged because it forces unnecessary clones when you static strs.

src/chainstate/stacks/db/mod.rs
```diff
@@ -94,6 +94,11 @@ use core::CHAINSTATE_VERSION;
 
 use chainstate::stacks::db::unconfirmed::UnconfirmedState;
 
+use crate::burnchains::bitcoin::address::BitcoinAddress;
+
+pub static EMPTY_GENESIS_CHAINSTATE_HASH: &str =
```
kantai, line 99: This static should be removed when the hash struct switches to `Sha256Sum` -- if you want an easy way to get the empty value, you should add an `empty()` method to the Sha256Sum impl in `util::hash`

kantai (approved):
This looks good to me -- just some minor comments about passing the hash around as `Sha256Sum` rather than a `String`

src/chainstate/stacks/db/mod.rs
zone117x: Good catch, fixed!

src/net/rpc.rs
zone117x: This and the related feedback should all be fixed now.

src/chainstate/stacks/db/mod.rs
```diff
@@ -595,17 +597,15 @@ pub const MINER_FEE_WINDOW: u64 = 24; // number of blocks (B) used to smooth ove
...
+pub struct ChainStateAccountBalance {
+    pub address: String,
     pub amount: u64,
-    pub block_height: u64,
 }
 
-#[derive(Debug, Clone, Deserialize)]
-pub struct AccountBalance {
-    pub address: StacksAddress,
+pub struct ChainStateAccountVesting {
```
jcnelson, line 605: Let's avoid calling this `vesting` -- @jessesoslow had a problem with this in Stacks 1.0.  The tokens aren't vesting in a legal sense -- they're already purchased and owned, but just not unlocked yet.

src/util/hash.rs
```diff
@@ -123,6 +123,12 @@ impl_array_newtype!(Sha256Sum, u8, 32);
 impl_array_hexstring_fmt!(Sha256Sum);
 impl_byte_array_newtype!(Sha256Sum, u8, 32);
 
+impl Default for Sha256Sum {
```
jcnelson, line 126: Is there a reason we need this?

src/util/hash.rs
```diff
@@ -317,6 +323,9 @@ impl Sha256Sum {
         tmp.copy_from_slice(sha2_1.result().as_slice());
         Sha256Sum(tmp)
     }
+    pub fn empty() -> Sha256Sum {
+        Sha256Sum([0u8; 32])
```
jcnelson, line 327: An empty Sha256 (the hash of no data) is not all 0's.

zone117x: Suggestion on alternative? `unanchored_tip` property is all 0's?

src/util/hash.rs
zone117x: `struct RPCHandlerArgs<'a>` does not compile without it.

src/chainstate/stacks/db/mod.rs
zone117x: This data is almost 1-1 from the Stacks 1.0 `account_vesting` table. Should it be renamed to `locked`?

jcnelson: Sounds good to me -- we can call it `ChainstateAccountLockup` or something (also, `Chainstate` is one word).

src/util/hash.rs
jcnelson: `zero()` is fine.

stx-genesis/src/lib.rs
```diff
@@ -3,65 +3,57 @@ use std::io::{self, BufReader};
...
-    chainstate::stacks::StacksAddress,
-};
+pub struct GenesisAccountBalance {
+    /// A STX or BTC address (BTC addresses should be converted to STX when used).
+    pub address: String,
+    /// Balance in microSTX.
+    pub amount: u64,
+}
+
+pub struct GenesisAccountVesting {
```
jcnelson, line 13: Same here -- we should avoid calling it `vesting`.

jcnelson (approved):
This LGTM, but on the following conditions:

* Don't use the word `vesting`.  Tokens are already vested; they're just not liquid yet.
* Sha256sum should have a `zero()` method (we do this elsewhere).
* Make sure all tests pass

src/util/hash.rs
zone117x: Done!

src/chainstate/stacks/db/mod.rs
zone117x: Done! Replaced the other `vesting` wording with `lockups`.

stx-genesis/src/lib.rs
zone117x: Done!

kantai (approved):
LGTM!

PR #2149, opened 2020-12-10
Commit production Stacks 1.0 chainstate.txt (as of block 653759)

zone117x:
Closes https://github.com/blockstack/stacks-blockchain/issues/2137
Ref https://github.com/blockstackpbc/devops/issues/523

* Change to the `account_genesis_credit` function to allow balance crediting (rather than just balance setting/replacing). 
* Uses a small "test" chainstate.txt for unit tests. The full production chainstate.txt is used in debug/release builds, as well as in integration tests (via a feature flag).
* On-chain name & namespace parsing functions implemented in preparation for loading into the bns.clar contract.


src/chainstate/stacks/db/accounts.rs
```diff
@@ -214,14 +214,19 @@ impl StacksChainState {
...
         clarity_tx
             .with_clarity_db(|ref mut db| {
-                let balance = STXBalance::initial(amount);
+                let mut balance = STXBalance::initial(amount);
                 let total_balance = balance.get_total_balance();
 
                 let mut snapshot = db.get_stx_balance_snapshot_genesis(principal);
+                let existing_balance = snapshot.balance().amount_unlocked;
+                if existing_balance > 0 {
+                    balance.amount_locked = balance
```
jcnelson, line 223: This is incorrect.  `amount_locked` refers to tokens locked through PoX, and has nothing to do with tokens unlocked through the `lockup` contract.

jcnelson (changes requested):
This doesn't do what you think it does.  The initial balance of an address should be set to whatever the liquid amount of tokens the address owns in the chainstate import.  Locked tokens that have yet to unlock should be managed by the `lockup` contract.  If you need to represent an address's non-liquid, locked tokens in `/v2/accounts`, then that's a separate issue.

src/chainstate/stacks/db/accounts.rs
zone117x: This is a typo, should be incrementing `amount_unlocked`, just fixed it. The change to this fn should just result in an increment of balance rather than replacement of balance (liquid, not locked). Does it look correct to you now?

jcnelson (approved):
LGTM!

jcnelson:
Looks like your tests are timing out waiting for the runloop to start.  How long does it take to instantiate the Stacks 2.0 chainstate from the Stacks 1.0 chainstate?

zone117x:
On my machine importing the genesis balances takes ~70 seconds.

The tests on circleCI were getting OOM killed after several concurrent instance of this process were started from a test, causing a big spike in memory usage. Those now pass when using an "init without genesis data" constructor for the helium node.

The neon integration tests ran on gh actions are stilling failing from a timeout:
`'Timed out waiting for run loop to start', testnet/stacks-node/src/tests/neon_integrations.rs:198:13`

I could create another "init without genesis data" constructor for these tests as well, although that partially defeats the purpose of an integration test. 

Thoughts?

zone117x:
Notes from conversation:
* Implement a small `chainstate.txt` for usage in unit tests.
* Ideally the full `chainstate.txt` should be used in the integration tests, although this may require running on our own CI to get around timeouts.

kantai (approved):
LGTM!

PR #2152, opened 2020-12-10
Refactor cost calculation to accept multiple size inputs

reedrosenbluth:
Refactor the `runtime_cost` function to take a `Vec<u64>` instead of a single `u64`, thus enabling Clarity cost functions to take multiple size input args.

### DIFF
--- a/src/vm/costs/mod.rs
+++ b/src/vm/costs/mod.rs
@@ -65,7 +65,7 @@ pub fn runtime_cost<T: TryInto<u64>, C: CostTracker>(
     input: T,
 ) -> Result<()> {
     let size: u64 = input.try_into().map_err(|_| CostErrors::CostOverflow)?;
-    let cost = tracker.compute_cost(cost_function, vec![size])?;
+    let cost = tracker.compute_cost(cost_function, &[size])?;
 
     tracker.add_cost(cost)
 }
@@ -87,7 +87,7 @@ pub fn analysis_typecheck_cost<T: CostTracker>(
     let t2_size = t2.type_size().map_err(|_| CostErrors::CostOverflow)?;
     let cost = track.compute_cost(
         ClarityCostFunction::AnalysisTypeCheck,
-        vec![cmp::max(t1_size, t2_size) as u64],
+        &[cmp::max(t1_size, t2_size) as u64],
     )?;
     track.add_cost(cost)
 }
@@ -106,7 +106,7 @@ pub trait CostTracker {
     fn compute_cost(
         &mut self,
         cost_function: ClarityCostFunction,
-        input: Vec<u64>,
+        input: &[u64],
     ) -> Result<ExecutionCost>;
     fn add_cost(&mut self, cost: ExecutionCost) -> Result<()>;
     fn add_memory(&mut self, memory: u64) -> Result<()>;
@@ -128,7 +128,7 @@ impl CostTracker for () {
     fn compute_cost(
         &mut self,
         _cost_function: ClarityCostFunction,
-        _input: Vec<u64>,
+        _input: &[u64],
     ) -> std::result::Result<ExecutionCost, CostErrors> {
         Ok(ExecutionCost::zero())
     }
@@ -739,7 +739,7 @@ fn parse_cost(
 fn compute_cost(
     cost_tracker: &mut LimitedCostTracker,
     cost_function_reference: ClarityCostFunctionReference,
-    input_sizes: Vec<u64>,
+    input_sizes: &[u64],
 ) -> Result<ExecutionCost> {
     let mut null_store = NullBackingStore::new();
     let conn = null_store.as_clarity_db();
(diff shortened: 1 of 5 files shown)


src/vm/costs/mod.rs
```diff
@@ -776,8 +826,12 @@ impl CostTracker for LimitedCostTracker {
         // grr, if HashMap::get didn't require Borrow, we wouldn't need this cloning.
         let lookup_key = (contract.clone(), function.clone());
         if let Some(cost_function) = self.contract_call_circuits.get(&lookup_key).cloned() {
-            let input_size = input.iter().fold(0, |agg, cur| agg + cur);
-            compute_cost(self, cost_function, input_size)?;
+            let input_sizes = if input.is_empty() {
```
kantai, line 829: This shouldn't be necessary, we can just require that a cost-function that is targeting a zero-arg function have zero args as well.

src/vm/costs/mod.rs
```diff
@@ -413,21 +419,30 @@ fn load_cost_functions(clarity_db: &mut ClarityDatabase) -> Result<CostStateSumm
...
+                    });
+
+                if cost_function_types.is_none() {
                     warn!("Confirmed cost proposal invalid: cost-function-name not defined";
                           "confirmed_proposal_id" => confirmed_proposal,
                           "contract_name" => %cost_contract,
                           "function_name" => %cost_function,
                     );
                     continue;
                 }
```
kantai, line 437: For these gets, the `if let` construct allows you to do this without an unwrap.

```suggestion
                let cost_function_opt =  c.read_only_function_types
                            .get(&cost_function)
                            .or_else(|| c.private_function_types.get(&cost_function))
                    });
                let cost_function_types = if let Some(result) = cost_function_opt {
                    result
                } else {
                    warn!("Confirmed cost proposal invalid: cost-function-name not defined";
                          "confirmed_proposal_id" => confirmed_proposal,
                          "contract_name" => %cost_contract,
                          "function_name" => %cost_function,
                    );
                    continue;
                }
                
                
                (
                    ClarityCostFunctionReference {
                        contract_id: cost_contract,
                        function_name: cost_function.to_string(),
                    },
                    cost_function_types.clone(),
                )
```

(Also, we don't need to check for public functions here, because the cost function couldn't be a public function anyways).

src/vm/costs/mod.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2013-2020 Blocstack PBC, a public benefit corporation
+// Copyright (C) 2013-2020 Blockstack PBC, a public benefit corporation
```
jcnelson, line 1: Lol are there any other instances of this?  If so can you just fix them in this PR?

diwakergupta: Yikes, I must have introduced these in my PR that updated copyright to Stacks foundation 🤦🏽 Happy to send a separate PR to correct them all.

reedrosenbluth: Not seeing them elsewhere...

src/vm/costs/mod.rs
```diff
@@ -716,10 +753,16 @@ fn compute_cost(
...
+    let mut program = vec![SymbolicExpression::atom(
+        cost_function_reference.function_name[..].into(),
+    )];
+
+    let mut args = input_sizes
+        .into_iter()
+        .map(|size| SymbolicExpression::atom_value(Value::UInt(size.into())))
+        .collect::<Vec<SymbolicExpression>>();
+
+    program.append(&mut args);
```
kantai, line 765: Instead of using `into_iter/map/collect/append`, you can do the following:

```suggestion
    for input_size in input_sizes.iter() {
        program.append(SymbolicExpression::atom_value(Value::UInt(input_size as u128)));
    }
```

And then all these functions can accept `input_sizes: &[u64]`, which means you wouldn't need to call `.to_vec()` at line 860 (and other places can just do `&[arg]` instead of constructing a vec.

src/vm/costs/mod.rs
diwakergupta: Opened https://github.com/blockstack/stacks-blockchain/pull/2172

kantai (approved):
LGTM!

src/vm/tests/costs.rs
```diff
@@ -502,12 +508,23 @@ fn test_cost_voting_integration() {
        })
     ";
 
+        let bad_cost_args_definer_src = "
+    (define-read-only (cost-definition (b uint) (b uint))
```
jcnelson, line 512: Won't this fail to parse?

jcnelson (commented):
I'm having a hard time seeing how the user-defined cost function's arguments are generated.  Can you point me to where that happens?

reedrosenbluth:
> I'm having a hard time seeing how the user-defined cost function's arguments are generated. Can you point me to where that happens?

Currently throughout the codebase we are using a function called `runtime_cost` to do the cost computation. This function, however, only takes a single `input_size` arg, since all of the cost functions loaded at boot only take one arg. The 

> I'm having a hard time seeing how the user-defined cost function's arguments are generated. Can you point me to where that happens?

Is this what you're looking for?

https://github.com/blockstack/stacks-blockchain/blob/0261f0a401414e67eb5519d1ff0354d661ba8501/src/vm/costs/mod.rs#L760

jcnelson (approved):
LGTM!

PR #2050, opened 2020-11-11
Minor event dispatcher updates

kantai:
## Description

1. Add event_index for ordering events (the dispatch matrix scrambles the event order).
2. Add `locked_address` field to stx_lock_event

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] API reference/documentation update
- [ ] Other

## Does this introduce a breaking change?

No.

## Are documentation updates required?

No -- the sample event payload in `event-dispatcher.md` has been updated.

Issue #1719, opened 2020-07-06
Mempool should accept out-of-order nonces up to some limit


kantai:
From @lgalabru and @jcnelson's suggestion in https://github.com/blockstack/stacks-blockchain/pull/1697,

It would be useful in general to support this because transaction messages might get lost or reordered as they are propagated through the p2p network. Balance checks would only work off of the highest contiguous sequence of transactions.

This doesn't _necessarily_ need to be supported from the mempool, nodes could instead detect an out-of-order broadcast and ask other nodes for the missing transactions.

diwakergupta:
@kantai is this still an issue?

kantai:
Yes, the mempool still requires that nonces be submitted in order.

kantai:
Closed via #2153

PR #2153, opened 2020-12-10
Inconsistent mempool admission: #2050, #1719

kantai:
This addresses #2050 by way of #1719 -- it allows "out of order" nonce chaining. This solves #2050 because it no longer considers which nonces are currently in the mempool when considering whether or not to admit a TX (if the same nonce is in the mempool, it performs replace-by-fee or replace-across-fork checks).


jcnelson (approved):
I love deleting code!  LGTM; just make sure the tests pass first.

lgalabru (approved):

PR #2155, opened 2020-12-11
[Draft] Load names in genesis boot

zone117x:
Added on-chain name and namespace reading inside the `stx-genesis` lib.
WIP: inserting into bns.clar during boot up


zone117x:
Closed in favor of https://github.com/blockstack/stacks-blockchain/pull/2149

PR #2156, opened 2020-12-11
fix: `mstx_balance` to `ustx_balance` in tomls and readmes

zone117x:
Closes https://github.com/blockstack/stacks-blockchain/issues/2151

We changed a config variable name from `mstx_balance` to `ustx_balance` in a previous PR. This PR fixes all the example tomls and readme references. 

For backwards compatibility, the `mstx_balance` option was fixed so that it still works but prints a deprecation notice -- added this after report of sudden broken balance config.

### DIFF
--- a/testnet/stacks-node/src/config.rs
+++ b/testnet/stacks-node/src/config.rs
@@ -1,6 +1,5 @@
 use std::convert::TryInto;
-use std::fs::File;
-use std::io::{BufReader, Read};
+use std::fs;
 use std::net::{SocketAddr, ToSocketAddrs};
 
 use rand::RngCore;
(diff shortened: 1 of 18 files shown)


testnet/stacks-node/src/config.rs
```diff
@@ -30,17 +29,28 @@ pub struct ConfigFile {
     pub block_limit: Option<BlockLimitFile>,
 }
 
+#[derive(Clone, Deserialize, Default)]
+pub struct LegacyMstxConfigFile {
```
jcnelson, line 33: I don't think we need to worry about legacy support -- we're pre-release still.  Maybe we can just do a `panic!()` if we see `mstx_balance` so users know to change it to `ustx_balance`.

jcnelson (approved):
LGTM!

lgalabru (approved):
Looks good to me! cc @wileyj @CharlieC3 for the devops repo

testnet/stacks-node/src/config.rs
zone117x: I think we should leave it for now to make things easier for devs. We can remove after next marge to master.

kantai (approved):
This LGTM!

I think we should keep the backwards compatibility path in place for now, it makes people's lives easier who already have miners running, and we can always remove the path from the next version.

PR #2159, opened 2020-12-12
fix --features=tx-log which no longer builds

psq:
with the addition of `TransactionOrigin`, `transaction.serialize_to_vec()` fails to build because it does not know how to process the new field.  This fixes that.


kantai (approved):
Thanks for catching this @psq!

Issue #2157, opened 2020-12-11
PoX: reward cycle must be proof-of-burn only, and should be shorter


jcnelson:
Per #1805, we need to make it so a PoX prepare phase is always PoB.  This is the consensus-critical portion of the solution to #1805.  In addition, we'll also make it so that the prepare phase is shorter (100 blocks instead of 240).  So, the todo's here are:

- [ ] Make the prepare phase 100 blocks
- [ ] Make the prepare phase PoB-only, and reset the mining smoothing parameter to 0 (#2141) so evaluating prepare-phase sortitions does not require any Stacks chain state
- [ ] Update SIP-007 to describe the above.

lgalabru:
I think this issue was addressed with https://github.com/blockstack/stacks-blockchain/pull/2161, can we close @jcnelson ?

jcnelson:
Yup!

Issue #2141, opened 2020-12-08
Detecting miner commitments in window amid invalid commitments


kantai:
Currently, the miner commitment windowing requires that each linked UTXO be between valid commitments, however, this may end up punishing an honest miner when one commitment "misses" the intended Bitcoin block, e.g.,

commit 1 -> goes in burn block 1
commit 2 -> goes in burn block 2
commit 3 -> goes in burn block 3
commit 4 -> goes in burn block 5
commit 5 -> goes in burn block 5

"commit 4" _isn't_ a valid block commit, and because of that, the miner is significantly punished -- not only is that commitment lost, but "commit 5" doesn't get to claim commits 1->3 as part of its window.

Fixing this would require more leniency in the way we enforce UTXO chaining across the commitments.

jcnelson:
Hmm, what if we had a way for a block-commit to indicate its targeted burnchain block?  Then, we could automatically detect this bumping behavior and _not_ punish a miner that tried to issue the block-commits correctly.  Regardless of how the bumping happens, it will always be the case that:

* commit _i_ will be mined before commit _i+1_ in the transaction history, and
* if commit _i_ is mined in burn block _b_, then commit _i+1_ may be mined in any burn block _b + k >= b_

So to calculate the sortition weight for block commit _i_ sent by miner _M_, all we'd need to do is search back 6 burn blocks and find the last 6 block-commits _M_ sent, and verify that each of these 6 block-commits targeted a unique burn block in the last 6 burn blocks (even if they didn't get mined in them).  If this was the case, then miner _M_ was trying to be compliant and should be allotted a _mean(median - min)_ burn weight.

How can we let miners target a burn block?  We'd only need two bytes (really only 11 bits), with which we can encode the index into the reward cycle's reward set that was targeted (which can be between 0 and 2000).  From there, we can calculate which burn block the block-commit had meant to be mined in, and validate its burn/PoX outputs against that burn block (which we will always have already processed or will be in the middle of processing, per the above two invariants).

[...]

Issue #2125, opened 2020-12-03
VM: Expanded token type methods


jcnelson:
The fungible and non-fungible token types currently do not have enough expressive power to allow fungible token supplies to be adjusted or queried, nor do they allow non-fungible tokens to be destroyed or re-created.  After talking with the community, it's clear that our token types need these properties in order to implement real-world projects safely.  To address this, the VM will need the following new built-ins for tokens:

* `(ft-get-supply (token-name (string-ascii 256)))) ;; returns (option uint)`:  This method should take the name of the fungible token and return how many units of that token are in circulation.  It should return `none` if the token is not defined.  This function should work for all fungible tokens, even those that don't have a fixed maximum supply.

* `(ft-burn? (token-name (string-ascii 256)) (amount uint) (sender principal)) ;; returns (response bool uint)`: This method works like `ft-transfer?`, but it _destroys_ the `amount` of tokens given.  This causes the total circulating supply of the token to decrease, which will be reflected in a subsequent call to `ft-get-supply`.  Like `ft-transfer?`, it returns `(err u1)` if the `sender` doesn't have `amount` tokens to burn, and returns `(err u3)` if `amount` is non-positive.  Returns `(ok true)` on success.

* `(nft-burn? (asset-class (string-ascii 256)) (asset-identifier T) (sender principal)) ;; returns (response bool uint)`:  This method works like `nft-transfer?`, but it _destroys_ the asset identified by `asset-identifier`.  This allows a subsequent call to `nft-mint?` with the same `asset-class` and `asset-identifier` to succeed in re-creating the asset.  Like `nft-transfer?`, it returns `(err u1)` if `sender` doesn't own the asset, and `(err u3)` if the asset identified with `asset-class` and `asset-identifier` does not exist.  Returns `(ok true)` on success.

Like transfers, we'll want to make it so these methods keep track of tokens moved through a transaction's `AssetMap`, so post-conditions can still enforce safe token burns.

friedger:
To solve expiring or shared ownership of tokens, there needs to be another function, e.g. to address expired BNS names (https://github.com/blockstack/stacks-blockchain/pull/2013#discussion_r532512302)

psq:
even though I haven't revised the `flexr` implementation with these in mind, it seems that having a `burn` and `mint` function as specified above would make it possible to use an `ft-token` by minting or burning tokens to account for past rebases.

[...]

PR #2161, opened 2020-12-13
Feat/pob prepare phase

jcnelson:
Make the prepare phase proof-of-burn only (#2157), as part of fixing #1805.  It tweaks `PoxConstants` so that `reward_slots()` is the _difference_ between the reward cycle length and prepare phase length, which led to most of the size of the diff here (since it affected several tests in several different ways).  It also makes it so a block-commit doesn't pass checks if not all of its outputs are burns and it's picked up in the prepare phase (this might need to be tweaked once #2141 merges).

I can't get the Atlas test to pass in CI (it passes locally); ~I can wait until #2125 merges if you prefer.~ EDIT: still having trouble getting this to reliably pass in CI.

I will update SIP-007 once I have a full solution worked out for #1805.


src/chainstate/coordinator/mod.rs
```diff
@@ -407,7 +409,7 @@ pub fn get_reward_cycle_info<U: RewardSetProvider>(
             }));
         }
 
-        info!("Beginning reward cycle. block_height={}", burn_height);
+        info!("Beginning reward cycle. block_height={}, reward_cycle_length={}, prepare_phase_length={}", burn_height, burnchain.pox_constants.reward_cycle_length, burnchain.pox_constants.prepare_length);
```
kantai, line 412: Can you update this `info!` log to use keys/values format:

```suggestion
        info!("Beginning reward cycle";
              "burn_height" => burn_height,
              "reward_cycle_length" => burnchain.pox_constants.reward_cycle_length,
              "prepare_phase_length" => burnchain.pox_constants.prepare_length);
```

src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -453,6 +455,14 @@ impl LeaderBlockCommitOp {
             //   we want to make sure that any TX that forces us to perform the check
             //   has either burned BTC or sent BTC to the PoX recipients
 
+            // if we're in the prepare phase, then this block-commit _must_ burn.
+            if burnchain.is_in_prepare_phase(self.block_height) {
```
kantai, line 459: Should we treat the prepare phase block commits the same as we do post-sunset commits? (i.e., single burn output, rather than 2)

src/chainstate/burn/db/sortdb.rs
```diff
@@ -1103,6 +1103,11 @@ impl<'a> SortitionHandleTx<'a> {
             if let PoxAnchorBlockStatus::SelectedAndKnown(ref anchor_block, ref reward_set) =
                 next_pox_info.anchor_status
             {
+                test_debug!(
```
kantai, line 1106: Should this method explicitly check if it's in the prepare phase and return burn addresses? If I understand correctly, right now it's kind of implicit that the prepare phase is always burns, because the reward set has `(reward_cycle_length - prepare_phase_length)*2` slots.

kantai (approved):
This LGTM, just a couple small comments, and a question about how we want to treat the burn commits during the prepare phase (whether we should allow just one commit output or force it to use two).

src/chainstate/burn/operations/leader_block_commit.rs
jcnelson: I'm not sure.  Minimally, we only need to care that we're burning, and that we can measure the burn amount.  A miner ought to be able to create But per #2094, maybe it makes sense to avoid compelling miners to use a particular format?

If anything in this method needs to be more explicit and rigid, I think it should be avoiding a PoX descent check altogether if it's in the prepare phase (will fix).

kantai: That's fine with me -- the rationale for requiring exactly two commit outputs during the reward cycles is that it levels the playing field between PoX'ers and PoB'ers -- both are required to have the same tx size (modulo the inputs they choose). During the prepare phase, that rationale doesn't apply any more: everyone is burning. So in theory, we could allow a single commit output. But we can also delay this change to a separate PR.

src/chainstate/burn/db/sortdb.rs
jcnelson: I went ahead and made this check explicit

src/chainstate/burn/operations/leader_block_commit.rs
jcnelson: I went ahead and made it so a prepare-phase block-commit must have a single burn output

src/chainstate/coordinator/mod.rs
jcnelson: Should be fixed now.

src/chainstate/stacks/boot/pox-testnet.clar
```diff
@@ -7,7 +7,7 @@
 (define-constant PREPARE_CYCLE_LENGTH u30)
 
 ;; Default length of the PoX reward cycle, in burnchain blocks.
-(define-constant REWARD_CYCLE_LENGTH u120)
+(define-constant REWARD_CYCLE_LENGTH u150)
```
lgalabru, line 10: Should we have the same name for our constants in Clarity and Rust? (`REWARD_CYCLE_LENGTH ` vs `POX_ REWARD_CYCLE_LENGTH`)?

lgalabru (approved):
Looks good to me!

PR #2173, opened 2020-12-14
Mainnet compatibility fixes

lgalabru:
This PR aims at making the `stacks-node` compatible with Bitcoin Mainnet.
The activity of the miner I'm using can be checked here: https://mempool.space/address/18aKMk7LfoVn9T6gAhhyA8hcVTDKvJ4YG6


testnet/stacks-node/src/config.rs
```diff
@@ -14,8 +14,11 @@ use stacks::util::secp256k1::Secp256k1PublicKey;
 use stacks::vm::costs::ExecutionCost;
 use stacks::vm::types::{AssetIdentifier, PrincipalData, QualifiedContractIdentifier};
 
-use super::neon_node::TESTNET_PEER_VERSION;
-use super::node::TESTNET_CHAIN_ID;
+pub const TESTNET_CHAIN_ID: u32 = 0x80000000;
+pub const TESTNET_PEER_VERSION: u32 = 0xfacade01;
+
+pub const MAINNET_CHAIN_ID: u32 = 0x00000001;
+pub const MAINNET_PEER_VERSION: u32 = 0xfacade01;
```
jcnelson, line 21: This should be the node's semantic version, e.g. `0x18000000` (represting `24.0.0.0`).  The software already treats the highest byte in the version as a major version number, and considers two version values equal if their highest bytes match for the purposes of hashing and equality (e.g. `0x18000000` == `0x18ffffff`).

testnet/stacks-node/src/config.rs
```diff
@@ -878,6 +970,8 @@ pub struct BurnchainConfigFile {
     pub chain: Option<String>,
     pub burn_fee_cap: Option<u64>,
     pub mode: Option<String>,
+    pub chain_id: Option<String>,
+    pub peer_version: Option<String>,
```
jcnelson, line 974: This probably shouldn't be configurable.

testnet/stacks-node/src/neon_node.rs
```diff
@@ -178,12 +187,19 @@ fn inner_generate_poison_microblock_tx(
...
     poison_payload: TransactionPayload,
+    is_mainnet: bool,
+    chain_id: u32,
 ) -> StacksTransaction {
     let mut tx_auth = keychain.get_transaction_auth().unwrap();
     tx_auth.set_origin_nonce(nonce);
 
-    let mut tx = StacksTransaction::new(TransactionVersion::Testnet, tx_auth, poison_payload);
-    tx.chain_id = TESTNET_CHAIN_ID;
+    let version = if is_mainnet {
```
jcnelson, line 196: Maybe we should implement `From<bool>` for `TransactionVersion` -- it feels like we use this sort of `if/else` a lot.

lgalabru: As in converting `mainnet` in `true` and `testnet` in `false`?
It feels like we'd been loosing in clarity.

testnet/stacks-node/src/config.rs
lgalabru: Address with https://github.com/blockstack/stacks-blockchain/pull/2173/commits/a7c78d4e28d380466449a8734056b0d06844f733

testnet/stacks-node/src/config.rs
lgalabru: Address with https://github.com/blockstack/stacks-blockchain/pull/2173/commits/a7c78d4e28d380466449a8734056b0d06844f733

Speaking of these `24.0.0.0` versions (also used for the DB versions), should we do a pass and reset to `1.0.0.0`, or `2.0.0.0`?

src/core/mod.rs
```diff
@@ -50,9 +50,21 @@ pub const INITIAL_MINING_BONUS_WINDOW: u16 = 10_000;
 // first burnchain block hash
 // TODO: update once we know the true first burnchain block
 pub const FIRST_BURNCHAIN_CONSENSUS_HASH: ConsensusHash = ConsensusHash([0u8; 20]);
-pub const FIRST_BURNCHAIN_BLOCK_HASH: BurnchainHeaderHash = BurnchainHeaderHash([0u8; 32]);
-pub const FIRST_BURNCHAIN_BLOCK_HEIGHT: u32 = 0;
-pub const FIRST_BURNCHAIN_BLOCK_TIMESTAMP: u64 = 0;
+
+pub const BITCOIN_MAINNET_FIRST_BLOCK_HEIGHT: u64 = 661389;
```
jcnelson, line 54: This is still really TBD, right?  We won't know what this value needs to be until the `.miners` namespace is fully populated.

testnet/stacks-node/src/config.rs
```diff
@@ -273,6 +276,48 @@ impl ConfigFile {
...
+            ..BurnchainConfigFile::default()
+        };
+
+        let node = NodeConfigFile {
+            bootstrap_node: Some("047435c194e9b01b3d7f7a2802d6684a3af68d05bbf4ec8f17021980d777691f1d51651f7f1d566532c804da506c117bbf79ad62eea81213ba58f8808b4d9504ad@mainnet.blockstack.org:20444".to_string()),
+            miner: Some(false),
+            ..NodeConfigFile::default()
+        };
+
+        let balances = vec![
```
jcnelson, line 294: If this is the function for booting the node to use mainnet, then shouldn't this be an empty vector?  I.e. all STX tokens should come from the genesis block, right?

testnet/stacks-node/src/neon_node.rs
jcnelson: Yeah, you're right.  Let's leave it as-is.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -2128,7 +2130,7 @@ fn atlas_integration_test() {
     // Poll GET v2/attachments/<attachment-hash>
     for i in 1..10 {
         let mut attachments_did_sync = false;
-        let mut timeout = 30;
+        let mut timeout = 60;
```
jcnelson, line 2133: This might need to be longer -- I've had to set it to 120 sometimes.

jcnelson (approved):
LGTM!

testnet/stacks-node/src/tests/neon_integrations.rs
kantai: A related commit is this one (https://github.com/blockstack/stacks-blockchain/commit/40908325be9449afe95b93fbe50b9153f93e1164) -- I think it may fix the other neon_integrations flakiness (noted here: https://github.com/blockstack/stacks-blockchain/issues/2207). Cherry picking that commit into other branches will probably fix the gh action flake.

src/core/mod.rs
kantai: Yep -- this should have a greppable comment like

```
// TODO: TO BE SET BY STACKS_V1_MINER_THRESHOLD
```

kantai (approved):
These changes look good to me. My only feedback is covered by Jude's: add a comment for setting the first block height, and the `mainnet()` config should only read values from the genesis import.

testnet/stacks-node/src/config.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2173/commits/dbe8f0a60efc8d6cb6844d466422018847724ef3

src/core/mod.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2173/commits/dbe8f0a60efc8d6cb6844d466422018847724ef3

PR #2177, opened 2020-12-15
Commit window fixes

kantai:
This PR addresses #2141, by introducing a "burn parent modulus" to allow the stacks-node to detect when a commitment may be missed. This PR also updates the smoothing function to:

```
min(last_burn, median)
```

Which is less punitive than `mean(min, median)`

## Type of Change
- Bug fix

Yes, this changes the wire format for block commits (change is documented in SIP-001), which would lead to a consensus-breaking change.

Documentation updates are probably required -- any mining references we have should be updated to reflect that the smoothing function is changing from

`mean(min, median)` to `min(last_burn, median)`

## Testing information

There's a few different ways that this is tested:

1. Changes to `distribution` are unit tested in that module.
2. Changes to the parsing of LeaderBlockCommitOp are tested in that module
3. An integration test for missed block commits is included in the chainstate::coordinator tests. This required adding `cfg(test)` enabled storage of burn distributions so that the evaluated distribution could be checked after a burn block is processed.


src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -168,7 +190,7 @@ impl LeaderBlockCommitOp {
         let parent_vtxindex = parse_u16_from_be(&data[68..70]).unwrap();
         let key_block_ptr = parse_u32_from_be(&data[70..74]).unwrap();
         let key_vtxindex = parse_u16_from_be(&data[74..76]).unwrap();
-        let memo = data[76..77].to_vec();
+        let burn_parent_modulus = ((data[76] as u64) % BURN_BLOCK_MINED_AT_MODULUS) as u8;
```
jcnelson, line 193: Can you add a bitmask here so that the burn parent modulus only comes from the lowest 3 bits?  This lets use use the remaining 5 bits as a memo field (which may be required for addressing #1805).

src/burnchains/burnchain.rs
```diff
@@ -111,6 +111,7 @@ impl BurnchainStateTransition {
         sort_tx: &mut SortitionHandleTx,
         parent_snapshot: &BlockSnapshot,
         block_ops: &Vec<BlockstackOperationType>,
+        missed_commits: &Vec<MissedBlockCommit>,
```
jcnelson, line 114: These are the `MissedBlockCommits` that showed up in _this_ block, right?  i.e. they were destined for some earlier block up to `BURN_BLOCK_MINED_AT_MODULUS` blocks ago, but they were bumped here?

jcnelson (approved):
This LGTM, but please add the bitmask in the block-commit before merging.  Thanks!

src/burnchains/burnchain.rs
kantai: Yep

src/chainstate/burn/operations/leader_block_commit.rs
kantai: Sure

kantai: Added in 5972a533a

src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -190,7 +191,12 @@ impl LeaderBlockCommitOp {
         let parent_vtxindex = parse_u16_from_be(&data[68..70]).unwrap();
         let key_block_ptr = parse_u32_from_be(&data[70..74]).unwrap();
         let key_vtxindex = parse_u16_from_be(&data[74..76]).unwrap();
-        let burn_parent_modulus = ((data[76] as u64) % BURN_BLOCK_MINED_AT_MODULUS) as u8;
+
+        let burn_parent_modulus_and_memo_byte = data[76];
+
+        let burn_parent_modulus = ((burn_parent_modulus_and_memo_byte & 0b111) as u64
+            % BURN_BLOCK_MINED_AT_MODULUS) as u8;
+        let memo = (burn_parent_modulus_and_memo_byte >> 3) & 0x1f;
```
jcnelson, line 199: The `& 0x1f` shouldn't be necessary, strictly speaking, because it's a `u8`.  If you wanted to be explicit about the masking though (which I support), you could do it as `(burn_parent_modulus_and_memo_byte & 0b11111000) >> 3`.

kantai: Yeah -- I don't think it's necessary, it's added here because the first source I found for rust's shift right operand said that it would shift in the high bit (i.e., `0b1111 >> 1 = 0b1111` rather than `0b0111`), but that's not the case -- right shift on uints shifts in `0` (`(burn_parent_modulus_and_memo_byte & 0b11111000) >> 3` wouldn't insure that 0 bits were shifted in).

lgalabru (approved):
Looks good to me!

PR #2178, opened 2020-12-15
BNS: read-only method for name prices

lgalabru:
This PR is addressing a request from @yknl - we essentially want to expose a readonly method that return a price, given a namespace and a name (it previously required to pass the price function).
Merging next into master produced a new build of the clarity js sdk (thank you @CharlieC3 for the automation!), this PR is updating the BNS js client used for the tests + fixing an issue in the contract +  adding some more tests covering namespace / name price calculations.


kantai (approved):
LGTM!

jcnelson (approved):
LGTM!

PR #2180, opened 2020-12-15
Implement transaction fees

jcnelson:
This PR adds transaction fees to the miner's block reward.  Per SIP-001, each miner gets all of its anchored block's transaction fees, as well as 40% of the microblocks it produces (awarded in each Stacks fork in which a subsequent anchored block confirms them), as well as 60% of the microblocks it confirms.  It also renames `fee_rate` to `tx_fee`, now that we're making it so transactions pay the fee they advertise (instead of a Kth-highest fee per epoch).


kantai (approved):
This LGTM!

lgalabru (approved):

PR #2184, opened 2020-12-16
Fix: chainstate import discrepancy

lgalabru:


jcnelson (approved):

jcnelson:
However, this is a consensus-breaking change -- we'll want to re-instantiate the Xenon follower and miner.

jcnelson:
Another thing occurs to me -- in addition to having the `chainstate.sha256`, we might also want to consider storing the root hash of the genesis block's MARF trie, and comparing the resulting MARF trie against it once the node finishes booting up. The node should crash if they don't match, since it would indicate non-determinism.

kantai (approved):
This looks good to me, but we should make sure that the bitcoin tests pass on CI, since this change could increase memory usage during genesis (the `sorted_keys` vec could be large?)

src/chainstate/stacks/db/mod.rs
```diff
@@ -962,6 +962,7 @@ impl StacksChainState {
             if let Some(get_schedules) = boot_data.get_bulk_initial_lockups.take() {
                 info!("Initializing chain with lockups");
                 let mut lockups_per_block: HashMap<u64, Vec<Value>> = HashMap::new();
```
kantai, line 964: Nitpick: can we just replace `lockups_per_block: HashMap<_, _> = HashMap::new()` with `BTreeMap` ?

`BTreeMap`s are sorted by key, so we wouldn't need a secondary structure here. It may be more efficient, though it's probably not important, because the expensive thing here is all the MARF edits, not this outer loop.

lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2184/commits/7ea12d1cc4474e731294fdaa39a530db540f7522

PR #2188, opened 2020-12-17
Test: add coverage for genesis consistency

lgalabru:
This PR is testing the predictability of the chainstate process, by ensuring that we're always reproducing the same MARF root hash.

### DIFF
--- a/src/chainstate/stacks/db/mod.rs
+++ b/src/chainstate/stacks/db/mod.rs
@@ -1822,7 +1822,7 @@ pub mod test {
     }
 
     #[test]
-    fn test_chainstate_genesis_consistency() {
+    fn test_chainstate_test_genesis_consistency() {
         // Test root hash for the test chainstate data set
         let mut boot_data = ChainStateBootData {
             initial_balances: vec![],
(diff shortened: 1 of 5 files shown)


src/chainstate/stacks/db/mod.rs
```diff
@@ -1819,4 +1820,125 @@ pub mod test {
...
+            marf.get_root_hash_at(&index_block_hash).unwrap()
+        });
+        assert_eq!(
+            format!("{}", genesis_block),
+            "e336d28e3bccdf8e5e77216e0dea26b60292f4577b05ab2d4414776f29a5e9fc"
+        );
+    }
+
+    #[test]
+    #[ignore]
```
jcnelson, line 1884: Why is this `ignore`?  Does it take a long time to run?

lgalabru: Yeah I had to mark this test as ignored in https://github.com/blockstack/stacks-blockchain/pull/2188/commits/5638835b9177a84964432a001dc7769d39348f6f after OOM'ing github's runner

kantai (approved):
This looks good to me -- just add a note to that test case that it would need to be updated in any PR that modifies the genesis data.

jcnelson (approved):
LGTM!

PR #2189, opened 2020-12-17
Add lockup events

lgalabru:
Per @zone117x request, we're generating a receipt with the lockup events inserted at genesis time.

### DIFF
--- a/src/chainstate/stacks/db/mod.rs
+++ b/src/chainstate/stacks/db/mod.rs
@@ -988,12 +988,26 @@ impl StacksChainState {
                         .with_clarity_db(|db| {
                             for (block_height, schedule) in lockups_per_block.into_iter() {
                                 let key = Value::UInt(block_height.into());
+                                let value = Value::list_from(schedule).unwrap();
                                 db.insert_entry(
                                     &lockup_contract_id,
                                     "lockups",
-                                    key,
-                                    Value::list_from(schedule).unwrap(),
+                                    key.clone(),
+                                    value.clone(),
                                 )?;
+
+                                // Attach some events
+                                let lookup_event = TupleData::from_data(vec![
+                                    (ClarityName::try_from("block-height").unwrap(), key),
+                                    (ClarityName::try_from("due-schedules").unwrap(), value)
+                                ]).unwrap();
+                                let lookup_event = StacksTransactionEvent::SmartContractEvent(
+                                    SmartContractEventData {
+                                        key: (boot_code_id("lockup"), "print".to_string()),
+                                        value: Value::Tuple(lookup_event)
+                                    }
+                                );
+                                allocation_events.push(lookup_event);
                             }
                             Ok(())
                         })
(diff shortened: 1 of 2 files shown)


src/chainstate/stacks/db/mod.rs
```diff
@@ -1001,12 +1003,27 @@ impl StacksChainState {
...
+                                    (ClarityName::try_from("due-schedules").unwrap(), value),
+                                ])
+                                .unwrap();
+                                let lookup_event = StacksTransactionEvent::SmartContractEvent(
+                                    SmartContractEventData {
+                                        key: (boot_code_id("lockup"), "print".to_string()),
+                                        value: Value::Tuple(lookup_event),
+                                    },
+                                );
+                                lockup_events.push(lookup_event);
```
jcnelson, line 1026: So, these are all "synthetic" events -- they don't correspond to a real transaction on the blockchain.  This could confuse users and block explorers -- they'll see a txid, but be unable to query it from the chainstate (we've had this problem before in Stacks 1.0 with "synthetic transactions" that achieved the same end).

Would it make sense to add a field in the `StacksTransactionEvent` struct to indicate whether or not the event came from a synthetic transaction, so people don't mistakenly believe that these genesis events correspond to on-chain transactions?

jcnelson: No need to address it in this PR if you're pressed for time, but please do open an issue for it if you're not going to address it here.

jcnelson (approved):
LGTM!

src/chainstate/stacks/db/mod.rs
jcnelson: Also, any insight into how big this list will be in RAM?

zone117x: I share the concern about these transactions and events being odd one-offs, however, the event broadcaster attaches these to the initial block in a way that does not require any special handling by the sidecar/explorer. We've already been doing this for a while in order to get the boot contracts exposed in a normalized way. 

I can imagine how an `is_synthetic` field could be useful, but I don't think it's worth the effort at the moment.

lgalabru: > Also, any insight into how big this list will be in RAM?

It's a couple of megabytes on disk, not sure about the representation in RAM.

zone117x: I'm doing some benchmarks on memory usage

zone117x: This PR appears to increase max memory usage from ~580MB to ~700MB. Measured a few times with 
`time -l stacks-node mocknet` and looking at the reported `maximum resident set size`.

jcnelson: That doesn't sound too bad.

zone117x: Also of note: with an event observer attached, memory usage is ~2.6GB. The content-size of the first block event payload is ~100MB.

PR #2190, opened 2020-12-17
Allow small test genesis file to be used in mocknet mode

zone117x:
Closes https://github.com/blockstack/stacks-blockchain/issues/2176

### DIFF
--- a/testnet/stacks-node/src/node.rs
+++ b/testnet/stacks-node/src/node.rs
@@ -184,7 +184,11 @@ impl Node {
     pub fn new(config: Config, boot_block_exec: Box<dyn FnOnce(&mut ClarityTx) -> ()>) -> Self {
         let use_test_genesis_data = if config.burnchain.mode == "mocknet" {
             // When running in mocknet mode allow the small test genesis chainstate data to be enabled.
-            env::var("BLOCKSTACK_USE_TEST_GENESIS_CHAINSTATE") == Ok("1".to_string())
+            if env::var("BLOCKSTACK_USE_TEST_GENESIS_CHAINSTATE") == Ok("1".to_string()) {
+                true
+            } else {
+                USE_TEST_GENESIS_CHAINSTATE
+            }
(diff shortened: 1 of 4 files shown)


testnet/stacks-node/src/node.rs
```diff
@@ -181,6 +182,17 @@ fn spawn_peer(
 impl Node {
     /// Instantiate and initialize a new node, given a config
     pub fn new(config: Config, boot_block_exec: Box<dyn FnOnce(&mut ClarityTx) -> ()>) -> Self {
+        let use_test_genesis_data = if config.burnchain.mode == "mocknet" {
+            // When running in mocknet mode allow the small test genesis chainstate data to be enabled.
+            if env::var("BLOCKSTACK_USE_TEST_GENESIS_CHAINSTATE") == Ok("1".to_string()) {
```
jcnelson, line 187: Instead of making this an envar, could we add this as a config option?

zone117x: We could, although env vars are easier to deal with in the tooling than config files. How strongly do you prefer config file vs env var?

friedger: It would be nice for the competition to set this as a config option

zone117x: Are we okay with allowing both methods? Config or env var

zone117x: @jcnelson @friedger Please see latest commit where support for both config toml and env var is added.

friedger: Great

kantai (approved):
LGTM!

testnet/stacks-node/src/node.rs
jcnelson: Both are fine.

jcnelson (approved):
LGTM! Thanks @zone117x

PR #2192, opened 2020-12-17
GH Action for Unit Tests with Code Coverage

kantai:
This is a draft PR for doing our unit tests in a github action, with code-coverage (these are the tests currently run on circle).

Closes https://github.com/blockstackpbc/devops/issues/539


CharlieC3:
@kantai Any reason you switched to using `grcov` in GH Actions from `tarpaulin` in CircleCI?

kantai:
I've gone back and forth between grcov and tarpaulin a bunch -- the benefit of grcov is that it lets you basically use the "normal" cargo and rustc build, which is useful for us because we use cargo workspaces, which didn't always work well in tarpaulin.

CharlieC3:
@kantai Was able to get further. No more compilation errors since [switching to LLVM source-based coverage](https://github.com/mozilla/grcov#example-how-to-generate-source-based-coverage-for-a-rust-project), but there might be a legitimate failing test:
https://github.com/blockstack/stacks-blockchain/runs/1602179356?check_suite_focus=true#step:3:1443

kantai:
Interesting -- LLVM source-based coverage looks awesome.

I don't know what's going on with that test failure -- I've never seen that particular test fail. I wonder if the stack depth is affected by the llvm coverage hooks? We could probably mark that test as ignored, and do something like we used to do with other ignored tests, where the "code coverage" pass runs with `cargo test`, and then we have another `all_units` pass which doesn't collect coverage information and runs with `cargo test --ignored --test-threads 1`.

kantai:
@CharlieC3 -- it looks like that test was allocating large buffers on the stack, which with the llvm instrumentation was causing a stack overflow. Switching to `vec` from `Box<[]>` remedied that situation though. Now it looks like the unit test action is just failing because of the output copying commands (I think those would need to change anyways to upload to codecov.io).

CharlieC3:
@kantai Cool, good catch. Will test a fix for the copying commands locally then push it up.

CharlieC3:
@kantai I rebased from the `next` branch and squashed all the commits to clean things up. It seems there are some actual test failures right now. I can revisit this after test fixes are merged into `next`.

kantai:
Yes -- there's a codecov account for the `blockstack` github organization, I think it just requires signing in via github.

kantai:
This LGTM! I can't approve this PR, because apparently I authored it :)

lgalabru (approved):
Nice! Looks good to me, thank you guys!

kantai:
Merging!

PR #2196, opened 2020-12-19
Feat/miner

psq:
This PR is extracting the basic mining logic from stacks-node (with more to come, see what's left to do section below).  The miner's code can then be improved more easily by miners to match the mining strategy they think would be most profitable.  It is also a lot easier to restart the miner without having to restart the stacks-node.

This PR introduces a new API endpoint (/v2/miner/build-block), which allows a miner to assemble a block with the transactions the miner would like to include (or let the miner decide for simple miners by passing an empty transaction list)

Implements https://github.com/blockstack/stacks-blockchain/issues/1969.

## Design guidelines
- don't duplicate data storage, so rely on what's in stacks-node instead
- keep private keys in miner (this worked well till we got a working implementation for microblocks, see below)
- miner sends its own transactions to BTC network

## Miner's logic flow
- generate and send leader key registration to btc network
- main loop
  - detect new tip/new btc block
  - generate a new vrf proof
  - generate new microblock keys
  - generate coinbase transaction
  - gather transactions to use (optional)
  - call `/v2/miner/build-block` endpoint to generate a new block
  - generate and send leader commit to btc network

The stacks-node will keep the generated block via `/v2/miner/build-block` and in case the block wins the next sortition, it will send it to the stacks network (nicely avoids adding the whole p2p layer to the miner)

## What's not covered
- Fork selection is primitive, just trying to mine off the most recent known tip
- Does not use a recommended fee service to help get block commit get into the next block
- Does not consider whether mining is profitable
- No fancy strategy trying to optimize how much to commit each block, just a constant value
- For more, see next section

No.

New endpoint needs to be documented
How to write a miner

## Left to do
- [ ] integration test(s?) with bitcoind, stacks-node and one, or more, miner(s)
- [ ] cleanup println!/replace with debug when useful
- [ ] cleanup TODO(psq) no longer relevant, commented out code
- [ ] some code could be better shared with stacks-node (by moving some code to a lib, either blockstack-core, or a new one, for things like `inner_generate_coinbase_tx`, or `keychain.rs`)
- [ ] run `cargo fmt`
- [ ] rebase one more time
- [ ] better tip selection strategy to avoid forks (separate PR?)
- [ ] improve configuration (either more options, or use a configuration file, so testnet value hardcoded)
- [ ] create a new issue to add the RBF logic currently in stacks-node (currently, the miner will not mine if no stacks block is received for the current btc tip)
- [ ] create a new issue to track new transactions & blocks (setting up the miner as an event obsever rather than have to rely on a separate api-node, and having to repeatedly query for transactions).  This will also help properly implement microblocks
- [ ] create a new issue to properly deal with microblocks as a separate PR so the private key does not need to sent to stacks-node
- [ ] create a new issue to create a node.js bridge so writing a mining strategy does not require knowning rust (possibly other languages?)
- [ ] create a new doc issue for using the new miner, and extend the code

cc @tyGavinZJU


src/chainstate/burn/db/sortdb.rs
```diff
@@ -1794,6 +1794,106 @@ impl<'a> SortitionHandleConn<'a> {
 
         Ok(result)
     }
+
+    pub fn get_last_anchor_block_hashPSQ(&self) -> Result<Option<BlockHeaderHash>, db_error> {
```
psq, line 1798: forgot about those, any reason why the original ones use a `&mut self` while it was not needed, which makes it impossible to call these without a mutable db ref.
So I added mines, obviously to be renamed, but I would think the mutable flavor should be removed, unless I'm missing something rather obvious

testnet/stacks-miner/tx.hex
```diff
@@ -0,0 +1 @@
...726c6420636f6e74726163740a0a28646566696e652d636f6e7374616e742073656e6465722027535a324a365a593438475631455a35563256355242394d5036365357383650594b4b51394836445052290a28646566696e652d636f6e7374616e7420726563697069656e742027535431584e4846455856374b5732455045464a34514b4d455445455358484a47454d39514d455a3741290a0a28646566696e652d66756e6769626c652d746f6b656e206e6f76656c2d746f6b656e2d3139290a28626567696e202866742d6d696e743f206e6f76656c2d746f6b656e2d3139207531322073656e64657229290a28626567696e202866742d7472616e736665723f206e6f76656c2d746f6b656e2d31392075322073656e64657220726563697069656e7429290a0a28646566696e652d6e6f6e2d66756e6769626c652d746f6b656e2068656c6c6f2d6e66742075696e74290a28626567696e20286e66742d6d696e743f2068656c6c6f2d6e66742075312073656e64657229290a28626567696e20286e66742d6d696e743f2068656c6c6f2d6e66742075322073656e64657229290a28626567696e20286e66742d7472616e736665723f2068656c6c6f2d6e66742075312073656e64657220726563697069656e7429290a0a28646566696e652d7075626c69632028746573742d656d69742d6576656e74290a2020202028626567696e0a2020202020202020287072696e7420224576656e74212048656c6c6f20776f726c6422290a2020202020202020286f6b2075312929290a28626567696e2028746573742d656d69742d6576656e7429290a
```
psq, line 1: need to remove

psq:
closing in favor of a new, simpler PR

Issue #2045, opened 2020-11-09
DB: merge Clarity side-store into its state DB, and remove savepoints


jcnelson:
Clarity uses a side-store database independent of its MARF'ed database.  These should really be stored to the same database, since it would remove a point of crash-consistency failure.  But in order to do this, we should remove the savepoint logic currently used to execute a block's-worth of Stacks transactions, and instead use the MARF'ed `begin`, `commit`, and `rollback` functions to instantiate, store, and abort bulk operations.

kantai:
This is the same issue as (or rather, this is the planned fix for) https://github.com/blockstack/stacks-blockchain/issues/1939, right?

jcnelson:
Yes; we can close #1939 in favor of this if you want.  If we do, then this issue should also be updated to reflect the need for adding WAL mode.  I think this also fixes my concerns in #1992 about concurrent RPC calls endpoint potentially mucking things up in the Clarity DB.

PR #2199, opened 2020-12-20
Clarity DB Fixes

kantai:
This addresses #2045 -- removing the savepoints, and using the same DB transaction to build MARF tries as "side-store" writes.

The primary changes here are to `vm::database::marf` (which propagate to `vm::clarity`):

* In order to be used, a caller of the `MarfedKV` struct needs to create either a Writable or a ReadOnly db connection. The Writable connection wraps a `MarfTransaction` instance -- this is used to ensure that a ClarityBlockConnection always corresponds with one MarfTransaction (guaranteeing atomicity of side-store operations, importantly the _metadata_ side store operations).
* Unconfirmed state still requires the ability to "undo" metadata side store operations. This is because unconfirmed state may be "sqlite committed", and then later dropped. This is implemented by doing a SQL delete from during the unconfirmed block rollback.

### DIFF
--- a/.github/workflows/stacks-blockchain.yml
+++ b/.github/workflows/stacks-blockchain.yml
@@ -35,22 +35,22 @@ jobs:
           status: STARTING
           color: warning
 
-  # Run large genesis test
-  large-genesis:
+  # Run full genesis test
+  full-genesis:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Run bitcoin tests
+      - name: Single full genesis integration test
         env:
           DOCKER_BUILDKIT: 1
         run: docker build -f ./.github/actions/bitcoin-int-tests/Dockerfile.large-genesis .
 
-  # Run bitcoind tests
-  bitcoin-tests:
+  # Run sampled genesis tests
+  sampled-genesis:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Run bitcoin tests
+      - name: All integration tests with sampled genesis
         env:
           DOCKER_BUILDKIT: 1
         run: docker build -f ./.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests .
@@ -166,8 +166,8 @@ jobs:
     outputs:
       upload_url: ${{ steps.create_release.outputs.upload_url }}
     needs:
-      - bitcoin-tests
-      - large-genesis
+      - sampled-genesis
+      - full-genesis
       - dist
       - build-publish
       - build-publish-stretch
@@ -219,8 +219,8 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - notify-start
-      - bitcoin-tests
-      - large-genesis
+      - sampled-genesis
+      - full-genesis
       - dist
       - build-publish
       - build-publish-stretch
--- a/.github/workflows/stacks-blockchain.yml
+++ b/.github/workflows/stacks-blockchain.yml
@@ -35,8 +35,18 @@ jobs:
           status: STARTING
           color: warning
 
-  # Run tests
-  test:
+  # Run large genesis test
+  large-genesis:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run bitcoin tests
+        env:
+          DOCKER_BUILDKIT: 1
+        run: docker build -f ./.github/actions/bitcoin-int-tests/Dockerfile.large-genesis .
+
+  # Run bitcoind tests
+  bitcoin-tests:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -156,7 +166,8 @@ jobs:
     outputs:
       upload_url: ${{ steps.create_release.outputs.upload_url }}
     needs:
-      - test
+      - bitcoin-tests
+      - large-genesis
       - dist
       - build-publish
       - build-publish-stretch
@@ -208,7 +219,8 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - notify-start
-      - test
+      - bitcoin-tests
+      - large-genesis
       - dist
       - build-publish
       - build-publish-stretch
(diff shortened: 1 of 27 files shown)


src/chainstate/coordinator/comm.rs
```diff
@@ -60,9 +60,6 @@ pub struct CoordinatorChannels {
     stacks_blocks_processed: Arc<AtomicU64>,
     /// how many sortitions have been processed by this Coordinator thread since startup?
     sortitions_processed: Arc<AtomicU64>,
-    /// Horrific kludgy clarity DB lock that should be removed once the Clarty DB doesn't
-    /// runtime-panic when two threads write to it.
-    kludgy_temporary_clarity_db_lock: Arc<Mutex<()>>,
```
jcnelson, line 65: :tada:

src/net/chat.rs
```diff
@@ -1811,6 +1811,13 @@ impl ConversationP2P {
                         break;
                     }
                 }
+                Err(net_error::PermanentlyDrained) => {
```
jcnelson, line 1814: I'd be curious to know what kinds of shenanigans you encountered that prompted you to debug connection closures :thinking:

src/net/p2p.rs
```diff
@@ -4276,7 +4276,15 @@ impl PeerNetwork {
         }
 
         if let Err(e) = PeerNetwork::setup_unconfirmed_state(chainstate, sortdb) {
-            warn!("Failed to instantiate unconfirmed state: {:?}", &e);
+            if let net_error::ChainstateError(ref err_msg) = e {
+                if err_msg == "Stacks chainstate error: NoSuchBlockError" {
```
jcnelson, line 4280: This makes me wonder if there should just be a dedicated error type for this particular condition.

src/vm/database/marf.rs
```diff
@@ -709,3 +660,233 @@ impl ClarityBackingStore for NullBackingStore {
...
+                    );
+                    Ok(None)
+                }
+                _ => Err(e),
+            })
+            .expect("ERROR: Unexpected MARF Failure on GET")
+            .map(|marf_value| {
+                let side_key = marf_value.to_hex();
+                trace!("MarfedKV get side-key for {:?}: {:?}", key, &side_key);
+                SqliteConnection::get(self.marf.sqlite_tx(), &side_key).expect(&format!(
```
jcnelson, line 789: Just so I understand -- this is a DB transaction instead of a normal connection in order to ensure that a concurrent read and write (e.g. from two separate threads in the node) are always serialized?

src/vm/database/marf.rs
```diff
@@ -709,3 +660,233 @@ impl ClarityBackingStore for NullBackingStore {
...
+        AnalysisDatabase::new(self)
+    }
+
+    pub fn rollback_block(self) {
+        self.marf.drop_current();
+    }
+
+    pub fn rollback_unconfirmed(self) {
+        SqliteConnection::drop_metadata(self.marf.sqlite_tx(), &self.chain_tip);
+        self.marf.drop_unconfirmed();
```
jcnelson, line 683: This all happens within the context of a single DB transaction, right?

src/vm/database/marf.rs
jcnelson: Or, am I misreading this, and both the `marf.get()` and the `SqliteConnection::get()` happen within the same transaction already?

src/net/chat.rs
kantai: Just trying to reduce some excessive log verbosity -- this PermanentlyDrained error was getting logged as `info` level, while not really conveying any useful information.

src/net/p2p.rs
kantai: Perhaps, but I'd leave refactoring this error variant to a different PR (this would be less hacky if the `net_error::ChainstateError` variant took a `ChainstateError` enum rather than a String).

src/vm/database/marf.rs
kantai: > both the marf.get() and the SqliteConnection::get() happen within the same transaction already?

Yep, that's correct --- `marf.sqlite_tx()` here is just returning a pointer to the currently open transaction. The `WritableMarf` struct here shares the same lifetime as the underlying SQLite transaction. Every read and write on the same instance of that struct shares a single transaction. That transaction is used not only for serialization, but atomization of the side store operations with the marf operations (it ensures that when the trie commits, the side store commits as well).

src/vm/database/marf.rs
kantai: Yep -- per my previous comment, the `MarfTransaction` instance held by this struct represents a single DB transaction, so both `drop_metadata` and `drop_unconfirmed` happen in the same transaction (as do all the prior operations applied during the lifetime of the struct).

For unconfirmed tries (unlike the anchored block tries), this can leave the side store _somewhat_ dirty, because unconfirmed tries need to commit (the anchored block tries just abort the transaction: the entirety of their writes exist in one sqlite tx) -- there would be left-over side store data, but not any metadata (because it's dropped here). The left-over side store data is safe, because that data is just mapping `hash(value) -> value` (though it is eating up storage space), but adding garbage collection there can be handled later (it won't affect consensus).

kantai:
Pull request authors can't approve their own PRs, but @jcnelson, these commits look good to me.

.github/workflows/stacks-blockchain.yml
```diff
@@ -35,8 +35,18 @@ jobs:
           status: STARTING
           color: warning
 
-  # Run tests
-  test:
+  # Run large genesis test
+  large-genesis:
```
lgalabru, line 39: 💅  what's our take on the semantic `full-genesis` vs `sampled-genesis`/`light-genesis`?

src/net/chat.rs
jcnelson: Ah.

src/vm/database/marf.rs
jcnelson: Thanks for explaining!  As long as there aren't any rows in the side-store that could conflict with subsequent unconfirmed (or confirmed) state, then this is fine.  Can you open an issue to add garbage-collection for unconfirmed side-store state?

jcnelson (approved):
LGTM!  Thanks @kantai!  Lots and lots of users are going to appreciate this :D

lgalabru (approved):

.github/workflows/stacks-blockchain.yml
kantai: That works for me -- I'll update the names in the workflow.

PR #2204, opened 2020-12-22
Atlas & BNS: Garbage collection, BNS constants + BNS genesis data

lgalabru:
This PR is addressing https://github.com/blockstack/stacks-blockchain/issues/2083, https://github.com/blockstack/stacks-blockchain/issues/2135 and https://github.com/blockstack/stacks-blockchain/issues/2165.

Once we merge https://github.com/blockstack/stacks-blockchain/pull/2188, I'll update this PR with the new trie root hash.

### DIFF
--- a/bns-test/test/name_transfer.test.ts
+++ b/bns-test/test/name_transfer.test.ts
@@ -142,7 +142,7 @@ describe("BNS Test Suite - NAME_TRANSFER", () => {
         sender: cases[0].nameOwner
       });
     expect(receipt.success).eq(true);
-    expect(receipt.result).include('u16');
+    expect(receipt.result).include('u150');
     block_height += 1;
 
     receipt = await bns.nameRegister(
@@ -164,7 +164,7 @@ describe("BNS Test Suite - NAME_TRANSFER", () => {
         sender: charlie
       });
     expect(receipt.success).eq(true);
-    expect(receipt.result).include('u18');
+    expect(receipt.result).include('u152');
     block_height += 1;
 
     receipt = await bns.nameRegister(
@@ -280,7 +280,7 @@ describe("BNS Test Suite - NAME_TRANSFER", () => {
       "charlie", {
         sender: cases[0].nameOwner
       });
-    expect(receipt.result).include('0x00');
+    expect(receipt.result).include('0x');
     expect(receipt.success).eq(true);
 
     // Bob should not be able to update 'charlie.blockstack'
@@ -321,7 +321,7 @@ describe("BNS Test Suite - NAME_TRANSFER", () => {
       "bob", {
         sender: cases[0].nameOwner
       });
-    expect(receipt.result).include('0x00');
+    expect(receipt.result).include('0x');
     expect(receipt.success).eq(true);
 
     // Bob should be able to update its zonefile
--- a/bns-test/test/name_transfer.test.ts
+++ b/bns-test/test/name_transfer.test.ts
@@ -280,7 +280,7 @@ describe("BNS Test Suite - NAME_TRANSFER", () => {
       "charlie", {
         sender: cases[0].nameOwner
       });
-    expect(receipt.result).include('0x');
+    expect(receipt.result).include('(zonefile-hash 0x)');
     expect(receipt.success).eq(true);
 
     // Bob should not be able to update 'charlie.blockstack'
@@ -321,7 +321,7 @@ describe("BNS Test Suite - NAME_TRANSFER", () => {
       "bob", {
         sender: cases[0].nameOwner
       });
-    expect(receipt.result).include('0x');
+    expect(receipt.result).include('(zonefile-hash 0x))');
     expect(receipt.success).eq(true);
 
     // Bob should be able to update its zonefile
--- a/src/chainstate/stacks/db/mod.rs
+++ b/src/chainstate/stacks/db/mod.rs
@@ -71,8 +71,8 @@ use chainstate::burn::db::sortdb::*;
 
 use chainstate::stacks::boot::*;
 
-use net::Error as net_error;
 use net::atlas::BNS_CHARS_REGEX;
+use net::Error as net_error;
 
 use vm::analysis::analysis_db::AnalysisDatabase;
 use vm::analysis::run_analysis;
@@ -1116,7 +1116,7 @@ impl StacksChainState {
                                 let components: Vec<_> =
                                     entry.fully_qualified_name.split(".").collect();
                                 assert_eq!(components.len(), 2);
-                                
+
                                 let namespace = {
                                     let namespace_str = components[1];
                                     if !BNS_CHARS_REGEX.is_match(&namespace_str) {
@@ -1125,7 +1125,7 @@ impl StacksChainState {
                                     let buffer = namespace_str.as_bytes();
                                     Value::buff_from(buffer.to_vec()).expect("Invalid namespace")
                                 };
-                                
+
                                 let name = {
                                     let name_str = components[0].to_string();
                                     if !BNS_CHARS_REGEX.is_match(&name_str) {
--- a/src/chainstate/stacks/db/mod.rs
+++ b/src/chainstate/stacks/db/mod.rs
@@ -1034,141 +1034,147 @@ impl StacksChainState {
             if let Some(get_namespaces) = boot_data.get_bulk_initial_namespaces.take() {
                 info!("Initializing chain with namespaces");
                 clarity_tx.connection().as_transaction(|clarity| {
-                    clarity.with_clarity_db(|db| {
+                    clarity
+                        .with_clarity_db(|db| {
+                            let initial_namespaces = get_namespaces();
+                            for entry in initial_namespaces {
+                                let namespace = {
+                                    let buffer = entry.namespace_id.as_bytes();
+                                    Value::buff_from(buffer.to_vec()).expect("Invalid namespace")
+                                };
 
-                        let initial_namespaces = get_namespaces();
-                        for entry in initial_namespaces {
-                            
-                            let namespace = {
-                                let buffer = entry.namespace_id.as_bytes();
-                                Value::buff_from(buffer.to_vec())
-                                    .expect("Invalid namespace")
-                            };
+                                let importer = {
+                                    let address = StacksChainState::parse_genesis_address(
+                                        &entry.importer,
+                                        mainnet,
+                                    );
+                                    Value::Principal(address)
+                                };
 
-                            let importer = {
-                                let address = StacksChainState::parse_genesis_address(&entry.importer, mainnet);
-                                Value::Principal(address)
-                            };
+                                let revealed_at = Value::UInt(entry.revealed_at.into());
+                                let launched_at = Value::UInt(entry.launched_at.into());
+                                let lifetime = Value::UInt(entry.lifetime.into());
+                                let price_function = {
+                                    let base = Value::UInt(entry.base.into());
+                                    let coeff = Value::UInt(entry.coeff.into());
+                                    let nonalpha_discount =
+                                        Value::UInt(entry.nonalpha_discount.into());
+                                    let no_vowel_discount =
+                                        Value::UInt(entry.no_vowel_discount.into());
+                                    let buckets: Vec<_> = entry
+                                        .buckets
+                                        .split(";")
+                                        .map(|e| Value::UInt(e.parse::<u64>().unwrap().into()))
+                                        .collect();
 
-                            let revealed_at = Value::UInt(entry.revealed_at.into());
-                            let launched_at = Value::UInt(entry.launched_at.into());
-                            let lifetime = Value::UInt(entry.lifetime.into());
-                            let price_function = {
-                                let base = Value::UInt(entry.base.into());
-                                let coeff = Value::UInt(entry.coeff.into());    
-                                let nonalpha_discount = Value::UInt(entry.nonalpha_discount.into());
-                                let no_vowel_discount = Value::UInt(entry.no_vowel_discount.into());
-                                let buckets: Vec<_> = entry.buckets
-                                    .split(";")
-                                    .map(|e| 
-                                        Value::UInt(e.parse::<u64>().unwrap().into()))
-                                    .collect();
-                                
-                                TupleData::from_data(vec![
-                                    ("buckets".into(), Value::list_from(buckets).unwrap()),
-                                    ("base".into(), base),
-                                    ("coeff".into(), coeff),
-                                    ("nonalpha-discount".into(), nonalpha_discount),
-                                    ("no-vowel-discount".into(), no_vowel_discount),
-                                ]).unwrap()
-                            };
+                                    TupleData::from_data(vec![
+                                        ("buckets".into(), Value::list_from(buckets).unwrap()),
+                                        ("base".into(), base),
+                                        ("coeff".into(), coeff),
+                                        ("nonalpha-discount".into(), nonalpha_discount),
+                                        ("no-vowel-discount".into(), no_vowel_discount),
+                                    ])
+                                    .unwrap()
+                                };
 
-                            let namespace_props = Value::Tuple(
-                                TupleData::from_data(vec![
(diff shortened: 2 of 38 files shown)


bns-test/test/name_transfer.test.ts
```diff
@@ -280,7 +280,7 @@ describe("BNS Test Suite - NAME_TRANSFER", () => {
       "charlie", {
         sender: cases[0].nameOwner
       });
-    expect(receipt.result).include('0x00');
+    expect(receipt.result).include('0x');
```
jcnelson, line 283: Would this result pass for any zonefile hash?

src/chainstate/stacks/db/mod.rs
```diff
@@ -1001,6 +1030,154 @@ impl StacksChainState {
...
+
+            if let Some(get_names) = boot_data.get_bulk_initial_names.take() {
+                info!("Initializing chain with names");
+                clarity_tx.connection().as_transaction(|clarity| {
+                    clarity
+                        .with_clarity_db(|db| {
+                            let initial_names = get_names();
+                            for entry in initial_names {
+                                let components: Vec<_> =
+                                    entry.fully_qualified_name.split(".").collect();
```
jcnelson, line 1112: Just for safety, you may want to assert that `components` has length 2, and that both `components[0]` and `components[1]` use only the permitted name characters.

src/core/mempool.rs
```diff
@@ -979,7 +979,7 @@ impl MemPoolDB {
             sponsor_nonce,
         )?;
 
-        Ok(())
+        Ok(tx)
```
jcnelson, line 982: Why is this returning `tx`?  If the caller needs the transaction after submitting it, then maybe it make more sense to just have this method take `tx: &StacksTransaction` as an argument, and continue to return `Ok(())`?

src/net/atlas/db.rs
```diff
@@ -113,9 +115,32 @@ impl AtlasDB {
         Ok(())
     }
 
+    pub fn should_keep_attachment(
+        &self,
+        contract_id: &QualifiedContractIdentifier,
+        attachment: &Attachment,
+    ) -> bool {
+        if self.atlas_config.contracts.contains(contract_id) {
```
jcnelson, line 123: Shouldn't this be `if !self.atlas_config.contracts.contains(contract_id) {`?

src/net/atlas/db.rs
```diff
@@ -113,9 +115,32 @@ impl AtlasDB {
...
+            info!(
+                "Atlas: will discard posted attachment - {} not in supported contracts",
+                contract_id
+            );
+            return false;
+        }
+        if attachment.content.len() as u32 > self.atlas_config.attachments_max_size {
```
jcnelson, line 130: You're going to want to check this while the attachment is being uploaded, so you can just cut off the client once the maximum attachment size is reached.  Otherwise, a client could DoS the node by uploading a humongous attachment, causing it to OOM, before this check is reached.

src/net/atlas/db.rs
```diff
@@ -231,30 +257,76 @@ impl AtlasDB {
         Ok(res)
     }
 
-    pub fn insert_new_attachment(&mut self, attachment: &Attachment) -> Result<(), db_error> {
+    pub fn insert_uninstanciated_attachment(
```
jcnelson, line 260: "uninstantiated"

src/net/atlas/db.rs
```diff
@@ -231,30 +257,76 @@ impl AtlasDB {
...
+        &mut self,
+        attachment: &Attachment,
+    ) -> Result<(), db_error> {
+        // Insert the new attachment
+        let uninstantiated_attachments = self.count_uninstantiated_attachments()?;
+        if uninstantiated_attachments >= self.atlas_config.max_uninstanciated_attachments {
+            let to_delete =
+                1 + uninstantiated_attachments - self.atlas_config.max_uninstanciated_attachments;
+            self.evict_k_oldest_uninstantiated_attachments(to_delete)?;
```
jcnelson, line 269: Won't this cause a lot of thrashing?  Suppose `self.atlas_config.max_uninstantiated_attachments` is 1024.  If I upload the 1025th uninstantiated attachment, won't this cause the node to delete all 1025 uninstantiated attachments, even if they were all otherwise valid and destined for inclusion?

jcnelson: I think what you want to do here is just evict the oldest attachment, in order to make space for this next one.  In addition to (or possibly instead of) using time, you might want to consider the transaction fee -- lower-fee-rate transactions have their attachments evicted before higher-fee-rate ones.  Open to discussion.

src/net/atlas/db.rs
```diff
@@ -231,30 +257,76 @@ impl AtlasDB {
...
+            "DELETE FROM attachments WHERE was_instantiated = 0 AND created_at < ?",
+            &[&cut_off as &dyn ToSql],
+        );
+        res.map_err(db_error::SqliteError)?;
+        tx.commit().map_err(db_error::SqliteError)?;
+        Ok(())
+    }
+
+    pub fn count_uninstantiated_attachments(&self) -> Result<u32, db_error> {
+        let qry = "SELECT COUNT(*) FROM attachments
```
jcnelson, line 312: `COUNT(*)` is a _O(n)_ operation in Sqlite.  Is there a way to avoid this call, or minimize its usage?  Over time, it will get more and more expensive.

If you're just looking to see if there are more than _k_ uninstantiated attachments, one way to do it is to create an index on the `was_instantiated` column and then do a `SELECT was_instantiated FROM attachments WHERE was_instantiated = 0 LIMIT ${k}`.  Then, it's a _O(k log n)_ operation, since you're doing an indexed B+-tree scan.

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -217,13 +222,17 @@ impl RunLoop {
...
+        let moved_atlas_config = atlas_config.clone();
+
         thread::spawn(move || {
             ChainsCoordinator::run(
                 chain_state_db,
                 coordinator_burnchain_config,
                 attachments_tx,
                 &mut coordinator_dispatcher,
                 coordinator_receivers,
+                moved_atlas_config,
```
jcnelson, line 235: why not just have `atlas_config.clone()` here?

src/chainstate/stacks/db/mod.rs
```diff
@@ -1001,6 +1030,154 @@ impl StacksChainState {
...
 
+            let bns_contract_id = boot_code_id("bns");
+            if let Some(get_namespaces) = boot_data.get_bulk_initial_namespaces.take() {
+                info!("Initializing chain with namespaces");
+                clarity_tx.connection().as_transaction(|clarity| {
+                    clarity
+                        .with_clarity_db(|db| {
+                            let initial_namespaces = get_namespaces();
+                            for entry in initial_namespaces {
+                                let namespace = {
```
jcnelson, line 1041: Just for safety, you may want to validate the characters in `namespace`.

src/chainstate/stacks/db/mod.rs
```diff
@@ -1001,6 +1030,154 @@ impl StacksChainState {
...
+                                let launched_at = Value::UInt(entry.launched_at.into());
+                                let lifetime = Value::UInt(entry.lifetime.into());
+                                let price_function = {
+                                    let base = Value::UInt(entry.base.into());
+                                    let coeff = Value::UInt(entry.coeff.into());
+                                    let nonalpha_discount =
+                                        Value::UInt(entry.nonalpha_discount.into());
+                                    let no_vowel_discount =
+                                        Value::UInt(entry.no_vowel_discount.into());
+                                    let buckets: Vec<_> = entry
```
jcnelson, line 1064: You may want to verify that there are 16 entries here.

src/net/atlas/download.rs
```diff
@@ -98,6 +98,9 @@ impl AttachmentsDownloader {
                         .resolve_attachment(&attachment.hash())
                 }
 
+                // Every once in a while, we delete uninstanciated attachments
+                network.atlasdb.evict_expired_uninstantiated_attachments()?;
```
jcnelson, line 102: Do you know how often this will run?  Is it once per state-machine pass (e.g. once every 5 calls to `PeerNetwork::run()`), or is it unbound?

src/net/atlas/db.rs
lgalabru: > Won't this cause a lot of thrashing? 
It's the other way, as shown in a test, if we receive a 1025th uninstantiated attachment, we delete the oldest one.

src/net/atlas/db.rs
lgalabru: Oups, I've been re-writing this condition a few times - will fix and test.

src/net/atlas/download.rs
lgalabru: > once per state-machine pass

Yes.

src/net/atlas/db.rs
lgalabru: > Over time, it will get more and more expensive. 

But it is bound by the `WHERE` condition so `max_uninstanciated_attachments`, no?

src/net/atlas/db.rs
lgalabru: If the transaction is not being admitted in the mempool, we would end up deleting legit attachments.

> Otherwise, a client could DoS the node by uploading a humongous attachment, causing it to OOM, before this check is reached

You don't need attachments for DDOSing a node, you can just use transactions, no?

testnet/stacks-node/src/run_loop/neon.rs
lgalabru: The `atlas_config` would be moved to the another thread.

src/net/atlas/db.rs
lgalabru: Fixed in test: add coverage for test_keep_uninstantiated_attachments

src/chainstate/stacks/db/mod.rs
lgalabru: Checks added in https://github.com/blockstack/stacks-blockchain/pull/2204/commits/52579d5dd4d88987a9dbc53543c8bb85ae971cbc

src/chainstate/stacks/db/mod.rs
lgalabru: Checks added in https://github.com/blockstack/stacks-blockchain/pull/2204/commits/52579d5dd4d88987a9dbc53543c8bb85ae971cbc

src/chainstate/stacks/db/mod.rs
lgalabru: Checks added in https://github.com/blockstack/stacks-blockchain/pull/2204/commits/52579d5dd4d88987a9dbc53543c8bb85ae971cbc

bns-test/test/name_transfer.test.ts
lgalabru: Assertion rewritten in https://github.com/blockstack/stacks-blockchain/pull/2204/commits/0dee5f178e09da48790e901ab1c903c98187ccc5

src/net/atlas/db.rs
lgalabru: Addressed in https://github.com/blockstack/stacks-blockchain/pull/2204/commits/fe89bce9fa000b3aad271fa6ae5404145cafa1f7

src/net/atlas/db.rs
lgalabru: Also, with the suggested SQL query, won't we end up returning / validating / instantiating N `i64`? Isn't better to just let sqlite doing a `COUNT`?

src/core/mempool.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2204/commits/7e4e9dffb1600aa28cbcc38c3cd798c67c0675e2

src/net/atlas/db.rs
jcnelson: > You don't need attachments for DoS-ing a node, you can just use transactions, no?

Looking at the code and `git blame`, the `parse_posttransaction()` code in `src/net/http.rs` _used to_ use a `BoundReader` to prevent this very problem.  It looks like this got removed by accident when `parse_posttransaction()` got updated to accommodate attachments.

You'll want to update `parse_posttransaction()` in `src/net/http.rs` as follows:

```
    fn parse_posttransaction<R: Read>(
        _protocol: &mut StacksHttp,
        preamble: &HttpRequestPreamble,
        _regex: &Captures,
        _query: Option<&str>,
        fd: &mut R,
    ) -> Result<HttpRequestType, net_error> {
        if preamble.get_content_length() == 0 {
                "Invalid Http request: expected non-zero-length body for PostTransaction"
                    .to_string(),
            ));
        }
        
        // NEW CODE
        if preamble.get_content_length() > MAX_PAYLOAD_LEN as u64 {
                "Invalid Http request: PostTransaction body is too big"
                    .to_string(),
            ));
        }

        // NEW CODE
        let mut bound_fd = BoundReader::from_reader(fd, preamble.get_content_length() as u64);

        match preamble.content_type {
            None => {
                    "Missing Content-Type for transaction".to_string(),
                ));
            }
            Some(HttpContentType::Bytes) => {
                HttpRequestType::parse_posttransaction_octets(preamble, bound_fd)  // NEW CODE
            }
            Some(HttpContentType::JSON) => {
                HttpRequestType::parse_posttransaction_json(preamble, bound_fd)  // NEW CODE
            }
            _ => {
                    "Wrong Content-Type for transaction; expected application/json".to_string(),
                ));
            }
        }
    }
```

(Note: I haven't checked that the code compiles or passes `rustfmt`).

The gist of it is that the network code needs to wrap the request body's input stream in a `BoundReader`, which will prevent it from reading more than `preamble.get_content_length()` bytes before encountering EOF.  `MAX_PAYLOAD_LEN` works out to be a little over 16MB; you may need to use a different constant if attachments are meant to be considerably bigger or smaller.

src/net/atlas/db.rs
jcnelson: > But it is bound by the WHERE condition so max_uninstantiated_attachments, no?

Only if you create an index that maps `was_instantiated` to `attachments.hash`.  Otherwise it's a full table scan.  Also, `COUNT(ROWID)` is much faster than `COUNT(*)`.  You'll want to verify this with `EXPLAIN QUERY PLAN`.

jcnelson (approved):
Overall this LGTM, but before merging, please address my feedback about the use of `COUNT(*)` in sqlite and the use of `BoundReader` in `src/net/http.rs` to prevent uploading arbitrarily-sized attachments.

src/net/atlas/db.rs
lgalabru: Thank you for the pointer, addressed with https://github.com/blockstack/stacks-blockchain/pull/2204/commits/27d0b1ce988f165eb92a9e8d0b672235eef1dc6c

src/net/atlas/db.rs
lgalabru: Thank you @jcnelson.
Since it's more usable, I went with 
```sql
SELECT COUNT(rowid) FROM attachments WHERE was_instantiated = 0;
```
that has the same query plan than

```sql
SELECT was_instantiated FROM attachments WHERE was_instantiated = 0 LIMIT N
```

as long as you add an index on `was_instantiated`

```bash
sqlite> EXPLAIN QUERY PLAN SELECT COUNT(*) FROM attachments WHERE was_instantiated = 0;
QUERY PLAN
--SEARCH TABLE attachments USING COVERING INDEX index_was_instanciated (was_instantiated=?)

sqlite> EXPLAIN QUERY PLAN SELECT was_instantiated FROM attachments WHERE was_instantiated = 0 LIMIT 2;
QUERY PLAN
--SEARCH TABLE attachments USING COVERING INDEX index_was_instanciated (was_instantiated=?)
```

jcnelson: Sounds good :+1: Thanks for checking with the query planner!  As long as we're using a `COVERING INDEX` to do the scan, then it's fine by me.

PR #2213, opened 2020-12-23
MARF RAM Improvement

kantai:
This PR does 2 things to trim the RAM usage of the MARF implementation. This has the most significant impact on genesis block processing, which previously had a maximum memory usage of ~3 GB, but in this PR has ~400MB max memory usage.

The two things are:

1. Make the `dump_traverse` routine used in flushing the TrieRAM to a buffer a consuming function. This allows the method to modify the `data` vec in place, rather than constructing a new vec. This provides about a 50% memory saving (when tested, went from 3GB to 1.6GB usage).
2. Use `Box` allocations for TrieNodeType variants for Node48 and Node256. This provides a significant memory savings, because it changes the size of `TrieNodeType` from ~2.5kb to ~100 bytes. That in turn allows the `data` vec in the TrieRAM to allocate significantly less memory (previously, this data vec would have ~340,000 entries in the genesis block, a vast majority of which are leaves, but allocated with enough space for Node256s and Node48s).

### DIFF
--- a/src/chainstate/stacks/index/bits.rs
+++ b/src/chainstate/stacks/index/bits.rs
@@ -228,8 +228,8 @@ pub fn get_nodetype_hash_bytes<T: MarfTrieId, M: BlockMap>(
     match node {
         TrieNodeType::Node4(ref data) => get_node_hash(data, child_hash_bytes, map),
         TrieNodeType::Node16(ref data) => get_node_hash(data, child_hash_bytes, map),
-        TrieNodeType::Node48(ref data) => get_node_hash(data, child_hash_bytes, map),
-        TrieNodeType::Node256(ref data) => get_node_hash(data, child_hash_bytes, map),
+        TrieNodeType::Node48(ref data) => get_node_hash(data.as_ref(), child_hash_bytes, map),
+        TrieNodeType::Node256(ref data) => get_node_hash(data.as_ref(), child_hash_bytes, map),
         TrieNodeType::Leaf(ref data) => get_node_hash(data, child_hash_bytes, map),
     }
 }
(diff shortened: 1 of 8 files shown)


jcnelson:
> Use Box allocations for TrieNodeType variants for Node48 and Node256. This provides a significant memory savings, because it changes the size of TrieNodeType from ~2.5kb to ~100 bytes.

I'm curious why this is?  The byte-for-byte representation hasn't changed; all that data is just now put on the heap instead of inline with the node data.

Also, why not make all inner node data `Box`'ed, including `Node4` and `Node16`?

kantai:
> I'm curious why this is? The byte-for-byte representation hasn't changed; all that data is just now put on the heap instead of inline with the node data.

The answer has to do with the way enum memory layouts work in Rust. Enums are sized by the maximum variant size. `Vec<T>` allocations _always_ lay out `sizeof(T) * vec_size` space. `sizeof(TrieNodeType)` was previously 2.5kb, because the Node256 and Node48 variants were that large. Replacing those variants with a Box makes those variants' sizes to be just a pointer, so the `sizeof(TrieNodeType)` is now much smaller. So whereas before the `data` field of the TrieRAM was (330,000 * 2.5kB), now it's more like (330,000 * 200B).

> Also, why not make all inner node data Box'ed, including Node4 and Node16?

There is a bit of a trade-off here between the size of the TrieNodeType enum and memory fragmentation. Making Node48 and Node256 boxed shaves off a huge amount of memory consumption, without really impacting performance. More tuning might be required if we pushed it further on the other nodes.

jcnelson:
> The answer has to do with the way enum memory layouts work in Rust. Enums are sized by the maximum variant size.

Huh, today I learned.  Also, seriously wtf rust.

jcnelson (approved):
LGTM!

lgalabru (approved):
Thank you @kantai!

Issue #2046, opened 2020-11-09
Using testground for large-scale testing


kantai:
Doing large-scale testing of the `stacks-blockchain` is possible using testground. A basic prototype plan is defined here:

https://github.com/kantai/stacks-blockchain-testground

It doesn't collect many metrics -- there's a couple different avenues for adding more metrics:

1. Parse the stderr logs out of the stacks-blockchain process (shouldn't be too hard, they are in JSON).
2. Have the Go process implement a EventDispatcher client to receive EventDispatcher events.

(1) is probably easier.

wileyj:
https://github.com/blockstack/stacks-blockchain-testground/tree/development
@jcnelson @lgalabru for visibility here. 


Problem: Occasionally, running a test will fail with some of the instances failing to start stacks-node.

```
Dec 21 02:53:01.463211 INFO 463.0579s    MESSAGE << single[003] (bc4b34) >> Get "http://localhost:20443/v2/info": dial tcp 127.0.0.1:20443: connect: connection refused
Dec 21 02:53:01.589109 INFO 463.1838s    MESSAGE << single[001] (d5ea45) >> Get "http://localhost:20443/v2/info": dial tcp 127.0.0.1:20443: connect: connection refused
Dec 21 02:53:02.113005 INFO 463.7076s    MESSAGE << single[002] (9e30c1) >> Stacks block height => 1 :: Burn block height => 308
Dec 21 02:53:02.166044 INFO 463.7607s    MESSAGE << single[002] (9e30c1) >> Inbound/Outbound Neighbors => 1/1
```

It doesn't happen all of the time, and cleaning the old runs, restarting the daemon etc doesn't seem to help. 
What does seem to help (and I have no idea why) is rebooting the VM. There doesn't seem to be any difference in resources used, so it may just be coincidental. 

When this does happen, the instances running stacks-node/bitcoin etc have multiple log files I can see in `/src/net-tools/mnt`

For the failed nodes, I only see a single file `cleanup.sh`:

```
root@dd48e860bd73:/src# ls -al /src/net-test/mnt/
total 16
drwxr-xr-x 2 root root 4096 Dec 14 15:34 .
drwxr-xr-x 1 root root 4096 Dec 19 02:38 ..
-rwxr-xr-x 1 root root  237 Dec 14 15:34 cleanup.sh
```

I do see the attempt to start the stacks-node as well:
```
[...]

PR #2214, opened 2020-12-23
increase the number of attempts to confrim BTC transaction from 10->20

wileyj:
## Description
Running tests with testground, occasionally an instance or two would fail to confirm the BTC faucet transaction after 10 attempts. 
Increasing this int from `10`->`20` has helped in local testing, no more failures of this type have been caught. 

For details refer to issue #2046 

## Testing information
Due to this change, it may take longer to confirm BTC faucet tx in `./net-test`, but functionally nothing should be affected.


jcnelson (approved):
LGTM!

PR #2218, opened 2020-12-24
Remove PoX state from consideration when validating prepare-phase sortitions

### LINKED ISSUE #2217: Chainstate: remove VRF key's consensus hash requirement
In order to address #1805, we need to make it so a PoB block-commit can be validated in a prepare phase without the node having access to any PoX state.  But in order to do this, we need to make it so that the miner's VRF key transaction is always validatable without PoX state, so we can validate the block-commit.  In order to achieve this we either need to:

* Make a VRF key registration transaction valid in all PoX forks, i.e. by removing its consensus hash, OR
* Make it so a block-commit can be valid without a VRF key registration existing, i.e. by altering the Stacks block validation logic to cause the Stacks block to be rejected by the network if its block-commit doesn't have a VRF key registration.

It's easier to remove the consensus hash from the VRF key registration (or to just ignore it) than to update the block-validation logic, so I recommend the former.

jcnelson:
Closes #2217, #2221, and #2222.

Regarding #2217, this change simply removes the check to see whether or not the VRF key register's consensus hash is valid.  The reason for doing this is that in order to address #1805, all sortitions in the prepare phase need to be valid on-chain -- i.e. they must attach to an existing Stacks fork.  But, one of the validation requirements for block-commits is that the VRF key it references exists.  Unfortunately, VRF keys are only accepted if their consensus hashes are valid, which means that the act of validating a VRF key (and thus a block-commit) requires the node to have processed all Stacks anchor blocks up to that VRF key's reward cycle.  This is problematic, since solving #1805 requires that the node be able to process all sortitions in all prepare phases without having any Stacks blocks (which means all block-commits must be validatable in each prepare phase).

The reason for simply removing the consensus hash check, but without touching the wire format, is to keep compatibility with existing clients.  I'm happy to remove the consensus hash altogether from the wire format, but that would be even more disruptive than this proposed change.

Regarding #2221, this change simply removes the block-wide filter that removes all block-commits that reference the same VRF key.

Regarding #2222, this change updates the `BurnSamplePoint::make_min_median_distribution()` function and the `BurnchainStateTransition::from_block_ops()` function so that (a) the windowing code will distinguish PoB-only block-commits from potentially-PoX block-commits when validating the consumed UTXO index (it will be 2 for PoB-only block-commits, and 3 for potentially-PoX block-commits), and (b) the commitment window size is set to 1 during a prepare phase (the window will reset to `MINING_COMMITMENT_WINDOW` on the start of a reward cycle, so the last 6 PoB-only commits contribute to their min and median commits as expected).


kantai (commented):
This approach looks good to me -- is this the only place where the consensus hash is validated for leader key registers?

A related check that we should remove is the VRF key "uniqueness" check on leader block commits. If multiple commits in the same block try to use the same key, only one is valid -- do you want to remove that check in this PR as well?

jcnelson:
> A related check that we should remove is the VRF key "uniqueness" check on leader block commits. If multiple commits in the same block try to use the same key, only one is valid -- do you want to remove that check in this PR as well?

I'm happy to remove that check.

While we're making consensus-breaking changes, I think I also need to modify the logic in `BurnchainStateTransition::from_block_ops()` to only do commit windowing during reward phases, and not prepare phases.

kantai:
> While we're making consensus-breaking changes, I think I also need to modify the logic in BurnchainStateTransition::from_block_ops() to only do commit windowing during reward phases, and not prepare phases.

Sure -- the other option is to have it "reset" the commit window at the start of the prepare phase (i.e., at the first block of the prepare phase, the window would be length 1, at the second block, the window would be length 2, etc.).

jcnelson:
> Sure -- the other option is to have it "reset" the commit window at the start of the prepare phase (i.e., at the first block of the prepare phase, the window would be length 1, at the second block, the window would be length 2, etc.).

Yup, either way works for me.  The code currently sets the window size to 1 during the prepare-phase, and re-sets it back to 6 once the reward phase starts, but I can change it if you think doing windowing during the prepare-phase is more desirable (the bulk of the complexity in parts of the diff that touch to windowing comes from having to accommodate prior block-commits that have only one burn output, instead of the usual two PoX outputs).

I was under the impression that windowing won't be necessary if we're only doing proof-of-burn (as will be the case during the prepare phase), because the main purpose of windowing was to compel would-be discount miners to have at least sent BTC to other stackers?

jcnelson:
I can confirm that the Neon integration tests that are failing are doing so intermittently -- they all pass on my local machine.

jcnelson:
I'm marking this as `[Do not merge]` for now.  In the mean time, I'm going to keep making progress on #1805 with this patch applied in order to verify that my solution to #1805 doesn't require any further consensus-altering modifications.  I'll keep all consensus-modifying modifications in this PR, so it can be more heavily scrutinized.

jcnelson:
This is ready for review -- we can hopefully merge it into the next Xenon release today.

src/chainstate/burn/distribution.rs
```diff
@@ -186,19 +201,64 @@ impl BurnSamplePoint {
...
-                false
+                None
             };
-            let expected_index = LeaderBlockCommitOp::expected_chained_utxo(sunset_finished);
+
             for linked_commit in commits_with_priors.iter_mut() {
                 let end = linked_commit.iter().rev().find_map(|o| o.as_ref()).unwrap(); // guaranteed to be at least 1 non-none entry
 
-                // check that the commit is using the right output index
-                if end.op.spent_output() != expected_index {
```
kantai, line 199: Previously, this branch checked that `end.op` spent the correct output from the "current" block. Now, this code checks that "current" spends the correct output from it's prior block. I think this shift causes a check to be missed between the latest commit in the window and the second-latest commit in the window (as well as requiring the introduction of an option type for the check, which shouldn't be necessary if we're always checking from the end to the "current").

src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -394,6 +394,18 @@ impl LeaderBlockCommitOp {
                 previous_is_burn && output_addr.is_burn()
             })
     }
+
+    pub fn spent_txid(&self) -> Txid {
+        self.input.0.clone()
```
kantai, line 399: Does this need to clone(), or could we make do with returning the `&Txid` here?

src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -439,6 +451,16 @@ pub struct MissedBlockCommit {
     pub intended_sortition: SortitionId,
 }
 
+impl MissedBlockCommit {
+    pub fn spent_txid(&self) -> Txid {
+        self.input.0.clone()
```
kantai, line 456: Same as above -- can this just return `&Txid`?

src/burnchains/burnchain.rs
```diff
@@ -2348,156 +2326,6 @@ pub mod tests {
         }
     }
 
-    #[test]
-    fn test_filter_block_commits_with_same_VRF_key() {
```
kantai, line 2352: Rather than deleting this test, should we update the assertions to accept multiple block commits?

src/burnchains/burnchain.rs
```diff
@@ -164,68 +165,107 @@ impl BurnchainStateTransition {
...
+        // figure out if the PoX sunset finished during the window,
+        // and/or which sortitions must be PoB due to them falling in a prepare phase.
         let window_end_height = parent_snapshot.block_height + 1;
         let window_start_height = window_end_height + 1 - (windowed_block_commits.len() as u64);
-        let sunset_finished_at = if sunset_end <= window_start_height {
-            Some(0)
+        let mut burn_blocks = vec![false; windowed_block_commits.len()];
+
+        // set burn_blocks flags to accomodate PoX sunset
+        if sunset_end <= window_start_height {
```
kantai, line 232: Nitpick: can you change the setting of `burn_blocks` to a single pass over the `burn_blocks`?

```rust
           for (i, b) in burn_blocks.iter_mut().enumerate() {
                if sunset_end <= window_start_height + (i as u64) {
                    // must burn here
                    *b = true;
                } else if burnchain.is_in_prepare_phase(window_start_height + (i as u64)) {
                    // must burn
                   *b = true;
                } else {
                    // may not burn here
                    *b = false;
                }
           }
```

I think it would make it more clear how the burn_blocks vec is set (as opposed to the various special cases on `sunset_end` and multiple passes over the vec)

src/burnchains/burnchain.rs
jcnelson: The only thing this test did was check to see that the now-removed `test_filter_block_commits_with_same_VRF_key()` function actually did the requisite filtering.  It didn't process or insert the block-commits into a database or anything that would make the kinds of checks you're asking for possible.

Instead of restoring this check, would it make more sense to expand the `test_check()` test in `leader_block_commit.rs` to verify that two block-commits can reference the same VRF key and be accepted?

kantai: > Instead of restoring this check, would it make more sense to expand the test_check() test in leader_block_commit.rs to verify that two block-commits can reference the same VRF key and be accepted?

Yep, that makes sense.

src/chainstate/burn/distribution.rs
jcnelson: Good catch!  Fixed in 9c6035c

src/chainstate/burn/operations/leader_block_commit.rs
jcnelson: Fixed in 9c6035c

src/chainstate/burn/operations/leader_block_commit.rs
jcnelson: Fixed in 9c6035c

src/burnchains/burnchain.rs
jcnelson: Addressed in 20f7304

src/burnchains/burnchain.rs
jcnelson: Addressed in c9285de

kantai (approved):
LGTM!

jcnelson:
Still having a hard time with those neon integration tests passing in CI, it seems.

lgalabru (approved):

Issue #2224, opened 2020-12-28
Introduce burnchain config constant for initial reward window


kantai:
Currently, the initial mining reward amount is calculated via the `first_burn_block`. However, the initial mining reward period should actually be configurable to a different value.

PR #2225, opened 2020-12-28
Address #2224 -- separate initial rewards start from first block height

kantai:
This PR separates the start of blocks counted towards the initial block reward from the first block height.

It does so by introducing a new field in the `Burnchain` struct, which is tested via a new unit test in `chainstate::burn::db::processing`.


jcnelson (approved):
LGTM!

lgalabru (approved):
Looks good to me!

Issue #2223, opened 2020-12-28
Cost voting contract vetos


kantai:
Currently, the cost voting contract works as follows:

1. A proposal is submitted
2. The proposal receives enough "locked STX votes" to clear a threshold. Once this happens, the proposal becomes stx-confirmed.
3. For some window of burn blocks, miners are able to "veto" the proposal. If some % of the window of burn blocks have miner vetos, then the proposal is aborted.
4. Anytime after the window, the proposal is now valid and may be "finalized"

One problem here is that miner vetos won't get counted in Stacks forks, e.g.,

```
a -> b -> c -> d -> e
 \_ a'
```

If a proposal in veto'd by block `b,c,d,e`, the proposal would not be valid in that fork. _However_, if `a'` is chosen in a much later burn block, the window would have passed without any vetos being applied.

I think the way to address this is as follows:

1. Switch to using `block-height` (i.e., _Stacks_ block height) for the veto window.
2. Add an expiration burn block for a proposal that clears the veto window. 

This makes it such that a proposal cannot be confirmed by miners if the network is facing heavily contested forks, and cannot be confirmed by a deep reorg.

jcnelson:
> If a proposal in veto'd by block b,c,d,e, the proposal would not be valid in that fork. However, if a' is chosen in a much later burn block, the window would have passed without any vetos being applied.

Having a hard time remembering at the moment -- what was the reason we consider the burn height at all?  What would happen if we took out consideration of the burn height entirely?  Then, two things would be true:

* In order to retroactively veto an already-ratified proposal, miners would need to reorg the Stacks chain back to the point in time where the veto period began, and then proceed to execute the veto.
* In order to retroactively approve a vetoed proposal, miners would need to reorg the Stacks chain back to the point in time where the veto period began, and then proceed to mine without veto transactions.

Either way, simply using `block-height` makes the act of changing a proposal outcome at least as expensive as taking control of the chain for a long period of time.  Also, using only `block-height` adds a fail-safe -- if an approved proposal turns out to be really damaging once it activates (let's say it triggers a chain split, or causes a denial of service), then miners can quickly and retroactively veto it this way for "only" the cost of re-mining the last `$VETO_PERIOD`'s transactions.

kantai:
Yep, switching to `block-height` works for me.

PR #2227, opened 2020-12-29
Update cost-voting to use block-height over burn-block-height

kantai:
This PR addresses #2223 by using `block-height` rather than `burn-block-height`.


jcnelson (approved):
This LGTM, but I'm surprised that no unit tests have been broken by this change?  Would it make sense to add a unit test case to see that a proposal can be vetoed in one fork, but permitted in another?

kantai:
Yeah, I can add a unit test for this in the presence of forks. Unrelated to this PR, I'm working on a neon_integration test for cost voting.

lgalabru (approved):
Looks good to me!

Issue #2228, opened 2020-12-29
Update reward duration details in SIP-007


agraebe:
Update proposal (for [this section](https://github.com/blockstack/stacks-blockchain/blame/master/sip/sip-007-stacking-consensus.md#L263)):

> These signed messages are valid for at most 25200 Bitcoin blocks (12
reward cycles, or >7 months). If the signed message specifies a lockup
period `x` less than 25200 blocks, then the signed message is only valid for
Stacking participation for `floor(x / 2100)` reward cycles (the minimum 
participation length is one cycle: 2100 blocks).

[bitcoin confirmation avg is at ~13min](https://www.blockchain.com/charts/median-confirmation-time), hence the 7 months.

cc @diwakergupta

diwakergupta:
Fixed with #2230

PR #2230, opened 2020-12-29
Address #2228: update reward cycle duration in SIP-007

diwakergupta:


sip/sip-007-stacking-consensus.md
```diff
@@ -260,19 +260,19 @@ reward period. If broadcast on the Bitcoin chain, they may be
 broadcast during the prepare phase, but must be included before
 the prepare phase finishes.
 
-These signed messages are valid for at most 12000 Bitcoin blocks (12
-reward cycles, or ~3 month). If the signed message specifies a lockup
-period `x` less than 12000 blocks, then the signed message is only valid for
-Stacking participation for `floor(x / 1000)` reward cycles (the minimum 
-participation length is one cycle: 1000 blocks).
+These signed messages are valid for at most 12 reward cycles (24000 Bitcoin
```
jcnelson, line 263: Reward cycles are 2100 blocks long.  This is 25200 blocks.

jcnelson (changes requested):

sip/sip-007-stacking-consensus.md
diwakergupta: I'm going by the convention in the original text: when the cycle was 1000 blocks, prepare phase was 240 blocks I believe. So technically original text should have used cycle length of 1240 blocks. However, my understanding is that prepare phase of the next cycle overlaps with the reward phase of the current cycle, so the "effective" cycle length was 1000 previously, and should be 2000 now.

Happy to change the convention once @kantai confirms.

kantai: The reward phase and prepare phase no longer overlap -- which means that each "cycle" is 2100 blocks. While the "effective" reward phase is 2000 blocks, the maximum length of a lockup is 12 * 2100 Stacks blocks (25,200).

diwakergupta: Yeah, I figured with prepare phase being PoB for #1805 they can no longer overlap. Will update, thanks!

kantai (approved):
LGTM!

jcnelson (approved):
LGTM!

PR #2233, opened 2020-12-30
PoX - Adjustments

lgalabru:
This PR is addressing a few issues that I've been facing while trying to complete some reward cycles, in the real life :).
As is, it's breaking a few tests - still need to check theses.


src/burnchains/burnchain.rs
```diff
@@ -464,12 +468,12 @@ impl Burnchain {
             .expect("Overflowed u64 in calculating expected sunset_burn")
     }
 
-    pub fn is_reward_cycle_start(&self, block_height: u64) -> bool {
-        if block_height <= (self.first_block_height + 1) {
-            // not a reward cycle start if we're the first block after genesis.
+    pub fn is_reward_cycle_start(&self, burn_height: u64, num_sortitions: u64) -> bool {
+        if num_sortitions == 0 {
```
kantai, line 472: Good catch!

src/burnchains/mod.rs
```diff
@@ -387,15 +387,27 @@ impl PoxConstants {
 
     pub fn testnet_default() -> PoxConstants {
         PoxConstants::new(
-            150, // 120 reward slots; 30 prepare-phase slots
-            30,
-            20,
+            50, // 120 reward slots; 30 prepare-phase slots
```
kantai, line 390: That's now 80 reward slots, and 10 prepare-phase blocks?

lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2233/commits/dc69365a483107053dc9a776b9998798c8c51578

lgalabru:
Looking at the tests (some of them are failing because this branch is based on `next`), but happy to start getting some eyes on this

src/chainstate/burn/db/sortdb.rs
```diff
@@ -1748,37 +1748,37 @@ impl<'a> SortitionHandleConn<'a> {
         &self,
         prepare_end_bhh: &BurnchainHeaderHash,
         pox_consts: &PoxConstants,
+        first_block_height: u64,
```
jcnelson, line 1751: Isn't this in `self.context` already?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -1748,37 +1748,37 @@ impl<'a> SortitionHandleConn<'a> {
...
 
         // if this block is the _end_ of a prepare phase,
-        if !(my_height % pox_consts.reward_cycle_length == 0) {
-            test_debug!(
-                "My height = {}, reward cycle length == {}",
-                my_height,
-                pox_consts.reward_cycle_length
+        let effective_height = block_height - first_block_height as u32;
+        let position_in_cycle = effective_height % pox_consts.reward_cycle_length;
+        if !(position_in_cycle == 0) {
```
jcnelson, line 1765: `position_in_cycle != 0`

src/chainstate/burn/db/sortdb.rs
```diff
@@ -1748,37 +1748,37 @@ impl<'a> SortitionHandleConn<'a> {
...
         }
 
-        let prepare_end = my_height;
-        // if this block isn't greater than prepare_length, then this shouldn't be the end of a prepare block
-        if prepare_end < pox_consts.prepare_length {
-            test_debug!(
-                "prepare_end = {}, pox_consts.prepare_length = {}",
-                prepare_end,
-                pox_consts.prepare_length
+        if !(effective_height == 0) {
```
jcnelson, line 1773: `effective_height != 0`

src/chainstate/coordinator/mod.rs
```diff
@@ -401,7 +403,11 @@ pub fn get_reward_cycle_info<U: RewardSetProvider>(
 
         let reward_cycle_info = {
             let ic = sort_db.index_handle(sortition_tip);
-            ic.get_chosen_pox_anchor(&parent_bhh, &burnchain.pox_constants)
+            ic.get_chosen_pox_anchor(
+                &parent_bhh,
+                &burnchain.pox_constants,
+                burnchain.first_block_height,
```
jcnelson, line 409: This should be equal to `sort_db.first_block_height`, and moreover, `sort_db.first_block_height` should be available in `ic.context`.

src/chainstate/burn/db/sortdb.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2233/commits/0e51df9b9122e660a18460d4dde02c426f946825

src/chainstate/burn/db/sortdb.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2233/commits/0e51df9b9122e660a18460d4dde02c426f946825

src/net/rpc.rs
```diff
@@ -255,55 +255,60 @@ impl RPCPoxInfoData {
...
+            .expect_u128() as u64;
 
         let total_required = total_liquid_supply_ustx
             .checked_div(rejection_fraction)
             .expect("FATAL: unable to compute total_liquid_supply_ustx/current_rejection_votes");
         let rejection_votes_left_required = total_required.saturating_sub(current_rejection_votes);
 
+        let burnchain_tip = SortitionDB::get_canonical_burn_chain_tip(sortdb.conn())?;
+
+        let next_reward_cycle_in = reward_cycle_length
```
jcnelson, line 309: Isn't this a method available in `Burnchain`?

lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2233/commits/fd507c59944e9f7d373b364b2beceea6cf00315a

src/chainstate/coordinator/mod.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2233/commits/fd507c59944e9f7d373b364b2beceea6cf00315a

src/chainstate/burn/db/sortdb.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2233/commits/fd507c59944e9f7d373b364b2beceea6cf00315a

src/burnchains/burnchain.rs
jcnelson: I'm a little confused why `num_sortitions` is needed here?  First, what's wrong with declaring that the first reward cycle in the system is always a PoB reward cycle?  It doesn't seem to require any special-case handling.  Second, if we cannot do this for some reason, then why not have `if block_height <= (self.first_block_height + self.pox_consts.reward_cycle_length + 1)` as the check, and remove `num_sortitions` entirely?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -1755,30 +1755,29 @@ impl<'a> SortitionHandleConn<'a> {
...
         }
 
-        let prepare_end = my_height;
-        // if this block isn't greater than prepare_length, then this shouldn't be the end of a prepare block
-        if prepare_end < pox_consts.prepare_length {
-            test_debug!(
-                "prepare_end = {}, pox_consts.prepare_length = {}",
-                prepare_end,
-                pox_consts.prepare_length
+        if effective_height == 0 {
```
jcnelson, line 1772: Wait, wasn't this `!(effective_height == 0)` before? Shouldn't this be `effective_height != 0`?

src/chainstate/coordinator/mod.rs
```diff
@@ -380,14 +381,15 @@ pub fn get_next_recipients<U: RewardSetProvider>(
 ///           * Was PoX anchor block known?
 pub fn get_reward_cycle_info<U: RewardSetProvider>(
     burn_height: u64,
+    num_sortitions: u64,
```
jcnelson, line 384: Can't this value be queried from the `sort_db`, instead of being passed via an argument?

lgalabru: It could, the intention was to avoid a query, since we have the data available (and beside tests, this method is invoked once or twice)

src/burnchains/burnchain.rs
lgalabru: From my understanding, and what I saw in the codebase, `self.first_block_height` is a constant for specifying that the genesis block can be mined anytime **after** `first_block_height`. I was under the impression that ensuring that we already did process some sortitions would be a robust way to check the required restriction.

src/chainstate/coordinator/mod.rs
jcnelson: I'd much rather prefer the query, since then we never have to worry about the `num_sortitions` argument being wrong (neither now nor in the future).

src/burnchains/burnchain.rs
jcnelson: The number of sortitions that have occurred is always equal to `current_burn_block_height - self.first_block_height`, because the sortition DB calculates a sortition for each burnchain block even if no Stacks block was mined.  Unless something has changed in SIP-007, reward cycles are defined as fixed intervals of burnchain blocks, starting from `self.first_block_height` (cc @kantai).

kantai: Hmmm.

So the intention of this check was to ensure that we're not starting a "new" reward cycle 1 block after genesis, so that we didn't have a 1 block reward cycle (comprising just the genesis block), followed by a burn reward cycle. Using `num_sortitions` achieves somewhat close to the same thing, except in the event that the "first sortition winner" is chosen > 1 reward cycle after `first_block_height`. Thinking about this more, though, I think it should be fine to allow that case -- after all, the code needs to deal with the case that a reward cycle didn't have any miners anyways, and allowing this case makes it so that we don't need to pass `num_sortitions` into this method.

lgalabru: @kantai ha, just saw your comment.
Should we also get rid of the check:
```

        if effective_height == 0 {
            debug!(
                "effective_height = {}, reward cycle length == {}",
                effective_height, pox_consts.reward_cycle_length
            );
            return Err(CoordinatorError::NotPrepareEndBlock);
        }
```
?

lgalabru: Addressed with latest

src/chainstate/coordinator/mod.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2233/commits/321fb172ae50c6a4921169d43725521e536fc19d

src/burnchains/burnchain.rs
kantai: Ooops -- I think maybe I wasn't very clear. I think that the original checks should be preserved, so `is_reward_cycle_start` would be unchanged versus next, i.e.:

```
    pub fn is_reward_cycle_start(&self, burn_height: u64) -> bool {
        if burn <= (self.first_block_height + 1) {
            // not a reward cycle start if we're the first block after genesis.
            false
        } else {
            let effective_height = burn_height - self.first_block_height;
            // first block of the new reward cycle
            (effective_height % (self.pox_constants.reward_cycle_length as u64)) == 1
        }
    }
```

And this check should still exist:

```rust
        if effective_height == 0 {
            debug!(
                "effective_height = {}, reward cycle length == {}",
                effective_height, pox_consts.reward_cycle_length
            );
            return Err(CoordinatorError::NotPrepareEndBlock);
        }
```

Because that makes sure that we don't start a prepare "end" in the first block either (as that would cause the node to try and count prepare blocks before the `first_burn_block`)

src/burnchains/burnchain.rs
```diff
@@ -487,15 +491,10 @@ impl Burnchain {
...
-            false
-        } else {
-            let effective_height = block_height - self.first_block_height;
-            // first block of the new reward cycle
-            (effective_height % (self.pox_constants.reward_cycle_length as u64)) == 1
-        }
+    pub fn is_reward_cycle_start(&self, burn_height: u64) -> bool {
+        let effective_height = burn_height - self.first_block_height;
+        // first block of the new reward cycle
+        (effective_height % (self.pox_constants.reward_cycle_length as u64)) == 1
```
kantai, line 497:
```suggestion
    pub fn is_reward_cycle_start(&self, burn_height: u64) -> bool {
        if burn_height <= (self.first_block_height + 1) {
            // not a reward cycle start if we're the first block after genesis.
            false
        } else {
            let effective_height = burn_height - self.first_block_height;
            // first block of the new reward cycle
            (effective_height % (self.pox_constants.reward_cycle_length as u64)) == 1
        }
```

src/burnchains/burnchain.rs
kantai: Thinking through this more -- it isn't essential that `is_reward_cycle_start` have this check -- it just means that there'll be an extra bit in the PoX bit vector, which isn't a problem in itself.

The other check though, should exist.

lgalabru: Check introduced in https://github.com/blockstack/stacks-blockchain/pull/2233/commits/438e0e7ab9dca697ff1ae6d26f61f48b02a5bf00

src/net/rpc.rs
lgalabru: Not sure, I couldn't find what I was looking for.

kantai (approved):
Looks good to me!

jcnelson (approved):
LGTM!

PR #2234, opened 2020-12-30
Developer / Miner Experience: improving logs

lgalabru:
Working on #2233 implied testing on testnet, meaning very slow feedback loops. As a result, I've spent quite some time staring at the logs, and decided to do something about it.
The `BLOCKSTACK_DEBUG` mode is really verbose, and the `BLOCKSTACK_DEBUG`-less mode is weird and giving an impression of flakiness, with very long pauses, or logs being repeated over and over, giving the impression that something is wrong.
This PR is iterating on:

- the logs themselves - it probably needs a few more iterations, my main focus was the boot sequence, downgrading some of our logs to `debug`, and adding some sense of progression on the bitcoin sync and the MARF bulk writes.

Screen Shot 2020-12-29 at 11 40 33 PM

- the way we show logs, by introducing a `STACKS_LOG_PP` mode, that looks like this:
Screen Shot 2020-12-30 at 12 35 31 AM

this mode can still be combined with the `BLOCKSTACK_DEBUG` (renamed to `STACKS_LOG_DEBUG`), for displaying the debug logs + source/line/thread informations.

Screen Shot 2020-12-30 at 12 36 45 AM

In terms of dependencies, this PR is explicitly requiring `chrono`, which was previously pulled by `slog` - nothing new. 

This new design was very helpful when investigating on #2233, I personally believe that we should make this `pretty print` default :)


src/burnchains/bitcoin/spv.rs
```diff
@@ -907,13 +907,27 @@ impl BitcoinMessageHandler for SpvClient {
...
-                    block_height + 2000,
-                    self.start_block_height,
-                    end_block_height
-                );
+                // if syncing requires to request more than one batch of 2000 headers, 
+                // we'll provide some progress in the logs
+                let total = end_block_height - self.start_block_height;
+                let batch_size = 2000;
+                if total > batch_size {
+                    let progress = (block_height - self.start_block_height) as f32 / total as f32 * 100.;
```
zone117x, line 915: Progress indicators are super helpful 💯

src/util/log.rs
```diff
@@ -61,6 +64,37 @@ fn print_msg_header(mut rd: &mut dyn RecordDecorator, record: &Record) -> io::Re
     Ok(count_rd.count() != 0)
 }
 
+fn pretty_print_msg_header(rd: &mut dyn RecordDecorator, record: &Record, debug: bool) -> io::Result<bool> {
+    rd.start_timestamp()?;
+    let now: DateTime<Utc> = Utc::now();
+    write!(
+        rd, "\x1b[0;90m{}", now.format("%b %e %T"))?;
```
zone117x, line 71: The pretty print looks great. I don't think this will work on Windows console by default, a win32 API has to be called at runtime to enable it, otherwise I think it will just print the escape codes.

jcnelson: Please remove terminal coloration altogether, or gate it with a feature flag or an environment variable.  It should be off by default.

First, when you have multiple threads writing to the same terminal device file (i.e. the pty behind your console), you can easily break coloration if one thread yields to another thread without first writing the escape sequence to disable coloration.  This can leave your console output looking corrupted.

Second, you're not using the `isatty(2)` check here, so if I redirect logging to a file, it will be filled with these escape sequences.

lgalabru: As explained in the introduction, this is off by default - gated with `STACKS_LOG_PP`

jcnelson: Ah, just saw that.  But, you'll still want want to guard this with `isatty(2)`.  In general, piping or redirecting the output of a program ought to disable coloration, since you (almost) never want to see the escape codes as actual text.

lgalabru: Yeah definitely will, thanks for the tip

zone117x: If we perform the `isatty` check, can we also enable the Win32 API for coloring? Would be super helpful. I for one would be running a Stacks 2.0 on Windows with pretty print enabled.

src/chainstate/burn/db/processing.rs
```diff
@@ -152,13 +152,13 @@ impl<'a> SortitionHandleTx<'a> {
         let mut next_pox = parent_pox;
         if let Some(ref next_pox_info) = next_pox_info {
             if next_pox_info.is_reward_info_known() {
-                info!(
+                debug!(
```
kantai, line 155: These are nice as `info` logs, right? Reward cycle starts are fairly rare.

lgalabru: Yeah we definitely have logs for new reward cycles `sortdb.rs:2857`, I believe.

src/util/log.rs
lgalabru: Addressed with latest. @zone117x I don't have what is needed for testing a windows setup, feel free to push a commit / follow-up with something tested :)

src/chainstate/burn/db/processing.rs
kantai: Hmm -- I think at the very least though, we should have an `info` log for when we start a reward cycle _without_ the anchor block.

lgalabru: Sounds good will add

testnet/stacks-node/src/main.rs
```diff
@@ -77,6 +77,13 @@ fn main() {
     let mut args = Arguments::from_env();
     let subcommand = args.subcommand().unwrap().unwrap_or_default();
 
+    let server_version = stacks::version_string(
+        option_env!("CARGO_PKG_NAME").unwrap_or("stacks-node"),
+        option_env!("CARGO_PKG_VERSION").unwrap_or("0.0.0.0"),
```
kantai, line 82: I believe the way we generate the `version_string` changed a little bit, we use:

```rust
                &stacks::version_string(
                    "stacks-node",
                    option_env!("STACKS_NODE_VERSION")
                        .or(option_env!("CARGO_PKG_VERSION"))
                        .unwrap_or("0.0.0.0")
                )
```

in the `--version` output in this file.

testnet/stacks-node/src/neon_node.rs
```diff
@@ -409,15 +400,8 @@ fn try_mine_microblock(
...
-                                if err_str.contains("NoTransactionsToMine") {
-                                    trace!("Failed to mine microblock because there are no transactions to mine");
-                                } else {
-                                    warn!(
-                                        "Failed to mine one microblock: {:?}",
-                                        &NetError::ChainstateError(err_str)
-                                    );
-                                }
+                            Err(ChainstateError::NoTransactionsToMine) => {
+                                info!("Will keep polling mempool for transactions to include in a microblock");
```
kantai, line 404: Should this be an `info!` log or `debug!` log (or even `trace!` log)? `NoTransactionsToMine` is pretty noisy, and doesn't provide much information other than the fact that the mempool is currently empty.

testnet/stacks-node/src/neon_node.rs
```diff
@@ -794,7 +778,7 @@ fn spawn_miner_relayer(
                                 && burn_hash == mined_burn_hash
                             {
                                 // we won!
-                                info!("Won sortition!";
+                                debug!("Won sortition!";
```
kantai, line 781: I'd suggest making this an `info!` log, everyone likes to know when they win!

kantai (approved):
These changes look good to me -- just a few nitpicks about some level changes.

We should also note that changing the env vars will have impacts for deployments (deployments that plan to consume JSON and DEBUG logs will have to change their env var settings).

testnet/stacks-node/src/neon_node.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2234/commits/445e25c5863951ce0606d59b4cf66b08529bf801

testnet/stacks-node/src/neon_node.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2234/commits/445e25c5863951ce0606d59b4cf66b08529bf801

testnet/stacks-node/src/main.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2234/commits/445e25c5863951ce0606d59b4cf66b08529bf801

src/chainstate/burn/db/processing.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2234/commits/445e25c5863951ce0606d59b4cf66b08529bf801

Cargo.toml
```diff
@@ -61,6 +61,8 @@ integer-sqrt = "0.1.3"
 slog = { version = "2.5.2", features = [ "max_level_trace" ] }
 slog-term = "2.6.0"
 slog-json = { version = "2.3.0", optional = true }
+chrono = "0.4.19"
+libc = "0.2.82"
```
jcnelson, line 65: This is already a transitive dependency, right?

src/util/log.rs
```diff
@@ -222,3 +305,31 @@ macro_rules! fatal {
         }
     })
 }
+
+fn color_if_atty(color: &str, isatty: bool) -> &str {
```
jcnelson, line 309: Minor nit -- `isatty(2)` is short for `is a tty` (a `tty` being the abbreviation for "teletype" device).  Might want to rename this to `color_if_tty`.

Cargo.toml
lgalabru: Correct

src/util/log.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2234/commits/7219536b3a6293a74ae0439c681ea4446bebd64c

jcnelson (approved):
LGTM!

jcnelson:
Just want to confirm -- `BLOCKSTACK_DEBUG` will still work, right?  Or, does devops know to use `STACKS_LOG_DEBUG`?

lgalabru:
Devops already changed their setup to use STACKS_LOG_DEBUG

PR #2235, opened 2020-12-30
Update Xenon magic bytes for reset

kantai:
This updates the magic bytes for Xenon, which will cause a Xenon reset.


jcnelson (approved):

lgalabru (approved):
Looks good to me!

kantai:
Merging -- these two neon_integration tests pass locally for me, so I think this is just test flake at this point.

PR #2238, opened 2020-12-30
Fix (append ...) documentation

lgalabru:
This PR is addressing https://github.com/blockstack/stacks-blockchain/issues/2231.


jcnelson (approved):
LGTM.  But, how does one append to a buffer now?  Is it through `concat`?

kantai (approved):

Issue #2240, opened 2020-12-31
Xenon crash: `blockstack_lib::chainstate::coordinator::get_reward_cycle_info`


diwakergupta:
Stack trace:

```
   0: backtrace::backtrace::libunwind::trace
             at /Users/diwaker/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.50/src/backtrace/libunwind.rs:95
      backtrace::backtrace::trace_unsynchronized
             at /Users/diwaker/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.50/src/backtrace/mod.rs:66
      backtrace::backtrace::trace
             at /Users/diwaker/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.50/src/backtrace/mod.rs:53
   1: backtrace::capture::Backtrace::create
             at /Users/diwaker/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.50/src/capture.rs:164
      backtrace::capture::Backtrace::new
             at /Users/diwaker/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.50/src/capture.rs:128
   2: stacks_node::main::{{closure}}
             at /Users/diwaker/Vault/blockstack/stacks-blockchain/testnet/stacks-node/src/main.rs:57
   3: std::panicking::rust_panic_with_hook
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:581
   4: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:484
   5: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/sys_common/backtrace.rs:153
   6: rust_begin_unwind
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:483
   7: core::panicking::panic_fmt
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/panicking.rs:85
   8: core::option::expect_failed
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/option.rs:1226
[...]

PR #2242, opened 2020-12-31
Fix: PoX Anchor Counting

kantai:
The PoX anchor selection logic assumes it can always find the parent of a sortition winner -- this is not true if the parent is the genesis block (this leads to a panic #2240)

### DIFF
--- a/testnet/stacks-node/src/neon_node.rs
+++ b/testnet/stacks-node/src/neon_node.rs
@@ -419,6 +419,16 @@ fn try_mine_microblock(
                                 // will need to relay this
                                 next_microblock = Some(microblock);
                             }
+                            Err(NetError::ChainstateError(err_str)) => {
+                                if err_str.contains("NoTransactionsToMine") {
+                                    trace!("Failed to mine microblock because there are no transactions to mine");
+                                } else {
+                                    warn!(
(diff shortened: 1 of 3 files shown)


testnet/stacks-node/src/neon_node.rs
```diff
@@ -419,6 +419,16 @@ fn try_mine_microblock(
                                 // will need to relay this
                                 next_microblock = Some(microblock);
                             }
+                            Err(NetError::ChainstateError(err_str)) => {
```
lgalabru, line 422: If not done in this PR, https://github.com/blockstack/stacks-blockchain/pull/2234 will turn the `NetError` returned by this function into a `ChainStateError` so we don't have to match against a string. For some reason, all the codepaths are returning ChainStateErrors, but they are being wrapped into a net_error

kantai: Ah -- yeah, that's better. I'm going to leave this check in this PR, and we can switch to matching with the ChainstateError enum in #2234 later on, because I'm irrationally annoyed by this warn :)

lgalabru (approved):
Thank you @kantai!

diwakergupta (approved):
Ship it!

PR #2244, opened 2020-12-31
Test fixes

kantai:
Opening this PR to try and fix some of the current test errors we're seeing. So far:

1. Fix the stack overflow caused by one of the pipe util tests.
2. DRY `neon_integrations` and `clear()` the event observer between runs.

### DIFF
--- a/src/util/pipe.rs
+++ b/src/util/pipe.rs
@@ -412,7 +412,8 @@ mod test {
 
     #[test]
     fn test_connection_pipe_producer_consumer() {
-        let mut buf = vec![0u8; 1048576]; // 1 MB
+        let mut buf = Vec::new();
+        buf.resize(1048576, 0);
 
         let mut rng = rand::thread_rng();
         rng.fill_bytes(&mut *buf);
--- a/src/util/pipe.rs
+++ b/src/util/pipe.rs
@@ -412,7 +412,8 @@ mod test {
 
     #[test]
     fn test_connection_pipe_producer_consumer() {
-        let mut buf = Box::new([0u8; 1048576]); // 1 MB
+        let mut buf = vec![0u8; 1048576]; // 1 MB
+
         let mut rng = rand::thread_rng();
         rng.fill_bytes(&mut *buf);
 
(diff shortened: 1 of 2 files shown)


src/util/pipe.rs
```diff
@@ -412,7 +412,8 @@ mod test {
 
     #[test]
     fn test_connection_pipe_producer_consumer() {
-        let mut buf = Box::new([0u8; 1048576]); // 1 MB
+        let mut buf = vec![0u8; 1048576]; // 1 MB
```
jcnelson, line 415: Wait, shouldn't this continue to be a box?  I can't even run this test locally without it -- I get a segfault due to stack overflow (Alpine + musl libc)

kantai: The `Box` version segfaults with stack overflow in `rust:buster` docker images (notably: Circle CI and the unit test PR that ports the Circle tests over to github actions both fail with a stack overflow with the box version). I am not 100% sure _why_ Box allocation does this (we'd need to look at the llvm codegen, I think), but it seems like the buffer is allocated on the heap and copied from the stack when using `Box`.

kantai: Either way-- the existing test is broken in CI and should be fixed or removed.

jcnelson: Can you try this?

```
// force it to be heap-allocated
let mut buf = Box::new(Vec::with_capacity(1048576));
for _ in 0..(1048576 / 4096) {
   // fill in one page at a time
   buf.append(&mut vec![0u8; 4096]);
}
```

kantai: What about the latest commit on musl? I don't think Boxing a Vec should have any impact on heap allocation -- Vecs are always heap allocated. I think it may just have been the `vec![]` macro that broke on musl.

jcnelson: Ah, yup -- it works now on latest musl.

jcnelson (approved):
LGTM!

lgalabru (approved):
Looks good to me!

PR #2255, opened 2021-01-04
Do not abort mining a block if the microblock stream failed to load

kantai:
The miner should be able to continue mining an anchor block even if the parent microblock stream fails to load -- it should default to mining without confirming the parent microblock stream.


jcnelson (approved):
LGTM!

PR #2260, opened 2021-01-04
Immediate updates of `total-liquid-ustx`

### LINKED ISSUE #2232: `stx-liquid-supply` broken
`stx-liquid-supply` evaluation is currently broken: it attempts to read from the HeadersDB at the current block height. This fails, causing the `clarity_db` to return `0` in all cases.

A fix and witnessing test are found in 8cbdabc71 (PR: #2227)

kantai:
This fixes #2232 by using the clarity_db MARF to store the current liquid supply of STX.

This PR also tries to play some tricks with our current integration tests workflows:
1. It creates a separate workflow for the integration tests, which allows them to be re-run without re-running the other jobs (the dist and build-release jobs are non-trivial executions).
2. It runs the tests in separate `cargo test` commands. This solves any issues that may be caused by using lazy_statics to track test state _and_ it makes the tests "fail fast".
3. It separates the atlas integration test into its own test runner.

### DIFF
--- a/.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
+++ b/.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
@@ -16,7 +16,6 @@ WORKDIR /src/testnet/stacks-node
 RUN cargo test -- --test-threads 1 --ignored tests::integrations::integration_test_get_info
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::bitcoind_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::size_check_integration_test
-RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::atlas_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::liquid_ustx_integration
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::stx_transfer_btc_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::bitcoind_forking_test
--- a/.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
+++ b/.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
@@ -12,4 +12,15 @@ RUN cd / && tar -xvzf bitcoin-0.20.0-x86_64-linux-gnu.tar.gz
 RUN ln -s /bitcoin-0.20.0/bin/bitcoind /bin/
 
 ENV BITCOIND_TEST 1
-RUN cd testnet/stacks-node && cargo test -- --test-threads 1 --ignored
+WORKDIR /src/testnet/stacks-node
+RUN cargo test -- --test-threads 1 --ignored tests::integrations::integration_test_get_info
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::bitcoind_integration_test
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::size_check_integration_test
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::atlas_integration_test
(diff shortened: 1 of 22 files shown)


.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
```diff
@@ -12,4 +12,14 @@ RUN cd / && tar -xvzf bitcoin-0.20.0-x86_64-linux-gnu.tar.gz
...
+RUN cargo test -- --test-threads 1 --ignored tests::integrations::integration_test_get_info
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::bitcoind_integration_test
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::size_check_integration_test
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::liquid_ustx_integration
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::stx_transfer_btc_integration_test
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::bitcoind_forking_test
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::microblock_integration_test
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::pox_integration_test
+RUN cargo test -- --test-threads 1 --ignored tests::bitcoin_regtest::bitcoind_integration_test
+RUN cargo test -- --test-threads 1 --ignored tests::should_succeed_handling_malformed_and_valid_txs
```
jcnelson, line 25: Is there a way to programmatically enumerate this list, so we don't accidentally forget to add new ones to the test runner as they get created?  Same question for Atlas tests.

src/chainstate/coordinator/mod.rs
```diff
@@ -211,20 +211,15 @@ impl RewardSetProvider for OnChainRewardSetProvider {
...
+        let liquid_ustx = chainstate.get_liquid_ustx(block_id);
 
         let (threshold, participation) = StacksChainState::get_reward_threshold_and_participation(
             &burnchain.pox_constants,
             &registered_addrs,
             liquid_ustx,
         );
 
-        test_debug!("PoX reward cycle threshold: {}, participation: {}, liquid_ustx: {}, num registered addrs: {}", threshold, participation, liquid_ustx, registered_addrs.len());
+        info!("PoX reward cycle threshold: {}, participation: {}, liquid_ustx: {}, num registered addrs: {}", threshold, participation, liquid_ustx, registered_addrs.len());
```
jcnelson, line 222: Structured logging? ;)

src/chainstate/stacks/miner.rs
```diff
@@ -761,23 +760,44 @@ impl StacksBlockBuilder {
...
-            StacksChainState::process_matured_miner_rewards(
+            let matured_ustx = StacksChainState::process_matured_miner_rewards(
                 clarity_tx,
                 miner_reward,
                 user_rewards,
                 parent_reward,
             )
             .expect("FATAL: failed to process miner rewards");
+
+            clarity_tx
```
jcnelson, line 777: I've seen this code block a few times.  Should `clarity_tx` just have its own method for incrementing the total uSTX supply?

jcnelson (approved):
This overall LGTM; just a few minor comments.  Thanks @kantai!

jcnelson:
It looks like the `sampled-genesis` test runner is still running the Atlas integration test?

.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
kantai: No, I don't believe so

kantai:
> It looks like the sampled-genesis test runner is still running the Atlas integration test?

It should only be running in the `stacks-bitcoin-integration-tests/atlas-test` runner, so now it has it's own docker container. The integration tests are all still woefully flaky, but at least now he execution ends more quickly.

lgalabru (approved):

Issue #2257, opened 2021-01-04
Clarity runtime errors should not abort block


kantai:
Clarity runtime errors _should not_ result in an invalid block, or cause a miner to not include the transaction. This incorrect behavior is currently relied on by tests in `chainstate::stacks::boot` that use `unwrap-panic!` to abort the transaction, which is then checked to ensure that the transaction is not included in the block.

diwakergupta:
Fixed by #2264

PR #2264, opened 2021-01-05
Smart contract transaction error handling

kantai:
This fixes the behavior in #2257

### DIFF
--- a/src/chainstate/stacks/db/transactions.rs
+++ b/src/chainstate/stacks/db/transactions.rs
@@ -856,15 +856,32 @@ impl StacksChainState {
 
                 let (result, asset_map, events) = match contract_call_resp {
                     Ok((return_value, asset_map, events)) => {
-                        info!("Contract-call to {}.{:?} args {:?} returned {:?}", &contract_id, &contract_call.function_name, &contract_call.function_args, &return_value);
-                        info!("Contract-call to {}.{:?} args {:?} cost {:?}", &contract_id, &contract_call.function_name, &contract_call.function_args, &total_cost);
+                        info!("Contract-call successfully processed";
+                              "contract_name" => %contract_id,
+                              "function_name" => %contract_call.function_name,
+                              "function_args" => ?contract_call.function_args,
+                              "return_value" => %return_value,
+                              "cost" => ?total_cost);
                         Ok((return_value, asset_map, events))
                     },
                     Err(e) => {
                         match e {
                             // runtime errors are okay -- we just have an empty asset map
                             clarity_error::Interpreter(InterpreterError::Runtime(ref runtime_error, ref stack)) => {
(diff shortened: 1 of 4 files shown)


src/chainstate/stacks/db/transactions.rs
```diff
@@ -856,15 +856,32 @@ impl StacksChainState {
...
-                                info!("Runtime error {:?} on contract-call {}.{:?} {:?}, stack trace {:?}", runtime_error, &contract_id, &contract_call.function_name, &contract_call.function_args, stack);
+                                info!("Contract-call processed with error";
+                                      "contract_name" => %contract_id,
+                                      "function_name" => %contract_call.function_name,
+                                      "function_args" => ?contract_call.function_args,
+                                      "error" => ?runtime_error,
+                                      "stack" => ?stack);
+                                Ok((Value::err_none(), AssetMap::new(), vec![]))
+                            },
+                            clarity_error::Interpreter(InterpreterError::ShortReturn(ref short_ret)) => {
```
jcnelson, line 879: There aren't any other such `InterpreterError`s we should handle here, right?

jcnelson (approved):
LGTM, but maybe there's some space to DRY up the error-handling code?  Right now, it looks like we catch all four non-block-invalidating Clarity errors twice.

src/chainstate/stacks/db/transactions.rs
kantai: I don't think so -- the other error kinds are Unchecked (which should have been prevented through analysis) and Interpreter (which should not occur at all), both of which a miner should not be allowed to mine a block with.

kantai:
I factored the error handling in 098fa3b6d -- there's still a bit of duplication in the two locations, but that's mostly due to the logging code, which should be different between them. The decision about which errors are acceptable versus rejectable is factored out, though.

src/chainstate/stacks/db/transactions.rs
```diff
@@ -2123,6 +2169,56 @@ pub mod test {
         conn.commit_block();
     }
 
+    #[test]
+    fn process_smart_contract_user_aborts_2257() {
+        let contract = "(asserts! false (err 1))";
```
lgalabru, line 2174: I think it would be super useful to see a test with all the different behavior - success, abort, panic - as a direct call,  in a nested `contract-call` - testing for writes being rollbacked and transaction inclusion.

lgalabru (approved):
Good catch, thank you @kantai!

src/chainstate/stacks/db/transactions.rs
kantai: I would agree -- and think we can add more test coverage for that after the next xenon reset. If that testing turns up an issue here, I think we can figure it out at that point.

Issue #2258, opened 2021-01-04
Net: unconfirmed microblock processing can prevent network activity


jcnelson:
If the time taken to create the unconfirmed microblock state trie takes too long, it can starve the network by preventing any network activity from completing in the allotted time.

The solution would be:
* Be less aggressive about rebuilding the unconfirmed microblock state trie.  A low-hanging win would be to not try and do anything with it unless there are actually new microblocks to deal with, and to throttle the rate at which we consider new microblocks.
* Move microblock-mining and unconfirmed miroblock-processing out of the p2p thread, and into the relayer thread.

jcnelson:
Fixed by #2266

PR #2266, opened 2021-01-05
Fix/net slow lock

jcnelson:
This PR addresses #2258 by moving management of unconfirmed state and microblock mining back into the relayer thread (now that it's safe to do this).

The p2p thread will drive microblock mining using a `RunMicroblockTenure` relayer directive, which will be sent to wake up the relayer every so often to mine the next microblock (according to the microblock interval from the config file).

The p2p thread can no longer manage the unconfirmed state trie on its own, so I modified the `UnconfirmedState` struct to have a "read-only" mode that will allow it to be used to query an existing trie in the underlying MARF.  To make this happen safely, I also added a new Clarity API method `begin_read_only_checked()`, which unlike `begin_read_only()` can fail if, for example, the given trie doesn't exist.  This is necessary because the relayer thread can delete the unconfirmed trie while the p2p thread is using it to service requests; these requests should fail gracefully if this happens.

Because the p2p thread needs to know which transactions went into the unconfirmed state trie (so it can handle `/v2/transactions/unconfirmed/{:txid}`), I took the path of least resistance and made it so that the relay thread will share a copy of its unconfirmed state's mined transaction list with the p2p thread via a mutex guard.  It's not the prettiest solution, but it requires fewer lines of code an introduces fewer failure modes than the alternatives I could think of.

### DIFF
--- a/src/vm/database/marf.rs
+++ b/src/vm/database/marf.rs
@@ -294,6 +294,27 @@ impl MarfedKV {
         }
     }
 
+    pub fn begin_read_only_checked<'a>(
+        &'a mut self,
+        at_block: Option<&StacksBlockId>,
+    ) -> Result<ReadOnlyMarfStore<'a>> {
+        let chain_tip = if let Some(at_block) = at_block {
+            self.marf.open_block(at_block).map_err(|_| {
+                debug!("Failed to open read only connection at {}", at_block);
+                InterpreterError::MarfFailure(IncomparableError {
+                    err: MarfError::NotFoundError,
+                })
+            })?;
+            at_block.clone()
+        } else {
+            self.chain_tip.clone()
+        };
+        Ok(ReadOnlyMarfStore {
+            chain_tip,
+            marf: &mut self.marf,
+        })
+    }
+
     /// begin, commit, rollback a save point identified by key
     ///    this is used to clean up any data from aborted blocks
     ///     (NOT aborted transactions that is handled by the clarity vm directly).
--- a/testnet/stacks-node/src/neon_node.rs
+++ b/testnet/stacks-node/src/neon_node.rs
@@ -8,6 +8,7 @@ use std::collections::{HashSet, VecDeque};
 use std::convert::{TryFrom, TryInto};
 use std::default::Default;
 use std::net::SocketAddr;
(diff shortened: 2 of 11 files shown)


src/vm/database/marf.rs
```diff
@@ -294,6 +294,27 @@ impl MarfedKV {
         }
     }
 
+    pub fn begin_read_only_checked<'a>(
```
kantai, line 297: Can you DRY `begin_read_only<'a>` by calling `.expect()` on this new function?

testnet/stacks-node/src/neon_node.rs
```diff
@@ -885,22 +928,26 @@ fn spawn_miner_relayer(
...
-                    if let Err(e) = relayer.broadcast_microblock(
-                        &parent_consensus_hash,
-                        &parent_block_hash,
-                        microblock,
-                    ) {
-                        error!(
-                            "Failure trying to broadcast microblock {}: {}",
-                            microblock_hash, e
-                        );
+                RelayerDirective::RunMicroblockTenure => {
```
kantai, line 931: Can `RunMicroblockTenure` directives be coalesced? As implemented, I think we'd end up with many `RunMicroblockTenure` directives in the channel at the same time, leading the mining thread to try mining a microblock tenure continuously, when the mempool is unlikely to have changed (and these 0-tx attempts to mine a microblock are not cheap).

jcnelson: Sure -- I can make the relayer aware of the expected time between microblocks and ignore these directives if the time since the last microblock was mined is earlier than the deadline to start mining the next one.

jcnelson: Fixed in 1d977c1

src/vm/database/marf.rs
jcnelson: Fixed in 1d977c1

kantai (approved):
This LGTM! Just make sure that the neon_integrations microblock test still works with the latest change.

jcnelson:
It passes locally at least.

lgalabru (approved):
LGTM!

jcnelson:
Flaky tests pass locally.

PR #2269, opened 2021-01-05
Update Xenon 1st bitcoin block height

lgalabru:
To merge before the next consensus breaking deployment.


jcnelson:
Pretty sure this should be opened against `next`, right?

kantai (approved):

jcnelson (approved):

PR #2271, opened 2021-01-05
Fix 2270

jcnelson:
Don't panic just because a network dispatch failed.  Instead, just drop the network dispatch result and proceed to drain the request queue to the relayer as if nothing happened.

Removes one explicit panic from the code.

### DIFF
--- a/testnet/stacks-node/src/neon_node.rs
+++ b/testnet/stacks-node/src/neon_node.rs
@@ -614,7 +614,7 @@ fn spawn_peer(
                 let _ = Relayer::setup_unconfirmed_state_readonly(&mut chainstate, &sortdb);
                 recv_unconfirmed_txs(&mut chainstate, unconfirmed_txs.clone());
 
-                let network_result = match this.run(
+                match this.run(
                     &sortdb,
                     &mut chainstate,
                     &mut mem_pool,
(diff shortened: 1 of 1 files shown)


jcnelson:
Okay this is weird -- I can't select @lgalabru as a reviewer.

kantai (approved):

testnet/stacks-node/src/neon_node.rs
```diff
@@ -624,35 +624,37 @@ fn spawn_peer(
                     &handler_args,
                     &mut expected_attachments,
                 ) {
-                    Ok(res) => res,
+                    Ok(network_result) => {
```
lgalabru, line 627: I think we might want to mitigate this approach. If the node is being dispatching its events to an event observer, crash, if not continue?
Otherwise, the event observer could potentially miss a few things:

```rust
RelayerDirective::HandleNetResult(ref mut net_result) => {
                    debug!("Relayer: Handle network result");
                    let net_receipts = relayer
                        .process_network_result(
                            &local_peer,
                            net_result,
                            &mut sortdb,
                            &mut chainstate,
                            &mut mem_pool,
                            Some(&coord_comms),
                        )
                        .expect("BUG: failure processing network results");

                    let mempool_txs_added = net_receipts.mempool_txs_added.len();
                    if mempool_txs_added > 0 {
                        event_dispatcher.process_new_mempool_txs(net_receipts.mempool_txs_added);
                    }

                    // Dispatch retrieved attachments, if any.
                    if net_result.has_attachments() {
                        event_dispatcher.process_new_attachments(&net_result.attachments);
                    }

                    // synchronize unconfirmed tx index to p2p thread
                    send_unconfirmed_txs(&chainstate, unconfirmed_txs.clone());
                }
```

jcnelson:
Got the all-clear to merge from @diwakergupta, given that this has now triggered a crash in production.  @lgalabru can you please open an issue for this so we can address it in a separate PR?

PR #2273, opened 2021-01-05
E2E Integration test for cost-voting

kantai:
This adds an end-to-end integration test for the cost-voting contract. It requires two changes:

1. Setting a different veto period for testnet and mainnet. To avoid causing a consensus change in mainnet, this code does string replacement to apply the difference in testnet and other testing environments.
2. Reporting the execution cost of contracts in transaction receipts (@zone117x -- does adding an extra field here break processing on the sidecar, or will it just ignore the new field?)


zone117x:
> does adding an extra field here break processing on the sidecar, or will it just ignore the new field?

It will be ignored 👍

reedrosenbluth (approved):

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1241,6 +1244,292 @@ fn size_check_integration_test() {
...
+    assert_eq!(res.balance, spender_bal as u128);
+    assert_eq!(res.nonce, 0);
+
+    let transactions = vec![
+        make_contract_publish(&spender_sk, 0, 1000, "cost-definer", cost_definer_src),
+        make_contract_publish(&spender_sk, 1, 1000, "caller", caller_src),
+        make_contract_publish(&spender_sk, 2, 1000, "voter", power_vote_src),
+    ];
+
+    // okay, let's push that stacking transaction!
```
jcnelson, line 1353: stacking?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1241,6 +1244,292 @@ fn size_check_integration_test() {
...
+                let raw_result = tx.get("raw_result").unwrap().as_str().unwrap();
+                let parsed = <Value as ClarityDeserializable<Value>>::deserialize(&raw_result[2..]);
+                assert_eq!(parsed.to_string(), "(ok u0)");
+                tested = true;
+            }
+        }
+    }
+    assert!(tested, "Should have found a contract call tx");
+
+    // try to confirm the passed vote (this will fail)
```
jcnelson, line 1421: How does this test know that this first `confirm_proposal` fails and the next `confirm_proposal` succeeds?  Isn't the first `confirm_proposal`'s failure masked?

jcnelson (approved):
This LGTM, but I had one question about the test design.

testnet/stacks-node/src/tests/neon_integrations.rs
kantai: Ah, yeah, this first one isn't checked -- the second one is checked via the event observer. A check could be added here via the event observer.

kantai: Resolved in 3877c953992d9259550db48b5b251cce389aaa75

testnet/stacks-node/src/tests/neon_integrations.rs
kantai: Copy-pasta -- fixed in 3877c95

PR #2276, opened 2021-01-05
Update build metadata. don't trigger from pushed tags anymore

CharlieC3:
* Updates the package name and version displayed in the `/v2/info` endpoint.
    * The `server_version` at `/v2/info` and `stacks-node version` command should now read something similar to:
    ```bash
    "stacks-node v24.3.2.0-xenon (build-metadata:0087561+, release build, linux [x86_64])"
    ```

    ```bash
    docker run --rm build-metadata-stretch stacks-node version
    stacks-node v24.3.2.0-xenon (build-metadata:0087561+, release build, linux [x86_64])
    ```

Screen Shot 2021-01-06 at 10 59 15 PM

* Disables triggering a release when a tag is pushed up. cc @kantai 
    * I disabled this because since we added a GH workflow which creates a new `clarity-js-sdk` PR upon each `stacks-blockchain` release, a new tag that gets created by the standard release workflow will create a loop, causing the release workflow to run again and attempt re-releasing the version.
    * With this disabled, we will no longer be able to push up a Git tag created locally to trigger a new release. [This GH workflow will need to be invoked manually](https://github.com/blockstack/stacks-blockchain/actions?query=workflow%3Astacks-blockchain), with a new tag passed into the textbox.
    * @kantai if you want this to remain enabled, an alternative solution is possible but will require more testing to ensure it doesn't break the clarity-js-sdk-pr workflow.

Closes https://github.com/blockstack/stacks-blockchain/issues/2275
Closes https://github.com/blockstackpbc/devops/issues/554

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] API reference/documentation update
- [X] Other

No

No

## Testing information

1. Run a new node via `docker run -p 20443:20443 --rm blockstack/stacks-blockchain:build-metadata-stretch`
2. Open `localhost:20443/v2/info` in a browser once to initial boot finishes.

## Checklist
- [ ] Code is commented where needed
- [ ] Unit test coverage for new or modified code paths
- [ ] `npm run test` passes
- [ ] Changelog is updated
- [ ] Tag 1 of @person1 or @person2

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -8,13 +8,12 @@ fn current_git_hash() -> Option<String> {
             .arg("--pretty=format:%h") // Abbreviated commit hash
             .current_dir(env!("CARGO_MANIFEST_DIR"))
             .output();
-    
+
         if let Ok(commit) = commit {
             if let Ok(commit) = String::from_utf8(commit.stdout) {
                 return Some(commit);
             }
         }
-
     } else {
         return option_env!("GIT_COMMIT").map(String::from);
     }
--- a/build.rs
+++ b/build.rs
@@ -1,32 +1,43 @@
 use std::process::Command;
 
 fn current_git_hash() -> Option<String> {
-    let commit = Command::new("git")
-        .arg("log")
-        .arg("-1")
-        .arg("--pretty=format:%h") // Abbreviated commit hash
-        .current_dir(env!("CARGO_MANIFEST_DIR"))
-        .output();
-
-    if let Ok(commit) = commit {
-        if let Ok(commit) = String::from_utf8(commit.stdout) {
-            return Some(commit);
+    if option_env!("GIT_COMMIT") == None {
+        let commit = Command::new("git")
+            .arg("log")
+            .arg("-1")
+            .arg("--pretty=format:%h") // Abbreviated commit hash
+            .current_dir(env!("CARGO_MANIFEST_DIR"))
+            .output();
+    
+        if let Ok(commit) = commit {
+            if let Ok(commit) = String::from_utf8(commit.stdout) {
+                return Some(commit);
+            }
         }
+
+    } else {
+        return option_env!("GIT_COMMIT").map(String::from);
     }
+
     None
 }
 
 fn current_git_branch() -> Option<String> {
-    let commit = Command::new("git")
-        .arg("rev-parse")
-        .arg("--abbrev-ref")
-        .arg("HEAD")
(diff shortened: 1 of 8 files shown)


kantai:
> Disables triggering a release when a tag is pushed up. cc @kantai 

This works for me

kantai (approved):

CharlieC3:
@kantai Did you see this part? 

> I tested this change out locally and it doesn't seem to do what we want it to do. The server_version in the /v2/info endpoint still reads blockstack-core 0.0.1 => 23.0.0.0. @kantai Can you double check this?

I'm wondering if the version and package name cannot be overridden by ENV vars because it's defined in the Cargo.toml.

diwakergupta:
> @kantai Did you see this part?
> 
> > I tested this change out locally and it doesn't seem to do what we want it to do. The server_version in the /v2/info endpoint still reads blockstack-core 0.0.1 => 23.0.0.0. @kantai Can you double check this?
> 
> I'm wondering if the version and package name cannot be overridden by ENV vars because it's defined in the Cargo.toml.

@CharlieC3 that's because code relies on `option_env!`, which inspects an environment variable at _compile time_, not run time: https://doc.rust-lang.org/std/macro.option_env.html

The `blockstack-core` and `0.0.1` are coming from the main Cargo.toml. I'm less sure about the `23.0.0.0`, but either way, if you set these variables when building locally, it should work.

CharlieC3:
@diwakergupta [The env vars are getting set before the build occurs](https://github.com/blockstack/stacks-blockchain/pull/2276/files#diff-18f6d44d9b68714128f52a7f925d93e239b1a98e31a034f0811ae92a8191c6f3R5-R6).

I also tried setting the env vars during runtime in the docker container just to test, but it doesn't seem to have any affect (as expected from the reasons you mentioned).

build.rs
```diff
@@ -1,32 +1,43 @@
...
+            .output();
+    
+        if let Ok(commit) = commit {
+            if let Ok(commit) = String::from_utf8(commit.stdout) {
+                return Some(commit);
+            }
         }
+
+    } else {
+        return option_env!("GIT_COMMIT").map(String::from);
```
CharlieC3, line 19: Not sure if there's a better way to convert a `Option<&'static str>` to `Option<String>`.

kantai: That's a perfect approach.

PR #2278, opened 2021-01-06
feat: export block height changes for names and namespaces

zone117x:
Closes https://github.com/blockstack/stacks-blockchain/issues/2274

> Stacks 2.0 BNS import needs the exported BNS data block heights to be deltas rather than an arbitrary Stacks 1.0 block height. This is already done for the lockup schedule block heights.

* We decided to reset name registrations to block 0 to avoid the negative integer issue. The `first_registered` column is now removed. The `expired_at` column is also removed as it's not used in the import process.
* We also decided to remove the `reveal_block` and `ready_block` from the `namespaces` table as these will always be negative.
* The rust import code will need some minor tweaks to work with a new `chainstate.txt` after this PR.

Snapshot of latest `chainstate.txt`: https://mega.nz/file/MNsQSQLb#P6Jpatjun-fhREzELK3YJHsTTOYrU13gzDZAndWRwgs

Here's a sample of the new output data:
```csv
-----BEGIN NAMESPACES-----
namespace_id,address,buckets,base,coeff,nonalpha_discount,no_vowel_discount,lifetime
app,SP24TC3Y58XKHZ7GX0N69X50BFYD9ECSR8PGAE3H6,6;5;4;3;2;1;0;0;0;0;0;0;0;0;0;0,4,250,4,4,0
blockstack,SP355AMV5R2A5C7VQCF4YPPAS05W93HTB68574W7S,7;6;5;4;3;2;1;1;1;1;1;1;1;1;1;1,4,250,4,4,0
btc,SP24TC3Y58XKHZ7GX0N69X50BFYD9ECSR8PGAE3H6,3;3;3;3;3;3;3;3;3;3;3;3;3;3;3;3,10,2,1,1,262800
graphite,SP3X2W6GTGVVG9RW78F1MT7M9AQFS2HWV2JETXPG2,6;5;4;3;2;1;0;0;0;0;0;0;0;0;0;0,4,250,4,4,52595
helloworld,SP3TJ4WRTXW0YFWF1TWKWCG7E6Z6MZDTWAVB6SW5V,15;14;13;12;11;10;9;8;7;6;5;4;3;2;1;0,2,4,2,5,0
id,SP364R1Y0XYC8PYXEK0PE3V752RAJB2GNWF28WP90,6;5;4;3;2;1;0;0;0;0;0;0;0;0;0;0,4,250,4,4,52595
miner,SP24TC3Y58XKHZ7GX0N69X50BFYD9ECSR8PGAE3H6,0;0;0;0;0;0;0;0;0;0;0;0;0;0;0;0,1,0,1,1,0
mining,SP24TC3Y58XKHZ7GX0N69X50BFYD9ECSR8PGAE3H6,6;5;4;3;2;1;0;0;0;0;0;0;0;0;0;0,4,250,4,4,0
podcast,SPHYHD2ZJ11HWRDKVZ160QX7FD7PCVNT2N9CT2H6,1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1,40,250,1,1,0
stacking,SP24TC3Y58XKHZ7GX0N69X50BFYD9ECSR8PGAE3H6,6;5;4;3;2;1;0;0;0;0;0;0;0;0;0;0,4,250,4,4,0
stacks,SP24TC3Y58XKHZ7GX0N69X50BFYD9ECSR8PGAE3H6,6;5;4;3;2;1;0;0;0;0;0;0;0;0;0;0,4,250,4,4,0
stx,SP24TC3Y58XKHZ7GX0N69X50BFYD9ECSR8PGAE3H6,6;5;4;3;2;1;0;0;0;0;0;0;0;0;0;0,4,250,4,4,0
-----END NAMESPACES-----
-----BEGIN NAMES-----
name,address,zonefile_hash
jude.blockstack,SP1JHJ56H4JTP1FR3TEWC4RS8D5T2VV53N9YZ2FE3,0c79ea594ac287aafe9c2a2c677d674b06066c34
0.id,SP1HPCXTGV31W5659M3WTBEFP5AN55HV4B1Q9T31F,c5f5b4197c94d4561f6332160f624d728b0da8bf
1.id,SP1FRDVA4JAZ4TC93XZC850QAY6R4MTHN10XZJ035,8961c32ad98df87af12b186c3d0be6ca22105936
10.id,SP12TFSBQFZWVVEAWJTVRW3K8V00R0S4895ABXRBR,8ae7173b9a2f96521ba077bbcbe9b56b1cf2449f
10x.id,SP3N5TVVEHGA8B6G3JXWBP476161E32654D9P9A3Y,3d0576cd134fa1ff8d54d8e2ef5da97f5cae36f2
111111111.id,SP15XBGYRVMKF1TWPXE6A3M0T2A87VYSVF9VFSZ1A,01a8c60c852b75f9c6417491ac89275fb9de774c
123.id,SP2PEZ0W4PFGYV6ZZ2DWDVNAVVTQPJ9KX8EK4VRVT,ae34427289a8865028eb2d475dafb085060d5625
12345.id,SP2TVQY1GZ81YPNMF7GEVQHB3XYQG6Q8D3YYJ7HMA,c2fe1fc7097eea0aa83662cb54f103bdcbdd1573
180.id,SP2RWN4GG3XBKZMGEW0E51FEQHNYG77DSG96F7DHS,d0757f74e5b86a31abfc7ad37a37c6296bb8a6a7
180canada.id,SP3CMWZWFE019VZ8JYVPJMYE4P4F4A2CHZ87VNWG5,df54539e135c5ebee49cac4d96724c1951258527
1glenco.id,SP2W5S34H18ECTX1292SFMR9W5R5CM82FBCVZ8RT8,5b3e2ec62deba8eab24b64cb26f475e69a75a335
21.id,SP1FRKP5DNGYH70AYEWRHM3J3Z6S1HPR85XSNFAXG,2fb73c0af99810c0431f198eae762f8d14ed29ee
4m8b.id,SP206DSBHDEJ8B3PNAFRE0GVJ3HTD2GD6CY4PVCKQ,d4cc424442defd5c7aa9231fc69e6919a6ef0d53
```

### DIFF
--- a/blockstack/blockstackd.py
+++ b/blockstack/blockstackd.py
@@ -3733,8 +3733,8 @@ def run_blockstackd():
         for namespace_str in namespaces_entries:
             namespace_info = db.get_namespace(namespace_str)
             namespace = {}
-            namespace['ready_block'] = namespace_info['ready_block']
-            namespace['reveal_block'] = namespace_info['reveal_block']
+            namespace['ready_block'] = namespace_info['ready_block'] - block
+            namespace['reveal_block'] = namespace_info['reveal_block'] - block
             namespace['namespace_id'] = namespace_info['namespace_id']
             namespace['address'] = b58ToC32(str(namespace_info['address']))
             namespace['buckets'] = ';'.join(str(x) for x in namespace_info['buckets'])
@@ -3757,7 +3757,7 @@ def run_blockstackd():
         name_zonefiles_txt_output_path = os.path.join(output_dir, 'name_zonefiles.txt')
         name_zonefiles_txt_hash_output_path = name_zonefiles_txt_output_path + '.sha256'
         chainstate_f.write('-----BEGIN NAMES-----\n')
-        chainstate_f.write('name,address,registered_at,expire_block,zonefile_hash\n')
+        chainstate_f.write('name,address,registered_at,expire_block,renewal_deadline,zonefile_hash\n')
         name_zonefiles = open(name_zonefiles_txt_output_path, 'w')
         for name_str in name_entries:
             name_info = load_name_info(db, name_str, block)
@@ -3768,8 +3768,16 @@ def run_blockstackd():
             name = {}
             name['name'] = name_info['name']
             name['address'] = b58ToC32(str(name_info['address']))
+
             name['expire_block'] = name_info['expire_block']
-            name['registered_at'] = max(name_info['first_registered'], name_info['last_renewed'])
+            if name['expire_block'] != -1: # for special flag `-1` do not subtract the block height
+                name['expire_block'] = name['expire_block'] - block
+
+            name['renewal_deadline'] = name_info['renewal_deadline']
+            if name['renewal_deadline'] != -1: # for special flag `-1` do not subtract the block height
+                name['renewal_deadline'] = name['renewal_deadline'] - block
+
+            name['registered_at'] = max(name_info['first_registered'], name_info['last_renewed']) - block
 
             has_zonefile_hash = 'value_hash' in name_info and name_info['value_hash'] is not None
             if has_zonefile_hash:
@@ -3781,8 +3789,8 @@ def run_blockstackd():
                 # print 'missing zonefile for {}'.format(name)
                 name_zonefiles.write(name_info['value_hash'] + '\n')
                 name_zonefiles.write(name_info['zonefile'].replace('\n', '\\n') + '\n')
-            chainstate_f.write('{},{},{},{},{}\n'.format(
-                name['name'], name['address'], name['registered_at'], name['expire_block'], name['zonefile_hash']
+            chainstate_f.write('{},{},{},{},{},{}\n'.format(
+                name['name'], name['address'], name['registered_at'], name['expire_block'], name['renewal_deadline'], name['zonefile_hash']
             ))
         name_zonefiles.flush()
         name_zonefiles.close()
--- a/blockstack/blockstackd.py
+++ b/blockstack/blockstackd.py
@@ -3773,7 +3773,7 @@ def run_blockstackd():
             if name['expire_block'] != -1: # for special flag `-1` do not subtract the block height
                 name['expire_block'] = name['expire_block'] - block
 
-            name['registered_at'] = max(name_info['first_registered'], name_info['last_renewed']) - block
+            name['registered_at'] = 0
 
             has_zonefile_hash = 'value_hash' in name_info and name_info['value_hash'] is not None
             if has_zonefile_hash:
--- a/blockstack/blockstackd.py
+++ b/blockstack/blockstackd.py
@@ -3729,12 +3729,10 @@ def run_blockstackd():
         namespaces_entries = db.get_all_namespace_ids()
         namespaces_entries.sort()
         chainstate_f.write('-----BEGIN NAMESPACES-----\n')
-        chainstate_f.write('namespace_id,address,reveal_block,ready_block,buckets,base,coeff,nonalpha_discount,no_vowel_discount,lifetime\n')
+        chainstate_f.write('namespace_id,address,buckets,base,coeff,nonalpha_discount,no_vowel_discount,lifetime\n')
         for namespace_str in namespaces_entries:
             namespace_info = db.get_namespace(namespace_str)
             namespace = {}
-            namespace['ready_block'] = namespace_info['ready_block'] - block
-            namespace['reveal_block'] = namespace_info['reveal_block'] - block
             namespace['namespace_id'] = namespace_info['namespace_id']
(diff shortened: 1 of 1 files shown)


blockstack/blockstackd.py
```diff
@@ -3782,8 +3782,16 @@ class Program {
...
             name['expire_block'] = name_info['expire_block']
-            name['registered_at'] = max(name_info['first_registered'], name_info['last_renewed'])
+            if name['expire_block'] != -1: # for special flag `-1` do not subtract the block height
+                name['expire_block'] = name['expire_block'] - block
+
+            name['renewal_deadline'] = name_info['renewal_deadline']
+            if name['renewal_deadline'] != -1: # for special flag `-1` do not subtract the block height
+                name['renewal_deadline'] = name['renewal_deadline'] - block
+
+            name['registered_at'] = max(name_info['first_registered'], name_info['last_renewed']) - block
```
jcnelson, line 3790: This looks like it could be negative?  Isn't `block` the current chain height, in which case, `first_registered` and `last_renewed` may precede it?

zone117x: Yes this is intended, see PR description. It could be absolute value if we don't want to deal with signed integers, but I figured we'd keep all the "block delta" fields logically consistent. 

Any suggestions?

lgalabru: Yeah I wonder if BNS block heights should be based on the burnchain block height, instead of the stacks chain block height.
Any opinion @jcnelson @kantai ?

kantai: I think it should be stacks chain height.

jcnelson: Seconded -- Stacks chain height.

jcnelson: > Any suggestions?

A couple things:
* for names whose `renewal_deadline` becomes 0 or negative after subtracting `block`, don't include the name (since it's expired).
* for names whose `first_registered` values become negative, set them to 0.  I don't think we need to worry about what height they were registered at before the launch of Stacks 2.0 -- anyone who wants to know this information is free to spin up a Stacks 1.0 node.

lgalabru: `expire_block` is not used (and we can remove it from that dump), because this property is calculated.
If we're setting the `registered at` to 0, then we would be extending all the leases.
If we're not basing on the burnchain height, I'll update the contract and add a `if` and a `legacy_expire_block` field to track the deadline set in v1.

jcnelson: The reason I recommended setting `first_registered` to `0`0 is because `registered at` can't be negative in BNS -- both `registered-at` and `imported-at` are `uint` in `bns.clar`.

zone117x: `renewal_deadline` is now removed. Expired names are already not included. 

Do we want to extend name expirations by setting `first_registered` to `0`, or change the contract to support signed ints? Seems like those are the only two options?

jcnelson: Yup, sounds like it.  Let's ask @diwakergupta.

diwakergupta: @zone117x let's do the former, and extend the expirations.

jcnelson: Got word from @diwakergupta -- let's just renew them all.  Set `first_registered` to 0.

zone117x: Great, done!

zone117x: Thoughts on if we should also drop `reveal_block` and `ready_block` from the `namespaces` table? These will also always be negative and I think require a contract change to handle. @jcnelson @lgalabru

zone117x: Removed the `reveal_block` and `ready_block` from the `namespaces` table as discussed on Discord.

blockstack/blockstackd.py
```diff
@@ -3782,8 +3780,12 @@ class Program {
             name = {}
             name['name'] = name_info['name']
             name['address'] = b58ToC32(str(name_info['address']))
+
             name['expire_block'] = name_info['expire_block']
-            name['registered_at'] = max(name_info['first_registered'], name_info['last_renewed'])
+            if name['expire_block'] != -1: # for special flag `-1` do not subtract the block height
```
jcnelson, line 3785: I'm not sure the Stacks chainstate boot code even uses `expire_block`.  Should we just remove it?  cc @lgalabru

lgalabru: correct

zone117x: `expire_block` column removed! The PR description includes an updated sample of what the export data now looks like.

lgalabru (approved):
Looks good to me, thank you @zone117x!

jcnelson (approved):
LGTM!

PR #2279, opened 2021-01-06
Migration: change export block duration from 500 to 300

zone117x:
At some point we decided to change the migration export waiting period from 500 blocks to 300 blocks, but never made the change. Here it is.

Ref: https://github.com/blockstackpbc/devops/issues/549#issuecomment-755407063

### DIFF
--- a/blockstack/lib/nameset/virtualchain_hooks.py
+++ b/blockstack/lib/nameset/virtualchain_hooks.py
@@ -580,7 +580,7 @@ def db_save( block_height, consensus_hash, ops_hash, accepted_ops, virtualchain_
         # threshold of required ID registrations
         v2_upgrade_signal_required_registrations = 20
         # number of blocks to wait after signal threshold is reached until block import and chainstate dump 
-        v2_upgrade_signal_import_threshold = 500
+        v2_upgrade_signal_import_threshold = 300
         if BLOCKSTACK_TEST:
             v2_upgrade_signal_import_threshold = 10
 


diwakergupta:
We documented it internally [here](https://paper.dropbox.com/doc/Stacks-v1-v2-Upgrade-Implementation--BCuqr0I9qAsWY4s34cS7xuv2Ag-bwo5CumbrcSGFvm3NBuzc#:uid=355636152862737218776293&h2=Forum-Post-Draft) and then on [the forum post](https://forum.stacks.org/t/rfc-stacks-1-0-2-0-upgrade-process/11346)

I'm actually wondering if we should reduce this even further. With 300 + 300 blocks, we have ~4.5 days between miner threshold and 2.0 genesis block, and ~2.05 days of total downtime. For Jan 14th launch, that means we should aim to hit the threshold on Jan 10th, which is a Sunday. Might be better to have that fall on Monday, I think 3-3.5 days total might be sufficient?

How do we feel about 250 blocks or 200 blocks? cc/ @kantai @jcnelson

jcnelson:
I'm personally in favor of sticking with what we've announced, lest we ruin other peoples' upgrade plans.  Also, given how frantic things have been on the Xenon testnet, I'd be in favor of having more time.

kantai (approved):

zone117x:
@diwakergupta looks like we should stick with 300?

diwakergupta:
Yep, ship it!

PR #2285, opened 2021-01-07
Pessimistic block limits

kantai:
This sets initial pessimistic block limits for mainnet. These are set based on evaluating:

* `stx-transfer?`
* `stx-stack`
* Simple large contract analysis
* `secp256k1-verify`

Far and way, of those, `stx-stack` is the most expensive, and governs the limits imposed on the `write-count` and `read-count` for the MARF. With these block settings, at most 50 such operations could occur in a block. On simple tests, even 50 such operations take around a minute to process (the test MARF is seeded with 500,000 account entries).

The benchmarking tests can be found in the `feat/block-limits` branch, which is separated here, because there's code in that branch to support benchmarking and profiling which should not necessarily be merged yet.

### DIFF
--- a/benches/block_limits.rs
+++ b/benches/block_limits.rs
@@ -1,149 +1,530 @@
 extern crate blockstack_lib;
+extern crate rand;
 extern crate serde_json;
 
 use blockstack_lib::{
-    chainstate::burn::BlockHeaderHash,
-    chainstate::stacks::index::storage::TrieFileStorage,
+    burnchains::BurnchainHeaderHash,
+    chainstate::{
+        self,
+        burn::BlockHeaderHash,
+        stacks::boot::STACKS_BOOT_POX_CONTRACT,
+        stacks::{index::MarfTrieId, StacksBlockId},
+    },
     vm::clarity::ClarityInstance,
     vm::costs::ExecutionCost,
-    vm::database::{MarfedKV, NULL_HEADER_DB},
-    vm::types::QualifiedContractIdentifier,
+    vm::database::MarfedKV,
+    vm::{
+        database::{HeadersDB, NULL_BURN_STATE_DB},
+        types::{PrincipalData, QualifiedContractIdentifier, StandardPrincipalData},
+        Value,
+    },
 };
+use chainstate::{burn::VRFSeed, stacks::StacksAddress};
 
-use std::env;
-use std::fmt::Write;
+use std::fs;
 use std::process;
+use std::{env, time::Instant};
 
-fn test_via_tx(scaling: u32, inner_loop: &str, other_decl: &str) -> ExecutionCost {
-    let marf = MarfedKV::temporary();
+use rand::Rng;
+
+struct TestHeadersDB;
+
+impl HeadersDB for TestHeadersDB {
+    fn get_stacks_block_header_hash_for_block(
+        &self,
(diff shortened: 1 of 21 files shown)


jcnelson (approved):
This LGTM.  Do you want to wait until `feat/block-limits` gets merged, though, so we can point to the scripts that justify the cost parameters?

kantai:
> Do you want to wait until feat/block-limits gets merged, though, so we can point to the scripts that justify the cost parameters?

Not sure -- but I can clean up `feat/block-limits` a bit to pull the benchmark changes into this branch pretty easily I think.

kantai:
This PR also just added a "cheap" performance optimization -- it turns out, that the large difference between say, the execution cost of `map-set?` `map-get`, etc., and `stx-transfer?` is that the VM type checks the former but not the latter. In order to do so, it needs to perform metadata reads (which are 2 MARF reads), and that the VM was previously unneccessarily performing two such loads (once to charge a cost, once to type check). Ultimately, both can be eliminated by storing this metadata in the contract metadata (which would already be loaded), but the current patch only eliminates the completely spurious load.

kantai:
https://github.com/blockstack/stacks-blockchain/pull/2285/commits/ff9e1c451b9ffa55a1990c1dfa53bdb0f5d847b1 includes the benchmarking suite used to produce limits.

Specifically, to understand marf read/write bounds, execute the stack-stx test on 50 stack-stx ops with a 500k simulated genesis and 100 block buildup:

```
$ cargo bench --bench block_limits -- stack-stx 100 500000 50
```

To understand the runtime bounds, execute the smart-contract test with a scaling factor of 10,000:

```
$ cargo bench --bench block_limits -- smart-contract 100 5000 10000
Finished buildup in 398ms
Completed smart-contract scaled at 10000 in 7547 ms, after 100 block buildup with a 5000 account genesis
{"write_length":518892,"write_count":2,"read_length":1,"read_count":1,"runtime":4947794000}
```

Write length and read length roughly correspond to bytes read and written, and aren't set by benchmarking, rather, those can be set by what is deemed as acceptable I/O per block.


I'm going to try to implement the second optimization of `map-get`, etc. The first optimization reduced runtime of stack-stx from `1.7s` to `1.0s`, and I'd expect the second optimization to do roughly the same. This would make the cost of MARF lookups _actually_ 1 MARF lookup (as opposed to now, where there are corresponding metadata fetches for each). This optimization, by itself, isn't consensus breaking, but it will probably allow us to increase the MARF read/write limit, which would be consensus breaking.

lgalabru (approved):

kantai:
> I'm going to try to implement the second optimization of map-get, etc. The first optimization reduced runtime of stack-stx from 1.7s to 1.0s, and I'd expect the second optimization to do roughly the same. This would make the cost of MARF lookups actually 1 MARF lookup (as opposed to now, where there are corresponding metadata fetches for each). This optimization, by itself, isn't consensus breaking, but it will probably allow us to increase the MARF read/write limit, which would be consensus breaking.

Implemented in https://github.com/blockstack/stacks-blockchain/pull/2285/commits/f76133decb76a5d8f3a66d763e4012227bc20d14 and it indeed reaped some savings -- now these operations take ~240ms on the same test. So, ultimately, we went from 1700ms->240ms.

PR #2277, opened 2021-01-06
feat: patch `btc` namespace price parameters

zone117x:
Closes https://github.com/blockstackpbc/devops/issues/540

This PR adds a patch to the `btc` namespace price parameters that get exported for Stacks 2.0. The goal is for names on the `btc` namespace to cost 2 STX per name.

@lgalabru can you verify the price parameters here are patched as intended? I ran into issues trying to test locally within the bns-test scripts.

PR #2286, opened 2021-01-07
BNS: fix & feats

lgalabru:
After having a closer look at https://github.com/blockstack/stacks-blockchain/pull/2277, it appeared to me that:
1) the namespace's price table is wrong
2) the import process is going to import some namespace v1 (BTC based) and silently/implicitly be converting them into namespace v2 (STX based). So #2277 should should probably apply a BTC / STX conversion - I think that could be achieved by tweaking the coefficient in the price function. 

The coefficient update would be arbitrarily fixed by us, set in the stone, and all this without consulting the namespace owners. I'm not super comfortable with this idea and would love to avoid a hard fork because of that :).
This PR is implementing the ability to update the price function + the possibility to forever renounce to this ability.


zone117x:
Note that the Stacks 1.0 image is blocked on this PR, as the `btc` namespace patch PR is blocked on this one.

src/chainstate/stacks/boot/bns.clar
```diff
@@ -497,7 +499,81 @@
...
+          p-func-b16))
+        (base p-func-base)
+        (coeff p-func-coeff)
+        (nonalpha-discount p-func-non-alpha-discount)
+        (no-vowel-discount p-func-no-vowel-discount))))
+    ;; The sender principal must match the namespace's import principal
+    (asserts!
+      (is-eq (get namespace-import namespace-props) tx-sender)
+      (err ERR_NAMESPACE_OPERATION_UNAUTHORIZED))
+    ;; The sender principal must match the namespace's import principal
```
jcnelson, line 554: Wrong comment text?

bns-test/test/namespace_update_price_function.test.ts
```diff
@@ -0,0 +1,181 @@
...
+
+      let priceFunctionUpdate2 = {
+        buckets: [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
+        base: 1,
+        coeff: 8, // <- update
+        noVowelDiscount: 0,
+        nonAlphaDiscount: 0,
+      };
+
+      // Updating the price function should succeed
```
jcnelson, line 166: Shouldn't this be `should fail`?

jcnelson (approved):
LGTM!  Please fix the comments before merging.

src/chainstate/stacks/boot/bns.clar
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2286/commits/d9831b806de4414eb16a99c1af714193f6ea9704

bns-test/test/namespace_update_price_function.test.ts
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2286/commits/d9831b806de4414eb16a99c1af714193f6ea9704

kantai (approved):

Issue #2281, opened 2021-01-06
P2P: cap the number of neighbors we handshake with when walking the neighbors-of-neighbor


jcnelson:
Right now, the p2p network naively tries to handshake with all neighbors of its current neighbor.  This leads to heavy and unnecessary I/O usage in practice, slowing the whole node (and host) down.

Issue #2280, opened 2021-01-06
P2P: cap number of peers we synchronize in the inventory state machine to `num_neighbors`


jcnelson:
Otherwise, the node goes crazy trying to synchronize inventories with _everyone_, which slows it down considerably.

Issue #2022, opened 2020-10-31
P2P: per-reward-cycle inventory refresh intervals


jcnelson:
Right now, the block inventory state-machine naively re-downloads all of the block inventories for each reward cycle for each peer every burnchain block.  In testing this with lots of public nodes, this can lead to lots of unnecessary CPU usage and network transmission -- inventories for older reward cycles are unlikely to change, and more often than not, the act of re-downloading and re-processing them does not cause the node to learn any new information.

What we could do instead is make it so that a node will only download the _N_ th prior block inventory every _N_ passes of the state machine.  So, if there are 100 reward cycles, the inventory for reward cycles 100 and 99 would be fetched on every pass; the inventory for the 98th reward cycle would be fetched every 2nd pass; the inventory for the 97th reward cycle would be fetched every 3rd pass, etc.  The node would unconditionally re-fetch all block inventory data from a peer if it is sufficiently stale (this expiry time could be configurable), but it would spend most of its time refreshing its view of the last two reward cycle's block inventories, which are by far the most likely block inventories to change in-between refreshes.

The refresh logic could be probabilistic.  When the inventory state-machine for a neighbor starts up, it could randomly sample the starting reward cycle to query based on the above (e.g. pick a reward cycle with probability inversely proportional to how old it is), and then proceed to query all subsequent reward cycles' inventories as it does now.

PR #2287, opened 2021-01-07
Fix performance regressions in the P2P network

### LINKED ISSUE #2272: P2P: node not pruning enough neighbors
On Xenon, I observed 182 outbound neighbors:

```bash
➜ curl -s http://127.0.0.1:20443/v2/neighbors | jq '.outbound[].public_key_hash' | sort | uniq | wc -l
     182
```

Per @jcnelson, [this seems too high](https://discord.com/channels/621759717756370964/730814571517968436/796107541703688222):

```
The node should have been pruning its inbound and outbound neighbor sets
Will look more into that.  182 is great to see, but it's really high
```

jcnelson:
This fixes #2272, #2281, #2280, #2022, and #2320.  It makes it so we prune connections _quickly_, we set sensible neighbor counts, we don't spend as much time fetching redundant block inventory data, and we limit the number of writes we do on the frontier DB per P2P state machine pass.  I have confirmed multiple times that I can boot up a Xenon node with this PR.

The PR sets and enforces a maximum number of outbound connections to 16 peers, but it permits up to 750 inbound p2p connections (and decouples this count from the number of inbound HTTP requests).  To ensure that a sparse peer set doesn't allow the node to get cut off from nodes that are likely to have anchor blocks (i.e. miners), the PR makes it so any bootstrap nodes the node has configured will _always_ be neighbors.  Moreover, the PoX sync watchdog will not permit the node to fetch the next reward cycle until it has confirmed that at least one bootstrap node has been scanned.  In addition, this PR makes it so that the neighbor-walk logic always tries to keep a bootstrap node as a neighbor, and makes it so the inventory synchronization logic always assumes that a bootstrap node that NACKs a request for inventory data due to an invalid consensus hash is correct  in doing so -- i.e. if this happens, then the local node, not the remote node, is assumed to have offered a bad consensus hash in the request (instead of the remote node having incomplete chainstate).

This PR effectively makes it so that a node's bootstrap nodes are always trusted to have exactly the right anchor blocks.  When we deploy this PR, we should be sure to have enough seed nodes that each miner gets reliably connected as an inbound peer to at least one seed node (so the network will stay in-sync and nodes can bootstrap without deep PoX reorgs).  Once we have #1805 ready, then this trust assumption will no longer be necessary -- #1805 will make it so nodes will be able to use the burnchain state to infer which neighbors' inventory NACKs are appropriate.

Please test with a Xenon node and verify that you can reach the chain tip without deep PoX reorgs (i.e. no more than 1 or 2 reward cycles).  In debug mode, should see the `PoxId(..)` occurrences in the log file grow with `true`'s at a linear pace.  At the time of this writing, there should be no `false`s.


jcnelson:
This is now ready for review.

testnet/stacks-node/src/syncctl.rs
```diff
@@ -395,7 +395,7 @@ impl PoxSyncWatchdog {
             // burnchain has advanced since the last call
             self.last_burnchain_height = burnchain_height;
             debug!("PoX watchdog: Wait for at least one inventory state-machine pass...");
-            self.relayer_comms.wait_for_inv_sync_pass(self.max_samples);
+            self.relayer_comms.wait_for_inv_sync_pass(3600); // wait an hour at most
```
kantai, line 398: Isn't waiting an hour way too long? Or is the goal here that if a node cannot complete a full inventory sync pass, it shouldn't process burnchain blocks at all?

kantai (approved):
This LGTM!

kantai:
These changes appear to have broken the `neon_integrations` tests -- it may be related to the `pox_sync_watchdog` changes. Can you fix this before merging?

jcnelson:
Yes, I'll fix them before merging.

testnet/stacks-node/src/syncctl.rs
jcnelson: The latter.

lgalabru: Should this value be coming from config?

src/net/p2p.rs
```diff
@@ -3004,14 +3030,16 @@ impl PeerNetwork {
...
+                    do_prune = true;
+                    /*
+                    do_prune = match self.walk {
+                        Some(ref mut walk) => {
+                            walk.do_prune = false;
+                            true
+                        }
+                        None => true
+                    };
+                    */
```
lgalabru, line 3042: Should we add a TODO or remove that snippet?

lgalabru (approved):

src/net/p2p.rs
jcnelson: The snippet should be removed; will do.

jcnelson:
Just waiting for the `nettest` to pass.  I'm also running it locally because I want to be reasonably sure this doesn't break anything.

PR #2288, opened 2021-01-07
Add on-chain name zonefiles

zone117x:
Embed the on-chain name zonefile content. This data is not consensus critical, but is needed to seed the atlas db.

As of this snapshot, there are ~1600 names. The compressed archive is ~100KB, so minimal impact to binary size.

@lgalabru do you know where in the init area is appropriate for loading this content into atlas?


lgalabru:
I can take it from there, thank you so @zone117x!

zone117x:
Thanks @lgalabru! Also beware the `name_zonefiles.txt` committed in this PR is from a Stacks 1.0 synced today, while the `chainstate.txt` is still months old. So it's very likely there will not be a 1-1 match in the zonefile hashes between the two. 

I'm waiting on a couple other PRs before committing a new `chainstate.txt` in order to avoid polluting git history with too many ~20MB files.

jcnelson (commented):
Is there a way to verify that each name zone file matches its corresponding name, or do you consider that out of scope since this will be (later) verified by Atlas? cc @lgalabru

src/net/atlas/db.rs
```diff
@@ -112,6 +114,21 @@ impl AtlasDB {
         )
         .map_err(db_error::SqliteError)?;
 
+        if let Some(attachments) = genesis_attachments {
+            let now = util::get_epoch_time_secs() as i64;
```
zone117x, line 118: Where is `created_at` used? If I decide to boot a Stacks 2.0 node a couple months after mainnet -- won't all my names be 2 months newer than the first node? Maybe this value should be based on the burnchain block time?

stx-genesis/src/lib.rs
```diff
@@ -81,6 +86,44 @@ impl GenesisData {
...
+            include_bytes!(concat!(env!("OUT_DIR"), "/name_zonefiles.gz"))
+        })
+    }
+}
+
+struct LinePairReader {
+    val: Lines<BufReader<Decoder<Cursor<&'static [u8]>>>>,
+}
+
+impl Iterator for LinePairReader {
```
zone117x, line 102: (For my own sanity) anyone know why `chunks` is implemented for slices and vecs, but not for iterators? This was implemented because `my_iter.chunks(2)` wasn't possible..

src/net/atlas/db.rs
lgalabru: It doesn't matter, the `created_at` is just used for pruning uninstantiated attachments.

lgalabru (commented):
Looks good to me :trollface:

stx-genesis/src/lib.rs
```diff
@@ -175,4 +218,14 @@ mod tests {
             assert!(name.owner.len() > 0);
         }
     }
+
+    #[test]
+    fn test_name_zonefiles_read() {
+        for zonefile in GenesisData::new(false).read_name_zonefiles() {
+            assert_eq!(zonefile.zonefile_hash.len(), 40);
```
jcnelson, line 225: Do you want to check at least that the zone file contents are correct?

testnet/stacks-node/Cargo.toml
```diff
@@ -23,6 +23,7 @@ libc = "0.2"
 slog = { version = "2.5.2", features = [ "max_level_trace" ] }
 
 [dev-dependencies]
+ring = "0.16.19"
```
jcnelson, line 26: Are we using this anywhere?

jcnelson (approved):
This LGTM, but please add a test to verify that we can load an on-chain name zone file's contents (don't just look at the length).

kantai (approved):
This looks good to me, though I agree if we don't need the `ring` dependency, we should drop it.

stx-genesis/src/lib.rs
kantai: I think the build itself fails if the zone file contents don't correspond to the sha sum.

testnet/stacks-node/Cargo.toml
lgalabru: ha the ring thing is already pulled by another crate, I'm just specifying a version to get this build working on Apple Silicon :)

stx-genesis/src/lib.rs
lgalabru: Yeah these tests are just checks against the sampled genesis.

PR #2297, opened 2021-01-08
Update Rusqlite (0.16 -> 0.24)

lgalabru:
This PR is addressing https://github.com/blockstack/stacks-blockchain/issues/1274.
We're looking at 3 different refactoring:
- `row::get_unwrap` is the new `row::get`
- `query_row` now returning a `Result<T, rusqlite::Error>`
- `rows::next` now returning a `Result<Option<T>>` (previously returning `Option<Result<T>>`)


src/chainstate/stacks/db/blocks.rs
```diff
@@ -914,16 +914,10 @@ impl StacksChainState {
...
-                Ok(row) => {
-                    let next_blob: Vec<u8> = row.get(0);
-                    blobs.push(next_blob);
-                }
-                Err(e) => {
-                    return Err(Error::DBError(db_error::SqliteError(e)));
-                }
-            };
+
+        while let Some(row) = rows.next().map_err(|e| db_error::SqliteError(e))? {
```
zone117x, line 918: Error handling in a rust destructuring loop 🥇

src/clarity.rs
```diff
@@ -203,21 +195,16 @@ fn get_cli_block_height(conn: &Connection, block_id: &StacksBlockId) -> Option<u
...
-                );
-                row_opt = Some(rowid);
-                break;
-            }
-            Err(e) => {
-                panic!("FATAL: could not read block hash: {:?}", e);
-            }
-        }
+
+    while let Some(row) = rows.next().expect("FATAL: could not read block hash") {
```
zone117x, line 199: The sqlite error message is no longer included, were they helpful?
`panic!("FATAL: could not read block hash: {:?}", e);`

zone117x (approved):
LGTM!

src/chainstate/stacks/db/blocks.rs
lgalabru: yeah chaining the `map_err` is not ideal. I've been trying a few loop patterns, this one was the most readable one I could come up with - open to something else.

jcnelson:
We'll want all the unit and integration tests to pass at least locally before merging, but otherwise this looks good to me.

src/chainstate/stacks/db/blocks.rs
zone117x: I think for anyone familiar with the destructuring and `?` syntax it is readable enough

lgalabru:
> We'll want all the unit and integration tests to pass at least locally before merging, but otherwise this looks good to me.

Tests are passing after a re-run!

jcnelson (approved):

PR #2301, opened 2021-01-08
Update default read-only limits for network calls

kantai:
The latest Xenon reset (2.0.0-rc1) increased the initial costs of Clarity functions (and increased the block limit), but did not increase the default runtime limit for read-only calls over the API interface. This PR does that.


zone117x (approved):
Fixes defaults for https://github.com/blockstack/stacks-blockchain/issues/2300

lgalabru (approved):

PR #2302, opened 2021-01-09
Fix version info in distributables

CharlieC3:
Fixes the version info in distributables:

```
./stacks-node version
stacks-node 92d9860 (binary-version-fix:92d9860+, release build, macos [x86_64])
```

Closes https://github.com/blockstack/stacks-blockchain/issues/2299

## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Other

No

No

## Testing information
[All distros can be downloaded and tested here](https://github.com/blockstack/stacks-blockchain/actions/runs/473111235)

### DIFF
--- a/.github/workflows/stacks-blockchain.yml
+++ b/.github/workflows/stacks-blockchain.yml
@@ -70,12 +70,24 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


kantai (approved):
Thanks for catching this, @CharlieC3

PR #2305, opened 2021-01-09
fix: Stacks 1.0 transaction halting condition

zone117x:
Reject transactions after the export block height instead of the name threshold block.


blockstack/lib/nameset/virtualchain_hooks.py
```diff
@@ -207,7 +207,7 @@ def db_parse( block_id, txid, vtxindex, op, data, senders, inputs, outputs, fee,
         
    # check if the v2 upgrade threshold block has already been reached.
```
diwakergupta, line 208: Can you update this comment and the log message below as well to avoid confusion? Otherwise LGTM.

diwakergupta:
LGTM

PR #2306, opened 2021-01-09
Feature: allow "missed" prepare commits

kantai:
This is a consensus breaking change that allows "missed" prepare commits to be parsed, so that they can later be counted by the sortition db as "missed block commits". This will not make those commitments valid, but rather allows them to be included in a miner's chain of commitments.

### DIFF
--- a/src/chainstate/burn/operations/leader_block_commit.rs
+++ b/src/chainstate/burn/operations/leader_block_commit.rs
@@ -297,8 +297,11 @@ impl LeaderBlockCommitOp {
             return Err(op_error::ParseError);
         }
 
+        let intended_burn_height =
+            LeaderBlockCommitOp::calculate_intended_height(data.burn_parent_modulus, block_height)?;
+
         // check if we've reached PoX disable
-        let (commit_outs, sunset_burn, burn_fee) = if block_height
+        let (commit_outs, sunset_burn, burn_fee) = if intended_burn_height
             >= burnchain.pox_constants.sunset_end
         {
             // should be only one burn output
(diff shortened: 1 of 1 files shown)


src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -645,6 +648,31 @@ impl LeaderBlockCommitOp {
         self.check_single_burn_output()
     }
 
+    fn calculate_intended_height(
+        parent_modulus: u8,
+        actual_block_height: u64,
+    ) -> Result<u64, op_error> {
+        let intended_modulus = (parent_modulus as u64 + 1) % BURN_BLOCK_MINED_AT_MODULUS;
+        let actual_modulus = actual_block_height % BURN_BLOCK_MINED_AT_MODULUS;
+        if actual_modulus != intended_modulus {
```
jcnelson, line 657: What happens if the block-commit is late by exactly `BURN_BLOCK_MINED_AT_MODULUS - 1` blocks, and `actual_block_height` is evenly divisible by `BURN_BLOCK_MINED_AT_MODULUS`?  This should never happen because the window size is smaller than `BURN_BLOCK_MINED_AT_MODULUS`, right?

kantai: In this case, this check would fail to detect that this was a "missed" block commit, and instead it would try to process it as if it were a block commit in the current sortition (which would fail the block commit check if the commit outputs were invalid, and otherwise may be selected but could not produce a valid block).

This is possible, _but_ it's okay to "penalize" the miner in this case, because the window size has already passed -- trying to link commits that old won't improve the miner's score in the window.

kantai:
I'm closing this -- on mainnet, a block that this would affect only occurs once every reward cycle (2000 blocks) and so the impact of this change is probably not worth the risk of introducing a new consensus-breaking change at this point. If this turns out to be a problem, the network can reassess in a later fork.

PR #2308, opened 2021-01-10
Ability to reuse SPV headers from previous run

lgalabru:
Trying to optimize the feedback look when working with Mainnet / Testnet.
With this patch, when setting a `working_dir` in the `[node]` node of a config, the stacks node will archive the previous chainstate with a timestamp, and just restore the `spv-headers.dat` file.
At some point, this piece of code will go away and the node will just be able to boot from its working_dir, but this patch sounds like a quick win today.

If we are fine with this approach, I'll clean that patch.

### DIFF
--- a/testnet/stacks-node/src/config.rs
+++ b/testnet/stacks-node/src/config.rs
@@ -1,5 +1,6 @@
 use std::convert::TryInto;
 use std::fs;
+use std::path::Path;
 use std::net::{SocketAddr, ToSocketAddrs};
 
 use rand::RngCore;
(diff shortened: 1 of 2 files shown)


testnet/stacks-node/src/config.rs
```diff
@@ -833,6 +835,43 @@ impl Config {
...
+            .collect::<Vec<String>>();
+
+        let archive_name = format!("{}-archive", get_epoch_time_secs());
+        let archive_path = format!("{}/{}", &self.node.working_dir, archive_name);
+        fs::create_dir(&archive_path).unwrap();
+
+        for dir in to_archive.iter() {
+            let src = format!("{}/{}", &self.node.working_dir, dir);
+            let dst = format!("{}/{}", archive_path, dir);
+            fs::rename(src, dst).unwrap();
```
zone117x, line 854: Is there any sanitation/checking done to the configured `working_dir` before this? If I set `working_dir=/home` seems like this would move all my `/home` files?

jcnelson (changes requested):
I don't think this patch belongs in the blockchain node.  I think this is something you'd put into a shell script that you'd use to launch the node.

lgalabru:
Lack of interest, closing now.

PR #2310, opened 2021-01-11
docs: add bns contract references

### LINKED ISSUE #2309: Add BNS contract reference generation
Just like for `pox.clar`, we should generate `bns.clar` references

agraebe:
fixes #2309


src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,68 @@ is expired, or if there's never been such a stacker, then returns none."),
...
-            descriptions: HashMap::from_iter(pox_descriptions.into_iter()),
-            skip_func_display: HashSet::from_iter(pox_skip_display.into_iter()),
-        },
-    )])
+    let bns_skip_display = vec![
+        "namespace-update-function-price",
+        "namespace-revoke-function-price-edition",
+        "check-name-ops-preconditions",
+        "is-name-in-grace-period",
+        "can-receive-name",
```
friedger, line 131: This function should be documented because it helps  to determine whether the current user has (already) registered a name

friedger (commented):
Can receive name should be documented

agraebe:
> Can receive name should be documented

sure, can do easily. I excluded it because `can-name-be-registered` uses that method. this seems like it should always be used for the check. @lgalabru in what case would the "can receive name" method be used individually? should we add or not?

friedger:
I have proposed to add get-name, then we don't need can-receive-name.

I want to use the username as the controlling property in my contract, not a prinicpal

agraebe:
Got it, I'll simply add it

src/vm/docs/contracts.rs
agraebe: added now

kantai (approved):

agraebe:
added an item for #2323 - this is simply not used until the method is available. it doesn't break but also doesn't export any new methods in the reference (so should be ok to merge before the referenced PR)

agraebe:
@lgalabru @kantai would be great if we could merge this in 🙏🏼

agraebe:
following up, this includes both `get-name` and `can-receive-name`. The tests fail, another method needs to be added (`resolve-principal`). I'll do that and ask for another review in a bit

agraebe:
@kantai this is now updated and ready for review

kantai (approved):

src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,70 @@ is expired, or if there's never been such a stacker, then returns none."),
         ("get-pox-info", "Returns information about PoX status.")
     ];
 
+    let bns_descriptions = vec![
+        ("namespace-preorder", "Registers the salted hash of the namespace with BNS nodes, and burns the requisite amount of cryptocurrency. Additionally, this step proves to the BNS nodes that user has honored the BNS consensus rules by including a recent consensus hash in the transaction. Returns pre-order's expiration date (in blocks)."),
+        ("namespace-reveal", "Reveals the salt and the namespace ID (pafter a namespace preorder). It reveals how long names last in this namespace before they expire or must be renewed, and it sets a price function for the namespace that determines how cheap or expensive names its will be."),
```
friedger, line 90: ~~pafter~~ after

src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,70 @@ is expired, or if there's never been such a stacker, then returns none."),
...
+        ("name-import", "Import name to a revealed namespace. Each imported name is given both an owner and some off-chain state."),
```
friedger, line 91: ~~Import~~ Imports

src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,70 @@ is expired, or if there's never been such a stacker, then returns none."),
...
+        ("namespace-ready", "Launches the namespace and makes it available to the public. Once a namespace is launched, anyone can register a name in it if they pay the appropriate amount of cryptocurrency."),
+        ("name-preorder", "Preorders a name by telling all BNS nodes the salted hash of the BNS name. It pays the registration fee to the namespace owner's designated address."),
+        ("name-register", "Reveals the salt and the name to all BNS nodes, and assigns the name an initial public key hash and zone file hash."),
+        ("name-update", "Changes the name's zone file hash. You would send one of these transactions if you wanted to change the name's zone file contents. For example, you would do this if you want to deploy your own Gaia hub and want other people to read from it."),
```
friedger, line 95: One of these? Which ones?

src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,70 @@ is expired, or if there's never been such a stacker, then returns none."),
...
+        ("name-transfer", "Changes the name's public key hash. You would send one of these transactions if you wanted to:
+* Change your private key
+* Send the name to someone else
```
friedger, line 98: or update your zone file.

src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,70 @@ is expired, or if there's never been such a stacker, then returns none."),
...
+
+When transferring a name, you have the option to also clear the name's zone file hash (i.e. set it to null). This is useful for when you send the name to someone else, so the recipient's name does not resolve to your zone file."),
+        ("name-revoke", "Makes a name unresolvable. The BNS consensus rules stipulate that once a name is revoked, no one can change its public key hash or its zone file hash.  The name's zone file hash is set to null to prevent it from resolving. You should only do this if your private key is compromised, or if you want to render your name unusable for whatever reason."),
```
friedger, line 101: Why revoking if the private key is compromised you want to transfer your name? Not clear to me.

src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,70 @@ is expired, or if there's never been such a stacker, then returns none."),
...
+        ("name-renewal", "Depending in the namespace rules, a name can expire. For example, names in the .id namespace expire after 2 years. You need to send a name reveal every so often to keep your name.
```
friedger, line 102: ~~name reveal~~ name renewal

src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,70 @@ is expired, or if there's never been such a stacker, then returns none."),
...
+
+You will pay the registration cost of your name to the namespace's designated burn address when you renew it.
```
friedger, line 104: It could be a normal address, couldn't it?

src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,70 @@ is expired, or if there's never been such a stacker, then returns none."),
...
+When a name expires, it enters a month-long \"grace period\" (5000 blocks).
```
friedger, line 105: depending on the namespace config, this can be shorter or longer I think

src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,70 @@ is expired, or if there's never been such a stacker, then returns none."),
...
+        ("name-revoke", "Makes a name unresolvable. The BNS consensus rules stipulate that once a name is revoked, no one can change its public key hash or its zone file hash.  The name's zone file hash is set to null to prevent it from resolving. You should only do this if your private key is compromised, or if you want to render your name unusable for whatever reason."),
+        ("name-renewal", "Depending in the namespace rules, a name can expire. For example, names in the .id namespace expire after 2 years. You need to send a name reveal every so often to keep your name.
+
+You will pay the registration cost of your name to the namespace's designated burn address when you renew it.
+When a name expires, it enters a month-long \"grace period\" (5000 blocks). 
+
+It will stop resolving in the grace period, and all of the above operations will cease to be honored by the BNS consensus rules.
+You may, however, send a NAME_RENEWAL during this grace period to preserve your name.
```
friedger, line 108: After the grace period, everybody can register that name again.

src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,70 @@ is expired, or if there's never been such a stacker, then returns none."),
...
+If your name is in a namespace where names do not expire, then you never need to use this transaction."),
+        ("get-namespace-price", "Get the price for a namespace."),
```
friedger, line 110: ~~Get~~ Gets

src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,70 @@ is expired, or if there's never been such a stacker, then returns none."),
...
+        ("get-name-price", "Get the price for a name."),
+        ("can-namespace-be-registered", "Returns true if the provided namespace is available."),
+        ("is-name-lease-expired", "Return true if the provided name lease is expired."),
+        ("can-name-be-registered", "Returns true if the provided name can be registered."),
```
friedger, line 114: What are the conditions? Otherwise the comment is not very helpful.

src/vm/docs/contracts.rs
```diff
@@ -85,20 +85,70 @@ is expired, or if there's never been such a stacker, then returns none."),
...
+        ("name-resolve", "Get name registration details."),
+        ("get-namespace-properties", "Get namespace properties."),
+        ("can-receive-name", "Returns true if the provided name can be received. That is, if it is not curretly owned, a previous lease is expired, and the name wasn't revoked."),
+        ("get-name", "Returns a response with the username that belongs to the user if any, otherwise an error ERROR_NOT_FOUND is returned."),
+        ("resolve-principal", "Returns the name for a principal, if there is one.")
```
friedger, line 119: Returns the registered name that a principal owns if there is one. A principal can only own one name at a time.

friedger (changes requested):
I added some comments. Not all is clear to me and I have to check the contract again..

src/vm/docs/contracts.rs
agraebe: `name-update` txs. updating

src/vm/docs/contracts.rs
agraebe: I copied this from the contract - not sure I have the right answer. @kantai might be able to help

src/vm/docs/contracts.rs
agraebe: added that it is configurable!

src/vm/docs/contracts.rs
agraebe: @kantai to confirm

src/vm/docs/contracts.rs
agraebe: which command explicitly? If you know the conditions, please let me know. I'm not an expert on the contract - I really just want to help with _some_ documentation

src/vm/docs/contracts.rs
agraebe: 👍🏼

agraebe:
> I added some comments. Not all is clear to me and I have to check the contract again..

Thanks @friedger!
 
IMHO, we don't need the _perfect_ documentation to begin with. something to start with would be great. do you need to check the contract for this to go live? could you prioritize this. every automated docs update PR from this repo is creating an issue and can't be merged until we have this PR in master

friedger:
Lgtm

friedger (approved):
Good for now, should be reviewed again.

agraebe:
I need @lgalabru or @jcnelson to approve as well I notice :)

agraebe:
ping - would love to get this done and merge into the docs. thanks!

kantai:
Pinging for a review from @jcnelson or @diwakergupta -- this PR's changes are already reflected in the docs repo, and without this, our documentation in this repo and the docs repo are out-of-sync, causing every push on `master` in this repo to open a new docs PR to re-sync.

diwakergupta (approved):
Ship it!

src/vm/docs/contracts.rs
jcnelson: @friedger It's just one example.  The attacker could just as easily transfer your name away from you before you can revoke it.  Revoking a name is more like burning a name -- you provably destroy a name in order to make some other thing happen (e.g. imagine a future version of BNS, where you can migrate your name to the new system by burning it on the old system, which in turn grants you the right to instantiate it on the new system).

PR #2312, opened 2021-01-11
feat: adjust export block height to 665750

zone117x:


diwakergupta:
Aaron's branch had a few more changes: https://github.com/blockstack/stacks-blockchain/commit/0948d65ed0e6143f000fc6100242f01086f531ea do we need to include them?

kantai:
Yep, @diwakergupta, I think that the changes in https://github.com/blockstack/stacks-blockchain/commit/0948d65ed0e6143f000fc6100242f01086f531ea should be applied -- they make sure that the database doesn't _try_ to set the export flag until the later block (it also outputs the export data at that point).

zone117x:
Closing in favor of https://github.com/blockstack/stacks-blockchain/pull/2314

PR #2316, opened 2021-01-11
feat: set impossible lock_transfer_block_id

zone117x:
@yknl 
> For the wallet issue that I mentioned during stand up, one way we can disable outgoing transactions from the wallet when we enter the freeze is to return a `lock_transfer_block_id` greater than the current block from the account status API endpoint on core:
https://core.blockstack.org/v1/accounts/1JSwmo6g8zgpKPMVRcrJBeH2oqJE8N9rax/STACKS/status
> 
> example return with `lock_transfer_block_id` set to 9999999:
   ```
address: "1JSwmo6g8zgpKPMVRcrJBeH2oqJE8N9rax"
block_id: 663762
credit_value: "1591022"
debit_value: "1270011"
lock_transfer_block_id: 9999999
txid: "e1d2702b865e180133690dd3dc022085168bf680f943853c53968572ed0a8d6c"
type: "STACKS"
vtxindex: 2263
   ```

### DIFF
--- a/blockstack/blockstackd.py
+++ b/blockstack/blockstackd.py
@@ -1433,10 +1433,6 @@ class BlockstackdRPC(BoundedThreadingMixIn, SimpleXMLRPCServer):
             'vtxindex': account_state['vtxindex'],
             'txid': account_state['txid'],
         }
-        # if block height is after the migration export threshold, return a lock height that will force the wallet to error when sending a tx
-        if result['block_id'] >= virtualchain_hooks.IMPORT_HEIGHT:
-            print log.warning('[v2-upgrade] Forcing lock_transfer_block_id to 9999999 to prevent wallet txs')
-            result['lock_transfer_block_id'] = 9999999
         return result
 
 
@@ -1456,10 +1452,16 @@ class BlockstackdRPC(BoundedThreadingMixIn, SimpleXMLRPCServer):
 
         db = get_db_state(self.working_dir)
         account = db.get_account(address, token_type)
+        last_block = db.lastblock
         db.close()
 
         if account is None:
             return {'error': 'No such account', 'http_status': 404}
+
+        # if block height is after the migration export threshold, return a lock height that will force the wallet to error when sending a tx
+        if last_block >= virtualchain_hooks.IMPORT_HEIGHT:
+            print log.warning('[v2-upgrade] Forcing lock_transfer_block_id to 9999999 to prevent wallet txs')
+            account['lock_transfer_block_id'] = 9999999
 
         state = self.export_account_state(account)
         return self.success_response({'account': state})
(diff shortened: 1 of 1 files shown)


yknl (approved):
Thanks @zone117x Looks good to me!

kantai (approved):

blockstack/blockstackd.py
```diff
@@ -1423,7 +1423,7 @@ def export_account_state(self, account_state):
         """
         Make an account state presentable to external consumers
         """
```
zone117x, line 1425: @jcnelson @kantai this location for a patch was not intuitive for me. There was another area that made more sense, but found this after some local testing. Please confirm

blockstack/blockstackd.py
```diff
@@ -1433,6 +1433,11 @@ def export_account_state(self, account_state):
             'vtxindex': account_state['vtxindex'],
             'txid': account_state['txid'],
         }
+        # if block height is after the migration export threshold, return a lock height that will force the wallet to error when sending a tx
+        if result['block_id'] >= 665750:
```
diwakergupta, line 1437: Can you use the constant IMPORT_BLOCK defined in Aaron's PR?

blockstack/blockstackd.py
kantai: I think @jcnelson has the most context on the appropriate place to apply this. Alternatively, you can catch the forwarding that happens in `api/server.py` and rewrite the lock time there.

blockstack/blockstackd.py
zone117x: idk how to import that from the other file. @kantai

diwakergupta: Not critical if that's not straight forward.

zone117x: Probably is but python module import is cancer

blockstack/blockstackd.py
jcnelson: All this change will do is alter what is (a) returned by the RESTful API and (b) what is returned by `blockstack/lib/client.py`.  If that's what you want, then this is fine.  I think it's more appropriate to do it here than in `api/server.py`.

jcnelson:
I'm a little confused as to what this is supposed to do.  Is it just supposed to trick the 1.0 wallet into thinking it can't spend STX because the tokens are all "locked"?

jcnelson (approved):

kantai:
> I'm a little confused as to what this is supposed to do. Is it just supposed to trick the 1.0 wallet into thinking it can't spend STX because the tokens are all "locked"? If so, then the changes proposed here should be fine, but you'll want to of course test them with the 1.0 wallet.

Yep -- that's the idea.

I think this PR needs a slight tweak -- the `block_id` checked here isn't the current block height, but rather the `block_id` that the account was last modified at, correct?

kantai:
Tested locally with an already passed block height, and this works correctly.

PR #2317, opened 2021-01-11
Ability to configure the satoshi_per_byte for BTC transactions

lgalabru:
Motivated by the current surges on Mainnet, this PR is introducing 3 new settings:
- `satoshis_per_byte`
- `leader_key_tx_estimated_size`
- `block_commit_tx_estimated_size`

And getting rid of `burnchain_op_tx_fee`.

The current default is 50 sats/bytes, it's been tested on Testnet for a [key registration](https://mempool.space/testnet/tx/2f4e3187b3b8bf9d7aafc8e618159f19e2616d667e5bf684d242600144563fa7) and a [block commit](https://mempool.space/testnet/tx/36292112c685da83a0066b8b9aaf4177a10964dfd31204d3003a6d536f17b1dc). 
I took some time to refactor the logic for satoshis allocation, which will squash a few bugs with the UTXOs filtering (minimum amount was incorrect).

I see a TODO about RBF, any action we should take here?

### DIFF
--- a/testnet/stacks-node/Stacks.toml
+++ b/testnet/stacks-node/Stacks.toml
@@ -23,7 +23,7 @@ rpc_bind = "127.0.0.1:20443"
 # password = "secret"
 # timeout = 30
 # local_mining_public_key = "04ee0b1602eb18fef7986887a7e8769a30c9df981d33c8380d255edef003abdcd243a0eb74afdf6740e6c423e62aec631519a24cf5b1d62bf8a3e06ddc695dcb77"
-# satoshis_per_block_commit_op = 1000
+# satoshis_per_block_commit_op = 5500
 # commit_anchor_block_within = 3000
 
 ## Settings for public testnet, relying on a remote bitcoind server
@@ -33,7 +33,7 @@ rpc_bind = "127.0.0.1:20443"
 # chain = "bitcoin"
 # mode = "argon"
 # peer_host = "argon.blockstack.org"
-# satoshis_per_block_commit_op = 1000
+# satoshis_per_block_commit_op = 5500
 # commit_anchor_block_within = 10000
 # rpc_port = 3000
 # peer_port = 18444
--- a/testnet/stacks-node/Stacks.toml
+++ b/testnet/stacks-node/Stacks.toml
@@ -23,7 +23,7 @@ rpc_bind = "127.0.0.1:20443"
 # password = "secret"
 # timeout = 30
 # local_mining_public_key = "04ee0b1602eb18fef7986887a7e8769a30c9df981d33c8380d255edef003abdcd243a0eb74afdf6740e6c423e62aec631519a24cf5b1d62bf8a3e06ddc695dcb77"
-# burnchain_op_tx_fee = 1000
+# satoshis_per_block_commit_op = 1000
 # commit_anchor_block_within = 3000
 
 ## Settings for public testnet, relying on a remote bitcoind server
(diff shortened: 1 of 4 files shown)


testnet/stacks-node/Stacks.toml
```diff
@@ -23,7 +23,7 @@ rpc_bind = "127.0.0.1:20443"
 # password = "secret"
 # timeout = 30
 # local_mining_public_key = "04ee0b1602eb18fef7986887a7e8769a30c9df981d33c8380d255edef003abdcd243a0eb74afdf6740e6c423e62aec631519a24cf5b1d62bf8a3e06ddc695dcb77"
-# burnchain_op_tx_fee = 1000
+# satoshis_per_block_commit_op = 1000
```
kantai, line 26: Event though this is commented out, I think this should probably be `= 5500`

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -730,7 +742,25 @@ impl BitcoinRegtestController {
...
 
-        let (mut tx, utxos) = self.prepare_tx(&public_key, payload.burn_fee, attempt)?;
+        let sunset_fee = if payload.sunset_burn > 0 {
+            cmp::max(payload.sunset_burn, DUST_UTXO_LIMIT)
+        } else {
+            0
+        };
+
+        let number_of_transfers = payload.commit_outs.len() as u64;
+        let value_per_transfer = payload.burn_fee / number_of_transfers;
```
kantai, line 752: Does `value_per_transfer` need to be `= cmp::max(value_per_transfer, DUST_UTXO_LIMIT)`?

kantai (approved):
This LGTM, just a couple comments regarding the dust limit.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: I think in that case, you would go over the cap that was fixed by the miner. I think this is being checked at a higher level, but we can abort the transaction here as well.

kantai: Ah, okay, as long as it gets checked somewhere, that works for me.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -730,7 +742,25 @@ impl BitcoinRegtestController {
...
+
+        let btc_miner_fee = self.config.burnchain.block_commit_tx_estimated_size
+            * self.config.burnchain.satoshis_per_byte
+            * attempt;
```
jcnelson, line 756: Is the `* attempt` at the end increasing the fee needlessly, considering that there's a separate `rbf_fee`?

testnet/stacks-node/Stacks.toml
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2317/commits/46c89a2f299a5593d5137fd25a8a210f6035dd1c

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: ha good catch, I've been doing a few back and forth there, thanks.
Addressed with https://github.com/blockstack/stacks-blockchain/pull/2317/commits/46c89a2f299a5593d5137fd25a8a210f6035dd1c

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2317/commits/c654e94be572e5edabf9a3205f1fe5cbb336a8d6

PR #2318, opened 2021-01-11
Fix key registration spanning over multiple blocks

lgalabru:
This PR is improving a patch for a bug that I spotted a few months ago (https://github.com/blockstack/stacks-blockchain/commit/750687e86a98b5f0820b1a5c818266d68e6c1ebb).
The patch above was essentially inefficient when a key_register op is sitting in the mempool during multiple blocks.

### DIFF
--- a/testnet/stacks-node/src/neon_node.rs
+++ b/testnet/stacks-node/src/neon_node.rs
@@ -957,7 +957,7 @@ fn spawn_miner_relayer(
 }
 
 enum LeaderKeyRegistrationState {
-    None,
+    Inactive,
     Pending,
     Active(RegisteredKey),
 }
@@ -1128,7 +1128,7 @@ impl InitializedNeonNode {
             is_miner,
             sleep_before_tenure,
             atlas_config,
-            leader_key_registration_state: LeaderKeyRegistrationState::None,
+            leader_key_registration_state: LeaderKeyRegistrationState::Inactive,
         }
     }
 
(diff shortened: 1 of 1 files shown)


kantai (approved):
This LGTM!

testnet/stacks-node/src/neon_node.rs
```diff
@@ -966,6 +956,12 @@ fn spawn_miner_relayer(
     Ok(())
 }
 
+enum LeaderKeyRegistrationState {
+    None,
```
jcnelson, line 960: Style nit -- can we avoid using reserved Rust keywords?  Maybe `Nothing` instead of `None`?

jcnelson (approved):
This LGTM, but please address the style nit.

I did have one other design question.  What happens if I stop the miner and re-start it?  Will it detect my previously-registered VRF key and proceed to use it?  Or will it force me to re-register a VRF key?

testnet/stacks-node/src/neon_node.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2318/commits/040777078f052f63d1654e8d4beb6fc11b878237

lgalabru:
At this point, it will force to re-register, but we should be able to add this feature by adding a `Recover` state to the enum. Happy to implement this optimization later this week.

jcnelson:
Yeah, it's not a mainnet blocker.  But please open an issue for it.

PR #2323, opened 2021-01-12
fix: add get-name #2322

### LINKED ISSUE #2322: [BNS] 1 to 1 mapping can't be resolved by contracts
**Describe the bug**
The `owner-name` map defines a 1-to-1 mapping between principals and user names. However, this mapping can't be used by a clarity contract due to https://github.com/clarity-lang/reference/issues/30

**Steps To Reproduce**
try to get the user name of the current tx-sender

**Expected behavior**
add a read-only function to return the name of the given principal.

friedger:
## Description

As a contract developer I want to check properties of the user based on the username
This PR
* adds a method `get-name` that returns the username of a given principal
* fixes #2322 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

- [ ] Link to documentation updates: https://github.com/blockstack/docs/pull/1024

## Testing information
No tests

## Checklist
- [x] Tag 1 of @lgalabru  or @agraebe

### DIFF
--- a/src/chainstate/stacks/boot/bns.clar
+++ b/src/chainstate/stacks/boot/bns.clar
@@ -883,7 +883,7 @@
     name (ok name)
     (err ERR_NAME_NOT_FOUND)))
 
-(define-read-only (get-resolvable-name (owner principal))
+(define-read-only (resolve-principal (owner principal))
   (match (map-get? owner-name owner)
     name (match (name-resolve (get namespace name) (get name name))
       resolved-name (ok name)
--- a/src/chainstate/stacks/boot/bns.clar
+++ b/src/chainstate/stacks/boot/bns.clar
@@ -879,7 +879,9 @@
           (<= block-height (+ (+ lifetime lease-started-at) NAME_GRACE_PERIOD_DURATION)))))))
 
 (define-read-only (get-name (owner principal))
-  (ok (map-get? owner-name owner)))
+  (match (map-get? owner-name owner)
+    name (ok name)
+    (err ERR_NAME_NOT_FOUND)))
 
 (define-read-only (can-receive-name (owner principal))
   (let ((current-owned-name (map-get? owner-name owner)))
--- a/src/chainstate/stacks/boot/bns.clar
+++ b/src/chainstate/stacks/boot/bns.clar
@@ -882,8 +882,8 @@
   (match (map-get? owner-name owner)
     name (match (name-resolve (get namespace name) (get name name))
       resolved-name (ok name)
-      error (err error))
-    (err ERR_NAME_NOT_FOUND)))
+      error (err {code: error, name: (some name)}))
+    (err {code: ERR_NAME_NOT_FOUND, name: none})))
 
 (define-read-only (can-receive-name (owner principal))
   (let ((current-owned-name (map-get? owner-name owner)))
--- a/src/chainstate/stacks/boot/bns.clar
+++ b/src/chainstate/stacks/boot/bns.clar
@@ -878,6 +878,10 @@
           (> block-height (+ lifetime lease-started-at)) 
           (<= block-height (+ (+ lifetime lease-started-at) NAME_GRACE_PERIOD_DURATION)))))))
 
+(define-read-only (get-name (owner principal))
+  (ok (map-get? owner-name owner))
+)
+
 (define-read-only (can-receive-name (owner principal))
   (let ((current-owned-name (map-get? owner-name owner)))
     (if (is-none current-owned-name)
--- a/src/chainstate/stacks/boot/bns.clar
+++ b/src/chainstate/stacks/boot/bns.clar
@@ -878,11 +878,6 @@
           (> block-height (+ lifetime lease-started-at)) 
           (<= block-height (+ (+ lifetime lease-started-at) NAME_GRACE_PERIOD_DURATION)))))))
 
-(define-read-only (get-name (owner principal))
-  (match (map-get? owner-name owner)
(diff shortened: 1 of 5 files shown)


src/chainstate/stacks/boot/bns.clar
```diff
@@ -878,6 +878,9 @@
           (> block-height (+ lifetime lease-started-at)) 
           (<= block-height (+ (+ lifetime lease-started-at) NAME_GRACE_PERIOD_DURATION)))))))
 
+(define-read-only (get-name (owner principal))
+  (ok (map-get? owner-name owner)))
```
kantai, line 882: Since this is a `define-read-only` function, this shouldn't need to return a response (especially since there's no error case here) -- can this function instead just return `(map-get? owner-name owner)`?

friedger: All the other read-only functions return a response value. 

Alternatively, I suggest to return `ERR_NAME_NOT_FOUND` in case of an "owner" without a name:
`(match (map-get? owner-name owner) username (ok username) (err ERR_NAME_NOT_FOUND))`

kantai: Yep, I think using `ERR_NAME_NOT_FOUND` would be better.

agraebe:
@friedger I'd appreciate if you could add another line and document this method in `src/vm/docs/contracts.rs`

jcnelson:
This is a consensus-breaking patch, so please don't merge until we're ready for mainnet (notice that the `test_chainstate_sampled_genesis_consistency` test has failed).

kantai:
> This is a consensus-breaking patch, so please don't merge until we're ready for mainnet (notice that the test_chainstate_sampled_genesis_consistency test has failed).

It should be fine to merge to `next`, right?

jcnelson:
Yeah, it would have to be `next`.  We'll include this along with the patch that sets the mainnet start block.

src/chainstate/stacks/boot/bns.clar
```diff
@@ -878,6 +878,11 @@
           (> block-height (+ lifetime lease-started-at)) 
           (<= block-height (+ (+ lifetime lease-started-at) NAME_GRACE_PERIOD_DURATION)))))))
 
+(define-read-only (get-name (owner principal))
+  (match (map-get? owner-name owner)
+    name (ok name)
```
lgalabru, line 883: Thank you @friedger. I think the current approach is incomplete, you'd want to make sure that this name is still resolving.
So you can pull from that map, use the method `name-resolve`, and then return a result if it returns `ok`.

friedger: @lgalabru for that I would add a function get-resolved-name to keep flexibility for contract developers

friedger: Then docs should be @agraebe :
("get-name", "Returns a response with the username that is associated with the user if any. The name can be resolvable, expired, revoked or in grace period. It is the responsibility of the function caller to act accordingly. If no name is associated with the user an error 2013 (`ERR_NAME_NOT_FOUND`) is returned.")

lgalabru: Thank you for adding the tests!
I like this feature. Can we rename to `resolve-principal` or equivalent, so we increase consistency with `resolve-name`?
I still have some reserves concerning the `get-name`. If we can do an analogy with DNS, it's basically equivalent to resolving an expired domain.
Can you unpack some use cases? 
Thanks!

friedger: @lgalabru I have renamed `get-resolvable-name` as suggested.

As use cases, a contract could act if a name is in grace period and maybe renew it. There are probably smart people out there that will find more use cases.

As the data is available through the RCP api anywhere, adding `get-name` unlocks this api to smart contracts until we have https://github.com/clarity-lang/reference/issues/30

lgalabru: Yeah I'm not saying that this data is not available, I'm just saying that the contract should be in charge of exposing relevant data, and the `get-name` method looks like a foot-gun for developers.

If the name is in grace period, or expired, great, with your current implementation of `resolve-principal`, you will get the right error code.

lgalabru: what you could do in resolve-principal, is, in case of an error, merge the tuple returned by resolve-name with the name that was queried.

friedger: I like that.

I have remove `get-name` and added the name in the error. 
Thank you for your input.

lgalabru: Yay, thank you. We don't have a CI for this tests, running on my machine and will follow-up.

lgalabru: @friedger This PR, by updating BNS, is modifying the signature of the Genesis block. 
We have some testing around that, can you update the tests `test_chainstate_sampled_genesis_consistency` and `test_chainstate_full_genesis_consistency`?
Thank you!

kantai: You should also merge (or rebase) the latest `next` into this branch, so that the hash in `test_chainstate_sampled_genesis_consistency` is up to date when we merge it into `next`.

friedger: merged and updated!

lgalabru (approved):
Looks good to me, thank you @friedger !

kantai (approved):
LGTM!

PR #2325, opened 2021-01-12
Fix: Handle the genesis import format with registered_at, etc. block info

kantai:
This PR updates the genesis import handling logic in the `stx-genesis` target. In the most recent versions of the genesis export in Stacks 1.0, 2 new fields were added to the `chainstate.txt` file. This PR allows the `stx-genesis` target to handle those fields while also remaining "backwards compatible" with the existing chainstate files. Once the newer version of the `chainstate.txt` is approved and merged, we can get rid of this backwards compatibility.

This behavior is tested by the new `chainstate-test.txt` which uses this "new format". And you can see that this does not impact the chainstate, because the consistency test did _not_ need to be updated.


kantai:
Also -- you can see the new format in the #2324 PR.

stx-genesis/src/lib.rs
```diff
@@ -108,24 +108,32 @@ fn read_lockups(deflate_bytes: &'static [u8]) -> Box<dyn Iterator<Item = Genesis
...
-        namespace_id: cols[0].to_string(),
-        importer: cols[1].to_string(),
-        buckets: cols[2].to_string(),
-        base: cols[3].parse::<i64>().unwrap(),
-        coeff: cols[4].parse::<i64>().unwrap(),
-        nonalpha_discount: cols[5].parse::<i64>().unwrap(),
-        no_vowel_discount: cols[6].parse::<i64>().unwrap(),
-        lifetime: cols[7].parse::<i64>().unwrap(),
+    let namespaces = iter_deflated_csv(deflate_bytes).map(|cols| {
+        let bucket_index = if cols.len() >= 10 { 4 } else { 2 };
```
jcnelson, line 112: Is there an example where `cols.len() < 10`?

stx-genesis/chainstate-test.txt
```diff
@@ -1,25 +1,25 @@
...
-10x.id,SP3N5TVVEHGA8B6G3JXWBP476161E32654D9P9A3Y,3d0576cd134fa1ff8d54d8e2ef5da97f5cae36f2
-111111111.id,SP15XBGYRVMKF1TWPXE6A3M0T2A87VYSVF9VFSZ1A,01a8c60c852b75f9c6417491ac89275fb9de774c
-123.id,SP2PEZ0W4PFGYV6ZZ2DWDVNAVVTQPJ9KX8EK4VRVT,ae34427289a8865028eb2d475dafb085060d5625
-zinai.id,SP19QQ0RZ2F3CKJ2G2FT4EA905MKV89FY9K5XB5M3,c7b0bbeb92c2575a3e4a82632d8ec4b7fc86cc19
-zlh.id,SP1E311NW655FXQMQQKEKQKZZJFAA58AVP3GBTAZT,da6a954fdf90c890966cde5ae808ca6c96082428
-zone117x.id,SP3R9402R2XB0WMG0BTEMZ5Q67QFJ7Y8YVJRQCJZN,578ed433dabeb24952c892d5f3fc0cd7b9844bb8
-zumminer_crux.id,SP3DGYX27MM12SSPVGK2NMTR4836DT9VWYJV844BK,fa602059cb01dbe8d1885d8fc2a7c6b265c57e64
-zumminer_dev_crux.id,SPSSW2FFJQ2AMR5KZASBY8XH3F5WXHJV0M22GJ9G,f9569aa25b6325b0f2e3a5f9edb552cd98260f9b
-zumrai.id,SP29EJ0SVM2TRZ3XGVTZPVTKF4SV1VMD8C0GA5SK5,853cd126478237bc7392e65091f7ffa5a1556a33
+name,address,registered_at,expire_block,zonefile_hash
```
jcnelson, line 10: It doesn't look like we're using `registered_at` or `expire_block`, right?

stx-genesis/src/lib.rs
kantai: The `chainstate.txt` file is that case, so it's tested by the `full-genesis` integration test.

stx-genesis/chainstate-test.txt
kantai: That's correct, but it is what the Stacks 1.0 node outputs.

lgalabru: I'm pretty sure that those fields were removed from the dump.
Is it possible that we mixed up commits in that branch @zone117x ?

kantai: Good catch -- yeah, I think that's what's happening -- an older `v1-migration` core node was used to digest the stacks-node backup data and open #2324, which also explains why things weren't re-mapped.

kantai:
Closing this PR, as it should not be necessary.

PR #2326, opened 2021-01-13
Fix: boot_contract_id with testnet address hard-coded

lgalabru:
PR is fixing:

```rust
pub const STACKS_BOOT_CODE_CONTRACT_ADDRESS_STR: &'static str = "ST000000000000000000002AMW42H";
```

Consensus critical and error prone. Eyes wide open during the review, please :)
I wonder if we should add an enum, enumerating the Pox contracts names.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,4 @@ net-test/mnt/bitcoin-*
 net-test/mnt/chainstate-*
 net-test/mnt/*.log
 net-test/mnt/archive/*
+*.log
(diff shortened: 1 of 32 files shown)


kantai (approved):
LGTM, thanks @lgalabru!

kantai:
(Also: A note on the `next` vs. `master` here -- this PR doesn't change the consensus state on testnet, so it can safely merge to `master`)

jcnelson (approved):

PR #2329, opened 2021-01-13
Update satoshis_per_byte comment in config

lgalabru:


diwakergupta (approved):
I believe there's another instance at line 26, can you fix that too? Otherwise LGTM!

PR #2333, opened 2021-01-14
Pick initial peers at random

jcnelson:
Pick initial peers at random, so when the neighbor walk tries to keep a bootstrap node in the neighbor set, it picks one at random.


lgalabru (approved):

diwakergupta:
LGTM once tests pass.

kantai (approved):

PR #2337, opened 2021-01-14
Move SIPs to Stacksgov

jcnelson:
This PR moves all SIPs to their new home at https://github.com/stacksgov/sips.


sip/README.md
```diff
@@ -0,0 +1,4 @@
+# Stacks Improvement Proposals
+
+This directory formally contained all of the in-progress Stacks Improvement
```
kantai, line 3: Typo: formally -> formerly

diwakergupta (changes requested):
If you've already copied over the SIPs, can we do this post launch?

A lot of places link to SIPs directly (esp. SIP-007) -- instead of deleting, my preference would be to just replace their text with a link to the new corresponding SIP location. After a few weeks of that we can delete the SIPs from here.

lgalabru (commented):
If we already have the new links, should we also update the  `README.md` for this repo?

jcnelson:
TODO: Add links to the new SIPs where the old SIPs are.

PR #2340, opened 2021-01-14
Feat: panicking enforcement of mainnet configs

kantai:
This PR forces a `mainnet` configured node to start with default settings for `magic_bytes`, `block_limit`, `use_test_genesis`, and without settings for `initial_balances` -- these settings, if applied, would cause the node to misbehave and other nodes in the network would consider the node to be invalid (and vice-versa).

### DIFF
--- a/testnet/stacks-node/src/config.rs
+++ b/testnet/stacks-node/src/config.rs
@@ -286,6 +286,7 @@ impl ConfigFile {
             peer_host: Some("bitcoin.blockstack.com".to_string()),
             username: Some("blockstack".to_string()),
             password: Some("blockstacksystem".to_string()),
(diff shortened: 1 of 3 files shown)


testnet/stacks-node/src/config.rs
```diff
@@ -286,6 +286,7 @@ impl ConfigFile {
             peer_host: Some("bitcoin.blockstack.com".to_string()),
             username: Some("blockstack".to_string()),
             password: Some("blockstacksystem".to_string()),
+            magic_bytes: Some("R5".to_string()),
```
jcnelson, line 289: I thought mainnet was something else?

jcnelson (approved):

lgalabru (approved):
LGTM

PR #2341, opened 2021-01-14
Add sample for mainnet config

lgalabru:
Add a file that mainnet miners can use as a base for their configuration


testnet/stacks-node/conf/mainnet-miner-conf.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[burnchain]
+chain = "bitcoin"
+mode = "mainnet"
+peer_host = "127.0.0.1"
+rpc_port = 8332
+peer_port = 8333
+username = "<USERNAME>"
+password = "<PASSWORD>"
+satoshis_per_byte = 50
```
jcnelson, line 17: This might need to be higher

testnet/stacks-node/conf/mainnet-miner-conf.toml
```diff
@@ -0,0 +1,18 @@
+[node]
+rpc_bind = "0.0.0.0:20443"
+p2p_bind = "0.0.0.0:20444"
+seed = "<YOUR_SEED>"
+local_peer_seed = "<YOUR_SEED>"
+miner = true
+bootstrap_node = "02da7a464ac770ae8337a343670778b93410f2f3fef6bea98dd1c3e9224459d36b@seed-0.mainnet.stacks.co:20444,02afeae522aab5f8c99a00ddf75fbcb4a641e052dd48836408d9cf437344b63516@seed-1.mainnet.stacks.co:20444,03652212ea76be0ed4cd83a25c06e57819993029a7b9999f7d63c36340b34a4e62@seed-2.mainnet.stacks.co:20444"
+
```
jcnelson, line 8: do we need `mine_microblocks = true`, or is that already `true` by default?

lgalabru: true by default

jcnelson (approved):

testnet/stacks-node/conf/mainnet-miner-conf.toml
lgalabru: Updated to `100`

kantai (approved):

PR #2342, opened 2021-01-14
Set mainnet first block hash

kantai:
This PR sets the first block hash, first block timestamp, and the magic bytes for Stacks 2.0.

This PR also bumps the Xenon magic bytes.


jcnelson (approved):

lgalabru (approved):
Looks good to me!

diwakergupta (approved):

PR #2350, opened 2021-01-15
Feat/mempool-pusher

jcnelson:
This PR adds a script in the `tools/` directory that, when run, will periodically scan a local mempool DB and push transactions to a given node's outbound neighbors.

### DIFF
--- a/tools/mempool-pusher.sh
+++ b/tools/mempool-pusher.sh
@@ -27,6 +27,7 @@ debug() {
    fi
 }
 
+# Prints out the list of mempool transactions with acceptance times later than a given time.
 # args:
 # $1 mempool path
 # $2 minimum accept time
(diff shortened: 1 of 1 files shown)


tools/mempool-pusher.sh
```diff
@@ -0,0 +1,168 @@
...
+
+debug() {
+   if [ $DEBUG -ne 0 ]; then
+      log "$1"
+   fi
+}
+
+# args:
+# $1 mempool path
+# $2 minimum accept time
```
lgalabru, line 32: Looks great! Any chance we can add a commented example for each function?

kantai (approved):

tools/mempool-pusher.sh
jcnelson: Added in 23a8adcf6c67cb76d6aaccf00516d7bc272ce11a

lgalabru: Thank you!!

lgalabru (approved):

lgalabru:
Should we merge to develop instead?

jcnelson:
Sure; done

PR #2351, opened 2021-01-15
Tests: fix the expected default xenon version

kantai:
We bumped the default xenon version, but the test that checks for this didn't have it's check bumped.


kantai:
This is addressed in #2352 as well, so closing this.

Issue #2354, opened 2021-01-17
Mempool should not admit transfers with `from = to`


kantai:
Currently, the mempool will admit certain semantically invalid transfers. For example, transactions where the `origin_address` is equal to the recipient address. Miners will correctly not include these transactions in blocks, but it leads to a pretty bad user experience -- an invalid transaction will show up as pending in explorers, etc., indefinitely and with no error message returned to a user wallet or elsewhere (and users would need to use replace-by-fee to correct the transaction or wait 256 blocks for the transaction to be garbage collected).

PR #2357, opened 2021-01-17
Feature: Check transfer semantics before mempool admission

kantai:
This implements #2354, applying some checks to transfer transactions before admitting them to the mempool:

1. If `recipient = origin`, then reject the transaction from the mempool
2. If `amt == 0`, then reject the transaction from the mempool

Either of these would already disqualify the transaction from being included in a block, but by applying these checks before mempool admission, clients are able to correct the error.

### DIFF
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -149,6 +149,8 @@ pub enum MemPoolRejection {
         principal: PrincipalData,
         is_origin: bool,
     },
+    TransferRecipientIsSender(PrincipalData),
+    TransferAmountMustBePositive,
     DBError(db_error),
     Other(String),
 }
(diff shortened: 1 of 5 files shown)


jcnelson (approved):
LGTM!

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4996,6 +5003,25 @@ impl StacksChainState {
...
+    fn can_admit_mempool_semantic(tx: &StacksTransaction) -> Result<(), MemPoolRejection> {
+        match tx.payload {
+            TransactionPayload::TokenTransfer(ref recipient, amount, ref _memo) => {
+                let origin = PrincipalData::from(tx.origin_address());
+                if &origin == recipient {
+                    return Err(MemPoolRejection::TransferRecipientIsSender(origin));
+                }
+                if amount == 0 {
+                    return Err(MemPoolRejection::TransferAmountMustBePositive);
+                }
```
lgalabru, line 5018: Could we also check that testnet addresses are used on testnets, and mainnet addresses on mainnet?

lgalabru (approved):

src/chainstate/stacks/db/blocks.rs
kantai: Yes, will add

Issue #2347, opened 2021-01-15
insufficient fee, rejecting replacement (Failed to submit won block 666155 off block 666154 instead used 666152 even though 666153 and 666154 were known)


friedger:
**Describe the bug**
I won a block and then submitted off an older block even though I have committed to the younger blocks. From stacks-dump:

```
666152 1 666151 69 br1$ s:425c38deca p:fe60d3cf13 c:8b80648142 i:3312a8d43a b:00000000000000000007 @+ 315000  69> SP137NZXD0 (30) SP1ER0V0WX (110) SP1K1A1PMG (20) SP25Y6NQXB*(40) SP2HFBWV0D (20) SP334VW9T9 (60) SP3F0CDFN2 (20) SPY1PP2KZ6 (15) 
666153 1 666152 70 br1$ s:fa7b09b77e p:425c38deca c:928ad57479 i:3840b588f4 b:00000000000000000000 @+ 315000  70> SP137NZXD0 (30) SP1ER0V0WX (110) SP1K1A1PMG (20) SP25Y6NQXB*(40) SP2HFBWV0D (20) SP334VW9T9 (60) SP3F0CDFN2 (20) SPY1PP2KZ6 (15) 
666154 1 666153 71 br1$ s:9753fa66b0 p:fa7b09b77e c:b5ba84fecc i:fe6ed1bfed b:00000000000000000006 @@ 315000  71> SP137NZXD0 (30) SP1ER0V0WX*(110) SP1K1A1PMG (20) SP25Y6NQXB (40) SP2HFBWV0D (20) SP334VW9T9 (60) SP3F0CDFN2 (20) SPY1PP2KZ6 (15) 
666155 3 666152 70 br7  s:fd4ca61313 p:425c38deca c:d688026dc2 i:a0a2079af8 b:0000000000000000000d    700000  70  SP137NZXD0 (30) SP1ER0V0WX (110) SP1K1A1PMG*(20) SP1RJH1641 (400) SP25Y6NQXB (40) SP2HFBWV0D (20) SP334VW9T9 (60) SP3F0CDFN2 (20) 
```
**Expected behavior**
Build off 666154

**Additional context**
From the log:
```
Succeeded assembling Stacks block #72: 6597d692fee25ebec60cd792aaa3c81e59623db56d5028e323dbd864b967a3ea, with 13 txs, attempt 13

[...]

PR #2363, opened 2021-01-19
Fix: address #2347

lgalabru:
This PR aims at fixing issue #2347.
Instead of using the computed size of the initial transaction for computing the RBF value, we'll reuse the constant, coming from the config.
On that note, after looking at a few transactions in the explorer, using 352 bytes instead of 350 and  291 instead of 290 would be more accurate.

### DIFF
--- a/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
+++ b/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
@@ -60,7 +60,6 @@ pub struct BitcoinRegtestController {
     use_coordinator: Option<CoordinatorChannels>,
     burnchain_config: Option<Burnchain>,
     last_utxos: Vec<UTXO>,
-    last_tx_len: u64,
     min_relay_fee: u64, // satoshis/byte
 }
 
@@ -121,7 +120,6 @@ impl BitcoinRegtestController {
             chain_tip: None,
             burnchain_config,
             last_utxos: vec![],
-            last_tx_len: 0,
             min_relay_fee: 1024, // TODO: learn from bitcoind
         }
     }
@@ -158,7 +156,6 @@ impl BitcoinRegtestController {
             chain_tip: None,
             burnchain_config: None,
             last_utxos: vec![],
-            last_tx_len: 0,
             min_relay_fee: 1024, // TODO: learn from bitcoind
         }
     }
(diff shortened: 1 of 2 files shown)


jcnelson (approved):
The changes look good to me, but what's the best way to test that this change is actually working?

Also, why can't we just use the actual size of the transaction, instead of using a hard-coded constant?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -757,7 +751,7 @@ impl BitcoinRegtestController {
         let btc_miner_fee = self.config.burnchain.block_commit_tx_estimated_size
             * self.config.burnchain.satoshis_per_byte;
 
-        let rbf_fee = (attempt.saturating_sub(1) * self.last_tx_len * self.min_relay_fee) / 1000;
+        let rbf_fee = attempt.saturating_sub(1) * btc_miner_fee;
```
kantai, line 754: This is a pretty steep fee increase for rbf, right? This would lead to, for example, something that looked like:

```
tx_fee attempt 1: 40,000 satoshis
tx_fee attempt 2: 80,000 satoshis
tx_fee attempt 3: 120,000 satoshis
```

Would  a better increase be like `attempt.saturating_sub(1) * self.config.burnchain.block_commit_tx_estimated_size`?

psq:
note: the transactions during the prepare phase are shorter (2 outputs instead of 3), but this seems fine to me

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
jcnelson: I think it is steep by design.  Consider rule 3 in BIP0125 [1], which defines RBF:

> The replacement transaction pays an absolute fee of at least the sum paid by the original transactions.

@lgalabru Can you test this out and verify that we cannot go lower than `attempt.saturating_sub(1) * btc_miner_fee`?

[1] https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki

lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2363/commits/eaacaa7cdeef23f562c41ea0b4401baff8df90fe.
Will give it a try on Xenon.

kantai (approved):

jcnelson (dismissed):
Re-approving.  Can we test this on Xenon before rolling out to production?

jcnelson:
I'm still seeing this error in production:

```
ERRO [1611380891.227426] [testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs:945] [relayer] Bitcoin RPC failure: transaction submission failed - Network("Bitcoin RPC: status(500) != success, body is \'Object({\"error\": Object({\"code\": Number(-26), \"message\": String(\"insufficient fee, rejecting replacement e88cc249435ac85e261795ca5979e514ed5f84667bed7f99c33a6b7aded4efc9; new feerate 0.00201000 BTC/kB <= old feerate 0.00201142 BTC/kB\")}), \"id\": String(\"stacks\"), \"result\": Null})\'")
```

jcnelson:
To fix this, we need to make sure that the fee per byte never decreases (in fact, it needs to strictly _increase_).

jcnelson:
Superseded by #2383

PR #2366, opened 2021-01-19
Developer tool: introducing Puppet chain

lgalabru:
The goal of this PR is to give us the ability to reproduce some weird cases, that can't really be reproduced in Testnet / Mainnet.
General idea is to recycle our old `neon-controller` into something that can "play" a bitcoin chain layout.

**Example**

With the following config:

```toml
[[blocks]]
count = 1
block_time = 20000
ignore_txs = false

[[blocks]]
count = 2
block_time = 1000
ignore_txs = true


[[blocks]]
count = 10
block_time = 30000
ignore_txs = false

```

**Puppet-chain** will:
1) mine 1 block, forward to bitcoind all the incoming transactions during the next 20 seconds, then 
2) mine 1 block, buffer submitted transactions and wait 1 second, 
3) mine 1 block, buffer submitted transactions and wait 1 second, 
4) mine 1 block, submit all the transactions buffered during 2) and 3) and wait 30 seconds,
etc.

This should help us reproduce, fix and test, at least manually, the current issues we're facing.


kantai (approved):
This LGTM, though it would be nice if there was a config setting like `continue_mining_after_last_block` (or something less verbose) that controlled the `get_block_time_at_height` such that after the last `[[ blocks ]]` definition, the puppet chain continued issuing blocks with the last block's definition.

testnet/puppet-chain/src/main.rs
```diff
@@ -92,38 +94,76 @@ async fn main() -> http_types::Result<()> {
...
-                eprintln!("{}", err);
+        if should_ignore_txs {
+            // Returns ok
+            println!("will buffer request from {}", stream.peer_addr()?);
+            async_h1::accept(&addr, stream.clone(), |_| async {
+                Ok(Response::new(StatusCode::Ok))
+            })
+            .await?;
+            // Enqueue request
+            buffered_requests.push_back((addr, stream));
```
jcnelson, line 148: This approach feels fragile to me.  It looks like you're accepting all inbound connections in the `should_ignore_txs` branch, and then just...idling them?  What happens if the client times out?  What happens if you have enough inbound connections during the block that you run out of file descriptors?  If the point of doing this is just to buffer transactions, but not to delay other RPCs to bitcoind (like `getnetworkinfo` or `getchaininfo`), then why not just accept and buffer up the transactions, and then when the next block gets generated, send them to the bitcoind node then?  Then, downstream clients won't be kept waiting.  You could even immediately reply to downstream clients with the txid, even though you've buffered up the transactions.

lgalabru: No, we're not idling them, we are blindly returning some `200s` (see line `144`).

jcnelson (approved):

Issue #2364, opened 2021-01-19
miner keeps burning commitment and tx fee without fully synced with current chain tip


sambit4111:
**Describe the bug**
Issue is observed when stacks node miner is restarted, before even syncing to current chain tip to match with seed nodes, miner starts burning commitment fee and transaction fees.

**Steps To Reproduce**
Restart the miner, have some UTXO on the btc address so that miner starts as miner mode.

**Expected behavior**
Miner should not spend commitment and tx fee until fully synced with current chain tip to avoid loss of funds to miners.

**Environment (please complete the following information):**
 - OS: Any
 - Rust version - 1.49.0
 - Version of the appropriate binary / software package - 2.0.2

**Additional context**
NA

If you are encountering an issue with a smart contract, please include the smart contract code
that demonstrates the issue.
NA

diwakergupta:
@sambit4111 can you please include some more details? Right now the issue doesn't have sufficient details to debug this further. DEBUG logs, relevant transaction IDs (BTC commitments for instance) etc.

diwakergupta:
Also include your config file (omit seed or any other sensitive details), thanks!

sambit4111:
DEBUG logs are not enabled as it grows very fast in size and also presumably slows down CPU, few times it has been highlighted in discord that slow CPU might be causing transaction delays, that's the reason DEBUG logs are not available.
BTC transaction IDs are as follows

https://www.blockchain.com/btc/tx/61073778866d40982f1576275359271750cee68bba3e53504377da004dbd3715
https://www.blockchain.com/btc/tx/c4bf08a2bb3b6aa4d95adac6fcafb9803edf6010a730990bae3fd923d5152d27

toml file -

[node]
rpc_bind = "0.0.0.0:20443"
p2p_bind = "0.0.0.0:20444"
bootstrap_node = "02da7a464ac770ae8337a343670778b93410f2f3fef6bea98dd1c3e9224459d36b@seed-0.mainnet.stacks.co:20444,02afeae522aab5f8c99a00ddf75fbcb4a641e052dd48836408d9cf437344b63516@seed-1.mainnet.stacks.co:20444,03652212ea76be0ed4cd83a25c06e57819993029a7b9999f7d63c36340b34a4e62@seed-2.mainnet.stacks.co:20444"

# Enter your private key here
seed = "YOUR-PRIVATE-KEY"
local_peer_seed = "YOUR-PRIVATE-KEY"
miner = true

[burnchain]
[...]

PR #2370, opened 2021-01-21
Feature: Do not mine while node is initially syncing Stacks blocks

kantai:
This is one method for addressing #2364 -- during initial bootup, a `stacks-node` instance will query a set of "height hint" peers. The node will _not_ begin mining until it reaches the max height returned by those peers.

This PR also sets a default set of height hint nodes, as well as providing more proactive default bootstrap node selection for mainnet configurations (i.e., if the `bootstrap_node` field is _not_ set in the config, but the config is a mainnet config, use the default set).

### DIFF
--- a/testnet/stacks-node/src/config.rs
+++ b/testnet/stacks-node/src/config.rs
@@ -292,8 +292,16 @@ impl ConfigFile {
             "02afeae522aab5f8c99a00ddf75fbcb4a641e052dd48836408d9cf437344b63516@seed-1.mainnet.stacks.co:20444",
             "03652212ea76be0ed4cd83a25c06e57819993029a7b9999f7d63c36340b34a4e62@seed-2.mainnet.stacks.co:20444"].join(",");
 
+        let height_hint_nodes = [
+            "seed-0.mainnet.stacks.co:20443",
+            "seed-1.mainnet.stacks.co:20443",
+            "seed-2.mainnet.stacks.co:20443",
+        ]
+        .join(",");
+
         let node = NodeConfigFile {
             bootstrap_node: Some(bootstrap_nodes),
+            height_hint_nodes: Some(height_hint_nodes),
             miner: Some(false),
             ..NodeConfigFile::default()
         };
@@ -421,7 +429,7 @@ pub const MAINNET_BLOCK_LIMIT: ExecutionCost = ExecutionCost {
 impl Config {
     pub fn from_config_file(config_file: ConfigFile) -> Config {
         let default_node_config = NodeConfig::default();
-        let (mut node, bootstrap_node, deny_nodes) = match config_file.node {
+        let (mut node, bootstrap_node, height_hint_nodes, deny_nodes) = match config_file.node {
             Some(node) => {
                 let rpc_bind = node.rpc_bind.unwrap_or(default_node_config.rpc_bind);
                 let node_config = NodeConfig {
@@ -437,6 +445,7 @@ impl Config {
                     p2p_bind: node.p2p_bind.unwrap_or(default_node_config.p2p_bind),
                     p2p_address: node.p2p_address.unwrap_or(rpc_bind.clone()),
                     bootstrap_node: vec![],
+                    height_hint_nodes: vec![],
                     deny_nodes: vec![],
                     data_url: match node.data_url {
                         Some(data_url) => data_url,
@@ -467,9 +476,14 @@ impl Config {
                         .unwrap_or(default_node_config.pox_sync_sample_secs),
                     use_test_genesis_chainstate: node.use_test_genesis_chainstate,
                 };
-                (node_config, node.bootstrap_node, node.deny_nodes)
+                (
+                    node_config,
+                    node.bootstrap_node,
+                    node.height_hint_nodes,
+                    node.deny_nodes,
+                )
             }
-            None => (default_node_config, None, None),
+            None => (default_node_config, None, None, None),
         };
 
         let default_burnchain_config = BurnchainConfig::default();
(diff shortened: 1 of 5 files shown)


diwakergupta:
> This is one method for addressing #2364 -- during initial bootup, a `stacks-node` instance will query a set of "height hint" peers. The node will _not_ begin mining until it reaches the max height returned by those peers.
> 
> This PR also sets a default set of height hint nodes, as well as providing more proactive default bootstrap node selection for mainnet configurations (i.e., if the `bootstrap_node` field is _not_ set in the config, but the config is a mainnet config, use the default set).

I'm curious if we expect scenarios where one would want *distinct* sets of "height hint" nodes that are different from bootstrap nodes. Put another way, what's the downside of always using bootstrap nodes provide height hints?

kantai:
> I'm curious if we expect scenarios where one would want distinct sets of "height hint" nodes that are different from bootstrap nodes. Put another way, what's the downside of always using bootstrap nodes provide height hints?

There's two reasons to separate them out (both of which could probably be solved differently):

1. The `rpc_port` and `p2p_port` are configurable separately. So, conceivably, while the bootstrap nodes listen on `p2p_port = 20444`, they may use a port like `20445` for RPC rather than the default `20443`.
2.  Making this configurable allows a miner to turn off the feature by setting the height_hint_nodes to `""`

lgalabru:
I'm also curious about this angle, does that mean that we can't safely rely on the boolean `ibd`, returned by `pox_watchdog.pox_sync_wait`?

kantai:
> I'm also curious about this angle, does that mean that we can't safely rely on the boolean ibd, returned by pox_watchdog.pox_sync_wait?

Yes -- when testing on my mock miner, the node usually tries to start mining once it reaches block height ~120 or so, and takes about another 20 minutes or so to reach the chain tip. So if I were really mining, that'd be 2-3 commits on very old chain tips.

testnet/stacks-node/src/config.rs
```diff
@@ -292,8 +292,16 @@ impl ConfigFile {
             "02afeae522aab5f8c99a00ddf75fbcb4a641e052dd48836408d9cf437344b63516@seed-1.mainnet.stacks.co:20444",
             "03652212ea76be0ed4cd83a25c06e57819993029a7b9999f7d63c36340b34a4e62@seed-2.mainnet.stacks.co:20444"].join(",");
 
+        let height_hint_nodes = [
+            "seed-0.mainnet.stacks.co:20443",
+            "seed-1.mainnet.stacks.co:20443",
+            "seed-2.mainnet.stacks.co:20443",
+        ]
+        .join(",");
```
jcnelson, line 300: So, I'm not convinced that we need `height_hint_nodes` at all.  This information is already available to you -- the `PeerNetwork` struct already tracks each of its neighbors' highest known burnchain block heights in its `ConversationP2P` structs.  You ought to be able to add a method to `PeerNetwork` that just goes through the set of `ConversationP2P` structs and extracts the highest reported burn block height (if there are any nodes connected at all).

jcnelson (changes requested):
I don't think this approach is the right way to solve this problem.  I think it is would be simpler and less error-prone to just pull the network's highest known burnchain block height from the p2p thread, since it's already tracking this information for you.

testnet/stacks-node/src/config.rs
kantai: But we are trying to get _Stacks_ height, not burnchain block heights. Using the PeerNetwork struct would require instantiating a new network object (and driving the events in it) during startup. That seems way more error prone than just directly querying the `v2/info` endpoint.

kantai:
> I don't think this approach is the right way to solve this problem. I think it is would be simpler and less error-prone to just pull the network's highest known burnchain block height from the p2p thread, since it's already tracking this information for you.

We're trying to obtain the Stacks height, not the burnchain height. Using the P2P thread would allow _any_ node in the network to increase this highest known Stacks height, which would be a bad outcome-- there is a fair amount of trust involved here: if a node reported a very high height, it would cause the miners to not mine.

jcnelson:
> Yes -- when testing on my mock miner, the node usually tries to start mining once it reaches block height ~120 or so, and takes about another 20 minutes or so to reach the chain tip. So if the node were really mining, that'd be 2-3 commits on very old chain tips.

So, the `PoxSyncWatchdog` returns `false` if it detects that (1) it doesn't need to download any more burnchain blocks, and (2) it detects that the node hasn't downloaded or processed any new blocks for a time.  I noticed this phenomenon as well, and I think it's due to the fact that the Stacks node currently just stops processing Stacks blocks altogether until the next burnchain block arrives (I fixed this in the WAL PR; @kantai commented on it).  Can you apply that small fix here and see if driving Stacks block processing makes this problem go away by itself?

jcnelson:
The consensus from the architecture meeting was to make the target Stacks block height a command-line option.  Then, users who want to pull the height from a node can do so with the usual shell command substitution, or they can supply their own.

kantai:
Updated to use `--mine-at-height` command line argument instead.

jcnelson (approved):
LGTM!  Thanks @kantai

lgalabru (approved):
Looks good to me!

MaksimalistT:
Thank you for working on this issue, this is really important
just a few comments:
>Updated to use --mine-at-height command line argument instead.
in this case miners would have to guess at what height miner will complete sync
hit lower - wasted bids
hit higher - wasted profits

could you please consider implementing commands to the process, like bitcoin-cli does
user will active mining when it consider miner is ready just send command for ex.: 
>stacks-node-cli start_mine
and for lazy ones we can add argument from which height to mine
>stacks-node-cli start_mine --at-height=1600

it would be great to do the same to preserve already loaded data in working_dir
>stacks-node-cli stop
process gracefully finish i/o operations and leave data in working dir not corrupted, so miner can start not from the scratch

Issue #2361, opened 2021-01-19
Networking: long delay during initial block download


jcnelson:
When bootstrapping a node, I noticed that it encountered a pathological corner case that cause it to stall its block download for a long period of time (e.g. 15-30 min).  It seemed to have downloaded all blocks and microblocks, except for one missing stream (whose absence prevented it from processing these blocks).  I suspect it's a bug or deficiency in the downloader logic that causes the node to accidentally cycle through all sortitions before trying a missing block or microblock stream again, even though its absence is holding up block-processing.  I've captured a full log of the boot-up process so I can dig more into this.

Issue #2334, opened 2021-01-14
Chainstate: use WAL mode for the database


jcnelson:
Make the chainstate DB use WAL mode so an I/O-heavy relay thread doesn't starve the P2P thread.

PR #2371, opened 2021-01-21
Feat/p2p no blocking

### LINKED ISSUE #2369: P2P: gracefully handle nodes that advertise private IP ranges
Just assume that the routable IP address we used to contact the neighbor is its correct address.

jcnelson:
This PR fixes #2369, #2362, #2361, and #2334.  This makes it so the sortition DB, burnchain DB, and chainstate MARFs all operate in WAL mode, which _significantly_ improves their write performance while also minimizing the chance of a read conflicting with a write.  It also tweaks a few things about the unconfirmed state trie management code to minimize the number of reads and writes that the p2p networking thread will do, which in turn makes the p2p thread much, much more responsive during periods of heavy I/O (I've never seen it wait for more than half a second when testing during bootstrapping).

These changes were guided by printing out a backtrace each time the `tx_busy_handler()` got called.  Using the backtrace, I was able to identify which points in the code were most responsible for causing database contention.  By far the biggest performance fix was making it so that we don't always create a MARF transaction when opening a MARF to initialize its metadata tables (in `src/vm/database/marf.rs`).  The p2p thread had been calling this each and every pass as part of refreshing its unconfirmed state trie view.

The second biggest performance fix was in minimizing the number of times the relayer or p2p threads refreshed the unconfirmed state trie view.  Now, refreshing the read-only view in the p2p thread happens without any I/O on the chainstate database, and refreshing the state trie while mining microblocks or processing microblocks only occurs when there are new microblocks to consider (whereas before, the relayer would create and close a MARF transaction even if there were no new microblocks to apply to the unconfirmed state trie).

There is still a little bit of database lock contention between the burnchain-db thread and the chains-coordinator thread during boot-up -- the burnchain-db thread will be storing burnchain operations while the chains-coordinator is trying to read them and process sortitions.  There is also still a little bit of lock contention between the p2p thread reading the sortition DB to load the canonical chain tip and some of the other threads, but they are quite rare (only four instances detected during a node boot-up).  This lock contention was higher in the past due to the way the p2p thread loaded a burnchain view -- it did a MARF read for each ancestor block hash it had to load.  I removed this entirely by instead making it so each `BlockSnapshot` has a `parent_sortition_id` field that points to its parent sortition in the same sortition fork.  Now, loading a burnchain view doesn't require any MARF reads.

### DIFF
--- a/src/chainstate/stacks/index/trie_sql.rs
+++ b/src/chainstate/stacks/index/trie_sql.rs
@@ -62,6 +62,7 @@ use std::convert::{TryFrom, TryInto};
 
 use chainstate::stacks::index::Error;
 
+use util::db::sql_pragma;
 use util::db::tx_begin_immediate;
 use util::log;
 
@@ -91,6 +92,8 @@ CREATE TABLE IF NOT EXISTS block_extension_locks (block_hash TEXT PRIMARY KEY);
 ";
 
 pub fn create_tables_if_needed(conn: &mut Connection) -> Result<(), Error> {
+    sql_pragma(conn, "PRAGMA journal_mode = WAL;")?;
+
     let tx = tx_begin_immediate(conn)?;
 
     tx.execute_batch(SQL_MARF_DATA_TABLE)?;
@@ -103,6 +106,15 @@ pub fn create_tables_if_needed(conn: &mut Connection) -> Result<(), Error> {
 pub fn get_block_identifier<T: MarfTrieId>(conn: &Connection, bhh: &T) -> Result<u32, Error> {
     conn.query_row(
         "SELECT block_id FROM marf_data WHERE block_hash = ?",
+        &[bhh],
+        |row| row.get("block_id"),
+    )
+    .map_err(|e| e.into())
+}
+
+pub fn get_mined_block_identifier<T: MarfTrieId>(conn: &Connection, bhh: &T) -> Result<u32, Error> {
+    conn.query_row(
+        "SELECT block_id FROM mined_blocks WHERE block_hash = ?",
         &[bhh],
         |row| row.get("block_id"),
     )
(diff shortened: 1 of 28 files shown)


kantai:
This should be based on `develop`, right?

src/chainstate/stacks/index/trie_sql.rs
```diff
@@ -171,13 +183,21 @@ pub fn write_trie_blob_to_mined<T: MarfTrieId>(
...
     data: &[u8],
 ) -> Result<u32, Error> {
-    let args: &[&dyn ToSql] = &[block_hash, &data];
-    let mut s =
-        conn.prepare("INSERT OR REPLACE INTO mined_blocks (block_hash, data) VALUES (?, ?)")?;
-    let block_id = s
-        .insert(args)?
-        .try_into()
-        .expect("EXHAUSTION: MARF cannot track more than 2**31 - 1 blocks");
+    if let Ok(block_id) = get_mined_block_identifier(conn, block_hash) {
```
kantai, line 186: Why not `INSERT OR REPLACE`?

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -373,6 +373,12 @@ impl RunLoop {
                     "Synchronized burnchain up to block height {} (chain tip height is {})",
                     block_height, burnchain_height
                 );
+            } else if ibd {
+                // drive block processing after we reach the burnchain tip.
+                // we may have downloaded all the blocks already,
+                // so we can't rely on the relayer alone to
+                // drive it.
+                coordinator_senders.announce_new_stacks_block();
```
kantai, line 381: I'm curious what scenario this is addressing -- `net::relay` calls this every time a Stacks block is received (which isn't driven by `node.relay_sortition_notify()`). How could we end up in a scenario where a Stacks block was received but the coordinator needs to be woken up again?

jcnelson:
Sorry, meant to open against `develop`.

src/chainstate/stacks/index/trie_sql.rs
jcnelson: `REPLACE` allocates a new rowid, even if the row already exists.

testnet/stacks-node/src/run_loop/neon.rs
jcnelson: I don't know.  All I know is that without this, I'm able to boot up a Stacks node and have it reach the burn chain tip, but then stall out when processing Stacks blocks.  It has all of them -- it downloaded them all -- but it just stops processing them.  It only resumes processing them once a new Bitcoin block is processed.

kantai: Ah, okay -- this probably needs to be invoked in #2370's "do not mine" branch as well, then.

kantai (approved):
LGTM! Very excited for WAL mode!

lgalabru (approved):
Looks good to me, thank you @jcnelson! Let's make sure that tests are green before merging - the failures look legit.

testnet/stacks-node/src/run_loop/neon.rs
jcnelson: Actually, I think I figured this out.  The node downloads microblock streams separately from blocks (it must do so, because it needs both the parent and child block to validate the microblock stream).  This means that in practice, the node spends the latter portion of its boot-up sequence downloading just microblock streams.  But, these don't trigger a Stacks block announcement in the relayer.  I've since patched it to do this.

kantai: Ah, yes, that makes sense

PR #2375, opened 2021-01-22
Remove redundant clones identified by Clippy.

diwakergupta:
Per https://rust-lang.github.io/rust-clippy/rust-1.49.0/index.html#redundant_clone:

	Checks for a redundant clone() (and its relatives) which clones an
	owned value that is going to be dropped without further use.

	It is not always possible for the compiler to eliminate useless
	allocations and deallocations generated by redundant clone()s.

No functional changes.

### DIFF
--- a/src/address/c32.rs
+++ b/src/address/c32.rs
@@ -74,8 +74,7 @@ fn c32_normalize(input_str: &str) -> String {
         .to_uppercase()
         .replace("O", "0")
         .replace("L", "1")
-        .replace("I", "1")
-        .to_string();
+        .replace("I", "1");
     norm_str
 }
 
--- a/src/burnchains/bitcoin/blocks.rs
+++ b/src/burnchains/bitcoin/blocks.rs
@@ -214,7 +214,7 @@ impl BitcoinMessageHandler for BitcoinBlockDownloader {
                 block_hash = ipc_header.block_header.header.bitcoin_hash();
             }
             _ => {
-                return Err(btc_error::UnhandledMessage(msg.clone()));
+                return Err(btc_error::UnhandledMessage(msg));
             }
         }
 
--- a/src/burnchains/bitcoin/indexer.rs
+++ b/src/burnchains/bitcoin/indexer.rs
@@ -277,14 +277,14 @@ impl BitcoinIndexerConfig {
                 ]);
 
                 let cfg = BitcoinIndexerConfig {
-                    peer_host: peer_host.to_string(),
+                    peer_host: peer_host,
                     peer_port: peer_port,
                     rpc_port: rpc_port,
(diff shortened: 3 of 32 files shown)


kantai (approved):

diwakergupta:
The `sampled-genesis` fails with:

```
#13 169.9 ---- tests::neon_integrations::microblock_integration_test stdout ----
#13 169.9 thread 'main' panicked at 'assertion failed: `(left == right)`
#13 169.9   left: `[88, 53]`,
#13 169.9  right: `[88, 52]`', testnet/stacks-node/src/tests/neon_integrations.rs:68:5
#13 169.9 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

That check seems related to magic-bytes:
```rust
    assert_eq!(magic_bytes.as_bytes(), &['X' as u8, '4' as u8]);
```

The `full-genesis` test fails with:
```
#12 706.7 ---- tests::neon_integrations::bitcoind_integration_test stdout ----
#12 706.7 thread 'main' panicked at 'assertion failed: `(left == right)`
#12 706.7   left: `[88, 53]`,
#12 706.7  right: `[88, 52]`', testnet/stacks-node/src/tests/neon_integrations.rs:68:5
#12 706.7 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Which is basically the same as above. Seems unrelated to this PR, does the check need to be updated to use `X5` instead of `X4` @kantai ?

```rust
pub fn xenon() -> ConfigFile {
        let burnchain = BurnchainConfigFile {
            mode: Some("xenon".to_string()),
            rpc_port: Some(18332),
            peer_port: Some(18333),
            peer_host: Some("bitcoind.xenon.blockstack.org".to_string()),
            magic_bytes: Some("X5".into()),
```

So unless there are any objections, I plan to merge this PR. I can update the integration test as a separate PR or fold it in this one.

jcnelson:
If it's just the magic bytes, then please fix the integration tests -- they're tiny fixes.

diwakergupta:
> If it's just the magic bytes, then please fix the integration tests -- they're tiny fixes.

Already did in my latest commit.

diwakergupta:
`full-genesis` passed. `sampled-genesis` failed with what looks like a real, but unrelated failure:

```
#14 197.5 thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Request, url: "http://127.0.0.1:26057/v2/transactions", source: TimedOut }', testnet/stacks-node/src/tests/neon_integrations.rs:230:10
```

I need one more approval to merge this @jcnelson @lgalabru :)

jcnelson (approved):
I think that last failing test case was fixed in my WAL PR.

PR #2377, opened 2021-01-22
Test: nearly full block

reedrosenbluth:
The aim of this PR is to test behavior when a block is approaching its `block-limit`, and is full of transactions that are expensive to analyze (in this case, mostly invoking `(contract-call? ...)`).

Additionally, it contains a benchmarking function for a similarly expensive contract.


jcnelson (approved):
LGTM!  Thanks for the new test @reedrosenbluth

kantai (approved):

PR #2383, opened 2021-01-26
Use RBF for recovering from fast blocks

lgalabru:
This PR is superseding  https://github.com/blockstack/stacks-blockchain/pull/2363 and addressing a few issues, with one refactoring;
- It allows a miner to recover from a "fast block", by RBF-ing submitted block commits trying to target a "flashed" block.
- It's preventing submission of multiple block commit / block. 
- Improving the RBF logic, by dynamically computing the size of the transactions (the values from config are still used but now only for initial estimation).

I was able to test the miner against a puppet chain and see the RBF logic successfully kicking in with multiple consecutive flash blocks (3 consecutive RBF, 1 final mined transaction), I should also be able to write some unit tests.

How to test:
1) start bitcoind
```bash
$ cat helium.conf
chain=regtest
disablewallet=0
txindex=1
server=1
rpcuser=helium-node
rpcpassword=secret%

$ bitcoind -conf=helium.conf
```

2) start puppet chain
```bash
$ cd testnet/puppet-chain
$ DYNAMIC_GENESIS_TIMESTAMP=1 cargo run local-leader.toml.default
```

3) start a stacks-node

in `core/mod.rs` update the constants 
```rust
pub const BITCOIN_REGTEST_FIRST_BLOCK_HEIGHT: u64 = 300;
pub const BITCOIN_REGTEST_FIRST_BLOCK_TIMESTAMP: u32 = 1611158538;
pub const BITCOIN_REGTEST_FIRST_BLOCK_HASH: &str =
    "47922cbadd8158ea999b6be9a31dd69daf58b33dd5be2d3d2d528274d5dd8ea2";
```

I'd also recommend enforce usage of a sampled genesis.

```bash
$ cd testnet/stacks-node
$ STACKS_LOG_PP=1 cargo run start --config=./conf/local-leader-conf.toml
```

The puppet chain will basically be messing around with bitcoind between block 307 and block 312, generating flash blocks and buffering transactions coming from the stacks-node.
When everything goes back to normal and inspecting block 313, we can see that all the block commit that happened during 307 and 312 have been RBF, and we only have one clean block commit.


testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -787,49 +911,137 @@ impl BitcoinRegtestController {
...
+        //  ii) If UTXOs initially used are sufficient for paying for a fee bump, then RBF
+
+        // Let's start by early returning 1)
+        if payload == ongoing_op.payload {
+            debug!("Abort attempt to re-submit identical LeaderBlockCommit");
+            return None;
+        }
+
+        // Let's proceed and early return 2) i)
+        let res = if ongoing_op.fees.estimated_amount_required() > ongoing_op.sum_utxos() {
```
kantai, line 1003: This case occurs if the `ongoing_op`'s fees are already too large to increase and replace the transaction via RBF?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -787,49 +911,137 @@ impl BitcoinRegtestController {
...
+        let ongoing_op = self.ongoing_block_commit.take().unwrap();
+
+        let _ = self.sortdb_mut();
+        let mined_op = self
+            .burnchain_db
+            .as_ref()
+            .expect("BurnchainDB not opened")
+            .get_burnchain_op(&ongoing_op.txid);
+
+        if mined_op.is_some() {
```
kantai, line 979: What happens in the following scenario:

1. `outgoing_op` is mined in Block N
2. `build_leader_block_commit_tx()` is called
3. burnchain_db receives Block N

Basically, if the `outgoing_op` is confirmed and this method is called _before_ the `burnchain_db` knows about the confirmation. It seems like this method would try to RBF that transaction, correct? And then, one of two things will happen: (1) the bitcoind instance communicated with would already know that the prior op was confirmed and there'd be an error, or (2) the bitcoind instance hasn't received the new block and accepts the RBF. In this case, the outgoing transaction would just be "wrong" -- the controller would be tracking a transaction that is going to be dropped by the network (which is possible no matter what with RBF). I think to remedy this, the `outgoing_op` struct may need to track multiple TXIDs?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: Naming is probably not optimal - `estimated_amount_required` is returning the estimated amount required for performing a RBF (so ongoing op + fee_rate increase + bandwidth fee).

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: I could be missing a subtlety, but I think that the case you mentioned is covered by the next lines, ensuring "idempotency":
```rust
        if payload == ongoing_op.payload {
            debug!("Abort attempt to re-submit identical LeaderBlockCommit");
            return None;
        }
```

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -787,50 +917,169 @@ impl BitcoinRegtestController {
...
+                debug!("Was able to retrieve ongoing TXID - {}", txid);
+                let res =
+                    self.send_block_commit_operation(payload, signer, None, None, None, &vec![]);
+                return res;
+            } else {
+                debug!("Was unable to retrieve ongoing TXID - {}", txid);
+            }
+        }
+
+        // Stop as soon as the fee_rate is 1.25 higher, stop RBF
```
jcnelson, line 999: Shouldn't this be a config option?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -847,26 +1096,68 @@ impl BitcoinRegtestController {
...
+        } else {
+            spent_in_rbf + tx_size // we're spending 1 sat / byte in RBF
+        };
+        self.serialize_tx(
+            tx,
+            spent_in_outputs + tx_size * fee_rate + rbf_fee,
+            utxos_set,
+            signer,
+        );
+        signer.dispose();
```
jcnelson, line 1141: Shouldn't this be called on a `impl Drop` for `BurnchainOpSigner`?  If not, can you make it so?

testnet/stacks-node/src/config.rs
```diff
@@ -18,6 +18,7 @@ use stacks::vm::costs::ExecutionCost;
 use stacks::vm::types::{AssetIdentifier, PrincipalData, QualifiedContractIdentifier};
 
 const DEFAULT_SATS_PER_VB: u64 = 50;
+const DEFAULT_RBF_FEE_RATE_INCREMENT: u64 = 5;
```
jcnelson, line 21: Shouldn't this be 1?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -59,9 +63,119 @@ pub struct BitcoinRegtestController {
...
+}
+
+impl OngoingBlockCommit {
+    fn sum_utxos(&self) -> u64 {
+        self.utxos.total_available()
+    }
+}
+
+#[derive(Clone)]
+struct LeaderBlockCommitFees {
```
jcnelson, line 83: Why are any of these fields important to track here?  The _tenure processor_ decides how much to commit and to burn.  This code should only be concerned with the Bitcoin transaction fees (the fee for the whole transaction, and any state required to replace-by-fee).

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -59,9 +63,119 @@ pub struct BitcoinRegtestController {
...
+    ) -> LeaderBlockCommitFees {
+        let mut fees = LeaderBlockCommitFees::estimated_fees_from_payload(payload, config);
+        fees.spent_in_attempts = cmp::max(1, self.spent_in_attempts);
+        fees.final_size = self.final_size;
+        fees.fee_rate = self.fee_rate + config.burnchain.rbf_fee_increment;
+        fees.is_rbf_enabled = true;
+        fees
+    }
+
+    pub fn estimated_fees_from_payload(
```
jcnelson, line 108: So, back in the world of Stacks 1.0 and `blockstack.js`, we spent the better part of a year trying to make fee estimation code like this work.  It ultimately failed -- no matter how refined our estimation heuristics were, they were never good enough to avoid corner cases where transactions didn't have enough fees to be mined.  We eventually scrapped them all and rewrite the code to calculate the transaction fee by _actually generating and signing_ the full transaction to get the payload length, and using _that_ to get the fee.

By far and away the biggest reasons for _not_ having an estimator like this are that you (1) cannot predict what kinds of UTXOs the transaction is going to consume, and (2) cannot predict how many UTXOs you need to consume to cover your fee.  If you get these wrong, then your transaction may not get mined in the next block, or even relayed at all.  The _only way_ you can reliably estimate your transaction fee is to try and assemble it from enough UTXOs that you can meet the targeted fee rate.

So, I think you should rewrite this method and estimate transaction fees only by actually generating them.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -638,18 +757,21 @@ impl BitcoinRegtestController {
...
                 },
-                vec![utxo],
+                UTXOSet {
+                    bhh: BurnchainHeaderHash::zero(),
+                    utxos: vec![utxo],
+                },
             )
         } else {
-            self.prepare_tx(&public_key, DUST_UTXO_LIMIT, 1)?
+            self.prepare_tx(&public_key, 3 * DUST_UTXO_LIMIT, None, None, 0)?
```
jcnelson, line 774: Where does the `3 * DUST_UTXO_LIMIT` constant come from?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -671,7 +793,15 @@ impl BitcoinRegtestController {
         tx.output
             .push(payload.recipient.to_bitcoin_tx_out(DUST_UTXO_LIMIT));
 
-        self.finalize_tx(&mut tx, DUST_UTXO_LIMIT, DUST_UTXO_LIMIT, utxos, signer, 1)?;
+        self.finalize_tx(
+            &mut tx,
+            DUST_UTXO_LIMIT,
+            0,
+            230,
```
jcnelson, line 800: What's `230` here?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -721,7 +851,15 @@ impl BitcoinRegtestController {
         tx.output = vec![consensus_output];
         tx.output.push(payload.output.to_bitcoin_tx_out(output_amt));
 
-        self.finalize_tx(&mut tx, output_amt, DUST_UTXO_LIMIT, utxos, signer, 1)?;
+        self.finalize_tx(
+            &mut tx,
+            output_amt,
+            0,
+            280,
```
jcnelson, line 858: What's `280` here?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -700,8 +830,8 @@ impl BitcoinRegtestController {
     ) -> Option<Transaction> {
         let public_key = signer.get_public_key();
 
-        let output_amt = 2 * DUST_UTXO_LIMIT;
-        let (mut tx, utxos) = self.prepare_tx(&public_key, output_amt, 1)?;
+        let output_amt = 4 * DUST_UTXO_LIMIT;
```
jcnelson, line 833: Why go from `2` to `4`?  Also, why is `output_amt` necessarily `4 * DUST_UTXO_LIMIT`?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -787,50 +917,169 @@ impl BitcoinRegtestController {
...
             signer,
-            attempt,
         )?;
 
-        increment_btc_ops_sent_counter();
+        let serialized_tx = SerializedTx::new(tx.clone());
+
+        let tx_size = serialized_tx.bytes.len() as u64;
+        estimated_fees.register_replacement(tx_size);
+        let mut txid = tx.txid().as_bytes().to_vec();
```
jcnelson, line 938: Might want to consider adding a constructor to our `Txid` that takes a `rust-bitcoin` txid value and does this low-level byte manipulation.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -787,50 +917,169 @@ impl BitcoinRegtestController {
...
+        }
+
+        // Stop as soon as the fee_rate is 1.25 higher, stop RBF
+        if ongoing_op.fees.fee_rate > (self.config.burnchain.satoshis_per_byte * 150 / 100) {
+            let res = self.send_block_commit_operation(payload, signer, None, None, None, &vec![]);
+            return res;
+        }
+
+        // Did a re-org occurred since we fetched our UTXOs
+        if let Err(e) = burnchain_db.get_burnchain_block(&ongoing_op.utxos.bhh) {
```
jcnelson, line 1006: I don't think this is ever going to fail.  We don't delete data from `BurnchainDB`, so a sibling Bitcoin block's block hash (and all its associated ops) would continue to be accessible through `get_burnchain_block`.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -787,50 +917,169 @@ impl BitcoinRegtestController {
...
+        //  ii) If UTXOs initially used are sufficient for paying for a fee bump, then RBF
+
+        // Let's start by early returning 1)
+        if payload == ongoing_op.payload {
+            info!("Abort attempt to re-submit identical LeaderBlockCommit");
+            return None;
+        }
+
+        // Let's proceed and early return 2) i)
+        let res = if ongoing_op.fees.estimated_amount_required() > ongoing_op.sum_utxos() {
```
jcnelson, line 1030: I think the code should try to avoid this case if at all possible.  Sending two block-commits ought to be strictly more expensive than sending one, and only one of these two block-commits is going to win sortition.  Maybe the code should instead attempt to figure out the total set of UTXOs to use in advance, given a maximum transaction fee, and then only try to RBF?

jcnelson: Also, I think the code should stop attempting to send block-commits altogether if the Bitcoin block height advances, since the block-commit will almost certainly become invalid.  Is this handled anywhere?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -1435,7 +1777,7 @@ impl BitcoinRPCRequest {
             }
         };
 
-        Ok(vec![])
+        Ok(UTXOSet { bhh, utxos })
     }
 
     pub fn send_raw_transaction(config: &Config, tx: String) -> RPCResult<()> {
```
jcnelson, line 1783: I think we're at the point where the transaction fee estimation logic is complex enough that it should probably have its own unit tests.  Namely, the code should have tests to ensure that:

* The RBF fee rate never decreases over successive replacements, and always increases by at least the minimal amount.
* The UTXO-consumption code can handle the case where UTXOs get invalidated by one or more arriving Bitcoin blocks (a "flash block").  It should stop attempting to send the block-commit, even if it's still pending in the mempool.
* The UTXO-consumption code can handle the case where UTXOs get invalidated by intermittent transactions.  For example, suppose my sends a block-commit, then spend its change output with another wallet outside the node.  The node should detect this and re-calculate its UTXO set.
* The fee estimation logic should strictly be based on the length of the generated transaction, including all the UTXOs it includes.
* The code should support an overall max transaction fee a block-commit can have.  The code should pick its UTXOs accordingly, so that it can always RBF its pending block-commits up to this maximum fee.  I don't think the code should attempt to send multiple, different block-commits for the same sortition.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: yeah this is exactly the approach taken in this PR - see line 1111.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: Not sure to understand the case you're challenging. `2) i)`?
> Sending two block-commits ought to be strictly more expensive than sending one, and only one of these two block-commits is going to win sortition.

If we're entering this codepath, it means that you have an ongoing transaction that was not mined in time, and the UTXOs used are too small for allowing a replacement by fee.
So the miner will indeed end up with 2 block commits in a block, but one of them will be outdated and won't be able to win the sortition.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: I see. I added this condition to have a green light from the test `bitcoind_forking_test`. Do you think we should leave that condition, or rewrite that test?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: `230` is the first estimated size we're using for the bitcoin `StackStx`  operations (it's a `#[cfg(test)]` method and used once, but I can use a constant if you prefer)

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: `280` is the first estimated size we're using for the bitcoin `TransferStx`  operations (it's a `#[cfg(test)]` method and used once, but I can use a constant if you prefer)

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: Is the concern being about having

```rust
struct Fee {
   fee_1: u64,
   fee_2: u64,
   fee_3: u64,
} 

impl Fee {
   fn get_total(&self) -> u64 {
       fee_1 + fee_2 + fee_3 
   }
}
```

vs 

```rust
struct Fee {
   total: u64,
} 

impl Fee {
   fn new(fee_1: u64, fee_2: u64, fee_3: u64) -> Fee {
       Fee {
           total: fee_1 + fee_2 + fee_3
       }
   }
}
```
?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: Refactored in https://github.com/blockstack/stacks-blockchain/pull/2383/commits/4d576c9ffe82b31c8efe4b4659f30b6fa7e576cd

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: Refactored in https://github.com/blockstack/stacks-blockchain/pull/2383/commits/4d576c9ffe82b31c8efe4b4659f30b6fa7e576cd

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: Refactored in https://github.com/blockstack/stacks-blockchain/pull/2383/commits/4d576c9ffe82b31c8efe4b4659f30b6fa7e576cd

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: Refactored in https://github.com/blockstack/stacks-blockchain/pull/2383/commits/4d576c9ffe82b31c8efe4b4659f30b6fa7e576cd

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
jcnelson: The concern is about why `sunset_fee`, `outputs_len`, and `sortition_fee` are even listed, as well as `amount_per_output()`.  The code shouldn't care about how the block-commit was structured, nor should it be trying to (re-)create the block-commit's outputs.  The purpose of this module is to provide a caller a way to construct a Bitcoin transaction, and then to attempt to send it.  It shouldn't be surreptitiously re-building it, because it can't know the caller's intentions.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
jcnelson: The code uses this method as a starting estimation (line 100 and 892) before it tries to estimate the fee from the serialized transaction body.  This is sub-optimal -- the code should calculate the required fee only from a "real" transaction (for the reasons listed above).  It's a little tricker in this case, because the code also needs to budget for a maximum RBF fee bump, so it needs to pick the transaction's UTXOs accordingly before it does any estimations.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
jcnelson: I think the error path here can arise and is definitely worth handling, but I think the check you want here is to verify whether or not `ongoing_op.utxos.bhh` is still on the canonical burnchain fork.  You can tell by either looking at the SPV headers, or by looking at the sortition DB.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
jcnelson: > it means that you have an ongoing transaction that was not mined in time

If this transaction has not been mined in time -- i.e. the Bitcoin blockchain advanced by one block -- then there's nothing more that this code can do.  The node needs to go and generate another Stacks block off of the new chain tip and re-try its bock-commit.

I'm saying that we should try to avoid the case where the transaction can no longer be RBF'ed because it's out of UTXOs to spend.  I think the node should instead have an "RBF budget" whereby it calculates the set of UTXOs required to pay for not only the initial transaction (with a typical fee), but also some additional number of RBFs that may be attempted in this Bitcoin block.  This way, the transaction would start out with enough UTXOs to pay for all of its RBF attempts up-front, and the node would simply decrement and re-sign the transaction's change address on each RBF.  If the Bitcoin block takes an unusually long time to be mined, and the node's RBF budget gets exceeded, then it should just give up attempting to win the block race for this block.  If it instead tried to send a second block-commit, it's forcing the miner to basically pay double for the chance to win the same block (since at least one of its in-flight block-commits is _guaranteed_ to fail).

kantai (approved):

jcnelson (approved):

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: We are clearly missing an abstraction, before this iteration, the code used to look like this:

```rust
fn build_leader_block_commit_tx(
        ...
        let number_of_transfers = payload.commit_outs.len() as u64;		         
        let value_per_transfer = payload.burn_fee / number_of_transfers;
        ...
```

It should not be up to the `build_leader_block_commit_tx` method to manipulate payload's fields. We should either be associating some additional methods to `payload`, or abstracting this somehow. This PR is introducing a new struct for centralizing the leader commit operation costs structure needed for RBF and it felt natural to add some helpers in there.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2383/commits/b03495b2c652bc1cfb084e089d590fc7e04e1446

PR #2386, opened 2021-01-28
Upload blocks via HTTP endpoint

jcnelson:
This PR adds an HTTP endpoint `POST /v2/blocks/upload/{consensus-hash}` that lets you upload Stacks blocks to a Stacks node via HTTP, instead of the P2P network.  If the node didn't have the block, it takes care of propagating it to its neighbors in the usual manner.  If a node already had the block, it takes no action.  Either way, it informs the client by replying with `{"stacks_block_id": "$stacks_block_id_hash, "accepted": $true_or_false}`.  The `accepted` field will be `true` if the block was new, and `false` if the node has already seen it.

I was originally going to include this with #1805, but I think we might want it available sooner.  It's meant to help a node recover if it's somehow missing blocks that we have locally (as we've seen with Xenon recently).


jcnelson:
Don't know how this dup was created, but closing

PR #2387, opened 2021-01-28
Upload blocks via HTTP endpoint

jcnelson:
This PR adds an HTTP endpoint `POST /v2/blocks/upload/{consensus-hash}` that lets you upload Stacks blocks to a Stacks node via HTTP, instead of the P2P network.  If the node recognizes the consensus hash but didn't have the block, it takes care of propagating it to its neighbors in the usual manner.  If a node already had the block, it takes no action.  Either way, it informs the client by replying with `{"stacks_block_id": "$stacks_block_id_hash, "accepted": $true_or_false}`.  The `accepted` field will be `true` if the block was new, and `false` if the node has already seen it.

I was originally going to include this with #1805, but I think we might want it available sooner.  It's meant to help a node recover if it's somehow missing blocks that we have locally (as we've seen with Xenon recently).

This also might be relevant to #2196.


kantai (commented):
The `net` changes look good to me, but I think the changes in `sortdb` should either be reverted from this PR, or have tests included -- the changes there alter the Sortition DB's behavior: conditions that used to cause panics seem to now result in `continue`s. The new endpoint should also be documented in `rpc-endpoints.md`

jcnelson:
> The net changes look good to me, but I think the changes in sortdb should either be reverted from this PR, or have tests included -- the changes there alter the Sortition DB's behavior: conditions that used to cause panics seem to now result in continues.

I reverted, and simplified to a one-liner (and did a little refactoring -- deleted some dead code and moved some MARF key/value formatters to `db_keys`).  The bug is really in `get_snapshot_by_arrival_index()` -- it shouldn't be returning `BlockSnapshot`s that aren't PoX-valid.  This would have been the case prior to adding PoX -- Stacks blocks that correspond to sortitions that aren't on the PoX fork haven't "arrived" since they no longer exist.

kantai (approved):

src/chainstate/burn/db/sortdb.rs
```diff
@@ -2862,7 +2854,7 @@ impl SortitionDB {
         conn: &Connection,
         consensus_hash: &ConsensusHash,
     ) -> Result<Option<SortitionId>, db_error> {
-        let qry = "SELECT sortition_id FROM snapshots WHERE consensus_hash = ?1";
+        let qry = "SELECT sortition_id FROM snapshots WHERE consensus_hash = ?1 AND pox_valid = 1 LIMIT 1";
```
lgalabru, line 2857: Are we sure that this changes are not breaking consensus?
It would impact the vectors returned by `process_new_block_arrivals`, which are then written to a MARF?

jcnelson: The sortition DB's MARF isn't part of the consensus code.

jcnelson: Also, block arrivals are going to be different for each node already, since nodes don't all download and process blocks in the same order.

lgalabru (approved):
LGTM

src/chainstate/burn/db/sortdb.rs
lgalabru: I see. So what are the properties that we're getting out of the MARF? Why not using a Hashmap?

jcnelson: We need to keep track of which Stacks blocks have arrived on which burnchain forks in order to determine which Stacks fork is the longest Stacks fork on each burnchain fork's PoX-valid sortitions.  The canonical Stacks fork is not the longest Stacks fork anywhere; it's the longest Stacks fork on the PoX-valid sortitions on the longest burnchain.

PR #2388, opened 2021-01-28
Feat: Mock mining mode

kantai:
This PR allows a miner to continue operating without UTXOs (and therefore never committing). This is useful for tracking how a miner _would_ behave in the network (i.e., what transactions it would try to include, and what transactions would fail to be included).

This is now ready for review -- the "mock mining" behavior is now configurable by setting `mock_mining = True` in the `[node]` config.


jcnelson (approved):
LGTM!

PR #2390, opened 2021-01-29
Fix: atlas attachment serving + broken test

lgalabru:
After investigation, it turns out the atlas test was broken for a good reason.
- Attachment serving broke with a recent refactoring
- The BNS pricing table changed and the value provided was too low, as a consequence the `namespace-preorder` contract call was failing.

### DIFF
--- a/.github/workflows/bitcoin-tests.yml
+++ b/.github/workflows/bitcoin-tests.yml
@@ -16,9 +16,6 @@ jobs:
           DOCKER_BUILDKIT: 1
         run: docker build -f ./.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests .
   atlas-test:
-    # disable this job/test for now, since we haven't seen this pass
-    #  on github actions in a while
-    if: ${{ false }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


src/net/atlas/db.rs
```diff
@@ -387,7 +387,7 @@ impl AtlasDB {
         content_hash: &Hash160,
     ) -> Result<Option<Attachment>, db_error> {
         let hex_content_hash = to_hex(&content_hash.0[..]);
-        let qry = "SELECT content, hash FROM attachments WHERE hash = ?1 AND was_instantiated = 0"
+        let qry = "SELECT content, hash FROM attachments WHERE hash = ?1 AND was_instantiated = 1"
```
jcnelson, line 390: There haven't been any attachments propagated to the mainnet network yet, right?  i.e. We're certain this change isn't going to break anything?

jcnelson (approved):
LGTM; just had one question I need some clarity on first.

src/net/atlas/db.rs
lgalabru: The BNS endpoints + registrar not being live at this point, the only interactions that happened with the contract were internal tests (https://explorer.stacks.co/txid/SP000000000000000000002Q6VF78.bns?chain=mainnet)

kantai (approved):

.github/workflows/bitcoin-tests.yml
```diff
@@ -16,9 +16,6 @@ jobs:
           DOCKER_BUILDKIT: 1
         run: docker build -f ./.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests .
   atlas-test:
-    # disable this job/test for now, since we haven't seen this pass
-    #  on github actions in a while
-    if: ${{ false }}
```
kantai, line 21: 🥳

Issue #2389, opened 2021-01-28
Miner implementation does not correctly walk mempool


kantai:
In order to "walk" the mempool and find transactions for inclusion in a block, a miner does the following (starting at block height _n_)

1.  Asks the mempool DB for all `(consensus_hash, block_hash)` pairs at height _n_
2. Figures out the `consensus_hash,  block_hash` pair for the current chain's view of height _n_
3. Finds the matching pair between 1 and 2, and then fetches all transactions accepted into the mempool at that block, if there's no such matching pair, stop walking
4. Attempts to include transactions from 3 in the block
5. Repeat from 1 at height _n-1_ until out of transactions or height = 0

Steps 1->2->3 have a (breaking) logic bug: if the mempool db doesn't have any entries for a block's ancestor, then the walk stops. This, however, is entirely possible, and would be expected to occur 50% of the time if there is a short-lived fork and this has happened a few times over the course of the Stacks chain so far.

For example, at block 1631:

In step (1), the mempool will only ever have entries for a given block height on the fork it followed (in 1631's case, the a node believed `51d33db6d82374122349314a1c582909dca473ce/adfa719db9a3ccb47b21e55a8a7f60d0bac4a51d9a266851ea8c2a41df49ae70` to be the chain tip).

But, 1632 is mined by someone with a different height 1631 block. 1632's parent block is `3152453237a1a030320b168d87555c18a6942893/8819f44ab05b1130cd053052400bcaaa64a27a58dbc55db600c8dbe878ae5428`

Once this occurs, any node that had a different 1631 height view will never consider for inclusion a transaction that they first received earlier than that block height.

Fixing this is _not_ consensus-breaking, but would require miners themselves to upgrade.

kantai:
In the `fix/mempool-walks`, you can experiment with a local chain state to see this activity in action (you'll need a `develop` chain state, because the database schema changed):

```
$ rm -r /tmp/stacks-node.clone/ && cp -r /tmp/stacks-node /tmp/stacks-node.clone && BAD_BEHAVIOR=1 ./target/debug/blockstack-core try-mine /tmp/stacks-node.clone
INFO [1611933620.471113] [src/chainstate/stacks/miner.rs:671] [main] Include tx 585345aa581299bd674040ace226d2ed81d4cf153888fc6b6fb91bfb790c54d7 (Coinbase) in anchor block
[...]

PR #2392, opened 2021-01-29
Fix: #2389 Mempool walks, add `try-mine` CLI command

kantai:
This PR addresses the most immediate mempool issue #2389, by modifying the walk logic to "try again" if there are mempool entries received at a particular height _but not_ in the same fork.

This PR also adds a `try-mine` CLI command, as well as a diagnostic envar `MEMPOOL_BAD_BEHAVIOR` which executes the mempool walk in the manner that causes #2389.

This leaves unaddressed two potential issues:

1. Transactions received into the mempool at block `A`, but block `A` is "forked away". Those transactions will not appear in the miner's walk. However, such transactions will be viable for "replace-across-forks", such that if the transaction is rebroadcasted (not re-signed or fee upgraded), the miner will replace the existing tx in the mempool. (Issue: #2394)
2. Mempool transactions are evaluated in "reverse" order: transactions received later are attempted first. This can impact chained transactions, because higher nonces will be tried first and fail, before the lower nonces are tried. In degenerate cases, this can lead to chained transactions only being included one-per-block. (Issue: #2393)

I think #2389 is much worse than those two issues at the moment, so I'm not sure if we should try to address those issues at the same time, or instead split it into multiple releases.

### DIFF
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -1113,6 +1113,15 @@ mod tests {
         C32_ADDRESS_VERSION_TESTNET_SINGLESIG,
     };
 
+    use crate::{
+        burnchains::BurnchainHeaderHash,
+        chainstate::stacks::{
+            db::StacksHeaderInfo, index::TrieHash, StacksBlockId, StacksWorkScore,
+        },
+        util::vrf::VRFProof,
+        vm::costs::ExecutionCost,
+    };
+
     use super::MemPoolDB;
     use util::db::{DBConn, FromRow};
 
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -188,6 +197,92 @@ fn main() {
...
+            TransactionPayload::Coinbase(CoinbasePayload([0u8; 32])),
+        );
+
+        coinbase_tx.chain_id = chain_id;
+        coinbase_tx.anchor_mode = TransactionAnchorMode::OnChainOnly;
+        let mut tx_signer = StacksTransactionSigner::new(&coinbase_tx);
+        tx_signer.sign_origin(&sk).unwrap();
+        let coinbase_tx = tx_signer.get_tx().unwrap();
+
+        let mainnet_block_limit = ExecutionCost {
```
jcnelson, line 254: Isn't this a constant somewhere?

jcnelson (approved):
LGTM, thanks for tackling this!  Just one minor comment.

src/main.rs
kantai: Yeah, in `stacks-node::config` unfortunately -- I can refactor it to be `core::` constant, though.

src/core/mempool.rs
```diff
@@ -548,26 +577,42 @@ impl MemPoolDB {
...
+                            break next_timestamp;
                         }
-                        None => {
-                            // no more transactions
+                        MemPoolWalkResult::NoneAtHeight(
+                            next_consensus_hash,
+                            next_block_hash,
+                            next_height,
+                        ) => {
+                            if std::env::var("MEMPOOL_BAD_BEHAVIOR") == Ok("1".into()) {
```
lgalabru, line 599: Should we prefix env var with `STACKS_`?

testnet/stacks-node/src/config.rs
```diff
@@ -7,7 +7,8 @@ use rand::RngCore;
 use stacks::burnchains::bitcoin::BitcoinNetworkType;
 use stacks::burnchains::{MagicBytes, BLOCKSTACK_MAGIC_MAINNET};
 use stacks::core::{
-    CHAIN_ID_MAINNET, CHAIN_ID_TESTNET, PEER_VERSION_MAINNET, PEER_VERSION_TESTNET,
+    CHAIN_ID_MAINNET, CHAIN_ID_TESTNET, MAINNET_BLOCK_LIMIT, PEER_VERSION_MAINNET,
```
lgalabru, line 10: 💅 Should `MAINNET_BLOCK_LIMIT` be `BLOCK_LIMIT_MAINNET`?

lgalabru (approved):

Issue #2405, opened 2021-02-03
Performance: `PeerNetwork::can_download_microblock_stream` is prohibitively slow


jcnelson:
The downloader in `develop` will make every effort to _not_ re-download the same data twice.  For microblock streams, this means verifying that (a) the parent anchored block exists, (b) the child anchored block exists, and (c) each microblock in-between them exists.  Part (c) is expensive -- it's one SQL query _per microblock_.

We can omit (c) in the happy path by simply checking that the child in (b) is processed and not orphaned, since this proves that the microblocks were present (or will never need to be fetched for this child).  We can omit (c) in the unhappy path by having the downloader remember in RAM that it did the expensive check already for the child in (b), and that this stream is already known to be present for that child block.

kantai:
PR #2411 speeds up this process (at least on SSD, I'd guess even more so on spinning disk) in two ways:

1. A lot of the overhead of the downloader was being spent fetching the block availability, which itself was suffering from the problems described in #2406, so just apply the wins from addressing that.
2. Add a sqlite index in the sortition db's snapshots table on the winning block hashes, because the downloader queries that for each block -- in profiling, that was the majority of the sqlite cost (i.e., unindexed scans). The remaining time in the tests appears to actually be IO from reading in the stacks headers.

We should see if just those changes address the performance issues in the downloader enough to avoid modifying any of the logic of the checks.

jcnelson:
Closed by #2411

PR #2411, opened 2021-02-03
Performance Fixes: get-blocks-inventory (#2406), can-download-microblock (#2405)

### LINKED ISSUE #2406: Performance: `ConversationP2P::handle_getblocksinv_response()` is prohibitively slow
I've noticed on my (still-bootstrapping) `develop` follower node that the `ConversationP2P::handle_getblocksinv_response()` method takes a whole 3 seconds (!!) to run, and that's with apparently zero MARF reads (100% block hash cache hits).  This is unacceptable, and should be fixed before release.

kantai:
This address #2406 by doing three things:

1. Use `parent_sortition_id` while back-walking the sortition db in `get_stacks_header_hashes`
2. Add sqlite index for `index_block_hash` for checking if a microblock is processed.
3. Fetch just `parent_microblock_hash` from the staging blocks db when testing whether or not a microblock parent has been processed.

The performance improvement is about 4x on an SSD in a release build, which can be tested with the new `get-block-inventory` CLI tool:

Before:
```
$ time ./target/release/blockstack-core get-block-inventory /tmp/stacks-node/
Fetched header hashes in 0.65074944
Fetched block inv in 0.65503883
Done!

real	0m1.610s
user	0m0.994s
sys	0m0.146s
```

With this patch:
```
$ time ./target/release/blockstack-core get-block-inventory /tmp/stacks-node/
Fetched header hashes in 0.092927
Fetched block inv in 0.24245426
Done!

real	0m0.427s
user	0m0.288s
sys	0m0.034s
```

This addresses #2405 by doing two things:
1. Using the performance gains from above in while obtaining the get_stacks_header_hashes (this is invoked in `get_block_availability`) -- not sure how much this impacts the performance of #2405 given the use of a header hash.
2. Adding an index in the sortition db for `winning_stacks_block_hash`, this is used by `load_parent_block_header`, and seems to be contributing more to the runtime of the downloader than the microblock fork check.

Before:
```
$ time ./target/release/blockstack-core can-download-microblock /tmp/stacks-node/
INFO [1612376077.127826] [src/net/db.rs:165] [main] Will be authenticating p2p messages with public key: 02c939aa049626470bfc67516127016a9630192504d52bdeab9a199d0d56a85c2c
Fetched header hashes in 0.108790584
Checked can_download in 1.159317
Done!

real	0m1.690s
user	0m0.881s
sys	0m0.151s

# this exhibits a lot of 'disk cache warmup' performance speedup -- probably because of the
# sqlite scanning happening 
$ time ./target/release/blockstack-core can-download-microblock /tmp/stacks-node/
INFO [1612376080.593005] [src/net/db.rs:165] [main] Will be authenticating p2p messages with public key: 03ae0387635db3eb61158693b2daa041662fd8adb234e879c9d68977a5e5aea27a
Fetched header hashes in 0.11902666
Checked can_download in 0.84514356
Done!

real	0m0.980s
user	0m0.858s
sys	0m0.120s
```

After:
```
$ time ./target/release/blockstack-core can-download-microblock /tmp/stacks-node/
INFO [1612376094.075604] [src/net/db.rs:165] [main] Will be authenticating p2p messages with public key: 0292cf97a7a160ae13a3c67e96df3b5e7adbc1f62794018e819fa87da7a665acf1
Fetched header hashes in 0.11809304
Checked can_download in 0.5653221
Done!

real	0m0.699s
user	0m0.580s
sys	0m0.117s
```

This is a speedup of about `0.8s` compared to the prior version after warmup (including the speedup from the changes to `get_stacks_header_hashes`)


jcnelson:
I can confirm that the performance of both of these methods is better!  But, it needs to be about 10x faster than it is even in this PR.  Here's what I'm seeing on my miner:

get-blocks-inventory:
```
Fetched header hashes in 0.17645259
Fetched block inv in 0.71278244
Done!
Fetched header hashes in 0.17876537
Fetched block inv in 0.77341956
Done!
Fetched header hashes in 0.17281035
Fetched block inv in 0.7165203
Done!
Fetched header hashes in 0.17799063
Fetched block inv in 0.72357416
Done!
Fetched header hashes in 0.1738727
Fetched block inv in 0.7506575
Done!
```

can-download-microblock
```
Fetched header hashes in 0.18426022
Checked can_download in 1.9242303
Done!
Fetched header hashes in 0.17765793
Checked can_download in 1.8006527
Done!
Fetched header hashes in 0.17775592
Checked can_download in 1.8218927
Done!
Fetched header hashes in 0.19738637
Checked can_download in 1.9065963
Done!
Fetched header hashes in 0.17920691
Checked can_download in 1.7992499
Done
```

I also tested both of these with _just_ adding the requisite indexes, and nearly all the performance gain is attributable to the indexes.

Replying to a blocks inventory happens synchronously in the p2p thread, as does determining whether or not to download a microblock stream.  So, each `GetBlocksInv` will stall the network for 0.7-0.8s, and when looking for things to download, the downloader can stall the network for 1.7s-2.0s.  Is it okay if I try and add a few things to this PR to see if I can speed it up some?

kantai:
> Replying to a blocks inventory happens synchronously in the p2p thread, as does determining whether or not to download a microblock stream. So, each GetBlocksInv will stall the network for 0.7-0.8s, and when looking for things to download, the downloader can stall the network for 1.7s-2.0s. Is it okay if I try and add a few things to this PR to see if I can speed it up some?

Works for me

jcnelson (approved):
So, in the interest of not blocking a release, I'm approving this PR.  But, please leave the two linked issues open, because I'll soon have some additional significant performance boosts.

lgalabru (approved):

PR #2412, opened 2021-02-03
Fix: serde_json depth limit

kantai:
This gets around `serde_json`'s default depth limit, using a `stacker` enabled deserializer.


jcnelson (approved):

lgalabru (approved):

kantai:
Merging -- the test flake in `sampled-genesis` is fixed in `develop`, but not on `master`, and these tests pass locally.

PR #2420, opened 2021-02-05
Indexes fixes + recoverable working_dir prep

lgalabru:
This PR is addressing https://github.com/blockstack/stacks-blockchain/issues/2417, by making sure that we're always executing one SQL statement per transaction.

Assuming that one of the next feature is the ability to re-use a working dir, this PR is also preparing the field, so that miners running this PR don't have to re-download all the blocks the next time they upgrade. In order to fully support this capability, the next version of the node will have to gracefully be able to handle DB migrations. 
In that regard, this PR is making sure that all the databases have a `db_config` table.
Any new db version will be a breaking version, so I simplified the version (`1` vs `23.0.0.0`) so that we don't have to write a custom parse for sorting migrations that would have been formatted as a semver.


src/burnchains/bitcoin/spv.rs
```diff
@@ -174,11 +177,16 @@ impl SpvClient {
     fn db_instantiate(conn: &mut DBConn) -> Result<(), btc_error> {
         let tx = tx_begin_immediate(conn)?;
 
-        for row_text in SPV_SQL {
+        for row_text in SPV_INITIAL_SCHEMA {
```
jcnelson, line 180: I think we should just migrate everything to use `execute_batch()` for DB instantiation.  Then this problem will never happen again, no matter how we structure the schema strings.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -470,22 +471,19 @@ const BURNDB_SETUP: &'static [&'static str] = &[
...
-    CREATE INDEX block_arrivals ON snapshots(arrival_index,burn_header_hash);
-    CREATE INDEX arrival_indexes ON snapshots(arrival_index);
-    "#,
+    "CREATE UNIQUE INDEX snapshots_block_hashes ON snapshots(block_height,index_root,winning_stacks_block_hash);",
+    "CREATE UNIQUE INDEX snapshots_block_stacks_hashes ON snapshots(num_sortitions,index_root,winning_stacks_block_hash);",
+    "CREATE UNIQUE INDEX snapshots_block_heights ON snapshots(burn_header_hash,block_height);",
+    "CREATE UNIQUE INDEX snapshots_burn_hashes ON snapshots(block_height,burn_header_hash);",
+    "CREATE INDEX snapshots_block_winning_hash ON snapshots(winning_stacks_block_hash);",
+    "CREATE INDEX block_arrivals ON snapshots(arrival_index,burn_header_hash);",
+    "CREATE INDEX arrival_indexes ON snapshots(arrival_index);",
```
jcnelson, line 480: Instead of turning each statement into its own string, why not use `execute_batch()`?

lgalabru: Yeah I don't have a strong opinion. Executing statement per statement is probably easier to debug?

lgalabru: This PR was about patching the issue, in my mind we need to revisit this strategy very soon with the ability of recovering a `working_dir`

lgalabru: Ideally, we execute the statements one by one, ensure that one statement contains only one statement, log the SQL statements failing, etc. This granularity would be hard to achieve with one big blob of SQL statements.

jcnelson: > Yeah I don't have a strong opinion. Executing statement per statement is probably easier to debug?

The problem was that adding additional statements into the same string was causing them to silently be ignored.  I think that the fix to this problem is to make it so that all statements get executed, no matter how they are represented in the code (either as separate strings, or grouped together into a single string).  You might argue that putting multiple statements into the same string is in and of itself a bug, but continuing with this path sets us up for future failure -- the compiler won't warn us when we do this, nor will the DB instantiation logic, and we'll have this problem again.

Also, the this PR's diff would be _smaller_ if we used `execute_batch()`.

src/burnchains/bitcoin/spv.rs
lgalabru: Well, we'll still have to debug bad SQL statements, like a missing `,`, `;` typos, etc.

jcnelson: That's fine.  The problem here isn't that we have to debug bad SQL statements; the problem is that SQL statements aren't being executed in the first place.

src/chainstate/burn/db/sortdb.rs
jcnelson: To be clear, I don't have an opinion on whether or not the PR should split the SQL schemas into one statement per string.  However, I have a very _strong_ opinion that we should use `execute_batch()`, since that would have prevented this problem in the first place.

lgalabru: Sounds good, addressed with https://github.com/blockstack/stacks-blockchain/pull/2420/commits/90b1bceb59313d54ce1a3a35a0361ce9f30ca448

src/burnchains/bitcoin/spv.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2420/commits/90b1bceb59313d54ce1a3a35a0361ce9f30ca448

src/chainstate/burn/db/sortdb.rs
```diff
@@ -470,22 +471,19 @@ const BURNDB_SETUP: &'static [&'static str] = &[
...
-    CREATE UNIQUE INDEX snapshots_block_hashes ON snapshots(block_height,index_root,winning_stacks_block_hash);
-    CREATE UNIQUE INDEX snapshots_block_stacks_hashes ON snapshots(num_sortitions,index_root,winning_stacks_block_hash);
-    CREATE UNIQUE INDEX snapshots_block_heights ON snapshots(burn_header_hash,block_height);
-    CREATE UNIQUE INDEX snapshots_burn_hashes ON snapshots(block_height,burn_header_hash)
-    CREATE INDEX snapshots_block_winning_hash ON snapshots(winning_stacks_block_hash);
-    CREATE INDEX block_arrivals ON snapshots(arrival_index,burn_header_hash);
-    CREATE INDEX arrival_indexes ON snapshots(arrival_index);
-    "#,
+    "CREATE UNIQUE INDEX snapshots_block_hashes ON snapshots(block_height,index_root,winning_stacks_block_hash);",
+    "CREATE UNIQUE INDEX snapshots_block_stacks_hashes ON snapshots(num_sortitions,index_root,winning_stacks_block_hash);",
```
kantai, line 475: I'm not sure that these tuples are unique in the presence of PoX -- if there were two PoX forks that both picked the same "winner" at the same height (I believe that is possible), then this constraint would be violated. I think that these indexes should drop the "unique" constraint.

Then, it seems like:

```
    "CREATE INDEX snapshots_block_heights ON snapshots(burn_header_hash,block_height);",
    "CREATE INDEX snapshots_burn_hashes ON snapshots(block_height,burn_header_hash);",
```

are both indexing more or less the same thing, so I'm pretty sure that one of them can be removed without affecting performance (and reducing the number of indexes managed by the db).

src/chainstate/stacks/db/mod.rs
```diff
@@ -489,9 +489,10 @@ const STACKS_CHAIN_STATE_SQL: &'static [&'static str] = &[
         block_size TEXT NOT NULL,       -- converted to/from u64
 
         PRIMARY KEY(consensus_hash,block_hash)
-    );
-    CREATE UNIQUE INDEX index_block_hash_to_primary_key(index_block_hash,consensus_hash,block_hash);
-    "#,
+    );"#,
+    "CREATE UNIQUE INDEX index_block_hash_to_primary_key ON block_headers(index_block_hash,consensus_hash,block_hash);",
```
kantai, line 493: We could safely drop the `UNIQUE` constraint here, because `(consensus_hash, block_hash)` is unique constrained and so is `index_block_hash`, and then this index is identical to `block_index_hash_index` below, so one of them should be dropped.

We _do_ query directly by `index_block_hash`, though, so it may be worthwhile to create an index on _just_ `index_block_hash`.

src/chainstate/burn/db/sortdb.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2420/commits/85822097eb7f13969a83f4d096df208b50896db7

src/chainstate/stacks/db/mod.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2420/commits/85822097eb7f13969a83f4d096df208b50896db7

src/chainstate/burn/db/sortdb.rs
jcnelson: Please restore this index:

```
 CREATE INDEX snapshots_block_winning_hash ON snapshots(winning_stacks_block_hash);
 ```

File: src/net/atlas/db.rs [hunk 5]
```diff
@@ -28,18 +28,17 @@ use net::StacksMessageCodec;
...
     CREATE TABLE attachments(
         hash TEXT UNIQUE PRIMARY KEY,
         content BLOB NOT NULL,
         was_instantiated INTEGER NOT NULL,
         created_at INTEGER NOT NULL
-    );
-    CREATE INDEX index_was_instanciated ON attachments(was_instantiated);
-    "#,
+    );"#,
+    "CREATE INDEX index_was_instanciated ON attachments(was_instantiated);",
```
jcnelson, line 41: Typo -- `index_was_instantiated`

jcnelson (approved):
This LGTM!  Just one minor comment to be addressed.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -474,7 +474,6 @@ const SORTITION_DB_INITIAL_SCHEMA: &'static [&'static str] = &[
     "CREATE UNIQUE INDEX snapshots_block_hashes ON snapshots(block_height,index_root,winning_stacks_block_hash);",
     "CREATE UNIQUE INDEX snapshots_block_stacks_hashes ON snapshots(num_sortitions,index_root,winning_stacks_block_hash);",
```
kantai: I think these two "CREATE UNIQUE INDEX" statements still need to be rewritten to drop the unique constraint, like:

```suggestion
    "CREATE INDEX snapshots_block_hashes ON snapshots(block_height,index_root,winning_stacks_block_hash);",
    "CREATE INDEX snapshots_block_stacks_hashes ON snapshots(num_sortitions,index_root,winning_stacks_block_hash);",
```

jcnelson: Ah, yes.  @lgalabru please address this before merging too.

src/net/atlas/db.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2420/commits/747297263d73b4c659df12dfce4af6c4f3cdd402

src/chainstate/burn/db/sortdb.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2420/commits/747297263d73b4c659df12dfce4af6c4f3cdd402
Thanks for the suggestion :) For some reason, I couldn't apply it.

kantai (approved):

PR #2421, opened 2021-02-05
[Boot from working_dir] Revisit working_dir architecture

lgalabru:
Assuming one of the next feature is the ability to re-use a working dir, this PR is revisiting the `working_dir` architecture (+ default name, now embedding a timestamp instead of random), so that miners can use the same working dir for a testnet node and a mainnet node (a la bitcoin). Finally it streamlines our file naming convention (`.sqlite`, instead of `.sqlite` / `.dat` / `.db` / no extension at all).
This will likely break 3rd party tools like `stacks-dump` or scripts that devops could be using for backups (cc @CharlieC3), but it feels like the more we wait, the harder it will be to migrate to a clean architecture.

Ideally, we ship this PR with the incoming `2.0.4`, so that the next reboot is the last reboot for miners.

Before:
```
➜  stacks-testnet-d09378c8197da090 tree .
.
├── burnchain
│   ├── db
│   │   └── bitcoin
│   │       └── regtest
│   │           ├── burnchain.db
│   │           └── sortition.db
│   │               ├── marf
│   │               ├── marf-shm
│   │               └── marf-wal
│   ├── spv-headers.dat
│   └── spv-headers.dat.reorg
├── chainstate
│   ├── atlas_db.sqlite
│   ├── chain-00000080-testnet
│   │   ├── blocks
│   │   │   ├── 244f
│   │   │   │   └── 5949
│   │   │   │       └── 244f59497d62dadc8b56e26bfab1b5402536202d28295826d5683824295d82fd
│   │   │   └── 6231
│   │   │       └── 9d93
│   │   │           └── 62319d931af3d42e2677b4a3a5732f571135df27b43419cb619e9f85f6a66b91
│   │   └── vm
│   │       ├── clarity
│   │       │   ├── marf
│   │       │   ├── marf-shm
│   │       │   └── marf-wal
│   │       ├── index
│   │       ├── index-shm
│   │       └── index-wal
│   ├── mempool.db
│   ├── mempool.db-shm
│   └── mempool.db-wal
└── peer_db.sqlite

```

After:
```
➜  stacks-node-1612546073 tree .
.
└── krypton
    ├── burnchain
    │   ├── burnchain.sqlite
    │   ├── sortition
    │   │   ├── marf.sqlite
    │   │   ├── marf.sqlite-shm
    │   │   └── marf.sqlite-wal
    │   ├── headers.sqlite
    │   └── headers.sqlite.reorg
    └── chainstate
        ├── atlas.sqlite
        ├── blocks
        │   ├── 7352
        │   │   └── ac30
        │   │       └── 7352ac30480398afa1d3940fe2998262a154c89c509af92c8660efae53d1b801
        │   ├── 8e51
        │   │   └── 5496
        │   │       └── 8e5154966e0b87aad4ad47ea7b7f164023d72717fac01b0a55dbc5eae9365656
        ├── mempool.sqlite
        ├── mempool.sqlite-shm
        ├── mempool.sqlite-wal
        ├── peer.sqlite
        └── vm
            ├── clarity
            │   ├── marf.sqlite
            │   ├── marf.sqlite-shm
            │   └── marf.sqlite-wal
            ├── index.sqlite
            ├── index.sqlite-shm
            └── index.sqlite-wal
```

Per suggestion from  https://github.com/blockstack/stacks-blockchain/issues/2414, it also rename the command

```bash
cargo testnet ...
```
to 
```bash
cargo stacks-node ...
```


jcnelson:
This layout looks great!  The only minor suggestions I would have are as follows:

* Use `atlas.sqlite` and `peer.sqlite` instead of `atlas_db.sqlite` and `peer_db.sqlite`.
* Use `headers.sqlite` instead of `spv-headers.sqlite` (the system is burnchain-agnostic, but SPV is a Bitcoin-specific term).

diwakergupta:
> Assuming one of the next feature is the ability to re-use a working dir, this PR is revisiting the `working_dir` architecture (+ default name, now embedding a timestamp instead of random), so that miners can use the same working dir for a testnet node and a mainnet node (a la bitcoin). Finally it streamlines our file naming convention (`.sqlite`, instead of `.sqlite` / `.dat` / `.db` / no extension at all).
> This will likely break 3rd party tools like `stacks-dump`

@psq FYI ⬆️

testnet/stacks-node/src/config.rs
```diff
@@ -831,36 +829,37 @@ impl Config {
...
-            "mocknet".to_string()
-        } else {
-            let (network, _) = self.burnchain.get_bitcoin_network();
-            network
-        };
-        format!(
-            "{}/burnchain/db/{}/{}/sortition.db/",
-            self.node.working_dir, self.burnchain.chain, dir_name
-        )
+        format!("{}sortition/", self.get_burnchain_path(),)
```
jcnelson, line 843: Defensive programming nitpick -- can you use a `PathBuf` here instead of `format!`, so we're always guaranteed that there will be a path separator and that it'll be the right character?

jcnelson: Same for the methods below.  We should have been doing this the whole time, but now's a good opportunity to address it.

jcnelson (approved):
This overall LGTM. Thanks @lgalabru!  But there's two things that I think have to happen before merging this:

* We'll want to add this to the changelog for the next release.  This change will require re-bootstrapping a node since the database layout on disk has changed.
* Whenever applicable, use the safer `PathBuf` struct to build paths instead of trying to combined path segments with `format!`.

psq:
> > This will likely break 3rd party tools like `stacks-dump`
> 
> @psq FYI ⬆️

I have something ready in https://github.com/psq/stacks-dump/tree/fix/working-fir (to mirror this pr name).  May need to tweak for things like regtest, and testnet, but works with mainnet

jcnelson:
N.B. we should clear the DB migration scheme for the sortition DB (in a separate PR) and Atlas DB (also a separate PR) after this merges, since it's no longer necessary (this is a breaking DB change).

kantai (approved):
LGTM!

testnet/stacks-node/src/config.rs
lgalabru: Addressed with latest

PR #2422, opened 2021-02-05
Fix #2416

### LINKED ISSUE #2416: Node: instantiate the mempool _once_
Right now, both the relayer thread and p2p thread try to instantiate the mempool DB.  This can reliably lead to a DoS (runtime panic) when the node attempts to process transactions, because the act of these two threads both attempting to create the same database will lead to its corruption.  Namely, the `mempool` table will not exist, and the WAL-mode `-shm` and `-wal` files will not be created.

What should instead happen is the node should create the mempool before spawning both of these threads.

jcnelson:
Instantiate the mempool before spawning threads, so the threads don't race each other to do so (and clobber the database in the process).

### DIFF
--- a/testnet/stacks-node/src/neon_node.rs
+++ b/testnet/stacks-node/src/neon_node.rs
@@ -1077,6 +1077,14 @@ impl InitializedNeonNode {
             _ => panic!("Unable to retrieve local peer"),
         };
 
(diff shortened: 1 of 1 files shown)


lgalabru (approved):
Looks good to me!

PR #2424, opened 2021-02-08
fix/atlas-crash

jcnelson:
Handle the case where rusqlite returns a `null` column value in a row, which it can do if querying a `MIN` or `MAX`.


lgalabru (approved):
Thank you @jcnelson

kantai (approved):

PR #2426, opened 2021-02-08
Disabling Atlas

lgalabru:


jcnelson (commented):
I think you should also disable `handle_getattachment` just to be safe.

src/net/rpc.rs
```diff
@@ -589,90 +589,96 @@ impl ConversationHttp {
...
-                let msg = format!("Unable to read Atlas DB");
-                warn!("{}", msg);
-                let response = HttpResponseType::ServerError(response_metadata, msg);
-                response.send(http, fd)?;
-                return Err(net_error::DBError(e));
-            }
-        }
+        let msg = format!("Atlas disabled");
+        warn!("{}", msg);
+        let response = HttpResponseType::ServerError(response_metadata, msg);
```
jcnelson, line 594: Can you make this 404 instead?

src/net/rpc.rs
```diff
@@ -589,90 +589,96 @@ impl ConversationHttp {
...
+        response.send(http, fd)?;
+        return Err(net_error::DBError(e));
```
jcnelson, line 596: What's `e` here?

lgalabru: Address with latest

src/net/rpc.rs
lgalabru: Address with latest

jcnelson (approved):

kantai (approved):

PR #2431, opened 2021-02-09
Move StacksString and UrlString to net module 

### LINKED ISSUE #2398: Refactor: Decouple Clarity VM
The Clarity `VM` is tightly coupled with the rest of the codebase. This has many implications, including that projects like the Clarity REPL can't depend on the Clarity `VM` alone. I'd like to decouple the `VM` to the point that it can be published as its own crate.

Remaining work:
- [ ] remove remaining `blockstack_lib` dependencies [M]
    - [ ] from `util`
    - [ ] from `vm`
- [ ] expose `vm` as its own crate [M]

reedrosenbluth:
Move string structs out of `util` module into `net`.

Changes:
- `mv util/strings.rs net/strings.rs`
- fix use statements

related issue: #2398


jcnelson (approved):
This refactoring is fine with me :+1: Just make sure the tests pass before merging.

src/vm/clarity.rs
```diff
@@ -1561,9 +1561,9 @@ mod tests {
     pub fn test_post_condition_failure_contract_publish() {
         use chainstate::stacks::db::*;
         use chainstate::stacks::*;
+        use net::strings::StacksString;
         use util::hash::Hash160;
         use util::secp256k1::MessageSignature;
-        use util::strings::StacksString;
```
lgalabru, line 1566: Not sure to follow, if we are moving `StacksString` to `net`, then `net` becomes a vm dependency.

reedrosenbluth: `net` was already a vm dependency as `vm` uses `StacksMessageCodec`and `NetError`. This is a step towards removing `util` as a dependency of `net`. 

I agree that introducing more coupling between `net` and `vm` goes against the ultimate goal of decoupling `vm`... Do you have any other proposals for where these things should live?

jcnelson:
Closing in favor of a future PR that will not only address all of these changes, but also move the codec stuff into its own module.

PR #2433, opened 2021-02-10
[Boot from working_dir] Better process termination

lgalabru:
This PR is improving the ability of terminating a stacks node gracefully, by implementing a handler terminating waiting for the following threads to achieve the ongoing processing:
- main thread
- relayer thread
- coordinator thread
- p2p thread

Would appreciate  getting some feedback on the approach.


testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -313,7 +317,30 @@ impl RunLoop {
...
+                info!("Terminating process");
+                // Terminate chains coordinator
+                info!("Stopping chains-coordinator");
+                coordinator_senders.stop_chains_coordinator();
+                // Terminate relayer
+                info!("Stopping relayer");
+                node.terminate_relayer();
+                // Commands handling is async, we'll sleep to make sure
+                // that the exit commands sent are ingested.
+                sleep(Duration::from_secs(10));
```
kantai, line 340: It'd be better to be able to join the thread handles here, so that we don't have to sleep and hope that the threads closed.

jcnelson: Seconded.  Definitely want to join threads.

testnet/stacks-node/Cargo.toml
```diff
@@ -21,6 +21,7 @@ base64 = "0.12.0"
 backtrace = "0.3.50"
 libc = "0.2"
 slog = { version = "2.5.2", features = [ "max_level_trace" ] }
+ctrlc = { version = "3.1.7", features = [ "termination" ] }
```
jcnelson, line 24: So, I took the time to look through this crate and its dependencies.  Here's what I found:

* `ctrlc` depends on `nix 0.18`, which in turn depends on `libc 0.2.72`.  We already have `libc 0.2.82` as a transitive dependency used consistently through blockstack-core and stacks-node's dependencies, so this crate definitely brings in a lot of extra code.  If it used `nix 0.20`, this problem would go away.
* `ctrlc`'s use of `nix` is questionable to me, because it's just using it to get access to signal handlers on UNIX platforms.  This is already provided by `libc`.  Using `nix` is odd here, because the value-add of having `nix` is to provide safe wrappers around `libc`'s unsafe code.  This benefit is lost on this crate, because this crate does a ton of necessarily-unsafe things with signal handling.  So, I'm not really a fan of this crate's packaging choices -- it should have just used `libc`.
* `ctrlc` commits a couple of somewhat bad library-writer sins that could come back to bite us later:
   * It has global state of its own -- namely, it doesn't let you register more than one signal handler via a global flag.  I'm willing to be lenient on this specific case, however, because signal handlers are already global state, and it's already bad programming practice for libraries to add their own signal handlers.
   * It uses the "signal pipe" trick to forward signals from a signal handler into the appropriate thread to handle them (a common tactic for dealing with signals in multi-threaded Unix programs), BUT, instead of allowing the application to handle incoming signal data from this pipe, it instead starts a thread internally to handle it.  This thread, in turn, calls the user-given callback function.
   * The fact that it uses a pipe _at all_ is odd to me, because Rust has safer IPC constructions for this purpose (like channels).  The code that forwards the signal data through the pipe is written without any error checking (!!), but it does at least have a lot of `TODO` comments to remind the author to fix them (for whatever that's worth).

Unfortunately, I don't think there's anything better for signal handling right now, short of writing our own..  The one redeeming feature of `ctrlc` is that it handles the Windows equivalent of signals, whereas none of the alternatives I've looked at do so:

* I looked at `signal_hook`, which doesn't depend on anything we don't already depend on, but it only works on Unix.
* I looked at `chan_signal`, but it's obsolete and superseded by `crossbeam-channel`, which has a crap-ton of dependencies.
* I considered just writing a small module for signal handling, but this again doesn't address the lack of Windows support (which is sadly much more complicated than Unix support).

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -313,7 +317,30 @@ impl RunLoop {
...
+                info!("Stopping stacks-node");
+                break;
+            }
```
jcnelson, line 343: Is there anything else that has to happen here?  Assuming all threads gracefully exit, are the node's databases all left in a consistent state such that the node can be started up again?

kantai: I think the only thread that won't _necessarily_ exit gracefully in the above implementation is the `p2p` thread and the `dns-resolver`. The `p2p` thread should exit gracefully once the `relayer` thread has exited (hanging up the relayer channel should cause the `p2p` thread to gracefully exit), but it may take it some time until the network receives data that needs to be passed to the relayer (which may not happen at all once the coordinator stops and the relayer stops). So this probably means that we need to submit an explicit exit command to the p2p thread somehow (the p2p thread _could_ leave bad db state if exited improperly).

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -313,7 +317,30 @@ impl RunLoop {
         // prepare to fetch the first reward cycle!
         target_burnchain_block_height = burnchain_height + pox_constants.reward_cycle_length as u64;
 
+        let should_keep_running = Arc::new(AtomicBool::new(true));
+        let keep_running_writer = should_keep_running.clone();
+
+        termination::set_handler(move || {
```
jcnelson, line 323: I would take a second look at where this is getting placed.  I think you'll want to do this _before_ you start any threads.

testnet/stacks-node/src/run_loop/neon.rs
jcnelson: Agreed.  The `dns-resolver` doesn't use any DB state, but the `p2p` thread can store blocks, microblocks, and transactions sent to it via the RPC interface.  So, it will need to be explicitly told to shut down.

testnet/stacks-node/src/run_loop/neon.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2433/commits/b2cc9e3b1fb15402af9191467a5197a3a160c69d

testnet/stacks-node/Cargo.toml
lgalabru: Thanks for auditing this @jcnelson!

testnet/stacks-node/src/run_loop/neon.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2433/commits/b2cc9e3b1fb15402af9191467a5197a3a160c69d

testnet/stacks-node/src/run_loop/neon.rs
lgalabru: Addressed with https://github.com/blockstack/stacks-blockchain/pull/2433/commits/b2cc9e3b1fb15402af9191467a5197a3a160c69d.
Also, instead of terminating the relayer from the main thread, the p2p thread is now in charge of sending the Exit command to the relayer, once all the network results from the ongoing p2p runloop epoch have been processed

kantai (approved):
LGTM!

jcnelson (approved):
LGTM!  Thanks @lgalabru!  I might try and vendor `ctrlc` at some point -- it's a very simple module, and it can easily be tweaked to work with our threads -- but doing this is not a release blocker.

PR #2436, opened 2021-02-11
Forward lockup events

lgalabru:
The goal of this PR is to fix the balance in the API, not taking into account lockups events, which end up being confusing for end users (balance outdated in explorer, wallet, etc). 
We're generating a synthetic receipt for wrapping all the lockup events, which will end up being forwarded to the stacks-blockchain-api.


src/chainstate/stacks/db/blocks.rs
```diff
@@ -4113,7 +4113,42 @@ impl StacksChainState {
                     let event = STXEventType::STXMintEvent(STXMintEventData { recipient, amount });
                     events.push(StacksTransactionEvent::STXEvent(event));
                 }
-                Ok((total_minted, events))
+
```
jcnelson, line 4116: Wait, we're _not_ going to report STX-transfers from unlocks?  Isn't this a breaking change to the event interface?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4113,7 +4113,42 @@ impl StacksChainState {
...
+                    }),
+                );
+                let lockup_receipt = StacksTransactionReceipt::from_stx_transfer(
+                    lockup_tx,
+                    events,
+                    Value::okay_true(),
+                    ExecutionCost::zero(),
+                );
+
+                Ok((total_minted, Some(lockup_receipt)))
```
jcnelson, line 4151: Why even include this in the event stream at all?  This information can be queried via the RESTful API already.

jcnelson:
I'm really confused by the intent of this PR.  Why can't the stacks blockchain API just call into the `.lockups` contract through the read-only contract-call RESTful API?  Why are we dropping the unlock events?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4113,7 +4113,42 @@ impl StacksChainState {
...
+                    }),
+                );
+
+                let lockup_tx = StacksTransaction::new(
+                    tx_version,
+                    boot_code_auth,
+                    TransactionPayload::ContractCall(TransactionContractCall {
+                        address: boot_code_address,
+                        contract_name: ContractName::try_from("lockup").unwrap(),
+                        function_name: ClarityName::try_from("get-lockups").unwrap(),
```
kantai, line 4140: Would "scheduled-unlock" be a better name for this phony contract call?

src/chainstate/stacks/db/blocks.rs
kantai: It's so that an event listener can build up an account balance history that includes unlocks, and maybe just as importantly, it allows an event listener to accurately derive an account balance without needing to have a kind of split-brain operation where some things need to be queried using the RESTful API, whereas other derived values are computed by the event stream.

kantai (approved):
This looks good to me, I just had a nitpick about the dummy transaction name.

src/chainstate/stacks/db/blocks.rs
jcnelson: That's part of why this is so confusing.  This `lockup_receipt` isn't actually returning any data -- it's just returning `(ok true)`, and isn't emitting any events of its own (`get-lockups` doesn't print anything or move any tokens).  So, none of the individual STX transfer events seem to be getting reported here.  Maybe that's a bug in this PR -- maybe the return value should be the `(list ..)` returned by `get-lockups`?

kantai: It's getting associated with the StxMint events.

lgalabru: Yeah, we are taking the events that were previously generated, and embedding them in a Receipt.
The reason for this is that the `stacks-blockchain-api` is not ingesting transactions events, it's ingesting transaction receipts.

src/chainstate/stacks/db/blocks.rs
lgalabru: Yeah, @zone117x to chime in and to confirm this, but I went with `get-lockups` because I think that the `stacks-blockchain-api` need to get a receipt associated with an existing method (and `get-lockups` is the only method in this contract).

src/chainstate/stacks/db/blocks.rs
jcnelson: Ah, yes, I see that now.  Thanks for helping me understand!

src/chainstate/stacks/db/blocks.rs
jcnelson: EDIT: I see that it's getting reported within the lockup receipt.

jcnelson (approved):
I don't want to block this, so I'll say this LGTM, but please first add a unit test for the new behavior before merging.

lgalabru:
Integration test added in https://github.com/blockstack/stacks-blockchain/pull/2436/commits/85116b912c3477dd6c10a59a7207754c52aa3650.
Will wait on @zone117x's approval before merging.

src/chainstate/stacks/db/blocks.rs
zone117x: I don't think associating these events to this fake tx will work. It looks like the API won't be aware of this tx and lots of things will break. I think your suggestion in the comment below about attaching to the coinbase transaction makes a lot more sense.

zone117x (changes requested):
Thanks @lgalabru. I think the events need to be attached to the coinbase tx in order for the API to consume these without issue. Other than that, looks good 👍

zone117x (approved):
Thanks @lgalabru lgtm

kantai:
I updated from `develop`, and cherry-picked a commit that should resolve the flakiness we were seeing in the `stacks-node` tests on CircleCI.

PR #2438, opened 2021-02-12
Fix: Miner heuristics for handling large transactions

kantai:
This PR fixes two issues for miners:

1. Block limit handling -- this fixes the way that miners assemble anchor blocks so that they use their configured block limit.
2. Expensive transaction heuristics -- this adds a heuristic to the miner for handling transactions that occupy over 80% of some dimension of the block limit which prevents those kinds of transactions from blocking txs that are behind the expensive in the miner's mempool "queue".


kantai:
Closing this, since we should be releasing from `develop`

PR #2439, opened 2021-02-12
Fix: Miner heuristics for handling large transactions

kantai:
This fixes some issues with transaction inclusion for miners -- this is the same patch as #2438, but this is opened against `develop`, so that we can do a release of `develop` alongside this.


src/core/mod.rs
```diff
@@ -48,6 +48,8 @@ pub const NETWORK_P2P_PORT: u16 = 6265;
 // its current block-commit in a sortition
 pub const MINING_COMMITMENT_WINDOW: u8 = 6;
 
+pub const TX_BLOCK_LIMIT_PROPORTION_HEURISTIC: u64 = 20;
```
lgalabru, line 51: Can we add a comment explaining how this constant is being used and why we came up with 20?

lgalabru (approved):
Just a tiny documentation request :)

src/core/mod.rs
kantai: Added in 973a6aad8b0896e3aa9abc6dd4aaf2ce5fc17a4f

jcnelson (approved):
This LGTM.  But before merging to `master`, I think we should try this out on a testnet miner.

PR #2452, opened 2021-02-15
Fix/2451

### LINKED ISSUE #2451: DB: load_descendant_staging_microblock_stream*() methods should verify completeness
Right now, the methods `load_descendant_staging_microblock_stream()` and `load_descendant_staging_microblock_stream_with_poison()` do not validate the stream's completeness.  So, they'll return a stream even if the stream is missing the first microblocks (e.g. 0..n).  This, in turn, has the downstream effect of causing the mining node to think that there's a new set of microblocks to confirm each time a new microblock arrives on the stream missing some of its initial blocks (since the length of the stream returned by these methods increases, despite the stream being incomplete).

jcnelson:
This fixes #2451 by making it so that we don't load a microblock stream if the stream's "prefix" is missing.  But, it still permits loading microblock stream forks, so we can mine poison-microblock transactions.  It also modifies the Neon node miner to look at a microblock stream's maximum sequence number (instead of its length) in order to decide when to RBF an in-flight block commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.0.6]
+
+### Fixed
+
+- The miner will no longer attempt to mine a new Stacks block if it receives a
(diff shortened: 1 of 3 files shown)


kantai (approved):

kantai:
This LGTM, but this should include an entry in the CHANGELOG.

lgalabru (approved):
LGTM

Issue #2444, opened 2021-02-14
Critical! Leader block commit sent multiple times for one block, which result to potentially infinite overspending


MaksimalistT:
**Describe the bug**
For some blocks leader block commit sent multiple times, which can result to committing as much as there are available BTC on the wallet if it takes long enough for next block to come.

Here is what happened for block:3710
`
INFO [1613339155.240835] [testnet/stacks-node/src/neon_node.rs:1589] [relayer] Succeeded assembling Stacks block #3710: 57590adf80f5c009d2f53ff5142d5a46af003969a7a10b3811d01887dd9407fc, with 17 txs, attempt 17
INFO [1613339155.291730] [testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs:1037] [relayer] Abort attempt to re-submit identical LeaderBlockCommit
WARN [1613339155.291759] [testnet/stacks-node/src/neon_node.rs:1654] [relayer] Failed to submit Bitcoin transaction
INFO [1613339158.551266] [src/net/http.rs:1558] [p2p] Handle GET /v2/blocks/7b8ccf2b757f2a8a80277ffc3b31d8bb58633e593fe1f77448918d17e81d233a
INFO [1613339165.308474] [src/net/http.rs:1558] [p2p] Handle GET /v2/blocks/d57ba9ebcabb807adc5f17c501b006dfb636c605bd02a4c1108f1740ea8d8c51
INFO [1613339212.853869] [src/net/chat.rs:1850] [p2p] convo:id=121,outbound=false,peer=18000000+00000001://47.242.239.96:56764: failed to send on P2P conversation: SendError("Failed to send 4 bytes")
INFO [1613339212.853899] [src/net/p2p.rs:2015] [p2p] Broken connection on event 121: SendError("Failed to send 4 bytes")
INFO [1613339225.942801] [src/net/http.rs:1558] [p2p] Handle GET /v2/neighbors
INFO [1613339277.782220] [src/net/http.rs:1558] [p2p] Handle GET /v2/blocks/0f0d3a5ba4c78694fddfb3c4e3b65120a46207abe8b41dbedf86be9aee5b720c
INFO [1613339279.783098] [src/net/http.rs:1558] [p2p] Handle GET /v2/blocks/6ce5929b90346383fa7f1bad556444c2136c0630324f5ea59e44b7d11ebb8836
[...]

Issue #2358, opened 2021-01-18
Burn values for sortition unexpected


friedger:
**Describe the bug**
Burn values for [1K1rSXoxAaJPQdWpm6XgZ6xwzkikxmWonW](https://btc.com/1K1rSXoxAaJPQdWpm6XgZ6xwzkikxmWonW) are 1 even though the miner is burning regularly.
```
666651 most_recent_burn: 205000, median_burn: 1, all_burns: [1, 1, 1, 1, 1, 205000]
666650 most_recent_burn: 205000, median_burn: 1, all_burns: [1, 1, 1, 1, 205000, 205000]
666649 BlockCommitNoParent
666649 most_recent_burn: 205000, median_burn: 1, all_burns: [1, 1, 1, 1, 1, 205000]
666646 most_recent_burn: 205000, median_burn: 1, all_burns: [1, 1, 1, 1, 1, 205000]
666644 leader key register 
```
**Expected behavior**

```
666651 most_recent_burn: 205000, median_burn: **205000**, all_burns: [1, 1, 1, **205000**, **205000**, 205000]
666650 most_recent_burn: 205000, median_burn: 1, all_burns: [1, 1, 1, 1, 205000, 205000]
666649 BlockCommitNoParent
666649 most_recent_burn: 205000, median_burn: 1, all_burns: [1, 1, 1, 1, 1, 205000]
666646 most_recent_burn: 205000, median_burn: 1, all_burns: [1, 1, 1, 1, 1, 205000]
666644 leader key register 
```

** Additional context**
```
stacks:~/programs/stacks-2.0.1$ grep a533b5670e99b28b9193f834a2f04074e410fd747c81897db11a6fc1337cfe6c miner.log  
DEBG [1611002220.456359] [src/chainstate/burn/distribution.rs:288] [chains-coordinator] Burn sample, txid: a533b5670e99b28b9193f834a2f04074e410fd747c81897db11a6fc1337cfe6c, most_recent_burn: 205000, median_burn: 1, all_burns: [1, 1, 1, 1, 1, 205000]
INFO [1611002220.494951] [src/chainstate/burn/db/sortdb.rs:3358] [chains-coordinator] ACCEPTED(666651) leader block commit a533b5670e99b28b9193f834a2f04074e410fd747c81897db11a6fc1337cfe6c at 666651,102

[...]

PR #2454, opened 2021-02-15
Fix: #2444, RBF fallback logic, and winning_txid fetching

kantai:
This PR addresses the superficial issues of #2444 (https://github.com/blockstack/stacks-blockchain/pull/2452 addresses a more underlying issue), by changing the RBF fallback logic. It also patches the way `winning_txid` is fetched by the sortition db (using the primary key fields to grab the winning block).

### DIFF
--- a/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
+++ b/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
@@ -56,6 +56,12 @@ use stacks::monitoring::{increment_btc_blocks_received_counter, increment_btc_op
 #[cfg(test)]
 use stacks::chainstate::burn::Opcodes;
 
+/// The number of bitcoin blocks that can have
+///  passed since the UTXO cache was last refreshed before
+///  the cache is force-reset.
+const UTXO_CACHE_STALENESS_LIMIT: u64 = 6;
+const DUST_UTXO_LIMIT: u64 = 5500;
+
 pub struct BitcoinRegtestController {
     config: Config,
     indexer_config: BitcoinIndexerConfig,
@@ -178,8 +184,6 @@ impl LeaderBlockCommitFees {
         self.final_size = new_size;
     }
 }
-
-const DUST_UTXO_LIMIT: u64 = 5500;
 
 impl BitcoinRegtestController {
     pub fn new(config: Config, coordinator_channel: Option<CoordinatorChannels>) -> Self {
@@ -1016,7 +1020,7 @@ impl BitcoinRegtestController {
         let mut traversal_depth = 0;
         let mut burn_chain_tip = burnchain_db.get_canonical_chain_tip().ok()?;
         let mut found_last_mined_at = false;
-        while traversal_depth < 6 {
+        while traversal_depth < UTXO_CACHE_STALENESS_LIMIT {
             if &burn_chain_tip.block_hash == &ongoing_op.utxos.bhh {
                 found_last_mined_at = true;
                 break;
--- a/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
+++ b/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
@@ -1005,10 +1005,11 @@ impl BitcoinRegtestController {
             }
         }
 
-        // Stop as soon as the fee_rate is 1.25 higher, stop RBF
+        // Stop as soon as the fee_rate is 1.50 higher, stop RBF
         if ongoing_op.fees.fee_rate > (self.config.burnchain.satoshis_per_byte * 150 / 100) {
-            let res = self.send_block_commit_operation(payload, signer, None, None, None, &vec![]);
-            return res;
+            warn!("RBF'd block commits reached 1.5x satoshi per byte fee rate, not resubmitting");
+            self.ongoing_block_commit = Some(ongoing_op);
+            return None;
         }
 
         // Did a re-org occurred since we fetched our UTXOs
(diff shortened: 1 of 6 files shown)


testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -805,6 +832,131 @@ fn stx_transfer_btc_integration_test() {
...
+            &mut tx,
+            &consensus_hash,
+            &stacks_block.header.block_hash(),
+            &garbage_block,
+        )
+        .unwrap();
+        tx.commit().unwrap();
+    }
+
+    thread::sleep(Duration::from_secs(30));
```
jcnelson, line 934: Is this long enough to prevent the node from trying lots of RBF's?  Is there a way to measure or reliably detect when the RBF limit gets reached, or is that no longer a concern with PR 2452?

jcnelson (approved):
This looks good to me.  Just had one minor question.

jcnelson:
I think you mean #2444 ;)

testnet/stacks-node/src/tests/neon_integrations.rs
kantai: The goal is to get it to do exactly that -- attempt tons of RBFs. Without the changes in this PR, the final assertion would fails with like `29 != 1` (i.e., the test miner would have issued 29 block commits).

lgalabru (approved):

kantai:
> I think you mean #2444 ;)

I'm pre-solving far future bugs :)

jcnelson:
> I'm pre-solving far future bugs :)

Oh well carry on then!

kantai:
Re-requesting reviews. There's another RBF fallback behavior that needed to be fixed here: in the presence of bitcoin forks, the UTXO set cache needs to be wiped, but the fork detection wasn't working correctly. The failing integration test surfaced this: the other RBF fallback issues were masking the problem.

diwakergupta:
> Re-requesting reviews. There's another RBF fallback behavior that needed to be fixed here: in the presence of bitcoin forks, the UTXO set cache needs to be wiped, but the fork detection wasn't working correctly. The failing integration test surfaced this: the other RBF fallback issues were masking the problem.

@kantai to confirm, there's no consensus-breaking change here right? It also sounded somewhat related to #2358 , is it?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -1005,20 +1005,34 @@ impl BitcoinRegtestController {
...
+            warn!("RBF'd block commits reached 1.5x satoshi per byte fee rate, not resubmitting");
+            self.ongoing_block_commit = Some(ongoing_op);
+            return None;
+        }
+
+        // Did a re-org occurred since we fetched our UTXOs, or are the UTXOs so stale that they should be abandoned?
+        let mut traversal_depth = 0;
+        let mut burn_chain_tip = burnchain_db.get_canonical_chain_tip().ok()?;
+        let mut found_last_mined_at = false;
+        while traversal_depth < 6 {
```
jcnelson, line 1019: This should probably be a constant somewhere, not `6`.

kantai:
Yep -- no consensus breaking change here. It's not directly related to #2358 -- rather, it's more closely related to the other RBF logic used by the neon_node when it submits commit ops. Basically, while tracking "pending transactions", the neon_node maintains a setting of pending UTXOs. That set needs to be wiped in the event of a bitcoin reorg. The previous logic for detecting this was not correct.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
kantai: Fixed in https://github.com/blockstack/stacks-blockchain/pull/2454/commits/bb3171d7cd2173c063a72c57dab7e69fdea23bea

jcnelson:
Is there an integration test for the new reorg-detection logic?

kantai:
> Is there an integration test for the new reorg-detection logic?

The existing `bitcoind_forking_test` covers the logic. It was failing with the other RBF fixes in place (but not failing without those fixes: the overzealous fallback to submitting a new commit op was masking this issue).

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
lgalabru: I trust your tests, but do you mind explaining what was wrong with the previous logic? I'm not sure to see how it was wrong.

jcnelson (approved):
LGTM!  Thanks for catching the UTXO invalidation bug

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
kantai: Yep, the prior logic had a few issues:

1. It should have been using the "stubbed" sortition ids -- those won't correspond to entries in the sortition db in any event. The stubbed sortition ids only exist for backwards compatibility while testing. In mainnet/testnet/regtest modes, the sortition id of a burn block will alway end up different than that stubbed id.
2. If the sortition is missing from the database, it wouldn't return an error, rather it would return `Ok(None)`.
3. The sortition db would continue to contain an entry for the "forked away" bitcoin blocks in most cases: data is deleted from it, and only PoX invalidations flip the "pox_valid" flag. if we wanted to use the sortition db to perform this check, we'd need to do something like `get_canonical_sortition_tip` to get the current sortition "chain tip", and then use that to invoke `get_sortition_id_for_bhh` which would look to see if it found the given burnchain tip within that sortition chain tip's fork history (it does this using a MARF lookup).

lgalabru: Got it, thank you!

lgalabru (approved):

PR #2457, opened 2021-02-16
Feat: Runtime-configurable tx logging, tx metrics

kantai:
This PR allows tx logging to be enabled with a runtime flag (`STACKS_TRANSACTION_LOG=1`), and adds prometheus metrics for tx processing:

```
# HELP stacks_node_mempool_outstanding_txs Number of still-unprocessed transactions received by this node since it started
# TYPE stacks_node_mempool_outstanding_txs gauge
stacks_node_mempool_outstanding_txs{handler="all"} 0
# HELP stacks_node_mempool_tx_confirm_times Time (seconds) between when a tx was received by this node's mempool and when a tx was first processed in a block
# TYPE stacks_node_mempool_tx_confirm_times histogram
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="300"} 2
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="600"} 4
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="900"} 4
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="1200"} 5
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="1500"} 7
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="1800"} 7
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="2100"} 7
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="2400"} 7
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="2700"} 7
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="3000"} 7
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="3600"} 7
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="4200"} 7
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="4800"} 7
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="6000"} 7
stacks_node_mempool_tx_confirm_times_bucket{handler="all",le="+Inf"} 7
```

As you can see, `stacks_node_mempool_outstanding_txs` isn't necessarily _truthful_: it's a runtime tracked variable, meaning that if the node was restarted, and one of its mempool transactions was confirmed, "outstanding_txs" could decrement even though the corresponding pending transaction wasn't including in the outstanding_txs count.


src/monitoring/mod.rs
```diff
@@ -117,6 +131,101 @@ pub fn increment_errors_emitted_counter() {
...
+}
+
+#[allow(unused_variables)]
+pub fn mempool_accepted(txid: &Txid, chainstate_root_path: &str) -> Result<(), DatabaseError> {
+    #[cfg(feature = "monitoring_prom")]
+    {
+        let tracking_db = txid_tracking_db(chainstate_root_path)?;
+
+        if txid_tracking_db_contains(&tracking_db, txid)? {
+            // processed by a previous block, do not track again
```
jcnelson, line 183: Couldn't this also be taken in the event a transaction winds up in two forks?

kantai: Yeah -- I think in that case as well, we wouldn't want the prom metrics to update. In general, our prom metrics _aren't_ fork aware. And this check just ensures that at least we aren't counting the same tx multiple times.

src/chainstate/stacks/db/mod.rs
```diff
@@ -753,11 +789,15 @@ impl StacksChainState {
             }
 
             if db_config.version != CHAINSTATE_VERSION {
-                error!(
-                    "Invalid chain state database: expected version = {}, got {}",
-                    CHAINSTATE_VERSION, db_config.version
-                );
-                return Err(Error::InvalidChainstateDB);
+                if db_config.version == "1" {
```
jcnelson, line 792: I almost wonder if this block should be its own method (or possibly its own module) that just executes migrations.  We'll surely add/change more tables in the future.

Related, do we have a plan or list of recommendations for schema changes that alter existing tables?  While we could always use `ALTER TABLE` to add new columns, I'm not sure that would result in the most maintainable code.  The way I've seen this done in other systems is they'll let developers create whatever schemas they want in-between successive versions, and then have their own `migrations` module that automatically calculates the schema delta between two schemas and executes code to morph one schema into another.  We don't have to solve this right now, but it's something to think about (I'll need to figure out something for #1805 because it adds new tables as well as new fields to existing tables).

Also, I think there should be a unit test for this (and for all schema migrations), so we can ensure that a node started from any DB schema version can automatically migrate to whatever the node's current schema version is.

src/monitoring/mod.rs
```diff
@@ -117,6 +131,101 @@ pub fn increment_errors_emitted_counter() {
...
+        }
+
+        prometheus::MEMPOOL_OUTSTANDING_TXS.inc();
+    }
+
+    Ok(())
+}
+
+#[allow(unused_variables)]
+pub fn log_transaction_processed(
```
jcnelson, line 194: I think we'll need to think about what'll happen if the user:

* starts the node with the tx log turned on, 
* stops the node,
* restarts the node with the tx log turned off
* stops the node
* starts the node with the tx log turned on again.

Is handling missing transactions a concern?  If so, then is there a plan to either force the node to re-index all unprocessed transactions on subsequent boot-up (bitcoin does this IIRC), or should this method simply stop processing transactions altogether if the node ever starts up with the tx log turned off?

kantai: I don't think there's necessarily an issue here -- tx's that are processed while tx log is turned off just won't appear in this table.

jcnelson: Right; I'm asking if that's going to be a problem.

src/chainstate/stacks/db/mod.rs
kantai: > I almost wonder if this block should be its own method (or possibly its own module) that just executes migrations. We'll surely add/change more tables in the future.

Probably could be, though my preference is always to save this kind of abstraction for when we have something to abstract across (i.e., more than one example). Otherwise, I worry about abstracting wrongly.

>  The way I've seen this done in other systems is they'll let developers create whatever schemas they want in-between successive versions, and then have their own migrations module that automatically calculates the schema delta between two schemas and executes code to morph one schema into another. We don't have to solve this right now, but it's something to think about (I'll need to figure out something for #1805 because it adds new tables as well as new fields to existing tables).

Yeah -- I've thought a little bit about this, and I think we ultimately will need something pretty generic. Looking at the last few db schema changes:

1. the addition of `parent_sortition_id` to the sortition db
2. new indexes
3. db_schema version table
4. this PR

Items 2, 3, and 4, would all be potentially solvable with a schema-computed migration. Item 1 _wouldn't_ be though, and neither would many of the changes for #1805. For those, migration may be possible with a bulk process (i.e., `parent_sortition_id` could have been written from available state), or it may not be, in which case, the migration process should just panic.

I think because of that we likely will want a pretty abstracted process, where schema versions just have associated functions that handle the migration from `vPrior -> vCurrent`, and we iterate those. If it turns out that many of those handlers could be rewritten with schema-computed migrations, then we can refactor those handlers, but I don't think we should try to put something like that in place before it's clear that there would be a benefit.

src/monitoring/mod.rs
kantai: I don't think so -- if someone is particularly interested in data for the tx log, they should run from a fresh start. If it turns out that the re-index support is a highly desired feature, we can always add support for that later on.

src/chainstate/stacks/db/mod.rs
jcnelson: Sounds good to me!

src/monitoring/mod.rs
jcnelson: Sounds good

jcnelson (approved):
LGTM!

lgalabru (approved):
Looks good to me!

Issue #2385, opened 2021-01-26
Stacking details in v2/pox


agraebe:
Over the last few weeks, we've identified a couple of things that could be added to the endpoint in order to make integrations easier:

- Add a field for the # bitcoin blocks before the next prepare phase starts - "next_prepare_phase_in"
- Add a field for the total ustx locked for the upcoming cycle - "next_cycle_ustx_locked"
- Add a field for the pox activation of the upcoming cycle - "next_cycle_pox_active"
- Add a field for the dynamically calculated minimum ustx required for a reward slot - "min_reward_slot_ustx"

cc @kantai @zone117x @kyranjamie

jcnelson:
I think this is a more-refined version of #2254.

agraebe:
as @lgalabru noted in the previous issue, if pox is not active, a reason would be helpful

lgalabru:
Addressed with https://github.com/blockstack/stacks-blockchain/pull/2459

PR #2459, opened 2021-02-17
Feature: `/v2/pox` endpoint updates + OpenAPI docs for RPC endpoints

kantai:
This improves the `/v2/pox` endpoint, returning data like:

```json
{
  "contract_id": "SP000000000000000000002Q6VF78.pox",
  "pox_activation_threshold_ustx": 52329761604388,
  "first_burnchain_block_height": 666050,
  "prepare_phase_block_length": 100,
  "reward_phase_block_length": 2000,
  "reward_slots": 4000,
  "rejection_fraction": 25,
  "total_liquid_supply_ustx": 1046595232087772,
  "current_cycle": {
    "id": 2,
    "min_threshold_ustx": 70000000000,
    "stacked_ustx": 202157971547640,
    "is_pox_active": true
  },
  "next_cycle": {
    "id": 3,
    "min_threshold_ustx": 70000000000,
    "min_increment_ustx": 52329761604,
    "stacked_ustx": 162057034977640,
    "prepare_phase_start_block_height": 672250,
    "blocks_until_prepare_phase": 407,
    "reward_phase_start_block_height": 672350,
    "blocks_until_reward_phase": 507,
    "ustx_until_pox_rejection": 261648808021925
  },
  "min_amount_ustx": 70000000000,
  "prepare_cycle_length": 100,
  "reward_cycle_id": 2,
  "reward_cycle_length": 2100,
  "rejection_votes_left_required": 261648808021925,
  "next_reward_cycle_in": 507
}
```

This includes the information requested in #2385 as well as some additional information:

* `contract_id` is the contract identifier for the PoX contract.
* `cur_cycle_threshold` is the threshold amount for obtaining a slot in the
  active reward cycle.
* `next_cycle_cur_threshold` is the _current_ microstacks threshold for the
  _next_ reward cycle. If more STX is locked for participation in that cycle,
  the threshold may increase.
* `first_burnchain_block_height` is the first burn block evaluated in this Stacks
  chain.
* `cur_cycle_stacked_ustx` is the total amount of stacked microstacks in the
  active reward cycle.
* `next_cycle_stacked_ustx` is the total amount of stacked microstacks in the
  next reward cycle.
* `reward_slots` is the number of reward slots in a reward cycle
* `next_rewards_start` is the burn block height when the next reward
  cycle begins
* `next_prepare_phase_start` is the burn block height when the next prepare
  phase begins. Any eligible stacks _must_ be stacked before this block.
* `min_stacking_increment_ustx` is the minimum amount that can be used to
  submit a `stack-stx` call.
* `pox_activation_threshold` is the threshold of stacking participation that
  must be reached for PoX to activate in any cycle.
* `prepare_cycle_length` is the length in burn blocks of the prepare phase
* `rejection_fraction` is the fraction of liquid STX that must vote to reject
  PoX in order to prevent the next reward cycle from activating.
* `rejection_votes_left_required` is the remaining amount of liquid
  STX that must vote to reject the next reward cycle to prevent the next
  reward cycle from activating.
* `reward_cycle_id` is the active reward cycle number
* `reward_cycle_length` is the length in burn blocks of a whole PoX
  cycle (reward phase and prepare phase)
* `total_liquid_supply_ustx` is the current total amount of liquid microstacks.
* `next_reward_cycle_in` is the number of burn blocks until the next reward
  cycle begins.
* `next_prepare_phase_in` is the number of burn blocks until the next prepare
  phase starts.

Note: as currently implemented, this endpoint change _is not_ backwards compatible. The previous field `min_amount_ustx` was removed because the field was particularly unhelpful. That value becomes `min_stacking_increment_ustx`, and I believe that downstream clients were using that field as if it were `next_cycle_cur_threshold`. I'm happy to add that field back in, and fill it with either one of the two threshold fields if that makes things easier for clients.

Update: based on feedback, this was updated for better consistency, and backwards-compatibility. See the OpenAPI specs for descriptions of the new response schema. The example in this comment was updated, but the field descriptions weren't.


src/net/rpc.rs
```diff
@@ -303,20 +307,90 @@ impl RPCPoxInfoData {
             .to_owned()
             .expect_u128() as u64;
 
-        let total_required = total_liquid_supply_ustx
-            .checked_div(rejection_fraction)
-            .expect("FATAL: unable to compute total_liquid_supply_ustx/current_rejection_votes");
+        let total_required = (total_liquid_supply_ustx as u128)
+            .checked_div(100)
+            .expect("FATAL: unable to compute total_liquid_supply_ustx * 100")
```
jcnelson, line 312: I'm not sure the `expect` is necessary here?  `checked_div(INTEGER_CONSTANT)` I don't think can ever fail, so why not just use plain old integer divide?

src/net/rpc.rs
```diff
@@ -559,10 +633,11 @@ impl ConversationHttp {
         sortdb: &SortitionDB,
         chainstate: &mut StacksChainState,
         tip: &StacksBlockId,
-        options: &ConnectionOptions,
+        burnchain: &Burnchain,
     ) -> Result<(), net_error> {
         let response_metadata = HttpResponseMetadata::from(req);
-        match RPCPoxInfoData::from_db(sortdb, chainstate, tip, options) {
+
+        match RPCPoxInfoData::from_db(sortdb, chainstate, tip, burnchain) {
```
jcnelson, line 640: If we're going to be updating this method anyway, can you take the opportunity to handle the `Err(net_error::NotFoundError)` variant as an HTTP 404, instead of an HTTP 500?  Right now, we do the latter in case of any error, even if the only problem is that the `tip=` query parameter refers to an unknown block.

src/net/rpc.rs
jcnelson: Actually, can we just make this entire `total_required` calculation just error out instead of runtime panic?  I see that we're already doing that below in the event of a numeric overflow.

src/net/rpc.rs
```diff
@@ -303,20 +307,90 @@ impl RPCPoxInfoData {
...
+
+        if reward_cycle_length != pox_consts.reward_cycle_length as u64 {
+            error!(
+                "PoX Constants in config mismatched with PoX contract constants: {} != {}",
+                reward_cycle_length, pox_consts.reward_cycle_length
+            );
+            return Err(net_error::DBError(db_error::Corruption));
+        }
+
+        let effective_height = burnchain_tip.block_height - first_burnchain_block_height;
```
jcnelson, line 339: These subtractions can _never_ underflow, right?

src/net/rpc.rs
```diff
@@ -303,20 +307,90 @@ impl RPCPoxInfoData {
...
+        let next_prepare_phase_start =
+            if burnchain_tip.block_height < next_reward_cycle_prepare_phase_start {
+                next_reward_cycle_prepare_phase_start
+            } else {
+                // currently in a prepare phase, so the next prepare phase start is actually the reward cycle after
+                //  next
+                next_reward_cycle_prepare_phase_start + reward_cycle_length
+            };
+
+        let next_prepare_phase_in = next_prepare_phase_start - burnchain_tip.block_height;
```
jcnelson, line 354: `next_reward_cycle_in - prepare_length >= 0` always, right?  Otherwise this could underflow

src/net/rpc.rs
kantai: Sure

src/net/rpc.rs
kantai: No -- but the `next_prepare_phase_start` calculation takes this into account:

```rust
        let next_prepare_phase_start = if burnchain_tip.block_height < next_reward_cycle_prepare_phase_start {
            next_reward_cycle_prepare_phase_start
        } else {
            // currently in a prepare phase, so the next prepare phase start is actually the reward cycle after
            //  next
            next_reward_cycle_prepare_phase_start + reward_cycle_length
        };
```

So, in the event that we're in a prepare phase,
`next_prepare_phase_in = burnchain_tip.block_height + next_reward_cycle_in + reward_cycle_length - prepare_cycle_length - burnchain_tip.block_height`

Which is:
`next_prepare_phase_in = next_reward_cycle_in + reward_cycle_length - prepare_cycle_length`

If `prepare_cycle_length > reward_cycle_length`, this could underflow, _but_ that would already have caused a panic in the `PoxConstants` constructor.

src/net/rpc.rs
kantai: Right, if the canonical burn chain tip is less than `first_burnchain_block_height`, there is something very wrong happening.

src/net/rpc.rs
kantai: Updated in 9e28b0a3851de7495c240c53595c65c477497033

src/net/rpc.rs
kantai: Updated in 9e28b0a3851de7495c240c53595c65c477497033

src/net/rpc.rs
jcnelson: thanks!

docs/rpc-endpoints.md
```diff
@@ -73,6 +73,72 @@ Possible values for the "reason" field and "reason_data" field are:
...
+* `cur_cycle_stacked_ustx` is the total amount of stacked microstacks in the
+  active reward cycle.
+* `next_cycle_stacked_ustx` is the total amount of stacked microstacks in the
+  next reward cycle.
+* `reward_slots` is the number of reward slots in a reward cycle
+* `next_rewards_start` is the burn block height when the next reward
+  cycle begins
+* `next_prepare_phase_start` is the burn block height when the next prepare
+  phase begins. Any eligible stacks _must_ be stacked before this block.
+* `min_stacking_increment_ustx` is the minimum amount that can be used to
```
jcnelson, line 123: I guess I'm a little bit confused by this value.  Is this the number of uSTX required to clinch a single reward slot?  Because if so, shouldn't it always be a multiple of 10,000 STX?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -2103,6 +2148,35 @@ fn pox_integration_test() {
...
+    assert_eq!(
+        pox_info.rejection_fraction,
+        pox_constants.pox_rejection_fraction
+    );
+    assert_eq!(pox_info.reward_cycle_id, 14);
+    assert_eq!(
+        pox_info.reward_cycle_length as u32,
+        pox_constants.reward_cycle_length
+    );
+    assert_eq!(pox_info.next_reward_cycle_in, 14);
```
jcnelson, line 2178: Should we be testing here whether or not the act of Stacking will bump the minimum required uSTX to lock per reward slot?

docs/rpc-endpoints.md
kantai: It's the amount that will result in an immediate rejection from `stack-stx`

testnet/stacks-node/src/tests/neon_integrations.rs
kantai: I'm not sure -- that's tested in the coordinator integration tests (and the other PoX unit tests), and I'm a little hesitant to add much more complexity to these integration tests, because they are much more painful to iterate on.

docs/rpc-endpoints.md
jcnelson: Ah, makes sense :+1: 

Should there be a field in /v2/pox that tells you how many uSTX you need to lock to get a reward slot?  I have a script [here](https://github.com/jcnelson/get-stacking-info.sh/blob/main/get-stacking-info.sh) that I use for this, but I could easily see this information getting added to this endpoint.

testnet/stacks-node/src/tests/neon_integrations.rs
jcnelson: Fair point.  I'm fine with not testing it here as long as it's tested adequately elsewhere.

jcnelson (approved):
This LGTM!  Just one last comment on whether or not /v2/pox should tell you how many uSTX will be required per reward address, but that shouldn't block this PR if you don't want it to.

docs/rpc-endpoints.md
kantai: That's _sort of_ what `next_cycle_cur_threshold` and `cur_cycle_threshold` tell you -- `cur_cycle_threshold` tells you the amount that would have needed to be stacked to get 1 reward address in the current cycle, `next_cycle_cur_threshold` tells you the amount that would be needed for the next cycle, assuming no more participation occurs.

jcnelson: Ah, okay. that's what I was looking for. Thanks!

agraebe:
thanks for the PR! wanted to bring up a few things:

**API versioning**
A backward-incompatible change will likely happen in the future again. I think we cannot assume such changes could be making trivially without a lot of follow-ups and significant lead time before we roll out such changes. I'd propose breaking this point out into two items:
- Leave the `min_amount_ustx` in and document it as "to be deprecated in the future". The downstream changes require updates in the docs, the API and the API client, the API and API client docs
- Align on a versioning strategy for the RPC endpoints. I'd suggest we find a sustainable way to making these changes without breaking existing integrations/clients/etc

**API endpoint design**
We use names in an inconsistent way (sometimes types are missing, prop names have types at the end or the beginning, we use phase and reward synonymously, we annotate cycle or not)... All that is to say: could we design a more consistent API? Maybe this could serve as a suggestion:

```json
{
    "contract_id": "SP000000000000000000002Q6VF78.pox",
    "pox_activation_threshold_ustx": 52251700044388,
    "first_burnchain_block_height": 666050,
    "prepare_phase_block_length": 100,
    "reward_phase_block_length": 2100,
    "reward_slots": 4000,
    "rejection_fraction": 25,
    "total_liquid_supply_ustx": 1045034000887772,
    "current_cycle": {
        "id": 2,
        "min_threshold_ustx": 70000000000,
        "stacked_ustx": 202157971547640,
        "is_pox_active": true // note: this is new
    },
    "next_cycle": {
        "id": 3,
        "min_threshold_ustx": 70000000000,
        "min_increment_ustx": 52251700044,
        "stacked_ustx": 151807880982060,
        "prepare_phase_start_block_height": 672250,
        "blocks_until_prepare_phase": 1218,
        "reward_phase_start_block_height": 672350,
        "blocks_until_reward_phase": 1318,
        "ustx_until_pox_rejection": 261258500221925,
        "is_pox_active": true // note: this is new
    },
    "min_amount_ustx": 70000000260, // needs to stay for backward compatibility
    "prepare_cycle_length": 100, // needs to stay for backward compatibility
    "reward_cycle_id": 2, // needs to stay for backward compatibility
    "reward_cycle_length": 2100, // needs to stay for backward compatibility
    "rejection_votes_left_required": 41864499875510, // needs to stay for backward compatibility
    "next_reward_cycle_in": 622 // needs to stay for backward compatibility
}
```

**API documentation**
Just hijacking this PR to reinforce that we need to always make docs updates if we update the RPC endpoints. The source of truth for developers coming to us is the documentation. The content in the docs itself relies on the OpenAPI spec in the API repo. I wanted to call out that a change to an RPC endpoint in this repo should not be enough to call an update completed. We should discuss how to make this process more streamlined, so we're not ending up with inconsistent docs (for instance, the prop `next_reward_cycle_in` is still not in the API docs)

kantai:
>     "min_amount_ustx": 70000000260, // needs to stay for backward compatibility

`min_amount_ustx` could be left in for compatibility, but it won't be this value -- it'd always be `next_cycle.min_threshold_ustx`.

> The content in the docs itself relies on the OpenAPI spec in the API repo. I wanted to call out that a change to an RPC endpoint in this repo should not be enough to call an update completed. We should discuss how to make this process more streamlined, so we're not ending up with inconsistent docs (for instance, the prop next_reward_cycle_in is still not in the API docs)

Hmm. I'd recommend that we move this spec over to this repo, then. It's a little strange to document the same API in two places (this PR _does_ document its changes, but it documents them in `docs/rpc-endpoints.md`), and I'd push back against making the source of truth for the endpoint live in a different repo, because that would then require two PRs (synchronized to releases) to be reviewed anytime an API endpoint here changes.

agraebe:
> I'd recommend that we move this spec over to this repo, then

I agree, we shouldn't have different places for the RPC refs. I don't think we're pointing anyone to the `rpc-endpoints.md` file at this stage. Maybe @zone117x could help us find a solution. Could we have an OpenAPI spec in the repo that the API repo pulls and rolls ups into the broader docs that are generated?

zone117x:
> Could we have an OpenAPI spec in the repo that the API repo pulls and rolls ups into the broader docs that are generated?

This would be nice, single source of truth and whatnot. But I don't mind keeping the API repo's spec updated with the latest stacks-node API. This task could be put in the backlog for now imo.

agraebe:
@zone117x I'll open a ticket to discuss a more automated way. Doing it manually is likely not a great solution. I can't imagine anyone of us having enough confidence to identify all endpoint changes manually, across all PRs. It wouldn't be required if we believe it's ok to have some inconsistency but I'd try to stay away from accepting that.

@kantai I'll move every other convo in new issues - just reviewing the prop names for consistency would be great (and keeping the fields clients already use).

kantai:
Okay cool -- it's also easy enough for me to just copy over the current OpenAPI specs that are relevant to the blockchain API into this repo in this PR (and update the /v2/pox spec), so I'll go ahead and do that. That'll definitely get us a little closer to an automated solution (and should make the manual process much less painful in the mean time).

kantai:
>         "is_pox_active": true // note: this is new

Didn't see this before in your comment, but this cannot be known before the reward cycle begins. Anchor selection occurs at the start of the reward cycle and determines whether or not PoX will be active.

kantai:
Yep -- will add to cur cycle info.

kantai:
Updated based on feedback -- example resonse is:

```

**Review by jcnelson — COMMENTED** [review]
Re-approving.  Question -- is there a way to automatically generate and/or test the documented API results against what the code actually does?  Should we add that (or at least create an issue for it if we're not going to address it here)?

**Comment by kantai** [comment]
> Re-approving. Question -- is there a way to automatically generate and/or test the documented API results against what the code actually does? Should we add that (or at least create an issue for it if we're not going to address it here)?

Very probably, though I think we'd need to do a fair amount of plumbing -- I believe we'd need to write a set of new unit tests (or one big unit test like the way the vm::docs tests work) which then used `jsonschema` and `openapi` libraries to check some results. I'll create an issue for it, because I think it's non-trivial enough that I wouldn't want to delay this PR much more for it.

**Comment by zone117x** [comment]
OpenAPI tooling can be pretty brittle. Especially with relatively local file system references -- there isn't really a standard, and the one that exists is not followed by most tooling. I recommend adding at least an OpenAPI file dereference + validator step in this PR before merging.

**Comment by kantai** [comment]
> OpenAPI tooling can be pretty brittle. Especially with relative local file system references -- there isn't really a standard, and the one that exists is not followed by most tooling. I recommend adding at least an OpenAPI file dereference + validator step in this PR before merging. Preferably the NodeJS tool used in the API repo (which I think is a good reference that covers the larger ecosystem).

Yep -- happy to add a validation action here. What validation are you referring to from the API repo? I see `validate:schemas`, but that seems to only check `*.example.json`, which aren't outputted by the gulp flattenSchemas pass:

```
$ find .tmp
.tmp
.tmp/api
.tmp/api/transaction
.tmp/api/transaction/post-core-node-transactions-error.schema.json
.tmp/api/contract
.tmp/api/contract/post-call-read-only-fn.schema.json
.tmp/api/core-node
.tmp/api/core-node/get-account-data.schema.json
.tmp/api/core-node/get-fee-transfer.schema.json
.tmp/api/core-node/get-contract-data-map-entry.schema.json
.tmp/api/core-node/get-contract-source.schema.json
.tmp/api/core-node/get-pox.schema.json
.tmp/api/core-node/get-contract-interface.schema.json
.tmp/api/core-node/get-info.schema.json
.tmp/entities
.tmp/entities/contracts
.tmp/entities/contracts/read-only-function-args.schema.json
```

And in `validate-schemas.ts`:

```
glob('.tmp/**/*.example.json', (err, files) => {
 ...
```

@zone117x, is one of the other run commands in that package.json doing the OpenAPI validation?

**Comment by zone117x** [comment]
The OpenAPI schema is validated in the API repo during the doc generation step. I think to run only the validation step, something like the following is needed:
```
npm install -g @apidevtools/swagger-cli
swagger-cli validate docs**/*.json
```

It may require the `swagger-cli bundle --dereference` step be ran first to deref all the files.

kantai:
> The OpenAPI schema is validated in the API repo during the doc generation step. I think to run only the validation step, something like the following is needed:

Hmm -- the docs generation in the API repo appears to use `redoc-cli` to do the docs generation, so I'll just use that to do the validation step here (`swagger-cli validate` appears to be angry about all sorts of things in the spec here as well as in the API repo).

zone117x (approved):

markmhx:
Has this PR been deployed into production, and does the legacy `min_amount_ustx` property now update dynamically or does it still map to a static value?

diwakergupta:
> Has this PR been deployed into production, and does the legacy `min_amount_ustx` property now update dynamically or does it still map to a static value?

No, not yet. Current plan is to deploy a release-candidate onto the testnet before next Tuesday. As for the re-mapping of `min_amount_ustx`, that is happening at the API layer -- `v2/pox` on a stacks-node _always_ returns a dynamic value.

Issue #2476, opened 2021-02-25
[p2p] less aggressive PoX watchdog behavior when not in the initial block download


jcnelson:
The PoX watchdog effectively prevents the node from downloading another burnchain block until it has done a full inventory sync with at least one always-allowed peer.  This has the undesirable side-effect of making it so that the time between a Bitcoin block getting mined and the node mining the next Stacks block can be somewhat long -- on the order of 1-2 minutes.  This makes it hard for the system to mine in the presence of Bitcoin flash blocks, in which one or more Bitcoin blocks arrive in rapid succession (i.e. before the node has a chance to mine).  The PoX watchdog should only require an inventory (and downloader) synchronization before processing new sortitions if the node is well and far behind the chain tip.

Issue #2475, opened 2021-02-25
[config] make timeouts and rlimits configurable


jcnelson:
Make the following fields configurable options:

* handshake_timeout -- time in which a node must complete a handshake before being declared idle and disconnected
* connect_timeout -- time in which a node must complete a socket connection
* max_sockets -- maximum number of file descriptors the node is allowed to open for sockets (currently hardcoded to 800)

Issue #2474, opened 2021-02-25
[p2p] advance PoX reward cycle synchronization only after both an inventory and downloader state-machine pass


jcnelson:
Right now, the PoX sync watchdog doesn't verify that the block downloader to enter a quiescent state before proceeding to download the next PoX reward cycle's burnchain blocks.  This can cause the node to process sortitions without having processed the anchor block, even though it exists and is available on the network, thereby leading to needless short-lived PoX reorgs.  To address this (in the absence of #1805), the node should at least wait until the downloader has finished trying to fetch blocks for a given reward cycle before deciding to move to the next reward cycle.

Issue #2473, opened 2021-02-25
[p2p] explicit initial block download behavior


jcnelson:
When booting up, a node will often get stalled by slow remote nodes that are slow to reply to block requests or inventory requests.  The way other blockchains solve this is to have an explicit "initial block download" behavior during boot-up whereby the node only talks to a fixed set of (usually just one) remote peers, and then proceeds to enmesh itself into the peer network after it has reached the chain tip.

PR #2477, opened 2021-02-25
Fix/Flashblocks

jcnelson:
This PR addresses a slew of small network fixes (#2473, #2474, #2475, #2476) that make the system better able to both boot up quickly, and to more-reliably mine in the presence of Bitcoin flash blocks.  The last fix is particularly pertinent because PoX activating at all is wholly dependent on the system's ability to reliably mine during the prepare phase.

**This PR is blocked on #2388**.  I merged #2388 into this PR so I could test the system's resilience to flash blocks without burning real BTC.  So, don't merge until #2388 is in `develop`.

EDIT: I have added an emergency fix that prevents valid Stacks blocks from getting deleted when recovering from a too-long microblock stream.


testnet/stacks-node/src/config.rs
```diff
@@ -443,6 +443,7 @@ impl Config {
                         None => default_node_config.local_peer_seed,
                     },
                     miner: node.miner.unwrap_or(default_node_config.miner),
+                    mock_mining: node.mock_mining.unwrap_or(default_node_config.mock_mining),
```
kantai, line 446: Thanks for catching this!

kantai (approved):
This LGTM! Also -- I wouldn't necessarily say this is blocked on #2388: if we merged this PR, that PR would be marked as "merged as well".

src/net/mod.rs
```diff
@@ -593,9 +593,9 @@ impl PeerAddress {
     pub fn is_in_private_range(&self) -> bool {
         if self.is_ipv4() {
             // 10.0.0.0/8, 172.16.0.0/12, or 192.168.0.0/16
-            self.0[13] == 10
-                || (self.0[13] == 172 && self.0[14] >= 16 && self.0[14] <= 31)
-                || (self.0[13] == 192 && self.0[14] == 168)
+            self.0[12] == 10
```
diwakergupta, line 596: Is this a regression, or fixes an existing bug?

src/net/p2p.rs
```diff
@@ -4025,7 +4037,7 @@ impl PeerNetwork {
         // set up sockets that have finished connecting
         self.process_connecting_sockets(&mut poll_state);
 
-        // find out who is inbound and unathenticed
+        // find out who is inbound and unauthenticed
```
diwakergupta, line 4040: [typo] "unauthenticated" :)

testnet/stacks-node/src/syncctl.rs
```diff
@@ -382,34 +426,40 @@ impl PoxSyncWatchdog {
...
+            burnchain_tip.block_snapshot.block_height,
+            burnchain_height,
+        );
+
         // unconditionally download the first reward cycle
         if burnchain_tip.block_snapshot.block_height
             < burnchain.first_block_height + (burnchain.pox_constants.reward_cycle_length as u64)
         {
             debug!("PoX watchdog in first reward cycle -- sync immediately");
             sleep_ms(10_000);
```
diwakergupta, line 444: Do we want to revisit this 10s sleep?

diwakergupta: I also recall we had added a 30s sleep somewhere just after the mainnet launch, do we still need that?

testnet/stacks-node/src/syncctl.rs
```diff
@@ -572,12 +619,19 @@ impl PoxSyncWatchdog {
                 }
             };
 
-            debug!("PoX watchdog: Wait for at least one inventory state-machine pass before resetting...");
-            self.relayer_comms.wait_for_inv_sync_pass(3600); // wait at most an hour, but should be far, far less
+            if ibbd || !steady_state {
+                debug!("PoX watchdog: Wait for at least one downloader state-machine pass before resetting...");
+                self.relayer_comms.wait_for_download_pass(3600); // wait at most an hour, but should be far, far less
```
lgalabru, line 624: Should `3600` be a constant / setting?

testnet/stacks-node/src/syncctl.rs
lgalabru: Also, should we turn it into a constant / setting?

src/net/mod.rs
jcnelson: Yes, both.

testnet/stacks-node/src/syncctl.rs
jcnelson: This is here mainly to prevent the node from eating 100% CPU (and slamming the Bitcoin node) while processing the first reward cycle.  Can change to the sync constant `self.steady_state_burnchain_sync_interval`.

testnet/stacks-node/src/syncctl.rs
jcnelson: It really should be "wait forever".

src/chainstate/stacks/miner.rs
```diff
@@ -256,7 +256,7 @@ impl<'a> StacksMicroblockBuilder<'a> {
             txs: txs,
         };
 
-        debug!(
+        info!(
             "\n\nMiner: Created microblock block {} (seq={}) off of {}/{}: {} transaction(s)\n",
```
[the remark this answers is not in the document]
jcnelson: Sure

jcnelson:
Only failing test is a bind error.  Going to merge once CI finishes.

testnet/stacks-node/src/syncctl.rs
jcnelson: Addressed in 08db53b

src/chainstate/stacks/miner.rs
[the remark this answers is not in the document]
jcnelson: Addressed in 08db53b

testnet/stacks-node/src/syncctl.rs
jcnelson: Addressed in 08db53b

src/net/p2p.rs
jcnelson: Addressed in 08db53b

jcnelson:
The failing sampled-genesis test is due to a bind error (address already in use on the CI server).

lgalabru (approved):
Looks good to me thank you @jcnelson!

Issue #2463, opened 2021-02-22
Emit "selected PoX recipient slot" details in `new_burn_block` events


kantai:
Right now, burn block events just report which addresses received rewards, but not which addresses had the "slots" for that block.

For example, if addresses A and B were chosen for block 1, addresses C and D were chosen for block 2, and then a miner commitment from block 1 ended up getting included in block 2, the stacks-node would report rewards for addresses A and B in both blocks.

With the improved RBF logic, this should happen less frequently, but we should also update the burn block events to report who claimed the actual slot -- that way we'd also be able to track when slots were "missed" (e.g., due to too-fast bitcoin blocks).

jcnelson:
Fixed in #2479

Issue #2446, opened 2021-02-15
Emit dropped mempool txs


zone117x:
Let event observers know when a mempool transaction has been dropped from the node, probably just a new event endpoint where the txid is passed?

(I thought we had an issue for this already, maybe we do and I couldn't find it, if so feel free to close this one)

jcnelson:
We can do this in the garbage-collection path in the relayer.

jcnelson:
This also applies to transactions that get dropped because they're too expensive or get rendered invalid (e.g. RBF'ed).

jcnelson:
This is now addressed in #2479

PR #2479, opened 2021-02-25
Feature: PoX Slot Events and Mempool Drop Tx Events

kantai:
This PR adds one new event type from dropped mempool txs (#2446), and adds a field to burn block events for the PoX reward slot holders (#2463).


docs/event-dispatcher.md
```diff
@@ -100,10 +100,22 @@ Example:
...
   "burn_amount": 12000
 }
 ```
 
+* `reward_recipients` is an array of all the rewards received during this burn block. It may
+  include recipients who did _not_ have reward slots during the block. This could happen if
+  a miner's commitment was included a block or two later than intended. Such commitments would
+  not be valid, but the reward recipient would still receive the burn `amt`.
+* `reward_slot_holders` is an array of the Bitcoin addresses that would validly receive
+  PoX commitments during this block. These addresses may not actually receive rewards during
```
jcnelson, line 116: Or, if miners simply don't mine in that block (e.g. fees are too high, etc.).

src/chainstate/stacks/miner.rs
```diff
@@ -7369,7 +7388,8 @@ pub mod test {
 
                 let coinbase_tx = make_coinbase(miner, tenure_id as usize);
 
-                let mut anchored_block = StacksBlockBuilder::build_anchored_block(chainstate, &sortdb.index_conn(), &mut mempool, &parent_tip, tip.total_burn, vrf_proof, Hash160([tenure_id as u8; 20]), &coinbase_tx, ExecutionCost::max_value()).unwrap();
+                let mut anchored_block = StacksBlockBuilder::build_anchored_block(chainstate, &sortdb.index_conn(), &mut mempool, &parent_tip, tip.total_burn, vrf_proof, Hash160([tenure_id as u8; 20]), &coinbase_tx, ExecutionCost::max_value(), None
```
jcnelson, line 7391: Wow, `cargo fmt` didn't split this into multiple lines?  Not a bug; just surprised at `cargo fmt`.

jcnelson (approved):
LGTM!  Thanks @kantai!  Should we require @zone117x sign off on this, so he can confirm the events are getting reported?

hstove:
Does this PR also mean that, for blocks with no miner commits, `new_burn_block` will be emitted with `reward_slot_holders`?

kantai:
> Does this PR also mean that, for blocks with no miner commits, new_burn_block will be emitted with reward_slot_holders?

Yep

docs/event-dispatcher.md
```diff
@@ -100,10 +100,22 @@ Example:
...
   ],
+  "reward_slot_holders": [
+    "1C56LYirKa3PFXFsvhSESgDy2acEHVAEt6",
+    "1C56LYirKa3PFXFsvhSESgDy2acEHVAEt6"
+  ],
   "burn_amount": 12000
 }
 ```
 
+* `reward_recipients` is an array of all the rewards received during this burn block. It may
```
lgalabru, line 111: I started working on similar in another branch a few weeks ago, and unless I missed something, if we want that list to be as exhaustive as possible, I think we'd need to tweak `LeaderBlockCommitOp::parse_from_tx` - ignoring a lot invalid block commit. I was thinking about producing some kind of `ConsensusReceipt` struct, that would help building a clear explanation of the consensus, and also returning the reasons for leader block commit being ignored.
Anyway this PR is way better than the approach I had in mind: it's live :).

lgalabru (approved):
LGTM!

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -204,6 +213,10 @@ impl EventObserver {
         self.send_payload(payload, PATH_MEMPOOL_TX_SUBMIT);
     }
 
+    fn send_dropped_mempool_txs(&self, payload: &serde_json::Value) {
+        self.send_payload(payload, PATH_MEMPOOL_TX_DROP);
```
zone117x, line 217: Note this sends a lot of `StaleGarbageCollect` payloads without any transactions. Not a blocker just a tiny inefficiency.

zone117x (approved):
LGTM.
Tested dropped mempool with the API in https://github.com/blockstack/stacks-blockchain-api/pull/471 and reward slot holders in https://github.com/blockstack/stacks-blockchain-api/pull/473

Note I was only able to test dropped txs for replace-by-fee and stale-garbage-collect, the others I was unable to replicate in mainnet.

testnet/stacks-node/src/event_dispatcher.rs
kantai: Noted -- will add a check to make sure its only sending payloads when they're non-empty.

PR #2480, opened 2021-02-25
Feature: add tx-sponsor? keyword to clarity

### LINKED ISSUE #1879: Proposal: enhance support for sponsored transaction
In order to fully support sponsored transaction schemes, I think smart contract developers should be able to get something like `tx-sponsor?`, in Clarity.
Along the same line, I think that sponsors should be able to set some post-conditions, so that they can protect themselves and make sure that they're getting what they want from the sponsorship.
Thought @jcnelson @kantai ?

pavitthrap:
## Description

Fixes #1879. 

Adds a keyword `tx-sponsor?` to clarity which is of type Optional Principal. If a transaction has a sponsor, this keyword will be equal to some(principal of sponsor), and none otherwise. 

### Notes:
I decided to avoid altering exising APIs. As an example, I did not modify `HttpRequestType::CallReadOnlyFunction` to take in a sponsor field. As such, the environment that a contract is executed in using this request type will have `tx-sponsor?` set to None. Similarly, I saw opportunities to pass sponsor through in `src/clarity.rs` (for the “launch” and “execute”), but I didn’t want to change the clarity CLI. 
Other calling locations from which sponsor is set to None: 
- In `src/chainstate/stacks/db/blocks.rs`, in `process_stacking_ops`
- In `src/vm/contexts.rs`, in `handle_poison_microblocks`, `stx_transfer`. 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

Yes

Yes, it adds a new keyword to clarity. Relevant details can be seen in `src/vm/docs/mod.rs`. 

## Testing information
Added tests for this change in `src/vm/tests/contracts.rs'.

## Checklist
- [x] `cargo test` passes


jcnelson:
Congratulations on your first significant PR @pavitthrap!

Just so you're aware, adding or changing Clarity's behavior is a _consensus-breaking_ change.  We cannot deploy it until the scheduled Stacks 2.1 upgrade, which will happen at Bitcoin block 700,000.  As such, can you please rebase your PR on `next`, which is the staging branch for backwards-incompatible changes?

src/vm/contexts.rs
```diff
@@ -742,23 +765,45 @@ impl<'a, 'b> Environment<'a, 'b> {
                 panic!("Tried to construct environment with bad caller {}", caller);
             }
         }
+        if let Some(ref sponsor) = sponsor {
+            if let Value::Optional(_) = sponsor {
+                let inner = Value::expect_optional(sponsor.clone());
```
jcnelson, line 770: Minor nitpick (not a bug) -- the `Value::expect_*` method will all print out panic messages if their arguments don't have the right types.  Up to you if you want to use them here instead.

src/chainstate/stacks/db/transactions.rs
```diff
@@ -874,11 +874,13 @@ impl StacksChainState {
...
                 let contract_id = contract_call.to_clarity_contract_id();
                 let cost_before = clarity_tx.cost_so_far();
+                let sponsor = tx.sponsor_address_as_optional_data();
 
                 let contract_call_resp = clarity_tx.run_contract_call(
                     &origin_account.principal,
-                    &contract_id,
+                    &sponsor,
                     &contract_call.function_name,
+                    &contract_id,
```
jcnelson, line 883: Not a bug, just suggested API consistency -- in Clarity and elsewhere in the codebase, the order is `contract_id, function_name, function_args`.  Can you update `run_contract_call()` to reflect this?  Thanks!

src/vm/contexts.rs
```diff
@@ -890,7 +937,17 @@ impl<'a, 'b> Environment<'a, 'b> {
             match res {
                 Ok(value) => {
                     let sender_principal = self.sender.clone().map(|v| v.expect_principal());
-                    handle_contract_call_special_cases(&mut self.global_context, sender_principal.as_ref(), contract_identifier, tx_name, &value)?;
+                    let sponsor_principal = match self.sponsor.clone() {
```
jcnelson, line 940: Minor nitpick -- you have a lot of dangling whitespace here.  I guess `cargo fmt` didn't catch it?

src/chainstate/stacks/db/transactions.rs
jcnelson: Also, can you make this the case in `execute_transaction()` below?

src/vm/tests/contracts.rs
```diff
@@ -486,6 +502,131 @@ fn test_contract_caller(owned_env: &mut OwnedEnvironment) {
...
+}
+
+fn test_sponsor_type_is_optional_principal(owned_env: &mut OwnedEnvironment) {
+    let p1 = execute("'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR");
+    let sponsor = Value::some(Value::Bool(true)).unwrap();
+    let env = owned_env.get_exec_environment(Some(p1.clone()), Some(sponsor.clone()));
+}
+
+#[test]
+#[should_panic(
```
jcnelson, line 517: I'm not sure I see why these are panicking failures?  Is it because the sender and sponsor are the same principal?

jcnelson: Ah nm, I see -- it should be `(some SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR)`.

src/vm/contexts.rs
pavitthrap: Good to know. I think I may leave it as is here because I pattern-matched right before calling.

src/vm/contexts.rs
pavitthrap: I ran `cargo fmt`, it didn't catch any issues here?

src/vm/mod.rs
```diff
@@ -335,7 +335,7 @@ fn eval_all(
                     global_context.execute(|global_context| {
                         let mut call_stack = CallStack::new();
                         let mut env = Environment::new(
-                            global_context, contract_context, &mut call_stack, Some(publisher.clone()), Some(publisher.clone()));
+                            global_context, contract_context, &mut call_stack, Some(publisher.clone()), Some(publisher.clone()), None);
```
kantai, line 338: `eval_all` is used for the initialization of a new contract (which should probably be pointed out somewhere in `eval_all`'s rustdocs) -- currently, the "tx-sender" for a contract initialization is the contract "publisher", which is always the origin account that broadcasted the contract publish tx. If you want `tx-sponsor?` to work during contract initialization as well, rather than passing a `None` here, I think we'd need to pass through the sponsor from the transaction.

kantai (commented):
This all looks great to me @pavitthrap! I had just one comment on `tx-sponsor?` in the context of contract publish transactions.

src/vm/variables.rs
```diff
@@ -50,11 +50,18 @@ pub fn lookup_reserved_variable(
...
-                    .ok_or(RuntimeErrorType::NoSenderInContext)?;
-                Ok(Some(sender))
+                    .ok_or(RuntimeErrorType::NoCallerInContext)?;
+                Ok(Some(caller))
+            }
+            NativeVariables::TxSponsor => {
+                let sponsor = env
+                    .sponsor
+                    .clone()
+                    .ok_or(RuntimeErrorType::NoSponsorInContext)?;
```
lgalabru, line 63: I have a double on my understanding here - are we throwing a runtime error when a sponsor can not be found?

lgalabru (commented):
Thank you @pavitthrap for taking on this feature, and congratulations for this 1st PR!
I had a few questions / feedbacks around the typing, curious how we feel about it. 
Usually, when we add new functions, we add test coverage in the type checker and runtime, I'm not sure we're currently fully covered.

Concerning the spec itself, are we 100% convinced by the fact that we don't want post-conditions for sponsors? I feel like it would be fair for sponsors to let them protect their interests.

hstove:
Thank you @pavitthrap , this is such a cool contribution!! Super helpful feature, and it's fantastic to see active development on Clarity. Keep it up!

src/vm/contexts.rs
```diff
@@ -57,12 +57,21 @@ pub const MAX_CONTEXT_DEPTH: u16 = 256;
...
+///   these different contexts can be mixed and matched (i.e., in a contract-call, you change
+///   contract context), a single "invocation" will end up creating multiple environment
+///   objects as context changes occur.
 pub struct Environment<'a, 'b> {
     pub global_context: &'a mut GlobalContext<'b>,
     pub contract_context: &'a ContractContext,
     pub call_stack: &'a mut CallStack,
     pub sender: Option<Value>,
     pub caller: Option<Value>,
+    pub sponsor: Option<Value>,
```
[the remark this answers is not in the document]
kantai: I think it should be `caller: PrincipalData`, but we'd want to confirm that it doesn't have any consensus-critical impacts.

src/vm/variables.rs
pavitthrap: Good point. I'll change this.

src/vm/contexts.rs
```diff
@@ -522,25 +537,32 @@ impl<'a> OwnedEnvironment<'a> {
         }
     }
 
+    // TODO: take in sponsor
+    // called from "launch" bran src/clarity.rs
     pub fn initialize_contract(
         &mut self,
         contract_identifier: QualifiedContractIdentifier,
         contract_content: &str,
+        sponsor: &OptionalData,
```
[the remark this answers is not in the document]
pavitthrap: I'll make another issue for refactoring `contexts.rs` to make caller/sender type `PrincipalData` instead of `Option<Value>`.

src/vm/mod.rs
```diff
@@ -234,13 +234,14 @@ pub fn is_reserved(name: &str) -> bool {
     }
 }
 
-/* This function evaluates a list of expressions, sharing a global context.
- * It returns the final evaluated result.
- */
+/// This function evaluates a list of expressions, sharing a global context.
+/// It returns the final evaluated result.!
+/// Used for the initialization of a new contract.
```
reedrosenbluth, line 239: nit: if you're refactoring to doc comments, perhaps use `/** ... */` instead of `///`, since this is a multiline comment

src/vm/contexts.rs
```diff
@@ -554,11 +574,12 @@ impl<'a> OwnedEnvironment<'a> {
     pub fn execute_transaction(
         &mut self,
         sender: Value,
+        sponsor: Option<PrincipalData>,
         contract_identifier: QualifiedContractIdentifier,
-        tx_name: &str,
         args: &[SymbolicExpression],
+        tx_name: &str,
```
reedrosenbluth, line 580: just curious, why did you reorder `args` and `tx_name`?

reedrosenbluth (approved):
a few nitpicky comments, but overall LGTM! Excellent work @pavitthrap!

src/vm/contexts.rs
```diff
@@ -890,7 +917,7 @@ impl<'a, 'b> Environment<'a, 'b> {
             match res {
                 Ok(value) => {
                     let sender_principal = self.sender.clone().map(|v| v.expect_principal());
-                    handle_contract_call_special_cases(&mut self.global_context, sender_principal.as_ref(), contract_identifier, tx_name, &value)?;
+                    handle_contract_call_special_cases(&mut self.global_context, sender_principal.as_ref(), self.sponsor.clone().as_ref(), contract_identifier, tx_name, &value)?;
```
kantai, line 920: Since you're using `as_ref()` here, you shouldn't need to `.clone()`

src/vm/mod.rs
```diff
@@ -255,7 +256,7 @@ fn eval_all(
             let try_define = global_context.execute(|context| {
                 let mut call_stack = CallStack::new();
                 let mut env = Environment::new(
-                    context, contract_context, &mut call_stack, Some(publisher.clone()), Some(publisher.clone()));
+                    context, contract_context, &mut call_stack, Some(publisher.clone()), Some(publisher.clone()), None);
```
kantai, line 259: These `None` values should take the passed `sponsor`, that way the contract initialization code can use `tx-sponsor?` as well.

src/vm/variables.rs
```diff
@@ -50,11 +50,18 @@ pub fn lookup_reserved_variable(
...
                     .clone()
-                    .ok_or(RuntimeErrorType::NoSenderInContext)?;
-                Ok(Some(sender))
+                    .ok_or(RuntimeErrorType::NoCallerInContext)?;
+                Ok(Some(caller))
+            }
+            NativeVariables::TxSponsor => {
+                let sponsor = match env.sponsor.clone() {
+                    None => Value::none(),
+                    Some(p) => Value::some(Value::Principal(p))?,
```
kantai, line 62: Nit: Here, the `Value::some(..)` constructor can throw an error in the event that the constructed object would be "too large" (or otherwise violate the Clarity object size limits). However, since we _know_ that a `Some(Principal)` is a valid Clarity object, we should treat such a failure as a panic (i.e., use `expect()`)

src/vm/tests/contracts.rs
```diff
@@ -486,6 +500,104 @@ fn test_contract_caller(owned_env: &mut OwnedEnvironment) {
...
+            &vec![],
+            false
+        )
+        .unwrap(),
+        Value::list_from(vec![sponsor.clone()]).unwrap()
+    );
+}
+
+fn test_tx_sponsor(owned_env: &mut OwnedEnvironment) {
+    let contract_a = "(define-read-only (get-sponsor)
```
kantai, line 551: You should also have a test that covers code executed during the contract initialization. For example:

```
(define-read-only (get-sponsor)
   (list tx-sponsor?))
(asserts! (is-eq tx-sponsor? (some 'SM2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQVX8X0G)) (err 1))
```

What this will do is that when the contract is initialized, it'll see if the contract publish transaction's tx sponsor was equal to `SM2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQVX8X0G`, and if not, will cause a runtime error during the contract initialization.

src/vm/mod.rs
pavitthrap: I was attempting to follow the style guide here: https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text 

Also there appears to be a lot of uses of the triple slash docs comment in our repo - lmk if I should change it anyway.

src/vm/contexts.rs
pavitthrap: Honestly probably happened by accident when I was using cLion's refactor tool - I'll switch it back.

src/vm/contexts.rs
```diff
@@ -966,16 +994,23 @@ impl<'a, 'b> Environment<'a, 'b> {
         &mut self,
         contract_identifier: QualifiedContractIdentifier,
         contract_content: &str,
+        sponsor: Option<PrincipalData>,
```
kantai, line 997: This function should just use `self.sponsor` here rather than taking it in as an argument -- the `Environment` struct should be constructed with the sponsor set to the correct value.

kantai: (And the same applies to `initialize_contract_from_ast`)

kantai (approved):
LGTM!

PR #2482, opened 2021-02-26
do NOT delete the parent anchored block if the microblock stream is too heavy

jcnelson:
This is an emergency hotfix that stops the node from accidentally deleting a valid parent block in the event that its descendant microblock stream is invalid for some reason.


kantai (approved):

psq (approved):
seems to do as advertised

PR #2486, opened 2021-02-27
Fix/clarity-cli-check-contract

jcnelson:
This adds an optional contract-id argument to `clarity-cli check`, so short-hand contract identifiers resolve correctly.


lgalabru (approved):
LGTM!

src/clarity.rs
```diff
@@ -449,6 +449,40 @@ struct InitialAllocation {
     amount: u64,
 }
 
+fn find_arg(args: &mut Vec<String>, argnames: &[&str], has_optarg: bool) -> Option<Option<String>> {
+    let mut idx = None;
+    let mut argval = None;
+    for (i, arg) in args.iter().enumerate() {
```
kantai, line 455: This search should use something like `.find()` or `.position()` (https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position).

Also, in what cases would this function return `Some(None)`? From looking at the control-flow, it seems like it isn't possible, in which case, this function should just return an `Option<String>` type.

jcnelson: It's possible -- it returns `Some(argument-value-option)`.

* `None` is the error condition -- it means that a switch was present and there was no argument given, but the switch required an argument.  For example, supplying `--contract_id` without the contract ID would return `None`.
* `Some(None)` is the success condition for switches without arguments -- it means that the switch was present and there was no argument given, and that's okay.  For example, supplying `--output_analysis` would return `Some(None)`.
* `Some(Some(_))` is the success condition for switches with an argument -- it means that the switch was present and there was an argument.  For example, supplying `--contract_id CONTRACT_ID` would return `Some(Some(CONTRACT_ID))`.

kantai: Can this be something like a `Result<Option<String>, String>` in that case?

* Err("Expected argument") - error condition
* ...

And this function should still use something like `.find()` or `.position()` to implement the search.

jcnelson: Addressed in 80d2bc2

kantai (approved):
LGTM!

PR #2488, opened 2021-03-02
feature: add memo field to stx-transfer clarity function

### LINKED ISSUE #2391: Add memo arg to stx-transfer? function
Currently Clarity's `stx-transfer?` function doesn't let you include a memo

### LINKED ISSUE #2485: all_burns in DEBG output in chronological order, not value-sorted order
Currently the all_burns in debug output of a stacks node is in value-sorted order:

Example: all_burns=[1,1,20000,20000,20000,20000]

This might be useful for determining median value. However, the median value is already indicated in the same output line. The requested functionality has the last 6 burns in chronological order so you can see which blocks you have missed:

e.g. all_burns=[20000,20000,1,1,20000,20000]

pavitthrap:
## Description
Added an optional memo field to the `stx-transfer?` clarity function (by _optional_ I mean you can choose to fully omit this argument when calling the function).

Fixes #2391.  

Also fixes #2485, which is a small DEBUG logging change.

### Notes
- I decided to make the memo input `buff(34)` instead of `option(buff(34))`. The reason for this is because memo was provided as a string directly through the struct `TransferStxOp` in `src/chainstate/stacks/db/blocks.rs`, so I wanted to keep things consistent with that. Moreover, for backwards compatibility, the memo field is completely optional to even provide, so making the type explicitly an option felt a little unnecessary. Let me know if anyone has any opinions on this. 
- In `special_stx_transfer`, I first check if there are 4 arguments. If there aren’t I then check if there are 3 arguments, which returns an error directly if there are not 3 arguments. Should I instead return an error message that the `stx-transfer?` function expects 4 arguments, instead of letting the `check_argument_count(3, args)?` return its error directly? Could be a source of confusion. 
- For some of the calls to `stx-transfer?` in the repo (such as in test functions), I passed in the memo field, but not for all the calls. I did this because the function can now take in either 3 or 4 arguments so I thought it would be good to have instances of both in our tests. Let me know if anyone thinks I should pass in the additional memo parameter to all the tests throughout the repo. 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

Yes

Yes, this adds an additional field to `stx-transfer?`. Refer to change in `src/vm/docs/mod.rs`. 

## Testing information
Run `cargo test`, there are various tests for this function in the code base.

For the debug logging change(issue #2485), I ran a stacks node locally and checked for a change in the DEBUG output for `all_burns`. Example of new output: 
`DEBG [1614734509.490227] [src/chainstate/burn/distribution.rs:289] [chains-coordinator] Burn sample, txid: 72389d8e5e512d72d4691db8ee98ada89559aeefe3bc72cb5d2f7e0144bb6997, most_recent_burn: 33332, median_burn: 1, all_burns: [33332, 1, 1, 33332, 1, 1]`


src/vm/clarity.rs
```diff
@@ -841,9 +841,14 @@ impl<'a, 'b> ClarityTransactionConnection<'a, 'b> {
         from: &PrincipalData,
         to: &PrincipalData,
         amount: u128,
+        memo: &ASCIIData,
```
zone117x, line 844: I think the memo in regular stx-transfer transactions is an arbitrary 34-byte array (e.g. not constrained to ASCII characters). This should be a buffer object rather than an ascii string.

zone117x:
I think we'd also need to a `memo` property to the `stx_transfer_event` data, otherwise I don't think the API and tooling will able to easily use this new feature
https://github.com/blockstack/stacks-blockchain/blob/fe81b98d81df289686140941a6d65ac49f2c63f4/src/chainstate/stacks/events.rs#L93

pavitthrap:
> I think we'd also need to a `memo` property to the `stx_transfer_event` data, otherwise I don't think the API and tooling will able to easily use this new feature
> 
> https://github.com/blockstack/stacks-blockchain/blob/fe81b98d81df289686140941a6d65ac49f2c63f4/src/chainstate/stacks/events.rs#L93

I added `memo` as a field in `STXTransferEventData`, so this should be covered.

src/vm/analysis/type_checker/natives/assets.rs
```diff
@@ -186,6 +190,41 @@ pub fn check_special_transfer_token(
...
+    args: &[SymbolicExpression],
+    context: &TypingContext,
+) -> TypeResult {
+    let memo_passed;
+    if let Ok(()) = check_argument_count(4, args) {
+        memo_passed = true;
+    } else {
+        check_argument_count(3, args)?;
+        memo_passed = false;
+    }
```
kantai, line 204: Style nit: rust's treatment of code blocks lets write these kinds of things without separating the declaration from the assignment:

```suggestion
    let memo_passed = let Ok(()) = check_argument_count(4, args) {
        true
    } else {
        check_argument_count(3, args)?;
        false
    };
```
```

File: src/vm/types/mod.rs [hunk 3]
```diff
@@ -342,7 +342,7 @@ impl SequenceData {
         macro_rules! drain_filter {
             ($data:expr, $seq_type:ident) => {
                 let mut i = 0;
-                while i != $data.data.len() {
+                while i < $data.data.len() {
```
kantai, line 345: This change shouldn't be necessary, right? The while loop can only ever either remove an element (decreasing len() by 1) or increment i, meaning that `i` should never cross `.len()`. I'm just a little wary of changing something like this (a loop condition deep in the type code) unless it's absolutely necessary.

kantai (commented):
These changes look great to me, thanks @pavitthrap, just had a couple of small nits.

src/vm/types/mod.rs
pavitthrap: Sure, I'll remove this.

kantai (approved):

zone117x:
This is great @pavitthrap!
Code changes LGTM. I think the changelog or similar needs updated with this change. That will be helpful for the API, for example, when implementing Stacks 2.1 upgrades.

zone117x (approved):

reedrosenbluth (approved):
Excellent work! One small comment but otherwise LGTM!

PR #2496, opened 2021-03-04
Fix/block push store tocttou (develop)

jcnelson:
Fix a TOCTTOU bug in the code that checks that the node doesn't yet have an anchored block (this is important now, because both the p2p thread and relayer thread can store new anchored blocks).


jcnelson:
Closing in favor of #2497

PR #2497, opened 2021-03-04
fix/block store push tocttou (master)

jcnelson:
Backport #2496 to master.

EDIT: Closing #2496 in favor of this one.  The original PR text follows.

Fix a TOCTTOU bug in the code that checks that the node doesn't yet have an anchored block (this is important now, because both the p2p thread and relayer thread can store new anchored blocks).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.0.9]
+
+This is a hotfix release for improved handling of arriving Stacks blocks through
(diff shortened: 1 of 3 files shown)


kantai:
Because this is against `master` directly, can you add a new changelog entry (`2.0.9`) for this? This update also won't require a new chainstate from `2.0.8`.

lgalabru (approved):
Looks good to me!
Any test that can be written to reproduce this issue?
I think the test failure can be fixed by including the changes from this commit https://github.com/blockstack/stacks-blockchain/pull/2457/commits/63d7497e888300d68f757e5f2b3dbd4ae47cc2c5

jcnelson:
> Because this is against master directly, can you add a new changelog entry (2.0.9) for this? This update also won't require a new chainstate from 2.0.8.

> I think the test failure can be fixed by including the changes from this commit 63d7497

Done and done.

kantai (approved):

PR #2498, opened 2021-03-04
feat: RPC endpoint /v2/traits/<contract-info>/<trait-info>

### LINKED ISSUE #2382: [Feature request] expose implemented traits via interface RPC API
I already have functionality in the Stacks Wallet for Web that allows users to natively transfer fungible tokens and see their balances. When the user goes to do a transfer, I fetch the interface for that contract, and then check if it has a function with the signature `(transfer (recipient principal) (amount uint))`. If it has this function, I allow the contract-call, add default post conditions, etc. It's pretty cool and all thanks to all the work already implemented in this repo and the API!

However, I don't fully know that the contract follows a trait. If I want to fetch the token's name, decimals, etc, I'd have to do a bunch of read-only calls. Or, I can create a wrapper contract that does `use-trait` and fetches that read-only info in the wrapper.

Ultimately, it would be awesome to expose the implemented traits via RPC. Wallets could fetch the interface and simply check if a "hard-coded" trait is implemented.

In `/v2/contracts/interface/SPXX/name`, expose in the JSON something like:

```js
{
  implemented_traits: [
    'SPXX.name.trait'
  ],
  // other existing fields
}
```

For example, I _could_ build a wrapper contract to fetch the info I need. What if I want to build a contract registry? All the stuff I need is already in Clarity (i.e. `use-trait`). However, if I was building a UI layer for the registry, it'd be best to first check that the trait is implemented. Otherwise, I'd have to make a contract call on my registry contract and wait to see if it was rejected.

Wrapper contracts, registry contracts, etc get us pretty much all the way there, but exposing via the interface would be better.

**Other context**

Friedger and I have started proposing SIPs for traits, which is a first step, and this would make traits even more powerful.

https://github.com/stacksgov/sips/pull/5 and https://github.com/stacksgov/sips/pull/3

pavitthrap:
Added a RPC endpoint to determine if a given trait is implemented in a contract, either explicitly or implicitly.
Fixes #2382 

## Description
For some background, a trait is explicitly implemented in a contract if the contract has `impl-trait <trait info>`. Implicit implementation means a contract implements the traits without the declaration above. 

### Notes 
- I made trait_id (type `TraitIdentifier`) a field in HttpRequestType::GetIsTraitImplemented, even though the GET request independently passes in a trait name, trait contract address, and trait contract name. I did this because I thought it would be cleaner to parse the API path at a higher level, and pass down consolidated information (in the form of the struct `TraitIdentifier`. However, this is not consistent with how the contract data is passed (which is passed separately as contract address and contract name), so let me know if I should avoid consolidating the trait info until the function `handle_get_is_trait_implemented` in `src/net/rpc.rs`. 
- For implicitly implemented traits, I have to check for trait compliance. However, since I am using a read only connection to the clarity database, I am not saving the results of this computation. If I were to save the computation, I would add the implicitly implemented trait by calling `contract_analysis.add_implemented_trait`, and save the updated contract analysis by calling `set_metadata`, which is a method of `ClarityDatabase`. Let me know if I should open a RW connection to the database instead, and commit this information. 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

List the APIs or describe the functionality that this PR breaks.
Workarounds for or expected timeline for deprecation

Want to update the RPC endpoint documentation [here](https://docs.blockstack.org/understand-stacks/stacks-blockchain-api#proxied-stacks-node-rpc-api-endpoints)

## Testing information
Ran `cargo test`.

### DIFF
--- a/src/net/http.rs
+++ b/src/net/http.rs
@@ -92,6 +92,8 @@ use std::time::SystemTime;
 use time;
 
 use chainstate::burn::ConsensusHash;
+use net::Error::ClarityError;
+use vm::types::{StandardPrincipalData, TraitIdentifier};
 
 lazy_static! {
     static ref PATH_GETINFO: Regex = Regex::new(r#"^/v2/info$"#).unwrap();
@@ -127,6 +129,11 @@ lazy_static! {
     static ref PATH_GET_CONTRACT_SRC: Regex = Regex::new(&format!(
         "^/v2/contracts/source/(?P<address>{})/(?P<contract>{})$",
         *STANDARD_PRINCIPAL_REGEX, *CONTRACT_NAME_REGEX
+    ))
+    .unwrap();
+    static ref PATH_GET_IS_TRAIT_IMPLEMENTED: Regex = Regex::new(&format!(
+        "^/v2/traits/(?P<address>{})/(?P<contract>{})/(?P<traitName>{})/(?P<traitContractAddr>{})/(?P<traitContractName>{})$",
+        *STANDARD_PRINCIPAL_REGEX, *CONTRACT_NAME_REGEX, *CLARITY_NAME_REGEX, *STANDARD_PRINCIPAL_REGEX, *CONTRACT_NAME_REGEX
     ))
     .unwrap();
     static ref PATH_GET_CONTRACT_ABI: Regex = Regex::new(&format!(
@@ -1505,6 +1512,11 @@ impl HttpRequestType {
             ),
             (
                 "GET",
+                &PATH_GET_IS_TRAIT_IMPLEMENTED,
+                &HttpRequestType::parse_get_is_trait_implemented,
+            ),
+            (
+                "GET",
                 &PATH_GET_CONTRACT_ABI,
                 &HttpRequestType::parse_get_contract_abi,
             ),
@@ -1868,6 +1880,45 @@ impl HttpRequestType {
                 HttpRequestType::GetContractSrc(preamble, addr, name, tip, with_proof)
             },
         )
+    }
+
+    fn parse_get_is_trait_implemented<R: Read>(
+        _protocol: &mut StacksHttp,
+        preamble: &HttpRequestPreamble,
+        captures: &Captures,
+        query: Option<&str>,
+        _fd: &mut R,
+    ) -> Result<HttpRequestType, net_error> {
+        let tip = HttpRequestType::get_chain_tip_query(query);
+        if preamble.get_content_length() != 0 {
+            return Err(net_error::DeserializeError(
+                "Invalid Http request: expected 0-length body".to_string(),
+            ));
+        }
+
+        let contract_addr = StacksAddress::from_string(&captures["address"]).ok_or_else(|| {
+            net_error::DeserializeError("Failed to parse contract address".into())
+        })?;
+        let contract_name = ContractName::try_from(captures["contract"].to_string())
+            .map_err(|_e| net_error::DeserializeError("Failed to parse contract name".into()))?;
+        let trait_name = ClarityName::try_from(captures["traitName"].to_string())
+            .map_err(|_e| net_error::DeserializeError("Failed to parse trait name".into()))?;
+        let trait_contract_addr = StacksAddress::from_string(&captures["traitContractAddr"])
+            .ok_or_else(|| net_error::DeserializeError("Failed to parse contract address".into()))?
+            .into();
+        let trait_contract_name = ContractName::try_from(captures["traitContractName"].to_string())
+            .map_err(|_e| {
+                net_error::DeserializeError("Failed to parse trait contract name".into())
+            })?;
+        let trait_id = TraitIdentifier::new(trait_contract_addr, trait_contract_name, trait_name);
+
+        Ok(HttpRequestType::GetIsTraitImplemented(
+            HttpRequestMetadata::from_preamble(preamble),
+            contract_addr,
+            contract_name,
+            trait_id,
+            tip,
+        ))
     }
 
     fn parse_getblock<R: Read>(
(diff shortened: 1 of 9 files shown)


kantai:
Thanks for this PR, @pavitthrap! One note before I take a closer look at the PR: this should be opened against `develop`. Generally speaking we follow something like a git-flow process for PRs, where features and other non-hotfix PRs should be based off of `develop`, and hotfixes are based off of `master`.

src/net/http.rs
```diff
@@ -129,6 +131,11 @@ lazy_static! {
         *STANDARD_PRINCIPAL_REGEX, *CONTRACT_NAME_REGEX
     ))
     .unwrap();
+    static ref PATH_GET_IS_TRAIT_IMPLEMENTED: Regex = Regex::new(&format!(
+        "^/v2/traits/(?P<address>{})/(?P<contract>{})/(?P<traitName>{})/(?P<traitContractAddr>{})/(?P<traitContractName>{})$",
```
kantai, line 135: For traits, they're usually reference like `<traitContractAddr>.<traitContractName>.<traitName>`, so it'd be more consistent for this path to match that order as well.

src/net/rpc.rs
```diff
@@ -1241,6 +1243,57 @@ impl ConversationHttp {
...
+            match chainstate.maybe_read_only_clarity_tx(&sortdb.index_conn(), tip, |clarity_tx| {
+                clarity_tx.with_clarity_db_readonly(|db| {
+                    let mut analysis = db.load_contract_analysis(&contract_identifier)?;
+                    if analysis.implemented_traits.contains(trait_id) {
+                        Some(GetIsTraitImplementedResponse {
+                            is_implemented: true,
+                        })
+                    } else {
+                        let trait_name = trait_id.name.to_string();
+                        let trait_definition = analysis.get_defined_trait(&trait_name)?;
```
kantai, line 1272: You should be able to use `&trait_id.name` directly here -- `ClarityName` implements `Deref<str>`, so rust can do automatic reference coercion.

kantai (commented):
This looks good to me, thanks @pavitthrap! I just had a few style nits. And since this PR adds a new RPC endpoint, you should update two places in the PR:

1. Add a brief changelog entry under the `[Unreleased]` section in `CHANGELOG.md`
2. Add an entry for the new endpoint to the OpenAPI spec `./docs/rpc/openapi.yaml`

src/net/rpc.rs
```diff
@@ -1241,6 +1243,57 @@ impl ConversationHttp {
...
+                        match analysis.check_trait_compliance(trait_id, trait_definition) {
+                            Err(_) => None,
```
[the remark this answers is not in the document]
pavitthrap: I changed the code to look like this.
For your second comments, if the contract analysis isn't found or the trait definition isn't found, a `None` is returned from this closure (I'm pretty sure).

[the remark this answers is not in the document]
kantai: Ah, I see. Yeah, that makes sense then.

kantai (approved):
LGTM! Just one comment on the endpoint's tag.

A note for the future: it'd be better not to do a force-push to update the PR once you've already opened the PR. It makes checking for resolution of comments harder because the reviewer can't see the specific changes that were made to address them.

pavitthrap:
> LGTM! Just one comment on the endpoint's tag.
> 
> A note for the future: it'd be better not to do a force-push to update the PR once you've already opened the PR. It makes checking for resolution of comments harder because the reviewer can't see the specific changes that were made to address them.

Ah got it. It seemed like the only way here since I needed to rebase to get the openAPI changes, but I'll avoid it in other scenarios.

src/net/rpc.rs
```diff
@@ -1333,6 +1335,54 @@ impl ConversationHttp {
...
+                clarity_tx.with_clarity_db_readonly(|db| {
+                    let mut analysis = db.load_contract_analysis(&contract_identifier)?;
+                    if analysis.implemented_traits.contains(trait_id) {
+                        Some(GetIsTraitImplementedResponse {
+                            is_implemented: true,
+                        })
+                    } else {
+                        let trait_definition = analysis.get_defined_trait(&trait_id.name)?;
+                        let is_implemented = analysis
+                            .check_trait_compliance(trait_id, trait_definition)
```
jcnelson, line 1365: Maybe this is a question for @kantai -- how expensive can this get?  Should this API call run with a given `ExecutionCost`?  I could imagine an attacker crafting an evil trait definition and contract that were specifically constructed to force this method to take a lot of CPU time.

testnet/stacks-node/src/tests/integrations.rs
```diff
@@ -642,6 +696,25 @@ fn integration_test_get_info() {
...
+
+                let path = format!("{}/v2/traits/{}/{}/{}/{}/{}", &http_origin, &contract_addr, "impl-trait-contract", "trait-1", &contract_addr, "get-info");
+                let res = client.get(&path).send().unwrap().json::<GetIsTraitImplementedResponse>().unwrap();
+                eprintln!("Test: GET {}", path);
+                assert!(res.is_implemented);
+
+                let path = format!("{}/v2/traits/{}/{}/{}/{}/{}", &http_origin, &contract_addr, "impl-trait-contract", "trait-2", &contract_addr, "get-info");
+                let res = client.get(&path).send().unwrap().json::<GetIsTraitImplementedResponse>().unwrap();
+                eprintln!("Test: GET {}", path);
+                assert!(res.is_implemented);
```
jcnelson, line 713: Is there a test you can add whereby `res.is_implemented` will be `false`?

src/net/rpc.rs
kantai: The trait compliance check is linear with respect to the size of the contract and the trait definition (and much faster than fetching either). This whole evaluation isn't exactly a "free" operation, but it's about the same as the account balance lookup API (it's 2 MARF reads: one for the trait definition and one for the contract in question), plus the read length of the two contracts (absolute maximum is two Stacks block sized reads). I'm not too worried about this becoming a too-expensive endpoint for that reason, it's actually doing much less work than the read-only call interface (which needs to do the contract fetch lookup before it begins any processing). If we wanted to add a bound on this, it isn't something it'd be obvious to add an ExecutionCost barrier to: during analysis, the data fetches are charged after fetch (because you cannot know how large the trait definition contract is until you've consulted the analysis db), so I think we'd need to place just a limit on the contract size that could be evaluated in this endpoint. But again, I don't think it's worthwhile -- an attack on this endpoint would need to make lots and lots of requests, and that's something that would need a proxy to protect against anyways.

jcnelson: Ah, okay.  Thanks for clarifying!  Sounds good to me.

jcnelson (approved):
LGTM!  Thanks @pavitthrap

PR #2501, opened 2021-03-06
Feature: keep invalidated block data

kantai:
This updates the two block truncation/deletion paths in the StacksChainState such that microblock is moved to an "invalid" table and block data is renamed to `.invalid`. This ensures that invalidated data is retained for debugging or recovery purposes.


jcnelson:
Let's rebase this off of #2502

jcnelson (approved):
Thanks @kantai!

kantai:
In the interest of reducing review load, I'm going to merge this with one review, as it's merging into a still-in-review PR anyways.

Issue #2483, opened 2021-02-26
[mining] microblock miner can produce a microblock stream that is too costly to be mined


jcnelson:
This appears to have been the trigger for #2482, even though the cause of #2482 was not this.  In general, the miner code shouldn't be doing this.

diwakergupta:
@jcnelson is this still an issue?

PR #2502, opened 2021-03-08
Fix/2483

jcnelson:
This is an attempt to address #2483.  I'm pretty sure it works, but I'm having a hard time getting a usable test-case.  Going to submit it as a draft for now.

EDIT: I think I have a good-enough test for this now.


src/chainstate/stacks/db/blocks.rs
```diff
@@ -1998,6 +2016,8 @@ impl StacksChainState {
         }
     }
 
+    /*
+    // NOTE: temporarily not used, so we can kee invalid data around for analysis.
```
kantai, line 2020: Would prefer to just delete rather than comment out -- the code will be stored in the git history in any case.

src/chainstate/stacks/miner.rs
```diff
@@ -460,7 +493,17 @@ impl<'a> StacksMicroblockBuilder<'a> {
 
 impl<'a> Drop for StacksMicroblockBuilder<'a> {
     fn drop(&mut self) {
-        debug!("Drop StacksMicroblockBuilder");
+        debug!(
```
kantai, line 496: This log entry would be a good candidate for a key-value pair log entry.

src/chainstate/stacks/miner.rs
```diff
@@ -287,6 +310,10 @@ impl<'a> StacksMicroblockBuilder<'a> {
             considered.insert(tx.txid());
         }
         if bytes_so_far + tx_len >= MAX_EPOCH_SIZE.into() {
+            warn!(
+                "Adding microblock tx {} would exceed epoch data size",
```
kantai, line 314: Should this be a warn or debug? It seems like this path is sort of expected behavior: a miner will try to mine until the threshold is met.

src/chainstate/stacks/db/blocks.rs
kantai: Actually -- would prefer if this path were updated like in #2501 (https://github.com/blockstack/stacks-blockchain/pull/2501/files#diff-36b251011e9d6834be636547f23dd28aa9b24b555331f637dd40cfd2a1552467R2018) --- the microblock data is still removed from the table, but it is inserted into a new invalid table as well.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -2321,10 +2337,14 @@ impl StacksChainState {
         )
         .map_err(Error::DBError)?;
 
+        /*
         // garbage-collect
+        // NOTE: temporarily disabled so we can keep invalid microblocks around for further
+        // analysis
```
kantai, line 2343: If `delete_microblock_data` is updated to move the microblock data to another table, this should also be uncommented.

src/chainstate/stacks/db/unconfirmed.rs
```diff
@@ -96,13 +98,15 @@ impl UnconfirmedState {
...
+        let cost_so_far = StacksChainState::get_stacks_block_anchored_cost(chainstate.db(), &tip)?
+            .ok_or(Error::NoSuchBlockError)?;
 
         Ok(UnconfirmedState {
             confirmed_chain_tip: tip,
             unconfirmed_chain_tip: unconfirmed_tip,
             clarity_inst: clarity_instance,
             mined_txs: UnconfirmedTxMap::new(),
-            cost_so_far: ExecutionCost::zero(),
+            cost_so_far: cost_so_far.clone(),
```
kantai, line 109: This clone can be dropped.

src/chainstate/stacks/db/blocks.rs
jcnelson: Copying invalid microblocks -- i.e. ones that could be invalid because they take up a lot of space -- to a different table isn't exactly a free operation.  I'm not sure what we'd be getting out of it?  IIRC sqlite doesn't do online defragmentation, so unless we explicitly `VACUUM` the database, deleting the blobs isn't going to improve query speed on the existing microblocks.  Putting them into a separate table doesn't really save us anything on query complexity, because you're just replacing a `WHERE orphaned = 1` with a different table name (so you're choosing between having a separate table index, or an index on `orphaned`).

src/chainstate/stacks/miner.rs
jcnelson: I'd prefer this not to be `debug` because I think it's useful information to miners -- it's a hint that they may want to slow down their microblock mining so they can get a chance to prioritize higher-fee transactions (instead of burning through their compute budget early while the network is congested).  Would you be okay with `info`?

kantai: Yeah, info works.

src/chainstate/stacks/db/blocks.rs
jcnelson: I'm pretty sure these codepaths only try to read microblocks where `orphaned = 0` already, but if not, then they should do so regardless of which way this goes (because an orphaned microblock can remain resident in the DB in-between two or more calls to `process_next_staging_block()`).

kantai: Right -- I agree that other codepaths _should_ only try to read where `orphaned = 0`, but it's a safer practice to make the more "pessimistic" change here.

jcnelson:
FINALLY!  Got a cost-overflow test that I think is sufficient to demonstrate that microblock mining now works correctly.  Please let me know what you think?

src/chainstate/stacks/miner.rs
kantai: I think this still should be changed to `info` -- `warn` indicates that something went wrong, whereas this path is expected behavior if blocks are full.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1585,6 +1587,584 @@ fn size_check_integration_test() {
     channel.stop_chains_coordinator();
 }
 
+#[test]
+#[ignore]
+fn size_overflow_unconfirmed_microblocks_integration_test() {
```
kantai, line 1592: Do you think these new tests should be added to the `bitcoin_tests` Github action? As long as they aren't flaky or add too much time to the action, we probably should.

kantai (approved):
This looks great to me! Just had a nitpick about a log level, and I think the new integration tests should be added (or decided not to add) to the `.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests` file.

src/chainstate/stacks/db/unconfirmed.rs
```diff
@@ -157,6 +162,8 @@ impl UnconfirmedState {
                 &self.confirmed_chain_tip,
             );
 
+            clarity_tx.reset_cost(cur_cost);
```
lgalabru, line 165: Top of mind, so sharing, but I think I remember having some issues a few months ago with this function, not doing what I was expecting, were you able to get the results you were looking for?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -1999,16 +2030,19 @@ impl StacksChainState {
...
-        // clear out the block data from staging
-        let clear_sql = "DELETE FROM staging_microblocks_data WHERE block_hash = ?1".to_string();
-        let clear_args = [&microblock_hash];
+        let args = [&microblock_hash];
 
-        tx.execute(&clear_sql, &clear_args)
-            .map_err(|e| Error::DBError(db_error::SqliteError(e)))?;
+        // copy into the invalidated_microblocks_data table
+        let copy_sql = "INSERT OR REPLACE INTO invalidated_microblocks_data SELECT * FROM staging_microblocks_data WHERE block_hash = ?1";
+        tx.execute(copy_sql, &args)?;
```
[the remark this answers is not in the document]
jcnelson: No; the point of this PR is to keep that data around indefinitely for the time being.  The problem was that it was getting deleted, and we almost ran into a case where we lost the chain's continuity.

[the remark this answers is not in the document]
lgalabru: I see. If we think that eviction in on our roadmap, should we be preparing the database schema with something like a `created_at` or `block_height`, so we can avoid a migration whenever we think we're ready?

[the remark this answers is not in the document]
jcnelson: I don't think we need to worry too much about eviction.  I'd rather keep all invalid blocks forever just in case (1) we accidentally mark a valid block invalid or (2) their absence somehow causes new nodes from bootstrapping.  To be clear, neither of these should be true, but we've already been bitten by (1) in production.

src/chainstate/stacks/db/unconfirmed.rs
jcnelson: I'm not sure I know what this is?  @kantai do you?

kantai: No -- `reset_cost` should work to set the cost tracker's current cost, and the implementation is pretty explicit.

lgalabru (approved):
LGTM!

Issue #2499, opened 2021-03-05
[networking] event ID leak in HTTP server error path on failure to get peer address


jcnelson:
Just jotting this down so we can remember to fix it.  The error-handling path in the HTTP server that handles the case where `peer_addr()` fails doesn't free the underlying event ID from the poller.

PR #2503, opened 2021-03-08
Fix/2499

jcnelson:
This is a fix for #2499.  It also contains `fix/clarity-cli-check-contract-id`, so let's wait to merge this until `fix/clarity-cli-check-contract-id` is merged.


src/net/poll.rs
```diff
@@ -270,14 +267,15 @@ impl NetworkState {
             self.event_map.len(),
             self.event_capacity
         );
+
+        if let Err(e) = sock.shutdown(Shutdown::Both) {
+            warn!("Failed to shut down socket {}: {:?}", event_id, &e);
+        }
+
```
kantai, line 274: Should failures in `sock.shutdown` or `poll.deregister` return an `Err` from this function?

lgalabru (approved):
LGTM!

Issue #2481, opened 2021-02-26
Replace by fee calculates new fee incorrectly


friedger:
**Describe the bug**
Will trying to replace a tx, it was rejected until I increased the fees drastically, like 10x

From conversation with @kantai :
```
message
    TX conflicts with sponsor/origin nonce in same fork with >= fee
new_fee
    37,800,000
new_txid
    910027a75cc93e954f4188a095bd93655c142acaecadc88180325c7ec957daa6
old_fee
    222,378,000
old_txid
    db6188708ceed6a957bf0a5ff9c4ffff23b4e0ec50ac510f8cec3cd4d0c27f14
```

jcnelson:
This sounds like there's some left-over code in the mempool that still treats the transaction fee as a fee _rate_, and incorrectly multiplies it by the transaction's byte length.

zone117x:
@jcnelson `fee_rate` is still used in (at least) variable names within the API. Should they all be `fee` instead?

jcnelson:
Sure, provided it doesn't lead to breakage.  It's called `fee` internally in the Stacks blockchain already.

pavitthrap:
Fixed in #2508

PR #2508, opened 2021-03-10
Fix/replace by fee incorrect

pavitthrap:
## Description

Altered logic in `mempool.rs` to avoid treating fee as a fee rate. 

Fixes issue #2481.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

No

## Testing information
Ran `cargo test`. 
Wrote test in `mempool.rs` that tries to replace a tx with a tx that has a smaller payload. It passes only if the fee is no longer treated as a fee rate. 

The test fails when ran on the tip of the `develop` branch. Here is an INFO message that prints before the test fails:
```
INFO [1615406406.228656] [src/core/mempool.rs:888] [core::mempool::mempool_db_test_rbf] TX conflicts with sponsor/origin nonce in same fork with >= fee, new_txid: ef76985e6e501e9f75f883c498ce2f8630e7906f6b055af6d67b070dceb1cdb0, old_txid: ac59200518dc3b5896ae36056f426ce7c188fac8dab1b5e7f766cf3343a7acca, origin_addr: SP1S7EZE5F588C3B6HE1M3MW95XA7QY145RGGFHBN, origin_nonce: 123, sponsor_addr: SP36V01MRY0NFV4MZZAZA621M1YMSNGNFN0KKE6CG, sponsor_nonce: 123, new_fee: 22320, old_fee: 24600 
```

### DIFF
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -130,7 +130,6 @@ pub struct MemPoolTxMetadata {
     pub txid: Txid,
     pub len: u64,
     pub tx_fee: u64,
-    pub estimated_fee: u64, // upper bound on what the fee to pay will be
     pub consensus_hash: ConsensusHash,
     pub block_header_hash: BlockHeaderHash,
     pub block_height: u64,
@@ -152,7 +151,6 @@ impl FromRow<MemPoolTxMetadata> for MemPoolTxMetadata {
         let txid = Txid::from_column(row, "txid")?;
         let consensus_hash = ConsensusHash::from_column(row, "consensus_hash")?;
         let block_header_hash = BlockHeaderHash::from_column(row, "block_header_hash")?;
-        let estimated_fee = u64::from_column(row, "estimated_fee")?;
         let tx_fee = u64::from_column(row, "tx_fee")?;
         let height = u64::from_column(row, "height")?;
         let len = u64::from_column(row, "length")?;
@@ -164,7 +162,6 @@ impl FromRow<MemPoolTxMetadata> for MemPoolTxMetadata {
 
         Ok(MemPoolTxMetadata {
             txid: txid,
-            estimated_fee: estimated_fee,
             tx_fee: tx_fee,
             len: len,
             consensus_hash: consensus_hash,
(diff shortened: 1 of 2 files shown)


src/core/mempool.rs
```diff
@@ -917,7 +916,7 @@ impl MemPoolDB {
             &u64_to_sql(origin_nonce)?,
             &sponsor_address.to_string(),
             &u64_to_sql(sponsor_nonce)?,
-            &u64_to_sql(estimated_fee)?,
+            &u64_to_sql(tx_fee)?,
```
jcnelson, line 919: Since this PR is headed to `develop` anyway, and `develop` already breaks the DB schema, can we just drop the `estimated_fee` column from the database?

src/core/mempool.rs
```diff
@@ -1174,7 +1167,10 @@ mod tests {
     use address::AddressHashMode;
     use burnchains::Address;
     use chainstate::burn::{BlockHeaderHash, VRFSeed};
+    // use chainstate::stacks::TransactionVersion::Testnet;
```
jcnelson, line 1170: We generally delete commented-out code :)  It lives on in the git history.

src/core/mempool.rs
```diff
@@ -2001,4 +1984,154 @@ mod tests {
...
+        let txid = tx.txid();
+        let mut tx_bytes = vec![];
+        tx.consensus_serialize(&mut tx_bytes).unwrap();
+        let expected_tx = tx.clone();
+        let tx_fee = tx.get_tx_fee();
+        let second_len = tx_bytes.len() as u64;
+
+        // these asserts are to ensure we are using the fee directly, not the fee rate
+        assert!(second_len < first_len);
+        assert!(second_len * tx_fee < first_len * old_tx_fee);
```
jcnelson, line 2094: Do you want to also add `assert!(tx_fee > old_tx_fee)`?

jcnelson (approved):
LGTM!  Just had a couple of minor comments.

kantai (approved):
This looks great to me! Just need to add an entry to the change log.

PR #2510, opened 2021-03-11
refactor: change typing for sender and caller

### LINKED ISSUE #1812: VM: Replace Environment::caller and Environment::sender with Option<Value::Principal(..)>
The `caller` and `sender` fields in `vm::context::Environment` are always principal types if they're not `None`.  Creating this as an issue because changing this potentially an invasive refactoring procedure.

pavitthrap:
## Description
Refactor types of sender/caller from either `Option<Value>` to `Option<PrincipalData` or `Value` to `PrincipalData`.

Fixes #1812 

### Notes 
- I used the method `expect_principal` multiple times in tests on `Value::Principal` values. It seems like we want to avoid use of this function (since it panics), but I thought it was fine since I used it in tests, where the input is definitely a principal type. 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

No

No 

## Testing information
Ran `cargo test`.


src/vm/functions/assets.rs
```diff
@@ -113,12 +113,7 @@ pub fn stx_transfer_consolidated(
         return clarity_ecode!(StxErrorCodes::SENDER_IS_RECIPIENT);
     }
 
-    if Some(from.clone())
-        != env
-            .sender
-            .as_ref()
-            .map(|pval| pval.clone().expect_principal())
-    {
+    if Some(from.clone()) != env.sender {
```
kantai, line 116: While we're modifying this line, I think we could avoid the clone here:

```rust
if Some(from) != env.sender.as_ref() {
...
```

kantai (approved):
Looks great to me! Just had one comment.

jcnelson (approved):
LGTM!  Thanks for taking the time to clean this up, @pavitthrap!

Issue #2455, opened 2021-02-16
[Feature request] possible improvements to local regtest setup


hstove:
After spending a bunch of time familiarizing myself with a local regtest setup with puppet chain, a miner, and a follower, I'm wondering if we can potentially update a few things:

- Support a configurable reward cycle length 
  - right now this is hard-coded for 5 blocks with a 1-block prepare phase. I believe I've heard this adds a lot of slowness while nodes sync, but I'm really not sure.
  - A configuration in the `toml` file would be great here
- Allow immediate mining on regtest
  - It seems like puppet-chain automatically mines 300 blocks - is this to give the Stacks miner funds? It takes quite a while for a node just to sync up to 300 blocks. Possibly related to the reward cycle length?
  - The nodes also seem to be doing the PoX `F*w` calculations during these blocks, even though there is no stacks chain tip yet
- Support `use_test_genesis_chainstate = true` in a regtest setup. 
  - It seems like this is ignored?

I'm really only doing this so I can test BNS stuff with Atlas/attachments, which are not supported on mocknet (please correct if I'm mistaken). After quite a few hours (mainly debugging and figuring things out), I haven't actually made progress on a fully successful end-to-end setup with a subdomain registrar, but hopefully my nodes will be all set in the morning.

lgalabru:
> I'm really only doing this so I can test BNS stuff with Atlas/attachments, which are not supported on mocknet (please correct if I'm mistaken).

I just pushed a fix in the PR `fix/atlas-fixes` and tested it, I was able to get attachments resolved + events from it.

reedrosenbluth:
> Support a configurable reward cycle length
From my testing, setting `pox_sync_sample_secs = 0` in your node's config toml drastically speeds up syncing

jcnelson:
I think this is all addressed now @hstove  -- please re-open if not.  Thanks!

PR #2513, opened 2021-03-15
Feat: allow neon node to use test_genesis_chainstate

reedrosenbluth:
Small PR enabling a neon node to boot up with the test genesis chainstate if `node.use_test_genesis_chainstate = true` in its config

issue: #2455

I don't think so... It shouldn't change the behavior of nodes that don't have `use_test_genesis_chainstate` set


testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -180,8 +180,18 @@ impl RunLoop {
...
+        // use test genesis data if use_test_genesis_chainstate is true in config
+        // otherwise default to USE_TEST_GENESIS_CHAINSTATE const
+        let use_test_genesis_data = if let Some(use_test_genesis_chainstate) =
+            self.config.node.use_test_genesis_chainstate
+        {
+            use_test_genesis_chainstate
+        } else {
+            USE_TEST_GENESIS_CHAINSTATE
+        };
+
```
pavitthrap, line 192: Here's a code snippet from `testnet/stacks-node/src/node.rs`:
```let use_test_genesis_data = if config.burnchain.mode == "mocknet" {
            // When running in mocknet mode allow the small test genesis chainstate data to be enabled.
            // First check env var, then config file, then use default.
            if env::var("BLOCKSTACK_USE_TEST_GENESIS_CHAINSTATE") == Ok("1".to_string()) {
                true
            } else if let Some(use_test_genesis_chainstate) =
                config.node.use_test_genesis_chainstate
            {
                use_test_genesis_chainstate
            } else {
                USE_TEST_GENESIS_CHAINSTATE
            }
        } else {
            USE_TEST_GENESIS_CHAINSTATE
        };
```

I'm wondering if you'd possibly want to (a) check the env var value here, and/or (b) create a helper function to determine the value of the var `use_test_genesis_data` and use it in both places.

kantai (approved):
This LGTM, but make sure that the tests all pass first.

pavitthrap (approved):
LGTM!

Issue #2394, opened 2021-01-29
Mempool transactions `accepted_at` different forks are not considered


kantai:
Transactions that received into the mempool at block `A`, but block `A` is "forked away", will not be considered for mining by the current miner implementation because those transactions will not appear in the miner's walk.

However, such transactions are viable for "replace-across-forks", such that if the transaction is rebroadcasted (not re-signed or fee upgraded), the miner will replace the existing tx in the mempool.

The rationale for this behavior is that the mempool admission checks that succeeded in one fork may have failed in the other. It isn't clear how large of a problem this behavior is currently (I think #2389 is a bit of confounding variable at the moment), and I think workarounds would be fairly simple _and_ effective, so this is probably a low priority change.

pavitthrap:
Fixed in `develop`

pavitthrap:
For #2393, I decided to implement the strict nonce-order with the following approach: (1) get a list of (consensus hash, block header hash) pairs that correspond to the last 256 blocks (or less) of the canonical chain tip, (2) select tx's in the mempool that have a consensus hash/ block header hash pair that correspond to one of the pairs in the list I created (used inner join for this), (3) order these tx by origin_nonce then timestamp, (4) query for the next "batch" of tx's to run the todo function on based on current nonce & timestamp, ordered by sponsor_nonce. 

However, I kind of am rethinking this approach based on this issue (which I was originally planning on implementing after I submitted this PR, but now I think I should do it at the same time). For a combined approach (doing nonce-order as well as including txs from different forks), I think I could do the following: (1) select all tx's with height > curr_height-MEMPOOL_DEPTH, and then (2), (3) correspond to (3), (4) from the steps above. Are there issues with how I select tx's for the second approach?

@jcnelson @kantai thoughts?

kantai:
> However, I kind of am rethinking this approach based on this issue (which I was originally planning on implementing after I submitted this PR - #2514, but now I think I should do it at the same time). For a combined approach (doing nonce-order as well as including txs from different forks), I think I could do the following: (1) select all tx's with height > curr_height-MEMPOOL_DEPTH, and then (2), (3) correspond to (3), (4) from the steps above. Are there issues with how I select tx's for the second approach?

I would be in favor of this approach -- I think basically what it would mean is:

1. Nodes accept transactions into their mempools based on their most up to date chain view.
2. Miners evaluate any transactions in their mempools.

[...]

PR #2514, opened 2021-03-15
Fix/mempool walk order

### LINKED ISSUE #2393: Mempool transactions should be evaluated in nonce-order, not walk-order
Currently, mempool transactions are evaluated in the "walk order" of the mempool, which is effectively a "reverse" order: transactions received later are attempted first.

This can impact chained transactions, because higher nonces will be tried first and fail, before the lower nonces are tried. In degenerate cases, this can lead to chained transactions only being included one-per-block, and will slow the rate that chained transactions are mined.

pavitthrap:
## Description
Altered `walk` function in `mempool.rs` so that 
1. transactions are executed in nonce order 
2. transactions in different forks are also considered

Changed the replace-by-fork logic to make it work in all cases.

Fixes #2393 and #2394. 

### Notes 
- Added pagination for getting all the transactions at a specific nonce & timestamp - I set the limit arbitrarily to 200, let me know if I should update this value. 

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

No

## Testing information
Ran `cargo test`. The test `net::server::test::test_http_getinfo` is failing, but it is also failing on the head of `develop` branch, so it isn't caused by the changes in this PR. 

Altered mempool walk tests, and added additional test for the replace-by-fork logic change.


src/core/mempool.rs
```diff
@@ -523,118 +618,105 @@ impl MemPoolDB {
...
+            max_tip_block_hash,
+            max_tip_height,
+            mut curr_tip_height,
+        ) = (
             tip_consensus_hash.clone(),
+            tip_consensus_hash.clone(),
+            tip_block_hash.clone(),
             tip_block_hash.clone(),
             tip_height,
+            None,
```
kantai, line 634: Something that could make this process a little more efficient (only a little, because the most costly part of the walk is the MARF lookups done in `get_index_tip_ancestor_conn`, which I think are avoided if a given block height doesn't have any transactions in the mempool) is to initialize `curr_tip_height` to `Some(tip_height - MEMPOOL_MAX_TRANSACTION_AGE)` (you'd want to check for underflow on that).

kantai (commented):
This looks great to me, thanks @pavitthrap! I only had one comment on the code, and it otherwise looks good.

However, because the mempool interacts so much with the miner implementation, this would be something good to have covered via a new (or modified) integration test in `neon_integrations` -- basically a test that would:

1. Broadcast tx with nonce=2 at block 1
2. Broadcast tx with nonce=3 at block 2
3. Broadcast tx with nonce=4 at block 3
4. Broadcast tx with nonce=1 at block 4

Then you could check that:

* At block 5, tx1 was included
* At block 6, tx2, tx3, and tx4 were included

src/core/mempool.rs
```diff
@@ -504,11 +538,72 @@ impl MemPoolDB {
         ))
     }
 
+    fn get_next_timestamp_at_next_block<E>(
+        &self,
+        chainstate: &mut StacksChainState,
+        mut curr_tip_consensus_hash: ConsensusHash,
+        mut curr_tip_block_hash: BlockHeaderHash,
+        mut curr_tip_height: Option<u64>,
```
lgalabru, line 546: Interesting and curious to hear about other opinions, I'm just sharing a practice that I personally prefer (coming from habits from other languages).
I usually try to avoid mutable arguments. In this case, the arguments are marked as mutable so that the inner loop can update them for next iterations, I think I would make some local, mutable copies of the arguments instead - in this case, it looks fine, since this inputs are owned, and not mutable references: thank you Rust :).

src/core/mempool.rs
```diff
@@ -504,11 +538,72 @@ impl MemPoolDB {
...
+                        next_height,
+                        next_timestamp,
+                    )));
+                }
+                MemPoolWalkResult::NoneAtHeight(
+                    next_consensus_hash,
+                    next_block_hash,
+                    next_height,
+                ) => {
+                    if std::env::var("STACKS_MEMPOOL_BAD_BEHAVIOR") == Ok("1".into()) {
```
lgalabru, line 583: Should this var being read once, before the beginning of the loop?

src/core/mempool.rs
```diff
@@ -399,34 +399,67 @@ impl MemPoolDB {
...
     }
 
-    fn walk(
+    /// Gets the lowest ancestor block(where `height > curr_tip_height && height <= max_tip_height`)
+    /// that is in the mempool of the "max" block (identified by `max_tip_block_hash` and
+    /// `max_tip_consensus_hash`). If found, this function returns a `MemPoolWalkResult::Chainstate`
+    /// enum along with identifying information about the ancestor block: its height
+    /// (`next_height`), its consensus hash, its block hash, and the earliest timestamp of the
+    /// transactions in it that are also in the mempool.
+    ///
```
lgalabru, line 408: I have some questions about the approach, but there could be some constraints that I'm not seeing - I'm not that familiar with this part of the codebase.
Instead of reversing the walk, I wonder if we could / should decouple things, and split the "mempool consumption" in 2 steps:
- Collect: we traverse the mempool using the most optimized approach, and collect in one big vector all the eligible transactions
- Sorting: we apply an algorithm, for ordering and packing transactions in blocks / microblocks, that can be smarter thanks to the knowledge of the whole load.
Thoughts?

src/core/mempool.rs
pavitthrap: I agree, I also try to avoid mutable arguments and prefer passing any mutated values back explicitly (which is what I did here) - only did it like this because these inputs are not references.

src/core/mempool.rs
pavitthrap: We only want to check for this when we are in the `NoneAtHeight` case, since this captures the behavior of the old walk algorithm. More details here (https://github.com/blockstack/stacks-blockchain/pull/2392)

src/core/mempool.rs
pavitthrap: This makes sense to me - @kantai @jcnelson any thoughts?

jcnelson: I'd want to be very careful about that, because a naive implementation could easily turn into a DDoS vector.  In general, I'd like to avoid loading _O(n)_ state into RAM if the network attacker controls _n_, as would be the case here.

I think it would be safer to structure the mempool's indexes so the miner could do something like "select each group of transactions that form a nonce chain that can potentially be attached to the chain state, and order the groups by total transaction fees in descending order."  This would let us stream chained transactions into RAM through an iterator, which takes only _O(1)_ RAM.  Then, the node would pick up and mine chained transactions that are contiguous with the current chain state, and would prioritize the transaction _groups_ with the highest total payouts (which also gives us the desirable property that a single very-high-fee transaction can get prioritized over a large chain with a lower total fee).  If the node runs out of block budget while mining, it just moves onto the next-highest-fee group and tries to go from there.

What do y'all think?

jcnelson: But, if you are looking for refactoring advice, one thing you might consider doing is turning the code that finds the ancestor consensus hash and block hash from the next block height (by using the chain tip's consensus hash and block hash, and the indexed connection to the sortition DB) into its own method.  This would be the code in lines 434-515.  Up to you though.

jcnelson (commented):
This overall looks good to me @pavitthrap, but let's get some clarity on what `walk_forwards()` should be doing here (waiting for @kantai's feedback).

src/core/mempool.rs
kantai: > I'd want to be very careful about that, because a naive implementation could easily turn into a DDoS vector. In general, I'd like to avoid loading O(n) state into RAM if the network attacker controls n, as would be the case here.

I definitely agree with this. However, I don't think that I agree with the next paragraph. Thinking ahead, I don't think we should try to make too opinionated a choice about mempool iteration, but rather provide an interface for the miner to iterate however they want -- providing different options for "sorting" does that without trying to make miner's choices for them (i.e., trying to compute fee-weighted transaction sets).

In thinking about this more, I think that this PR doesn't really solve issue #2393: providing the ability to iterate transactions in nonce-order, rather than walk-order. This PR just reverses the walk which we expect to help alleviate the problem. If we wanted to apply some other kind of sort in the traversal, I'm not sure how to do that iteratively with the way the mempool is currently structured -- I suspect we'd need to do something like construct the full set of blocks in the current fork (up to the mempool age limit) and use that to filter the SQLite query. I do not know how well SQLite would perform with a large set query like that.

jcnelson: > Thinking ahead, I don't think we should try to make too opinionated a choice about mempool iteration, but rather provide an interface for the miner to iterate however they want -- providing different options for "sorting" does that without trying to make miner's choices for them (i.e., trying to compute fee-weighted transaction sets).

Yes, agreed!  But, the issue this PR addresses _is_ an opinionated walk order (it just happens to be more sensible than the current walk order).  Are you saying you think that providing an interface for walk orders should be done in this PR?

> I suspect we'd need to do something like construct the full set of blocks in the current fork (up to the mempool age limit) and use that to filter the SQLite query. I do not know how well SQLite would perform with a large set query like that.

What might a "transaction selector" need?  I think it really just needs the following API functions:

* Get the "best" chain tip the Stacks node knows about.  This is _O(1)_.
* Get all block header hashes (i.e. consensus hashes and block hashes) on the fork from which a given "best" chain tip descends, down to the lowest block we keep transactions for.  This is _O(1)_ time and space to gather (_O(n log n)_ for _n_ blocks if we're counting the index searches, but _n_ is _O(1)_ due to the garbage collection limit).
* Get the newest timestamp of a transaction on a given block header hash history (this is at worst a traversal of the header hash history, but that's still _O(1)_ time and space).
* Find all transaction IDs and fees in a chain tip whose transactions have a given origin address, a nonce in a given range, and a "min-age" timestamp.  This is _O(1) time and space to gather (_O(n log n)_ for _n_ transactions from the same origin, including index searches, but _n_ is _O(1)_ due to the chaining depth limit).

Once you have an interface for these things, I think you can build any sort of mempool walker logic.  Moreover, you could easily expose this via the RPC interface, so a downstream miner program could just iterate through transactions it is interested in (each RPC call is _O(1)_ time and space, and also, the node could gate or rate-limit them if they became too burdensome).  Maybe there's room here to further constrain the kinds of transactions you'd want to expose -- e.g. filtering transactions by type variant or by minimum fee or maximum estimated size or cost, and so on.

I don't think the API calls need to be executed within a single Sqlite transaction -- I think you can get the necessary concurrency control in other ways.  You could guard against an intermittent chain reorg changing your view of the mempool by just checking your best chain tip before and after you produce a block (i.e. optimistic concurrency control), and you could guard against new transactions arriving and potentially invalidating your view by selecting transactions no newer than a given timestamp (effectively giving you read snapshot consistency).  In order for the latter to work correctly, we'd need to also alter the RBF logic in the mempool so it doesn't actually replace the old transaction, but just stores both the old and new transactions (existing code paths that query mempool transactions would be updated to always just take the newest version of a transaction, given an origin and a nonce).  This way, an RBF won't invalidate your mempool view.

What do y'all think?

src/core/mempool.rs
```diff
@@ -675,18 +521,91 @@ impl MemPoolDB {
...
         block_header_hash: &BlockHeaderHash,
+    ) -> Result<usize, db_error> {
+        let sql = "SELECT * FROM mempool WHERE consensus_hash = ?1 AND block_header_hash = ?2";
+        let args: &[&dyn ToSql] = &[consensus_hash, block_header_hash];
+        let rows = query_rows::<MemPoolTxInfo, _>(conn, &sql, args)?;
+        Ok(rows.len())
+    }
+
+    /// Get the next nonce/timestamp after this nonce that occurs in the height range.
+    pub fn get_next_nonce(
```
jcnelson, line 538: Minor style nit -- because this returns both a nonce and timestamp, this method name is misleading.  Can you call it something like `get_next_nonce_and_timestamp` or something similar?

src/core/mempool.rs
```diff
@@ -675,18 +521,91 @@ impl MemPoolDB {
...
+    pub fn get_next_timestamp_at_nonce(
+        conn: &DBConn,
+        min_height: Option<u64>,
+        max_height: u64,
         timestamp: u64,
-    ) -> Result<Option<u64>, db_error> {
-        let sql = "SELECT accept_time FROM mempool WHERE accept_time > ?1 AND consensus_hash = ?2 AND block_header_hash = ?3 ORDER BY accept_time ASC LIMIT 1";
-        let args: &[&dyn ToSql] = &[&u64_to_sql(timestamp)?, consensus_hash, block_header_hash];
+        nonce: u64,
+    ) -> Result<Option<(u64)>, db_error> {
```
jcnelson, line 566: Minor style nit: `Result<Option<u64>, db_error>`.  No need for `(u64)`.

src/core/mempool.rs
```diff
@@ -675,18 +521,91 @@ impl MemPoolDB {
...
+        nonce: u64,
+        timestamp: u64,
+    ) -> Result<Vec<MemPoolTxInfo>, db_error> {
+        let min_height_sql_arg = match min_height {
+            None => -1,
+            Some(h) => u64_to_sql(h)?,
+        };
+
+        let sql = "SELECT * FROM mempool WHERE origin_nonce = ?1 AND accept_time = ?2 AND \
+        height > ?3 AND height <= ?4 ORDER BY sponsor_nonce ASC";
```
jcnelson, line 598: What happens if there are a _lot_ of transactions with the same nonce and timestamp?  Should there be some kind of pagination logic here to prevent this from becoming a DDoS vector?  If not, then perhaps some kind of upstream insert rate-limiting should be in place to prevent this list from getting too large.

src/util/db.rs
```diff
@@ -169,6 +169,17 @@ impl FromRow<u64> for u64 {
     }
 }
 
+impl FromRow<(u64, u64)> for (u64, u64) {
```
jcnelson, line 172: If this is only ever used in the mempool, can you move it to `mempool.rs`?  From a code organization standpoint, `db.rs` is used for implementations that are used by multiple different databases; the `FromRow` / `FromColumn` implementations for subsystem-specific structs (which `(u64, u64)` appears to be) go into the modules that use them.

testnet/stacks-node/src/tests/integrations.rs
```diff
@@ -1025,6 +1025,172 @@ fn contract_stx_transfer() {
...
+                }
+                3 => {
+                    assert_eq!(chain_tip.metadata.block_height, 4);
+                    assert_eq!(chain_tip.block.txs.len(), 1);
+                }
+                4 => {
+                    assert_eq!(chain_tip.metadata.block_height, 5);
+                    assert_eq!(chain_tip.block.txs.len(), 2);
+
+                    // check that 1000 stx _was_ transfered to the contract principal
```
jcnelson, line 1136: I'm confused by this test.  I would have thought that all four transactions would have been mined into the same block, even though they were submitted out of order?  Why is it the case that the first STX-transfer (the one sent in round 4) is mined in a separate block from the three others?

src/core/mempool.rs
```diff
@@ -255,31 +256,36 @@ impl<'a> MemPoolTx<'a> {
...
 
-        let height_result = chainstate
-            .with_clarity_marf(|marf| marf.get_block_height_of(&index_block, &admitter_block));
-        match height_result {
-            Ok(x) => {
-                eprintln!("{} from {} => {:?}", &index_block, &admitter_block, x);
-                Ok(x.is_some())
-            }
-            Err(x) => Err(db_error::IndexError(x)),
+        let height_of_first_with_second_tip = chainstate
```
jcnelson, line 276: I'd prefer that you didn't use `with_clarity_marf()` for this, and instead used `get_index_tip_ancestor_conn()` and used that instead.  The reason is that `with_clarity_marf()` requires a _mutable_ reference to a `StacksChainState` instance, whereas `get_index_tip_ancestor_conn()` does not.  We're trying to keep the number of places where a `&mut StacksChainState` is needed outside of the relayer thread to a minimum, and this code path is executed through the networking thread.

src/core/mempool.rs
pavitthrap: Yeah, I can definitely add some type of pagination logic.

jcnelson:
This is going to be merged for 2.0.12 -- let's hold off on merging to `develop` until 2.0.11 is out the door. Thanks!

src/chainstate/stacks/db/headers.rs
```diff
@@ -324,6 +307,15 @@ impl StacksChainState {
         Ok(ancestors)
     }
 
+    /// Get an ancestor's height given the tip index hash and the ancestor index hash
+    pub fn get_block_height_of_ancestor(
```
kantai, line 311: I think the callers of this method should be able to just call `conn.get_ancestor_block_height`, right? That way, we wouldn't need to add this static function.

src/core/mempool.rs
```diff
@@ -400,244 +391,85 @@ impl MemPoolDB {
...
-            tip_height,
-        );
-
-        debug!(
-            "Begin scanning transaction mempool at {}/{} height={}",
-            &tip_consensus_hash, &tip_block_hash, tip_height
-        );
+        // Want to consider transactions with
+        // height > max(0, `tip_height - MEMPOOL_MAX_TRANSACTION_AGE`) - 1
+        let min_height = match tip_height.checked_sub(MEMPOOL_MAX_TRANSACTION_AGE) {
```
kantai, line 410: Does this make more sense as:

```rust
let min_height = tip_height.checked_sub(MEMPOOL_MAX_TRANSACTION_AGE + 1);
```

src/core/mempool.rs
```diff
@@ -400,244 +391,85 @@ impl MemPoolDB {
...
-                                tip_height = next_height;
-                            }
-                        }
-                        MemPoolWalkResult::Done => {
-                            return Ok(());
-                        }
-                    }
-                },
-            };
+        let (mut next_nonce, mut next_timestamp) = match MemPoolDB::get_next_nonce_and_timestamp(
```
kantai, line 417: My understanding of why we previously were using the timestamp in iteration was as a way to do some paging -- with the new structure of the mempool with explicit `cur_page` parameters, I don't think we need to use the timestamp anymore, instead we could just iterate using the nonce and the cur_page parameter (still including the timestamp in the sort though). This would simplify the loop and also reduce the number of SQLite queries.

pavitthrap: Oh ok, I thought we were using the timestamp for a different reason. I'll simplify the logic here then.

kantai (approved):
LGTM, thanks for bearing with all the changes @pavitthrap!

Note: we should wait to merge until `2.0.11` release has finished.

src/core/mempool.rs
```diff
@@ -195,6 +196,24 @@ impl FromRow<MemPoolTxInfo> for MemPoolTxInfo {
...
+        let t1: i64 = row.get_unwrap(0);
+        let t2: i64 = row.get_unwrap(1);
+        if t1 < 0 || t2 < 0 {
+            return Err(db_error::ParseError);
+        }
+        Ok((t1 as u64, t2 as u64))
+    }
+}
+
+pub fn u32_to_sql(x: u32) -> Result<i32, Error> {
```
jcnelson, line 210: This should not be necessary.  `ToSql` is already implemented for `i32` and `u32`.

jcnelson (approved):
LGTM! Thanks @pavitthrap!

jcnelson:
Just a reminder: please don't merge until 2.0.11 is released.

PR #2516, opened 2021-03-17
fix: fixed block attachability logic

### LINKED ISSUE #2212: Chainstate: don't mark a block as attachable if we either don't have its parent, or haven't processed its parent
Following discussion on Discord: https://discord.com/channels/621759717756370964/730814571517968436/791052952524816482

Copying notes from @kantai and @jcnelson 

> Right around here, attachable is computed by counting row results from: https://github.com/blockstack/stacks-blockchain/blob/master/src/chainstate/stacks/db/blocks.rs#L1529
> 
> but if the parent isn't in the staging_blocks table at all, `rows.len() == 0`
> so the child would get marked as attachable
> 
> ah, so it's a bug
> but probably not a show-stopper -- all it would mean is the node would try to process the child, and then (correctly) fail to do so because it can't find its parent
> but it does mean that the node will keep trying to process it, ad infinitum
> our unit tests already cover random block arrivals pretty extensively (basically, every miner.rs test is re-tested with blocks arriving in random order to verify that all blocks get processed), so it would be surprising to me to see a bug like this that was a show-stopper
>

pavitthrap:
## Description
Currently when we check if a block is attachable, we check if there exists an unprocessed parent. This PR additionally checks that the block's parent has been received. 

Fixes #2212.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

## Testing information
Ran `cargo test`. The test `chainstate::stacks::miner::test::test_mempool_mining_heuristics` is failing, but it is also failing on the head of `develop` branch, so it isn't caused by the changes in this PR.

### DIFF
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -1545,18 +1545,28 @@ impl StacksChainState {
 
         let attachable = {
             // if this block has an unprocessed staging parent, then it's not attachable until its parent is.
-            let has_parent_sql = "SELECT anchored_block_hash FROM staging_blocks WHERE anchored_block_hash = ?1 AND consensus_hash = ?2 AND processed = 0 AND orphaned = 0 LIMIT 1".to_string();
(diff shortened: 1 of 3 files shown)


lgalabru (approved):
Looks good to me, thank you @pavitthrap!

reedrosenbluth (approved):

jcnelson (approved):
LGTM!

Issue #2491, opened 2021-03-02
`stacks-node` appears to stall during bootup in `develop`


kantai:
This impacted the Xenon miner running an `atlas+develop` branch: the node appeared to stall during bootup due to "nack"s received from the miner's bootstrap node. The miner was able to be nudged out of this state using the chain pusher tool, but there's likely a bug somewhere here that needs to be addressed.

jcnelson:
I think I've figured out why this is -- the relayer isn't waking up the chains coordinator when it receives a _confirmed_ microblock stream, and nothing else.  This happens more often than not, and really should be the cause of a wake-up event.  The current code erroneously uses either the arrival of a Stacks block or an _unconfirmed_ microblock, but ignores the arrival of _confirmed_ microblocks.  Will send a PR once my testnet node finishes sync'ing.

jcnelson:
Yup, that was the problem.  Confirmed that my testnet node continuously streams blocks and microblocks with the fix.

![blocks-and-microblocks](https://user-images.githubusercontent.com/459947/110670019-dd38ea80-81c4-11eb-8a1d-e250ca9d7b66.png)

jcnelson:
PR sent

PR #2520, opened 2021-03-18
Fix/2518 and 2491

### LINKED ISSUE #2518: [p2p] downloader skips some microblock streams while downloading
There's a bug in the microblock download logic that can cause it to skip microblock streams.  It only appears to manifest if there's a large range of Stacks blocks that don't have microblock stream tails.

### LINKED ISSUE #2519: [p2p] don't report microblock streams that are orphaned in blockinv replies
We're not omitting microblock streams that got orphaned.  Oops.

jcnelson:
This PR fixes #2518 and #2491, both of which are hindering nodes from bootstrapping (2518 is far more serious than 2491).  It also fixes #2519, and temporarily makes it so a node reporting a microblock stream that turns out to be missing is not a ban-able offense (until which point we can roll this upgrade out).


kantai (approved):
This looks great to me!

Question about release scheduling (and somewhat this PR) -- should we open a patch of this against master for a hotfix? I think this qualifies as a low priority hotfix: spinning up a node right now will not work without this fix applied.

lgalabru (approved):
Looks good to me, thank you @jcnelson. Any tests that we can add to confirm that expectations are met?

PR #2521, opened 2021-03-18
fix #2517

### LINKED ISSUE #2517: Stacking contract reference: Call out unsupported bitcoin address formats
The [stacking contract refs](https://github.com/blockstack/stacks-blockchain/blob/master/src/vm/docs/contracts.rs#L41) don't explicitly call out that native segwit addresses are not supported. It might be worth adding a brief note. Thoughts?

jcnelson:
Update the PoX documentation to explicitly state that we do NOT take native Segwit addresses at this time.


agraebe (approved):
🙏

kantai (approved):

lgalabru (approved):

PR #2525, opened 2021-03-19
Fix/blocks.rs

pavitthrap:
## Description
-  resolved bug in blocks.rs: previously didn't check if the sponsor could afford a tx fee if the payload was a token transfer
- made some style changes in download.rs

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

## Testing information
Ran `cargo test`. Added test to check for bug.

### DIFF
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -4706,13 +4706,13 @@ impl StacksChainState {
     ) -> Result<Vec<StacksMicroblock>, Error> {
         // NOTE: since we got the microblocks from staging, where their signatures were already
         // validated, we don't need to validate them again.
-        let (microblock_terminus, _) = match StacksChainState::validate_parent_microblock_stream(
+        let microblock_terminus = match StacksChainState::validate_parent_microblock_stream(
             &parent_block_header_info.anchored_header,
             &block.header,
             &next_microblocks,
             false,
         ) {
-            Some((terminus, poison_opt)) => (terminus, poison_opt),
+            Some((terminus, _)) => terminus,
             None => {
                 debug!(
                     "Stopping at block {}/{} -- discontiguous header stream",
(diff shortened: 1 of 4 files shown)


src/net/download.rs
```diff
@@ -1797,18 +1794,22 @@ impl PeerNetwork {
...
-                        next_block_sortition_height = next_block_sortition_height
-                            + (network.burnchain.pox_constants.reward_cycle_length as u64);
-                    }
-                    if downloader.microblocks_to_try.len() == 0 {
-                        next_microblock_sortition_height = next_microblock_sortition_height
-                            + (network.burnchain.pox_constants.reward_cycle_length as u64);
-                    }
-
-                    debug!("{:?}: Pessimistically increase block and microblock sortition heights to ({},{})", &network.local_peer, next_block_sortition_height, next_microblock_sortition_height);
+                // nothing in this range, so advance sortition range to try for next time
```
jcnelson, line 1797: Technically, this comment should go under both of these `if`-statements, since the logic here is meant to advance the pointers into the chainstate at which to begin the next scan for available blocks and microblock streams.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5385,6 +5385,19 @@ impl StacksChainState {
                             .get_available_balance_at_burn_block(block_height),
                     ));
                 }
+
+                // if the payer for the tx is different from owner, check if they can afford fee
+                if origin != payer {
```
jcnelson, line 5390: Great catch on this!  Do we have a bunch of un-mineable transactions in the mempool that need to be cleared out?

jcnelson (approved):
LGTM! Thanks for catching this!

kantai (approved):
This looks good to me!

src/chainstate/stacks/db/blocks.rs
pavitthrap: I looked at pending transactions in the Stacks explorer and the oldest pending transaction is from 3 days ago, so I *think* we should be fine on this?

PR #2527, opened 2021-03-22
Implement #1641 rpc call public functions as read-only - take 2

psq:
second attempt in fixing #1641 to allow calling any function as read-only.  the call will fail if any write happened during the call.

For example, this allows calling public functions that call use `contract-call?`, which can not be used in a read-only function, even if the called contract function itself is read-only.

### DIFF
--- a/src/net/rpc.rs
+++ b/src/net/rpc.rs
@@ -89,10 +89,12 @@ use util::hash::{hex_bytes, to_hex};
 use vm::database::clarity_store::make_contract_hash_key;
 use vm::types::TraitIdentifier;
 use vm::{
+    analysis::errors::CheckErrors,
     costs::{ExecutionCost, LimitedCostTracker},
     database::{
         clarity_store::ContractCommitment, ClarityDatabase, ClaritySerializable, STXBalance,
     },
+    errors::Error::Unchecked,
     errors::Error as ClarityRuntimeError,
     errors::InterpreterError,
     types::{PrincipalData, QualifiedContractIdentifier, StandardPrincipalData},
@@ -1215,7 +1217,7 @@ impl ConversationHttp {
                     })?;
 
                 clarity_tx.with_readonly_clarity_env(mainnet, sender.clone(), cost_track, |env| {
-                    env.execute_contract(&contract_identifier, function.as_str(), &args, true)
+                    env.execute_contract(&contract_identifier, function.as_str(), &args, false)
                 })
             });
 
@@ -1228,14 +1230,26 @@ impl ConversationHttp {
                     cause: None,
                 },
             ),
-            Ok(Some(Err(e))) => HttpResponseType::CallReadOnlyFunction(
-                response_metadata,
-                CallReadOnlyResponse {
-                    okay: false,
-                    result: None,
-                    cause: Some(e.to_string()),
+            Ok(Some(Err(e))) => match e {
+                Unchecked(CheckErrors::CostBalanceExceeded(actual_cost, _)) if actual_cost.write_count > 0 => {
+                    HttpResponseType::CallReadOnlyFunction (
+                        response_metadata,
+                        CallReadOnlyResponse {
+                            okay: false,
+                            result: None,
+                            cause: Some("NotReadOnly".to_string()),
+                        },
+                    )
                 },
-            ),
+                _ => HttpResponseType::CallReadOnlyFunction (
+                    response_metadata,
+                    CallReadOnlyResponse {
+                        okay: false,
+                        result: None,
+                        cause: Some(e.to_string()),
+                    },
+                )
+            },
             Ok(None) | Err(_) => {
                 HttpResponseType::NotFound(response_metadata, "Chain tip not found".into())
             }
@@ -2867,7 +2881,7 @@ mod test {
         (define-public (set-bar (x int) (y int))
           (begin (var-set bar (/ x y)) (ok (var-get bar))))
         (define-public (add-unit)
-          (begin 
+          (begin
             (map-set unit-map { account: tx-sender } { units: 1 } )
             (ok 1)))
         (begin
--- a/src/net/rpc.rs
+++ b/src/net/rpc.rs
@@ -1202,15 +1202,15 @@ impl ConversationHttp {
             .map(|x| SymbolicExpression::atom_value(x.clone()))
             .collect();
         let mainnet = chainstate.mainnet;
+        let mut cost_limit = options.read_only_call_limit.clone();
+        cost_limit.write_length = 0;
+        cost_limit.write_count = 0;
+
         let data_opt_res =
             chainstate.maybe_read_only_clarity_tx(&sortdb.index_conn(), tip, |clarity_tx| {
                 let cost_track = clarity_tx
                     .with_clarity_db_readonly(|clarity_db| {
-                        LimitedCostTracker::new_mid_block(
-                            mainnet,
-                            options.read_only_call_limit.clone(),
-                            clarity_db,
-                        )
+                        LimitedCostTracker::new_mid_block(mainnet, cost_limit, clarity_db)
                     })
                     .map_err(|_| {
                         ClarityRuntimeError::from(InterpreterError::CostContractLoadFailure)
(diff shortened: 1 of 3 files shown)


jcnelson:
Thanks for sending this @psq!  We'll try to get this included in 2.0.12.

src/net/rpc.rs
```diff
@@ -1267,14 +1269,28 @@ impl ConversationHttp {
...
-            Ok(Some(Err(e))) => HttpResponseType::CallReadOnlyFunction(
-                response_metadata,
-                CallReadOnlyResponse {
-                    okay: false,
-                    result: None,
-                    cause: Some(e.to_string()),
-                },
-            ),
+            Ok(Some(Err(e))) => match e {
+                Unchecked(CheckErrors::CostBalanceExceeded(actual_cost, _))
```
pavitthrap, line 1273: I just want to clarify the "path" you're expecting `execute_contract` to follow to get to this branch of the match statement (where we get the `CostBalanceExceeded` error). 

To me, it seems you want `execute_contract` to fail at its call to `runtime_cost`, which will check that the execution costs do not exceed those set by the limits the environment has set. However, the limit is equal to `options.read_only_call_limit` (as set in line 1250 of this file), and the value for this struct (type `ExecutionCost`) ultimately comes from `ConnectionOptions`, which is specified in the `Config` passed to initialize a new node. So, it seems like there is not a guarantee that `write_count` or `write_length` would be 0, meaning the call to `runtime_cost` in `execute_contract` may be okay even if there is a write.

jcnelson: I also had a question about this.  I could actually see this going either way:

* This API endpoint explicitly forbids any writes, and causes a `CostBalanceExceeded` error the moment the code attempts to wrote.
* This API endpoint explicitly permits writes, but never materializes them.  This could be used to allow clients to simulate what would happen if they called a function that performed a write, even though the write never commits.

I can see value in both approaches.

psq: as described in #1641 

> Make it so that the RPC endpoint used for calling read-only functions can be used to call any functions. Make it so that any mutations are discarded

this makes it possible to call any function, and get a result if no write occurred.  Furthermore, a read-only function that makes a `contrac-call?` is flagged as non read only, even if it is calling other read-only functions and the `call-read` endpoint will fail.  So this only allow calling functions that are so basic it is not so useful in practice, and it prevents composing contracts.  You have to use `define-public` instead, but this is not callable via the `call-read` endpoint.

So this extends the `call-read` endpoint to try the function, and return a result if no writes occurred, or a failure if any writes did occur.

So, @jcnelson, this is definitely the former.

@pavitthrap I'm not sure I understand how the node configuration would come into the picture.  Either `write_count` is 0 after the execution and that's a read-only call, or it will fail because it was not read-only.

jcnelson: I think @pavitthrap's concern (correct me if I'm wrong) is that `options.read_only_call_limit` may have a non-zero values, since this `ExecutionCost` instance is instantiated from the node's config file, and you're allowed to specify non-zero write sizes.  I think the intention for even permitting this at all is that one day, there would be an interface that lets you call read/write contract-calls such that the writes never materialize (or even hit disk).

I think the solution here would be to constract an `ExecutionCost` from `options.read_only_call_limit` and explicitly set the `write_length` and `write_count` fields to 0.

psq: oh, @pavitthrap comment makes sense now.  I wasn't expecting these values to be other that 0, but to be safe, I'll set them to 0.

psq: addressed in 57b29c8eece8f3d9ef89bb5fcb2ad51847dfbb7f

testnet/stacks-node/src/tests/integrations.rs
```diff
@@ -32,6 +32,21 @@ use super::{
...
 
+const OTHER_CONTRACT: &'static str = "
+  (define-data-var x uint u0)
+  (define-public (f1)
+    (ok (var-get x)))
+  (define-public (f2 (val uint))
+    (ok (var-set x val)))
+";
+
+const MAIN_CONTRACT: &'static str = "
```
pavitthrap, line 43: Could you change the variables here to be more descriptive of what we are trying to test? (ex: `READ_ONLY_MAIN_CONTRACT`)

testnet/stacks-node/src/tests/integrations.rs
```diff
@@ -174,9 +189,29 @@ fn integration_test_get_info() {
 
             if round == 1 {
                 // block-height = 2
+                eprintln!("Tenure in 1 started!");
```
pavitthrap, line 192: I'm wondering if we should rethink how we are adding tests in `integrations.rs` in general. The test function `integration_test_get_info` is now over 600 lines long, and its original use seems to be to test the `GET_INFO_CONTRACT`. 

Maybe we should break tests for this change into its own function (or maybe not, because it might be time-consuming to run many individual integrations tests). Thoughts @jcnelson ?

src/net/rpc.rs
```diff
@@ -1239,22 +1241,22 @@ impl ConversationHttp {
...
-                        )
+                        LimitedCostTracker::new_mid_block(mainnet, cost_limit, clarity_db)
                     })
                     .map_err(|_| {
                         ClarityRuntimeError::from(InterpreterError::CostContractLoadFailure)
                     })?;
 
                 clarity_tx.with_readonly_clarity_env(mainnet, sender.clone(), cost_track, |env| {
-                    env.execute_contract(&contract_identifier, function.as_str(), &args, true)
+                    env.execute_contract(&contract_identifier, function.as_str(), &args, false)
```
pavitthrap, line 1259: Could you add a comment here explaining why we are setting this to false here (something about how we are using the limits imposed by the cost tracker, which is owned by the environment, to detect any writes).

psq: sure, will do.

testnet/stacks-node/src/tests/integrations.rs
psq: I can do this, but this will probably increase the overall testing time, given that you'd need to re-setup the same thing over and over.

testnet/stacks-node/src/tests/integrations.rs
psq: will do

testnet/stacks-node/src/tests/integrations.rs
psq: after thinking this further, this is something that could be considered separately, but is outside the scope of this PR

src/net/rpc.rs
psq: addressed in 1e2ae7a

testnet/stacks-node/src/tests/integrations.rs
psq: addressed in 1e2ae7a

pavitthrap (approved):

jcnelson (approved):
LGTM! Thanks @psq!

jcnelson:
Just a reminder: please don't merge until 2.0.11 is released.

psq:
thank you both for the review and the back and forth, which made this PR better than what it started as.

philipdesmedt:
thanks for the PR @psq 

would be great if this can land in 2.0.12 as this blocks arkadiko from dynamically versioning contracts 👀

psq:
> thanks for the PR @psq
> 
> would be great if this can land in 2.0.12 as this blocks arkadiko from dynamically versioning contracts 👀

swapr also relies heavily on this, but would run my own node with this PR deployed if necessary to extract data from contract, but let Connect send transactions through the regular API node (but would rather not have to do this, of course).

Not available yet on mainnet, just on my own mocknet.  Happy to share if 2.0.12 gets delayed for any reason.

kantai (commented):
@psq -- these changes all look good to me. Can you resolve the conflicts in `integrations` and make sure that the test passes?

psq:
@kantai sure, will rebase

psq:
@kantai rebased and all green

PR #2548, opened 2021-03-25
Light versus full inventory rescans in pox syncs

kantai:
This is a network inventory sync heuristic change that alters the rescan behavior for PoX inventory syncs. Every 6 rescans with a peer, the inventory sync will do a full rescan. Otherwise, it will only rescan the last 5 PoX cycles.

### DIFF
--- a/src/net/inv.rs
+++ b/src/net/inv.rs
@@ -86,6 +86,11 @@ pub const INV_SYNC_INTERVAL: u64 = 150;
 #[cfg(test)]
 pub const INV_SYNC_INTERVAL: u64 = 0;
 
+/// For light PoX rescans, how many cycles back should we rescan?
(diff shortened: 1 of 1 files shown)


jcnelson:
See https://github.com/blockstack/stacks-blockchain/pull/2549.  I think we worked on the problem at the same time.

jcnelson:
Before working on this further, can we confirm that the inventory state machine with #2549 is still taking too long to do a step? There's two sources of slowness we're seeing on boot-up: the inventory state machine doing a full reward cycle scan, and the downloader state machine doing a full reward cycle rescan.  #2549 only fixes the inventory state machine.  #2550 builds on it and attempts to fix the downloader state machine as well.

kantai:
Yep -- #2549 is the approach we should be taking here.

PR #2549, opened 2021-03-25
Fix/2407

### LINKED ISSUE #2407: Performance: don't ask for another peer's inventory if we already have all the blocks for it
If the peer has all of the blocks for a given reward cycle, then it has no need to ask any other peer for its inventory as long as its PoX bitvector agrees with that peer.  The only time the local node needs to ask a peer for a block inventory vector is if the local node is (a) missing one or more blocks or microblocks in that reward cycle, and (b) does _not_ have the anchor block for that reward cycle.

jcnelson:
Fixes #2407.  The node will only scan the last N reward cycles of a remote peer's inventory, but every so often (e.g. 12 hours), it will scan everyone's.

### DIFF
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -1970,6 +1970,10 @@ impl PoxId {
         }
         (ret, count)
     }
+
+    pub fn num_inventory_reward_cycles(&self) -> usize {
+        self.0.len() - 1
+    }
 }
 
 impl FromStr for PoxId {
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -1972,7 +1972,7 @@ impl PoxId {
     }
 
     pub fn num_inventory_reward_cycles(&self) -> usize {
-        self.0.len() - 1
+        self.0.len().saturating_sub(1)
     }
 }
 
--- a/src/net/download.rs
+++ b/src/net/download.rs
@@ -336,12 +336,12 @@ impl BlockDownloader {
         // preserve download accounting
     }
 
-    pub fn restart_scan(&mut self) -> () {
+    pub fn restart_scan(&mut self, sortition_start: u64) -> () {
         // prepare to restart a full-chain scan for block downloads
-        self.block_sortition_height = 0;
-        self.microblock_sortition_height = 0;
-        self.next_block_sortition_height = 0;
-        self.next_microblock_sortition_height = 0;
+        self.block_sortition_height = sortition_start;
+        self.microblock_sortition_height = sortition_start;
+        self.next_block_sortition_height = sortition_start;
+        self.next_microblock_sortition_height = sortition_start;
         self.empty_block_download_passes = 0;
         self.empty_microblock_download_passes = 0;
     }
@@ -2082,6 +2082,12 @@ impl PeerNetwork {
 
         let now = get_epoch_time_secs();
 
+        let inv_sortition_start = self
+            .inv_state
+            .as_ref()
+            .map(|inv_state| inv_state.block_sortition_start)
+            .unwrap_or(0);
+
         PeerNetwork::with_downloader_state(self, |ref mut network, ref mut downloader| {
             // extract blocks and microblocks downloaded
             for (request_key, block) in downloader.blocks.drain() {
@@ -2209,11 +2215,12 @@ impl PeerNetwork {
                     > network.chain_view.burn_block_height
                 {
                     debug!(
-                        "{:?}: Downloader for blocks has reached the chain tip",
-                        &network.local_peer
+                        "{:?}: Downloader for blocks has reached the chain tip; wrapping around to {}",
+                        &network.local_peer,
+                        inv_sortition_start
                     );
-                    downloader.block_sortition_height = 0;
-                    downloader.next_block_sortition_height = 0;
+                    downloader.block_sortition_height = inv_sortition_start;
+                    downloader.next_block_sortition_height = inv_sortition_start;
 
                     if downloader.num_blocks_downloaded == 0 {
                         downloader.empty_block_download_passes += 1;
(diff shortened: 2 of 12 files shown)


jcnelson:
Okay, this PR does the following:

* Makes it so the node will only synchronize the last few reward cycles in the inventory state machine (configurable under `[connection_options]` as `inv_reward_cycles`, default is 6).  Every 12 hours, it will do a full inventory synchronization (configurable under `[connection_options]` as `full_inv_sync_interval`).
* Makes it so the block downloader will always begin scanning for blocks to download at the block sortition height that the inventory scan last started at.  This means that most of the time, the downloader will by default only scan the last 6 reward cycles (but it will scan all of them on a full inv sync).
* Makes it so the block antientropy protocol will only consider at most `inv_sync_interval` reward cycles at once, but also make it cycle through all reward cycles (just through multiple p2p state machine passes).  In doing so, I discovered and fixed a couple of bugs that was preventing the antientropy protocol from pushing data to nodes that were more than one reward cycle out-of-sync, and was preventing it from sending the same block more than once (even if there were other nodes that needed it).  I added an antientropy stress test as a Neon integration test.

jcnelson:
Still not ready quite yet -- the Neon node run loop isn't correctly inferring whether or not we're in the initial block download when in the first reward cycle.  This, in turn, is causing the anti-entropy protocol to just blast out blocks and microblocks to all its neighbors as they arrive (which is what we want when it's _not_ in IBD mode, but is completely inappropriate while in IBD mode due to its frequency).

src/net/p2p.rs
```diff
@@ -3010,6 +3063,29 @@ impl PeerNetwork {
...
+                            // height we just synchronized
+                            let start_download_sortition =
+                                if let Some(ref inv_state) = self.inv_state {
+                                    debug!(
+                                    "{:?}: Begin downloader synchronization at sortition height {}",
+                                    &self.local_peer, inv_state.block_sortition_start
+                                );
+                                    inv_state.block_sortition_start
+                                } else {
+                                    // really unreachable, but why tempt fate?
```
lgalabru, line 3077: Should we emit an error log in this case?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -1970,6 +1970,10 @@ impl PoxId {
         }
         (ret, count)
     }
+
+    pub fn num_inventory_reward_cycles(&self) -> usize {
+        self.0.len() - 1
```
pavitthrap, line 1975: Use `checked_sub` here maybe?

src/net/download.rs
```diff
@@ -2347,7 +2358,12 @@ impl PeerNetwork {
             self.init_block_downloader();
         }
 
-        let last_inv_update_at = self.inv_state.as_ref().unwrap().last_change_at;
+        let mut last_inv_update_at = 0;
```
pavitthrap, line 2361: style nit: could do `(last_inv_update_at, inv_start_sortition) = match self.inv_state { Some(x) => ...`

jcnelson: Done

src/net/p2p.rs
jcnelson: Done

jcnelson:
> Still not ready quite yet -- the Neon node run loop isn't correctly inferring whether or not we're in the initial block download when in the first reward cycle (it thinks we're not). This, in turn, is causing the anti-entropy protocol to just blast out blocks and microblocks to all its neighbors as they arrive (which is what we want when it's not in IBD mode, but is completely inappropriate while in IBD mode due to its frequency).

Okay, I think both of these are fixed.  I leave a mainnet and testnet follower running for a few days and confirm, but so far it looks good.

src/net/p2p.rs
```diff
@@ -2789,22 +2828,34 @@ impl PeerNetwork {
                                         block_stats,
                                     )?
                                 {
+                                    let index_block_hash = StacksBlockHeader::make_index_block_hash(
+                                        &parent_consensus_hash,
+                                        &parent_block_hash
+                                    );
+
+                                    if parent_consensus_hash == highest_snapshot.consensus_hash {
+                                        // This parent block was just sortition'ed
```
kantai, line 2837: What happens if the antientropy protocol pushes blocks that have just been chosen?

jcnelson: It shouldn't, and I've confirmed this when running the node on mainnet.  This condition should make it so the anti-entropy protocol only pushes blocks and microblocks after at least one subsequent sortition has happened (since the block or confirmed microblock stream to be pushed must have a different consensus hash than the burnchain's canonical tip).

kantai: Right -- my question is, why not? What's the downside of pushing the just-chosen block?

jcnelson: It's a DDoS vector.  If we did this, then all nodes you're connected to will all push the block to you at about the same time.  We avoid this today by making it so that all your connected neighbors will only push you a (much smaller) `BlocksAvailable` message when they have obtained a block.  This, in turn, causes you to go and ask the neighbors for the block data if you do not yet have it already.

jcnelson: Hilariously, I've witnessed this DDoS-y behavior already.  Remember the network outage from a few weeks ago when we lost a couple blocks for a few hours?  Turns out, a new nodes will have obtained, but not processed, these blocks because the invalid microblock stream that they confirm has been lost forever.  So new nodes don't know that old nodes already have this block, and have simply orphaned it because it confirmed a too-big stream.  As such, a new node will periodically send the invalid block to old nodes that did process the (now-lost) microblock stream over and over because the old nodes report it missing while the new nodes report it as present but unprocessed.

This behavior doesn't pose a systemic risk simply because new nodes booting from this branch will need a new chainstate anyway, so they'll all obtain this unprocessable block and will all believe that their peers already have it.

kantai: Ah, okay, that makes sense

kantai (approved):

lgalabru (approved):
LGTM, Thank you @jcnelson !

PR #2550, opened 2021-03-26
Fix/2407 net+downloader

jcnelson:
This is a follow-up PR against #2549. It modifies the downloader to only scan for new blocks over the range of newly-sync'ed inventories (so it only looks at the list _N_ reward cycles of blocks).  It will do a full download re-scan only when it happens for the inventories.


jcnelson:
Skip reward cycles with no sortitions (affects the sync watchdog).

jcnelson:
The functionality for this has been folded into #2549.

PR #2562, opened 2021-04-02
src/vm/docs/mod: correct example

ehildenb:
Fixes: https://github.com/blockstack/stacks-blockchain/issues/2560


kantai:
This (and the class of related issues with the docs examples) are addressed in https://github.com/blockstack/stacks-blockchain/pull/2568, so I'm closing this PR. Thanks for raising this issue @ehildenb !

Issue #2560, opened 2021-04-01
Example for `match` in documentation is ill-typed


ehildenb:
The example provided for the `(match ...)` construct: https://docs.blockstack.org/references/language-functions#match
contains an ill-typed program:

```
(define-private (add-10 (x (optional int)))
  (match x
  value (+ 10 value)
  10))
(add-10 (some 5)) ;; Returns 15
(add-10 none) ;; Returns 10

(define-private (add-or-pass-err (x (response int (string-ascii 10))) (to-add int))
  (match x
   value (+ to-add value)
   err-value (err err-value)))
(add-or-pass-err (ok 5) 20) ;; Returns 25
(add-or-pass-err (err "ERROR") 20) ;; Returns (err "ERROR")
```

Specifically, the `(match ...)` in `(add-or-pass-err ...)` should return the _same_ type on both branches, but on the first branch it returns an `int`, and on the second branch it returns a `(response bot (string-ascii 5))`. Another problem is the use of the `(string-ascii ...)` type, which I think may be outdated?

This example can be correctly simply by applying this diff:

```
@@ -5,9 +5,9 @@
 (add-10 (some 5)) ;; returns 15
 (add-10 none) ;; returns 10
 
-(define-private (add-or-pass-err (x (response int (string-ascii 10))) (to-add int))
+(define-private (add-or-pass-err (x (response int (buff 10))) (to-add int))
   (match x
-   value (+ to-add value)
+   value (ok (+ to-add value))
    err-value (err err-value)))
 (add-or-pass-err (ok 5) 20) ;; returns 25
 (add-or-pass-err (err "ERROR") 20) ;; returns (err "ERROR")
```

This changes:

-   The input type of `x` from `(response int (string-ascii 10))` to `(response int (buff 10))`, and
-   The first branch of `(match ...)` now returns an `(ok (+ to-add value))` instead of just a `(+ to-add value)`.

This makes this example type-check correctly, and the resulting type of `(add-or-pass-err ...)` is now something like `(response int (buff 10)) -> int -> (response int (buff 10))`.

lgalabru:
Hi @ehildenb! Thank you for reporting this, we need indeed to fix the documentation, as 

```
(define-private (add-or-pass-err (x (response int (string-ascii 10))) (to-add int))
  (match x
   value (+ to-add value)
   err-value (err err-value)))
```
[...]

PR #2568, opened 2021-04-05
Fix: Typecheck in `vm::docs` and correct examples

kantai:
This PR adds a typechecker pass to the unit-tests in `vm::docs` and corrects the mis-typed examples in the docs (addressing #2560).

This PR also includes a nice extension to the `decode-tx` function in `main`, verifying the inputted transaction and outputting the origin address as a C32 address.


jcnelson (approved):
Thanks @kantai !

reedrosenbluth (approved):
LGTM!

PR #2570, opened 2021-04-05
Boot looping

lgalabru:
I'm opening this PR for addressing the issue raised here https://github.com/blockstack/stacks-blockchain/issues/2546.
Depending on when you terminate the node, the boot can be stuck in a loop on the subsequent relaunch.
With this patch, the node is able to recover, however I'm having a hard time measuring side effects, and would appreciated more opinions on this approach.

### DIFF
--- a/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
+++ b/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
@@ -1305,6 +1305,8 @@ impl BitcoinRegtestController {
         &self,
         height_to_wait: Option<u64>,
     ) -> Result<BurnchainTip, BurnchainControllerError> {
+        let mut timeout = 0;
+        let step = 100;
         loop {
             let canonical_burnchain_tip = self
                 .burnchain_db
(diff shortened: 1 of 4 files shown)


testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -1314,12 +1316,18 @@ impl BitcoinRegtestController {
                 .unwrap();
             let canonical_sortition_tip =
                 SortitionDB::get_canonical_burn_chain_tip(self.sortdb_ref().conn()).unwrap();
-            if canonical_burnchain_tip.block_height == canonical_sortition_tip.block_height {
+
+            if canonical_burnchain_tip.block_height == canonical_sortition_tip.block_height
```
jcnelson, line 1320: What are these values on each loop pass?  Like, why does the timeout break out of this loop? Would love some more insight on why it is never the case on restart that `canonical_burnchain_tip.block_height == canonical_sortition_tip.block_height`.

kantai: This is the scenario that I _think_ this is catching, but should perhaps be caught differently:

1. A burn block is processed by the burnchain db
2. Coordinator is notified
3. Before coordinator processes the sortition, the node exits gracefully

On restart:
1. The neon runloop syncs the burnchain, starts the coordinator (but _does not_ tell it to try to process burn blocks or stacks blocks)
2. The neon runloop invokes `wait_for_sortitions`, which will wait for the sortition db to catch up to the burnchain db. However, that won't happen until the coordinator is told to process a new burn block.

I _think_ this stall would be pushed forward by the arrival of a new burn block. An alternative solution would be for the neon runloop to just tell the chains coordinator to wake up and try to process a burn block. I think that approach would be better: the issue isn't that there's a stall, the issue is that the sortition db hasn't caught up with the burnchain db.

lgalabru: @kantai yes you're right, we are looking at the use case that you've described. And when the stacks-node is being interrupted when the burnchain block pipeline is being fed, chances that you end up in this situations are high.

kantai: Okay -- in that case, I'd suggest addressing the issue by issuing a `new_burn_block` notification to the coordinator rather than adding a timeout here.

lgalabru: I can confirm that this new approach is working. Thank you for that tip @kantai!

kantai (approved):
LGTM, just make sure the tests now pass!

jcnelson (approved):
LGTM!  Can't wait to test it in rc2!

Issue #2553, opened 2021-03-29
Event broadcasts for pending microblocks


pavitthrap:
Make a new event type to transmit streamed microblocks.

PR #2571, opened 2021-04-05
Feat: Added event for new microblocks

pavitthrap:
## Description

Added events for unconfirmed microblocks for issue #2553. The event data includes the `index_block_hash` of the preceding block, filtered events (only includes the events from the receipts that each event observer is interested in), and transaction receipts.

Each payload may contain transactions from multiple microblocks. Each tx receipt comes with a `tx_index`, which starts from 0 for each microblock. Each tx receipt also includes the sequence number of the microblock it belongs to.

### Notes 
- One foreseeable "issue" is that different stacks nodes might have different microblock streams (in the case the leader is creating a microblock fork). If someone is querying the API through a load balancer endpoint, the load balancer could conceivably direct a user to different stacks nodes on each page refresh, which could lead to different data on the page. This might not be an important case to consider if most leaders are acting honestly, but something to consider. 
- I could add more in-depth testing

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

## Testing information
Ran `microblock_integration_test` in `neon_integrations.rs`. Ran `cargo test`.


testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -111,6 +113,20 @@ impl EventObserver {
...
+        let mut tx_index = 0;
+        let mut serialized_txs = Vec::new();
+        for receipt in receipts.iter() {
+            let payload = EventObserver::make_new_block_txs_payload(receipt, tx_index);
+            serialized_txs.push(payload);
+            tx_index += 1;
+        }
+
+        json!({
+            "transactions": serialized_txs,
```
zone117x, line 126: We'll also need the transaction events -- I think the type is `StacksTransactionEvent`.

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -111,6 +113,20 @@ impl EventObserver {
         }
     }
 
+    fn make_new_microblocks_payload(receipts: Vec<StacksTransactionReceipt>) -> serde_json::Value {
+        let mut tx_index = 0;
+        let mut serialized_txs = Vec::new();
+        for receipt in receipts.iter() {
+            let payload = EventObserver::make_new_block_txs_payload(receipt, tx_index);
+            serialized_txs.push(payload);
+            tx_index += 1;
```
zone117x, line 122: Is this ordering deterministic? As in would all running stacks-nodes report microblock transactions with the same `tx_index`?

jcnelson: Yes -- the transaction receipts from the unconfirmed microblock processing are generated in the order that they will also be (potentially) confirmed.  However, because the microblocks aren't confirmed yet, it's possible that you'll get receipts for transactions that will ultimately get dropped (e.g. because the next anchored block miner doesn't have to confirm them).

testnet/stacks-node/src/event_dispatcher.rs
zone117x, line 126: Is there `StacksMicroblockHeader` info that we should also send, in case the API needs to expose it to the explorer?

zone117x, line 126: I think we'd also need to include the `index_block_hash` in order for the API to handle re-orgs, and I think the microblock header to handle micro re-orgs?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1295,6 +1393,41 @@ fn microblock_integration_test() {
             .unwrap()
     );
 
+    // wait at least two p2p refreshes so it can produce the microblock
+    for i in 0..30 {
```
jcnelson, line 1397: You should be able to use the `PoxWatchdog` to poll for this, so you don't have to worry about races.

src/chainstate/stacks/db/unconfirmed.rs
```diff
@@ -217,6 +241,7 @@ impl UnconfirmedState {
                     }
                     total as u64
                 };
+                self.bytes_so_far += new_bytes;
```
jcnelson, line 244: Can you remove this, and put it back where it was?  I'm intentionally trying to _avoid_ mutating `self` here.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1117,6 +1139,32 @@ fn bitcoind_forking_test() {
...
+    parent_block: BlockHeaderHash,
+    seq: u16,
+) -> StacksMicroblock {
+    let txid_vecs = txs.iter().map(|tx| tx.txid().as_bytes().to_vec()).collect();
+    let merkle_tree = MerkleTree::<Sha512Trunc256Sum>::new(&txid_vecs);
+    let tx_merkle_root = merkle_tree.root();
+
+    let mut mblock = StacksMicroblock {
+        header: StacksMicroblockHeader {
+            version: 0x12,
```
jcnelson, line 1156: Just for fun, can you make `.version` a random value each time this test runs? I just want to make sure that the version field can be arbitrary -- i.e. it's not consensus-critical.

src/chainstate/stacks/db/unconfirmed.rs
pavitthrap: Oh I see. The reason I changed it is because `bytes_so_far` was only adding the number of new bytes from the last microblock. 

I could keep track of a total new bytes and add that to `bytes_so_far` below.

testnet/stacks-node/src/tests/neon_integrations.rs
jcnelson: On second thought, plumbing this through would probably be more effort than it's worth.

src/chainstate/stacks/db/unconfirmed.rs
jcnelson: Yes, I believe there's a bug there as well (that should be fixed in 2.0.11.0.0-rc4).  The `new_bytes` variable should represent the total number of bytes added to the unconfirmed state.

testnet/stacks-node/src/tests/neon_integrations.rs
pavitthrap: Discussed in discord that this change isn't strictly necessary.

src/chainstate/stacks/db/unconfirmed.rs
```diff
@@ -43,6 +43,26 @@ use vm::costs::ExecutionCost;
 
 pub type UnconfirmedTxMap = HashMap<Txid, (StacksTransaction, BlockHeaderHash, u16)>;
 
+pub struct ProcessedUnconfirmedState {
+    pub total_burns: u128,
+    pub total_fees: u128,
+    pub receipts: Vec<StacksTransactionReceipt>,
+    // tuple representing (index, sequence #), where the index corresponds to the starting index of
+    // receipts corresponding to that sequence number in the receipts vector
+    pub sequence_indices_for_receipts: Vec<(usize, u16)>,
```
jcnelson, line 52: This seems like a very round-about way of just turning `receipts` into `Vec<Vec<StacksTransactionReceipt>>`.  Want to do that instead, and in doing so, reduce the complexity of `process_new_microblocks()` in `event_observer.rs`?

testnet/stacks-node/src/neon_node.rs
```diff
@@ -811,6 +811,19 @@ fn spawn_miner_relayer(
                         event_dispatcher.process_new_mempool_txs(net_receipts.mempool_txs_added);
                     }
 
+                    let num_unconfirmed_microblock_tx_receipts = net_receipts.processed_unconfirmed_state.receipts.len();
+                    if num_unconfirmed_microblock_tx_receipts > 0 {
+                        let (canonical_consensus_hash, canonical_block_hash) =
+                            SortitionDB::get_canonical_stacks_chain_tip_hash(
```
jcnelson, line 817: I think you can just use this here:

```rust
let canonical_tip = chainstate
   .as_ref()
   .map(|unconfirmed_state| unconfirmed_state.confirmed_chain_tip.clone())
   .unwrap_or(StacksBlockId([0u8; 32]));

if canonical_tip != StacksBlockId([0u8; 32]) {
   event_dispatcher.process_new_microblocks(canonical_tip, net_receipts.processed_unconfirmed_state);
}
else {
   // this really should be unreachable
   warn!("OOPS: unconfirmed state is uninitialized but there are microblock events");
}
```

The `unwrap_or()` case should never be reached if `new_unconfirmed_microblock_tx_receipts.len() > 0`, since the unconfirmed state will always be initialized and the `confirmed_chain_tip` will correspond to the Stacks chain tip on top of which the unconfirmed microblocks are built.

jcnelson (approved):
LGTM! Just one minor comment to address.

testnet/stacks-node/src/neon_node.rs
pavitthrap: My question about this change is that it seems like `process_new_microblocks` will be called regardless of the value of `net_receipts.processed_unconfirmed_state.receipts.len()`. I can make this change, but I guess I should add a check in `process_new_microblocks` to ensure the length of the receipts is non-zero before executing any logic?

jcnelson: Oh, I was only talking about removing the call to `SortitionDB::get_canonical_stacks_chain_tip_hash()`.  The above code would run only if `net_receipts.processed_unconfirmed_state.receipts.len() > 0`.

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -199,6 +202,17 @@ impl EventObserver {
                 None => json!(null),
             }
         };
+        (txid, success, raw_result, raw_tx, contract_interface_json)
```
kantai, line 205: This is a pretty large tuple response (with several of the entries of the same type) -- I think it'd make sense here to add a new non-public struct like `ReceiptPayloadInfo` so that there isn't any confusion in the future if we need to add another entry here.

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -501,14 +554,47 @@ impl EventDispatcher {
                         );
                     }
                 }
-                events.push((!receipt.post_condition_aborted, tx_hash, event));
+                events.push((!receipt.post_condition_aborted, tx_hash, event.clone()));
```
kantai, line 557: Is it possible to avoid this clone? These events could get large memory-wise, and this clone would end up magnifying the memory consumption of the event dispatcher.

kantai (approved):
LGTM! Just needs to do some merge conflict resolution.

Issue #2572, opened 2021-04-06
New anti-entropy integration test times out


jcnelson:
The antientropy integration test in `develop` is timing out.  This should be investigated -- it shouldn't take that long.

jcnelson:
Working now!

Issue #2427, opened 2021-02-08
Atlas: get_attachments_missing_at_page_index() can create an unbound-length SQL query


jcnelson:
The method `AtlasDB::get_attachments_missing_at_page_index()` takes a `Vec<StacksBlockId>` as an argument, and feeds it into a SQL statement.  The trouble is, the length of this vector is network-controlled -- the requester may be able to request a particularly long block range be queried, leading to a potential DoS (crash) or a degradation of service (since this RPC handler blocks on querying all of the block IDs, and there doesn't seem to be any upper bound on how many block IDs may be queried).

PR #2574, opened 2021-04-06
Fix/2.0.11 rc2 integration tests

### LINKED ISSUE #2077: Atlas: stream attachment inventories and attachments
For context https://github.com/blockstack/stacks-blockchain/pull/2013#discussion_r525528001
Screen Shot 2020-11-18 at 5 11 40 PM

### LINKED ISSUE #2569: Speed up `runtime_overflow_unconfirmed_microblocks_integration_test` (currently > 40 minutes)
We should figure out if we can speed up the `runtime_overflow_unconfirmed_microblocks_integration_test` -- it's currently taking over 40 minutes, and it runs as part of our integration tests in Github actions.

jcnelson:
This fixes #2569 and #2572, and adds a new Atlas stress-test that looks for performance regressions that, if passing, will allow us to close #2077 and #2427 (since we will have confirmed that the maximum runtime of these methods is acceptable).

### DIFF
--- a/.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
+++ b/.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
@@ -27,3 +27,4 @@ RUN cargo test -- --test-threads 1 --ignored tests::should_succeed_handling_malf
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::size_overflow_unconfirmed_microblocks_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::runtime_overflow_unconfirmed_microblocks_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::antientropy_integration_test
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::atlas_stress_integration_test
--- a/.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
+++ b/.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
@@ -27,4 +27,3 @@ RUN cargo test -- --test-threads 1 --ignored tests::should_succeed_handling_malf
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::size_overflow_unconfirmed_microblocks_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::runtime_overflow_unconfirmed_microblocks_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::antientropy_integration_test
-RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::atlas_stress_integration_test
--- a/testnet/stacks-node/src/tests/neon_integrations.rs
+++ b/testnet/stacks-node/src/tests/neon_integrations.rs
@@ -3998,6 +3998,7 @@ fn atlas_stress_integration_test() {
         conf_bootstrap_node.clone(),
         None,
         Some(burnchain_config.clone()),
+        None,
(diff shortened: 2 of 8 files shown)


testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -3886,3 +3892,778 @@ fn antientropy_integration_test() {
...
+            "Failed to mine all transactions: nonces = {:?}, expected {:?} + {}",
+            &account_after_nonces, account_before_nonces, batch_size
+        );
+        panic!();
+    }
+}
+
+#[test]
+#[ignore]
+fn atlas_stress_integration_test() {
```
kantai, line 3949: This test doesn't appear to work correctly in the Github action.

lgalabru: I think there's an issue line 4630, when retrieving the expected content hash - it looks like the tuple for retrieving the content hash would be invalid.
Do you want me to take over?

.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
```diff
@@ -27,3 +27,4 @@ RUN cargo test -- --test-threads 1 --ignored tests::should_succeed_handling_malf
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::size_overflow_unconfirmed_microblocks_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::runtime_overflow_unconfirmed_microblocks_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::antientropy_integration_test
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::atlas_stress_integration_test
```
lgalabru, line 30: Can we move this entry to `Dockerfile.atlas-test` instead?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -3886,3 +3892,778 @@ fn antientropy_integration_test() {
...
+            );
+        }
+
+        for i in 0..l {
+            if attachments[i] == 0 {
+                continue;
+            }
+            let page_index = attachments[i] % 64; // AttachmentInstance::ATTACHMENTS_INV_PAGE_SIZE;
+            let content_hash = attachment_hashes
+                .get(&(*ibh, page_index))
```
lgalabru, line 4630: Looking at the insertions, I think it should be:
```suggestion
                .get(&(*ibh, attachments[i]))
```
Will test this patch locally.

jcnelson: Ah, I was wondering why this wasn't working reliably.  I'll try it too -- I want to understand the system better.

lgalabru: Confirming that this test is green on my machine, with this patch.

.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
jcnelson: Done

testnet/stacks-node/src/tests/neon_integrations.rs
jcnelson: Fixed here as well; let's see if the CI takes it.

testnet/stacks-node/src/tests/neon_integrations.rs
jcnelson: I think I've fixed it, but let's see if the CI takes it.

lgalabru (approved):
LGTM, thank you @jcnelson!

kantai (approved):
LGTM!

PR #2576, opened 2021-04-07
don't wait forever while waiting for an inv-sync pass

jcnelson:
Fixes https://discord.com/channels/621759717756370964/730814571517968436/829186890388340816

### DIFF
--- a/testnet/stacks-node/src/syncctl.rs
+++ b/testnet/stacks-node/src/syncctl.rs
@@ -62,28 +62,17 @@ impl PoxSyncWatchdogComms {
         self.last_ibd.load(Ordering::SeqCst)
     }
 
-    /// Wait for at least one p2p state-machine passes
-    pub fn wait_for_p2p_state_pass(&self, timeout: u64) -> bool {
-        let current = self.get_p2p_state_passes();
-
-        let now = get_epoch_time_secs();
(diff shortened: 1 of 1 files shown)


testnet/stacks-node/src/syncctl.rs
```diff
@@ -62,28 +62,17 @@ impl PoxSyncWatchdogComms {
...
-        }
-        return true;
-    }
-
     /// Wait for at least one inv-sync state-machine passes
     pub fn wait_for_inv_sync_pass(&self, timeout: u64) -> Result<bool, burnchain_error> {
         let current = self.get_inv_sync_passes();
 
+        let now = get_epoch_time_secs();
         while current >= self.get_inv_sync_passes() {
```
jcnelson, line 70: TL;DR We have to keep polling `self.get_inv_sync_passes()` or this loop will never exit.

lgalabru (approved):
LGTM, thank you @jcnelson!

kantai (approved):
LGTM!

diwakergupta:
Thanks @jcnelson ! Is it possible to add an integration or E2E test that would catch a regression going forward? The fact we had built and deployed an RC2 without catching this type of "dead on arrival" behavior isn't great.

PR #2581, opened 2021-04-08
Fix/updated bitcoind

### LINKED ISSUE #2559: Bitcoind: Need to explicitly create wallet for tests
As of bitcoin 0.21.0, a default wallet is no longer automatically created (https://github.com/bitcoin/bitcoin/pull/15454). 

This change causes the tests in neon_integrations.rs to fail with the following message:
```
ERRO [1617206434.490494] [testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs:589] [ThreadId(13)] Bitcoin RPC failure: error listing utxos Network("Bitcoin RPC: status(500) != success, body is \'Object({\"error\": Object({\"code\": Number(-18), \"message\": String(\"No wallet is loaded. Load a wallet using loadwallet or create a new one with createwallet. (Note: A default wallet is no longer automatically created)\")}), \"id\": String(\"stacks\"), \"result\": Null})\'")
```

pavitthrap:
## Description
Makes tests in `neon_integrations.rs` compatible with both old/newest version of bitcoind. This change was needed because the newest version of `bitcoind` does not automatically create a default wallet anymore. 
Fixes #2559.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

## Testing information
Ran tests in `neon_integrations.rs` using versions 0.20.1 & 0.21.0 of bitcoind.

### DIFF
--- a/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
+++ b/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
@@ -567,6 +567,11 @@ impl BitcoinRegtestController {
         result_vec
     }
 
+    pub fn create_wallet(&self, wallet_name: &str) {
+        BitcoinRPCRequest::create_wallet(&self.config, wallet_name)
+            .expect("Wallet creation failed");
+    }
+
     pub fn get_utxos(
         &self,
         public_key: &Secp256k1PublicKey,
@@ -1872,6 +1877,18 @@ impl BitcoinRPCRequest {
         let payload = BitcoinRPCRequest {
             method: "importaddress".to_string(),
             params: vec![address.to_b58().into(), label.into(), rescan.into()],
+            id: "stacks".to_string(),
+            jsonrpc: "2.0".to_string(),
+        };
+
+        BitcoinRPCRequest::send(&config, payload)?;
+        Ok(())
+    }
+
+    pub fn create_wallet(config: &Config, wallet_name: &str) -> RPCResult<()> {
+        let payload = BitcoinRPCRequest {
+            method: "createwallet".to_string(),
+            params: vec![wallet_name.into()],
             id: "stacks".to_string(),
             jsonrpc: "2.0".to_string(),
         };
--- a/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
+++ b/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
@@ -570,12 +570,25 @@ impl BitcoinRegtestController {
     /// Checks if there is a default wallet with the name of "".
     /// If the default wallet does not exist, this function creates a wallet with name "".
     pub fn create_wallet_if_dne(&self) {
-        let wallets =
-            BitcoinRPCRequest::list_wallets(&self.config).expect("Call to listwallets failed");
+        let wallet_res = BitcoinRPCRequest::list_wallets(&self.config);
+        let wallets = match wallet_res {
+            Ok(wallets) => wallets,
+            Err(e) => {
+                warn!("Call to listwallets failed during wallet creation: {:?}", e);
+                return;
+            }
+        };
 
         if !wallets.contains(&("".to_string())) {
-            BitcoinRPCRequest::create_wallet(&self.config, "")
-                .expect("Call to createwallet failed");
+            match BitcoinRPCRequest::create_wallet(&self.config, "") {
+                Err(e) => {
+                    warn!(
+                        "Call to createwallet failed during wallet creation: {:?}",
+                        e
+                    );
+                }
+                _ => {}
+            }
         }
     }
 
@@ -1919,7 +1932,7 @@ impl BitcoinRPCRequest {
                     }
                 }
                 _ => {
-                    warn!("Failed to get wallet");
+                    warn!("Failed to get wallets");
                 }
             },
             _ => {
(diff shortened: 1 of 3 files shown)


testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -121,6 +121,8 @@ impl RunLoop {
             Some(should_keep_running.clone()),
         );
         let pox_constants = burnchain.get_pox_constants();
+        // Initialize the wallet.
+        burnchain.create_wallet_if_dne();
```
jcnelson, line 125: This should only be invoked if the Stacks node is running as a miner.  If you try this on a public Bitcoin node, it will fail.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -567,6 +567,18 @@ impl BitcoinRegtestController {
         result_vec
     }
 
+    /// Checks if there is a default wallet with the name of "".
+    /// If the default wallet does not exist, this function creates a wallet with name "".
+    pub fn create_wallet_if_dne(&self) {
+        let wallets =
+            BitcoinRPCRequest::list_wallets(&self.config).expect("Call to listwallets failed");
```
jcnelson, line 574: This code could panic if the bitcoin node was slow, or if your device was having connectivity issues.  Could you instead have this method return an error if the RPC fails, so the caller can report an appropriate error message?

jcnelson: (Same goes for the call to `create_wallet`.  Generally speaking, networking code that could fail should not panic).

pavitthrap: Ok, now handling the error with a warn statement.

pavitthrap: Added a call to `warn!` instead of panicking now.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -567,6 +567,31 @@ impl BitcoinRegtestController {
...
 
+    /// Checks if there is a default wallet with the name of "".
+    /// If the default wallet does not exist, this function creates a wallet with name "".
+    pub fn create_wallet_if_dne(&self) {
+        let wallet_res = BitcoinRPCRequest::list_wallets(&self.config);
+        let wallets = match wallet_res {
+            Ok(wallets) => wallets,
+            Err(e) => {
+                warn!("Call to listwallets failed during wallet creation: {:?}", e);
+                return;
```
jcnelson, line 578: I think this should return a `Result<..>`, and the caller -- i.e. the run loop -- should handle the `Err(..)` case.  This is because the way to handle the error case could be to try e.g. retrying it a few times before giving up, or erroring out quickly, or printing helpful diagnostics before exiting, and so on.

jcnelson (approved):
This LGTM! Just had one last minor comment.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -567,6 +567,31 @@ impl BitcoinRegtestController {
         result_vec
     }
 
+    /// Checks if there is a default wallet with the name of "".
+    /// If the default wallet does not exist, this function creates a wallet with name "".
+    pub fn create_wallet_if_dne(&self) {
```
kantai, line 572: Because these new methods are only called via tests, can you mark them all `cfg(test)`?

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -123,6 +123,10 @@ impl RunLoop {
         let pox_constants = burnchain.get_pox_constants();
 
         let is_miner = if self.config.node.miner {
+            // Initialize the wallet.
+            info!("Miner node: creating the wallet if it does not exist");
+            burnchain.create_wallet_if_dne();
```
kantai, line 128: Because this is a pretty expensive operation in Bitcoin mainnet and testnet, I think we should avoid trying to do this automatically for users. Can this call be moved to the `neon_integrations` setup code? That way the new functions in the regtest controller could also be marked `cfg(test)`

pavitthrap: I don't know what setup code you're referring to in `neon_integrations`? For now I replicated the way `bump_blocks_processed` handles test/non-test code in `neon.rs`.

kantai: All of the `neon_integrations` tests do some form of the following:

```rust
    let mut btcd_controller = BitcoinCoreController::new(conf.clone());
    btcd_controller
        .start_bitcoind()
        .map_err(|_e| ())
        .expect("Failed starting bitcoind");

    let mut btc_regtest_controller = BitcoinRegtestController::new(conf.clone(), None);
    let http_origin = format!("http://{}", &conf.node.rpc_bind);

    btc_regtest_controller.bootstrap_chain(201);
```

Maybe we could call `create_wallet_if_dne` from the `bootstrap_chain` implementation of the `bitcoin_regtest_controller`? That way it could also be non-public.

kantai (approved):
LGTM!

PR #2582, opened 2021-04-08
refactor: move StacksMessageCodec into new codec module

reedrosenbluth:
This PR adds a new `codec` module for the `StacksMessageCodec` trait, and related types and functions.

This means that we can remove `net` as a dependency from `vm` and `util`, instead depending on the more minimal `codec`.

TODO:

- [ ] move `StacksMessageCodec` impls out of `net::codec.rs`


src/vm/types/serialization.rs
```diff
@@ -14,29 +14,31 @@
...
+use std::{error, fmt, str};
+use std::borrow::Borrow;
+use std::collections::HashMap;
+use std::convert::{TryFrom, TryInto};
+use std::io::{Read, Write};
+
+use serde_json::Value as JSONValue;
+
+use codec::Error as codec_error;
+use net::Error as NetError;
```
lgalabru, line 26: can we get rid of this import?

src/chainstate/stacks/address.rs
```diff
@@ -278,15 +268,15 @@ impl Address for StacksAddress {
...
 #[cfg(test)]
 mod test {
-    use super::*;
-
     use chainstate::stacks::*;
-    use net::codec::test::check_codec_and_corruption;
-    use net::codec::*;
     use net::*;
+    use net::codec::*;
+    use net::codec::test::check_codec_and_corruption;
```
lgalabru, line 274: `use codec::`?

src/codec/mod.rs
```diff
@@ -0,0 +1,160 @@
+use std::{io, mem, fmt, error};
+use std::io::{Read, Write};
+
+use net::{MAX_MESSAGE_LEN};
```
lgalabru, line 4: Could MAX_MESSAGE_LEN be defined in this module, instead of defining it in `net`?

src/codec/mod.rs
```diff
@@ -0,0 +1,188 @@
+use std::io::{Read, Write};
```
jcnelson, line 1: This code is lifted verbatim from `src/net/codec.rs`, right?

src/net/mod.rs
```diff
@@ -1554,17 +1517,19 @@ pub const BLOCKS_PUSHED_MAX: u32 = 32;
 
 macro_rules! impl_byte_array_message_codec {
     ($thing:ident, $len:expr) => {
-        impl ::net::StacksMessageCodec for $thing {
+        impl ::codec::StacksMessageCodec for $thing {
```
jcnelson, line 1520: Should this macro go into `src/codec/macros.rs`?

jcnelson (approved):
This overall looks good to me! :+1: Just had a couple of clarifying questions.  Thanks @reedrosenbluth!

src/codec/mod.rs
reedrosenbluth: yes and `src/net/mod.rs`

src/chainstate/stacks/address.rs
```diff
@@ -30,32 +30,31 @@ use burnchains::bitcoin::address::{
 };
 use burnchains::{Address, BurnchainSigner, PublicKey};
 use chainstate::stacks::StacksPublicKey;
+use chainstate::stacks::{
+    C32_ADDRESS_VERSION_MAINNET_MULTISIG, C32_ADDRESS_VERSION_MAINNET_SINGLESIG,
+    C32_ADDRESS_VERSION_TESTNET_MULTISIG, C32_ADDRESS_VERSION_TESTNET_SINGLESIG,
```
lgalabru, line 35: Should we move this constants to the address module?

lgalabru (approved):
Looks good to me, thank you for making some progress on that front @reedrosenbluth!

src/chainstate/stacks/address.rs
reedrosenbluth: I think it's ok for these to remain in `chainstate` so long as `StacksAddress` is there.

lgalabru: Should StacksAddress be in types?

reedrosenbluth: Yes, `StacksAddress` should definitely be in `types`. These constants should probably live there too. Will make those changes in a future PR

PR #2587, opened 2021-04-13
Feat/clarity cli execute json

hstove:
## Description

This PR implements some functionality that I've been using to develop a tool for running running Clarity tests in a JS environment. It's built on top of `clarity-js-sdk`, but it uses the raw Clarity hex result of different contract interactions. This allows for more robust handling of contracts, without having to write a parser for Clarity output.

It adds two new commands to the `clarity-cli` CLI:

- `eval_at_chaintip_json`: just like `eval_at_chaintip`, but it returns a JSON-serialized output to stdout, with a field for the raw Clarity hex result
- `execute_json`: like `execute`, but it returns JSON to stdout, along with the raw Clarity hex result

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

### DIFF
--- a/src/clarity.rs
+++ b/src/clarity.rs
@@ -25,6 +25,7 @@ use std::path::PathBuf;
 use std::process;
 
 use util::log;
+use util::hash::hex_bytes;
 
 use chainstate::burn::BlockHeaderHash;
 use chainstate::stacks::index::{storage::TrieFileStorage, MarfTrieId};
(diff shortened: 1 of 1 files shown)


jcnelson:
While I agree that `clarity_cli` would be improved by printing out a machine-readable result, I don't think this is best achieved by adding new commands.  Instead, I think the `invoke_command` function should simply return the data to be displayed -- either a `serde::Value` or a `Vec<String>` -- based on the presence or absence of a `--json` flag.  I'm already working on this.

jcnelson:
Going to go ahead and close this -- #2597 supersedes it.

PR #2589, opened 2021-04-13
Refactor: initial vm decoupling

reedrosenbluth:
issue: #2398

## Goal
The aim of this PR is to take first steps towards decoupling the `vm` modules from the rest of the codebase, specifically `chainstate`. I'm approaching this by splitting `vm` into two modules. One will closely resemble the existing `vm` module, and will at some point be its own published crate available to be utilized by other projects (such as the Clarity repl). The other module (`clarity_vm`) will contain Stacks blockchain specific interfaces and functionality for interacting with the Clarity `vm` (such as `ClarityInstance` and `ClarityBlockConnection`)

## Overview
Currently `vm` is pretty tightly coupled with the rest of our codebase. The following dependency graphs show rough approximations of the current state of affairs, what things will look like after this initial refactor, and the eventual refactoring goal. They don't show dependencies on `vm` or `clarity_vm` (except for the temporary circular dependency between `vm` and `clarity_vm` in the second graph), but only show the relevant dependencies that these two modules have on other `stacks-blockchain` modules.

```
Current dependency graph (ignoring most dependencies between the upper level):

                                        -----------------------------------------------
                                       |                                               v 
+-------------+ +-------------+ +-------------+ +-------------+ +-------------+ +-------------+
|  rusqlite   | | chainstate  | |    util     | | burnchains  | |    core     | |     net     |
+-------------+ +-------------+ +-------------+ +-------------+ +-------------+ +-------------+
       ^               ^               ^               ^               ^              ^
       |               |               |               |               |              |
+-------------+        |               |               |               |              |
|     vm      |------------------------------------------------------------------------
+-------------+

Initial refactor (after this PR):
                                        -----------------------------------------------
                                       |                                               v 
+-------------+ +-------------+ +-------------+ +-------------+ +-------------+ +-------------+
|  rusqlite   | | chainstate  | |    util     | | burnchains  | |    core     | |     net     |
+-------------+ +-------------+ +-------------+ +-------------+ +-------------+ +-------------+
       ^               ^               ^               ^               ^               ^
       |               |               |               |               |               |
       |-------------------------------------------------------------------------------
       |                   |
+-------------+     +-------------+   
| clarity_vm  |<--->|     vm      |
+-------------+     +-------------+

Eventual Target:
                                        -----------------------------------------------
                                       |                                               v 
+-------------+ +-------------+ +-------------+ +-------------+ +-------------+ +-------------+
|  rusqlite   | | chainstate  | |    util     | | burnchains  | |    core     | |     net     |
+-------------+ +-------------+ +-------------+ +-------------+ +-------------+ +-------------+
       ^               ^               ^               ^               ^               ^
       |               |               |               |               |               |
       |-------------------------------------------------------------------------------
       |                   
+-------------+     +-------------+   
| clarity_vm  |---->|     vm      |
+-------------+     +-------------+
```

Changes:
- created new `clarity_vm` module for Stacks specific code for utilizing the Clarity VM
- moved most of the `MARF` related code in `vm` to `clarity_vm`
- moved `clarity.rs` from `vm` to `clarity_vm`
- hoisted various `chainstate` types and constants into new `types` module
- new `boot.rs` file in `util` module. moved some boot contract related helper functions into there

## Next Steps
Currently `vm` still has some dependencies on `chainstate` and the newly created `clarity_vm`, namely on `MemoryBackingStore` and `MarfedKV`. These dependencies can mostly be eliminated by moving tests into `clarity_vm`.

Future refactoring work (make issues for these):
- [ ] Excise `Environment::handle_poison_microblock()`
- [ ] figure out what to do with dependencies on `MemoryBackingStore` and `MarfedKV`
- [ ] figure out how to get rid of `vm`'s dependency on `rusqlite`
- [ ] deal with `vm::functions::special::handle_pox_api_contract_call()`
- [ ] move `C32_ADDRESS_VERSION_...` constants out of chainstate
- [ ] deal with `chainstate::stacks::boot::STACKS_BOOT_CODE_MAINNET` in `vm::docs`
- [ ] hoist remainint constants and types out of `chainstate` to eliminate it as a dependency from `vm`


benches/block_limits.rs
```diff
@@ -2,33 +2,33 @@ extern crate blockstack_lib;
...
-    vm::database::MarfedKV,
     vm::{
         database::{HeadersDB, NULL_BURN_STATE_DB},
         types::{PrincipalData, QualifiedContractIdentifier, StandardPrincipalData},
         Value,
     },
+    vm::clarity::ClarityInstance,
+    vm::costs::ExecutionCost,
 };
+use blockstack_lib::vmlib::database::marf::MarfedKV;
```
lgalabru, line 26: what about `clarity_vm`?

lgalabru:
thank you for opening this PR @reedrosenbluth! do you think you can share a rough graph of our dependency graph today, how the graph should look like after this PR, and what's our final target? That would be super helpful for reviewing this PR.

reedrosenbluth:
@lgalabru yes absolutely, will work on this!

kantai:
@reedrosenbluth -- can you make sure that the various test scripts we use still pass with these changes?

reedrosenbluth:
@kantai yep, working on this now

src/clarity_vm/database/marf.rs
```diff
@@ -826,11 +546,11 @@ impl<'a> ClarityBackingStore for WritableMarfStore<'a> {
             })
     }
 
-    fn get_with_proof(&mut self, key: &str) -> Option<(String, TrieMerkleProof<StacksBlockId>)> {
+    fn get_with_proof(&mut self, key: &str) -> Option<(String, String)> {
```
jcnelson, line 549: Why is this changed to a `String`?

jcnelson:
So, this overall looks good to me, but I do want some clarity (no pun intended) on why the API changed to emitting the serialized `TrieMerkleProof` instead of the structured `TrieMerkleProof`.

src/clarity_vm/database/marf.rs
reedrosenbluth: This change was made in order to remove an instance of the `vm`'s dependency on `chainstate`, and by extension, `rusqlite`.

```Rust
// chainstate/stacks/index/proofs.rs
pub struct TrieMerkleProof<T: MarfTrieId>(pub Vec<TrieMerkleProofType<T>>);

// chainstate/stacks/index/mod.rs
pub trait MarfTrieId:
    PartialEq
    + Clone
    + std::fmt::Display
    + std::fmt::Debug
    + rusqlite::types::ToSql
    + rusqlite::types::FromSql
    + std::convert::From<[u8; 32]>
    + std::convert::From<MARFValue>
    + ::net::StacksMessageCodec
{
...
}
```

Changing the return type of the `get_with_proof` function was a simple way to get rid of these dependencies. Another solution would be to put the `ToSql` and `FromSql` trait inheritance behind a feature flag, and hoist the `MarfTrieId` type out of `chainstate` and into the new `types` module.

jcnelson: Why is it a problem that the `clarity_vm` crate depend on `rusqlite`?  Doesn't it do so already?  Can't you just hoist `MarfTrieId` into the `types` module?

kantai: I think we can worry about the structured merkle proof response when we actually need them -- for now, the only use case of this method immediately serializes the proof (and this change doesn't actually destroy the structure of the response: it could be deserialized). Once we're implementing things that actually want these proofs, we'll have a much better idea about how to change the response (does an associated type make sense here for example? a non-marf implementation of the clarity db, for example, probably would want to return a totally different type).

pavitthrap (commented):
LGTM!

With the goal of making the output of rustdoc more useful for this repository, could you add a description of the modules you modified/added (`vm`, `clarity_vm`, `clarity`, and `database`). For example, in `src/lib.rs`, something like 
```
/// This module is ___
pub mod clarity_vm;
```

pavitthrap (approved):
🎉🎉🎉

src/clarity_vm/database/marf.rs
reedrosenbluth: @jcnelson I went ahead and implemented this as it wasn't that much of an effort. @kantai given that I've made this change, do you still have a preference either way?

kantai: Yep, that works for me

lgalabru (approved):
Looks good to me, thank you for making some progress on that front @reedrosenbluth!

kantai (approved):
LGTM!

jcnelson (approved):
LGTM! :tada:

PR #2590, opened 2021-04-13
Fix/atlas replication retry

jcnelson:
Make it so Atlas chunks are re-tried in an exponential back-off fashion with random jitters.  Also, reduce the inv reward cycle sync to 3 reward cycles by default on mainnet (you can get away with just 1 in practice -- I've tried it several times), and fix the `new_bytes` calculation in microblock mining.

### DIFF
--- a/src/chainstate/stacks/db/unconfirmed.rs
+++ b/src/chainstate/stacks/db/unconfirmed.rs
@@ -57,6 +57,10 @@ pub struct UnconfirmedState {
     readonly: bool,
     dirty: bool,
     num_mblocks_added: u64,
+
+    // fault injection for testing
+    pub disable_cost_check: bool,
+    pub disable_bytes_check: bool,
 }
 
 impl UnconfirmedState {
@@ -85,6 +89,9 @@ impl UnconfirmedState {
             readonly: false,
             dirty: false,
             num_mblocks_added: 0,
+
+            disable_cost_check: check_fault_injection(FAULT_DISABLE_MICROBLOCKS_COST_CHECK),
+            disable_bytes_check: check_fault_injection(FAULT_DISABLE_MICROBLOCKS_BYTES_CHECK),
         })
     }
 
@@ -115,6 +122,9 @@ impl UnconfirmedState {
             readonly: true,
             dirty: false,
             num_mblocks_added: 0,
+
+            disable_cost_check: check_fault_injection(FAULT_DISABLE_MICROBLOCKS_COST_CHECK),
+            disable_bytes_check: check_fault_injection(FAULT_DISABLE_MICROBLOCKS_BYTES_CHECK),
         })
     }
 
@@ -237,6 +247,16 @@ impl UnconfirmedState {
         self.bytes_so_far += new_bytes;
         self.num_mblocks_added += num_new_mblocks;
 
+        // apply injected faults
+        if self.disable_cost_check {
+            warn!("Fault injection: disabling microblock miner's cost tracking");
+            self.cost_so_far = ExecutionCost::zero();
+        }
+        if self.disable_bytes_check {
+            warn!("Fault injection: disabling microblock miner's size tracking");
+            self.bytes_so_far = 0;
+        }
+
         Ok((total_fees, total_burns, all_receipts))
     }
 
@@ -310,6 +330,14 @@ impl UnconfirmedState {
         txid: &Txid,
     ) -> Option<(StacksTransaction, BlockHeaderHash, u16)> {
         self.mined_txs.get(txid).map(|x| x.clone())
+    }
+
+    pub fn num_microblocks(&self) -> u64 {
+        if self.last_mblock.is_some() {
+            (self.last_mblock_seq as u64) + 1
+        } else {
+            0
+        }
     }
 }
 
--- a/src/chainstate/stacks/db/unconfirmed.rs
+++ b/src/chainstate/stacks/db/unconfirmed.rs
@@ -207,7 +207,7 @@ impl UnconfirmedState {
 
                 last_mblock = Some(mblock_header);
                 last_mblock_seq = seq;
-                new_bytes = {
+                new_bytes += {
                     let mut total = 0;
                     for tx in mblock.txs.iter() {
                         let mut bytes = vec![];
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -348,6 +348,7 @@ impl<'a> StacksMicroblockBuilder<'a> {
         return Ok(false);
     }
 
+    /// NOTE: this is only used in integration tests.
     pub fn mine_next_microblock_from_txs(
         &mut self,
         txs_and_lens: Vec<(StacksTransaction, u64)>,
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -65,6 +65,10 @@ struct MicroblockMinerRuntime {
     considered: Option<HashSet<Txid>>,
     num_mined: u64,
     tip: StacksBlockId,
+
+    // fault injection, inherited from unconfirmed
+    disable_bytes_check: bool,
+    disable_cost_check: bool,
 }
 
 #[derive(PartialEq)]
@@ -87,6 +91,9 @@ impl From<&UnconfirmedState> for MicroblockMinerRuntime {
             considered: Some(considered),
             num_mined: 0,
             tip: unconfirmed.confirmed_chain_tip.clone(),
+
+            disable_bytes_check: unconfirmed.disable_bytes_check,
+            disable_cost_check: unconfirmed.disable_cost_check,
         }
     }
 }
@@ -386,6 +393,16 @@ impl<'a> StacksMicroblockBuilder<'a> {
             }
         }
 
+        // do fault injection
+        if self.runtime.disable_bytes_check {
+            warn!("Fault injection: disabling miner limit on microblock stream size");
+            bytes_so_far = 0;
+        }
+        if self.runtime.disable_cost_check {
+            warn!("Fault injection: disabling miner limit on microblock runtime cost");
+            clarity_tx.reset_cost(ExecutionCost::zero());
+        }
+
         self.runtime.bytes_so_far = bytes_so_far;
         self.clarity_tx.replace(clarity_tx);
         self.runtime.considered.replace(considered);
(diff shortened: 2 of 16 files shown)


src/chainstate/stacks/db/unconfirmed.rs
```diff
@@ -207,7 +207,7 @@ impl UnconfirmedState {
 
                 last_mblock = Some(mblock_header);
                 last_mblock_seq = seq;
-                new_bytes = {
+                new_bytes += {
```
kantai, line 210: Should we add a regression test for this?

jcnelson: Yes, we should.

src/net/atlas/download.rs
```diff
@@ -43,6 +65,29 @@ impl AttachmentsDownloader {
         }
     }
 
+    pub fn has_ready_batches(&self) -> bool {
+        for batch in self.priority_queue.iter() {
```
lgalabru, line 69: I think we need to have some kind of symmetry between this method, traversing the priority queue and, the method  `pop_next_ready_batch` just `peek`ing.
With this approach, you could end up returning `None`, which would result in a panic because of the expect, line 137.

jcnelson: I'm not sure I follow? The previous code used `priority_queue.is_empty()` followed by a `priority_queue.pop().expect(..)`.  If anything, I'm adhering to the semantics that were already there.

lgalabru: I could be missing something, but I think that this new conditions are equivalent to writing:
```
queue = [1, 1, 0]
if queue.contains(1) {
    assert!(queue.last() == 1);
}
```

jcnelson: I changed the implementation of `Ord` for `AttachmentsBatch` so that the "first" `AttachmentsBatch` is the one with the earliest deadline.  If `has_ready_batches()` returns `true`, then not only is `self.priority_queue.len() > 0`, but also `self.priority_queue.front().as_ref().unwrap().retry_deadline < get_epoch_time_secs()`.

src/chainstate/stacks/db/unconfirmed.rs
jcnelson: I think this test can actually be expanded.  The difficulty with testing microblock mining is that the current Neon node design makes it very difficult to construct invalid or malicious microblock streams; the bugs we've had with microblocks stem from the node confirming an invalid stream.

I think I've found a better way to test this through fault injection.  Going to spend cycles tomorrow doing this not only for size, but also compute budget.

src/net/atlas/download.rs
```diff
@@ -88,8 +133,7 @@ impl AttachmentsDownloader {
                 }
 
                 let attachments_batch = self
```
kantai, line 135: With the new interface methods, I think this expect can be avoided (and combined with the `is_empty` check) above while maintaining the same behavior:

```rust
let attachments_batch = match self.pop_next_ready_batch() {
    None => {
         warn!("Atlas: could not get a peer to sync with");
         // Nothing can be done!
         return Ok((vec![], vec![]));
    },
    Some(ready_batch) => ready_batch
}
```

While I think this is pretty much the exact same behavior, I think we're better off avoiding the `expect` if we can.

kantai:
It looks like the microblock size-check integration test in sampled-genesis is failing.

src/chainstate/stacks/db/unconfirmed.rs
```diff
@@ -237,6 +247,16 @@ impl UnconfirmedState {
         self.bytes_so_far += new_bytes;
         self.num_mblocks_added += num_new_mblocks;
 
+        // apply injected faults
```
pavitthrap, line 250: Can you separate this code out into a separate function to make it clearer that it is "testing" code?

src/chainstate/stacks/miner.rs
```diff
@@ -386,6 +393,16 @@ impl<'a> StacksMicroblockBuilder<'a> {
             }
         }
 
+        // do fault injection
```
pavitthrap, line 396: same here about separate function

src/net/atlas/download.rs
pavitthrap: could you add a comment for the fn `pop_next_ready_batch` explaining this?

src/net/atlas/download.rs
```diff
@@ -1048,6 +1094,14 @@ impl AttachmentsBatch {
 
     pub fn bump_retry_count(&mut self) {
         self.retry_count += 1;
+        let delay = cmp::min(
+            MAX_RETRY_DELAY,
+            2u64.saturating_pow(self.retry_count as u32)
+                + (thread_rng().gen::<u64>() % 2u64.saturating_pow((self.retry_count - 1) as u32)),
```
pavitthrap, line 1100: use `saturated_add`?

src/net/atlas/mod.rs
```diff
@@ -20,6 +36,7 @@ use std::convert::TryFrom;
 use std::hash::{Hash, Hasher};
 
 pub const MAX_ATTACHMENT_INV_PAGES_PER_REQUEST: usize = 8;
+pub const MAX_RETRY_DELAY: u64 = 600;
```
pavitthrap, line 39: Perhaps add a comment about unit of time here (seconds?)

testnet/stacks-node/src/neon_node.rs
```diff
@@ -375,92 +384,82 @@ fn try_mine_microblock(
     chainstate: &mut StacksChainState,
     sortdb: &SortitionDB,
     mem_pool: &MemPoolDB,
-    winning_tip_opt: Option<&(ConsensusHash, BlockHeaderHash, Secp256k1PrivateKey)>,
+    winning_tip: (ConsensusHash, BlockHeaderHash, Secp256k1PrivateKey),
 ) -> Result<Option<StacksMicroblock>, NetError> {
+    let ch = winning_tip.0;
```
pavitthrap, line 389: style nit: pattern match whole tuple at once

testnet/stacks-node/src/neon_node.rs
```diff
@@ -87,9 +88,9 @@ struct MicroblockMinerState {
 enum RelayerDirective {
     HandleNetResult(NetworkResult),
     ProcessTenure(ConsensusHash, BurnchainHeaderHash, BlockHeaderHash),
-    RunTenure(RegisteredKey, BlockSnapshot),
+    RunTenure(RegisteredKey, BlockSnapshot, u128),
```
pavitthrap, line 91: Add comment for `RunTenure` and `RunMicroblockTenure` described what data is in tuple (like what is the `u128` for).

src/chainstate/stacks/db/unconfirmed.rs
jcnelson: We can't :(  This is because the `#[cfg(test)]` pragma doesn't apply across crates.  The `blockstack_lib` crate is built without `#[cfg(test)]` when running integration tests.

testnet/stacks-node/src/neon_node.rs
jcnelson: Rust doesn't let you destruct tuples if the lhs is a `let`-binding.

testnet/stacks-node/src/neon_node.rs
```diff
@@ -914,10 +930,13 @@ fn spawn_miner_relayer(
...
                                     // proceed to mine microblocks
                                     debug!(
-                                        "Microblock miner tip is now {}/{}",
-                                        &consensus_hash, &block_header_hash
+                                        "Microblock miner tip is now {}/{} ({})",
+                                        &consensus_hash, &block_header_hash, StacksBlockHeader::make_index_block_hash(&consensus_hash, &block_header_hash)
                                     );
                                     miner_tip = Some((ch, bh, microblock_privkey));
+
+                                    Relayer::refresh_unconfirmed(&mut chainstate, &mut sortdb);
```
pavitthrap, line 938: I'm confused why we'd want to call `refresh_unconfirmed` (and also `send_unconfirmed_txs` for `RelayerDirective::ProcessTenure` - could you explain?

testnet/stacks-node/src/neon_node.rs
```diff
@@ -1680,6 +1738,52 @@ impl InitializedNeonNode {
...
             Ok(block) => block,
+            Err(ChainstateError::InvalidStacksMicroblock(msg, mblock_header_hash)) => {
+                // part of the parent microblock stream is invalid, so try again
+                info!("Parent microblock stream is invalid; trying again without the offender {} (msg: {})", &mblock_header_hash, &msg);
+
+                // truncate the stream
+                stacks_parent_header.microblock_tail = match microblock_info_opt {
+                    Some((microblocks, _)) => {
+                        let mut tail = None;
+                        for mblock in microblocks.into_iter() {
```
pavitthrap, line 1749: Maybe the iteration logic here could be avoided by modifying `process_microblocks_transactions` (which seems is like what is ultimately returning the data for the `InvalidStacksMicroblock` error here) to return an option with the previous microblock hash?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1813,6 +1844,372 @@ fn size_overflow_unconfirmed_microblocks_integration_test() {
...
+                    &format!("small-{}", i),
+                    &small_contract,
+                );
+                ret.push(tx);
+            }
+            ret
+        })
+        .collect();
+
+    let flat_txs: Vec<_> = txs.iter().fold(vec![], |mut acc, a| {
```
pavitthrap, line 1884: There's a [`flat_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map) function that could be helpful here.

testnet/stacks-node/src/neon_node.rs
jcnelson: This code path is invoked when the node wins the block race.  This will invalidate all unconfirmed microblocks, because the canonical Stacks chain tip will have changed (and the node only tracks unconfirmed microblocks for the canonical tip).  Therefore, we need to call `refresh_unconfirmed()` in order to reset the unconfirmed microblock state.  This in turn has the intended effect of changing which transactions are represented in the unconfirmed microblock stream (via `inner_process_tenure()`, which _may_ process _multiple_ Stacks blocks), so we must then call `send_unconfirmed_txs()` to synchronize them with the p2p thread.

testnet/stacks-node/src/neon_node.rs
jcnelson: As a general rule, I try to write methods that only return data the caller can't figure out on its own; otherwise, the method could be doing excess work that the caller doesn't need.

jcnelson:
Looks like the antientropy test is stalling out for some reason.  Will take a look tomorrow.

jcnelson:
I think this is ready for merging, if there's no other review comments.

kantai (approved):
These changes look good to me -- however, you should add an entry to the changelog for this before merging.

I'm also concerned that the integration tests now take 84 minutes to run -- while they weren't exactly speedy before (around 30 minutes), the extra 54 minutes will make PRs and releases pretty painful. This doesn't need to be solved immediately, but if it seems like something that can be solved pretty quickly, then I think it's worth taking the time to do so.

lgalabru (approved):
LGTM

PR #2603, opened 2021-04-21
Remove mine_microblock from config

lgalabru:


pavitthrap:
Do you maybe also want to remove mentions of `mine_microblocks` in `net-test/bin/start.sh` (I don't know if this is in current use, I just saw mentions there).

jcnelson (changes requested):
I don't think this is something we want to do.  If we needed to disable microblock mining in a hurry, having a config setting available in the config file would make this very straightforward -- we wouldn't need to cut a release, for example.  I think we should instead plumb through the `mine_microblocks` setting into the node.

PR #2631, opened 2021-04-27
Update documentation of define-map

friedger:
This PR is a small update of the documentation of `define-map`. Clarifying that maps take any key types and any value types.


kantai (approved):
Thanks for patching this!

pavitthrap (approved):
LGTM!

PR #2634, opened 2021-04-28
fix: v2/pox endpoint for mocknet

### LINKED ISSUE #2600: Mocknet pox endpoint not working
## Describe the bug

`/v2/pox` endpoint not working when running in mocknet.

## Steps To Reproduce

Booting mocknet on develop 7ba919a52, running

`cargo run --release --bin=stacks-node --package=stacks-node --  mocknet`

Waited for it to start producing coinbase blocks.

`curl "http://127.0.0.1:20443/v2/pox"` Responds with 500 error

```
HTTP/1.1 500 Internal Server Error
Access-Control-Allow-Headers: origin, content-type
Access-Control-Allow-Methods: POST, GET, OPTIONS
Access-Control-Allow-Origin: *
Content-Length: 25
Content-Type: text/plain
Date: Wed, Apr 21 2021 11:55:56 GMT
Server: stacks/2.0
X-Request-Id: 2564431540

Failed to query peer info
```

cc/ @zone117x

pavitthrap:
## Description
Fixes #2600 

Added constant `MAINNET_2_0_GENESIS_ROOT_HASH`, which is the expected genesis hash on mainnet. Added assertion for this value in `install_boot_code`. 

In addition, added an env var in circle CI named `CIRCLE_CI_TEST` to denote tests that should be run in circle ci but not locally.

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

## Testing information
Running `curl "http://127.0.0.1:20443/v2/pox"` after `cargo run --release --bin=stacks-node --package=stacks-node -- mocknet` no longer failed.


src/net/rpc.rs
```diff
@@ -324,15 +324,15 @@ impl RPCPoxInfoData {
 
         let pox_consts = &burnchain.pox_constants;
 
-        if prepare_cycle_length != pox_consts.prepare_length as u64 {
+        if mainnet && prepare_cycle_length != pox_consts.prepare_length as u64 {
```
jcnelson, line 327: While this would work insofar as stopping the bug from manifesting, I don't think this is correct.  Why is `pox_consts` not getting initialized to the same values as the result of calling `get-pox-info`?  I think they really need to be the same, since clients that use `/v2/info` to learn about the PoX configuration expect that this endpoint reflects how the `.pox` contract actually behaves.

pavitthrap: We are using the PoxConstants that are defined for regtest for the mocknet. These are defined in `src/burnchain/mod.rs`: 
```
pub fn regtest_default() -> PoxConstants {
        PoxConstants::new(
            5,
            1,
            1,
            3333333333333333,
            1,
            BITCOIN_REGTEST_FIRST_BLOCK_HEIGHT + POX_SUNSET_START,
            BITCOIN_REGTEST_FIRST_BLOCK_HEIGHT + POX_SUNSET_END,
        )
    }
```

I can change it so that the constants for regtest (and testnet?) match the values obtained from calling `get-pox-info`.

jcnelson: It's the other way around -- you'll need to make sure that the `.pox` contract is initialized with these `PoxConstant` values.  Then this code path should work without modification.

src/chainstate/stacks/db/mod.rs
```diff
@@ -672,6 +673,7 @@ pub struct ChainStateBootData {
     pub first_burnchain_block_height: u32,
     pub first_burnchain_block_timestamp: u32,
     pub initial_balances: Vec<(PrincipalData, u64)>,
+    pub pox_constants: Option<PoxConstants>,
```
jcnelson, line 676: I don't think this should be `Option<..>`.  I think the boot data should always have PoX constants.

pavitthrap: addressed this

src/chainstate/stacks/db/mod.rs
```diff
@@ -1205,6 +1208,27 @@ impl StacksChainState {
                 callback(&mut clarity_tx);
             }
 
+            // Setup burnchain parameters for pox contract
```
jcnelson, line 1211: Minor nit -- can you put this *before* the call to `post_flight_callback`?   This code used to be the post-flight callback; we might even consider removing the `post_flight_callback` now that the code lives here.

jcnelson:
The failing unit test -- `test_chainstate_sampled_genesis_consistency` -- is breaking because it does _not_ expect the `post_flight_callback` to be invoked.  There's two ways to go about this:

* Put the PoX contract setup logic back into a callback function, and make that function the _default value_ of the `post_flight_callback` function in `ChainStateBootData`.  Then, have `test_chainstate_sampled_genesis_consistency` simply set this callback to `None` before running.  The resulting MARF root hash should then match.

* Update the `test_chainstate_sampled_genesis_consistency` hash so it reflects the instantiation of the PoX smart contract.

I'm biased towards the former because figuring out the latter will be extremely tedious.  Sorry that this is a round-about process!

kantai:
> The failing unit test -- test_chainstate_sampled_genesis_consistency -- is breaking because it does not expect the post_flight_callback to be invoked. There's two ways to go about this:

My preference would be us dealing with it as follows:

1. Update the `test_chainstate_sampled_genesis_consistency` hash value to whatever the current value is.
2. Update the `test_chainstate_full_genesis_consistency` test in two ways:
    a. Remove the `#[ignore]` and instead gate it with an envvar that we set in CircleCI. This test shouldn't run by default locally, but we should run it in CI.
    b. Update the test itself so that it expects _exactly_ the actual Stacks 2.0 genesis hash. If that test fails, it should block any PR.

pavitthrap:
> > The failing unit test -- test_chainstate_sampled_genesis_consistency -- is breaking because it does not expect the post_flight_callback to be invoked. There's two ways to go about this:
> 
> My preference would be dealing with it as follows:
> 
> 1. Update the `test_chainstate_sampled_genesis_consistency` hash value to whatever the current value is.
> 2. Update the `test_chainstate_full_genesis_consistency` test in two ways:
>    a. Remove the `#[ignore]` and instead gate it with an envvar that we set in CircleCI. This test shouldn't run by default locally, but we should run it in CI.
>    b. Update the test itself so that it expects _exactly_ the actual Stacks 2.0 genesis hash. If that test fails, it should block any PR.

@jcnelson thoughts on Aaron's approach?

jcnelson:
Point (2) is fine by me.  Point (1) is the part I still have questions about.  How did you independently verify that the genesis hash was `9653c92b1ad726e2dc17862a3786f7438ab9239c16dd8e7aaba8b0b5c34b52af` in the mainnet test?

kantai:
> How did you independently verify that the genesis hash was 9653c92b1ad726e2dc17862a3786f7438ab9239c16dd8e7aaba8b0b5c34b52af in the mainnet test?

You can verify the genesis root hash by applying this patch:

```diff
diff --git a/src/chainstate/stacks/db/mod.rs b/src/chainstate/stacks/db/mod.rs
index 5f7c9c159..005d8fc70 100644
--- a/src/chainstate/stacks/db/mod.rs
+++ b/src/chainstate/stacks/db/mod.rs
@@ -1218,6 +1218,18 @@ impl StacksChainState {
             clarity_tx.commit_to_block(&FIRST_BURNCHAIN_CONSENSUS_HASH, &FIRST_STACKS_BLOCK_HASH);
         }
 
+        {
+            let genesis_root_hash = chainstate.clarity_state.with_marf(|marf| {
+                let index_block_hash = StacksBlockHeader::make_index_block_hash(
+                    &FIRST_BURNCHAIN_CONSENSUS_HASH,
+                    &FIRST_STACKS_BLOCK_HASH,
+                );
+                marf.get_root_hash_at(&index_block_hash).unwrap()
+            });
+
+            info!("Computed Clarity state genesis"; "root_hash" => %genesis_root_hash);
+        }
+
         {
             // add a block header entry for the boot code
             let mut tx = chainstate.index_tx_begin()?;
```

And running `stacks-node mainnet`

jcnelson:
> And running stacks-node mainnet -- maybe the last change here is to add this as a constant in core/mod.rs and assert!/panic! on boot if the computed state genesis doesn't match?

Yes, I like this approach.  Then we'll know really quickly if the boot code changes here broke something when we go to test it.

kantai (approved):
LGTM!

jcnelson (approved):
LGTM!

Issue #2607, opened 2021-04-26
Epoch switch: gate contract-calls by epoch


jcnelson:
Certain contract-calls will have different behaviors depending on which epoch they are in.  These contract-call sites need to be updated to be aware of the epoch switch.  These include:

* special calls to PoX stacking operations
* BNS name renewals

...etc...

This issue covers just the gating logic.  Other issues will cover gating specific contract-calls.

jcnelson:
Intead, we're gating the VM itself.

PR #2638, opened 2021-04-30
Feat/stacks 2.1 gate contract calls

jcnelson:
This PR implements #2607.  It provides the boilerplate infrastructure for replacing a well-known `contract-call?`-able function with a native implementation, optionally gated by which epoch the system is running in.

The PR works by associating a Stacks epoch ID with the `GlobalContext`, as well as a `HashMap` that binds fully-qualified function identifiers to native methods that do their work.  In doing so, any time the function is called -- either through `contract-call?` or even through a normal function invocation -- the native method is used instead.  It does so by making it so the `execute_apply()` method in `DefinedFunction` may take an optional function pointer to the native method -- if given, the `execute_apply()` method calls it instead.

Most of the diff is plumbing through a handful of API changes.  In particular, you have to supply an epoch ID to `OwnedEnvironment` when constructing it.  The new features are mostly in `src/vm/extensions/mod.rs`, `src/vm/callables.rs`, and `src/vm/contexts.rs`, with tests in `src/vm/tests/extensions.rs`.  The test contains a very simple end-to-end proof-of-concept whereby an extension contract gets defined in `src/chainstate/boot/ext-2_1.clar`, and overridden with a native implementation with the same type signature in `src/vm/extensions/mod.rs` if the `GlobalContext` is running in the 2.1 epoch.

I did it this way because this is essentially the same logic that we'll be doing to replace well-known public Clarity methods with native implementations as part of cost-voting.  Replacing an extension contract's body, deployed with a no-op Clarity body, with a native body is really just a special case of this.


jcnelson:
Switching this to draft status until I can figure out why the remaining unit tests are failing.

jcnelson:
Don't worry about reviewing this; we'll just hold this on stand-by in case we can't do #2643 for some (unlikely) reason.

Issue #2611, opened 2021-04-26
Epoch switch: instantiate the PoX 2.1 contract and extension contracts at the epoch boundary


jcnelson:
When the Stacks 2.1 epoch switch happens, the version of the PoX contract will need to be instantiated, as well as any extension contracts.

kantai:
I'm currently working on the process of PoX 2.1 contract instantiation. The extension contracts are somewhat similar, but probably sufficiently different that they'll need their own issue as well.

Issue #2610, opened 2021-04-26
Epoch switch: PoX unlock/relock


jcnelson:
When the Stacks 2.1 epoch goes live, the _next reward cycle_ will use the Stacks 2.1 PoX contract to calculate the reward set.  The 2.0 contract will be ignored.  Users will have a window of time (TBD) to re-lock their STX into the 2.1 contract.  This obviates the need to migrate state from the 2.0 contract, which will be extremely expensive and would have to happen each time two Stacks blocks are mined that cross the epoch boundary.

PR #2639, opened 2021-04-30
Stacks 2.1: epoch initialization

kantai:
This PR implements a basic structure for doing contract initialization at the "beginning" of Epoch 2.1. Importantly, "beginning" here isn't a one-time event: any Stacks block that is the first Stacks block in its fork after the epoch _must_ call the initialization method. Right now, the initialization method only reinitializes the original PoX contract as "pox-2", but it would eventually need to do more than just initialize the new PoX contract (e.g., initialize analysis storage for extension functions). This PR also changes the #2659 behavior so that Clarity version 2 default check uses the epoch transition application rather than using a burn db lookup.

The other things implemented in this PR is the start of the PoX 1 -> PoX 2 transition. Right now, it implements:

#2610 - automatic unlock for PoX 1 lockups before the PoX 2 reward cycles begin.
#2611 - instantiation of PoX 2 at the epoch boundary, and the transition to PoX 2 reward cycles after the v1 automatic unlock.

This PR adds an end-to-end test case for the epoch and PoX transitions.


src/burnchains/mod.rs
```diff
@@ -342,13 +347,23 @@ impl PoxConstants {
...
     pub fn test_default() -> PoxConstants {
         // 20 reward slots; 10 prepare-phase slots
-        PoxConstants::new(10, 5, 3, 25, 5, 5000, 10000)
+        PoxConstants::new(10, 5, 3, 25, 5, 5000, 10000, u32::max_value())
+    }
+
+    /// Returns the PoX contract that is "active" at the given burn block height
+    pub fn active_pox_contract(&self, burn_height: u64) -> &'static str {
+        if burn_height >= (self.v1_unlock_height as u64) {
+            "pox-2"
```
jcnelson, line 363: Do we want to label the PoX contract by the Stacks major/minor version, or by the contract version?  i.e. `pox-2` vs `pox-2.1`?  Either one is fine with me, just as long as it's consistent.

kantai: My preference is by the contract version -- it's not the case that future versions of the Stacks chain (if they exist) will necessarily create new PoX contracts, so it'd be kind of confusing that `Stacks 2.9` uses `pox-2.4` (or whatever the versioning mismatch becomes).

kantai: However, I could be convinced otherwise if people believe one or the other will create more confusion.

src/chainstate/stacks/boot/mod.rs
```diff
@@ -813,6 +835,34 @@ pub mod test {
...
+        addr_bytes: Hash160,
+        lock_period: u128,
+        burn_ht: u64,
+    ) -> StacksTransaction {
+        // (define-public (stack-stx (amount-ustx uint)
+        //                           (pox-addr (tuple (version (buff 1)) (hashbytes (buff 20))))
+        //                           (lock-period uint))
+        let payload = TransactionPayload::new_contract_call(
+            boot_code_test_addr(),
+            "pox-2",
```
jcnelson, line 852: Minor style nit -- should probably have the contract name as a constant somewhere.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -0,0 +1,322 @@
...
+            let reward_addrs = with_sortdb(&mut peer, |ref mut chainstate, ref sortdb| {
+                get_reward_addresses_with_par_tip(chainstate, &burnchain, sortdb, &tip_index_block)
+            })
+            .unwrap();
+            assert_eq!(reward_addrs.len(), 0);
+
+            // record the first reward cycle when Alice's tokens get stacked
+            let tip_burn_block_height =
+                get_par_burn_block_height(peer.chainstate(), &tip_index_block);
+            alice_reward_cycle = 1 + burnchain
```
jcnelson, line 252: I think this is going to get assigned twice, since `tenure_id <= 1` is true twice?  But, I don't think the value is different either way, since `1 + burnchain.block_height_to_reward_cycle(tip_burn_block_height)` will be the same in both cases?  Can we confirm?

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -0,0 +1,322 @@
...
+             ref parent_opt,
+             ref parent_microblock_header_opt| {
+                let parent_tip = get_parent_tip(parent_opt, chainstate, sortdb);
+                let coinbase_tx = make_coinbase(miner, tenure_id as usize);
+
+                let mut block_txs = vec![coinbase_tx];
+
+                if tenure_id == 1 {
+                    // Alice locks up exactly 25% of the liquid STX supply, so this should succeed.
+                    // this locks for cycles 6, 7, 8, 9.
```
jcnelson, line 116: Why is the first cycle -- i.e. at `tenure_id == 1` -- the 6th reward cycle?  Wouldn't it be the locked starting in the first reward cycle (if we start counting reward cycles from 0)?

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -0,0 +1,322 @@
...
+                        &alice,
+                        1,
+                        512 * POX_THRESHOLD_STEPS_USTX,
+                        AddressHashMode::SerializeP2PKH,
+                        key_to_stacks_addr(&alice).bytes,
+                        12,
+                        tip.block_height,
+                    );
+                    block_txs.push(alice_lockup);
+                } else if tenure_id == 31 {
```
jcnelson, line 166: Is this condition reachable?  It looks like `tenure_id` maxes out at 25?

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -0,0 +1,322 @@
...
+                        0,
+                        512 * POX_THRESHOLD_STEPS_USTX,
+                        AddressHashMode::SerializeP2PKH,
+                        key_to_stacks_addr(&bob).bytes,
+                        6,
+                        tip.block_height,
+                    );
+                    block_txs.push(bob_lockup);
+                } else if tenure_id == 11 {
+                    // Bob tries to lock in PoX v1 too, but it shouldn't work!
```
jcnelson, line 143: I'm not sure this is checked below?  We're checking the same address in the reward set either way, so if either `make_pox_2_lockup()` or `make_pox_lockup()` succeed, the test will pass.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -0,0 +1,322 @@
...
+                            .unwrap(),
+                    )
+                });
+
+            eprintln!("\ntenure: {}\nreward cycle: {}\nmin-uSTX: {}\naddrs: {:?}\ntotal_liquid_ustx: {}\ntotal-stacked: {}\n", tenure_id, cur_reward_cycle, min_ustx, &reward_addrs, total_liquid_ustx, total_stacked);
+
+            if cur_reward_cycle >= alice_reward_cycle {
+                if cur_reward_cycle < first_v2_cycle as u128 {
+                    let (amount_ustx, pox_addr, lock_period, first_reward_cycle) =
+                        get_stacker_info(&mut peer, &key_to_stacks_addr(&alice).into()).unwrap();
```
jcnelson, line 287: Does `get_stacker_info()` know to read `.pox-2` if the height is in the 2.1 epoch?

src/chainstate/stacks/boot/pox_2_tests.rs
jcnelson: Also, how would this error manifest if Bob called into `.pox` when 2.1 rolls around?  Would the block itself be unmineable because it calls into a contract that is no longer "active," or is this just a runtime error in the transaction, or is Bob's call re-routed to the `.pox-2` contract automatically?

src/chainstate/stacks/db/accounts.rs
```diff
@@ -287,20 +279,58 @@ impl StacksChainState {
...
+        unlock_burn_height: u64,
+    ) -> Result<(), Error> {
+        assert!(unlock_burn_height > 0);
+        assert!(lock_amount > 0);
+
+        let mut snapshot = db.get_stx_balance_snapshot(principal);
+
+        if snapshot.balance().was_locked_by_v2() {
+            debug!("PoX Lock attempted on an account locked by v2");
+            return Err(Error::DefunctPoxContract);
```
jcnelson, line 317: Would it make sense to call this error `DefunctContract`?  I suspect we'll replace other contracts this way as well as time goes on.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -3889,6 +3889,49 @@ impl StacksChainState {
...
+                        );
+                        clarity_tx.block.initialize_epoch_2_1()?;
+                    }
+                    StacksEpochId::Epoch21 => {
+                        panic!("No defined transition from Epoch21 forward")
+                    }
+                }
+            }
+        }
+        Ok(vec![])
```
jcnelson, line 3932: Are there going to be receipts from this?

src/clarity_vm/clarity.rs
```diff
@@ -626,6 +632,56 @@ impl<'a> ClarityBlockConnection<'a> {
...
+
+        self.as_transaction(|tx_conn| {
+            tx_conn
+                .with_clarity_db(|db| {
+                    db.set_clarity_epoch_version(StacksEpochId::Epoch21);
+                    Ok(())
+                })
+                .unwrap();
+
+            let pox_2_code = if mainnet {
```
jcnelson, line 650: I take it this will be replaced by the actual `.pox-2` code at some point?

src/clarity_vm/clarity.rs
```diff
@@ -626,6 +632,56 @@ impl<'a> ClarityBlockConnection<'a> {
...
     }
 
+    pub fn initialize_epoch_2_1(&mut self) -> Result<(), Error> {
+        let mainnet = self.mainnet;
+        let first_block_height = self.burn_state_db.get_burn_start_height();
+        let pox_prepare_length = self.burn_state_db.get_pox_prepare_length();
+        let pox_reward_cycle_length = self.burn_state_db.get_pox_reward_cycle_length();
+        let pox_rejection_fraction = self.burn_state_db.get_pox_rejection_fraction();
+
+        self.as_transaction(|tx_conn| {
```
jcnelson, line 642: Is there a block cost being applied here?  If so, we probably want to disable it if instantiating `.pox-2` is expensive.

src/clarity_vm/clarity.rs
```diff
@@ -626,6 +632,56 @@ impl<'a> ClarityBlockConnection<'a> {
...
+            let pox_2_code = if mainnet {
+                &*BOOT_CODE_POX_MAINNET
+            } else {
+                &*BOOT_CODE_POX_TESTNET
+            };
+
+            let pox_2_contract_id = boot_code_id("pox-2", mainnet);
+            let (ast, _) = tx_conn
+                .analyze_smart_contract(&pox_2_contract_id, pox_2_code)
+                .unwrap();
```
jcnelson, line 659: Minor nit: might want to replace `.unwrap()` with `.expect()` or even `.map_err().unwrap()` to print out what error happened.

src/chainstate/stacks/boot/pox_2_tests.rs
kantai: I think in the TestPeer, there's 25 bitcoin blocks before the first Stacks block.

src/chainstate/stacks/boot/pox_2_tests.rs
kantai: This test does 35 tenures

src/chainstate/stacks/boot/pox_2_tests.rs
kantai: As currently implemented, it's a runtime error when someone attempts to lock via PoX after PoX 2 becomes active. If you look at the logs from this test, you'll see the tx being mined and aborted via a runtime error. But really, I should add a TestObserver to be able to check tx receipts in the test.

kantai: (This is, I think, the correct behavior in the event that we _don't_ re-route pox calls to pox-2)

src/chainstate/stacks/boot/pox_2_tests.rs
kantai: No, it doesn't -- that function always reads data from `.pox` directly.

src/vm/database/structures.rs
```diff
@@ -141,44 +152,124 @@ type Result<T> = std::result::Result<T, Error>;
...
+                    amount: amount_unlocked,
+                }
+            } else {
+                STXBalance::LockedPoxOne {
+                    amount_unlocked,
+                    amount_locked,
+                    unlock_height,
+                }
+            }
+        } else if bytes.len() == STXBalance::v2_size {
```
jcnelson, line 236: Since we're probably going to be versioning this going forward, want to just make this into something like this?

```
} else {
   assert_eq!(bytes.len(), STXBalance::v2_size, "Unrecognized serialized STXBalance");
   let version = bytes[0];
   match version {
      STXBalance::pox_2_version => {
          // ...
      }
      _ => {
          // add more versions as supported
          panic!("Bad version byte in STXBalance serialization = {}", version);
       }
    }
 }
 ```

File: src/vm/database/structures.rs [hunk 14]
```diff
@@ -218,11 +309,9 @@ impl<'db, 'conn> STXBalanceSnapshot<'db, 'conn> {
...
             .unwrap_or(STXBalance::zero());
 
-        recipient_balance.amount_unlocked =
-            recipient_balance
-                .amount_unlocked
-                .checked_add(amount)
-                .ok_or(Error::Runtime(RuntimeErrorType::ArithmeticOverflow, None))?;
+        recipient_balance
+            .checked_add_unlocked_amount(amount)
+            .ok_or(Error::Runtime(RuntimeErrorType::ArithmeticOverflow, None))?;
```
jcnelson, line 314: Is this even an error we want to recover from?  If there are this many STX piled into an address (which shouldn't be possible for thousands of years), then we probably have a show-stopping problem.

src/chainstate/stacks/boot/pox_2_tests.rs
jcnelson: Ah, yes, that's right.

src/chainstate/stacks/boot/pox_2_tests.rs
jcnelson: Yes, that sounds right -- we don't do contract-call interception for pox.

But my other point stands -- it's not clear from what is being tested which contract-call is succeeding.  I don't think you need to go so far as to use a TestObserver (unless you want to truly confirm that the tx is getting rejected at runtime) -- I think you can get away with having Bob use one BTC address in pox, and use another in pox-2, and see which one got accepted.

kantai: Hmmm, adding a TestObserver is better I think -- it allows for more direct/clear assertions (checking the PoX set is somewhat indirect for this, and there's other reasons Bob might not end up being included in the PoX1 reward set) and also richer assertions. It'll probably be useful for other PRs as well.

src/burnchains/mod.rs
jcnelson: Up to you. I'm fine with it as-is as long as we believe users won't get confused.

src/vm/database/structures.rs
kantai: But that code structure assumes that the size would remain the same in version 3, etc., which isn't necessarily case, so I'd prefer a code structure that fits more closely with the current check (i.e., without a single case match).

src/vm/database/structures.rs
kantai: This is preserving the existing behavior on unlock. This will just make the transaction invalid. We could change this to a panic, but that would alter the current behavior for this case.

jcnelson: Ah, right. This is fine then.

src/vm/database/structures.rs
jcnelson: Sounds good

src/clarity_vm/clarity.rs
kantai: Yep, exactly. The next PR, which implements `stack-extend` and `delegate-stack-extend` will be the new `pox-2` code.

src/chainstate/stacks/db/blocks.rs
kantai: Ah, yes, there should be.

src/chainstate/stacks/boot/pox_2_tests.rs
kantai: Yep -- it got assigned to twice, fixed in c46422e

src/chainstate/stacks/db/accounts.rs
kantai: I think it's more descriptive as DefunctPoxContract for now. If we use similar things for other contracts, at some point, we can change this to like `DefunctContract(QualifiedContractIdentifier)`.

src/chainstate/stacks/db/blocks.rs
kantai: Added in d6458eccc939ccf7806470012568b90a5470fda9

src/clarity_vm/clarity.rs
kantai: Yep -- made it free in d6458eccc939ccf7806470012568b90a5470fda9

src/clarity_vm/clarity.rs
kantai: Addressed in d6458eccc939ccf7806470012568b90a5470fda9

src/chainstate/stacks/db/blocks.rs
jcnelson: The receipt from `initialize_epoch_2_1()` needs to be plumbed through here.

jcnelson (approved):
This LGTM; just had one minor comment to make sure you plumb through the transaction receipt from the initialization.  Super excited for Clarity v2!

src/chainstate/stacks/db/blocks.rs
kantai: Fixed in 92caa6d

src/chainstate/stacks/boot/mod.rs
kantai: Addressed in d6458eccc939ccf7806470012568b90a5470fda9

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -0,0 +1,419 @@
...
+
+    let num_blocks = 35;
+
+    let alice = keys.pop().unwrap();
+    let bob = keys.pop().unwrap();
+    let charlie = keys.pop().unwrap();
+
+    let mut alice_reward_cycle = 0;
+
+    for tenure_id in 0u32..num_blocks {
```
gregorycoppola, line 102: I think this test would be easier to understand if it were not written in a for loop. Instead, it could be written as an alternating series of 1) execute a tenure, 2) check the relevant invariants after that tenure.

E.g.:

// Do tenure 1 by running code "code for tenure 1"
// Test what is true after tenure 1

// Do tenure 1 by running code "code for tenure 2"
// Test what is true after tenure 2

... etc


If it were done this way, without a loop, and with a more declarative syntax, I think it would have the following benefits:
1) Easier to read - locality: Right now, the logic that changes the state is not near the logic that checks the state. So, I have to scroll back and forth and try to figure out what is being checked after each test. I think it would be easier to read if you could organize the code as a sequence of statements, where what is being checked is checked right after the relevant change.
2) Easier to read - no thinking: This test has a lot of logic to understand what is tested when. It's confusing to have to do so much logic to read a test. The whole point of the test is to test the logic of the code under test, so it's harder to know what the test result should be if the test also has a lot of logic.
2)More fool-proof test: With so much logic going into the test, we run the risk of the test-logic having a bug.

gregorycoppola (commented):
Hey Aaron,

I was pretty interested in the `pox_2_tests` test. I had a few comments about that.

src/burnchains/mod.rs
```diff
@@ -282,7 +283,7 @@ pub struct BurnchainBlockHeader {
     pub timestamp: u64,
 }
 
-#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Clone)]
```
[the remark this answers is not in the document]
kantai: Serialize/Deserialize were never used for this struct, and including them would require that the new Account enum implement serialize.

src/burnchains/mod.rs
```diff
@@ -317,6 +318,10 @@ pub struct PoxConstants {
     pub sunset_end: u64,
     /// first block height of sunset phase
     pub sunset_start: u64,
+    /// The auto unlock height for PoX v1 lockups before transition to PoX v2. This
+    /// also defines the burn height at which PoX reward sets are calculated using
+    /// PoX v2 rather than v1
+    pub v1_unlock_height: u32,
```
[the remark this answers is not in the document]
kantai: It's v1 balances from that all unlock, so that's "the height at which all v1 locks expire"

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -0,0 +1,419 @@
...
+///        but is forbidden because he has already placed an account lock via PoX v2.
+///
+#[test]
+fn test_simple_pox_lockup_transition_pox_2() {
+    let AUTO_UNLOCK_HT = 12;
+    let mut burnchain = Burnchain::default_unittest(0, &BurnchainHeaderHash::zero());
+    burnchain.pox_constants.reward_cycle_length = 5;
+    burnchain.pox_constants.prepare_length = 2;
+    burnchain.pox_constants.anchor_threshold = 1;
+    burnchain.pox_constants.v1_unlock_height = AUTO_UNLOCK_HT + 25;
```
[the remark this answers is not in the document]
kantai: The simmed environment for these tests produces 25 blocks with no sortitions, which don't iterate the "tenure" count.

src/chainstate/stacks/boot/pox_2_tests.rs
kantai: Yes, I agree with that. I'll unroll the test.

kantai: Unrolled in 6c69759 -- I *think* it improved the clarity of the test.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -0,0 +1,423 @@
...
+
+    for _i in 0..7 {
+        peer.tenure_with_txs(&[], &mut coinbase_nonce);
+
+        // alice is still locked, balance should be 0
+        let alice_balance = get_balance(&mut peer, &key_to_stacks_addr(&alice).into());
+        assert_eq!(alice_balance, 0);
+    }
+
+    // now, try to use PoX 2 contract. this should _fail_.
```
gregorycoppola, line 236: I'm not sure which of the asserts shows that this "failed".. is there one? If so, can a stronger connection be made with comments, e.g., does "failing" mean Alice has a balance of 0?

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -0,0 +1,423 @@
...
+                }
+            }
+        }
+    }
+
+    assert!(tested_charlie, "Charlie TX must be tested");
+    assert_eq!(alice_txs.len(), 3, "Alice should have 3 confirmed txs");
+    assert_eq!(bob_txs.len(), 2, "Bob should have 2 confirmed txs");
+
+    assert!(
```
gregorycoppola, line 388: For each assert in this section, where you are checking transactions, can you put a quick comment about *why* we expect these things to happen?

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -0,0 +1,423 @@
...
+    let alice_lockup = make_pox_lockup(
+        &alice,
+        0,
+        1024 * POX_THRESHOLD_STEPS_USTX,
+        AddressHashMode::SerializeP2PKH,
+        key_to_stacks_addr(&alice).bytes,
+        4,
+        tip.block_height,
+    );
+    let tip_index_block = peer.tenure_with_txs(&[alice_lockup], &mut coinbase_nonce);
```
gregorycoppola, line 197: looks pretty clean the way that you were able to use "tenure_with_txs".. is it possible to indicate, at least on some more instances, where the "counter" is? It's easier to get lost as teh number of iterations goes up.

gregorycoppola (approved):
Hey Aaron.. I think it's very elegant the way you could use `feat/stacks-2.1-initialization` here. I just had some requests for things to be commented a bit more, so that it's easier for other readers (like me) to follow along. Not asking for any code changes.

PR #2647, opened 2021-05-05
Private local testnet support

kantai:
These changes were required for supporting a private local testnet with the latest version of bitcoind


src/burnchains/bitcoin/indexer.rs
```diff
@@ -497,6 +497,11 @@ impl BitcoinIndexer {
                 Err(btc_error::ConnectionBroken) => {
                     do_handshake = true;
                 }
+                Err(btc_error::SerializationError(
+                    btc_serialization_err::UnrecognizedNetworkCommand(s),
+                )) => {
+                    debug!("Received unrecognized netowkr command while receiving a message: {}, ignoring", s);
```
psq, line 503: network

testnet/stacks-node/src/neon_node.rs
```diff
@@ -1383,6 +1402,101 @@ impl InitializedNeonNode {
...
+            SortitionDB::get_block_snapshot_consensus(burn_db.conn(), mine_tip_ch)
+                .expect("Failed to look up block's parent snapshot")
+                .expect("Failed to look up block's parent snapshot");
+
+        let parent_sortition_id = &parent_snapshot.sortition_id;
+        let parent_winning_vtxindex =
+            SortitionDB::get_block_winning_vtxindex(burn_db.conn(), parent_sortition_id)
+                .expect("SortitionDB failure.")
+                .ok_or_else(|| {
+                    error!(
```
jcnelson, line 1439: Is this error reachable without being indicative of database corruption?  The only way I can think it could be reached is if you're mining off of the genesis block, which has no parent.

kantai: I think that's correct, but this replicates the previous behavior of the miner (which would also just "not mine" rather than panic here).

pavitthrap (approved):
LGTM!

jcnelson (approved):
LGTM! Let's get this tested on the testnet before merging to master.

kantai:
> LGTM! Let's get this tested on the testnet before merging to master.

Yep -- will hold off on merging until 2.0.11.1.0 is released

PR #2648, opened 2021-05-06
Initial balances for local setups

lgalabru:
This PR is updating the config to make it so the command `stacks-node mocknet` start with some initialized balances that developers can use via private keys or mnemonics.
Key pairs are coming from https://github.com/blockstack/stacks-local-dev/blob/master/stacks-node-follower/Config.toml.template.


kantai (approved):
LGTM!

jcnelson (approved):
This LGTM, but what's the rationale for the change?  Was the number of initial balances too big?

lgalabru:
@jcnelson Balances are fine, rationale, is exposing mnemonic along with keypairs derived with our default derivation path, so that the genesis balances can be used by other products like the wallet or clarinet.

PR #2652, opened 2021-05-12
Update configs

CharlieC3:
## Description
Updates the example configs for `stacks-node`, removes old or irrelevant configs.

I noticed the connection option defaults we've been explicitly defining for our nodes for read-only smart-contract calls 
 (the `[connection_options]` block) were already the default [as shown here](https://github.com/blockstack/stacks-blockchain/blob/master/src/net/connection.rs#L435-L441). So, I didn't make any changes to the default config values, nor did I explicitly configure these properties in the new example configs.

Closes https://github.com/hirosystems/devops/issues/698

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

No.

Yes
- [x] Link to documentation updates: https://github.com/blockstack/docs/pull/1155

## Checklist


testnet/stacks-node/conf/mainnet-follower-conf.toml
```diff
@@ -0,0 +1,20 @@
+[node]
+# working_dir = "/dir/to/save/chainstate"
+rpc_bind = "0.0.0.0:20443"
+p2p_bind = "0.0.0.0:20444"
+bootstrap_node = "02da7a464ac770ae8337a343670778b93410f2f3fef6bea98dd1c3e9224459d36b@seed-0.mainnet.stacks.co:20444,02afeae522aab5f8c99a00ddf75fbcb4a641e052dd48836408d9cf437344b63516@seed-1.mainnet.stacks.co:20444,03652212ea76be0ed4cd83a25c06e57819993029a7b9999f7d63c36340b34a4e62@seed-2.mainnet.stacks.co:20444"
+
+[burnchain]
```
jcnelson, line 7: Isn't `bitcoin.blockstack.com` what we want here?

testnet/stacks-node/conf/testnet-miner-conf.toml
```diff
@@ -1,19 +1,21 @@
...
 miner = true
-bootstrap_node = "047435c194e9b01b3d7f7a2802d6684a3af68d05bbf4ec8f17021980d777691f1d51651f7f1d566532c804da506c117bbf79ad62eea81213ba58f8808b4d9504ad@xenon.blockstack.org:20444"
+bootstrap_node = "047435c194e9b01b3d7f7a2802d6684a3af68d05bbf4ec8f17021980d777691f1d51651f7f1d566532c804da506c117bbf79ad62eea81213ba58f8808b4d9504ad@testnet.stacks.co:20444"
+wait_time_for_microblocks = 10000
 
 [burnchain]
 chain = "bitcoin"
 mode = "xenon"
 peer_host = "127.0.0.1"
+username = "<USERNAME>"
```
jcnelson, line 15: Don't we want to set up a public bitcoin testnet node to use?

testnet/stacks-node/conf/mainnet-follower-conf.toml
CharlieC3: Good catch, updated to `bitcoind.stacks.co` to avoid confusion with using the old blockstack name.

jcnelson (approved):
LGTM! Thanks @CharlieC3!

diwakergupta (approved):

PR #2653, opened 2021-05-13
fix: emit microblocks mined by the node itself, add microblock hashes

### LINKED ISSUE #2662: Add `burn_block_*` properties to `/new_microblocks` payload
After talking with @kantai , it sounds like microblock transactions use the same `burn_block_hash` (and height and time) as their subsequent anchor block? 

These properties need added to the `/new_microblocks` payload. These are pretty important properties exposed to clients. At the moment, the API is unable to populate them for unanchored transactions, which creates significant breaking changes.

zone117x:
Microblock events were being sent to the event observer when microblock data was received by a peer, but were not emitted if the node mined the microblocks itself. This made something like the private-testnet setup incapable of emitting microblock events.

Fixes https://github.com/blockstack/stacks-blockchain/issues/2668
- Adds microblock specific information to transactions packaged into the `new_block` event. This information is useful since `new_microblock` events are not emitted when the chain is syncing. 

Fixes #2662
- Adds burn_block_{height, hash, timestamp} data to the `new_microblock` event. 
- For the `new_block` event, added prev_burn_block_{height, hash, timestamp}.


src/chainstate/stacks/db/blocks.rs
```diff
@@ -4245,6 +4252,26 @@ impl StacksChainState {
...
+                        sn.burn_header_hash,
+                        sn.block_height as u32,
+                        sn.burn_header_timestamp,
+                    ),
+                    None => {
+                        // shouldn't happen
+                        panic!(
+                            "CORRUPTION: staging block {}/{} does not correspond to a burn block",
+                            &parent_consensus_hash, &parent_block_hash
+                        );
```
kantai, line 4271: Does this panic on the first block? (i.e., if `parent_consensus_hash` is `FIRST_BURNCHAIN_CONSENSUS_HASH`)?

src/chainstate/coordinator/mod.rs
```diff
@@ -120,6 +120,9 @@ pub trait BlockEventDispatcher {
         winner_txid: Txid,
         matured_rewards: Vec<MinerReward>,
         matured_rewards_info: Option<MinerRewardInfo>,
+        prev_burn_block_hash: BurnchainHeaderHash,
+        prev_burn_block_height: u32,
+        prev_burn_block_timestamp: u64,
```
kantai, line 125: Rather than `prev_*`, I'd call these `parent_*` -- this isn't necessarily the immediately prior burn block: it's the burn block that selected this block's parent.

testnet/stacks-node/src/node.rs
```diff
@@ -761,6 +761,26 @@ impl Node {
...
+                    sn.block_height as u32,
+                    sn.burn_header_timestamp,
+                ),
+                None => {
+                    // shouldn't happen
+                    panic!(
+                        "CORRUPTION: staging block {}/{} does not correspond to a burn block",
+                        &parent_consensus_hash, &anchored_block.header.parent_block
+                    );
+                }
```
kantai, line 781: Same goes for here: I'm not convinced this should panic.

src/chainstate/stacks/db/blocks.rs
pavitthrap: I left a panic, but now I'm also explicitly checking whether or not the block is the first.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4245,6 +4253,30 @@ impl StacksChainState {
...
+                        &parent_consensus_hash,
+                    )? {
+                        Some(sn) => (
+                            sn.burn_header_hash,
+                            sn.block_height as u32,
+                            sn.burn_header_timestamp,
+                        ),
+                        None => {
+                            // shouldn't happen
+                            panic!(
```
jcnelson, line 4272: This panic code is reachable under normal conditions.  If we're doing a PoX reorg, then because this isn't carried out in a transaction, it's possible that the coordinator thread will invalidate the block snapshot just before `get_block_snapshot_consensus()` gets called, leading to the `None` case.

jcnelson: Generally speaking, it's a Bad Idea to panic on network code paths.

testnet/stacks-node/src/node.rs
```diff
@@ -761,6 +761,30 @@ impl Node {
...
+                    .unwrap()
+                {
+                    Some(sn) => (
+                        sn.burn_header_hash,
+                        sn.block_height as u32,
+                        sn.burn_header_timestamp,
+                    ),
+                    None => {
+                        // shouldn't happen
+                        panic!(
```
jcnelson, line 780: Same here -- this `panic!` is reachable under normal circumstances, like a PoX reorg.

src/chainstate/stacks/db/blocks.rs
pavitthrap: Ok, will remove the panic, and use an option type for the burn block stats instead. 

Also, I have questions about the invalidation logic you mention. The function `get_block_snapshot_consensus` queries with `SELECT * FROM snapshots WHERE consensus_hash = ?1`, and the invalidation function I saw just updates the field `pox_valid` to 0. Also, this code block is (I'm pretty sure), being called in the coordinator thread. I'm unsure in what cases that `get_block_snapshot_consensus` would return `None` here.

pavitthrap: @jcnelson

jcnelson: Ah, I stand corrected.  I had forgotten that this method ignores the `pox_valid` flag.

But just to be safe, let's remove the `panic!()` and log a warning.  I've had a couple sleepless nights now that can basically be summarized as stories in three acts:

Act 1: "There's no way this code is reachable"
Act 2: "The life, death, and resurrection of the blockchain"
Act 3: "It was totally the unreachable code"

src/chainstate/stacks/db/blocks.rs
```diff
@@ -3875,10 +3876,12 @@ impl StacksChainState {
         for microblock in microblocks.iter() {
             debug!("Process microblock {}", &microblock.block_hash());
             for tx in microblock.txs.iter() {
-                let (tx_fee, tx_receipt) =
+                let (tx_fee, mut tx_receipt) =
                     StacksChainState::process_transaction(clarity_tx, tx, false)
                         .map_err(|e| (e, microblock.block_hash()))?;
 
+                // TODO: also keep track of a real tx_index here and add to receipt
```
jcnelson, line 3883: Does this `TODO` need to be addressed before merge?

jcnelson (approved):
LGTM! Just had one small question

src/chainstate/stacks/db/blocks.rs
pavitthrap: @zone117x I remember you mentioning we need the tx indices of tx's within microblocks? Is that still true?

jcnelson: Let's just make an issue for this so we can merge it.

kantai (approved):
LGTM!

Issue #2643, opened 2021-05-03
Epoch switch: Clarity VM versioning via pragmas


jcnelson:
We'll be taking the "pragma" approach to adding new Clarity keywords and functions.  See https://github.com/blockstack/stacks-blockchain/discussions/2637

PR #2659, opened 2021-05-18
2.1: Clarity VM Versioning

kantai:
This PR implements the Clarity versioning component of #2643 -- it does not add the pragma, rather it just creates a contract with the default Clarity version of that epoch. We will need to add that pragma, but that is a separable task that we can address in a subsequent PR.

This PR also introduces a new 2.1 function for obtaining the whole account entry for a STX account (i.e., returning the lockup status). This function may need to change as we flesh out the implementation of PoX 2.1, but it exists in this PR as an example of how a new function can be added via Clarity versioning.

For testing, this PR adds a new integration test to the Sim environment which simulates cross the epoch boundary. The test constructs contracts before and after the epoch boundary, which is used to demonstrate the correct behavior of variable name shadowing in the different Clarity versions. This is the test that should be generalized as we add more features that cross epoch boundaries.


src/vm/functions/mod.rs
```diff
@@ -49,95 +51,128 @@ mod sequences;
...
+        NativeFunctions::lookup_by_name(name).and_then(|native_function| {
+            if &native_function.get_version() == version {
+                Some(native_function)
+            } else {
+                None
+            }
+        })
+    }
+
+    pub fn lookup_by_name_before_version(
```
pavitthrap, line 154: Could rename to `lookup_by_name_at_version`.

src/vm/analysis/contract_interface_builder/mod.rs
```diff
@@ -365,6 +371,7 @@ impl ContractInterface {
             maps: Vec::new(),
             fungible_tokens: Vec::new(),
             non_fungible_tokens: Vec::new(),
+            clarity_version: ClarityVersion::Clarity1,
```
pavitthrap, line 374: Why do we want to default to `ClarityVersion::Clarity1` here?

src/vm/database/structures.rs
```diff
@@ -351,6 +355,21 @@ impl STXBalance {
         }
     }
 
+    /// Returns a canonicalized STXBalance at a given burn_block_height
+    /// (i.e., if burn_block_height >= unlock_height, then return struct where
+    ///   amount_unlocked = 0, unlock_height = 0)
```
pavitthrap, line 360: `amount_locked = 0` instead?

src/vm/mod.rs
```diff
@@ -355,7 +359,7 @@ fn eval_all(
  */
 pub fn execute(program: &str) -> Result<Option<Value>> {
     let contract_id = QualifiedContractIdentifier::transient();
-    let mut contract_context = ContractContext::new(contract_id.clone());
+    let mut contract_context = ContractContext::new(contract_id.clone(), ClarityVersion::Clarity1);
```
pavitthrap, line 362: Should we use `info!` or some type of debug statement here (to alert a cli user to what clarity version is being used by default?)

src/vm/functions/mod.rs
kantai: Yep -- renamed in ab8204133

src/vm/analysis/contract_interface_builder/mod.rs
kantai: Yep, there's no need for a default here, changed in ab8204133

src/vm/database/structures.rs
kantai: Comment fixed in ab8204133

src/vm/mod.rs
kantai: Yep, added in ab8204133

src/clarity_vm/clarity.rs
```diff
@@ -743,6 +758,14 @@ impl<'a, 'b> ClarityTransactionConnection<'a, 'b> {
         identifier: &QualifiedContractIdentifier,
         contract_content: &str,
     ) -> Result<(ContractAST, ContractAnalysis), Error> {
+        let epoch_id = self
+            .with_clarity_db_readonly(|db| db.get_current_stacks_epoch())
+            .expect("Failed to load current epoch")
+            .epoch_id;
+
+        // ClarityVersionPragmaTodo: need to use contract's declared version or default
```
jcnelson, line 766: Is there an issue somewhere tracking these TODOs, if they're not going to happen in this PR?

src/core/mod.rs
```diff
@@ -159,6 +159,7 @@ pub enum StacksEpochId {
 }
 
 impl PartialOrd for StacksEpochId {
+    // Note: this comparison makes Epoch10 > Epoch21 > Epoch20. Is that the intention?
```
jcnelson, line 162: Lol nope.

src/vm/analysis/mod.rs
```diff
@@ -38,47 +39,36 @@ use self::read_only_checker::ReadOnlyChecker;
 use self::trait_checker::TraitChecker;
 use self::type_checker::TypeChecker;
 
-pub fn mem_type_check(snippet: &str) -> CheckResult<(Option<TypeSignature>, ContractAnalysis)> {
+pub fn mem_type_check(
```
jcnelson, line 42: Is this method used anywhere besides tests?  Should it be `#[cfg(test)]`?

src/vm/ast/definition_sorter/mod.rs
```diff
@@ -210,8 +212,12 @@ impl<'a> DefinitionSorter {
                                 }
                             }
                         } else if let Some(native_function) =
-                            NativeFunctions::lookup_by_name(function_name)
-                        {
+                            // NOTE: can use ClarityVersion::latest() here only as long as NO NEW FUNCTIONS are special cased
```
jcnelson, line 215: Isn't this something that will realistically happen in a hypothetical Clarity3?  Does it make sense to front-load the handling for this case now, while it's still fresh in our minds?

src/vm/contexts.rs
```diff
@@ -1107,11 +1122,22 @@ impl<'a, 'b> Environment<'a, 'b> {
             let memory_use = contract_string.len() as u64;
             self.add_memory(memory_use)?;
 
+            let version = ClarityVersion::default_for_epoch(
```
jcnelson, line 1125: This code will eventually only be if the contract doesn't supply a `(clarity-version VERSION)` pragma, right?

src/vm/database/clarity_db.rs
```diff
@@ -561,10 +584,7 @@ impl<'a> ClarityDatabase<'a> {
     pub fn get_current_burnchain_block_height(&mut self) -> u32 {
         let cur_stacks_height = self.store.get_current_block_height();
         let last_mined_bhh = if cur_stacks_height == 0 {
-            StacksBlockHeader::make_index_block_hash(
-                &FIRST_BURNCHAIN_CONSENSUS_HASH,
-                &FIRST_STACKS_BLOCK_HASH,
-            )
+            return self.burn_state_db.get_burn_start_height();
```
jcnelson, line 587: While this should be correct (I checked in prod), this feels risky.  Do we at least want to assert somewhere that `.get_burn_start_height()` returns the same value as `.get_burnchain_block_height()` on the initial block hash?

src/vm/functions/mod.rs
```diff
@@ -49,95 +51,128 @@ mod sequences;
...
+    BurnAsset("nft-burn?", ClarityVersion::Clarity1),
+    GetStxBalance("stx-get-balance", ClarityVersion::Clarity1),
+    StxTransfer("stx-transfer?", ClarityVersion::Clarity1),
+    StxBurn("stx-burn?", ClarityVersion::Clarity1),
+    StxGetAccount("stx-account", ClarityVersion::Clarity2),
 });
 
-pub fn lookup_reserved_functions(name: &str) -> Option<CallableType> {
+impl NativeFunctions {
+    pub fn lookup_by_name_exact_version(
```
jcnelson, line 141: I'm curious how this method will be used?  It doesn't appear to be called anywhere yet.  Will there be cases where the VM has a defined native function that is available in _exactly one_ version of the VM?

src/vm/costs/cost_functions.rs
```diff
@@ -130,4 +130,5 @@ define_named_enum!(ClarityCostFunction {
     NftOwner("cost_nft_owner"),
     NftBurn("cost_nft_burn"),
     PoisonMicroblock("poison_microblock"),
+    StxGetAccount("cost_stx_get_account"),
```
jcnelson, line 133: I didn't see where this cost function is defined?  Is this waiting on a hypothetical `.costs-2` contract that will add it in, as well as changes to the runtime cost-tracker to know which `.costs-X` contract to use?

src/vm/database/clarity_db.rs
kantai: In the genesis block, `.get_burnchain_block_height()` would previously _panic_, which wasn't previously a problem, because none of the contracts deployed or called in the genesis block queried the burnchain block height. However, because the versioning check _does_ query the burnchain block height during the genesis initialization, that case had to be handled.

src/core/mod.rs
kantai: That's what I figured -- the #2639 PR fixes that comparison.

src/vm/contexts.rs
kantai: Yep, that's correct

src/clarity_vm/clarity.rs
kantai: Opened https://github.com/blockstack/stacks-blockchain/issues/2719

src/vm/ast/definition_sorter/mod.rs
kantai: I'm not sure -- I think ultimately the way to handle this would be to just create different definition sorters for each Clarity version. We could create a top-level match that then routed the version to the correct definition sorter.

src/vm/functions/mod.rs
kantai: There's probably several theoretical future use cases, but none at the moment, so I just removed this function.

src/vm/costs/cost_functions.rs
kantai: Removed -- this needs to be added when addressing #2717.

kantai:
I think all the comments have been addressed and edited -- can we do another round of reviews? Several PRs and features for 2.1 and Clarity 2 depend on this PR, and it'd be good to unblock those.

jcnelson (approved):
LGTM. Thanks @kantai !

src/chainstate/stacks/boot/contract_tests.rs
```diff
@@ -207,6 +250,60 @@ fn test_sim_hash_to_height(in_bytes: &[u8; 32]) -> Option<u64> {
...
+    fn get_burn_header_hash(
+        &self,
+        height: u32,
+        sortition_id: &crate::types::chainstate::SortitionId,
+    ) -> Option<BurnchainHeaderHash> {
+        panic!("Not implemented in TestSim");
+    }
+
+    fn get_stacks_epoch(&self, height: u32) -> Option<StacksEpoch> {
+        let epoch_begin_index = match self.epoch_bounds.binary_search(&(height as u64)) {
```
pavitthrap, line 270: In the structs that use `epoch_bounds`, I think it would be useful to leave a note about how it should be constructed (for example, don't include 0 for the start of Epoch10).

pavitthrap (approved):
LGTM!

src/chainstate/stacks/boot/contract_tests.rs
kantai: Added in https://github.com/blockstack/stacks-blockchain/pull/2659/commits/1b38084371d2788e878b8c2785d7a501544a9333

PR #2660, opened 2021-05-18
fix: block limit should not be configurable

lgalabru:
This PR is removing the ability to custom cost limits from a config file, along with aligning cost limits on mocknet with cost limits on testnet / mainnet.


jcnelson (approved):
LGTM! Thanks @lgalabru

PR #2661, opened 2021-05-18
Fix/2645

### LINKED ISSUE #2645: Mining: emit a warning when there are multiple UTXOs available, and exit
It has come to my attention that the Stacks miner will flip-flop between consuming different UTXOs when creating block-commit transactions, if there are multiple UTXOs available.  This can happen if there are, for example, two similarly-valuable UTXOs, such that spending one of them will cause the other one to be preferred on the next block-commit transaction.  This is very problematic, because it causes the `min(mean,median)` sampling algorithm to penalize miners that have been continuously online.

To remedy this, the miner should check on boot-up whether or not there are multiple UTXOs, and if so, *crash* with a warning to the user to re-consolidate their UTXOs and try again.

jcnelson:
This fixes #2645 by altering the miner to prioritize spending the _most recent_ UTXO when building a transaction, instead of the _largest_ UTXO.  In the event of a tie, it uses the _smallest_ UTXO first.  This way, if you refill your miner, your block-commits will remain chained together -- your new, large UTXO will be consolidated with your outstanding small UTXO in the subsequent block-commit.

### DIFF
--- a/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
+++ b/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
@@ -1,15 +1,15 @@
+use async_h1::client;
 use async_std::io::ReadExt;
+use async_std::net::TcpStream;
+use base64::encode;
+use http_types::{Method, Request, Url};
+use std::cmp::Ordering as CmpOrdering;
 use std::io::Cursor;
 use std::sync::{
     atomic::{AtomicBool, Ordering},
     Arc,
 };
 use std::time::Instant;
-
-use async_h1::client;
-use async_std::net::TcpStream;
-use base64::encode;
-use http_types::{Method, Request, Url};
 
 use serde::Serialize;
 use serde_json::value::RawValue;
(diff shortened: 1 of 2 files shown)


testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -1173,9 +1174,24 @@ impl BitcoinRegtestController {
...
+                // for block-commits, the smaller value is likely the UTXO-chained value, so
+                // continue to prioritize it as the first spend in order to avoid breaking the
+                // miner commit chain.
+                if u1.amount < u2.amount {
+                    CmpOrdering::Greater
+                } else if u1.amount > u2.amount {
+                    CmpOrdering::Less
+                } else {
+                    CmpOrdering::Equal
+                }
```
lgalabru, line 1192: Is this equivalent to
```rust
u1.amount.cmp(&u2.amount)
```
?

lgalabru (approved):

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
jcnelson: Yup; fixed

kantai (approved):
LGTM! Just had a nit about deriving PartialEq

PR #2664, opened 2021-05-19
Feat: added mining-related prometheus stats

### LINKED ISSUE #2651: Emit miner-relevant prometheus metrics
It would be helpful for miners to be able to get more useful miner-relevant prom metrics.

For example:

1. Their computed relative miner score (I think this would be a gauge metric?)
2. Their computed commitment (again a gauge metric)
3. Their current median commitment

There may be some other useful miner-relevant prom metrics, but this would be a start

pavitthrap:
## Description

Fixes #2651 

Added the following metrics in prometheus: 
- Miner's computed relative miner score as a percentage (Gauge)
- Miner's computed commitment, the min of their previous commitment and their median commitment
- Miner's current median commitment

The latter two metrics are u128 values. Prometheus's IntGauge only captures 64 bit values, so I used two IntGauges for both of the metrics (one for the high 64 bits, and another for the low 64 bits). 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

## Testing information
Ran `BITCOIND_TEST=1 cargo test --release --features monitoring_prom -- --test-threads 1 --ignored neon_integrations::bitcoind_integration_test` in the `testnet/stacks-node` directory.  Pinged the prometheus endpoint and got the value of all tracked metrics, which I then checked against expected value.
> Note: When running an integration test with a specific feature enabled, running the command above in the root directory of the repo doesn't work. If anyone understands why this is, let me know! 

Ran `cargo test`.

### DIFF
--- a/src/chainstate/burn/distribution.rs
+++ b/src/chainstate/burn/distribution.rs
@@ -30,6 +30,7 @@ use chainstate::burn::operations::{
 };
 use chainstate::stacks::StacksPublicKey;
 use core::MINING_COMMITMENT_WINDOW;
+use monitoring;
 use util::hash::Hash160;
 use util::log;
 use util::uint::BitArray;
@@ -163,6 +164,7 @@ impl BurnSamplePoint {
         assert!(window_size > 0);
         BurnSamplePoint::sanity_check_window(&block_commits, &missed_commits);
         assert_eq!(burn_blocks.len(), block_commits.len());
+        let global_burnchain_signer = monitoring::get_burnchain_signer();
 
         // first, let's link all of the current block commits to the priors
         let mut commits_with_priors: Vec<_> =
@@ -285,6 +287,13 @@ impl BurnSamplePoint {
                        "median_burn" => %median_burn,
                        "all_burns" => %format!("{:?}", all_burns));
 
+                // prometheus: log miner commitment
+                if let Some(signer) = global_burnchain_signer {
+                    if candidate.apparent_sender == *signer {
+                        monitoring::update_computed_miner_commitment(burns);
+                        monitoring::update_miner_current_median_commitment(median_burn);
+                    }
+                }
                 BurnSamplePoint {
                     burns,
                     range_start: Uint256::zero(), // To be filled in
@@ -297,6 +306,21 @@ impl BurnSamplePoint {
 
         // calculate burn ranges
         BurnSamplePoint::make_sortition_ranges(&mut burn_sample);
+
+        // prometheus: calculate miner relative score
+        if let Some(signer) = global_burnchain_signer {
+            let mut range_total = Uint256::zero();
+            let mut signer_seen = false;
+            for burn in burn_sample.iter() {
+                if burn.candidate.apparent_sender == *signer {
+                    signer_seen = true;
+                    range_total = range_total + burn.range_end - burn.range_start;
+                }
+            }
+            if signer_seen {
+                monitoring::update_computed_relative_miner_score(range_total);
+            }
+        }
         burn_sample
     }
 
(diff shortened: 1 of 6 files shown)


src/chainstate/burn/distribution.rs
```diff
@@ -297,6 +306,21 @@ impl BurnSamplePoint {
...
         BurnSamplePoint::make_sortition_ranges(&mut burn_sample);
+
+        // prometheus: calculate miner relative score
+        if let Some(signer) = global_burnchain_signer {
+            let mut range_total = Uint256::zero();
+            let mut signer_seen = false;
+            for burn in burn_sample.iter() {
+                if burn.candidate.apparent_sender == *signer {
+                    signer_seen = true;
+                    range_total = range_total + burn.range_end - burn.range_start;
```
jcnelson, line 317: This could overflow.  I think you want this: `range_total + (burn.range_end - burn.range_start)`.

src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -598,14 +598,14 @@ impl LeaderBlockCommitOp {
                             error!("Failed to check whether parent (height={}) is descendent of anchor block={}: {}",
                                    parent_block_height, &reward_set_info.anchor_block, e);
                             op_error::BlockCommitAnchorCheck})?;
-                    if descended_from_anchor != expect_pox_descendant {
-                        if descended_from_anchor {
-                            warn!("Invalid block commit: descended from PoX anchor, but used burn outputs");
-                        } else {
-                            warn!(
+                    if descended_from_anchor && !expect_pox_descendant {
```
jcnelson, line 601: THIS IS POTENTIALLY CONSENSUS BREAKING.  Please undo this!

src/monitoring/mod.rs
```diff
@@ -284,3 +296,126 @@ pub fn increment_contract_calls_processed() {
     #[cfg(feature = "monitoring_prom")]
     prometheus::CONTRACT_CALLS_PROCESSED_COUNT.inc();
 }
+
+/// Given a value (type uint256), return value/uint256::max() as an f64 value.
+/// The precision of the percentage is determined by the input `precision_points`.
+fn convert_uint256_to_f64_percentage(value: Uint256, precision_points: u32) -> f64 {
+    let base = 10;
+    let multiplier = Uint512::from_u128(100 * u128::pow(base, precision_points));
```
jcnelson, line 304: I think this can overflow.  What if `precision_points` is `u32::MAX`?

src/chainstate/burn/operations/leader_block_commit.rs
pavitthrap: Removed.

src/monitoring/mod.rs
```diff
@@ -284,3 +299,130 @@ pub fn increment_contract_calls_processed() {
...
+        let low_bits = value as u64;
+        prometheus::MINER_CURRENT_MEDIAN_COMMITMENT_HIGH.set(high_bits as i64);
+        prometheus::MINER_CURRENT_MEDIAN_COMMITMENT_LOW.set(low_bits as i64);
+    }
+}
+
+/// Function sets the global variable `GLOBAL_BURNCHAIN_SIGNER`.
+/// Fails if there are multiple attempts to set this variable.
+pub fn set_burnchain_signer(signer: BurnchainSigner) -> Result<(), SetGlobalBurnchainSignerError> {
+    if BURNCHAIN_SIGNER_INIT.compare_and_swap(
```
kantai, line 393: Can this use the mutex'ed Option to check for initialization, i.e.,

```rust
let mut signer_mutex = GLOBAL_BURNCHAIN_SIGNER.lock().unwrap();
if signer_mutex.is_some() {
   Err(SetGlobalBurnchainSignerError)
} else {
   *signer_mutex = Some(signer);
   Ok(())
}
```

kantai (approved):
LGTM!

jcnelson (approved):
LGTM!

PR #2666, opened 2021-05-20
fix: use .sqlite instead of .db for tx_tracking

lgalabru:
This DB was introduced in the previous version. I'm not sure if a correction is possible or not, will withdraw the PR if we think that it's problematic.


jcnelson (approved):

kantai:
Changing the name of the database shouldn't be much of an issue -- it should be added to the changelog though.

kantai (approved):

PR #2670, opened 2021-05-25
Fix/2667

### LINKED ISSUE #2667: Add another log field to the http handler
Current http log events are vary basic. 
I propose adding an additional field to identify the originating request, ex: `real-ip	123.123.123.123`

In the exposed log, it would look similar to this:
```json
{
    "msg":"Handle GET /v2/info",
    "level":"INFO",
    "ts":"2021-05-19T23:36:55.968303260+00:00",
    "thread":"p2p",
    "line":1575,
    "file":"src/net/http.rs",
    "real-ip": "123.123.123.123"
}
```

jcnelson:
This fixes #2667, and also makes it so the unconfirmed chain tip in `/v2/info` is set to all `0`'s if there are no unconfirmed microblocks.  This is the API contract we've promised, but it had regressed at some point; this PR fixes it by having the `/v2/info` handler verify that the unconfirmed state in the p2p thread's `StacksChainState` reflects the transactions mined in the relayer thread (whereas before, the p2p thread had only been checking that its own unconfirmed state was in a readable state -- something that is always true).  This PR additionally makes a couple of small fixes to the `net-tests` that was dependent on this API contract.

### DIFF
--- a/src/net/http.rs
+++ b/src/net/http.rs
@@ -1568,10 +1568,7 @@ impl HttpRequestType {
                     } else {
                         "".to_string()
                     };
-                    info!(
-                        "Handle [{}] {} {}{}",
-                        verb, &protocol.peer_addr, decoded_path, query
-                    );
+                    info!("Handle HTTPRequest"; "verb" => %verb, "peer_addr" => %protocol.peer_addr, "path" => %decoded_path, "query" => %query);
                     return Ok(request);
                 }
                 None => {
--- a/src/net/http.rs
+++ b/src/net/http.rs
@@ -1568,7 +1568,10 @@ impl HttpRequestType {
                     } else {
                         "".to_string()
                     };
-                    info!("Handle {} {}{}", verb, decoded_path, query);
+                    info!(
+                        "Handle [{}] {} {}{}",
+                        verb, &protocol.peer_addr, decoded_path, query
+                    );
                     return Ok(request);
                 }
                 None => {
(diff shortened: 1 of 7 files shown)


ynrAYF3yPuymfkx7WRJRn6:
Мне как получить доступ к своим биткоинам которые мне перевели???

ср, 26 мая 2021 г., 6:47 Jude Nelson ***@***.***>:

src/net/http.rs
```diff
@@ -1568,7 +1568,10 @@ impl HttpRequestType {
                     } else {
                         "".to_string()
                     };
-                    info!("Handle {} {}{}", verb, decoded_path, query);
+                    info!(
+                        "Handle [{}] {} {}{}",
+                        verb, &protocol.peer_addr, decoded_path, query
+                    );
```
kantai, line 1574: Can this use the k-v logging syntax? i.e.,

```rust
info!("Handle HTTPRequest"; "verb" => %verb, "peer_addr" => %protocol.peer_addr, "path" => %decoded_path, "query" => %query);
```

kantai (approved):
LGTM! Just had one comment that this log line is a good candidate for the k-v logging syntax.

lgalabru (approved):
LGTM, thank you @jcnelson!

lgalabru:
Can you also please update the CHANGELOG to include this fix? Thanks!

src/net/http.rs
jcnelson: Done

PR #2681, opened 2021-06-01
feat(clarity): introduce slice function

lgalabru:
Address https://github.com/blockstack/stacks-blockchain/issues/2627

### DIFF
--- a/src/vm/analysis/arithmetic_checker/tests.rs
+++ b/src/vm/analysis/arithmetic_checker/tests.rs
@@ -39,7 +39,7 @@ fn arithmetic_check(contract: &str) -> Result<(), Error> {
         contract_identifier,
         expressions,
         LimitedCostTracker::new_free(),
-        ClarityVersion::Clarity1,
+        ClarityVersion::Clarity2,
     );
 
     ArithmeticOnlyChecker::run(&analysis)
@@ -206,6 +206,9 @@ fn test_functions() {
          FunctionNotPermitted(NativeFunctions::Sha512)),
         ("(define-private (foo) (sha512/256 0))",
          FunctionNotPermitted(NativeFunctions::Sha512Trunc256)),
+        ("(define-private (foo (a (list 3 uint)))
+         (slice a u2 u1))",
+         FunctionNotPermitted(NativeFunctions::Slice)),
 
     ];
 
--- a/src/vm/analysis/type_checker/natives/sequences.rs
+++ b/src/vm/analysis/type_checker/natives/sequences.rs
@@ -391,3 +391,65 @@ pub fn check_special_index_of(
 
     TypeSignature::new_option(TypeSignature::UIntType).map_err(|e| e.into())
 }
+
+pub fn check_special_slice(
+    checker: &mut TypeChecker,
+    args: &[SymbolicExpression],
+    context: &TypingContext,
+) -> TypeResult {
+    check_argument_count(3, args)?;
+
+    // Position
+    let position = match args[1].expr {
+        SymbolicExpressionType::LiteralValue(Value::UInt(position)) => position,
+        _ => {
+            let position_type = checker.type_check(&args[1], context)?;
+            return Err(CheckErrors::TypeError(TypeSignature::UIntType, position_type).into());
+        }
+    };
+    checker
+        .type_map
+        .set_type(&args[1], TypeSignature::UIntType)?;
+
+    let position = u32::try_from(position).map_err(|_e| CheckErrors::MaxLengthOverflow)?;
+
+    // Length
+    let length = match args[2].expr {
+        SymbolicExpressionType::LiteralValue(Value::UInt(expected_len)) => expected_len,
+        _ => {
+            let expected_len_type = checker.type_check(&args[2], context)?;
+            return Err(CheckErrors::TypeError(TypeSignature::UIntType, expected_len_type).into());
+        }
+    };
+    checker
+        .type_map
+        .set_type(&args[2], TypeSignature::UIntType)?;
+
+    let length = u32::try_from(length).map_err(|_e| CheckErrors::MaxLengthOverflow)?;
+
+    // Sequence
+    let seq_type = checker.type_check(&args[0], context)?;
+
+    let (origin_len, resized_seq) = match &seq_type {
+        TypeSignature::SequenceType(ListType(list)) => {
+            (list.get_max_len(), TypeSignature::list_of(list.get_list_item_type().clone(), length)?)
+        },
+        TypeSignature::SequenceType(BufferType(len)) => {
+            (len.0, TypeSignature::SequenceType(BufferType(BufferLength(length))))
+        },
+        TypeSignature::SequenceType(StringType(ASCII(len))) => {
+            (len.0, TypeSignature::SequenceType(StringType(ASCII(BufferLength(length)))))
+        },
+        TypeSignature::SequenceType(StringType(UTF8(len))) => {
+            (len.0, TypeSignature::SequenceType(StringType(UTF8(StringUTF8Length(length)))))
+        },
+        _ => return Err(CheckErrors::ExpectedSequence(seq_type.clone()).into()),
+    };
+
+    if (position + length) > origin_len {
+        return Err(CheckErrors::ConstructedListTooLarge.into())
+    } 
+
+    Ok(resized_seq)
+}
+
--- a/src/vm/analysis/type_checker/tests/mod.rs
+++ b/src/vm/analysis/type_checker/tests/mod.rs
@@ -974,12 +974,12 @@ fn test_buff_fold() {
     let good = [
         "(define-private (get-len (x (buff 1)) (acc uint)) (+ acc u1))
         (fold get-len 0x000102030405 u0)",
-        "(define-private (slice (x (buff 1)) (acc (tuple (limit uint) (cursor uint) (data (buff 10)))))
+        "(define-private (slice-v1 (x (buff 1)) (acc (tuple (limit uint) (cursor uint) (data (buff 10)))))
             (if (< (get cursor acc) (get limit acc))
                 (let ((data (default-to (get data acc) (as-max-len? (concat (get data acc) x) u10))))
                     (tuple (limit (get limit acc)) (cursor (+ u1 (get cursor acc))) (data data)))
                 acc))
-        (fold slice 0x00010203040506070809 (tuple (limit u5) (cursor u0) (data 0x)))"];
+        (fold slice-v1 0x00010203040506070809 (tuple (limit u5) (cursor u0) (data 0x)))"];
     let expected = [
         "uint",
         "(tuple (cursor uint) (data (buff 10)) (limit uint))",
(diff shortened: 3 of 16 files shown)


src/vm/analysis/type_checker/natives/sequences.rs
```diff
@@ -391,3 +391,39 @@ pub fn check_special_index_of(
...
+) -> TypeResult {
+    check_argument_count(3, args)?;
+
+    // Check sequence
+    let seq_type = checker.type_check(&args[0], context)?;
+    let seq = match &seq_type {
+        TypeSignature::SequenceType(seq) => TypeSignature::SequenceType(seq.clone()),
+        _ => return Err(CheckErrors::ExpectedSequence(seq_type.clone()).into()),
+    };
+    // TODO: runtime_cost
```
pavitthrap, line 408: Is this ready for review? I see a few todos in this function

src/vm/analysis/type_checker/natives/sequences.rs
```diff
@@ -391,3 +391,39 @@ pub fn check_special_index_of(
...
+    checker: &mut TypeChecker,
+    args: &[SymbolicExpression],
+    context: &TypingContext,
+) -> TypeResult {
+    check_argument_count(3, args)?;
+
+    // Check sequence
+    let seq_type = checker.type_check(&args[0], context)?;
+    let seq = match &seq_type {
+        TypeSignature::SequenceType(seq) => TypeSignature::SequenceType(seq.clone()),
```
jcnelson, line 405: Do we need to clone this?  Can we just use `seq_type` as-is?

src/vm/functions/sequences.rs
```diff
@@ -311,3 +311,30 @@ pub fn native_element_at(sequence: Value, index: Value) -> Result<Value> {
...
+
+    // TODO: runtime_cost
+
+    let seq = eval(&args[0], env, context)?;
+    let position = eval(&args[1], env, context)?;
+    let length = eval(&args[2], env, context)?;
+
+    let sliced_seq = match (seq, position, length) {
+        (Value::Sequence(seq), Value::UInt(position), Value::UInt(length)) => {
+            let (position, length) = match (usize::try_from(position), usize::try_from(length)) {
```
jcnelson, line 330: `usize` has a maximum value that is dependent on the host architecture -- it's not guaranteed to have a given number of bits.  Can you use something else, like `u32` or `u128` or something?

jcnelson:
Do we have some clarity (no pun intended) on when it will be possible to add cost functions for Clarity2 functions and analysis?

src/vm/docs/mod.rs
```diff
@@ -581,6 +581,20 @@ function returns `none`.",
 ",
 };
 
+const SLICE_API: SpecialAPI = SpecialAPI {
+    input_type: "buff|list A|string, uint, uint",
```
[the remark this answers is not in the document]
pavitthrap: Looking at the tests for slice, it appears that the return type is not optional - could you either update the docs or the behavior of `special_slice` to return an optional?

src/vm/analysis/type_checker/natives/sequences.rs
```diff
@@ -391,3 +391,39 @@ pub fn check_special_index_of(
...
+        _ => Err(CheckErrors::TypeError(TypeSignature::UIntType, position)),
+    }?;
+    // TODO: runtime_cost
+
+    // Check length argument
+    let length = checker.type_check(&args[2], context)?;
+    match length {
+        TypeSignature::UIntType => Ok(()),
+        _ => Err(CheckErrors::TypeError(TypeSignature::UIntType, length)),
+    }?;
```
kantai, line 423: I think both of these can use `checker.type_check_expects` to simplify the check:

```suggestion
    let position = checker.type_check_expects(&args[1], context, &TypeSignature::UIntType)?;
    // TODO: runtime_cost

    // Check length argument
    let length = checker.type_check_expects(&args[2], context, &TypeSignature::UIntType)?;
```

src/vm/analysis/arithmetic_checker/tests.rs
```diff
@@ -39,7 +39,7 @@ fn arithmetic_check(contract: &str) -> Result<(), Error> {
         contract_identifier,
         expressions,
         LimitedCostTracker::new_free(),
-        ClarityVersion::Clarity1,
+        ClarityVersion::Clarity2,
```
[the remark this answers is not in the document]
jcnelson: Yeah, we'll want to preserve tests for both Clarity v1 and v2.

src/vm/analysis/type_checker/natives/sequences.rs
pavitthrap: The error arm uses `seq_type`, which is why the match is done with `&seq_type`. So, that makes `seq` a `&SequenceSubtype`, which leads to the clone - is there a way around this?

src/vm/analysis/type_checker/tests/mod.rs
```diff
@@ -2302,12 +2435,12 @@ fn test_string_ascii_fold() {
     let good = [
         "(define-private (get-len (x (string-ascii 1)) (acc uint)) (+ acc u1))
         (fold get-len \"blockstack\" u0)",
-        "(define-private (slice (x (string-ascii 1)) (acc (tuple (limit uint) (cursor uint) (data (string-ascii 10)))))
+        "(define-private (slice-v1 (x (string-ascii 1)) (acc (tuple (limit uint) (cursor uint) (data (string-ascii 10)))))
```
gregorycoppola, line 2438: are you renaming these to avoid a conflict?

maybe rather than renaming these, you can run these old contracts in Clarity1?

src/vm/docs/mod.rs
```diff
@@ -815,6 +815,21 @@ supplied), this function returns `none`.
...
 };
 
+const SLICE_API: SpecialAPI = SpecialAPI {
+    input_type: "sequence_A, uint, uint",
+    output_type: "(optional sequence_A)",
+    signature: "(slice sequence position length)",
+    description:
+        "The `slice` function attempts to return a sub-sequence of size `length` starting from
+index `position` in the provided sequence.
+If `length` is 0 or `position + length` is greater than `(len sequence)`, this function returns `none`.",
```
gregorycoppola, line 825: maybe there is some variation in languages, but it's more usual to return a partial list if `length` will overflow the original sequence.

also, if `position > length` I would return an empty list instead of "none"... where did you get this API from?

Also, why are we using "length" rather than "left_index"/"right_index"?

This is what happens in python, for example:

```
>>> l=[1,2,3,4,5]
>>> l[10:]
[]
>>> l[3:10]
[4, 5]
```

src/vm/types/mod.rs
```diff
@@ -397,6 +397,49 @@ impl SequenceData {
...
+                };
+                Value::string_ascii_from_bytes(data)
+            }
+            SequenceData::String(CharType::UTF8(data)) => {
+                let data = if empty_seq {
+                    vec![]
+                } else {
+                    data.data[position..(length + position)].to_vec()
+                };
+                Ok(Value::Sequence(SequenceData::String(CharType::UTF8(
```
gregorycoppola, line 435: does this compile? How come this one is wrapped in Ok?

src/vm/types/signatures.rs
```diff
@@ -71,7 +71,7 @@ pub struct TupleTypeSignature {
 pub struct BufferLength(pub u32);
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct StringUTF8Length(u32);
+pub struct StringUTF8Length(pub u32);
```
gregorycoppola, line 74: This goes in the opposite direction of  https://github.com/blockstack/stacks-blockchain/pull/2757

I guess he merged this into `develop`, so the example he set is not present on the line 71 of this file.

Basically, don't make this public, and use a constructor that will check the length is not outside of the bounds..

https://github.com/blockstack/stacks-blockchain/blob/986a210ba63db49d838d73ce916b86b7bc6b9958/src/vm/types/signatures.rs#L266

src/vm/analysis/arithmetic_checker/tests.rs
```diff
@@ -284,13 +287,90 @@ fn test_functions_clarity1() {
...
-            r#"(stx-transfer-memo? u100 'SPAXYA5XS51713FDTQ8H94EJ4V579CXMTRNBZKSF 'SPAXYA5XS51713FDTQ8H94EJ4V579CXMTRNBZKSF 0x010203)"#,
-            Err(FunctionNotPermitted(NativeFunctions::StxTransferMemo)),
-        ),
+        (r#"(stx-transfer-memo? u100 'SPAXYA5XS51713FDTQ8H94EJ4V579CXMTRNBZKSF 'SPAXYA5XS51713FDTQ8H94EJ4V579CXMTRNBZKSF 0x010203)"#,
+        Err(FunctionNotPermitted(NativeFunctions::StxTransferMemo))),
+        ("(define-private (foo (a (list 3 uint)))
+              (slice a u2 u1))",
+         Err(FunctionNotPermitted(NativeFunctions::Slice))),
+
+        // Clarity1 functions.
```
gregorycoppola, line 299: good idea to test this..

gregorycoppola (commented):
thanks for the change.. my main question is about the interface to this function.

are we trying to mimic the behavior of a particular language? if so which one and why?

src/vm/analysis/type_checker/tests/mod.rs
pavitthrap: The point of this test was not to test across clarity versions - it is a coincidence that we had an old test that defined a function with the name `slice`, so I think renaming it is the most clear thing to do.

src/vm/docs/mod.rs
pavitthrap: Also I'm not too attached to using position/length over left_index/right_index, we can see what others think

reedrosenbluth: I'm also a fan of returning a result/option when the slice indices are out of range. I think this is more in the spirit of the Clarity lang, and is consistent with how functions like `element-at` behave

gregorycoppola: I guess in Rust it just panics if you try to use an inappropriate range:

```
fn main() {
    // Statements here are executed when the compiled binary is called

    // Print text to the console
    println!("Hello World!");

		let a = vec!(1,2,3,4);
    println!("a {:?}", a);

		let b = &a[1..3];
    println!("b {:?}", b);

		// Panics.
		let c = &a[1..10];
    println!("c {:?}", c);
}
```

c++ seems to use undefined behavior as well:
https://en.cppreference.com/w/cpp/container/span/span

reedrosenbluth: That being said, I think that if `length` is `0`, an empty list should be returned, rather than `none`. I think it's still best to return `none` if `position + length > (len sequence)` though.

reedrosenbluth: I'm also curious about why you decided to use `position/length` instead of indices.

I think that returning an empty list is better for when the slice `length` is `0`, since the parameters are not technically invalid in the way that they are when you provide a length that puts you out of bounds.

reedrosenbluth: Don't have an opinion about `position/length` vs indices

lgalabru: +1 on the current approach, I think that taking a Rust-like approach is more in line with Clarity's design, than Python.

src/vm/tests/sequences.rs
```diff
@@ -448,6 +448,96 @@ fn test_simple_map_append() {
...
+        assert_eq!(expected.clone(), execute_v2(test).unwrap().unwrap());
+    }
+}
+
+#[test]
+fn test_slice_utf8() {
+    let tests = [
+        "(slice u\"hello \\u{1F98A}\" u0 u5)",
+        "(slice u\"hello \\u{1F98A}\" u6 u1)",
+        "(slice u\"hello \\u{1F98A}\" u6 u0)",
```
gregorycoppola, line 523: Are you still returning "None" here? I agree with Reed's comment that it should be empty list.

src/vm/docs/mod.rs
gregorycoppola: 1) If, as Ludo says, we are being Rust style, that would mean:
a) Fail on invalid indices. Why not just return an error for bad indices?
b) Use start/end, e.g. `let b = &a[1..3]` in the example I posted has value `b [2, 3]`. It seems like we didn't really make a clear decision on this.

2) As Reed says, I think list of length 0 should be an empty list, not a None.

Also please include the 0 length sub-sequence as an example here.

gregorycoppola (commented):
more on the api

src/vm/docs/mod.rs
pavitthrap: @kantai @jcnelson any opinions here?

kantai: I'm a pretty firm believer that we should stick to a rust-like error style here -- that is, providing an out of bounds index should result in a `none` result.

However, I do agree with @reedrosenbluth for slice length 0 -- we should just return an empty list rather than `none`.

pavitthrap: @kantai Thanks for your input! Any opinions on index/length vs. left/right indices?

kantai: Yep -- I prefer left-right indexes, rather than a length arg.

Issue #2682, opened 2021-06-01
Epoch switch: new cost function contract


jcnelson:
We need a cost function contract for all of the new keywords that will become available in Stacks 2.1.

jcnelson:
Closing in favor of #3115

PR #2691, opened 2021-06-03
[Clarity2] Convert buffer to uint and int

gregorycoppola:
## Description
Fixes 2616

This PR adds Clarity functions to convert a 16-byte buffer to 1) int and 2) uint, using encodings 1) big-endian and 2) little-endian.

The functions are guarded to require ClarityVersion::Clarity2.

## Testing
I tested this by creating and running the new tests in src/vm/tests/conversions.rs

## Here are some issues to consider
- [ ] To make this work, I added the function 'BuffData::as_slice'.. is that a good idea?
- [ ] I added new descriptions to the vm::docs, but didn't add any examples because when I tried I got the error NonFunctionApplication. Should we be able to add new functions and new docs at the same time?
- [ ] Should we allow < 16 bytes?


src/chainstate/stacks/boot/costs.clar
```diff
@@ -160,6 +160,9 @@
 (define-read-only (cost_filter (n uint))
     (runtime u1000))
 
+(define-read-only (cost_buff_to_int (n uint))
+    (runtime u1000))
+
```
kantai, line 165: Unfortunately we can't make cost-contract changes this way -- with this change, the "costs" contract published at genesis would change, which would change the consensus value for the stacks blockchain starting at height = 0. This is basically issue #2682. For now, I'd recommend just reusing an existing cost function for the new methods, and leaving a todo comment.

src/vm/analysis/type_checker/natives/conversions.rs
```diff
@@ -0,0 +1,59 @@
...
+pub fn check_special_buff_to_int(
+    checker: &mut TypeChecker,
+    args: &[SymbolicExpression],
+    context: &TypingContext,
+) -> TypeResult {
+    check_argument_count(1, args)?;
+    runtime_cost(ClarityCostFunction::BuffToInt, checker, 0)?;
+
+    let collection_type = checker.type_check(&args[0], context)?;
+    match collection_type {
```
kantai, line 36: For these functions, the type checker should do an admission check, rather than an exact type check.

e.g.,

`(buff-to-uint 0x00)` should work, because we'll do left-padding, but the type of that input is `(buff 1)`, which would fail the exact type check.

Instead, you can do something like:

`checker.type_check_expects(&args[0], context, &TypeSignature::SequenceType(SequenceSubType::BufferType(16)))`

Rather than the current match check.

The same applies to `buff-to-uint`.

src/vm/docs/mod.rs
```diff
@@ -489,6 +489,38 @@ input list, and returns the same list with any elements removed for which the `f
     example: "(filter not (list true false true false)) ;; Returns (false false)"
 };
 
+const BUFF_TO_INT_LE_API: SpecialAPI = SpecialAPI {
+    input_type: "(buff 16)",
+    output_type: "int",
+    signature: "(buff-to-int-le (buff 16))",
+    description: "Converts a 16-byte buffer to a signed integer use a little-endian encoding.",
+    example: "",
```
kantai, line 497: It'd be good to include some examples for these functions -- they'd double as tests as well.

src/vm/functions/conversions.rs
```diff
@@ -0,0 +1,109 @@
...
+    args: &[SymbolicExpression],
+    env: &mut Environment,
+    context: &LocalContext,
+) -> Result<Value> {
+    fn convert_to_int_le(buffer: [u8; 16]) -> Value {
+        let value = i128::from_le_bytes(buffer);
+        return Value::Int(value);
+    }
+    return buff_to_int_generic(args, env, context, convert_to_int_le);
+}
```
kantai, line 73: Do these methods need to be "special"? I think they could be implemented as `CallableType::NativeFunction`, because they don't need to do any special type checks (normal type checks work fine) and they don't need any special access to the `env` or `context`.

src/chainstate/stacks/boot/costs.clar
gregorycoppola: I reverted this now. The cost should now be specified in src/vm/functions/mod.rs, ClarityCostFunction::Mul.

src/vm/analysis/type_checker/natives/conversions.rs
gregorycoppola: I reverted this file. Checking should now be done in src/vm/analysis/type_checker/natives/mod.rs

Is there a way to test the type checking?

src/vm/functions/conversions.rs
gregorycoppola: These should be native functions now.. see src/vm/analysis/type_checker/natives/mod.rs. I also changed the function signatures to not use the 'env' and 'context'.

src/vm/docs/mod.rs
gregorycoppola: When I try to add the examples, they fail with the error

`err: NonFunctionApplication, expressions: Some([SymbolicExpression { expr: Atom(ClarityName("buff-to-int-le"))...`

Is this a clarity version issue?

src/vm/analysis/type_checker/natives/mod.rs
```diff
@@ -568,6 +570,18 @@ impl TypedNativeFunction {
                 )],
                 returns: TypeSignature::IntType,
             }))),
+            BuffToIntLe | BuffToIntBe => Simple(SimpleNativeFunction(FunctionType::UnionArgs(
+                vec![TypeSignature::SequenceType(SequenceSubtype::BufferType(
+                    BufferLength(16),
```
gregorycoppola, line 575: Is this right? Is there a good way to test it?

src/vm/tests/costs.rs
```diff
@@ -79,6 +79,10 @@ pub fn get_simple_test(function: &NativeFunctions) -> &'static str {
         SetVar => "(var-set var-foo 1)",
         Map => "(map not list-foo)",
         Filter => "(filter not list-foo)",
+        BuffToIntLe => "(buff-to-int-le 0x00000000000000000000000000000001)",
+        BuffToUIntLe => "(buff-to-uint-le 0x00000000000000000000000000000001)",
+        BuffToIntBe => "(buff-to-int-be 0x00000000000000000000000000000001)",
+        BuffToUIntBe => "(buff-to-uint-be 0x00000000000000000000000000000001)",
```
gregorycoppola, line 85: Should any more thought be given to this test?

gregorycoppola:
I updated the code with 1) answers to the comments, and 2) logic to support < 16 byte input.

The only thing that wasn't addressed is the question of the docs, because it doesn't seem to work when I put in examples for these new functions. Is it a clarity version isssue? see https://github.com/blockstack/stacks-blockchain/pull/2691#discussion_r647003318

src/vm/analysis/errors.rs
```diff
@@ -131,6 +131,7 @@ pub enum CheckErrors {
     // expect a function, or applying a function to a list
     NonFunctionApplication,
     ExpectedListApplication,
+    ExpectedBuffer16(TypeSignature),
```
gregorycoppola, line 134: Probably I should use an existing error like "TypeError(TypeSignature, TypeSignature)" instead of this.

src/vm/analysis/type_checker/natives/conversions.rs
kantai: Yep -- look at the unit tests in `vm/analysis/type_checker/tests/mod.rs` -- you should be able to add some tests for different input types pretty easily, and test them against either expected errors or expected result types.

src/vm/docs/mod.rs
kantai: There shouldn't be a versioning issue -- what's the example that you're trying to add?

src/vm/tests/costs.rs
kantai: This is fine for now -- we'll want more precise testing once the benchmarking work progresses though.

src/vm/analysis/errors.rs
kantai: Yes -- if you use `checker.type_check_expects(..)` to perform the type checks, then it will create the `TypeError(..)` error for you, so you should be able to remove the `ExpectedBuffer16` variant.

gregorycoppola: The error in 'src/vm/functions/conversions.rs' is now:

                return Err(CheckErrors::TypeError(
                    SequenceType(BufferType(BufferLength(16))),
                    TypeSignature::type_of(&value),
                )

src/vm/analysis/type_checker/natives/conversions.rs
gregorycoppola: I added some tests in `vm/analysis/type_checker/tests/mod.rs`

src/vm/docs/mod.rs
gregorycoppola: It was a version issue.. there was a Clarity1 assumption in the test.. I changed it and added a bunch of examples.

gregorycoppola:
Thanks for the notes. I added some fixes.

src/vm/docs/mod.rs
```diff
@@ -1886,7 +1938,7 @@ mod test {
 
         let conn = store.as_clarity_db(&DOC_HEADER_DB, &DOC_POX_STATE_DB);
         let mut contract_context =
-            ContractContext::new(contract_id.clone(), crate::vm::ClarityVersion::Clarity1);
+            ContractContext::new(contract_id.clone(), crate::vm::ClarityVersion::Clarity2);
```
gregorycoppola, line 1941: note.. Clarity version referenced here.

gregorycoppola:
@kantai Hopefully most major issues are fixed now.

src/vm/analysis/arithmetic_checker/mod.rs
```diff
@@ -189,7 +189,11 @@ impl<'a> ArithmeticOnlyChecker<'a> {
             | Modulo | Power | Sqrti | Log2 | BitwiseXOR | And | Or | Not | Equals | If
             | ConsSome | ConsOkay | ConsError | DefaultTo | UnwrapRet | UnwrapErrRet | IsOkay
             | IsNone | Asserts | Unwrap | UnwrapErr | IsErr | IsSome | TryRet | ToUInt | ToInt
-            | Len | Begin | TupleMerge => self.check_all(args),
+            | BuffToIntLe | BuffToUIntLe | BuffToIntBe | BuffToUIntBe | Len | Begin
+            | TupleMerge => {
+                // Check all arguments.
+                self.check_all(args)
+            }
```
kantai, line 196: I think that these buff conversion functions should not pass the arithmetic only checker -- so these new functions would be "FunctionNotPermitted".

src/vm/functions/mod.rs
```diff
@@ -267,6 +272,27 @@ pub fn lookup_reserved_functions(name: &str, version: &ClarityVersion) -> Option
             SetVar => SpecialFunction("special_set-var", &database::special_set_variable),
             Map => SpecialFunction("special_map", &sequences::special_map),
             Filter => SpecialFunction("special_filter", &sequences::special_filter),
+            BuffToIntLe => NativeFunction(
+                "native_buff_to_int_le",
+                NativeHandle::SingleArg(&conversions::native_buff_to_int_le),
+                // TODO: Create a dedicated cost fn for all BuffToInt-type functions.
```
kantai, line 278: Can you copy this TODO comment to the other instances, too? It'll help us make sure we don't accidentally forget to update one.

src/vm/functions/conversions.rs
```diff
@@ -0,0 +1,107 @@
...
+    direction: EndianDirection,
+    conversion_fn: fn([u8; 16]) -> Value,
+) -> Result<Value> {
+    match value {
+        Value::Sequence(SequenceData::Buffer(ref sequence_data)) => {
+            if sequence_data.len() > BufferLength(16) {
+                return Err(CheckErrors::TypeError(
+                    SequenceType(BufferType(BufferLength(16))),
+                    TypeSignature::type_of(&value),
+                )
```
kantai, line 54: You can use the `TypeValue` error to throw a CheckError without doing `type_of`.

src/vm/docs/mod.rs
```diff
@@ -173,6 +173,54 @@ const TO_INT_API: SimpleFunctionAPI = SimpleFunctionAPI {
     example: "(to-int u238) ;; Returns 238"
 };
 
+const BUFF_TO_INT_LE_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: None,
+    signature: "(buff-to-int-le (buff 16))",
+    description: "Converts a byte buffer to a signed integer use a little-endian encoding.",
```
kantai, line 179: Can you add a description of the padding behavior to the descriptions of these functions?

kantai (commented):
This looks good to me -- just some slight tweaks and a request for some more description in the docs.

src/vm/analysis/arithmetic_checker/mod.rs
gregorycoppola: done

src/vm/docs/mod.rs
gregorycoppola: done

src/vm/functions/mod.rs
gregorycoppola: done

src/vm/functions/conversions.rs
gregorycoppola: is this what you're thinking?

gregorycoppola:
> This looks good to me -- just some slight tweaks and a request for some more description in the docs.

Thanks.. addressed the comments. Not sure if the new type checking is right? I made the s/TypeError/TypeValueError/ change in the production code, but left the test unchanged. The same tests are still passing.

src/vm/functions/mod.rs
```diff
@@ -267,6 +272,30 @@ pub fn lookup_reserved_functions(name: &str, version: &ClarityVersion) -> Option
             SetVar => SpecialFunction("special_set-var", &database::special_set_variable),
             Map => SpecialFunction("special_map", &sequences::special_map),
             Filter => SpecialFunction("special_filter", &sequences::special_filter),
+            BuffToIntLe => NativeFunction(
+                "native_buff_to_int_le",
+                NativeHandle::SingleArg(&conversions::native_buff_to_int_le),
+                // TODO: Create a dedicated cost function for this case.
```
pavitthrap, line 278: just want to double check if we want to merge these "todos" as well

src/vm/mod.rs
```diff
@@ -354,12 +354,9 @@ fn eval_all(
 
 /* Run provided program in a brand new environment, with a transient, empty
  *  database.
- *
- *  Only used by CLI.
```
pavitthrap, line 358: Do we still want this comment above `execute`

pavitthrap (approved):
lgtm!

src/vm/docs/mod.rs
```diff
@@ -173,6 +173,66 @@ const TO_INT_API: SimpleFunctionAPI = SimpleFunctionAPI {
     example: "(to-int u238) ;; Returns 238"
 };
 
+const BUFF_TO_INT_LE_API: SimpleFunctionAPI = SimpleFunctionAPI {
```
jcnelson, line 176: Do we want to annotate the documentation somehow to indicate that this is only available in Clarity2?  Doesn't have to be in this PR, but if not, can you make a separate issue so we remember to do this?

src/vm/docs/mod.rs
```diff
@@ -173,6 +173,66 @@ const TO_INT_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+    signature: "(buff-to-int-be (buff 16))",
+    description: "Converts a byte buffer to a signed integer use a big-endian encoding.
+The byte buffer can be up to 16 bytes in length. If there are fewer than 16 bytes, as
+this function uses a big-endian encoding, the input behaves as if it is
+zero-padded on the _left_.",
+    example: r#"
+(buff-to-int-be 0x01) ;; Returns 1
+(buff-to-int-be 0x00000000000000000000000000000001) ;; Returns 1
+(buff-to-int-be 0xffffffffffffffffffffffffffffffff) ;; Returns -1
+(buff-to-int-le 0x) ;; Returns 0
```
jcnelson, line 217: Typo -- `buff-to-int-be`

src/vm/docs/mod.rs
```diff
@@ -173,6 +173,66 @@ const TO_INT_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+    signature: "(buff-to-uint-be (buff 16))",
+    description: "Converts a byte buffer to an unsigned integer use a big-endian encoding.
+The byte buffer can be up to 16 bytes in length. If there are fewer than 16 bytes, as
+this function uses a big-endian encoding, the input behaves as if it is
+zero-padded on the _left_.",
+    example: r#"
+(buff-to-uint-be 0x01) ;; Returns u1
+(buff-to-uint-be 0x00000000000000000000000000000001) ;; Returns u1
+(buff-to-uint-be 0xffffffffffffffffffffffffffffffff) ;; Returns u340282366920938463463374607431768211455
+(buff-to-int-le 0x) ;; Returns 0
```
jcnelson, line 232: Typo -- `buff-to-uint-le`

src/vm/functions/conversions.rs
```diff
@@ -0,0 +1,107 @@
...
+    match value {
+        Value::Sequence(SequenceData::Buffer(ref sequence_data)) => {
+            if sequence_data.len() > BufferLength(16) {
+                return Err(CheckErrors::TypeValueError(
+                    SequenceType(BufferType(BufferLength(16))),
+                    value,
+                )
+                .into());
+            } else {
+                let mut buf = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
```
jcnelson, line 57: Minor style nit: this could be `let mut buf = [0u8; 16]`.

src/vm/functions/conversions.rs
```diff
@@ -0,0 +1,107 @@
...
+use vm::representations::SymbolicExpression;
+use vm::types::BufferLength;
+use vm::types::SequenceSubtype::{BufferType, StringType};
+use vm::types::StringSubtype::ASCII;
+use vm::types::TypeSignature::SequenceType;
+use vm::types::{SequenceData, TypeSignature, Value};
+use vm::{apply, eval, lookup_function, Environment, LocalContext};
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum EndianDirection {
```
jcnelson, line 29: Do we need a specific `EndianDirection` type?  Could we get away with instead defining four different `conversion_fn` functions that simply wrapped `::from_le_bytes` and `::from_be_bytes` for `u128` and `i128`?

src/vm/functions/conversions.rs
```diff
@@ -0,0 +1,107 @@
...
+        Value::Sequence(SequenceData::Buffer(ref sequence_data)) => {
+            if sequence_data.len() > BufferLength(16) {
+                return Err(CheckErrors::TypeValueError(
+                    SequenceType(BufferType(BufferLength(16))),
+                    value,
+                )
+                .into());
+            } else {
+                let mut buf = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+                let mut original_slice = sequence_data.as_slice().to_vec();
```
jcnelson, line 58: You might be able to get away with doing an internal `malloc()` here if you instead copied the original slice directly into `buf` (no need to reverse it), and passed _that_ into the relevant `::from_be_bytes()` or `::to_be_bytes()` methods.

src/vm/functions/mod.rs
jcnelson: I don't think there's yet a way to represent cost functions for Clarity2.  If not, then @gregorycoppola please create issues to do so, and tag them as `stacks-2.1`.  Thanks!

src/vm/docs/mod.rs
gregorycoppola: see https://github.com/blockstack/stacks-blockchain/issues/2716

I also added some text for now:
`    Note: This function is only available starting with Stacks 2.1.`

src/vm/functions/mod.rs
gregorycoppola: @pavitthrap I had previously merged the todos but @kantai asked to expand them.
@jcnelson I added https://github.com/blockstack/stacks-blockchain/issues/2717

src/vm/functions/conversions.rs
gregorycoppola: The endian direction has a use apart from this, because we need to know which side to do the padding on.

src/vm/functions/conversions.rs
gregorycoppola: I made some changes to maybe make it easier to understand. And, we avoid the 'reverse'.

src/vm/mod.rs
kantai: Oh, yes, we still want this comment -- should be changed to `Only used by CLI and unit tests`

gregorycoppola: done

jcnelson (approved):
LGTM! Thanks @gregorycoppola

kantai (approved):
This LGTM! For merging, please wait until #2659 merges, and then you can merge this directly to `next`

PR #2692, opened 2021-06-06
Update version of tini dependency

citrux:
Sup guys! I'm one of the `tini` authors and downloading stats of my library make me cry 😢 

изображение

You're main consumers of this smal but epic library, but why you're using realy outdated version of this masterpiece? I can't bear it anymore, I feel useless and depressed 😞 

I have updated version, don't thank me.
изображение

Have a nice day!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ name = "block_limits"
 harness = false
 
 [dependencies]
-tini = "0.2"
+tini = "1.2"
 rand = "=0.7.2"
 rand_chacha = "=0.2.2"
 serde = "1"


jcnelson:
Hey, thanks for the PR!  Can you please agree to the CLA so we can merge it?  Thanks!

citrux:
Screenshot 2021-06-09 at 23 41 47

kantai:
Thanks for opening this @citrux, but it turns out that our `tini` dependency was only necessary for an unused code path, so rather than updating our `tini` dependency, we're going to remove it via #2709.

PR #2694, opened 2021-06-08
Fix: #2644 reorder UTXO staleness and RBF-limit check

### LINKED ISSUE #2644: RBF logic fails to reap stale outstanding TXs if RBF limit had been reached
A testnet miner ended up stalling seemingly due to an issue in the RBF logic.

I believe the relevant code is here:

https://github.com/blockstack/stacks-blockchain/blob/67dccdf6d1c28e482d3d71eb616cf8682dac863d/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs#L1033-L1073

In the first code block: we check if an outstanding TX has been committed.
In the second code block: we check if we're already exceeding the RBF limit.
In the third code block: we check if we should abandon any outstanding TXs because they've been reorged or stale.

I think that in the following scenario:
* Outstanding TXs cause the miner to reach the "RBF limit"
* Those outstanding TXs are reorged or never mined

The second code block would prevent the miner from ever getting to the third block and therefore some stale outstanding TXs would prevent the miner from ever producing a new commitment.

kantai:
## Description

This fixes #2644 by simply re-ordering the UTXO staleness check so that it occurs before the RBF-limit check. This ensures that if stale UTXOs reached the "RBF limit" a miner will recover by resetting the UTXO cache.

## Type of Change
- Bug fix

No

## Testing information

This can be tested on testnet, where current tBTC mining behavior rapidly exposes the #2644 issue.

### DIFF
--- a/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
+++ b/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
@@ -1043,18 +1043,6 @@ impl BitcoinRegtestController {
             }
         }
 
-        // Stop as soon as the fee_rate is ${self.config.burnchain.max_rbf} percent higher, stop RBF
-        if ongoing_op.fees.fee_rate
-            > (self.config.burnchain.satoshis_per_byte * self.config.burnchain.max_rbf / 100)
-        {
-            warn!(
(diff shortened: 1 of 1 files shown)


jcnelson (approved):
Thanks @kantai!  LGTM

lgalabru (approved):
LGTM! thank you @kantai

PR #2697, opened 2021-06-08
[Clarity 2] Adds Clarity functions to convert between string's and int's

gregorycoppola:
# Description

This PR adds clarity language functions to convert between strings and integers. Conversion can work in both directions, and supports both uint and int, and both string-ascii and string-utf8.

Fixes issues/2617

Function signatures are:
    signature: "(string-to-int (string-ascii|string-utf8))",
    signature: "(string-to-uint (string-ascii|string-utf8))",
    signature: "(int-to-ascii (int|uint))",
    signature: "(int-to-utf8 (int|uint))",

Note that the input types can be inferred (e.g. 'uint' vs 'int'). But, the output types are part of the function name, to avoid impossible ambiguity.

# Testing
We created four new test suites in src/vm/tests/conversions.rs


src/vm/analysis/read_only_checker/mod.rs
```diff
@@ -180,12 +180,12 @@ impl<'a, 'b> ReadOnlyChecker<'a, 'b> {
...
             | Secp256k1Verify | ConsSome | ConsOkay | ConsError | DefaultTo | UnwrapRet
             | UnwrapErrRet | IsOkay | IsNone | Asserts | Unwrap | UnwrapErr | Match | IsErr
-            | IsSome | TryRet | ToUInt | ToInt | Append | Concat | AsMaxLen | ContractOf
-            | PrincipalOf | ListCons | GetBlockInfo | TupleGet | TupleMerge | Len | Print
-            | AsContract | Begin | FetchVar | GetStxBalance | StxGetAccount | GetTokenBalance
-            | GetAssetOwner | GetTokenSupply | ElementAt | IndexOf => {
-                self.check_all_read_only(args)
-            }
+            | IsSome | TryRet | ToUInt | ToInt | BuffToIntLe | BuffToUIntLe | BuffToIntBe
+            | BuffToUIntBe | StringToInt | StringToUInt | IntToAscii | IntToUtf8 | Append
```
kantai, line 184: Do the new functions use 2 or 1 argument? It looks like this branch performs a "2" argument check, rather than a 1 argument check.

gregorycoppola:
I'm deleting this PR for now.

I opened it against branch 'feat/conversion21', which is another branch I have out for review.

But, the diffs are confusing because that branch is changing.

I can reopen later.

PR #2699, opened 2021-06-08
Fix the warning messages in 'cargo build'

### LINKED ISSUE #2698: Fix Rust Build Warnings
**Describe the bug**
There are 12 warnings when we run 'cargo build'. These make it hard to see what the real errors are.

We should remove these.

**Steps To Reproduce**
'cargo build'

gregorycoppola:
Fixes 2698

## The Change
This PR fixes 12 build warnings in the rust code that we get when we run 'cargo build'. This greatly simplifies the output.

## Testing
After this change, the output from running 'cargo build' is just.

```
Gregs-MacBook-Pro:work1 gregcoppola$ cargo build
Compiling blockstack-core v0.0.1 (/Users/gregcoppola/work1)
Finished dev [unoptimized + debuginfo] target(s) in 23.71s
```

### DIFF
--- a/src/chainstate/coordinator/comm.rs
+++ b/src/chainstate/coordinator/comm.rs
@@ -167,7 +167,7 @@ impl CoordinatorChannels {
                 return false;
             }
             thread::sleep(Duration::from_millis(100));
-            std::sync::atomic::spin_loop_hint();
+            std::hint::spin_loop();
         }
         return true;
     }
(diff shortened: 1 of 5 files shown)


kantai (commented):
This LGTM, but are these all present in `develop` as well? If so, I think this PR should be rebased/opened against `develop`.

gregorycoppola:
OK. I'll reopen against 'develop'.

PR #2704, opened 2021-06-11
Chore: update default read_count

lgalabru:
In this PR, we're fine tuning the default value used for controlling how many times a readonly RPC contract-call can hit storage.
With the previous default value, sophisticated contracts like Arkadiko would end up being unusable by default.

### DIFF
--- a/src/net/connection.rs
+++ b/src/net/connection.rs
@@ -436,7 +436,7 @@ impl std::default::Default for ConnectionOptions {
                 write_length: 0,
                 write_count: 0,
                 read_length: 100000,
-                read_count: 30,
+                read_count: 64,
                 runtime: 1_000_000_000,
             },
             maximum_call_argument_size: 20 * BOUND_VALUE_SERIALIZATION_HEX,


jcnelson (approved):

reedrosenbluth (approved):

pavitthrap:
How did you choose the new value / is there a reason not to make it any higher?

kantai:
> How did you choose the new value / is there a reason not to make it any higher?

+1 here, do we have any sense for the trade-offs for this as a default? This could be tested by just changing the config setting in deployments as well.

lgalabru:
We're not bumping the marf_read by an order of magnitude, just adding a few more reads. After getting some feedbacks from Arkadiko (which is a good candidate for Goldilocks-ing this values), that for certain operations involving contract-calls + traits, this specific limit is by default a bit aggressive.

jcnelson:
Did Arkadiko give a tighter upper-bound on the `read_count` they need?  Each MARF index lookup can be on the order of a millisecond, so I wouldn't be comfortable pushing this beyond 64 for now.

lgalabru:
Follow up - Arkadiko was able to do an optimization pass. Closing this issue.

PR #2709, opened 2021-06-14
Chore: remove tini dependency

kantai:
This removes the `tini` dependency.

The only codepath that actually used the tini reader was a CLI command for replaying chainstate (this CLI command may already have been defunct), but that codepath only needed the tini reader for determining the first burn block hash and first burn timestamp, both of which were otherwise available.


lgalabru (approved):
LGTM, thank you @kantai!

jcnelson (approved):
LGTM!

kantai:
Merging -- the failing `sampled-genesis` test is a flapping test. I'm going to work on a PR to improve the integration tests.

PR #2715, opened 2021-06-15
chore: try to speed up some integration tests

kantai:
This is a simple attempt to tweak some of the settings in `neon_integrations` to speed those tests up some. So far, this just tweaked the microblock integrations tests (picked up about 15-20 minutes locally).

### DIFF
--- a/testnet/stacks-node/src/tests/neon_integrations.rs
+++ b/testnet/stacks-node/src/tests/neon_integrations.rs
@@ -1555,7 +1555,7 @@ fn microblock_integration_test() {
             .json::<RPCPeerInfoData>()
             .unwrap();
         eprintln!("{:#?}", tip_info);
-        if tip_info.unanchored_tip != StacksBlockId([0; 32]) {
+        if tip_info.unanchored_tip == StacksBlockId([0; 32]) {
             iter_count += 1;
             assert!(
                 iter_count < 10,
--- a/testnet/stacks-node/src/tests/neon_integrations.rs
+++ b/testnet/stacks-node/src/tests/neon_integrations.rs
@@ -310,12 +310,12 @@ fn wait_for_runloop(blocks_processed: &Arc<AtomicU64>) {
 fn wait_for_microblocks(microblocks_processed: &Arc<AtomicU64>, timeout: u64) -> bool {
     let mut current = microblocks_processed.load(Ordering::SeqCst);
     let start = Instant::now();
-
+    info!("Waiting for next microblock");
     loop {
         let now = microblocks_processed.load(Ordering::SeqCst);
         if now == 0 && current != 0 {
             // wrapped around -- a new epoch started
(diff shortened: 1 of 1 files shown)


pavitthrap (approved):

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -2280,7 +2277,7 @@ fn size_overflow_unconfirmed_stream_microblocks_integration_test() {
     let mut ctr = 0;
     while ctr < flat_txs.len() {
         submit_tx(&http_origin, &flat_txs[ctr]);
-        if !wait_for_microblocks(&microblocks_processed, 240) {
+        if !wait_for_microblocks(&microblocks_processed, 60) {
```
jcnelson, line 2280: Due to the time it takes for gigantic microblocks to process, 60 seconds isn't enough to run it locally.  It really takes about 4 minutes.

jcnelson:
So, the reason the size-overflow unconfirmed microblock test is failing is _precisely because_ the timeout values were tweaked.  The timeout values give the node a chance to find the expensive transactions in the mempool and then mine them into microblocks before the next Stacks block gets mined.

kantai:
> So, the reason the size-overflow unconfirmed microblock test is failing is precisely because the timeout values were tweaked. The timeout values give the node a chance to find the expensive transactions in the mempool and then mine them into microblocks before the next Stacks block gets mined.

The only change in that test case was removing two `wait_ms(150_000)` calls, which did have the effect of usually getting a microblock mined. I added `wait_for_microblock` calls there now instead, which should have the same effect, but not require an unnecessary wait.

jcnelson:
> I added wait_for_microblock calls there now instead, which should have the same effect, but not require an unnecessary wait.

The effect we're waiting for is for the node to mine a microblock _with the big transaction_.  Just waiting for a microblock isn't enough, nor is speeding up the time between microblocks.  The code really needs to wait until the big transaction gets picked up in the unconfirmed microblock stream.  Ideally, it'd wait until the miner tries to pick up _two_ big transactions, and then _fails_ to mine the second one because it would overflow the stream budget, but I don't know of a good way to test that as it happens.

kantai:
> The effect we're waiting for is for the node to mine a microblock with the big transaction. Just waiting for a microblock isn't enough, nor is speeding up the time between microblocks. The code really needs to wait until the big transaction gets picked up in the unconfirmed microblock stream. Ideally, it'd wait until the miner tries to pick up two big transactions, and then fails to mine the second one because it would overflow the stream budget, but I don't know of a good way to test that as it happens.

Calling `wait_for_microblocks` until it returns false would have this effect, and it would have the benefit of not requiring a magic 2.5 minute wait.

jcnelson (approved):
Thanks for cleaning this up @kantai!

PR #2721, opened 2021-06-18
feat(Clarity2): Functions to convert between strings and ints

### LINKED ISSUE #2617: Extension contract: string-ascii/string-utf8 to/from int and uint
There needs to be an extension contract that converts a `string-ascii` or a `string-utf8` to/from a `int` or `uint`.

gregorycoppola:
Fixes #2617

## About this change
This PR adds functions` string-to-int`, `string-to-uint`, `int-to-ascii`, and `int-to-utf8`. These functions allow conversion between integers and string representations.

## Testing
The following tests have been added to:
```
M     src/vm/docs/mod.rs
M	src/vm/tests/conversions.rs
M	src/vm/tests/costs.rs
```


jcnelson:
Is this a dup of #2720?

gregorycoppola:
> Is this a dup of #2720?

Yes.. this is not mailed out yet, but it's almost ready. I was just using the PR to look at the diffs.

src/vm/analysis/type_checker/tests/mod.rs
```diff
@@ -1328,6 +1328,60 @@ fn test_buffer_to_ints() {
...
+fn test_string_to_ints() {
+    let good = [
+        r#"(int-to-ascii 1)"#,
+        r#"(int-to-ascii u1)"#,
+        r#"(int-to-utf8 1)"#,
+        r#"(int-to-utf8 u1)"#,
+        r#"(string-to-int "1")"#,
+        r#"(string-to-int u"1")"#,
+        r#"(string-to-uint "1")"#,
+        r#"(string-to-uint u"1")"#,
```
jcnelson, line 1341: Can you add some tests here for negative numbers?

src/vm/analysis/type_checker/tests/mod.rs
```diff
@@ -1328,6 +1328,60 @@ fn test_buffer_to_ints() {
...
+        "int",
+        "uint",
+        "uint",
+    ];
+
+    let bad = [
+        r#"(int-to-ascii 0x0001 0x0001)"#,
+        r#"(int-to-ascii)"#,
+        r#"(int-to-ascii 0x000102030405060708090a0b0c0d0e0f00)"#,
+        r#"(int-to-ascii "a")"#,
```
jcnelson, line 1359: Can you add coverage for `int-to-utf8`?

src/vm/docs/mod.rs
```diff
@@ -241,6 +241,55 @@ Note: This function is only available starting with Stacks 2.1.",
...
+const STRING_TO_UINT_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: None,
+    signature: "(string-to-uint (string-ascii|string-utf8))",
+    description:
+        "Converts a string, either `string-ascii` or `string-utf8`, to an unsigned integer.
+
+        Note: This function is only available starting with Stacks 2.1.",
+    example: r#"
+(string-to-uint "1") ;; Returns u1
+(string-to-uint u"1") ;; Returns u1
```
jcnelson, line 265: What happens if you give it an invalid numeric representation, like `"abc"`?  What's the error code?

src/vm/functions/conversions.rs
```diff
@@ -18,11 +18,13 @@ use vm::costs::cost_functions::ClarityCostFunction;
...
 use vm::representations::SymbolicExpression;
-use vm::types::BufferLength;
 use vm::types::SequenceSubtype::{BufferType, StringType};
 use vm::types::StringSubtype::ASCII;
 use vm::types::TypeSignature::SequenceType;
-use vm::types::{SequenceData, TypeSignature, Value};
+use vm::types::{
+    ASCIIData, BuffData, BufferLength, CharType, SequenceData, TypeSignature, UTF8Data, Value,
+};
+// use vm::types::{SequenceData, TypeSignature, Value};
```
jcnelson, line 27: Please remove all commented-out code

src/vm/functions/conversions.rs
```diff
@@ -113,3 +115,107 @@ pub fn native_buff_to_uint_be(value: Value) -> Result<Value> {
...
     return buff_to_int_generic(value, EndianDirection::BigEndian, convert_to_uint_be);
 }
+
+pub fn native_string_to_int_generic(
+    value: Value,
+    conversion_fn: fn(String) -> Result<Value>,
+) -> Result<Value> {
+    match value {
+        Value::Sequence(SequenceData::String(CharType::ASCII(ASCIIData { data }))) => {
+            let as_string = String::from_utf8(data).unwrap();
```
jcnelson, line 125: This is a runtime crash waiting to happen.  There are exactly zero circumstances where a `.unwrap()` or `.expect()` is acceptable when processing user-submitted input.

src/vm/functions/conversions.rs
```diff
@@ -113,3 +115,107 @@ pub fn native_buff_to_uint_be(value: Value) -> Result<Value> {
...
+            return conversion_fn(as_string);
+        }
+        Value::Sequence(SequenceData::String(CharType::UTF8(UTF8Data { data }))) => {
+            let flat = data.into_iter().flatten().collect();
+            let as_string = String::from_utf8(flat).unwrap();
```
jcnelson, line 130: Same here.  Never `.unwrap()` something that's from a (potentially-hostile) user.

src/vm/functions/conversions.rs
```diff
@@ -113,3 +115,107 @@ pub fn native_buff_to_uint_be(value: Value) -> Result<Value> {
...
+                ],
+                value,
+            )
+            .into())
+        }
+    };
+}
+
+pub fn native_string_to_int(value: Value) -> Result<Value> {
+    fn convert_to_int(raw_string: String) -> Result<Value> {
```
jcnelson, line 147: Style nit -- we don't have nested functions.  Please either make this into a top-level function, or turn it into an anonymous function argument to `native_string_to_int_generic()`.  Same goes for `native_string_to_uint` below.

src/vm/functions/conversions.rs
```diff
@@ -113,3 +115,107 @@ pub fn native_buff_to_uint_be(value: Value) -> Result<Value> {
...
+        let possible_int = raw_string.parse::<i128>();
+        match possible_int {
```
jcnelson, line 149: This `match` statement really should not return an `Err(..)`, because the user may submit a code snippet like `(string-to-uint "this is definitely not an integer")`.  An `InterpreterError` will cause the transaction to be invalid, which is not what we want.

What we really want is for `(string-to-uint)` and `(string-to-int)` to return either `(optional int|uint)` or `(response int|uint int)` -- i.e. a type that represents a possibly-successful but not guaranteed-successful conversion of the string type to an integer type.  I'm not too picky on `optional` vs `response`, but the function must evaluate to either a "success" or "failed" state that the caller can handle.

As to what this means for this `match` statement, you should be returning `Value::some(Value::Int(val))` or `Value::none()`, or `Value::okay(Value::Int(val))` or `Value::error(Value::Int(error_code))`.  Same goes for `native_string_to_uint` below.

src/vm/functions/conversions.rs
```diff
@@ -113,3 +115,107 @@ pub fn native_buff_to_uint_be(value: Value) -> Result<Value> {
...
+        })));
+    }
+    return native_int_to_string_generic(value, convert_to_ascii);
+}
+
+pub fn native_int_to_utf8(value: Value) -> Result<Value> {
+    // Given a string representing an integer, convert this to Clarity UTF8 value.
+    // The conversion of an integer into a string is going to be ASCII-compliant, therefor, we just need
+    // to wrap each individual character in another vector.
+    fn convert_to_utf8(raw_string: String) -> Value {
```
jcnelson, line 208: Instead of creating a bespoke consensus-critical conversion routine from a utf8 literal to a `CharType::UTF8()`, can you just use the existing `Value::string_utf8_from_string_utf8_literal()`?

src/vm/functions/conversions.rs
```diff
@@ -113,3 +115,107 @@ pub fn native_buff_to_uint_be(value: Value) -> Result<Value> {
...
+                value,
+            )
+            .into())
+        }
+    };
+}
+
+pub fn native_int_to_ascii(value: Value) -> Result<Value> {
+    // Given a string representing an integer, convert this to Clarity ASCII value.
+    fn convert_to_ascii(raw_string: String) -> Value {
```
jcnelson, line 195: Instead of creating a bespoke consensus-critical conversion routine from a utf8 literal (because in Rust, all Strings are utf8) to a `CharType::ASCII()`, can you just use the existing `Value::string_ascii_from_bytes()` routine, and convert the `InterpreterResult` into a `response`?

src/vm/tests/conversions.rs
```diff
@@ -16,11 +16,13 @@
 
 pub use vm::analysis::errors::{CheckError, CheckErrors};
 use vm::execute_v2;
-use vm::types::BufferLength;
 use vm::types::SequenceSubtype::{BufferType, StringType};
 use vm::types::StringSubtype::ASCII;
 use vm::types::TypeSignature::SequenceType;
-use vm::types::{ASCIIData, BuffData, CharType, SequenceData, Value};
+// use vm::types::{ASCIIData, BuffData, CharType, SequenceData, Value};
```
jcnelson, line 22: Please remove

src/vm/tests/conversions.rs
```diff
@@ -230,3 +232,137 @@ fn test_simple_buff_to_uint_be() {
...
+        execute_v2(wrong_type_error_test).unwrap_err(),
+        CheckErrors::UnionTypeValueError(
+            vec![
+                TypeSignature::max_string_ascii(),
+                TypeSignature::max_string_utf8(),
+            ],
+            Value::Int(1)
+        )
+        .into()
+    );
```
jcnelson, line 267: Can you add test coverage for _all_ ASCII characters, including the unprintable ones?  Also, can you try converting unprintable UTF-8 strings, or nonsensical UTF-8 strings like emojis and the like, into integers, and verify that they fail with `(err ...)` or `none`?

src/vm/tests/conversions.rs
```diff
@@ -230,3 +232,137 @@ fn test_simple_buff_to_uint_be() {
...
```
jcnelson, line 301: Same here -- please check a wider array of bad strings.

src/vm/functions/conversions.rs
```diff
@@ -113,3 +115,107 @@ pub fn native_buff_to_uint_be(value: Value) -> Result<Value> {
...
+    }
+    return native_string_to_int_generic(value, convert_to_int);
+}
+
+pub fn native_string_to_uint(value: Value) -> Result<Value> {
+    fn convert_to_uint(raw_string: String) -> Result<Value> {
+        let possible_int = raw_string.parse::<u128>();
+        match possible_int {
+            Ok(val) => return Ok(Value::UInt(val)),
+            Err(error) => return Err(CheckErrors::InvalidCharactersDetected.into()),
```
jcnelson, line 162: Same as with `natve_string_to_int()`, this *should not abort the transaction* if the user submits a string that is too big (or small) to be represented in a `i128` or `u128`.  This should be a runtime error that the caller can handle, so the return type here should be a `(optional int|uint)` or `(response int|uint int)`.

src/vm/analysis/type_checker/tests/mod.rs
gregorycoppola: i added fuller coverage for all four functions

src/vm/docs/mod.rs
gregorycoppola: I changed this so now "string to int" functions will return an "optional", and updated the docs.

src/vm/functions/conversions.rs
gregorycoppola: good point

src/vm/functions/conversions.rs
gregorycoppola: good catch! I believe I have removed all of the "unwrap" instances from the production code in this change.

src/vm/functions/conversions.rs
gregorycoppola: good catch

src/vm/functions/conversions.rs
gregorycoppola: I removed the nested functions from this change.

src/vm/functions/conversions.rs
gregorycoppola: i've gone with the "optional approach", and changed the documentation and tests accordingly.. what do you think?

src/vm/functions/conversions.rs
gregorycoppola: yes.. it's an optional now

src/vm/functions/conversions.rs
gregorycoppola: yes.. good idea.. thanks!

src/vm/functions/conversions.rs
gregorycoppola: is it possible to just use `Value::string_utf8_from_bytes` here, because we know the output of "int -> string" will be within the ASCII subset?

Also, in case this call fails, I have added a new InternalError to the erros in `src/vm/analysis/errors.rs`. Do you think that is appropriate?

src/vm/tests/conversions.rs
gregorycoppola: I added some weird character examples, as well as an integer that has a very long length.

src/vm/tests/conversions.rs
gregorycoppola: i added some more

src/vm/analysis/errors.rs
```diff
@@ -175,6 +175,8 @@ pub enum CheckErrors {
 
     WriteAttemptedInReadOnly,
     AtBlockClosureMustBeReadOnly,
+
+    InternalError,
```
gregorycoppola, line 179: Note: I added "InternalError" here, because I didn't see it.

src/vm/functions/conversions.rs
```diff
@@ -113,3 +114,93 @@ pub fn native_buff_to_uint_be(value: Value) -> Result<Value> {
...
+pub fn native_int_to_string_generic(
+    value: Value,
+    conversion_fn: fn(bytes: Vec<u8>) -> Result<Value>,
+) -> Result<Value> {
+    match value {
+        Value::Int(ref int_value) => {
+            let as_string = int_value.to_string();
+            match conversion_fn(as_string.into()) {
+                Ok(value) => Ok(value),
+                Err(_error) => Err(CheckErrors::InternalError.into()),
```
kantai, line 180: Is it possible to reach this case? `int_value.to_string()` will always produce a string that would fit ASCII or UTF8, right? If it can't be reached, this should be switched to an `expects("PANIC: ...")` statement, and then we won't need the `InternalError` variant anymore.

gregorycoppola: This call to "conversion_fn" should always work... that's why I was going to use InternalError.

It was based on the suggestion of @jcnelson not to use "unwrap" or "expect" in production code.

https://github.com/blockstack/stacks-blockchain/pull/2721#discussion_r656549977

gregorycoppola:
Looks like there is some discussion around the error handling so I'm sending it back with some comments.

src/vm/functions/conversions.rs
kantai: Hmmm, I think in this case, we want to expect or unwrap -- in general, we should expect/unwrap in places that should never fail: it's better for the process to end abruptly than continue to operate with broken assumptions. I think from the discussion linked, the main issue was for `string -> int`. But for `int -> string`, we should be confident that the conversion never fails.

src/vm/functions/conversions.rs
gregorycoppola: Note: I removed InternalError.

src/vm/analysis/errors.rs
gregorycoppola: I removed InternalError, using 'expect' instead as discussed.

gregorycoppola:
I removed the InternalError under discussion and substituted some "expect"s.

src/vm/functions/mod.rs
```diff
@@ -283,6 +287,30 @@ pub fn lookup_reserved_functions(name: &str, version: &ClarityVersion) -> Option
                 // TODO: Create a dedicated cost function for this case.
                 ClarityCostFunction::Mul,
             ),
+            StringToInt => NativeFunction(
+                "native_string_to_int",
+                NativeHandle::SingleArg(&conversions::native_string_to_int),
+                // TODO: Create a dedicated cost function for this case.
+                ClarityCostFunction::Mul,
```
reedrosenbluth, line 294: I wonder if it would be worth creating an `Unimplemented` cost function enum member to make it easy to keep track of the cost functions we will need to implement before 2.1

src/vm/functions/conversions.rs
```diff
@@ -113,3 +114,91 @@ pub fn native_buff_to_uint_be(value: Value) -> Result<Value> {
...
+    }
+}
+
+pub fn native_string_to_uint(value: Value) -> Result<Value> {
+    native_string_to_int_generic(value, safe_convert_string_to_uint)
+}
+
+pub fn native_int_to_string_generic(
+    value: Value,
+    conversion_fn: fn(bytes: Vec<u8>) -> Result<Value>,
```
reedrosenbluth, line 173: nit: I think it's a bit confusing that this conversion function expects a byte vector instead of an int, while the previous one (for string->int) expects a string.

reedrosenbluth (approved):
a few minor comments, but otherwise LGTM!

kantai (approved):
This LGTM!

src/vm/functions/mod.rs
kantai: This could be useful, I can open a PR that adds the `Unimplemented` enum member pretty quickly -- I think it just needs to be special cased in the cost tracker to be "always free", rather than introducing a new vote-able function.

kantai: https://github.com/blockstack/stacks-blockchain/pull/2751

src/vm/functions/conversions.rs
gregorycoppola: I added some comments to the functions that receive these functional inputs. Even though it's counter-intuitive, these are, I think, the most natural inputs.

In the case of the function that takes bytes, this is to interface with `Value::string_ascii_from_bytes` and `Value::string_utf8_from_bytes`. So, otherwise, I would need to introduce wrappers around these.

src/vm/functions/mod.rs
gregorycoppola: OK. If your change is in first I can update this.

gregorycoppola:
Thanks for the LGTM's @kantai and @reedrosenbluth . I'll wait for @jcnelson 's LGTM before submitting because he has been commenting.

src/vm/tests/conversions.rs
jcnelson: Can you add a test for `(string-to-uint -1)` to verify that it fails typecheck?

src/vm/types/signatures.rs
```diff
@@ -637,6 +644,24 @@ impl TypeSignature {
...
+                u32::try_from(MAX_VALUE_SIZE)
+                    .expect("FAIL: Max Clarity Value Size is no longer realizable in ASCII Type"),
+            ),
+        )))
+    }
+
+    pub fn max_string_utf8() -> TypeSignature {
+        SequenceType(SequenceSubtype::StringType(StringSubtype::UTF8(
+            StringUTF8Length(
+                u32::try_from(MAX_VALUE_SIZE)
```
jcnelson, line 659: I think this should be `MAX_VALUE_SIZE / 4`, since Clarity UTF-8 strings represent each character as 4 bytes (even if the character itself can be represented in fewer bytes).

src/vm/tests/conversions.rs
jcnelson: It also might be worth adding a test that tries to convert a too-big string into an integer (or an unsigned integer), and verifies that it succeeds for the maximal size but fails for a string with one byte more.  That way, we'll be covering the boundary condition.

src/vm/functions/conversions.rs
jcnelson: Works for me!

src/vm/functions/conversions.rs
gregorycoppola: done

src/vm/types/signatures.rs
gregorycoppola: OK. That makes sense to me. @kantai do you have an opinion on this?

kantai: Hmm -- this actually raises a slightly separate but good point. What we want to do here is change this to something like:


```rust
    pub fn max_string_ascii() -> TypeSignature {
        SequenceType(SequenceSubtype::StringType(StringSubtype::ASCII(
            BufferLength::try_from(MAX_VALUE_SIZE)
                .expect("FAIL: Max Clarity Value Size is no longer realizable in ASCII Type"),
            ),
        ))
    }

    pub fn max_string_utf8() -> TypeSignature {
        SequenceType(SequenceSubtype::StringType(StringSubtype::UTF8(
            StringUTF8Length::try_from(MAX_VALUE_SIZE)
                .expect("FAIL: Max Clarity Value Size is no longer realizable in UTF8 Type"),
            ),
        ))
    }
```

Which will use the `try_from` check to enforce that the actual Clarity value size is legal (I believe @jcnelson is right that MAX_VALUE_SIZE isn't, but this way we'd get a panic indicating such).

Somewhat relatedly, we really should change `BufferLength(pub u32)` back to `BufferLength(u32)` -- I'll open a PR on `develop` that does exactly that. Nothing in the codebase violates the checked invariant, but the whole idea of the various `Length` classes is that their constructors enforce legal lengths.

jcnelson (approved):
Just need to address the last comment about `max_value_size` and this LGTM.  Thanks @gregorycoppola!

src/vm/types/signatures.rs
gregorycoppola: OK. Using StringUTF8Length::try_from(MAX_VALUE_SIZE / 4).

gregorycoppola (commented):
thanks for the review!

PR #2722, opened 2021-06-18
docs: added text on the return value of var-set

pgray-hiro:
## Description

This PR adds text about the return value of the `var-set` Clarity function. This clarification came from discussion that I had with @reedrosenbluth regarding the usage of `asserts!` with the `var-set` function, where we discovered that it was not possible for `var-set` to return any other value than `true`.

## Type of Change

- [ ] New feature
- [ ] Bug fix
- [ ] Other

No

No

## Testing information

No tests are required for this PR.

## Checklist
- [x] `npm run test` passes
- [x] Changelog is updated


jcnelson (approved):

kantai (approved):
LGTM!

PR #2723, opened 2021-06-18
feat(Clarity2): "is-standard" a function to determine whether a 'principal' matches the network type

### LINKED ISSUE #2642: Extension contract: determine whether or not an address is a "standard" address
Make an extension contract that takes a `principal` and determines whether or not its version byte is recognized by the network.

gregorycoppola:
Fixes #2642

## The Change
This change adds `is-standard`, a Clarity function that tests whether a principal addresses matches the network type.

That is, the network is either
of type `mainnet`, or `testnet`. And, `principal` is either suited for the
mainnet or a testnet.  That is, `SPxxxx` and `SMxxxx` addresses are only usable
on mainnet, whereas `STxxxx` and `SNxxxx` are only usable on testnet and
regtest. This method will return `true` iff the principal matches the network-type.

## Testing
The following tests have been added to:
```
M       src/vm/docs/mod.rs
M	src/vm/tests/costs.rs
M	src/vm/tests/mod.rs
A	src/vm/tests/principals.rs
```


src/vm/docs/mod.rs
```diff
@@ -241,6 +241,25 @@ Note: This function is only available starting with Stacks 2.1.",
 "#,
 };
 
+const IS_STANDARD_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: None,
+    signature: "(is-standard principal)",
+    description:
+        "Tests whether `principal` matches the current network type. That is, the network is either
```
jcnelson, line 248: `principal` is the name of a type, so we should probably use a different word here.

Also, the copy should be more specific about what this means, like `represents a principal that can spend tokens`.  The thing is, principal addresses with unspendable version bytes will be accepted as literals (i.e. you can `stx-transfer?` to an address with a non-standard version byte); it's just that any tokens they receive won't be spendable.

src/vm/docs/mod.rs
```diff
@@ -241,6 +241,25 @@ Note: This function is only available starting with Stacks 2.1.",
...
+of type `mainnet`, or `testnet`. And, `principal` is either suited for the
+mainnet or a testnet.  That is, `SPxxxx` and `SMxxxx` addresses are only usable
+on mainnet, whereas `STxxxx` and `SNxxxx` are only usable on testnet and
+regtest. This method will return `true` iff the principal matches the network-type.
```
jcnelson, line 252: Might want to clarify that no other version bytes are currently spendable.  Also, I don't know if the typical reader knows that `iff` means `if and only if`.

src/vm/docs/mod.rs
```diff
@@ -241,6 +241,25 @@ Note: This function is only available starting with Stacks 2.1.",
...
+
+Note: This function is only available starting with Stacks 2.1.",
+    example: r#"
+(is-standard 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6) ;; returns true on testnet and false on mainnet
+(is-standard 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6.foo) ;; returns true on testnet and false on mainnet
+(is-standard 'SP3X6QWWETNBZWGBK6DRGTR1KX50S74D3433WDGJY) ;; returns true on mainnet and false on testnet
+(is-standard 'SP3X6QWWETNBZWGBK6DRGTR1KX50S74D3433WDGJY.foo) ;; returns true on mainnet and false on testnet
```
jcnelson, line 259: Can you add an example of an address that is not standard on either mainnet or testnet?

src/vm/functions/principals.rs
```diff
@@ -0,0 +1,51 @@
...
+    C32_ADDRESS_VERSION_TESTNET_MULTISIG, C32_ADDRESS_VERSION_TESTNET_SINGLESIG,
+};
+
+pub fn special_is_standard(
+    args: &[SymbolicExpression],
+    env: &mut Environment,
+    context: &LocalContext,
+) -> Result<Value> {
+    check_argument_count(1, args)?;
+    runtime_cost(ClarityCostFunction::StxTransfer, env, 0)?;
```
jcnelson, line 27: This is not the correct cost function to use here, since this is a Stacks 2.1 method.

src/vm/tests/costs.rs
```diff
@@ -83,6 +83,7 @@ pub fn get_simple_test(function: &NativeFunctions) -> &'static str {
         BuffToUIntLe => "(buff-to-uint-le 0x00000000000000000000000000000001)",
         BuffToIntBe => "(buff-to-int-be 0x00000000000000000000000000000001)",
         BuffToUIntBe => "(buff-to-uint-be 0x00000000000000000000000000000001)",
+        IsStandard => "(buff-to-uint-be 0x00000000000000000000000000000001)",
```
jcnelson, line 86: This code snippet is wrong.

src/vm/mod.rs
```diff
@@ -352,6 +352,21 @@ fn eval_all(
     })
 }
 
+/* Run provided program in an environment specified by the user. This version of the function
+  gives the most leeway to the caller to create a custom environment.
+*/
+pub fn execute_program_with_context(
```
jcnelson, line 358: If you just needed this for testing, can you instead revert the changes to this file and put the function into your test module?

gregorycoppola: 1)  we would then have to expose "eval_all"
2) we can use this in other tests

However, it is possible to just expose a version that is a function of (clarity_version, is_mainnet), and avoid exposing such a complicated method. What would you say about that?

src/vm/functions/principals.rs
gregorycoppola: which cost function would you use?

src/vm/docs/mod.rs
gregorycoppola: what do you think of the current wording?

gregorycoppola:
@jcnelson thanks for the comments... I'm wondering what you think of the current docs wording?

src/vm/functions/principals.rs
kantai: Once #2751 merges, we should use Unimplemented here.

src/vm/mod.rs
kantai: I'd support just adding an argument to `execute_against_version`

kantai (approved):
This LGTM

gregorycoppola (commented):
Thanks for the review!

pavitthrap (approved):
LGTM!

src/vm/docs/mod.rs
```diff
@@ -241,6 +241,29 @@ Note: This function is only available starting with Stacks 2.1.",
...
+    signature: "(is-standard principal1)",
+    description:
+"Tests whether `principal1` matches the current network type, and therefor
+represents a principal that can spend tokens on the current network type. That is, the network is either of
+type `mainnet`, or `testnet`. And, `principal1` is either suited for,
+or able to spend tokens on, the mainnet or a testnet.  `SPxxxx` and
+`SMxxxx` addresses can only spend tokens on mainnet, whereas `STxxxx` and `SNxxxx` addresses can only
+spend tokens on a testnet. Other than these four forms of address, no other forms of addresses are valid.
+This method will return `true` if and only if the principal matches the
+network-type, and false otherwise.
```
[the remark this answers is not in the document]
gregorycoppola: good points 👍

src/vm/docs/mod.rs
```diff
@@ -241,6 +241,29 @@ Note: This function is only available starting with Stacks 2.1.",
 "#,
 };
 
+const IS_STANDARD_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: None,
+    signature: "(is-standard principal1)",
```
jcnelson, line 246: Not that this is wrong, but I'm curious why this is `principal1` and not something more communicative, like `standard-or-contract-principal` or something?

src/vm/docs/mod.rs
```diff
@@ -241,6 +241,29 @@ Note: This function is only available starting with Stacks 2.1.",
...
+    description:
+"Tests whether `principal1` matches the current network type, and therefore
+represents a principal that can spend tokens on the current network type. That is, the network is either of
+type `mainnet`, or `testnet`. And, `principal1` is either suited for,
+or able to spend tokens on, the mainnet or a testnet.  `SPxxxx` and
+`SMxxxx` addresses can only spend tokens on mainnet, whereas `STxxxx` and `SNxxxx` addresses can only
+spend tokens on a testnet. Other than these four forms of address, no other forms of addresses are valid.
```
jcnelson, line 253: All c32check address forms are valid in that they can receive tokens, but only these four c32check forms can spend them.

src/vm/functions/principals.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    let address_is_mainnet = version == C32_ADDRESS_VERSION_MAINNET_MULTISIG
+        || version == C32_ADDRESS_VERSION_MAINNET_SINGLESIG;
+    let address_is_testnet = version == C32_ADDRESS_VERSION_TESTNET_MULTISIG
+        || version == C32_ADDRESS_VERSION_TESTNET_SINGLESIG;
+    let context_is_mainnet = env.global_context.mainnet;
+
+    if address_is_mainnet || address_is_testnet {
+        // We can only return true if the address is mainnet or testnet.
+        Ok(Value::Bool(address_is_mainnet == context_is_mainnet))
```
jcnelson, line 46: I had to read this logic a few times to verify that it works.  For legibility sake, can you change this line to the following?

```
Ok(Value::Bool((address_is_mainnet && context_is_mainnet) || (address_is_testnet && !context_is_mainnet)))
```

The compiler should optimize this out, so there's no expected performance hit.

jcnelson (approved):
Just had a few minor comments, but this LGTM.  Thanks @gregorycoppola!

PR #2732, opened 2021-06-22
fix(Clarity2):Factor out 'stx-transfer-memo' into a separate function

### LINKED ISSUE #2615: Extension contract: stx-transfer-memo?
In Stacks 2.1 right now, `stx-transfer?` can take an optional memo field.  Unfortunately, the presence of this change would mean the parser would need to be aware of the epoch rules in order to know which invocations are allowed.

Instead of doing this, I propose refactoring `stx-transfer?` in 2.1 into an extension contract with the public function `stx-transfer-memo?`.  This would do the same things that `stx-transfer` does, but would use the epoch-switching gates that already would exist for all extension contracts (and wouldn't require modifying the analysis logic).

Open to push-back on this.

gregorycoppola:
Fixes #2615

## Motivation
Currently, in the next branch, it is possible to supply a memo to `stx-transfer?`. However, it has been decided that it might be better to force the user to use a *different* function,`stx-transfer-memo?` if they are going to add a memo. The reason for this is to allow better type checking -- the user will either have to add a memo or not, depending on which version of the function they use, and they won't be able to forget to put a memo if they are using the "memo" version.

We also version-guard this function, so that one has to be using Stacks 2.1 in order to use it.

## Testing

Tested using unit tests. The following tests have been modified:

```
M	src/vm/tests/costs.rs
M	src/vm/tests/events.rs
M	src/vm/tests/simple_apply_eval.rs
```


src/vm/analysis/type_checker/tests/mod.rs
```diff
@@ -188,26 +188,24 @@ fn test_stx_ops() {
...
         r#"(stx-transfer? u4 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR 0x7759)"#,
-        r#"(stx-transfer? 4 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR 0x000000)"#,
+        r#"(stx-transfer? 4 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR)"#,
         r#"(stx-transfer? 4 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR true 0x00)"#,
-        r#"(stx-transfer? u4 u3  'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR 0x00)"#,
-        r#"(stx-transfer? u4 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR true 0x00)"#,
-        r#"(stx-transfer? u4 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR true)"#,
-        "(stx-transfer? u10 tx-sponsor? 'SM2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQVX8X0G)",
+        r#"(stx-transfer? u4 u3  'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR)"#,
+        r#"(stx-transfer? u4 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR true)"#,
```
kantai, line 201: Can you add two new tests for:

```
(stx-transfer? u4 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR 'SM2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQVX8X0G 0x000000)
(stx-transfer-memo? u4 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR 'SM2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQVX8X0G)
```

i.e., make sure that the two functions don't accept one another's inputs.

kantai (approved):
LGTM! Just commented about a small test addition.

src/vm/analysis/type_checker/tests/mod.rs
gregorycoppola: done

gregorycoppola (commented):
thanks for review!

src/vm/costs/cost_functions.rs
```diff
@@ -120,6 +120,7 @@ define_named_enum!(ClarityCostFunction {
     BlockInfo("cost_block_info"),
     StxBalance("cost_stx_balance"),
     StxTransfer("cost_stx_transfer"),
+    StxTransferMemo("cost_stx_transfer"),
```
jcnelson, line 123: Should this be `unimplemented` for now?

src/vm/functions/assets.rs
```diff
@@ -147,24 +147,40 @@ pub fn special_stx_transfer(
...
+        Err(CheckErrors::BadTransferSTXArguments.into())
     }
+}
 
+pub fn special_stx_transfer_memo(
+    args: &[SymbolicExpression],
+    env: &mut Environment,
+    context: &LocalContext,
+) -> Result<Value> {
+    check_argument_count(4, args)?;
```
jcnelson, line 177: This has the wrong cost function -- it should be `StxTransferMemo`.

jcnelson (changes requested):
I didn't see any gating for `stx-transfer-memo?`.  I think you'll want to add that in first before merging -- `stx-transfer-memo?` is not valid in Clarity 1.

src/vm/functions/assets.rs
gregorycoppola: replaced with Unimplemented

src/vm/costs/cost_functions.rs
gregorycoppola: this line is gone because I used Unimplemented in src/vm/functions/assets.rs

src/vm/analysis/arithmetic_checker/tests.rs
```diff
@@ -130,95 +130,123 @@ fn test_variables_fail_arithmetic_check() {
 }
 
 #[test]
-fn test_functions() {
-    let bad_tests = [
+fn test_functions_clarity1() {
```
gregorycoppola, line 133: note: I separated these tests into "clarity1" and "clarity2", similar to what was suggested in https://github.com/blockstack/stacks-blockchain/pull/2735

src/vm/functions/mod.rs
```diff
@@ -138,6 +138,7 @@ define_versioned_named_enum!(NativeFunctions(ClarityVersion) {
     BurnAsset("nft-burn?", ClarityVersion::Clarity1),
     GetStxBalance("stx-get-balance", ClarityVersion::Clarity1),
     StxTransfer("stx-transfer?", ClarityVersion::Clarity1),
+    StxTransferMemo("stx-transfer-memo?", ClarityVersion::Clarity2),
```
gregorycoppola, line 141: note: version guarding here

gregorycoppola (commented):
@jcnelson  thanks for the comments.

I rebased against 'next' so now this PR has version guarding. 

Refactored tests in `src/vm/analysis/arithmetic_checker/tests.rs` test the version guarding.

src/vm/analysis/arithmetic_checker/tests.rs
jcnelson: Thanks for doing this -- I think it's a healthy habit to maintain, especially since there will likely be future versions of Clarity.

jcnelson (commented):
LGTM now. Thanks @gregorycoppola!

jcnelson (approved):

gregorycoppola:
Thanks for the reviews everyone.

PR #2735, opened 2021-06-22
feat(Clarity2): Version-guard variables, starting with 'tx-sponsor?'

### LINKED ISSUE #2614: Extension contract: tx-sponsor
Right now, Stacks 2.1 has a `tx-sponsor` keyword that resolves to the `(optional principal)` that represents the Stacks transaction sponsor.  Trouble is, this can't be a keyword -- contracts today can use this variable name today, and in doing so, make it impossible to represent this functionality as a keyword without breaking these existing contracts.  The `tx-sponsor` keyword will need to be refactored into an extension contract.

The extension contract can use the same functionality in the VM; it just needs to be handled via a `contract-call?` interception instead of a built-in variable.

This is partially blocked on #2607

gregorycoppola:
Fixes #2614 

## Background
Right now, Stacks 2.1 has a tx-sponsor keyword that resolves to the (optional principal) that represents the Stacks transaction sponsor. Trouble is, this can't be a keyword -- contracts today can use this variable name today, and in doing so, make it impossible to represent this functionality as a keyword without breaking these existing contracts.

## Change

We introduce version-guarding for variables similar to what pull/2659 does for versioning of function names.

## Testing
So far, the only new tests are in 
`src/vm/analysis/arithmetic_checker/tests.rs`


src/vm/contexts.rs
```diff
@@ -223,7 +223,7 @@ pub struct ContractContext {
     pub meta_ft: HashMap<ClarityName, FungibleTokenMetadata>,
     pub data_size: u64,
     /// track the clarity version of the contract
-    clarity_version: ClarityVersion,
+    pub clarity_version: ClarityVersion,
```
gregorycoppola, line 226: Note: Making this public! What do you think @kantai ?

gregorycoppola (commented):
Leaving one note.

src/vm/contexts.rs
kantai: I'd prefer if consumers just used `.get_clarity_version()` -- this clarity_version instance variable should be immutable, and making it private helps ensure that's the case.

kantai (approved):
LGTM -- only comment is about the visibility of the ContractContext instance variable.

src/vm/contexts.rs
gregorycoppola: oh.. :/... I guess this wasn't even needed in this PR. Must have been copied from another PR. However, in future PR's I'll de-pub this line and use `get_clarity_version`.

src/vm/variables.rs
```diff
@@ -19,20 +19,41 @@ use vm::contexts::{Environment, LocalContext};
...
+    ContractCaller("contract-caller", ClarityVersion::Clarity1),
+    TxSender("tx-sender", ClarityVersion::Clarity1),
+    BlockHeight("block-height", ClarityVersion::Clarity1),
+    BurnBlockHeight("burn-block-height", ClarityVersion::Clarity1),
+    NativeNone("none", ClarityVersion::Clarity1),
+    NativeTrue("true", ClarityVersion::Clarity1),
+    NativeFalse("false", ClarityVersion::Clarity1),
+    TotalLiquidMicroSTX("stx-liquid-supply", ClarityVersion::Clarity1),
+    Regtest("is-in-regtest", ClarityVersion::Clarity1),
+    TxSponsor("tx-sponsor?", ClarityVersion::Clarity2),
```
jcnelson, line 37: So, we don't have a coding convention anywhere, but this is the first time I'm seeing `?` getting used in a variable name.  Is this a convention we want?

src/vm/analysis/arithmetic_checker/tests.rs
```diff
@@ -129,6 +129,22 @@ fn test_variables_fail_arithmetic_check() {
...
 
+#[test]
+fn test_version_controlled_variables() {
+    // Certain variables are controlled by the ClarityVersion. E.g., in Clarity1, 'tx-sponsor?'
+    // is not a variable, so can be defined over.
+    let ok_tests = ["(define-private (foo) tx-sponsor?)"];
+
+    for contract in ok_tests.iter() {
+        assert_eq!(
+            arithmetic_check(contract, ClarityVersion::Clarity1),
```
jcnelson, line 140: Wait, why does this pass?  `tx-sponsor?` only exists as a built-in in `Clarity2`, but this is running in `Clarity1`.

src/vm/analysis/arithmetic_checker/tests.rs
```diff
@@ -112,7 +112,7 @@ fn test_variables_fail_arithmetic_check() {
 
     for (contract, error) in tests.iter() {
         assert_eq!(
-            arithmetic_check(contract),
+            arithmetic_check(contract, ClarityVersion::Clarity2),
```
jcnelson, line 115: Can you actually test both Clarity versions here?  I don't want to lose coverage for `Clarity1`.

src/vm/analysis/arithmetic_checker/tests.rs
```diff
@@ -69,7 +69,7 @@ fn test_bad_defines() {
 
     for (contract, error) in tests.iter() {
         assert_eq!(
-            arithmetic_check(contract),
+            arithmetic_check(contract, ClarityVersion::Clarity1),
```
jcnelson, line 72: Can you actually test both Clarity versions here? I don't want to lose coverage for `Clarity2`.

src/vm/ast/definition_sorter/mod.rs
```diff
@@ -213,12 +213,11 @@ impl<'a> DefinitionSorter {
...
                             // NOTE: can use ClarityVersion::latest() here only as long as NO NEW FUNCTIONS are special cased
-                            //        in the definition sorter.
+                        //        in the definition sorter.
                             NativeFunctions::lookup_by_name_at_version(
-                                    function_name,
-                                    &ClarityVersion::latest(),
-                                )
-                        {
+                            function_name,
+                            &ClarityVersion::latest(),
```
jcnelson, line 219: Hold up -- this needs to be gated. You should be using whatever `ClarityVersion` the outermost runtime context is using.

gregorycoppola: Hmm.. this change was not supposed to be part of this PR. It snuck in as because of some weird behavior of "cargo fmt". @kantai ?

src/vm/variables.rs
gregorycoppola: I think it means that the return value is "(option X)"?

https://github.com/blockstack/stacks-blockchain/blob/2c424d9142a1b8a8ad53ceca6c4c96a33bc201a1/src/vm/variables.rs#L59

src/vm/analysis/arithmetic_checker/tests.rs
gregorycoppola: I made clarity1 and clarity2 versions of this. In general this is a good idea, but I'm not sure it's that needed in this particular instance, but can't really hurt.

src/vm/analysis/arithmetic_checker/tests.rs
gregorycoppola: Good idea.. I added tests for both versions.

src/vm/analysis/arithmetic_checker/tests.rs
gregorycoppola: I dropped this test method in the refactor. However, the answer is that the contract "passes" because "tx-sponsor?" is NOT reserved in Clarity1, so you can define something over it.

gregorycoppola (commented):
Thanks for the comments.

src/vm/ast/definition_sorter/mod.rs
kantai: Right -- so the `definition_sorter` is a bit of a special case when it comes to the analysis -- it doesn't actually do any special checks for native functions vs. defined functions except for a few special cases. As long as no new special cases are added in Clarity2, it's fine to just use `::latest` here. It may be safer in the long run to gate this, but I don't feel strongly that it needs to be done in this PR.

gregorycoppola: It probably shouldn't be in this PR, but should we make an issue for it and make sure to do it?

kantai: Issue created #2759

src/vm/analysis/arithmetic_checker/tests.rs
```diff
@@ -78,42 +78,62 @@ fn test_bad_defines() {
 }
 
 #[test]
-fn test_variables_fail_arithmetic_check() {
+fn test_bad_defines_clarity2() {
```
pavitthrap, line 81: Not sure if it makes sense to add a "bad defines" test specifically for clarity2, considering we aren't planning on adding any define statements to Clarity2. 

Instead, maybe just wrap the for loop on line 70 (`for (contract, error) in tests.iter() {`), in a for loop that iterates through clarity versions 1 and 2

pavitthrap (approved):
LGTM, just had a comment on testing.

gregorycoppola (commented):
Thanks for the reviews!

PR #2742, opened 2021-06-28
Chore: reset testnet

lgalabru:
This PR, when merged and ran, we will trigger a Testnet reset by:
- rotating the genesis account keys,
- updating the POX constants,
- updating the magic bytes (from X6 to T2),
- anchoring to a more recent bitcoin testnet height (to accelerate the boot sequence).

This PR is also updating the subcommand used for launching a testnet node (from `stacks-node xenon` to `stacks-node testnet`)


src/core/mod.rs
```diff
@@ -81,10 +81,10 @@ pub const BITCOIN_MAINNET_FIRST_BLOCK_HASH: &str =
     "0000000000000000000ab248c8e35c574514d052a83dbc12669e19bc43df486e";
 pub const BITCOIN_MAINNET_INITIAL_REWARD_START_BLOCK: u64 = 651389;
 
-pub const BITCOIN_TESTNET_FIRST_BLOCK_HEIGHT: u64 = 1931620;
-pub const BITCOIN_TESTNET_FIRST_BLOCK_TIMESTAMP: u32 = 1612282029;
+pub const BITCOIN_TESTNET_FIRST_BLOCK_HEIGHT: u64 = 2000000;
```
jcnelson, line 84: This has already passed

gregorycoppola:
@lgalabru I don't have too much too add in terms of comments but will follow along to add this to the release.

kantai (approved):
LGTM!

src/core/mod.rs
lgalabru: We have to use a block from the past, I kind of liked the simple and clean `2000000` value (from just 2-3 weeks ago).

src/chainstate/stacks/boot/contract_tests.rs
```diff
@@ -341,6 +341,7 @@ fn recency_tests() {
 fn delegation_tests() {
     let mut sim = ClarityTestSim::new();
     let delegator = StacksPrivateKey::new();
+    const REWARD_CYCLE_LENGTH: u128 = 1050;
```
pavitthrap, line 344: Want to calculate this with `POX_REWARD_CYCLE_LENGTH / 2` instead of using 1050?

src/chainstate/stacks/boot/contract_tests.rs
```diff
@@ -512,7 +513,7 @@ fn delegation_tests() {
                 "(ok {{ stacker: '{}, lock-amount: {}, unlock-burn-height: {} }})",
                 Value::from(&USER_KEYS[0]),
                 Value::UInt(*MIN_THRESHOLD - 1),
-                Value::UInt(450)
+                Value::UInt(3150)
```
pavitthrap, line 516: possibly calculate 3150 instead?

src/chainstate/stacks/boot/contract_tests.rs
```diff
@@ -70,7 +70,7 @@ lazy_static! {
     )
     .unwrap();
     static ref LIQUID_SUPPLY: u128 = USTX_PER_HOLDER * (POX_ADDRS.len() as u128);
-    static ref MIN_THRESHOLD: u128 = *LIQUID_SUPPLY / 480;
+    static ref MIN_THRESHOLD: u128 = *LIQUID_SUPPLY / 8000;
```
pavitthrap, line 73: Could use const `TESTNET_STACKING_THRESHOLD_25` instead of 8000

testnet/stacks-node/src/main.rs
```diff
@@ -100,19 +100,7 @@ fn main() {
...
-        "argon" => {
-            args.finish().unwrap();
-            ConfigFile::argon()
-        }
-        "krypton" => {
-            args.finish().unwrap();
-            ConfigFile::krypton()
-        }
-        "xenon" => {
+        "testnet" => {
```
pavitthrap, line 103: Can delete these functions in `testnet/stacks-node/src/config.rs` ( neon, argon, krypton)

pavitthrap (approved):
Just left some minor comments, but lgtm!!

src/chainstate/stacks/boot/contract_tests.rs
lgalabru: Address with latest

testnet/stacks-node/src/main.rs
lgalabru: Good idea, addressed with latest

src/chainstate/stacks/boot/contract_tests.rs
lgalabru: In this case, I don't think that it would improve readability

src/chainstate/stacks/boot/contract_tests.rs
lgalabru: If we are fine hard coding this in our contracts, then hard code once in our tests sounds acceptable :)

PR #2744, opened 2021-06-28
build: add libvm workspace member

reedrosenbluth:
Expose `vm` as separate library, so it can be directly depended on by external projects (such as Clarinet)


src/libvm.rs
```diff
@@ -0,0 +1,127 @@
...
+const GIT_TREE_CLEAN: Option<&'static str> = option_env!("GIT_TREE_CLEAN");
+
+#[cfg(debug_assertions)]
+const BUILD_TYPE: &'static str = "debug";
+#[cfg(not(debug_assertions))]
+const BUILD_TYPE: &'static str = "release";
+
+pub fn version_string(pkg_name: &str, pkg_version: &str) -> String {
+    let git_branch = GIT_BRANCH
+        .map(|x| format!("{}", x))
```
pavitthrap, line 111: Doesn't seem like this map is doing anything - can this line be removed?

pavitthrap (approved):
LGTM!

Cargo.lock
```diff
@@ -1,5 +1,7 @@
 # This file is automatically @generated by Cargo.
```
gregorycoppola, line 1: Not about this line, but should this be merged into `next` instead?

Cargo.toml
```diff
@@ -110,7 +110,8 @@ sha2-asm = "0.5.3"
 
 [workspace]
 members = [
-    ".", 
+    ".",
+    "libvm",
```
gregorycoppola, line 114: is anything being removed from this file or from anywhere? How do we know it's not the old code paths that are producing the behavior.

gregorycoppola: Moreover.. what is the observed behavior? Is the idea that you ran all the old tests and they pass? Can you add information about what was tested to the commit message for posterity?

gregorycoppola (commented):
Thanks, Reed.. just had a few high level questions.

Cargo.lock
kantai: I don't think so -- there's nothing consensus breaking about this PR, so it doesn't need to wait until Stacks 2.1.

src/libvm.rs
kantai: Yeah -- this could probably be rewritten as `git_branch = GIT_BRANCH.unwrap_or("");`

This whole file was just copied from `src/lib.rs`.

lgalabru (commented):
Looks good to me, thank you @reedrosenbluth. Just one feedback on the name!

Cargo.toml
reedrosenbluth: @lgalabru I think those are potentially confusing since we named the `stacks` specific vm module `clarity_vm`, and the more general vm module remained `vm`. Perhaps we should rename as follows:

`clarity_vm` -> `stacks_vm`
`libvm` ->  `clarity-vm-lib`

thoughts?

lgalabru: Sounds good to me!

gregorycoppola: why is it "-lib" though? most crates are "libs" but they don't have "-lib" at the end

Cargo.toml
```diff
@@ -110,7 +110,8 @@ sha2-asm = "0.5.3"
 
 [workspace]
 members = [
-    ".", 
+    ".",
+    "clarity",
```
gregorycoppola, line 114: Not about this line but it looks like "rust fmt" test is failing.

gregorycoppola, line 114: Not about this line but it looksl ike "unit tests" is failing.. Do you know why that would be?

```
failures:

---- clarity::test::test_assets stdout ----
initialize
INFO [1629087718.866954] [clarity/../src/vm/functions/mod.rs:445] [clarity::test::test_assets] "``... to be a completely separate network and separate block chain, yet share CPU power with Bitcoin`` - Satoshi Nakamoto"
check tokens
Error reading file: sample-contracts/tokens-ft.clar
Caused by: No such file or directory (os error 2)
thread 'clarity::test::test_assets' panicked at 'explicit panic', clarity/../src/clarity.rs:122:9

---- clarity::test::test_samples stdout ----
initialize
INFO [1629087720.500553] [clarity/../src/vm/functions/mod.rs:445] [clarity::test::test_samples] "``... to be a completely separate network and separate block chain, yet share CPU power with Bitcoin`` - Satoshi Nakamoto"
check tokens
Error reading file: sample-contracts/tokens.clar
Caused by: No such file or directory (os error 2)
thread 'clarity::test::test_samples' panicked at 'explicit panic', clarity/../src/clarity.rs:122:9

---- vm::docs::test::test_examples stdout ----
thread 'vm::docs::test::test_examples' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', clarity/../src/vm/docs/mod.rs:1967:77


failures:
    clarity::test::test_assets
    clarity::test::test_samples
    vm::docs::test::test_examples

test result: FAILED. 1218 passed; 3 failed; 94 ignored; 0 measured; 0 filtered out; finished in 922.25s

error: test failed, to rerun pass '-p clarity --lib'

Exited with code exit status 101

CircleCI received exit code 101

```

gregorycoppola (commented):
Hey Reed.. few technical things.

Cargo.toml
reedrosenbluth: looking into this...

circle.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - run:
           no_output_timeout: 200m
           command: |
-            cargo test -j 1 --workspace
+            cargo test -j 1 --workspace --exclude clarity-lib
```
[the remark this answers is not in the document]
reedrosenbluth: this should just be `clarity`. changing...

lgalabru (approved):
LGTM, thanks Reed!

PR #2745, opened 2021-06-28
fix(Clarity2): Fix "principal-of" to be sensitive to mainnet/textnet

### LINKED ISSUE #2619: Epoch switch: fix and gate behavior of `principal-of`
`principal-of` is broken on mainnet.  This needs to be fixed in 2.1, but the fixed logic must only activate when the epoch switches.

gregorycoppola:
Fixes #2619

## Description
Currently, "principal-of" has a bug, in that it always creates a testnet address.

For Clarity2, we want to be sensitive to whether the network is mainnet vs. testnet.

But, we want to gate this behavior on the ClarityVersion, so that the behavior does not change for Clarity1.

## Testing
New tests in
```
src/vm/tests/simple_apply_eval.rs
```


kantai (commented):
These changes look good to me, the only question I have is whether or not it makes more sense to change the current function or add a new one?

I think changing the current function makes sense -- it only changes in v2 contracts. However, in this case, I think you need to update the function's documentation with a description of the differences.

gregorycoppola:
> These changes look good to me, the only question I have is whether or not it makes more sense to change the current function or add a new one?
> 
> I think changing the current function makes sense -- it only changes in v2 contracts. However, in this case, I think you need to update the function's documentation with a description of the differences.

I agree it makes more sense to just fix the behavior rather than add a new function. Otherwise we'll just have a buggy function around forever. So, I update the docs to discuss this.

jcnelson (approved):
LGTM! Thanks for fixing this @gregorycoppola!

src/vm/docs/mod.rs
```diff
@@ -1029,6 +1029,12 @@ const PRINCIPAL_OF_API: SpecialAPI = SpecialAPI {
     signature: "(principal-of? public-key)",
     description: "The `principal-of?` function returns the principal derived from the provided public key.
     If the `public-key` is invalid, it will return the error code `(err u1).`.
+
+Note: Before Stacks 2.1, this function has a bug, in that the principal returned would always
+be a testnet single-signature principal, even if the function were run on the mainnet. Starting
+with Stacks 2.1, this bug is fixed, so that if this function will return a principal suited to
```
kantai, line 1035: "so that if this function" -> "so that this function"

kantai (approved):
LGTM! Just a minor typo fix.

hstove:
Question about this - I'm not clear about how this will effect contracts deployed before the epoch switch. If I deploy a contract today with `principal-of` on mainnet, will the result change after the switch? Or no?

If not, could I deploy a contract today with the 2.1 pragma, and then the behavior _will_ change after the switch?

Basically I'm trying to determine a way to deploy contracts today that can take advantage of this after the switch. Depending on the answers to the above, then the contracts might need to be "upgradeable".

Issue #2738, opened 2021-06-25
server: cache /v2/info responses


jcnelson:
These are surprisingly slow -- on the order of 10s of milliseconds -- but they don't change very often.  Let's cache some of this data.

Issue #2730, opened 2021-06-22
p2p: in ibd, only require the downloader to check the last few reward cycles before advancing to the next reward cycle


jcnelson:
Right now, the PoX sync watchdog requires the node to do a full pass over the block availability data in the downloader state machine.  Instead, it should only require a pass over the last O(1) reward cycles.

Issue #2728, opened 2021-06-22
p2p: don't do a full inv scan on PoX divergence


jcnelson:
It is instead sufficient to do an inv-sync from the reward cycle just prior to the divergence.  A full one can take a long, long time.

Issue #2725, opened 2021-06-21
Slow/failed syncing with testnet


wileyj:
Running a follower node for testnet was failing to sync stacks blocks as expected. 
Initially, the burnchain sync went through fine - but then it took roughly 48 hours of continued operation to sync a single stacks block (the only event that happened on my side was a disconnection due to laptop lid being closed for a bit - reopened the lid, service became active again and blocks started to download). 

## Steps To Reproduce
Run a testnet node with the following config.toml:
```toml
[node]
working_dir = "/root/stacks-node/data"
rpc_bind = "0.0.0.0:20443"
p2p_bind = "0.0.0.0:20444"
bootstrap_node="047435c194e9b01b3d7f7a2802d6684a3af68d05bbf4ec8f17021980d777691f1d51651f7f1d566532c804da506c117bbf79ad62eea81213ba58f8808b4d9504ad@testnet.stacks.co:20444"
wait_time_for_microblocks = 10000

[[events_observer]]
endpoint = "stacks-node-api:3700"
retry_count = 255
events_keys = ["*"]

[burnchain]
chain = "bitcoin"
mode = "xenon"
peer_host = "bitcoin.testnet.blockstack.com"
username = "blockstack"
password = "blockstacksystem"
rpc_port = 18332
peer_port = 18333

[[ustx_balance]]
address = "STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6"
amount = 10000000000000000

[[ustx_balance]]
address = "ST11NJTTKGVT6D1HY4NJRVQWMQM7TVAR091EJ8P2Y"
amount = 10000000000000000

[[ustx_balance]]
address = "ST1HB1T8WRNBYB0Y3T7WXZS38NKKPTBR3EG9EPJKR"
amount = 10000000000000000

[[ustx_balance]]
address = "STRYYQQ9M8KAF4NS7WNZQYY59X93XEKR31JP64CP"
amount = 10000000000000000

[connection_options]
read_only_call_limit_write_length = 0
read_only_call_limit_read_length = 100000
read_only_call_limit_write_count = 0
read_only_call_limit_read_count = 30
read_only_call_limit_runtime = 1000000000
```

## Expected behavior
Node will sync stacks blocks in a timely manner

## Environment
 - OS: Docker (`Docker version 20.10.6, build 370c289`)
 - Image: `blockstack/stacks-blockchain:latest` (same as tagged image `2.0.11.1.0`)
[...]

PR #2746, opened 2021-06-29
Fix/2725

### LINKED ISSUE #2701: Network dispatch doesn't handle PoX invalidations
The `PeerNetwork::run` method can error in the event of PoX invalidations. This mostly only happens on testnet (where PoX cycles are _much_ more frequent, so nodes do occasionally perform a PoX "reorg"), but it's possible on mainnet too. Usually this isn't a problem, if `PeerNetwork::run` fails, the loop just keeps running. However, when the node is "event driven" (i.e., the event dispatcher is connected to an event consumer) like in the case of an API node, `PeerNetwork::run` errors are panicking. That's because the event dispatcher cannot recover from an arbitrary PeerNetwork error.

The code path that causes issues in testnet followers is `get_block_availability`:

```rust
        // what blocks do we have in this range?
        let local_blocks = {
            let ic = sortdb.index_conn();
            let tip = SortitionDB::get_canonical_burn_chain_tip(&ic)?;

...

            let local_blocks = ic.get_stacks_header_hashes(
                sortition_height_end - sortition_height_start,
                &last_ancestor.consensus_hash,
                header_cache,
            )?;
```

The bug witnessed is as follows:

1. `get_canonical_burn_chain_tip` is called and returns the then-canonical sortition tip.
2. the chains coordinator detects a PoX reorg and invalidates the sortition tip
3. `get_stacks_header_hashes` is called and returns an `Err` because the supplied tip is marked `pox_valid = 0`.

The invalidations are performed in a sqlite transaction, so they are atomic. However, 1 and 3 are not, these operations are performed on just an `index_conn`, not a sqlite transation. So perhaps the solution is to use a transaction here.

jcnelson:
This PR fixes a lot of issues underpinning #2725.  Namely, it fixes #2701, #2728, #2730, and #2738 as well.

The gist of the changes is that we make it so that when the block inventory state-machine determines that the node is in the initial block download, and one of its bootstrap nodes NACK's its request for a `PoxInv` message or a `BlocksInv` message at a given reward cycle, it will _only_ refresh its view of the last `INV_REWARD_CYCLES` reward cycles from its bootstrap peers instead of _all_ prior reward cycles.  This situation arises more often than not -- for example, whenever the anchor block is "late" relative to the PoX sync watchdog -- which leads to a significant slow-down in synchronization as the node gets closer to the testnet chain tip (which has thousands of reward cycles).  In addition, the changes here make it so that if this happens, the download state machine will begin scanning the local chain state from the sortition height at the start of the highest reward cycle synchronized with the bootstrap peer (instead of the very first sortition).  This significantly speeds up block downloads, because like the inventory state machine, the download state machine only processes one reward cycle at a time (and making it start right where it's bound to find new blocks makes synchronization go faster towards the chain tip).

The PR also fixes a few odds and ends, like making it so `/v2/info` doesn't require any I/O to complete and making it so the inventory state machine doesn't have to load the PoX bitvector on each pass (which is surprisingly slow).

I'm in the process of testing this on a local testnet follower; will report back when it's ready.


gregorycoppola:
@jcnelson Are there any tests that we should add as part of the release process in order to test this change?

jcnelson:
Yes -- per the blockchain engineering meeting, it's imperative that we spin up a release candidate from genesis on both mainnet and testnet and verify that it reaches the chain tip.

jcnelson:
@wileyj Want to take this branch for a spin?

src/blockstack_cli.rs
```diff
@@ -18,6 +18,7 @@
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 #![allow(non_upper_case_globals)]
+#![allow(non_fmt_panic)]
```
kantai, line 21: Didn't we fix all these warnings in `master`? If so, we should just merge `master` -> `develop` and leave this warning as is. Technically, this warning is telling us that we have a spurious string construction.

src/net/mod.rs
```diff
@@ -303,6 +305,7 @@ impl fmt::Display for Error {
             Error::StaleView => write!(f, "State view is stale"),
             Error::ConnectionCycle => write!(f, "Tried to connect to myself"),
             Error::NotFoundError => write!(f, "Requested data not found"),
+            Error::Transient(ref s) => fmt::Display::fmt(s, f),
```
kantai, line 308: I think `write!(f, "Transient: {}", s)` would be more descriptive here

testnet/stacks-node/src/neon_node.rs
```diff
@@ -735,9 +735,14 @@ fn spawn_peer(
                         }
                     }
                     Err(e) => {
-                        error!("P2P: Failed to process network dispatch: {:?}", &e);
-                        if config.is_node_event_driven() {
-                            panic!();
+                        if let NetError::Transient(ref msg) = &e {
+                            error!("P2P: Failed to process network dispatch due to transient error: {}", &msg);
+                        }
```
kantai, line 740: Are you sure that these transient errors won't break the event dispatcher's guarantees? The reason that this panics in `event_driven` mode is if an error would cause a given block or block's events to be sent multiple times to the same observer. That would break the event dispatcher's interface contract with observers.

wileyj:
> @wileyj Want to take this branch for a spin?

I can do that, sure!

kantai (commented):
This LGTM once the comments are resolved

wileyj:
> @wileyj Want to take this branch for a spin?

spinning up this morning, will check back in a bit

jcnelson:
I tested it overnight.  It's still getting stuck :(

wileyj:
> I tested it overnight. It's still getting stuck :(

working on my end so far, continuing to monitor though:
```
{
  "peer_version": 4207599105,
  "pox_consensus": "52310389ee679fd45852a559237e653215ba8cd3",
  "burn_block_height": 1936970,
  "stable_pox_consensus": "ce9e716823b659be1ed136168afb9b84e249ea55",
  "stable_burn_block_height": 1936963,
  "server_version": "stacks-node 473da85 (fix/2725:473da85+, release build, linux [x86_64])",
  "network_id": 2147483648,
  "parent_network_id": 118034699,
  "stacks_tip_height": 2620,
  "stacks_tip": "c6d15a0b3401c0e56ad40b355979f00c70500d8c82bc9f98e783e9b781e6db7a",
  "stacks_tip_consensus_hash": "52310389ee679fd45852a559237e653215ba8cd3",
  "genesis_chainstate_hash": "74237aa39aa50a83de11a4f53e9d3bb7d43461d1de9873f402e5453ae60bc59b",
  "unanchored_tip": "5476f7fe3fa6294bf3be2f402654bd23653e8b896c53d8811ec842bddb72c6a0",
  "unanchored_seq": 0,
  "exit_at_block_height": null
}
```

src/blockstack_cli.rs
gregorycoppola: FYI: I checked my "fix rust warnings" PR into `develop`, but i still get the warnings in next `branches`.

https://github.com/blockstack/stacks-blockchain/pull/2702

wileyj:
up to block `3557` currently, seems fine

testnet/stacks-node/src/neon_node.rs
jcnelson: > if an error would cause a given block or block's events to be sent multiple times to the same observer

Wait, is this part of the API contract?  If so, this sounds incredibly brittle, especially since an event observer can easily avoid re-processing a block's events if it's already seen the block.  Moreover, this API contract can't be guaranteed in the first place, since the same Stacks block can be processed on multiple PoX forks under normal circumstances.  Namely, if there are two identical Stacks blocks in two different PoX forks that share a parent block, then both Stacks blocks can be processed in their respective PoX forks since the state root calculation only depends on the _parent_ being in the same PoX fork as the miner when the miner calculated the block header.  If the event observer can handle this case already -- i.e. by using the block's consensus hash or index hash to distinguish two otherwise-identical blocks -- then it can easily handle the case where a block's receipts get sent more than once.

My impression was that the API contract only requires that the node not _drop_ block receits, since the observer has no way to poll for them (the node doesn't store them nor replay them).

kantai: This guarantee is provided for `(consensus_hash, block_hash)` pairs. I'm not sure if that guarantee is strictly necessary, but this panic would protect the event observer from "dropped" data as well -- but I think we should get some insight from @lgalabru as well -- looking at the git history for this line, I think this was added for Atlas attachments. Could an error from the network.run command cause an attachment event to be missed and never reprocessed?

wileyj:
@jcnelson ready to second spin-up test with the commits above?

jcnelson:
Not yet -- still getting stuck (investigating)

jcnelson:
Okay, this time it managed to reach the testnet chain tip.   Going to spin up from testnet genesis again to verify that it can do so without getting stuck even once.

wileyj:
> Okay, this time it managed to reach the testnet chain tip. Going to spin up from testnet genesis again to verify that it can do so without getting stuck even once.

👍  spinning up our node from genesis now

gregorycoppola:
> Okay, this time it managed to reach the testnet chain tip. Going to spin up from testnet genesis again to verify that it can do so without getting stuck even once.

Nice work, @jcnelson . Thanks :).

jcnelson:
Okay, my node reached the chain tip on testnet and is in steady state.

src/blockstack_cli.rs
kantai: Yep -- warnings were still present in tests and other workspace members, PR https://github.com/blockstack/stacks-blockchain/pull/2758 addresses it. We should not disable this warning.

jcnelson:
Both my testnet and mainnet nodes sync'ed with this patch, without issue.

kantai (commented):
This looks good to me, but the allowed warning should be re-enabled.

jcnelson:
> This looks good to me, but the allowed warning should be re-enabled.

As in 2a1e8f5?

testnet/stacks-node/src/neon_node.rs
jcnelson: I've made it so that `network.run()` does not fail from `net::Error::Transient(..)` -- all the call sites where this error could be emitted simply absorb it and turn it into an `Ok(..)`.

src/net/mod.rs
jcnelson: Done

src/blockstack_cli.rs
jcnelson: Reverted

kantai (approved):

gregorycoppola (commented):
Hey Jude

I was talking a look at this PR to review.

Even though I lack a lot of context, I was planning to just review based on the testing.

I was surprised to see that no tests were affected, or added.

Is it possible to add any unit tests that would exercise some of the new behavior? Perhaps this could be in a future PR.

Otherwise, can we document in the PR description how this was tested and why we think that is sufficient?

jcnelson:
In terms of lines changed, the bulk of the delta in this PR comes from refactoring.  Namely, we now wrap access to the HTTP network state with the new `PeerNetwork::with_http()` method, and we avoid reloading PoX and burnchain state in the RPC handler and the inventory state machine if we don't need to.  The existing test battery already covers both areas of the codebase.

The truly new functionality -- which is simply concerned with making sure the block downloader will start downloading blocks at the right reward cycle -- is really only testable against a live system.  You'd need a live Bitcoin node and a Stacks node that not only has all the blocks, but also will occasionally disconnect from bootstrapping nodes and render them in a state where they have no choice but to process reward cycles for which they don't have the anchor block.  You'd also need to instrument the Stacks node so that it can be stopped and restarted while it is bootstrapping (clearing its inventory state), in order to verify that the node does _not_ spend an unreasonable amount of time bootstrapping.  Implementing such an integration test that faithfully simulates this condition would not only take more work than this PR, but also would be redundant -- this environment is _exactly_ what the live testnet is supposed to provide.  The release procedure explicitly calls for booting up a new node from genesis on both testnet and mainnet in order to give PRs like this their due testing.

gregorycoppola (approved):
@jcnelson If you think this has had the full test-suite run on it then LGTM.

jcnelson:
The tests are failing is because bitcoin.org is down right now.  However, it previously passed with the current patches with the exception of only `size_overflow_unconfirmed_stream_microblocks_integration_test`, which passed locally for me.

PR #2726, opened 2021-06-21
Docs: adding release process to README.md

kantai:
This adds a description of the release process to README.md for non-consensus-breaking releases.

Tagging @gregorycoppola, @jcnelson, @pavitthrap, @lgalabru, @reedrosenbluth, @CharlieC3, @wileyj for input -- if you'd like to make changes, please feel free to push commits to this branch!

PR #2711, opened 2021-06-14
fix: logic bug in mocknet/helium miner #2710

kantai:
This fixes #2710 -- the mocknet/helium miner had a logic bug that would lead to a panic if a burn block occurred without a sortition in it.

PR #2684, opened 2021-06-01
Feat/clarity cli serialized output

hstove:
This is some of the same functionality from #2587, but now built on top of #2597, which includes the new JSON output of `clarity-cli`.

The output now includes `output_serialized` where applicable. This allows for tighter integration with other packages that can parse clarity values. I've also included a `success` boolean.

PR #2750, opened 2021-06-29
Release 2.0.11.2.0

gregorycoppola:
## [2.0.11.2.0]

NOTE: This change resets the `testnet`. Users running a testnet node will need
to reset their chain states.

### Added

- `clarity-cli` will now also print a serialized version of the resulting
  output from `eval` and `execute` commands. This serialization is in
  hexademical string format and supports integration with other tools. (#2684)
- The creation of a Bitcoin wallet with BTC version `> 0.19` is now supported
  on a private testnet. (#2647)
- `lcov`-compatible coverage reporting has been added to `clarity-cli` for
  Clarity contract testing. (#2592)
- The `README.md` file has new documentation about the release process. (#2726)

### Changed

- This change resets the testnet. (#2742)
- Caching has been added to speed up `/v2/info` responses. (#2746)

### Fixed

- PoX syncing will only look back to the reward cycle prior to divergence,
  instead of looking back over all history. This will speed up running a
  follower node. (#2746)
- The UTXO staleness check is re-ordered so that it occurs before the RBF-limit
  check. This way, if stale UTXOs reached the "RBF limit" a miner will recover
  by resetting the UTXO cache. (#2694)
- Microblock events were being sent to the event observer when microblock data
  was received by a peer, but were not emitted if the node mined the
  microblocks itself. This made something like the private-testnet setup
  incapable of emitting microblock events. Microblock events are now sent
  even when self-mined. (#2653)
- A bug is fixed in the mocknet/helium miner that would lead to a panic if a
  burn block occurred without a sortition in it. (#2711)
- Two bugs that caused problems syncing with the bitcoin chain during a
  bitcoin reorg have been fixed (#2771, #2780).
- Documentation is fixed in cases where string and buffer types are allowed
  but not covered in the documentation.  (#2676)


jcnelson:
Spinning up a public mainnet node right now with this branch

jcnelson:
The Foundation's public mainnet follower has reached the chain tip successfully

kantai (approved):
Ship it! 🚀

lgalabru (approved):
Ship it 🚀

Issue #2717, opened 2021-06-17
[costs] Represent cost functions for Clarity 2 functions


gregorycoppola:
**Is your feature request related to a problem? Please describe.**
We don't have a way to represent costs for Clarity 2 functions. This means that existing cost functions have to be arbitrarily reused.

**Additional context**
I think the reason this is non-trivial is because it would be a consensus breaking change?

https://github.com/blockstack/stacks-blockchain/pull/2691#discussion_r653091807

kantai:
Thanks for opening @gregorycoppola -- yeah, I think this is just a little subtle to implement without altering the consensus before the epoch switch.

I think that in order to implement this, we would need to:

1. Add a new default cost contract instantiation at the "epoch-switch" boundary.
2. Change the CostTracker implementation to use that default cost contract for Clarity2 functions.
3. Change the CostTracker implementation so that it only matches new cost functions based on the active Clarity version (not the contract version)

jcnelson:
Closing in favor of #3115.  Most of this work is done already in 2.05.

PR #2751, opened 2021-06-29
Clarity 2: Add `Unimplemented` cost variant

kantai:
This PR adds an `Unimplemented` cost variant to the Clarity CostFunctions, which is useful while #2717 is still open.


jcnelson (approved):
LGTM! Thanks @kantai

reedrosenbluth (approved):
LGTM!

PR #2753, opened 2021-06-29
feat(Clarity2): "is-in-mainnet", version-guarded variable that says if running in mainnet

### LINKED ISSUE #2620: Converting and querying valid `principal` representations
There needs to be an extension contract that parses a standard `principal` into its `(buff 20)` and `version` representation, and creates a `principal` from a `version` and `(buff 20)`.  In addition, there should be a way to query the `version` values for mainnet and testnet, and determine whether or not the chain is running on mainnet or testnet.

gregorycoppola:
Fixes #2620

## Description
We add a new variable, `is-in-mainnet`, which is true iff the contract is running on the mainnet.

## Testing
Added tests in:
```
M	src/vm/tests/simple_apply_eval.rs
```


pavitthrap (approved):
lgtm! 

quick question though: what was the reasoning behind adding the regtest keyword?

src/vm/analysis/arithmetic_checker/tests.rs
```diff
@@ -133,7 +137,10 @@ fn test_variables_fail_arithmetic_check() {
 fn test_version_controlled_variables() {
     // Certain variables are controlled by the ClarityVersion. E.g., in Clarity1, 'tx-sponsor?'
     // is not a variable, so can be defined over.
-    let ok_tests = ["(define-private (foo) tx-sponsor?)"];
+    let ok_tests = [
+        "(define-private (foo) tx-sponsor?)",
+        "(define-private (foo) is-in-mainnet)",
```
jcnelson, line 142: Same question here as with `tx-sponsor?` in your other PR -- why does this pass with `Clarity1` rules?

src/vm/ast/definition_sorter/mod.rs
```diff
@@ -213,11 +213,12 @@ impl<'a> DefinitionSorter {
...
                         } else if let Some(native_function) =
                             // NOTE: can use ClarityVersion::latest() here only as long as NO NEW FUNCTIONS are special cased
-                        //        in the definition sorter.
+                            //        in the definition sorter.
                             NativeFunctions::lookup_by_name_at_version(
-                            function_name,
-                            &ClarityVersion::latest(),
-                        ) {
+                                    function_name,
+                                    &ClarityVersion::latest(),
```
jcnelson, line 219: Hold up -- you need to gate this.  The Clarity version you should be passing here is whatever the transaction's `GlobalContext` is running in.

src/vm/mod.rs
```diff
@@ -366,6 +366,28 @@ pub fn eval_all(
     })
 }
 
+#[derive(Debug, Clone, PartialEq, Copy, Serialize, Deserialize)]
+pub enum StacksNetworkType {
```
jcnelson, line 370: Please remove -- we just use `true` or `false` for mainnet everywhere else in the codebase.  No need to introduce two different ways to represent the same concept.

src/vm/tests/mod.rs
```diff
@@ -103,6 +109,16 @@ pub fn execute(s: &str) -> Value {
     vm_execute(s).unwrap().unwrap()
 }
 
+pub fn execute_against_version_and_network(
```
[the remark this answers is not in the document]
jcnelson: If you insist on keeping this (I agree with @pavitthrap -- we just `unwrap()` in-place in every other context), then can you at least call this something else to indicate that it will panic on failure?  Maybe `execute_against_version_and_network_nofail` or something?

src/vm/analysis/arithmetic_checker/tests.rs
kantai: It's because the arithmetic checker doesn't do type checking -- it's only looking for calls to native functions that are disallowed. In Clarity1, `is-in-mainnet` is a legal variable name that could be used as a function argument, for example. In Clarity2, however, it disqualifies a contract from being "arithmetic only". These example tests would fail the type checker, because they reference unbounded variables, but the arithmetic checker doesn't do variable dereferencing.

src/vm/ast/definition_sorter/mod.rs
kantai: I think this came up in a different PR review -- the definition sorter is not a "normal" analysis pass (it's not an analysis pass at all -- that's why it's part of the AST module). All it is trying to do in this loop is figure out whether or not a particular function application is a "special" application where the arguments aren't dereferenced (e.g., the tuple keys in a tuple constructor). As long as the new Clarity version doesn't add a new branch to this match, it is fine to just use `ClarityVersion` latest here. However, it may make sense to future-proof this better by gating the `definition-sorter`, but I think that's sufficiently different that it should be done in it's own PR. Issue here: https://github.com/blockstack/stacks-blockchain/issues/2759

src/vm/mod.rs
kantai: Yep-- I agree with this. We use a boolean `is_mainnet` everywhere else (which maybe we shouldn't have), so we should be consistent with that.

kantai (approved):
This LGTM -- but you should resolve the comments first.

gregorycoppola:
Thanks for the reviews everyone..

This PR is dependent on https://github.com/blockstack/stacks-blockchain/pull/2735 going in, so I'll wait for that.

src/vm/analysis/arithmetic_checker/tests.rs
jcnelson: Ah, gotcha. Thanks!

src/vm/tests/mod.rs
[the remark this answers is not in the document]
gregorycoppola: I removed this function.

jcnelson (approved):
LGTM!

gregorycoppola (commented):
thanks 4 the review :)

Issue #2532, opened 2021-03-23
PoX Updates: `stack-extend` and `delegate-stack-extend`


kantai:
As part of updating PoX in Stacks 2.1, the PoX contract should implement:

```
;; extend the "tx-sender"s active lockup
(define-public (stack-extend (extend-count uint) (start-at uint)))
;; extend the delegated active lockup
(define-public (delegate-stack-extend (stacker principal) (extend-count uint) (start-at uint)))
```

in order to support extensions of existing Stacking locks.

hstove:
- Is `start-at` a cycle or a block height? 
- Why not only have the parameter `extend-count`, which extends from the end of the last currently stacked cycle?
- Can `stack-extend` be called at any time during a current stacking position? i.e. if I'm locked for 12 cycles, and in the 4th cycle, can I `stack-extend`?
  - Is it safe to assume that I'll never be able to `stack-extend` that would result in my last cycle being more than 12 blocks in the future? i.e. if I have 8 cycles left, I can only call `(stack-extend u4)`?

jcnelson:
Implemented with #2755

Issue #2533, opened 2021-03-23
PoX Updates: `stack-increase`


kantai:
As part of updating PoX in Stacks 2.1, the PoX contract should implement:

```
;; increase the amount in "tx-sender"s active lockup
(define-public (stack-increase (increase-by uint)))
;; increase the amount in a delegated active lockup
(define-public (delegate-stack-increase (stacker principal) (increase-by uint)))
```

to enable increasing the amount that a particular stacker has locked.

jcnelson:
The interface should be as follows:

```clarity
;; increase the amount in "tx-sender"s active lockup
(define-public (stack-increase (increase-by uint) (index uint)))
;; increase the amount in a delegated active lockup
(define-public (delegate-stack-increase (stacker principal) (increase-by uint) (index uint)))
```

The `index` indexes the `reward-cycle-pox-address-list` map.

saralab:
@jcnelson : @kantai  can help with along with https://github.com/stacks-network/stacks-blockchain/issues/2534 , Let me know if I can mark him as the assignee here .

saralab:
Can we close this issue? The PR was merged .

jcnelson:
Yes

PR #2755, opened 2021-06-30
Stacks 2.1: PoX 2 `stack-extend`, `delegate-stack-extend`

### LINKED ISSUE #2522: pox contract: parameter in allow-contract-caller for upper limit does not work
**Describe the bug**
Calling `allow-contract-caller` with a `until-burn-ht` value in the future, results in stacks function to fail with permission denied.

**Steps To Reproduce**
1. deploy a contract that calls e.g. `delegate-stx`
1. make a `allow-contract-caller` call with a `until-burn-ht`
1. call the contract function that calls the genesis `delegate-stx` function
1. see the error `9` permission denied.

**Expected behavior**
The contract call should fail only after `until-burn-ht` block has passed.

Reporting it here as discovered with @yknl

kantai:
This implements #2532 and fixes #2522 in a new `pox-2` contract.

Test coverage is supplied through two testing structures.

`contract_tests` contains "unit" tests for these new functions and fixes. This tests the contract purely using a Clarity VM -- so it can check how the contract itself is working, but not how it interacts with the SortitionDB / ChainsCoordinator when constructing reward sets.

`test_pox_extend_transition_pox_2` in the `pox_2_tests` module does integration testing for extension across the epoch boundary -- testing that a user that locks via `pox` in Epoch 2.0 can `stack-extend` in `pox-2` in Epoch 2.1.


src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -0,0 +1,792 @@
...
+      ;;  to "post-date" their `stack-stx` transaction
+      (asserts! (is-eq first-reward-cycle specified-reward-cycle)
+                (err ERR_INVALID_START_BURN_HEIGHT))
+
+      ;; must be called directly by the tx-sender or by an allowed contract-caller
+      (asserts! (check-caller-allowed)
+                (err ERR_STACKING_PERMISSION_DENIED))
+
+      ;; tx-sender principal must not be stacking
+      (asserts! (is-none (get-stacker-info tx-sender))
```
jcnelson, line 467: This code doesn't look any different from `.pox` -- in particular, the `get-stacker-info` code is the same, as are the checks here.  Were any changes made in `.pox-2` from `.pox`, or is the goal of this PR just to add `stack-extend` and `delegate-stack-extend` without e.g. implementing continuous stacking?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -0,0 +1,792 @@
...
+        first-burnchain-block-height: (var-get first-burnchain-block-height),
+        reward-cycle-length: (var-get pox-reward-cycle-length),
+        rejection-fraction: (var-get pox-rejection-fraction),
+        current-rejection-votes: (next-cycle-rejection-votes),
+        total-liquid-supply-ustx: stx-liquid-supply,
+    })
+)
+
+
+(define-public (stack-extend (extend-count uint)
```
jcnelson, line 686: Can you add a method-level comment block to describe this new method, and indicate that it's a new change in 2.1?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -0,0 +1,792 @@
...
+         (new-unlock-ht (reward-cycle-to-burn-height (+ u1 last-extend-cycle))))
+
+      ;; must be called directly by the tx-sender or by an allowed contract-caller
+      (asserts! (check-caller-allowed)
+                (err ERR_STACKING_PERMISSION_DENIED))
+
+      ;; tx-sender must be locked
+      (asserts! (> amount-ustx u0)
+        (err ERR_STACK_EXTEND_NOT_LOCKED))
+
```
jcnelson, line 704: It doesn't look like we're checking whether or not `tx-sender` is stacking.  Should this method only work if `tx-sender` is already stacking?  What would be the consequence of calling this method over and over again?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -0,0 +1,792 @@
...
+         (cur-cycle (current-pox-reward-cycle))
+         (lock-period (- last-extend-cycle cur-cycle))
+         (new-unlock-ht (reward-cycle-to-burn-height (+ u1 last-extend-cycle))))
+
+      ;; must be called directly by the tx-sender or by an allowed contract-caller
+      (asserts! (check-caller-allowed)
+                (err ERR_STACKING_PERMISSION_DENIED))
+
+      ;; tx-sender must be locked
+      (asserts! (> amount-ustx u0)
```
jcnelson, line 702: Shouldn't we also be checking that this meets the minimum amount?  I don't see a call to `can-stack-stx`, for example.

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -0,0 +1,792 @@
...
+)
+
+
+(define-public (stack-extend (extend-count uint)
+                             (pox-addr { version: (buff 1), hashbytes: (buff 20) }))
+   (let ((stacker-info (stx-account tx-sender))
+         (amount-ustx (get locked stacker-info))
+         (unlock-height (get unlock-height stacker-info))
+         (first-extend-cycle (burn-height-to-reward-cycle unlock-height))
+         (last-extend-cycle  (- (+ first-extend-cycle extend-count) u1))
```
jcnelson, line 692: This can't underflow, right?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -0,0 +1,792 @@
...
+         (cur-cycle (current-pox-reward-cycle))
+         (lock-period (- last-extend-cycle cur-cycle))
```
jcnelson, line 694: This can't underflow either, right?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -0,0 +1,792 @@
...
+
+(define-public (delegate-stack-extend
+                    (stacker principal)
+                    (pox-addr { version: (buff 1), hashbytes: (buff 20) })
+                    (extend-count uint))
+    (let ((stacker-info (stx-account stacker))
+         (amount-ustx (get locked stacker-info))
+         (unlock-height (get unlock-height stacker-info))
+         (first-extend-cycle (burn-height-to-reward-cycle unlock-height))
+         (last-extend-cycle  (- (+ first-extend-cycle extend-count) u1))
```
jcnelson, line 736: Same underflow question applies here too.

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -0,0 +1,792 @@
...
+      (asserts! (> amount-ustx u0)
+        (err ERR_STACK_EXTEND_NOT_LOCKED))
+
+      ;; stacker must have delegated to the caller
+      (let ((delegation-info (unwrap! (get-check-delegation stacker) (err ERR_STACKING_PERMISSION_DENIED))))
+        ;; must have delegated to tx-sender
+        (asserts! (is-eq (get delegated-to delegation-info) tx-sender)
+                  (err ERR_STACKING_PERMISSION_DENIED))
+        ;; must have delegated enough stx
+        (asserts! (>= (get amount-ustx delegation-info) amount-ustx)
```
jcnelson, line 759: Is this error condition ever reachable?  How would I go about increasing the amount of uSTX I lock through a delegate without also increasing the `amount-ustx` field in `delegation-info`?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -0,0 +1,792 @@
...
+        ;; if pox-addr is set, must be equal to pox-addr
+        (asserts! (match (get pox-addr delegation-info)
+                         specified-pox-addr (is-eq pox-addr specified-pox-addr)
+                         true)
+                  (err ERR_DELEGATION_POX_ADDR_REQUIRED))
+        ;; delegation must not expire before lock period
+        (asserts! (match (get until-burn-ht delegation-info)
+                         until-burn-ht (>= until-burn-ht
+                                           new-unlock-ht)
+                      true)
```
jcnelson, line 770: Why would this be an error condition?  Couldn't I revoke the delegate and re-authorize the delegate in the mean time?

src/chainstate/stacks/boot/pox-2.clar
kantai: The only changes to PoX made in this PR are #2532 (stack-extend) and #2522 (fixing the contract allowance expiration).

`stack-extend` _is_ how continuous stacking is implemented -- a caller that is currently locked can call extend to lengthen their lockup and "restack" without the current cooldown period.

src/chainstate/stacks/boot/pox-2.clar
kantai: It probably makes more sense to document the method in the `docs` module, but I can add a marker here to indicate that it is new in 2.1.

src/chainstate/stacks/boot/pox-2.clar
kantai: I'm not so worried about an underflow -- it would just abort the tx with a runtime error. But we should check to make sure that extend-count is positive and first-extend-cycle is in the future.

src/chainstate/stacks/boot/pox-2.clar
kantai: You could change your delegation relationship between extensions. i.e., you stack with a delegator for 6 cycles, and on the 5th cycle, you update your delegation map with a lower limit. Then, if the delegator tried to extend, they would hit this error case (it's covered in one of the unit tests).

src/chainstate/stacks/boot/pox-2.clar
kantai: Same as above -- if your delegation has expired with the delegator, then they shouldn't be allowed to extend.

kantai: (And to your question about `revoke -> reauthorize -> delegate-stack-extend`, yes, that would be allowed, the delegation info here is checked at the time of the `delegate-stack-extend` call, not the original `delegate-stack-stx` call).

src/chainstate/stacks/boot/contract_tests.rs
```diff
@@ -429,6 +432,908 @@ impl HeadersDB for TestSimHeadersDB {
...
+                                                           (pox-addr (tuple (version (buff 1)) (hashbytes (buff 20))))
+                                                           (start-burn-ht uint)
+                                                           (lock-period uint))
+                                   (contract-call? .pox-2 stack-stx amount-ustx pox-addr start-burn-ht lock-period))",
+                                None)
+            .unwrap();
+
+        let burn_height = env.eval_raw("burn-block-height").unwrap().0;
+
+        assert_eq!(
```
gregorycoppola, line 465: can you put a brief comment for each of these cases to explain why we get these results? because I think it's not obvious.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -417,3 +417,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
         "Bob tx1 should have resulted in a runtime error"
     );
 }
+
```
gregorycoppola, line 420: This is the old version of the test?

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -417,3 +417,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+/// In this test case, two Stackers, Alice and Bob stack and interact with the
```
gregorycoppola, line 421: What does this method `test_pox_extend_transition_pox_2` test now? Does it test anything new? If this test was testing the last PR, then how does it exercise the new `.clar` code?

gregorycoppola (commented):
Hey Aaron.. looks good but had some misc comments/questions.

src/chainstate/stacks/boot/pox-2.clar
philipdesmedt: @kantai would the following use case supported by this?

1. Arkadiko has a smart contract where users deposit their STX tokens to borrow our stablecoin against that STX position (called an Arkadiko vault)
2. As soon as the Arkadiko contract has the minimum amount of uSTX, we will call `stack-stx`
3. Whenever a new vault (i.e. new STX is deposited) is created or destroyed (i.e. STX gets withdrawn), we can call `stack-extend` to increase/decrease the amount of STX stacked in PoX through our contract.

Any thoughts?

kantai: That use case is intended to be serviced via the `stack-increase` #2533, but it seems likely that the use case would also want to use `stack-extend` and (potentially) `stack-decrease`.

philipdesmedt: Nice, great to hear that this will be possible. In the current implementation, we always stack 1 cycle and then have a cool down cycle, so there's no optimal usage of PoX & bitcoin yield within Arkadiko yet.

Stacks 2.1 is going to be a big improvement on this then 👍

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+                        0,
+                        512 * POX_THRESHOLD_STEPS_USTX,
+                        AddressHashMode::SerializeP2PKH,
+                        key_to_stacks_addr(&bob).bytes,
+                        4,
+                        tip.block_height,
+                    );
+                    block_txs.push(bob_lockup);
+
+                    // Alice _will_ auto-unlock, so stack in PoX v2
```
gregorycoppola, line 581: how are you using the word "so" here? she is stacking becasue she's unlocking, or are these two different events?

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+///
+/// Alice: stacks via PoX v1 for 4 cycles. The third of these cycles occurs after
+///        the PoX v1 -> v2 transition, and so Alice gets "early unlocked".
+///        Before the early unlock, Alice invokes `stack-extend` in PoX v2
+///        Alice tries to stack again via PoX v1, which is allowed by the contract,
+///        but forbidden by the VM (because PoX has transitioned to v2)
+/// Bob:   stacks via PoX v2 for 6 cycles.
+///        Bob extends 6 cycles
+#[test]
+fn test_pox_extend_transition_pox_2() {
```
gregorycoppola, line 498: how does "delegate extend" fit into this?

src/vm/database/structures.rs
```diff
@@ -407,6 +407,29 @@ impl<'db, 'conn> STXBalanceSnapshot<'db, 'conn> {
         };
     }
 
+    pub fn extend_lock_v2(&mut self, unlock_burn_height: u64) {
```
gregorycoppola, line 410: funciton comment, include panic conditions

src/vm/functions/special.rs
```diff
@@ -200,6 +224,54 @@ fn handle_pox_v2_api_contract_call(
...
+                    panic!(
+                        "FATAL: failed to extend lock from {} until {}: '{:?}'",
+                        stacker, unlock_height, &e
+                    );
+                }
+            }
+
+            return Ok(());
+        } else {
+            // nothing to do -- the function failed
```
gregorycoppola, line 272: why does the error get suppressed? please add to comment

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -0,0 +1,792 @@
...
+                  (err ERR_DELEGATION_EXPIRES_DURING_LOCK)))
```
gregorycoppola, line 771: is this case where they can't extend in one of the tests?

src/chainstate/stacks/boot/contract_tests.rs
```diff
@@ -429,6 +432,908 @@ impl HeadersDB for TestSimHeadersDB {
...
+    sim.execute_next_block(|_env| {});
+    sim.execute_next_block(|_env| {});
+    sim.execute_next_block(|_env| {});
+
+    sim.execute_next_block_as_conn(|conn| {
+        test_deploy_smart_contract(conn, &POX_2_CONTRACT_TESTNET, &POX_2_TESTNET_CODE).unwrap()
+    });
+
+    sim.execute_next_block(|env| {
+        assert_eq!(
```
gregorycoppola, line 860: are the different tests inside of this `execute_next_block` block meant to interact? if not, could they be separated to highlight that they are not interacting?

src/chainstate/stacks/boot/pox-2.clar
gregorycoppola: the other methods are documented in this file

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -0,0 +1,792 @@
...
+         (amount-ustx (get locked stacker-info))
+         (unlock-height (get unlock-height stacker-info))
+         (first-extend-cycle (burn-height-to-reward-cycle unlock-height))
+         (last-extend-cycle  (- (+ first-extend-cycle extend-count) u1))
+         (cur-cycle (current-pox-reward-cycle))
+         (lock-period (- last-extend-cycle cur-cycle))
+         (new-unlock-ht (reward-cycle-to-burn-height (+ u1 last-extend-cycle))))
+
+      ;; must be called directly by the tx-sender or by an allowed contract-caller
+      (asserts! (check-caller-allowed)
```
gregorycoppola, line 698: do these various asserts have corresponding tests?

src/chainstate/stacks/mod.rs
```diff
@@ -228,6 +229,9 @@ impl fmt::Display for Error {
             Error::MemPoolError(ref s) => fmt::Display::fmt(s, f),
             Error::NoTransactionsToMine => write!(f, "No transactions to mine"),
             Error::PoxAlreadyLocked => write!(f, "Account has already locked STX for PoX"),
+            Error::PoxExtendNotLocked => {
+                write!(f, "Account has not already locked STX for PoX extend")
```
gregorycoppola, line 233: newbie question: is "stacking" == "locking"?

src/vm/database/structures.rs
```diff
@@ -407,6 +407,29 @@ impl<'db, 'conn> STXBalanceSnapshot<'db, 'conn> {
...
+            // caller needs to have checked this
+            panic!("FATAL: account does not have locked tokens");
+        }
+
+        if unlock_burn_height <= self.burn_block_height {
+            // caller needs to have checked this
+            panic!("FATAL: cannot set a lock with expired unlock burn height");
+        }
+
+        self.balance = STXBalance::LockedPoxTwo {
```
gregorycoppola, line 426: i'm curious why it is right to overwrite this.. is there a test that calls this twice and three times?

src/burnchains/burnchain.rs
```diff
@@ -504,13 +504,22 @@ impl Burnchain {
         self.first_block_height + reward_cycle * (self.pox_constants.reward_cycle_length as u64) + 1
     }
 
-    pub fn block_height_to_reward_cycle(&self, block_height: u64) -> Option<u64> {
-        if block_height < self.first_block_height {
+    pub fn static_block_height_to_reward_cycle(
```
gregorycoppola, line 507: function comment.. in particular, what is `first_block_ht` in relation to `block_ht`?

gregorycoppola (commented):
Huge change. I think it looks pretty good.. trying to follow along with what is tested, and make sure each of the edge cases in the ".clar" file is tested.

src/chainstate/stacks/boot/pox-2.clar
kantai: > Back to Jude's question.. could this have been written as a module, that didn't copy-and-paste code, but instead just adds some new code?

I think this could be interpreted in two ways. First, could this have been written as a contract that just adds the new code? The answer is "yes" but with a huge caveat: the new contract would need to share its data-space with the original PoX contract -- i.e., it would need direct access to all the underlying data-var, data-map, etc. structures of the original PoX contract. This is problematic for a couple reasons. Firstly, it only works if those structures don't need schema changes. This is true for `stack-extend`, but once we try introducing things like allowing unlocks or stacking increases, we either need to do schema changes, or introduce new auxiliary data-structures that would need to be updated/maintained through contract-call interposition, etc. Secondly, it would require special-casing the data-structure lookup code in the VM so that the new contract used the old contract's data-space. This is possible, but would introduce an otherwise unnecessary potential source for a pretty scary class of bugs (contracts that use prior contracts' data-spaces).

I think the second way to interpret this is as follows: could this have been written as a contract that just includes the code of the prior contract as a rust literal string include? The answer here is "probably", but I don't think it'd be the best way to implement the new contract. Doing that would obscure the contract implementation, and if the original functions end up being changed, it'd be hard to follow exactly how they're being changed (you'd need to mentally track the changes to the string inclusion).

src/burnchains/burnchain.rs
jcnelson: `first_block_ht` == first burnchain block height at which Stacks blocks are mined
`block_ht` == current burnchain block height

src/chainstate/stacks/boot/mod.rs
```diff
@@ -866,6 +874,30 @@ pub mod test {
...
     }
 
+    pub fn make_pox_2_extend(
+        key: &StacksPrivateKey,
+        nonce: u64,
+        addr_version: AddressHashMode,
+        addr_bytes: Hash160,
+        lock_period: u128,
+    ) -> StacksTransaction {
+        // (define-public (stack-stx (amount-ustx uint)
```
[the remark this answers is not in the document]
jcnelson: This docstring helps clarify the structure of `new_contract_call` below.

src/chainstate/stacks/boot/pox-2.clar
jcnelson: > Back to Jude's question.. could this have been written as a module, that didn't copy-and-paste code, but instead just adds some new code?

I was not asking this.  I very much agree with @kantai that having `.pox-2` be the same as `.pox` plus the new code is a lot less error-prone since you don't have to share data-spaces between contracts.

I was asking specifically if `stack-stx` needed to change to accommodate continuous stacking.  The answer is "no."

src/chainstate/stacks/boot/pox-2.clar
jcnelson: More generally -- do we want to permit someone to `stack-extend` their locked STX indefinitely, even if the quantity stacked becomes smaller than the minimum amount?  I don't think it matters much either way from a technical standpoint, but it does affect how the user finds out that they won't clinch a reward address -- they can either get an error here, or they can wait until a reward cycle completes where they have no reward slots.

src/chainstate/stacks/boot/pox-2.clar
jcnelson: That's fair.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+    let num_blocks = 35;
+
+    let alice = keys.pop().unwrap();
+    let bob = keys.pop().unwrap();
+    let charlie = keys.pop().unwrap();
+
+    let mut alice_reward_cycle = 0;
+
+    for tenure_id in 0u32..num_blocks {
+        let microblock_privkey = StacksPrivateKey::new();
```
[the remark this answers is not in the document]
jcnelson: You need a microblock public key hash to make a block.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+             vrf_proof,
+             ref parent_opt,
+             ref parent_microblock_header_opt| {
+                let parent_tip = get_parent_tip(parent_opt, chainstate, sortdb);
+                let coinbase_tx = make_coinbase(miner, tenure_id as usize);
+
+                let mut block_txs = vec![coinbase_tx];
+
+                if tenure_id == 1 {
+                    // Alice locks up exactly 25% of the liquid STX supply, so this should succeed.
```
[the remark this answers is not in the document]
jcnelson: This is explained in SIP-007

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+                    let alice_lockup = make_pox_2_extend(
+                        &alice,
+                        1,
+                        AddressHashMode::SerializeP2PKH,
+                        key_to_stacks_addr(&alice).bytes,
+                        6,
+                    );
+                    block_txs.push(alice_lockup);
+                } else if tenure_id == 31 {
+                    // Alice would have unlocked under v1 rules, so try to stack again via PoX 1 and expect a runtime error
```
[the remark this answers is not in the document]
jcnelson: The latter

src/chainstate/stacks/boot/pox-2.clar
kantai: I'm not sure if it ultimately makes much difference in terms of alerting the user -- the `can-stack-stx` check is performed at the time of the extension, so if the liquid supply of STX increases during their stacking (e.g., they stack for 12 cycles, but the supply increases during cycle 6), they wouldn't be alerted either way.

The biggest difference is in how the user needs to deal with the consequences of this. In any event, to meet the new threshold, they will need to invoke `stack-increase` to increase their lock and stacking amount. However, if the `can-stack-stx` check is in place, then it imposes an ordering constraint: they would need to invoke `stack-increase` before extending.

I think it probably *does* make sense to add the check here, as it seems the least surprising behavior.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+                        512 * POX_THRESHOLD_STEPS_USTX,
+                        AddressHashMode::SerializeP2PKH,
+                        key_to_stacks_addr(&alice).bytes,
+                        12,
+                        tip.block_height,
+                    );
+                    block_txs.push(alice_lockup);
+                }
+
+                let block_builder = StacksBlockBuilder::make_regtest_block_builder(
```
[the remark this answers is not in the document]
jcnelson: Yes -- you make one of these per block created

src/chainstate/stacks/boot/contract_tests.rs
```diff
@@ -429,6 +432,908 @@ impl HeadersDB for TestSimHeadersDB {
...
+                &symbols_from_values(vec![
+                    Value::UInt(10),
+                    POX_ADDRS[2].clone(),
+                ])
+            )
+            .unwrap()
+            .0
+            .to_string(),
+            "(err 2)".to_string()
+        );
```
jcnelson, line 779: Just be sure to add a test in here to make sure that you can't extend "into the past".  `stack-extend` should only allow you to extend into the future.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+    );
+
+    assert!(
+        match bob_txs.get(&0).unwrap().result {
+            Value::Response(ref r) => r.committed,
+            _ => false,
+        },
+        "Bob tx0 should have committed okay"
+    );
+}
```
jcnelson, line 806: Do you think there could be a corresponding test for `delegate-stack-extend` to verify that Alice can delegate some STX to Bob, and then extend the delegation across the v1 -> v2 upgrade?

src/vm/functions/special.rs
```diff
@@ -200,6 +224,54 @@ fn handle_pox_v2_api_contract_call(
...
+                                unlock_height,
+                                locked_address: stacker,
+                            }),
+                        ));
+                    }
+                }
+                Err(ChainstateError::DefunctPoxContract) => {
+                    return Err(Error::Runtime(RuntimeErrorType::DefunctPoxContract, None))
+                }
+                Err(e) => {
```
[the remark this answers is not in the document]
jcnelson: AIUI the `stack-extend` function should have returned an error if the caller didn't have any tokens locked in the first place.  So if `stack-extend` succeeds but there are no tokens locked, then we have a safety failure and thus should panic.

src/vm/functions/special.rs
jcnelson: This is the case where `stack-extend` or `delegate-stack-extend` return an `(err ...)` response.  No additional work needs to happen in this case.

src/chainstate/stacks/mod.rs
jcnelson: yes, in this context

src/vm/functions/special.rs
```diff
@@ -200,6 +224,54 @@ fn handle_pox_v2_api_contract_call(
...
+        // applying a pox lock at this point is equivalent to evaluating a transfer
+        debug!(
+            "Handle special-case contract-call to {:?} {} (which returned {:?})",
+            boot_code_id("pox-2", global_context.mainnet),
+            function_name,
+            value
+        );
+
+        runtime_cost(
+            ClarityCostFunction::StxTransfer,
```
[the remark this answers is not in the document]
kantai: I don't think so -- we use the same cost when applying a lock, which is very similar cost-wise to what's happening here with the lock extend.

src/chainstate/stacks/boot/pox_2_tests.rs
kantai: Yes -- `test_pox_extend_transition_pox_2` covers the new functionality from the `stack-extend` function, in the function doc, it describes how Alice and Bob each interact with `stack-extend`. Will add a bit more detail to this comment.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+
+    let num_blocks = 35;
+
+    let alice = keys.pop().unwrap();
+    let bob = keys.pop().unwrap();
+    let charlie = keys.pop().unwrap();
+
+    let mut alice_reward_cycle = 0;
+
+    for tenure_id in 0u32..num_blocks {
```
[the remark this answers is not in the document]
kantai: No -- the magic 25 number is just the amount of burn blocks that the testing network goes through before the first Stacks block (i.e., in the test environment, Stacks block 0 occurs at "Bitcoin" height 25).

src/burnchains/burnchain.rs
```diff
@@ -504,13 +504,22 @@ impl Burnchain {
         self.first_block_height + reward_cycle * (self.pox_constants.reward_cycle_length as u64) + 1
```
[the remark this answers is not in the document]
kantai: They don't absolutely have to be, but it seemed easy to address #2522 at the same time.

src/chainstate/stacks/boot/pox-2.clar
kantai: Yes, `check-caller-allowed` is tested via unit tests in `pox_2_contract_caller_units`.

src/vm/database/structures.rs
kantai: Added a comment to this function -- this method is invoked by PoX2 as a result of `stack-extend` or `delegate-stack-extend`. The account structure reflects whether or not the account was locked by PoX1 or PoX2 -- this method will upgrade an account to PoX2. The two `pox_2_tests` attempt to invoke PoX functions in PoX1 after this has been called -- the expectation is that if PoX1 attempt to re-lock an account that was locked by PoX2 already, then it will abort with a runtime exception (which is checked in the tests).

src/chainstate/stacks/boot/pox_2_tests.rs
kantai: Yes, I will add

src/chainstate/stacks/boot/pox-2.clar
kantai: Checks added in bbd826d4bde81aee6c26f134439b9d73d8885e34

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+    );
+
+    let num_blocks = 35;
+
+    let alice = keys.pop().unwrap();
+    let bob = keys.pop().unwrap();
+    let charlie = keys.pop().unwrap();
+
+    let mut alice_reward_cycle = 0;
+
```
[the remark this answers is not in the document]
kantai: Yes -- addressed in 4a5677cbb463100b0283ea6c89e03420a1070534

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+            })
+            .unwrap();
+            assert_eq!(reward_addrs.len(), 0);
+
+            // record the first reward cycle when Alice's tokens get stacked
+            let tip_burn_block_height =
+                get_par_burn_block_height(peer.chainstate(), &tip_index_block);
+
+            if tenure_id == 1 {
+                alice_reward_cycle = 1 + burnchain
```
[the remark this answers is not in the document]
kantai: Done via unrolling in 4a5677cbb463100b0283ea6c89e03420a1070534

src/vm/functions/special.rs
```diff
@@ -73,6 +73,31 @@ fn parse_pox_stacking_result(
     }
 }
 
+fn parse_pox_extend_result(result: &Value) -> std::result::Result<(PrincipalData, u64), i128> {
```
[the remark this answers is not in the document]
kantai: Added in bbd826d4bde81aee6c26f134439b9d73d8885e34

src/vm/database/structures.rs
kantai: Added in bbd826d4bde81aee6c26f134439b9d73d8885e34

src/burnchains/burnchain.rs
kantai: Added in bbd826d4bde81aee6c26f134439b9d73d8885e34

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+
+    assert!(
+        match alice_txs.get(&1).unwrap().result {
+            Value::Response(ref r) => r.committed,
+            _ => false,
+        },
+        "Alice tx1 should have committed okay"
+    );
+
+    assert_eq!(
```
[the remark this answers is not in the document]
kantai: Added in bbd826d4bde81aee6c26f134439b9d73d8885e34

src/vm/functions/special.rs
```diff
@@ -73,6 +73,31 @@ fn parse_pox_stacking_result(
...
+                .get("unlock-burn-height")
+                .expect(&format!("FATAL: no 'unlock-burn-height'"))
+                .to_owned()
+                .expect_u128()
+                .try_into()
+                .expect("FATAL: 'unlock-burn-height' overflow");
+
+            Ok((stacker, unlock_burn_height))
+        }
+        Err(e) => Err(e.expect_i128()),
```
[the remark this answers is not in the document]
kantai: Added in bbd826d4bde81aee6c26f134439b9d73d8885e34

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+/// Alice: stacks via PoX v1 for 4 cycles. The third of these cycles occurs after
+///        the PoX v1 -> v2 transition, and so Alice gets "early unlocked".
+///        Before the early unlock, Alice invokes `stack-extend` in PoX v2
+///        Alice tries to stack again via PoX v1, which is allowed by the contract,
+///        but forbidden by the VM (because PoX has transitioned to v2)
+/// Bob:   stacks via PoX v2 for 6 cycles.
+///        Bob extends 6 cycles
+#[test]
+fn test_pox_extend_transition_pox_2() {
+    let AUTO_UNLOCK_HT = 12;
```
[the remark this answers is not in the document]
kantai: Added in bbd826d4bde81aee6c26f134439b9d73d8885e34

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+    let mut burnchain = Burnchain::default_unittest(0, &BurnchainHeaderHash::zero());
+    burnchain.pox_constants.reward_cycle_length = 5;
+    burnchain.pox_constants.prepare_length = 2;
+    burnchain.pox_constants.anchor_threshold = 1;
+    burnchain.pox_constants.v1_unlock_height = AUTO_UNLOCK_HT + 25;
```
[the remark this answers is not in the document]
kantai: Yep, added in bbd826d4bde81aee6c26f134439b9d73d8885e34

src/chainstate/stacks/boot/pox_2_tests.rs
kantai: Added in 20f156fe9c9434ba6c98d61f1c51ab6a96ed8dcc

src/chainstate/stacks/boot/contract_tests.rs
kantai: Added in 199ce4126

src/vm/docs/mod.rs
```diff
@@ -1645,6 +1645,18 @@ principal isn't materialized, it returns 0.
...
+const STX_GET_ACCOUNT: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: None,
+    signature: "(stx-account owner)",
+    description: "`stx-account` is used to query the STX account of the `owner` principal.
+
+This function returns a tuple with the canonical account representation for an STX account.
+This includes the current amount of unlocked STX, the current amount of locked STX, and the
+unlock height for any locked STX.
+",
+    example: "",
```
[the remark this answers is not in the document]
kantai: Added in 199ce4126

src/chainstate/stacks/boot/pox-2.clar
kantai: Added documentation for these methods in bbd826d4bde81aee6c26f134439b9d73d8885e34

src/chainstate/stacks/boot/pox-2.clar
kantai: Added `can-stack-stx` and `minimal-can-stack-stx` checks to `stack-extend` and `delegate-stack-extend` respectively.

src/chainstate/stacks/boot/mod.rs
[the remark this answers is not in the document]
kantai: Updated the comment for the two `stack-stx` callers to reflect the actual type signature (they left out burn height), removed the copy-pasta comment from the extend callers.

src/chainstate/stacks/boot/pox_2_tests.rs
kantai: I'm not sure what this comment refers to -- the `test_simple_pox_lockup_transition_pox_2` test is, I think, unchanged in this  PR. The new tests were based on the loop-rolled version of that test, but I've since unrolled them.

src/chainstate/stacks/boot/pox_2_tests.rs
kantai: This was documented further in bbd826d4bde81aee6c26f134439b9d73d8885e34, and the test was updated to be a bit more clear in what it is doing.

src/chainstate/stacks/boot/contract_tests.rs
kantai: Added a test in e0f607764 that covers the closest to this case that I think is possible -- it attempts to call `stack-extend` _after_ the lock for a user expired (that would be the only way to extend into the past). This case is caught early in invocation, because the user has no locked up funds.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -483,3 +483,324 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+                        get_reward_addresses_with_par_tip(c, &burnchain, sortdb, &tip_index_block)
+                            .unwrap(),
+                        c.test_get_total_ustx_stacked(sortdb, &tip_index_block, cur_reward_cycle)
+                            .unwrap(),
+                    )
+                });
+
+            eprintln!("\ntenure: {}\nreward cycle: {}\nmin-uSTX: {}\naddrs: {:?}\ntotal_liquid_ustx: {}\ntotal-stacked: {}\n", tenure_id, cur_reward_cycle, min_ustx, &reward_addrs, total_liquid_ustx, total_stacked);
+
+            if cur_reward_cycle >= alice_reward_cycle {
```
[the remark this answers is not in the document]
kantai: Updated in 1b97e6cc7

kantai:
Thanks for the reviews -- the existing comments have been addressed, and this PR is ready for review again.

PR #2756, opened 2021-07-01
feat(Clarity2): Parse and assemble principals

### LINKED ISSUE #2620: Converting and querying valid `principal` representations

gregorycoppola:
Fixes #2620

## Description
We add two new Clarity2 functions "parse-principal" and "assemble-principal". assemble-principal allows a principal representation to be created from its version and pub_key_hash constituents. parse-principal works in the other direction, parsing a principal representation into its constituent parts.

EDIT(@jcnelson): The functions are now called `principal-parse` and `principal-construct`.  They allow the caller to both construct standard and contract principals from `(buff 1) (buff 20)` and `(buff 1) (buff 20) (string-ascii 40)`, respectively, and to parse principal values into values with type `{ version: (buff 1), hash-bytes: (buff 20), name: (optional (string-ascii 40)) }`.  They both return `response` types, and permit the user to parse and construct principals with version bytes that are not _currently_ supported by the consensus rules but are otherwise well-formed by representing them in the `(err ..)` variant.

## Testing
See:

```
M	src/vm/tests/principals.rs
```


pavitthrap:
I think maybe you need to switch the branch to merge into from `master` to something else - there's a lot of unrelated commits right now

src/vm/analysis/errors.rs
```diff
@@ -125,6 +125,8 @@ pub enum CheckErrors {
     // get-block-info? errors
     NoSuchBlockInfoProperty(String),
     GetBlockInfoExpectPropertyName,
+    NoSuchParsePrincipalProperty(String),
+    ParsePrincipalPropertyName,
```
pavitthrap, line 129: Maybe `ParsePrincipalPropertyNameExpected` instead?

src/vm/docs/mod.rs
```diff
@@ -264,6 +264,22 @@ Note: This function is only available starting with Stacks 2.1.",
...
+    name: None,
+    signature: "(parse-principal property_name principal)",
+    description: "Note: This function is only available starting with Stacks 2.1.",
+    example: r#"
+"#,
+};
+
+const ASSEMBLE_PRINCIPAL_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: None,
+    signature: "(parse-principal property_name principal)",
```
pavitthrap, line 277: change signature here 

also could add examples for both PARSE_PRINCIPAL_API and ASSEMBLE_PRINCIPAL_API?

src/vm/functions/mod.rs
```diff
@@ -286,6 +288,15 @@ pub fn lookup_reserved_functions(name: &str, version: &ClarityVersion) -> Option
...
             ),
             IsStandard => SpecialFunction("special_is_standard", &principals::special_is_standard),
+            ParsePrincipal => SpecialFunction(
+                "special_parse_principal",
+                &principals::special_parse_principal,
+            ),
+            AssemblePrincipal => NativeFunction(
+                "native_assemble_principal",
+                NativeHandle::DoubleArg(&principals::native_assemble_principal),
+                ClarityCostFunction::Mul,
```
pavitthrap, line 298: Perhaps add a todo here for the cost function

src/vm/functions/principals.rs
```diff
@@ -49,3 +52,78 @@ pub fn special_is_standard(
...
+
+pub fn special_parse_principal(
+    args: &[SymbolicExpression],
+    env: &mut Environment,
+    context: &LocalContext,
+) -> Result<Value> {
+    check_argument_count(2, args)?;
+    runtime_cost(ClarityCostFunction::StxTransfer, env, 0)?;
+
+    info!("inside");
```
pavitthrap, line 64: remove info statement

src/vm/functions/principals.rs
```diff
@@ -49,3 +52,78 @@ pub fn special_is_standard(
...
+    // Handle the block property name input arg.
+    let property_name = args[0]
+        .match_atom()
+        .ok_or(CheckErrors::ParsePrincipalPropertyName)?;
+
+    info!("property_name: {:?}", property_name);
```
pavitthrap, line 70: remove this info too?

src/vm/functions/principals.rs
```diff
@@ -49,3 +52,78 @@ pub fn special_is_standard(
...
     }
 }
+
+pub fn special_parse_principal(
+    args: &[SymbolicExpression],
+    env: &mut Environment,
+    context: &LocalContext,
+) -> Result<Value> {
+    check_argument_count(2, args)?;
+    runtime_cost(ClarityCostFunction::StxTransfer, env, 0)?;
```
pavitthrap, line 62: also perhaps add a todo about the cost function here as well

src/vm/tests/costs.rs
```diff
@@ -84,6 +84,8 @@ pub fn get_simple_test(function: &NativeFunctions) -> &'static str {
         BuffToIntBe => "(buff-to-int-be 0x00000000000000000000000000000001)",
         BuffToUIntBe => "(buff-to-uint-be 0x00000000000000000000000000000001)",
         IsStandard => "(is-standard 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6)",
+        ParsePrincipal => "(parse-principal version 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6)",
+        AssemblePrincipal => "(assemble-principal 26 0x164247d6f2b425ac5771423ae6c80c754f7172b0)",
```
pavitthrap, line 88: should 26 be `u26`?

src/vm/tests/principals.rs
```diff
@@ -262,3 +265,117 @@ fn test_simple_is_standard_undefined_cases() {
...
+        execute_against_version_and_network(
+            mainnet_addr_test,
+            ClarityVersion::Clarity2,
+            StacksNetworkType::Mainnet
+        )
+        .unwrap()
+        .unwrap()
+    );
+
+    // Note: Still works, even though the address is invalid.
```
pavitthrap, line 296: Why is the behavior described in this note what we want?

pavitthrap (commented):
Just a few minor comments!

src/vm/docs/mod.rs
gregorycoppola: yes.. good points.. what do you think now?

gregorycoppola (commented):
Thanks for the comments!

src/vm/docs/mod.rs
pavitthrap: looks good to me now!

src/vm/tests/principals.rs
pavitthrap: I *think* the more constrained the function is, the better? I think it would be a reasonable assumption for a user that if a principal is parseable by this native function, then the principal is valid - this assumption could lead to confusion possibly.

src/vm/analysis/type_checker/tests/mod.rs
```diff
@@ -2425,3 +2425,79 @@ fn test_string_utf8_negative_len() {
         _ => false,
     });
 }
+
+#[test]
+fn test_parse_principal() {
+    let good = [
+        r#"(parse-principal version 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6)"#,
```
lgalabru, line 2432: A while back, we settled on a C convention for naming our function (map-get, var-set, etc) - the rationale can be found here https://github.com/blockstack/stacks-blockchain/issues/989.
If we're going to stick to our conventions, I think this functions should be named `principal-parse` and `principal-assemble` instead.

src/vm/functions/principals.rs
```diff
@@ -45,3 +48,76 @@ pub fn special_is_standard(
...
+
+pub fn special_parse_principal(
+    args: &[SymbolicExpression],
+    env: &mut Environment,
+    context: &LocalContext,
+) -> Result<Value> {
+    check_argument_count(2, args)?;
+    runtime_cost(ClarityCostFunction::Unimplemented, env, 0)?;
+
+    // Handle the block property name input arg.
```
lgalabru, line 60: `block property` 
copy pasta?

src/vm/docs/mod.rs
```diff
@@ -264,6 +264,46 @@ Note: This function is only available starting with Stacks 2.1.",
 "#,
 };
 
+const PARSE_PRINCIPAL_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: None,
+    signature: "(parse-principal property-name principal-address)",
+    description: "A principal value is a concatenation of two things: a single *version byte*
```
lgalabru, line 270:
```suggestion
    description: "A principal value is a concatenation of two components: a single *version byte*
```

src/vm/docs/mod.rs
```diff
@@ -264,6 +264,46 @@ Note: This function is only available starting with Stacks 2.1.",
...
+(indicating the type of account and the type of network that this principal can spend tokens on),
+and a 20-byte *public key hash* (indicating the principal's unique identity).
+`parse-principal` is used to decompose a principal into its component parts.
```
lgalabru, line 273: This piece of documentation would be a great place for exhaustively be enumerating all the versions.

lgalabru: Ah nevermind seeing it in the other function.

src/vm/docs/mod.rs
```diff
@@ -264,6 +264,46 @@ Note: This function is only available starting with Stacks 2.1.",
...
+    description: "A principal value is a concatenation of two things: a single *version byte*
+(indicating the type of account and the type of network that this principal can spend tokens on),
+and a 20-byte *public key hash* (characterizing the principal's unique identity).
+`assemble-principal` takes such a *version byte*, of type `uint`, 
+and a *public key hash*, of type `(buff 20)`, and creates a corresponding object of type `principal`.
+
+The version byte should be `22` for a single-signature account on mainnet, `20`
+for a multi-signature account on mainnet, `26` for a single-signature account on
+a testnet, `21` for a multi-signature account on a testnet. The public key hash
+should be a 20-byte buffer containing the hash of a public key.
```
lgalabru, line 299: Can you also give the hexadecimal representation (0x..) for this values? Developers are manipulating bytes.

src/vm/docs/mod.rs
```diff
@@ -264,6 +264,46 @@ Note: This function is only available starting with Stacks 2.1.",
...
+Note: This function is only available starting with Stacks 2.1.",
+    example: r#"
+(parse-principal version 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6) ;; Returns u26
+(parse-principal pub-key-hash 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6) ;; Returns 0x164247d6f2b425ac5771423ae6c80c754f7172b0
+"#,
+};
+
+const ASSEMBLE_PRINCIPAL_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: None,
+    signature: "(assemble-principal version-byte public-key-hash)",
```
lgalabru, line 289: I'm confused by the `version-byte` name. If it's a byte, it should be `(buff 1)`, not a uint?

src/vm/functions/mod.rs
```diff
@@ -88,6 +88,8 @@ define_versioned_named_enum!(NativeFunctions(ClarityVersion) {
     BuffToIntBe("buff-to-int-be", ClarityVersion::Clarity2),
     BuffToUIntBe("buff-to-uint-be", ClarityVersion::Clarity2),
     IsStandard("is-standard", ClarityVersion::Clarity2),
+    ParsePrincipal("parse-principal", ClarityVersion::Clarity2),
+    AssemblePrincipal("assemble-principal", ClarityVersion::Clarity2),
```
lgalabru, line 92: What's your take on `principal-new` instead of `principal-assemble`. this function looks like a constructor to me.

src/vm/functions/principals.rs
```diff
@@ -45,3 +48,76 @@ pub fn special_is_standard(
...
+    let property_name = args[0]
+        .match_atom()
+        .ok_or(CheckErrors::ParsePrincipalExpectPropertyName)?;
+
+    let principal_property = PrincipalProperty::lookup_by_name(property_name).ok_or(
+        CheckErrors::NoSuchParsePrincipalProperty(property_name.to_string()),
```
lgalabru, line 66: `version` and `hashbytes` sounds more like components than properties to me. What do you think?

src/vm/functions/mod.rs
lgalabru: Or `principal-construct` and `principal-destruct` ?

src/vm/functions/principals.rs
```diff
@@ -45,3 +48,76 @@ pub fn special_is_standard(
...
+        PrincipalProperty::Version => Value::UInt(version as u128),
+        PrincipalProperty::PubKeyHash => Value::Sequence(SequenceData::Buffer(BuffData {
+            data: pub_key_hash.into(),
+        })),
+    };
+    Ok(result)
+}
+
+pub fn native_assemble_principal(version: Value, pub_key_hash: Value) -> Result<Value> {
+    let verified_version = match version {
```
lgalabru, line 93: Should you actually verify here that the `version` is valid?

src/vm/tests/principals.rs
```diff
@@ -178,3 +181,116 @@ fn test_simple_is_standard_undefined_cases() {
...
+            true
+        )
+        .unwrap()
+        .unwrap()
+    );
+}
+
+#[test]
+fn test_simple_assemble_principal() {
+    let normal_case_test = r#"(assemble-principal u22 0xfa6bf38ed557fe417333710d6033e9419391a320)"#;
```
lgalabru, line 261: Could we add the following tests?
```
(assemble-principal u5904934 0x0102030405060708091011121314151617181920)
(assemble-principal u22 0x01020304050607080910111213141516171819)
(assemble-principal u22 0x010203040506070809101112131415161718192021)
(assemble-principal 22 0x0102030405060708091011121314151617181920)
(assemble-principal 0x00 0x0102030405060708091011121314151617181920)
(assemble-principal u0 0x0102030405060708091011121314151617181920)
(let ((version u9) (hash-bytes 0x0102030405060708091011121314151617181920))
    (assemble-principal version hash-bytes))
```

lgalabru: (execution + type analysis)

src/vm/tests/principals.rs
```diff
@@ -178,3 +181,116 @@ fn test_simple_is_standard_undefined_cases() {
...
+            data: hex_bytes("fa6bf38ed557fe417333710d6033e9419391a320").unwrap()
+        })),
+        execute_against_version_and_network(
+            mainnet_addr_test,
+            ClarityVersion::Clarity2,
+            true
+        )
+        .unwrap()
+        .unwrap()
+    );
```
lgalabru, line 256: Could we also have for multisig testnet / mainnet?

lgalabru (commented):
Thank you for taking on this feature @gregorycoppola!
I had some feedbacks mainly around testing.
I wonder if we should go with some different naming for the function, I think that I'm starting to like `principal-new` and `principal-get`.

src/vm/functions/mod.rs
pavitthrap: I like `principal-construct` & `principal-parse`

gregorycoppola: I don't have strong feelings. Who wants to make the call?

src/vm/docs/mod.rs
gregorycoppola: You think I should change the signature to accept `(buff 1)`? I can do it as long as we're sure and I don't have to change it back.

src/vm/functions/principals.rs
gregorycoppola: What's teh difference between an "component" and a "property"?

What code changes are you suggesting?

src/vm/tests/principals.rs
gregorycoppola: OK. Can do.

src/vm/functions/principals.rs
gregorycoppola: Pavitthra also said same thing. We can check this.

gregorycoppola (commented):
Hello.. just had a couple questions about the function signature before I started doing any coding.

Major questions are:

1) What should the function names be?
2) Should the "version byte" be a "(buff 1)" instead of "uint"?
3) Any other interface issues to discuss? E.g., "component" vs "property"?

src/vm/functions/mod.rs
lgalabru: Pick something you like, as long as it works with the convention. We will do a pass before 2.1, to make sure that there is some naming consistency with all the the functions that we introduced :)

src/vm/docs/mod.rs
lgalabru: Yeah I think we should change and use a `buff 1`, if possible.

src/vm/functions/principals.rs
lgalabru: What's your take shifting gears a bit here, and instead of going with `(principal-parse <something> principal) -> (buff 1) || (buff 20)`, go with something like `(principal-destruct principal) -> { version: (buff 1), hashbytes: (buff 20) }` ?
That way we don't have to do this lookup, and introduce a new error variant / pitfall for developers.

src/vm/docs/mod.rs
lgalabru: (for developer experience + cost optimization)

gregorycoppola (commented):
Thanks for all the comments..

This version contains major revisions including:

1) function names renamed to start with the noun: `principal-construct` and `principal-parse`

2) `principal-parse` now takes a single argument, and returns a tuple, instead of having a "get by property" interface as before

3) we now fail any time that a version byte is invalid, for both construct and parse

4) many new test cases

pavitthrap (approved):
lgtm! make sure to address my comment about avoiding the use of `unwrap` though

gregorycoppola (commented):
Thanks for the review

src/vm/functions/principals.rs
```diff
@@ -38,13 +53,100 @@ pub fn special_is_standard(
...
+    }
+
+    let buffer_data = match Value::buff_from(hash_bytes.to_vec()) {
+        Ok(data) => data,
+        Err(err) => return Err(err),
+    };
+
+    let tuple_data = TupleData::from_data(vec![
+        (
+            ClarityName::try_from("version".to_owned()).unwrap(),
```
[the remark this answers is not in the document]
pavitthrap: gotcha - could you switch the unwrap to an `expect` with a failure message?

src/vm/analysis/type_checker/tests/mod.rs
```diff
@@ -2529,3 +2530,101 @@ fn test_comparison_types() {
         assert_eq!(expected, &type_check_helper(&bad_test).unwrap_err().err);
     }
 }
+
+#[test]
+fn test_principal_parse() {
+    let good = [
+        // Standard good example.
+        r#"(principal-parse 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6)"#,
```
lgalabru, line 2538: Could we have a test case using a contract id (`'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6.contract`) ?

lgalabru: Speaking of which, should we have a third component `name: (option (string-ascii))`?

src/vm/functions/principals.rs
```diff
@@ -38,13 +53,97 @@ pub fn special_is_standard(
...
+        Value::Principal(PrincipalData::Contract(QualifiedContractIdentifier { issuer, name })) => {
+            (issuer.0, issuer.1)
+        }
+        _ => {
+            return Err(CheckErrors::TypeValueError(TypeSignature::PrincipalType, principal).into())
+        }
+    };
+
+    if !version_matches_mainnet(version_byte) && !version_matches_testnet(version_byte) {
+        return Err(CheckErrors::InvalidVersionByte.into());
```
kantai, line 79: I think this should be a `RuntimeErrorType`. In general, `CheckErrors` are errors that _should_ be caught by a type checker, and if they aren't, they _invalidate_ the transaction (meaning that the transactions can't even be included in a block).

src/vm/functions/principals.rs
```diff
@@ -38,13 +53,97 @@ pub fn special_is_standard(
...
+        _ => return Err(CheckErrors::TypeValueError(TypeSignature::UIntType, version).into()),
+    };
+
+    if verified_version.len() != 1 {
+        return Err(CheckErrors::TypeValueError(
+            TypeSignature::SequenceType(SequenceSubtype::BufferType(BufferLength(1))),
+            version,
+        )
+        .into());
+    }
```
kantai, line 114: Same as above -- I think this should be a runtime error, if I pass this method a variable for the version byte, and that variable holds a 0-length buff, the type checker will accept the input (because the type system only check max-length), and this case will fail.

src/vm/functions/principals.rs
```diff
@@ -38,13 +53,97 @@ pub fn special_is_standard(
...
+        }
+    };
+
+    if verified_hash_bytes.len() != 20 {
+        return Err(CheckErrors::TypeValueError(
+            TypeSignature::SequenceType(SequenceSubtype::BufferType(BufferLength(20))),
+            hash_bytes,
+        )
+        .into());
+    }
```
kantai, line 140: Same as above -- this should be a runtime error.

src/vm/analysis/type_checker/tests/mod.rs
kantai: Yep, I agree with @lgalabru's suggestion here: a `name` component would be useful for contract principals.

src/vm/functions/principals.rs
jcnelson: Is `(principal-parse)` supposed to terminate the transaction if given an invalid version byte?  That sounds needlessly egregious to me.

Also, if all we're doing here is *parsing* principals, then why make an unexpected version byte into an error at all?  Isn't there a separate method for testing the version byte to see if it's standard?  There's no guarantee that the set of version bytes we support today will be the *only* version bytes supported in the future.

jcnelson: Could we make it so that at least `(principal-parse)` returns `(response (tuple (....)) uint)` like the other methods that could possibly fail already do?

src/vm/docs/mod.rs
```diff
@@ -271,6 +271,46 @@ Note: This function is only available starting with Stacks 2.1.",
...
+and a 20-byte *public key hash*, characterizing the principal's unique identity.
+`principal-construct` takes such a `version-byte`, of type `(buff 1)`, 
+and `hash-bytes`, of type `(buff 20)`, and creates a corresponding object of type `principal`.
+
+The version byte should be `0x16` for a single-signature account on mainnet, `0x14`
+for a multi-signature account on mainnet, `0x1a` for a single-signature account on
+a testnet, `0x15` for a multi-signature account on a testnet. The public key hash
+should be a 20-byte buffer containing the hash of a public key.
+
+This method will return an `InvalidVersionByte` error if the input `version-byte` does
```
jcnelson, line 305: I really don't think this function should outright fail if the user gives bad input.  Like other functions that can fail in this way, this function could easily return a `(response)` type that tells the caller why it won't work.  If the caller wants the transaction to abort or short-circuit, they can use `unwrap-panic` or `try!`.

jcnelson (changes requested):
I think I have to overall questions about the design of this PR:

* **Should it be possible to parse and assemble principals using non-standard version bytes?**  I'm in favor of "yes" here for forwards compatibility, since we also have a function for telling the user if a version byte is standard with respect to the network the code is running on (or if not, we can easily add one).  But I can be convinced otherwise.

* **Should building or parsing a principal cause the transaction to abort?**  I think I'm a solid "no" on both of these.  Building or parsing a principal can easily take user-supplied data, so these functions shouldn't cause a transaction to become wholesale invalid.  Instead, these functions can return a `(response)` type that may wrap an error code identifying why a principal could not be parsed or assembled, so that the caller can unwrap it later to take the appropriate flow-control action.

src/vm/analysis/type_checker/tests/mod.rs
gregorycoppola: OK.. We can add optional third argument.

src/vm/docs/mod.rs
gregorycoppola: I had it as not failing initially. Then I think @lgalabru and @pavitthrap wanted it to fail for bad input.

Returning a `response` is a separate question. I can implement either but we should get some kind of consensus or at least decision as to what to do.

src/vm/functions/principals.rs
gregorycoppola: Again.. how to decide? I'm fine to do either on both questions.

gregorycoppola (commented):
Looks like reviewers have conflicting opinions about some things. How can we reach a resolution.

1) Should invalid version bytes be allowed? @jcnelson wants it to be allowed. @pavitthrap and @lgalabru suggested it be an error.

2) If there are errors, should we runtime error or return a response? @kantai 

They're all similar to implement but how to make a decision?

kantai:
This is what I think @jcnelson suggested:

* `parse-principal` returns a `tuple` type, being the components of the principal. 
* `assemble-principal` returns a `optional` type, where the some type is the constructed principal (and not doing a byte version check -- just allowing the caller to do that check separately). the `none` result would be used in the event that a too-small buffer is supplied for the version or the hash bytes.

The alternative variants are:

1. Error in the event of non-current version byte. This error should also be captured in a response type (rather than an optional, because now there's more than one reason a call could fail).
2. Rather than surface the error with a response type, surface via a _runtime_ error (I think that in no event, should we surface as a check error).


I think that the option @jcnelson suggested provides the most flexibility to contract authors. However, it may expose a bit of a trap to authors, which is that it makes it very easy to create and parse invalid principals. An alternative option, which still provides flexibility, but avoids the invalid principal trap is the following:

* `parse-principal` returns a `(response tuple tuple)` type. The `ok` type is the components of the principal, and returns `ok` _if and only if_ the supplied principal is valid in the current network. The `err` type is _also_ the components of the principal, but is used _if and only if_ the supplied principal is invalid in the current network.
* `assemble-principal` returns a `(response principal tuple)` type. The `ok` type is the  constructed principal and returns _if and only if_ the assembled principal is valid in the current network. The `err` type is a tuple containing an error code, and an `(optional principal)` entry. This `err` returns if the supplied buffers aren't valid or if the assembled principal is in the wrong network. If the err was because the assembled principal was in the wrong network, then the `err` type's `(optional principal)` entry would hold that assembled principal.

While this is a little more complex of a type signature, in most "normal" cases, callers are just checking for the `ok` type. If a contract author _really_ wants to accept invalid-network principals, they can do so, but it would require a bit more sophisticated `match`, etc. manipulation.

I'm in favor of that proposal.

As for coming to a decision, I think we should shoot for consensus among the people that have weighed in so far. If that doesn't seem possible, then we can set a decision deadline and try to force a decision.

jcnelson:
I like @kantai's proposal.

gregorycoppola:
I'm worried that @kantai 's proposal makes it hard for a new user to understand what this function does in the main case. I say this because the return type has these kind of parallel worlds to digest.

If we only want to admit invalid version bytes for forwards compatibility, will it be the case that we will be on a new Clarity or Stacks version by the time a new version byte is valid? E.g., "is-valid-version-byte" will change only when we get to Clarity3? In that case, the validity of the byte can be determined relative to the Clarity or Stacks version.

Are there use cases beyond forwards compability @jcnelson ?

kantai:
> I'm worried that @kantai 's proposal makes it hard for a new user to understand what this function does in the main case. I say this because the return type has these kind of parallel worlds to digest.

I think I understand that, but I believe that in the main case, the user just needs to know that the function is capable of failing (which shouldn't be too surprising: in one case you are parsing something, and parsing can fail. in the other case, you're destructuring, and that it can fail also shouldn't be too surprising), and like other functions capable of failing, it returns a response type.

jcnelson:
> I'm worried that @kantai 's proposal makes it hard for a new user to understand what this function does in the main case. I say this because the return type has these kind of parallel worlds to digest.

Parsing and building can both fail.  If we don't provide a way for these functions to gracefully return an error type, then we have to take the more-surprising path of aborting the transaction (or even the block).

> If we only want to admit invalid version bytes for forwards compatibility

A major use-case for admitting different version bytes is that they provide a convenient way to represent the same keys as different principals (i.e. `principal` type instances with different values), based on the context in which they operate.  For example, today an off-chain BNS name and an on-chain BNS name can both be owned by the same key, but their owner addresses are intentionally different (they have the same 20-byte hash, but different version bytes).  This allows a Clarity contract to look at a name owner address and know that it can only correspond to an off-chain name or only to an on-chain name, but not both.

gregorycoppola (commented):
@kantai @jcnelson . I have implemented the return-type/version-byte checking as discussed in https://github.com/blockstack/stacks-blockchain/pull/2756#issuecomment-900415272

The specifics of the return types are discussed in `src/vm/docs/mod.rs`

I did not yet implement @lgalabru 's suggestion for an optional string for constructing contracts because I felt like a lot of changes had already happened to make the version-byte/response-type changes, and maybe we could use a review now.

src/vm/docs/mod.rs
```diff
@@ -271,6 +271,58 @@ Note: This function is only available starting with Stacks 2.1.",
...
+`hash-bytes`, and returns a principal.
+
+This function returns a `Response`. On success, the `ok` value is a `Principal`.
+The `err` value is a value tuple with the form `{err_int:UInt,principal_opt:Option<Principal>}`.
+
+If the single-byte `version-byte` is in the valid range `0x00` to `0x1f`, but is not an appropriate
+version byte for the current network, then the error will be `u0`, and `principal_opt` will contain
+`Some<Principal>`, where the wrapped value is the principal.
+
+If the `version-byte` is a `buff` of length less than 1, if the single-byte `version-byte` is a
```
jcnelson, line 315: Um, `version-byte` is _always_ `(buff 1)`; otherwise this won't even type-check.

Also, what's with the cut-off at 0x1f?  Don't you mean 0xff?

src/vm/functions/principals.rs
```diff
@@ -38,13 +66,199 @@ pub fn special_is_standard(
...
+            version,
+        )
+        .into());
+    }
+
+    // If the version byte buffer has 0 bytes, this is a recoverable error, because it wasn't the
+    // job of the type system.
+    if verified_version.len() < 1 {
+        // do some kind of error
+        return Ok(create_principal_true_error_response(1));
```
jcnelson, line 206: Can the `1` be a meaningful `const` variable somewhere, like we do with other Clarity error codes?

src/vm/functions/principals.rs
```diff
@@ -38,13 +66,199 @@ pub fn special_is_standard(
...
+    }
+
+    // Assume: verified_version.len() == 1
+    let version_byte = (*verified_version)[0];
+
+    // If the version byte is >= 32, this is a recoverable error, because it wasn't the job of the
```
jcnelson, line 212: Why 32?

src/vm/docs/mod.rs
```diff
@@ -271,6 +271,58 @@ Note: This function is only available starting with Stacks 2.1.",
...
+value greater than `0x1f`, or the `hash-bytes` is a `buff` of length less than 20, then `err_int`
+will be `u1` and `principal_opt` will be `None`.
+
+Note: This function is only available starting with Stacks 2.1.",
+    example: r#"
+(principal-construct 0x16 0xfa6bf38ed557fe417333710d6033e9419391a320) ;; Returns (err (tuple (error_int u0) (value (some SP3X6QWWETNBZWGBK6DRGTR1KX50S74D3433WDGJY))))
+(principal-construct 0x1a 0xfa6bf38ed557fe417333710d6033e9419391a320) ;; Returns (ok ST3X6QWWETNBZWGBK6DRGTR1KX50S74D3425Q1TPK)
+"#,
+};
```
jcnelson, line 324: Can you add some examples of the error outcomes?

src/vm/functions/principals.rs
```diff
@@ -38,13 +66,199 @@ pub fn special_is_standard(
...
+///
+/// The response is an error Response, where the `err` value is a tuple `{error_int,parse_tuple}`.
+/// `error_int` is of type `UInt`, `parse_tuple` is None.
+fn create_principal_true_error_response(error_int: u32) -> Value {
+    Value::Response(ResponseData {
+        committed: false,
+        data: Box::new(Value::Tuple(
+            TupleData::from_data(vec![
+                ("error_int".into(), Value::UInt(error_int.into())),
+                ("value".into(), Value::none()),
```
jcnelson, line 108: I thought this was `(err_int: xxx, principal_opt: xxx)`, per the documentation?

src/vm/tests/principals.rs
```diff
@@ -178,3 +182,353 @@ fn test_simple_is_standard_undefined_cases() {
...
+                TupleData::from_data(vec![
+                    ("error_int".into(), Value::UInt(1 as u128)),
+                    ("value".into(), Value::none()),
+                ])
+                .expect("FAIL: Failed to initialize tuple."),
+            )),
+        }),
+    );
+
+    // The version byte 0x20 is too big, even for the future. So, we get no result.
```
jcnelson, line 517: I would bet Bitcoins to Satoshis that 32 address bytes is not going to be enough.

gregorycoppola: I thought you could only represent 32 options with the "c32" address.

src/vm/docs/mod.rs
gregorycoppola: I changed the text to "buff length of 0"

for the 32: isn't it implied by this? https://github.com/blockstack/stacks-blockchain/blob/master/src/address/c32.rs

can it really be anything?

src/vm/docs/mod.rs
```diff
@@ -271,6 +271,58 @@ Note: This function is only available starting with Stacks 2.1.",
...
+
+If the `version-byte` is a `buff` of length less than 1, if the single-byte `version-byte` is a
+value greater than `0x1f`, or the `hash-bytes` is a `buff` of length less than 20, then `err_int`
+will be `u1` and `principal_opt` will be `None`.
+
+Note: This function is only available starting with Stacks 2.1.",
+    example: r#"
+(principal-construct 0x16 0xfa6bf38ed557fe417333710d6033e9419391a320) ;; Returns (err (tuple (error_int u0) (value (some SP3X6QWWETNBZWGBK6DRGTR1KX50S74D3433WDGJY))))
+(principal-construct 0x1a 0xfa6bf38ed557fe417333710d6033e9419391a320) ;; Returns (ok ST3X6QWWETNBZWGBK6DRGTR1KX50S74D3425Q1TPK)
+"#,
```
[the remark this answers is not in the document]
gregorycoppola: Not sure what you mean here.. The error codes are what we were discussing in the last paragraph, and third to last paragraph of this doc.

Can you give some example text of what you mean?

src/vm/docs/mod.rs
gregorycoppola: I added one.. I'm wondering if we really need to dwell that much on the error cases in the docs.

src/vm/functions/principals.rs
gregorycoppola: whoops.. changed the docs.

src/vm/functions/principals.rs
gregorycoppola: comes from this I thought https://github.com/blockstack/stacks-blockchain/blob/master/src/address/c32.rs

can the byte really be anything?

src/vm/functions/principals.rs
gregorycoppola: do you have an example of this? I'm seeing stuff like 

```
        return Ok(Value::err_uint(1));
```

`src/vm/functions/crypto.rs`

hstove:
> I did not yet implement @lgalabru 's suggestion for an optional string for constructing contracts because I felt like a lot of changes had already happened to make the version-byte/response-type changes, and maybe we could use a review now.

Is there an issue to track this? It would be very handy.

src/vm/tests/principals.rs
jcnelson: The version *byte* can encode up to 256 possible value.

src/vm/docs/mod.rs
jcnelson: This is the go-to source of Clarity function behavior, so we *absolutely* must document the error types.

src/vm/functions/principals.rs
jcnelson: Yes

src/vm/docs/mod.rs
jcnelson: Yes, it can be anything.

src/vm/docs/mod.rs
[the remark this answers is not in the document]
jcnelson: Asking for examples that represent all possible error paths.  Looks like you added one.

src/vm/functions/principals.rs
jcnelson: I don't know what `1` means.  Is it documented somewhere?  We document this kind of thing in `vm/docs/mod.rs` at least, for e.g. `stx-transfer?` error codes.

r0zar:
I have a use case where I'd like to ensure two traits passed into a function call are from the same principal, so I'm looking forward to this PR. :]

jcnelson:
I'm taking over this PR.

jcnelson:
This PR is now ready for review @kantai @obycode @lgalabru @gregorycoppola @pavitthrap

clarity/src/vm/docs/mod.rs
```diff
@@ -279,6 +279,93 @@ Note: This function is only available starting with Stacks 2.1.",
 "#,
 };
 
+const PRINCIPAL_PARSE_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: None,
+    signature: "(principal-parse principal-address)",
```
kantai, line 284: The name of this function seems backwards to me -- I usually think of parsing as "buffer-to-structure", but this is kind of "structure-to-buffers" -- maybe a better name is `principal-destruct`? This also conveys that it is the inverse of `principal-construct`.

clarity/src/vm/types/mod.rs
```diff
@@ -1287,6 +1289,16 @@ impl From<TupleData> for Value {
...
 
+impl From<ContractName> for Value {
+    fn from(name: ContractName) -> Self {
+        // ContractName is guaranteed to be between 5 and 40 bytes and contains only printable
+        // ASCII already, so this conversion should not fail.
+        Value::Sequence(SequenceData::String(CharType::ASCII(ASCIIData {
+            data: name.as_str().as_bytes().to_vec(),
+        })))
+    }
+}
```
kantai, line 1300: It may end up being more verbose, but I think it would add more clarity to the conversion to split this into two conversions:

```suggestion
impl From<ASCIIData> for Value {
    fn from(ascii: ASCIIData) -> Self {
        Value::Sequence(SequenceData::String(CharType::ASCII(ascii)))
    }
}
impl From<ContractName> for ASCIIData {
    fn from(name: ContractName) -> Self {
        // ContractName is guaranteed to be between 5 and 40 bytes and contains only printable
        // ASCII already, so this conversion should not fail.
        ASCIIData {
            data: name.as_str().as_bytes().to_vec(),
        }
    }
}
```

Then, instead of `Value::from(name)`, callers would need to perform `Value::from(ASCIIData::from(name))`. I know this seems like it is an unnecessary inflation of the code, but I don't think it's obvious that `Value::from(name)` should produce the ASCII variant.

clarity/src/vm/functions/principals.rs
```diff
@@ -38,13 +69,259 @@ pub fn special_is_standard(
...
+    Value::Response(ResponseData {
+        committed: false,
+        data: Box::new(Value::Tuple(
+            TupleData::from_data(vec![
+                ("error_int".into(), Value::UInt(error_int.into())),
+                ("value".into(), Value::none()),
+            ])
+            .expect("FAIL: Failed to initialize tuple."),
+        )),
+    })
```
kantai, line 125:
```suggestion
    Value::error(
        Value::Tuple(
            TupleData::from_data(vec![
                ("error_int".into(), Value::UInt(error_int.into())),
                ("value".into(), Value::none()),
            ])
            .expect("FAIL: Failed to initialize tuple."),
        )
    )
```

clarity/src/vm/functions/principals.rs
```diff
@@ -38,13 +69,259 @@ pub fn special_is_standard(
...
+        return Ok(create_principal_true_error_response(1));
+    }
+
+    // Assume: verified_version.len() == 1
+    let version_byte = (*verified_version)[0];
+
+    // If the version byte is >= 32, this is a runtime error, because it wasn't the job of the
+    // type system.  This is a requirement for c32check encoding.
+    if version_byte >= 32 {
+        return Ok(create_principal_true_error_response(1));
```
kantai, line 228: The error codes should have consts (this method uses a couple different error codes)

clarity/src/vm/functions/principals.rs
```diff
@@ -38,13 +69,259 @@ pub fn special_is_standard(
...
+    };
 
-    Ok(Value::Bool(
-        (address_is_mainnet && context_is_mainnet) || (address_is_testnet && !context_is_mainnet),
-    ))
+    if version_byte_is_valid {
+        Ok(Value::Response(ResponseData {
+            committed: version_byte_is_valid,
+            data: Box::new(principal),
+        }))
```
kantai, line 323:
```suggestion
        Value::okay(principal)
```

clarity/src/vm/functions/principals.rs
```diff
@@ -38,13 +69,259 @@ pub fn special_is_standard(
...
+        // (err ..).
+        let name_string = match name {
+            // destruct again to avoid a .clone() on the inner ascii_data.data
+            Value::Sequence(SequenceData::String(CharType::ASCII(ascii_data))) => {
+                String::from_utf8(ascii_data.data).expect("FAIL: could not convert bytes of type (string-ascii 40) back to a UTF-8 string")
+            },
+            _ => {
+                unreachable!()
+            }
+        };
```
kantai, line 300: This code can be cleaned up so that it doesn't do a second destruct while still avoiding a clone, though it adds a "restructuring" in an unreachable error path (it shouldn't be possible in production to get that `TypeValue` error).

```suggestion
        let name_bytes = match name {
            Value::Sequence(SequenceData::String(CharType::ASCII(ascii_data))) => {
                ascii_data
            }
            _ => {
                return Err(CheckErrors::TypeValueError(
                    TypeSignature::contract_name_string_ascii(),
                    name,
                )
                .into())
            }
        };

        // If it's not long enough, then it's a runtime error that warrants an (err ..) response.
        if name_bytes.len() < CONTRACT_MIN_NAME_LENGTH {
            return Ok(create_principal_true_error_response(2));
        }

        // if it's too long, then this should have been caught by the type-checker
        if name_bytes.len() > CONTRACT_MAX_NAME_LENGTH {
            return Err(CheckErrors::TypeValueError(
                TypeSignature::contract_name_string_ascii(),
                Value::from(name_bytes),
            )
            .into());
        }

        // The type-checker can't verify that the name is a valid ContractName, so we'll need to do
        // it here at runtime.  If it's not valid, then it warrants this function evaluating to
        // (err ..).
        let name_string = String::from_utf8(name_bytes.data).expect("FAIL: could not convert bytes of type (string-ascii 40) back to a UTF-8 string");
```

clarity/src/vm/functions/principals.rs
```diff
@@ -38,13 +69,259 @@ pub fn special_is_standard(
...
+                ("error_int".into(), Value::UInt(error_int.into())),
+                ("value".into(), Value::none()),
+            ])
+            .expect("FAIL: Failed to initialize tuple."),
+        )),
+    })
+}
+
+/// Creates Response return type, to wrap a *return value returned as an error* result of a
+/// `principal-construct` or `principal-parse`.
```
pavitthrap, line 129: this function seems to be used for `principal-construct` - same thing for `create_principal_true_error_response`

clarity/src/vm/functions/principals.rs
pavitthrap: `Ok(Value::okay(principal))`

clarity/src/vm/tests/principals.rs
```diff
@@ -284,3 +289,815 @@ fn test_simple_is_standard_undefined_cases() {
...
+// is still valid.
+fn test_principal_construct_version_byte_future() {
+    // The version byte 0x1f is unrecognized today, but is valid for the future.
+    let input = r#"(principal-construct 0x1f 0x0102030405060708091011121314151617181920)"#;
+    assert_eq!(
+        Value::Response(ResponseData {
+            committed: false,
+            data: Box::new(Value::Tuple(
+                TupleData::from_data(vec![
+                    ("error_int".into(), Value::UInt(0 as u128)),
```
pavitthrap, line 850: Once you create constants for the error codes you can use them in these tests too

pavitthrap (commented):
Just a few nits, agree with Aaron's code suggestions too

clarity/src/vm/functions/principals.rs
jcnelson: `version_byte_is_valid` may be false, so this won't do.

clarity/src/vm/docs/mod.rs
jcnelson: Fixed

clarity/src/vm/types/mod.rs
jcnelson: Fixed

clarity/src/vm/functions/principals.rs
jcnelson: Fixed

clarity/src/vm/functions/principals.rs
jcnelson: Fixed

clarity/src/vm/functions/principals.rs
```diff
@@ -38,13 +69,259 @@ pub fn special_is_standard(
...
+
+    // If the version byte buffer has 0 bytes, this is a recoverable error, because it wasn't the
+    // job of the type system.
+    if verified_version.len() < 1 {
+        // do some kind of error
+        return Ok(create_principal_true_error_response(1));
+    }
+
+    // Assume: verified_version.len() == 1
+    let version_byte = (*verified_version)[0];
```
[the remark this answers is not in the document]
jcnelson: Fixed

clarity/src/vm/functions/principals.rs
jcnelson: Fixed

clarity/src/vm/functions/principals.rs
jcnelson: Fixed

clarity/src/vm/types/signatures.rs
```diff
@@ -673,6 +677,21 @@ impl TypeSignature {
         ))
     }
 
+    pub fn contract_name_string_ascii() -> TypeSignature {
```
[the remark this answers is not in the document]
jcnelson: Fixed

clarity/src/vm/types/signatures.rs
```diff
@@ -673,6 +677,21 @@ impl TypeSignature {
...
+        TypeSignature::bound_string_ascii(
+            CONTRACT_MAX_NAME_LENGTH
+                .try_into()
+                .expect("FAIL: contract name max length exceeds u32 space"),
+        )
+    }
+
+    pub fn bound_string_ascii(max_len: u32) -> TypeSignature {
```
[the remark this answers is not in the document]
jcnelson: Fixed

clarity/src/vm/functions/principals.rs
jcnelson: Fixed

clarity/src/vm/tests/principals.rs
jcnelson: Fixed

clarity/src/vm/docs/mod.rs
```diff
@@ -279,6 +279,93 @@ Note: This function is only available starting with Stacks 2.1.",
...
+depending on whether or not the third `(string-ascii 40)` argument is given.
+
+This function returns a `Response`. On success, the `ok` value is a `Principal`.
+The `err` value is a value tuple with the form `{ error_int: uint, value: (optional principal) }`.
+If the `value` field in the `err` variant is `(some ..)`, then the wrapped `Principal` will be 
+the well-formed principal that could have been formed, had the version byte been supported.
+
+If the single-byte `version-byte` is in the valid range `0x00` to `0x1f`, but is not an appropriate
+version byte for the current network, then the error will be `u0`, and `value` will contain
+`(some principal)`, where the wrapped value is the principal.
```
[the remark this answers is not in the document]
jcnelson: Fixed

kantai (approved):

jcnelson (approved):

PR #2757, opened 2021-07-01
Chore: BufferLength code quality

kantai:
This PR makes the BufferLength instance variable private. The rationale for this is that the BufferLength constructor enforces the Clarity TypeSignature invariant that a TypeSignature fails to construct if the described type is too large.

As part of this, this also refactored the BufferLength constants to be lazy_statics, so that the length check is applied to the constants on first use.

### DIFF
--- a/src/burnchains/mod.rs
+++ b/src/burnchains/mod.rs
@@ -167,7 +167,7 @@ pub trait PrivateKey: Clone + fmt::Debug + serde::Serialize + serde::de::Deseria
 
 pub trait Address: Clone + fmt::Debug + fmt::Display {
     fn to_bytes(&self) -> Vec<u8>;
-    fn from_string(&str) -> Option<Self>
+    fn from_string(from: &str) -> Option<Self>
     where
         Self: Sized;
     fn is_burn(&self) -> bool;
--- a/src/vm/functions/crypto.rs
+++ b/src/vm/functions/crypto.rs
@@ -80,11 +80,11 @@ pub fn special_principal_of(
     let pub_key = match param0 {
         Value::Sequence(SequenceData::Buffer(BuffData { ref data })) => {
             if data.len() != 33 {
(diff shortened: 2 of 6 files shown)


src/burnchains/mod.rs
```diff
@@ -167,7 +167,7 @@ pub trait PrivateKey: Clone + fmt::Debug + serde::Serialize + serde::de::Deseria
 
 pub trait Address: Clone + fmt::Debug + fmt::Display {
     fn to_bytes(&self) -> Vec<u8>;
-    fn from_string(&str) -> Option<Self>
+    fn from_string(from: &str) -> Option<Self>
```
kantai, line 170: These changes are to address warnings about anonymous function arguments.

src/vm/functions/crypto.rs
```diff
@@ -80,11 +80,11 @@ pub fn special_principal_of(
     let pub_key = match param0 {
         Value::Sequence(SequenceData::Buffer(BuffData { ref data })) => {
             if data.len() != 33 {
-                return Err(CheckErrors::TypeValueError(BUFF_33, param0).into());
+                return Err(CheckErrors::TypeValueError(BUFF_33.clone(), param0).into());
```
gregorycoppola, line 83: Why do these become clone? Was it copy before? Is copy better? Can we make this copyable?

kantai: Previously, BUFF_33 was a const, which means the line here was an inlined constructor, so more or less equivalent to an auto-impl'd clone. But since it's no longer a const, the clone needs to be explicit.

TypeSignature cannot implement `Copy` -- `Copy` works for types that can be safely `memcpy`ed without changing their semantics. That's not true of TypeSignatures (mostly because it's a recursive type).

jcnelson (approved):
LGTM!

gregorycoppola (approved):
sounds good

pavitthrap (approved):

PR #2758, opened 2021-07-06
Chore: fix all warns in `cargo check --tests --workspace`

kantai:
This PR fixes all the warnings generated by current rust stable in `develop` when run with `cargo check --workspace --tests` (previously, warnings were only addressed in `cargo check`)

### DIFF
--- a/src/net/http.rs
+++ b/src/net/http.rs
@@ -4894,12 +4894,11 @@ mod test {
         for (data, errstr) in tests.iter() {
             let res = HttpRequestPreamble::consensus_deserialize(&mut data.as_bytes());
             test_debug!("Expect '{}'", errstr);
-            let expected_errstr = format!("{:?}", &res);
-            assert!(res.is_err(), "{}", expected_errstr);
+            assert!(res.is_err(), "{:?}", &res);
             assert!(
-                res.unwrap_err().to_string().find(errstr).is_some(),
-                "{}",
-                expected_errstr
+                res.as_ref().unwrap_err().to_string().find(errstr).is_some(),
+                "{:?}",
+                &res
             );
         }
     }
@@ -4947,13 +4946,16 @@ mod test {
 
         for (data, errstr) in tests.iter() {
             let sres = StacksHttpPreamble::consensus_deserialize(&mut data.as_bytes());
-            let expected_serrstr = format!("{:?}", &sres);
             test_debug!("Expect '{}'", errstr);
-            assert!(sres.is_err(), "{}", expected_serrstr);
+            assert!(sres.is_err(), "{:?}", &sres);
             assert!(
-                sres.unwrap_err().to_string().find(errstr).is_some(),
-                "{}",
-                expected_serrstr
+                sres.as_ref()
+                    .unwrap_err()
+                    .to_string()
+                    .find(errstr)
+                    .is_some(),
+                "{:?}",
+                &sres
             );
         }
     }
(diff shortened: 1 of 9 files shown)


src/net/http.rs
```diff
@@ -4895,9 +4895,10 @@ mod test {
             let res = HttpRequestPreamble::consensus_deserialize(&mut data.as_bytes());
             test_debug!("Expect '{}'", errstr);
             let expected_errstr = format!("{:?}", &res);
```
pavitthrap, line 4897: could simplify to `assert!(res.is_err(), "{:?}", &res);` and delete previous line

src/net/http.rs
```diff
@@ -4948,9 +4949,10 @@ mod test {
             let sres = StacksHttpPreamble::consensus_deserialize(&mut data.as_bytes());
             let expected_serrstr = format!("{:?}", &sres);
             test_debug!("Expect '{}'", errstr);
-            assert!(sres.is_err(), expected_serrstr);
+            assert!(sres.is_err(), "{}", expected_serrstr);
```
pavitthrap, line 4952: can simplify the assert here as well

src/net/http.rs
```diff
@@ -5213,9 +5215,10 @@ mod test {
             let sres = StacksHttpPreamble::consensus_deserialize(&mut data.as_bytes());
             let expected_serrstr = format!("{:?}", &sres);
             test_debug!("Expect '{}', got: {:?}", errstr, &sres);
-            assert!(sres.is_err(), expected_serrstr);
+            assert!(sres.is_err(), "{}", expected_serrstr);
```
pavitthrap, line 5218: can simplify here

src/net/http.rs
```diff
@@ -5464,12 +5467,13 @@ mod test {
             let e = http.read_payload(&preamble, &bad_content_length.as_bytes()[offset..]);
             let estr = format!("{:?}", &e);
 
-            assert!(e.is_err(), estr);
+            assert!(e.is_err(), "{}", estr);
```
pavitthrap, line 5470: can simplify here

src/util/retry.rs
```diff
@@ -169,11 +169,12 @@ mod test {
         let mut tmp_buf = [0u8; 3];
         let e = retry_reader.read_exact(&mut tmp_buf);
         let e_str = format!("{:?}", &e);
-        assert!(e.is_err(), e_str);
+        assert!(e.is_err(), "{}", e_str);
```
pavitthrap, line 172: can simplify here

pavitthrap (approved):
Just left some minor comments about removing some calls to `format!` - but LGTM!

gregorycoppola:
Should we get all these changes in `next` as well?

Is the easiest way to do that two PR's?

jcnelson (approved):

kantai:
> Should we get all these changes in next as well?
> Is the easiest way to do that two PR's?

Yes -- and sort of. For "upstream" updates, i.e., `master -> develop` or `develop -> next`, we shouldn't really need a review. However, branch protection rules mean that we in practice need that rubber stamp. So yeah, opening a PR is probably the easiest way to achieve that.

src/util/retry.rs
kantai: Addressed in 93a7c393069ca4420433e03338a6534bee19dc8f

src/net/http.rs
kantai: Addressed in 93a7c393069ca4420433e03338a6534bee19dc8f

src/net/http.rs
kantai: Addressed in 93a7c393069ca4420433e03338a6534bee19dc8f

src/net/http.rs
kantai: Addressed in 93a7c393069ca4420433e03338a6534bee19dc8f

src/net/http.rs
kantai: Addressed in 93a7c393069ca4420433e03338a6534bee19dc8f

PR #2768, opened 2021-07-14
Feat/health check endpoint

### LINKED ISSUE #1229: RESTful API for querying blockchain status
something like `/v1/health` with a return code of 200 if everything is synced and up to date.

pavitthrap:
## Description
Fixes #1229 

The get health endpoint first gets a list of all valid initial peers of the node. It then gets the max height amongst those nodes. If the block height of the node is greater than or equal to that max height, the health check endpoint returns `true` for the field `is_healthy`, and `false` otherwise. The endpoint also returns the fraction `(block height of node/ max height of initial peers)` as a percentage. 

In addition to the rpc endpoint, I updated some function and variable names in `http.rs` for consistency (because there is a discrepancy between newly added rpc calls and older ones). I switched older methods to use pure snake case in naming (so `get_poxinfo` would now be `get_pox_info`). Let me know if this change is unwanted, I can switch it back. If it is wanted, I will make a similar change in `rpc.rs` as well. 

### To think about: 
- It's possible a node isn't keeping track of the block stats for any of its initial peers - might need to consider fallback scenario (where we consider the block heights of any node we are currently keeping the stats of)

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

Added docs to existing rpc documentation. 

## Testing information
Added typical RPC test for this. Mocked out block stats for the peer server. 

Need to test this on a live node.

### DIFF
--- a/docs/rpc-endpoints.md
+++ b/docs/rpc-endpoints.md
@@ -344,3 +344,11 @@ object of the following form:
 Determine whether a given trait is implemented within the specified contract (either explicitly or implicitly).
 
 See OpenAPI [spec](./rpc/openapi.yaml) for details.
+
+### GET /v2/health
+
+Determine whether a node is healthy. The notion of health used for this 
+endpoint is whether or not the block height of a node is greater than or 
+equal to the max block height of its initial peers. 
+
+See OpenAPI [spec](./rpc/openapi.yaml) for details.
--- a/src/net/db.rs
+++ b/src/net/db.rs
@@ -1296,13 +1296,36 @@ impl PeerDB {
     /// -- always include all allowed neighbors
     /// -- never include denied neighbors
     /// -- for neighbors that are neither allowed nor denied, sample them randomly as long as they're fresh.
-    pub fn get_initial_neighbors(
+    pub fn get_randomized_initial_neighbors(
         conn: &DBConn,
         network_id: u32,
         count: u32,
         block_height: u64,
     ) -> Result<Vec<Neighbor>, db_error> {
         PeerDB::get_random_neighbors(conn, network_id, count, block_height, true)
+    }
+
+    pub fn get_valid_initial_neighbors(
+        conn: &DBConn,
+        network_id: u32,
+        block_height: u64,
+    ) -> Result<Vec<Neighbor>, db_error> {
+        // UTC time
+        let now_secs = util::get_epoch_time_secs();
+
+        let query = "SELECT * FROM frontier WHERE initial = 1 AND (allowed < 0 OR ?1 < allowed) \
+        AND network_id = ?2 AND denied < ?3 AND ?4 < expire_block_height"
+            .to_string();
+
+        let args: &[&dyn ToSql] = &[
+            &u64_to_sql(now_secs)?,
+            &network_id,
+            &u64_to_sql(now_secs)?,
+            &u64_to_sql(block_height)?,
+        ];
+
+        let initial_peers = query_rows::<Neighbor, _>(conn, &query, args)?;
+        Ok(initial_peers)
     }
 
     /// Get a randomized set of peers for walking the peer graph.
@@ -1633,17 +1656,20 @@ mod test {
         )
         .unwrap();
 
-        let n5 = PeerDB::get_initial_neighbors(db.conn(), 0x9abcdef0, 5, 23455).unwrap();
+        let n5 = PeerDB::get_randomized_initial_neighbors(db.conn(), 0x9abcdef0, 5, 23455).unwrap();
         assert!(are_present(&n5, &initial_neighbors));
 
-        let n10 = PeerDB::get_initial_neighbors(db.conn(), 0x9abcdef0, 10, 23455).unwrap();
+        let n10 =
+            PeerDB::get_randomized_initial_neighbors(db.conn(), 0x9abcdef0, 10, 23455).unwrap();
         assert!(are_present(&n10, &initial_neighbors));
 
-        let n20 = PeerDB::get_initial_neighbors(db.conn(), 0x9abcdef0, 20, 23455).unwrap();
+        let n20 =
+            PeerDB::get_randomized_initial_neighbors(db.conn(), 0x9abcdef0, 20, 23455).unwrap();
         assert!(are_present(&initial_neighbors, &n20));
 
         let n15_fresh =
-            PeerDB::get_initial_neighbors(db.conn(), 0x9abcdef0, 15, 23456 + 14).unwrap();
+            PeerDB::get_randomized_initial_neighbors(db.conn(), 0x9abcdef0, 15, 23456 + 14)
+                .unwrap();
         assert!(are_present(
             &n15_fresh[10..15].to_vec(),
             &initial_neighbors[10..20].to_vec()
--- a/src/net/http.rs
+++ b/src/net/http.rs
@@ -2860,19 +2860,33 @@ impl HttpResponseType {
     ) -> Result<HttpResponseType, net_error> {
         if preamble.status_code < 400 || preamble.status_code > 599 {
             return Err(net_error::DeserializeError(
-                "Inavlid response: not an error".to_string(),
+                "Invalid response: not an error".to_string(),
             ));
         }
 
-        if preamble.content_type != HttpContentType::Text {
+        if preamble.content_type != HttpContentType::Text
+            && preamble.content_type != HttpContentType::JSON
+        {
             return Err(net_error::DeserializeError(
-                "Invalid error response: expected text/plain".to_string(),
+                "Invalid error response: expected text/plain or json".to_string(),
             ));
         }
 
         let mut error_text = String::new();
+        let mut json_val: serde_json::Value = Default::default();
         fd.read_to_string(&mut error_text)
             .map_err(net_error::ReadError)?;
+        if preamble.content_type == HttpContentType::JSON {
+            let json_result = serde_json::from_str(&error_text.clone());
+            json_val = match json_result {
+                Ok(val) => val,
+                Err(serde_json::Error { .. }) => {
+                    return Err(net_error::DeserializeError(
+                        "Invalid response: unable to deserialize json response".to_string(),
+                    ))
+                }
+            }
+        }
 
         let md = HttpResponseMetadata::from_preamble(request_version, preamble);
         let resp = match preamble.status_code {
@@ -2883,6 +2897,8 @@ impl HttpResponseType {
             404 => HttpResponseType::NotFound(md, error_text),
             500 => HttpResponseType::ServerError(md, error_text),
             503 => HttpResponseType::ServiceUnavailable(md, error_text),
+            512 => HttpResponseType::GetHealthError(md, json_val),
+            513 => HttpResponseType::GetHealthQueryError(md, error_text),
             _ => HttpResponseType::Error(md, preamble.status_code, error_text),
         };
         Ok(resp)
@@ -3553,6 +3569,8 @@ impl HttpResponseType {
             HttpResponseType::ServerError(ref md, _) => md,
             HttpResponseType::ServiceUnavailable(ref md, _) => md,
             HttpResponseType::Error(ref md, _, _) => md,
+            HttpResponseType::GetHealthError(ref md, _) => md,
+            HttpResponseType::GetHealthQueryError(ref md, _) => md,
         }
     }
 
@@ -3799,6 +3817,21 @@ impl HttpResponseType {
             HttpResponseType::ServiceUnavailable(_, ref msg) => {
                 self.error_response(fd, 503, msg)?
             }
+            HttpResponseType::GetHealthError(ref md, ref data) => {
+                HttpResponsePreamble::new_serialized(
+                    fd,
+                    512,
+                    HttpResponseType::error_reason(512),
+                    md.content_length.clone(),
+                    &HttpContentType::JSON,
+                    md.request_id,
+                    |ref mut fd| keep_alive_headers(fd, md),
+                )?;
+                HttpResponseType::send_json(protocol, md, fd, data)?;
+            }
+            HttpResponseType::GetHealthQueryError(_, ref msg) => {
+                self.error_response(fd, 513, msg)?
+            }
             HttpResponseType::Error(_, ref error_code, ref msg) => {
                 self.error_response(fd, *error_code, msg)?
             }
@@ -3916,6 +3949,8 @@ impl MessageSequence for StacksHttpMessage {
                 HttpResponseType::NotFound(_, _) => "HTTP(404)",
                 HttpResponseType::ServerError(_, _) => "HTTP(500)",
                 HttpResponseType::ServiceUnavailable(_, _) => "HTTP(503)",
+                HttpResponseType::GetHealthError(..) => "HTTP(512)",
+                HttpResponseType::GetHealthQueryError(..) => "HTTP(513)",
                 HttpResponseType::Error(_, _, _) => "HTTP(other)",
             },
         }
@@ -5990,27 +6025,14 @@ mod test {
 
     #[test]
     fn test_http_response_type_codec_err() {
-        let request_paths = vec![
-            "/v2/blocks/1111111111111111111111111111111111111111111111111111111111111111",
-            "/v2/transactions",
-            "/v2/neighbors",
-            "/v2/neighbors",
-            "/v2/neighbors",
-        ];
+        let request_paths = vec!["/v2/transactions", "/v2/neighbors", "/v2/neighbors"];
         let bad_request_payloads = vec![
-            "HTTP/1.1 200 OK\r\nServer: stacks/v2.0\r\nX-Request-Id: 123\r\nContent-Type: application/json\r\nContent-length: 2\r\n\r\nab",
             "HTTP/1.1 200 OK\r\nServer: stacks/v2.0\r\nX-Request-Id: 123\r\nContent-Type: application/json\r\nContent-length: 4\r\n\r\n\"ab\"",
             "HTTP/1.1 200 OK\r\nServer: stacks/v2.0\r\nX-Request-Id: 123\r\nContent-Type: application/json\r\nContent-length: 1\r\n\r\n{",
             "HTTP/1.1 200 OK\r\nServer: stacks/v2.0\r\nX-Request-Id: 123\r\nContent-Type: application/json\r\nContent-length: 1\r\n\r\na",
-            "HTTP/1.1 400 Bad Request\r\nServer: stacks/v2.0\r\nX-Request-Id: 123\r\nContent-Type: application/json\r\nContent-length: 2\r\n\r\n{}",
         ];
-        let expected_bad_request_payload_errors = vec![
-            "Invalid content-type",
-            "Invalid txid:",
-            "Not enough bytes",
-            "Failed to parse",
-            "expected text/plain",
-        ];
+        let expected_bad_request_payload_errors =
+            vec!["Invalid txid:", "Not enough bytes", "Failed to parse"];
         for (test, (expected_error, request_path)) in bad_request_payloads.iter().zip(
             expected_bad_request_payload_errors
                 .iter()
--- a/src/net/http.rs
+++ b/src/net/http.rs
@@ -89,21 +89,21 @@ use crate::codec::{
 use crate::types::chainstate::{BlockHeaderHash, StacksAddress, StacksBlockId};
 
 lazy_static! {
-    static ref PATH_GET_INFO: Regex = Regex::new(r#"^/v2/info$"#).unwrap();
-    static ref PATH_GET_POX_INFO: Regex = Regex::new(r#"^/v2/pox$"#).unwrap();
-    static ref PATH_GET_NEIGHBORS: Regex = Regex::new(r#"^/v2/neighbors$"#).unwrap();
-    static ref PATH_GET_BLOCK: Regex = Regex::new(r#"^/v2/blocks/([0-9a-f]{64})$"#).unwrap();
-    static ref PATH_GET_MICROBLOCKS_INDEXED: Regex =
+    static ref PATH_GETINFO: Regex = Regex::new(r#"^/v2/info$"#).unwrap();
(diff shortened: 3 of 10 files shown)


docs/rpc-endpoints.md
```diff
@@ -343,4 +343,12 @@ object of the following form:
 
 Determine whether a given trait is implemented within the specified contract (either explicitly or implicitly).
 
-See OpenAPI [spec](./rpc/openapi.yaml) for details.
\ No newline at end of file
+See OpenAPI [spec](./rpc/openapi.yaml) for details.
+
+### GET /v2/health
+
+Determine whether a node is healthy. The notion of health used for this
```
gregorycoppola, line 350: Where did you get this operationalization for "healthy"?

I see https://github.com/blockstack/stacks-blockchain/issues/1229. Is this an analogy to "v1 health"?

src/net/http.rs
```diff
@@ -1447,44 +1448,48 @@ impl HttpRequestType {
...
             ) -> Result<HttpRequestType, net_error>,
         )] = &[
-            ("GET", &PATH_GETINFO, &HttpRequestType::parse_getinfo),
-            ("GET", &PATH_GETPOXINFO, &HttpRequestType::parse_getpoxinfo),
+            ("GET", &PATH_GET_INFO, &HttpRequestType::parse_get_info),
             (
                 "GET",
-                &PATH_GETNEIGHBORS,
-                &HttpRequestType::parse_getneighbors,
+                &PATH_GET_POX_INFO,
```
gregorycoppola, line 1454: Is this "snake case" change inherently tied to your other changes? Could this be two PR's?

gregorycoppola (commented):
Hello Pavi,

Just has a couple of high-level questions before looking at the details.

docs/rpc-endpoints.md
pavitthrap: Yes - discussed this notion of "health" here with Aaron (https://discord.com/channels/621759717756370964/730814571517968436/837017500763422817)

src/net/http.rs
pavitthrap: I could separate it out into another PR, but since it was just renaming, it seemed like a minor enough refactor to group with this change

src/net/rpc.rs
```diff
@@ -3397,6 +3498,175 @@ mod test {
...
+            "test_rpc_get_health",
+            40010,
+            40011,
+            50010,
+            50011,
+            |ref mut peer_client,
+             ref mut convo_client,
+             ref mut peer_server,
+             ref mut convo_server| {
+                // // create block stats for peer server
```
gregorycoppola, line 3645: Is this comment really more central than for any other line? Is it copied from somewhere? Why is it double quotes?

src/net/rpc.rs
```diff
@@ -3397,6 +3498,175 @@ mod test {
...
+                    }
+                }
+            },
+        );
+    }
+
+    #[test]
+    #[ignore]
+    fn test_rpc_get_health_true() {
+        test_rpc(
```
gregorycoppola, line 3555: Is this comment really more central than for any other line? Is it copied from somewhere?

src/net/rpc.rs
```diff
@@ -3397,6 +3498,175 @@ mod test {
         );
     }
 
+    #[test]
```
gregorycoppola, line 3501: For each of these tests, can you please put some commenting that helps the reader understand what situation you are setting up and why we expect the given results?

I don't find it obvious.

src/net/rpc.rs
```diff
@@ -1509,6 +1510,86 @@ impl ConversationHttp {
...
+                    response_metadata.clone(),
+                    "Couldn't obtain stats on any bootstrap peers, unable to determine health"
+                        .to_string(),
+                );
+                return response.send(http, fd).map(|_| ());
+            }
+
+            let data = if block_height >= (max_height - 1) {
+                GetHealthResponse {
+                    is_healthy: true,
```
gregorycoppola, line 1575: Rather than "is_healthy", I wonder if you could call this something like "matches_peers" or something.

That way, we can eventually add different statistics here, and then the client can determine whether they think the node "is health" based on the other signals, like "matches_peers".

docs/rpc-endpoints.md
gregorycoppola: Sounds good.. only comment I had below was what if your "is_healthy" is rewritten to something like "matches_peers", leaving open the possibility of adding multiple "health"  metrics to this endpoint over time?

src/net/http.rs
gregorycoppola: Personally, I would prefer to see conceptually separable PR's separated. For reviewers, it's easier to know whether a change fits in with the necessary purpose of the change. For readers, they have an easier to follow record of why the lines changed.

But, I'm not asking you to separate them in *this* case, but maybe for the future you can consider it.

src/net/rpc.rs
```diff
@@ -1509,6 +1510,86 @@ impl ConversationHttp {
...
+            Some(tip) => tip.height,
+            None => 0,
+        };
+
+        let initial_neighbors =
+            PeerDB::get_valid_initial_neighbors(peerdb.conn(), network_id, block_height).unwrap();
+
+        if initial_neighbors.len() == 0 {
+            // no bootstrap nodes found, unable to determine health.
+            let response = HttpResponseType::NotFound(
```
gregorycoppola, line 1539: I'm not sure if this should be an "error". Maybe it can be a valid GetHealthResponse, that simply indicates things are bad.

That way:
1) This function can have a simpler definition.
1) The client can expect a standard form, especially if they want to write code to parse the response.

What are your thoughts on that?

src/net/rpc.rs
```diff
@@ -1509,6 +1510,86 @@ impl ConversationHttp {
...
+                            max_height = max_height.max(height);
+                            stats_obtained = true;
+                        }
+                    }
+                    None => {}
+                }
+            }
+
+            if !stats_obtained {
+                let response = HttpResponseType::NotFound(
```
gregorycoppola, line 1565: Same as above.. should this be an error or a special kind of status?

gregorycoppola (commented):
Overall, this is a very good functionality and should help us a lot in monitoring what is going on!

I had a few more comments on the API, the tests, and the idea that maybe we can leave the notion of "healthy" open to including multiple metrics.

src/net/http.rs
pavitthrap: Ya I understand this - I'll separate out the renames to a different pr!

docs/rpc-endpoints.md
pavitthrap: I think we can add more notions of health to this same endpoint as the need arises

src/net/rpc.rs
pavitthrap: Yes I see. I'll make a decision on this after discussing what the returns should look like with devops.

src/net/rpc.rs
pavitthrap: Are you referring to `// set up block stats for peer server`? It isn't copied from anywhere

src/net/rpc.rs
pavitthrap: removed the double quotes, also not copied from anywhere. I added it to give context to the setup code.

pavitthrap:
@wileyj @CharlieC3 Wondering what your thoughts are on the response type in `handle_get_health` (in `rpc.rs`). Would you prefer 
1) always getting a HttpResponseType of GetHealth, with a detailed error field
2) always want a 400 level error if either there is an issue querying for the health or if the node is not healthy
3) a 400 level error if there is an issue querying for health, and a GetHealth response otherwise (includes instances when query succeeds, but it is determined the node is not healthy)

src/net/rpc.rs
```diff
@@ -1509,6 +1510,86 @@ impl ConversationHttp {
...
+                    response_metadata.clone(),
+                    "Couldn't obtain stats on any bootstrap peers, unable to determine health"
+                        .to_string(),
+                );
+                return response.send(http, fd).map(|_| ());
+            }
+
+            let data = if block_height >= (max_height - 1) {
```
wileyj, line 1573: I would defer to @kantai or @lgalabru here, but `max_height - 1` may give us a flapping health output over time.

src/net/rpc.rs
wileyj: i agree with @gregorycoppola here, but i don't have an opinion on how to represent this data with the returned json. On the other hand, if the node didn't have any neighbors to sync with...that would be a larger problem for the node to deal with anyways. 

perhaps in this case, the output could be returned like this:
```
{
  "is_healthy": false,
  "percent_of_blocks_synced":  null
}
```
or some other identifier that signals the issue is with a lack of neighbors

src/net/rpc.rs
wileyj: > @wileyj @CharlieC3 Wondering what your thoughts are on the response type in `handle_get_health` (in `rpc.rs`). Would you prefer
> 
> 1. always getting a HttpResponseType of GetHealth, with a detailed error field
> 2. always want a 400 level error if either there is an issue querying for the health or if the node is not healthy
> 3. a 400 level error if there is an issue querying for health, and a GetHealth response otherwise (includes instances when query succeeds, but it is determined the node is not healthy)

if the node is indeed "not healthy" or still syncing - it would need to be non-200 status in order to prevent traffic from being sent to it by a load balancer health check.

wileyj: i wouldn't think we need an "error" message, the same json output would be good to have for all requests, with the only difference between a "healthy" and "not healthy" response being the HTTP status code (ex: `200` status for a healthy node, `503` for unhealthy. subsets of 5xx can be used for more detailed reasons if you prefer)

CharlieC3:
> @wileyj @CharlieC3 Wondering what your thoughts are on the response type in `handle_get_health` (in `rpc.rs`). Would you prefer
> 
> 1. always getting a HttpResponseType of GetHealth, with a detailed error field
> 2. always want a 400 level error if either there is an issue querying for the health or if the node is not healthy
> 3. a 400 level error if there is an issue querying for health, and a GetHealth response otherwise (includes instances when query succeeds, but it is determined the node is not healthy)

Agreed with @wileyj.

We plan on modeling the health endpoint to be compatible with basic standards that health-checking tools align on, and typically that's status codes between `200` and `399` means the service is ready to be used, and anything above `400` means it's not. So while a `500` could mean the stacks-node is broken or it's just not synced to the stacks tip, the distinction between these two doesn't matter much because the action we'd need to take for either of them would be the same.

The plan is to have a liveliness probe and a readiness probe, each with different functions. The liveliness probe would check the health endpoint for a passing or failing HTTP status code. The liveliness probe would have to fail N times over a probing interval of M minutes before it restarts the stacks-node.

The readiness probe usually has a lower threshold before it takes action. So it may have to fail (1/2 * N) times over a probing interval of M minutes before it stops allowing traffic to reach the stacks-node, but not restart it yet until the liveliness probe triggers later.

src/net/rpc.rs
CharlieC3: > I would defer to @kantai or @lgalabru here, but `max_height - 1` may give us a flapping health output over time.

The thresholds we incorporate in the health probes can account for this to only take action after seeing failing health for a while, like 30 minutes for example (considering how long a node may take to download a block from neighbors in rare times). It may only look strange to others hitting this endpoint in their browser and seeing a failed response every so often.

If the gap is widened beyond `max_height - 1`, to `max_height - 3` for example, it will take us longer to start detecting issues with the node since blocks can sometimes take an hour to be produced.

src/net/rpc.rs
gregorycoppola: I guess, in this case, that the "health" endpoint has the responsibility to make a binary decision on healthiness.

I was thinking maybe the client would want to weigh several factors and decide on healthiness. But, I guess now after learning about the context, that I guess the endpoint does need to weight whatever factors and make a decision itself.

pavitthrap: If I return type HttpResponseType:::GetHealth, the status code will be 200, which doesn't seem like the desired behavior here. 

I could make a specific response type HttpResponseType:::GetHealthError that has a 400 status code. This will deviate from how we "fail" in the rest of our rpc calls, but I think that's fine because we are trying to use this endpoint with Kubernetes probes.

gregorycoppola: But we can still return a lot of info in the response in case we want to understand why the health check failed.

docs/rpc-endpoints.md
wileyj: yes! as we find more things we want to know, it's easy within this structure to add a new k:v

src/net/rpc.rs
wileyj: in this case though, we're only concerned with the "healthiness" of the node itself at a specific moment in time. 
I would say that the node making those binary decisions *about itself* are fine. 
@pavitthrap yes - if you wanted to convey specific reasons for a non-healthy node (or even why it's healthy), more k:v pairs and differing HTTP response codes would be the best way to accomplish this i would think. 
My only ask if we do add more and more response codes is that we document what each code means in the context of this endpoint.

wileyj:
> > @wileyj @CharlieC3 Wondering what your thoughts are on the response type in `handle_get_health` (in `rpc.rs`). Would you prefer
> > 
> > 1. always getting a HttpResponseType of GetHealth, with a detailed error field
> > 2. always want a 400 level error if either there is an issue querying for the health or if the node is not healthy
> > 3. a 400 level error if there is an issue querying for health, and a GetHealth response otherwise (includes instances when query succeeds, but it is determined the node is not healthy)
> 
> Agreed with @wileyj.
> 
> We plan on modeling the health endpoint to be compatible with basic standards that health-checking tools align on, and typically that's status codes between `200` and `399` means the service is ready to be used, and anything above `400` means it's not. So while a `500` could mean the stacks-node is indefinitely stalled or it's just not synced to the stacks tip, the distinction between these two doesn't matter much because the action we'd need to take for either of them would be the same.
> 
> The plan is to have a liveliness probe and a readiness probe, each with different functions. The liveliness probe would check the health endpoint for a passing or failing HTTP status code. The liveliness probe would have to fail N times over a probing interval of M minutes before it restarts the stacks-node.
> 
> The readiness probe usually has a lower threshold before it takes action. So it may have to fail (1/2 * N) times over a probing interval of M minutes before it stops allowing traffic to reach the stacks-node, but not restart it yet until the liveliness probe triggers later.

yes, 4XX can be used for sure - but in this case i would argue 5XX is more appropriate and standard for what we're expecting. That said, there is no concrete rule on what to use here @pavitthrap

src/net/http.rs
```diff
@@ -2837,19 +2860,33 @@ impl HttpResponseType {
...
             ));
         }
 
-        if preamble.content_type != HttpContentType::Text {
+        if preamble.content_type != HttpContentType::Text
+            && preamble.content_type != HttpContentType::JSON
+        {
             return Err(net_error::DeserializeError(
-                "Invalid error response: expected text/plain".to_string(),
+                "Invalid error response: expected text/plain or json".to_string(),
```
lgalabru, line 2871:
```suggestion
                "Invalid error response: expected text/plain or application/json".to_string(),
```

src/net/rpc.rs
```diff
@@ -1509,6 +1510,91 @@ impl ConversationHttp {
...
+        req: &HttpRequestType,
+        network_id: u32,
+        sortdb: &SortitionDB,
+        chainstate: &StacksChainState,
+        peerdb: &PeerDB,
+        block_stats: Option<&HashMap<NeighborKey, NeighborBlockStats>>,
+    ) -> Result<(), net_error> {
+        let response_metadata = HttpResponseMetadata::from(req);
+
+        let block_height = match chainstate.get_stacks_chain_tip(sortdb)? {
```
lgalabru, line 1529: Instead of hitting the DB, should we be using the caching layer introduced by Jude in https://github.com/blockstack/stacks-blockchain/pull/2746 ?

src/net/rpc.rs
```diff
@@ -1509,6 +1510,91 @@ impl ConversationHttp {
...
+
+        // get max block height amongst bootstrap nodes
+        let response = if let Some(block_stats) = block_stats {
+            let mut max_height: u64 = 1;
+            let mut stats_obtained = false;
+            for neighbor in initial_neighbors.iter() {
+                let nk = &neighbor.addr;
+                match block_stats.get(nk) {
+                    Some(neighbor_block_stats) => {
+                        if neighbor_block_stats.status == NodeStatus::Online {
```
lgalabru, line 1554: Should we add a helper method in `NeighborBlockStats`?

src/net/rpc.rs
```diff
@@ -1509,6 +1510,91 @@ impl ConversationHttp {
...
+                let data = GetHealthResponse {
+                    matches_peers: true,
+                    percent_of_blocks_synced: 100,
+                };
+                HttpResponseType::GetHealth(response_metadata, data)
+            } else {
+                let data = GetHealthResponse {
+                    matches_peers: false,
+                    percent_of_blocks_synced: ((block_height as f32 / max_height as f32) * (100.0))
+                        as u8,
```
lgalabru, line 1584: You can probably remove these casting with: `block_height * 100 / max_height`

src/net/rpc.rs
pavitthrap: For what logic do you think we should add a helper method (it's not clear to me)?

src/net/rpc.rs
pavitthrap: I would get that we would want an un-cached version of block height. This value is ultimately used in the comparison `block_height >= (max_height - 1)`, which determines whether or not the node returns a 200 level response or not.

src/net/db.rs
```diff
@@ -1305,6 +1305,29 @@ impl PeerDB {
...
+    pub fn get_valid_initial_neighbors(
+        conn: &DBConn,
+        network_id: u32,
+        block_height: u64,
+    ) -> Result<Vec<Neighbor>, db_error> {
+        // UTC time
+        let now_secs = util::get_epoch_time_secs();
+
+        let query = "SELECT * FROM frontier WHERE initial = 1 AND (allowed < 0 OR ?1 < allowed) \
+        AND network_id = ?2 AND denied < ?3 AND ?4 < expire_block_height"
```
jcnelson, line 1317: You should probably also filter by the `peer_version` major version byte -- i.e. `peer_version & 0xff000000`.  It should be less than or equal to that of the local peer's version.

This doesn't happen often, but when we have a breaking upgrade like Stacks 2.1, the major peer version byte will change, and for long-running nodes that will re-use the same chainstate, this means that their frontier table will have some peers that are still Stacks 2.0 peers.  Stacks 2.0 peers shouldn't try and talk to Stacks 2.1 peers, but Stacks 2.1 peers may be able to talk to Stacks 2.0 peers.

src/net/http.rs
```diff
@@ -3756,6 +3817,21 @@ impl HttpResponseType {
             HttpResponseType::ServiceUnavailable(_, ref msg) => {
                 self.error_response(fd, 503, msg)?
             }
+            HttpResponseType::GetHealthError(ref md, ref data) => {
+                HttpResponsePreamble::new_serialized(
+                    fd,
+                    512,
```
jcnelson, line 3823: HTTP 512 (and HTTP 513 below) aren't standard HTTP status codes.  Do you want to put something more standard here, or is there a downstream tool that expects these?

src/net/rpc.rs
```diff
@@ -1509,6 +1510,90 @@ impl ConversationHttp {
...
+                // no bootstrap nodes found, unable to determine health.
+                let response = HttpResponseType::GetHealthQueryError(
+                    response_metadata.clone(),
+                    "Couldn't obtain stats on any bootstrap peers, unable to determine health"
+                        .to_string(),
+                );
+                return response.send(http, fd).map(|_| ());
+            }
+
+            if block_height >= (max_height - 1) {
```
jcnelson, line 1574: Minor style not -- parentheses aren't needed here.

src/net/http.rs
jcnelson: Also, `GetHealthError` reports that the node is still bootstrapping, right?  Why is this an internal server error if the node is still bootstrapping?  Wouldn't this still be a 200 OK, since the node's bootstrapping status is an application-level concern above the HTTP transport layer?

src/net/rpc.rs
```diff
@@ -3397,6 +3502,186 @@ mod test {
...
+            },
+        );
+    }
+
+    /// The block height of the peer server and its peer should match, so `matches_peers` is true.
+    #[test]
+    #[ignore]
+    fn test_rpc_get_health_true() {
+        test_rpc(
+            "test_rpc_get_health",
```
jcnelson, line 3566: This string is going to need to be different for each unit test, since it ultimately gets used to create a directory.  Tests running in parallel will clobber each other.  Also, the numbers below are port numbers -- they need to be different from other unit tests' port numbers in order to avoid `EADDRINUSE` errors.

src/net/rpc.rs
jcnelson: The "online-ness" of a node for the purposes of deciding whether or not it qualifies as a valid bootstrap peer is not represented by `neighbor_block_stats.status`.  This field represents what the local node tentatively believes about the status of the remote node.  When bootstrapping, a local node will occasionally think that its remote node has diverged from it -- i.e. it is online, but it reports a different consensus hash.  This can happen if the local node manages to start processing the sortitions in reward cycle N+1 before it has processed the PoX anchor block for reward cycle N.  Moreover, the way this situation is handled depends on whether or not the local node is doing the initial block download -- if so, then the remote node's connection is preserved, but if not, then it's severed.

What would happen from the vantage point of this health endpoint is that a node's bootstrap peers might suddenly appear to disappear, since none of them are "online" anymore.  This probably isn't what you want to report.  Instead, you might want to add a method like this:

```rust
pub fn is_peer_healthy(&self, ibd: bool) -> bool {
   // apply de Morgan's law as you see fit
   (ibd && (self.status == NodeStatus::Online ||  self.status == NodeStatus::Diverged)) || (!ibd && self.status == NodeStatus::Online)
}
```

pavitthrap:
Re-implemented in #3729

PR #2770, opened 2021-07-15
feat(Clarity2) Get burnchain_header_hash by burnchain_height

gregorycoppola:
Fixes https://github.com/blockstack/stacks-blockchain/issues/2663

## Description

This change allows a clarity contract to get the **header hash** of a **burnchain block** using the ***burchain* height**. Currently, you can get the header hash of the burnchain only using the ***Stacks* height**.

## Testing

Tests have been added in the following files:
```
M	src/chainstate/stacks/boot/contract_tests.rs
M	src/vm/tests/contracts.rs
M	src/vm/tests/costs.rs
```

A TODO was added to modify the following integration test. We avoid doing this now because it will require having a way to configure the Stacks Epoch for the Mocknet miner.
```
M	testnet/stacks-node/src/tests/integrations.rs
```


src/vm/database/clarity_db.rs
```diff
@@ -680,6 +713,22 @@ impl<'a> ClarityDatabase<'a> {
             .expect("Failed to get block data.")
     }
```
gregorycoppola, line 715: Note: New implementation.

src/clarity_vm/database/mod.rs
```diff
@@ -23,12 +24,17 @@ use core::StacksEpoch;
...
 
 impl HeadersDB for DBConn {
+    // Is this it?
     fn get_stacks_block_header_hash_for_block(
         &self,
         id_bhh: &StacksBlockId,
     ) -> Option<BlockHeaderHash> {
+        // I think it's this one.
         get_stacks_header_info(self, id_bhh).map(|x| x.anchored_header.block_hash())
     }
```
gregorycoppola, line 34: New implementation

gregorycoppola (commented):
added some pointers to implementations

src/vm/database/clarity_db.rs
```diff
@@ -680,6 +713,22 @@ impl<'a> ClarityDatabase<'a> {
...
+        let consensus_hash = self
+            .headers_db
+            .get_consensus_hash_for_block(&id_bhh)
+            .expect("Failed to get block data.");
+
+        self.burn_state_db
+            .get_burn_header_hash_using_consensus_hash(burnchain_block_height, &consensus_hash)
+            .expect("Failed to get block data.")
+    }
+
```
kantai, line 731: This looks like the right approach to me!

src/clarity_vm/database/mod.rs
```diff
@@ -146,6 +162,21 @@ impl BurnStateDB for SortitionDBConn<'_> {
...
+    ) -> Option<BurnchainHeaderHash> {
+        let bt = backtrace::Backtrace::new();
+        warn!("look2 {:?}", bt);
+        let db_handle = SortitionHandleConn::open_reader_consensus(self, consensus_hash).ok()?;
+        match db_handle.get_block_snapshot_by_height(height as u64) {
+            Ok(Some(x)) => Some(x.burn_header_hash),
+            _ => return None,
+        }
+    }
+
```
kantai, line 179: I think this is right approach. You should be more careful about the error handling here. Basically, we want the "burn header hash" to:

* Return `Some(burn_header_hash)` if the given height is `0 <= height < current_burn_block_height`
* Return `None` if the given height is `>= current_burn_block_height`

Anything else should _panic_ (including errors from the SortitionDB: any such error means the database is corrupt, and the node should not be trying to continue processing Stacks transactions).

testnet/stacks-node/src/tests/integrations.rs
```diff
@@ -362,12 +363,26 @@ fn integration_test_get_info() {
                 assert_eq!(
                     chain_state.clarity_eval_read_only(
                         burn_dbconn, bhh, &contract_identifier, "(test-6)"),
-                    Value::some(Value::buff_from(last_burn_header).unwrap()).unwrap());
+                    Value::some(Value::buff_from(last_burn_header.clone()).unwrap()).unwrap());
                 assert_eq!(
                     chain_state.clarity_eval_read_only(
                         burn_dbconn, bhh, &contract_identifier, "(test-7)"),
                     Value::some(Value::buff_from(last_vrf_seed).unwrap()).unwrap());
```
gregorycoppola, line 371: this is the test so far

src/vm/docs/mod.rs
```diff
@@ -1346,14 +1346,20 @@ const GET_BLOCK_INFO_API: SpecialAPI = SpecialAPI {
     description: "The `get-block-info?` function fetches data for a block of the given block height. The
```
pavitthrap, line 1346: I think you should clarify that the block height input might refer to either the burn chain or the. stacks blockchain, depending on the property

pavitthrap (commented):
You should add tests to type checker and arithmetic checker, as well as `vm/tests/contracts.rs`. Trying seeing where else `get-block-info?` is tested.

src/vm/docs/mod.rs
gregorycoppola: Aaron suggested making a new function called "get-burn-block-info?" instead.

Do you think it would be easier to make the docs if we did that?

https://github.com/blockstack/stacks-blockchain/issues/2663#issuecomment-866154516

src/vm/database/clarity_db.rs
```diff
@@ -106,6 +106,10 @@ pub trait BurnStateDB {
         height: u32,
         sortition_id: &SortitionId,
     ) -> Option<BurnchainHeaderHash>;
+    fn get_burn_header_hash_using_canonical_sortition(
```
pavitthrap, line 109: should this function be named something more like `get_burn_header_hash_using_height` instead?

src/vm/docs/mod.rs
pavitthrap: Oh - yes, I definitely prefer the approach of putting this functionality into a separate clarity function, I think that is way less prone to user error

pavitthrap (commented):
Agree with Aaron's idea of putting this functionality into `get-burn-block-info?`

Also, need to add type checker & arithmetic checker tests still I think

src/vm/docs/mod.rs
friedger: Yes. In the future `get-burn-block-info?` could also take `stacks-tip` and return the stacks tip.

src/vm/database/clarity_db.rs
```diff
@@ -101,11 +101,34 @@ pub trait BurnStateDB {
...
     fn get_pox_reward_cycle_length(&self) -> u32;
     fn get_pox_rejection_fraction(&self) -> u64;
+
+    /// Returns the header hash of the burnchain block at burnchain height `height`
+    /// using the sortition `sortition_id`.
+    ///
+    /// Returns None if `height` is negative or greater than the canonical burn
+    /// chain height.
+    ///
+    /// Panics if there is an error with the underlying database connection or
```
pavitthrap, line 111: Do you want to add a markdown section for "Panics"?

src/vm/database/clarity_db.rs
pavitthrap: what were your thoughts here?

src/vm/docs/mod.rs
```diff
@@ -1354,6 +1354,10 @@ and block times are accurate only to within two hours. See [BIP113](https://gith
 
 The `header-hash`, `burnchain-header-hash`, `id-header-hash`, and `vrf-seed` properties return a 32-byte buffer.
 
+`header-hash` returns the header hash of a Stacks node, given a Stacks chain height.
+
+`burnchain-header-hash` returns header hash of a burnchain (Bitcoin) node, given a Stacks chain height.
```
pavitthrap, line 1359: nit: returns *the* header hash

src/vm/docs/mod.rs
```diff
@@ -1364,6 +1368,23 @@ The `id-header-hash` is the block identifier value that must be used as input to
 "
 };
 
+const GET_BURN_BLOCK_INFO_API: SpecialAPI = SpecialAPI {
+    input_type: "BlockInfoPropertyName, BlockHeightInt",
```
pavitthrap, line 1372: Should we name the second arg `BlockHeightUint` instead?

src/vm/functions/database.rs
```diff
@@ -497,3 +497,52 @@ pub fn special_get_block_info(
...
+    // Handle the block-height input arg clause.
+    let height_eval = eval(&args[1], env, context)?;
+    let height_value = match height_eval {
+        Value::UInt(result) => Ok(result),
+        x => Err(CheckErrors::TypeValueError(TypeSignature::UIntType, x)),
+    }?;
+
+    let height_value = match u32::try_from(height_value) {
+        Ok(result) => result,
+        _ => return Ok(Value::none()),
```
pavitthrap, line 528: Do we want to be returning an error here? Returning Value::none() is not too informative of what went wrong?

src/vm/tests/contracts.rs
```diff
@@ -164,6 +172,89 @@ fn test_get_block_info_eval() {
     }
 }
 
+/// `BurnBlockTestDB` is a mocked out implementation of `BurnStateDB` for test.
```
pavitthrap, line 175: Could we use `TestSimBurnStateDB` here?

src/vm/functions/database.rs
gregorycoppola: This is copy-pasted.. I guess the assumption was that we would never overflow u32 in practice. I.e., if `result` is u128, and we can't cast to u32, then it is bigger than u32::MAX.

Can we assume that?

src/vm/docs/mod.rs
gregorycoppola: I copy/pasted this from "get-block-info".

You could argue that the "Height" implies that it is unsigned and so "BlockHeightUInt" is redundant.

src/vm/docs/mod.rs
gregorycoppola: done

src/vm/database/clarity_db.rs
gregorycoppola: Oh yeah.. if that's our convention I should.. you know it's going to add a lot of vertical height to the comments if we do this. The docs also suggest a special section for "#Errors":
https://rust-lang.github.io/api-guidelines/documentation.html

src/vm/database/clarity_db.rs
gregorycoppola: oh whoops.. Well, the reason it is "using_canonical_sortition" is because the "canonical sortiiton" is the hidden argument. That is, the `height` part is in the signature. Compare with `get_burn_header_hash`. The idea is that these two functions are similar but this one uses an implicit sortition. `get_burn_header_hash` also doesn't mention height.

src/vm/docs/mod.rs
gregorycoppola: Did this.. turned out to be a big delta but worth it!

src/clarity_vm/database/mod.rs
gregorycoppola: did this

gregorycoppola (commented):
@pavitthrap thanks for the comments.

src/vm/tests/contracts.rs
pavitthrap: I mean I guess I'm wondering if another mock of `BurnStateDB` is strictly necessary - if it seems necessary to you, then go ahead with `BurnBlockTestDB`.

src/vm/docs/mod.rs
pavitthrap: I see your point. 

Looking more broadly at this file though, it seems that `input_type` often just uses the actual clarity types, leaving the more descriptive names to the `signature` field. I think I would prefer that convention here.

src/vm/database/clarity_db.rs
pavitthrap: Ya, I think that makes sense to me, since that what the convention would be with rustdoc.

src/vm/functions/database.rs
pavitthrap: Right, but this branch represents the error case of overflow, so I feel like the result should be a corresponding error.

src/vm/database/clarity_db.rs
pavitthrap: ahh I see, sounds good to me.

pavitthrap (commented):
Looks good, just awaiting your replies on the ongoing comment threads we have.

src/vm/database/clarity_db.rs
gregorycoppola: ok.. can start doing this

src/vm/docs/mod.rs
gregorycoppola: I was actually copying `get-block-info`.. should I just change both?

src/vm/functions/database.rs
gregorycoppola: This would have to be a user facing error? which one would it be? https://github.com/blockstack/stacks-blockchain/blob/4650aae6ffb111d487ce324fd8591c529e60c24e/src/vm/analysis/errors.rs#L27

src/vm/tests/contracts.rs
gregorycoppola: it's necessary.. we need a mock otherwise we would have to spin up a whole database.. which is a good integration test, but wouldn't be a unit test.

also, we can't use NullBurnStateDB because it always returns None and then we can't do a meaningufl test

gregorycoppola (commented):
@pavitthrap  I made the "#Panics" changes.. had a few questions.

src/vm/docs/mod.rs
pavitthrap: I think that would be good, it would make the documentation more consistent overall

src/vm/functions/database.rs
pavitthrap: I think `CheckErrors::ValueTooLarge` makes sense here

src/vm/docs/mod.rs
gregorycoppola: done

gregorycoppola (commented):
@pavitthrap just the "None" vs "Error" issue remaining

src/vm/functions/database.rs
pavitthrap: You're right, that error type doesn't make sense. 

I'm alright with returning `Value::none()` here.

pavitthrap (approved):
LGTM! Thanks for addressing all the feedback :)

src/vm/functions/database.rs
```diff
@@ -497,3 +497,55 @@ pub fn special_get_block_info(
 
     Ok(Value::some(result)?)
 }
+
+pub fn special_get_burn_block_info(
+    args: &[SymbolicExpression],
+    env: &mut Environment,
+    context: &LocalContext,
+) -> Result<Value> {
+    // (get-block-info? property-name block-height-int)
```
jcnelson, line 506: comment typo -- did you mean "`get-burn-block-info?`"?

src/vm/analysis/type_checker/natives/mod.rs
```diff
@@ -521,6 +521,27 @@ fn check_get_block_info(
...
+fn check_get_burn_block_info(
+    checker: &mut TypeChecker,
+    args: &[SymbolicExpression],
+    context: &TypingContext,
+) -> TypeResult {
+    check_arguments_at_least(2, args)?;
+
+    let block_info_prop_str = args[0]
+        .match_atom()
+        .ok_or(CheckError::new(CheckErrors::GetBlockInfoExpectPropertyName))?;
```
jcnelson, line 533: Up to you how you decide to do this, but do you want a separate `CheckErrors` error for burn block properties not being known to the system?

src/vm/database/clarity_db.rs
```diff
@@ -680,6 +719,22 @@ impl<'a> ClarityDatabase<'a> {
...
+    /// chain height.
+    ///
+    /// # Panics
+    /// Panics if the underlying database is not available.
+    pub fn get_burnchain_block_header_hash_for_burnchain_height(
+        &mut self,
+        burnchain_block_height: u32,
+    ) -> Option<BurnchainHeaderHash> {
+        self.burn_state_db
+            .get_burn_header_hash_using_canonical_sortition(burnchain_block_height)
```
kantai, line 735: I don't think that it is sufficient to use the canonical sortition for looking up the burnchain block at a given burnchain height here. The way that these lookups are performed is that the provided "sortition" defines the burnchain fork that the lookup is performed from. The canonical sortition _may not_ be the same fork that the currently-being-evaluated Stacks block was contained within (I believe that with the way the ChainsCoordinator currently works, this _is_ the case, but it's not a guarantee that the ChainsCoordinator explicitly provides).

For example, the burnchain could look like this:

```
A -> B -> C -> D
 \_> B' -> C'
```

Where the currently-being-evaluated Stacks block was elected in `C'`, but the canonical sortition is `D`. We'd want `get-burn-block-info u1` to return `B'` in this case and _not_ `B`.

The second important thing here is that the height supplied to `get-burn-block-info` _must_ be strictly less than the current burn block height. This is because miners have to evaluate the block _before_ the burn block is mined, so they could not _know_ that `(get-burn-block-info u2)` should return `C'`. Basically, if the supplied height is `>= current_burn_block_height` it should return `none`.

jcnelson: Beat me to it.

You're going to need to track the `StacksBlockId` of the chain tip at which this code is running, and use *that* to look up the associated burnchain block hash.  Then, if the VM is in the process of building block `C'` off of `B'`, then this method would be able to report the hash of `B'` as the hash of the block at height 2.  Correspondingly, if the VM is in the process of building block `C` off of `B`, then this method would be able to report the hash of `B` as the hash of the block at height 2.

src/vm/database/clarity_db.rs
```diff
@@ -101,11 +101,36 @@ pub trait BurnStateDB {
...
+    /// Returns None if `height` is negative or greater than the canonical burn
+    /// chain height.
+    ///
+    /// # Panics
+    /// Panics if there is an error with the underlying database connection or
+    /// query.
+    fn get_burn_header_hash_using_canonical_sortition(
+        &self,
+        height: u32,
+    ) -> Option<BurnchainHeaderHash>;
```
jcnelson, line 132: You should probably just drop this method altogether.  This method is going to need to take the `StacksBlockId` of the chain tip that the VM is currently working on, in addition to the burnchain block height.  It would look something like this:

```
// just an example; feel free to rename
fn get_burn_header_hash_from_tip(&self, stacks_tip: &StacksBlockId, burn_height: u32) -> Option<BurnchainHeaderHash>)
```

src/vm/database/clarity_db.rs
```diff
@@ -680,6 +716,41 @@ impl<'a> ClarityDatabase<'a> {
...
 
+    /// Gets the `StacksBlockId` for the *current* chain tip Stacks block height, converts this
+    /// `StacksBlockId` to a `SortitionId` and then gets the `BurnchainHeaderHash` from the
+    /// `BurnStateDB` using this `SortitionId`.
+    ///
+    /// Returns Some() if `0 <= height < current_burnblock_height`, and None otherwise.
+    pub fn get_burnchain_block_header_hash_for_burnchain_height(
+        &mut self,
+        burnchain_block_height: u32,
+    ) -> Option<BurnchainHeaderHash> {
```
gregorycoppola, line 727: @kantai said "Basically, if the `supplied height is >= current_burn_block_height` it should return none."

I conceptually get why this is, I think, because we are "building the block" and the snapshot wouldn't exist in sqlite3 yet.

But, what is the literal API call for `current_burn_block_height`.

src/vm/database/clarity_db.rs
```diff
@@ -680,6 +716,41 @@ impl<'a> ClarityDatabase<'a> {
...
+        let current_stacks_height = self.get_current_block_height();
+        let id_bhh = self.get_index_block_header_hash(current_stacks_height);
+
+        // Is this block in the database?
+        // Should this return an error/panic if it fails?
+        let consensus = match self.headers_db.get_consensus_hash_for_block(&id_bhh) {
+            None => {
+                return None;
```
gregorycoppola, line 735: Should these None cases 1) propagate None, 2) return Error, 3) panic?

src/vm/database/clarity_db.rs
gregorycoppola, line 727: @kantai @jcnelson can you review the logic in this method before I re-write the tests?

gregorycoppola (commented):
@kantai @jcnelson I'm just asking for an *intermediate* review of the logic of `ClarityDatabase::get_burnchain_block_header_hash_for_burnchain_height` before I re-write the tests and clean up the PR.

The reset of the PR doesn't need/isn't ready for review.

jcnelson:
Blocked on getting a mini-LGTM on the logic for obtaining a burnchain header hash off of a particular chain tip. CC @jcnelson @kantai

src/vm/database/clarity_db.rs
```diff
@@ -680,6 +716,41 @@ impl<'a> ClarityDatabase<'a> {
...
+    /// Gets the `StacksBlockId` for the *current* chain tip Stacks block height, converts this
+    /// `StacksBlockId` to a `SortitionId` and then gets the `BurnchainHeaderHash` from the
+    /// `BurnStateDB` using this `SortitionId`.
+    ///
+    /// Returns Some() if `0 <= height < current_burnblock_height`, and None otherwise.
+    pub fn get_burnchain_block_header_hash_for_burnchain_height(
+        &mut self,
+        burnchain_block_height: u32,
+    ) -> Option<BurnchainHeaderHash> {
+        let current_stacks_height = self.get_current_block_height();
```
kantai, line 728: Is this the current Stack height or the current burn height? Each Stacks block has both (and they aren't equal), i.e., a Stacks block could be Stacks height 100, but burn height 667000.

src/vm/database/clarity_db.rs
```diff
@@ -680,6 +716,41 @@ impl<'a> ClarityDatabase<'a> {
...
+        let id_bhh = self.get_index_block_header_hash(current_stacks_height);
```
kantai, line 729: You should use the _parent_ ID block hash, rather than the current stacks height. The reason for that is that the current block hash is evaluated with a placeholder by the miner, so its consensus hash, etc. isn't contained in the headers db yet (which only contains information about already-mined blocks).

src/vm/database/clarity_db.rs
kantai: It should panic. _Except_ in the case that the current block height is 0. That is, a parent block should _always_ have a consensus hash (unless it's at height 0 and there is no parent).

src/vm/database/clarity_db.rs
```diff
@@ -680,6 +716,41 @@ impl<'a> ClarityDatabase<'a> {
...
+        // Is this block in the database?
+        // Should this return an error/panic if it fails?
+        let consensus = match self.headers_db.get_consensus_hash_for_block(&id_bhh) {
+            None => {
+                return None;
+            }
+            Some(consensus) => consensus,
+        };
+
+        // Should this return an error/panic if it fails?
```
kantai, line 740: Panic

src/vm/database/clarity_db.rs
```diff
@@ -680,6 +716,41 @@ impl<'a> ClarityDatabase<'a> {
             .expect("Failed to get block data.")
     }
 
+    /// Gets the `StacksBlockId` for the *current* chain tip Stacks block height, converts this
+    /// `StacksBlockId` to a `SortitionId` and then gets the `BurnchainHeaderHash` from the
+    /// `BurnStateDB` using this `SortitionId`.
+    ///
+    /// Returns Some() if `0 <= height < current_burnblock_height`, and None otherwise.
```
kantai, line 723: I believe this returns `Some() if first_burn_block_height <= height < current_burn_block_height` and None otherwise. The sortition DB _does not_ track information about burn blocks before `first_burn_block_height`.

src/vm/database/clarity_db.rs
```diff
@@ -680,6 +716,41 @@ impl<'a> ClarityDatabase<'a> {
...
+            .burn_state_db
+            .get_block_snapshot_from_consensus_hash(&consensus)
+        {
+            None => {
+                return None;
+            }
+            Some(snapshot) => snapshot,
+        };
+        self.burn_state_db
+            .get_burn_header_hash(burnchain_block_height, &snapshot.sortition_id)
```
kantai, line 751: As part of this PR, can you add a rustdoc to the interface method `get_burn_header_hash` that indicates that it returns the burnchain header has for the given burn block height, as queried from the given Sortition ID?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -3695,6 +3705,7 @@ impl<'a> SortitionHandleTx<'a> {
             &snapshot.accumulated_coinbase_ustx.to_string(),
         ];
 
+        // Note: insert into snap shots.. is this done by the time that we have this block?
```
jcnelson, line 3708: "snapshots"

Yes, this happens once we have processed and stored the bitcoin block and extracted and processed all of its stacks-specific transactions.  There is one snapshot per sortition, which is a view calculated over the set of block-commits in a bitcoin block.

src/clarity_vm/database/mod.rs
```diff
@@ -140,12 +158,32 @@ impl BurnStateDB for SortitionDBConn<'_> {
         sortition_id: &SortitionId,
     ) -> Option<BurnchainHeaderHash> {
         let db_handle = SortitionHandleConn::open_reader(self, &sortition_id).ok()?;
+
+        let current_height = self.get_burn_block_height(sortition_id) {
+            None => { return None; }
+            Some(height) => height,
+        }
+
+        if height < 0 || height >= current_height {
```
jcnelson, line 167: How is `height < 0` if it's a `u32`?

src/vm/database/clarity_db.rs
jcnelson: This code does not work in the presence of Bitcoin forks.  The sortition DB and burnchain DB will process and store state for multiple different Bitcoin blocks at the same height, should a Bitcoin reorg happen.

If you want to query a burnchain block hash at a particular height, you need to know which burnchain fork you're on.  You can do this given the Stacks chain tip's ID block hash -- i.e. the ID block hash of the parent block that this Clarity code is building on top of.

The `id_bhh` variable you load below should be the _same ID hash_ that was passed to the inner `MarfedKV`'s `begin()` method as `current`.  This, in turn, is the ID hash passed as `current` to `ClarityInstance::begin_block()`.  You should *not* be trying to deduce `id_bhh` from a block height.

src/vm/database/clarity_db.rs
jcnelson: Per my other comment, this would be the same as the value of `current` passed into `ClarityInstance::begin_block()`.

src/vm/docs/mod.rs
```diff
@@ -1364,6 +1368,23 @@ The `id-header-hash` is the block identifier value that must be used as input to
...
+    description: "The `get-burn-block-info?` function fetches data for a block of the given *burnchain* block height. The
+value and type returned are determined by the specified `BlockInfoPropertyName`. If the provided `block-height` does
+not correspond to an existing block prior to the current block, the function returns `none`. The only available property
+name so far is `header-hash`.
+
+The `header-hash` property returns a 32-byte integer representing the header hash of the burnchain block at
+burnchain height `block-height`.
+",
+    example: "
+(get-burn-block-info? header-hash u0) ;; Returns (some 0xe67141016c88a7f1203eca0b4312f2ed141531f59303a1c267d7d83ab6b977d8)
```
jcnelson, line 1384: I'm not sure `u0` is a good example here.

src/vm/docs/mod.rs
```diff
@@ -1364,6 +1368,23 @@ The `id-header-hash` is the block identifier value that must be used as input to
 "
 };
 
+const GET_BURN_BLOCK_INFO_API: SpecialAPI = SpecialAPI {
+    input_type: "BlockInfoPropertyName, uint",
+    output_type: "(optional buff)",
+    signature: "(get-burn-block-info? prop-name block-height)",
+    description: "The `get-burn-block-info?` function fetches data for a block of the given *burnchain* block height. The
+value and type returned are determined by the specified `BlockInfoPropertyName`. If the provided `block-height` does
+not correspond to an existing block prior to the current block, the function returns `none`. The only available property
```
jcnelson, line 1377: It will also return `none` for any burnchain block value prior to the start of the Stacks chain (i.e. 666050).  We can discuss whether or not this is desirable, but that's how the code currently behaves.

testnet/stacks-node/src/tests/integrations.rs
```diff
@@ -49,6 +49,8 @@ const CALL_READ_CONTRACT: &'static str = "
     (ok (contract-call? .other f2 u5)))
 ";
 
+// TODO(gregorycoppola) Add "burn-block-info" section to this test, once a way to configure the
+// Mocknet default Stacks epoch has been decided.
```
jcnelson, line 53: Specific to this test, can you verify that `burn-block-info` will continue to work even if there was no Stacks block chosen at that height?  Also, verify that we get `(none)` for any block height prior to the first burnchain block height.

gregorycoppola (commented):
OK. I implemented the algo changes here

hstove:
Oh wow this is so handy, thanks! Very critical for validating burnchain transactions.

hstove:
Question on this - if the burnchain block at that height does not correspond to a stacks block (i.e. because of a flash block), will this return `none`? From reading the code, it seems like it would return `none`, but I'm not sure.

jcnelson:
I'm going to take over working on this.

jcnelson:
This is now ready for review @kantai @gregorycoppola @lgalabru @obycode

clarity/src/vm/functions/database.rs
```diff
@@ -768,3 +768,63 @@ pub fn special_get_block_info(
...
 }
+
+/// Interprets `args` as variables `[property_name, burnblock_height]`, and returns
+/// a property value determined by `property_name`:
+/// - `header_hash` returns the burnblock header hash at `burnblock_height`
+///
+/// # Errors:
+/// - CheckErrors::IncorrectArgumentCount if there aren't 2 arguments.
+/// - CheckErrors::GetBlockInfoExpectPropertyName if `args[0]` isn't a ClarityName.
+/// - CheckErrors::NoSuchBurnBlockInfoProperty if `args[0]` isn't a BurnBlockInfoProperty.
```
obycode, line 779:
```suggestion
/// - CheckErrors::NoSuchBurnBlockInfoProperty if `args[0]` isn't a BurnBlockInfoProperty.
/// - CheckErrors::TypeValueError if `args[1]` isn't a uint
```

clarity/src/vm/analysis/type_checker/natives/mod.rs
```diff
@@ -524,6 +524,30 @@ fn check_get_block_info(
...
 
+// # Errors
+// - `CheckErrors::GetBurnBlockInfoExpectPropertyName` when `args[0]` is not a valid `ClarityName`.
+// - `CheckErrors::NoSuchBlockInfoProperty` when `args[0]` doesnot name a `BurnBlockInfoProperty`.
+fn check_get_burn_block_info(
+    checker: &mut TypeChecker,
+    args: &[SymbolicExpression],
+    context: &TypingContext,
+) -> TypeResult {
+    check_arguments_at_least(2, args)?;
```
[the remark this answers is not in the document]
jcnelson: So, `special_get_burn_block_info()` uses `check_argument_count()`, meaning that passing the wrong number of arguments would be a runtime error, not an analysis error.  Is this desired @gregorycoppola @kantai?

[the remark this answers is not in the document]
jcnelson: To be clear, we can totally change it for those functions too as part of 2.1; we just can't do it today ;)  Happy to use `check_argument_count()` here.

clarity/src/vm/analysis/type_checker/tests/mod.rs
```diff
@@ -116,6 +116,42 @@ fn test_get_block_info() {
...
+
+    for (good_test, expected) in good.iter().zip(expected.iter()) {
+        assert_eq!(
+            expected,
+            &format!("{}", type_check_helper(&good_test).unwrap())
+        );
+    }
+
+    for (bad_test, expected) in bad.iter().zip(bad_expected.iter()) {
+        warn!("bad_test {:?}", bad_test);
```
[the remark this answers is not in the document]
jcnelson: Removed

clarity/src/vm/analysis/type_checker/natives/mod.rs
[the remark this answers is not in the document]
jcnelson: Yup, agreed.  Using `check_argument_count()` here.  Also: https://github.com/stacks-network/stacks-blockchain/issues/3117

clarity/src/vm/analysis/type_checker/natives/mod.rs
```diff
@@ -524,6 +524,30 @@ fn check_get_block_info(
     Ok(TypeSignature::new_option(block_info_prop.type_result())?)
 }
 
+// # Errors
+// - `CheckErrors::GetBurnBlockInfoExpectPropertyName` when `args[0]` is not a valid `ClarityName`.
+// - `CheckErrors::NoSuchBlockInfoProperty` when `args[0]` doesnot name a `BurnBlockInfoProperty`.
```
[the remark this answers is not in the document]
jcnelson: Done

clarity/src/vm/docs/mod.rs
```diff
@@ -1384,6 +1388,25 @@ The `id-header-hash` is the block identifier value that must be used as input to
 "
 };
 
+const GET_BURN_BLOCK_INFO_API: SpecialAPI = SpecialAPI {
+    input_type: "BlockInfoPropertyName, uint",
```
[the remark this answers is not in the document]
jcnelson: Yup, fixed

clarity/src/vm/functions/database.rs
jcnelson: Fixed

obycode (approved):

Issue #2771, opened 2021-07-15
Bitcoin reorgs can impact run loop synchronization, stalling bitcoin downloads


kantai:
A specific interaction between `target_block_height` and bitcoin reorgs can cause a node to stall in downloading Bitcoin blocks. The only remedy if a node gets caught in this situation is to restart the node with a fresh chainstate (in theory, the chainstate could be manually fixed, but it would require changes across a few different SQLite dbs).

The interaction in question:

https://github.com/blockstack/stacks-blockchain/blob/master/src/burnchains/burnchain.rs#L1223

Here, the `sync_with_indexer` function will correctly "shorten" the range for block downloads if passed a target block height. However, this does _not_ stop the SPV client from downloading the headers from the original range. Once it has done this, subsequent invocations of `sync_with_indexer` will set the `start_height` initially to the end of that range. Naively, this would mean the indexer would _never_ download, parse and process the blocks in that shortened portion of the range. In normal operation (and even in the presence of short-lived Bitcoin forks), this isn't an issue because the current height in `burnchain_db` is used to reduce the `start_height`. However, in the event of a bitcoin reorg, the `burnchain_db` height _may already_ be higher than the blocks needed to download. This would cause the indexer to never download those blocks.

Solving this could be done in a number of different ways:

1. On reorgs, make entries in the `burnchain_db` as `reorged`, and ignore them in `canonical_burnchain_tip` lookups. _(I think this is a bad approach)_.
2. Apply `target_block_height` to the SPV client's `sync_headers` command.
3. Track which blocks have been processed by the burnchain_db in the SPV headers db. Use that information to determine `start_height`.

gregorycoppola:
Fixed.

PR #2773, opened 2021-07-16
Fix: 2771

lgalabru:
Address #2771.
The test added consists in producing a canonical chain with 210 blocks. Once done, we invalidate the block 208, and instead of rebuilding directly a longer fork with N blocks (at done in the `bitcoind_forking_test`) we slowly add some more blocks.
Without the patch, this behavior ends up crashing the node with errors like:

```
WARN [1626791307.078061] [src/chainstate/coordinator/mod.rs:535] [chains-coordinator] ChainsCoordinator: could not retrieve  block burnhash=40bdbf0dda349642bdf4dd30dd31af4f0c9979ce12a7c17485245d0a6ddd970b
WARN [1626791307.078098] [src/chainstate/coordinator/mod.rs:308] [chains-coordinator] Error processing new burn block: NonContiguousBurnchainBlock(UnknownBlock(40bdbf0dda349642bdf4dd30dd31af4f0c9979ce12a7c17485245d0a6ddd970b))
```

And the burnchain db end up in the same state we ended up while investing 2771.

With this patch, the node is able to entirely register this new canonical fork, and then able to make some progress.

### DIFF
--- a/.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
+++ b/.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
@@ -21,6 +21,7 @@ RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::bitcoind_
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::liquid_ustx_integration
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::stx_transfer_btc_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::bitcoind_forking_test
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::should_fix_2771
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::pox_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::bitcoin_regtest::bitcoind_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::should_succeed_handling_malformed_and_valid_txs
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -1187,9 +1187,9 @@ impl Burnchain {
         // get latest headers.
         debug!("Sync headers from {}", sync_height);
 
+        // fetch all headers, no matter what
         let mut end_block = indexer.sync_headers(sync_height, end_height)?;
 
-        // fetch all headers, no matter what
         let mut start_block = sync_height;
         if db_height < start_block {
             start_block = db_height;
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -1184,10 +1184,10 @@ impl Burnchain {
             None
         };
 
-        let mut end_block = indexer.sync_headers(sync_height, end_height)?;
-
         // get latest headers.
         debug!("Sync headers from {}", sync_height);
+
+        let mut end_block = indexer.sync_headers(sync_height, end_height)?;
 
         // fetch all headers, no matter what
         let mut start_block = sync_height;
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -1170,8 +1170,8 @@ impl Burnchain {
         let db_height = burn_chain_tip.block_height;
 
         // handle reorgs
-        let (mut sync_height, did_reorg) = Burnchain::sync_reorg(indexer)?;
-        if did_reorg {
+        let (sync_height, did_reorg) = Burnchain::sync_reorg(indexer)?;
+        let end_height = if did_reorg {
             // a reorg happened
             warn!(
                 "Dropping headers higher than {} due to burnchain reorg",
@@ -1179,16 +1179,17 @@ impl Burnchain {
             );
             indexer.drop_headers(sync_height)?;
             // Patch issue #2771
-            if let Some(target_block_height) = target_block_height_opt {
-                sync_height = target_block_height;
-            }
-        }
+            target_block_height_opt
+        } else {
+            None
+        };
+
+        let mut end_block = indexer.sync_headers(sync_height, end_height)?;
 
         // get latest headers.
         debug!("Sync headers from {}", sync_height);
 
         // fetch all headers, no matter what
-        let mut end_block = indexer.sync_headers(sync_height, None)?;
         let mut start_block = sync_height;
         if db_height < start_block {
             start_block = db_height;
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -891,7 +891,7 @@ impl Burnchain {
     }
 
     /// Determine if there has been a chain reorg, given our current canonical burnchain tip.
-    /// Return the new chain tip and a boolen signaling the presence of a reorg
+    /// Return the new chain tip and a boolean signaling the presence of a reorg
     fn sync_reorg<I: BurnchainIndexer>(indexer: &mut I) -> Result<(u64, bool), burnchain_error> {
         let headers_path = indexer.get_headers_path();
 
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -892,7 +892,7 @@ impl Burnchain {
 
     /// Determine if there has been a chain reorg, given our current canonical burnchain tip.
     /// Return the new chain tip
-    fn sync_reorg<I: BurnchainIndexer>(indexer: &mut I) -> Result<u64, burnchain_error> {
+    fn sync_reorg<I: BurnchainIndexer>(indexer: &mut I) -> Result<(u64, bool), burnchain_error> {
         let headers_path = indexer.get_headers_path();
 
         // sanity check -- what is the height of our highest header
@@ -905,7 +905,7 @@ impl Burnchain {
         })?;
 
         if headers_height == 0 {
-            return Ok(0);
+            return Ok((0, false));
         }
 
         // did we encounter a reorg since last sync?  Find the highest common ancestor of the
@@ -921,10 +921,10 @@ impl Burnchain {
                 "Burnchain reorg detected: highest common ancestor at height {}",
                 reorg_height
             );
-            return Ok(reorg_height);
+            return Ok((reorg_height, true));
         } else {
             // no reorg
-            return Ok(headers_height);
+            return Ok((headers_height, false));
         }
     }
 
@@ -959,7 +959,7 @@ impl Burnchain {
 
         // handle reorgs
         let orig_header_height = indexer.get_headers_height()?; // 1-indexed
-        let sync_height = Burnchain::sync_reorg(indexer)?;
+        let (sync_height, _) = Burnchain::sync_reorg(indexer)?;
         if sync_height + 1 < orig_header_height {
             // a reorg happened
             warn!(
@@ -1170,15 +1170,18 @@ impl Burnchain {
         let db_height = burn_chain_tip.block_height;
 
         // handle reorgs
-        let orig_header_height = indexer.get_headers_height()?; // 1-indexed
-        let sync_height = Burnchain::sync_reorg(indexer)?;
-        if sync_height + 1 < orig_header_height {
+        let (mut sync_height, did_reorg) = Burnchain::sync_reorg(indexer)?;
+        if did_reorg {
             // a reorg happened
             warn!(
                 "Dropping headers higher than {} due to burnchain reorg",
                 sync_height
             );
             indexer.drop_headers(sync_height)?;
+            // Patch issue #2771
+            if let Some(target_block_height) = target_block_height_opt {
+                sync_height = target_block_height;
+            }
         }
 
         // get latest headers.
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -1349,14 +1349,14 @@ impl Burnchain {
                         }
 
                         if is_mainnet {
-                            if block_height == STACKS_2_0_LAST_BLOCK_TO_PROCESS {
+                            if last_processed.block_height == STACKS_2_0_LAST_BLOCK_TO_PROCESS {
                                 info!("Reached Stacks 2.0 last block to processed, ignoring subsequent burn blocks";
-                                      "block_height" => block_height);
+                                      "block_height" => last_processed.block_height);
                                 continue;
-                            } else if block_height > STACKS_2_0_LAST_BLOCK_TO_PROCESS {
+                            } else if last_processed.block_height > STACKS_2_0_LAST_BLOCK_TO_PROCESS {
                                 debug!("Reached Stacks 2.0 last block to processed, ignoring subsequent burn blocks";
                                        "last_block" => STACKS_2_0_LAST_BLOCK_TO_PROCESS,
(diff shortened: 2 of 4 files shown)


testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1310,6 +1310,73 @@ fn bitcoind_forking_test() {
...
+
+    let mut sort_height = channel.get_sortitions_processed();
+    eprintln!("Sort height: {}", sort_height);
+
+    while sort_height < 210 {
+        next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+        sort_height = channel.get_sortitions_processed();
+        eprintln!("Sort height: {}", sort_height);
+    }
+    // let's query the miner's account nonce:
```
gregorycoppola, line 1358: is this really what this line does?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1310,6 +1310,73 @@ fn bitcoind_forking_test() {
...
+
+    eprintln!("Miner account: {}", miner_account);
+
+    // okay, let's figure out the burn block we want to fork away.
+    warn!("Will trigger re-org at block {}", 208);
```
gregorycoppola, line 1363: can you make a constant for `208` since you use it twice?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1310,6 +1310,73 @@ fn bitcoind_forking_test() {
...
+    // first block wakes up the run loop
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    // first block will hold our VRF registration
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    let mut sort_height = channel.get_sortitions_processed();
+    eprintln!("Sort height: {}", sort_height);
+
+    while sort_height < 210 {
```
gregorycoppola, line 1353: Can you add some more comments about the logic of this test for the benefit of someone who might come across it later?

What are scenario are you setting up? What do we expect to happen?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1310,6 +1310,73 @@ fn bitcoind_forking_test() {
...
+    warn!("Will trigger re-org at block {}", 208);
+    let burn_header_hash_to_fork = btc_regtest_controller.get_block_hash(208);
+    btc_regtest_controller.invalidate_block(&burn_header_hash_to_fork);
+    btc_regtest_controller.build_next_block(1);
+    thread::sleep(Duration::from_secs(5));
+
+    // We invalidated a fork 210 long, and the new one is 207 blocks long.
+    // Let's dispatch the blocks one by one, instead of mining 210-207 blocks
+    while sort_height < 213 {
+        next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
```
gregorycoppola, line 1372: I guess we cannot have this test fail by actually "failing" an assert and crashing.

On the failure mode of this test, it just keeps running.

So, maybe you can explain to the reader how they will know if there is a failure? Or how do we know when the test has succeeded?

src/burnchains/burnchain.rs
```diff
@@ -892,7 +892,7 @@ impl Burnchain {
 
     /// Determine if there has been a chain reorg, given our current canonical burnchain tip.
     /// Return the new chain tip
```
gregorycoppola, line 894: What is the new return type? You added a boolean, right? What is it doing?

src/burnchains/burnchain.rs
```diff
@@ -1327,19 +1343,20 @@ impl Burnchain {
...
 
-                        if burnchain_block.block_height() == 0 {
+                        let block_height = burnchain_block.block_height();
+                        if block_height == 0 {
                             continue;
                         }
 
                         if is_mainnet {
-                            if last_processed.block_height == STACKS_2_0_LAST_BLOCK_TO_PROCESS {
+                            if block_height == STACKS_2_0_LAST_BLOCK_TO_PROCESS {
```
gregorycoppola, line 1352: Is this line tested by any existing tests? This is the logic to stop the chain, right?

src/burnchains/burnchain.rs
```diff
@@ -1186,6 +1185,23 @@ impl Burnchain {
...
+            // is on a smaller fork than the one we just
+            // invalidated. Wait for more blocks.
+            while end_block < db_height {
+                if let Some(ref should_keep_running) = should_keep_running {
+                    if !should_keep_running.load(Ordering::SeqCst) {
+                        return Err(burnchain_error::CoordinatorClosed);
+                    }
+                }
+                let end_height = target_block_height_opt.unwrap_or(0).max(db_height);
+                info!("Burnchain reorg happened at height {} invalidating chain tip {} but only {} headers presents on canonical chain. Retry in 1s", sync_height, db_height, end_block);
```
gregorycoppola, line 1199: 1s or 2s?

gregorycoppola (commented):
Thanks for the fix, @lgalabru !

Looks pretty good assuming that the test passes.. Just had a few requests mostly for comments and readability.

src/burnchains/burnchain.rs
pavitthrap: should we remove this snippet since we know we no longer want to stop at that block height?

testnet/stacks-node/src/tests/neon_integrations.rs
pavitthrap: +1. I think adding a comment to the top with what you put in the pr description would be helpful

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1310,6 +1310,73 @@ fn bitcoind_forking_test() {
     channel.stop_chains_coordinator();
 }
 
+#[test]
+#[ignore]
+fn should_fix_2771() {
```
pavitthrap, line 1315: Do you need to add this test to `Dockerfile.bitcoin-tests`

lgalabru: Good catch, probably. Addressed with latest.

src/burnchains/burnchain.rs
lgalabru: Addressed with latest.

src/burnchains/burnchain.rs
lgalabru: Addressed with latest

testnet/stacks-node/src/tests/neon_integrations.rs
lgalabru: Addressed with latest

testnet/stacks-node/src/tests/neon_integrations.rs
lgalabru: Copy pasta / addressed with latest

testnet/stacks-node/src/tests/neon_integrations.rs
lgalabru: Addressed with latest

testnet/stacks-node/src/tests/neon_integrations.rs
lgalabru: Addressed with latest

src/burnchains/burnchain.rs
lgalabru: I reverted these changes, and will let Jude remove this snippet when he implements the on-chain exit-at-block-height. 
In addition to the `is_mainnet` issue, there is a bug in there where instead of looking at block height of the block we just downloaded, we're using the height of the block that was initially the chain tip when the sync started.
Fixing it does not matter since this code path is dead anyway.

.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
```diff
@@ -21,6 +21,7 @@ RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::bitcoind_
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::liquid_ustx_integration
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::stx_transfer_btc_integration_test
 RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::bitcoind_forking_test
+RUN cargo test -- --test-threads 1 --ignored tests::neon_integrations::should_fix_2771
```
gregorycoppola, line 24: I'm not sure how these `bitcoin-tests` work, but my understanding is is that this test does not actually "fail" on failure.. it just kind of hangs.. or is that wrong?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1307,6 +1307,97 @@ fn bitcoind_forking_test() {
...
+    let reorg_height = 208;
+    warn!("Will trigger re-org at block {}", reorg_height);
+    let burn_header_hash_to_fork = btc_regtest_controller.get_block_hash(reorg_height);
+    btc_regtest_controller.invalidate_block(&burn_header_hash_to_fork);
+    btc_regtest_controller.build_next_block(1);
+    thread::sleep(Duration::from_secs(5));
+
+    // The test here consists in producing a canonical chain with 210 blocks.
+    // Once done, we invalidate the block 208, and instead of rebuilding directly
+    // a longer fork with N blocks (at done in the bitcoind_forking_test)
```
gregorycoppola, line 1388: what is N? maybe put "N=3" or whatever it really is?

gregorycoppola, line 1388: typo: "as done"

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1307,6 +1307,97 @@ fn bitcoind_forking_test() {
...
+    // we slowly add some more blocks.
+    // Without the patch, this behavior ends up crashing the node with errors like:
+    // WARN [1626791307.078061] [src/chainstate/coordinator/mod.rs:535] [chains-coordinator] ChainsCoordinator: could not retrieve  block burnhash=40bdbf0dda349642bdf4dd30dd31af4f0c9979ce12a7c17485245d0a6ddd970b
+    // WARN [1626791307.078098] [src/chainstate/coordinator/mod.rs:308] [chains-coordinator] Error processing new burn block: NonContiguousBurnchainBlock(UnknownBlock(40bdbf0dda349642bdf4dd30dd31af4f0c9979ce12a7c17485245d0a6ddd970b))
+    // And the burnchain db end up in the same state we ended up while investing 2771.
```
gregorycoppola, line 1393: "ends up"

what does "investing" mean?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1307,6 +1307,97 @@ fn bitcoind_forking_test() {
...
+    // With this patch, the node is able to entirely register this new canonical fork, and then able to make some progress.
```
gregorycoppola, line 1394: Maybe just add that it will finish successfully on success.

"make progress and finish successfully"

gregorycoppola (commented):
Hello... Mostly my comments are small wording suggestions for comments.

But, I have one bigger questions, which is: can we really add this test to automated tests? Is that what you're doing in the Dockerfile change?

Because, I thought this test doesn't actually "fail" on failure.. it just kind of hangs.. is that OK with the GitHub actions framework?

.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
lgalabru: When hanging for more for more than N minutes (10 I believe), tests end up aborting and failing.

testnet/stacks-node/src/tests/neon_integrations.rs
lgalabru: N could be 3, 5, 42, 10_000.

lgalabru:
```suggestion
    // a longer fork with N blocks (as done in the bitcoind_forking_test)
```

testnet/stacks-node/src/tests/neon_integrations.rs
lgalabru:
```suggestion
    // And the burnchain db ends up in the same state we ended up while investigating 2771.
```

testnet/stacks-node/src/tests/neon_integrations.rs
lgalabru:
```suggestion
    // With this patch, the node is able to entirely register this new canonical fork, and then able to make progress and finish successfully.
```

.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
lgalabru: https://github.com/blockstack/stacks-blockchain/blob/3382ecdc73dc456f51fe40b5094d321a7f6e8407/testnet/stacks-node/src/tests/neon_integrations.rs#L273

gregorycoppola (approved):
thanks for the fix!

src/burnchains/burnchain.rs
```diff
@@ -1186,6 +1185,23 @@ impl Burnchain {
 
         // fetch all headers, no matter what
         let mut end_block = indexer.sync_headers(sync_height, None)?;
+        if did_reorg && sync_height > 0 {
+            // a reorg happened, and the last header fetched
+            // is on a smaller fork than the one we just
+            // invalidated. Wait for more blocks.
+            while end_block < db_height {
```
pavitthrap, line 1192: my only concern is the use of a while loop here (in case the new blocks take a while to reach the `db_height`. It's my impression that elsewhere in the code base we "fail" and try again - for example, by returning `burnchain_error::TrySyncAgain`.  

I'm still approving, but wondering what your thoughts are on this.

pavitthrap (approved):

src/burnchains/burnchain.rs
lgalabru: There are a few places in the codebase where we block on operations. 
This patch is clearly an edge case that we saw on Bitcoin Testnet (way more instable), and will probably never see on Mainnet. The TrySyncAgain was my initial approach, however on the next retry, we would be loosing the information "we just reorgd and waiting for more blocks from the new canonical fork", and would be trying to fetch blocks using the burnchain tip of the previous fork as a point of reference (which is the real issue).
This approach is limiting the footprint, I think the clean solution was introducing a DB schema modification, where we would update the BurnchainDB when we reorg, and marking the blocks from former canonical fork as such.
This approach is way more invasive and subject to introduce regression, which is something I'd like to avoid for the next few weeks :).

src/burnchains/burnchain.rs
```diff
@@ -891,8 +891,8 @@ impl Burnchain {
     }
 
     /// Determine if there has been a chain reorg, given our current canonical burnchain tip.
-    /// Return the new chain tip
-    fn sync_reorg<I: BurnchainIndexer>(indexer: &mut I) -> Result<u64, burnchain_error> {
+    /// Return the new chain tip and a boolen signaling the presence of a reorg
```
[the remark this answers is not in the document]
lgalabru:
```suggestion
    /// Return the new chain tip and a boolean signaling the presence of a reorg
```

Issue #2774, opened 2021-07-22
Mempool admission logic change: do admission check against unconfirmed state


pavitthrap:
Currently, the admission check for the mempool for a given transaction is executed against the anchored chain tip (not the unconfirmed tip). As a result, a transaction that has a balance that is valid based only on the unconfirmed state would be rejected with the current logic. Right now, the logic only refers to the unconfirmed state if the transaction has an unexpected nonce. 

Reference the function `will_admit_mempool_tx` in `src/chainstate/stacks/db/blocks.rs` for the current logic. 

This issue is to track whether or not doing the admission check against the unconfirmed tip is a change we want.

zone117x:
@blockstack/ux-team can chime in, but my understanding is that if txs cannot be submitted against unconfirmed state, it renders the whole microblock work stream much less useful or interesting. 

It sounds like maybe the tx POST endpoint can default to the current behavior, but also support a query param to enable unconfirmed state?

pavitthrap:
After thinking about it, it seems like allowing the mempool admission check to occur against unconfirmed state could be disruptive. 

Imagine the following sequence of transactions in subsequent microblocks, where user A starts off with 0 STX: 
- user B transfers some STX to user A (nonce 1500) 
- user A transfers STX to user C (nonce 23) 
- user A calls contract-example (nonce 24)  
- ...
- user A calls contract-example-2 (nonce 400)

If the following miner does not build off this stream of microblocks (and instead builds off the previous anchor block for example), then all of user A's transactions become invalid. This is because the next miner will mine transactions in nonce-order, and will thus attempt to mine the transaction with a nonce of 23 first. 

So, every time the next miner doesn't build off the previous microblock stream, there could be a pretty significant re-org if we do the mempool admission check against unconfirmed state.

zone117x:
> So, every time the next miner doesn't build off the previous microblock stream, there could be a pretty significant re-org if we do the mempool admission check against unconfirmed state.

Yep, this has been an on going discussion with the UX team for a while, and are trying to structure the UI/UX to represent the risk/state around this.

diwakergupta:
Based on my reading of the code (search call sites and implementation of `mine_next_microblock`), I believe that microblock transactions don't go through the same mempool admission checks as regular transactions. @lgalabru can you confirm? I started with the neon_noder miner and worked my way down the call graph.

lgalabru:
[...]

PR #2778, opened 2021-07-23
Feat: added use_latest_tip query parameter to the relevant v2 endpoints

pavitthrap:
## Description
Updated `tip` query parameter to the v2 RPC endpoints to accept the value of "latest". 

When set to "latest", the latest tip is used - this includes the unconfirmed state, if there is any unconfirmed state. 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

Added changes to the openapi yaml file 

## Testing information
Added tests mostly in `rpc.rs`. 
Added test in `integrations.rs`.


src/net/rpc.rs
```diff
@@ -3086,67 +3143,73 @@ mod test {
...
-                )
-                .unwrap();
-                let microblock = microblock_builder
-                    .mine_next_microblock_from_txs(
-                        vec![
-                            (tx_cc_signed, tx_cc_len),
-                            (tx_unconfirmed_contract_signed, tx_unconfirmed_contract_len),
-                        ],
-                        &microblock_privkey,
+        // begin microblock section
```
gregorycoppola, line 3146: does it make sense to test this new code? Is it possible? Do you think it's self-testing?

src/net/rpc.rs
```diff
@@ -4137,6 +4349,48 @@ mod test {
...
+             ref mut convo_client,
+             ref mut peer_server,
+             ref mut convo_server| {
+                convo_client.new_getaccount(
+                    StacksAddress::from_string("ST2DS4MSWSGJ3W9FBC6BVT0Y92S345HY8N3T6AV7R")
+                        .unwrap()
+                        .to_account_principal(),
+                    None,
+                    false,
+                    true,
```
gregorycoppola, line 4393: Can we have a few comments about why we expect the answers in the tests to be the way they are? And why there are not expected to be what they are not?

gregorycoppola (commented):
Hello. Seems like a good change but I just had a few nits about the comments, including for the tests to add some info about the logic of the test for people new to the file.

src/net/rpc.rs
pavitthrap: The code itself is a test, so I'm not sure - also this is not new code, I added an indented section to make including microblocks optional.

src/net/rpc.rs
pavitthrap: Added comment.

docs/rpc/openapi.yaml
```diff
@@ -343,39 +395,47 @@ paths:
...
+            type: string
+        - name: tip
+          in: query
+          schema:
+            type: string
+          description: The Stacks chain tip to query from
+        - name: use_latest_tip
+          in: query
+          schema:
+            type: string
```
zone117x, line 437: Thoughts on making this more OpenAPI-compliant with a `boolean` rather than a `string`? It helps with doc and client generation. 

In the API, I usually document these as `boolean` (`true`/`false`), but in the route handling code also support parsing `1`/`0`.

pavitthrap: sure, I'll make this change

src/net/rpc.rs
gregorycoppola: this is a very helpful comment!, but can you please do the same for the other tests you are adding?

gregorycoppola (commented):
Looks good.. can you just add comments for all tests you are adding?

docs/rpc/openapi.yaml
```diff
@@ -74,7 +74,14 @@ paths:
         schema:
           type: string
         description: The Stacks chain tip to query from
-
+      - name: use_latest_tip
+        in: query
+        schema:
+          type: boolean
+        description: |
+          If use_latest_tip == true, the query will be run from the latest tip (includes unconfirmed state).
```
lgalabru, line 82:
```suggestion
          If use_latest_tip == true, the query will be run from the latest known tip (includes unconfirmed state).
```

src/net/rpc.rs
gregorycoppola: ok

src/net/rpc.rs
```diff
@@ -3321,17 +3398,22 @@ mod test {
     #[test]
     #[ignore]
     fn test_rpc_getpoxinfo() {
+        /// Test v2/pox (aka GetPoxInfo) endpoint.
+        /// In this test, we don't set any tip parameters, and we expect that querying for pox info
```
gregorycoppola, line 3402: so it defaults to `use_latest_tip` is true? If so please update comment

src/net/rpc.rs
```diff
@@ -4076,13 +4229,62 @@ mod test {
 
     #[test]
     #[ignore]
-    fn test_rpc_get_contract_src_unconfirmed() {
+    fn test_rpc_get_contract_src_unconfirmed_with_canonical_tip() {
+        /// Test v2/contracts/source (aka GetContractSrc) endpoint.
+        /// In this test, we don't set any tip parameters, and allow the endpoint to execute against
+        /// the canonical Stacks tip.
+        /// The contract source we are querying for exists in the unconfirmed state, so we expect
```
gregorycoppola, line 4236: "only exists"

src/net/rpc.rs
gregorycoppola: nice comments, thank you!

gregorycoppola (approved):
Thanks for all the new test comments.. I feel like I can understand it a lot better and hopefully will be the same for other readers as well!

src/net/rpc.rs
```diff
@@ -1512,37 +1511,66 @@ impl ConversationHttp {
...
+    /// - `use_latest_tip` is also an optional query parameter, and defaults to false.
+    ///
     /// The order of chain tips this method prefers is as follows:
-    /// * tip_opt, if it's Some(..),
-    /// * the unconfirmed canonical stacks chain tip, if initialized
-    /// * the confirmed canonical stacks chain tip
+    /// * If `use_latest_tip` is true, the method attempts to return the unconfirmed chain tip,
+    ///     and returns the confirmed stacks chain tip if that doesn't work.
+    /// * If `tip_opt` is Some(tip), the method returns tip.
+    /// * The last option the function considers is the canonical Stacks chain tip.
```
gregorycoppola, line 1524: This is kind of confusing, how about "Otherwise, the function considers the canonical Stacks chain tip."

gregorycoppola (approved):
Thanks for the change! One small nit on the comments but LGTM.

src/net/http.rs
```diff
@@ -1674,6 +1675,19 @@ impl HttpRequestType {
         !no_proof
     }
 
+    /// Check whether the given option query string sets use_latest_tip=1 (setting use_latest_tip to true).
+    /// Defaults to false.
+    fn get_use_latest_chain_tip(query: Option<&str>) -> bool {
+        if let Some(query_string) = query {
+            form_urlencoded::parse(query_string.as_bytes())
+                .find(|(key, _v)| key == "use_latest_tip")
+                .map(|(_k, value)| value == "1" || value == "true")
```
jcnelson, line 1684: Do we gain anything concrete by accepting either/or `1` or `true`?  Might as well just pick one or the other if not.

src/net/http.rs
```diff
@@ -2516,46 +2564,55 @@ impl HttpRequestType {
                 _key,
                 tip_opt,
                 with_proof,
+                use_latest_tip,
```
jcnelson, line 2567: You don't have to address it here, but could you open an issue to consolidate `tip_opt`, `with_proof`, and `use_latest_tip` into a single struct that contains the standard query parameters that most API endpoints accept?  Otherwise PRs like this are going to get untenable ;)

pavitthrap: sure!

src/net/http.rs
zone117x: Yep, if we want to adhere to OpenAPI schema and general API design guidelines, this should be a boolean rather than an `number` or `string`. I suggested also allowing a number for both consistency purposes (some existing endpoints use a number rather than a boolean), as well as the robustness principal (permissive in what is accepted) because this isn't really a security-sensitive feature. 

In the stacks-blockchain-api, similar boolean query params are pretty liberal in what is accepted, because it's not entirely unreasonable for someone to do any of the following:
`curl http://example.com/?use_latest_tip=true`,
`curl http://example.com/?use_latest_tip=1`, or even
`curl http://example.com/?use_latest_tip`. 

But, if there are reservations around accepting multiple forms of query param input, it should be a proper boolean.

jcnelson: What's here is fine.

src/net/rpc.rs
```diff
@@ -4076,13 +4230,62 @@ mod test {
...
     #[ignore]
-    fn test_rpc_get_contract_src_unconfirmed() {
+    fn test_rpc_get_contract_src_unconfirmed_with_canonical_tip() {
+        /// Test v2/contracts/source (aka GetContractSrc) endpoint.
+        /// In this test, we don't set any tip parameters, and allow the endpoint to execute against
+        /// the canonical Stacks tip.
+        /// The contract source we are querying for only exists in the unconfirmed state, so we
+        /// expect the query to fail.
+        test_rpc(
+            "test_rpc_get_contract_src",
```
jcnelson, line 4240: Please be sure to change the test name and the ports used here.

src/net/rpc.rs
```diff
@@ -4097,16 +4300,60 @@ mod test {
...
+
+    #[test]
+    #[ignore]
+    fn test_rpc_get_contract_src_use_latest_tip() {
+        /// Test v2/contracts/source (aka GetContractSrc) endpoint.
+        /// In this test, we set `use_latest_tip` to true.
+        /// The contract source we are querying for exists in the unconfirmed state, so we expect
+        /// the query to succeed.
+        test_rpc(
+            "test_rpc_get_contract_src_unconfirmed",
```
jcnelson, line 4333: Please be sure to change the test name and ports here.

src/net/rpc.rs
```diff
@@ -4137,6 +4385,96 @@ mod test {
...
+    }
+
+    /// In this test, the query parameter `use_latest_tip` is set to true, and so we expect the
+    /// tip used for the query to be the latest microblock.
+    /// We check that the account state matches the state in the most recent microblock.
+    #[test]
+    #[ignore]
+    fn test_rpc_get_account_use_latest_tip() {
+        test_rpc(
+            "test_rpc_get_account",
```
jcnelson, line 4416: Please be sure to change the test name and ports here.

src/net/rpc.rs
```diff
@@ -4291,6 +4640,62 @@ mod test {
...
+            },
+        );
+    }
+
+    #[test]
+    #[ignore]
+    fn test_rpc_get_map_entry_use_latest_tip() {
+        test_rpc(
+            "test_rpc_get_map_entry_unconfirmed",
+            40140,
```
jcnelson, line 4674: Same here please.

src/net/rpc.rs
```diff
@@ -4377,6 +4791,42 @@ mod test {
...
+                }
+            },
+        );
+    }
+
+    #[test]
+    #[ignore]
+    fn test_rpc_get_contract_abi_use_latest_tip() {
+        test_rpc(
+            "test_rpc_get_contract_abi_unconfirmed",
```
jcnelson, line 4814: Same here

src/net/rpc.rs
```diff
@@ -4436,6 +4891,58 @@ mod test {
...
     }
 
+    #[test]
+    #[ignore]
+    fn test_rpc_call_read_only_use_latest_tip() {
+        /// Test /v2/contracts/call-read (aka CallReadOnlyFunction) endpoint.
+        /// In this test, we set `use_latest_tip` to true , and we expect that querying
+        /// against the unconfirmed state will succeed.
+        test_rpc(
+            "test_rpc_call_read_only_unconfirmed",
```
jcnelson, line 4901: Same here

src/net/rpc.rs
```diff
@@ -1512,37 +1511,66 @@ impl ConversationHttp {
...
+                        None
+                    }
+                }
+                None => None,
+            };
+
+            if let Some(unconfirmed_chain_tip) = unconfirmed_chain_tip_opt {
+                Ok(Some(unconfirmed_chain_tip))
+            } else {
+                let tip = chainstate.get_stacks_chain_tip(sortdb).unwrap().unwrap();
```
jcnelson, line 1549: Please don't `.unwrap()` in network-facing code.

src/net/rpc.rs
```diff
@@ -1512,37 +1511,66 @@ impl ConversationHttp {
...
         sortdb: &SortitionDB,
         chainstate: &StacksChainState,
+        use_latest_tip: bool,
     ) -> Result<Option<StacksBlockId>, net_error> {
-        match tip_opt {
-            Some(tip) => Ok(Some(*tip).clone()),
-            None => match chainstate.get_stacks_chain_tip(sortdb)? {
-                Some(tip) => Ok(Some(StacksBlockHeader::make_index_block_hash(
+        if use_latest_tip {
+            let unconfirmed_chain_tip_opt = match &chainstate.unconfirmed_state {
```
jcnelson, line 1535: I just want to confirm that the following scenarios are tested:

* The caller passes `use_latest_tip=1` with an existing unconfirmed microblock stream, and Clarity state from the unconfirmed microblock stream is successfully loaded.
* The caller passes `use_latest_tip=1` with an empty unconfirmed microblock stream, and Clarity state from the canonical chain tip is successfully loaded (i.e. you don't get a 404 since the unconfirmed chain tip points to a nonexistent MARF trie).
* The caller passes `use_latest_tip=0`, and the canonical chain tip is used regardless of the state of the unconfirmed microblock stream.

jcnelson: That second bullet is going to be a bit thorny.  It may work in these unit tests, but the unit test framework doesn't reflect how the Stacks node actually deals with unconfirmed chain state.  This is because the Stacks node itself has a very weird way of representing unconfirmed state, due to the fact that unconfirmed state is curated both by the p2p thread (to service read-only queries off of it) and by the relayer thread (to process new microblocks).  I think you might want to create a nominal integration test for these bullet points in `neon_integrations.rs`.  It doesn't have to exhaustively test all of the RPC methods; it just needs to verify that the right chain tip is used in each situation (and nothing 404's).

jcnelson (changes requested):
Overall this looks good to me @pavitthrap!  I just had a couple of small but must-have testing asks.  After that, I'll approve.

src/net/rpc.rs
jcnelson: @pavitthrap This issue isn't resolved yet, so I'm unresolving the conversation.

src/net/rpc.rs
jcnelson: Unresolving this conversation, since the issue is not resolved yet.

src/net/rpc.rs
jcnelson: Unresolving this conversation since the issue is not yet resolved.

src/net/rpc.rs
jcnelson: Unresolving this conversation since the issue is not yet resolved.

src/net/rpc.rs
jcnelson: Unresolving

src/net/rpc.rs
jcnelson: Unresolving

src/net/rpc.rs
jcnelson: Unresolving

jcnelson:
Review is blocked on #2774

src/net/rpc.rs
pavitthrap: Addressed in commit 0c5f6aa

src/net/rpc.rs
```diff
@@ -1512,37 +1511,97 @@ impl ConversationHttp {
...
+                        warn!("Failed to load Stacks chain tip");
+                        let response = HttpResponseType::ServerError(
+                            response_metadata,
+                            format!("Failed to load Stacks chain tip"),
+                        );
+                        response.send(http, fd).and_then(|_| Ok(None))
+                    }
+                }
+            }
+        } else {
```
jcnelson, line 1586: This API feels a little weird.  According to this code path, if I requested the query string `?tip=xxxxxxx&use_latest_tip=true`, not only would the request potentially succeed, but also the `tip=` argument would be ignored in favor of `use_latest_tip`.  Is this what we want?  Would it make more sense to require the caller to use *either* `tip=` *or* `use_latest_tip` but *not* both?

jcnelson: It makes me wonder if instead of having a separate `use_latest_tip` query argument, we could instead overload `tip=` so that `tip=latest` becomes a valid argument.  Then, this ambiguity goes away.

pavitthrap: I think I like adding the option for `tip=latest`. I'll refactor this PR to do that.

src/net/rpc.rs
```diff
@@ -1532,23 +1531,75 @@ impl ConversationHttp {
...
-        chainstate: &StacksChainState,
+        chainstate: &mut StacksChainState,
     ) -> Result<Option<StacksBlockId>, net_error> {
-        match tip_opt {
-            Some(tip) => Ok(Some(*tip).clone()),
-            None => match chainstate.get_stacks_chain_tip(sortdb)? {
+        match tip_req {
+            TipRequest::UseLatestUnconfirmedTip => {
+                let unconfirmed_chain_tip_opt = match &mut chainstate.unconfirmed_state {
+                    Some(unconfirmed_state) => {
```
jcnelson, line 1550: Minor refactoring nit: can you add a method to `UnconfirmedState` called `get_unconfirmed_tip_if_exists()` that does all this work here?  Then, if it fails -- i.e. the underlying trie doesn't exist -- this code block would simply return the appropriate HTTP response type.

src/net/rpc.rs
```diff
@@ -1532,23 +1531,75 @@ impl ConversationHttp {
...
+                            // Check if underlying MARF trie exists before returning unconfirmed chain tip
+                            let trie_exists = match unconfirmed_state
+                                .clarity_inst
+                                .trie_exists_for_block(&unconfirmed_state.unconfirmed_chain_tip)
+                            {
+                                Ok(res) => res,
+                                Err(e) => {
+                                    let response_metadata = HttpResponseMetadata::from(req);
+                                    warn!("Failed to load Stacks chain tip; error checking underlying trie");
+                                    let response = HttpResponseType::ServerError(
```
jcnelson, line 1561: This wouldn't be a server error (500).  It would be a not-found error (404).

src/net/rpc.rs
```diff
@@ -1532,23 +1531,75 @@ impl ConversationHttp {
...
+                } else {
+                    match chainstate.get_stacks_chain_tip(sortdb)? {
+                        Some(tip) => Ok(Some(StacksBlockHeader::make_index_block_hash(
+                            &tip.consensus_hash,
+                            &tip.anchored_block_hash,
+                        ))),
+                        None => {
+                            let response_metadata = HttpResponseMetadata::from(req);
+                            warn!("Failed to load Stacks chain tip");
+                            let response = HttpResponseType::ServerError(
```
jcnelson, line 1592: This would also be a 404, not a 500.  A stacks node that hasn't downloaded any blocks yet is not in an error state.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -671,6 +671,33 @@ fn get_chain_tip_height(http_origin: &str) -> u64 {
...
 }
 
+fn get_contract_src(
+    http_origin: &str,
+    contract_addr: StacksAddress,
+    contract_name: String,
+    use_latest_tip: bool,
+) -> Result<String, String> {
+    let client = reqwest::blocking::Client::new();
+    let query_string = if use_latest_tip {
```
jcnelson, line 681: I think this PR got rid of `use_latest_tip`, and now just uses `tip=latest`?

docs/rpc/openapi.yaml
```diff
@@ -74,7 +74,14 @@ paths:
         schema:
           type: string
         description: The Stacks chain tip to query from
-
+      - name: use_latest_tip
```
jcnelson, line 77: I think this PR changes this from `use_latest_tip` to `tip=latest`?

jcnelson:
This PR is looking a lot better.  Thanks for adding all the thorough test coverage!  Just a few minor nits to address, and then I'll approve.

jcnelson (approved):
LGTM.  I think the one thing that still stands out to me on this PR is that there's a race condition whereby the unconfirmed tip obtained from `handle_load_stacks_chain_tip()` can get invalidated before getting passed into `maybe_read_only_clarity_tx()`.  However, the result of this race occurring isn't too bad -- the caller would simply receive a 404 response.  I don't think this is a show-stopping problem; if it becomes a problem, the fix would be to simply load the unconfirmed chain tip hash again at the call for `maybe_read_only_clarity_tx()` if it got invalidated.  Can you add a TODO comment describing this on `handle_load_stacks_chain_tip()` before merging (and open an issue for it)?

PR #2780, opened 2021-07-26
Only consider target_block_height when relevant

lgalabru:
With this fix, we're only considering the `target_block_height` if this value is relevant.
The `sync_with_indexer` function is reading the burnchain state from the bitcoin-node and should be considered as the source of truth, and should ignore / correct the hint when it's determined as off.
Address #2783.

### DIFF
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -1213,7 +1213,10 @@ impl Burnchain {
         );
 
         if let Some(target_block_height) = target_block_height_opt {
-            if target_block_height < end_block {
+            // target_block_height is used as a hint, but could also be completely off
+            // in certain situations. The current function is directly reading the 
+            // headers and syncing with the bitcoin-node.
+            if target_block_height > start_block && target_block_height < end_block {
                 debug!(
                     "Will download up to max burn block height {}",
                     target_block_height
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -1214,7 +1214,7 @@ impl Burnchain {
 
         if let Some(target_block_height) = target_block_height_opt {
             // target_block_height is used as a hint, but could also be completely off
-            // in certain situations. The current function is directly reading the 
+            // in certain situations. The current function is directly reading the
             // headers and syncing with the bitcoin-node.
             if target_block_height > start_block && target_block_height < end_block {
                 debug!(
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -1215,13 +1215,19 @@ impl Burnchain {
         if let Some(target_block_height) = target_block_height_opt {
             // target_block_height is used as a hint, but could also be completely off
             // in certain situations. The current function is directly reading the
-            // headers and syncing with the bitcoin-node.
+            // headers and syncing with the bitcoin-node, and the interval of blocks
+            // to download computed here should be considered as our source of truth.
             if target_block_height > start_block && target_block_height < end_block {
                 debug!(
                     "Will download up to max burn block height {}",
                     target_block_height
                 );
                 end_block = target_block_height;
+            } else {
+                debug!(
+                    "Ignoring target block height {} considered as irrelevant",
+                    target_block_height
+                );
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/burnchains/burnchain.rs
```diff
@@ -1213,12 +1213,21 @@ impl Burnchain {
         );
 
         if let Some(target_block_height) = target_block_height_opt {
-            if target_block_height < end_block {
+            // target_block_height is used as a hint, but could also be completely off
```
gregorycoppola, line 1216: Maybe put the variable name `target_block_height` in some kind of quotes.

src/burnchains/burnchain.rs
```diff
@@ -1213,12 +1213,21 @@ impl Burnchain {
         );
 
         if let Some(target_block_height) = target_block_height_opt {
```
gregorycoppola, line 1215: So, there's no test? How was this change tested and why do we know it is an improvement?

src/burnchains/burnchain.rs
```diff
@@ -1213,12 +1213,21 @@ impl Burnchain {
...
-            if target_block_height < end_block {
+            // target_block_height is used as a hint, but could also be completely off
+            // in certain situations. The current function is directly reading the
```
gregorycoppola, line 1217: s/"The current function"/"This function"

src/burnchains/burnchain.rs
```diff
@@ -1213,12 +1213,21 @@ impl Burnchain {
...
+            // headers and syncing with the bitcoin-node, and the interval of blocks
+            // to download computed here should be considered as our source of truth.
```
gregorycoppola, line 1219: Does the comment on line 1144 need to change?

gregorycoppola (commented):
Hey Ludo.. just a few comments about the comments.. and some questions about clarifying how this was tested.

gregorycoppola:
Also.. is there an Issue for this?

src/burnchains/burnchain.rs
```diff
@@ -1213,12 +1213,21 @@ impl Burnchain {
         );
```
pavitthrap, line 1213: I think the debug statement above should be 
```
 debug!(
            "Sync'ed headers from {} to {}. DB at {}",
            sync_height, end_block, db_height
         );
```

src/burnchains/burnchain.rs
pavitthrap: +1

lgalabru: So, no, unfortunately there's no test. 
This change was tested locally with a copy the chainstate of a miner that was stuck because of this bug, and this branch was then directly deployed by @CharlieC3 to the miner on the same chainstate, and that helped us unblocking the miner - looks like an improvement. 

The code coverage for this routine is pretty bad because this function is doing way too many things. 
I think we could rewrite this routine, extract the logic dealing with the boundaries and target block (there are a lot of `if` in there that would deserve some testing), and unit test that logic.

src/burnchains/burnchain.rs
lgalabru:
```suggestion
            // `target_block_height` is used as a hint, but could also be completely off
```

src/burnchains/burnchain.rs
lgalabru: Note: you can actually propose some changes using the suggestion syntax - https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request

lgalabru:
```suggestion
            // in certain situations. This function is directly reading the
```

src/burnchains/burnchain.rs
lgalabru: I don't think so. This function is still returning a burnchain tip, at least as high as target_block_height_opt if given.

src/burnchains/burnchain.rs
lgalabru: Addressed with latest

lgalabru:
@gregorycoppola yes there's an issue - https://github.com/blockstack/stacks-blockchain/issues/2783

src/burnchains/burnchain.rs
gregorycoppola: OK. Can you just summarize this in the PR description?

gregorycoppola (approved):
Can you link the issue in the description?

pavitthrap (approved):

PR #2785, opened 2021-07-28
Add doc guide

pavitthrap:
I added preliminary documentation guidelines to `CONTRIBUTORS.md`. Since rust comes with a tool called rustdoc, I am proposing we write our docs in a style that complies with it (for example, formatting comments with markdown). Here's what I propose we follow (thanks @diwakergupta for linking me to this): https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html 

I also added a section for overarching "documentation rules", so feel free to propose modifications/ more rules. 

I also wanted to collect people's thoughts on adding doc coverage to the release process. The nightly build of rustdoc has a flag `--show-coverage`, which calculates the percent of items with documentation. I'm thinking we could list the percent coverage in `CHANGELOG.md` when we do releases. 

Overall, feel free to chime in with any opinions here!

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -35,6 +35,30 @@ This project and everyone participating in it is governed by this [Code of Condu
 - Any major changes should be added to the [CHANGELOG](CHANGELOG.md).
 - Mention any required documentation changes in the description of your pull request.
 - If adding an RPC endpoint, add an entry for the new endpoint to the OpenAPI spec `./docs/rpc/openapi.yaml`.
+- If your code adds or modifies any major features (struct, trait, test, module, etc.), each should be 
+  documented according to the [Rust style guide](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html) 
+
+### Documentation Style 
+We will try to follow the [Rust style guide](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html). 
+Here's a contrived example of code following the guide: 
+```rust
+/// Returns 0 if the input is less than or equal to 5, and panics otherwise.
+/// 
+/// # Panics 
+/// The function panics if the input is greater than 5.
+pub fn example(i: int) -> int {
+    if i > 5 {
+        panic!("Panicking")
+    } else {
+        0
+    }
+}
+```
+
+Overview of documentation rules
+- Use markdown to format comments. 
+- Start with a high-level description of the function, adding more sentences with details if necessary. 
+- For some functions, you might want to add a markdown section for panics or for examples.
 
 ### Each file should include relevant unit tests
 


CONTRIBUTORS.md
```diff
@@ -35,6 +35,30 @@ This project and everyone participating in it is governed by this [Code of Condu
...
+
+### Documentation Style 
+We will try to follow the [Rust style guide](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html). 
+Here's a contrived example of code following the guide: 
+```rust
+/// Returns 0 if the input is less than or equal to 5, and panics otherwise.
+/// 
+/// # Panics 
+/// The function panics if the input is greater than 5.
+pub fn example(i: int) -> int {
```
lgalabru, line 49: Should we be describing the inputs?

diwakergupta: From [the guide](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components):

> Because the type system does a good job of defining what types a function passes and returns, there is no benefit of explicitly writing it into the documentation, especially since rustdoc adds hyper links to all types in the function signature.

reedrosenbluth (approved):
This looks good to me!

CONTRIBUTORS.md
gregorycoppola: I agree we shouldn't document things that are redundant in the context, and the type declarations can save a lot of comments. In some cases, though, there remains ambiguity even given the type.

Here are some examples, from the rust code base, of methods without comments that I believe are ambiguous without comments:

## Example 1

What is "input"?
```
    fn compute_cost(
        &mut self,
        cost_function: ClarityCostFunction,
        input: &[u64],
    ) -> Result<ExecutionCost, CostErrors> {
```

## Example 2
Height of what? Burn chain or Stacks chain?
```
    fn get_burn_header_hash(
        &self,
        height: u32,
        sortition_id: &SortitionId,
    ) -> Option<BurnchainHeaderHash>;
```

# Example 3
What is `mblock_header_1` vs `mblock_header_2`?
```
    pub fn run_poison_microblock(
        &mut self,
        sender: &PrincipalData,
        mblock_header_1: &StacksMicroblockHeader,
        mblock_header_2: &StacksMicroblockHeader,
    ) -> Result<Value, Error> {
```

CONTRIBUTORS.md
```diff
@@ -35,6 +35,30 @@ This project and everyone participating in it is governed by this [Code of Condu
...
+    } else {
+        0
+    }
+}
+```
+
+Overview of documentation rules
+- Use markdown to format comments. 
+- Start with a high-level description of the function, adding more sentences with details if necessary. 
+- For some functions, you might want to add a markdown section for panics or for examples.
```
gregorycoppola, line 61: it's not just panic, there is also Error results and None results.

pavitthrap: agreed, I think we should document inputs when there seems to be ambiguity.

pavitthrap:
Moving everything in my PR over to greg's docs pr (#2786 ) for simplicity.

PR #2786, opened 2021-07-28
feat(Style): Initial Guidelines for Comments

gregorycoppola:
We update the documentation to include guidelines for comments. This includes guidance and examples for comments on:

1. components
2. functions
3. data members
4. tests

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -158,73 +158,81 @@ Moreover, Pull requests where a large number of changes only deal with whitespac
 rejected.
 
 # Comments
-Comments are very important. Not only are they are an aid to the reader, but in many
-cases the comments specify the contract against which the code is evaluated.
-Without a clear contract for a function, for example, it is *impossible* to determine
-whether the function is implemented correctly.
-Here are high level rules for creating comments:
+Information-adding comments are very important for the readability and correctness of the codebase. The purpose of comments is:
+
+* Allow readers to understand the roles of components and functions without having to check how they are used.
+* Allow readers to check the correctness of the code against the comments.
+* Allow readers to follow tests.
+
+In the limit, if there are no comments, the problems that arise are:
+
+* Understanding one part of the code requires understanding *many* parts of the code. E.g., understanding what a function does requires reading its definition, which requires reading the definitions of all the functions it depends on, and so on recursively.
+* The user cannot be certain if there is a bug in the code, because there is no distinction between the contract of a function, and its definition.
+* The user cannot be sure if a test is correct, because the logic of the test is not specified, and the functions do not have contracts.
+
+## Formatting
+
+Comments are to be formatted in typical `rust` style, specifically:
+
 - Use markdown to format comments.
+
 - Use the triple forward slash "///" for modules, structs, enums, traits and functions. Use double forward slash "//" for comments on individual lines of code.
+
 - Start with a high-level description of the function, adding more sentences with details if necessary.
-- For some functions, you might want to add a markdown section for panics, errors, none results, or for examples.
-- Rust has explicit typing, so it most cases explicit documentation for input and outputs of functions is not necessary. However, if there is any ambiguity, document it in the function comment. 
-- Comments are not meant to recapitulate things that are obvious from the context, whether the type signature or otherwise. Use your best judgement to document that which is not obvious.
+
+- When documenting panics, errors, or other conceptual sections, introduce a Markdown section with a single `#`, e.g.:
+
+  - ```
+    # Errors
+    * ContractTooLargeError: Thrown when `contract` is larger than `MAX_CONTRACT_SIZE`.
+    ```
+
+## What to Comment
 The following kinds of things should have comments.
 
-## Components (`struct`'s, `trait`'s, and `enum`'s)
+### Components
 Comments for a component (`struct`, `trait`, or `enum`) should explain what the overall
-purpose of that component is. This is usually a concept, and not a formal contract.
-
-Help the reader understand why this group of data or methods are grouped
-together. By explaining what a component *does* we also implicitly explain what
-it *does not do* in a modular architecture.
+purpose of that component is. This is usually a concept, and not a formal contract. Include anything that is not obvious about this component.
 
 Example:
 
 ```rust
-/// The `BurnStateDB` allows the user to query the state of the underlying
-/// "burn chain", which in practice is the Bitcoin chain.
-trait BurnChainDB {
+/// The `ReadOnlyChecker` analyzes a contract to determine whether 
+/// there are any violations of read-only declarations. By a "violation"
+/// we mean a function that is marked as "read only" but which tries
+/// to modify chainstate.
+pub struct ReadOnlyChecker<'a, 'b> {
 ```
 
-## Functions
-The comments on a function should explain the contract that the function
-implements. This is very important. Without a clear contract, we cannot
-evaluate the correctness of the implementation.
+### Functions
+The comments on a function should explain what the function does, without having to read it. Wherever practical, it should specify the contract of a function, such that a bug in the logic could be discovered by a discrepancy between contract and implementation, or such that a test could be written with only access to the function comment.
 
 Without being unnecessarily verbose, explain how the output is calculated
-from the inputs. Explain any restrictions on the inputs. Explain error
+from the inputs. Explain the side effects. Explain any restrictions on the inputs. Explain failure
 conditions, including when the function will panic, return an error
 or return an empty value.
 
 Example:
 
 ```rust
-pub trait BurnStateDB {
-    /// Returns the *header hash* of the burnchain block at the
-    /// burnchain height `height` in sortition `sortition_id`.
+/// A contract that does not violate its read-only declarations is called
+/// *read-only correct*.
+impl<'a, 'b> ReadOnlyChecker<'a, 'b> {
+    /// Checks each top-level expression in `contract_analysis.expressions`
+    /// for read-only correctness.
     ///
-    /// Returns None if there is no block at this height.
-    fn get_burn_header_hash(
-        &self,
-        height: u32,
-        sortition_id: &SortitionId,
-    ) -> Option<BurnchainHeaderHash>;
+    /// Returns successfully iff this function is read-only correct.
+    ///
+    /// # Errors
+    ///
+    /// - Returns CheckErrors::WriteAttemptedInReadOnly if there is a read-only
+    ///   violation, i.e. if some function marked read-only attempts to modify
+    ///   the chainstate.
+    pub fn run(&mut self, contract_analysis: &ContractAnalysis) -> CheckResult<()>
 ```
-```rust
-/// Returns 0 if the input is less than or equal to 5, and panics otherwise.
-/// 
-/// # Panics 
-/// The function panics if the input is greater than 5.
-pub fn example(i: int) -> int {
-    if i > 5 {
-        panic!("Panicking")
-    } else {
-        0
-    }
-}
+
 Note that, if a function implements an interface, the comments should not
-repeat what was specified on the interface declaration.
+repeat what was specified on the interface declaration. The comment should only add information specific to that implementation.
 
 ## Data Members
 Each data member in a struct should have a comment describing what that member
@@ -235,20 +243,21 @@ name and type.
 Example:
 
 ```rust
-pub struct ContractAnalysis {
-    /// The identifier associated with the contract represented in
-    /// this `ContractAnalysis`.
-    pub contract_identifier: QualifiedContractIdentifier,
+pub struct ReadOnlyChecker<'a, 'b> {
+    /// Mapping from function name to a boolean indicating whether
+    /// the function with that name is read-only.
+    /// This map contains all functions in the contract analyzed.
+    defined_functions: HashMap<ClarityName, bool>,
 ```
 
-## Tests
+### Tests
 Each test should have enough comments to help an unfamiliar reader understand:
 
 1. what is conceptually being tested
-1. why the answer is expected
+1. why a given answer is expected
 
-Some times this can be obvious without much comments, perhaps from the context,
-or because the test is very simple. Usually it isn't.
+Sometimes this can be obvious without much comments, perhaps from the context,
+or because the test is very simple. Often though, comments are necessary.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTORS.md
```diff
@@ -154,6 +154,93 @@ All contributions should use the same whitespacing as the rest of the project.
...
+evaluate the correctness of the implementation.
+
+Without being unnecessarily verbose, explain how the output is calculated
+from the inputs. Explain any restrictions on the inputs. Explain error
+conditions, including when the function will panic, return an error
+or return an empty value.
+
+Example:
+
+```
```
pavitthrap, line 193:
```suggestion
```rust
```

pavitthrap: switch the code blocks to be rust

CONTRIBUTORS.md
```diff
@@ -154,6 +154,93 @@ All contributions should use the same whitespacing as the rest of the project.
...
 
+# Comments
+Comments are very important. Not only are they are an aid to the reader, but in many
+cases the comments specify the contract against which the code is evaluated.
+Without a clear contract for a function, for example, it is *impossible* to determine
+whether the function is implemented correclty.
+
+The following kinds of things should have comments.
+
+## Components (`struct`'s and `trait`'s)
```
pavitthrap, line 165:
```suggestion
## Components (`struct`'s, `trait`'s, and `enum`'s)
```

CONTRIBUTORS.md
```diff
@@ -154,6 +154,93 @@ All contributions should use the same whitespacing as the rest of the project.
...
+Comments for a component (`struct` or `trait`) should explain what the overall
```
pavitthrap, line 166:
```suggestion
Comments for a component (`struct`, `trait`, or `enum`) should explain what the overall
```

CONTRIBUTORS.md
```diff
@@ -154,6 +154,93 @@ All contributions should use the same whitespacing as the rest of the project.
...
+    /// Returns the *header hash* of the burnchain block at the
+    /// burnchain height `height` in sortition `sortition_id`.
+    ///
+    /// Returns None if there is no block at this height.
+    fn get_burn_header_hash(
+        &self,
+        height: u32,
+        sortition_id: &SortitionId,
+    ) -> Option<BurnchainHeaderHash>;
+```
```
pavitthrap, line 204: Do we want to give an example of a function with a markdown subheading (like with a panics section?)

gregorycoppola: I can add one or you can make some revisions.

gregorycoppola (commented):
Thanks for the comments, esp. @reedrosenbluth . Here is another draft.

CONTRIBUTORS.md
```diff
@@ -154,6 +154,93 @@ All contributions should use the same whitespacing as the rest of the project.
...
+
```
pavitthrap, line 205:
```suggestion
```rust
/// Returns 0 if the input is less than or equal to 5, and panics otherwise.
/// 
/// # Panics 
/// The function panics if the input is greater than 5.
pub fn example(i: int) -> int {
    if i > 5 {
        panic!("Panicking")
    } else {
        0
    }
}
```
```

File: CONTRIBUTORS.md (hunk 4 above)
Comment (pavitthrap) [reply CONTRIBUTORS.md:204]: I am just adding the example from my pr here as another example for documentation for a function

**Review by pavitthrap — APPROVED** [review]

File: CONTRIBUTORS.md [hunk 6]
```diff
@@ -154,6 +154,93 @@ All contributions should use the same whitespacing as the rest of the project.
 Moreover, Pull requests where a large number of changes only deal with whitespace will be
 rejected.
 
+# Comments
+Comments are very important. Not only are they are an aid to the reader, but in many
+cases the comments specify the contract against which the code is evaluated.
+Without a clear contract for a function, for example, it is *impossible* to determine
+whether the function is implemented correctly.
+
```
pavitthrap, line 162:
```suggestion
Here are high level rules for creating comments:
- Use markdown to format comments, and use the triple backlash "///"
- Start with a high-level description of the function, adding more sentences with details if necessary.
- For some functions, you might want to add a markdown section for panics, errors, none results, or for examples.
- Rust has explicit typing, so it most cases explicit documentation for input and outputs of functions is not necessary. However, if there is any ambiguity, document it in the function comment. 
```

pavitthrap:
I'm going to just close my docs pr in favor of yours. 

Could you add the following bullet after line 37 in the Documentation Updates section: 
"- If your code adds or modifies any major features (struct, trait, test, module, etc.), each should be 
   documented according to our [style rules](#comments) "

And can you also add the comments subheading to the table of comments?

pavitthrap:
> I'm going to just close my docs pr in favor of yours.
> 
> Could you add the following bullet after line 37 in the Documentation Updates section:
> "- If your code adds or modifies any major features (struct, trait, test, module, etc.), each should be
> documented according to our [style rules](#comments) "
> 
> And can you also add the comments subheading to the table of comments?

Also in the Documentation Updates section, can you note the following:
- It's possible to check the percentage of code coverage by (a) switching to the nightly version of rust (can run `rustup default nightly`, and also might need to edit `rust-toolchain` file to say "nightly" instead of "stable"), and (b) running `RUSTDOCFLAGS='-Z unstable-options --show-coverage' cargo doc`

CONTRIBUTORS.md
kantai: I agree with adding @pavitthrap's suggestion, but should it read `Here are high level rules for creating function comments` (because it doesn't seem to apply to things like inline comments, etc.)?

jcnelson: While what you have here looks good, I want to also make sure that our contributor guide doesn't become a form of bureaucratic gatekeeping that generates unnecessary work for contributors and results in low-entropy documentation for maintainers, all just to fulfill a PR checkbox.  While I like what you have here, I think there should be a bullet to the effect of "use your best judgement to document that which is not obvious."  It's up to the PR reviewers to ask for more comments if they don't understand how the code works, or if they feel that a 3rd party reviewer wouldn't readily understand.  But simple functions whose name describes what it does (e.g. getters and setters) I don't think need comment blocks.

gregorycoppola (commented):
Hello everyone,

I did kind of a re-write to address @jcnelson 's concern that comments not become a bureaucratic or time-wasting exercise in listing redundant information in the following ways:

* I chose some new examples that are less contrived and less vacuous, so it's more clear how a comment can add value.
* I updated the wording to emphasize that we don't want 'redundant' comments, that more comments aren't always better, and that it's important to use judgment.

A preponderance of examples were taken from #2804 only because it was fresh in my mind.

CONTRIBUTORS.md
jcnelson: > The goal should be to not put nothing,

I guess this is a difference between our personal philosophies, but I think that this should be *exactly* the goal.  The goal for *code* should be to write it so clearly that it's self-documenting -- the code itself should tell you what's happening.  This is obviously not realizable in practice, but it's an ideal I think we should hold.

The purpose of documentation, then, would be to explain that which cannot be immediately deduced from the code itself.  Namely, the code describes *what* happens, and the documentation describes what *should* happen.  Documentation, then, includes higher-level information like where a particular method or struct fits into the overall system architecture, and what its intended use-cases are and are not.  This would also include module-level documentation, as well as (in our case) documentation about what is and is not consensus-critical code.

You can see this in my code, for example.  I tend to write more free-form documentation blocks that try to describe the code's intended purpose and behavior in the overall system, and I generally eschew things like item-by-item documentation for method arguments and struct members (seeking instead to give them names and types that describe their purpose).  I also expect the reader to have some familiarity with the SIPs.  But in all cases, I absolutely expect the reader to read and comprehend the code in addition to reading the documentation.  If the code is so non-obvious that it needs lots of documentation to explain, then it's a bug in the code -- the code should be made more comprehensible.

kantai: > I guess this is a difference between our personal philosophies, but I think that this should be exactly the goal. The goal for code should be to write it so clearly that it's self-documenting -- the code itself should tell you what's happening. This is obviously not realizable in practice, but it's an ideal I think we should hold.

The problem with this approach is two-fold:

1. There's a lot of room for difference in what people think is clear or unclear. For the example of function arguments, there are dozens, if not more, examples in the codebase of function arguments that are named such that they describe sufficiently what they are (e.g., `clarity_tx`), but not really at all indicating what their relationship to the function's calling contract is. Writing rustdoc comments for function arguments is not terribly difficult, and something like "`clarity_tx` should be a transaction opened to chain tip" helps a lot to indicate how a given function's lookup is affected by the open block, etc.
2. @gregorycoppola's points about helping to review should not be understated -- code _is only_ capable of describing what it does. Not what it should do, or why it does. You need look no further than our own PRs to see that a lot of explanatory text occurs _beyond_ the code. Having that content all captured in comments enables future readers to see that explanation in the code, rather than it being lost to time in a merged PR. For a current reviewer, they can see a described calling contract or description of a parameter, struct, global, etc., and know to look for a correct implementation of that description _and_ sufficient testing.

I agree with you, @jcnelson, that the described guidelines here may encourage some needless verbosity in comments, but the burden of writing comments is not significant, and given that a major bottleneck in development currently is PR review time, I think the project would benefit from experimenting with forcing-functions that potentially trade-off some increased development burden for improvements to the review process.

gregorycoppola:
Just to emphasize as well, I wrote this Markdown document in a "this is what should happen" tone. That's because,I thought,  if checked in, the style guide would have to be authoritative and not say "one way to write comments would be..".

But, the meta-context is that all this is up for review. It's not what "should" happen, it's a proposal for what "should" happen, that will either get accepted, rejected or modified.


That said, "design by contract" isn't an idea I made up myself. It was drilled into us in undergrad so much I didn't know there was a debate about it.

CONTRIBUTORS.md
jcnelson: Here's a pithy description of the problem I'm trying to avoid.

What I want to see in PR reviews:

* Hi $SUBMITTER, I don't quite understand what's going on here. Can you add a comment here to describe why this code does what it does?

What I do *not* want to see in PR reviews:

* Hi $SUBMITTER, your comment here exceeds 80 columns. Please fix.
* Hi $SUBMITTER, you forgot to add a 10-line comment block for this 1-line self-explanatory getter
* Hi $SUBMITTER, you didn't add a comment for this struct field member even though its usage is obvious

The above sorts of nitpicking not only wastes both the reviewer's and submitter's time, but also fail to actually solve the problem that documentation solves: provide clarity to the reader as to what the code is supposed to be doing.

> given that a major bottleneck in development currently is PR review time

I agree.  In the spirit of addressing this, I'm trying to point out that opening the domain of discourse for code review to potentially hundreds of documentation nitpicking comments like the above will make this worse.  Can I get an agreement here that this will *not* happen?

jcnelson:
> That said, "design by contract" isn't an idea I made up myself. It was drilled into us in undergrad so much I didn't know there was a debate about it.

The idea isn't controversial.  Its execution in this style guide, however, is.

CONTRIBUTORS.md
jcnelson: Like, I don't want this style guide to require us to spend non-trivial amounts of cognitive overhead policing other peoples' comments, especially if the requested changes don't actually add any value.

gregorycoppola:
> The idea isn't controversial. Its execution in this style guide, however, is.

Well, I emphasize it's just a draft for discussion.

At least we are getting some common context on what people think is the ideal by discussing it.

But, if we get to a point where the discussion is more annoying than productive I am alright dropping it.


I'm personally getting to a point where I understand the code with or without that additional comments.


Also, I think I was heavily influenced by google style of comments, which are on the chatty side:
https://google.github.io/styleguide/cppguide.html#Function_Comments

CONTRIBUTORS.md
gregorycoppola: I think the notion of "nitpicking" boils down to a situation where two people differ on what they think is important in code style.

This what an explicitly agreed on style guide is meant to address. 

If something is considered important enough to be in the style guide, then insisting on it isn't nitpicking. And if something isn't covered by the style guide, then insisting on it *is* nitpicking.

Individuals might disagree with individual points in the style guide, just like they could disagree on any product decision, but at least everyone agrees on what to ask for and what to include in PR's.

gregorycoppola: I guess some of the material I have been putting in comments is redundant, e.g.:

```
impl<T: MarfTrieId> MARF<T> {
    /// Reopen this MARF with readonly storage.
    ///
    /// Returns Err if:
    ///   1) This class is already in the process of writing.
    ///   2) A new underlying SQLite database connection cannot be established.
    pub fn reopen_readonly(&self) -> Result<MARF<T>, Error> {
```

You could make a rule like `don't restate the sentence in the function name; assume the user has read it, and interpreted it as English`, in which case the above could be reduced to:

```
impl<T: MarfTrieId> MARF<T> {
    /// Returns Err if:
    ///   1) This class is already in the process of writing.
    ///   2) A new underlying SQLite database connection cannot be established.
    pub fn reopen_readonly(&self) -> Result<MARF<T>, Error> {
```

jcnelson: I'm less concerned about the case where an individual contributor chooses to go above and beyond the documentation requirements.  This example here is fine by me -- if I were a reviewer, I would be happy with either docstring, since in both cases I come away from reading the docstring and function signature understanding the code's contract.

What I'm more concerned about is whether or not the contributor guide will require us to spend more time worrying about aspects of documentation that *aren't adding value*.  For example, if the contributor guide mandated that each function have a sentence describing what the function does (even if it's obvious in the name), then (a) the contributor would be required to add unnecessary documentation and (b) the reviewers would be obliged to make sure the contributor does this for each function they add, even if it's not necessary.

I do think that reviewers should ask for more documentation if they believe it's warranted, and I agree with @kantai that the burden of adding more documentation isn't significant on the contributor.  However, I don't think that reviewers should have to spend an inordinate amount of time verifying that the documentation is consistent with a contributor guide.  **As long as the reviewer can read the docstring and function or struct signature and correctly comprehend the code's contract, then the documentation as given should be considered sufficient.**

jcnelson: Clarifying from offline sync:

* Reviewers should ask for comments and documentation where appropriate.  Such asks are PR-blocking.
* Reviewers are not required to rigorously check that a comment block adheres to a particular formatting standard, although self-imposed adherence to rustdoc is encouraged.

gregorycoppola: What about the *content* of comments though? 

Should we say that the **contracts** of functions should be included by default? We can make a rule like "don't restate redundant information from the declaration in the contract", to address the redundancy concern.

But, to me, the reviewing process would involve such steps as:
* compare a function's implementation to its contract
* compare a function's tests to its contract
* compare the contracts of smaller functions, to see how they build into larger functions

I guess I'm curious what your reviewing algorithm is if it doesn't involve the function contracts? Or, maybe you have such a model of the overall codebase and the work being done that you implicitly know what the contract of each method is even if its not documented? @jcnelson 

I mean, the concept of a function having a "bug" sort of implies to me that the contract exists somewhere (either in someone's mind, or PR description, if not the comments). If the definition itself were the contract, there would be no such thing as a bug, but we do find bugs. (Is this getting too philosophical?)

So, if functions do have contracts that exist independently of the implementations, I think it helps the reviewers and the author verify accuracy if we state the contracts explicitly in the comments. Cf., @kantai reviewing my comments on the ReadOnlyChecker.

gregorycoppola (commented):
There hadn't been any activity on this for a while so I pushed a small update. Basically, I felt like I was trying to push for **always** documenting the contract of a function.

And @jcnelson was pushing for **never documenting redundant info**.

So, what about a synthesis of the two, which is at the you should:
1) **always** document the contracts of functions precisely, but 
2) **always** try to use context wherever possible to reduce the size of the comments, to zero when possible.

I tried to summarize this in this new discussion.

What do you think about this as a policy?

---

#### Don't Restate the Function Names

The contracts of functions should be implemented precisely enough that tests could be written looking only at the declaration and the comments (and without looking at the definition!). However:

* **the author should assume that the reader has already read and understood the function name, variable names, type names, etc.**
* **the author should only state information that is new**

So, if a function and its variables have very descriptive names, then there may be nothing to add in the comments at all!

**Bad Example**

```
/// Appends a transaction to a block.
fn append_transaction_to_block(transaction:Transaction, &mut Block) -> Result<()>
```

This is considered bad because the function name already says "append transaction to block", so it doesn't add anything to restate it in the comments. However, *do* add anything that is not redundant, such as elaborating what it means to "append" (if there is more to say), or what conditions will lead to an error.

**Good Example**

```
/// # Errors
///
/// - BlockTooBigError: Is returned if adding `transaction` to `block` results
/// in a block size bigger than MAX_BLOCK_SIZE.
fn append_transaction_to_block(transaction:Transaction, block:&mut Block) -> Result<()>
```

This is considered good because the reader builds on the context created by the function and variable names. Rather than restating them, the function just adds elements of the contract that are not implicit in the declaration.

jcnelson:
Can someone else approve this so we can merge it?

diwakergupta (approved):

PR #2787, opened 2021-07-29
Feat/mempool admission logic

### LINKED ISSUE #2774: Mempool admission logic change: do admission check against unconfirmed state

pavitthrap:
## Description
Addresses #2774. 
With this change, any transaction posted to the `v2/transactions` will be accepted into the mempool if it is able to be validated against either the unconfirmed state or the confirmed state. 

In addition, this PR adds the optional query parameter `mempool_admission_check` to the `v2/transactions` endpoint. 
If `mempool_admission_check=="unconfirmed"`, the transaction will only be validated against unconfirmed state (and then accepted into the mempool if that validation succeeds). If `mempool_admission_check=="confirmed"`, the transaction will only be validated against confirmed state (and then accepted into the mempool if that validation succeeds).

Note that if the unconfirmed state is uninitialized, then specifying `mempool_admission_check=="unconfirmed"` will validate the transaction against the previous anchored tip. 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

Added docs to openapi.yaml 

## Testing information
Added test to `neon_integrations.rs`


testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -742,10 +748,10 @@ fn liquid_ustx_integration() {
 
     let publish = make_contract_publish(&spender_sk, 0, 1000, "caller", caller_src);
 
-    let replaced_txid = submit_tx(&http_origin, &publish);
+    let replaced_txid = submit_tx(&http_origin, &publish, TransactionAnchorMode::Any);
```
gregorycoppola, line 751: Why does `Any` match the previous behavior? I would have thought it was maybe `OnChainOnly`? I thought the new behavior you were adding was more permissive, but this makes it seem like the old behavior was more permissive.

gregorycoppola (commented):
Interesting change, assuming there is consensus we want to do this.

I had some comments that might be helpful but I'm kind of in a new part of the codebase for me here.

It seems like maybe there are not enough tests for such an important change? Are all the new code paths tested? Are you testing the output and showing different answers for different conditions? Is it possible to add unit tests in addition to the integration test?

I'm still trying to understand the old and new behavior of this code, so maybe some of my questsions are like newbie questisons.

testnet/stacks-node/src/tests/neon_integrations.rs
pavitthrap: `Any` does not match the previous behavior. However, based on the discussion in the linked issue, we want `Any` to be the new default behavior.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1915,6 +1921,157 @@ fn microblock_integration_test() {
...
+
+    // first block will hold our VRF registration
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    // second block will be the first mined Stacks block
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    // let's query our first spender
+    let account = get_account(&http_origin, &spender_addr);
+    assert_eq!(account.balance, 100300);
```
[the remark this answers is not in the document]
pavitthrap: this is the balance that is initialized in the beginning of the function

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1915,6 +1921,157 @@ fn microblock_integration_test() {
...
+        .post(&path)
+        .header("Content-Type", "application/octet-stream")
+        .body(call_tx.clone())
+        .send()
+        .unwrap();
+    let err_str = res.text().unwrap();
+    assert!(err_str.contains("NoSuchContract"));
+
+    // call the contract defined in the microblock, but make mempool admission check against unconfirmed state
+    submit_tx(&http_origin, &call_tx, TransactionAnchorMode::OffChainOnly);
```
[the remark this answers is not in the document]
pavitthrap: So the `submit_tx` function checks for a successful status code, which is why we don't need an assert! here - the function itself will panic if the call fails.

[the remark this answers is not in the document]
pavitthrap: That's what I do in the lines before - I don't call submit_tx for `OnChainOnly` because that function panics on failure

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5272,28 +5273,45 @@ impl StacksChainState {
...
+            }
+        };
+
+        let res = match self.maybe_read_only_clarity_tx(&NULL_BURN_STATE_DB, &current_tip, |conn| {
             StacksChainState::can_include_tx(conn, &conf, has_microblock_pubk, tx, tx_size)
         }) {
-            Some(r) => r,
-            None => Err(MemPoolRejection::NoSuchChainTip(
+            Ok(Some(r)) => r,
+            Ok(None) | Err(_) => Err(MemPoolRejection::NoSuchChainTip(
```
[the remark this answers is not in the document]
pavitthrap: You're right, I should propagate the underlying error here. 

@lgalabru do you think it would be appropriate to
- add an error type to `MemPoolRejection` that takes in type Error. Right now, the only enum that does that in `MemPoolRejection` is `FailedToValidate(Error)`, but that error doesn't make sense here.
- convert the error here to a string and have the error be `MemPoolRejection::Other(stringified_err)`

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5272,28 +5273,45 @@ impl StacksChainState {
...
+                    if unconfirmed_state.is_readable() {
+                        unconfirmed_state.unconfirmed_chain_tip
+                    } else {
+                        StacksChainState::get_parent_index_block(
+                            current_consensus_hash,
+                            current_block,
+                        )
+                    }
+                } else {
+                    StacksChainState::get_parent_index_block(current_consensus_hash, current_block)
```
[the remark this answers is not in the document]
pavitthrap: We want to avoid using `unwrap` in non-test code, which is why I wrote it like this

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1915,6 +1921,157 @@ fn microblock_integration_test() {
...
+    // let's query our first spender
+    let account = get_account(&http_origin, &spender_addr);
+    assert_eq!(account.balance, 100300);
+    assert_eq!(account.nonce, 0);
+
+    // this call wakes up our node
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    // open chainstate
+    // TODO (hack) instantiate the sortdb in the burnchain
```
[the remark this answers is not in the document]
pavitthrap: I put this comment here bc I duplicated this code from elsewhere: 
https://github.com/blockstack/stacks-blockchain/commit/27381d64b1e6c8f308537a2c397958a97fdea337

I can leave a todo in the code and revisit when Jude is back in office so I can ask him what the underlying todo is

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1915,6 +1921,157 @@ fn microblock_integration_test() {
...
+        HttpRequestType::make_query_string_for_post_tx_endpoint(TransactionAnchorMode::OnChainOnly);
+    let path = format!("{}/v2/transactions{}", http_origin, query);
+    let res = client
+        .post(&path)
+        .header("Content-Type", "application/octet-stream")
+        .body(call_tx.clone())
+        .send()
+        .unwrap();
+    let err_str = res.text().unwrap();
+    assert!(err_str.contains("NoSuchContract"));
```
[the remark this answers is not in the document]
pavitthrap: explained in code comment.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5234,6 +5234,7 @@ impl StacksChainState {
         current_block: &BlockHeaderHash,
         tx: &StacksTransaction,
         tx_size: u64,
+        mempool_admission_check: TransactionAnchorMode,
```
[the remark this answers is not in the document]
pavitthrap: The "no such chain tip" and various mem pool rejection reasons are the errors relevant to this new parameter. 

There aren't any "new" errors that result from this parameter, since this function attempted to validate the given tx against the confirmed chain tip and the unconfirmed chain tip (in the case of a bad nonces error) before. Thus, I'm not making any "new" function calls, just switching the ordering.

src/chainstate/stacks/db/blocks.rs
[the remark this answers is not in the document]
gregorycoppola: but, that's why i suggested to first check `self.unconfirmed_state.is_some()`... it should be fine to unwrap after checking that, no? I think what we don't want is unchecked unwraps.

src/chainstate/stacks/db/blocks.rs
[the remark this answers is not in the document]
gregorycoppola: it might be easier to just deal with the error in this block. how does it add value to return an error and then re-match? seems like it takes more mental RAM to keep track of it here, because you have to understand the correspondence between two match blocks instead of just looking at one.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5226,14 +5226,25 @@ impl StacksChainState {
...
+    /// If `mempool_admission_check`==OnChainOnly, the transaction will only be validated against confirmed state.
+    /// If `mempool_admission_check`=Any, the transaction will be accepted if it validates against either unconfirmed or confirmed tip
+    ///
+    /// This function returns an error if
+    ///     - The transaction fails a basic semantic check (for example, if the sender & recipient
+    ///       of a token transfer are the same).
+    ///     - If there was a failure in querying the stacks block height.
+    ///     - If the tx is a poison microblock tx, and there is a failure in recovering the
+    ///       public key.
+    ///     -
```
gregorycoppola, line 5240: this line is empty

src/chainstate/stacks/db/blocks.rs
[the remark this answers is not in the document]
gregorycoppola: but, isn't the format for documenting errors (suggested by you :)), that there be a section called `#Errors` and then we actually list the class names of the errors.

src/chainstate/stacks/mod.rs
```diff
@@ -360,7 +360,9 @@ impl Txid {
     }
 }
 
-/// How a transaction may be appended to the Stacks blockchain
+/// Used to determine
+/// - how a transaction may be appended to the Stacks blockchain
+/// - how a transaction will be validated for the the post transaction endpoint
```
gregorycoppola, line 365: nit: "the the post transaction"? is there a noun missing between the two "the's"

gregorycoppola: why are you reusing this enum?

In your `openapi.yaml` description you say

```
            If mempool_admission_check=="unconfirmed", the transaction will only be validated against unconfirmed state.
            If mempool_admission_check=="confirmed", the transaction will only be validated against confirmed state.
            If unspecified, the transaction will be accepted if it validates against either unconfirmed or confirmed state.
```

It seems kind of confusing to me that we went from confirmed/unconfirmed to onchain/offchain.

Is that like too newbie of a confusion?

testnet/stacks-node/src/tests/neon_integrations.rs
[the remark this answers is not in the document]
gregorycoppola: OK. Can you put that in the comment then, for readers who don't realize that this method panics.

testnet/stacks-node/src/tests/neon_integrations.rs
gregorycoppola: discussed offline: as long as we are sure some "decision maker" understands that we are shipping this product change

testnet/stacks-node/src/tests/neon_integrations.rs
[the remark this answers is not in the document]
gregorycoppola: ok

testnet/stacks-node/src/tests/neon_integrations.rs
gregorycoppola: OK. as discussed offline as sure as we are aware the "product decision making body" understands this change is happening.

docs/rpc/openapi.yaml
```diff
@@ -39,6 +39,17 @@ paths:
                 $ref: ./api/transaction/post-core-node-transactions-error.schema.json
               example:
                 $ref: ./api/transaction/post-core-node-transactions-error.example.json
+      parameters:
```
gregorycoppola, line 42: Not for this line but...

Can you add what you want the CHANGELOG entry for this change to be? Either in the description or in the CHANGELOG file?

gregorycoppola (commented):
Thanks, Pavitthra.. left some replies.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5272,28 +5283,45 @@ impl StacksChainState {
...
         };
 
-        let current_tip =
-            StacksChainState::get_parent_index_block(current_consensus_hash, current_block);
-        let res = match self.with_read_only_clarity_tx(&NULL_BURN_STATE_DB, &current_tip, |conn| {
+        let current_tip = match mempool_admission_check {
+            TransactionAnchorMode::OnChainOnly | TransactionAnchorMode::Any => {
+                StacksChainState::get_parent_index_block(current_consensus_hash, current_block)
+            }
+            TransactionAnchorMode::OffChainOnly => {
```
gregorycoppola, line 5290: this whole OffChainOnly  branch is not covered by unit tests

src/chainstate/stacks/db/blocks.rs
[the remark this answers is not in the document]
gregorycoppola: this error branch isn't covered by unit tests

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5272,28 +5283,45 @@ impl StacksChainState {
...
         };
 
         match res {
             Ok(x) => Ok(x),
-            Err(MemPoolRejection::BadNonces(mismatch_error)) => {
-                // try again, but against the _unconfirmed_ chain tip, if we
-                // (a) have one, and (b) the expected nonce is less than the given one.
-                if self.unconfirmed_state.is_some()
-                    && mismatch_error.expected < mismatch_error.actual
+            Err(e) => {
```
gregorycoppola, line 5318: this whole error branch isn't covered by unit tests

src/net/http.rs
```diff
@@ -1696,6 +1698,36 @@ impl HttpRequestType {
...
+    /// If `mempool_admission_check` == "unconfirmed", the transaction is only validated against
+    /// the unconfirmed state.
+    /// If `mempool_admission_check` == "confirmed", the transaction is only validated against the
+    /// confirmed state.
+    /// If `mempool_admission_check` is unspecified, the transaction is validated against both the
+    /// confirmed state and the unconfirmed state.
+    /// This function takes the first value we can parse.
+    fn get_mempool_admission_check_query(query: Option<&str>) -> TransactionAnchorMode {
+        match query {
+            Some(query_string) => {
```
gregorycoppola, line 1711: this whole some branch isn't covered by unit tests

src/net/http.rs
```diff
@@ -2111,19 +2143,24 @@ impl HttpRequestType {
...
                     "Missing Content-Type for transaction".to_string(),
                 ));
             }
-            Some(HttpContentType::Bytes) => {
-                HttpRequestType::parse_posttransaction_octets(preamble, &mut bound_fd)
-            }
-            Some(HttpContentType::JSON) => {
-                HttpRequestType::parse_posttransaction_json(preamble, &mut bound_fd)
-            }
+            Some(HttpContentType::Bytes) => HttpRequestType::parse_posttransaction_octets(
```
gregorycoppola, line 2154: not covered by unit tests

src/net/http.rs
```diff
@@ -2111,19 +2143,24 @@ impl HttpRequestType {
...
+                preamble,
+                mempool_admission_check,
+                &mut bound_fd,
+            ),
+            Some(HttpContentType::JSON) => HttpRequestType::parse_posttransaction_json(
```
gregorycoppola, line 2159: not covered by unit tests

src/net/http.rs
```diff
@@ -2472,6 +2513,19 @@ impl HttpRequestType {
...
     }
 
+    /// Formats the query string for a post transaction request.
+    pub fn make_query_string_for_post_tx_endpoint(
+        mempool_admission_check: TransactionAnchorMode,
+    ) -> String {
+        match mempool_admission_check {
+            TransactionAnchorMode::OnChainOnly => "?mempool_admission_check=confirmed".to_string(),
+            TransactionAnchorMode::OffChainOnly => {
+                "?mempool_admission_check=unconfirmed".to_string()
```
gregorycoppola, line 2523: not tested

src/net/p2p.rs
```diff
@@ -4393,8 +4393,14 @@ impl PeerNetwork {
             return false;
         }
 
-        if let Err(e) = mempool.submit(chainstate, consensus_hash, block_hash, &tx, event_observer)
-        {
+        if let Err(e) = mempool.submit(
```
gregorycoppola, line 4396: not tested

src/net/rpc.rs
```diff
@@ -2067,7 +2069,12 @@ impl ConversationHttp {
                 }
                 None
             }
-            HttpRequestType::PostTransaction(ref _md, ref tx, ref attachment) => {
+            HttpRequestType::PostTransaction(
```
gregorycoppola, line 2072: not tested

gregorycoppola (commented):
I've added some comments based on the results of the coverage check.

See https://www.dropbox.com/s/ea4ht8ztd9o7emh/mempool-admission-logic.colored-ascii.zip?dl=0

Feel free to push back if you think there are errors, since I'm still new to the tool.

To view the coverage report for yourself unzip and run `less -R mempool-admission-logic.colored-ascii`

src/chainstate/stacks/db/blocks.rs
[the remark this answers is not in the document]
pavitthrap: I renamed the section  to errors. The failure reasons include all of the possible errors in `MemPoolRejection`, which is captured in the return type of the function.

src/chainstate/stacks/db/blocks.rs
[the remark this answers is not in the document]
pavitthrap: The rematch also deals with validating a tx against a different type of state, if appropriate. For example, if a tx is unable to be validated against the confirmed chain tip, in some cases we would attempt to validate it against the unconfirmed chain tip. We would want this second validation to happen in the case of an error, hence the reason for the match below.

src/chainstate/stacks/mod.rs
pavitthrap: They are pretty much synonyms, although I do understand your confusion here. 
I think we don't typically use the phrase OffChainOnly/OnChainOnly very often, but it's something we learn with exposure.

src/net/http.rs
pavitthrap: In line 2074 of `neon_integrations.rs`, I call `submit_tx` with the parameter `mempool_admission_check` set to `TransactionAnchorMode::OffChainOnly`. That function then calls this function. 

Did you run the neon_integrations test with the code coverage tool?

kantai (commented):
This LGTM, @pavitthrap, but I am a little concerned that the `Any` case can end up doing 2x admission checks for an invalid tx. Would it make sense for the `Any` case to just check against the unconfirmed state (if it exists)? Basically, eliminate the `Any` case and make the default `OffChainOnly`.

gregorycoppola (approved):

pavitthrap:
> This LGTM, @pavitthrap, but I am a little concerned that the `Any` case can end up doing 2x admission checks for an invalid tx. Would it make sense for the `Any` case to just check against the unconfirmed state (if it exists)? Basically, eliminate the `Any` case and make the default `OffChainOnly`.

Sure, just implemented this. So I decided to leave the type of `mempool_admission_check` as the `TransactionAnchorMode` enum. I wanted to keep it as an enum (as opposed to a bool) since I thought that was clearer.

kantai (approved):
LGTM!

kantai (changes requested):
Ooops -- I spoke too soon. It looks like the failing integration test revealed an interesting consequence of the change: it appears that the mempool admission failed because it could not find the unconfirmed MARF trie. I think this error scenario should be explored before merging.

src/chainstate/stacks/db/blocks.rs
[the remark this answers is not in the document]
jcnelson: Please don't `.unwrap()` in network-reachable code.

src/chainstate/stacks/mod.rs
jcnelson: I actually think it would be better to use a separate enum for this (or even just a bool would suffice, since you have only two options).  This enum is specifically meant to represent the way we encode the anchor mode for transactions.  It's conceptually unclear what the `::Any` option means, for example.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5226,14 +5226,24 @@ impl StacksChainState {
     }
 
     /// Check to see if a transaction can be (potentially) appended on top of a given chain tip.
-    /// Note that this only checks the transaction against the _anchored chain tip_, not the
-    /// unconfirmed microblock stream trailing off of it.
+    /// The input `mempool_admission_check` determines what the transaction is validated against.
+    /// If `mempool_admission_check`==OffChainOnly, the transaction will only be validated against unconfirmed state.
+    /// If `mempool_admission_check`==OnChainOnly, the transaction will only be validated against confirmed state.
+    /// If `mempool_admission_check`==Any, the transaction will be accepted if it validates against the unconfirmed tip.
```
jcnelson, line 5232: What's the difference between the `Any` and `OffChainOnly` case here?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1915,6 +1921,163 @@ fn microblock_integration_test() {
...
+        .send()
+        .unwrap();
+    let err_str = res.text().unwrap();
+    // Expect to get "NoSuchContract" because the function we are attempting to call is in a
+    // contract that only exists on unconfirmed state (and we set `mempool_admission_check` to
+    // OnChainOnly).
+    assert!(err_str.contains("NoSuchContract"));
+
+    // The function `submit_tx` panics if the tx is unable to be submitted.
+    // Call the contract defined in the microblock, but make mempool admission check against unconfirmed state.
```
jcnelson, line 2074: Do we want to explicitly check for a mempool admission error here?  I think we want to be making sure that, going forward, a transaction is actually rejected for the right reasons.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1915,6 +1921,163 @@ fn microblock_integration_test() {
...
+    submit_tx(&http_origin, &call_tx, TransactionAnchorMode::OffChainOnly);
+    // Call the contract defined in the microblock. Set the transaction anchor mode to `Any`, which
+    // makes mempool admission check occur against the unconfirmed state.
+    submit_tx(&http_origin, &call_tx, TransactionAnchorMode::Any);
```
jcnelson, line 2078: Can you expand this test to verify that posting a transaction that will validate against the confirmed chain tip will work, even if there's no unconfirmed state available?  This would be the case where `mempool_admission_check` is not given, and there are no unconfirmed microblocks to consider.  This is to make sure that the inner call to `maybe_read_only_clarity_tx()` gets the correct chain tip, since the unconfirmed state will exist but will not be readable (since it hasn't processed any microblocks yet).

src/chainstate/stacks/db/blocks.rs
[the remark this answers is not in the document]
gregorycoppola: Even if you check it, you still shouldn't unwrap? Is that because there is some way for the unwrap to fail after checking `is_some`? Or because it is an error prone construction and we might one day lose the `is_some` check?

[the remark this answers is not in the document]
jcnelson: The latter -- it's an error-prone construction.

jcnelson:
> Ooops -- I spoke too soon. It looks like the failing integration test revealed an interesting consequence of the change: it appears that the mempool admission failed because it could not find the unconfirmed MARF trie. I think this error scenario should be explored before merging.

Right -- I was afraid of this.  The code needs to verify that the underlying MARF trie actually exists before proceeding to use the unconfirmed chain tip.

src/chainstate/stacks/db/blocks.rs
pavitthrap: There is no difference between the `Any` and `OffChainOnly` case - I can consolidate the comments. 

I left the enum as is, and now I take it as what the user has specified. `Any` signifies the user has no specific preference, and `OffChainOnly`/`OnChainOnly` signify that the query parameter was set. Let me know if I should instead specify a new enum that omits the `Any` case.

jcnelson: I'd very much prefer a separate enum, since this enum here is used for an entirely different purpose that does not represent what the `mempool_admission_check` query parameter is doing.  The `mempool_admission_check` parameter is ultimately specifying a non-consensus-critical admission policy, and we can expect to expand the policy in the future.  The `TransactionAnchorMode` enum, by contrast, is a consensus-critical enum that lists the ways transaction anchor modes are represented and encoded in the system.  These are not even close to the same things in my mind.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5272,50 +5281,46 @@ impl StacksChainState {
             _ => false, // unused
         };
 
-        let current_tip =
-            StacksChainState::get_parent_index_block(current_consensus_hash, current_block);
-        let res = match self.with_read_only_clarity_tx(&NULL_BURN_STATE_DB, &current_tip, |conn| {
+        let current_tip = if use_unconfirmed_tip
+            && self.unconfirmed_state.is_some()
+            && self.unconfirmed_state.as_ref().unwrap().is_readable()
+        {
```
jcnelson, line 5287: Actually, I almost wonder if this code path here is actually fixing a bug in `StacksChainState::maybe_read_only_clarity_tx()` -- perhaps this method should instead be checking to see if the unconfirmed state trie actually exists?

jcnelson (changes requested):
This overall LGTM, but I must insist on not using `.unwrap()` in network code paths.

markmhx:
@pavitthrap I'm curious about where this PR stands these days and whether a release ETA might be available?

pavitthrap:
> @pavitthrap I'm curious about where this PR stands these days and whether a release ETA might be available?

Update here: https://github.com/blockstack/stacks-blockchain/issues/2774#issuecomment-922077240

jcnelson:
This will be unblocked soon with @pavitthrap's work on the microblock anti-entropy protocol.

saralab:
Stale, closing this PR

PR #2788, opened 2021-07-29
fix: switched all occurrences of max_value/min_value to MAX/MIN

pavitthrap:
## Description
Switched all occurrences of {i, u}{8, 16, 32, 64, 128}::max_value() to {i, u}{8, 16, 32, 64, 128}::MAX. Did the same for the function min_value. I did this by searching for occurrences of max_value/min_value and replacing them one by one (since `ExecutionCost` also has a method named max_value).

These functions are set to be deprecated in a future rust version: https://doc.rust-lang.org/std/primitive.i64.html#method.min_value

## Testing Information 
Ran `cargo test`.


gregorycoppola (approved):
You could add to the description how you made this change, since it is mechanical.

I mean, I'm guess it was search and replace, and maybe it's obvious, but still.

lgalabru (approved):
LGTM, thanks @pavitthrap!

PR #2799, opened 2021-08-03
chore: Replace code that `cargo fmt` always wants to fix

gregorycoppola:
# Description
This change moves a comment so that `cargo fmt` will think the file is well-formatted.

# Motivation
With `cargo --version` at `cargo 1.52.0 (69767412a 2021-04-21)`, my `cargo fmt` wants to format this file *every time* I run it. It seems this is because there is a comment inside an "else if" block and inside of an assignment.

# Testing
Should have no effect.

### DIFF
--- a/src/vm/ast/definition_sorter/mod.rs
+++ b/src/vm/ast/definition_sorter/mod.rs
@@ -211,9 +211,10 @@ impl<'a> DefinitionSorter {
                                     return Ok(());
                                 }
                             }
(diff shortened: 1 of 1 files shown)


pavitthrap (approved):

reedrosenbluth (approved):

gregorycoppola:
thanks for the reviews, everyone!

PR #2802, opened 2021-08-04
Read Only Functions for Traits

gregorycoppola:


gregorycoppola:
Barely at draft stage, really.

PR #2803, opened 2021-08-04
ci: update clarity-js-sdk gh workflow

CharlieC3:
## Description

Updates the clarity-js-sdk Github workflow to:
* Remove Hank as a reviewer of new PRs
* Update the same PR if it's still open instead of creating a new PR for each release
* Only run on releases and not pre-releases.
* Add a `dependencies` label to the PR

For details refer to issue https://github.com/hirosystems/devops/issues/750

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

No

No

## Checklist

### DIFF
--- a/.github/workflows/clarity-js-sdk-pr.yml
+++ b/.github/workflows/clarity-js-sdk-pr.yml
@@ -11,7 +11,7 @@ env:
 on:
   release:
     types:
-      - published
+      - released
   workflow_dispatch:
 
 jobs:      
@@ -29,7 +29,6 @@ jobs:
         run: |
           RELEASE_VERSION=$(echo ${GITHUB_REF#refs/*/} | tr / -)
           echo "RELEASE_VERSION=$RELEASE_VERSION" >> $GITHUB_ENV
-          echo "UPDATE_BRANCH=auto/update-bin-$RELEASE_VERSION" >> $GITHUB_ENV
 
       - name: Update SDK Tag
         run: sed -i "s@CORE_SDK_TAG = \".*\"@CORE_SDK_TAG = \"$RELEASE_VERSION\"@g" packages/clarity-native-bin/src/index.ts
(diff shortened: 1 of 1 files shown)


gregorycoppola (approved):
don't really understand every line but looks ok

.github/workflows/clarity-js-sdk-pr.yml
```diff
@@ -11,12 +11,13 @@ env:
...
     types:
-      - published
+      - released
   workflow_dispatch:
 
-jobs:      
+jobs:
   run:
     runs-on: ubuntu-latest
+    if: "!contains(github.ref, '-rc')"
```
pavitthrap, line 20: What is this line doing?

.github/workflows/clarity-js-sdk-pr.yml
```diff
@@ -11,12 +11,13 @@ env:
 on:
   release:
     types:
-      - published
+      - released
```
pavitthrap, line 14: What exactly is the difference between the `released` event and the `published` event in GitHub actions?

pavitthrap (approved):
lgtm, just had a few questions out of curiousity

.github/workflows/clarity-js-sdk-pr.yml
CharlieC3: It's a condition to ensure the newly created release does not have `-rc` in its name, otherwise it'll exit. We don't want the `clarity-native-bin` to use release candidate pre-releases, so this is how it's being circumnavigated for now. This is sort of a hack to make up for maintenance that needs to be completed on the main `stacks-blockchain` workflow.

Ideally the workflow would create a pre-release instead of an actual release when a tag with `-rc` in the name is passed in. If it worked this way, then the type change from `published` to `released` would've sufficed (see my other comment). But since it doesn't do that at the moment, and instead it creates a real release which I then have to modify later to make it a pre-release.

Once the workflow is improved to create pre-releases when appropriate, this condition can be removed.

.github/workflows/clarity-js-sdk-pr.yml
CharlieC3: The `released` type will trigger when a new release is created, the `published` type will trigger when a new release **or pre-release** is published. The difference between created vs published isn't so important here, but I don't want this job running when new pre-releases are created. I've been using `released` in `clarinet` and it seems to work much better!

.github/workflows/clarity-js-sdk-pr.yml
pavitthrap: Cool, could possibly be helpful to leave a short comment with some of this info in the yaml file, but up to you!

CharlieC3: Good idea, comment added!

Issue #2797, opened 2021-08-02
Create a Comments Policy for Stacks Blockchain Code


gregorycoppola:
We need a more specific style guide for the `stacks-blockchain` repo.

There has been some confusion about what should be in the comments.

gregorycoppola:
Make sure that this supports Rust Docs.

PR #2804, opened 2021-08-05
feat: Add comments and clarify function names for `ReadOnlyChecker`

gregorycoppola:
We are trying to add more comments to the code (#2797).

Ahead of changing `ReadOnlyChecker`, we make the following two kinds of change:

1. add class-level and function-level comments, especially to clarify how the roles of the different private functions differ
2. change the names of the internal private functions to reflect how the roles of the functions differ, whereas previously the names were things like `check_read_only`, which is not descriptive for an internal function name, because the entire point of the class is to test for read-only compliance


src/vm/analysis/read_only_checker/mod.rs
```diff
@@ -77,97 +93,132 @@ impl<'a, 'b> ReadOnlyChecker<'a, 'b> {
...
+    fn check_each_expression_is_read_only(
+        &mut self,
+        expressions: &[SymbolicExpression],
+    ) -> CheckResult<bool> {
         let mut result = true;
-        for expr in expressions.iter() {
-            let expr_read_only = self.check_read_only(expr)?;
+        for expression in expressions.iter() {
+            let expr_read_only = self.check_atomic_expression_is_read_only(expression)?;
             result = result && expr_read_only;
```
pavitthrap, line 198: Do you understand why we don't do an early return is `expr_read_only` is ever false?

src/vm/analysis/read_only_checker/mod.rs
```diff
@@ -290,27 +341,31 @@ impl<'a, 'b> ReadOnlyChecker<'a, 'b> {
...
                     .map(|args_read_only| args_read_only && is_function_read_only)
             }
         }
     }
 
-    fn check_function_application_read_only(
+    /// Checks the native function application implied by `expressions`. The first
+    /// argument is used as the function name, and the tail is used as the arguments.
+    ///
+    /// Returns `true` iff the function application is read-only.
```
pavitthrap, line 353: Could add section for Errors here & mention the possible `Unknown Function` error.

pavitthrap (approved):
LGTM!

src/vm/analysis/read_only_checker/mod.rs
```diff
@@ -290,27 +341,31 @@ impl<'a, 'b> ReadOnlyChecker<'a, 'b> {
...
 
-                self.check_all_read_only(&args[2..])
+                self.check_each_expression_is_read_only(&args[2..])
                     .map(|args_read_only| args_read_only && is_function_read_only)
             }
         }
     }
 
-    fn check_function_application_read_only(
+    /// Checks the native function application implied by `expressions`. The first
```
kantai, line 350: This checks any function application -- not _just_ native function applications (e.g. if `expressions` is a user-defined-function application).

gregorycoppola: done

src/vm/analysis/read_only_checker/mod.rs
gregorycoppola: thanks!

src/vm/analysis/read_only_checker/mod.rs
gregorycoppola: Not sure.. @kantai ? Should we return early?

gregorycoppola (commented):
thanks 4 the reviews.. Just had one question for @kantai about @pavitthrap 's early return question.

src/vm/analysis/read_only_checker/mod.rs
kantai: Yes -- this method shouldn't return early, because it's responsible not only for checking that a given expression isn't read-only, but also enforcing a read-only constraint (see the thrown CheckErrors)

gregorycoppola: ok.. i added a note too

gregorycoppola (commented):
I tried to make sure the `#Errors` were captured for every method, allowing cascading errors to be captured under broad categories like "contract parsing errors".

Jude brought up the issue of how to document inherited errors in a way that isn't very burdensome for the author so maybe characterizing inherited errors with broad classes would be a workable way to do this.

kantai (approved):
LGTM

PR #2809, opened 2021-08-10
test: using rstest, made it simpler to test multiple clarity versions

pavitthrap:
Added way to easily run the same test for multiple clarity versions.

### Impact 
Added the crates `rstest` and `rstest_reuse` as dev-dependencies.


gregorycoppola (approved):
Looks OK as far as I can tell, not knowing before about `rstest`. Thanks for finding this resource.

pavitthrap:
I tried to move the macro generated by the rstest package to `src/vm/version.rs` and then import the generated macro in the test file. I tried many strategies to fix the import, and after discussing with @kantai, the import seems too difficult to fix. 

I also considered writing a procedural macro to accomplish what the library does. However, macros of this type must be defined in a procedural macro crate, so I ruled out this possibility. 

I'm going to go ahead with the approach of defining a new macro in each test file that needs it.

kantai (approved):

PR #2812, opened 2021-08-13
chore: Log all outcomes for a transaction

gregorycoppola:
## Description
Currently, it is possible for transactions to fail without any error message being put out by the miner. This makes it harder to debug. With this PR we attempt to log all outcomes.

## Testing
We'll have to run this in the field, with a mock miner, to know if we are still missing cases. It should have no impact on the code.

### DIFF
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -254,6 +254,7 @@ impl<'a> StacksMicroblockBuilder<'a> {
         let miner_pubkey_hash =
             Hash160::from_node_public_key(&StacksPublicKey::from_private(miner_key));
         if txs.len() == 0 {
+            warn!("Transaction skipped: Error::NoTransactionsToMine");
             return Err(Error::NoTransactionsToMine);
         }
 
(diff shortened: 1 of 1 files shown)


gregorycoppola:
1) Make sure we can write a query for each transaction and see the outcome.

2) Consider topic-based logging, which means we can filter logs per topic. This avoids being "flooded" with logs. This would be a parameter added when the executable is run.

gregorycoppola:
@kantai This PR is missing the "structured logging" and doesn't include block height but I thought it could use some intermediate feedback, because there are a lot of choices made already.

gregorycoppola:
@jcnelson @kantai I converted this PR to a draft because, after investigating the `STACKS_TRANSACTION_LOG` feature that Jude brought up, I believe maybe we should expand that approach, rather than what I have done here.

But, some feedback would be appreciated.

The idea is that, rather than have O(N) haphazard log points in the code, we expand/arrange the interface so we can capture all the events in a batch, and have O(1), or literally 1, log points at the end, as the `STACKS_TRANSACTION_LOG` does. The only thing is we need to be sure to cover all cases.

This is discussed in the issue comment:
https://github.com/blockstack/stacks-blockchain/issues/2815#issuecomment-901188898

gregorycoppola:
I reverted this so it should match the logs in https://kibana.blockstack.xyz/s/blockstack/goto/7340f2ad155ad49e0969a721a2f90dc1

PR #2818, opened 2021-08-19
feat: gate ast passes

### LINKED ISSUE #2759: 2.1: Gate `definition-sorter` pass
The definition sorter is not a "normal" analysis pass (it's not an analysis pass at all -- that's why it's part of the AST module). All it is trying to do in this loop is figure out whether or not a particular function application is a "special" application where the arguments aren't dereferenced (e.g., the tuple keys in a tuple constructor). As long as the new Clarity version doesn't add a new branch to this match, it is fine to just use `ClarityVersion` latest here. However, it would make sense to future-proof this better by gating the `definition-sorter` and other AST passes.

pavitthrap:
Fixes #2759 

I added clarity version as a parameter to all of the ast passes. For the time being, this parameter is unused.

For all tests that used functions related to the ast passes, such as `build_ast` or `parse`, I defined and used a macro that enabled me to test the same snippet of code across multiple clarity versions. 

## Notes 
- In `vm/docs/mod.rs`, the documentation examples are now tested against the latest version of clarity only 
- The way the clarity version is determined in different parts of the code could be worth scrutiny during review. In some places, the "default for epoch" is used to get the version, and in other places, the version is extracted from the contract context.

### DIFF
--- a/src/clarity.rs
+++ b/src/clarity.rs
@@ -79,6 +79,7 @@ use crate::types::chainstate::StacksAddress;
 use crate::types::chainstate::StacksBlockId;
 use crate::types::chainstate::VRFSeed;
 use crate::types::proof::ClarityMarfTrieId;
+use std::str::FromStr;
 
 #[cfg(test)]
 macro_rules! panic_test {
@@ -141,8 +142,9 @@ struct EvalInput {
 fn parse(
     contract_identifier: &QualifiedContractIdentifier,
     source_code: &str,
+    version: ClarityVersion,
 ) -> Result<Vec<SymbolicExpression>, Error> {
-    let ast = build_ast(contract_identifier, source_code, &mut ())
+    let ast = build_ast(contract_identifier, source_code, &mut (), version)
         .map_err(|e| RuntimeErrorType::ASTError(e))?;
     Ok(ast.expressions)
 }
@@ -692,6 +694,7 @@ fn consume_arg(
     }
 }
 
+/// This function uses Clarity1 to parse the boot code.
 fn install_boot_code<C: ClarityStorage>(header_db: &CLIHeadersDB, marf: &mut C) {
     let mainnet = header_db.is_mainnet();
     let boot_code = if mainnet {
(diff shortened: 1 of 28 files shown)


src/clarity.rs
```diff
@@ -692,6 +694,7 @@ fn consume_arg(
     }
 }
 
+/// This function uses Clarity1 to parse the boot code.
```
gregorycoppola, line 697: Is this going to use Clarity1 until some point in time? Or will always do so?

gregorycoppola (approved):
I agree that it seems like a potential problem that ClarityVersion sometimes comes from the epoch and sometimes is a parameter. But, seems like the answer, if any, would be a separate PR.

src/clarity.rs
pavitthrap: I'm not sure, I'd like feedback on this as well. I updated the description to reflect this.

kantai: Yes -- this should always use Clarity1, unless there's some sort of catastrophic bug. The "boot contracts" (or maybe more aptly named genesis contracts) shouldn't be modified in any future network upgrades, so they'd remain Clarity1.

src/vm/types/signatures.rs
```diff
@@ -1054,6 +1045,12 @@ impl TypeSignature {
         }
         Ok(trait_signature)
     }
+
+    pub fn from_string(val: &str, version: ClarityVersion) -> Self {
+        use vm::ast::parse;
+        let expr = &parse(&QualifiedContractIdentifier::transient(), val, version).unwrap()[0];
+        TypeSignature::parse_type_repr(expr, &mut ()).unwrap()
+    }
```
kantai, line 1053: Can you mark this `#[cfg(test)]`? The previous `From<&str>` implementation was mostly just used as a convenience for constructing test assertions, and I wouldn't want to see this getting used in production.

kantai (approved):
This LGTM -- just had a small comment about marking something as testing-only.

PR #2820, opened 2021-08-25
feat(Miner): Structured Type System for Logging

gregorycoppola:
## Description

The goal of this PR is to provide more robust logging for the miner, so that all outcomes for mining a transaction are logged, and to make it as "hard as possible" for neglect to log an outcome.

To do this, we use the type system. `MemPoolDB::iterate_candidates` now requires a function that returns a new kind of object `TransactionResult`.

There are three kinds of `TransactionResult`: `Success`, `Error` and `Skipped`.

In order to create one of these types, the user *should* use a factory function, which will do the logging as a side-effect.

Also, the different kinds of events are collected, and can be used for further logging later.

## Testing
I am running this in a local mock miner and observing the logs. We will also run this on Hiro's mock miner.

It doesn't really admit of unit tests right now, I think.

## Limitations

It is currently still possible to create a `TransactionResult` without using the factory methods, and that would mean that nothing gets logged. I'm not sure if it's technically possible to force the user to use the factory methods the first time an object is created, while also allowing them to decompose and reassemble objects in `match` blocks in Rust.


gregorycoppola:
I should have written this PR against `develop` instead of `master`.

PR #2821, opened 2021-08-25
feat: Compiler-Enforced Logging of All Miner Transaction Outcomes

gregorycoppola:
## Description

The goal of this PR is to provide more robust logging for the miner, so that:
1) all outcomes for mining a transaction are logged, and
2) we make it as "hard as possible" to neglect to log an outcome

To do this, we use the type system. `MemPoolDB::iterate_candidates` will now require to be given a new kind of object called `TransactionResult`.

There are three kinds of `TransactionResult`: `TransactionSuccess`, `TransactionError` and `TransactionSkipped`.

In order to create one of these types, the user *should* use a factory function, which will do the logging as a side-effect.

The logs look like this:

```
# Error
WARN [1629921614.330687] [src/chainstate/stacks/db/queryable_logging.rs:31] [relayer] Transaction outcome for tx=491493d01fc8c0f21f0891ad7c378bfb9c7b83142524f3320ea609e6b5cffa90: failed with error: Bad nonce: origin account SP3WV3VC6GM1WF215SDHP0MESQ3BNXHB1N6TPB70S nonce of tx 491493d01fc8c0f21f0891ad7c378bfb9c7b83142524f3320ea609e6b5cffa90 is 9656 (expected 9680)

# Success
INFO [1629921614.162689] [src/chainstate/stacks/db/queryable_logging.rs:25] [relayer] Transaction outcome for tx=d2fa7f52421891a6e0f0c5c1a6a7db71e812637b4b12ae1e49524c8656c48b89: successfully processed.
```

Also, the different kinds of events are collected, and can be used for further logging later.

## Testing
I am running this in a local mock miner and observing the logs. We will also run this on Hiro's mock miner.

It doesn't really admit of unit tests right now, I think.

## Limitations

It is currently still possible to create a `TransactionResult` without using the factory methods, and that would mean that nothing gets logged. I'm not sure if it's technically possible to force the user to use the factory methods the first time an object is created, while also allowing them to decompose and reassemble objects in `match` blocks in Rust.

### DIFF
--- a/src/chainstate/stacks/db/queryable_logging.rs
+++ b/src/chainstate/stacks/db/queryable_logging.rs
@@ -10,8 +10,14 @@ use chainstate::stacks::Error;
 /// Logs a queryable message for the case where `txid` has succeeded.
 use chainstate::stacks::StacksTransaction;
 
+/// Returns a prefix for a "queryable logging" statement. This is the "queryable" part, which is to
+/// say that the prefix will be the same no matter what the outcome is, so that a fixed regex will
+/// match this part for any transaction.
+///
+/// Format looks like this:
+/// "Transaction outcome for tx=491493d01fc8c0f21f0891ad7c378bfb9c7b83142524f3320ea609e6b5cffa90: "
 fn create_transaction_key(tx: &StacksTransaction) -> String {
-    format!("Transaction outcome {}:", &tx.txid())
+    format!("Transaction outcome for tx={}:", &tx.txid())
 }
 
 /// Logs a queryable message for the case where `txid` has succeeded.
--- a/src/chainstate/stacks/db/queryable_logging.rs
+++ b/src/chainstate/stacks/db/queryable_logging.rs
@@ -1,0 +1,36 @@
+//! The methods in this file implement a "queryable" form for logging the outcomes of transactions.
+//!
+//! Each processed transaction should be logged if it succeeds, logged if it fails with an error,
+//! and should be logged each time it is skipped.
+//!
+//! This way, when debugging a miner, we can run a standard query, and see what happened to the
+//! transaction.
+//!
+use chainstate::stacks::Error;
+/// Logs a queryable message for the case where `txid` has succeeded.
+use chainstate::stacks::StacksTransaction;
+
+fn create_transaction_key(tx: &StacksTransaction) -> String {
+    format!("Transaction outcome {}:", &tx.txid())
+}
+
+/// Logs a queryable message for the case where `txid` has succeeded.
+pub fn log_transaction_success(tx: &StacksTransaction) {
+    info!("{} successfully processed.", create_transaction_key(tx));
+}
+
+/// Logs a queryable message for the case where `txid` has failed
+/// with error `err`.
+pub fn log_transaction_error(tx: &StacksTransaction, err: &Error) {
+    warn!("{} failed with error: {}", create_transaction_key(tx), err);
+}
+
+/// Logs a queryable message for the case where `tx` has been skipped
+/// for reason `reason`.
+pub fn log_transaction_skipped(tx: &StacksTransaction, reason: String) {
+    info!(
+        "{} skipped for reason: {}",
+        create_transaction_key(tx),
+        reason
+    );
+}
--- a/src/chainstate/stacks/db/queryable_logging.rs
+++ b/src/chainstate/stacks/db/queryable_logging.rs
@@ -15,9 +15,9 @@ use chainstate::stacks::StacksTransaction;
 /// match this part for any transaction.
 ///
 /// Format looks like this:
-/// "Transaction outcome for tx=491493d01fc8c0f21f0891ad7c378bfb9c7b83142524f3320ea609e6b5cffa90: "
+/// "Transaction outcome 491493d01fc8c0f21f0891ad7c378bfb9c7b83142524f3320ea609e6b5cffa90: "
 fn create_transaction_key(tx: &StacksTransaction) -> String {
-    format!("Transaction outcome for tx={}:", &tx.txid())
+    format!("Transaction outcome {}:", &tx.txid())
 }
 
 /// Logs a queryable message for the case where `txid` has succeeded.
--- a/src/chainstate/stacks/db/transactions.rs
+++ b/src/chainstate/stacks/db/transactions.rs
@@ -1123,9 +1123,9 @@ impl StacksChainState {
         }
     }
 
-		/// This method implements the contract of `process_transaction`, and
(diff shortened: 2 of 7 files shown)


src/chainstate/stacks/db/queryable_logging.rs
```diff
@@ -0,0 +1,42 @@
...
+/// Logs a queryable message for the case where `txid` has succeeded.
+use chainstate::stacks::StacksTransaction;
+
+/// Returns a prefix for a "queryable logging" statement. This is the "queryable" part, which is to
+/// say that the prefix will be the same no matter what the outcome is, so that a fixed regex will
+/// match this part for any transaction.
+///
+/// Format looks like this:
+/// "Transaction outcome 491493d01fc8c0f21f0891ad7c378bfb9c7b83142524f3320ea609e6b5cffa90: "
+fn create_transaction_key(tx: &StacksTransaction) -> String {
```
jcnelson, line 19: If we're going through the effort of building a queryable logging system for transactions, you might as well use the existing structured logging logic to ensure that downstream consumers receive a semantic representation of the transaction status (i.e. something machine-readable).

src/chainstate/stacks/db/transactions.rs
```diff
@@ -1122,8 +1123,10 @@ impl StacksChainState {
         }
     }
 
-    /// Process a transaction.  Return the fee and the transaction receipt
-    pub fn process_transaction(
+    /// This method implements the contract of `process_transaction`, and
+    /// exists as an intermediate step so that its output can be captured and
+    /// logged.
+    fn process_transaction_internal(
```
jcnelson, line 1129: The naming convention in this codebase for such "internal" functions is to prefix it with `inner_`.  This should be `inner_process_transaction`.

src/chainstate/stacks/events.rs
```diff
@@ -63,6 +65,128 @@ pub struct StacksTransactionReceipt {
...
+}
+
+/// Represents a transaction that was skipped, but might succeed later.
+#[derive(Debug, Clone, PartialEq)]
+pub struct TransactionSkipped {
+    pub tx: StacksTransaction,
+    pub reason: String,
+}
+
+/// `TransactionResult` represents the outcome of transaction processing.
```
jcnelson, line 92: Really, this is a `MiningResult`.  A transaction cannot be skipped by the blockchain-processing logic.  That the miner uses the blockchain-processing logic to produce a block does not mean that the blockchain-processing logic needs to be aware of the decision to include or exclude the transaction when it is mined.

src/chainstate/stacks/miner.rs
```diff
@@ -1350,6 +1425,105 @@ impl StacksBlockBuilder {
...
+    ///  On `BlockTooBigError` or `TransactionTooBigError`, we increment the value of `block_limit_hit`.
+    fn process_candidate_for_anchored_block<'a>(
+        txinfo: MemPoolTxInfo,
+        epoch_tx: &mut ClarityTx<'a>,
+        builder: &mut StacksBlockBuilder,
+        considered: &mut HashSet<Txid>,
+        mined_origin_nonces: &mut HashMap<StacksAddress, u64>,
+        mined_sponsor_nonces: &mut HashMap<StacksAddress, u64>,
+        invalidated_txs: &mut Vec<Txid>,
+        block_limit_hit: &mut BlockLimitFunction,
```
jcnelson, line 1446: The fact that you have to pass so many `&mut` references to this method from the caller's scope suggests to me that this is not the right way factor this.

src/chainstate/stacks/miner.rs
```diff
@@ -731,94 +775,115 @@ impl StacksBlockBuilder {
...
+                    ),
+                );
             }
 
-            let (fee, _receipt) = StacksChainState::process_transaction(clarity_tx, tx, quiet)
-                .map_err(|e| match e {
+            let transaction_result = StacksChainState::process_transaction(clarity_tx, tx, quiet);
+            match transaction_result {
+                TransactionResult::Success(TransactionSuccess { tx, fee, receipt }) => {
+                    info!("Include tx";
```
jcnelson, line 790: Why isn't this logging code in the `queryable_logging.rs` file?

src/chainstate/stacks/miner.rs
```diff
@@ -292,21 +304,37 @@ impl<'a> StacksMicroblockBuilder<'a> {
     }
 
     /// Mine the next transaction into a microblock.
-    /// Returns true/false if the transaction was/was not mined into this microblock.
+    ///
+    /// This is a wrapper around `StacksChainState::process_transaction` that also checks certain
```
jcnelson, line 308: "Wrapper" is not really the right word here.  The `mine_next_transaction` method creates a block by processing a sequence of transactions it pulls from the mempool.  The inner call to `StacksChainState::process_transaction` is used to faithfully create the MARF trie for this block and calculate the MARF state root for the block header.  This is not the same as being a wrapper, since `mine_next_transaction` is not meant to be used to process blocks.

src/chainstate/stacks/events.rs
```diff
@@ -63,6 +65,128 @@ pub struct StacksTransactionReceipt {
...
+    ///
+    /// Otherwise crashes.
+    pub fn unwrap_err(self) -> Error {
+        match self {
+            TransactionResult::Error(TransactionError { tx: _, error }) => error,
+            _ => panic!("Tried to `unwrap_error` a non-error result."),
+        }
+    }
+}
+
```
jcnelson, line 189: Are these transaction results pushed to the event observer in any way?

jcnelson (changes requested):
There are a couple of problems with this PR, but they're easily fixed.

First, it doesn't seem to actually be logging anything in a queryable way, beyond just grepping the log file.  It does seem to send the structured transaction-mining results to the event observer either.  Do you intend to add either of these things?  If not -- if the point of this PR is just to change what the miner prints out -- then a lot of this PR can be dropped and consolidated into `src/chainstate/stacks/miner.rs` as simply a set of structured types to represent each of the miner's decisions for a transaction.

Second, all of the transaction outcomes we're interested in are the concern of the _miner_, not the block database.  The files in `src/chainstate/stacks/db` should not be touched by this PR at all, since they are not related to deciding which transactions to include and which to skip.  The miner, which uses the block database to produce blocks, should instead interpret the result of processing each transaction it considers for mining on its own before deciding how to record the outcome of considering it.

gregorycoppola (commented):
I'm leaving some responses to the reviewers.

This PR has some bugs that I will leave comments on in the a new round.

gregorycoppola:
Closed for https://github.com/blockstack/stacks-blockchain/pull/2975

PR #2823, opened 2021-08-27
Fix/2772

jcnelson:
This alters the mempool mining heuristic so as to consider high-fee transactions before low-fee transactions, and alters the mining behavior to try and mine a "small" block as soon as possible after discovering a new burnchain block or Stacks block, and then later on, try and mine a "big" block by RBF'ing the small block's commit later.  It achieves this by making both mempool iteration and block-building *interruptable*, so that once a soft deadline passes, the node will attempt to mine the transactions it had considered so far in the allotted time.

This is a draft for now, since it's undergoing live testing.


src/core/mempool.rs
```diff
@@ -386,52 +435,136 @@ impl MemPoolDB {
...
-                    Some(next_nonce),
-                )? {
-                    None => {
-                        return Ok(());
+            if origin_addresses.len() == 0 {
+                debug!("No more origin addresses to consider");
+                break;
+            }
+
+            for origin_address in origin_addresses.iter() {
```
gregorycoppola, line 500: It seems to me like what you are trying to do here is:
1) get the origin addresses that are trying to spend the most
2) for those addresses, find the lowest nonces.

There is something to be said for the "simpler" strategy that you just sort by transaction fee.

For exmaple, to "cancel" a transaction in ETH, you submit a "0 transfer" transaction, with the same nonce, and a higher transaction fee than the transaction you want to cancel.

https://info.etherscan.com/how-to-cancel-ethereum-pending-transactions/

Maybe you can 1) sort by transaction fee, and then, to break ties 2) sort by nonce. This would support chaining.

But, if there is no tie on transaction fee, then don't involve the nonces, to allow transaction cancelation.

Users should have the awareness on their side to not submit the chained transaction (nonce = n + k) with a higher fee than the base transaction that is still to process (nonce = n).

gregorycoppola (commented):
Hey Jude,

I just had a question about the sorting criterion...

src/core/mempool.rs
jcnelson: None of the Hiro tooling currently supports RBF (ethereum's "cancel" operation), so if the logic only ordered by highest fee, your transactions could get stuck indefinitely under mempool congestion unless you're willing to manually change the fee of your transaction and re-sign and re-submit it (something only developers can be expected to do, and even they wouldn't be happy about it).  In this query, because the cumulative transaction fee for an origin is what gets used to prioritize transactions, you can use the current tooling to un-stuck your transactions simply by sending more of them.

jcnelson:
Tests are currently failing due to an unrelated build error, seemingly brought by the `proc-macro-hack` package:

```
#10 9.489   = note: /usr/bin/ld: /usr/local/rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libprofiler_builtins-3db8bb50be1aec25.rlib(InstrProfilingPlatformLinux.o): relocation R_X86_64_PC32 against undefined hidden symbol `__ehdr_start' can not be used when making a shared object
#10 9.489           /usr/bin/ld: final link failed: nonrepresentable section on output
#10 9.489           collect2: error: ld returned 1 exit status
#10 9.489           
#10 9.490 
#10 9.494 error: could not compile `proc-macro-hack` due to previous error
```

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1915,6 +1915,191 @@ fn microblock_integration_test() {
     channel.stop_chains_coordinator();
 }
 
+#[test]
+#[ignore]
+fn filter_low_fee_tx_integration_test() {
```
kantai, line 1920: Should these new tests be added to the integrations dockerfile?

testnet/stacks-node/src/config.rs
```diff
@@ -1085,6 +1106,25 @@ impl NodeConfig {
     }
 }
 
+#[derive(Clone, Debug, Default)]
+pub struct MinerConfig {
+    pub min_tx_fee: u64,
+    pub min_cumulative_tx_fee: u64,
+    pub first_attempt_time_ms: u64,
+    pub subsequent_attempt_time_ms: u64,
+}
```
kantai, line 1115: Can you document these new config options somewhere in this repo? Does the README.md make sense?

testnet/stacks-node/src/neon_node.rs
```diff
@@ -155,6 +158,31 @@ struct MiningTenureInformation {
...
+                // first attempt to mine a block -- do so right away
+                config.miner.first_attempt_time_ms
+            } else {
+                // second or later attempt to mine a block -- give it some time
+                config.miner.subsequent_attempt_time_ms
+            },
+        },
+    }
+}
+
```
kantai, line 185: I think this makes more sense as a method of `BlockBuilderSettings`, e.g., `BlockBuilderSettings::from_config` or the like, and separated from the `neon_node` module.

testnet/stacks-node/src/neon_node.rs
```diff
@@ -1356,12 +1399,6 @@ impl InitializedNeonNode {
         if let Some(burnchain_tip) = self.last_burn_block.clone() {
             match self.leader_key_registration_state {
                 LeaderKeyRegistrationState::Active(ref key) => {
-                    debug!(
-                        "Tenure: will wait for {}s before running tenure off of {}",
-                        self.sleep_before_tenure / 1000,
-                        &burnchain_tip.burn_header_hash
-                    );
-                    sleep_ms(self.sleep_before_tenure);
```
kantai, line 1364: Does this change remove `sleep_before_tenure` entirely as a config option?

src/core/mempool.rs
```diff
@@ -386,52 +435,136 @@ impl MemPoolDB {
...
+        test_debug!(
+            "Mempool walk for {}ms, min tx fee {}, min cumulative tx fees {}",
+            settings.max_walk_time_ms,
+            min_tx_fee,
+            min_cumulative_fees
+        );
+
+        while !done {
+            if deadline <= get_epoch_time_ms() {
+                test_debug!("Mempool iteration deadline exceeded");
```
kantai, line 475: Should be a `debug`?

src/core/mempool.rs
```diff
@@ -699,6 +756,15 @@ impl MemPoolDB {
         let add_tx = if let Some(ref prior_tx) = prior_tx {
             if tx_fee > prior_tx.tx_fee {
                 // is this a replace-by-fee ?
+                test_debug!(
```
kantai, line 759: Should be a `debug`?

src/core/mempool.rs
```diff
@@ -709,6 +775,13 @@ impl MemPoolDB {
                 block_header_hash,
             )? {
                 // is this a replace-across-fork ?
+                test_debug!(
```
kantai, line 778: Should be a `debug`?

src/core/mempool.rs
kantai: I think that using the raw fee of transactions, rather than the cumulative fee of chained transactions would be a better approach here. With the way it is currently written, this mining logic will _strongly_ favor the chattiest transaction submitters, such that if blocks are all full, you'd expect that any origin that only submits one transaction at a time will almost always get crowded out. I know that most tooling doesn't support RBF yet, but that is an easier change for wallets, libraries, etc. to roll out (and a feature that they'd want to support anyways), rather than miner block assembly, so I'd definitely advocate for using high fees.

testnet/stacks-node/src/neon_node.rs
jcnelson: Yup

kantai:
Can you add a `CHANGELOG.md` entry for this PR?

testnet/stacks-node/src/neon_node.rs
jcnelson: You can't `impl` a method for a struct in a different crate.

testnet/stacks-node/src/config.rs
jcnelson: Is there any mining documentation in the repo anywhere I can add to?

src/core/mempool.rs
jcnelson: Done

src/core/mempool.rs
jcnelson: Done

src/core/mempool.rs
jcnelson: Done

testnet/stacks-node/src/neon_node.rs
jcnelson: Removed

testnet/stacks-node/src/config.rs
jcnelson: Added a section to the README

testnet/stacks-node/src/tests/neon_integrations.rs
jcnelson: Done

src/core/mempool.rs
jcnelson: Yup; changed to prioritizing by just tx fee

jcnelson:
> Can you add a CHANGELOG.md entry for this PR?

Added

testnet/stacks-node/src/neon_node.rs
kantai: Right -- I think this code shouldn't be in `neon_node` module (or stacks-node crate) -- it should live with the BlockBuilderSettings. But I guess that'd introduce a circular dependency on `Config`, so maybe just move this to `config.rs` and make it a member of `Config` like `config.make_block_builder_settings(attempt)`.

CHANGELOG.md
```diff
@@ -7,6 +7,24 @@ and this project adheres to the versioning scheme outlined in the [README.md](RE
 
 ## [2.0.11.2.0]
```
kantai, line 8: Change the subheader to either `## [Not Yet Released]` or `## [2.0.11.3.0]`

src/core/mempool.rs
kantai: Can you change the iteration to no longer consider all of an origin's transaction as a block? This new ordering still would end up favoring chatty submitters (if I have a single high fee transaction, then all of my transactions would be considered for inclusion before considering anyone else's transactions).

jcnelson: The downside to this is that it will lead to starvation if there are few distinct origin addresses (i.e. when the mempool isn't congested).  Maybe I can fix this too -- now that `iterate_candidates()` takes a `ClarityTx`, the miner ought to be able to do multiple mempool passes when filling a block, thereby mining multiple transactions from the same origin but only after all other origins have been considered.

kantai: I think that's fine -- I'm more worried about the current behavior where a single origin address could monopolize the miner's attempts just by having one transaction be the highest fee tx.

jcnelson: I made it so that all origin addresses that clear the minimum fee will be considered in a round-robin fashion, and that if time and space permits, the miner will continue to cycle through origin addresses until the block is filled.  I've added an integration test that verifies that multiple transactions from the same origin will be included in a block, but *only after* all origins have been considered first.  Sound good?

kantai: Perfect!

kantai (approved):
This LGTM once the tests pass!

testnet/stacks-node/src/config.rs
```diff
@@ -1085,6 +1106,25 @@ impl NodeConfig {
     }
 }
 
+#[derive(Clone, Debug, Default)]
```
gregorycoppola, line 1106: The main part of this PR changes the ranking algorithm for the transactions.

Do we know that if you change that algorithm, you still need all these other changes?

gregorycoppola (approved):
LGTM because whatever you want to submit is fine with me.

However, I did have a question I am curious about. This PR does two things:
1) change the ranking algorithm for picking transactions
2) add settings that miners can configure

Do you know the relative contributions of the two? Would it be enough to just do (1)?

SyAsteria (approved):
LGTM. This should help clear things up in the mempool rn ⚡️

testnet/stacks-node/src/neon_node.rs
```diff
@@ -990,27 +1016,44 @@ fn spawn_miner_relayer(
...
                         "height" => last_burn_block.block_height,
-                        "burn_header_hash" => %burn_header_hash
+                        "burn_header_hash" => %burn_chain_tip,
+                        "last_burn_header_hash" => %burn_header_hash
                     );
 
-                    let burn_chain_tip = SortitionDB::get_canonical_burn_chain_tip(sortdb.conn())
-                        .expect("FATAL: failed to query sortition DB for canonical burn chain tip")
-                        .burn_header_hash;
-                    if config.node.mock_mining && failed_to_mine_in_block.as_ref() == Some(&burn_chain_tip) {
```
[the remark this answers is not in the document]
jcnelson: Because I want the mock miner to try and re-try building better and better blocks, such as when new microblocks come in.

src/core/mempool.rs
```diff
@@ -135,6 +139,30 @@ pub struct MemPoolTxMetadata {
     pub accept_time: u64,
 }
 
+#[derive(Debug, Clone)]
+pub struct MemPoolWalkSettings {
+    /// Minimum transaction fee that will be considered
+    pub min_tx_fee: u64,
```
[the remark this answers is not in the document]
jcnelson: I think it's obvious from the codebase that it's microSTX.  Without exception, all of the variables we use to represent a quantity of STX are in units of microSTX.

testnet/stacks-node/src/neon_node.rs
```diff
@@ -1586,11 +1618,28 @@ impl InitializedNeonNode {
...
+                    "Consider in-flight block {} on Stacks tip {}/{} in {} with {} txs",
+                    &prev_block.anchored_block.block_hash(),
                     &prev_block.parent_consensus_hash,
                     &prev_block.anchored_block.header.parent_block,
-                    &prev_block.my_burn_hash
+                    &prev_block.my_burn_hash,
+                    &prev_block.anchored_block.txs.len()
                 );
+                if prev_block.anchored_block.txs.len() == 1 {
+                    if last_mined_blocks.len() == 1 {
```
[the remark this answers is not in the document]
jcnelson: These two checks are sufficient.  If there is only one transaction in this block, and there has only been one block mined, then it means that the miner only ran tenure once and it resulted in an empty block (i.e. because it didn't have time to actually consider any transactions).

src/core/mempool.rs
```diff
@@ -386,52 +434,127 @@ impl MemPoolDB {
...
                     }
-                    Some(nonce) => nonce,
-                };
+                }
+            }
+            offset += 1;
+
+            if total_added == 0 {
+                debug!("No origin address had a minable transaction");
+                break;
```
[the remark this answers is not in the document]
jcnelson: Good catch! No, we don't want that.

src/core/mempool.rs
```diff
@@ -386,52 +434,127 @@ impl MemPoolDB {
...
+
+                    if !todo(clarity_tx, vec![tx])? {
+                        test_debug!("Mempool early return from iteration");
+                        break;
                     }
-                    Some(nonce) => nonce,
-                };
+                }
+            }
+            offset += 1;
```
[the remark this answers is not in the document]
jcnelson: This method only does one pass through the origins. The callers in `miner.rs` repeatedly call `iterate_candidates()` with the same `epoch_tx`.  This way, if origin _O_ gets considered more than once, its expected nonces _n_ and _n+1_ will automatically be deduced by way of the inner `get_account()` call.

src/core/mempool.rs
```diff
@@ -1164,63 +1220,97 @@ mod tests {
...
+                    &origin,
+                    origin_nonce,
+                    &sponsor,
+                    sponsor_nonce,
+                    None,
+                )
+                .unwrap();
 
-            assert!(MemPoolDB::db_has_tx(&mempool_tx, &txid).unwrap());
+                assert!(MemPoolDB::db_has_tx(&mempool_tx, &txid).unwrap());
```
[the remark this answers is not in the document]
jcnelson: Yes.  Eventually, we'll have a separate, _global_ minimum relay fee that will determine whether or not the transaction will be admitted to the mempool.  But it's important that this value is the same for all nodes; otherwise, we run the risk of transactions not getting well-propagated due to next-hop nodes having higher min-relay fees than the miners.

src/core/mempool.rs
```diff
@@ -386,52 +434,127 @@ impl MemPoolDB {
     ///
     /// Consider transactions across all forks where the transactions have
     /// height >= max(0, tip_height - MEMPOOL_MAX_TRANSACTION_AGE) and height <= tip_height.
-    pub fn iterate_candidates<F, E>(&self, tip_height: u64, mut todo: F) -> Result<(), E>
+    pub fn iterate_candidates<F, E, C>(
```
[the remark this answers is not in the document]
jcnelson: Fixed

src/core/mempool.rs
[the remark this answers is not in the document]
jcnelson: Fixed

src/core/mempool.rs
```diff
@@ -1329,11 +1466,11 @@ mod tests {
         // check that the number of transactions at b_2 and b_4 starts at 1 each
```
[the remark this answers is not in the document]
jcnelson: Fixed

src/core/mempool.rs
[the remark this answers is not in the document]
gregorycoppola: I think it makes sense to use context to save on comments, but I would agree with @pavitthrap in the sense that I didn't realize this was assumed until this discussion.

jcnelson:
The atlas integration test repeatedly passes locally for me.  Can I just merge this without that specific test passing?

kantai:
> The atlas integration test repeatedly passes locally for me. Can I just merge this without that specific test passing?

The atlas integration test does appear to just be flaky in GH actions -- it's passed at different times for this PR. The sampled_genesis tests, however, don't appear to just be flaky-- they have not all passed yet for this PR.

src/core/mempool.rs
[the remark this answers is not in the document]
pavitthrap: Looking at the code, I see that iterate_candidates is called by `build_anchored_block` and `mine_next_microblock`. 

In `build_anchored_block`, I see that `epoch_tx` is initialized, `iterate_candidates` is called once at the top level of the function, and afterwards, `epoch_finish` is called. I'm confused how `iterate_candidates` is being called repeatedly?

PR #2838, opened 2021-09-09
Chore: Integration tests cleanup

kantai:
Opening this PR to clean up some of the integration testing in `develop`.

The first cleaned up test is `fix_2771`, which as previously written could lead to an infinite loop if an event got dropped by the test (leading to 6 hour GH actions for failures).

I suspect one of the microblock tests also requires a fix, as the merge from `master -> develop` caused an issue there.


kantai:
Requesting reviews from @lgalabru and @jcnelson to make sure that these changes don't impact the testing logic for the two changed tests.

lgalabru (approved):
LGTM, thank you @kantai!

jcnelson (approved):

PR #2840, opened 2021-09-10
Feature: Initial CostEstimator and FeeEstimator implementations

kantai:
This PR implements initial (fairly pessimistic) cost and fee estimators, per the described plan in #2825.

The CostEstimator and FeeEstimator traits operate on newly validated StacksBlocks (specifically, the block receipts). The naive pessimistic cost estimator just takes an average of ten samples for a given transaction payload, only allowing samples that increase the cost estimate. The fee estimator uses an exponential decay windowing function to track the 5th, 50th, and 95th percentile fee rate in each block.

`ExecutionCost` and tx length are translated into a scalar by computing a "proportional dot product", that is:

```
sum (resolution * cost_d / limit_d  for each dimension d)
```

Where resolution is `1000`. This means that the single dimension "cost" for a transaction that used the whole block would be `6000` (`1000` for each dimension of `ExecutionCost` plus `1000` for the proportion of the block byte length consumed).

Unit tests cover the new structs, but to get a sense for how well these costs perform, we'd need to analyze logged error rates as a node boots with this code from genesis.

### DIFF
--- a/src/cost_estimates/fee_scalar.rs
+++ b/src/cost_estimates/fee_scalar.rs
@@ -2,6 +2,7 @@ use std::cmp;
 use std::convert::TryFrom;
 use std::{iter::FromIterator, path::Path};
 
+use rusqlite::Transaction as SqlTransaction;
 use rusqlite::{
     types::{FromSql, FromSqlError},
     Connection, Error as SqliteError, OptionalExtension, ToSql,
@@ -9,7 +10,9 @@ use rusqlite::{
 use serde_json::Value as JsonValue;
 
 use chainstate::stacks::TransactionPayload;
+use util::db::tx_begin_immediate_sqlite;
 use util::db::u64_to_sql;
+
 use vm::costs::ExecutionCost;
 
 use core::BLOCK_LIMIT_MAINNET;
@@ -27,7 +30,7 @@ CREATE TABLE scalar_fee_estimator (
     estimate_key NUMBER PRIMARY KEY,
     fast NUMBER NOT NULL,
     medium NUMBER NOT NULL,
-    slow NUMBER NOT NULL,
+    slow NUMBER NOT NULL
 )";
 
 /// This struct estimates fee rates by translating a transaction's `ExecutionCost`
@@ -52,8 +55,10 @@ impl<M: CostMetric> ScalarFeeRateEstimator<M> {
             .or_else(|e| {
                 if let SqliteError::SqliteFailure(ref internal, _) = e {
                     if let rusqlite::ErrorCode::CannotOpen = internal.code {
-                        let db = Connection::open(p)?;
-                        Self::instantiate_db(&db)?;
+                        let mut db = Connection::open(p)?;
+                        let tx = tx_begin_immediate_sqlite(&mut db)?;
+                        Self::instantiate_db(&tx)?;
+                        tx.commit()?;
                         Ok(db)
                     } else {
                         Err(e)
@@ -65,23 +70,48 @@ impl<M: CostMetric> ScalarFeeRateEstimator<M> {
         Ok(Self {
             db,
             metric,
-            decay_rate_fraction: (3, 4),
+            decay_rate_fraction: (1, 2),
         })
     }
 
-    fn instantiate_db(c: &Connection) -> Result<(), SqliteError> {
+    fn instantiate_db(c: &SqlTransaction) -> Result<(), SqliteError> {
         c.execute(CREATE_TABLE, rusqlite::NO_PARAMS)?;
         Ok(())
     }
 
-    fn update_estimate(&self, new_measure: FeeRateEstimate) {
+    fn update_estimate(&mut self, new_measure: FeeRateEstimate) {
         let next_estimate = match self.get_rate_estimates() {
             Ok(old_estimate) => {
-                let prior_component =
-                    (old_estimate / self.decay_rate_fraction.1) * self.decay_rate_fraction.0;
-                let next_component = (new_measure / self.decay_rate_fraction.1)
+                // compute the exponential windowing:
+                // estimate = (a/b * old_estimate) + ((1 - a/b) * new_estimate)
+                //
+                // in order to avoid overflows, we do the division by `b` first, but to preserve
+                //  some integer precision, include the remainder by taking the remainder and doing
+                //  multiplication first (this also cannot overflow, because this calculation is at most
+                //  a * b, which are both u16)
+                let prior_component = (old_estimate.clone() / self.decay_rate_fraction.1)
+                    * self.decay_rate_fraction.0;
+                let prior_remainder = ((old_estimate % self.decay_rate_fraction.1)
+                    * self.decay_rate_fraction.0)
+                    / self.decay_rate_fraction.1;
+
+                let next_component = (new_measure.clone() / self.decay_rate_fraction.1)
                     * (self.decay_rate_fraction.1 - self.decay_rate_fraction.0);
-                prior_component + next_component
+                let next_remainder = (new_measure.clone() % self.decay_rate_fraction.1)
+                    * (self.decay_rate_fraction.1 - self.decay_rate_fraction.0)
+                    / self.decay_rate_fraction.1;
+
+                let mut next_computed =
+                    prior_component + prior_remainder + next_component + next_remainder;
+
+                // because of integer math, we can end up with some edge effects
+                // when the estimate is < decay_rate_fraction.1, so just saturate
+                // on the low end at a rate of "1"
+                next_computed.fast = cmp::max(next_computed.fast, 1);
+                next_computed.medium = cmp::max(next_computed.medium, 1);
+                next_computed.slow = cmp::max(next_computed.slow, 1);
+
+                next_computed
             }
             Err(EstimatorError::NoEstimateAvailable) => new_measure.clone(),
             Err(e) => {
@@ -90,19 +120,31 @@ impl<M: CostMetric> ScalarFeeRateEstimator<M> {
             }
         };
 
+        debug!("Updating fee rate estimate for new block";
+               "new_measure_fast" => new_measure.fast,
+               "new_measure_medium" => new_measure.medium,
+               "new_measure_slow" => new_measure.slow,
+               "new_estimate_fast" => next_estimate.fast,
+               "new_estimate_medium" => next_estimate.medium,
+               "new_estimate_slow" => next_estimate.slow);
+
         let sql = "INSERT OR REPLACE INTO scalar_fee_estimator
                      (estimate_key, fast, medium, slow) VALUES (?, ?, ?, ?)";
-        self.db
-            .execute(
-                sql,
-                rusqlite::params![
-                    SINGLETON_ROW_ID,
-                    u64_to_sql(next_estimate.fast).unwrap_or(i64::MAX),
-                    u64_to_sql(next_estimate.medium).unwrap_or(i64::MAX),
-                    u64_to_sql(next_estimate.slow).unwrap_or(i64::MAX)
-                ],
-            )
-            .expect("SQLite failure");
+
+        let tx = tx_begin_immediate_sqlite(&mut self.db).expect("SQLite failure");
+
+        tx.execute(
+            sql,
+            rusqlite::params![
+                SINGLETON_ROW_ID,
+                u64_to_sql(next_estimate.fast).unwrap_or(i64::MAX),
+                u64_to_sql(next_estimate.medium).unwrap_or(i64::MAX),
+                u64_to_sql(next_estimate.slow).unwrap_or(i64::MAX)
+            ],
+        )
+        .expect("SQLite failure");
+
+        tx.commit().expect("SQLite failure");
     }
 }
 
@@ -136,7 +178,7 @@ impl<M: CostMetric> FeeEstimator for ScalarFeeRateEstimator<M> {
                             .from_cost_and_len(&tx_receipt.execution_cost, tx_size)
                     }
                 };
-                let fee_rate = fee / cmp::max(1, scalar_cost);
+                let fee_rate = cmp::max(1, fee / cmp::max(1, scalar_cost));
                 Some(fee_rate)
             })
             .collect();
--- a/src/cost_estimates/fee_scalar.rs
+++ b/src/cost_estimates/fee_scalar.rs
@@ -90,9 +90,21 @@ impl<M: CostMetric> ScalarFeeRateEstimator<M> {
                 // because of integer math, we can end up with some edge effects
                 // when the estimate is < decay_rate_fraction.1, so just saturate
                 // on the low end at a rate of "1"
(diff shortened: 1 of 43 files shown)


kantai:
Data from running this estimator on mainnet...


This is the CDF of the "error" for the estimation: basically, the percentage of the actual cost value that the estimator was off by. 

 Error | CDF
-- | --
1.00% | 33.49%
5.00% | 39.04%
10.00% | 45.60%
15.00% | 67.27%
20.00% | 72.65%
30.00% | 73.64%
40.00% | 74.41%
50.00% | 75.38%
2x | 77.34%
3x | 78.93%
4x | 80.19%
5x | 80.50%
10x | 89.42%
20x | 95.72%
50x | 98.54%

So -- for ~40% of transactions, the estimator is within 5 percent of the actual cost. For 67% of transactions, the estimator is within 15% of the actual cost. However, for ~24% of transactions, the estimator estimates a cost that is over 2x the actual cost of executing the transaction (and up to much more than that in fact for a few transactions).

Looking at the methods that get over estimated:

```
cc:boom-nfts.mint-series
cc:boomboxes-cycle-12.delegate-stx
cc:boomboxes-cycle-14.delegate-stx
cc:boomboxes-cycle-16.delegate-stx
cc:boomboxes-cycle-6.delegate-stx
cc:fpwr-v03-depot.add-rewards
cc:free-punks-v0.mint
cc:free-punks-v0.transfer
cc:homely-lime-bass.sell-mia
cc:miamicoin-core-v1.claim-mining-reward
cc:miamicoin-core-v1.mine-many
cc:miamicoin-core-v1.stack-tokens
cc:oracle-v1.add-prices
cc:pool-tool-v1.delegate-stack-stx-and-commit
cc:power-ball.bet
cc:power-ball.step_draw
cc:powerball.bet
cc:powerball.step_draw
cc:pox.stack-stx
cc:send-many-memo.send-many
cc:send-many.send-many
cc:stacks-art-market.buy-item
cc:stacks-art-market.list-item
cc:stacks-art-market.unlist-item
cc:stacks-punks-bulk-transfer-v3.bulk-transfer
cc:stacks-punks-market.buy-punk
cc:stacks-punks-market.list-punk
cc:stacks-punks-market.unlist-punk
cc:stacks-punks-v3.claim-v1-punks
cc:stacks-punks-v3.mint
cc:stacks-punks-v3.transfer
cc:stacks-punks.mint
cc:stacks-punks.transfer
cc:stackswap-one-step-mint-v0d.create-pair-new-poxl-token-with-stx
cc:stx-ft-swap-v1.submit-swap
cc:vvss.bet
cc:will.bet
cc:willend2.bet
cc:xverse-pool-v1.delegate-stack-stx
cc:xverse-pool-v2.delegate-stack-stx
```

What it appears these all have in common is that their runtime can be very high when given large inputs (i.e., if `send-many` is given a large list of transfers), and this pessimistic estimator treats everything as a constant, so these methods will always get estimated at their mostly costly possibility.


Looking at _how_ the estimator errors, we have:

Error type | %
-- | --
Over | 68.47%
Exact | 30.40%
Under | 1.13%


So the estimator is behaving pretty much as expected: it is accurate or over estimating in the vast majority of cases, and the estimate is good for about 2/3 transactions. But for a lot of transactions, it is extremely pessimistic, suggesting that a worthwhile extension for the estimator may be to factor in the argument length.

gregorycoppola (commented):
Thanks for this initiative. Just a few comments.

src/cost_estimates/pessimistic.rs
```diff
@@ -0,0 +1,304 @@
...
+
+        return false;
+    }
+
+    /// Return the integer mean of the sample, uses iterative
+    /// algorithm to avoid overflow. The iterative algorithm
+    /// does have some error around *underflows* on the update,
+    /// but only when the new value is close to the average relative
+    /// to the window size.
+    fn mean(&self) -> u64 {
```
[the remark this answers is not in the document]
kantai: I'm not positive it would overflow u64 current, but it's possible if the block limits changed, the resolution of the scalar changed, or the window size changed.

Computing as a float is certainly possible, though.

src/cost_estimates/pessimistic.rs
```diff
@@ -0,0 +1,304 @@
...
+                          "value" => field.select_key(actual_cost));
+                }
+            }
+        }
+
+        let sql_tx = tx_begin_immediate_sqlite(&mut self.db)?;
+        for field in CostField::ALL.iter() {
+            let key = PessimisticEstimator::get_estimate_key(tx, field);
+            let field_cost = field.select_key(actual_cost);
+            let mut current_sample = Samples::get_sqlite(&sql_tx, &key);
```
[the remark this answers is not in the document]
kantai: It doesn't appear to be particularly costly -- simple gets / inserts in SQLite are much faster than most other kinds of operations performed while processing blocks, and in spinning up nodes that use this feature branch, I haven't noticed any performance issues related to this path. If this did lead to a slow down in node spinup, though, I think it would be useful to benchmark.

src/cost_estimates/fee_scalar.rs
```diff
@@ -0,0 +1,207 @@
...
+    db: Connection,
+    /// how quickly does the current estimate decay
+    /// compared to the newly received block estimate
+    ///      new_estimate := (decay_rate) * old_estimate + (1 - decay_rate) * new_measure
+    decay_rate: f64,
+    metric: M,
+}
+
+impl<M: CostMetric> ScalarFeeRateEstimator<M> {
+    /// Open a pessimistic estimator at the given db path. Creates if not existent.
```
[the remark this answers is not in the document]
kantai: Added in b05b2332e

src/cost_estimates/pessimistic.rs
```diff
@@ -0,0 +1,304 @@
...
+use util::db::u64_to_sql;
+use vm::costs::ExecutionCost;
+
+use core::BLOCK_LIMIT_MAINNET;
+
+use crate::util::db::tx_begin_immediate_sqlite;
+
+use super::{CostEstimator, EstimatorError};
+
+pub struct PessimisticEstimator {
```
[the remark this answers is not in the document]
kantai: Added in b05b2332e

src/cost_estimates/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+    fn add(self, rhs: Self) -> FeeRateEstimate {
+        FeeRateEstimate {
+            fast: self.fast.saturating_add(rhs.fast),
+            medium: self.medium.saturating_add(rhs.medium),
+            slow: self.slow.saturating_add(rhs.slow),
+        }
+    }
+}
+
+pub trait CostEstimator {
```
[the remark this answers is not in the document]
kantai: Added in b05b2332e

gregorycoppola:
> Data from running this estimator on mainnet...
> 
 
This is a very interesting analysis.

I think we should push on this analysis to get a metric, or small set of metrics, that each condense what we want to get out of these estimators into a number. Then, we can compare different estimators using the metrics.

For example, are we *looking for* a certain ratio of under-estimates to over-estimates? What ratio?
Are we *looking for* a certain behavior with respect to over-estimates vs under-estimates? What behavior?

One way to do this is to penalize under-estimates at a rate of F (F > 0) to 1. But there would be others as well.

kantai:
In almost all cases, over-estimates are preferred to under-estimates -- the loss function to optimize may be hard to specify. I worry about over-indexing on this, though. I think it is far more important to get _something_ in place, and we can optimize it further later. Looking at this:

```
What it appears these all have in common is that their runtime can be very high when given large inputs (i.e., if send-many is given a large list of transfers), and this pessimistic estimator treats everything as a constant, so these methods will always get estimated at their mostly costly possibility.
```

I believe the next step (after getting this PR merged) for improving the estimator is to incorporate argument size into the estimator somehow.

gregorycoppola:
> I believe the next step (after getting this PR merged) for improving the estimator is to incorporate argument size into the estimator somehow.

Well, the metric is kind of just a decision. It's not like implementation work.

We are going to want some way to compare the argument size-aware model to the current model?

kantai:
I think for the first version of this (and perhaps a good default going forward) is to optimize for minimizing under-estimates. For now, the `PessimisticEstimator` wins on that metric.

gregorycoppola (approved):

pavitthrap (approved):
LGTM! Thanks for all the code comments, they were super helpful :)

src/cost_estimates/fee_scalar.rs
```diff
@@ -0,0 +1,207 @@
+use std::cmp;
```
jcnelson, line 1: Minor Q on code organization -- should `src/cost_estimates` be moved under `src/chainstate/stacks`, since it pertains mainly to processing Stacks chainstate?

src/cost_estimates/fee_scalar.rs
```diff
@@ -0,0 +1,207 @@
...
+               "new_estimate_slow" => next_estimate.slow);
+
+        let sql = "INSERT OR REPLACE INTO scalar_fee_estimator
+                     (estimate_key, fast, medium, slow) VALUES (?, ?, ?, ?)";
+
+        let tx = tx_begin_immediate_sqlite(&mut self.db).expect("SQLite failure");
+
+        tx.execute(
+            sql,
+            rusqlite::params![
```
jcnelson, line 121: Ooooh, new macro for `&[&dyn ToSql]`?

src/cost_estimates/fee_scalar.rs
```diff
@@ -0,0 +1,207 @@
...
+                        Err(e)
+                    }
+                } else {
+                    Err(e)
+                }
+            })?;
+        Ok(Self {
+            db,
+            metric,
+            decay_rate: 0.5_f64,
```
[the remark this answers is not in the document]
kantai: Yeah, we could at some point.

src/cost_estimates/fee_scalar.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        let sql = "INSERT OR REPLACE INTO scalar_fee_estimator
+                     (estimate_key, fast, medium, slow) VALUES (?, ?, ?, ?)";
+
+        let tx = tx_begin_immediate_sqlite(&mut self.db).expect("SQLite failure");
+
+        tx.execute(
+            sql,
+            rusqlite::params![
+                SINGLETON_ROW_ID,
```
jcnelson, line 122: So, it looks like this database table `scalar_fee_estimator` only ever has one row?

src/cost_estimates/fee_scalar.rs
```diff
@@ -0,0 +1,207 @@
...
+                    TransactionPayload::PoisonMicroblock(_, _)
+                    | TransactionPayload::ContractCall(_)
+                    | TransactionPayload::SmartContract(_) => {
+                        // These transaction payload types all "work" the same: they have associated ExecutionCosts
+                        // and contibute to the block length limit with their tx_len
+                        self.metric
+                            .from_cost_and_len(&tx_receipt.execution_cost, tx_size)
+                    }
+                };
+                let fee_rate = cmp::max(1, fee / cmp::max(1, scalar_cost));
```
jcnelson, line 164: Just want to confirm that `scalar_cost` can never be zero here, right?

jcnelson: Also, because fees can be low (i.e. in small numbers of uSTX), should `fee_rate` be a float here?  We're not worried about any loss of precision that might arise?

src/cost_estimates/fee_scalar.rs
kantai: Old macro, though we don't use it much, we should, because it's a lot nicer than peppering &dyn ToSqls

src/cost_estimates/fee_scalar.rs
kantai: Yep

src/cost_estimates/fee_scalar.rs
kantai: Even if it could, this wouldn't cause an overflow, because of the `cmp::max`'s.

kantai: Fee rate is microstacks per (1/60k)th of the block limit. So a typical contract-call would be about 300 of the execution cost units, meaning that an integer fee rate would have a precision of 300 microstacks (0.0003 stacks). At worst (e.g., a block-filling tx), the precision would be 0.06 stacks. A float could provide more precision, though I don't think it would add much. However, it's easy enough to change to a float, so I'll just do that.

src/cost_estimates/pessimistic.rs
```diff
@@ -0,0 +1,297 @@
...
+
+impl Samples {
+    fn to_json(&self) -> JsonValue {
+        JsonValue::from(self.items.as_slice())
+    }
+
+    /// Add a new sample to this struct. The pessimistic sampler only adds to the sample set
+    ///  if the sample set is less than SAMPLE_SIZE or the new sample is greater than the current min.
+    /// Boolean return indicates whether or not the sample was included.
+    fn update_with(&mut self, sample: u64) -> bool {
```
jcnelson, line 102: Just to be clear - there's no way for the sample values to get smaller?  So, if we deploy bigger and bigger smart contracts, for example, the cost estimate for deploying them will continue to get bigger and bigger (given the `mean()` calculation below)?

kantai: Yeah, that's true for smart contract deployment in this naively pessimistic estimator --- smart contract deployment will always be estimated as pretty costly, and perhaps a linear model that takes into account contract size would work better for deploy type transactions. However contract-call's are estimated independently, so this isn't the case for those kinds of transactions.

testnet/stacks-node/src/config.rs
```diff
@@ -1010,6 +1023,155 @@ pub struct NodeConfig {
...
+            cost_estimator: Some(cost_estimator),
+            fee_estimator: Some(fee_estimator),
+            cost_metric: Some(cost_metric),
+            log_error,
+        }
+    }
+}
+
+impl FeeEstimationConfig {
+    pub fn make_pessimistic_cost_estimator(
```
jcnelson, line 1147: Do we want to maybe put the cost and fee estimator state into the same database?  Or is there some sort of pathological locking behavior we're trying to avoid here?  Seems like the state isn't that big in both these databases.

jcnelson (approved):
This LGTM; just had a couple of minor comments. Is there data out there for mainnet that shows how effective the pessimistic cost estimator is over time?  i.e. does it get progressively more pessimistic once the sample vector is full?

src/cost_estimates/fee_scalar.rs
jcnelson: Fair enough

testnet/stacks-node/src/config.rs
kantai: Just trying to avoid dependence between the two if it isn't necessary -- there aren't any queries that use both databases.

PR #2849, opened 2021-09-17
data: Python scripts to extract CostEstimator dataset, python implementation of PessimisticEstimator

gregorycoppola:
We might not want this to actually go in this repo, but until we decide where to put it...


jcnelson:
I'm fine with keeping this in this repo, but could you put the scripts into `tools/cost_estimator/analysis` instead?

PR #2851, opened 2021-09-17
WIP: add `multi-stack-stx` function to pox-2

hstove:
## Description

I'm working on a set of smart contracts that essentially allows a large pool of STX to be 'rented out' to multiple various other parties, so that the other party can use some amount of the pool's STX for a stacking cycle.

These funds are held by a single smart contract, so I'm trying to find a solution to allow a single principal to stack to multiple different reward addresses. After learning a lot about the underlying DB and how it interacts with the PoX contract to perform locking, I believe this solution works well in tandem with existing functionality.

Because the snapshot DB does not allow for multiple different locking periods, this new function only allows you to:

- Perform this "multi-stack" once
- Only for one cycle

The idea is that this function will work well for an initial stack, but would need further changes to support "continuous stacking" in this use case. I am more than willing to follow along with future implementations for things like `stack-increase` to support this "multi stack" use case, and to work on that implementation.

This is more of a proposal than anything - I'm hopeful that we can figure out a solution for this use case, and it doesn't have to be this solution.

I have other tests in a different repo (using clarity-cli), and I can work on translating them all into these rust tests.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

Yes, which is why it's intended for the `next` branch

Eventually, pox-2 contract documentation

### DIFF
--- a/src/chainstate/stacks/boot/contract_tests.rs
+++ b/src/chainstate/stacks/boot/contract_tests.rs
@@ -730,6 +730,7 @@ fn pox_2_lock_extend_units() {
         );
 
         let burn_height = env.eval_raw("burn-block-height").unwrap().0;
+        eprintln!("{}", POX_ADDRS[1].clone());
         assert_eq!(
             env.execute_transaction(
                 (&delegator).into(),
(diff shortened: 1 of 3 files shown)


jcnelson:
We generally require SIPs for breaking changes like this. Can you send over a draft copy that explains not only what this does, but also why it cannot be done as a non-breaking change?

hstove:
Do we have SIPs for the other pox-2 functions? Maybe me calling it a "breaking change" is ambiguous, as the whole pox-2 contract is a breaking change

hstove:
For more context on the use case here:

- This would enable something like [stacking futures](https://github.com/jcnelson/stx-future) that can be sold on a cycle-by-cycle basis (to multiple buyers)
- Some of this is discussed in the SIP issue for [multiple commitments per principal](https://github.com/stacksgov/sips/issues/8)

I am 100% not tied to this exact API. I believe this is a more efficient (and simpler to handle on the node) implementation than something like "call stack-increase a bunch of times with different pox addrs"

hstove:
I will close this and open an issue to start a more broad discussion. Apologies if it felt like I was trying to force this in - the intention was just to show a proof-of-concept, but I think I jumped the gun before describing the situation more broadly. Thanks for taking a look!

PR #2856, opened 2021-09-22
Feat/mempool pusher improvements

wileyj:
## Description

To help solve some of the recent mempool "freshness", we deployed this script as-is and noticed a few limitations that would have made it challenging to be effective for multiple followers. There was also an issue noted where the filename was incorrect and has been adjusted. 
One thing to note - if you use a large list of hosts, it will take a several minutes per run to process all of the neighbors so it' s probably best to keep the list small. 

`usage()` function has been updated to address this new functionality, i.e.:
`log "      $0 daemon PATH_TO_MEMPOOL LOCAL_PEER_HOST1:PORT,LOCAL_PEER_HOST2:PORT"`

1.  Fixed a typo in the `antientropy_pass` function where an incorrect filename was referenced:
`/tmp/last_mempool_scan.tx.$$` changed to : `/tmp/last_mempool_scan.txs.$$`
2. Added new functionality to allow the script to be run against multiple hosts when provided as a comma separated list. ex: `host1:20443,host2:20443...`

## Type of Change
- [X] New feature
- [X] Bug fix
- [ ] Other


tools/mempool-pusher.sh
```diff
@@ -106,10 +106,16 @@ antientropy_pass() {
...
    local local_peer="$2"
    local last_sent="$3"
+   local peer_count=0
 
    # do this sequentially so we don't hold open the DB
-   get_mempool_txs "$mempool_path" "$last_sent" > "/tmp/last_mempool_scan.txs.$$"
-   antientropy_push "/tmp/last_mempool_scan.tx.$$" "$local_peer"
+   for peer in ${local_peer//,/ }; do
+         if [ $peer_count -eq 0 ]; then
+         get_mempool_txs "$mempool_path" "$last_sent" > "/tmp/last_mempool_scan.txs.$$"
```
jcnelson, line 114: Minor style nit -- please indent here

tools/mempool-pusher.sh
```diff
@@ -162,7 +168,10 @@ usage() {
    log "Usage: $0 MODE [ARGS...]"
    log "MODE can be the following:"
    log "      $0 daemon PATH_TO_MEMPOOL LOCAL_PEER_HOST:PORT"
+   log "      $0 daemon PATH_TO_MEMPOOL LOCAL_PEER_HOST1:PORT,LOCAL_PEER_HOST2:PORT"
```
jcnelson, line 171: For the usage section, can it have an ellipsis at the end, so it reads `LOCAL_PEER_HOST1:PORT,LOCAL_PEER_HOST2:PORT,...` to indicate that a list of host:ports will be taken?

jcnelson (approved):
LGTM; just a couple of minor changes.

tools/mempool-pusher.sh
wileyj: Indented - good catch!

tools/mempool-pusher.sh
wileyj: Added elipsis, agreed that it makes it more clear more than 2 hosts can be in the list

kantai (approved):

Issue #2834, opened 2021-09-07
Fee estimation: Update block assembly to use estimator and fee rate ordering


kantai:
Per #2825, miner block assembly needs to be updated such that iteration is ordered by estimated fee rate. This work would also include _updating_ the estimator with the output of tentatively executing transactions: that is, if a particular contract-call has never been seen before, it should be tentatively executed to obtain a cost estimate, and rolled back if the fee rate turned out to be too low.

kantai:
In order to make the mempool iteration a little easier to reason about, in implementing this, I am proposing to restructure the mempool iterator to do the following:

1. Add a new method `MempoolDB::prepare(&self, &mut ClarityConnection)` that will:
    1. Read the current account nonce for every origin (and sponsor) in the mempool
    2. Estimate the ExecutionCost for each transaction in the mempool
    3. Write this data to a side table in the mempool db (which will get wiped before the write)
2. `MempoolDB::iterate_candidates` will, instead of the current `find_origin_addresses_by_descending_fees` call, perform a SQLite query like:
    1. `SELECT tx FROM mempool as m INNER JOIN sidetable as s ON m.txid = s.txid WHERE m.origin_nonce = s.origin_nonce + 1 AND m.sponsor_nonce = s.sponsor_nonce + 1 AND s.scalar_cost IS NOT NULL ORDER BY CAST(m.tx_fee as REAL) / CAST(s.scalar_cost as REAL) DESC LIMIT 1`
    2. After applying a transaction, update the nonces stored in the side table.
    3. With some probability, instead of the above query to get the next TX, explore a `s.scalar_cost IS NULL` transaction, or if there are only such transactions.

The advantage of this approach is that it makes it _much_ more clear how the mempool is being ordered/considered. The disadvantage is that it front-loads a lot of MARF reads (the account nonce reads) -- would be happy to consider alternatives, but the current approach has a _lot_ of downsides: iterating by selecting the origin adress can have many unintended consequences for sorting, for example, clumping origin transactions together, or rewarding origin addresses that have one high-fee transaction (possibly already confirmed) by confirming their other low-fee transactions before other accounts.

jcnelson:
Some feedback

[...]

PR #2859, opened 2021-09-29
Draft: Use fee rate estimation in block assembly

kantai:
This **draft** PR alters the miner block assembly logic to use configured cost estimation and cost metrics to produce fee rate estimations for candidate transactions in the mempool. This fee rate estimation is used as the sort order for transaction consideration.

This PR also implements the expected nonce calculation proposed in #2834.

There isn't a new end to end test case yet for the block assembly -- I'm still trying to understand the best way to test the end to end changes here (many of the existing e2e tests cover these code paths, but they don't specifically test for the desired behavior here).

### DIFF
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -1496,7 +1496,7 @@ impl StacksBlockBuilder {
         builder.try_mine_tx(&mut epoch_tx, coinbase_tx)?;
 
         mempool.reset_last_known_nonces()?;
-        mempool.estimate_tx_rates(estimator, metric, 100);
+        mempool.estimate_tx_rates(estimator, metric, 100)?;
 
         let mut considered = HashSet::new(); // txids of all transactions we looked at
         let mut mined_origin_nonces: HashMap<StacksAddress, u64> = HashMap::new(); // map addrs of mined transaction origins to the nonces we used
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -448,7 +448,7 @@ impl<'a> StacksMicroblockBuilder<'a> {
         return self.make_next_microblock(txs_included, miner_key);
     }
 
-    pub fn mine_next_microblock<CE: CostEstimator, CM: CostMetric>(
+    pub fn mine_next_microblock<CE: CostEstimator + ?Sized, CM: CostMetric + ?Sized>(
         &mut self,
         mem_pool: &mut MemPoolDB,
         miner_key: &Secp256k1PrivateKey,
@@ -1444,7 +1444,7 @@ impl StacksBlockBuilder {
 
     /// Given access to the mempool, mine an anchored block with no more than the given execution cost.
     ///   returns the assembled block, and the consumed execution budget.
-    pub fn build_anchored_block<CE: CostEstimator, CM: CostMetric>(
+    pub fn build_anchored_block<CE: CostEstimator + ?Sized, CM: CostMetric + ?Sized>(
         chainstate_handle: &StacksChainState, // not directly used; used as a handle to open other chainstates
         burn_dbconn: &SortitionDBConn,
         mempool: &mut MemPoolDB,
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -475,7 +475,7 @@ impl<'a> StacksMicroblockBuilder<'a> {
         let deadline = get_epoch_time_ms() + (self.settings.max_miner_time_ms as u128);
 
         mem_pool.reset_last_known_nonces()?;
-        mem_pool.estimate_tx_rates(estimator, metric, 100);
+        mem_pool.estimate_tx_rates(estimator, metric, 100)?;
 
         debug!(
             "Microblock transaction selection begins (child of {}), bytes so far: {}",
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -20,6 +20,8 @@ use std::convert::From;
 use std::fs;
 use std::mem;
 
+use crate::cost_estimates::metrics::CostMetric;
+use crate::cost_estimates::CostEstimator;
 use crate::types::StacksPublicKeyBuffer;
 use burnchains::PrivateKey;
 use burnchains::PublicKey;
@@ -322,21 +324,22 @@ impl<'a> StacksMicroblockBuilder<'a> {
     }
 
     /// Mine the next transaction into a microblock.
-    /// Returns true/false if the transaction was/was not mined into this microblock.
+    /// Returns Some(StacksTransactionReceipt) or None if the transaction was
+    /// or was not mined into this microblock.
     fn mine_next_transaction(
         clarity_tx: &mut ClarityTx<'a>,
         tx: StacksTransaction,
         tx_len: u64,
         considered: &mut HashSet<Txid>,
         bytes_so_far: u64,
-    ) -> Result<bool, Error> {
+    ) -> Result<Option<StacksTransactionReceipt>, Error> {
         if tx.anchor_mode != TransactionAnchorMode::OffChainOnly
             && tx.anchor_mode != TransactionAnchorMode::Any
         {
-            return Ok(false);
+            return Ok(None);
         }
         if considered.contains(&tx.txid()) {
-            return Ok(false);
+            return Ok(None);
         } else {
             considered.insert(tx.txid());
         }
@@ -345,11 +348,11 @@ impl<'a> StacksMicroblockBuilder<'a> {
                 "Adding microblock tx {} would exceed epoch data size",
                 &tx.txid()
             );
-            return Ok(false);
+            return Ok(None);
         }
         let quiet = !cfg!(test);
         match StacksChainState::process_transaction(clarity_tx, &tx, quiet) {
-            Ok(_) => return Ok(true),
+            Ok((_, receipt)) => return Ok(Some(receipt)),
             Err(e) => match e {
                 Error::CostOverflowError(cost_before, cost_after, total_budget) => {
                     // note: this path _does_ not perform the tx block budget % heuristic,
@@ -367,7 +370,7 @@ impl<'a> StacksMicroblockBuilder<'a> {
                 }
             },
         }
-        return Ok(false);
+        return Ok(None);
     }
 
     /// NOTE: this is only used in integration tests.
@@ -401,12 +404,12 @@ impl<'a> StacksMicroblockBuilder<'a> {
                 &mut considered,
                 bytes_so_far,
             ) {
-                Ok(true) => {
+                Ok(Some(_)) => {
                     bytes_so_far += tx_len;
                     num_txs += 1;
                     txs_included.push(tx);
                 }
-                Ok(false) => {
+                Ok(None) => {
                     continue;
                 }
                 Err(e) => {
@@ -445,10 +448,12 @@ impl<'a> StacksMicroblockBuilder<'a> {
         return self.make_next_microblock(txs_included, miner_key);
     }
 
-    pub fn mine_next_microblock(
+    pub fn mine_next_microblock<CE: CostEstimator, CM: CostMetric>(
         &mut self,
-        mem_pool: &MemPoolDB,
+        mem_pool: &mut MemPoolDB,
         miner_key: &Secp256k1PrivateKey,
+        estimator: &mut CE,
+        metric: &CM,
     ) -> Result<StacksMicroblock, Error> {
         let mut txs_included = vec![];
         let mempool_settings = self.settings.mempool_settings.clone();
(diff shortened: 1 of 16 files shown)


src/core/mempool.rs
```diff
@@ -103,6 +108,21 @@ pub enum MemPoolDropReason {
...
 
+pub struct ConsiderTransaction {
+    /// Transaction to consider in block assembly
+    pub tx: MemPoolTxInfo,
+    /// If `update_estimator` is set, the iteration should update the estimator
+    /// after considering the tx.
+    pub update_estimate: bool,
+}
+
+enum ConsiderTransactionResult {
```
jcnelson, line 119: Not a bug, but a style question -- is there a reason these are all caps?

src/core/mempool.rs
```diff
@@ -255,6 +281,28 @@ const MEMPOOL_INITIAL_SCHEMA: &'static [&'static str] = &[
...
+    );
+    "#,
+    r#"
+    ALTER TABLE mempool ADD COLUMN last_known_origin_nonce INTEGER;
+    "#,
+    r#"
+    ALTER TABLE mempool ADD COLUMN last_known_sponsor_nonce INTEGER;
+    "#,
+    "CREATE INDEX fee_by_txid ON fee_estimates(txid);",
+];
```
jcnelson, line 304: Can you also update the mempool DB schema so that these columns simply get created in the table if the table is being made for the first time?

jcnelson: Also, is the mempool DB versioned?  If not, we should add it, so we know what migrations to apply.

src/core/mempool.rs
```diff
@@ -425,6 +572,53 @@ impl MemPoolDB {
         query_row_columns(self.conn(), sql, args, "origin_address")
     }
 
+    /// Add estimated fee rates to the mempool rate table using
+    /// the supplied `CostMetric` and `CostEstimator`. Will update
+    /// at most `max_updates` entries in the database before returning.
+    ///
+    /// Returns `Ok(number_updated)` on success
+    pub fn estimate_tx_rates<CE: CostEstimator + ?Sized, CM: CostMetric + ?Sized>(
```
jcnelson, line 580: The mempool really can grow unbound.  Is there a strategy for prioritizing transactions for fee reconsideration beyond just doing so in the order in which they were inserted into the mempool (which is what's implicitly happening in the SQL query here)?

src/chainstate/stacks/miner.rs
```diff
@@ -469,6 +474,9 @@ impl<'a> StacksMicroblockBuilder<'a> {
         let mut num_selected = 0;
         let deadline = get_epoch_time_ms() + (self.settings.max_miner_time_ms as u128);
 
+        mem_pool.reset_last_known_nonces()?;
```
jcnelson, line 477: Won't this obliterate the known nonces for the *entire* mempool?  Isn't that going to make microblock mining expensive, if we have to re-calculate nonces all the time?  Or, is that bound by the subsequent call to `estimate_tx_rates` with `max_updates = 100`?

src/chainstate/stacks/miner.rs
```diff
@@ -1464,6 +1495,9 @@ impl StacksBlockBuilder {
         let mut epoch_tx = builder.epoch_begin(&mut chainstate, burn_dbconn)?;
         builder.try_mine_tx(&mut epoch_tx, coinbase_tx)?;
 
+        mempool.reset_last_known_nonces()?;
+        mempool.estimate_tx_rates(estimator, metric, 100)?;
```
jcnelson, line 1499: Because there can be far more than 100 transactions available for consideration, what's our strategy for helping miners find the *maximum-fee-rate* transactions, if we're only going to estimate fees for 100 of them per attempt at building an anchored block?

jcnelson (changes requested):
So, I think this PR is on the right track, but I think there's a couple of architectural changes that would make this PR a lot stronger and less brittle.  They are:

1. Make maintenance of the `fee_estimates` table _entirely_ the responsibility of the miner.  While the table needs to be in the same database as the mempool in order for the `LEFT OUTER JOIN`s to work, I think the miner should be explicitly managing that table's lifetime -- including creating it, curating it, and deleting it.

The only compelling reason I can see so far to keep the fee estimation data life cycle in the mempool is if we also intend to use the fee estimations to make decisions about whether or not to accept a transaction pushed to us (which we very well may want to do in the near future).  But if this is a world we're trying to go for, then I think managing fee estimates should be handled by the chains coordinator or relayer instead of the miner (since transaction relaying decisions need to be made by followers as well as miners).

2. Separate fee estimation and nonce (re-)calculation from building blocks.  The size of the mempool can grow unbound, and thus so can the amount of work a mining node needs to do to estimate fees.  Having a fixed-size "update 100 transactions per block attempt" I think makes it hard for miners to do what they really want: prioritize the globally highest-fee-rate transactions for mining.  To this end, I wonder if instead there should be a background process in the chains coordinator that asynchronously updates nonces and fees, so that when the miner wakes up, there's a better chance that the fee estimates for the highest value transactions are known already.  If we intend to use fee estimates to make relay decisions, then updating fee estimates outside of the miner is going to be required.

What do you think?

src/core/mempool.rs
kantai: Defunct enum style memory on the part of the author. Will change to the rust-y standard.

src/core/mempool.rs
kantai: Nope, there isn't, we could instead sort by random (or by raw fee, though that has the potential to skew things).

src/chainstate/stacks/miner.rs
kantai: Yes, it will obliterate the nonces for the entire mempool. However, it doesn't make mining any more expensive (in terms of nonce lookups) than the current implementation -- nonce calculation is done lazily. When a transaction is considered for inclusion, if the expected nonces aren't known, the origin and sponsor nonces are looked up, and the database is updated with those nonces, then the transaction is thrown back in for consideration, and the iterator polls for the next transaction to be considered (which could be the one it just saw, if it was valid given the expected nonce). That is:

1. Get the transaction from the mempool with the highest fee rate whose expected nonce is not known _or_ the transaction nonce matches the known expected nonce.
2. If the transaction's nonce was not known, look it up and update the table, go back to 1.
3. Otherwise, try to add the transaction to the block, then bump the expected nonce, go back to 1.

src/chainstate/stacks/miner.rs
kantai: The estimates aren't wiped out between blocks --- given the growth rate of the mempool, in practice, the miner would construct estimates for all the transactions in the mempool. Alternatively, the miner could do this iteratively -- if there's more than 100 transactions without estimates in the mempool, the miner could try to estimate more of the transactions after doing one pass of block assembly.

kantai:
> Take maintenance of the fee_estimates out of the mempool. While the table needs to be in the same database as the mempool in order for the LEFT OUTER JOINs to work, I think at a minimum, the miner should be explicitly managing that table's lifetime -- including creating it, curating it, and deleting it. However, if we're ultimately aiming to make it so that transactions in the mempool have fee estimations for the purposes of both finding high-value transactions and making transaction relay decisions, then I think that makes a compelling case for having the chains coordinator (or some background worker thread the chains coordinator controls) responsible for building up the fee estimations. Then, the estimates are available to the miner for mining nodes, and are also available to help the relayer decide if a transaction's fee rate is high enough to be added to the mempool (or more likely, replace a low-fee transaction in the mempool, since we will eventually want to bound the size of the mempool for the dual purposes of avoiding DDoS attacks and making mempool set difference calculations tractable).

The `fee_estimates` table is explicitly a secondary storage of estimate information for the mempool for use in mempool iteration -- it makes sense that the mempool module would have the code that managed and controlled mempool iteration (as is currently the case).

However, the "ground truth" of cost estimation and fee estimation does in fact happen in the chains coordinator -- that's where the estimators receive updates from the block receipts. The `fee_estimates` table must be located alongside the mempool in order to query it efficiently (and, I would argue, in a tractable order). As for the impact on the relayer, I'd prefer to cross that bridge when we get to it, because I'm not sure that using this information in the relayer is a good idea at all -- in the current framework (and probably any workable framework for estimation), it's entirely possible (and, indeed, expected) that nodes could have different estimates for the same transaction. Making transaction relay depend on that opens the door to pretty bad network behavior.

> Separate fee estimation and nonce (re-)calculation from building blocks. The size of the mempool can grow unbound, and thus so can the amount of work a mining node needs to do to estimate fees. Having a fixed-size "update 100 transactions per block attempt" I think makes it hard for miners to do what they really want: prioritize the globally highest-fee-rate transactions for mining. 

The design of the nonce tracking in this PR is lazy specifically to avoid unbounded work -- it should perform almost identically to the current system with regards to nonce lookups. The problem with trying to persist and maintain nonce data between block assembly invocations is that this state is _highly_ dependent on the Stacks fork (unlike the cost estimates, which are also dependent on the fork, but less critically so), and so the maintenance of a SQL table with that information would be at once laborious and error-prone. It may be worthwhile to separate the fee rate estimation of transactions in the mempool from block assembly, but I'm not convinced yet; while updating just a subset per block attempt is naive and insufficient to span the mempool, approaching like iteratively estimating more transactions in the mempool while assembling the block would be best implemented during block assembly. Finally, for transactions that are not yet even possible to estimate (i.e., the called contract is new and has not been invoked before), the only way to provide an initial measurement to the estimator is by trying to evaluate the transaction in a block-- again, a job best done in block assembly (see the `update_estimate` boolean in the mempool iterator `ConsiderTransaction` struct).


> To this end, I wonder if instead there should be a background process in the chains coordinator that asynchronously updates nonces and fees, so that when the miner wakes up, there's a better chance that the fee estimates for the highest value transactions are known already. 

Ultimately, my answer here is "maybe", but I'm not sufficiently convinced that I believe such a design should be a requirement for a first version of a miner that uses cost estimation.

jcnelson:
> However, the "ground truth" of cost estimation and fee estimation does in fact happen in the chains coordinator -- that's where the estimators receive updates from the block receipts. The fee_estimates table must be located alongside the mempool in order to query it efficiently

Yup, that all makes sense.  But then, should the miner be responsible for carrying out fee estimations at all?  It's unclear to me why the responsibility for curating the state of this table is spread between these two modules.

> As for the impact on the relayer, I'd prefer to cross that bridge when we get to it ... 
> Making transaction relay depend on that opens the door to pretty bad network behavior.

I'm in a position where I have to think about this now, because bounding the size of the mempool is an important part of my proposal for a mempool sync protocol (https://docs.google.com/document/d/1uHLUZEkzJJA8HtKfKVZmZIn5n7sY0X9j6gg90IyixYM/edit#).  I'd welcome approaches that don't require bounding the size of the mempool, but supposing for a second that the bound must be in place, we're going to need a way to determine which transactions get evicted if the mempool is full.  I'm not sure what criteria could be used *besides* the transaction's value to the miner; otherwise we run the risk of a DDoS attack whereby cheap transactions crowd out expensive transactions.  Other blockchains use the fee rate -- lower-value transactions get evicted by higher-value transactions.  But, other blockchains also have a well-defined fee estimation procedure -- Bitcoin has satoshis/byte, and Ethereum has wei/opcode.  We may never have such an estimator, but that's what motivated me to ask about the possibility of exposing fee data to the relayer.

Now, it's entirely possible that we always *relay* transactions if they pay more than some minimum absolute fee, but only *store* them if they furthermore clear a minimum fee *rate*.  The mempool synchronization logic would still work, but it would simply be a little bit less efficient -- Alice could send Bob some transactions that he ultimately won't store.  Do you think that would be alright?

> while updating just a subset [of nonces] per block attempt is naive and insufficient to span the mempool, approaching like iteratively estimating more transactions in the mempool while assembling the block would be best implemented during block assembly

The downside to this -- more generally, the downside of lazy estimation -- is that the probability that the highest-value transaction will be considered at all (let alone mined in the next block) is independent of the value of the other transactions.  The highest-value transaction is not guaranteed to be considered at all if the mempool grows unbound -- it's possible that new transactions arrive frequently enough that it's simply unlikely (less-than-50% chance) that the miner will "get around" to this high-value transaction within 256 blocks.  This leads to a very frustrating user experience and miner experience under load, precisely when the miners really ought to be searching for the highest-value transactions and users who pay more really ought to be prioritized.

Would it be unreasonable from a performance standpoint to simply analyze the transaction once it's stored to the mempool (or shortly afterwards)?

kantai:
> Yup, that all makes sense. But then, should the miner be responsible for carrying out fee estimations at all? It's unclear to me why the responsibility for curating the state of this table is spread between these two modules.

Because one module is _only_ concerned with constructing `ExecutionCost` estimates -- translating that into mempool iteration orders is squarely a job for the mining relevant modules. For example, miners may wish to do block fill using some kind of sat solver to optimally bin pack the various dimensions of the block limit, rather than using a scalar cost metric to approximate fee rate. Those kinds of decisions don't impact the cost estimator module at all -- the block assembly and mempool iteration logic are _consumers_ of the cost estimator API/trait, and that separation allows for far more flexibility in the future. Furthermore, I'd argue that updating the internal data storage of the mempool is definitively something that the mempool module should be responsible for, delegating that task to a cost estimator would break abstraction layers.

> Would it be unreasonable from a performance standpoint to simply analyze the transaction once it's stored to the mempool (or shortly afterwards)?

Nope, not unreasonable at all -- I can update the code to do that as well.

PR #2865, opened 2021-10-06
Profiling work

pavitthrap:
Here is the branch through which I collected profiling data for the stacks node. 

Although the collection of logs is controlled by setting an environment variable, I don't think we'll want to merge this into the normal node. We can leave it as a side branch. 

I will add documentation for how to enable & use this feature shortly (to this same PR).


pavitthrap:
Not going to merge this into the main code base - will leave this feature as a side branch.

Issue #2836, opened 2021-09-07
Fee estimation: New endpoint for obtaining estimates


kantai:
The fee estimation endpoint needs to be updated to provide a true fee rate estimate for transfer operations. Fee rate estimation differs from `ExecutionCost` estimation and the kind of estimation done during block assembly in that fee rate estimation looks at the observed fee rates in the last block produced (and potentially the estimated fee rates in the mempool) and produces an estimated fee rate for inclusion in the next block.

Once a fee rate estimate is obtained, communicating this to clients also requires some changes:

1. The current fee rate endpoint is really only suitable for _transfer_ transactions. This endpoint can be updated to be useful for transfer transactions, and provide a better estimate for other transactions than it currently does. However, this endpoint should probably be marked deprecated.
1. A new endpoint could accept a partial transaction and use that to return the estimated execution cost and estimated fee for the transaction.

Strategy two would require updating dependent libraries: `stacks.js` first, and then any integrators would need to update their scripts, integrations, etc. Wallets like the web wallet and desktop wallet would also need to be updated (though they may just need to update their stacks.js version).

markmhx:
What do we all think about providing three estimate values:

1. **Fast**: Expensive / high value for broadcasting a transaction that will _very_ likely (>95% chance) be picked up by the next anchor block / has large time premium for user
2. **Normal / medium** (for default in wallets, etc): Modest value for broadcasting a transaction that will likely (>60%) be picked up by the next anchor block / has moderate time premium for user
3. **Slow**: Low value for broadcasting a transaction that will likely be picked up within 3-5 anchor blocks / has little to no time premium for user. 

cc @jasperjansz

diwakergupta:
> What do we all think about providing three estimate values:
> 
> 1. **Fast**: Expensive / high value for broadcasting a transaction that will _very_ likely (>95% chance) be picked up by the next anchor block / has large time premium for user
> 2. **Normal** (for default in wallets, etc): Modest value for broadcasting a transaction that will likely (>60%) be picked up by the next anchor block / has moderate time premium for user
> 3. **Slow**: Low value for broadcasting a transaction that will likely be picked up within 3-5 anchor blocks / has little to no time premium for user.

This is an application level concern that IMO doesn't belong in an endpoint exposed by stacks-node -- something like this might be more appropriate for the API.

[...]

PR #2872, opened 2021-10-13
Fee estimation: RPC interface for estimating transaction costs and fees

kantai:
This PR implements #2836, adding a new endpoint for obtaining fee estimation information for transactions.

The endpoint is documented in the openapi specification, but I'll include that in this PR description:

### Description

`POST /v2/fees/transaction`

Get an estimated fee for that supplied transaction.  This estimates the execution cost of the transaction, the current fee rate of the network, and returns estimates for fee amounts. The posted data includes the following information:

* `transaction_payload` is a hex-encoded serialization of the TransactionPayload for the transaction.
* `estimated_len` is an optional argument that provides the endpoint 

If the node cannot provide an estimate for the transaction (e.g., if the node has never seen a contract-call for the given contract and function) or if estimation is not configured on this node, a 400 response is returned.

The 200 response contains the following data:

* `estimated_cost` - the estimated multi-dimensional cost of
          executing the Clarity VM on the provided transaction.
* `estimated_cost_scalar` - an integer that captures the total
          proportion of the block limit that a transaction would be
          estimated to consume. This value is multiplied by a fee rate
          to suggest the total fee amount to be paid by the node. This
          incorporates the estimated transaction length.
* `cost_scalar_change_by_byte` - a float value that indicates how
          much the `estimated_cost_scalar` value would increase for every
          additional byte in the final transaction.
* `estimated_fee_rates` - three estimated values for the current
          fee rates in the network
* `estimated_fees` - three estimated values for the total fee that
          the given transaction should pay. These values are the result of
          computing: `estimated_fee_rates` x `estimated_cost_scalar`

Note: If the final transaction's byte size is larger than supplied to `estimated_len`, then applications should increase this fee amount by:

          `fee_rate` x `cost_scalar_change_by_byte` x (`final_size` - `estimated_size`)

### Example request data

```json
{
  "estimated_len": 350,
  "transaction_payload": "021af942874ce525e87f21bbe8c121b12fac831d02f4086765742d696e666f0b7570646174652d696e666f00000000"
}
```

### Example response data

```json
{
  "cost_scalar_change_by_byte": 0.00476837158203125,
  "estimated_cost": {
    "read_count": 19,
    "read_length": 4814,
    "runtime": 7175000,
    "write_count": 2,
    "write_length": 1020
  },
  "estimated_cost_scalar": 14,
  "estimated_fee_rates": {
    "high": 10,
    "low": 1.2410714285714286,
    "middle": 8.958333333333332
  },
  "estimated_fees": {
    "high": 140,
    "low": 17,
    "middle": 125
  }
}
```

### DIFF
--- a/docs/rpc/openapi.yaml
+++ b/docs/rpc/openapi.yaml
@@ -315,12 +315,15 @@ paths:
 
         * `estimated_cost` - the estimated multi-dimensional cost of
           executing the Clarity VM on the provided transaction.
-        * `estimated_cost_scalar` - a unitless integer that captures
-          the total proportion of the block limit that a transaction
-          would be estimated to consume. In order to compute an
-          estimate of total fee amount for the transaction, this value
-          is multiplied by the estimated fee rate. This value
-          incorporates the estimated transaction length.
+        * `estimated_cost_scalar` - a unitless integer that the Stacks
+          node uses to compare how much of the block limit is consumed
+          by different transactions. This value incorporates the
+          estimated length of the transaction and the estimated
+          execution cost of the transaction. The range of this integer
+          may vary between different Stacks nodes. In order to compute
+          an estimate of total fee amount for the transaction, this
+          value is multiplied by the same Stacks node's estimated fee
+          rate.
         * `cost_scalar_change_by_byte` - a float value that indicates how
           much the `estimated_cost_scalar` value would increase for every
           additional byte in the final transaction.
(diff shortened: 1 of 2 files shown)


kyranjamie:
An array of objects might be better than an object for high/mid/low values, better enabling non-breaking changes later if we want 1) more increments, 2) some metadata beyond value itself

```json
{
  "estimated_fee_rates": [
    { fee_rate: 1 },
    { fee_rate: 2 },
    { fee_rate: 3 },
  ],
}
```

PR #2884, opened 2021-10-16
Feat/mempool sync

jcnelson:
This PR implements an anti-entropy protocol for querying transactions from other nodes' mempools, based on https://docs.google.com/document/d/1uHLUZEkzJJA8HtKfKVZmZIn5n7sY0X9j6gg90IyixYM/edit?usp=sharing.

It does the following:

* Adds a bloom filter and counting bloom filter implementation.  Importantly, the counting bloom filter is (a) disk-backed -- specifically, backed by a Sqlite blob -- and (b) can be converted into a bloom filter for issuing queries to other nodes.

* Adds `POST /v2/mempool/query`, which takes a `MemPoolSyncData` structure as input and streams back zero or more recently-arrived transactions from the mempool (as determined by block height).

* Adds a state machine to the main P2P state machine, which will cause a node to periodically ask a randomly-chosen outbound neighbor for any transactions in its mempool that it does *not* have locally.

The protocol in this PR is a simple set-reconciliation protocol.  A client node will periodically ask a peer node for the list of recent transactions that it does *not* have in its own mempool.  It does this by sending the remote peer a sketch of its mempool, which the remote peer uses to iterate through its set of recent transactions and stream back transactions it has that are *not* in the sketch.

A `MemPoolSyncData` structure encodes this sketch.  Specifically, it encodes either a list of 8-byte transaction ID prefixes ("tags"), or a bloom filter with a protocol-defined error rate and size.  The `/v2/mempool/query` handler will handle either variant; the client uses the variant that is cheaper for it to generate with an acceptable error rate.

The `MemPoolSyncData` represents a sketch calculated from node-specific transaction tags.  These are calculated as `siphash(node_seed, txid)`, which is an 8-byte value.  The `node_seed` is a node-specific random 32-byte value.  The reason for doing this is to ensure that a transaction has a different identifier for each node, so that even if a transaction is "masked" in one node's sketch via a false positive, it will almost certainly not be masked in other nodes' sketches.  It also ensures that no one can create a "malicious" transaction whose txid masks another txid in all nodes' mempools.  Note that this is an important anti-censorship tactic, because if the remote peer sees a positive match for a transaction in the client peer's `MemPoolSyncData`, it will *not* serve that transaction (even if it's a false positive).

Nodes can individually decide how many transactions they'll reply in a mempool query (the maximum is 8192, which is a protocol constant).  To make sure that different nodes serve different transactions even though they can serve different numbers of transactions, a node iterates through transactions in the sketch in a deterministic random order by using its *own* sketch seed to calculate a random permutation of transactions to consider.  Specifically, each `txid` in the mempool has a `randomized_txid` (i.e. derived as `siphash(node_seed, txid)`), and transactions are iterated through in `randomized_txid` order.  This way, a heavily-loaded node can get away with serving fewer than the maximum transactions, and downstream client peers are still just as likely to receive a transaction as any other.

Given that everyone's attention is currently on SIP-012, I'm leaving this in draft status for now.  I'd like to leave it running in production for a while to verify that a set of nodes reliably keep their mempools in sync.


jcnelson:
Okay, I have some data on this.  It looks like this is working!

I set up and ran a public node and a NAT'ed node and looked at the *differences* between their mempools every 5 minutes.  Specifically, I pulled the mempools from both nodes every 5 minutes, pulled out all the transactions in both at the current Stacks tip height, and took a diff between them (so the data here is a little noisy, due to the fact that the same tx can arrive at different heights when there's a block being propagated).  Here's what the differences look like between the public and NAT'ed nodes *without* this PR:

Mempool Tx Uniqueness, No Sync

And here's what the same differences look like with this PR applied:

Mempool Tx Unqiueness, With Sync

As you can see, this PR is a big improvement.

src/core/mempool.rs
```diff
@@ -299,13 +361,14 @@ const MEMPOOL_INITIAL_SCHEMA: &'static [&'static str] = &[
...
     "CREATE INDEX by_txid ON mempool(txid);",
+    "CREATE INDEX by_txid_and_height ON mempool(txid,height);",
     "CREATE INDEX by_sponsor ON mempool(sponsor_address, sponsor_nonce);",
     "CREATE INDEX by_origin ON mempool(origin_address, origin_nonce);",
     "CREATE INDEX by_timestamp ON mempool(accept_time);",
     "CREATE INDEX by_chaintip ON mempool(consensus_hash,block_header_hash);",
 ];
 
-const MEMPOOL_SCHEMA_2: &'static [&'static str] = &[
+const MEMPOOL_SCHEMA_COST_ESTIMATOR: &'static [&'static str] = &[
```
kantai, line 371: While `MEMPOOL_SCHEMA_2` isn't particularly descriptive, the name _does_ correspond to the DB version, which is helpful. Maybe these schema arrays should be named like `MEMPOOL_SCHEMA_2_COST_ESTIMATOR`?

src/core/mempool.rs
```diff
@@ -1360,16 +1640,231 @@ impl MemPoolDB {
...
+        Ok(num_written)
+    }
 }
 
 #[cfg(test)]
 mod tests {
+    use std::cmp;
+    use std::collections::HashSet;
+    use std::io;
+
```
kantai, line 1859: Can you move this module to another file? After this PR, roughly 60% of this file will be tests, and 3.5k line files with mixtures of tests and code make the codebase much harder to navigate.

src/net/chat.rs
```diff
@@ -621,6 +621,13 @@ impl ConversationP2P {
         self.burnchain_stable_tip_burn_header_hash.clone()
     }
 
+    /// Does thos remote neighbor support the mempool query interface?  It will if it has both
```
kantai, line 624: thos -> this

src/net/chat.rs
```diff
@@ -1127,7 +1134,7 @@ impl ConversationP2P {
         };
 
         debug!(
-            "Handshake from {:?} {} public key {:?} expires at {:?}",
+            "Handshake from {:?} {} public key {:?} services 0x{} expires at {:?}",
```
kantai, line 1137: This could be a good structured log:

```
debug!("Handling handshake"; 
             "neighbor" => ?self,
             "authentic_msg" => &authentic_msg,
             "public_key" =>  &to_hex(&hanshake_data...),
             "services" => &to_hex(&handshake_data.services.to_be_bytes()),
             "expires_block_height" => handshake_date.expire_block_height,
             "supports_mempool_query" => Self::supports_mempool_query(&handshake_data.services),
);
```

Using more structured logs in the networking code would help with debugging and tracking what's happening in the network stack (note, I'm also suggesting combining that later "supports_mempool_query" log into this log entry). Combining logs where possible would also go a long way towards making the debug logs more generally useful.

src/net/mod.rs
```diff
@@ -838,6 +839,19 @@ pub struct NatPunchData {
     pub nonce: u32,
 }
 
+#[repr(u8)]
+#[derive(Debug, Clone, PartialEq)]
+pub enum MemPoolSyncDataID {
+    BloomFilter = 0x01,
+    TxTags = 0x02,
+}
```
kantai, line 847: This should use the `define_u8_enum`. It provides a `to_u8()` method for variants, and a `from_u8(v: u8) -> Option<Self>` method. That avoids needing to do cast-matches, because you can check the result of `from_u8`, and then match on the enum, which allows the compiler to enforce that all variants are handled.

src/net/codec.rs
```diff
@@ -761,6 +766,42 @@ impl StacksMessageCodec for NatPunchData {
...
+            }
+            x if x == MemPoolSyncDataID::TxTags as u8 => {
+                let seed: [u8; 32] = read_next(fd)?;
+                let txtags: Vec<TxTag> = read_next(fd)?;
+                Ok(MemPoolSyncData::TxTags(seed, txtags))
+            }
+            _ => Err(codec_error::DeserializeError(format!(
+                "Unrecognized MemPoolSyncDataID {}",
+                &data_id
+            ))),
```
kantai, line 800: I dropped a note at the definition for MemPoolSyncDataID, but this match should take advantage of the `from_u8` method from the `define_u8_enum` variant. That way, we don't need to do the `x if x == .. as u8` matching pattern (which circumvent's rust's variant coverage checks).

src/net/connection.rs
```diff
@@ -374,6 +375,9 @@ pub struct ConnectionOptions {
     pub max_buffered_microblocks_available: u64,
     pub max_buffered_blocks: u64,
     pub max_buffered_microblocks: u64,
+    pub mempool_sync_interval: u64,
+    pub mempool_max_tx_query: u64,
+    pub mempool_sync_timeout: u64,
```
kantai, line 380: Can you add rustdocs for these fields?

src/net/db.rs
```diff
@@ -161,11 +161,11 @@ impl LocalPeer {
...
         let addr = addrbytes;
         let port = port;
-        let services = ServiceFlags::RELAY;
+        let services = (ServiceFlags::RELAY as u16) | (ServiceFlags::RPC as u16);
 
         info!(
-            "Will be authenticating p2p messages with public key: {}",
-            Secp256k1PublicKey::from_private(&pkey).to_hex()
+            "Will be authenticating p2p messages with public key: {}, service: {}",
+            Secp256k1PublicKey::from_private(&pkey).to_hex(), &to_hex(&(services as u16).to_be_bytes());
```
kantai, line 168: Use a structured log here (the log entry already is implicitly written as 2 keys and values).

src/util/bloom.rs
```diff
@@ -0,0 +1,1034 @@
...
+//  * https://github.com/crepererum/pdatastructs.rs
+//      * Problems:
+//          * No invertable bloom filter
+//          * Brings in lots of other code
+//
+//  * https://github.com/CodeChain-io/rust-iblt
+//      * Problems:
+//          * Can't control serialization/deserialization
+//          * Can't induce a bloom filter from the counts
+//          * Hilariously inefficient
```
kantai, line 33: While useful to know what alternatives were considered, I'm not sure this should live on in the comment-- perhaps just in the PR body?

src/chainstate/stacks/db/mod.rs
```diff
@@ -473,6 +474,17 @@ impl<'a> DerefMut for ChainstateTx<'a> {
     }
 }
 
+#[derive(Debug, PartialEq, Clone)]
+pub struct TxStreamData {
```
kantai, line 478: Can you add a rustdoc for this struct?

kantai: Actually, I think in `develop`, these stream structs became an enum, right? So when you merge, add a rustdoc for the variant.

src/core/mempool.rs
```diff
@@ -78,6 +90,56 @@ use crate::util::db::table_exists;
...
+// how far back in time (in Stacks blocks) does the bloom counter maintain tx records?
+pub const BLOOM_COUNTER_DEPTH: usize = 2;
+
+// maximum many tx tags we'll send before sending a bloom filter instead.
+// The parameter choice here is due to performance -- calculating a tag set can be slower than just
+// loading the bloom filter, even though the bloom filter is larger.
+const DEFAULT_MAX_TX_TAGS: u32 = 2048;
+
+#[derive(Debug, Clone, PartialEq, Hash, Eq)]
+pub struct TxTag(pub [u8; 8]);
```
kantai, line 111: Can you add a rustdoc for this struct?

src/core/mempool.rs
```diff
@@ -78,6 +90,56 @@ use crate::util::db::table_exists;
...
+
+impl TxTag {
+    pub fn from_seed_and_txid(seed: &[u8], txid: &Txid) -> TxTag {
```
kantai, line 114: Some function name bike shedding -- names like `from` are really meant to be conversion methods, and this is more like a constructor than a conversion, which usually should use `new`. Even so, it's clear from the function definition that it is "from" a seed and a txid, and there's no overloading happening (i.e., there's no other `from_x` methods), so we'd end up with more readable code if it were just `from` instead.

src/core/mempool.rs
```diff
@@ -956,11 +1114,124 @@ impl MemPoolDB {
...
 
+    /// Remove all txids at the given height from the bloom counter.
+    /// Used to clear out txids that are now outside the bloom counter's depth.
+    fn prune_bloom_counter(tx: &mut MemPoolTx, target_height: u64) -> Result<(), MemPoolRejection> {
+        let sql = "SELECT txid FROM mempool WHERE height = ?1 AND NOT EXISTS (SELECT 1 FROM removed_txids WHERE txid = mempool.txid)";
+        let args: &[&dyn ToSql] = &[&u64_to_sql(target_height)?];
+        let txids: Vec<Txid> = query_rows(tx, sql, args)?;
+        let num_txs = txids.len();
+
+        debug!("Prune bloom counter from height {}", target_height);
```
kantai, line 1125: Can this log entry be removed? It doesn't appear to add much that the log at 1137 isn't already.

src/core/mempool.rs
```diff
@@ -956,11 +1114,124 @@ impl MemPoolDB {
...
+
+        // keep borrow-checker happy
+        let bloom_counter = tx.take_bloom_state();
+        for txid in txids.into_iter() {
+            bloom_counter.remove_raw(&mut tx.tx, &txid.0)?;
+
+            let sql = "INSERT OR REPLACE INTO removed_txids (txid) VALUES (?1)";
+            let args: &[&dyn ToSql] = &[&txid];
+            tx.execute(sql, args).map_err(db_error::SqliteError)?;
```
kantai, line 1134: If these early exit, it seems like the bloom state could be corrupted (i.e., not replaced). I think the easiest way to make this "safe" would be to add a `using!` macro (like in `clarity_vm/clarity.rs`).

src/core/mempool.rs
```diff
@@ -956,11 +1114,124 @@ impl MemPoolDB {
         }
     }
 
+    /// Remove all txids at the given height from the bloom counter.
+    /// Used to clear out txids that are now outside the bloom counter's depth.
+    fn prune_bloom_counter(tx: &mut MemPoolTx, target_height: u64) -> Result<(), MemPoolRejection> {
```
kantai, line 1119: Can this be an instance method of `impl MemPoolTx` rather than a static?

src/core/mempool.rs
```diff
@@ -956,11 +1114,124 @@ impl MemPoolDB {
...
+        if present.is_none() && height > (BLOOM_COUNTER_DEPTH as u64) {
+            // this is the first-ever tx at this height.
+            // which means, the bloom filter window has advanced.
+            // which means, we need to remove all the txs that are now out of the window.
+            MemPoolDB::prune_bloom_counter(tx, height - (BLOOM_COUNTER_DEPTH as u64))?;
+        }
+
+        // keep borrow-checker happy
+        let bloom_counter = tx.take_bloom_state();
+
```
kantai, line 1170: Same comment as above -- most of these uses of `take_bloom_state` are pretty unsafe. Any thrown errors will cause the bloom state to be lost.

src/core/mempool.rs
```diff
@@ -956,11 +1114,124 @@ impl MemPoolDB {
...
+    }
+
+    /// Add the txid to the bloom counter in the mempool DB.
+    /// If this is the first txid at this block height, then also garbage-collect the bloom counter to remove no-longer-recent transactions.
+    /// If the bloom counter is saturated -- i.e. it represents more than MAX_BLOOM_COUNTER_TXS
+    /// transactions -- then pick another transaction to evict from the bloom filter and return its txid.
+    /// (Note that no transactions are ever removed from the mempool; we just don't prioritize them
+    /// in the bloom filter).
+    fn update_bloom_counter(
+        tx: &mut MemPoolTx,
```
kantai, line 1152: Same as above, can this be an instance method?

src/core/mempool.rs
```diff
@@ -956,11 +1114,124 @@ impl MemPoolDB {
         }
     }
 
+    /// Remove all txids at the given height from the bloom counter.
+    /// Used to clear out txids that are now outside the bloom counter's depth.
+    fn prune_bloom_counter(tx: &mut MemPoolTx, target_height: u64) -> Result<(), MemPoolRejection> {
+        let sql = "SELECT txid FROM mempool WHERE height = ?1 AND NOT EXISTS (SELECT 1 FROM removed_txids WHERE txid = mempool.txid)";
```
kantai, line 1120: Is this query implementable as a `JOIN`, e.g.,

```
SELECT A.txid FROM mempool as A LEFT OUTER JOIN removed_txids as B on A.txid = B.txid WHERE B.txid IS NULL AND A.height = ?1
```

which I think is usually more efficient than inner selects. Of course, this may be totally unnecessary optimization.

src/core/mempool.rs
```diff
@@ -956,11 +1114,124 @@ impl MemPoolDB {
...
+    /// If this is the first txid at this block height, then also garbage-collect the bloom counter to remove no-longer-recent transactions.
+    /// If the bloom counter is saturated -- i.e. it represents more than MAX_BLOOM_COUNTER_TXS
+    /// transactions -- then pick another transaction to evict from the bloom filter and return its txid.
+    /// (Note that no transactions are ever removed from the mempool; we just don't prioritize them
+    /// in the bloom filter).
+    fn update_bloom_counter(
+        tx: &mut MemPoolTx,
+        height: u64,
+        txid: &Txid,
+        prior_txid: Option<Txid>,
```
kantai, line 1155: Can you add a description of the `prior_txid` field to the rustdoc comments?

src/core/mempool.rs
kantai: One nice benefit of making these instance methods is that the MemPoolTx becomes "self" instead of "tx" (which can be muddled with "txid")

src/core/mempool.rs
```diff
@@ -956,11 +1114,124 @@ impl MemPoolDB {
...
+        Ok(evict_txid)
+    }
+
+    /// Add the txid to our randomized page order
+    fn update_mempool_pager(tx: &mut MemPoolTx, txid: &Txid) -> Result<(), MemPoolRejection> {
+        let mut randomized_buff = vec![];
+
+        let bloom_counter = tx.take_bloom_state();
+        randomized_buff.extend_from_slice(bloom_counter.get_seed());
+        tx.replace_bloom_state(bloom_counter);
```
kantai, line 1216: take/replace here looks like it could be `tx.bloom_state.as_ref().expect().get_seed()` instead, which is maybe more verbose, but it makes this method no longer need a mutable reference.

src/core/mempool.rs
kantai: (Same comment applies to other cases of the inner select)

src/core/mempool.rs
```diff
@@ -1360,16 +1640,231 @@ impl MemPoolDB {
...
+        };
+        let min_height = max_height.saturating_sub(BLOOM_COUNTER_DEPTH as u64);
+        let sql = "SELECT mempool.txid FROM mempool WHERE height > ?1 AND height <= ?2 AND NOT EXISTS (SELECT 1 FROM removed_txids WHERE txid = mempool.txid)";
+        let args: &[&dyn ToSql] = &[&u64_to_sql(min_height)?, &u64_to_sql(max_height)?];
+        query_rows(&self.conn(), sql, args)
+    }
+
+    /// Get the transaction tag list that represents the set of recent transactions we have.
+    /// Generate them with our node-local seed so that our txtag list is different from anyone
+    /// else's, w.h.p.
```
kantai, line 1679: `w.h.p.` = ?

src/core/mempool.rs
```diff
@@ -1360,16 +1640,231 @@ impl MemPoolDB {
...
+            if (ret.len() as u64) >= max_txs {
+                break;
+            }
+        }
+
+        Ok(ret)
+    }
+
+    /// Stream transaction data
+    pub fn stream_txs<W: Write>(
```
kantai, line 1789: Can you make this more descriptive? Is this method supposed to (1) query the mempool using the MemPoolSyncData query contained in the TxStreamData, (2) buffer the resulting transactions inside of the TxStreamData struct, and then (3) write from the TxStreamData struct's buffer into `fd`? That seems like something that could be called `handle_mempool_transaction_query` or something similar.

I also wonder about the structure of this streaming -- why use the TxStreamData to buffer data before writing to the fd? First, why buffer at all instead of just writing from the `find_next_missing_transactions` result into the fd. Second, why use the TxStreamData struct for the buffer, instead of a local buffer?

src/core/mempool.rs
```diff
@@ -1360,16 +1640,231 @@ impl MemPoolDB {
...
+    /// Get the next batch of transactions from our mempool that are *not* represented in the given
+    /// MemPoolSyncData.  Transactions are ordered lexicographically by randomized_txids.hashed_txid, since this allows us
+    /// to use the txid as a cursor while ensuring that each node returns txids in a deterministic random order
+    /// (so if some nodes are configured to return fewer than MAX_BLOOM_COUNTER_TXS transactions,
+    /// a requesting node will still have a good chance of getting something useful).
+    pub fn find_next_missing_transactions(
+        &self,
+        data: &MemPoolSyncData,
+        height: u64,
+        last_txid: &Txid,
```
kantai, line 1731: Can you name this so that we know that this isn't an actual txid, rather it's a randomized txid.

src/core/mempool.rs
```diff
@@ -1360,16 +1640,231 @@ impl MemPoolDB {
...
+                // find next page
+                let sql = "SELECT hashed_txid FROM randomized_txids WHERE txid = ?1 LIMIT 1";
+                let args: &[&dyn ToSql] = &[&next_tx.txid()];
+                let last_txid = match query_row(&self.conn(), sql, args)? {
+                    Some(txid) => txid,
+                    None => {
+                        // done!
+                        break;
+                    }
+                };
```
kantai, line 1843: The next page should be returned from the `find_next_missing_transactions` -- it's not only slightly more efficient to just return the data from that SQL query, but it also preserves the abstraction boundaries of that function to return the "next page", rather than relying on the caller to know how to figure out the next page.

src/net/p2p.rs
```diff
@@ -3028,11 +3050,355 @@ impl PeerNetwork {
...
+                            }
+                        },
+                    }
+                }
+            }
+        })
+    }
+
+    /// Do a mempool sync
+    /// Return true if we're done and can advance to the next state.
```
kantai, line 3319: Can you add more description to the rustdoc about the state transitions here? From reading the code, it seems like `do_mempool_sync` will allow the P2P state to advance after each of the following steps:

1. Picking a peer to sync with
2. Getting the data url for that peer
3. Sending the mempool query to the peer
4. Receiving data from the peer

And then once these steps have completed, it doesn't advance the P2P state? Is there a specific reason not to advance the P2P state at the end (or, to just directly invoke the next step of the mempool sync state?)

src/net/p2p.rs
```diff
@@ -3028,11 +3050,355 @@ impl PeerNetwork {
...
+                            // still waiting
+                            debug!(
+                                "Mempool sync event {} still waiting for a response",
+                                event_id
+                            );
+                            return Ok(false);
+                        }
+                        Some(http_response) => match http_response {
+                            HttpResponseType::MemPoolTxs(_, txs) => {
+                                network.mempool_sync_transactions = Some(txs);
```
kantai, line 3303: Is it possible to just return the synced transactions rather than adding them to the network state? It seems like this caller's caller is checking the `Option<>` field after every invocation to shuttle the data out into the network result object. It should be possible (and would simplify the data flow in the code) to just return this data instead, right?

src/net/p2p.rs
```diff
@@ -3028,11 +3050,355 @@ impl PeerNetwork {
...
+
+        // 2. resolve its data URL
+        // sets self.mempool_sync_sockaddr on success
+        if self.mempool_sync_sockaddr.is_none() {
+            if let Some(url_str) = self.mempool_sync_data_url.clone() {
+                if self.mempool_sync_resolve_data_url(&url_str, dns_client_opt)? {
+                    return Ok(true);
+                }
+            }
+        }
```
kantai, line 3362: The control flow of these if statements seems like it could lead to some unintended paths. For example:

```
self.mempool_sync_data_url = None
self.mempool_sync_pick_outbound_peer() => returns false
```

The control flow would then check all the other if statements in this method. Presumably they all return false (because at some point `self.mempool_sync_reset` was invoked), and so this method returns `Ok(false)`, but it seems like an unsafe pattern. In most of these cases, does it make sense to just return the inner function response, e.g.,

```rust
        if self.mempool_sync_data_url.is_none() {
            return self.mempool_sync_pick_outbound_peer(dns_client_opt);
        }
```

?

gregorycoppola:
@jcnelson the overall concept and results seem great. Very important component.

I'm really only interested to review the testing. Can you please add to your doc, or something, a summary of the test coverage we have for this. And, which tests we might like to have but don't at the moment.

src/core/mempool.rs
jcnelson: Sure; done.

src/core/mempool.rs
jcnelson: Yup, done

src/net/chat.rs
jcnelson: Done

src/net/chat.rs
jcnelson: Yup, done

src/net/mod.rs
jcnelson: Done

src/net/codec.rs
jcnelson: Yup, done

src/net/connection.rs
jcnelson: added

src/net/db.rs
jcnelson: added

src/util/bloom.rs
jcnelson: ha, yup, removing

src/chainstate/stacks/db/mod.rs
jcnelson: Yup, I refactored this as a `StreamCursor::MempoolTxs` data payload and added rust docs

src/core/mempool.rs
jcnelson: added

src/core/mempool.rs
jcnelson: sure, will use `from`

src/core/mempool.rs
jcnelson: yup, removed

src/core/mempool.rs
jcnelson: yup, will replace with safer code

src/core/mempool.rs
jcnelson: Yup, refactored

src/core/mempool.rs
jcnelson: Fixed

src/core/mempool.rs
jcnelson: Yup, refactored

src/core/mempool.rs
jcnelson: Yup, added

src/core/mempool.rs
jcnelson: Yup, refactored

src/core/mempool.rs
jcnelson: With high probability.  My math professors used it a lot

src/core/mempool.rs
```diff
@@ -1360,16 +1640,231 @@ impl MemPoolDB {
...
+    ) -> Result<u64, ChainstateError> {
+        let mut num_written = 0;
+        while num_written < count {
+            if query.num_txs >= query.max_txs {
+                // don't serve more than this many txs
+                break;
+            }
+
+            // write out bufferred tx
+            while query.tx_buf.len() > 0 && query.tx_buf_ptr < query.tx_buf.len() {
```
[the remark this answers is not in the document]
jcnelson: It does iterate, but it shouldn't -- the API contract is that "no more than `count` bytes shall be written out."  This code was violating this contract, but was successfully streaming out data.  I didn't notice it because the underlying `NetworkConnection<StacksHttp>` (a.k.a. `ConnectionHttp`) instance uses a `Pipe` instance to buffer writes from serialization code (which it will gladly fill up with a pending `StacksTransaction`'s bytes even if the HTTP server, which owns the `ConversationHttp` which owns the `NetworkConnection<StacksHttp>`, asked for less).  Will fix.

src/core/mempool.rs
jcnelson: So, this method is supposed to feed at most `count` bytes into `fd` from a stream of transactions loaded from the mempool that match a `MemPoolSyncData` query (owned by the given `TxStreamData`).  The `TxStreamData` is just a cursor -- it's all the state the HTTP server needs to remember in-between attempts to fill the underlying socket buffer with more transaction data as it becomes writeable.

While `fd` is just something that implements `Write`, in practice `fd` is a `PipeWrite` instance, which permits the networking code to buffer up messages for transmission without duplicating a lot of buffer book-keeping work.  The socket server consumes data from the corresponding `PipeRead` and writes *that* to the actual socket.  The reason `Pipe`, `PipeRead`, and `PipeWrite` exist as this layer of indirection is because the socket server doesn't know how many bytes are going to be available in the socket buffer until after it becomes writeable.  By passing a `PipeWrite` to `consensus_serialize()`, we ensure that `consensus_serialize()` always runs to completion without blocking, and the socket server takes care of reading the data out from its `PipeRead` in order to fill up socket buffer bytes as more space becomes available.  The various streaming methods, including this one, take a `count` argument to instruct the implementation to only push at most `count` bytes into the `PipeWrite` in order to keep memory usage under control.  The alternative is far messier -- we'd either have to write all our code assuming that `consensus_serialize()` could block for a long, long time, or implement `consensus_serialize()` to be "restartable" at a given byte offset.

Now, the reason we have a `tx_buf` at all in the `TxStreamData` is because we need the transaction being queried to stick around across multiple calls to `stream_txs` (which can be interleaved with mempool garbage collection), which necessitates copying it out of the mempool and into this cursor via `find_next_missing_transactions()`.  We could just as easily stored the `StacksTransaction` in the `TxStreamData`, but that would necessitate re-serializing the transaction each time the socket server asked for more bytes.  Using `tx_buf` and `tx_buf_ptr` is an optimization that ensures we only load and serialize the transaction once.

jcnelson: Also, the reason it's called `stream_txs` is that it follows the naming scheme of `stream_headers`, `stream_block`, `stream_microblocks_confirmed`, and `stream_microblocks_unconfirmed`.  In all cases, these method write at most `count` bytes to a given `Write` called `fd` (really a `PipeWrite`), and use state within their respective `StreamCursor` variants to keep track of what to write on the subsequent call.

src/core/mempool.rs
jcnelson: Yup, done

src/core/mempool.rs
jcnelson: Yup, refactored

src/net/p2p.rs
jcnelson: Your understanding of `do_mempool_sync()` is correct.  Also, the network state machine's state is advanced in `do_network_work()`, where all such writes to `self.work_state` take place (and is the only place I allow them to happen).

I'm considering refactoring this so that the mempool sync happens "in parallel" to the network state machine, much like how the neighbor walk does.  This is because I don't want to slow down inv and block syncs any more than need be.

src/net/p2p.rs
jcnelson: Sure; refactored

src/net/p2p.rs
jcnelson: I went ahead and refactored this so that the mempool state machine runs in parallel to the main state machine, like how neighbor walks do

src/net/p2p.rs
jcnelson: So, I admit I was a bit lazy here.  The more robust way of doing this would be to have an explicit `mempool_state` that indicated which state the mempool synchronizer was actually in.  Right now, the code paths here do the same work as having an explicit `mempool_state`, but implicitly.  Instead, I can have this code just run off a `match self.mempool_state {}` block that updates `self.mempool_state`, so that it's always clear what the state machine should be doing in reaction to attempted state transitions (like how we do in `do_network_work()`).

jcnelson:
@kantai Thanks for all your comments!  I think I've addressed them all now.

src/net/p2p.rs
jcnelson: I went ahead and refactored this so that there's now an explicit state machine loop for the mempool, and each state-transition function will return the data it generated instead of trying to store it to the `PeerNetwork`.  This way, all state-transition code and all state book-keeping (including resets) happen exclusively in the top-level mempool sync state machine loop.

src/net/rpc.rs
```diff
@@ -5077,4 +5199,69 @@ mod test {
...
+                    }
+                    _ => false,
+                }
+            },
+        );
+    }
+
+    #[test]
+    #[ignore]
+    fn test_rpc_mempool_query_bloom() {
```
gregorycoppola, line 5235: what is this testing?

src/net/rpc.rs
```diff
@@ -5077,4 +5199,69 @@ mod test {
...
+                    BLOOM_COUNTER_ERROR_RATE,
+                    MAX_BLOOM_COUNTER_TXS,
+                    BloomNodeHasher::new(&[0u8; 32]),
+                )))
+            },
+            |ref http_request, ref http_response, ref mut peer_client, ref mut peer_server| {
+                let req_md = http_request.metadata().clone();
+                println!("{:?}", http_response);
+                match http_response {
+                    HttpResponseType::MemPoolTxs(_, txs) => {
```
gregorycoppola, line 5257: why is this the answer?

src/util/bloom.rs
```diff
@@ -0,0 +1,1016 @@
+// Copyright (C) 2013-2020 Blockstack PBC, a public benefit corporation
```
gregorycoppola, line 1: https://app.codecov.io/gh/blockstack/stacks-blockchain/compare/2884/tree/src/util/bloom.rs

Code coverage is pretty good for this file. 

BitField::clear isn't tested and some error conditions aren't tested. Might as well add a test for clear while you're in the area, up to you about error conditions.

gregorycoppola (commented):
Again, empirical data looks good.

Are there any integration tests being checked in for this?

src/core/mempool.rs
kantai: Ah, well, I think it's probably uncommon enough to warrant expansion in the comment!

src/core/mempool.rs
kantai: Okay, I see why `query.tx_buff` is used in this case. I still think this function needs to be named something else and have an expanded rustdoc: `stream_txs` tells us relatively little about what this function is doing -- a function named like that implies that it's just blindly streaming transactions from one buffer into another, but this function is doing substantially more work than that -- maintaining request state, querying the database, etc., that's much closer to `handle_mempool_transaction_query`. (I'm not sure the other `stream_x` functions are well-named, either, but they're not directly related to this PR).

kantai (approved):
This looks good to me, my only remaining comment is (possibly bikeshedding?) about the name of `stream_transactions`. I do feel pretty strongly that it needs to have more details in its rustdoc at a minimum, and really should be called something other than `stream_transactions`.

pavitthrap (approved):
lgtm, just a few minor comments

src/net/rpc.rs
```diff
@@ -3489,6 +3543,63 @@ mod test {
         peer_1.sortdb = Some(sortdb1);
         peer_2.sortdb = Some(sortdb2);
 
+        // stuff some transactions into peer_2's mempool
```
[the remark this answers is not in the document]
jcnelson: Done

src/net/p2p.rs
```diff
@@ -2109,6 +2142,42 @@ impl PeerNetwork {
         Ok(done)
     }
 
+    /// Do a mempool sync. Return any transactions we might receive.
+    /// Return true if we finish the sync.
```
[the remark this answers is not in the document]
jcnelson: Fixed

src/net/p2p.rs
```diff
@@ -3030,11 +3100,433 @@ impl PeerNetwork {
...
+            debug!("Mempool sync state is {:?}", &cur_state);
+            match cur_state {
+                MempoolSyncState::PickOutboundPeer => {
+                    // 1. pick a random outbound conversation.
+                    // Sets self.mempool_sync_data_url on success.
+                    match self.mempool_sync_pick_outbound_peer(dns_client_opt)? {
+                        (false, Some(url)) => {
+                            // success! can advance
+                            self.mempool_sync_data_url = Some(url);
+                            self.mempool_state = MempoolSyncState::ResolveURL;
```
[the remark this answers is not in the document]
jcnelson: Sounds like a good idea.  Done.

src/core/mempool.rs
```diff
@@ -1360,1007 +1640,226 @@ impl MemPoolDB {
...
-            &TransactionAnchorMode::Any,
-            &TransactionPostConditionMode::Allow,
-        );
-        let num_txs = txs.len() as u64;
-
-        let mut mempool_tx = mempool.tx_begin().unwrap();
+    /// Make a mempool sync request.
+    /// If sufficiently sparse, use a MemPoolSyncData::TxTags variant
+    /// Otherwise, use a MemPoolSyncData::BloomFilter variant
+    /// If force_bloom_filter is true, then always make a bloom filter.  The reason for doin this
```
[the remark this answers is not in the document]
jcnelson: Fixed

src/core/mempool.rs
```diff
@@ -1360,1007 +1640,226 @@ impl MemPoolDB {
...
-            let txid = tx.txid();
-            let mut tx_bytes = vec![];
-            tx.consensus_serialize(&mut tx_bytes).unwrap();
-            let expected_tx = tx.clone();
+        let mut tags_table = HashSet::new();
+        if let MemPoolSyncData::TxTags(_, ref tags) = data {
+            for tag in tags.iter() {
+                tags_table.insert(tag.clone());
+            }
+        }
```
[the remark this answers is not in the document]
jcnelson: The size of `tags` is small enough in practice that it shouldn't matter.

src/core/tests/mod.rs
```diff
@@ -0,0 +1,1764 @@
...
+                    for txid in txids.iter() {
+                        if !recent_set.contains(&txid) && bf.contains_raw(&txid.0) {
+                            fp_count += 1;
+                        }
+                        if bf.contains_raw(&txid.0) {
+                            present_count += 1;
+                        } else {
+                            absent_count += 1;
+                        }
+                    }
```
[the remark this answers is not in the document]
jcnelson: This code path is verifying that the bloom filter's false positive rate is sufficiently small.  This does not apply to `TxTags`, since the `TxTags` query variant does not use a bloom filter.

src/core/tests/mod.rs
```diff
@@ -0,0 +1,1764 @@
...
+                        assert!(recent_set.contains(tag));
+                    }
+                }
+            }
+
+            let mut nonrecent_fp_rate = 0.0f64;
+            let recent_txids = mempool.get_bloom_txids().unwrap();
+            if recent_txids.len() < (present_count + absent_count) as usize {
+                nonrecent_fp_rate = (fp_count as f64)
+                    / ((present_count + absent_count - (recent_txids.len() as u32)) as f64);
```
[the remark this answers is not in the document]
jcnelson: No, this is deliberate -- we don't care to test the code in the if-block `if recent_txids.len() < (present_count + absent_count) as usize { /* ... */ }` if the query is a `TxTags` variant.

Issue #2874, opened 2021-10-13
Cost updates: New cost defaulting behavior


kantai:
_Note:_ As discussed in [More Accurate Cost Functions for Clarity Native Function](https://forum.stacks.org/t/more-accurate-cost-functions-for-clarity-native-functions/12386) and recent blockchain meetings, one path forward for updating the costs is a Pre-2.1 fork that _only_ covers cost and block limit updates. Separate from the process of formally proposing, voting on, and confirming such a proposal is the task of actually implementing those changes. This issue is related only to the task of implementing those changes in the `next-costs` branch of this repository.

Currently, there is a single default "costs" contract: `SP000000000000000000002Q6VF78.costs` that is used as the default when the Clarity VM loads up the cost functions.

This "defaulting" behavior happens in [src/vm/costs/mod.rs#L639](https://github.com/blockstack/stacks-blockchain/blob/master/src/vm/costs/mod.rs#L639):

```rust
let boot_costs_id = boot_code_id("costs", self.mainnet);

...

for f in ClarityCostFunction::ALL.iter() {
    let cost_function_ref = cost_function_references.remove(&f).unwrap_or_else(|| {
        ClarityCostFunctionReference::new(boot_costs_id.clone(), f.get_name())
		});

...
```

Basically, if a cost vote hasn't defined a new cost function for a particular function, then a default is used (`boot_costs_id` refers to `SP0000..costs`).

In `next-costs`, this behavior needs to change. After the new epoch begins, if `SP000.costs-2-05` has been instantiated in the current fork, then the default should be *that* contract. If the new contract has not been instantiated, then the old default should be used.

Issue #2873, opened 2021-10-13
Cost updates: Implement new cost contract instantion


kantai:
_Note:_ As discussed in [More Accurate Cost Functions for Clarity Native Function](https://forum.stacks.org/t/more-accurate-cost-functions-for-clarity-native-functions/12386) and recent blockchain meetings, one path forward for updating the costs is a Pre-2.1 fork that _only_ covers cost and block limit updates. Separate from the process of formally proposing, voting on, and confirming such a proposal is the task of actually implementing those changes. This issue is related only to the task of implementing those changes in the `next-costs` branch of this repository.

Currently, there is a single default "costs" contract: `SP000000000000000000002Q6VF78.costs` that is used as the default when the Clarity VM loads up the cost functions.

In the `next-costs` fork, a new `SP000000000000000000002Q6VF78.costs-2-05` contract must be instantiated when some specific block height has been reached. This must be apply across all forks.

Implementing this is equivalent to "backporting" PR #2639 to the cost-only fork branch: `next-costs`. This will, I think, need to pull in work that #2639 depended on as well (PR #2630).

PR #2630, opened 2021-04-27
Feat/stacks 2.1 epoch change

jcnelson:
This PR implements #2606 ~~and merges `develop` into `next`.~~

PR #2888, opened 2021-10-19
Feat/cost contract instantiation

pavitthrap:
Implements #2873 and #2874.

This PR backports the changes in #2630  and #2639. This change has the logic for epoch transitions, and introduces `Epoch2_05` as the next epoch. 

TODOs from the PR: 
- setting the exit block height (modify `STACKS_2_0_LAST_BLOCK_TO_PROCESS`) 
- adding `costs-2.clar`
- adding schema transition function for the sortition db

### DIFF
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -2031,12 +2031,11 @@ impl SortitionDB {
         };
 
         // TODO - add schema application method for sortition DB
+        let expected_version = SORTITION_DB_VERSION.to_string();
         match db.get_schema_version() {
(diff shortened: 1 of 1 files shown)


jcnelson:
I'm taking a crack at #2829 right now, and I'm discovering that there's not an easy way to get the current Stacks epoch within an `Environment` (or even a `GlobalContext`).  I'm happy to add this in a separate PR on top of this if you want -- ideally, the `GlobalContext` would just have the current Stacks epoch as a member, so it doesn't need to get loaded up each time we have some epoch-specific logic that needs to run.  Thoughts?

PR #2898, opened 2021-10-22
Stacks 2.05: ExecutionCost depends on StacksEpochId

### LINKED ISSUE #2875: Cost Updates: Implementing block limit change
_Note:_ As discussed in [More Accurate Cost Functions for Clarity Native Function](https://forum.stacks.org/t/more-accurate-cost-functions-for-clarity-native-functions/12386) and recent blockchain meetings, one path forward for updating the costs is a Pre-2.1 fork that _only_ covers cost and block limit updates. Separate from the process of formally proposing, voting on, and confirming such a proposal is the task of actually implementing those changes. This issue is related only to the task of implementing those changes in the `next-costs` branch of this repository.

Currently, a Stacks node has a single notion of the block limit. When the node run loop begins, the chainstate and any other structs that need a view of the block limit are configured with the current block limit (see: [src/run_loop/neon.rs#L236](https://github.com/blockstack/stacks-blockchain/blob/master/testnet/stacks-node/src/run_loop/neon.rs#L236)).

However, in order to change the block limit in a cost-changing fork, nodes will need to switch the block limit used when a given burnchain block has been reached.

gregorycoppola:
Fixes #2875 

## Description

Stacks 2.05 will introduce a new "epoch" at a certain burnblock height to be decided. When this epoch starts, we want to be able to change the block limits.

Thus, we replace `ExecutionCost` with `ExecutionCostSchedule`, which maps a `StacksEpochId` to a `ExecutionCost`. When the actual limit is needed, the call site can get the current epoch and use this to pick a specific `ExecutionCost` object to enforce as the limit.

## Testing

* new integration test in `neon_integrations`


jcnelson (changes requested):
Respectfully, I think you're overthinking this.

* The schedule of execution costs should be a system-wide constant, since they don't change.  Maybe you should put an `ExecutionCost` instance into a `StacksEpoch`, so it'll just be available with a call to `get_stacks_epoch()`?  There wouldn't be a need for a separate `ExecutionCostSchedule` if you do this -- it should just be part of the epoch itself.

* All instantiations of a Clarity connection created from a `StacksChainState` should continue to inherit the block limit set by the `StacksChainState`.  That's the whole point of putting it there -- the block limit schedule is a property of the chain state, not the VM (the VM merely enforces it on a per-block basis).  In turn, `StacksChainState` should use the block limit schedule defined by the stacks epochs constants, unless you're testing (or are building an appchain, but that's another story).

Issue #2864, opened 2021-10-05
Cost Updates: Use dynamic vs. static data sizes


kantai:
The cost assessment in Clarity for most data-handling functions (e.g., `map-get?`) use the _static_ cost of the fetch rather than the dynamic cost. See the following:

https://github.com/blockstack/stacks-blockchain/blob/1f85faf3ccc14823782d4ef482897b3566f97791/src/vm/functions/database.rs#L395-L423

There were two motivating reasons for doing this:

1. It makes static analysis of costs easier, because the cost assessed at runtime would always use the declared size of the map entry.
2. It allows the cost to be assessed _before_ running the operation.

However, it's pretty easy to argue that these reasons aren't all that essential. For (1), static analysis is always going to overestimate anyways, so why not allow the real cost to match the actual runtime overhead more closely. For (2), allowing a _single_ "speculative" evaluation before aborting a block due to cost overflow is not particularly burdensome to the network.

The benefit of using dynamic costs, however, could be great. Many contracts use patterns where potentially long lists are stored in data maps, but in practice the stored lists are relatively short.

PR #2901, opened 2021-10-26
Stacks 2.05: Use dynamic type size in db cost methods

kantai:
This PR implements #2864, using the actual size of stored objects as inputs to the database cost methods (and `concat`)

As part of this change, the current Stacks epoch is added to the `GlobalContext` (technically, the Stacks epoch that elected the parent Stacks block).


src/vm/functions/assets.rs
```diff
@@ -332,7 +353,70 @@ pub fn special_mint_asset(
...
+            &asset,
+            expected_asset_type,
+        ) {
+            Err(Error::Runtime(RuntimeErrorType::NoSuchToken, _)) => Ok(()),
+            Ok(_owner) => return clarity_ecode!(MintAssetErrorCodes::ALREADY_EXIST),
+            Err(e) => Err(e),
+        }?;
+
+        env.add_memory(TypeSignature::PrincipalType.size() as u64)?;
+        env.add_memory(expected_asset_type.size() as u64)?;
```
jcnelson, line 397: Argh, sorry, commented on the commit itself.  Just wanted some clarity (ha ha), this appears to be assessing the memory usage by the NFT's type signature, instead of the actual size (which is now used for the runtime cost). Is this intentional?

jcnelson: This appears to apply to all of the new NFT methods.

jcnelson:
This LGTM, but I think there's still a bug in the way we're assessing runtime vs memory costs for NFTs -- the runtime cost is taking the NFT value's size, but the memory cost assessment is based on the type signature (whereas the new methods for getting, storing, and deleting data use the value size for both runtime and memory cost assessment).

src/vm/functions/assets.rs
kantai: Yep -- good catch, we should use the actual size for the memory usage. Fixed in b83b8e1

jcnelson (approved):
LGTM!

PR #2902, opened 2021-10-27
Stacks 2.05: ExecutionCost inside of StacksEpoch

### LINKED ISSUE #2875: Cost Updates: Implementing block limit change

gregorycoppola:
Fixes #2875

This PR allows the `block_limit` to vary according to the epoch.

This is done by adding `block_limit: ExecutionCost` to `StacksEpoch`.


pavitthrap:
Is this ready for review? Looks like some of Aaron's comments are still outstanding

src/core/mod.rs
```diff
@@ -123,23 +123,6 @@ pub const POX_THRESHOLD_STEPS_USTX: u128 = 10_000 * (MICROSTACKS_PER_STACKS as u
...
-
-pub const HELIUM_BLOCK_LIMIT: ExecutionCost = ExecutionCost {
-    write_length: 15_0_000_000,
-    write_count: 5_0_000,
-    read_length: 1_000_000_000,
-    read_count: 5_0_000,
-    // allow much more runtime in helium blocks than mainnet
-    runtime: 100_000_000_000,
-};
-
```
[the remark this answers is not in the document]
gregorycoppola: I tried this once but you can't clone into this kind of `pub const` constant.

```
error[E0015]: calls in constants are limited to constant functions, tuple structs and tuple variants
   --> src/core/mod.rs:324:22
    |
324 |         block_limit: BLOCK_LIMIT_MAINNET_20.clone(),
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

What are your thoughts on that?

src/clarity_vm/clarity.rs
```diff
@@ -294,13 +294,11 @@ impl ClarityInstance {
             None => None,
         };
 
-        if epoch_found.is_some() {
-            epoch_found.unwrap()
-        } else {
+        epoch_found.unwrap_or_else(||
             burn_state_db
                 .get_stacks_epoch_by_epoch_id(&StacksEpochId::Epoch20)
                 .expect(&format!("Failed to get Stacks epoch for Epoch20"))
```
kantai, line 300: You don't need the `format!` call here: `.expect("Failed to get Stacks epoch for Epoch20")`

src/chainstate/burn/db/sortdb.rs
```diff
@@ -6807,16 +6822,19 @@ pub mod tests {
                     epoch_id: StacksEpochId::Epoch10,
                     start_height: 0,
                     end_height: 8,
+                    block_limit: ExecutionCost::max_value(),
                 },
                 StacksEpoch {
                     epoch_id: StacksEpochId::Epoch20,
                     start_height: 8,
                     end_height: 12,
+                    block_limit: ExecutionCost::max_value(),
```
jcnelson, line 6831: Shouldn't these use the proper max values for their respective epochs?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -5820,3 +5819,313 @@ fn atlas_stress_integration_test() {
...
+
+    true
+}
+
+/// This test checks that the block limit is changed at Stacks 2.05. We lower the allowance, and
+/// check that we can 1) afford the function call before the target height and
+/// 2) cannot afford the function call after the target height.
+#[test]
+#[ignore]
+fn test_cost_limit_switch_version205() {
```
kantai, line 5889: Can you move this test to the `epoch_205` module?

And also add a line to `.github/workflows/bitcoin-tests.yml` to add this test to the continuous integrations.

src/chainstate/burn/db/sortdb.rs
kantai: This test isn't doing any checks of the `block_limit`, it's just testing the sortition db's epoch tracking functionality, so using `max_value()` (or really any value) doesn't matter in this test.

src/chainstate/stacks/db/mod.rs
```diff
@@ -1327,35 +1320,31 @@ impl StacksChainState {
...
         chain_id: u32,
         path_str: &str,
         boot_data: Option<&mut ChainStateBootData>,
-        block_limit: ExecutionCost,
     ) -> Result<(StacksChainState, Vec<StacksTransactionReceipt>), Error> {
-        StacksChainState::open_and_exec(false, chain_id, path_str, boot_data, block_limit)
+        StacksChainState::open_and_exec(false, chain_id, path_str, boot_data)
     }
 
     pub fn open_with_block_limit(
```
jcnelson, line 1335: Does this method still need to exist if it no longer takes a block limit?

src/chainstate/stacks/db/mod.rs
```diff
@@ -1327,35 +1320,31 @@ impl StacksChainState {
     /// parameters, but with a block limit
     pub fn reopen_limited(
```
jcnelson, line 1321: There's no longer a need to call this `reopen_limited()` if an execution cost is not going to be given to constrain the execution budget.

src/clarity.rs
```diff
@@ -58,8 +58,22 @@ use burnchains::Txid;
 use chainstate::stacks::boot::{STACKS_BOOT_CODE_MAINNET, STACKS_BOOT_CODE_TESTNET};
 use util::boot::{boot_code_addr, boot_code_id};
 
-use core::BLOCK_LIMIT_MAINNET;
-use core::HELIUM_BLOCK_LIMIT;
+const BLOCK_LIMIT_MAINNET_STACKS20: ExecutionCost = ExecutionCost {
```
jcnelson, line 61: Style nit: `BLOCK_LIMIT_MAINNET_2_0`.

src/clarity.rs
```diff
@@ -612,6 +626,7 @@ impl HeadersDB for CLIHeadersDB {
         }
     }
     fn get_burn_block_height_for_block(&self, id_bhh: &StacksBlockId) -> Option<u32> {
+        warn!("check");
```
jcnelson, line 629: Please remove

src/clarity_vm/clarity.rs
```diff
@@ -274,28 +272,33 @@ impl ClarityInstance {
...
         stacks_block: &StacksBlockId,
         header_db: &dyn HeadersDB,
         burn_state_db: &dyn BurnStateDB,
-    ) -> StacksEpochId {
-        match header_db.get_burn_block_height_for_block(stacks_block) {
+    ) -> StacksEpoch {
+        let epoch_found = match header_db.get_burn_block_height_for_block(stacks_block) {
             Some(burn_height) => {
                 // special case the Stacks 2.0 genesis block -- it occurs at a zero burn block height
                 if burn_height == 0 {
```
jcnelson, line 285: Woah, this is absolutely wrong and broken.  The first-ever `burn_height` is going to be 666050 on mainnet, and something entirely different on testnet.  It's only 0 on regtest.  Also, why is this branch even needed?  Doesn't `get_burn_block_height_for_block()` work for all index block hashes, including the genesis one?

src/clarity.rs
jcnelson: Also, why is this here, and not in `src/core/mod.rs`?

src/cost_estimates/tests/cost_estimators.rs
```diff
@@ -55,6 +60,14 @@ impl CostMetric for TestCostMetric {
     }
 }
 
+const BLOCK_LIMIT_MAINNET_STACKS20: ExecutionCost = ExecutionCost {
```
jcnelson, line 63: Why is this duplicated here?

src/core/mod.rs
```diff
@@ -292,16 +290,38 @@ pub const STACKS_EPOCHS_MAINNET: &[StacksEpoch] = &[
...
+            read_length: 100_000_000,
+            read_count: 7_750,
+            runtime: 5_000_000_000,
+        },
     },
     StacksEpoch {
         epoch_id: StacksEpochId::Epoch2_05,
         start_height: STACKS_2_0_LAST_BLOCK_TO_PROCESS + 1,
         end_height: STACKS_EPOCH_MAX,
+        // TODO: Fill this in with the determined 2.05 costs.
```
jcnelson, line 317: Don't we already know what the limits will be?  They're specified in SIP 012.

src/core/mod.rs
```diff
@@ -292,16 +290,38 @@ pub const STACKS_EPOCHS_MAINNET: &[StacksEpoch] = &[
         epoch_id: StacksEpochId::Epoch10,
         start_height: 0,
         end_height: BITCOIN_MAINNET_FIRST_BLOCK_HEIGHT,
+        block_limit: ExecutionCost {
```
jcnelson, line 293: Stacks 1.0 didn't have smart contracts, so the ExecutionCost for it should be zero in all dimensions.

src/core/mod.rs
```diff
@@ -323,16 +357,37 @@ pub const STACKS_EPOCHS_REGTEST: &[StacksEpoch] = &[
         epoch_id: StacksEpochId::Epoch10,
         start_height: 0,
         end_height: 0,
+        block_limit: ExecutionCost {
```
jcnelson, line 360: Should be zeros.

src/cost_estimates/mod.rs
```diff
@@ -103,7 +107,8 @@ pub fn estimate_fee_rate<CE: CostEstimator + ?Sized, CM: CostMetric + ?Sized>(
     metric: &CM,
 ) -> Result<f64, EstimatorError> {
     let cost_estimate = estimator.estimate_cost(&tx.payload)?;
-    let metric_estimate = metric.from_cost_and_len(&cost_estimate, tx.tx_len());
+    // DO NOT SUBMIT: fix this
```
jcnelson, line 110: Please fix this before merging.  Also, as a general practical rule, please don't submit PRs for review if they have stuff like this in them.

src/main.rs
```diff
@@ -489,7 +489,15 @@ simulating a miner.
         tx_signer.sign_origin(&sk).unwrap();
         let coinbase_tx = tx_signer.get_tx().unwrap();
 
-        let mut settings = BlockBuilderSettings::limited(core::BLOCK_LIMIT_MAINNET.clone());
+        let stacks20_block_limit_mainnet = ExecutionCost {
```
jcnelson, line 492: Why is this duplicated here?

src/net/rpc.rs
```diff
@@ -1618,7 +1619,14 @@ impl ConversationHttp {
                         .send(http, fd);
                 }
             };
-            let scalar_cost = metric.from_cost_and_len(&estimated_cost, estimated_len);
+            let tip = SortitionDB::get_canonical_burn_chain_tip(sortdb.conn()).unwrap();
+            use vm::database::BurnStateDB;
```
jcnelson, line 1623: Please don't do this.  Put all module `use` statements at the beginning of the file, unless it's truly impossible to do so (e.g. there's an unavoidable name conflict).

src/cost_estimates/tests/metrics.rs
```diff
@@ -115,7 +115,14 @@ fn test_proportional_dot_product() {
 
 #[test]
 fn test_proportional_dot_product_with_mainnet_lims() {
-    let metric = ProportionalDotProduct::new(MAX_BLOCK_LEN as u64, BLOCK_LIMIT_MAINNET.clone());
+    let block_limit: ExecutionCost = ExecutionCost {
```
jcnelson, line 118: If you're going to use the mainnet limit, just use the constant from `core`.

src/net/rpc.rs
```diff
@@ -1618,7 +1619,14 @@ impl ConversationHttp {
                         .send(http, fd);
                 }
             };
-            let scalar_cost = metric.from_cost_and_len(&estimated_cost, estimated_len);
+            let tip = SortitionDB::get_canonical_burn_chain_tip(sortdb.conn()).unwrap();
```
jcnelson, line 1622: I've said this before, but I guess I need to say it again:  we do *NOT* `unwrap()` in network-facing code.  If it can fail, it needs to be handled properly, such as by returning an HTTP 500.

src/vm/docs/mod.rs
```diff
@@ -1847,6 +1849,7 @@ mod test {
             Some(1557860301)
         }
         fn get_burn_block_height_for_block(&self, _id_bhh: &StacksBlockId) -> Option<u32> {
+            warn!("check");
```
jcnelson, line 1852: Please remove

testnet/stacks-node/src/burnchains/mocknet_controller.rs
```diff
@@ -89,16 +90,22 @@ impl BurnchainController for MocknetController {
         &mut self,
         _ignored_target_height_opt: Option<u64>,
     ) -> Result<(BurnchainTip, u64), BurnchainControllerError> {
+        // If `config` sets a value, use that. Otherwise, use a default.
```
jcnelson, line 93: Why is the epoch even configurable in the first place?

testnet/stacks-node/src/node.rs
```diff
@@ -879,12 +879,18 @@ impl Node {
         let mut cost_estimator = self.config.make_cost_estimator();
         let mut fee_estimator = self.config.make_fee_estimator();
 
+        // REVIEW QUESTION: Should this go in the "if block"?
```
jcnelson, line 882: which `if` block?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -5820,3 +5819,313 @@ fn atlas_stress_integration_test() {
...
+    return result;
+}
+
+/// This function will call `next_block_and_wait` until the burnchain height underlying `BitcoinRegtestController`
+/// reaches *at least* `target_height`.
+///
+/// Returns `false` if `next_block_and_wait` times out.
+///
+/// Note: The user cannot assume that the block height reached is *exactly* `target_height`. It
+/// seems that the chain may often advanced `1` block past `target_height`.
```
jcnelson, line 5853: Isn't this a bug, then?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -5820,3 +5819,313 @@ fn atlas_stress_integration_test() {
...
+        &make_contract_publish(
+            &creator_sk,
+            0,
+            1100000,
+            "increment-contract",
+            &giant_contract,
+        ),
+    );
+
+    // Wait to make sure the contract is published.
```
jcnelson, line 6031: If this is unreliable, then you can simply query the node after each burnchain block to see if the contract was instantiated.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -5820,3 +5819,313 @@ fn atlas_stress_integration_test() {
...
+            }
+            _ => false,
+        },
+    );
+    assert_eq!(increment_calls_alice.len(), 1);
+
+    // Clear the observer so we can look for Bob's transaction.
+    test_observer::clear();
+
+    // The cost contract was switched at height 220. So, now we expect Bob's call to fail.
```
jcnelson, line 6090: Can you test the edge case here?  Does the new block limit come into effect on block 220, or block 221?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -5820,3 +5819,313 @@ fn atlas_stress_integration_test() {
...
+
+    // Wait for the contract to finish.
+    run_until_burnchain_height(
+        &mut btc_regtest_controller,
+        &blocks_processed,
+        &burnchain_height,
+        235,
+    );
+
+    // Bob's calls didn't work because he called after the block limit was lowered.
```
jcnelson, line 6118: Is there a way to query the miner for transactions that got rejected?  Then you could verify that the miner tried to apply Bob's transaction, but failed to do so.

jcnelson (changes requested):
There is at least one show-stopping consensus bug in this PR, as well as a potential network DDoS bug.

testnet/stacks-node/src/burnchains/mocknet_controller.rs
kantai: It's very useful for testing to make the epochs configurable (see the epoch switch tests in the other PR), as well as for things like private testnet testing -- having exactly one vector of epochs for testnet and regtest is pretty painful.

src/clarity_vm/clarity.rs
kantai: This ends up being invoked a _huge_ amount in the unit tests. I think it ends up getting invoked in all of the TestPeer tests and some of the other tests as well.

I agree that it should not ever be invoked in production, so I think it should be okay to make this branch something like `cfg(test)` only.

kantai:
Before addressing the current feedback in the PR, @gregorycoppola, can you also pull in updates and resolve conflicts from the `feat/2864...` branch?

src/cost_estimates/mod.rs
gregorycoppola: This one is still TODO.

testnet/stacks-node/src/tests/neon_integrations.rs
gregorycoppola: these are possibilities but do you want to hold up the PR for this?

gregorycoppola (commented):
Here are some interim responses, the call to `cost_estimates::estimate_fee_rate` remains TBD.

src/core/mod.rs
kantai: Yes -- the block limits in SIP-012 are more or less finalized -- there's no new benchmarking happening that would impact the block limit definition there.

testnet/stacks-node/src/tests/neon_integrations.rs
kantai: See my test in `epoch_205::test_dynamic_db_method_costs` -- it does an exact check on the cut over. It does "overrun" a bit, but it explicitly checks the receipts  make sure that the transition is done at the right block.

testnet/stacks-node/src/tests/neon_integrations.rs
kantai: No, I don't believe that there's a way to do this.

src/clarity_vm/clarity.rs
jcnelson: Yeah, at the very least this needs to be `#[cfg(test)]`.  In all non-test cases, `get_stacks_epoch()` should cover the full domain of burnchain block heights.  I.e., `get_stacks_epoch(0)` should be `Some(..)`.

gregorycoppola: I'm trying to take out the special case, and testing it out on GitHub to see what fails on there. We can roll back.

But, seems like maybe the bandaid shouldn't go here but instead inside the mock BurnStateDB.

gregorycoppola: I put a "cfg(test)" part in the if statement.. or if there is a different style maybe soeone can suggest it.

jcnelson:
If you look at the error transcript when you remove `burn_height == 0`, it's because of a cargo build error, not a logic error.

gregorycoppola:
@jcnelson Do you have a link to what you're looking at?

> If you look at the error transcript when you remove `burn_height == 0`, it's because of a cargo build error, not a logic error.

Locally I get during tests:
```
thread panicked while panicking. aborting.
error: test failed, to rerun pass '--lib'
```

And I saw that once online but now I'm not sure where so I just re-pushed it.

src/core/mod.rs
gregorycoppola: updated

testnet/stacks-node/src/tests/neon_integrations.rs
gregorycoppola: Now stopping on exact block and asserting this

src/clarity_vm/clarity.rs
gregorycoppola: Proposal is to:
1) use the "cfg(test)" fix for now
2) fix in issue #2907

src/net/rpc.rs
kantai: As of 340985579, there should be no more panics in the `net` paths.

src/clarity.rs
kantai: All these cases just use the `core` definition as of 340985579

src/core/mod.rs
kantai: Updated as of 340985579

src/core/mod.rs
kantai: Updated as of 340985579

src/chainstate/stacks/db/mod.rs
kantai: Yep, that's correct-- we still want to `reopen`.

src/clarity_vm/clarity.rs
```diff
@@ -275,27 +281,38 @@ impl ClarityInstance {
...
-            None => GENESIS_EPOCH,
+            None => None,
+        };
+
+        // Step 2: If not found, default to Epoch20.
+        match epoch_found {
+            Some(epoch) => epoch,
+            None => burn_state_db
+                .get_stacks_epoch_by_epoch_id(&GENESIS_EPOCH)
+                .expect("Failed to get Stacks epoch for GENESIS_EPOCH"),
```
kantai, line 315: Continuing the thread from the comment chain on the commit with @jcnelson and @gregorycoppola.

Right now, this performs the special case handling only in `#cfg(test)`. This special case handling was actually already present in my previous PR, so I think it's somewhat orthogonal to the work of this PR. I opened an issue #2907 for resolving this. I do not think that this issue should block this PR, though.

Does that sound okay to you, @jcnelson ?

src/cost_estimates/tests/cost_estimators.rs
kantai: This has been resolved at least by commit 340985579

src/core/mod.rs
kantai: Resolved at least by 340985579

src/cost_estimates/mod.rs
kantai: This has also been fixed by at least 340985579

src/clarity_vm/clarity.rs
jcnelson: > This special case handling was actually already present in my previous PR

I would have raised this issue then if I had caught it then.

I'm very uncomfortable as to why we can't answer the question "why does `burn_state_db.get_stacks_epoch(0)` return `None`?"  Aren't epochs defined for each and every block height in the `u64` domain?

testnet/stacks-node/src/tests/neon_integrations.rs
kantai: Bounds have been tightened up on this test, so marking this thread resolved -- if there's still issues with this test, we should discuss them on the PR, not the commit (this test has moved to a different file entirely in the PR).

src/clarity_vm/clarity.rs
kantai: These codepaths only occur when using mocked burn state databases. Fixing this will involve synchronizing all of the "first burnchain height" parameters passed to different unit tests with their instantiated mock header dbs. It's possible, but it's a task that's pretty unrelated to this PR.

testnet/stacks-node/src/tests/neon_integrations.rs
kantai: Similar to the other comment threads -- the tighter bounds on the burnchain height addresses some of this issue, so if there's still more we want out of this integration test, we should discuss on the PR itself.

testnet/stacks-node/src/tests/neon_integrations.rs
kantai: I think this is something that may be worthwhile to add to our integration testing framework, but it's pretty unrelated to this PR -- many of our integration tests already do this kind of "assumed transaction rejection" testing (all of the overflow tests).

kantai: If we want to continue discussing this, let's do it on the PR, not the commit!

src/core/mod.rs
```diff
@@ -289,54 +324,68 @@ impl Ord for StacksEpoch {
...
+            epoch_id: StacksEpochId::Epoch20,
+            start_height: 0,
+            end_height: 1000,
+            block_limit: HELIUM_BLOCK_LIMIT_20.clone()
+        },
+        StacksEpoch {
+            epoch_id: StacksEpochId::Epoch2_05,
+            start_height: 1000,
+            end_height: STACKS_EPOCH_MAX,
+            block_limit: HELIUM_BLOCK_LIMIT_20.clone()
```
jcnelson, line 385: Is there a 2.05 helium block limit?

testnet/stacks-node/src/tests/mod.rs
```diff
@@ -439,31 +440,41 @@ fn should_succeed_mining_valid_txs() {
...
+                ).unwrap();
             },
             2 => {
                 // On round 2, publish a "get:foo" transaction
                 // ./blockstack-cli --testnet contract-call 043ff5004e3d695060fa48ac94c96049b8c14ef441c50a184a6a3875d2a000f3 10 1 STGT7GSMZG7EA0TS6MVSKT5JC1DCDFGZWJJZXN8A store get-value -e \"foo\"
                 let get_foo = "8080000000040021a3c334fc0ee50359353799e8b2605ac6be1fe40000000000000001000000000000000a0100b7ff8b6c20c427b4f4f09c1ad7e50027e2b076b2ddc0ab55e64ef5ea3771dd4763a79bc5a2b1a79b72ce03dd146ccf24b84942d675a815819a8b85aa8065dfaa030200000000021a21a3c334fc0ee50359353799e8b2605ac6be1fe40573746f7265096765742d76616c7565000000010d00000003666f6f";
-                tenure.mem_pool.submit_raw(&mut chainstate_copy, &consensus_hash, &header_hash,hex_bytes(get_foo).unwrap().to_vec()).unwrap();
+                tenure.mem_pool.submit_raw(&mut chainstate_copy, &consensus_hash, &header_hash,hex_bytes(get_foo).unwrap().to_vec(),
+                                &ExecutionCost::max_value()
+                                           ).unwrap();
```
jcnelson, line 453: Holy whitespace Batman! Does this past `cargo fmt`?

src/core/mod.rs
kantai: Not unless there's a reason for one?

testnet/stacks-node/src/tests/mod.rs
kantai: I think `cargo fmt` specifically wanted this 🤷‍♂️

src/clarity_vm/clarity.rs
jcnelson: Okay, but I'd like to ship #2907 in `next-costs` if at all possible.  Adding test-specific special cases to production code paths -- especially consensus-critical code paths like this -- makes me deeply uncomfortable.

jcnelson (approved):
LGTM; just minor comments at this point.

src/core/mod.rs
jcnelson: Works for me.

testnet/stacks-node/src/tests/mod.rs
gregorycoppola: Yeah seems like `cargo fmt` bug.

I would rather leave cases like this than start manually formatting code. But it can be manually formatted.

kantai: Yeah, let's definitely leave cases like this

src/clarity_vm/clarity.rs
kantai: Yeah, it should be possible -- I already started working on the patchset for the change. I think the best approach is to do something like:

1. Create a new NullHeadersDB and NullBurnstateDB which are like UnitTestHeadersDB and UnitTestBurnstateDB, which are `#[cfg(test)]`
2. Make NullHeadersDB and NullBurnstateDB actually just always panic on invocation. I need to enumerate all of their usage outside of testing, but I'm pretty sure its just in the evaluation of cost functions (which panic on database lookups: cost functions should never invoke db methods).
3. Make sure that unit test headers/burn dbs return correctly enough mocked information to pass tests.

kantai: Okay b90a5ba20 contains the changes:

1. Fix the panic while panicking issue which made running down test failures of this kind really annoying.
2. Explicit special case for when the parent is the "FIRST_STACKS_BLOCK". That "block" _isn't ever_ elected, so it would fail to lookup a burn block height (or the height wouldn't be sane). I don't see a way around this special case.
3. Add `UnitTest..` versions of the mock burn and headers interfaces. These versions return some sane-ish mock results for block height.
4. Alter the read-only connections to use the provided `at_block`'s _evaluated_ epoch version, rather than trying to figure out the epoch that elected the block. The difference is subtle, but I believe this is the desired behavior (the read-only connections will be evaluated under the same rules that `at_block` was evaluated under, rather than the rules that `at_block`'s descendants will be evaluated under).

jcnelson: Re (4), does this go against what we were thinking for 2.1 -- i.e. the `at-block` code body runs with the VM rules of the contract containing the `at-block` invocation, instead of the VM rules of the block indicated by `at-block`?

jcnelson: Re (4), so I understand correctly (since this is what the code is now doing) -- the VM rules that get applied by an `at-block` are the rules in which the `at-block` is running, not the rules that were in place in the ancestor block in which the `at-block` code body is meant to be run?  I.e. if `(at-block X (print "hello"))` runs in block Y, then the rules for Y, not X, are applied to `(print "hello")`, right?

kantai: Ah, point (4) above doesn't impact `at-block`, rather I'm talking about the `eval_read_only` style methods that things like RPCs use (and the chains coordinator uses to get the amount of liquid stx). Those methods open a read-only connection, but need to supply an EpochId for the GlobalContext.

When it comes to 2.1, we'll need to be careful to make sure that it uses the same method of figuring the epoch that 2.05 uses -- basically, the epoch of the parent governs the current epoch.

gregorycoppola:
thanks for review, @jcnelson

src/clarity_vm/clarity.rs
kantai: The one thing left I want to highlight is the change in 85a42f0816857d8c307acf3aa8cf215711d29680. The goal there was just to eliminate an uneccessary usage of the NULL_BURN_STATE_DB by doing the `rollback_unconfirmed` invocation directly. Does that change make sense @jcnelson ?

jcnelson: Yes, I think that's fine.  As long as the call to `datastore.rollback_unconfirmed()` is preceded by a `datastore.begin_unconfirmed()`, then the behavior should be the same -- i.e. the inner MARF will be pointed to the unconfirmed state trie, so when we roll it back, the unconfirmed state trie gets deleted.

kantai (approved):
LGTM!

Issue #2892, opened 2021-10-20
Cost updates: Add transition function for sortition DB schema 


pavitthrap:
The latest schema version for the sortition DB is 2. 

We need a "transition" function that applies the newest schema changes if the version of the db is less than 2.

PR #2903, opened 2021-10-28
Added sortition db schema transition logic

pavitthrap:
Addresses #2892 - adds a transition function for the sortition DB schema. 

Ran existing tests, and they passed.

### DIFF
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -2262,7 +2262,7 @@ impl SortitionDB {
         Ok(version)
     }
 
-    fn apply_schema_2(tx: &SortitionHandleTx) -> Result<(), db_error> {
+    fn apply_schema_2(tx: &SortitionDBTx) -> Result<(), db_error> {
         for sql_exec in SORTITION_DB_SCHEMA_2 {
             tx.execute_batch(sql_exec)?;
         }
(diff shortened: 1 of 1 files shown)


src/chainstate/burn/db/sortdb.rs
```diff
@@ -2267,6 +2377,37 @@ impl SortitionDB {
             .optional()?;
         Ok(version)
     }
+
+    fn apply_schema_2(tx: &SortitionDBTx) -> Result<(), db_error> {
+        for sql_exec in SORTITION_DB_SCHEMA_2 {
+            tx.execute_batch(sql_exec)?;
+        }
+
+        tx.execute("INSERT INTO db_config (version) VALUES (?1)", &["2"])?;
```
jcnelson, line 2386: Should be `INSERT OR REPLACE INTO`

src/chainstate/burn/db/sortdb.rs
```diff
@@ -597,14 +597,15 @@ const SORTITION_DB_INITIAL_SCHEMA: &'static [&'static str] = &[
     );"#,
     "CREATE INDEX canonical_stacks_blocks ON canonical_accepted_stacks_blocks(tip_consensus_hash,stacks_block_hash);",
     "CREATE TABLE db_config(version TEXT NOT NULL);",
-    r#"
+];
```
jcnelson, line 600: There should only ever be one row in `db_config`, and it should be the current version.  Can you make the schema for `db_config` to be `db_config(version TEXT PRIMARY KEY)` instead, and use `INSERT OR REPLACE` when updating the version?

jcnelson (approved):
LGTM; just had two easily-addressable comments about making sure there's only ever one version row in `db_config`.

kantai (approved):
LGTM -- just address the `insert or replace into` feedback before merging.

PR #2906, opened 2021-11-01
Track tx_index for blocks and microblocks transactions

asimm241:
## Description
Track tx index in recept for `micrblocks` and blocks transactions. 

For details:
 - https://github.com/blockstack/stacks-blockchain/issues/2712
 - https://github.com/blockstack/stacks-blockchain-api/issues/662

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

### DIFF
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -4237,17 +4237,20 @@ impl StacksChainState {
         let mut receipts = vec![];
         for microblock in microblocks.iter() {
             debug!("Process microblock {}", &microblock.block_hash());
+            let mut tx_index: u32 = 0;
             for tx in microblock.txs.iter() {
                 let (tx_fee, mut tx_receipt) =
                     StacksChainState::process_transaction(clarity_tx, tx, false)
                         .map_err(|e| (e, microblock.block_hash()))?;
 
                 tx_receipt.microblock_header = Some(microblock.header.clone());
+                tx_receipt.tx_index = tx_index;
                 fees = fees.checked_add(tx_fee as u128).expect("Fee overflow");
                 burns = burns
                     .checked_add(tx_receipt.stx_burned as u128)
                     .expect("Burns overflow");
                 receipts.push(tx_receipt);
+                tx_index += 1;
             }
         }
         Ok((fees, burns, receipts))
@@ -4359,6 +4362,7 @@ impl StacksChainState {
                             contract_analysis: None,
                             execution_cost,
                             microblock_header: None,
+                            tx_index: 0,
                         };
 
                         all_receipts.push(receipt);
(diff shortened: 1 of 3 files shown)


src/chainstate/stacks/db/blocks.rs
```diff
@@ -4031,14 +4036,17 @@ impl StacksChainState {
...
-            let (tx_fee, tx_receipt) =
+            let (tx_fee, mut tx_receipt) =
                 StacksChainState::process_transaction(clarity_tx, tx, false)?;
             fees = fees.checked_add(tx_fee as u128).expect("Fee overflow");
+            tx_receipt.tx_index = tx_index;
             burns = burns
                 .checked_add(tx_receipt.stx_burned as u128)
                 .expect("Burns overflow");
             receipts.push(tx_receipt);
+            tx_index += 1;
```
jcnelson, line 4049: Style nit: instead of having a distinct `mut tx_index` counter, this code should use `for (tx_index, tx) in blocks.txs.iter().enumerate()`

jcnelson: Same goes for `process_microblock_transactions` above.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4031,14 +4036,17 @@ impl StacksChainState {
         let mut fees = 0u128;
         let mut burns = 0u128;
         let mut receipts = vec![];
+        let mut tx_index: u32 = 0;
```
jcnelson, line 4039: The `tx_index` value refers to the index into the epoch at which this transaction occurs.  One epoch is comprised of the microblock stream that a Stacks miner confirms, plus the anchored block transactions it produces.  So, the value of the first `tx_index` in the receipts for an anchored block should be the one plus the `tx_index` of the last microblock transaction it confirms.  It shouldn't start at 0.  cc @zone117x

zone117x (approved):
@jcnelson can you confirm that the latest code in this PR is doing what you described w.r.t anchor block tx index starting _after_ confirmed microblock txs? If so, looks good to me.

src/chainstate/stacks/events.rs
```diff
@@ -61,6 +61,7 @@ pub struct StacksTransactionReceipt {
     pub contract_analysis: Option<ContractAnalysis>,
     pub execution_cost: ExecutionCost,
     pub microblock_header: Option<StacksMicroblockHeader>,
+    pub tx_index: u16,
```
jcnelson, line 64: This needs to be a `u32` at least.  It's possible that in Stacks 2.1, an epoch can have more than 65536 transactions (right now, the max is 11650).  Let's make the field big enough that it's inconceivable that we'll overflow it.

jcnelson:
Logically, it looks fine.  But `tx_index` should be a `u32` at least.  Also, let's get this PR up-to-date with `develop`, which now includes all the code for Stacks 2.05.

asimm241:
> Logically, it looks fine. But `tx_index` should be a `u32` at least.

Done!

> Also, let's get this PR up-to-date with `develop`, which now includes all the code for Stacks 2.05.

Done!

jcnelson (approved):

asimm241:
@pavitthrap is this PR released to mainnet?

Issue #2871, opened 2021-10-12
Hash and tuple_merge functions ignore argument size when computing cost


reedrosenbluth:
The hash functions and `tuple_merge` currently don't take into account the size of their arguments when calculating runtime cost.

PR #2909, opened 2021-11-05
2.05: use serialized_size() aggregation for cost inputs of linear methods

kantai:
This PR implements #2871 by adding a new `Callable` variant for methods whose cost _input_ calculation changes from 2.0 to 2.05. The methods changed are:

-     native_eq
-     index_of
-     tuple_merge
-     hash160
-     sha256
-     sha512
-     sha512trunc256
-     keccak256

This PR also adds a method to Clarity values `.serialized_size()` which returns the size in bytes of the consensus serialization of the Clarity value. This uses a non-recording writer to just count the number of bytes the serialize method would write. The runtime of this method is linear in the size of the Clarity value, but it doesn't involve creating a new buffer. While the runtime of this method is small, because it does depend on the input size, it should only be invoked for functions that already have a linear associated cost.

### DIFF
--- a/src/vm/callables.rs
+++ b/src/vm/callables.rs
@@ -32,9 +32,19 @@ use vm::types::{
 };
 use vm::{eval, Environment, LocalContext, Value};
 
+use super::costs::CostOverflowingMath;
+
 pub enum CallableType {
     UserFunction(DefinedFunction),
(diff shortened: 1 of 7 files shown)


pavitthrap (approved):
LGTM

reedrosenbluth (approved):
LGTM!

Issue #2913, opened 2021-11-08
[chainstate] test that miner and follower have *exactly* the same block costs


jcnelson:
Add an integration test to determine that the execution cost calculated by a miner matches the execution cost that a follower would see.

kantai:
Something interesting!

In building some integration tests for this, it seems like the miner code's cost tracking **is** doing something different than what the `append_block` code is doing while processing microblocks. The miner is more pessimistic -- whereas `append_block` (correctly) resets the cost tracking between the confirmed microblocks (billing them to the microblock miner's block limit) and the beginning of the anchored block, the miner `epoch_begin` logic _doesn't_. This makes it so that the next miner will actually use less of their block limit.

jcnelson:
Wow that's an interesting oversight!  Is this going to be fixed in #2914?

kantai:
It should be, but I'll also open a PR with the new integration test that applies the fix without a refactor, since it should be a pretty straightforward fix.

PR #2916, opened 2021-11-10
2.05 integration testing: exact costs matching in miner and follower: #2913

kantai:
This PR does a few things:

1. Merges `develop` to `next-costs`
2. Adds the confirmed microblock cost and anchored block cost to `new_block` events
3. Adds a new `mined_block` event that relays information about mined, but not necessarily won blocks from a miner.
4. Adds an integration test that asserts exact cost consumption matching between the miner block assembly and the follower block validation paths (#2913)
5. Fixes a bug in the miner block assembly path that led miners to be overly pessimistic in their own block costs when confirming microblocks

### DIFF
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -1663,6 +1663,15 @@ impl StacksBlockBuilder {
 
         let ts_end = get_epoch_time_ms();
 
+        if let Some(observer) = event_observer {
+            observer.mined_block_event(
+                SortitionDB::get_canonical_burn_chain_tip(burn_dbconn.conn())?.block_height + 1,
+                &block,
+                size,
+                &consumed,
+            );
+        }
+
         debug!(
             "Miner: mined anchored block";
             "block_hash" => %block.block_hash(),
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -1136,6 +1136,8 @@ impl StacksBlockBuilder {
     }
 
     /// Begin mining an epoch's transactions.
+    /// Returns an open ClarityTx for mining the block, as well as the ExecutionCost of any confirmed
+    ///  microblocks.
     /// NOTE: even though we don't yet know the block hash, the Clarity VM ensures that a
     /// transaction can't query information about the _current_ block (i.e. information that is not
     /// yet known).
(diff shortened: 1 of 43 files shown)


src/chainstate/stacks/miner.rs
```diff
@@ -1263,7 +1281,18 @@ impl StacksBlockBuilder {
...
             self.set_parent_microblock(&last_mblock_hdr.block_hash(), last_mblock_hdr.sequence);
-        }
+
+            let mut microblock_cost = tx.cost_so_far();
+            microblock_cost
+                .sub(&parent_block_cost)
+                .expect("BUG: block_cost + microblock_cost < block_cost");
+
+            // if we get here, then we need to reset the block-cost back to 0 because this begins the
+            // block defined by this miner.
```
jcnelson, line 1291: lol copypasta

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -59,10 +60,21 @@ const STATUS_RESP_POST_CONDITION: &str = "abort_by_post_condition";
 pub const PATH_MICROBLOCK_SUBMIT: &str = "new_microblocks";
 pub const PATH_MEMPOOL_TX_SUBMIT: &str = "new_mempool_tx";
 pub const PATH_MEMPOOL_TX_DROP: &str = "drop_mempool_tx";
+pub const PATH_MINED_BLOCK: &str = "mined_block";
 pub const PATH_BURN_BLOCK_SUBMIT: &str = "new_burn_block";
 pub const PATH_BLOCK_PROCESSED: &str = "new_block";
 pub const PATH_ATTACHMENT_PROCESSED: &str = "attachments/new";
 
+#[derive(Clone, Serialize, Deserialize)]
+pub struct MinedBlockEvent {
```
jcnelson, line 69: Is this something the sidecar will use, or is this only meant for testing?  Fine either way; just curious.

jcnelson (approved):
LGTM!  Thanks @kantai 

EDIT: can we make sure that this change doesn't break any of the overflow tests?

kantai:
> EDIT: can we make sure that this change doesn't break any of the overflow tests?

Yep -- that's the remaining item here, I need to look at that microblock integration test.

testnet/stacks-node/src/event_dispatcher.rs
kantai: Probably not API operators, but it's something that could be useful down the road for miners (or running a mock miner for analysis) -- event observers could get all sorts of information about their miner this way. Right now, it's really only useful for testing, but I figured it could be made into something more generally useful at some point.

src/chainstate/stacks/miner.rs
kantai: I specifically altered "epoch" to "block"! Is it more confusing? I'm not sure. But having two definitions of epochs is terrifying to me!

jcnelson: No, not a problem at all.  Just happened to notice that almost-the-same comment is in `append_block()` ;)

gregorycoppola (approved):

Issue #2904, opened 2021-10-29
2.05: no microblocks allowed across an epoch boundary


jcnelson:
To simplify a lot of corner cases, make it so that (a) an anchored block cannot have any microblock parents if its parent anchored block is in a different epoch, and (b) the default miner doesn't try to mine any microblocks if it's building an anchored block in a different epoch.

PR #2920, opened 2021-11-10
2.05: make it so a block whose anchored parent is in a different epoch cannot confirm a microblock stream

jcnelson:
This PR implements #2904.  It adds the following three changes:

* It makes it so that when pre-processing an anchored block (such as one downloaded or pushed to the node), the node will not store it if its parent is in a different epoch and it confirms a microblock stream.

* It makes it so the block-processing logic in `append_block()` will reject a block whose parent is in a different epoch and who has a parent microblock.

* It makes it so the anchored block miner will not confirm a microblock stream if the anchored block it produces builds on a parent in a different epoch.

The test coverage here is only in `miner.rs` -- there is no neon integration test.  This is because in order to test this new consensus behavior, we need fine-grained control over how the blocks and microblocks get produced.  Namely, we need to *force* the system to produce an anchored block in epoch 2.05 whose parent is in epoch 2.0 and who confirms a microblock stream in order to test that the consensus rules reject the block.  The neon integration test framework cannot (easily) test this, because it only passes anchored blocks to the relayer, which would simply never store the invalid block per the first change.

### DIFF
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -3104,7 +3104,32 @@ impl StacksChainState {
         return Some((end, None));
     }
 
-    /// Validate an anchored block against the burn chain state.
+    /// Determine whether or not a block is in a different epoch than its parent, given their
+    /// burnchain heights.
+    pub fn block_crosses_epoch_boundary(
+        sortdb_conn: &DBConn,
+        parent_burn_height: u64,
+        child_burn_height: u64,
+    ) -> Result<bool, db_error> {
+        let child_epoch =
+            SortitionDB::get_stacks_epoch(sortdb_conn, child_burn_height)?.expect(&format!(
+                "FATAL: no epoch defined at burn height {}",
+                child_burn_height
+            ));
+
+        let parent_epoch =
+            SortitionDB::get_stacks_epoch(sortdb_conn, parent_burn_height)?.expect(&format!(
+                "FATAL: no epoch defined at burn height {}",
+                parent_burn_height
+            ));
+
+        Ok(child_epoch.epoch_id != parent_epoch.epoch_id)
+    }
+
+    /// Validate an anchored block against the burn chain state.  Determines if this given Stacks
+    /// block can attach to the chainstate.  Called before inserting the block into the staging
+    /// DB.
+    ///
     /// Returns Some(commit burn, total burn) if valid
     /// Returns None if not valid
     /// * consensus_hash is the PoX history hash of the burnchain block whose sortition
@@ -3190,6 +3215,22 @@ impl StacksChainState {
             return Ok(None);
         }
 
+        // NEW in 2.05
+        // if the parent and child epoch differ, then this block cannot descend from a microblock
+        // stream.  Don't store it if this is the case.
+        if StacksChainState::block_crosses_epoch_boundary(
+            db_handle,
+            burn_chain_tip.block_height,
+            block_commit.block_height,
+        )? {
+            if block.has_microblock_parent() {
+                warn!(
+                    "Invalid block, mined in different epoch than parent but confirms microblocks"
+                );
+                return Ok(None);
+            }
+        }
+
         let sortition_burns =
             SortitionDB::get_block_burn_amount(db_handle, &penultimate_sortition_snapshot)
                 .expect("FATAL: have block commit but no total burns in its sortition");
@@ -3204,7 +3245,10 @@ impl StacksChainState {
     /// to the blockchain.  The consensus_hash is the hash of the burnchain block whose sortition
     /// elected the given Stacks block.
     ///
-    /// If we find the same Stacks block in two or more burnchain forks, insert it there too
+    /// If we find the same Stacks block in two or more burnchain forks, insert it there too.
+    ///
+    /// (New in 2.05+) If the anchored block descends from a parent anchored block in a different
+    /// system epoch, then it *must not* have a parent microblock stream.
     ///
     /// sort_ic: an indexed connection to a sortition DB
     /// consensus_hash: this is the consensus hash of the sortition that chose this block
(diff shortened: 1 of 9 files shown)


src/chainstate/stacks/db/blocks.rs
```diff
@@ -3190,6 +3215,22 @@ impl StacksChainState {
...
+            block_commit.block_height,
+        )? {
+            if block.has_microblock_parent() {
+                warn!(
+                    "Invalid block, mined in different epoch than parent but confirms microblocks"
+                );
+                return Ok(None);
+            }
+        }
+
```
kantai, line 3233: This should probably be documented somewhere in this repo (I'd appreciate a suggestion on _where_), but the way that the 2.05 transition (and other epoch transitions) will work is roughly as follows:

When elected burn height is in an epoch later than its parent:
* The Clarity block connection is initiated with the _parent's_ epoch. In this case, 2.0. That means that the loaded cost functions are `costs.clar`, the block limit is the 2.0 block limit, and the inputs used for cost methods are 2.0 inputs (static sizes for db ops, etc.)
* The `append_block` method invokes `StacksChainState::process_epoch_transition`
* This invokes `ClarityBlockConnection::initialize_epoch_2_05`
* This instantiates the `costs-2` contract, and sets the Clarity epoch version in the Clarity DB
* The cost tracker will still use cost functions set at the beginning of this block (i.e., before `costs-2` was instantiated). The Clarity Tx will still use StacksEpoch = 2.00. The block limit will still be set using 2.00

When the parent block is the first block in a new epoch (i.e., the Stacks _parent_ was a block that applied `initialize_epoch_2_05`):
* The cost tracker will use cost functions of the new epoch
* The ClarityTx will use StacksEpoch = new epoch (so the cost input changes will be applied)
* The block limit will be set using the new epoch
* This is actually the behavior of _every_ block (i.e., all blocks load that information from their Stacks parent), but there's only a change from the behavior of the parent when the parent applied an epoch transition.

Perhaps to make this more concrete, imagine the following scenario:

Epoch 2.05 begins at `n`
Stacks Block A: elected at burn block `n-1`
Stacks Block B: elected at burn block `n`, parent is Stacks Block A
Stacks Block C: elected at burn block `n+1`, parent is Stacks Block A
Stacks Block D: elected at burn block `n+2`, parent is Stacks Block B
Stacks Block E: elected at burn block `n+3`, parent is Stacks Block C
Stacks Block F: elected at burn block `n+4`, parent is Stacks Block D

* Stacks Block A is in epoch 2.0, and evaluates exactly like any other block in 2.0
* Stacks Block B discovers that it is in epoch 2.05, but its Clarity DB hasn't had the epoch change applied. It invokes `initialize_epoch_2_05`. Otherwise, the cost functions applied during B are the epoch 2.0 cost functions, the block limit is a epoch 2.0 block limit.
* Stacks Block C discovers that it is in epoch 2.05, but its Clarity DB hasn't had the epoch change applied. It invokes `initialize_epoch_2_05`. Otherwise, the cost functions applied during C are the epoch 2.0 cost functions, the block limit is a epoch 2.0 block limit.
* Stacks Block D is in epoch 2.05, and its ClarityDB has had the epoch change applied, so it does not invoke `initialize_epoch_2_05`. The cost functions and block limit applied are based on the parent's epoch, which was 2.05.
* Stacks Block E is in epoch 2.05, and its ClarityDB has had the epoch change applied, so it does not invoke `initialize_epoch_2_05`. The cost functions and block limit applied are based on the parent's epoch, which was 2.05.
* Stacks Block F is in epoch 2.05, and its ClarityDB has had the epoch change applied, so it does not invoke `initialize_epoch_2_05`. The cost functions and block limit applied are based on the parent's epoch, which was 2.05.


Given all of that -- how do you think microblocks should be handled? Should we disallow microblocks between A->B, A->C? What about B->D and C->E?

jcnelson: Ah, I'd forgotten this.  I thought B and C would have invoked `initialize_epoch_2_05` *and* would use the 2.05 cost functions and block limit?

kantai: Yep -- I agree. So I think this PR needs to be updated to check the parent's parent and the parent, rather than checking the parent and the current.

jcnelson: Okay, this should be addressed now.  I made it so that the chainstate remembers which block(s) encounter an epoch transition, and then requires that any immediate children *not* confirm microblocks.  Then, it's no longer a function of burn block height -- it's solely a function of whether or not the block processed an epoch transition.

src/chainstate/stacks/miner.rs
```diff
@@ -1183,24 +1185,37 @@ impl StacksBlockBuilder {
...
+                &parent_consensus_hash,
+                &parent_header_hash,
+                &parent_index_hash,
+            ) {
+                Ok(x) => x,
+                Err(e) => {
+                    warn!("Miner failed to load parent microblock, mining without parent microblock tail";
+                              "parent_block_hash" => %parent_header_hash,
+                              "parent_index_hash" => %parent_header_hash,
+                              "parent_consensus_hash" => %parent_header_hash,
```
kantai, line 1211: These look like they were wrong before, but might as well correct them here -- I think two of these should be `%parent_consensus_hash` and `%parent_index_hash`.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -3104,12 +3104,36 @@ impl StacksChainState {
...
+            parent_consensus_hash,
+            parent_block_hash,
+        )];
+        let res = block_conn
+            .query_row(sql, args, |_r| Ok(()))
+            .optional()
+            .map(|x| x.is_some())?;
+
+        Ok(res)
+    }
```
kantai, line 3125: Can you add an assertion to one of the `stacks-node::tests::epoch_205` tests that asserts this function returns the expected boolean at the appropriate block?

kantai (approved):
This LGTM, but we should add an assertion to one of the epoch_205 integration tests that checks that the new chainstate database field is being correctly set.

gregorycoppola (approved):

src/chainstate/stacks/miner.rs
jcnelson: Haha wow. Fixed that.

src/chainstate/stacks/db/blocks.rs
jcnelson: Yup, added

kantai (approved):

PR #2923, opened 2021-11-11
Stacks 2.05: Add Epoch to Cost Estimate Key

### LINKED ISSUE #2908: Cost estimations must reset between 2.0 and 2.05 epoch
Cost estimations for contract-call invocations in 2.05 and 2.00 must be reset -- the costs of contract-calls will all change because of the new cost functions, methods of calculating input, etc.

This should be relatively easy to implement -- the cost estimator should just add an epoch identifier to the key used as a cost identifier:

```rust
    fn get_estimate_key(tx: &TransactionPayload, field: &CostField, evaluated_epoch: &StacksEpochId) -> String {
        let tx_descriptor = match tx {
            TransactionPayload::TokenTransfer(..) => "stx-transfer".to_string(),
            TransactionPayload::ContractCall(cc) => {
                let epoch_marker = match evaluated_epoch {
                     StacksEpochId::Epoch10 => "",
                     StacksEpochId::Epoch20 => "",
                     StacksEpochId::Epoch2_05 => ":2.05",
                };
                format!("cc{}:{}.{}", epoch_marker, cc.contract_name, cc.function_name)
            }
            TransactionPayload::SmartContract(_sc) => "contract-publish".to_string(),
            TransactionPayload::PoisonMicroblock(_, _) => "poison-ublock".to_string(),
            TransactionPayload::Coinbase(_) => "coinbase".to_string(),
        };

        format!("{}:{}", &tx_descriptor, field)
    }
```

This will separate the estimates of functions evaluated in 2.05 from those evaluated in 2.00.

gregorycoppola:
Fixes #2908

## Description
We make the cost estimate keys dependent on the epoch.

This effectively resets the cost estimates for Stacks 2.05.

## Testing

New unit tests in `src/cost_estimates/tests/cost_estimators.rs`.

### DIFF
--- a/src/cost_estimates/tests/cost_estimators.rs
+++ b/src/cost_estimates/tests/cost_estimators.rs
@@ -648,3 +648,56 @@ fn test_pessimistic_cost_estimator() {
         }
     );
 }
+
+/// Test that we forget the "learnings" from previous Stacks epoch on next epoch.
+#[test]
+fn test_cost_estimator_forget_previous() {
+    // Setup: Do "notify" in Epoch20.
+    let mut estimator = instantiate_test_db();
+    let block = vec![
+        make_dummy_coinbase_tx(),
+        make_dummy_transfer_tx(),
+        make_dummy_transfer_tx(),
+        make_dummy_cc_tx(
+            "contract-1",
+            "func1",
+            ExecutionCost {
+                write_length: 10,
+                write_count: 10,
+                read_length: 10,
+                read_count: 10,
+                runtime: 10,
+            },
+        ),
+    ];
+    estimator.notify_block(&block, &BLOCK_LIMIT_MAINNET_20, &StacksEpochId::Epoch20);
+
+    // Test 1: We should get *non-zero* estimates back when we test in Epoch20.
+    assert_eq!(
+        estimator
+            .estimate_cost(
+                &make_dummy_cc_payload("contract-1", "func1"),
+                &StacksEpochId::Epoch20
+            )
+            .expect("Should be able to provide cost estimate now"),
+        ExecutionCost {
+            write_length: 10,
+            write_count: 10,
+            read_length: 10,
+            read_count: 10,
+            runtime: 10,
+        }
+    );
+
+    // Test 2: We should get *zero* estimates back when we test in Epoch20.
+    assert_eq!(
+        error,
+        estimator
+            .estimate_cost(
+                &make_dummy_cc_payload("contract-1", "func1"),
+                &StacksEpochId::Epoch2_05
+            )
+            .unwrap_err(),
+        EstimatorError::NoEstimateAvailable
+    );
+}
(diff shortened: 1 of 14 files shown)


src/cost_estimates/tests/cost_estimators.rs
```diff
@@ -589,3 +648,56 @@ fn test_pessimistic_cost_estimator() {
...
+        ExecutionCost {
+            write_length: 10,
+            write_count: 10,
+            read_length: 10,
+            read_count: 10,
+            runtime: 10,
+        }
+    );
+
+    // Test 2: We should get *zero* estimates back when we test in Epoch20.
```
kantai, line 692: Comment typo -- Epoch20 -> Epoch2_05

src/cost_estimates/tests/cost_estimators.rs
```diff
@@ -589,3 +648,56 @@ fn test_pessimistic_cost_estimator() {
...
+    assert_eq!(
+        error,
+        estimator
+            .estimate_cost(
+                &make_dummy_cc_payload("contract-1", "func1"),
+                &StacksEpochId::Epoch2_05
+            )
+            .unwrap_err(),
+        EstimatorError::NoEstimateAvailable
+    );
```
kantai, line 702: Can you expand this test so that it sends some measurements for `Epoch2_05` for `contract-1.func1`, and assert that you get an estimate back that's different from the `Epoch20` measurement (which should stay the same)?

src/cost_estimates/tests/cost_estimators.rs
gregorycoppola: done

gregorycoppola (commented):
thanks.

kantai (approved):
This LGTM -- just had one comment about expanding the unit test.

src/cost_estimates/tests/cost_estimators.rs
gregorycoppola: added two new tests

src/net/rpc.rs
```diff
@@ -1609,8 +1609,19 @@ impl ConversationHttp {
         estimated_len: u64,
     ) -> Result<(), net_error> {
         let response_metadata = HttpResponseMetadata::from(req);
+        let tip = SortitionDB::get_canonical_burn_chain_tip(sortdb.conn())?;
+        let stacks_epoch = sortdb
```
jcnelson, line 1613: Can't this just be `SortitionDB::get_stacks_epoch(tip.block_height)`?  Why involve an indexed connection, which potentially has a MARF read?

jcnelson (approved):
LGTM; just please change the call to `get_stacks_epoch()` in the fee estimator endpoint to avoid using an index connection.  This network code path is a hot path, so if we can keep it fast, we should.

gregorycoppola:
thanks for the reviews everybody

Issue #2894, opened 2021-10-20
Cost updates: add new costs contract


pavitthrap:
Tasks 
- Add contract 
- Must determine what to name the file. 

In #2873, I named the contract "costs-2", but "costs-2.05" might have more consensus on the team.

Issue #2890, opened 2021-10-20
Cost updates: Update function instantiate_epoch_2_05 once costs-2 contract is added


pavitthrap:
Update the code body for the costs-2 contract in the function `instantiate_epoch_2_05` (in `src/clarity_vm/clarity.rs`) once the contract is added to the repo.

jcnelson:
Still todo:

* ~Update the way the benchmark runs for the AST (ClarityString vs Value)~
* NFT burn must have the same cost as NFT transfer (same for FTs and STX)
    * Functions that should have the same cost must have the same cost

PR #2924, opened 2021-11-12
Instantiate `costs-2.clar` with latest results from benchmarking

kantai:
This PR implements #2894 and #2890 -- using the cost contract from the latest results in the `clarity-benchmarking` repository.

SIP-012 hasn't been updated with these costs yet, but will be.

This PR expands the costs unit tests to instantiate and use the costs-2 contract for cost tracking.

### DIFF
--- a/src/chainstate/stacks/boot/costs-2.clar
+++ b/src/chainstate/stacks/boot/costs-2.clar
@@ -1,1 +1,567 @@
 ;; the .costs-2 contract
+
+;; Helper Functions
+
+;; Return a Cost Specification with just a runtime cost
+(define-private (runtime (r uint))
+    {
+        runtime: r,
+        write_length: u0,
+        write_count: u0,
+        read_count: u0,
+        read_length: u0,
+    })
+
+;; Linear cost-assessment function
+(define-private (linear (n uint) (a uint) (b uint))
+    (+ (* a n) b))
+
+;; LogN cost-assessment function
+(define-private (logn (n uint) (a uint) (b uint))
+    (+ (* a (log2 n)) b))
+
+;; NLogN cost-assessment function
+(define-private (nlogn (n uint) (a uint) (b uint))
+    (+ (* a (* n (log2 n))) b))
+
+
+;; Cost Functions
+(define-read-only (cost_analysis_type_annotate (n uint))
+    (runtime (linear n u1 u9)))
+
+(define-read-only (cost_analysis_type_check (n uint))
+    (runtime (linear n u113 u1)))
+
+(define-read-only (cost_analysis_type_lookup (n uint))
+    (runtime (linear n u1 u6)))
+
+(define-read-only (cost_analysis_visit (n uint))
+    (runtime u1))
+
+(define-read-only (cost_analysis_iterable_func (n uint))
+    (runtime (linear n u2 u14)))
+
+(define-read-only (cost_analysis_option_cons (n uint))
+    (runtime u6))
+
+(define-read-only (cost_analysis_option_check (n uint))
+    (runtime u3))
+
+(define-read-only (cost_analysis_bind_name (n uint))
+    (runtime (linear n u2 u176)))
+
+(define-read-only (cost_analysis_list_items_check (n uint))
+    (runtime (linear n u2 u4)))
+
+(define-read-only (cost_analysis_check_tuple_get (n uint))
+    (runtime (logn n u1 u2)))
+
+(define-read-only (cost_analysis_check_tuple_merge (n uint))
+    (runtime (linear n u1000 u1000)))
+
+(define-read-only (cost_analysis_check_tuple_cons (n uint))
+    (runtime (nlogn n u3 u5)))
+
+(define-read-only (cost_analysis_tuple_items_check (n uint))
+    (runtime (linear n u1 u59)))
+
+(define-read-only (cost_analysis_check_let (n uint))
+    (runtime (linear n u1 u12)))
+
+(define-read-only (cost_analysis_lookup_function (n uint))
+    (runtime u20))
+
+(define-read-only (cost_analysis_lookup_function_types (n uint))
+    (runtime (linear n u1 u28)))
+
+(define-read-only (cost_analysis_lookup_variable_const (n uint))
+    (runtime u15))
+
+(define-read-only (cost_analysis_lookup_variable_depth (n uint))
+    (runtime (nlogn n u1 u34)))
+
+(define-read-only (cost_ast_parse (n uint))
+    (runtime (linear n u172 u287441)))
+
+(define-read-only (cost_ast_cycle_detection (n uint))
+    (runtime (linear n u141 u72)))
+
+(define-read-only (cost_analysis_storage (n uint))
+    {
+        runtime: (linear n u2 u100),
+        write_length: (linear n u1 u1),
+        write_count: u1,
+        read_count: u1,
+        read_length: u1
+    })
+
+(define-read-only (cost_analysis_use_trait_entry (n uint))
+    {
+        runtime: (linear n u9 u723),
+        write_length: (linear n u1 u1),
+        write_count: u0,
+        read_count: u1,
+        read_length: (linear n u1 u1)
+    })
+
+
+(define-read-only (cost_analysis_get_function_entry (n uint))
+    {
+        runtime: (linear n u81 u1303),
+        write_length: u0,
+        write_count: u0,
+        read_count: u1,
+        read_length: (linear n u1 u1)
+    })
+
+
+(define-read-only (cost_analysis_fetch_contract_entry (n uint))
+    {
+        runtime: (linear n u1000 u1000),
+        write_length: u0,
+        write_count: u0,
+        read_count: u1,
+        read_length: (linear n u1 u1)
+    })
+
+(define-read-only (cost_lookup_variable_depth (n uint))
+    (runtime (linear n u2 u14)))
+
+(define-read-only (cost_lookup_variable_size (n uint))
+    (runtime (linear n u2 u1)))
+
+(define-read-only (cost_lookup_function (n uint))
+    (runtime u16))
+
+(define-read-only (cost_bind_name (n uint))
+    (runtime u256))
+
+(define-read-only (cost_inner_type_check_cost (n uint))
+    (runtime (linear n u2 u9)))
+
+(define-read-only (cost_user_function_application (n uint))
+    (runtime (linear n u26 u140)))
+
+(define-read-only (cost_let (n uint))
+    (runtime (linear n u146 u862)))
+
+(define-read-only (cost_if (n uint))
+    (runtime u200))
+
+(define-read-only (cost_asserts (n uint))
+    (runtime u158))
+
+(define-read-only (cost_map (n uint))
+    (runtime (linear n u1210 u3314)))
+
+(define-read-only (cost_filter (n uint))
+    (runtime u460))
+
+(define-read-only (cost_len (n uint))
+    (runtime u486))
+
+(define-read-only (cost_element_at (n uint))
+    (runtime u619))
+
+(define-read-only (cost_index_of (n uint))
+    (runtime (linear n u1 u243)))
+
+(define-read-only (cost_fold (n uint))
+    (runtime u483))
+
+(define-read-only (cost_list_cons (n uint))
+    (runtime (linear n u14 u198)))
+
+(define-read-only (cost_type_parse_step (n uint))
+    (runtime u5))
+
+(define-read-only (cost_tuple_get (n uint))
+    (runtime (nlogn n u4 u1780)))
+
+(define-read-only (cost_tuple_merge (n uint))
+    (runtime (linear n u4 u646)))
+
+(define-read-only (cost_tuple_cons (n uint))
+    (runtime (nlogn n u11 u1101)))
+
+(define-read-only (cost_add (n uint))
+    (runtime (linear n u12 u156)))
+
+(define-read-only (cost_sub (n uint))
+    (runtime (linear n u12 u156)))
+
+(define-read-only (cost_mul (n uint))
+    (runtime (linear n u14 u157)))
+
+(define-read-only (cost_div (n uint))
+    (runtime (linear n u14 u157)))
+
+(define-read-only (cost_geq (n uint))
+    (runtime u166))
+
+(define-read-only (cost_leq (n uint))
+    (runtime u166))
+
+(define-read-only (cost_le (n uint))
+    (runtime u166))
+
+(define-read-only (cost_ge (n uint))
+    (runtime u166))
+
+(define-read-only (cost_int_cast (n uint))
+    (runtime u164))
+
+(define-read-only (cost_mod (n uint))
+    (runtime u168))
+
+(define-read-only (cost_pow (n uint))
+    (runtime u170))
+
+(define-read-only (cost_sqrti (n uint))
+    (runtime u167))
+
+(define-read-only (cost_log2 (n uint))
+    (runtime u161))
+
+(define-read-only (cost_xor (n uint))
+    (runtime u167))
+
+(define-read-only (cost_not (n uint))
+    (runtime u162))
+
+(define-read-only (cost_eq (n uint))
+    (runtime (linear n u7 u172)))
+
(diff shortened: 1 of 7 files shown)


src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -1 +1,567 @@
-;; the .costs-2 contract
\ No newline at end of file
+;; the .costs-2 contract
```
jcnelson, line 1: Can you also put this into the SIP 012 text?  Thanks!

src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -1 +1,567 @@
...
+    {
+        runtime: (linear n u1 u1539),
+        write_length: u0,
+        write_count: u0,
+        read_count: u1,
+        read_length: (linear n u1 u1)
+    })
+
+
+(define-read-only (cost_set_entry (n uint))
```
jcnelson, line 400: This is the same for deletions, right?

src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -1 +1,567 @@
...
+    {
+        runtime: u6321,
+        write_length: u0,
+        write_count: u0,
+        read_count: u1,
+        read_length: u1
+    })
+
+
+(define-read-only (cost_stx_balance (n uint))
```
jcnelson, line 450: I'm surprised that getting the STX balance and transferring STX is much more expensive (2x) than doing so for a fungible token.  Do we know why this is?

src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -1 +1,567 @@
...
+        write_length: u1,
+        write_count: u1,
+        read_count: u1,
+        read_length: u1
+    })
+
+
+(define-read-only (poison_microblock (n uint))
+    {
+        runtime: u29568,
```
jcnelson, line 562: Can we make this as close to free as possible?  I think we want people reporting poison microblocks even if there's congestion, and we always want the payoff for reporting a poison microblock to be positive.

src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -1 +1,567 @@
...
+        write_length: u1,
+        write_count: u1,
+        read_count: u1,
+        read_length: u1
+    })
+
+
+(define-read-only (cost_nft_transfer (n uint))
+    {
+        runtime: (linear n u9 u647),
```
jcnelson, line 512: I notice that this has a lower runtime than `nft-get-owner`, which feels wrong to me.  Do we know why?

src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -1 +1,567 @@
...
+        write_length: u1,
+        write_count: u1,
+        read_count: u1,
+        read_length: u1
+    })
+
+
+(define-read-only (cost_ft_mint (n uint))
+    {
+        runtime: u1645,
```
jcnelson, line 472: Interesting.  The higher runtime for minting implies that efficient contracts that don't care about `ft-get-supply` might just mint all the tokens to themselves, and then `ft-transfer?` them to owners instead of minting them on the fly.

src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -1 +1,567 @@
...
+    (runtime u460))
+
+(define-read-only (cost_len (n uint))
+    (runtime u486))
+
+(define-read-only (cost_element_at (n uint))
+    (runtime u619))
+
+(define-read-only (cost_index_of (n uint))
+    (runtime (linear n u1 u243)))
```
jcnelson, line 167: I know that most Clarity code that uses the VRF or does some kind of parsing makes heavy use of `index-of` to convert 1-byte buffers into uints.  Since proper slicing and conversion methods won't be available until 2.1, is there a world in which we just declare that this function has O(1) cost, since lists are finite length anyway and we could always implement lists internally as hashmaps or some such if it became a problem?

src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -1 +1,567 @@
...
+    (runtime u158))
+
+(define-read-only (cost_map (n uint))
+    (runtime (linear n u1210 u3314)))
+
+(define-read-only (cost_filter (n uint))
+    (runtime u460))
+
+(define-read-only (cost_len (n uint))
+    (runtime u486))
```
LNow, line 161: How come this one is cheaper than `filter`?

src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -1 +1,567 @@
...
+    (runtime u167))
+
+(define-read-only (cost_log2 (n uint))
+    (runtime u161))
+
+(define-read-only (cost_xor (n uint))
+    (runtime u167))
+
+(define-read-only (cost_not (n uint))
+    (runtime u162))
```
LNow, line 230: Shouldn't this one be cheaper than `log2` ?

src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -1 +1,567 @@
...
+    (runtime u287))
+
+(define-read-only (cost_is_none (n uint))
+    (runtime u212))
+
+(define-read-only (cost_is_err (n uint))
+    (runtime u282))
+
+(define-read-only (cost_is_some (n uint))
+    (runtime u223))
```
LNow, line 290: I think `is_ok` should have same cost as `is_err`, and `is_none` same as `is_none`.

src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -1 +1,567 @@
...
+    (runtime (linear n u75 u244)))
+
+(define-read-only (cost_as_max_len (n uint))
+    (runtime u475))
+
+(define-read-only (cost_contract_call (n uint))
+    (runtime u153))
+
+(define-read-only (cost_contract_of (n uint))
+    (runtime u13400))
```
LNow, line 323: Why `contract_of` is so expensive? It only unwraps principal from trait reference.

src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -1 +1,567 @@
...
+
+(define-read-only (cost_principal_of (n uint))
+    (runtime u39))
```
LNow, line 326: This one is a puzzle for me.
It calculates principal derived from the public key. Computational wise it is far more complex than many other functions yet it looks like it is the cheapest one.

src/chainstate/stacks/boot/costs-2.clar
kantai: Yep -- though deletion costs will almost always be cheaper, because the input is just the key size, not the value size as well.

src/chainstate/stacks/boot/costs-2.clar
kantai: There is more work being done for STX balances -- fetching the current burnchain height and applying any necessary checks for Stacking unlocks to get the current account balance. The difference in runtime of ~700 isn't all that much: that's equivalent to ~4 comparison operations. So this isn't terribly surprising to me given that fetching STX balances requires the additional work on the unlock checks.

src/chainstate/stacks/boot/costs-2.clar
kantai: Sure -- What would you suggest for the runtime cost? `u1` ?

src/chainstate/stacks/boot/costs-2.clar
kantai: Not really -- we can set the NFT operations to be matched like the comparison operations, though, such that NFT transfer, mint, and get owner are the same cost.

src/chainstate/stacks/boot/costs-2.clar
kantai: The benefit would be _very_ marginal -- these operations all incur 2 MARF reads and writes, each of which is 2/15k of the block limit (1.3*10^-4), whereas the savings in runtime is 1/5m (2*10^-7) -- it would have very little impact on the cost of the transaction to perform a transfer instead of a mint.

src/chainstate/stacks/boot/costs-2.clar
kantai: I'm not sure what the benefit would be of doing so -- is it to just make this operation cheaper than we're measuring?

src/chainstate/stacks/boot/costs-2.clar
kantai: `filter` and `len` are ultimately doing pretty dissimilar things, I'm not sure that I can say definitively one way or another which should be cheaper. The `filter` cost is the cost of the overhead of the `filter` function: checking that its inputs are the appropriate type, dissassembling the list struct to apply the filter function and reassemble an output list. It basically measures everything but the individual invocations of filter. `len` does a lot of the same work, but instead of dissassembling the list struct, it is doing function invocation on the object to get a length. I'm not sure exactly what the difference in runtime would come from here, but I'm not surprised that there is a difference.

src/chainstate/stacks/boot/costs-2.clar
kantai: I'm not sure -- `log2` is a pretty fast operation on integers: just a bitshift, `not` is fast too, but I'm not sure its easy to say which should be faster.

src/chainstate/stacks/boot/costs-2.clar
kantai: Yep -- fixed.

src/chainstate/stacks/boot/costs-2.clar
kantai: It also checks that the contract exists -- this is where most of the cost is.

src/chainstate/stacks/boot/costs-2.clar
LNow: Thanks for the explanation. 
I thought that cost of `filter` is the cost of all filter operation including individual function invocation on every single element not just the overhead of the `filter`.

src/chainstate/stacks/boot/costs-2.clar
kantai: Yeah -- I think this may require looking more into the structure of the benchmark. Ultimately, this isn't _that_ complex of an operation: calculating a principal is just a hash160 of the inputted public key (and checking that the public key is valid) -- but this measurement is about 10x cheaper than the hash160 operation, so we should look into it.

src/chainstate/stacks/boot/costs-2.clar
LNow: `not` inverses boolean value so it changes `true` to `false` or the other way around. It operates on 1bit.
It should be the cheapest or one of the cheapest functions that we have in Clarity.

src/chainstate/stacks/boot/costs-2.clar
LNow: I think this check is done a little bit earlier.

With function like this:
```clojure
(define-read-only (test (argument <my-trait>))
  (contract-of argument)
)
```
I can't pass to it address of a contract that do not exists. It must be already deployed and it must be compliant with `my-trait`.
And there is no other way to pass trait reference to `contract-of` other than defining it first as function argument (like in the example above).

src/chainstate/stacks/boot/costs-2.clar
reedrosenbluth: working on it!

jcnelson:
Let's be sure to also review the cost contract instantiation code path. Let's not block on the cost functions themselves -- we need to get this migration code tested on testnet ASAP.

src/chainstate/stacks/boot/costs-2.clar
jcnelson: Yeah, I think that would be fine as long as the work they all do internally is about the same.

src/chainstate/stacks/boot/costs-2.clar
jcnelson: On second thought, I think this is fine as-is.  Looking at the code, there's currently no penalty to submitting a poison-microblock transaction for a fork that occurs later in the stream than the highest-reported fork, so making these too cheap could open a DoS vector.  We can revisit whether or not gratuitous submissions should be penalized (i.e. cause the block to be invalid) in 2.1 if we want.

src/vm/analysis/tests/costs.rs
```diff
@@ -90,6 +91,21 @@ pub fn test_tracked_costs(prog: &str) -> ExecutionCost {
...
+        conn.commit_block();
+    }
+
+    {
+        let mut conn = clarity_instance.begin_block(
+            &StacksBlockId([1 as u8; 32]),
+            &StacksBlockId([2 as u8; 32]),
+            &TEST_HEADER_DB,
+            &TEST_BURN_STATE_DB,
+        );
```
jcnelson, line 108: Do you want to add an assertion here to verify that this block is indeed executing in the proper epoch?

src/vm/tests/mod.rs
```diff
@@ -103,7 +103,7 @@ impl HeadersDB for UnitTestHeaderDB {
         {
             Some(BITCOIN_REGTEST_FIRST_BLOCK_TIMESTAMP as u64)
         } else {
-            None
+            Some(1 + 10 * (id_bhh.as_bytes()[0] as u64))
```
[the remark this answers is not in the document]
kantai: Yeah -- that's fine for the unit tests

src/vm/analysis/tests/costs.rs
kantai: Added in https://github.com/blockstack/stacks-blockchain/pull/2924/commits/4972224e49852c4739ac56dd1a5a0c4f34f4221a

src/vm/tests/mod.rs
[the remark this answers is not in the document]
gregorycoppola: How did you pick this value? Can you answer in a brief comment?

src/vm/tests/costs.rs
```diff
@@ -651,68 +666,52 @@ fn test_tracked_costs(prog: &str) -> ExecutionCost {
...
 
-    owned_env
-        .initialize_contract(trait_contract_id.clone(), contract_trait)
-        .unwrap();
-    owned_env
-        .initialize_contract(other_contract_id.clone(), contract_other)
-        .unwrap();
-    owned_env
-        .initialize_contract(self_contract_id.clone(), &contract_self)
+        eprintln!("{}", &contract_self);
```
gregorycoppola, line 682: do you mean to check these `eprintln` lines in? just checking

src/vm/tests/costs.rs
```diff
@@ -651,68 +666,52 @@ fn test_tracked_costs(prog: &str) -> ExecutionCost {
...
+        let test = get_simple_test(f);
+        let cost = test_tracked_costs(test, StacksEpochId::Epoch20);
+        assert!(cost.exceeds(&baseline));
+    }
 }
 
 #[test]
-fn test_all() {
-    let baseline = test_tracked_costs("1");
+fn epoch_205_test_all() {
```
gregorycoppola, line 709: clarification: What is this testing? What remains to be tested?

gregorycoppola (commented):
A few small comments/questions.

src/vm/tests/costs.rs
kantai: Yeah -- `eprintln`s are useful in test code (which this line is in a tests module), because they get captured by the test runner and only display if the test fails.

src/vm/tests/costs.rs
kantai: It tests each individual method's cost function invocation

kantai: (Added a comment in 2103e9f96)

src/vm/tests/mod.rs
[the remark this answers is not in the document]
kantai: Added a comment in 2103e9f96

gregorycoppola (approved):

kantai:
After merging, `next-costs` and the SIP-012 proposal will have identical `costs-2.clar` content -- any further discussion of the proposed costs should occur in the proposal, and then any decided changes can propagate back over to the stacks-blockchain repo.

PR #2925, opened 2021-11-12
Stacks 2.05: make p2p network reject connections from 2.0 peers after the epoch switch

jcnelson:
This PR repurposes the lowest byte in the `peer_version` field in the `Preamble` struct so that p2p messages can inform remote peers of their _maximum supported_ epoch.  For 2.0, this field is ignored, but the value in all 2.0 nodes right now is `0x00` on mainnet and `0x01` on testnet.  For 2.05 and up, this value must be `0x05` or later once the 2.05 epoch activates (on both mainnet and testnet).  Before the epoch activates, a 2.05 peer will accept connections from either a 2.0 peer or a 2.05 peer.

This PR alters the `is_preamble_valid()` method in `ConversationP2P` to enforce the above constraints. If the remote peer's epoch marker in the `peer_version` field is greater than or equal to the local node's epoch marker, then the connection is accepted.  If it isn't, then as long as the remote peer's epoch marker is greater than or equal to the _current_ epoch's epoch marker, then the connection will be accepted.  This way, a 2.0 node can talk to a 2.05 node during the 2.0 epoch, and while the 2.05 node is bootstrapping, since `0x00 <= 0x05` (and `0x01 <= 0x05` on testnet).  However, once the 2.05 node passes the 2.05 Bitcoin block, it will reject subsequent messages from the 2.0 nodes.

This PR expands the `convo_is_preamble_valid()` test in `src/net/chat.rs` to cover the new code paths.  In addition, this PR adds `test_step_walk_1_neighbor_bad_epoch()` to `src/net/neighbors.rs` to verify that two peers that believe they run in different epochs will never talk to each other.

### DIFF
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -429,12 +429,15 @@ impl FromRow<StacksEpoch> for StacksEpoch {
         let start_height = u64::from_column(row, "start_block_height")?;
         let end_height = u64::from_column(row, "end_block_height")?;
 
+        let network_epoch: u8 = row.get_unwrap("network_epoch");
+
         let block_limit = row.get_unwrap("block_limit");
         Ok(StacksEpoch {
             epoch_id,
             start_height,
             end_height,
             block_limit,
+            network_epoch,
         })
     }
 }
@@ -609,6 +612,7 @@ const SORTITION_DB_SCHEMA_2: &'static [&'static str] = &[r#"
          end_block_height INTEGER NOT NULL,
          epoch_id INTEGER NOT NULL,
          block_limit TEXT NOT NULL,
+         network_epoch INTEGER NOT NULL,
          PRIMARY KEY(start_block_height,epoch_id)
      );"#];
 
@@ -2240,7 +2244,7 @@ impl SortitionDB {
         first_burn_header_timestamp: u64,
         epochs_ref: &[StacksEpoch],
     ) -> Result<(), db_error> {
-        debug!("Instantiate SortDB");
+        debug!("Instantiate sortition DB");
 
         sql_pragma(self.conn(), "PRAGMA journal_mode = WAL;")?;
 
@@ -2274,9 +2278,10 @@ impl SortitionDB {
                 &u64_to_sql(epoch.start_height)?,
                 &u64_to_sql(epoch.end_height)?,
                 &epoch.block_limit,
+                &epoch.network_epoch,
             ];
             db_tx.execute(
-                "INSERT INTO epochs (epoch_id,start_block_height,end_block_height,block_limit) VALUES (?1,?2,?3,?4)",
+                "INSERT INTO epochs (epoch_id,start_block_height,end_block_height,block_limit,network_epoch) VALUES (?1,?2,?3,?4,?5)",
                 args
             )?;
         }
@@ -3464,6 +3469,12 @@ impl SortitionDB {
             &u64_to_sql(burn_block_height)?,
         ];
         query_row(conn, sql, args)
+    }
+
+    /// Get all StacksEpochs
+    pub fn get_stacks_epochs(conn: &DBConn) -> Result<Vec<StacksEpoch>, db_error> {
+        let sql = "SELECT * FROM epochs ORDER BY start_block_height ASC";
+        query_rows(conn, sql, NO_PARAMS)
     }
 
     pub fn get_stacks_epoch_by_epoch_id(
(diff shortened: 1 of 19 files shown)


src/net/chat.rs
```diff
@@ -641,16 +647,63 @@ impl ConversationP2P {
...
+                my_epoch
+            );
+            return true;
+        }
+
+        test_debug!(
+            "Remote peer has old network version {} (epoch {})",
+            remote_peer_version,
+            remote_epoch
+        );
```
kantai, line 673: Would these `test_debug`s be useful as `debug`s?

kantai: Nevermind -- they probably won't because there's a `debug` after this returns.

kantai (approved):
LGTM!

src/chainstate/burn/db/sortdb.rs
```diff
@@ -3467,6 +3472,12 @@ impl SortitionDB {
         query_row(conn, sql, args)
     }
 
+    /// Get all StacksEpochs
```
gregorycoppola, line 3475: "Get all StacksEpochs, in ascending order of start block height."

src/core/mod.rs
```diff
@@ -219,9 +226,19 @@ pub struct StacksEpoch {
     pub start_height: u64,
     pub end_height: u64,
     pub block_limit: ExecutionCost,
+    pub network_epoch: u8,
 }
 
 impl StacksEpoch {
+    pub fn find_epoch(epochs: &[StacksEpoch], height: u64) -> Option<usize> {
```
gregorycoppola, line 233: Suggested contract: ```
/// Returns the epoch in `epochs` that contains burn block height `height`, or `None` if none found.
///
/// Note: Does a linear search through `epochs`.
```

File: src/net/neighbors.rs [hunk 4]
```diff
@@ -3390,6 +3394,103 @@ mod test {
...
+            let mut peer_1_config = TestPeerConfig::from_port(31998);
+            let mut peer_2_config = TestPeerConfig::from_port(31990);
+
+            peer_1_config.connection_opts.walk_retry_count = 10;
+            peer_2_config.connection_opts.walk_retry_count = 10;
+            peer_1_config.connection_opts.walk_interval = 1;
+            peer_2_config.connection_opts.walk_interval = 1;
+
+            // peer 1 thinks its always epoch 2.0
+            peer_1_config.peer_version = 0x18000000;
```
gregorycoppola, line 3410: what happens if the 2.0 node hasn't upgraded? Will this be 0x0?

src/net/neighbors.rs
```diff
@@ -3390,6 +3394,103 @@ mod test {
...
+            peer_1_config.epochs = Some(vec![StacksEpoch {
```
gregorycoppola, line 3411: What kind of testing do we need to do? or can we do? in order to fully understand this feature?

gregorycoppola (commented):
Just a few superficial questions, and wondering about testing.

src/net/chat.rs
```diff
@@ -641,16 +647,63 @@ impl ConversationP2P {
         false
     }
 
+    /// Determine whether or not a remote node has the proper epoch marker in its peer version
+    /// * If the local and remote nodes are in the same system epoch, then yes
+    /// * If they're in different epochs, but the epoch shift hasn't happened yet, then yes
+    /// * Otherwise, no
+    fn has_acceptable_epoch(&self, cur_burn_height: u64, remote_peer_version: u32) -> bool {
+        // which epochs do I support, and which epochs does the remote peer support?
+        let my_epoch = (self.version & 0x000000ff) as u8;
```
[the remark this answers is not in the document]
jcnelson: Making it a function would require _more_ text than just taking a literal bitmask.

src/core/mod.rs
```diff
@@ -219,9 +226,19 @@ pub struct StacksEpoch {
     pub start_height: u64,
     pub end_height: u64,
     pub block_limit: ExecutionCost,
+    pub network_epoch: u8,
```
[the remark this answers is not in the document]
jcnelson: `epoch_id` and `network_epoch` are not the same values. Also, an epoch can change but the network epoch can stay the same.

src/net/neighbors.rs
jcnelson: The unit tests added here should be sufficient.  If you want, you can leave a 2.0 testnet node running and verify that it loses all its neighbor connections on the epoch transition

src/net/neighbors.rs
jcnelson: The peer version `0x18000000` *is* the 2.0 peer version.  The 2.05 one is `0x18000005`.

src/net/chat.rs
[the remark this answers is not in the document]
gregorycoppola: It's not to save text but to avoid code duplication and make the code self-explanatory.

src/net/neighbors.rs
gregorycoppola: I can walk through the test if there is a possible one. Like if we can construct a test with the private testnet or something.

gregorycoppola (approved):
suggested changes are optional.. please follow up on testing if there is some test I can set up for this.

Issue #2889, opened 2021-10-20
2.05 Testing: Perform 2.05 test fork on xenon / testnet


pavitthrap:
Need to define `BITCOIN_TESTNET_2_0_LAST_BLOCK_HEIGHT` and update `STACKS_EPOCHS_TESTNET` in `src/core/mod.rs`.

jcnelson:
Let's try and get this done by this Thursday at the latest (ideally Wednesday)

kantai:
To recap the action items before a launch on testnet can occur:

- [ ] Review and Merge #2925 -- requires two approvals (pinging @kantai and @gregorycoppola)
- [ ] Review and Merge #2924 -- requires two approvals (pinging @jcnelson and @gregorycoppola)
- [ ] Review and Merge #2923 -- requires one more approval (pinging @jcnelson)
- [ ] Merge #2920 -- approved, should be ready to merge (pinging @jcnelson)
- [ ] Deploy `next-costs` branch to a testnet follower, confirm that it can follow the current testnet, reuse existing chainstate. This can be tested before the above PRs merge.
- [ ] Pick testnet block height, (proposal: 2,104,364 = 48 * 6 + 2104076 ~= 48 hours from now) and open configuration PR (@kantai)
- [ ] Review and Merge configuration PR
- [ ] Tag a next-costs build and deploy to testnet nodes

kantai:
Testnet fork activated! 

https://explorer.stacks.co/txid/0xef13fb69ad25f49400fad61d82fb678cb59f8e584acd70450acd84b3e49ef706?chain=testnet

PR #2928, opened 2021-11-15
chore: select 2.05 xenon testnet start height, dead code elimination

kantai:
This picks a 2.05 xenon/testnet start height, and does some dead code elimination

This implements #2889


src/core/mod.rs
```diff
@@ -364,9 +343,15 @@ lazy_static! {
         StacksEpoch {
             epoch_id: StacksEpochId::Epoch20,
             start_height: BITCOIN_MAINNET_FIRST_BLOCK_HEIGHT,
-            end_height: STACKS_2_0_LAST_BLOCK_TO_PROCESS + 1,
+            end_height: BITCOIN_TESTNET_STACKS_2_05_BURN_HEIGHT,
```
kantai: Note that the prior value of end_height here would have made the `next-costs` branch broken when running on Xenon testnet: the end_height is long before the first bitcoin block height that the testnet considers

jcnelson (approved):
LGTM!  Do you also want to set the mainnet block height for 2.05 here in this PR, or do you want that to be separate?

kantai:
> LGTM! Do you also want to set the mainnet block height for 2.05 here in this PR, or do you want that to be separate?

Done!

jcnelson (approved):

gregorycoppola (commented):
@kantai can we get a clean change for the new height? And do the dead code elimination in separate PR? For easier review and possible future anthropology?

gregorycoppola:
Oh i see Jude already lgtmed.. nevermind the idea of splitting in up if it's more overhead @kantai

gregorycoppola (approved):

src/core/mod.rs
```diff
@@ -91,11 +88,13 @@ pub const BITCOIN_MAINNET_FIRST_BLOCK_TIMESTAMP: u32 = 1610643248;
 pub const BITCOIN_MAINNET_FIRST_BLOCK_HASH: &str =
     "0000000000000000000ab248c8e35c574514d052a83dbc12669e19bc43df486e";
 pub const BITCOIN_MAINNET_INITIAL_REWARD_START_BLOCK: u64 = 651389;
+pub const BITCOIN_MAINNET_STACKS_2_05_BURN_HEIGHT: u64 = 713_000;
 
 pub const BITCOIN_TESTNET_FIRST_BLOCK_HEIGHT: u64 = 2000000;
 pub const BITCOIN_TESTNET_FIRST_BLOCK_TIMESTAMP: u32 = 1622691840;
 pub const BITCOIN_TESTNET_FIRST_BLOCK_HASH: &str =
     "000000000000010dd0863ec3d7a0bae17c1957ae1de9cbcdae8e77aad33e3b8c";
+pub const BITCOIN_TESTNET_STACKS_2_05_BURN_HEIGHT: u64 = 2_104_076;
```
gregorycoppola, line 97: Is this in the past? @CharlieC3 https://www.blockchain.com/btc-testnet/block/0000000000000017bd07ae4a153830268e39d744efce605c69af5ccf6adf79a7

CharlieC3: Seems like it, is that intentional @kantai ?

kantai: Yep -- good catch -- updated to 2104380 -- that's the current block height (2104096) + 284, which should be about 48 hours.

PR #2931, opened 2021-11-16
next-costs fix: invoke connect early to avoid race-to-connect

kantai:


jcnelson (approved):
I was just about to add this

kantai:
This, it turns out, isn't *strictly* necessary, but it does make it much more explicit exactly when the db instantiation occurs.

pavitthrap (approved):
lgtm

PR #2934, opened 2021-11-17
Ci/next costs coverage

CharlieC3:
## Description

Cherry-picking changes from [this PR](https://github.com/blockstack/stacks-blockchain/pull/2932) to `next-costs` branch. Also, adding code coverage for one more integration test, `bitcoind_integration_test`.

Fixes code coverage uploads.
About 7 months ago test coverage reports stopped uploading to codecov - not going to go into the details here, but they had a MiM and had to stop using the original method of upload.

This PR fixes the codecoverage upload step using their new CLI.

Code coverage of this repo can be found here:
https://app.codecov.io/gh/blockstack/stacks-blockchain/

## Type of Change
- Other 

No

### DIFF
--- a/.github/actions/bitcoin-int-tests/Dockerfile.generic.bitcoin-tests
+++ b/.github/actions/bitcoin-int-tests/Dockerfile.generic.bitcoin-tests
@@ -13,7 +13,8 @@ RUN rustup override set nightly && \
 ENV RUSTFLAGS="-Zinstrument-coverage" \
     LLVM_PROFILE_FILE="stacks-blockchain-%p-%m.profraw"
 
-RUN cargo build
+RUN cargo test --norun && \
+    cargo build
 
 RUN cd / && wget https://bitcoin.org/bin/bitcoin-core-0.20.0/bitcoin-0.20.0-x86_64-linux-gnu.tar.gz
 RUN cd / && tar -xvzf bitcoin-0.20.0-x86_64-linux-gnu.tar.gz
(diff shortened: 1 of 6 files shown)


jcnelson (approved):
LGTM!

gregorycoppola:
Eagerly awaiting the results of this test!

.github/actions/bitcoin-int-tests/Dockerfile.generic.bitcoin-tests
```diff
@@ -5,7 +5,15 @@ WORKDIR /src/
...
-RUN cargo test --no-run
+
+RUN rustup override set nightly && \
+    rustup component add llvm-tools-preview && \
+    cargo install grcov
+
+ENV RUSTFLAGS="-Zinstrument-coverage" \
+    LLVM_PROFILE_FILE="stacks-blockchain-%p-%m.profraw"
+
+RUN cargo build
```
kantai, line 16: I think this line should remain:

`RUN cargo test --no-run`

Because the test build is different than the normal cargo build outputs. That way, each individual integration test doesn't also have to perform the build.

CharlieC3: I changed it here because I need a binary to be produced because it's required for codecov to interpret the generated report properly. I could use `cargo test --no-run`, but I'd have to run `cargo build` following it to get the binary. Unless there's a way to generate the binary using `cargo test`?

CharlieC3: As discussed offline, updated to run `cargo test --no-run` first then `cargo build` to generate the needed stacks-node binary.

kantai (approved):
LGTM! Thanks @CharlieC3 !

gregorycoppola (approved):
Really helps to understand the 2.05 launch!

Coverage is looking pretty good for the 2.05 code.

PR #2935, opened 2021-11-17
Stacks 2.05: make the node abort if it starts in epoch 2.05 but still has epoch 2.0 chainstate

### LINKED ISSUE #2899: 2.05 node should panic if it detects 2.0 chainstate post-2.05
If a 2.05 node (1) boots from 2.0 chainstate and (2) detects that the 2.05 epoch change should have happened already, then it should panic and tell the user to restart from genesis (since it will have invalid chainstate data written by the 2.0 node post-2.05).

jcnelson:
This fixes #2899 by adding a top-level `check_chainstate_db_versions()` method in `src/chainstate/coordinators/mod.rs` that inspects the sortition DB and chainstate DB, and verifies that their current schemas are supported in the given epoch (if they exist at all).  I made the `RunLoop::start()` method in the neon node invoke this early in its initialization, so that if the node is starting from existing chainstate, it will abort if the epoch of the last sortition processed is not supported by either of the chainstate schemas.  I added a unit test to `src/chainstate/stacks/coordinator/tests.rs` that verifies that an epoch 2.0 sortition DB and epoch 2.0 chainstate DB cannot be opened if the system is in epoch 2.05.

In writing this, I also refactored the code for generating directories and paths to the various files in the chainstate directory to make them a bit more approachable (and to make it possible to query the `DBConfig` directly, from the chainstate path).

### DIFF
--- a/src/chainstate/coordinator/tests.rs
+++ b/src/chainstate/coordinator/tests.rs
@@ -3855,3 +3855,78 @@ fn preprocess_block(
         )
         .unwrap();
 }
+
+#[test]
+fn test_check_chainstate_db_versions() {
+    let path = "/tmp/stacks-blockchain-check_chainstate_db_versions";
+    let _ = std::fs::remove_dir_all(path);
+
+    let sortdb_path = format!("{}/sortdb", &path);
+    let chainstate_path = format!("{}/chainstate", &path);
+
+    let epoch_2 = StacksEpoch {
+        epoch_id: StacksEpochId::Epoch20,
+        start_height: 0,
+        end_height: 10000,
+        block_limit: BLOCK_LIMIT_MAINNET_20.clone(),
+        network_epoch: PEER_VERSION_EPOCH_2_0,
+    };
+    let epoch_2_05 = StacksEpoch {
+        epoch_id: StacksEpochId::Epoch2_05,
+        start_height: 0,
+        end_height: 10000,
+        block_limit: BLOCK_LIMIT_MAINNET_205.clone(),
+        network_epoch: PEER_VERSION_EPOCH_2_05,
+    };
+
+    // should work just fine in epoch 2 if the DBs don't exist
+    assert!(
+        check_chainstate_db_versions(&[epoch_2.clone()], &sortdb_path, &chainstate_path).unwrap()
+    );
+
+    // should work just fine in epoch 2.05 if the DBs don't exist
+    assert!(
+        check_chainstate_db_versions(&[epoch_2_05.clone()], &sortdb_path, &chainstate_path)
+            .unwrap()
+    );
+
+    StacksChainState::make_chainstate_dirs(&chainstate_path).unwrap();
+
+    let sortdb_v1 =
+        SortitionDB::connect_v1(&sortdb_path, 100, &BurnchainHeaderHash([0x00; 32]), 0, true)
+            .unwrap();
+    let chainstate_v1 = StacksChainState::open_db_without_migrations(
+        false,
+        0x80000000,
+        &StacksChainState::header_index_root_path(PathBuf::from(&chainstate_path))
+            .to_str()
+            .unwrap(),
+    )
+    .unwrap();
+
+    assert!(fs::metadata(&chainstate_path).is_ok());
+    assert!(fs::metadata(&sortdb_path).is_ok());
+    assert_eq!(
+        StacksChainState::get_db_config_from_path(&chainstate_path)
+            .unwrap()
+            .version,
+        "1"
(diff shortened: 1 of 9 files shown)


testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -167,6 +167,18 @@ impl RunLoop {
...
+            &epochs,
+            &self.config.get_burn_db_file_path(),
+            &self.config.get_chainstate_path_str(),
+        )
+        .expect("FATAL: unable to query filesystem or databases for version information")
+        {
+            panic!(
+                "FATAL: chainstate database(s) are not compatible with the current system epoch"
+            );
+        }
```
kantai, line 181: Have you run this branch on an empty chainstate directory on mainnet or testnet? When I made the `connect_dbs()` change in #2931, calling it before the `burnchain.start()` led to a panic, because the burnchain doesn't have a `first_block_hash` yet, so I had to move the `connect_dbs` invocation after `burnchain.start()`, and I wonder if the same problem would crop up here.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -1480,6 +1481,16 @@ impl BurnchainController for BitcoinRegtestController {
...
 
+    fn get_stacks_epochs(&self) -> Vec<StacksEpoch> {
+        match self.config.burnchain.epochs.as_ref() {
+            Some(epochs) => epochs.clone(),
+            None => {
+                let (_, indexer) = self.setup_indexer_runtime();
+                indexer.get_stacks_epochs()
+            }
+        }
+    }
```
kantai, line 1492: While I think this is always the same, I think we should just always consult the indexer for the stacks epochs: it should always be configured with the correct epochs from either the config or the static variables.

src/chainstate/stacks/db/mod.rs
```diff
@@ -762,15 +773,30 @@ impl StacksChainState {
...
+        let index_pathbuf =
+            StacksChainState::header_index_root_path(PathBuf::from(chainstate_root_path));
+        let index_path = index_pathbuf
+            .to_str()
+            .ok_or_else(|| db_error::ParseError)?
+            .to_string();
+
+        let marf = StacksChainState::open_index(&index_path)?;
+        StacksChainState::load_db_config(marf.sqlite_conn())
+    }
```
kantai, line 797: This is only ever used in the new test, right? Can you move it to `coordinator::tests`?

kantai: Nevermind -- I see it gets used in the schema checks.

kantai (approved):
This looks good to me, just make sure that the branch doesn't try to `get_first_block_header_hash()` before the burnchain sync starts.

src/chainstate/coordinator/tests.rs
```diff
@@ -3855,3 +3855,78 @@ fn preprocess_block(
...
+    );
+
+    StacksChainState::make_chainstate_dirs(&chainstate_path).unwrap();
+
+    let sortdb_v1 =
+        SortitionDB::connect_v1(&sortdb_path, 100, &BurnchainHeaderHash([0x00; 32]), 0, true)
+            .unwrap();
+    let chainstate_v1 = StacksChainState::open_db_without_migrations(
+        false,
+        0x80000000,
```
gregorycoppola, line 3900: nit: Could use the constant for this `CHAIN_ID_TESTNET`.

testnet/stacks-node/src/run_loop/neon.rs
gregorycoppola: +1 The change looks good to me, but does it work on a real node?

gregorycoppola (commented):
Change looks good and easy to follow.

Just wondering about the testing on full testnet/mainnet nodes, especially since Charlie's testnet tests are already in flight.

testnet/stacks-node/src/run_loop/neon.rs
jcnelson: Yup, it definitely works.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
jcnelson: Yup, fixed

src/chainstate/coordinator/tests.rs
jcnelson: Done

gregorycoppola (approved):

PR #2937, opened 2021-11-18
Stacks 2.05: only report neighbors in the current or later epoch

jcnelson:
This PR updates the peer database query logic to only select for neighbors whose peer version epoch marker meets or exceeds a given epoch marker, which in turn is derived from whatever epoch the node is currently in.  A 2.05 node running in the 2.0 epoch will return neighbors with either `0x00` or `0x05` as their epoch marker, but a 2.05 node running in the 2.05 epoch will only return neighbors with `0x05` (note that the neighbors returned depends not on the epochs the node supports per se, but on where the node is in processing the burnchain).  The logic here prevents a node from serving neighbors that are no longer compatible with the current consensus rules to peers that ask for them -- either via `/v2/neighbors` or via a `GetNeighbors` p2p request.

### DIFF
--- a/src/net/db.rs
+++ b/src/net/db.rs
@@ -1242,6 +1242,7 @@ impl PeerDB {
     pub fn get_random_neighbors(
         conn: &DBConn,
         network_id: u32,
+        network_epoch: u8,
         count: u32,
         block_height: u64,
         always_include_allowed: bool,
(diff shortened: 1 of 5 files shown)


kantai (approved):
LGTM!

src/net/db.rs
```diff
@@ -1663,6 +1674,126 @@ mod test {
...
+                allowed: -1,
+                denied: -1,
+                asn: (34567 + i) as u32,
+                org: (45678 + i) as u32,
+                in_degree: 1,
+                out_degree: 1,
+            });
+        }
+
+        fn are_present(ne: &Vec<Neighbor>, nei: &Vec<Neighbor>) -> bool {
```
gregorycoppola, line 1723: suggestion: 
```
/// Checks that each element of `ne` is also present in `nei`.
```

gregorycoppola (approved):
thanks for taking this one

Issue #2939, opened 2021-11-18
2.05: Update costs-2.clar with final proposal from SIP-012


kantai:
We should update costs-2.clar with the final values from SIP-012 (I think only one function has changed).

In the process of this, we should also make sure our `vm::tests::costs` and `vm::analysis::tests::costs` tests cover the _mainnet_ flag being set to true (this should be pretty easy to do -- those tests all setup their instance specifically with `mainnet=false`, so we can just add test variants that set it to true instead).

kantai:
This was done in #2942 -- the testing PR was separated from the costs sync.

PR #2940, opened 2021-11-18
testing: Unit tests for cost contracts on mainnet

gregorycoppola:
Addresses #2939 

We update `test analysis::tests::costs` to test the `BOOT_CODE_COSTS_2` can load on mainnet.

Previously, code coverage analysis showed we only were testing the testnet cost loading path.

New coverage results:
https://app.codecov.io/gh/blockstack/stacks-blockchain/blob/b5c50a4f03550108f63e7b6e1cbf9fc64f34a1e6/src/clarity_vm/clarity.rs

### DIFF
--- a/clarity/Cargo.toml
+++ b/clarity/Cargo.toml
@@ -80,6 +80,8 @@ assert-json-diff = "1.0.0"
 #  but it isn't necessary for tests: only benchmarks. therefore, commenting out for now.
 # criterion = "0.3"
 stx_genesis = { package = "stx-genesis", path = "../stx-genesis/."}
+rstest = "0.11.0"
+rstest_reuse = "0.1.3"
 
 [features]
 default = ["developer-mode"]
--- a/clarity/Cargo.toml
+++ b/clarity/Cargo.toml
@@ -80,8 +80,6 @@ assert-json-diff = "1.0.0"
 #  but it isn't necessary for tests: only benchmarks. therefore, commenting out for now.
 # criterion = "0.3"
 stx_genesis = { package = "stx-genesis", path = "../stx-genesis/."}
-rstest = "0.11.0"
-rstest_reuse = "0.1.3"
 
 [features]
 default = ["developer-mode"]
--- a/src/clarity_vm/clarity.rs
+++ b/src/clarity_vm/clarity.rs
@@ -378,7 +378,7 @@ impl ClarityInstance {
 
         let epoch = GENESIS_EPOCH;
 
-        let cost_track = Some(LimitedCostTracker::new_free());
+        let cost_track = Some(LimitedCostTracker::new_free_on_network(self.mainnet));
 
         let mut conn = ClarityBlockConnection {
             datastore: writable,
--- a/src/clarity_vm/clarity.rs
+++ b/src/clarity_vm/clarity.rs
@@ -389,7 +389,7 @@ impl ClarityInstance {
             epoch,
         };
 
-        let use_mainnet = self.mainnet;
+        let use_mainnet = self.mainnet.clone();
         conn.as_transaction(|clarity_db| {
             let (ast, _) = clarity_db
                 .analyze_smart_contract(&boot_code_id("costs", use_mainnet), BOOT_CODE_COSTS)
--- a/src/clarity_vm/clarity.rs
+++ b/src/clarity_vm/clarity.rs
@@ -378,7 +378,7 @@ impl ClarityInstance {
 
         let epoch = GENESIS_EPOCH;
 
-        let cost_track = Some(LimitedCostTracker::new_free_on_network(self.mainnet));
+        let cost_track = Some(LimitedCostTracker::new_free());
 
         let mut conn = ClarityBlockConnection {
             datastore: writable,
--- a/src/clarity_vm/clarity.rs
+++ b/src/clarity_vm/clarity.rs
@@ -389,7 +389,7 @@ impl ClarityInstance {
             epoch,
         };
 
-        let use_mainnet = self.mainnet.clone();
+        let use_mainnet = self.mainnet;
         conn.as_transaction(|clarity_db| {
             let (ast, _) = clarity_db
                 .analyze_smart_contract(&boot_code_id("costs", use_mainnet), BOOT_CODE_COSTS)
--- a/src/clarity_vm/clarity.rs
+++ b/src/clarity_vm/clarity.rs
@@ -385,17 +385,18 @@ impl ClarityInstance {
             header_db,
             burn_state_db,
             cost_track,
-            mainnet: false,
+            mainnet: self.mainnet,
             epoch,
         };
 
+        let use_mainnet = self.mainnet.clone();
         conn.as_transaction(|clarity_db| {
             let (ast, _) = clarity_db
-                .analyze_smart_contract(&boot_code_id("costs", false), BOOT_CODE_COSTS)
+                .analyze_smart_contract(&boot_code_id("costs", use_mainnet), BOOT_CODE_COSTS)
                 .unwrap();
             clarity_db
                 .initialize_smart_contract(
-                    &boot_code_id("costs", false),
+                    &boot_code_id("costs", use_mainnet),
                     &ast,
                     BOOT_CODE_COSTS,
                     |_, _| false,
@@ -406,13 +407,13 @@ impl ClarityInstance {
         conn.as_transaction(|clarity_db| {
             let (ast, analysis) = clarity_db
                 .analyze_smart_contract(
-                    &boot_code_id("cost-voting", false),
+                    &boot_code_id("cost-voting", use_mainnet),
                     &*BOOT_CODE_COST_VOTING,
                 )
                 .unwrap();
             clarity_db
                 .initialize_smart_contract(
-                    &boot_code_id("cost-voting", false),
+                    &boot_code_id("cost-voting", use_mainnet),
                     &ast,
                     &*BOOT_CODE_COST_VOTING,
                     |_, _| false,
@@ -420,17 +421,17 @@ impl ClarityInstance {
                 .unwrap();
 
             clarity_db
-                .save_analysis(&boot_code_id("cost-voting", false), &analysis)
+                .save_analysis(&boot_code_id("cost-voting", use_mainnet), &analysis)
                 .unwrap();
         });
 
         conn.as_transaction(|clarity_db| {
             let (ast, _) = clarity_db
-                .analyze_smart_contract(&boot_code_id("pox", false), &*BOOT_CODE_POX_TESTNET)
+                .analyze_smart_contract(&boot_code_id("pox", use_mainnet), &*BOOT_CODE_POX_TESTNET)
                 .unwrap();
             clarity_db
                 .initialize_smart_contract(
-                    &boot_code_id("pox", false),
+                    &boot_code_id("pox", use_mainnet),
                     &ast,
                     &*BOOT_CODE_POX_TESTNET,
                     |_, _| false,
--- a/src/libclarity.rs
+++ b/src/libclarity.rs
@@ -66,6 +66,14 @@ extern crate serde_json;
 #[macro_use]
 extern crate assert_json_diff;
 
+#[cfg(test)]
+#[macro_use]
+extern crate rstest;
+
+#[cfg(test)]
+#[macro_use]
+extern crate rstest_reuse;
+
 #[cfg(feature = "monitoring_prom")]
 #[macro_use]
 pub extern crate prometheus;
--- a/src/libclarity.rs
+++ b/src/libclarity.rs
@@ -66,14 +66,6 @@ extern crate serde_json;
 #[macro_use]
 extern crate assert_json_diff;
 
-#[cfg(test)]
-#[macro_use]
-extern crate rstest;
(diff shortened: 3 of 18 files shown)


kantai:
This is great -- can you also update `vm::tests::costs`?

src/clarity_vm/clarity.rs
```diff
@@ -385,17 +385,18 @@ impl ClarityInstance {
             header_db,
             burn_state_db,
             cost_track,
-            mainnet: false,
+            mainnet: self.mainnet,
             epoch,
         };
 
+        let use_mainnet = self.mainnet.clone();
```
jcnelson, line 392: `.clone()` should be unnecessary, since `bool` implements `Copy`.

kantai:
Looking at 1a5164c7a745f3fdb1382681567d47d476972093 -- I think the issue is the following:

`vm::tests::costs::with_owned_env` invokes:

```rust
    to_do(OwnedEnvironment::new_max_limit(
        store.as_clarity_db(&TEST_HEADER_DB, &TEST_BURN_STATE_DB),
        epoch,
    ))
```

Which is defined in `src/vm/costs/mod.rs`:

```rust
    #[cfg(test)]
    pub fn new_max_limit(
        clarity_db: &mut ClarityDatabase,
        epoch: StacksEpochId,
    ) -> Result<LimitedCostTracker> {
        assert!(clarity_db.is_stack_empty());
        LimitedCostTracker::new(false, ExecutionCost::max_value(), clarity_db, epoch)
    }
```

That `false` will need to be a parameter `is_mainnet`.

src/vm/contexts.rs
```diff
@@ -531,7 +531,7 @@ impl EventBatch {
 
 impl<'a> OwnedEnvironment<'a> {
     #[cfg(test)]
-    pub fn new(database: ClarityDatabase<'a>) -> OwnedEnvironment<'a> {
+    pub fn new(database: ClarityDatabase<'a>, use_mainnet: bool) -> OwnedEnvironment<'a> {
         let epoch = StacksEpochId::Epoch2_05;
         OwnedEnvironment {
             context: GlobalContext::new(false, database, LimitedCostTracker::new_free(), epoch),
```
kantai, line 537: I think `false` on line 537 should be updated to `use_mainnet` -- I don't think it's causing your test failures though.

kantai:
It looks like the error is coming from lines 1207 and 921 of the `src/tests/costs.rs` file:

```rust
        db.set_variable_unknown_descriptor(
            &COST_VOTING_TESTNET_CONTRACT,
            "confirmed-proposal-count",
            Value::UInt(1),
        )
        .unwrap();
```

and

```rust
        db.set_variable_unknown_descriptor(
            &COST_VOTING_TESTNET_CONTRACT,
            "confirmed-proposal-count",
            Value::UInt(bad_proposals as u128),
        )
        .unwrap();
```

I think those lines need to use `COST_VOTING_MAINNET_CONTRACT` when running with the mainnet flag.

For future reference, it's usually helpful when trying to debug test failures to run with the environment variable `RUST_BACKTRACE=1`, which will tell rust to output the full backtrace:

```
-------------- TEST START --------------
thread 'vm::tests::costs::test_cost_voting_integration::case_1' panicked at 'called `Result::unwrap()` on an `Err` value: Unchecked(NoSuchDataVariable("confirmed-proposal-count"))', src/vm/tests/costs.rs:1212:10
stack backtrace:
   0: rust_begin_unwind
             at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/core/src/panicking.rs:92:14
   2: core::result::unwrap_failed
             at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/core/src/result.rs:1355:5
   3: core::result::Result<T,E>::unwrap
             at /Users/aaron/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/result.rs:1037:23
   4: blockstack_lib::vm::tests::costs::test_cost_voting_integration
             at ./src/vm/tests/costs.rs:1207:9
   5: blockstack_lib::vm::tests::costs::test_cost_voting_integration::case_1
             at ./src/vm/tests/costs.rs:296:1
   6: blockstack_lib::vm::tests::costs::test_cost_voting_integration::case_1::{{closure}}
             at ./src/vm/tests/costs.rs:296:1
   7: core::ops::function::FnOnce::call_once
             at /Users/aaron/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/function.rs:227:5
   8: core::ops::function::FnOnce::call_once
             at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

And you can usually find the offending line of code pretty quickly that way.

kantai:
Also -- note that the use of `rstest` seems to have broken the clarity build:

```
cargo check --workspace --tests
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/aaron/devel/stacks-blockchain/clarity/Cargo.toml
workspace: /Users/aaron/devel/stacks-blockchain/Cargo.toml
    Checking blockstack-core v0.0.1 (/Users/aaron/devel/stacks-blockchain)
    Checking clarity v0.0.1 (/Users/aaron/devel/stacks-blockchain/clarity)
    Checking stacks-node v0.1.0 (/Users/aaron/devel/stacks-blockchain/testnet/stacks-node)
error[E0432]: unresolved import `rstest`
  --> clarity/../src/vm/tests/costs.rs:49:5
   |
49 | use rstest::rstest;
   |     ^^^^^^ maybe a missing crate `rstest`?

...
```

So you'll need to add the dev-dependencies to `clarity/Cargo.toml` and the extern-crate statements to `src/libclarity.rs`.

src/vm/tests/costs.rs
```diff
@@ -1287,7 +1365,8 @@ fn test_cost_voting_integration() {
         let (_db, tracker) = owned_env.destruct().unwrap();
 
         // cost of `le` should be less now, because the proposal made it free
-        assert!(le_cost_without_interception.exceeds(&tracker.get_total()));
```
gregorycoppola, line 1290: assertion failed

src/vm/tests/costs.rs
```diff
@@ -1306,12 +1385,13 @@ fn test_cost_voting_integration() {
 
         for (target, referenced_function) in tracker.cost_function_references().into_iter() {
             if target == &ClarityCostFunction::Le {
-                assert_eq!(&referenced_function.contract_id, &cost_definer);
-                assert_eq!(&referenced_function.function_name, "cost-definition-le");
+                // DO NOT SUBMIT: uncomment
```
gregorycoppola, line 1388: assertion failed

clarity/Cargo.toml
```diff
@@ -80,6 +80,8 @@ assert-json-diff = "1.0.0"
 #  but it isn't necessary for tests: only benchmarks. therefore, commenting out for now.
 # criterion = "0.3"
 stx_genesis = { package = "stx-genesis", path = "../stx-genesis/."}
+rstest = "0.11.0"
+rstest_reuse = "0.1.3"
```
kantai, line 84: You can drop the rstest dependency from here too -- and then the Cargo.lock changes should be zero.

src/libclarity.rs
```diff
@@ -66,6 +66,14 @@ extern crate serde_json;
...
 extern crate assert_json_diff;
 
+#[cfg(test)]
+#[macro_use]
+extern crate rstest;
+
+#[cfg(test)]
+#[macro_use]
+extern crate rstest_reuse;
+
```
kantai, line 76: Like I say above -- you can drop `rstest` from libclarity now.

src/vm/contexts.rs
gregorycoppola: I reverted this change. It wasn't necessary to get this round of tests to pass, so I think it's better to keep the PR as small as possible.

clarity/Cargo.toml
gregorycoppola: Yes. I dropped the rstest dependency and "manually" built the grids.

src/clarity_vm/clarity.rs
gregorycoppola: done

src/vm/analysis/tests/costs.rs
```diff
@@ -173,24 +173,42 @@ pub fn test_tracked_costs(prog: &str, epoch: StacksEpochId) -> ExecutionCost {
...
 }
+
+#[test]
+fn epoch_205_test_all_mainnet() {
+    epoch_205_test_all(true)
+}
+
+#[test]
+fn epoch_205_test_all_testnet() {
+    epoch_205_test_all(true)
```
kantai, line 213: Should be false

kantai (approved):
LGTM!

src/vm/costs/mod.rs
```diff
@@ -640,10 +641,14 @@ impl LimitedCostTracker {
             memory: 0,
             memory_limit: CLARITY_MEMORY_LIMIT,
             non_free: None,
-            mainnet: false,
+            mainnet: use_mainnet,
         }
     }
 
+    pub fn new_free() -> LimitedCostTracker {
+        LimitedCostTracker::new_free_on_network(false)
```
jcnelson, line 649: I don't think this is right.  We use `LimitedCostTracker::new_free()` in mainnet-reachable code, such as in `src/net/rpc.rs`, in `src/vm/analysis/mod.rs` (in `type_check()`), and in `src/clarity_vm/clarity.rs` (i.e. in `begin_genesis_block()`).

kantai: The free cost tracker never uses the mainnet flag — prior to this PR, I think the mainnet flag was always set to false. The cleanest way to have implemented the free tracker would probably to make the LimitedCostTracker an enum type, but that doesn’t end up having much to do with this PR.

jcnelson: This feels unsafe and brittle.  Why should a cost tracker's `mainnet` flag be set to `false`, when the node is actually running on mainnet?

Currently, `LimitedCostTracker'`s mainnet flag is only used in calls to `boot_code_id()` and `load_cost_functions()` in `LimitedCostTracker::load_costs()`.  While it is the case that `LimitedCostTracker::load_costs()` is only called by `LimitedCostTracker::new()` and `LimitedCostTracker::new_mid_block()`, it seems that the only reason this doesn't cause immediate problems is because all of the `CostTracker` trait methods in the `LimitedCostTracker` implementation short-circuit if the `self.free` flag is set.  This short-circuiting prevents these trait method implementations from ever trying to access the loaded cost functions.  However, the act of trying to load a cost function in a `LimitedCostTracker` created by `new_free()` doesn't seem to panic; the trait method implementations would instead return `CostErrors::CostComputationFailed("Computation not defined: {}")`.  This, to me, looks like an accident waiting to happen -- for example, suppose we accidentally try to rely on `LimitedCostTracker`'s mainnet flag indicating the mainnet status of the system, like we did with `Burnchain`'s mainnet flag.

Is there a way we can de-risk this here without doing a lot of refactoring?  For example, could we make `load_costs()` panic if `self.free` is set, and make the `CostTracker` trait method implementations panic if `mainnet` and `free` are both true?

kantai: The refactor is pretty straight-forward: https://github.com/blockstack/stacks-blockchain/pull/2945

gregorycoppola: As I said on #2945 , this PR is meant for test coverage, but it introduces new changes to production code, which really should induce a new round of testing.

One option is to just *look* at the test coverage, but not actually merge this into the release, so that we don't introduce new changes to prod over the weekend.

gregorycoppola: > This feels unsafe and brittle. Why should a cost tracker's mainnet flag be set to false, when the node is actually running on mainnet?

This new code is just making explicit the previous behavior. It's not changing the behavior, as kantai said. In other words, it was effectively using the value `mainnet: false` before.

src/vm/costs/mod.rs
```diff
@@ -640,10 +641,14 @@ impl LimitedCostTracker {
             memory: 0,
             memory_limit: CLARITY_MEMORY_LIMIT,
             non_free: None,
-            mainnet: false,
```
gregorycoppola, line 643: note: previous behavior of new_free

src/vm/costs/mod.rs
kantai: > One option is to just look at the test coverage, but not actually merge this into the release, so that we don't introduce new changes to prod over the weekend.

True, though I do believe we should merge the tests and refactor — this is useful to integrate into CI. Certainly doesn’t need to be merged on the weekend, though (and merging to next-costs by itself doesn’t mean production yet either).

gregorycoppola: @kantai @jcnelson  If you guys want to merge it this weekend, go ahead.

Otherwise can defer.

jcnelson (approved):

PR #2941, opened 2021-11-18
2.05 Chore: add CHANGELOG.md entry

kantai:
This adds a CHANGELOG.md entry for 2.05

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,11 @@ and this project adheres to the versioning scheme outlined in the [README.md](RE
 This software update is a consensus changing release and the
 implementation of the proposed cost changes in SIP-012. This release's
 chainstate directory is compatible with chainstate directories from
-2.0.11.4.0, however, it is only compatible with chainstate directories
-before the 2.05 consensus changes activate (Bitcoin height 713,000). If
-you run a 2.00 stacks-node beyond this point, and wish to run a 2.05 node
-afterwards, you must start from a new chainstate directory.
+2.0.11.4.0. However, this release is only compatible with chainstate
+directories before the 2.05 consensus changes activate (Bitcoin height
+713,000). If you run a 2.00 stacks-node beyond this point, and wish to
+run a 2.05 node afterwards, you must start from a new chainstate
+directory.
 
 ## Added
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,36 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
+## [2.05.0.0.0]
+
+This software update is a consensus changing release and the
+implementation of the proposed cost changes in SIP-012. This release's
+chainstate directory is compatible with chainstate directories from
+2.0.11.4.0, however, it is only compatible with chainstate directories
```
jcnelson, line 13: Run-on sentence.  `However, it should only ...` should be a new sentence.

Also, can you drop a link here to the SIP 012 text?  Thanks

CHANGELOG.md
```diff
@@ -5,6 +5,36 @@ All notable changes to this project will be documented in this file.
...
+before the 2.05 consensus changes activate (Bitcoin height 713,000). If
+you run a 2.00 stacks-node beyond this point, and wish to run a 2.05 node
+afterwards, you must start from a new chainstate directory.
+
+## Added
+
+- At height 713,000 a new `costs-2` contract will be launched by the
+  null Stacks address.
```
jcnelson, line 21: Is there an official name for the zero-address?  The code calls it the "boot code address" for example.

kantai: Yeah, I liked boot code address, but it's no longer boot code only :(

Not sure otherwise, happy to call it whatever.

CHANGELOG.md
kantai: Will fix the run on sentence.

As for a SIP-012 text link -- it hasn't merged yet, and I'm hesitant to link to a branch.

CHANGELOG.md
jcnelson: I think of it like a BIOS or a ROM -- it's the address where all the _pre-defined_ smart contract code lives that ships with the node.  It's true that it's mainly involved in bootup, but it's randomly accessed and called by other running code loaded later and it's allowed to receive new code as part of a system-wide upgrade.  That to me is consistent with calling it the "boot code address."

I guess I'm more anxious about the code and documentation calling it two different things than the address having a more appropriate name like "system address."

kantai: Okay, updated to "Stacks boot address"

jcnelson (approved):
LGTM!

gregorycoppola (approved):

PR #2945, opened 2021-11-20
chore: refactor LimitedCostTracker into Free and Limited variants

kantai:
Per some discussion on #2940 -- the existing LimitedCostTracker's split free/limited implementation was very brittle, and a simple refactor into an enum would fix that. This PR is a pure refactor of the LimitedCostTracker into that enum. The scope of the changes are pretty limited: none of the LimitedCostTracker's fields were exposed outside the `costs/mod.rs` file.

### DIFF
--- a/src/clarity_vm/clarity.rs
+++ b/src/clarity_vm/clarity.rs
@@ -378,7 +378,7 @@ impl ClarityInstance {
 
         let epoch = GENESIS_EPOCH;
 
-        let cost_track = Some(LimitedCostTracker::new_free_on_network(self.mainnet));
+        let cost_track = Some(LimitedCostTracker::new_free());
 
         let mut conn = ClarityBlockConnection {
             datastore: writable,
--- a/src/vm/costs/mod.rs
+++ b/src/vm/costs/mod.rs
@@ -223,7 +223,8 @@ impl CostStateSummary {
 }
 
 #[derive(Clone)]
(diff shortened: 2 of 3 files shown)


jcnelson (approved):
LGTM! Thanks for taking the time to do this refactoring -- it significantly reduces my anxiety around `LimitedCostTracker`.

gregorycoppola (approved):

gregorycoppola:
@jcnelson @kantai  since the whole point of this change is to increase test coverage, and not to add functionality, another option is to just look at the test coverage but avoid pushing any of this before the 2.05 release.

I agree that there are some sensitive prod-touching changes in here, and these changes arent tested by, e.g., the testnet work we have done so far. I would hate to have this testing PR introduce a bug that we don't understand.

kantai:
> @jcnelson @kantai since the whole point of this change is to increase test coverage, and not to add functionality, another option is to just look at the test coverage but avoid pushing any of this before the 2.05 release.

Well, I do think that these test cases ended up highlighting some code that, if not problematic now, could be a trap in the future. The refactored code is well covered in our tests, and there’s not actually new functionality added here— just a cleaner separation between the free and non-free tracker.

gregorycoppola:
I think we should merge these changes. I'm just saying that we can wait until after the release. Ir merge them to develop.

PR #2948, opened 2021-11-22
Stacks 2.05: add integration test to verify that the miner can pack a 2.05-sized epoch into a microblock stream

### LINKED ISSUE #2912: [mining] add integration test to confirm that microblock streams can consume the 2.05 block limit in 2.05
Once 2.05 goes live, a miner should be able to mine microblock streams off of a 2.05 block that consume the 2.05 block budget.

jcnelson:
No new features in this PR; this just fixes #2912 by adding an integration test to verify that we can pack a *lot* more compute power into a microblock stream after the 2.05 epoch transition.

### DIFF
--- a/testnet/stacks-node/src/tests/epoch_205.rs
+++ b/testnet/stacks-node/src/tests/epoch_205.rs
@@ -1130,11 +1130,11 @@ fn bigger_microblock_streams_in_2_05() {
     microblocks_processed.store(0, Ordering::SeqCst);
 
     // this test can sometimes miss a mine block event.
-    sleep_ms(80_000);
+    sleep_ms(120_000);
     next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
 
-    let blocks = test_observer::get_blocks();
-    assert!(blocks.len() >= 5, "Should have produced at least 5 blocks");
+    let mut epoch_20_stream_cost = ExecutionCost::zero();
+    let mut epoch_205_stream_cost = ExecutionCost::zero();
 
     // max == largest number of transactions per stream in a given epoch (2.0 or 2.05)
     // total == number of transactions across all streams in a given epoch (2.0 or 2.05)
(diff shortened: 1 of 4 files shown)


kantai (commented):
Awesome! Can you add the new test to the integration test matrix in `.github/workflows/bitcoin-tests.yml`?

jcnelson:
Yup; added.  Thanks for catching!

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -331,7 +331,7 @@ pub fn wait_for_runloop(blocks_processed: &Arc<AtomicU64>) {
     }
 }
 
-fn wait_for_microblocks(microblocks_processed: &Arc<AtomicU64>, timeout: u64) -> bool {
+pub fn wait_for_microblocks(microblocks_processed: &Arc<AtomicU64>, timeout: u64) -> bool {
```
gregorycoppola, line 334: while we are here, can we add a concise but precise contract for this method?

It's not clear from the signature what is going to happen.

testnet/stacks-node/src/tests/epoch_205.rs
```diff
@@ -913,3 +916,292 @@ fn test_cost_limit_switch_version205() {
...
+                        num_big_microblock_txs += 1;
+                        total_big_txs_per_microblock_20 += 1;
+                    }
+                }
+            }
+        }
+        if in_205 && num_big_microblock_txs > max_big_txs_per_microblock_205 {
+            max_big_txs_per_microblock_205 = num_big_microblock_txs;
+        }
+        if !in_205 && num_big_microblock_txs > max_big_txs_per_microblock_20 {
```
gregorycoppola, line 1185: else?

testnet/stacks-node/src/tests/epoch_205.rs
```diff
@@ -913,3 +916,292 @@ fn test_cost_limit_switch_version205() {
...
+
+    thread::spawn(move || run_loop.start(None, 0));
+
+    // give the run loop some time to start up!
+    wait_for_runloop(&blocks_processed);
+
+    // first block wakes up the run loop
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    // first block will hold our VRF registration
```
gregorycoppola, line 1075: we have two "first block"s in the comments

you can optionally use `run_until_burnchain_height` in `testnet/stacks-node/src/tests/epoch_205.rs`. this takes a "burnchain height" argument, so if the tests passes the reader will be able to know exactly which block we're on.

the numbers then can't be out of date because they'll be part of the test

testnet/stacks-node/src/tests/epoch_205.rs
```diff
@@ -913,3 +916,292 @@ fn test_cost_limit_switch_version205() {
...
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+    wait_for_microblocks(&microblocks_processed, 180);
+
+    // this test can sometimes miss a mine block event.
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    let blocks = test_observer::get_blocks();
+    assert!(blocks.len() >= 5, "Should have produced at least 5 blocks");
+
+    let mut max_big_txs_per_microblock_20 = 0;
```
gregorycoppola, line 1145: seems like you are using the word "large" in the code so maybe `s/big/large/` for these variables

testnet/stacks-node/src/tests/epoch_205.rs
```diff
@@ -913,3 +916,292 @@ fn test_cost_limit_switch_version205() {
...
+    );
+    eprintln!(
+        "max_big_txs_per_microblock_205: {}, total_big_txs_per_microblock_205: {}",
+        max_big_txs_per_microblock_205, total_big_txs_per_microblock_205
+    );
+
+    assert!(max_big_txs_per_microblock_20 < 2);
+    assert!(total_big_txs_per_microblock_20 < 2);
+
+    assert!(max_big_txs_per_microblock_205 >= 18);
```
gregorycoppola, line 1202: I'm curious why this is `>=`? Do we not have an exact expectation?

testnet/stacks-node/src/tests/epoch_205.rs
```diff
@@ -913,3 +916,292 @@ fn test_cost_limit_switch_version205() {
...
+            }
+            let tx_bytes = hex_bytes(&raw_tx[2..]).unwrap();
+            let parsed = StacksTransaction::consensus_deserialize(&mut &tx_bytes[..]).unwrap();
+            if let TransactionPayload::SmartContract(tsc) = parsed.payload {
+                if tsc.name.to_string().find("costs-2").is_some() {
+                    in_205 = true;
+                }
+                if tsc.name.to_string().find("large").is_some() {
+                    if in_205 {
+                        num_big_microblock_txs += 1;
```
gregorycoppola, line 1173: move this line outside if it's in both if/else

testnet/stacks-node/src/tests/epoch_205.rs
gregorycoppola, line 1202: can a brief comment explain why we have these expectations?  Including how do `max` and `total` differ?

testnet/stacks-node/src/tests/epoch_205.rs
```diff
@@ -913,3 +916,292 @@ fn test_cost_limit_switch_version205() {
...
+    eprintln!(
+        "max_big_txs_per_microblock_20: {}, total_big_txs_per_microblock_20: {}",
+        max_big_txs_per_microblock_20, total_big_txs_per_microblock_20
+    );
+    eprintln!(
+        "max_big_txs_per_microblock_205: {}, total_big_txs_per_microblock_205: {}",
+        max_big_txs_per_microblock_205, total_big_txs_per_microblock_205
+    );
+
+    assert!(max_big_txs_per_microblock_20 < 2);
```
gregorycoppola, line 1199: interesting that the numbers are `2` and `18`.. you only made the difference in limit 2x, but the difference in trandactions is > 9x.

 it's optoinal but you could pick a smaller number of transactions or a bigger block limit, so that there is some correspondence between the two

gregorycoppola (commented):
Thanks for the test.

I had mostly readability/factoring comments. You can treat them as optional but I'm curious to hear your thoughts.

testnet/stacks-node/src/tests/epoch_205.rs
```diff
@@ -913,3 +916,292 @@ fn test_cost_limit_switch_version205() {
...
+                read_count: 7750,
+                runtime: 5000000000,
+            },
+            network_epoch: PEER_VERSION_EPOCH_2_0,
+        },
+        StacksEpoch {
+            epoch_id: StacksEpochId::Epoch2_05,
+            start_height: 208,
+            end_height: 9223372036854775807,
+            block_limit: ExecutionCost {
```
pavitthrap, line 1031: Why are you doubling the default limits in 2.05?

pavitthrap (approved):
lgtm, just confused about the limit doubling for epoch 2.05 in the test.

testnet/stacks-node/src/tests/neon_integrations.rs
jcnelson: Added

testnet/stacks-node/src/tests/epoch_205.rs
jcnelson: Fixed

testnet/stacks-node/src/tests/epoch_205.rs
jcnelson: Done

testnet/stacks-node/src/tests/epoch_205.rs
jcnelson: Sure; added.

testnet/stacks-node/src/tests/epoch_205.rs
jcnelson: That's because the cost functions get applied at the epoch transition, which means that the previously-expensive `index-of` Clarity function gets a *lot* cheaper :)

testnet/stacks-node/src/tests/epoch_205.rs
jcnelson: I can just double the read/write counts and leave everything else untouched

testnet/stacks-node/src/tests/epoch_205.rs
jcnelson: It can vary +/- 1.  The point of the test is to verify that the microblock stream can get filled with more than the 2.0 block capacity.

testnet/stacks-node/src/tests/epoch_205.rs
```diff
@@ -913,3 +916,287 @@ fn test_cost_limit_switch_version205() {
...
+        "max_big_txs_per_microblock_205: {}, total_big_txs_per_microblock_205: {}",
+        max_big_txs_per_microblock_205, total_big_txs_per_microblock_205
+    );
+
+    assert!(max_big_txs_per_microblock_20 <= 1);
+    assert!(total_big_txs_per_microblock_20 <= 1);
+
+    // can be 9; allow a fudge factor
+    assert!(max_big_txs_per_microblock_205 >= 8);
+    assert!(total_big_txs_per_microblock_205 >= 8);
```
kantai, line 1198: Is testing the transaction receipt costs instead of testing the tx count sufficient? That is, assert that the big transactions in the microblock consume X cost before and Y cost after the epoch switch? That's a little more robust of an assertion which should cut down on test flapping and sensitivity to the runtime environment, but if its important to capture the block limit itself in the assertion somehow, perhaps just test that the sum of the MARF operations is less than the new limit but greater than the old limit?

gregorycoppola (approved):
Thanks for the change!

gregorycoppola:
... whichever way you want to handle remaining questions is up to you from my perspective.

testnet/stacks-node/src/tests/epoch_205.rs
jcnelson: Sure, done.

kantai (approved):
LGTM -- just make sure that the tests are all passing before merging.

kantai:
Changed the base to `develop`, because `next-costs` has already been merged.

PR #2956, opened 2021-11-30
merge: `next-costs` into `next`

gregorycoppola:
This PR merges the `next-costs` branch (Stacks 2.05) into the `next` branch (Stacks 2.1).

Some code re-writes required for reasons, e.g.:
* `next-costs` transitions 2->2.05 and `next` transitions 2->2.1, so we need to make sure we make both transitions
* `next-costs` code/tests aren't aware of ClarityVersion's
* new values that need to be added to structs like `StacksEpoch`
* integrating support for peer versioning into `next`


src/vm/costs/mod.rs
```diff
@@ -842,45 +906,53 @@ impl CostTracker for LimitedCostTracker {
...
-                &cost_function
-            )))?
-            .clone();
+        match self {
+            Self::Free => {
+                // tracker is free, return zero!
+                return Ok(ExecutionCost::zero());
+            }
+            Self::Limited(ref mut data) => {
+                if cost_function == ClarityCostFunction::Unimplemented {
```
jcnelson, line 915: Should this branch be a runtime panic?

src/vm/database/clarity_db.rs
```diff
@@ -279,15 +289,18 @@ impl BurnStateDB for NullBurnStateDB {
     }
 
     fn get_pox_prepare_length(&self) -> u32 {
-        panic!("NullBurnStateDB should not return PoX info");
+        1
```
jcnelson, line 292: Why did you remove all of these panics?  If it's for test compatibility, we have a separate `UnitTestBurnStateDB` for this (IIRC use `NullBurnStateDB` in production, or at least used to).

src/burnchains/mod.rs
```diff
@@ -1053,6 +1053,15 @@ pub mod test {
             txop.txid =
                 Txid::from_test_data(txop.block_height, txop.vtxindex, &txop.burn_header_hash, 0);
 
+            let epoch = SortitionDB::get_stacks_epoch(ic, txop.block_height)
+                .unwrap()
+                .expect(&format!("BUG: no epoch for height {}", &txop.block_height));
+            if epoch.epoch_id == StacksEpochId::Epoch2_05
```
jcnelson, line 1059: You should be able to use `epoch.epoch_id >= StacksEpochId::Epoch2_05`.  Also, we'll have a different epoch marker for 2.1, so please open an issue to add it and fix this test.

jcnelson:
You don't have to fix it right now, but in the future, please don't commit and then delete test artifacts.  If you do by mistake, then unwind the commits and remove them entirely from the history.  Right now, I'm paging through the list of changes you made since the merge, I see a whole lot of lines from now-deleted files called `a/out.txt`.

jcnelson:
Also, in the future, please don't commit things like `panic!("here we are")` or `let bt = backtrace::Backtrace::new(); warn!("process:bt {:?}", &bt);`.  It's totally fine to do this locally, but it makes it that much harder to read the delta between the last merge commit and `HEAD`.

jcnelson:
In addition, please remember to try and compile the code before committing it, so we don't have a slew of commits that fix compiler errors.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -2133,20 +2148,84 @@ impl SortitionDB {
...
+    }
+
     /// Validate all Stacks Epochs. Since this is data that always comes from a static variable,
     /// any invalid StacksEpoch structuring should result in a runtime panic.
     fn validate_epochs(epochs_ref: &[StacksEpoch]) -> Vec<StacksEpoch> {
         // sanity check -- epochs must all be contiguous, each epoch must be unique,
         // and the range of epochs should span the whole non-negative i64 space.
         let mut epochs = epochs_ref.to_vec();
         let mut seen_epochs = HashSet::new();
-        epochs.sort();
```
jcnelson, line 2149: Why is ts removed?

src/clarity_vm/clarity.rs
```diff
@@ -1908,25 +2099,84 @@ mod tests {
...
+                _epoch_id: &StacksEpochId,
+            ) -> Option<StacksEpoch> {
+                self.get_stacks_epoch(0)
+            }
+            fn get_v1_unlock_height(&self) -> u32 {
+                u32::max_value()
+            }
+
+            fn get_pox_prepare_length(&self) -> u32 {
+                panic!("NullBurnStateDB should not return PoX info");
```
jcnelson, line 2149: This is `BlockLimitBurnStateDB`, not `NullBurnStateDB`.

src/core/mod.rs
```diff
@@ -85,11 +95,19 @@ pub const BITCOIN_MAINNET_FIRST_BLOCK_TIMESTAMP: u32 = 1610643248;
 pub const BITCOIN_MAINNET_FIRST_BLOCK_HASH: &str =
     "0000000000000000000ab248c8e35c574514d052a83dbc12669e19bc43df486e";
 pub const BITCOIN_MAINNET_INITIAL_REWARD_START_BLOCK: u64 = 651389;
+pub const BITCOIN_MAINNET_STACKS_2_05_BURN_HEIGHT: u64 = 713_000;
+
+// TODO: Pick a real height for Stacks 2.1.
+pub const BITCOIN_MAINNET_STACKS_21_BURN_HEIGHT: u64 = 2_000_000;
```
jcnelson, line 101: Can you just make this `u64::MAX`?  Same for `BITCOIN_TESTNET_STACKS_2_05_BURN_HEIGHT`

src/core/mod.rs
```diff
@@ -159,18 +202,20 @@ pub fn check_fault_injection(fault_name: &str) -> bool {
...
-#[derive(Debug, Clone, Eq, PartialEq, PartialOrd, Ord, Hash, Copy)]
+#[derive(Debug, Clone, Eq, PartialEq, PartialOrd, Ord, Hash, Copy, Deserialize)]
 pub enum StacksEpochId {
-    Epoch10 = 0x0100,
-    Epoch20 = 0x0200,
-    Epoch21 = 0x0201,
+    Epoch10 = 0x01000,
+    Epoch20 = 0x02000,
+    Epoch2_05 = 0x02005,
+    Epoch21 = 0x02010,
```
jcnelson, line 210: This should be `0x0200a`.  `0x0a` is 10, not `0x10`.

src/cost_estimates/pessimistic.rs
```diff
@@ -0,0 +1,336 @@
...
+    ) -> String {
+        let tx_descriptor = match tx {
+            TransactionPayload::TokenTransfer(..) => "stx-transfer".to_string(),
+            TransactionPayload::ContractCall(cc) => {
+                // Epoch key is "" before Epoch2_05 for backwards compatibility.
+                let epoch_marker = match evaluated_epoch {
+                    StacksEpochId::Epoch10 => "",
+                    StacksEpochId::Epoch20 => "",
+                    StacksEpochId::Epoch2_05 => ":2.05",
+                    // Note: Reuse 2.05 in 2.1.
```
jcnelson, line 233: This should be `:2.1`.  Estimates reset across epochs.

jcnelson (approved):
Lots of small things to fix and address, but this otherwise LGTM.  Thanks for doing all the legwork of getting this merged in!  Please answer my questions before merging.

gregorycoppola:
> Also, in the future, please don't commit things like `panic!("here we are")` or `let bt = backtrace::Backtrace::new(); warn!("process:bt {:?}", &bt);`. It's totally fine to do this locally, but it makes it that much harder to read the delta between the last merge commit and `HEAD`. These changes do not belong in the git history.

I didn't realize people were using the individual commits as part of a review.

I was planning to "squash and merge" into `next` and remove all these details.

I was using the commits to just be like checkpoints between local changes.

If the individual commits have meaning for reviewers, I can think about it in the future.

I guess there must be a command to collapse multiple commits.

src/vm/database/clarity_db.rs
kantai: Yep -- seconding this: the NullBurnStateDB should continue to panic here. Any test cases that are failing due to this should use the UnitTestBurnStateDB instead.

src/vm/types/serialization.rs
```diff
@@ -702,28 +702,6 @@ impl ClarityDeserializable<Value> for Value {
...
-    fn deserialize(input: &str) -> Self {
-        let bytes = hex_bytes(&input).expect("u32 deserialization: failed decoding bytes.");
-        assert_eq!(bytes.len(), 4);
-        u32::from_be_bytes(
-            bytes[0..4]
-                .try_into()
-                .expect("u32 deserialization: failed reading."),
-        )
-    }
-}
```
kantai: This needs to be preserved, and the `clarity_serializable!(u32);` macro invocation needs to be removed: these serialization methods are already used in Stacks 2.05, so switching how u32's are serialized in Clarity (which this change would do) would be consensus-breaking to 2.05.

src/vm/analysis/tests/costs.rs
```diff
@@ -198,33 +194,35 @@ pub fn test_tracked_costs(prog: &str, use_mainnet: bool, epoch: StacksEpochId) -
...
+        let cost = test_tracked_costs(test, use_mainnet, StacksEpochId::Epoch21);
         assert!(cost.exceeds(&baseline));
     }
 }
 
 #[test]
-fn test_all_mainnet() {
-    test_all(true)
+fn epoch_21_test_all_mainnet() {
+    epoch_21_test_all(true)
```
kantai: Can you make this merge additive, so that there's `epoch_21_test_all`, `epoch_20_test_all` and `epoch_205_test_all` in both mainnet and testnet versions?

src/vm/types/serialization.rs
jcnelson: Yikes, I can't believe I missed that.  Yeah, in general, *any* code that does serialization or deserialization needs to be treated as consensus-critical.

jcnelson:
Retracting approval until all consensus bugs are fixed

src/chainstate/stacks/db/blocks.rs
```diff
@@ -3920,20 +3974,30 @@ impl StacksChainState {
                         panic!("Clarity VM believes it was running in 1.0: pre-Clarity.")
                     }
                     StacksEpochId::Epoch20 => {
+                        assert_eq!(
+                            sortition_epoch.epoch_id,
+                            StacksEpochId::Epoch2_05,
+                            "Should only transition from Epoch20 to Epoch2_05"
+                        );
```
kantai, line 3981: I think here we need to match on `sortition_epoch.epoch_id`, e.g.:

```rust
match sortition_epoch.epoch_id {
    StacksEpochId::Epoch2_05 => {
        receipts.push(clarity_tx.block.initialize_epoch_2_05()?);
        applied = true;
    }
    StacksEpochId::Epoch21 => {
        receipts.push(clarity_tx.block.initialize_epoch_2_05()?);
        receipts.push(clarity_tx.block.initialize_epoch_2_1()?);
        applied = true;
    }
    _ => panic!("Bad Stacks epoch transition; parent_epoch = {}, current_epoch = {}", ...),
}
```

This is because a miner could try mining off of a very deep ancestor block (i.e., a 2.0 block) from the 2.1 epoch, and would need to apply _both_ epoch transitions in that block.

jcnelson: Seconded

src/vm/costs/mod.rs
kantai: Not yet -- the 2.10 branch uses `Unimplemented` for cost functions that haven't been specified yet. There's an open issue for resolving.

kantai (commented):
There's a couple issues that need to be addressed before being merged, but otherwise looks great. Thanks for taking this on, @gregorycoppola!

src/chainstate/burn/db/sortdb.rs
gregorycoppola: Something was going on where it wasn't sorting by epoch_id.

Why would it not be that the epochs are passed in in the correct order?

I could update the code, but we can also just leave it as it is and require them to be passed in in order.

Either way the method contract can be updated to reflect the input requirements.

src/clarity_vm/clarity.rs
gregorycoppola: done

src/vm/tests/costs.rs
```diff
@@ -876,9 +877,12 @@ fn test_all(use_mainnet: bool) {
     let baseline = test_tracked_costs("1", use_mainnet, StacksEpochId::Epoch20);
 
     for f in NativeFunctions::ALL.iter() {
-        let test = get_simple_test(f);
-        let cost = test_tracked_costs(test, use_mainnet, StacksEpochId::Epoch20);
-        assert!(cost.exceeds(&baseline));
+        // Note: The 2.05 test assumes Clarity1.
+        if f.get_version() == ClarityVersion::Clarity1 {
```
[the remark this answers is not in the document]
gregorycoppola: Functions running in Stacks 2.0 can't be Clarity2, because the ClarityVersion is sometimes derived from the epoch, c.f. ```
impl ClarityVersion {
    pub fn latest() -> ClarityVersion {
        ClarityVersion::Clarity2
    }
    pub fn default_for_epoch(epoch_id: StacksEpochId) -> ClarityVersion {
        match epoch_id {
            StacksEpochId::Epoch10 => {
                warn!("Attempted to get default Clarity version for Epoch 1.0 where Clarity does not exist");
                ClarityVersion::Clarity1
            }
            StacksEpochId::Epoch20 => ClarityVersion::Clarity1,
            StacksEpochId::Epoch2_05 => ClarityVersion::Clarity1,
            StacksEpochId::Epoch21 => ClarityVersion::Clarity2,
        }
    }
}
```

src/chainstate/burn/db/sortdb.rs
jcnelson: > Something was going on where it wasn't sorting by epoch_id.

That sounds like a bug in this PR, then.  It was working before, so you should find and fix the bug instead of deleting non-regressed code.

> Why would it not be that the epochs are passed in in the correct order?

Why should we trust that no one will ever screw up their end of the API contract, when we can easily defend against it?

src/cost_estimates/pessimistic.rs
jcnelson: We'll be adding new cost functions for new Clarity methods, and I can't promise that we won't change any existing cost functions.

src/vm/tests/costs.rs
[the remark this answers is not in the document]
jcnelson: > Functions running in Stacks 2.0 can't be Clarity2, because the ClarityVersion is sometimes derived from the epoch, c.f.

I don't understand what you're saying here.  All Clarity functions that worked in Stacks 2.0 should continue to work in all future epochs.

src/chainstate/burn/db/sortdb.rs
gregorycoppola: I think the different sort behavior is an artifact of there being new fields in this struct in the 2.05 branch.

I can take a deeper look if we are curious.

src/core/mod.rs
gregorycoppola: yikes

src/chainstate/burn/db/sortdb.rs
jcnelson: Right, this sounds indicative of something deeper.  Epochs should be totally ordered, and if something changed that removes this guarantee, then we have a problem.

src/cost_estimates/pessimistic.rs
gregorycoppola: guess it can't hurt much to make its own key... except for the fact that estimates will reset for a brief period.

src/vm/tests/costs.rs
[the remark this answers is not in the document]
gregorycoppola: Clarity2 functions didn't exist in Stacks 2.0.

They are going to start existing in Stacks 2.1.

src/vm/database/clarity_db.rs
gregorycoppola: I switched this back. We'll see which tests are failing on GitHub and follow up in another round.

gregorycoppola:
I pushed changes to most but not all comments in case anyone wants to review.

I kept the replies to one commit per comment addressed. cc @jcnelson 

The extra merges of "merge/nc2" into itself are because I was working in parallel from three different directories.

There will be failures due to `
reinstate panics in pox methods for NullBurnStateDB`.

src/chainstate/burn/db/sortdb.rs
gregorycoppola: reverting the change. I forget now which test it was that is broken so we'll see.

src/chainstate/stacks/db/blocks.rs
gregorycoppola: done

src/vm/database/clarity_db.rs
gregorycoppola: @kantai pushed the fixes for this so these tests are working now.

src/chainstate/burn/db/sortdb.rs
gregorycoppola: I reverted this change and no tests broke so I'm not sure what happened.

The test are all passing.

kantai (approved):
LGTM! Thanks for bearing with us on all the PR updates.

jcnelson (approved):
LGTM! Thanks @gregorycoppola

gregorycoppola:
Thanks for the reviews everyone! Probably harder to review than write.

PR #2959, opened 2021-11-30
feat: build multi-platform docker images

zone117x:
Build multi-architecture docker images -- specifically build `linux/arm64` images for improved support on Apple Silicon. 

See https://docs.docker.com/desktop/multi-arch/ and https://docs.docker.com/desktop/mac/apple-silicon/

> In summary, running Intel-based containers on Arm-based machines should be regarded as “best effort” only. We recommend running arm64 containers on Apple Silicon machines whenever possible, and encouraging container authors to produce arm64, or multi-arch, versions of their containers.


zone117x:
Well, it built the multi-arch stretch images successfully, but...
image


Compiling inside qemu is incredibly slow. I'm not sure what to do here. We should attempt some clever caching of intermediate compilation artifacts? @kantai @CharlieC3?

diwakergupta:
> Well, it built the multi-arch stretch images successfully, but... image
> 
> Compiling inside qemu is incredibly slow. I'm not sure what to do here. We could attempt some clever caching of intermediate compilation artifacts? @kantai @CharlieC3?

How critical is building these multi-platform Docker images? Can we generate these images only for releases? If so, slow builds might be less of a concern -- we can even separate it out into a workflow that has to be triggered manually.

My sense is we shouldn't be spending too many cycles trying to optimize this unless there's a strong reason to do so.

zone117x:
> How critical is building these multi-platform Docker images? Can we generate these images only for releases? 

Yep we only need them on release. Arm64 images are becoming more important as m1 Macs gain adoption. Right now the stretch image is broken on Docker for Mac (on m1) due to a qemu bug. The alpine image works but still runs in slow emulation mode.

CharlieC3:
@zone117x Updated the build to only produce an arm64 image on releases, so it should help with the build time elsewhere. However, we should be cognizant of the downside of this. Since we won't be building arm64 images regularly and often, this means we may encounter a build failure for that platform on an actual release since we wouldn't be able to detect it sooner.

In that case we'd simply have to patch and re-run the build, so not a major issue, just not something we should get comfortable doing.

zone117x:
@CharlieC3 should we also build multi-platform for the alpine image?

CharlieC3:
@zone117x Couldn't hurt! ✅

zone117x:
The main image (alpine) and debian-stretch image are both successfully built and pushed to Docker hub for x86_64 and arm64:
image


And I can confirm that the debian stretch image pushed by this PR is now working on an M1. The arm64 image is automatically picked in the API docker-compose setup, and is snappy.

```
INFO [1638549850.210417] [testnet/stacks-node/src/main.rs:81] [main] stacks-node 9dfbae4 (2959/merge:9dfbae4+, release build, linux [aarch64])
INFO [1638549850.210523] [testnet/stacks-node/src/main.rs:114] [main] Loading config at path /app/config/Stacks-dev.toml
INFO [1638549850.239107] [src/chainstate/stacks/db/mod.rs:837] [main] Migrating chainstate schema from version 1 to 2
INFO [1638549850.269175] [src/chainstate/stacks/db/mod.rs:955] [main] Building genesis block
INFO [1638549850.361399] [src/vm/functions/mod.rs:472] [main] "``... to be a completely separate network and separate block chain, yet share CPU power with Bitcoin`` - Satoshi Nakamoto"
WARN [1638549850.361590] [src/chainstate/stacks/db/mod.rs:1019] [main] Seeding 6 balances coming from the config
INFO [1638549850.362206] [src/chainstate/stacks/db/mod.rs:1043] [main] Importing accounts from Stacks 1.0
INFO [1638549850.362649] [src/chainstate/stacks/db/mod.rs:1066] [main] Seeding 24 balances coming from chain dump
INFO [1638549850.362661] [src/chainstate/stacks/db/mod.rs:1071] [main] Initializing chain with lockups
INFO [1638549850.363482] [src/chainstate/stacks/db/mod.rs:1116] [main] Initializing chain with namespaces
INFO [1638549850.363884] [src/chainstate/stacks/db/mod.rs:1190] [main] Initializing chain with names
INFO [1638549850.364852] [src/chainstate/stacks/db/mod.rs:1280] [main] Saving Genesis block. This could take a while
INFO [1638549850.378724] [src/chainstate/stacks/db/mod.rs:1348] [main] Computed Clarity state genesis, root_hash: b7f854e48f3691d719d2b823e3d62571255527134c92955ecf178d68f1434eec
INFO [1638549850.393287] [testnet/stacks-node/src/event_dispatcher.rs:888] [main] Registering event observer at: host.docker.internal:3700
INFO [1638549850.412594] [src/chainstate/burn/db/sortdb.rs:3692] [main] ACCEPTED(1) leader key register a17565bb827d6466ffcbef8bc0bdd650a12afa5c6914dd00c4c8c77602660827 at 1,1
INFO [1638549850.422935] [src/net/db.rs:166] [main] Will be authenticating p2p messages with public key: 025e3fc946d26ccb23b2c494cf418287ea9f5ed44e3783de4e5cd27a14115c7243
INFO [1638549850.439196] [testnet/stacks-node/src/node.rs:557] [main] Start HTTP server on: 0.0.0.0:20443
```

.github/workflows/stacks-blockchain.yml
```diff
@@ -139,23 +139,58 @@ jobs:
...
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            STACKS_NODE_VERSION=${{ github.event.inputs.tag || env.GITHUB_SHA_SHORT }}
+            GIT_BRANCH=${{ env.GITHUB_REF_SHORT }}
+            GIT_COMMIT=${{ env.GITHUB_SHA_SHORT }}
           # Only push if (a tag was passed in) or (we're building a non-master branch which isn't a PR)
-          push: ${{ github.event.inputs.tag != '' || (github.ref != 'refs/heads/master' && !contains(github.ref, 'refs/pull')) }}
+          # push: ${{ github.event.inputs.tag != '' || (github.ref != 'refs/heads/master' && !contains(github.ref, 'refs/pull')) }}
+          push: true # (TEMP) always push this branch
```
CharlieC3, line 193: @zone117x Before merging, let's just make sure this and the other get reverted :)

jcnelson:
* Might need a separate build process in order to speed up builds
* Benefits users of Graviton VMs, Raspberry Pis, and Apple M1s
* Can have a release-only build trigger that's manually deployed

.github/workflows/stacks-blockchain.yml
CharlieC3: @zone117x Is this ready to be reverted?

CharlieC3:
@kantai @jcnelson [Thoughts about this current setup?](https://github.com/stacks-network/stacks-blockchain/runs/4939061601?check_suite_focus=true) I extracted the additional docker platform build jobs into a separate workflow as to not clutter the main one. It's only triggered by the standard CI workflow when a tag is passed in, so there's no need to remember to trigger two workflows. Also, the second workflow is non-blocking for any jobs in the main workflow (e.g. dist builds, tests, releases, etc).

Also, the latest commit will fail because of [this line](https://github.com/stacks-network/stacks-blockchain/actions/runs/1747371010/workflow#L127) until this PR is merged to `master`. But the workflow run I linked above should be what a tagged workflow run will resemble after this is merged.

Some other changes made:
* Changed the name of the main workflow to just "CI". It was originally given the same name as the repository when GH Actions was much newer, and was a requirement for something hacky we were doing back then. It's not necessary anymore.
* Removed the Slack notify jobs from the CI workflow.
* Remove the old `v1-v2-migration.yml` workflow since it's not needed anymore.

wileyj (approved):

kantai (approved):
LGTM!

Issue #2822, opened 2021-08-25
Add stacks tip height to http headers


pavitthrap:
For the `v2/health` endpoint, a node will need to determine whether it is caught up with its peer nodes (its Stacks tip height should be comparable to its peers' tip height). 

This information is currently not readily available. 

Can have a node include a `X-Stacks-Tip` header in its responses, which would contain the tip block height and block hash. This requires modification to the node's HTTP server.

pavitthrap:
From blockchain meeting: 
We would include the stacks tip height within http headers for messages sent in the data place by default. 
We would also want to cache the tip height.

jcnelson:
This is now in `master`.

PR #2961, opened 2021-12-02
Feat/add stacks height to preamble

pavitthrap:
In order to implement the health check endpoint of a node (#2768), the node must have knowledge of its peer's stacks block height. By comparing the height of the tip of the stacks blockchain with that of its startup peers, a node can determine if it is "caught up" with the rest of the network, and is thus "healthy". 

In this PR, I added the canonical stacks tip height to the Preamble, which is sent along with every stacks message sent between peers. The node stores this received information in the `ConversationP2P` object for each peer it is in contact with. 

## Testing 
I updated relevant tests, did not add any new tests.

### DIFF
--- a/src/codec/mod.rs
+++ b/src/codec/mod.rs
@@ -165,7 +165,7 @@ pub const MAX_PAYLOAD_LEN: u32 = 1 + 16 * 1024 * 1024;
 pub const MAX_MESSAGE_LEN: u32 =
     MAX_PAYLOAD_LEN + (PREAMBLE_ENCODED_SIZE + MAX_RELAYERS_LEN * RELAY_DATA_ENCODED_SIZE);
 
-/// P2P preamble length (addands correspond to fields above)
+/// P2P preamble length (addands correspond to fields in `Preamble`)
 pub const PREAMBLE_ENCODED_SIZE: u32 = 4
     + 4
     + 4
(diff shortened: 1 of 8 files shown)


jcnelson (changes requested):
I'm sorry, but this approach is basically unworkable.  You can't change the preamble structure, since all peers assume it's a fixed-sized structure with pre-determined byte fields at pre-determined offsets in the buffer.

kantai:
> I'm sorry, but this approach is basically unworkable. You can't change the preamble structure, since all peers assume it's a fixed-sized structure with pre-determined byte fields at pre-determined offsets in the buffer.

Doesn't that just mean that a form of versioning would be necessary for the preamble structure?

jcnelson:
A couple things:

* When we originally discussed addressing the problem of communicating the stacks block height between nodes, we agreed to do it via the data plane (https://github.com/blockstack/stacks-blockchain/issues/2822) because doing so would be so much easier than trying to change the p2p wire formats.

* The `Preamble` is not meant to be versioned -- it must be decodable to all nodes. Instead, the `Preamble` struct has an `additional_data` field which is meant to be used as a pointer to an offset in the message payload buffer where to start parsing additional data in a backwards-compatible way (i.e. by adding a trailer to p2p messages beyond the end of the payload).  Then, old nodes would just ignore the trailer, while new nodes would parse it and use it somehow.

This PR does not address #2822 as originally discussed; moreover, even if we wanted to include the Stacks block height in p2p messages (and it's not clear to me that we do), it goes about it in the wrong way.

gregorycoppola (commented):
Thanks, @pavitthrap .

I don't have any new notes, but waiting to see how @jcnelson 's question is resolved.

pavitthrap:
Closed - will open new PR that adds the stacks height to the HTTP headers.

PR #2965, opened 2021-12-08
Chore: remove token from codecov.yml

kantai:
Codecov does not need a token to be provided when uploads are from a public repository:

https://docs.codecov.com/docs/frequently-asked-questions#where-is-the-repository-upload-token-found

This removes the codecov token (which is also a defunct token) from our codecov.yaml.


pavitthrap (approved):
lgtm!

jcnelson (approved):

Issue #2954, opened 2021-11-29
Improved fee estimation


kantai:
The current fee estimator looks at _mined blocks_, and ranks the transactions in the block by their fee rate, taking the 5th, 50th, and 95th percentiles, and uses those three values as fee rate estimates.

This can lead to some problems -- it seems like very high fee estimates can be commonly obtained when STX transfers are broadcasted with fee amounts around 1 STX (STX transfers are much cheaper operations than contract-calls, like 50-100x cheaper, so when most transfers use fees around 1 STX, they imply that contract-calls should be 50-100 STX).

timstackblock:
From todays meeting, here are some of the Test contracts from the test app. It would be great if these could returned some fee estimate data on testnet

Test Contract call
ST1X6M947Z7E58CNE0H8YJVJTVKS9VW0PHEG3NHN3.faker
Test NFT
ST9VQ21ZEGG54JDFE39B99ZBTSFSWMEC323MENFG.animal
Test BNS 
ST000000000000000000002AMW42H.bns
website
https://btc-website.tintashlabs.com/
Test stell transfer 
ST6G7N19FKNW24XH5JQ5P5WR1DN10QWMKQSPSTK7.stella-the-cat
Rocket tokens
ST1X6M947Z7E58CNE0H8YJVJTVKS9VW0PHEG3NHN3.dull-sapphire-bird

In Prod we test with the following coins:
USDC
MIA
DIKA

@Eshwari007 feel free to elaborate

Eshwari007:
> 

Thats quite accurate list of what we test in both Testnet and Mainnet along with contract deployment which we test via Testnet demo app https://stacks-wallet-web.vercel.app/ and also via the Sandbox in Explorer.QA also test BNS functions in Sandbox along with the contract deployment mentioned in first line above.Believe all of the use cases has fee estimate impact on it hence listing them all out for your perusal.

cc @timstackblock

kantai:
This is a graph of data from a stacks-node over the last two days. It shows the median fee rate (measured as ustx per 1/60kth of the block limit) in each Stacks block in blue, and the corresponding "middle" estimate in the node in purple.

<img width="1294" alt="median-rates" src="https://user-images.githubusercontent.com/1153552/144084402-8be4bf36-e141-40e4-a90e-8b5c053755c7.png">

[...]

PR #2966, opened 2021-12-08
feat: Improved Fee Rate Computation

gregorycoppola:
Addresses #2954

https://github.com/blockstack/stacks-blockchain/issues/2954#issuecomment-983017453 discusses four proposals for improving fee estimation.

We address these first three:

1. Address outlier block sensitivity by using a median of a window, rather than exponential windowing.
2. Address impact of STX transfers by using percentiles weighted by the proportion of the block limit consumed rather than 5th, 50th, and 95th percentiles. See weighted percentiles.
3. Address impact of non-full blocks by treating the non-full proportion of the block as a transaction at the relay fee minimum (a fee of 0 is too low for admission to the mempool).

These changes are tested with a suite of new unit tests.

### DIFF
--- a/src/cost_estimates/fee_medians.rs
+++ b/src/cost_estimates/fee_medians.rs
@@ -308,9 +308,7 @@ fn fee_rate_and_weight_from_receipt(
     block_limit: &ExecutionCost,
 ) -> Option<FeeRateAndWeight> {
     let (payload, fee, tx_size) = match tx_receipt.transaction {
-        TransactionOrigin::Stacks(ref tx) => {
-            Some((&tx.payload, tx.get_tx_fee(), tx.tx_len()))
-        }
+        TransactionOrigin::Stacks(ref tx) => Some((&tx.payload, tx.get_tx_fee(), tx.tx_len())),
         TransactionOrigin::Burn(_) => None,
     }?;
     let scalar_cost = match payload {
--- a/src/cost_estimates/fee_medians.rs
+++ b/src/cost_estimates/fee_medians.rs
@@ -53,7 +53,7 @@ pub struct WeightedMedianFeeRateEstimator<M: CostMetric> {
     /// We only look back `window_size` fee rates when averaging past estimates.
     window_size: u32,
     /// The weight of a "full block" in abstract scalar cost units. This is the weight of
-    /// a block that is filled on each dimension.
+    /// a block that is filled *one single* dimension.
     full_block_weight: u64,
     /// Use this cost metric in fee rate calculations.
     metric: M,
@@ -86,7 +86,7 @@ impl<M: CostMetric> WeightedMedianFeeRateEstimator<M> {
             db,
             metric,
             window_size,
-            full_block_weight: 6 * PROPORTION_RESOLUTION,
+            full_block_weight: PROPORTION_RESOLUTION,
         })
     }
 
--- a/src/cost_estimates/fee_medians.rs
+++ b/src/cost_estimates/fee_medians.rs
@@ -52,6 +52,9 @@ pub struct WeightedMedianFeeRateEstimator<M: CostMetric> {
     db: Connection,
     /// We only look back `window_size` fee rates when averaging past estimates.
     window_size: u32,
+    /// The weight of a "full block" in abstract scalar cost units. This is the weight of
+    /// a block that is filled on each dimension.
+    full_block_weight: u64,
     metric: M,
 }
 
@@ -82,6 +85,7 @@ impl<M: CostMetric> WeightedMedianFeeRateEstimator<M> {
             db,
             metric,
             window_size,
+            full_block_weight: 6 * PROPORTION_RESOLUTION,
         })
     }
 
@@ -208,13 +212,20 @@ impl<M: CostMetric> FeeEstimator for WeightedMedianFeeRateEstimator<M> {
             .collect();
 
         // If necessary, add the "minimum" fee rate to fill the block.
-        maybe_add_minimum_fee_rate(&mut working_fee_rates, PROPORTION_RESOLUTION);
+        maybe_add_minimum_fee_rate(&mut working_fee_rates, self.full_block_weight);
 
-        // Compute a FeeRateEstimate from the sorted, adjusted fee rates.
-        let block_estimate = fee_rate_esimate_from_sorted_weights(&working_fee_rates);
+        // If fee rates non-empty, then compute an update.
+        if working_fee_rates.len() > 0 {
+            working_fee_rates.sort_by(|a, b| {
+                a.fee_rate
+                    .partial_cmp(&b.fee_rate)
+                    .unwrap_or(Ordering::Equal)
+            });
 
-        // Update the running estimate using this rounds estimate.
-        self.update_estimate(block_estimate);
+            let block_estimate =
+                fee_rate_estimate_from_sorted_weighted_fees(&working_fee_rates);
+            self.update_estimate(block_estimate);
+        }
 
         Ok(())
     }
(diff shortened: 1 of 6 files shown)


src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,348 @@
...
+        Self::instantiate_db(&tx)?;
+        tx.commit()?;
+
+        Ok(Self {
+            db,
+            metric,
+            window_size,
+            full_block_weight: 6 * PROPORTION_RESOLUTION,
+        })
+    }
```
kantai, line 91: Can you also update the open behavior in `cost_estimates::fee_scalar::ScalarFeeRateEstimator` so that it behaves the same as this open (basically, on every open, invoke `instantiate_db` to check if the table exists). The reason to do this check rather than the one it currently does (`instantiate` only if `OPEN_CREATE` was necessary) is that the node configuration will use the same SQLite database for both of these estimators, but the estimators only create their own tables (so the database may exist, but not have the table for the opened estimator yet).

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,348 @@
...
+/// 3) "Pad" the block, so that any unused spaces is considered to have an associated fee rate of
+///    1f, the minimum. Ignoring the amount of empty space leads to over-estimates because it
+///    ignores the fact that there was still space in the block.
+pub struct WeightedMedianFeeRateEstimator<M: CostMetric> {
+    db: Connection,
+    /// We only look back `window_size` fee rates when averaging past estimates.
+    window_size: u32,
+    /// The weight of a "full block" in abstract scalar cost units. This is the weight of
+    /// a block that is filled on each dimension.
+    full_block_weight: u64,
```
kantai, line 57: I think that the "fill with empty" behavior is a little more honest if the "full" block weight is  a block with just _one_ dimension filled. Because a block cannot exceed any dimension, a block with a weight > a single dimension fill may not be able to accept any more transactions. Currently, I think this only impacts the docstring comment, but when it's integrated into the `testnet/stacks-node`, that will impact how this value is set.

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,348 @@
...
+///
+/// `sorted_fee_rates` must be non-empty.
+fn fee_rate_estimate_from_sorted_weighted_fees(
+    sorted_fee_rates: &Vec<FeeRateAndWeight>,
+) -> FeeRateEstimate {
+    if sorted_fee_rates.is_empty() {
+        panic!("`sorted_fee_rates` cannot be empty.");
+    }
+
+    let mut total_weight = 0u64;
```
kantai, line 245: You can avoid worrying about overflows in `total_weight` and `cumulative_weight` by using `f64` for them (especially because they get cast to f64 anyways)

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,348 @@
...
+    }
+    let mut cumulative_weight = 0u64;
+    let mut percentiles = Vec::new();
+    for rate_and_weight in sorted_fee_rates {
+        cumulative_weight += rate_and_weight.weight;
+        let percentile_n: f64 =
+            (cumulative_weight as f64 - rate_and_weight.weight as f64 / 2f64) / total_weight as f64;
+        percentiles.push(percentile_n);
+    }
+
```
kantai, line 257: Can you add an `assert_eq!` here for `percentiles.len()` and `sorted_fee_rates.len()`?

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,348 @@
...
+    let target_percentiles = vec![0.05, 0.5, 0.95];
+    let mut fees_index = 0; // index into `sorted_fee_rates`
+    let mut values_at_target_percentiles = Vec::new();
+    for target_percentile in target_percentiles {
+        while fees_index < percentiles.len() && percentiles[fees_index] < target_percentile {
+            fees_index += 1;
+        }
```
kantai, line 264: What happens when `percentiles[fees_index] == target_percentile`?

I think in the correct operation of the weighted percentile algorithm, `value` should equal `sorted_fee_rates[fee_index].fee_rate` in that case. Is that the case in this implementation?

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,348 @@
...
+        low: values_at_target_percentiles[0],
+    }
+}
+
+/// If the weights in `working_rates` do not add up to `full_block_weight`, add a new entry **in
+/// place** that takes up the remaining space.
+fn maybe_add_minimum_fee_rate(working_rates: &mut Vec<FeeRateAndWeight>, full_block_weight: u64) {
+    let mut total_weight = 0u64;
+    for rate_and_weight in working_rates.into_iter() {
+        total_weight += rate_and_weight.weight;
```
kantai, line 292: You should avoid an overflow here by doing checked_add:

```rust
total_weight = match total_weight.checked_add(rate_and_weight.weight) {
   Some(result) => result
   None => return,
};
```

Which just returns from this function if there's an overflow (because if there's an overflow the block definitely does not need any filling help).

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,348 @@
...
+        //  which would not have created the necessary table for this estimator.
+        let tx = tx_begin_immediate_sqlite(&mut db)?;
+        Self::instantiate_db(&tx)?;
+        tx.commit()?;
+
+        Ok(Self {
+            db,
+            metric,
+            window_size,
+            full_block_weight: 6 * PROPORTION_RESOLUTION,
```
kantai, line 89: Regarding my comment above about the block weight to use -- I think this should just be a single dimension fill (so just `PROPORTION_RESOLUTION`), because a block with that weight may not be able to add any more transactions.

kantai (commented):
This looks mostly good to me, I had a couple comments in the current implementation, and I think you should have a test case for some of the corner cases of weighted percentiles (i.e., the percentile requested is exactly equal to one of the percentiles, explicit test cases for capturing too-low of a target percentile and too-high of a target percentile).

Additionally, you should integrate this fee estimator into the testnet/stacks-node configuration: the `config.rs` file handles instantiating estimators, and this new estimator should be added there (and probably set to the default fee estimator). And please also add an entry to the `CHANGELOG.md`, under a new `## [Unreleased]` subsection.

src/cost_estimates/fee_medians.rs
gregorycoppola: I exposed `fee_rate_estimate_from_sorted_weighted_fees` for test, and added test cases in `src/cost_estimates/tests/fee_medians.rs`.

@kantai I just put the answers from the program in to the right-hand side of the "expect".. What do you think?

src/cost_estimates/fee_medians.rs
gregorycoppola: done

src/cost_estimates/fee_medians.rs
gregorycoppola: copied.. we'll see if it causes any regressions in tests.

src/cost_estimates/fee_medians.rs
gregorycoppola: done in latest

src/cost_estimates/fee_medians.rs
gregorycoppola: done in latest

src/cost_estimates/fee_medians.rs
gregorycoppola: dropped the 6. some of the parameters in the tests changed, because the blocks got different amounts full.

src/cost_estimates/fee_medians.rs
gregorycoppola: changed the text to say its any dimension

gregorycoppola:
see https://github.com/blockstack/stacks-blockchain/pull/2972

PR #2968, opened 2021-12-09
feat: Fuzzing for FeeRate Estimates

### LINKED ISSUE #2954: Improved fee estimation

gregorycoppola:
Addresses #2954

We are implementing the four points in https://github.com/blockstack/stacks-blockchain/issues/2954#issuecomment-983017453

#2966 addresses the first three points.

Here, we address:
4. We should address fee changes in the face of congestion using a strategy of **"fuzzing"** the estimated fee rate: when queried, nodes will add some random noise to the estimated fee rate. In periods of congestion, clients that used the higher fee would be included in the next block, pushing the fee estimate for the next block up.

This is tested just by the new unit tests.

### DIFF
--- a/src/cost_estimates/fee_rate_fuzzer.rs
+++ b/src/cost_estimates/fee_rate_fuzzer.rs
@@ -32,43 +32,61 @@ use rand::RngCore;
 use rand::SeedableRng;
 
 pub struct FeeRateFuzzer {
+    /// We will apply a random "fuzz" on top of the estimates given by this.
     underlying: Box<dyn FeeEstimator>,
-    rng: Box<dyn RngCore>,
-}
-
-fn fuzz_esimate(input: &FeeRateEstimate) -> FeeRateEstimate {
-    input.clone()
+    /// Creator function for a new random generator. For prod, use `thread_rng`. For test,
+    /// pass in a contrived generator.
+    rng_creator: Box<dyn Fn() -> Box<dyn RngCore>>,
 }
 
 impl FeeRateFuzzer {
-    /// To get strong random numbers, pass in None for `seed`.
-    /// To get predictable random numbers for test, pass in a non-empty numeric `seed`.
-    pub fn new(underlying: Box<dyn FeeEstimator>, seed: Option<[u8; 32]>) -> FeeRateFuzzer {
-        let rng: Box<dyn RngCore> = match seed {
-            Some(seed) => {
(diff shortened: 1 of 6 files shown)


jcnelson:
How to test:

* During high congestion, we want users with higher fees to be accepted with higher probability.
* The integration test we need is to submit enough transactions to congest the network, and we're able to measure the average? median? fee rate going up.

gregorycoppola:
> How to test:
> 
>     * During high congestion, we want users with higher fees to be accepted with higher probability.
For the integration test, I plan to mostly just test that the program:
1) starts
2) allows us to call the right estimator

I didn't plan to validate the entire strategy round-trip.

We can add that.


> 
>     * The integration test we need is to submit enough transactions to congest the network, and we're able to measure the average? median? fee rate going up.

gregorycoppola:
Well, we *could* add an experiment for the round-trip working.

But, we don't need an integration test for that, I think. We just need to verify:
1) From the unit tests, any calculations we want
2) From the integration test, that we can boot up with the estimator we wanted

kantai:
> Well, we could add an experiment for the round-trip working.
> But, we don't need an integration test for that, I think. We just need to verify:
>
>    From the unit tests, any calculations we want
>    From the integration test, that we can boot up with the estimator we wanted

I think we do need an integration test. The behavior we're trying to measure is that (1) the fuzzer introduces a distribution into client's fee rates, (2) based on this distribution, the default miner will prefer the higher fee rate transactions, and (3) the assembled blocks will cause fee rates estimates to increase.

Testing that requires testing that all the pieces of this (miner, fee estimator, fee estimation RPC, and mempool iterator) work together to get the behavior we want: that's an integration test. It may be possible to use the `TestPeer` based testing structure to achieve this as well, but that's similarly a kind of integration test.

gregorycoppola:
see https://github.com/blockstack/stacks-blockchain/pull/2972

PR #2972, opened 2021-12-14
feat: Improved Fee Rate Computation

### LINKED ISSUE #2954: Improved fee estimation

gregorycoppola:
Fixes #2954

https://github.com/blockstack/stacks-blockchain/issues/2954#issuecomment-983017453 discusses four proposals for improving fee estimation.

We address these here:

1. Address outlier block sensitivity by using a median of a window, rather than exponential windowing.
2. Address impact of STX transfers by using percentiles weighted by the proportion of the block limit consumed rather than 5th, 50th, and 95th percentiles. See weighted percentiles.
3. Address impact of non-full blocks by treating the non-full proportion of the block as a transaction at the relay fee minimum (a fee of 0 is too low for admission to the mempool).
4. We should address fee changes in the face of congestion using a strategy of **"fuzzing"** the estimated fee rate: when queried, nodes will add some random noise to the estimated fee rate. In periods of congestion, clients that used the higher fee would be included in the next block, pushing the fee estimate for the next block up.

## Testing
* There are new unit tests for the two FeeEstimator components, to test the computations.
* There is an integration in `neon_integrations.rs` that tests that we can load this new estimator, and use it.


testnet/stacks-node/src/config.rs
```diff
@@ -42,6 +44,7 @@ pub struct ConfigFile {
     pub ustx_balance: Option<Vec<InitialBalanceFile>>,
     pub events_observer: Option<Vec<EventObserverConfigFile>>,
     pub connection_options: Option<ConnectionOptionsFile>,
+    // Note: put the name here
```
kantai, line 47: What does this note refer to?

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,352 @@
...
+    sorted_fee_rates: &Vec<FeeRateAndWeight>,
+) -> FeeRateEstimate {
+    if sorted_fee_rates.is_empty() {
+        panic!("`sorted_fee_rates` cannot be empty.");
+    }
+
+    let mut total_weight = 0f64;
+    for rate_and_weight in sorted_fee_rates {
+        total_weight += rate_and_weight.weight as f64;
+    }
```
kantai, line 248: Can you add a check if `total_weight` is `<= 0`? I think in that case, the estimator should throw an `EstimatorError`.

src/cost_estimates/fee_rate_fuzzer.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /// Add a uniform fuzz to input.
+    fn fuzz_estimate(&self, input: &FeeRateEstimate) -> FeeRateEstimate {
+        FeeRateEstimate {
+            high: self.fuzz_individual_scalar(input.high),
+            middle: self.fuzz_individual_scalar(input.middle),
+            low: self.fuzz_individual_scalar(input.low),
+        }
+    }
```
kantai, line 81: the `high` `middle` and `low` estimates should always use the same fuzz amount -- otherwise, the estimator could easily produce a `high` estimate that is lower than the `middle` estimate, etc.

kantai:
```suggestion
    fn fuzz_estimate(&self, input: FeeRateEstimate) -> FeeRateEstimate {
        let mut rng = (self.rng_creator)();
        let uniform = Uniform::new(-self.uniform_fuzz_bound, self.uniform_fuzz_bound);
        let fuzz_scale = 1f64 + uniform.sample(&mut rng);
        input * fuzz_scale
    }
```

src/cost_estimates/fee_rate_fuzzer.rs
```diff
@@ -0,0 +1,101 @@
...
+    ) -> Result<(), EstimatorError> {
+        self.underlying.notify_block(receipt, block_limit)
+    }
+
+    /// Call underlying estimator and add some fuzz.
+    fn get_rate_estimates(&self) -> Result<FeeRateEstimate, EstimatorError> {
+        match self.underlying.get_rate_estimates() {
+            Ok(underlying_estimate) => Ok(self.fuzz_estimate(&underlying_estimate)),
+            Err(e) => Err(e),
+        }
```
kantai, line 99: If you're just going to return the error, the try syntax in rust is generally preferred, i.e.,

```suggestion
        let underlying_estimate = self.underlying.get_rate_estimates()?;
        Ok(self.fuzz_estimate(&underlying_estimate)
```

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -5847,3 +5873,199 @@ fn atlas_stress_integration_test() {
...
+
+/// This is an integration test for the "fuzzed_weighted_median_fee_rate" estimation.  It tests that
+/// 1) We can load up the "fuzzed_weighted_median_fee_rate" fee estimator.
+/// 2) We get results that look "somewhat right", but we can't expect exact values due to
+///    pseudo-random fuzz.
+/// We test that, if fees are increasing every block round for the same transaction, that we
+/// expect to see the fee rate estimates rise on every round, modulo the random noise.
+#[test]
+#[ignore]
+fn fuzzed_median_fee_rate_estimation_test() {
```
kantai, line 5916: Can you add this test to the github integration tests workflow?

testnet/stacks-node/src/config.rs
```diff
@@ -1202,19 +1210,39 @@ impl FeeEstimationConfig {
...
+        &self,
+        mut chainstate_path: PathBuf,
+        metric: CM,
+    ) -> Box<dyn FeeEstimator> {
+        if let Some(FeeEstimatorName::FuzzedWeightedMedianFeeRate) = self.fee_estimator.as_ref() {
+            chainstate_path.push("fee_fuzzed_weighted_median.sqlite");
+            let underlying_estimator =
+                WeightedMedianFeeRateEstimator::open(&chainstate_path, metric, 5)
+                    .expect("Error opening fee estimator");
+            Box::new(FeeRateFuzzer::new(underlying_estimator, 0.5))
```
kantai, line 1241: I think a tighter range than +/- 50% should be used for the fuzzer -- I think something as low as 10% would still work to be responsive during congestion.

kantai: Better still, can this be a configuration variable, with default = 10%?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -5847,3 +5873,199 @@ fn atlas_stress_integration_test() {
...
+
+            let body = json!({ "transaction_payload": payload_hex.clone() });
+
+            let client = reqwest::blocking::Client::new();
+            let fee_rate_result = client
+                .post(&path)
+                .json(&body)
+                .send()
+                .expect("Should be able to post")
+                // .json::<serde_json::Value>()
```
pavitthrap, line 6043: remove line here

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -5847,3 +5873,199 @@ fn atlas_stress_integration_test() {
...
+
+    // Check that:
+    // 1) The cost is always the same.
+    // 2) The estimated fee rate only either i) moves up, ii) stays within random noise of where it
+    //    was.
+    for i in 1..response_estimated_costs.len() {
+        let curr_rate = response_top_fee_rates[i] as f64;
+        let last_rate = response_top_fee_rates[i - 1] as f64;
+        // Check that either 1) curr_rate is bigger, or 2) is only smaller by random 11% noise.
+        assert!(curr_rate - last_rate > -1.0 * curr_rate * 0.11);
```
pavitthrap, line 6062: I'm a little confused by the reasoning behind this assertion - could you explain?

src/cost_estimates/tests/fee_medians.rs
```diff
@@ -0,0 +1,425 @@
...
+    let mut path = env::temp_dir();
+    let random_bytes = rand::thread_rng().gen::<[u8; 32]>();
+    path.push(&format!("fee_db_{}.sqlite", &to_hex(&random_bytes)[0..8]));
+
+    let window_size = 5;
+    WeightedMedianFeeRateEstimator::open(&path, m, window_size)
+        .expect("Test failure: could not open fee rate DB")
+}
+
+fn make_block_receipt(tx_receipts: Vec<StacksTransactionReceipt>) -> StacksEpochReceipt {
```
pavitthrap, line 58: Is this function a dupe? Pretty sure I saw this in a different file.

pavitthrap (commented):
just some minor feedback!

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,352 @@
...
+        total_weight = match total_weight.checked_add(rate_and_weight.weight) {
+            Some(result) => result,
+            None => return,
+        };
+    }
+
+    if total_weight < full_block_weight {
+        let weight_remaining = full_block_weight - total_weight;
+        working_rates.push(FeeRateAndWeight {
+            fee_rate: 1f64,
```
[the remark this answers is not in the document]
gregorycoppola: good idea

src/cost_estimates/tests/common.rs
```diff
@@ -0,0 +1,50 @@
...
+use vm::costs::ExecutionCost;
+
+use crate::chainstate::stacks::{
+    CoinbasePayload, StacksTransaction, TokenTransferMemo, TransactionAuth,
+    TransactionContractCall, TransactionPayload, TransactionSpendingCondition, TransactionVersion,
+};
+use crate::core::StacksEpochId;
+
+/// Make a block receipt from `tx_receipts` with some dummy values filled for test.
+pub fn make_block_receipt(tx_receipts: Vec<StacksTransactionReceipt>) -> StacksEpochReceipt {
```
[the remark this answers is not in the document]
gregorycoppola: done

gregorycoppola (commented):
Thanks for the thoughtful reviews. I left two conversation issues open.

gregorycoppola:
I added to the CHANGELOG, but I left the README change until we have decided how this feature will be used.

Should we also be updating default configs?

kantai:
> I added to the CHANGELOG, but I left the README change until we have decided how this feature will be used.

I think that the point of README changes is to document the new configuration settings. It's not necessarily important to figure out how this feature will be used first -- the README should just describe how to alter the fee rate configuration from the existing method to the new method.

> Should we also be updating default configs?

Not yet -- I think that on the first release, the default configs _shouldn't_ change, because we want to be able to test the fee rate changes without changing the default behavior of other nodes on the network.

testnet/stacks-node/src/config.rs
kantai: Yes, we could make both fields configurable by adding two fields to the `FeeEstimationConfig`:

```
fee_rate_fuzzer_fraction: f64,
fee_rate_window_size: u64,
```

And these fields would only ever be used by the `WeightedMedianFeeRateEstimator` and `FeeRateFuzzer`.

In order for these to be configurable via the config file, the fields would also need to be added to the `FeeEstimationConfigFile` struct.

testnet/stacks-node/src/tests/neon_integrations.rs
gregorycoppola: This comment has been addressed in the latest revision, by using the new config parameters to simplify the test.

testnet/stacks-node/src/config.rs
gregorycoppola: I've added configuration options for this.

gregorycoppola (commented):
1) added config options so this can be specified from file

2) simplified the integration tests, using the config options

3) added to the README

CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
...
+
+### Changed
+
+- A new fee estimator intended to produce fewer over-estimates, by having less
+  sensitivity to outliers. Its characteristic features are: 1) use a window to
+  forget past estimates instead of exponential averaging, 2) use weighted
+  percentiles, so that bigger transactions influence the estimates more, 3)
+  assess empty blocks as having paid the "minimum fee", so that empty space is
+  accounted for, 4) use random "fuzz" so that in busy times we will not have
+  ties. We will now default to this estimator.
```
kantai, line 18: We're not defaulting to this estimator yet.

README.md
```diff
@@ -378,6 +380,11 @@ are **not** consensus-critical components, but rather can be used by miners to
 rank transactions in the mempool or client to determine appropriate fee rates
 for transactions before broadcasting them.
 
+The default estimator is now `fuzzed_weighted_median_fee_rate`. This uses a
```
kantai, line 383:
```suggestion
The `fuzzed_weighted_median_fee_rate` uses a
```

README.md
```diff
@@ -378,6 +380,11 @@ are **not** consensus-critical components, but rather can be used by miners to
...
+median estimate from a window the past `fee_rate_window_size` fees paid.
```
kantai, line 384:
```suggestion
median estimate from a window of the fees paid in the last `fee_rate_window_size` blocks.
```

testnet/stacks-node/src/config.rs
```diff
@@ -1235,10 +1246,16 @@ impl FeeEstimationConfig {
...
         if let Some(FeeEstimatorName::FuzzedWeightedMedianFeeRate) = self.fee_estimator.as_ref() {
             chainstate_path.push("fee_fuzzed_weighted_median.sqlite");
-            let underlying_estimator =
-                WeightedMedianFeeRateEstimator::open(&chainstate_path, metric, 5)
-                    .expect("Error opening fee estimator");
-            Box::new(FeeRateFuzzer::new(underlying_estimator, 0.5))
+            let underlying_estimator = WeightedMedianFeeRateEstimator::open(
+                &chainstate_path,
+                metric,
+                self.fee_rate_window_size.try_into().unwrap(),
```
kantai, line 1252: Can you use `expect` to add an error message like `"Configured fee rate window size out of bounds"`?

gregorycoppola: ok

testnet/stacks-node/src/config.rs
```diff
@@ -1455,6 +1472,8 @@ pub struct FeeEstimationConfigFile {
     pub cost_metric: Option<String>,
     pub disabled: Option<bool>,
     pub log_error: Option<bool>,
+    pub fee_rate_fuzzer_fraction: f64,
+    pub fee_rate_window_size: u64,
```
kantai, line 1476: I think it may make more sense for these to be `Option<>` fields, then if they're `None`, the `FeeEstimationConfig` struct should use _it's_ default values for these fields.

kantai: I believe if you do not use the `Option` types here, the config parser will always expect the field to be present in the file, see this playground gist: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=32c5c340dde70bf7b210e13ffc95da8b

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -6085,19 +6083,278 @@ fn fuzzed_median_fee_rate_estimation_test() {
...
+fn fuzzed_median_fee_rate_estimation_test_window10() {
+    fuzzed_median_fee_rate_estimation_test(10, 90080.5496)
+}
+
+#[test]
+#[ignore]
+fn use_latest_tip_integration_test() {
+    // The purpose of this test is to check if setting the query parameter `tip` to `latest` is working
+    // as expected. Multiple endpoints accept this parameter, and in this test, we are using the
+    // GetContractSrc method to test it.
```
kantai, line 6126: Why is this new test here? Is this testing some new feature for this PR or is it just adding coverage for something previously not covered?

gregorycoppola: something weird is going on with git/github here because this is in `develop` I'm pretty sure

kantai: Ah, yeah, you're right.

testnet/stacks-node/src/config.rs
gregorycoppola: marking as resolved because the discussion points have moved elsewhere

testnet/stacks-node/src/config.rs
gregorycoppola: I switched these to toptions but now the questions is, is the "default" handling correct?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -6085,19 +6083,278 @@ fn fuzzed_median_fee_rate_estimation_test() {
...
-    // 2) The estimated fee rate only either i) moves up, ii) stays within random noise of where it
-    //    was.
+    // 2) Only grows.
     for i in 1..response_estimated_costs.len() {
         let curr_rate = response_top_fee_rates[i] as f64;
         let last_rate = response_top_fee_rates[i - 1] as f64;
-        // Check that either 1) curr_rate is bigger, or 2) is only smaller by random 11% noise.
-        assert!(curr_rate - last_rate > -1.0 * curr_rate * 0.11);
+        // Check that the rate is growing monotonically.
+        assert!(curr_rate >= last_rate);
```
gregorycoppola, line 6091: need to investigate whether this is going to be flaky.. might need to wait a few extra rounds to be sure that the tx's have all been incorporated

testnet/stacks-node/src/config.rs
```diff
@@ -1202,19 +1221,49 @@ impl FeeEstimationConfig {
...
+        mut chainstate_path: PathBuf,
+        metric: CM,
+    ) -> Box<dyn FeeEstimator> {
+        if let Some(FeeEstimatorName::FuzzedWeightedMedianFeeRate) = self.fee_estimator.as_ref() {
+            chainstate_path.push("fee_fuzzed_weighted_median.sqlite");
+            let underlying_estimator = WeightedMedianFeeRateEstimator::open(
+                &chainstate_path,
+                metric,
+                self.fee_rate_window_size
+                    .expect("Must set `fee_rate_window_size`.")
```
gregorycoppola, line 1253: I put an `expect` in here.. but where would the defaults go?

gregorycoppola (commented):
Added some questions of my own.

testnet/stacks-node/src/config.rs
kantai: The most straight-forward way to handle the defaults is probably to change the `FeeEstimationConfig` struct fields back to non-optional, and then update the `impl From<FeeEstimationConfigFile> for FeeEstimationConfig` method (lines 1166 and 1167) to something like:

```rust
            fee_rate_fuzzer_fraction: f.fee_rate_fuzzer_fraction.unwrap_or(0.1f64),
            fee_rate_window_size: f.fee_rate_window_size.unwrap_or(5),
```

That way, if the config _file_ doesn't define those values, the parsed config struct uses the defaults.

kantai (approved):
This LGTM! Just make sure that tests are passing on this PR.

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,359 @@
...
+        let sql =
+            "SELECT high, middle, low FROM median_fee_estimator ORDER BY measure_key DESC LIMIT ?";
+        let mut stmt = conn.prepare(sql).expect("SQLite failure");
+
+        // shuttle high, low, middle estimates into these lists, and then sort and find median.
+        let mut highs = Vec::with_capacity(window_size as usize);
+        let mut mids = Vec::with_capacity(window_size as usize);
+        let mut lows = Vec::with_capacity(window_size as usize);
+        let results = stmt
+            .query_and_then::<_, SqliteError, _, _>(&[window_size], |row| {
```
jcnelson, line 122: In rusqlite, you can get columns by name instead of index.  It's generally safer to do this, since it's far less brittle.

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,359 @@
...
+}
+
+/// Computes a `FeeRateEstimate` based on `sorted_fee_rates` using a "weighted percentile" method
+/// described in https://en.wikipedia.org/wiki/Percentile#Weighted_percentile
+///
+/// The percentiles computed are [0.05, 0.5, 0.95].
+///
+/// `sorted_fee_rates` must be non-empty.
+pub fn fee_rate_estimate_from_sorted_weighted_fees(
+    sorted_fee_rates: &Vec<FeeRateAndWeight>,
```
jcnelson, line 241: Style nit: we generally prefer `&[T]` over `&Vec<T>` if possible.

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,359 @@
...
+    if sorted_fee_rates.is_empty() {
+        panic!("`sorted_fee_rates` cannot be empty.");
+    }
+
+    let mut total_weight = 0f64;
+    for rate_and_weight in sorted_fee_rates {
+        total_weight += rate_and_weight.weight as f64;
+    }
+
+    if total_weight <= 0f64 {
```
jcnelson, line 252: Can this just be an `assert!`?

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,359 @@
...
+        low: values_at_target_percentiles[0],
+    }
+}
+
+/// If the weights in `working_rates` do not add up to `full_block_weight`, add a new entry **in
+/// place** that takes up the remaining space.
+fn maybe_add_minimum_fee_rate(working_rates: &mut Vec<FeeRateAndWeight>, full_block_weight: u64) {
+    let mut total_weight = 0u64;
+    for rate_and_weight in working_rates.into_iter() {
+        total_weight = match total_weight.checked_add(rate_and_weight.weight) {
```
jcnelson, line 300: Could we realistically encounter an overflow here?  If so, would it make more sense to use `.saturating_add()`?

src/cost_estimates/tests/fee_medians.rs
```diff
@@ -0,0 +1,391 @@
...
+use crate::cost_estimates::FeeRateEstimate;
+use crate::types::chainstate::StacksAddress;
+use crate::vm::types::{PrincipalData, StandardPrincipalData};
+use crate::vm::Value;
+use cost_estimates::fee_medians::fee_rate_estimate_from_sorted_weighted_fees;
+use cost_estimates::fee_medians::FeeRateAndWeight;
+use cost_estimates::tests::common::*;
+
+/// Tolerance for approximate comparison.
+const error_epsilon: f64 = 0.1;
```
jcnelson, line 35: If we're comparing fee rates, then won't this be a somewhat big epsilon for when blocks aren't full?  i.e. the fee rate in this case would be small.

pavitthrap:
Will review again after you address Jude's comments!

src/cost_estimates/tests/fee_medians.rs
```diff
@@ -0,0 +1,391 @@
...
+            make_dummy_cc_tx(i * 10 * half_operation_cost_basis, &half_operation_cost),
+            make_dummy_cc_tx(i * 10 * half_operation_cost_basis, &half_operation_cost),
+        ]);
+
+        estimator
+            .notify_block(&single_tx_receipt, &block_limit)
+            .expect("Should be able to process block receipt");
+    }
+
+    // The fee should be 30, because it's the median of [10, 20, .., 50].
```
jcnelson, line 243: Do you want to use a tighter epsilon here, if we're certain it's 30?

testnet/stacks-node/src/config.rs
```diff
@@ -1107,6 +1112,12 @@ pub struct FeeEstimationConfig {
     pub fee_estimator: Option<FeeEstimatorName>,
     pub cost_metric: Option<CostMetricName>,
     pub log_error: bool,
+    /// If using FeeRateFuzzer, the amount of random noise, as a percentage of the base value (in
```
jcnelson, line 1115: Will it be problematic if different nodes use different fuzzer fractions?  I'm inclined to think "no, unless they're wildly different," but I'm curious to hear your thoughts.

testnet/stacks-node/src/config.rs
```diff
@@ -1202,19 +1221,47 @@ impl FeeEstimationConfig {
...
+
+    // Creates a fuzzed WeightedMedianFeeRateEstimator with window_size 5. The fuzz
+    // is uniform with bounds [+/- 0.5].
+    pub fn make_fuzzed_weighted_median_fee_estimator<CM: 'static + CostMetric>(
+        &self,
+        mut chainstate_path: PathBuf,
+        metric: CM,
+    ) -> Box<dyn FeeEstimator> {
+        if let Some(FeeEstimatorName::FuzzedWeightedMedianFeeRate) = self.fee_estimator.as_ref() {
+            chainstate_path.push("fee_fuzzed_weighted_median.sqlite");
```
jcnelson, line 1248: So, I see that there are now several fee estimation databases in the top-level chainstate directory.  It doesn't have to be done in this PR, but can we at least move these databases to their own `fee_estimators` directory to de-clutter the chainstate layout?

testnet/stacks-node/src/config.rs
gregorycoppola: I'm kind of unclear on what would happen if we didn't do the random fuzzing now, actually.

If people submitted the same fees, would that clog the system or just mean that the miners had to choose who wins arbitrarily?

Because, right now we are going to decide who gets which random fuzz arbitrarily. So, both would be arbitrary.

src/cost_estimates/tests/fee_medians.rs
gregorycoppola: I was being kind of lazy in not just making this a reusable util function.

We should make a reusable `is_close` for floats, and make epsilon to the base value, i.e., `abs(a-b)/a < eps`... then make it 0.001 or something.

Then, this derivative is_close can use the float one.

testnet/stacks-node/src/config.rs
kantai: The goal of fuzzing isn't to break ties -- it's so that in periods of congestion, there's a set of higher fee rate transactions in the mempool, which miners will end up preferring, which will in turn cause fee estimates to go up naturally.

kantai: > Will it be problematic if different nodes use different fuzzer fractions? I'm inclined to think "no, unless they're wildly different," but I'm curious to hear your thoughts.

No, even if they're wildly different. These fee estimates are only used by clients to determine how to price their transactions. If someone decides to run a wildly fuzzy node, and use it for their transactions, _their_ transactions might be priced entirely wrong, but it'd have little to no impact on other users. Setting the fuzz setting correctly for commonly queried nodes would be important though.

gregorycoppola: Oh right. Yeah that makes sense.

testnet/stacks-node/src/config.rs
gregorycoppola: done

gregorycoppola (commented):
replying to @jcnelson

src/cost_estimates/fee_medians.rs
jcnelson: `&Vec<T>` will deref to `&[T]`, but `[T; N]` can't be coerced into `&Vec<T>` without a `clone()` (i.e. via `.to_vec()`).

src/cost_estimates/fee_medians.rs
jcnelson: Per the blockchain meeting, the code with `.checked_add()` is fine.  I didn't know that `.saturating_add()` wasn't yet stable for `f64`.  In general, if you're dealing with data whose values can be influenced by the network, you will want to use checked math operations (otherwise, you're inviting a DoS vulnerability).  "Basically impossible" isn't nearly as strong of a guarantee as "impossible" to someone who's motivated to make an unlikely but catastrophic thing happen.

src/cost_estimates/tests/fee_medians.rs
jcnelson: The code itself is fine for now; I was only commenting on the `epsilon` value.

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,355 @@
...
+        }
+        TransactionPayload::PoisonMicroblock(_, _)
+        | TransactionPayload::ContractCall(_)
+        | TransactionPayload::SmartContract(_) => {
+            // These transaction payload types all "work" the same: they have associated ExecutionCosts
+            // and contibute to the block length limit with their tx_len
+            metric.from_cost_and_len(&tx_receipt.execution_cost, &block_limit, tx_size)
+        }
+    };
+    let denominator = if scalar_cost >= 1 {
```
jcnelson, line 338: Should this be `if (scalar_cost as f64) >= MINIMUM_TX_FEE_RATE` here instead, if the point is to simply clamp `denominator` to be at least `MINIMUM_TX_FEE_RATE`?

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,355 @@
...
+        scalar_cost as f64
+    } else {
+        MINIMUM_TX_FEE_RATE
+    };
+    let fee_rate = fee as f64 / denominator;
+    if fee_rate >= MINIMUM_TX_FEE_RATE && fee_rate.is_finite() {
```
jcnelson, line 344: Wait, what happens if `fee_rate` is `inf`?  Surely, the `fee_rate` field in `FeeRateAndWeight` should simply be a large but finite number then, right?  Right now, this code handles this case by setting `fee_rate` to `MINIMUM_TX_FEE_RATE`, per this `if`-condition.

jcnelson: I think that if you have a `fee_rate` that is `inf`, you should set the `.fee_rate` field to the largest finite `f64` value.

jcnelson (approved):
This mostly LGTM.  Just had two comments about the fee math corner cases that must be addressed before merging. Thanks @gregorycoppola!

src/cost_estimates/fee_medians.rs
gregorycoppola: ok

src/cost_estimates/fee_medians.rs
gregorycoppola: I think intention is that it should be positive, rather than `0`. I don't think it refers to `minimum_fee_rate`. Line 341 should probably be changed to NOT be `MINIMUM_TX_FEE_RATE`, maybe that is the mistake.

We could instead use `scalar_cost > 0` to make this clear.

I copied this from code written by @kantai  so maybe he wants to weigh in?

src/cost_estimates/fee_medians.rs
gregorycoppola: I don't think fee rate can be infinite, because demoninator must be positive. maybe we could panic here instead if it's infinite? @kantai  , thoughts?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -668,6 +711,25 @@ fn get_pox_info(http_origin: &str) -> RPCPoxInfoData {
         .unwrap()
 }
 
+fn get_fee_rate_info(
```
pavitthrap, line 714: function name isn't so indicative function - maybe rename to `get_pox_info`?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -5880,6 +5942,192 @@ fn atlas_stress_integration_test() {
...
+      (unwrap! (increment) (err u1))
+      (ok (var-get counter))))
+    "#
+    .to_string();
+
+    let spender_sk = StacksPrivateKey::new();
+    let spender_addr = to_addr(&spender_sk);
+
+    // 1) a basic test `Config`
+    // 2) the miner's
```
pavitthrap, line 5978: unfinished comment here?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -442,6 +480,11 @@ fn find_microblock_privkey(
     return None;
 }
 
+// Returns true if relative difference between `a` and `b` is less than 0.01.
+fn is_near(a: f64, b: f64) -> bool {
+    (a - b).abs() < 0.01
```
pavitthrap, line 485: Do you want to use `error_epsilon` constant from fee_medians.rs in here?

pavitthrap (commented):
just a few minor comments left!

src/cost_estimates/fee_rate_fuzzer.rs
```diff
@@ -0,0 +1,92 @@
...
+        Self {
+            underlying,
+            rng_creator,
+            uniform_fuzz_fraction,
+        }
+    }
+
+    /// Add a uniform fuzz to input. Each element is multiplied by the same random factor.
+    fn fuzz_estimate(&self, input: FeeRateEstimate) -> FeeRateEstimate {
+        if self.uniform_fuzz_fraction > 0f64 {
```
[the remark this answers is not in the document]
gregorycoppola: 1) We only do this check because `Uniform::new` will fail if you pass in 0.

2) Returning `input` would not be correct if the randomness was 1.0. The correct answer would be to add random fuzz, up to 100% of the original value.

gregorycoppola (commented):
replying to @pavitthrap

src/cost_estimates/fee_rate_fuzzer.rs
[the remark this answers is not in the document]
pavitthrap: I suggested doing this check because otherwise, if `self.uniform_fuzz_fraction > 1.0`, the line `let fuzz_scale = 1f64 + uniform.sample(&mut rng);` could be negative

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,359 @@
...
+        let results = stmt
+            .query_and_then::<_, SqliteError, _, _>(&[window_size], |row| {
+                let high: f64 = row.get(0)?;
+                let middle: f64 = row.get(1)?;
+                let low: f64 = row.get(2)?;
+                Ok((low, middle, high))
+            })
+            .expect("SQLite failure");
+
+        for result in results {
```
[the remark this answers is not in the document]
pavitthrap: seems drastic to panic? maybe only iterate window_size times and print a warn if results is longer than expected. thoughts @kantai ?

testnet/stacks-node/src/tests/neon_integrations.rs
pavitthrap: I'm not sure where else it would go. I think it's alright to leave it in this file.

pavitthrap (commented):
responded to comments

testnet/stacks-node/src/tests/neon_integrations.rs
gregorycoppola: just copy-and-pasted for now

src/cost_estimates/fee_medians.rs
```diff
@@ -0,0 +1,359 @@
...
+/// Computes a `FeeRateEstimate` based on `sorted_fee_rates` using a "weighted percentile" method
+/// described in https://en.wikipedia.org/wiki/Percentile#Weighted_percentile
+///
+/// The percentiles computed are [0.05, 0.5, 0.95].
+///
+/// `sorted_fee_rates` must be non-empty.
+pub fn fee_rate_estimate_from_sorted_weighted_fees(
+    sorted_fee_rates: &Vec<FeeRateAndWeight>,
+) -> FeeRateEstimate {
+    if sorted_fee_rates.is_empty() {
```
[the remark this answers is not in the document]
gregorycoppola: yes

testnet/stacks-node/src/tests/neon_integrations.rs
gregorycoppola: fixed

testnet/stacks-node/src/config.rs
gregorycoppola: fixed

src/cost_estimates/fee_medians.rs
gregorycoppola: I chose to panic if the fee_rate is infite, because I think it should never be, unless the user had paid an infinte amount. does that sound ok @kantai

src/cost_estimates/fee_medians.rs
gregorycoppola: I rewrote using cmp::min, maybe it's clearer now.

gregorycoppola (commented):
Two unresolved points.

src/cost_estimates/fee_medians.rs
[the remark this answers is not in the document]
kantai: I think this would be an appropriate panic/assert, `assert!(results.len() <= window_size)`, because the SQL statement has a `LIMIT`, there's nothing short of a bug in sqlite or a misunderstanding in our code that should cause that assertion to fail.

src/cost_estimates/fee_medians.rs
kantai: No, this shouldn't panic -- while I'm pretty sure that we couldn't end up with an infinite fee rate here, I'm not convinced that's actually impossible (and it definitely isn't immediately obvious from just the surrounding code), so it's safer to return `None` in the event that fee_rate isn't finite (returning `MINIMUM_TX_FEE_RATE` wouldn't be accurate either). This should also emit a `warn!` log.

src/cost_estimates/fee_medians.rs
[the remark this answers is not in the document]
gregorycoppola: Let's just forget about this assert as:

1) we have a limit anyway, so we probably don't need to worry

2) this structure doesn't have a `len`:
```
error[E0599]: no method named `len` found for struct `AndThenRows` in the current scope
   --> src/cost_estimates/fee_medians.rs:130:25
    |
130 |         assert!(results.len() <= window_size);
    |                         ^^^ method not found in `AndThenRows<'_, [closure@src/cost_estimates/fee_medians.rs:122:69: 127:14]>`
```

src/cost_estimates/fee_medians.rs
gregorycoppola: ok, return none and log

src/cost_estimates/fee_medians.rs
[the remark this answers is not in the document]
kantai: That's fine with me.

src/cost_estimates/fee_rate_fuzzer.rs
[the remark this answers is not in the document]
pavitthrap: @gregorycoppola

pavitthrap (approved):
lgtm - thanks for addressing my feedback!

gregorycoppola:
Thanks for many rounds of reviews, everyone.

There was a broken that seemed flaky so re-running.

gregorycoppola:
@kantai sorry about the no-op commits. Next time I'll do it the right way but was just tired of this PR and took a short cut.

PR #2975, opened 2021-12-16
Feat/miner structured logging

pavitthrap:
## Description
Re-implements #2821: This PR adds the type `TransactionResult`, which is used to track outcomes of transaction inclusion by miners in both blocks and microblocks. 
There are three result types for a transactions: 
- Success - the tx is included in the (micro)block being mined
- Skipped - the tx is skipped over (this could be due to space constraints, or the tx might be invalid) 
- ProcessingError - the miner attempt to process the tx for inclusion in the block, but there was an error

Moreover, this PR updates the `mined_blocks` event and adds the `mined_microblocks` event. Each of these events has a field `tx_events`, which is a list of `TransactionEvents` (which are constructed from `TransactionResults`; they contain a subset of that data)

## Areas for Feedback
- The event dispatcher uses the set `miner_observers_lookup` to determine which observers to send `mined_blocks` events too. Instead of using this set, I created another set, `mined_microblocks_observers_lookup` for the new `mined_microblocks` event. I could:
     -  remove `mined_microblocks_observers_lookup` and just use `miner_observers_lookup` for both `mined_blocks` and `mined_microblocks` events
     - update the name of `miner_observers_lookup` to be `mined_blocks_observers_lookup`
     -  have three categories, `miner_observers_lookup`, `mined_blocks_observers_lookup` and `mined_microblocks_observers_lookup`. The first group `miner_observers_lookup` would subscribe to both the `mined_blocks` and `mined_microblocks` events
- Can use feedback on the payload for the following types (do we want any more data transmitted for the tx event?): 
     - `TransactionSuccessEvent`
     - `MinedMicroblockEvent`

## Testing 
Added `mining_events_integration_test` to `neon_integrations`

### DIFF
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -89,10 +89,17 @@ struct MicroblockMinerRuntime {
     disable_cost_check: bool,
 }
 
+/// The value of `BlockLimitFunction` holds the state of the size of the block being built.
+/// As the value increases, the less we can add to blocks.
 #[derive(PartialEq)]
 enum BlockLimitFunction {
+    /// The block size limit has not been hit, and there are no restrictions on what can be added to
+    /// a block.
     NO_LIMIT_HIT,
+    /// We have got a pretty full block, and so will not allow any more contract call or
+    /// contract publish transactions to be added to this block.
     CONTRACT_LIMIT_HIT,
+    /// We have a completely full block. No new transactions can be added to the block.
     LIMIT_REACHED,
 }
 
@@ -121,6 +128,163 @@ impl From<&UnconfirmedState> for MicroblockMinerRuntime {
 
             disable_bytes_check: unconfirmed.disable_bytes_check,
             disable_cost_check: unconfirmed.disable_cost_check,
+        }
+    }
+}
+
+/// Represents a successful transaction. This transaction should be added to the block.
+#[derive(Debug, Clone, PartialEq)]
+pub struct TransactionSuccess {
+    pub tx: StacksTransaction,
+    // The fee that was charged to the user for doing this transaction.
+    pub fee: u64,
+    pub receipt: StacksTransactionReceipt,
+}
+
+/// Represents a failed transaction. Something concreteley went wrong.
+#[derive(Debug)]
+pub struct TransactionError {
+    pub tx: StacksTransaction,
+    pub error: Error,
+}
+
+/// Represents a transaction that was skipped, but might succeed later.
+#[derive(Debug, Clone, PartialEq)]
+pub struct TransactionSkipped {
+    pub tx: StacksTransaction,
+    pub reason: String,
+}
+
+/// `TransactionResult` represents the outcome of transaction processing.
+/// We use this enum to involve the compiler in forcing us to always clearly
+/// indicate the outcome of a transaction.
+///
+/// There are currently three outcomes for a transaction:
+/// 1) succeed
+/// 2) fail, may be tried again later
+/// 3) be skipped for now, to be tried again later
+#[derive(Debug)]
+pub enum TransactionResult {
+    /// Transaction has already succeeded.
+    Success(TransactionSuccess),
+    /// Transaction failed. It may succeed later depending on the error.
+    ProcessingError(TransactionError),
+    /// Transaction wasn't ready to be be processed, but might succeed later.
+    /// The bool represents whether mempool propagation should halt or continue
+    Skipped(bool, TransactionSkipped),
+}
+
+impl TransactionResult {
+    /// Logs a queryable message for the case where `txid` has succeeded.
+    pub fn log_transaction_success(tx: &StacksTransaction) {
+        let tx_id_str = format!("{}", &tx.txid());
+        info!("Tx successfully processed.";
+            "event_name" => %"transaction_result",
+            "tx_id" => %tx_id_str,
+            "event_type" => %"success",
+        );
+    }
+
+    /// Logs a queryable message for the case where `txid` has failed
+    /// with error `err`.
+    pub fn log_transaction_error(tx: &StacksTransaction, err: &Error) {
+        let tx_id_str = format!("{}", &tx.txid());
+        warn!("Tx processing failed with error: {}", err;
+            "event_name" => %"transaction_result",
+            "tx_id" => %tx_id_str,
+            "event_type" => %"error",
+        );
+    }
+
+    /// Logs a queryable message for the case where `tx` has been skipped
+    /// for reason `reason`.
+    pub fn log_transaction_skipped(tx: &StacksTransaction, reason: String) {
+        let tx_id_str = format!("{}", &tx.txid());
+        info!(
+            "Tx processing skipped for reason: {}", reason;
+            "event_name" => %"transaction_result",
+            "tx_id" => %tx_id_str,
+            "event_type" => %"skip"
+        );
+    }
+
+    /// Creates a `TransactionResult` backed by `TransactionSuccess`.
+    /// This method logs "transaction success" as a side effect.
+    pub fn success(
+        transaction: &StacksTransaction,
+        fee: u64,
+        receipt: StacksTransactionReceipt,
+    ) -> TransactionResult {
+        Self::log_transaction_success(transaction);
+        Self::Success(TransactionSuccess {
+            tx: transaction.clone(),
+            fee: fee,
+            receipt: receipt,
+        })
+    }
+
+    /// Creates a `TransactionResult` backed by `TransactionError`.
+    /// This method logs "transaction error" as a side effect.
+    pub fn error(transaction: &StacksTransaction, error: Error) -> TransactionResult {
+        Self::log_transaction_error(transaction, &error);
+        TransactionResult::ProcessingError(TransactionError {
+            tx: transaction.clone(),
+            error: error,
+        })
+    }
+
+    /// Creates a `TransactionResult` backed by `TransactionSkipped`.
+    /// This method logs "transaction skipped" as a side effect.
+    pub fn skipped(
+        transaction: &StacksTransaction,
+        reason: String,
+        continue_processing: bool,
+    ) -> TransactionResult {
+        Self::log_transaction_skipped(transaction, reason.clone());
+        TransactionResult::Skipped(
+            continue_processing,
+            TransactionSkipped {
+                tx: transaction.clone(),
+                reason: reason,
+            },
+        )
+    }
+
+    /// Returns true iff this enum is backed by `TransactionSuccess`.
+    pub fn is_ok(&self) -> bool {
+        match &self {
+            TransactionResult::Success(_) => true,
+            _ => false,
+        }
+    }
+
+    /// Returns a TransactionSuccess result as a pair of 1) fee and 2) receipt.
+    /// Otherwise crashes.
+    pub fn unwrap(self) -> (u64, StacksTransactionReceipt) {
+        match self {
+            TransactionResult::Success(TransactionSuccess {
+                tx: _,
+                fee,
+                receipt,
+            }) => (fee, receipt),
+            _ => panic!("Tried to `unwrap` a non-success result."),
+        }
+    }
+
+    /// Returns true iff this enum is backed by `Error`.
+    pub fn is_err(&self) -> bool {
+        match &self {
+            TransactionResult::ProcessingError(_) => true,
+            _ => false,
+        }
+    }
+
+    /// Returns an Error result as an Error.
+    /// Otherwise crashes.
+    pub fn unwrap_err(self) -> Error {
+        match self {
+            TransactionResult::ProcessingError(TransactionError { tx: _, error }) => error,
+            _ => panic!("Tried to `unwrap_error` a non-error result."),
         }
     }
 }
@@ -332,22 +496,46 @@ impl<'a> StacksMicroblockBuilder<'a> {
     }
 
     /// Mine the next transaction into a microblock.
-    /// Returns Some(StacksTransactionReceipt) or None if the transaction was
-    /// or was not mined into this microblock.
+    /// Returns Ok(TransactionResult::Success) if the transaction was mined into this microblock.
+    /// Returns Ok(TransactionResult::Skipped) if the transaction was not mined, but can be mined later.
+    /// Returns Ok(TransactionResult::Error) if the transaction was not mined due to an error.
+    /// Returns Err(e) if an error occurs during the function.
+    ///
+    /// This calls `StacksChainState::process_transaction` and also checks certain pre-conditions
+    /// and handles errors.
+    ///
+    /// # Pre-Checks
+    /// - skip if the `anchor_mode` rules out micro-blocks
+    /// - skip if 'tx.txid()` is already in `considered`
+    /// - skip if adding the block would result in a block size bigger than `MAX_EPOCH_SIZE`
+    ///
+    /// # Error Handling
+    /// - If the error when processing a tx is `CostOverflowError`, reset the cost of the block.
     fn mine_next_transaction(
         clarity_tx: &mut ClarityTx<'a>,
         tx: StacksTransaction,
         tx_len: u64,
         considered: &mut HashSet<Txid>,
         bytes_so_far: u64,
-    ) -> Result<Option<StacksTransactionReceipt>, Error> {
+    ) -> Result<TransactionResult, Error> {
         if tx.anchor_mode != TransactionAnchorMode::OffChainOnly
             && tx.anchor_mode != TransactionAnchorMode::Any
         {
-            return Ok(None);
+            return Ok(TransactionResult::skipped(
+                &tx,
+                format!(
+                    "tx.anchor_mode does not support microblocks, anchor_mode={:?}.",
+                    tx.anchor_mode
+                ),
+                true,
+            ));
         }
         if considered.contains(&tx.txid()) {
-            return Ok(None);
+            return Ok(TransactionResult::skipped(
+                &tx,
+                "Already considered.".to_string(),
+                true,
+            ));
         } else {
             considered.insert(tx.txid());
         }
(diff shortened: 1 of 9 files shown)


src/chainstate/stacks/miner.rs
```diff
@@ -356,29 +616,37 @@ impl<'a> StacksMicroblockBuilder<'a> {
...
+                            tx.txid(),
+                            &cost_after,
+                            &total_budget
+                        );
+                        clarity_tx.reset_cost(cost_before.clone());
+                        Ok(TransactionResult::error(&tx, e))
+                    }
+                    _ => {
+                        warn!("Error processing TX {}: {}", tx.txid(), e);
+                        // TODO: might want to change this to error.
```
gregorycoppola, line 644: This TODO is still here.

src/chainstate/stacks/miner.rs
```diff
@@ -332,22 +568,46 @@ impl<'a> StacksMicroblockBuilder<'a> {
     }
 
     /// Mine the next transaction into a microblock.
-    /// Returns Some(StacksTransactionReceipt) or None if the transaction was
-    /// or was not mined into this microblock.
+    /// Returns Ok(TransactionResult::Success) if the transaction was mined into this microblock.
+    /// Returns Ok(TransactionResult::Skipped) if the transaction was not mined, but can be mined later.
+    /// Returns Ok(TransactionResult::Error) if the transaction was not mined due to an error.
```
gregorycoppola, line 573: what is the difference between `Ok(TR:Error)` and `Err(e)`?

I.e., it's `transaction was not mined due to an error` vs. `if an error occurs during the function`.

I'm not understanding the dividing line from the comment text.

src/chainstate/stacks/miner.rs
```diff
@@ -816,56 +1117,75 @@ impl StacksBlockBuilder {
...
 
         let quiet = !cfg!(test);
-        let receipt = if !self.anchored_done {
+        let result = if !self.anchored_done {
             // building up the anchored blocks
             if tx.anchor_mode != TransactionAnchorMode::OnChainOnly
                 && tx.anchor_mode != TransactionAnchorMode::Any
             {
-                return Err(Error::InvalidStacksTransaction(
+                return TransactionResult::skipped(
```
gregorycoppola, line 1153: is this a change in logic from error to skipped?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -3717,6 +3739,222 @@ fn cost_voting_integration() {
...
+    assert_eq!(account.balance, 10000000);
+
+    let account = get_account(&http_origin, &addr_2);
+    assert_eq!(account.nonce, 0);
+    assert_eq!(account.balance, 10000000);
+
+    submit_tx(&http_origin, &tx); // should succeed
+    submit_tx(&http_origin, &tx_2); // should fail since it tries to publish contract with same name
+    submit_tx(&http_origin, &mb_tx); // should be in microblock bc it is microblock only
+    sleep_ms(20_000);
```
gregorycoppola, line 3839: is the sleep needed? is `next_block_and_wait` enough?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -3717,6 +3739,222 @@ fn cost_voting_integration() {
...
+    // first block wakes up the run loop
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    // first block will hold our VRF registration
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    // second block will be the first mined Stacks block
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    let account = get_account(&http_origin, &miner_account);
```
gregorycoppola, line 3824: we copy these "get_account" checks around but i don't think it's needed for the test

gregorycoppola (commented):
Thanks for taking this up and careful thought about it.

I just had a few questions about the control flow and comments.

gregorycoppola:
@kantai @jcnelson is there an "ontology" of different error subtypes in the transaction processor?

How do we know which errors would cause a block to stop processing (ie `break`) vs keep going with the block ( ie `continue`?

Do we basically always continue if the block isn't full?

Are we panicking on all unrecoverable errors so that any Err error is recoverable?

src/chainstate/stacks/miner.rs
```diff
@@ -125,6 +132,224 @@ impl From<&UnconfirmedState> for MicroblockMinerRuntime {
...
+            "tx_id" => %tx_id_str,
+            "event_type" => %"success",
+        );
+    }
+
+    /// Logs a queryable message for the case where `txid` has failed
+    /// with error `err`.
+    pub fn log_transaction_error(tx: &StacksTransaction, err: &Error) {
+        let tx_id_str = format!("{}", &tx.txid());
+        warn!("Tx processing failed with error: {}", err;
```
kantai, line 228: I don't think this should be `warn` level. A transaction having an error is a "normal" event, and doesn't indicate that there's either something wrong with the network or the node (which is what `warn` would usually indicate).

src/chainstate/stacks/miner.rs
```diff
@@ -356,29 +616,37 @@ impl<'a> StacksMicroblockBuilder<'a> {
...
+                        // note: this path _does_ not perform the tx block budget % heuristic,
+                        //  because this code path is not directly called with a mempool handle.
+                        warn!(
+                            "Transaction {} reached block cost {}; budget was {}",
+                            tx.txid(),
+                            &cost_after,
+                            &total_budget
+                        );
+                        clarity_tx.reset_cost(cost_before.clone());
+                        Ok(TransactionResult::error(&tx, e))
```
kantai, line 640: This shouldn't be a `TransactionResult::error`, but instead a `TransactionResult::skipped` with the budget message. The transaction itself didn't necessarily do anything to create this error, rather the block is just full.

src/chainstate/stacks/miner.rs
```diff
@@ -125,6 +132,224 @@ impl From<&UnconfirmedState> for MicroblockMinerRuntime {
...
+    /// Logs a queryable message for the case where `tx` has been skipped
+    /// for reason `reason`.
+    pub fn log_transaction_skipped(tx: &StacksTransaction, reason: String) {
+        let tx_id_str = format!("{}", &tx.txid());
+        info!(
+            "Tx processing skipped for reason: {}", reason;
+            "event_name" => %"transaction_result",
+            "tx_id" => %tx_id_str,
+            "event_type" => %"skip"
+        );
```
kantai, line 244: Here's a suggested change for these lines:

```suggestion
        info!(
            "Tx processing skipped";
            "event_name" => "transaction_result",
            "tx_id" => %tx.txid(),
            "event_type" => "skip",
            "reason" => reason,
        );
```

Basically, make "reason" a field in the structured log. Also, the `%` prefix just tells the macro to invoke the `Display` format implementation to write the passed object as a str, which the str arguments don't need.

src/chainstate/stacks/miner.rs
```diff
@@ -125,6 +132,224 @@ impl From<&UnconfirmedState> for MicroblockMinerRuntime {
...
+
+    /// Logs a queryable message for the case where `txid` has failed
+    /// with error `err`.
+    pub fn log_transaction_error(tx: &StacksTransaction, err: &Error) {
+        let tx_id_str = format!("{}", &tx.txid());
+        warn!("Tx processing failed with error: {}", err;
+            "event_name" => %"transaction_result",
+            "tx_id" => %tx_id_str,
+            "event_type" => %"error",
+        );
```
kantai, line 232:
```suggestion
        warn!("Tx processing failed with error";
            "event_name" => "transaction_result",
            "reason" => %err,
            "tx_id" => %tx.txid(),
            "event_type" => "error",
        );
```

gregorycoppola:
If we're able to change in the production code "whether or not an error-like event aborts processing for the batch", and not break any tests, then it would probably add a lot of value to add some integration tests that would capture that. 

I.e., the test would break if we changed the error class of each error event.

Probably that's a big thing to ask but eventually would be very useful. This is really a key component of the blockchain flow.

src/chainstate/stacks/miner.rs
```diff
@@ -804,9 +1101,13 @@ impl StacksBlockBuilder {
...
         limit_behavior: &BlockLimitFunction,
-    ) -> Result<StacksTransactionReceipt, Error> {
+    ) -> TransactionResult {
         if self.bytes_so_far + tx_len >= MAX_EPOCH_SIZE.into() {
-            return Err(Error::BlockTooBigError);
+            return TransactionResult::skipped(
+                &tx,
+                "Adding this tx would exceed epoch data size".to_string(),
+                true,
+            );
```
kantai, line 1110: I don't know if this will actually impact the miner's control-flow, but these changes do alter the error returned by `try_mine_tx` in this case from `Error::BlockTooBigError` to `Error::StacksTransactionSkipped`. I think we may need to add an `Error` field to the `StacksTransactionSkipped` struct, which is just `Error::StacksTransactionSkipped` when constructed using `TransactionResult::skipped`, and then there's another variant like `TransactionResult::skipped_on_block_done()` that accepts an error like `Error::BlockTooBigError` as an argument.

src/chainstate/stacks/miner.rs
```diff
@@ -816,56 +1117,75 @@ impl StacksBlockBuilder {
...
+                            return TransactionResult::error(&tx, Error::TransactionTooBigError);
                         } else {
                             warn!(
                                 "Transaction {} reached block cost {}; budget was {}",
                                 tx.txid(),
                                 &cost_after,
                                 &total_budget
                             );
-                            Error::BlockTooBigError
+                            return TransactionResult::error(&tx, Error::BlockTooBigError);
```
kantai, line 1183: Kind of like in my comment on the microblock transaction mining, the `BlockTooBigError` isn't really the transaction's error, so this should be something like `TransactionResult::skipped_on_block_done()` with `Error::BlockTooBigError` supplied.

src/chainstate/stacks/miner.rs
kantai: This isn't exactly a "transaction error", but I agree that the error passed up from this method should be the same. I think this another case where the `::skipped` variant should be constructed with an Error type.

src/chainstate/stacks/miner.rs
```diff
@@ -875,43 +1195,48 @@ impl StacksBlockBuilder {
...
```
kantai, line 1234: Same comment here as above.

src/chainstate/stacks/miner.rs
```diff
@@ -875,43 +1195,48 @@ impl StacksBlockBuilder {
...
                 && tx.anchor_mode != TransactionAnchorMode::Any
             {
-                return Err(Error::InvalidStacksTransaction(
+                return TransactionResult::skipped(
+                    tx,
                     "Invalid transaction anchor mode for streamed data".to_string(),
-                    false,
-                ));
+                    true,
+                );
```
kantai, line 1208: Same comment here as above.

src/chainstate/stacks/miner.rs
```diff
@@ -1575,40 +1908,54 @@ impl StacksBlockBuilder {
...
-                                    debug!("Switch to mining stx-transfers only");
-                                } else if block_limit_hit == BlockLimitFunction::CONTRACT_LIMIT_HIT
-                                {
-                                    debug!("Stop mining anchored block due to limit exceeded");
-                                    block_limit_hit = BlockLimitFunction::LIMIT_REACHED;
-                                    return Ok(false);
+                            TransactionResult::Skipped(..) => {}
+                            TransactionResult::ProcessingError(TransactionError {
+                                error, ..
+                            }) => {
```
kantai, line 1914: I'm pretty sure that if you change the `Skipped` variant to also have an `error` field, this could be matched as:
```
TransactionResult::Skipped(TransactionSkipped { error, .. }) | TransactionResult::ProcessingError(TransactionError { error, .. }) =>
```

And then you'd just need to add a `Error::StacksTransactionSkipped => {}` branch to the `match &error` block to silence the warnings.

src/core/mempool.rs
```diff
@@ -714,7 +723,8 @@ impl MemPoolDB {
     ///  highest-fee-first order.  This method is interruptable -- in the `settings` struct, the
     ///  caller may choose how long to spend iterating before this method stops.
     ///
-    ///  Returns the number of transactions considered on success.
+    ///  `todo` must return a result of type StacksTransactionResult, indicating the outcome of the
+    ///  transaction.
```
kantai, line 727: These comment changes don't seem accurate, right? `todo` still returns a `bool` indicating whether or not to continue iterating or exit early.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -3717,6 +3739,222 @@ fn cost_voting_integration() {
     channel.stop_chains_coordinator();
 }
 
+#[test]
+#[ignore]
+fn mining_events_integration_test() {
```
kantai, line 3744: Can you add this test to the `sampled-genesis` test matrix in `.github/workflows/bitcoin-tests.yml`?

gregorycoppola:
Thanks again for doing a fast job on this, @pavitthrap .

Part of the goal of making this PR was to discover an error ontology by trying this change and having it reviewed. So, hopefully doesn't seem like a drag to be iterating on the solution.

We're clarifying the API a lot and also got the events out of it.

kantai (commented):
These changes look good to me, thanks @pavitthrap! Just had a few comments about the TransactionResult handling in the miner.

gregorycoppola:
Thanks for the help finishing this, @kantai .

src/chainstate/stacks/miner.rs
```diff
@@ -412,15 +681,27 @@ impl<'a> StacksMicroblockBuilder<'a> {
...
+                            test_debug!("Include tx {} in microblock", tx.txid());
+                            bytes_so_far += tx_len;
+                            num_txs += 1;
+                            txs_included.push(tx);
+                        }
+                        TransactionResult::Skipped(..) => {
+                            test_debug!("Exclude tx {} from microblock", tx.txid());
+                            continue;
+                        }
+                        TransactionResult::ProcessingError(TransactionError { error, .. }) => {
```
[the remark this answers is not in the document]
pavitthrap: It's not possible - here, we have a match for the result of `mine_next_transaction`, which did not return an Err even if the transaction was skipped or failed.

src/chainstate/stacks/miner.rs
```diff
@@ -445,23 +726,21 @@ impl<'a> StacksMicroblockBuilder<'a> {
         self.runtime.num_mined = num_txs;
 
         match result {
-            Err(Error::BlockTooBigError) => {
```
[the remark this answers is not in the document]
pavitthrap: I moved this code block to the `ProcessingError` branch of the match just above. 
This piece of code was actually never executed before, because `mine_next_transaction` never returned an Err.

src/chainstate/stacks/miner.rs
pavitthrap: The latter refers to an unexpected "operating" error in the function. An example of an operating error would be if a database query failed.

testnet/stacks-node/src/tests/neon_integrations.rs
pavitthrap: True... but I guess it doesn't hurt to ensure that the starting state of all the relevant accounts are what you expect them to be? Let me know what you think.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -158,6 +162,14 @@ pub mod test_observer {
         Ok(warp::http::StatusCode::OK)
     }
 
+    async fn handle_mined_microblock(
```
gregorycoppola, line 165: what does it mean to "handle"? can we have a  brief comment here?

src/core/mempool.rs
```diff
@@ -714,7 +723,7 @@ impl MemPoolDB {
     ///  highest-fee-first order.  This method is interruptable -- in the `settings` struct, the
     ///  caller may choose how long to spend iterating before this method stops.
     ///
-    ///  Returns the number of transactions considered on success.
+    ///  `todo` returns the number of transactions considered on success.
```
gregorycoppola, line 726: I think `todo` returns a bool?

gregorycoppola (commented):
Eagerly awaiting the landing of this PR.

I just had a few very small points and I am LGTM.

However, with so little test coverage in the area, I leave it up Aaron and Pavitthra's knowledge of the code paths and ability to read code carefully in order to ensure we don't break existing logic/consensus.

src/core/mempool.rs
pavitthrap: good catch!

gregorycoppola (approved):

src/chainstate/stacks/miner.rs
```diff
@@ -356,29 +622,50 @@ impl<'a> StacksMicroblockBuilder<'a> {
...
+                        if total_budget.proportion_largest_dimension(&cost_before)
+                            < TX_BLOCK_LIMIT_PROPORTION_HEURISTIC
+                        {
+                            warn!(
+                                "Transaction {} consumed over {}% of block budget, marking as invalid; budget was {}",
+                                tx.txid(),
+                                100 - TX_BLOCK_LIMIT_PROPORTION_HEURISTIC,
+                                &total_budget
+                            );
+                            return Ok(TransactionResult::error(
```
jcnelson, line 648: Should this be `::skipped_due_to_error()`?  It's not the transaction's fault that the block was nearly full when the miner attempted to apply it.

jcnelson: Ah nm -- this is fine.  I had thought `::skipped_due_to_error()` was for skipping transactions that could be mined, but were not for some other reason.

jcnelson (approved):
LGTM!  Thanks @pavitthrap!

PR #2981, opened 2021-12-22
Fix: update microblock mining behavior 

pavitthrap:
While working on #2975, I noticed some unexpected logic in the miner. 

Here is a summary of changes in this PR: 
- Updated `mine_next_microblock` to keep track of `block_limit_hit`. This variable toggles between `NO_LIMIT_HIT`, `CONTRACT_LIMIT_HIT`, and `LIMIT_REACHED`. Keeping track of this signals the microblock builder to stop processing transactions once the block limit is hit. 
- Updated `mine_next_transaction` (which is used by `mine_next_microblock`) to take `block_limit_hit` as an input. This allows the function to skip over appropriate transactions (for example when `block_limit_hit=CONTRACT_LIMIT_HIT`, contract call and smart contract payloads will be automatically skipped). 
- Moved logic that updates mined_origin_nonces in `build_anchored_block` so that it only gets executed if the transaction being considered was successfully mined. As it was before, these values were sometimes being updated even when the transaction processing failed. 
- Changed location of where transactions were being added to the `considered` set for the microblock mining flow. It was originally in `mine_next_transaction`, but I moved the logic out to the calling function because I thought it was (a) more readable, and (b) it now mirrors the logic for the anchored block mining flow. 

I need feedback on the following change: 
- Currently, the `block_limit_hit` variable moves onto its next state if a transaction errors with either `BlockTooBigError` or `TransactionTooBigError`. I removed this transition for the latter case, since I felt that it didn't make sense to change the block's limit behavior for an invalid tx that consumes >80% of the block budget that we are anyway dropping from the mempool. 

Tests: 
Ran existing tests. 
Added 4 tests: 
- block limit behavior 
- microblock limit behavior 
- transaction too large in anchored block mining 
- transaction too large in microblock mining


src/chainstate/stacks/miner.rs
```diff
@@ -709,8 +740,29 @@ impl<'a> StacksMicroblockBuilder<'a> {
...
                         }
-                        TransactionResult::Skipped(..) | TransactionResult::ProcessingError(..) => {
+                        TransactionResult::Skipped(TransactionSkipped { error, .. })
+                        | TransactionResult::ProcessingError(TransactionError { error, .. }) => {
                             test_debug!("Exclude tx {} from microblock", tx.txid());
+                            match &error {
+                                Error::BlockTooBigError => {
+                                    // done mining -- our execution budget is exceeded.
+                                    // Make the block from the transactions we did manage to get
+                                    test_debug!("Block budget exceeded on tx {}", &tx.txid());
```
jcnelson, line 750: Can this `test_debug!` statement be merged with the ones in either `if` branch, so that there's just one debug line emitted here?  Makes it easier to `grep` for this event in test output.

jcnelson (approved):
LGTM!  Let's just make sure all the tests pass first.

src/chainstate/stacks/miner.rs
```diff
@@ -861,8 +922,40 @@ impl<'a> StacksMicroblockBuilder<'a> {
...
+                                                        BlockLimitFunction::CONTRACT_LIMIT_HIT;
+                                                } else if block_limit_hit
+                                                    == BlockLimitFunction::CONTRACT_LIMIT_HIT
+                                                {
+                                                    debug!(
+                                                        "Stop mining anchored block due to limit exceeded"
+                                                    );
+                                                    block_limit_hit = BlockLimitFunction::LIMIT_REACHED;
+                                                    return Ok(false);
+                                                }
```
kantai, line 950: In addition to combining the logs, I recommend moving some of the data to a key-value pair:

```suggestion
                                                if block_limit_hit == BlockLimitFunction::NO_LIMIT_HIT {
                                                    debug!("Block budget exceeded while mining microblock"; "tx" => %tx.txid(), "next_behavior" => "Switch to mining stx-transfers only");
                                                    block_limit_hit =
                                                        BlockLimitFunction::CONTRACT_LIMIT_HIT;
                                                } else if block_limit_hit
                                                    == BlockLimitFunction::CONTRACT_LIMIT_HIT
                                                {
                                                    debug!("Block budget exceeded while mining microblock"; "tx" => %tx.txid(), "next_behavior" => "Stop mining microblock");
                                                    block_limit_hit = BlockLimitFunction::LIMIT_REACHED;
                                                    return Ok(false);
                                                }
```

kantai (approved):
LGTM -- just had some comments on log entries.

PR #2984, opened 2022-01-06
Feat: update pessimistic cost estimator to differentiate by contract sender

kantai:
This change adds a field to the pessimistic cost estimator's lookup key. It doesn't change the table schema, so it's compatible with existing cost estimator tables, though it will end up resetting cost estimates. This change allows the cost estimator to differentiate between contracts with different contract senders. In most respects, this PR is a fix for incorrect behavior on the part of the pessimistic estimator, but I'm not sure this rises to the level of necessitating a hotfix, as the cost estimator's current behavior still allows miners to progress the network, albeit slightly less efficiently. If necessary, though, this could be rebased as a hotfix very easily (the patch is essentially a one line change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
+## [Unreleased]
+
+### Changed
+
+- The pessimistic execution cost estimator differentiates between
+  contracts with different origin addresses.
+
 ## [2.05.0.0.0]
 
 This software update is a consensus changing release and the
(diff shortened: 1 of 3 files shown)


gregorycoppola:
So you are adding the address of the user calling the contract?

What motivates this change? What problem is it solving and what data suggests it is working?

kantai:
> So you are adding the address of the user calling the contract?

No -- this is the contract publisher address.

> What motivates this change? What problem is it solving and what data suggests it is working?

Without differentiating by contract publisher, two contracts with the same name, but different publishers, would share an estimate, even though the contracts are actually different contracts. This isn't rooted in data, but it is definitely incorrect behavior (which is why this is isn't necessarily a hotfix), and the added unit test demonstrates the correct behavior (separation between different contracts).

gregorycoppola (approved):
Oh I see. yes this would seem to be a logical bug.

pavitthrap (approved):
lgtm!

jcnelson (approved):

PR #2985, opened 2022-01-07
libclarity separation

kantai:
This draft PR performs a strict separation between `libclarity` and the main Cargo outputs of this repository. It makes `libclarity` a workspace member, removing the source files from the main source directory. In doing this, this PR also separates out a `stacks-common` library for usage in both the main source tree and the `libclarity` source tree.

The separation boundary is mostly based on the logic of what should live in a general-purpose Clarity instance, versus what components are Stacks chain specific. Because of this, things like the genesis contracts _are not_ included in the Stacks chain, nor are the "special case handlers" for contract-calls (i.e., handlers for the PoX locks). Some separation boundaries are confusing, but are required by the handling of lifetime bounds in rust generics. Notably, the high-level `instance -> block connection -> transaction connection` interface in `src/clarity_vm/clarity.rs` cannot be made generic in terms of the backing database. I believe that doing so would require support for higher-ordered generics in rust. Instead, `libclarity` defines a trait for transaction connections, which supplies auto-implementations for common transaction types.

(Note: closed the prior PR, because it seems like, for whatever reason, that PR wasn't getting picked up by Github Actions, so none of the CI was running on it, except for the Circle unit tests).


clarity/src/vm/analysis/arithmetic_checker/tests.rs
```diff
@@ -44,10 +43,11 @@ fn check_good(contract: &str) {
     ArithmeticOnlyChecker::run(&analysis).expect("Should pass arithmetic checks");
 }
 
-#[test]
-fn test_boot_definitions() {
-    check_good(BOOT_CODE_COSTS);
-}
+// #[test]
```
jcnelson, line 46: :eyes:

jcnelson, line 46: :eyes:

clarity/src/vm/analysis/tests/mod.rs
```diff
@@ -19,7 +19,7 @@ use vm::analysis::{mem_type_check, AnalysisDatabase};
 use vm::analysis::{type_check, ContractAnalysis};
 use vm::ast::parse;
 
-mod costs;
+// mod costs;
```
jcnelson, line 22: :eyes:

clarity/src/vm/ast/mod.rs
```diff
@@ -70,68 +70,13 @@ pub fn build_ast<T: CostTracker>(
...
-    use clarity_vm::clarity::ClarityInstance;
-    use clarity_vm::database::marf::MarfedKV;
+mod test {
     use std::collections::HashMap;
-    use vm::costs::*;
-    use vm::database::*;
-    use vm::representations::depth_traverse;
-    use vm::tests::{TEST_BURN_STATE_DB, TEST_HEADER_DB};
-
-    fn dependency_edge_counting_runtime(iters: usize) -> u64 {
```
jcnelson, line 85: Where did this code get moved to? Or was it deleted?

clarity/src/vm/clarity.rs
```diff
@@ -0,0 +1,305 @@
...
+    ) -> Result<(Value, AssetMap, Vec<StacksTransactionEvent>), Error> {
+        self.with_abort_callback(
+            |vm_env| vm_env.stx_transfer(from, to, amount).map_err(Error::from),
+            |_, _| false,
+        )
+        .and_then(|(value, assets, events, _)| Ok((value, assets, events)))
+    }
+
+    /// Execute a contract call in the current block.
+    ///  If an error occurs while processing the transaction, it's modifications will be rolled back.
```
jcnelson, line 234: "its modifications"

clarity/src/vm/costs/mod.rs
```diff
@@ -46,6 +45,10 @@ type Result<T> = std::result::Result<T, CostErrors>;
 
 pub const CLARITY_MEMORY_LIMIT: u64 = 100 * 1000 * 1000;
 
+// TODO: factor out into a boot lib?
```
jcnelson, line 48: This is probably the right thing to do, if possible: create a "boot" crate that has a top-level trait that gets invoked at bootup time and then once per block thereafter, and will (in the former case) instantiate `.pox`, `.costs`, `.bns`, etc. and fill in the initial genesis balances as well as (in the latter case) check for the appropriate block heights at which to deploy new system boot code like `.costs-2` and (eventually) `.pox-2`.  Subnets and appchains can have their own boot implementations this way, and `libclarity` doesn't need to be peppered with bootup code.

clarity/src/vm/database/clarity_db.rs
```diff
@@ -17,11 +17,11 @@
 use std::collections::{HashMap, VecDeque};
 use std::convert::{TryFrom, TryInto};
 
-use core::{
-    StacksEpoch, StacksEpochId, BITCOIN_REGTEST_FIRST_BLOCK_HASH,
-    BITCOIN_REGTEST_FIRST_BLOCK_HEIGHT, BITCOIN_REGTEST_FIRST_BLOCK_TIMESTAMP,
-    FIRST_BURNCHAIN_CONSENSUS_HASH, FIRST_STACKS_BLOCK_HASH, POX_REWARD_CYCLE_LENGTH,
-};
+// use core::{
```
jcnelson, line 20: :eyes:

clarity/src/vm/database/clarity_db.rs
```diff
@@ -700,6 +687,10 @@ impl<'a> ClarityDatabase<'a> {
         Ok(())
     }
 
+    pub fn get_cc_special_cases_handler(&self) -> Option<SpecialCaseHandler> {
+        self.store.get_cc_special_cases_handler()
+    }
+
```
jcnelson, line 693: I'm confused here -- why is `insert_microblock_poison` preserved in the clarity DB, but the top-level method for processing it in an `OwnedEnvironment` is dropped?

clarity/src/vm/database/clarity_store.rs
```diff
@@ -228,3 +246,75 @@ impl ClarityBackingStore for NullBackingStore {
...
+
+    fn get_open_chain_tip_height(&mut self) -> u32 {
+        0
+    }
+
+    fn get_current_block_height(&mut self) -> u32 {
+        0
+    }
+
+    fn get_cc_special_cases_handler(&self) -> Option<SpecialCaseHandler> {
```
jcnelson, line 311: If this implementation is the one used by the clarity CLI, it will need to use the same `SpecialCaseHandler`s that the production implementation does.

clarity/src/vm/docs/contracts.rs
```diff
@@ -1,13 +1,13 @@
-use chainstate::stacks::boot::STACKS_BOOT_CODE_MAINNET;
+//use chainstate::stacks::boot::STACKS_BOOT_CODE_MAINNET;
```
jcnelson, line 1: :eyes:

clarity/src/vm/docs/contracts.rs
```diff
@@ -44,6 +44,8 @@ struct ContractSupportDocs {
     skip_func_display: HashSet<&'static str>,
 }
 
+const STACKS_BOOT_CODE_MAINNET: [(&'static str, &'static str); 0] = [];
```
jcnelson, line 47: Why not just drop this entirely?

clarity/src/vm/docs/mod.rs
```diff
@@ -1889,79 +1886,79 @@ mod test {
...
-                        assert_eq!(expected, result.unwrap().to_string());
-                    }
-                }
-                Ok(())
-            })
-            .unwrap();
-
-        store.rollback_block();
-    }
+    // fn docs_execute(marf: &mut MarfedKV, program: &str) {
```
jcnelson, line 1889: Is this going away?

clarity/src/vm/mod.rs
```diff
@@ -32,19 +32,28 @@ pub mod database;
...
 pub mod tests;
 
-use crate::clarity_vm::database::MemoryBackingStore;
-use crate::core::StacksEpochId;
+#[cfg(any(test, feature = "testing"))]
+pub mod test_util;
+
+pub mod clarity;
+
+// publish the non-generic StacksEpoch form for use throughout module
```
jcnelson, line 53: So, there are *two* `StacksEpoch`s now?

clarity/src/vm/tests/contracts.rs
```diff
@@ -167,255 +163,6 @@ fn test_get_block_info_eval() {
...
-        _ => false,
-    }
-}
-
-fn test_block_headers(n: u8) -> StacksBlockId {
-    StacksBlockId([n as u8; 32])
-}
-
-#[test]
-fn test_simple_token_system() {
```
jcnelson, line 189: This get moved somewhere, or is it gone now?

clarity/src/vm/tests/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+mod assets;
+mod contracts;
+mod datamaps;
+mod defines;
+mod events;
+mod sequences;
+mod simple_apply_eval;
+mod traits;
+
+// pub struct TestBlockConnection {
```
jcnelson, line 56: Not sure what to make of all this commented code

src/chainstate/burn/db/sortdb.rs
```diff
@@ -91,19 +93,19 @@ pub const REWARD_WINDOW_END: u64 = 144 * 90 + REWARD_WINDOW_START;
 
 pub type BlockHeaderCache = HashMap<ConsensusHash, (Option<BlockHeaderHash>, ConsensusHash)>;
 
-// for using BurnchainHeaderHash values as block hashes in a MARF
-impl From<BurnchainHeaderHash> for BlockHeaderHash {
-    fn from(bhh: BurnchainHeaderHash) -> BlockHeaderHash {
-        BlockHeaderHash(bhh.0)
-    }
-}
+// // for using BurnchainHeaderHash values as block hashes in a MARF
```
jcnelson, line 96: Is there a reason these aren't being used anymore?

src/chainstate/stacks/address.rs
```diff
@@ -34,156 +34,62 @@ use chainstate::stacks::{
...
-        // if segwit, then keys must all be compressed
-        match *hash_mode {
-            AddressHashMode::SerializeP2WPKH | AddressHashMode::SerializeP2WSH => {
-                for pubkey in pubkeys {
-                    if !pubkey.compressed() {
-                        return None;
-                    }
-                }
+    fn to_bitcoin_tx_out(&self, value: u64) -> TxOut {
+        let btc_version = to_b52_version_byte(self.version)
```
jcnelson, line 67: Do you mean `to_b58_version_byte`?

jcnelson: I think this method is inappropriately named, repo-wide.

clarity/src/vm/ast/mod.rs
jcnelson: Ah I see it now.

clarity/src/vm/tests/contracts.rs
jcnelson: Ah nm I see it now

src/net/mod.rs
```diff
@@ -724,10 +724,13 @@ pub struct PoxInvData {
     pub pox_bitvec: Vec<u8>, // a bit will be '1' if the node knows for sure the status of its reward cycle's anchor block; 0 if not.
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub struct BlocksDatum(pub ConsensusHash, pub StacksBlock);
```
jcnelson, line 728: Is there a reason for this particular refactoring?  If it's necessary, can you call it something more specific to indicate that the Stacks block is paired to a particular PoX history (which is the reason we include the consensus hash)?  Maybe `UniqueStacksBlock`?

src/util_lib/db.rs
```diff
@@ -224,13 +230,48 @@ impl FromColumn<QualifiedContractIdentifier> for QualifiedContractIdentifier {
     }
 }
 
+/// Make public keys loadable from a sqlite database
+impl FromColumn<Secp256k1PublicKey> for Secp256k1PublicKey {
```
jcnelson, line 234: Is there a reason this isn't in `secp256k1.rs` anymore?

stacks-common/src/types/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+            version: if mainnet {
+                C32_ADDRESS_VERSION_MAINNET_SINGLESIG
+            } else {
+                C32_ADDRESS_VERSION_TESTNET_SINGLESIG
+            },
+            bytes: Hash160([0u8; 20]),
+        }
+    }
+
+    // pub fn from_burnchain_signer(o: &BurnchainSigner, mainnet: bool) -> Option<StacksAddress> {
```
jcnelson, line 140: :eyes:

jcnelson (commented):
So, I think the overall strategy of this refactoring makes sense.  I think the new traits added to libclarity make sense in the absence of higher-order generics.

What I'm still confused about are the `*Extension` traits and implementations.  Why are these necessary, and why can't they live in `stacks_common`?

clarity/src/vm/database/clarity_db.rs
kantai: `insert_microblock_poison` should probably also be moved out of this module.

The reason for removing the method for processing microblock poison is that it is a method that depends on the `StacksChainState` struct, and the Clarity crate cannot depend on anything in the main source tree.

clarity/src/vm/tests/mod.rs
kantai: That the PR is still a draft 😜

The goal of the draft is get some insight on the overall separation, and how it impacts a couple different things (requirements for `...Extension` traits, movement of some implementations, etc.).

src/chainstate/burn/db/sortdb.rs
kantai: Yes -- these implementations, if required, need to moved to `stacks-common`, because this is a foreign trait that is being defined for (now) foreign types. The "foreign trait implementation" constraint in rust is one of the biggest things that necessitates moving implementations around in this PR (e.g., it's why a lot of the codec implementations moved to stacks-common).

src/chainstate/stacks/address.rs
kantai: Yes, this is just inappropriately named in the base branch -- I can rename this method in this refactor.

src/net/mod.rs
kantai: Yes -- the `StacksMessageCodec` trait is defined in `stacks-common`, which means it's a foreign trait, so we cannot implement it for a tuple that contains a foreign type (in this case, this would have been `(ConsensusHash, StacksBlock)`, with the `ConsensusHash` being the foreign type).

src/util_lib/db.rs
kantai: Yes, `FromColumn` and `FromRow` aren't defined in the `stacks-common` crate, which is where `secp256k1` lives.

kantai:
> What I'm still confused about are the *Extension traits and implementations. Why are these necessary, and why can't they live in stacks_common?

The `*Extension` traits are used for crates to be able to define some new instance methods for structs defined in `stacks_common` (because you cant `impl ForeignType { }`). The reason that the various methods in these extension traits aren't just implemented in `stacks-common`, is that they are pretty specific to the crate that defines the extension. For example, `libclarity` has an extension for `StacksAddress` that converts it into a Clarity principal. That's a really useful method and gets invoked a bunch of times in the codebase. However, implementing that in `stacks-common` would require that `stacks-common` contain the Clarity type system. Not an outrageous idea, but with this PR, I'm really trying to limit how much movement in and out of the existing source tree we need to do--- basically just get a first version where at least there's no code duplication. Once the `stacks-blockchain`'s build is separated out, we can do further refactoring, figuring out what else we might want separated, etc., and I think it'll be much less painful than this big refactor.

clarity/src/vm/costs/mod.rs
jcnelson: Is this going to be addressed in this PR?  If not, can you create an issue for it?

clarity/src/vm/database/clarity_db.rs
jcnelson: Thanks, makes sense to me.

clarity/src/vm/costs/mod.rs
kantai: No, I don't think so, so I'll open an issue for it.

clarity/src/vm/docs/mod.rs
kantai: Nope -- docs tests and generation reenabled in 84b4369485e13bc1127bb2eb65fbd053343917d2

clarity/src/vm/docs/contracts.rs
kantai: Dropped now -- this docs generation now gets invoked from the boot module.

clarity/src/vm/docs/contracts.rs
kantai: This is dropped now -- docs tests and generation were reenabled in 84b4369485e13bc1127bb2eb65fbd053343917d2

stacks-common/src/types/mod.rs
kantai: Removed

clarity/src/vm/database/clarity_db.rs
kantai: Removed

clarity/src/vm/analysis/tests/mod.rs
kantai: Moved to `clarity_vm` testing in 9e1f96ad4

clarity/src/vm/analysis/arithmetic_checker/tests.rs
kantai: Moved to `boot::contract_tests` in 9e1f96ad4

kantai: Moved to boot::contract_tests in https://github.com/stacks-network/stacks-blockchain/commit/9e1f96ad419cd6a620c52f4bc040c144fd71e826

src/chainstate/burn/db/sortdb.rs
kantai: Removed the commented out region

jcnelson:
So, this PR overall looks good to me.  Only asks are:

* Make sure that the special case handlers for PoX are installed and working in the `clarity-cli` binary, since there are folks who rely on it today (including CityCoins and some of Marvin's projects).

* Fix `to_b58_version_byte`

* Make sure the integration tests pass

jcnelson (approved):

kantai:
> Make sure that the special case handlers for PoX are installed and working in the clarity-cli binary, since there are folks who rely on it today (including CityCoins and some of Marvin's projects).

The `clarity-cli` uses the `MemoryBackingStore`, which has the special case handler installed.

gregorycoppola (approved):

kantai:
```
    Make sure that the special case handlers for PoX are installed and working in the clarity-cli binary, since there are folks who rely on it today (including CityCoins and some of Marvin's projects).

    Fix to_b58_version_byte

    Make sure the integration tests pass
```

All addressed. The one failing integration test is test flakiness -- it passes locally.

Merging!!

Issue #2969, opened 2021-12-12
Multiple commits for btc block, missed target block


MaksimalistT:
**Describe the bug**
Multiple commits for series of btc blocks detected, missed target block
https://btc.com/btc/transaction/10d26d88f6d2fd18e1b593b74c99886b8da5519ac4b5fd36035043e44ee26345
https://btc.com/btc/transaction/02740d28e351394ce7dac1f7440c59e69d5544ebe3f05f65870f24d7033d3aaf
https://btc.com/btc/transaction/99e7e232c915b120d7d553c92a46f7366d8cd2747db893b53e5c3685d78bcff3
https://btc.com/btc/transaction/7daf7a50fb69abd7c0c3ff2558bc627c9579c246f42c721297680a52184ebd18
https://btc.com/btc/transaction/15643a45ac954a88c11e8d80e76752226fd5dae56ac3698290393a8b67a6f5ff
https://btc.com/btc/transaction/d5dafd4661c4fb3a5bdf8f15f13420193abf6e102b1c657464f7309cceee346b
https://btc.com/btc/transaction/97383a59e877a537d19fbb7b6f4bdc143f98dbf2ab0f0bc47ab4bb09f2cff4ad
https://btc.com/btc/transaction/a264a1508bf1f89bdc2f989d2eb686b968db2b2c7ea07ed023834b35f6a7787b
https://btc.com/btc/transaction/01d6169694017c48b1b6495f91c5b3f3fcd000419812e769f7f26536400dcd6d
https://btc.com/btc/transaction/1ae5682ff9346234c1cb1928999d6b684f393fd24cec826fec4f3b59043d96f9

logs attached, can't provide full debug log

**Steps To Reproduce**
no info

**Expected behavior**
one commit per block

**Environment (please complete the following information):**
- OS: Amazon Linux
- Rust version: rustc 1.49.0 (e1884a8e3 2020-12-29)
- Stacks-node 2.05.0.0.0

**Additional context**
[multiple_commits_main.log](https://github.com/blockstack/stacks-blockchain/files/7698441/multiple_commits_main.log)

jcnelson:
Yup, looks like a bug:

```
INFO [1639286229.697231] [testnet/stacks-node/src/neon_node.rs:1930] [relayer] Succeeded assembling Stacks block #41217: beb76f416f8a6747d9dda5a2c580f2c425fb5a5f3417159a50dec480e2e944f9, with 1 txs, attempt 1
[...]

Issue #2944, opened 2021-11-20
[burnchain-download] download burnchain blocks in parallel to processing sortitions


jcnelson:
Right now, the neon run loop downloads all of a reward cycle's burnchain blocks before it begins processing any of their sortitions.  It *might* be possible to download burnchain blocks in parallel to processing sortiitons.  We'd need to refactor the main loop in `RunLoop::start()` so that we only invoke the PoX sync watchdog once the reward cycle has been downloaded, but we could in theory call `burnchain.sync()` with one block at a time before attempting to process any stacks blocks.  Will need to test the feasibility of this.

jcnelson:
Fixed in #2989

PR #2989, opened 2022-01-09
Minimize time between sortitions and block-acceptance

### LINKED ISSUE #2986: [net] high latency when fetching blocks in certain cases
This issue is meant to track a set of issues related to a high latency observed between processing a sortition and processing the associated Stacks epochs.  In watching a node process blocks both during bootup and steady-state for a few weeks, I've reached the following conclusions:

* During bootup, nodes can wait a _long_ time to fetch a Stacks block for a burnchain block, because the node will process all sortitions before it attempts to download blocks (#2944)

* During steady-state, there are two principal sources of latency: doing a periodic full block inventory sync, and a tendency for blocks to simply propagate slowly.

   * Nodes will periodically (every 12 hours) synchronize *all* block inventories with their neighbors.  If there are slow nodes, this can take a while.  In my analysis of a node's operation for 3,000 blocks, at least 6 blocks took over *10 minutes* to arrive once their sortitions were processed, because the node was spending all that time in the block inventory synchronization step.
   
   * Nodes rarely push blocks directly to one another.  Instead, they send `BlocksAvailable` and `MicroblocksAvailable` messages to remote nodes, with the expectation that the remote node will turn around and request the block and microblock data via the HTTP interface.  The only times they'll push a block or microblock stream directly is when they either (1) mine the block, or (2) notice that a neighbor is missing a block or stream and push it over via the anti-entropy protocol.  The latency induced by not pushing blocks has a very wide distribution, and can add as much as 120 *seconds* of delay between when a sortition is processed and when the block is downloaded.

* (via @kantai) If the node is mining, the node can spend an inordinate amount of time in the `RunTenure` step, but in doing so, will starve itself from running the `ProcessTenure` step (especially if there are many `RunTenure` steps in the pipeline).  The node should immediately broadcast a block or microblock it produces at the end of `RunTenure`, instead of waiting for `ProcessTenure`.

The required fixes are as follows:

- [X] Address #2944 
- [X] Remove the full inventory sync feature.  The node can get away with doing a single full inventory sync when it boots up, and then in the unlikely event that a block or microblock stream from over 2 reward cycles ago becomes available, the anti-entropy protocol can take care of propagating it.  No need to delay block downloads actively searching for missing data in prior reward cycles.
- [X] Forward blocks and microblock streams to outbound peers, unconditionally.  Send `BlocksAvailable` / `MicroblocksAvailable` messages to inbound peers.
- [X] If the miner mined a block in this sortition, then immediately try to push the block to any new neighbors that connect and don't have the block.
- [X] Before trying to mine, verify that the target parent block is still the chain tip.  Drop `RunTenure` requests for which this is not true.
- [X] Immediately broadcast a mined block or microblock once it is produced; don't do so in a subsequent relayer loop pass.
- [ ] Do not query a peer's block inventory more than once per reward cycle during initial block download, since this stalls block downloads.

jcnelson:
This PR fixes #2986, #2944, and #2969.

The overarching goal of this PR is to improve chain quality by minimizing the time between when a sortition happens and when the miner can start producing the next block.  This is achieved in three principal ways:

* **Aggressive block replication.**  This PR makes it so that peers directly push blocks and confirmed microblock streams to their outbound neighbors if they detect that they are missing them from their inventories.  The original behavior was to send a `BlocksAvailable` or `MicroblocksAvailable`.  Instead, these two messages will only be sent to inbound neighbors (whose inventories are not tracked).

* **Minimize time spent doing things besides block downloads.**  This PR removes the full inventory synchronization paths from the inventory state machine, which could cause a node to miss several sortitions' blocks every 12 hours (which had a detrimental effect on miners' ability to produce blocks on the chain tip).  Instead, this PR makes it so that a node will only do a full inventory synchronization with its neighbors on boot-up, and from then on, it will only query the past two reward cycles' inventories.

* **Aggressively cancel stale `RunTenure` events and `RunMicroblockTenure` events.**  This PR updates the mining code-paths to check the current chain tip against each is7ued `RunTenure`, and drop the `RunTenure` if it refers to a now-stale chain tip.  In addition, it checks the canonical Stacks chain tip _after_ block assembly but _before_ sending the block-commit, and cancels the operation if the canonical chain tip has changed.  In addition, a similar set of checks are now done on `RunMicroblockTenure`.  The reason for these changes is that the act of processing a `RunTenure` or `RunMicroblockTenure` directive in the relayer thread can take on the order of 30 seconds, during which time the chain tip they targeted can become stale.  I believe this was the root cause of #2969.  With this change, the node's mock-miner reliably mines atop block-commits on the canonical chain.

  

---

In addition to making these improvements, this PR fixes three bugs, one of which is a show-stopper for 2.05.1.0.

* A **denial-of-service bug** in the mempool synchronization logic.  This **must be merged** into 2.05.0.1.0.  Specifically, the mempool synchronization logic would do a full mempool scan on a query to `/v2/mempool/query`, which could take on the order of 500ms.  This in turn would stall the network.  We didn't notice it before because the effect was small during testing; ALEX's launch revealed that this is a problem.  The fix is to make it so that the mempool synchronization happens in pages, where each page is only permitted to do a fixed number of mempool database queries.

* A **denial-of-service bug** in the Bitcoin indexer whereby it would not be able to recover from a deep reorg.  I actually discovered this first with my appchain MVP, and since noticed that this can happen in Bitcoin (and indeed, occasionally the `bitcoind_forking_test` flaps from it).  Basically, there are cases where the reorg is deeper than the maximum number of blocks the indexer is allowed to synchronize in one go.  If this happens, then the node gets stuck because it forgets to download the blocks in-between the _reorg depth + maximum number_ height and the _chain tip_ height.  This PR fixes this by having the indexer drop the headers after _reorg depth + maximum number_, so the indexer is forced to re-fetch them and their blocks.

* A lot of I/O paths are needlessly slow due to accidental database scans.  This PR adds a set of indexes to the various databases that _significantly_ improve boot-up and steady-state performance.  In my testing, they cut boot-up time _in half_.  I've also addressed #2944, so that the node does not need to wait for all of a reward cycle's sortitions to be processed before downloading Stacks blocks.  I already merged the highest-impact index update, but these ones also help.  In addition, I've instrumented the test code so that if the `BLOCKSTACK_DB_TRACE` envar is set, the node will print out `EXPLAIN QUERY PLAN` statements in the log file that you can go and execute yourself against the test's databases.  Using this new feature, I systematically went through each query that high I/O tests were doing and found queries that were doing (un-`LIMIT`-ed) table scans or creating temporary B-trees, and added indexes to prevent them.

  In order to achieve the above, I refactored `neon_node.rs` and `runloop.rs` in order to gather related state into the `RunLoop` struct, and to break up a lot of path-dependent spaghetti code into single-purpose methods.  This should improve maintainability going forward.
---

I realize that there's a lot here, but time really is of the essence with the mempool being as full as it is.  I have three nodes running with this right now (one from boot-up, two from existing chain state), and they're all doing fine.

### DIFF
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -1855,8 +1855,8 @@ impl MemPoolDB {
 
                 if query.num_txs >= query.max_txs {
                     // no more space in this stream
-                    test_debug!(
-                        "No more space in this query after {:?}",
+                    debug!(
+                        "No more space in this query after {:?}. Corking tx stream.",
                         &query.last_randomized_txid
                     );
 
@@ -1865,10 +1865,6 @@ impl MemPoolDB {
                     query.tx_buf.clear();
                     query.corked = true;
 
-                    test_debug!(
-                        "Cork tx stream with next page {}",
-                        &query.last_randomized_txid
-                    );
                     query
                         .last_randomized_txid
                         .consensus_serialize(&mut query.tx_buf)
@@ -1886,13 +1882,14 @@ impl MemPoolDB {
                         query.max_txs.saturating_sub(query.num_txs),
                     )?;
 
-                test_debug!(
-                    "Visited {} rows from {}, num_txs = {}, max txs = {}",
-                    num_rows_visited,
-                    &query.last_randomized_txid,
-                    query.num_txs,
-                    query.max_txs
+                debug!(
+                    "Streaming mempool propagation stepped";
+                    "rows_visited" => num_rows_visited,
+                    "last_rand_txid" => %query.last_randomized_txid,
+                    "num_txs" => query.num_txs,
+                    "max_txs" => query.max_txs
                 );
+                
                 query.num_txs += num_rows_visited;
                 if let Some(next_tx) = next_txs.pop() {
                     query.tx_buf_ptr = 0;
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -371,6 +371,7 @@ const MEMPOOL_INITIAL_SCHEMA: &'static [&'static str] = &[
     );
     "#,
     "CREATE INDEX by_txid ON mempool(txid);",
+    "CREATE INDEX by_height ON mempool(height);",
     "CREATE INDEX by_txid_and_height ON mempool(txid,height);",
     "CREATE INDEX by_sponsor ON mempool(sponsor_address, sponsor_nonce);",
     "CREATE INDEX by_origin ON mempool(origin_address, origin_nonce);",
@@ -417,6 +418,7 @@ const MEMPOOL_SCHEMA_3_BLOOM_STATE: &'static [&'static str] = &[
         hashed_txid TEXT NOT NULL,
         FOREIGN KEY(txid) REFERENCES mempool(txid) ON DELETE CASCADE
     );
+    CREATE INDEX IF NOT EXISTS by_ordered_hashed_txid ON randomized_txids(hashed_txid ASC);
     CREATE INDEX IF NOT EXISTS by_hashed_txid ON randomized_txids(txid,hashed_txid);
     "#,
     r#"
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -1808,7 +1808,8 @@ impl MemPoolDB {
         Ok((ret, next_last_randomized_txid))
     }
 
-    /// Stream transaction data
+    /// Stream transaction data.
+    /// Send back one transaction at a time.
     pub fn stream_txs<W: Write>(
         &self,
         fd: &mut W,
@@ -1873,7 +1874,7 @@ impl MemPoolDB {
                     // if we can send at least one transaction, or at least one byte, by the time
                     // we finish sending a transaction, then stop early in order to limit the
                     // amount of I/O that happens in a stream.
-                    break;
+                    return Err(ChainstateError::Yield(num_written));
                 }
             }
         }
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -1740,6 +1740,7 @@ impl MemPoolDB {
     /// (so if some nodes are configured to return fewer than MAX_BLOOM_COUNTER_TXS transactions,
     /// a requesting node will still have a good chance of getting something useful).
     /// Also, return the next value to pass for `last_randomized_txid` to load the next page.
+    /// Also, return the number of rows considered.
     pub fn find_next_missing_transactions(
         &self,
         data: &MemPoolSyncData,
@@ -1747,7 +1748,7 @@ impl MemPoolDB {
         last_randomized_txid: &Txid,
         max_txs: u64,
         max_run: u64,
-    ) -> Result<(Vec<StacksTransaction>, Option<Txid>), db_error> {
+    ) -> Result<(Vec<StacksTransaction>, Option<Txid>, u64), db_error> {
         let mut ret = vec![];
         let sql = "SELECT mempool.txid as txid, mempool.tx as tx \
                    FROM mempool JOIN randomized_txids \
@@ -1773,9 +1774,18 @@ impl MemPoolDB {
 
         let mut stmt = self.conn().prepare(sql)?;
         let mut rows = stmt.query(args)?;
+        let mut num_rows_visited = 0;
+        let mut last_txid = None;
         while let Some(row) = rows.next()? {
             let txid = Txid::from_column(row, "txid")?;
-            test_debug!("Consider txid {}", &txid);
+            last_txid = Some(txid.clone());
+            num_rows_visited += 1;
+
+            test_debug!(
+                "Consider txid {} at or after {}",
+                &txid,
+                last_randomized_txid
+            );
             let contains = match data {
                 MemPoolSyncData::BloomFilter(ref bf) => bf.contains_raw(&txid.0),
                 MemPoolSyncData::TxTags(ref seed, ..) => {
@@ -1799,13 +1809,13 @@ impl MemPoolDB {
         }
 
         // find next page, if needed
-        let next_last_randomized_txid = if let Some(last_tx) = ret.last() {
-            self.get_randomized_txid(&last_tx.txid())?
+        let next_last_randomized_txid = if let Some(last_txid) = last_txid {
+            self.get_randomized_txid(&last_txid)?
         } else {
             None
         };
 
-        Ok((ret, next_last_randomized_txid))
+        Ok((ret, next_last_randomized_txid, num_rows_visited))
     }
 
     /// Stream transaction data.
@@ -1817,13 +1827,7 @@ impl MemPoolDB {
         count: u64,
     ) -> Result<u64, ChainstateError> {
         let mut num_written = 0;
-        let mut num_txs = 0;
         while num_written < count {
-            if query.num_txs >= query.max_txs {
-                // don't serve more than this many txs
-                break;
-            }
-
             // write out bufferred tx
             let start = query.tx_buf_ptr;
             let end = cmp::min(query.tx_buf.len(), ((start as u64) + count) as usize);
@@ -1836,20 +1840,59 @@ impl MemPoolDB {
             num_written += nw;
 
             if query.tx_buf_ptr >= query.tx_buf.len() {
+                if query.corked {
+                    // we're done
+                    test_debug!(
+                        "Finished streaming txs; last page was {:?}",
+                        &query.last_randomized_txid
+                    );
+                    break;
+                }
+
+                if query.num_txs >= query.max_txs {
+                    // no more space in this stream
+                    test_debug!(
+                        "No more space in this query after {:?}",
+                        &query.last_randomized_txid
+                    );
+
+                    // send the next page ID
+                    query.tx_buf_ptr = 0;
+                    query.tx_buf.clear();
+                    query.corked = true;
+
+                    test_debug!(
+                        "Cork tx stream with next page {}",
+                        &query.last_randomized_txid
+                    );
+                    query
+                        .last_randomized_txid
+                        .consensus_serialize(&mut query.tx_buf)
+                        .map_err(ChainstateError::CodecError)?;
+                    break;
+                }
+
                 // load next
-                let (mut next_txs, next_last_randomized_txid_opt) = self
+                let (mut next_txs, next_last_randomized_txid_opt, num_rows_visited) = self
                     .find_next_missing_transactions(
                         &query.tx_query,
                         query.height,
                         &query.last_randomized_txid,
                         1,
-                        MAX_BLOOM_COUNTER_TXS.into(),
+                        MAX_BLOOM_COUNTER_TXS as u64,
                     )?;
+
+                test_debug!(
+                    "Visited {} rows from {}, num_txs = {}, max txs = {}",
+                    num_rows_visited,
+                    &query.last_randomized_txid,
+                    query.num_txs,
+                    query.max_txs
+                );
+                query.num_txs += num_rows_visited;
                 if let Some(next_tx) = next_txs.pop() {
                     query.tx_buf_ptr = 0;
                     query.tx_buf.clear();
-                    query.num_txs += 1;
-                    num_txs += 1;
 
                     next_tx
                         .consensus_serialize(&mut query.tx_buf)
@@ -1861,20 +1904,36 @@ impl MemPoolDB {
                         test_debug!("No more txs after {}", &next_tx.txid());
                         break;
                     }
-                } else {
-                    // no more
+                } else if let Some(next_txid) = next_last_randomized_txid_opt {
+                    test_debug!("No rows returned for {}", &query.last_randomized_txid);
+
+                    // no rows found
+                    query.last_randomized_txid = next_txid;
+
+                    // send the next page ID
+                    query.tx_buf_ptr = 0;
+                    query.tx_buf.clear();
+                    query.corked = true;
+
                     test_debug!(
-                        "No more txs in query after {:?}",
+                        "Cork tx stream with next page {}",
                         &query.last_randomized_txid
                     );
+                    query
+                        .last_randomized_txid
+                        .consensus_serialize(&mut query.tx_buf)
+                        .map_err(ChainstateError::CodecError)?;
                     break;
-                }
+                } else if next_last_randomized_txid_opt.is_none() {
+                    // no more transactions
+                    test_debug!(
+                        "No more txs to send after {:?}",
+                        &query.last_randomized_txid
+                    );
 
-                if num_written > 0 && num_txs > 1 {
-                    // if we can send at least one transaction, or at least one byte, by the time
-                    // we finish sending a transaction, then stop early in order to limit the
-                    // amount of I/O that happens in a stream.
-                    return Err(ChainstateError::Yield(num_written));
+                    query.tx_buf_ptr = 0;
+                    query.tx_buf.clear();
+                    query.corked = true;
                 }
             }
         }
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -1777,6 +1777,10 @@ impl MemPoolDB {
         let mut num_rows_visited = 0;
         let mut last_txid = None;
         while let Some(row) = rows.next()? {
+            if num_rows_visited >= max_run {
+                break;
+            }
+
             let txid = Txid::from_column(row, "txid")?;
             last_txid = Some(txid.clone());
             num_rows_visited += 1;
@@ -1869,7 +1873,7 @@ impl MemPoolDB {
                         .last_randomized_txid
                         .consensus_serialize(&mut query.tx_buf)
                         .map_err(ChainstateError::CodecError)?;
-                    break;
+                    continue;
                 }
 
                 // load next
@@ -1879,7 +1883,7 @@ impl MemPoolDB {
                         query.height,
                         &query.last_randomized_txid,
                         1,
-                        MAX_BLOOM_COUNTER_TXS as u64,
+                        query.max_txs.saturating_sub(query.num_txs),
                     )?;
 
                 test_debug!(
@@ -1923,11 +1927,10 @@ impl MemPoolDB {
                         .last_randomized_txid
                         .consensus_serialize(&mut query.tx_buf)
                         .map_err(ChainstateError::CodecError)?;
-                    break;
                 } else if next_last_randomized_txid_opt.is_none() {
                     // no more transactions
                     test_debug!(
-                        "No more txs to send after {:?}",
+                        "No more txs to send after {:?}; corking stream",
                         &query.last_randomized_txid
                     );
 
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -1816,6 +1816,7 @@ impl MemPoolDB {
         count: u64,
     ) -> Result<u64, ChainstateError> {
         let mut num_written = 0;
+        let mut num_txs = 0;
         while num_written < count {
             if query.num_txs >= query.max_txs {
                 // don't serve more than this many txs
@@ -1847,6 +1848,7 @@ impl MemPoolDB {
                     query.tx_buf_ptr = 0;
                     query.tx_buf.clear();
                     query.num_txs += 1;
+                    num_txs += 1;
 
                     next_tx
                         .consensus_serialize(&mut query.tx_buf)
@@ -1864,6 +1866,13 @@ impl MemPoolDB {
                         "No more txs in query after {:?}",
                         &query.last_randomized_txid
                     );
+                    break;
+                }
+
+                if num_written > 0 && num_txs > 1 {
+                    // if we can send at least one transaction, or at least one byte, by the time
+                    // we finish sending a transaction, then stop early in order to limit the
+                    // amount of I/O that happens in a stream.
                     break;
                 }
             }
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -1889,7 +1889,7 @@ impl MemPoolDB {
                     "num_txs" => query.num_txs,
                     "max_txs" => query.max_txs
                 );
-                
+
                 query.num_txs += num_rows_visited;
                 if let Some(next_tx) = next_txs.pop() {
                     query.tx_buf_ptr = 0;
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -1750,7 +1750,7 @@ impl MemPoolDB {
         max_run: u64,
     ) -> Result<(Vec<StacksTransaction>, Option<Txid>, u64), db_error> {
         let mut ret = vec![];
-        let sql = "SELECT mempool.txid as txid, mempool.tx as tx \
+        let sql = "SELECT mempool.txid AS txid, mempool.tx AS tx, randomized_txids.hashed_txid AS hashed_txid \
                    FROM mempool JOIN randomized_txids \
                    ON mempool.txid = randomized_txids.txid \
                    WHERE randomized_txids.hashed_txid > ?1 \
@@ -1775,21 +1775,24 @@ impl MemPoolDB {
         let mut stmt = self.conn().prepare(sql)?;
         let mut rows = stmt.query(args)?;
         let mut num_rows_visited = 0;
-        let mut last_txid = None;
+        let mut next_page = None;
         while let Some(row) = rows.next()? {
             if num_rows_visited >= max_run {
                 break;
             }
 
             let txid = Txid::from_column(row, "txid")?;
-            last_txid = Some(txid.clone());
             num_rows_visited += 1;
 
+            let hashed_txid = Txid::from_column(row, "hashed_txid")?;
             test_debug!(
-                "Consider txid {} at or after {}",
+                "Consider txid {} ({}) at or after {}",
                 &txid,
+                &hashed_txid,
                 last_randomized_txid
             );
+            next_page = Some(hashed_txid);
+
             let contains = match data {
                 MemPoolSyncData::BloomFilter(ref bf) => bf.contains_raw(&txid.0),
                 MemPoolSyncData::TxTags(ref seed, ..) => {
@@ -1812,14 +1815,7 @@ impl MemPoolDB {
             }
         }
 
-        // find next page, if needed
-        let next_last_randomized_txid = if let Some(last_txid) = last_txid {
-            self.get_randomized_txid(&last_txid)?
-        } else {
-            None
-        };
-
-        Ok((ret, next_last_randomized_txid, num_rows_visited))
+        Ok((ret, next_page, num_rows_visited))
     }
 
     /// Stream transaction data.
@@ -1873,13 +1869,14 @@ impl MemPoolDB {
                 }
 
                 // load next
-                let (mut next_txs, next_last_randomized_txid_opt, num_rows_visited) = self
+                let remaining = query.max_txs.saturating_sub(query.num_txs);
+                let (next_txs, next_last_randomized_txid_opt, num_rows_visited) = self
                     .find_next_missing_transactions(
                         &query.tx_query,
                         query.height,
                         &query.last_randomized_txid,
                         1,
-                        query.max_txs.saturating_sub(query.num_txs),
+                        remaining,
                     )?;
 
                 debug!(
@@ -1891,22 +1888,33 @@ impl MemPoolDB {
                 );
 
                 query.num_txs += num_rows_visited;
-                if let Some(next_tx) = next_txs.pop() {
+                if next_txs.len() > 0 {
                     query.tx_buf_ptr = 0;
                     query.tx_buf.clear();
 
-                    next_tx
-                        .consensus_serialize(&mut query.tx_buf)
-                        .map_err(ChainstateError::CodecError)?;
-
+                    for next_tx in next_txs.iter() {
+                        next_tx
+                            .consensus_serialize(&mut query.tx_buf)
+                            .map_err(ChainstateError::CodecError)?;
+                    }
                     if let Some(next_last_randomized_txid) = next_last_randomized_txid_opt {
                         query.last_randomized_txid = next_last_randomized_txid;
                     } else {
-                        test_debug!("No more txs after {}", &next_tx.txid());
+                        test_debug!(
+                            "No more txs after {}",
+                            &next_txs
+                                .last()
+                                .map(|tx| tx.txid())
+                                .unwrap_or(Txid([0u8; 32]))
+                        );
                         break;
                     }
                 } else if let Some(next_txid) = next_last_randomized_txid_opt {
-                    test_debug!("No rows returned for {}", &query.last_randomized_txid);
+                    test_debug!(
+                        "No rows returned for {}; cork tx stream with next page {}",
+                        &query.last_randomized_txid,
+                        &next_txid
+                    );
 
                     // no rows found
                     query.last_randomized_txid = next_txid;
@@ -1916,10 +1924,6 @@ impl MemPoolDB {
                     query.tx_buf.clear();
                     query.corked = true;
 
-                    test_debug!(
-                        "Cork tx stream with next page {}",
-                        &query.last_randomized_txid
-                    );
                     query
                         .last_randomized_txid
                         .consensus_serialize(&mut query.tx_buf)
--- a/src/net/connection.rs
+++ b/src/net/connection.rs
@@ -464,8 +464,8 @@ impl std::default::Default for ConnectionOptions {
             max_buffered_blocks: 1,
             max_buffered_microblocks: 10,
             mempool_sync_interval: 30, // number of seconds in-between mempool sync
-            mempool_max_tx_query: 128, // maximum number of transactions to return
-            mempool_sync_timeout: 120, // how long a mempool sync can go for
+            mempool_max_tx_query: 128, // maximum number of transactions to visit per mempool query
+            mempool_sync_timeout: 180, // how long a mempool sync can go for (3 minutes)
 
             // no faults on by default
             disable_neighbor_walk: false,
@@ -1093,15 +1093,13 @@ impl<P: ProtocolFamily> ConnectionOutbox<P> {
 
                     self.socket_out_buf.extend_from_slice(&buf[0..nr_input]);
 
-                    if nr_input >= 0 {
-                        test_debug!(
-                            "Connection buffered {} bytes from pipe ({} total, ptr = {}, blocked = {})",
-                            nr_input,
-                            self.socket_out_buf.len(),
-                            self.socket_out_ptr,
-                            blocked
-                        );
-                    }
+                    test_debug!(
+                        "Connection buffered {} bytes from pipe ({} total, ptr = {}, blocked = {})",
+                        nr_input,
+                        self.socket_out_buf.len(),
+                        self.socket_out_ptr,
+                        blocked
+                    );
                     nr_input
                 }
                 None => {
--- a/src/net/connection.rs
+++ b/src/net/connection.rs
@@ -464,7 +464,7 @@ impl std::default::Default for ConnectionOptions {
             max_buffered_blocks: 1,
             max_buffered_microblocks: 10,
             mempool_sync_interval: 30, // number of seconds in-between mempool sync
-            mempool_max_tx_query: MAX_BLOOM_COUNTER_TXS.into(),
+            mempool_max_tx_query: 128, // maximum number of transactions to return
             mempool_sync_timeout: 120, // how long a mempool sync can go for
 
             // no faults on by default
--- a/src/net/connection.rs
+++ b/src/net/connection.rs
@@ -1060,7 +1060,10 @@ impl<P: ProtocolFamily> ConnectionOutbox<P> {
                             message_eof = true;
                             0
                         }
-                        Ok(read_len) => read_len,
+                        Ok(read_len) => {
+                            test_debug!("Connection message pipe returned {} bytes", read_len);
+                            read_len
+                        }
                         Err(ioe) => match ioe.kind() {
                             io::ErrorKind::WouldBlock => {
                                 // no data consumed, but we may need to make a break for it
@@ -1090,12 +1093,13 @@ impl<P: ProtocolFamily> ConnectionOutbox<P> {
 
                     self.socket_out_buf.extend_from_slice(&buf[0..nr_input]);
 
-                    if nr_input > 0 {
-                        trace!(
-                            "Connection buffered {} bytes from pipe ({} total, ptr = {})",
+                    if nr_input >= 0 {
+                        test_debug!(
+                            "Connection buffered {} bytes from pipe ({} total, ptr = {}, blocked = {})",
                             nr_input,
                             self.socket_out_buf.len(),
-                            self.socket_out_ptr
+                            self.socket_out_ptr,
+                            blocked
                         );
                     }
                     nr_input
@@ -1136,7 +1140,7 @@ impl<P: ProtocolFamily> ConnectionOutbox<P> {
 
                 self.socket_out_ptr += num_written;
 
-                trace!(
+                test_debug!(
                     "Connection wrote {} bytes to socket (buffer len = {}, ptr = {})",
                     num_written,
                     self.socket_out_buf.len(),
@@ -1157,7 +1161,7 @@ impl<P: ProtocolFamily> ConnectionOutbox<P> {
             }
         }
 
-        trace!(
+        test_debug!(
             "Connection send_bytes finished: blocked = {}, disconnected = {}",
             blocked,
             disconnected
--- a/src/net/connection.rs
+++ b/src/net/connection.rs
@@ -465,7 +465,7 @@ impl std::default::Default for ConnectionOptions {
             max_buffered_microblocks: 10,
             mempool_sync_interval: 30, // number of seconds in-between mempool sync
             mempool_max_tx_query: MAX_BLOOM_COUNTER_TXS.into(),
-            mempool_sync_timeout: 60, // how long a mempool sync can go for
+            mempool_sync_timeout: 120, // how long a mempool sync can go for
 
             // no faults on by default
             disable_neighbor_walk: false,
--- a/src/net/http.rs
+++ b/src/net/http.rs
@@ -1729,6 +1729,27 @@ impl HttpRequestType {
         }
     }
 
+    /// get the mempool page ID optional query argument (`page_id`)
+    /// Take the first value we can parse.
+    fn get_mempool_page_id_query(query: Option<&str>) -> Option<Txid> {
+        match query {
+            Some(query_string) => {
+                for (key, value) in form_urlencoded::parse(query_string.as_bytes()) {
+                    if key != "page_id" {
+                        continue;
+                    }
+                    if let Ok(page_id) = Txid::from_hex(&value) {
+                        return Some(page_id);
+                    }
+                }
+                return None;
+            }
+            None => {
+                return None;
+            }
+        }
+    }
+
     fn parse_get_account<R: Read>(
         _protocol: &mut StacksHttp,
         preamble: &HttpRequestPreamble,
@@ -2556,7 +2577,7 @@ impl HttpRequestType {
         _protocol: &mut StacksHttp,
         preamble: &HttpRequestPreamble,
         _regex: &Captures,
-        _query: Option<&str>,
+        query: Option<&str>,
         fd: &mut R,
     ) -> Result<HttpRequestType, net_error> {
         if preamble.get_content_length() == 0 {
@@ -2590,10 +2611,12 @@ impl HttpRequestType {
 
         let mut bound_fd = BoundReader::from_reader(fd, preamble.get_content_length() as u64);
         let mempool_query = MemPoolSyncData::consensus_deserialize(&mut bound_fd)?;
+        let page_id_opt = HttpRequestType::get_mempool_page_id_query(query);
 
         Ok(HttpRequestType::MemPoolQuery(
             HttpRequestMetadata::from_preamble(preamble),
             mempool_query,
+            page_id_opt,
         ))
     }
 
@@ -2672,7 +2695,7 @@ impl HttpRequestType {
         }
     }
 
-    fn make_query_string(tip_req: &TipRequest, with_proof: bool) -> String {
+    fn make_tip_query_string(tip_req: &TipRequest, with_proof: bool) -> String {
         match tip_req {
             TipRequest::UseLatestUnconfirmedTip => {
                 format!("?tip=latest{}", if with_proof { "" } else { "&proof=0" })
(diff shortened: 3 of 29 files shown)


kantai:
Any progress on this?

The network has been experiencing very high orphan rates, so I think the priority of resolving https://github.com/blockstack/stacks-blockchain/issues/2986 should be pretty high. There may be other contributing factors, so it would be useful to figure out how much this PR would improve the current situation.

kantai:
> First, as you know, the chains-coordinator thread will process either a Stacks block, a burnchain block, or neither, but never both at the same time. What this means is that when the node is booting, there's no way to download and process burnchain blocks in parallel to processing Stacks blocks -- the coordinator serializes them regardless of what we do in the sync_with_indexer() method or the Neon run loop. While this PR does change the Neon run loop so that burnchain blocks are downloaded in parallel to Stacks block processing, this does not address this bottleneck.

I think that's fine -- I'm hesitant to try to parallelize the processing in the coordinator (and would definitely want any such change isolated) because the logic of tracking PoX cycles relies on that serialization. Not that there aren't ways to parallelize this, but I think it's non-trivial. I'm also somewhat skeptical of this being a serious bottleneck--- if this were indeed a bottleneck, you'd expect to see _sustained_ high CPU usage during bootup, which we don't currently.

> Second, I discovered that as the bootup progresses, it becomes more and more likely that the chains-coordinator thread spends a lot of time in the handle_new_burnchain_block() loop, which queries multiple unprocessed burnchain blocks: https://github.com/blockstack/stacks-blockchain/blob/master/src/chainstate/coordinator/mod.rs#L570. In the worst cases, this stall can take on the order of 100 seconds, and the returned list of unprocessed burnchain blocks is on the order of 1,000 entries long. This loop returns multiple entries about 5% of the time during bootup, and they get longer as the node runs. This could be due to the fact that a slew of Stacks blocks getting processed at once can stall sortition processing, but not stall burnchain block downloads.

Right, there may be some simple heuristics that could speed up that fetch loop. For example, before the loop, that path could lookup the canonical sortition db height (i.e., the highest processed burn block in the sortition db), and then find the canonical burn block's parent at that height from the burn db. That way, in most cases, the loop only needs to consult a marf once or twice. I'm still kind of skeptical that this is really impacting the boot-up, though. It's true that the list of unprocessed blocks can get really long, but I don't see how it would stall progress -- those burn blocks really do need to have their sortitions processed, and I'm not sure why it would impact boot-up time to have them all processed in a large chunk (it may even speed up progress: stacks blocks won't be downloaded until those sortitions are processed, right?)

> I haven't had a chance yet to modify the miner run loop behavior to make it coalesce ProcessTenure and RunTenure requests. I'll get to that hopefully tomorrow.

There may be some other behaviors that should also be looked at regarding the orphan rate. For example, miner's "wait" behavior (different from the first-commit time, I think) would impact orphaning --- if miners issue their next commitment 10-20 seconds after receiving the last burn block, we should expect a lot of orphans on that first commit regardless of block propagation: validating a stacks blocks' Clarity code can take up to ~30 seconds.

jcnelson:
> I think that's fine -- I'm hesitant to try to parallelize the processing in the coordinator (and would definitely want any such change isolated) because the logic of tracking PoX cycles relies on that serialization.

Yup, that's about where I'm at as well.

> I'm also somewhat skeptical of this being a serious bottleneck--- if this were indeed a bottleneck, you'd expect to see sustained high CPU usage during bootup, which we don't currently.

Wait, you don't?  I see high CPU load all the time on `release`.

Also, why wouldn't it be a bottleneck?  The node currently alternates between processing sortitions and Stacks blocks, but it could conceivably process a whole reward cycle's sortitions concurrently with downloading its blocks.  So instead of taking _O(m + n)_ time for _m_ sortitions and _n_ Stacks blocks, bootup could go as fast as _O(max(m, n))_.  Sortition-processing would only ever need to block in order to wait for the anchor block for the next reward cycle.

>  That way, in most cases, the loop only needs to consult a marf once or twice. I'm still kind of skeptical that this is really impacting the boot-up, though. It's true that the list of unprocessed blocks can get really long, but I don't see how it would stall progress -- those burn blocks really do need to have their sortitions processed, and I'm not sure why it would impact boot-up time to have them all processed in a large chunk (it may even speed up progress: stacks blocks won't be downloaded until those sortitions are processed, right?)

Right -- these blocks need to be processed one way or another.  But the problem is that while this loop is running, it blocks processing for _both_ Stacks blocks and sortitions -- even Stacks blocks that are otherwise ready to be processed from already-processed sortitions.  But I don't think it's a major source of slowness, since it doesn't happen very often (and the worst offending outlier is on the order of two minutes).

> There may be some other behaviors that should also be looked at regarding the orphan rate. For example, miner's "wait" behavior (different from the first-commit time, I think) would impact orphaning --- if miners issue their next commitment 10-20 seconds after receiving the last burn block, we should expect a lot of orphans on that first commit regardless of block propagation: validating a stacks blocks' Clarity code can take up to ~30 seconds.

I can look into this more.

jcnelson:
Note to self: still to do on this: update the `inv.rs` state machine so that it only pulls inventories once per complete reward cycle during initial boot-up.  Requires passing highest reward cycle known at boot-up into the `PeerNetwork`.

jcnelson:
Okay, we're good to go on reviews here.

gregorycoppola (approved):
if you think it's working, sgtm

kantai:
Can you update the `CHANGELOG.md` file? There's several updates in this PR that seem like they would qualify for entries there.

src/core/mempool.rs
```diff
@@ -1832,19 +1844,59 @@ impl MemPoolDB {
...
+
+                    // send the next page ID
+                    query.tx_buf_ptr = 0;
+                    query.tx_buf.clear();
+                    query.corked = true;
+
+                    test_debug!(
+                        "Cork tx stream with next page {}",
+                        &query.last_randomized_txid
+                    );
```
kantai, line 1871: Can this `test_debug!` be a `debug!` log, and combined with the `test_debug!` log line at 1858 -- this could be useful in debugging a running node's behavior, but it seems like they can be one log line rather than two.

src/core/mempool.rs
```diff
@@ -1832,19 +1844,59 @@ impl MemPoolDB {
...
+                        query.max_txs.saturating_sub(query.num_txs),
                     )?;
+
+                test_debug!(
+                    "Visited {} rows from {}, num_txs = {}, max txs = {}",
+                    num_rows_visited,
+                    &query.last_randomized_txid,
+                    query.num_txs,
+                    query.max_txs
+                );
```
kantai, line 1895: Can this also be a `debug!` log, with something more descriptive than `Visited` as it's message, perhaps:

```suggestion
                debug!(
                    "Streaming mempool propagation stepped";
                    "rows_visited" => num_rows_visited,
                    "last_rand_txid" => %query.last_randomized_txid,
                    "num_txs" => query.num_txs,
                    "max_txs" => query.max_txs
                );
```

src/net/connection.rs
```diff
@@ -457,15 +457,15 @@ impl std::default::Default for ConnectionOptions {
...
+            antientropy_retry: 60,   // retry pushing data once every minute
             antientropy_public: true, // run antientropy even if we're NOT NAT'ed
             max_buffered_blocks_available: 1,
             max_buffered_microblocks_available: 1,
             max_buffered_blocks: 1,
             max_buffered_microblocks: 10,
             mempool_sync_interval: 30, // number of seconds in-between mempool sync
-            mempool_max_tx_query: MAX_BLOOM_COUNTER_TXS.into(),
-            mempool_sync_timeout: 60, // how long a mempool sync can go for
+            mempool_max_tx_query: 128, // maximum number of transactions to return
```
kantai, line 467: This used to be `MAX_BLOOM_COUNTER_TXS`, is that constant no longer used anywhere? Should it be updated to 128?

src/net/rpc.rs
```diff
@@ -2087,14 +2089,22 @@ impl ConversationHttp {
...
             .map(|blk| blk.height)
             .unwrap_or(0);
-        let stream = Some(StreamCursor::new_tx_stream(query, max_txs, height));
+
+        test_debug!(
+            "Begin mempool stream from height {} page_id {:?} for up to {} txs",
+            height,
+            &page_id,
+            max_txs
+        );
```
kantai, line 2106: This also seems useful as a `debug!` (and k-v formatted)

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -6569,3 +6577,69 @@ fn use_latest_tip_integration_test() {
     )
     .is_ok());
 }
+
+#[test]
+#[ignore]
+fn test_flash_block_skip_tenure() {
```
kantai, line 6583: Should this test be added to the CI test matrix? Look at `.github/workflows/bitcoin-tests.yml`

testnet/stacks-node/src/config.rs
```diff
@@ -362,6 +362,9 @@ impl Config {
                     microblock_frequency: node
                         .microblock_frequency
                         .unwrap_or(default_node_config.microblock_frequency),
+                    microblock_attempt_time_ms: node
+                        .microblock_attempt_time_ms
+                        .unwrap_or(default_node_config.microblock_attempt_time_ms),
```
kantai, line 367: What does `microblock_attempt_time_ms` control? Can you add something to the Mining section in `README.md` that describes the new config option?

kantai:
Rerunning the CI actions to see if it can produce code coverage correctly.

src/net/http.rs
```diff
@@ -3747,36 +3775,141 @@ impl HttpResponseType {
...
+            let pos = retry_reader.position();
+            let next_msg: Result<StacksTransaction, _> = read_next(&mut retry_reader);
+            match next_msg {
+                Ok(tx) => {
+                    if expect_eof {
+                        // this should have failed
+                        test_debug!("Expected EOF; got transaction {}", tx.txid());
+                        return Err(net_error::DeserializeError(
+                            "Expected end-of-stream".to_string(),
+                        ));
```
kantai, line 3841: Would it make sense to add a `net_error::ExpectedEndOfStream` variant? This same error string is produced three times here.

src/net/relay.rs
```diff
@@ -932,35 +937,44 @@ impl Relayer {
...
                 Relayer::preprocess_downloaded_blocks(&sort_ic, network_result, chainstate);
-            for new_dled_block in new_dled_blocks.into_iter() {
-                debug!("Received downloaded block for {}", &new_dled_block);
-                new_blocks.insert(new_dled_block);
+            for (new_dled_block_ch, block_data) in new_dled_blocks.into_iter() {
+                debug!(
+                    "Received downloaded block for {}/{}",
+                    &new_dled_block_ch,
+                    &block_data.block_hash()
+                );
```
kantai, line 945:
```suggestion
                debug!(
                    "Received downloaded block",
                    "consensus_hash" => %new_dled_block_ch,
                    "anchor_block_hash" => %block_data.block_hash(),
                );
```

Same suggestion for the other debug lines in the function too.

testnet/stacks-node/src/neon_node.rs
```diff
@@ -96,13 +92,14 @@ enum RelayerDirective {
...
     Exit,
 }
 
-pub struct InitializedNeonNode {
+pub struct StacksNode {
     config: Config,
     relay_channel: SyncSender<RelayerDirective>,
+    last_sortition_mutex: Arc<Mutex<Option<BlockSnapshot>>>,
     burnchain_signer: BurnchainSigner,
     last_burn_block: Option<BlockSnapshot>,
```
kantai, line 104: What's the difference between `last_burn_block` and `last_sortition_mutex`?

kantai: (Also, naming peeve: `last_sortition_mutex` doesn't need to tell us it's a mutex-- it's typed as one)

testnet/stacks-node/src/neon_node.rs
```diff
@@ -2007,6 +2039,41 @@ impl InitializedNeonNode {
...
+                    &anchored_block.header.parent_block,
+                    &anchored_block.header.parent_microblock,
+                    &anchored_block.header.parent_microblock_sequence,
+                    &burn_block.burn_header_hash,
+                    burn_block.block_height,
+                    burn_block.sortition_id,
+                    attempt,
+                    stacks_tip.anchored_block_hash,
+                    stacks_tip.consensus_hash,
+                    cur_burn_chain_tip.sortition_id
```
kantai, line 2071: This should be a k-v structured log. Message should just be "Canceling block-commit for block", and then all of the rest of these values should be k-v pairs.

testnet/stacks-node/src/neon_node.rs
```diff
@@ -112,34 +109,28 @@ pub struct InitializedNeonNode {
...
+        Ok(tenure_str) => match tenure_str.parse::<u64>() {
+            Ok(tenure_time) => {
+                info!(
+                    "Fault injection: sleeping for {} milliseconds to simulate a long tenure",
+                    tenure_time
+                );
+                stacks::util::sleep_ms(tenure_time);
+            }
+            Err(_) => {
+                error!("Parse error for STX_TEST_SLOW_TENURE");
```
kantai, line 125: Should this panic the test?

testnet/stacks-node/src/neon_node.rs
```diff
@@ -1475,11 +1493,22 @@ impl InitializedNeonNode {
 
         if let Some(ref snapshot) = &self.last_burn_block {
             debug!(
-                "Tenure: Notify sortition! Last snapshot is {}/{} ({})",
+                "Tenure: Notify sortition! Last snapshot is {}/{} ({}) sortition {}",
                 &snapshot.consensus_hash,
                 &snapshot.burn_header_hash,
-                &snapshot.winning_stacks_block_hash
+                &snapshot.winning_stacks_block_hash,
+                &snapshot.sortition_id
```
kantai, line 1500: This would be much more readable as a k-v pattern.

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -394,156 +472,229 @@ impl RunLoop {
...
 
                 info!("Exiting stacks-node");
                 break;
             }
 
+            let remote_chain_height = burnchain.get_headers_height();
+
             // wait for the p2p state-machine to do at least one pass
-            debug!("Wait until we reach steady-state before processing more burnchain blocks...");
+            debug!("Wait until we reach steady-state before processing more burnchain blocks (chain height is {}, we are at {})...", remote_chain_height, burnchain_height);
```
kantai, line 577: This log could be more descriptive about what's happening here. What does "reach steady-state" mean?

Would this describe what's happening at this step:

```
debug!("Waiting for Stacks block downloads to reach steady-state before processing more burnchain blocks..."; "known_burn_height" => remote_chain_height, "our_burn_height" => burnchain_height);
```

But isn't this step waiting on more _Stacks_ blocks to be processed, rather than bitcoin blocks? If the log line was providing a status update, isn't the current stacks block more important than the current burnchain blocks?

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -394,156 +472,229 @@ impl RunLoop {
...
+                            warn!("Burnchain controller stopped: {}", e);
+                            continue;
+                        }
+                    };
 
-            let sortition_tip = &burnchain_tip.block_snapshot.sortition_id;
-            let next_height = burnchain_tip.block_snapshot.block_height;
+                // *now* we know the burnchain height
+                burnchain_tip = next_burnchain_tip;
+                burnchain_height = cmp::min(burnchain_height + 1, tip_burnchain_height);
```
kantai, line 617: What does limiting the increase of `burnchain_height` by 1 do in this loop? It seems like `burnchain_height` is only used as a parameter to `bitcoin_controller.sync(...)`, so each trip through the loop will always invoke with +1 of the prior invocation. If `burnchain.sync` returns a more advanced chain tip, why not progress burnchain_height to the new sortition height?

kantai:
This looks good to me -- mostly just superficial comments. I would like to be able to see some useful code coverage results though (not sure why codecov is so borked on this PR right now), as these changes do touch a lot of different areas, and there's some corner cases in the networking changes that we should make sure are covered.

src/core/mempool.rs
jcnelson: Yup, fixed

src/core/mempool.rs
jcnelson: Yup, changed

src/net/connection.rs
jcnelson: Yes; this constant is used in the mempool bloom filter implementation.

This number controls how many transactions will be _visited_ in the DB, not returned.  I'll update the field member.

src/net/rpc.rs
jcnelson: Fixed

src/net/rpc.rs
```diff
@@ -2087,14 +2089,22 @@ impl ConversationHttp {
         chainstate: &StacksChainState,
         query: MemPoolSyncData,
         max_txs: u64,
+        page_id: Option<Txid>,
     ) -> Result<Option<StreamCursor>, net_error> {
```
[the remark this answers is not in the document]
jcnelson: Nope; will change it to `Result<StreamCursor, net_error>`

src/util/db.rs
```diff
@@ -264,13 +264,58 @@ macro_rules! impl_byte_array_from_column {
     };
 }
 
+/// Load the path of the database from the connection
+fn get_db_path(conn: &Connection) -> Result<String, Error> {
```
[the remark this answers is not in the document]
jcnelson: Fixed

testnet/stacks-node/src/tests/neon_integrations.rs
jcnelson: Ha, yes, forgot. Will add

testnet/stacks-node/src/config.rs
jcnelson: It controls how long the miner will spend iterating through transactions when building a microblock (soft deadline).  It's analogous to `first_attempt_time_ms` and `subsequent_attempt_time_ms` in the `[miner]` section.

jcnelson: Also, it should be in `config.miner`.  Will move.

src/net/http.rs
jcnelson: Sure, I can try adding that.  I don't *think* anything will break, but we'll see.

src/net/relay.rs
```diff
@@ -619,13 +619,14 @@ impl Relayer {
...
-    /// Returns the set of consensus hashes for the sortitions that selected these blocks
+    /// Returns the set of consensus hashes for the sortitions that selected these blocks, and the
+    /// blocks themselves
     fn preprocess_downloaded_blocks(
         sort_ic: &SortitionDBConn,
         network_result: &mut NetworkResult,
         chainstate: &mut StacksChainState,
-    ) -> HashSet<ConsensusHash> {
-        let mut new_blocks = HashSet::new();
+    ) -> HashMap<ConsensusHash, StacksBlock> {
```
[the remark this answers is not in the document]
jcnelson: This can only be up to `ConnectionOptions::max_inflight_blocks` entries, which is currently 6.

src/net/relay.rs
jcnelson: Can I keep both?  I have scripts that expect `${consensus_hash}/${block_hash}`, since it's a pattern we use pervasively in the code.

testnet/stacks-node/src/neon_node.rs
jcnelson: `last_sortition_mutex` can be accessed by the relayer thread, whereas `last_burn_block` cannot (it's only available to the main thread).

I think I can just drop `last_burn_block` altogether, since it gets set immediately before `last_sortition_mutex` in practice (the former is done by `.process_burnchain_state()`, whereas the latter is done by `.relayer_sortition_notify()`; both are called one right after the other in the runloop thread).

src/net/relay.rs
kantai: Yeah, that works.

testnet/stacks-node/src/neon_node.rs
kantai: Yep, dropping `last_burn_block` makes sense

testnet/stacks-node/src/neon_node.rs
jcnelson: Yeah, probably a good idea. Will fix.

testnet/stacks-node/src/run_loop/neon.rs
jcnelson: This change makes sure that the burnchain indexer only downloads _one_ burnchain block per call.  The `.sync()` method takes an optional maximum burnchain block height, at which `.sync()` will return if it is reached.  Before, the node would download a whole reward cycle of burnchain blocks, and then run sortitions on all of them, which prevented the node from downloading Stacks blocks while burnchain blocks were being downloaded.  With this new loop, a node will download a burnchain block, process its sortition (waking up the chains coordinator), and then proceed to download the next burnchain block in the next loop pass while the p2p and relayer threads work on downloading the sortition's Stacks block.

testnet/stacks-node/src/run_loop/neon.rs
jcnelson: No, it's waiting for Stacks blocks.  The PoX sync watchdog observes the chainstate database and p2p network behavior and deduces when the node has downloaded all of the Stacks blocks it is going to download for the foreseeable future.  Namely, if there are some sortitions with Stacks blocks that are simply missing (and these *do exist*), the PoX sync watchdog needs to deduce that they are actually absent and never going to realistically arrive, so the node can go ahead and get the next reward cycle processed (with or without a potentially-missing anchor block).  The comment refers to the call to `pox_sync_wait()` below.

jcnelson: (Btw, this whole strategy for determining when to advance from one reward cycle to another is going away once #1805 lands, since then we'll be able to infer whether or not to wait on a reward cycle's blocks based on whether or not we have the next reward cycle's anchor block on the canonical anchor block history.)

jcnelson: I'll update the comment to this:

```rust
debug!("Wait until Stacks block downloads reach a quiescent state before processing more burnchain blocks"; "remote_chain_height" => remote_chain_height, "local_chain_height" => burnchain_height);
```

testnet/stacks-node/src/neon_node.rs
jcnelson: Fixed

testnet/stacks-node/src/neon_node.rs
jcnelson: Yup; fixed

src/net/rpc.rs
pavitthrap: I think you need to push?

jcnelson:
Okay, addressed @kantai's feedback.

jcnelson:
Okay, codecov is misbehaving, which is causing everything to break.

kantai:
> Okay, codecov is misbehaving, which is causing everything to break.

Hmm-- I'd guess possibly due to the Github org name change

src/core/mempool.rs
```diff
@@ -1856,13 +1905,35 @@ impl MemPoolDB {
...
                     test_debug!(
-                        "No more txs in query after {:?}",
+                        "Cork tx stream with next page {}",
                         &query.last_randomized_txid
                     );
-                    break;
+                    query
+                        .last_randomized_txid
+                        .consensus_serialize(&mut query.tx_buf)
+                        .map_err(ChainstateError::CodecError)?;
```
kantai, line 1926: According to code coverage, this branch isn't covered -- is it possible to add coverage for this to the unit tests?

src/net/http.rs
```diff
@@ -3747,36 +3775,135 @@ impl HttpResponseType {
...
+                            return Err(e.into());
+                        }
+                    }
+                    _ => {
+                        return Err(e.into());
+                    }
+                },
+                e => {
+                    return Err(e.into());
+                }
```
kantai, line 3806: None of these error branches appear to be covered-- the condition `pos != retry_reader.position()` seems like it could be important to check, can you add unit test coverage for that?

src/net/http.rs
```diff
@@ -3747,36 +3775,135 @@ impl HttpResponseType {
...
+        };
+
+        test_debug!("Read page_id {:?}", &next_page);
+        Ok(Some(next_page))
+    }
+
+    /// Decode a transaction stream, returned from /v2/mempool/query.
+    /// The wire format is a list of transactions (no SIP-003 length prefix), followed by an
+    /// optional 32-byte page ID.  Obtain both the transactions and page ID, if it exists.
+    pub fn decode_tx_stream<R: Read>(
```
kantai, line 3817: In this method, it looks like none of the `ExpectedEndOfStream` error cases have coverage -- can these be covered by unit tests?

src/net/p2p.rs
```diff
@@ -2165,8 +2178,19 @@ impl PeerNetwork {
...
+            (false, txs_opt) => {
+                // did we get some transactions, but have more to get?
+                if let Some(txs) = txs_opt {
+                    debug!(
+                        "{:?}: Mempool sync obtained {} transactions from mempool sync, but have more",
+                        &self.local_peer,
+                        txs.len()
+                    );
+
+                    return Ok(Some(txs));
```
kantai, line 2190: The `if let Some(...)` branch here appears to not be covered

src/net/p2p.rs
```diff
@@ -3381,7 +3425,10 @@ impl PeerNetwork {
                 get_epoch_time_secs() + self.connection_opts.mempool_sync_timeout;
         } else {
             if get_epoch_time_secs() > self.mempool_sync_timeout {
```
kantai, line 3427: What's the different between `self.mempool_sync_timeout` and `self.mempool_sync_deadline`?

src/net/p2p.rs
```diff
@@ -3446,23 +3509,45 @@ impl PeerNetwork {
...
+                            let ret = match next_page_id_opt {
+                                Some(next_page_id) => {
+                                    // get the next page
+                                    self.mempool_state = MempoolSyncState::SendQuery(
+                                        url.clone(),
+                                        addr.clone(),
+                                        next_page_id,
+                                    );
+                                    false
+                                }
```
kantai, line 3532: This branch appears to be uncovered as well.

MaksimalistT:
Hi, @jcnelson, @kantai 
my node is struggling with **high frequency of invalid blocks** and would be happy to **increase number of txs mined per block** to help the network and earn more in fees
**willing to upgrade my node before official release**
which branches are addressing this issues and safe enough to use?

src/net/p2p.rs
jcnelson: `self.connection_opts.mempool_sync_timeout` is the maximum amount of time a mempool sync is allowed to take before the peer gives up.  `self.mempool_sync_deadline` is the wall-clock time when the current mempool sync must terminate because it's taking too long.

kantai: But `self.mempool_sync_timeout` is checked directly against the epoch time here -- so it must also be a deadline, right?

jcnelson: Ah, I misread what you were asking.  `self.mempool_sync_timeout` is the absolute time at which the mempool sync will be terminated if it takes too long.  `self.mempool_sync_deadline` is the time at which the _next_ mempool sync will start.  So, once a mempool sync terminates, at least `self.mempool_sync_deadline` seconds will pass until the next mempool sync starts up.

src/net/http.rs
jcnelson: Yup, can do

src/net/http.rs
jcnelson: Yup; it's working locally

src/net/p2p.rs
jcnelson: I don't think that code is even reachable.  I will just remove it instead.

jcnelson: Hmm, it's definitely getting exercised.  I wonder why codecov didn't report it.

src/net/p2p.rs
jcnelson: That's definitely getting exercised in `test_mempool_sync_2_peers_paginated`.

src/net/p2p.rs
kantai: That makes sense, thanks -- can you add some rustdocs to those fields in the `PeerNetwork` struct that convey that?

src/net/p2p.rs
kantai: Yes -- that'd be why -- I think the `#[ignore]` tests aren't getting executed by the code coverage unit tests.

kantai (approved):
This LGTM, thanks @jcnelson!

pavitthrap (approved):
lgtm - just some minor comments

testnet/stacks-node/src/neon_node.rs
```diff
@@ -1055,6 +1091,7 @@ fn spawn_miner_relayer(
                         burn_tenure_snapshot = burn_chain_sn;
                         if issue_timestamp_ms + (config.node.wait_time_for_microblocks as u128) < get_epoch_time_ms() {
                             // still waiting for microblocks to arrive
+                            debug!("Relayer: will NOT run tenure since still waiting for microblocks to arrive ({} <= {})", (issue_timestamp_ms + (config.node.wait_time_for_microblocks as u128)) / 1000, get_epoch_time_secs());
```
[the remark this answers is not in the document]
jcnelson: The debug statement is comparing the time to the current epoch time in _seconds_, whereas the `if`-condition is comparing it to the current epoch time in _milliseconds_.

testnet/stacks-node/src/neon_node.rs
```diff
@@ -675,20 +670,18 @@ fn spawn_peer(
             )
             .expect("Database failure opening mempool");
 
-            // create estimators, metric instances for RPC handler
             let cost_estimator = config
                 .make_cost_estimator()
                 .unwrap_or_else(|| Box::new(UnitEstimator));
             let metric = config
                 .make_cost_metric()
                 .unwrap_or_else(|| Box::new(UnitMetric));
```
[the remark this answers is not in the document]
jcnelson: The first `cost_estimator` and `metric` are owned by the mempool, whereas the second instances are owned by the `RPCHandlerArgs`.

src/net/p2p.rs
```diff
@@ -2833,7 +2862,8 @@ impl PeerNetwork {
             reward_cycle_finish
         );
 
-        for reward_cycle in reward_cycle_start..reward_cycle_finish {
+        // go from latest to earliest reward cycle
+        for reward_cycle in (reward_cycle_finish..reward_cycle_start).rev() {
```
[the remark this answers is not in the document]
jcnelson: Ah, but `reward_cycle_finish <= reward_cycle_start`.  The anti-entropy protocol works backwards, from latest reward cycle to earliest reward cycle.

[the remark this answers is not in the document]
pavitthrap: if `reward_cycle_finish <= reward_cycle_start`, wouldn't that make `reward_cycle_start` the latest reward cycle?

src/util/db.rs
```diff
@@ -264,13 +264,58 @@ macro_rules! impl_byte_array_from_column {
...
+    }
+
+    let mut parts = sql_query.clone().split(" ");
+    let mut full_sql = if let Some(part) = parts.next() {
+        part.to_string()
+    } else {
+        sql_query.to_string()
+    };
+
+    while let Some(part) = parts.next() {
```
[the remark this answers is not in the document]
jcnelson: I'm not sure what you're asking?  The code is constructing a new string because trying to replace `?X` with `"mock_arg"` would in effect be doing the same thing.

src/net/p2p.rs
```diff
@@ -5856,4 +5957,197 @@ mod test {
...
+            }
+
+            info!("Completed mempool sync in {} step(s)", round);
+
+            let mp = peer_2.mempool.take().unwrap();
+            let peer_2_mempool_txs = MemPoolDB::get_all_txs(mp.conn()).unwrap();
+            peer_2.mempool.replace(mp);
+
+            // peer 2 has all the recent txs
+            // peer 2 has none of the old ones
```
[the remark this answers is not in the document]
jcnelson: Yeah, it's copypasta'ed from the above test, which does check for this. Will remove.

src/net/relay.rs
```diff
@@ -619,13 +619,14 @@ impl Relayer {
     /// Preprocess all our downloaded blocks.
     /// Return burn block hashes for the blocks that we got.
```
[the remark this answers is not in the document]
jcnelson: Indeed it is.

src/net/p2p.rs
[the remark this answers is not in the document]
jcnelson: Yes, it would.  Here's an execution trace of some of the debug statements from a unit test:

```
$ grep -E 'Local blocks inventory|over reward cycles' /tmp/test.out
DEBG [1643228629.268898] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 6-5
DEBG [1643228629.273409] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 5 is BlocksInvData { bitlen: 5, block_bitvec: [31], microblocks_bitvec: [30] }
DEBG [1643228630.151754] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 6-5
DEBG [1643228630.155805] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 5 is BlocksInvData { bitlen: 5, block_bitvec: [3], microblocks_bitvec: [2] }
DEBG [1643228631.350438] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 5-4
DEBG [1643228631.353977] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 4 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228632.425082] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 5-4
DEBG [1643228632.428814] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 4 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228633.348980] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 4-3
DEBG [1643228633.352538] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 3 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228634.376591] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 4-3
DEBG [1643228634.380029] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 3 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228635.396177] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 3-2
DEBG [1643228635.399907] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 2 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228636.350704] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 3-2
DEBG [1643228636.354618] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 2 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228637.306919] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 2-1
DEBG [1643228637.310311] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 1 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228638.418983] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 2-1
DEBG [1643228638.423283] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 1 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228639.353688] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 1-0
DEBG [1643228639.357670] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 0 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228640.394823] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 1-0
DEBG [1643228640.399527] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 0 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228641.443450] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 6-5
DEBG [1643228641.452464] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 5 is BlocksInvData { bitlen: 5, block_bitvec: [31], microblocks_bitvec: [30] }
DEBG [1643228642.554711] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 6-5
DEBG [1643228642.560304] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 5 is BlocksInvData { bitlen: 5, block_bitvec: [15], microblocks_bitvec: [14] }
DEBG [1643228643.357928] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 5-4
DEBG [1643228643.363271] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 4 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228644.372366] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 5-4
DEBG [1643228644.376326] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 4 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228645.404950] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 4-3
DEBG [1643228645.408651] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 3 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228646.427323] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 4-3
DEBG [1643228646.432209] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 3 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228647.400470] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 3-2
DEBG [1643228647.405313] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 2 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228648.366515] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 3-2
DEBG [1643228648.371189] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 2 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228649.301985] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 2-1
DEBG [1643228649.306562] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 1 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228650.444922] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 2-1
DEBG [1643228650.449579] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 1 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228651.346471] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 1-0
DEBG [1643228651.352485] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 0 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228652.412420] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 1-0
DEBG [1643228652.417380] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4242)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 0 is BlocksInvData { bitlen: 5, block_bitvec: [0], microblocks_bitvec: [0] }
DEBG [1643228653.391679] [src/net/p2p.rs:2857] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: run protocol for 1 neighbors, over reward cycles 6-5
DEBG [1643228653.400161] [src/net/p2p.rs:2879] [ThreadId(2)] local.80000000://(bind=127.0.0.1:4240)(pub=None): AntiEntropy: Local blocks inventory for reward cycle 5 is BlocksInvData { bitlen: 5, block_bitvec: [31], microblocks_bitvec: [30] }
```

As you can see, the system starts from the highest reward cycle, looks for data to push, and on the next pass, starts from the next-highest reward cycle, and so on and so forth until it wraps around.  The idea is that other nodes are most likely to be missing _recent_ blocks, so the antientropy system should prioritize them first.

src/net/relay.rs
```diff
@@ -1403,48 +1451,203 @@ impl PeerNetwork {
...
+                        }
                     }
                 }
             }
             Ok(wanted)
         })?;
 
-        if microblocks {
-            self.advertize_to_peer(recipient, &wanted, |payload| {
+        let mut advertize = vec![];
```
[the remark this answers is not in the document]
jcnelson: Yup, fixed

Issue #2990, opened 2022-01-10
Unexpected Panic due to empty http response from github


wileyj:
**Describe the bug**
`Process abort due to thread panic: panicked at 'Empty response', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/async-h1-1.0.2/src/client.rs:146:9`

**Steps To Reproduce**
intermittent, and dependent on the external service provider

**Expected behavior**
No panic, perhaps an indefinite retry that is logged as to what the issue might be?

**Environment (please complete the following information):**
 - alpine linux
 - stacks-node 2.05.0.0.0 (master:a783b27, release build, linux [x86_64])
 - binaries from official dockerhub image
 - 
**Additional context**
```

Jan 10 09:04:01 AM  INFO [1641834241.102900] [src/net/relay.rs:972] [relayer] Processing newly received Stacks blocks: 6, microblocks: 0, confirmed microblocks: 4
Jan 10 09:04:01 AM  Process abort due to thread panic: panicked at 'Empty response', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/async-h1-1.0.2/src/client.rs:146:9
Jan 10 09:04:02 AM     0: stacks_node::main::{{closure}}
Jan 10 09:04:02 AM               at /src/testnet/stacks-node/src/main.rs:58
Jan 10 09:04:02 AM     1: std::panicking::rust_panic_with_hook
Jan 10 09:04:02 AM               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:628
Jan 10 09:04:02 AM     2: std::panicking::begin_panic::{{closure}}
Jan 10 09:04:02 AM               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:544
Jan 10 09:04:02 AM     3: std::sys_common::backtrace::__rust_end_short_backtrace
Jan 10 09:04:02 AM               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/sys_common/backtrace.rs:141
Jan 10 09:04:02 AM     4: std::panicking::begin_panic
Jan 10 09:04:02 AM               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:543
Jan 10 09:04:02 AM     5: async_h1::client::decode::{{closure}}
[...]

PR #2995, opened 2022-01-12
Chore: Update async-h1 and related dependencies

kantai:
It looks like #2990 was caused by a bug in an dependency (used by the event observer, so it's only being invoked on configured listeners).

This PR updates the relevant dependencies, which required a few lines of code changes as well.

### DIFF
--- a/testnet/puppet-chain/src/main.rs
+++ b/testnet/puppet-chain/src/main.rs
@@ -147,7 +147,7 @@ async fn main() -> http_types::Result<()> {
         if should_ignore_txs {
             // Returns ok
             println!("Buffering request from {}", stream.peer_addr()?);
-            async_h1::accept(&addr, stream.clone(), |_| async {
+            async_h1::accept(stream.clone(), |_| async {
                 Ok(Response::new(StatusCode::Ok))
             })
             .await?;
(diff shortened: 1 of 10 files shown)


testnet/puppet-chain/src/main.rs
```diff
@@ -181,8 +181,8 @@ async fn main() -> http_types::Result<()> {
 }
 
 // Take a TCP stream, and convert it into sequential HTTP request / response pairs.
-async fn accept(addr: String, stream: TcpStream, config: &ConfigFile) -> http_types::Result<()> {
-    async_h1::accept(&addr, stream.clone(), |mut req| async {
+async fn accept(_addr: String, stream: TcpStream, config: &ConfigFile) -> http_types::Result<()> {
```
pavitthrap, line 184: Could you refactor to remove this argument altogether?

pavitthrap (dismissed):
lgtm - just one nit

kantai:
Hmmm -- it looks like updating the dependency is not as simple as I initially thought... will have to look into this a bit more.

jcnelson:
Is this something that we're going to try and get into 2.05.0.1.0?

kantai:
> Is this something that we're going to try and get into 2.05.0.1.0?

That depends on the priority of fixing #2990 I suppose.

This is also something that could go in a patch release.

gregorycoppola (approved):

kantai:
Looks like all it needed was an update of the tokio version (only used in the integration tests for the integration event observer). Merging!

Issue #2970, opened 2021-12-12
stacks-node crashed


MaksimalistT:
**Describe the bug**
stacks-node process crashed
works after restart

**Steps To Reproduce**
no info

**Environment (please complete the following information):**
- OS: Amazon Linux
- Rust version: rustc 1.49.0 (e1884a8e3 2020-12-29)
- Stacks-node 2.05.0.0.0

**Additional context**
[debug_with_crash_tail.log](https://github.com/blockstack/stacks-blockchain/files/7698481/debug_with_crash_tail.log)

jcnelson:
Thanks for reporting!  The full stack trace was:

```
Process abort due to thread panic: panicked at 'ERROR: Failed to begin new unconfirmed MARF block for dd756178192819bb98b1948e976f5f89e92c98a8ee5f4cfcdb870aa9bfe65819): InProgressError', src/clarity_vm/database/marf.rs:199:39
   0: stacks_node::main::{{closure}}
             at /src/testnet/stacks-node/src/main.rs:58
   1: std::panicking::rust_panic_with_hook
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:628
   2: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:521
   3: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/sys_common/backtrace.rs:141
   4: rust_begin_unwind
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:517
   5: core::panicking::panic_fmt
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:101
   6: core::result::unwrap_failed
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/result.rs:1617
   7: core::result::Result<T,E>::expect
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/result.rs:1259
      blockstack_lib::clarity_vm::database::marf::MarfedKV::begin_unconfirmed
             at /src/src/clarity_vm/database/marf.rs:199
[...]

PR #2996, opened 2022-01-12
Changes for version 2.05.0.1.0

pavitthrap:
This PR merges changes from develop into master for version 2.05.0.1.0 of the stacks node.


CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
+## [Unreleased]
+
+### Changed
+
+- The pessimistic execution cost estimator differentiates between
+  contracts with different origin addresses.
```
jcnelson, line 13: Needs to be fleshed out with the fee estimation work and mempool sync

pavitthrap: Yup made a separate PR for changelog: https://github.com/blockstack/stacks-blockchain/pull/2997

kantai:
Tagged [2.05.0.1.0-rc2](https://github.com/blockstack/stacks-blockchain/releases/tag/2.05.0.1.0-rc2)

jcnelson:
Well, I'm definitely seeing some slowness in /v2/mempool/query.  Not sure where yet, but some queries can make the p2p thread stall out for a few seconds.  I'm guessing there's some slowness in the walk logic; will investigate further.

jcnelson:
Okay, we need to block the release here until /v2/mempool/query can be fixed.  I think I know what's going on -- we're filling the socket buffer for a mempool request, but not limiting the amount of mempool DB I/O that happens to do so.  I can make it so that the node will stream back data such that it does not attempt to fill the socket buffer (i.e. it'll just send back one transaction at a time), but for some reason this is causing the *client* to time out.  Will look more into this tomorrow.

In the mean time, I'm happy to report that https://github.com/blockstack/stacks-blockchain/pull/2989 is working quite well.  The mock-miner is correctly dropping in-flight `RunTenure` requests for stale chain tips; the mock-miner only builds on blocks in the canonical chain.  In addition, all the new indexes in this PR significantly speed up the node boot-up (like, it cuts the time in half), as well as node steady-state behavior.  I'd like to get this PR merged into this release if possible; it will contain the fix to /v2/mempool/query.

jcnelson:
Fix for the mempool query logic: https://github.com/blockstack/stacks-blockchain/pull/2989

jcnelson:
Should also probably block on #3007 (as well as #2970, which appears to be the same issue).  We've now seen it on the Foundation miner.

kantai:
> Should also probably block on #3007 (as well as #2970, which appears to be the same issue). We've now seen it on the Foundation miner.

Why block on that issue? That issue appears to have existed before this RC, so I don't think this is a regression from this RC. I think we should assess how easy that issue will be to fix before deciding to block the RC.

kantai:
I added a release checklist to the PR description to help track the progress of the release.

jcnelson (approved):
LGTM!

PR #2998, opened 2022-01-12
fix: Create a separate directory for `estimates` tables

gregorycoppola:
fixes https://github.com/blockstack/stacks-blockchain/issues/2991

1) Create a directory `chainstate/estimates`
2) Ensure this exists
3) Use this for the new estimation db's

### DIFF
--- a/testnet/stacks-node/src/config.rs
+++ b/testnet/stacks-node/src/config.rs
@@ -776,6 +776,13 @@ impl Config {
         path
     }
 
+    pub fn get_estimates_path(&self) -> PathBuf {
+        let mut path = self.get_chainstate_path();
+        path.push("estimates");
+        assert!(fs::create_dir_all(&path).is_ok());
+        path
+    }
+
     pub fn get_chainstate_path_str(&self) -> String {
         self.get_chainstate_path()
             .to_str()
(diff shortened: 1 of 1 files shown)


kantai:
I think this PR is going to need to add a `fs::create_dir_all` call to create the new directory (https://doc.rust-lang.org/std/fs/fn.create_dir_all.html)

testnet/stacks-node/src/config.rs
```diff
@@ -776,6 +776,14 @@ impl Config {
         path
     }
 
+    /// Returns the path `{get_chainstate_path()}/estimates`, and ensures it exists.
+    pub fn get_estimates_path(&self) -> PathBuf {
+        let mut path = self.get_chainstate_path();
+        path.push("estimates");
+        assert!(fs::create_dir_all(&path).is_ok());
```
gregorycoppola, line 783: @kantai note the call to `create_dir_all`

What do you think about this location for it?

kantai: Ah, oops, sorry I missed this -- yeah, this works.

kantai: Though, I would suggest changing from the assert to an expect:

```suggestion
        fs::create_dir_all(&path).expect("Failed to create cost estimates path in node chainstate directory.");
```

gregorycoppola: updated

pavitthrap (approved):
lgtm!

kantai (approved):
LGTM!

gregorycoppola:
Thanks for the reviews.

I added Issue https://github.com/blockstack/stacks-blockchain/issues/3000 for the broken test because I think it's unrelated and flaky.

PR #2999, opened 2022-01-12
CI: add cancel in progress behavior to PR actions

kantai:
This adds cancel-in-progress behavior to our PR Github actions, where if you push new commits to a PR, it cancels any in progress actions associated with that PR.


pavitthrap (approved):
lgtm!

gregorycoppola (approved):
Not sure if this is correct, but limited downside to try.

I would have thought Charlie would be a better reviewer than me, but to save Charlie-bandwidth can approve.

PR #3003, opened 2022-01-14
Feat/add stacks height http header

pavitthrap:
This PR adds information about a node's canonical stacks tip height to its http headers. This way, nodes will have visibility of their peers' heights. 
This work will be used in #2768, to implement a health check endpoint. Part of the notion of health of a node is being caught up to their peers.  

## Testing 
Added test in `rpc.rs`

### DIFF
--- a/src/net/atlas/download.rs
+++ b/src/net/atlas/download.rs
@@ -361,8 +361,11 @@ impl AttachmentsBatchStateContext {
                         reliability_report: reliability_report.clone(),
                         contract_id: contract_id.clone(),
                         pages: pages.clone(),
-                        block_height: self.attachments_batch.block_height,
+                        stacks_block_height: self.attachments_batch.stacks_block_height,
                         index_block_hash: self.attachments_batch.index_block_hash,
+                        canonical_stacks_tip_height: self
+                            .attachments_batch
+                            .canonical_stacks_tip_height,
                     };
                     queue.push(request);
                 }
@@ -446,7 +449,8 @@ impl AttachmentsBatchStateContext {
                 let request = AttachmentRequest {
                     sources,
                     content_hash: content_hash.clone(),
-                    block_height: self.attachments_batch.block_height,
+                    stacks_block_height: self.attachments_batch.stacks_block_height,
+                    canonical_stacks_tip_height: self.attachments_batch.canonical_stacks_tip_height,
                 };
                 enqueued.insert(content_hash);
                 queue.push(request);
@@ -955,9 +959,10 @@ pub struct AttachmentsInventoryRequest {
     pub url: UrlString,
     pub contract_id: QualifiedContractIdentifier,
     pub pages: Vec<u32>,
-    pub block_height: u64, // block height on Stacks chain
+    pub stacks_block_height: u64,
     pub index_block_hash: StacksBlockId,
     pub reliability_report: ReliabilityReport,
+    pub canonical_stacks_tip_height: Option<u64>,
 }
 
 impl Hash for AttachmentsInventoryRequest {
@@ -965,7 +970,7 @@ impl Hash for AttachmentsInventoryRequest {
         self.contract_id.hash(state);
         self.pages.hash(state);
         self.index_block_hash.hash(state);
-        self.block_height.hash(state);
+        self.stacks_block_height.hash(state);
     }
 }
 
@@ -1002,7 +1007,7 @@ impl Requestable for AttachmentsInventoryRequest {
             pages_indexes.insert(*page);
         }
         HttpRequestType::GetAttachmentsInv(
-            HttpRequestMetadata::from_host(peer_host, Some(self.block_height)),
+            HttpRequestMetadata::from_host(peer_host, self.canonical_stacks_tip_height),
             self.index_block_hash,
             pages_indexes,
         )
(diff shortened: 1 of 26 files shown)


src/net/atlas/download.rs
```diff
@@ -446,6 +446,7 @@ impl AttachmentsBatchStateContext {
                 let request = AttachmentRequest {
                     sources,
                     content_hash: content_hash.clone(),
+                    block_height: self.attachments_batch.block_height,
```
jcnelson, line 449: Is `self.attachments_batch.block_height` is the current Stacks block height, is it?

src/net/download.rs
```diff
@@ -157,12 +157,12 @@ impl Requestable for BlockRequestKey {
     fn make_request_type(&self, peer_host: PeerHost) -> HttpRequestType {
         match self.kind {
             BlockRequestKeyKind::Block => HttpRequestType::GetBlock(
-                HttpRequestMetadata::from_host(peer_host),
+                HttpRequestMetadata::from_host(peer_host, Some(self.sortition_height)),
```
jcnelson, line 160: `self.sortition_height` is not the Stacks block height; it's the Bitcoin block height, minus the start Bitcoin block height.

src/net/http.rs
```diff
@@ -883,6 +906,13 @@ fn keep_alive_headers<W: Write>(fd: &mut W, md: &HttpResponseMetadata) -> Result
             }
         }
     }
+    match md.canonical_stacks_tip_height {
+        Some(height) => {
+            fd.write_all(format!("Canonical-Stacks-Tip-Height: {}\r\n", height).as_bytes())
```
jcnelson, line 911: This isn't a standard HTTP header, so it should start with `X-`.  Also, do we need to propagate the canonical tip's index block hash as well?  Multiple Stacks blocks can have the same height.

src/net/p2p.rs
```diff
@@ -3294,12 +3295,17 @@ impl PeerNetwork {
         &mut self,
         url: &UrlString,
         addr: &SocketAddr,
+        sortdb: &SortitionDB,
         mempool: &MemPoolDB,
         chainstate: &mut StacksChainState,
     ) -> Result<(bool, Option<usize>), net_error> {
+        let burn_tip = SortitionDB::get_canonical_burn_chain_tip(sortdb.conn())?;
```
jcnelson, line 3302: Instead of pulling through the `SortitionDB` and doing additional database I/O on this hot path, can you instead use the cached chain tip in `self.burnchain_tip`?

src/net/rpc.rs
```diff
@@ -2113,6 +2174,7 @@ impl ConversationHttp {
         handler_opts: &RPCHandlerArgs,
     ) -> Result<Option<StacksMessageType>, net_error> {
         let mut reply = self.connection.make_relay_handle(self.conn_id)?;
+        let burn_tip = SortitionDB::get_canonical_burn_chain_tip(sortdb.conn())?;
```
jcnelson, line 2177: Instead of doing database I/O on this hot path, can you just instead use `network.burnchain_tip` for the burnchain tip?  Thanks!

src/net/rpc.rs
```diff
@@ -2982,27 +3096,30 @@ impl ConversationHttp {
     }
 
     /// Make a new getinfo request to this endpoint
-    pub fn new_getinfo(&self) -> HttpRequestType {
-        HttpRequestType::GetInfo(HttpRequestMetadata::from_host(self.peer_host.clone()))
+    pub fn new_getinfo(&self, stacks_height: Option<u64>) -> HttpRequestType {
```
jcnelson, line 3099: Why does `getinfo` need a stacks height?  Doesn't the message body _already_ return this information?

jcnelson: Also, `/v2/info` isn't used by any of the p2p networking protocols, so it doesn't really help peers here.

gregorycoppola (commented):
I don't have too much to add right now beyond what @jcnelson said.

src/net/download.rs
pavitthrap: ohh I see.

src/net/http.rs
pavitthrap: The health check endpoint will be checking that the node's height is in the neighborhood of its peers' heights, so I don't think that info will be necessary.

src/net/rpc.rs
pavitthrap: I added this info just for the test - I do a check in the test for get info that checks that the client and server's ConversationHttp objects get updated appropriately. 

I can test in a different and separate function if this seems confusing.

src/net/atlas/download.rs
```diff
@@ -1059,7 +1061,10 @@ impl Requestable for AttachmentRequest {
     }
 
     fn make_request_type(&self, peer_host: PeerHost) -> HttpRequestType {
-        HttpRequestType::GetAttachment(HttpRequestMetadata::from_host(peer_host), self.content_hash)
+        HttpRequestType::GetAttachment(
+            HttpRequestMetadata::from_host(peer_host, Some(self.block_height)),
```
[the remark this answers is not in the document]
pavitthrap: I'll have to review this - I made an incorrect assumption here.

src/net/atlas/download.rs
pavitthrap: It looks like it. The value here seems to originate from the block_height field in `StacksHeaderInfo`

src/net/rpc.rs
jcnelson: Ah, makes sense

jcnelson (approved):
LGTM.  Thanks @pavitthrap!

kantai (approved):
LGTM!

gregorycoppola (commented):
Already approved.

PR #3005, opened 2022-01-16
fix: significantly increase the speed of most RPC endpoints

jcnelson:
This PR improves querying the canonical chain tip by a couple orders of magnitude, using this One Weird Trick (tm).

It improves network responsiveness considerably, and probably also helps chain boot-up (but I need to test this to get an actual number).

It's so much faster on the Foundation's public node that RPC queries that used to take ~200 ms now take 0 ms.  I don't even know what the true multiplier is.  I'm dumbfounded that we didn't find this sooner.

Please consider adding to 2.05.0.1.0.


jcnelson:
The only failed check was `codecov`, since I guess it doesn't know how to deal with a 1-line change to a global variable?

gregorycoppola (approved):
nice find!

kantai (approved):
LGTM!

Regarding the failing codecov check, it looks like the codecov checker just kind of bugged out -- the codecov report here (https://app.codecov.io/gh/blockstack/stacks-blockchain/branch/fix%2Fpox-valid-index) shows that the coverage is perfectly fine, but the PR check and robot comment look really weird. This kind of PR should normally work fine, I just think there's some kind of bug in codecov's GH robot.

jcnelson:
Going to go ahead and merge.  Will follow up in #2989 with code that instantiates this index on an already-sync'ed node.

PR #3010, opened 2022-01-21
CI: Replace codecov with coveralls

kantai:
Codecov has been really flaky recently, and never terribly reliable in any event. This replaces that with coveralls so we can see how well that might work.


kantai:
Okay -- so this PR _doesn't_ fix the current coverage issue where we are getting 14% coverage from the lcov reports (which isn't accurate): that seems like it may be a regression in the current rust nightly, but I'll look more into it.

This PR _does_, however, fix the issue where all of our tests fail because of a failure to upload to codecov. I think that issue was caused by the `blockstack` -> `stacks-network` Github organization rename. It may be possible to fix that in codecov by reauthing with codecov's app, but it requires some really pretty invasive permissions (repo write, in particular), that coveralls _doesn't_. So, I'd rather we just switched to coveralls. Thoughts, @jcnelson and @CharlieC3?

kantai:
Actuailly, it looks like just having someone log into codecovs interfaces may have fixed the codecov integration. We can postpone talking about this PR for now.

PR #3016, opened 2022-01-26
Fix/2970

### LINKED ISSUE #2970: stacks-node crashed

jcnelson:
This fixes #2970 by ensuring that the unconfirmed state MARF trie is *always* committed when processing unconfirmed microblocks, *even if* the microblocks are invalid.

Thanks to @wileyj, I was able to identify the root cause of #2970: the node was processing an invalid microblock and then leaving the MARF open, which would then cause the node to crash when it attempted to mine a microblock (via `InProgressError`).  The solution is to unconditionally close the unconfirmed state MARF trie, even if the unconfirmed microblock stream is bad.  I have added a unit test that fails in `develop` but succeeds in this PR to demonstrate this.

Almost as concerning is how the invalid microblock came to be.  The microblock in question was invalid because it contained a transaction with a bad nonce.  Moreover, this node mined it -- it produced a microblock that was valid at the time it was created, but later became invalid through some unknown means.  I need to figure out why this is, but for the time being, this PR addresses the crash.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ and this project adheres to the versioning scheme outlined in the [README.md](RE
   this PR adds to represent this data). (#2862)
 - Adds the /v2/data_var endpoint, which returns a contract's data variable 
   value and a MARF proof of its existence. (#2862)
+- Fixed a bug in the unconfirmed state processing logic that could lead to a
+  denial of service (node crash) for nodes that mine microblocks (#2970)
 
 ### Changed
 - Updated the mined block event. It now includes information on transaction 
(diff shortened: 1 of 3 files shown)


kantai (approved):

pavitthrap (approved):
lgtm

jcnelson:
@pavitthrap can you sign the CLA? Thanks!

pavitthrap:
@jcnelson could you add a note about this in `CHANGELOG.md`?

PR #3018, opened 2022-01-27
Fix arg count checks for map functions

obycode:
## Description

The checks in the type checker for the map functions use `check_arguments_at_least` when they should use `check_argument_count`, since passing too many arguments should be reported as an error.

1. Motivation for change
hirosystems/clarinet#228 (@friedger)

2. What was changed
I modified those checks to look for the exact argument count.

3. How does this impact application developers
If existing contracts were passing too many arguments to these functions, they will now get errors.

4. Link to relevant issues and documentation
hirosystems/clarinet#228

5. Provide examples of use cases with code samples and applicable acceptance criteria

This program should report an error for passing too many arguments to `map-delete`:
```clarity
(define-map user-profile principal principal)
(define-public (delete)
  (ok (map-delete user-profile tx-sender contract-caller)))
```

## Type of Change
- Bug fix - this is reporting an error where it was previously ignored

Technically, yes. If any existing contracts passed too many arguments to these functions, they would now report an error.

No, the documentation already describes the appropriate arguments.

## Testing information

I added cases to check for wrong numbers of arguments to these functions.

### DIFF
--- a/src/vm/analysis/type_checker/natives/maps.rs
+++ b/src/vm/analysis/type_checker/natives/maps.rs
@@ -21,8 +21,7 @@ use vm::functions::tuples;
 
 use super::check_special_tuple_cons;
 use vm::analysis::type_checker::{
-    check_arguments_at_least, no_type, CheckError, CheckErrors, TypeChecker, TypeResult,
-    TypingContext,
+    check_argument_count, no_type, CheckError, CheckErrors, TypeChecker, TypeResult, TypingContext,
 };
 
 use vm::costs::cost_functions::ClarityCostFunction;
@@ -33,7 +32,7 @@ pub fn check_special_fetch_entry(
     args: &[SymbolicExpression],
     context: &TypingContext,
 ) -> TypeResult {
-    check_arguments_at_least(2, args)?;
+    check_argument_count(2, args)?;
 
     let map_name = args[0].match_atom().ok_or(CheckErrors::BadMapName)?;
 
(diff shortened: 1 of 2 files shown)


jcnelson (changes requested):
This PR cannot be accepted as-is, because it alters the consensus rules and, if applied, would lead to a chain split.  A node with this patch applied would reject a block that other nodes without this patch would accept.

If we want to address this problem, then it needs to happen separately from the production blockchain code (e.g. a fork of `libclarity`, perhaps); it needs to instead only emit a warning that Clarinet tooling can intercept; or, it needs to wait until Stacks 2.1.  If the latter, then this change needs to be feature-gated.

obycode:
Thanks for the explanation @jcnelson. I am fine with implementing this as only a check in the REPL -- we already have a check for passing the wrong number of arguments to user-defined functions.

I don't know how to decide if this should be a consensus-breaking change for 2.1, but would be happy to discuss. This issue was originally posted by @friedger, who may also have input.

friedger:
If tooling can handle this, then it is enough. For me, it was surprising that my code was check successfully but wouldn't run.

jcnelson:
Yeah, I'm all for making it so the tooling reports this error.  But we can't change the running VM implementation at this time (more details here: https://github.com/stacks-network/stacks-blockchain/issues/2683).

obycode:
Ok, I opened hirosystems/clarity-repl#86 to add this check in the REPL analysis for now. I don't know if you want to keep this PR around to consider for 2.1 or close it.

PR #3021, opened 2022-01-27
fix: download headers before instantiating burnchain

jcnelson:
Download headers *before* trying to find the first sortition block header.


kantai (approved):

pavitthrap (approved):
lgtm

PR #3022, opened 2022-01-28
Feat/faster inv generation

jcnelson:
This PR adds a new method `StacksChainState::get_blocks_inventory_for_reward_cycle()`, which is a faster version of `StacksChainState::get_blocks_inventory()` for the purposes of producing `BlocksInv` messages.  I noticed during testing 2.05.0.1.0 RCs that the Foundation nodes are spending up to 100ms (!!) generating `BlocksInv` messages; this PR should rectify this by significantly decreasing the amount of disk I/O needed to produce them.

EDIT: This PR improves block inventory generation by a factor of 10.  In live testing, the new `StacksChainState::get_blocks_inventory_for_reward_cycle()` takes around 7 milliseconds to run on the same machine.  I have booted a node from genesis to verify that the new `BlocksInv` messages are well-formed, and that they don't result in the node getting banned by existing peers (which would happen if the node announced blocks and microblocks that it claimed to have, but did not have).

We'll need to preserve both methods for the time being, because the antientropy protocol relies on `StacksChainState::get_blocks_inventory()`'s ability to query fine-grained block availability.


pavitthrap (approved):

src/chainstate/stacks/db/mod.rs
```diff
@@ -665,7 +665,7 @@ const CHAINSTATE_INITIAL_SCHEMA: &'static [&'static str] = &[
     CREATE TABLE staging_blocks(anchored_block_hash TEXT NOT NULL,
                                 parent_anchored_block_hash TEXT NOT NULL,
                                 consensus_hash TEXT NOT NULL,
-                                -- parent_consensus_hash is the consensus hash of the parent sortition of the sortition that chose this block
+                                -- parent_consensus_hash is the consensus hash of the parent sortition that chose this block
```
kantai, line 668: Is this comment update correct? What is the "parent sortition that chose this block"?

src/net/chat.rs
```diff
@@ -1439,10 +1439,27 @@ impl ConversationP2P {
...
+            .expect("FATAL: no reward cycle for a valid BlockSnapshot");
+        let blocks_inv_data = chainstate
+            .get_blocks_inventory_for_reward_cycle(burnchain, reward_cycle, &block_hashes)
             .map_err(|e| net_error::from(e))?;
 
+        if cfg!(test) {
+            // make *sure* the behavior stays the same
+            let original_blocks_inv_data: BlocksInvData = chainstate
+                .get_blocks_inventory(&block_hashes)
+                .map_err(|e| net_error::from(e))?;
```
kantai, line 1453: `map_err` should be unnecessary here -- the try operator (`?`) automatically includes a `map_err(|e| ReturnErrVariant::from(e)` statement.

kantai (commented):
This LGTM, just some superficial comments on the code, and a question about the DB schema comment changes.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -2212,6 +2212,141 @@ impl StacksChainState {
...
+            .query(args)
+            .map_err(|e| Error::DBError(db_error::SqliteError(e)))?;
+
+        while let Some(row) = rows.next()? {
+            let start_height_i64: i64 = row.get_unwrap(0);
+            let end_height_i64: i64 = row.get_unwrap(1);
+            return Ok((start_height_i64 as u64, end_height_i64 as u64));
+        }
+
+        return Err(Error::DBError(db_error::NotFoundError));
```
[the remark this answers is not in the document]
jcnelson: Fixed

src/chainstate/stacks/db/blocks.rs
```diff
@@ -2212,6 +2212,141 @@ impl StacksChainState {
...
+        let sql = "SELECT staging_blocks.consensus_hash, staging_blocks.processed, staging_blocks.orphaned, staging_microblocks.processed, staging_microblocks.orphaned \
+                   FROM staging_blocks LEFT JOIN staging_microblocks \
+                   ON staging_blocks.parent_microblock_hash = staging_microblocks.microblock_hash \
+                   WHERE staging_blocks.height >= ?1 AND staging_blocks.height <= ?2";
+        let args: &[&dyn ToSql] = &[&u64_to_sql(start_height)?, &u64_to_sql(end_height)?];
+
+        let mut stmt = self
+            .db()
+            .prepare(sql)
+            .map_err(|e| Error::DBError(db_error::SqliteError(e)))?;
```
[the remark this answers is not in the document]
jcnelson: Fixed

src/net/chat.rs
jcnelson: Fixed

src/net/chat.rs
```diff
@@ -1429,10 +1429,27 @@ impl ConversationP2P {
...
+
+            if original_blocks_inv_data != blocks_inv_data {
+                warn!(
```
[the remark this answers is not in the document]
jcnelson: This only gets executed in testing, though.

src/chainstate/stacks/db/mod.rs
jcnelson: It's the consensus hash of the sortition of the parent Stacks block.

jcnelson: Fixed

src/chainstate/stacks/db/blocks.rs
```diff
@@ -10056,9 +10211,306 @@ pub mod test {
             }
             for j in (i + 1)..blocks.len() {
                 assert!(block_inv_all.has_ith_block(j as u16));
-                assert!(block_inv_all.has_ith_microblock_stream(j as u16));
+                if j > i + 1 {
```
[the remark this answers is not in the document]
jcnelson: Microblock stream `j` is the microblock stream confirmed by block `j` (i.e. block `j` is a child of microblock stream `j`).  The reason it works this way is because multiple blocks can descend from the same microblock stream (possibly even from different microblocks), so if a node has block `j`, then it can look at the inventory vector of microblocks and see if `j`'s parent microblock stream is available by checking the `j`th bit.

kantai (approved):
LGTM!

PR #3023, opened 2022-01-28
feat: Prometheus Block Fullness Metrics

gregorycoppola:
## Motivation
We want to log a bunch of new metrics.

## Change
This PR logs the 5-dimensional `ExecutionCost` for just seen block in `append_block`.

This supports ingestion of this data into prometheus without needing to launch a custom server.

## Testing
No testing yet! Since this is for diagnostics, we can hook it up to the prometheus monitoring system and see if it looks right.


gregorycoppola:
I'll try reopening, maybe the CLA will be happy then.

Issue #3024, opened 2022-01-28
feat: Implement Diagnostics and Monitoring


gregorycoppola:
We want to eventually implement the following metrics:

1. Block orphan rate at that point in time. This is the percentage of Stacks blocks over the last, say, 24 bitcoin blocks, that were mined but did not get included in the main chain.
2. Flash blocks at that point in time. This is the percentage of Bitcoin blocks over the last 24 bitcoin blocks *which did not have any Stacks commitments*.
3. Confirmable mempool transaction count. This is number of transactions in the mempool that were confirmable at this point in time. For transactions that have a “too high” nonce, they should still be included as confirmable if the correct nonce transaction is also in the mempool.
4. Unconfirmable mempool transaction count.
5. Transactions considered by the mock miner.
6. Transactions successfully processed by the mock miner.
7. Average estimated cost of transactions considered by the mock miner
8. Average cost of transactions included in a block
    1. For each block
        1. How full is it on each dimension?
9. Transactions confirmed in a block or microblock
10. Median transaction confirmation delay over the last 24 bitcoin blocks: the number of blocks after a transaction is broadcasted before it is confirmed. 
    1. 90% transaction confirmation delay over the last 24 bitcoin blocks
    2. 95% transaction confirmation delay over the last 24 bitcoin blocks

PR #3025, opened 2022-01-28
feat: Prometheus Block Fullness Metrics

gregorycoppola:
## Motivation
We want to log a bunch of new metrics (#3024). This one adds "block fullness", which we have been talking about lately. 

This supports ingestion of this data into prometheus without needing to launch a custom server.

## Change
This PR logs the 5-dimensional `ExecutionCost` for the just seen block in the `append_block` function run by the followers and miners.

## Testing
No testing yet! Since this is for diagnostics, we can hook it up to the prometheus monitoring system and see if it looks right.

### DIFF
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -5009,12 +5009,10 @@ impl StacksChainState {
             None,
         )?;
 
-        let block_limit = clarity_tx
-            .block_limit()
-            .unwrap_or_else(|| {
-                 warn!("Failed to read transaction block limit");
-                 ExecutionCost::max_value()
-            });
+        let block_limit = clarity_tx.block_limit().unwrap_or_else(|| {
+            warn!("Failed to read transaction block limit");
+            ExecutionCost::max_value()
+        });
 
         let (
             scheduled_miner_reward,
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -5003,6 +5003,10 @@ impl StacksChainState {
             None,
         )?;
 
+        let block_limit = clarity_tx
+            .block_limit()
+            .unwrap_or(ExecutionCost::max_value());
+
         let (
             scheduled_miner_reward,
             block_execution_cost,
@@ -5241,7 +5245,7 @@ impl StacksChainState {
 
         chainstate_tx.log_transactions_processed(&new_tip.index_block_hash(), &tx_receipts);
 
-        set_last_execution_cost_observed(&block_execution_cost);
+        set_last_execution_cost_observed(&block_execution_cost, &block_limit);
 
         let epoch_receipt = StacksEpochReceipt {
             header: new_tip,
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -82,6 +82,7 @@ use crate::types::chainstate::{
     StacksAddress, StacksBlockHeader, StacksBlockId, StacksMicroblockHeader,
 };
 use crate::{types, util};
+use monitoring::set_last_execution_cost_observed;
 use types::chainstate::BurnchainHeaderHash;
 
 #[derive(Debug, Clone, PartialEq)]
@@ -5239,6 +5240,8 @@ impl StacksChainState {
         .expect("FATAL: failed to advance chain tip");
 
         chainstate_tx.log_transactions_processed(&new_tip.index_block_hash(), &tx_receipts);
+
+        set_last_execution_cost_observed(&block_execution_cost);
 
         let epoch_receipt = StacksEpochReceipt {
             header: new_tip,
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -5005,7 +5005,10 @@ impl StacksChainState {
 
         let block_limit = clarity_tx
             .block_limit()
-            .unwrap_or(ExecutionCost::max_value());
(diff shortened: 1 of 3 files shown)


src/monitoring/mod.rs
```diff
@@ -104,6 +105,20 @@ pub fn increment_btc_blocks_received_counter() {
     prometheus::BTC_BLOCKS_RECEIVED_COUNTER.inc();
 }
 
+#[allow(unused_variables)]
+pub fn set_last_execution_cost_observed(execution_cost: &ExecutionCost) {
+    #[cfg(feature = "monitoring_prom")]
+    prometheus::LAST_EXECUTION_READ_COUNT.set(execution_cost.read_count.try_into().unwrap());
```
gregorycoppola, line 111: We can try to make these percentages.

But, is this the right structure to get the data into prometheus?

src/monitoring/prometheus.rs
```diff
@@ -91,6 +91,31 @@ lazy_static! {
         "Total number of error logs emitted by node"
     )).unwrap();
 
+    pub static ref LAST_EXECUTION_READ_COUNT: IntGauge = register_int_gauge!(opts!(
+        "execution_cost_read_count",
```
kantai, line 95: I think these could be slightly more descriptively named for prometheus as `last_block_read_count`, etc.

src/monitoring/mod.rs
kantai: Yeah, this works.

Instead of panicking on failing to convert from the costs to `i64`, can you just saturate to `i64::max`?

kantai (commented):
This looks like it is failing to build with the prometheus features enabled.

You can test the build locally by enabling the `monitoring_prom` feature specifically, or by running:

```
$ cargo check --all-features
```

src/monitoring/mod.rs
gregorycoppola: what is the command to saturate?

gregorycoppola: what would be a good way to get the active ExecutionCost limits in this `append_block` context @kantai ?

kantai:
```rust
execution_cost.read_count.try_into().unwrap_or(i64::MAX)
```

kantai: > what would be a good way to get the active ExecutionCost limits in this append_block context @kantai ?

`append_block` creates a `clarity_tx` variable for execution of the block. That `ClarityTx` type has a `block_limit()` instance method -- it returns an `Option`, but it should _always_ be `Some`, as long as there is no transaction currently being executed.

gregorycoppola: should be as a percentage right now

src/monitoring/prometheus.rs
gregorycoppola: In the variable name or the caption part?

kantai: In the prometheus metric name, i.e.

```rust
    pub static ref LAST_EXECUTION_READ_COUNT: IntGauge = register_int_gauge!(opts!(
        "last_block_read_count",
```

CharlieC3: Can we also prepend `stacks_node_` to each metric name like we do for the other stacks-node metrics? It helps with search and discovery, especially when your Prometheus deployment could have hundreds of metrics from all kinds of services.
Screen Shot 2022-01-31 at 3 24 23 PM

gregorycoppola: Went with `stacks_node_last_block_read_count`-style prefix.

kantai (approved):

kantai:
LGTM! Thanks @gregorycoppola !

src/monitoring/mod.rs
```diff
@@ -104,6 +106,28 @@ pub fn increment_btc_blocks_received_counter() {
     prometheus::BTC_BLOCKS_RECEIVED_COUNTER.inc();
 }
 
+/// Log `execution_cost` as a ratio of `block_limit`.
+#[allow(unused_variables)]
+pub fn set_last_execution_cost_observed(
+    execution_cost: &ExecutionCost,
+    block_limit: &ExecutionCost,
+) {
+    #[cfg(feature = "monitoring_prom")]
```
jcnelson, line 115: Is there a way to just group all of these statements inside a single `#[cfg(feature = "monitoring_prom")]`?  Something like:

```
#[cfg(feature = "monitoring_prom")] {
   /* do stuff here
}
```

jcnelson (approved):
LGTM.  Just a small style nit.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5002,6 +5003,10 @@ impl StacksChainState {
             None,
         )?;
 
+        let block_limit = clarity_tx
+            .block_limit()
+            .unwrap_or(ExecutionCost::max_value());
```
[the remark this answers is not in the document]
gregorycoppola: hmm... what would it mean if `block_limit()` failed?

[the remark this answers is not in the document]
kantai: It would mean that a transaction was currently being evaluated. I don't think it's possible in this code path, but it'd be good to warn --

```suggestion
            .unwrap_or_else(|| {
                 warn!("Failed to read transaction block limit");
                 ExecutionCost::max_value()
            });
```

src/monitoring/mod.rs
gregorycoppola: did this

CharlieC3 (approved):
:shipit:

gregorycoppola:
Thanks for the reviews, everyone.

PR #3026, opened 2022-01-28
fix: Update Benchmarks

reedrosenbluth:
Fixed the benchmarks so that they compile and run. Update Criterion to `v0.3.5`. Not sure if the large contracts bench is working as expected however...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,9 +103,9 @@ dependencies = [
  "async-attributes",
  "async-task",
  "broadcaster",
- "crossbeam-channel",
- "crossbeam-deque",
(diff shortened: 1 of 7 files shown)


jcnelson (approved):
Thanks for doing this @reedrosenbluth!

kantai (changes requested):
It would be preferable if these benches could work without enabling these `#[cfg(test)]` methods -- they're `cfg(test)` because they'd be dangerous if used in production. Can `execute` just be replicated in the benchmark? And can other commit and begin methods be used in place of these?

kantai (approved):
LGTM!

PR #3029, opened 2022-01-31
Feat/add all indexes

jcnelson:
This PR makes it so all database indexes in all databases are instantiated unconditionally, in an idempotent fashion, after all migrations have been applied.  This way, if we want to add future indexes in subsequent PRs and have them take effect on existing chain state, we can do so just by adding the index-creation command to the list of indexes to create.

While testing this, I uncovered and fixed two small bugs that arise only when a node is stopped and restarted during boot-up:

* When booting up, the neighbor-walk logic should always try to walk to a bootstrap peer during IBD.  This way, the inventory state-machine will be able to resume crawling bootstrap peers right away, which causes block downloads to resume right away.  Before, it would take potentially a long time for the node to resume downloading blocks, because it would only stumble upon a bootstrap peer by chance.

* When booting up from existing chainstate, the burnchain controller would download 2100 Bitcoin blocks (1 reward cycle) unconditionally from when it left off, before starting the runloop and chains coordinator.  This PR makes it so that the node will instead just resume the main run loop and chains coordinator, so that both Bitcoin and Stacks blocks can proceed in parallel.

### DIFF
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -655,9 +655,7 @@ impl MemPoolDB {
         MemPoolDB::apply_schema_migrations(&mut tx)?;
 
         // add all indexes
-        for cmd in MEMPOOL_INDEXES {
-            tx.execute_batch(cmd).map_err(db_error::SqliteError)?;
-        }
+        MemPoolDB::add_indexes(&mut tx)?;
 
         tx.commit().map_err(db_error::SqliteError)?;
         Ok(())
@@ -700,6 +698,14 @@ impl MemPoolDB {
                     panic!("Unknown schema version {}", version);
                 }
             }
+        }
+        Ok(())
+    }
+
+    /// Add indexes
+    fn add_indexes(tx: &mut DBTx) -> Result<(), db_error> {
+        for cmd in MEMPOOL_INDEXES {
+            tx.execute_batch(cmd).map_err(db_error::SqliteError)?;
         }
         Ok(())
     }
@@ -794,6 +800,7 @@ impl MemPoolDB {
         } else {
             let mut tx = tx_begin_immediate(&mut conn)?;
             MemPoolDB::apply_schema_migrations(&mut tx)?;
+            MemPoolDB::add_indexes(&mut tx)?;
             tx.commit().map_err(db_error::SqliteError)?;
         }
 
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -350,8 +350,7 @@ impl FromRow<(u64, u64)> for (u64, u64) {
     }
 }
 
-const MEMPOOL_INITIAL_SCHEMA: &'static [&'static str] = &[
-    r#"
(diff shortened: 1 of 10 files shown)


src/core/mempool.rs
```diff
@@ -653,6 +654,11 @@ impl MemPoolDB {
         // apply all migrations
         MemPoolDB::apply_schema_migrations(&mut tx)?;
```
pavitthrap, line 656: Wouldn't you want to add the index creation code snippet to the `open` function as well? Same comment for other dbs (like `open_db` for StacksChainState)

src/net/neighbors.rs
```diff
@@ -1890,10 +1890,16 @@ impl PeerNetwork {
...
-    /// Instantiate the neighbor walk to an always-allowed node
-    fn instantiate_walk_to_always_allowed(&mut self) -> Result<(), net_error> {
-        let allowed_peers =
-            PeerDB::get_always_allowed_peers(self.peerdb.conn(), self.local_peer.network_id)?;
+    /// Instantiate the neighbor walk to an always-allowed node.
+    /// If we're in the initial block download, then this must also be a *bootstrap* peer.
+    fn instantiate_walk_to_always_allowed(&mut self, ibd: bool) -> Result<(), net_error> {
+        let allowed_peers = if ibd {
+            // only get bootstrap peers
+            PeerDB::get_bootstrap_peers(&self.peerdb.conn(), self.local_peer.network_id)?
```
pavitthrap, line 1898: Do we want to return the subset of bootstrap peers that are allowed?

src/core/mempool.rs
jcnelson: Yes; thanks for catching that.

src/net/neighbors.rs
jcnelson: A peer can be a bootstrap peer without being allowed.  Specifically, if I put a node in the `bootstrap_node` list in my config file, it should override any other decision the node software makes about whether or not that peer should otherwise be ignored.

src/core/mempool.rs
jcnelson: Fixed

pavitthrap (approved):

kantai (approved):
LGTM -- just had one comment

src/net/neighbors.rs
jcnelson: A node can be either "allowed" or "denied" for an interval of time between now, and a future timestamp _t_.  This is how the node temporarily bans misbehaving peers, for example -- it marks them as "denied" until _t_, where _t_ grows exponentially up to a cap.  However, a node can also be marked as "always-allowed", which means "allowed, and will never be denied".  This includes bootstrap nodes.

In addition, a node can be marked as "initial", in which case, it will always be considered a candidate for inventory synchronization during initial block download.  Once the node reaches the chain tip, this consideration stops.

PR #3031, opened 2022-02-02
Initial burnchains replacement

kantai:
This PR does two major things:

1. Removes the existing bitcoin processor, replacing the Bitcoin implementation of the Burnchain with a new `StacksEventBlock` implementation. This required extensive changes in the sortition db, as many of the previously used types and methods are removed. Existing sortition and `TestPeer` tests cover these changes. 
2. Adds a parser for Stacks event dispatcher events for constructing the new `StacksEventBlock` types. This parser is tested with new testing.

Of note: item 1 above touches a lot of things, and this PR deals with the replacement in a very quick and dirty way. Ultimately, cleaning up the sortition db and the burnchains interface will take a bit more effort, but most of that work can be done in other PRs as we march towards milestone 0.


jcnelson:
Heh, wrong repo? ;)

PR #3033, opened 2022-02-03
updating metrics for issue 2987

wileyj:
2 metrics weren't as correct as they should be.
- `stacks_node_stacks_tip_height` was moved to the relayer. in tests, this proved to be a more effective location to track the data vs what `/v2/info` was reporting
- moved `stacks_node_neighbors_*` out of the pruning loop. this resulted in the metric being set more often, but the actual count of neighbors was far more accurate when compared to `/v2/neighbors`
- no changes were made to the mempool tx metric, to get this more accurate will be a far heavier lift. leaving for future work
- script used for testing: https://gist.github.com/wileyj/881eda423366134bc25e3fb5ec39ccbb

## Description
Some existing metrics were not as correct/up to date as they should be compared with `/v2` output.
  - `stacks_node_stacks_tip_height` was moved to the relayer. 
  - moved `stacks_node_neighbors_*` out of the pruning loop

In tests, the new locations of these directives showed much more accurate metric data when compared to the `/v2` output of a running/syncing stacks blockchain node. This change has no impact application developers and is not a breaking change. 
For details refer to issue https://github.com/stacks-network/stacks-blockchain/issues/2987

## Type of Change
- Bug fix

No

No

## Testing information
No blockchain tests are necessary for this change. However, I used this script https://gist.github.com/wileyj/881eda423366134bc25e3fb5ec39ccbb to help determine the current state of accuracy, compared to the updates proposed in this PR. The changes proposed are much more closely aligned with what the blockchain reports via http over `/v2` endpoints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
+## [Unreleased]
+
+### Added
+
+### Changed
+
+### Fixed
+- Fixed a few prometheus metrics to be more accurate comapared to `/v2` endpoints when polling data (#2987)
+
 ## [2.05.0.1.0]
 
 ### Added 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,15 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
-## [Unreleased]
-
-### Added
-
-### Changed
-
-### Fixed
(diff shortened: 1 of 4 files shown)


jcnelson (approved):
LGTM, as long as it's getting you the data you need.

kantai (approved):
LGTM -- please add a changelog entry, though

wileyj:
> LGTM -- please add a changelog entry, though

Added - let me know if it needs to be changed, wasn't entirely sure about the standard here for unreleased changes

kantai:
> Added - let me know if it needs to be changed, wasn't entirely sure about the standard here for unreleased changes

Thanks -- because this PR is going to be included in 2.05.0.1.0, I think that entry should go under that section. I went ahead and moved it.

wileyj:
> > Added - let me know if it needs to be changed, wasn't entirely sure about the standard here for unreleased changes
> 
> Thanks -- because this PR is going to be included in 2.05.0.1.0, I think that entry should go under that section. I went ahead and moved it.

Got it, thanks! wasn't sure if this PR would be included

Issue #3034, opened 2022-02-03
Update contract identifier for exit contract


pavitthrap:
Am in the process of implementing the exit contract (where stackers can vote on an exit height). 
Currently, the contract is initialized with the rest of the boot contracts. 

Eventually, I will have to remove it from boot data and deploy it live. When that happens I need to replace the contract identifier in the code.

PR #3035, opened 2022-02-04
Feat/exit at reward cycle contract

pavitthrap:
This PR implements vote to exit network contract & related logic. It is built off of the ideas in this discussion: https://github.com/stacks-network/stacks-blockchain/discussions/2845

## Overview of protocol 
- Stackers vote for a reward cycle at which the node should next exit. 
- For a vote to gain consensus in a particular cycle: 
    - PoX must be active 
    - 50% of stackers must have voted in the exit contract 
    - Voting is aggregated as follows:
         - Before an exit height is chosen, a vote for RC N is interpreted as voting to exit at RC N or higher. If there is an equal number of votes for RC 10, 11, and 20, RC 11 will be selected as the exit height (those who voted for RC 10 will have their votes counted towards RC 11 in this example). 
         - After some exit height H is chosen, only votes for block heights higher than H will be counted as valid in the election. 
- Once there is consensus for some reward cycle, the cycle after consensus was reached is a "veto" cycle. 
- Miners call the veto function to signal that they want to reject a particular proposal. 80% of the blocks in the veto reward cycle must contain a veto for the veto to pass. 

## Feedback 
- In `process_ready_blocks` and `process_exit_reward_cycle`, I currently do nothing if the node is unable to find the exit cycle info for a particular Stacks block. This should only happen for the first Stacks block. What should this behavior look like?

## Future steps 
- Need to deploy the exit contract, and once I do so, I must "refer" to the contract differently within the code (issue #3034). 
- Need to think about whether miners will have difficulty getting their veto transactions in the current blocks. Might need to prioritize miner veto transactions?
- Update SIP-015 draft to mention the contract

## Testing 
- new tests in contract tests, chains coordinator tests, and added a neon_integrations test

### DIFF
--- a/src/burnchains/mod.rs
+++ b/src/burnchains/mod.rs
@@ -52,6 +52,7 @@ use self::bitcoin::Error as btc_error;
 use self::bitcoin::{
     BitcoinBlock, BitcoinInputType, BitcoinTransaction, BitcoinTxInput, BitcoinTxOutput,
 };
+use vm::types::QualifiedContractIdentifier;
 
 /// This module contains drivers and types for all burn chains we support.
 pub mod bitcoin;
@@ -296,6 +297,7 @@ pub struct Burnchain {
     pub first_block_timestamp: u32,
     pub pox_constants: PoxConstants,
     pub exit_contract_constants: ExitContractConstants,
+    // pub exit_contract_id: Option<QualifiedContractIdentifier>,
     pub initial_reward_start_block: u64,
 }
 
@@ -355,11 +357,11 @@ impl ExitContractConstants {
     }
 
     pub fn mainnet_default() -> ExitContractConstants {
-        ExitContractConstants::new(80, 50, 50)
+        ExitContractConstants::new(60, 50, 50)
     }
 
     pub fn testnet_default() -> ExitContractConstants {
-        ExitContractConstants::new(80, 50, 50)
+        ExitContractConstants::new(60, 50, 50)
     }
 }
 
--- a/src/burnchains/mod.rs
+++ b/src/burnchains/mod.rs
@@ -328,9 +328,9 @@ pub struct PoxConstants {
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 pub struct ExitContractConstants {
+    // This percent of mined blocks in the rejection period must have rejected the proposal for the rejection to succeed.
+    pub rejection_confirmation_percent: u8,
     // The chosen exit reward cycle must have this percent of stackers that voted for it or for a lower reward cycle to be confirmed
-    pub veto_confirmation_percent: u8,
-    // This percent of mined blocks in the veto period must have vetoed the proposal for the veto to succeed.
     pub vote_confirmation_percent: u8,
     // For a vote to be valid in a certain reward cycle, this percent of STX must be stacked
     pub percent_stacked_stx_for_valid_vote: u8,
@@ -345,12 +345,12 @@ pub struct ExitContractConstants {
 impl ExitContractConstants {
     /// Default constants were chosen in the exit contract proposal in the stacks-blockchain repo.
     pub fn new(
-        veto_confirmation_percent: u8,
+        rejection_confirmation_percent: u8,
         vote_confirmation_percent: u8,
         percent_stacked_stx_for_valid_vote: u8,
     ) -> ExitContractConstants {
         ExitContractConstants {
-            veto_confirmation_percent,
+            rejection_confirmation_percent,
             vote_confirmation_percent,
             percent_stacked_stx_for_valid_vote,
             absolute_minimum_exit_rc: EXIT_RC_ABSOLUTE_MINIMUM_EXIT_RC,
--- a/src/burnchains/mod.rs
+++ b/src/burnchains/mod.rs
@@ -297,6 +297,9 @@ pub struct Burnchain {
     pub first_block_timestamp: u32,
     pub pox_constants: PoxConstants,
     pub exit_contract_constants: ExitContractConstants,
+    /// The contract ID of the 'exit-at-rc' contract. For tests, this contract ID is
+    /// boot_code_addr("exit-at-rc", false). For mainnet, this should be set to the
+    /// contract deployed on Stacks mainnet.
     pub exit_contract_id: Option<QualifiedContractIdentifier>,
     pub initial_reward_start_block: u64,
 }
--- a/src/burnchains/mod.rs
+++ b/src/burnchains/mod.rs
@@ -297,7 +297,7 @@ pub struct Burnchain {
     pub first_block_timestamp: u32,
     pub pox_constants: PoxConstants,
     pub exit_contract_constants: ExitContractConstants,
-    // pub exit_contract_id: Option<QualifiedContractIdentifier>,
+    pub exit_contract_id: Option<QualifiedContractIdentifier>,
     pub initial_reward_start_block: u64,
 }
 
--- a/src/burnchains/mod.rs
+++ b/src/burnchains/mod.rs
@@ -336,26 +336,23 @@ impl ExitContractConstants {
         veto_confirmation_percent: u8,
         vote_confirmation_percent: u8,
         percent_stacked_stx_for_valid_vote: u8,
-        absolute_minimum_exit_rc: u64,
-        minimum_rc_buffer_from_present: u64,
-        maximum_rc_buffer_from_present: u64,
     ) -> ExitContractConstants {
         ExitContractConstants {
             veto_confirmation_percent,
             vote_confirmation_percent,
             percent_stacked_stx_for_valid_vote,
-            absolute_minimum_exit_rc,
-            minimum_rc_buffer_from_present,
-            maximum_rc_buffer_from_present,
+            absolute_minimum_exit_rc: EXIT_RC_ABSOLUTE_MINIMUM_EXIT_RC,
+            minimum_rc_buffer_from_present: EXIT_RC_MINIMUM_RC_BUFFER_FROM_PRESENT,
+            maximum_rc_buffer_from_present: EXIT_RC_MAXIMUM_RC_BUFFER_FROM_PRESENT,
         }
     }
 
     pub fn mainnet_default() -> ExitContractConstants {
-        ExitContractConstants::new(80, 50, 50, 50, 4, 52)
+        ExitContractConstants::new(80, 50, 50)
     }
 
     pub fn testnet_default() -> ExitContractConstants {
-        ExitContractConstants::new(80, 50, 50, 15, 4, 52)
+        ExitContractConstants::new(80, 50, 50)
     }
 }
 
--- a/src/burnchains/mod.rs
+++ b/src/burnchains/mod.rs
@@ -323,15 +323,22 @@ pub struct PoxConstants {
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 pub struct ExitContractConstants {
+    // The chosen exit reward cycle must have this percent of stackers that voted for it or for a lower reward cycle to be confirmed
     pub veto_confirmation_percent: u8,
+    // This percent of mined blocks in the veto period must have vetoed the proposal for the veto to succeed.
     pub vote_confirmation_percent: u8,
+    // For a vote to be valid in a certain reward cycle, this percent of STX must be stacked
     pub percent_stacked_stx_for_valid_vote: u8,
+    // The lowest allowable exit reward cycle.
     pub absolute_minimum_exit_rc: u64,
+    // Votes for reward cycles less than the current reward cycle plus this buffer are not allowed.
     pub minimum_rc_buffer_from_present: u64,
+    // Votes for reward cycles greater than the current reward cycle plus this buffer are not allowed.
     pub maximum_rc_buffer_from_present: u64,
 }
 
 impl ExitContractConstants {
+    /// Default constants were chosen in the exit contract proposal in the stacks-blockchain repo.
     pub fn new(
         veto_confirmation_percent: u8,
         vote_confirmation_percent: u8,
--- a/src/burnchains/mod.rs
+++ b/src/burnchains/mod.rs
@@ -295,6 +295,7 @@ pub struct Burnchain {
     pub first_block_hash: BurnchainHeaderHash,
     pub first_block_timestamp: u32,
     pub pox_constants: PoxConstants,
+    pub exit_contract_constants: ExitContractConstants,
     pub initial_reward_start_block: u64,
 }
 
@@ -318,6 +319,44 @@ pub struct PoxConstants {
     /// first block height of sunset phase
     pub sunset_start: u64,
     _shadow: PhantomData<()>,
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub struct ExitContractConstants {
+    pub veto_confirmation_percent: u8,
+    pub vote_confirmation_percent: u8,
+    pub percent_stacked_stx_for_valid_vote: u8,
+    pub absolute_minimum_exit_rc: u64,
+    pub minimum_rc_buffer_from_present: u64,
+    pub maximum_rc_buffer_from_present: u64,
+}
+
+impl ExitContractConstants {
+    pub fn new(
+        veto_confirmation_percent: u8,
+        vote_confirmation_percent: u8,
+        percent_stacked_stx_for_valid_vote: u8,
+        absolute_minimum_exit_rc: u64,
+        minimum_rc_buffer_from_present: u64,
+        maximum_rc_buffer_from_present: u64,
+    ) -> ExitContractConstants {
+        ExitContractConstants {
+            veto_confirmation_percent,
+            vote_confirmation_percent,
+            percent_stacked_stx_for_valid_vote,
+            absolute_minimum_exit_rc,
+            minimum_rc_buffer_from_present,
+            maximum_rc_buffer_from_present,
+        }
+    }
+
+    pub fn mainnet_default() -> ExitContractConstants {
+        ExitContractConstants::new(80, 50, 50, 50, 4, 52)
+    }
+
+    pub fn testnet_default() -> ExitContractConstants {
+        ExitContractConstants::new(80, 50, 50, 15, 4, 52)
+    }
 }
 
 impl PoxConstants {
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -3790,33 +3790,30 @@ impl<'a> SortitionHandleTx<'a> {
         &self,
         exit_at_reward_cycle_info: BlockExitRewardCycleInfo,
     ) -> Result<(), db_error> {
-        let sql = "INSERT or REPLACE INTO exit_at_reward_cycle_info (block_id, parent_block_id, block_reward_cycle, stacks_block_height_in_cycle, invalid_reward_cycles) VALUES (?, ?, ?, ?, ?)";
+        let exit_proposal = if let Some(exit_prop) = exit_at_reward_cycle_info.curr_exit_proposal {
+            Some(u64_to_sql(exit_prop)?)
+        } else {
+            None
+        };
+        let exit_at_reward_cycle =
+            if let Some(curr_exit) = exit_at_reward_cycle_info.curr_exit_at_reward_cycle {
+                Some(u64_to_sql(curr_exit)?)
+            } else {
+                None
+            };
+
+        let sql = "INSERT or REPLACE INTO exit_at_reward_cycle_info (block_id, parent_block_id, block_reward_cycle, stacks_block_height_in_cycle, invalid_reward_cycles, curr_exit_proposal, curr_exit_at_reward_cycle) VALUES (?, ?, ?, ?, ?, ?, ?)";
         let args: &[&dyn ToSql] = &[
             &exit_at_reward_cycle_info.block_id,
             &exit_at_reward_cycle_info.parent_block_id,
             &u64_to_sql(exit_at_reward_cycle_info.block_reward_cycle)?,
             &u64_to_sql(exit_at_reward_cycle_info.stacks_block_height_in_cycle)?,
             &serde_json::to_string(&exit_at_reward_cycle_info.invalid_reward_cycles).unwrap(),
+            &exit_proposal,
+            &exit_at_reward_cycle,
         ];
         self.execute(sql, args)?;
 
-        if let Some(exit_proposal) = exit_at_reward_cycle_info.curr_exit_proposal {
-            let sql =
-                "UPDATE exit_at_reward_cycle_info SET curr_exit_proposal = ? WHERE block_id = ?";
-            let args: &[&dyn ToSql] = &[
-                &u64_to_sql(exit_proposal)?,
-                &exit_at_reward_cycle_info.block_id,
-            ];
-            self.execute(sql, args)?;
-        }
-        if let Some(exit_at_reward_cycle) = exit_at_reward_cycle_info.curr_exit_at_reward_cycle {
-            let sql = "UPDATE exit_at_reward_cycle_info SET curr_exit_at_reward_cycle = ? WHERE block_id = ?";
-            let args: &[&dyn ToSql] = &[
-                &u64_to_sql(exit_at_reward_cycle)?,
-                &exit_at_reward_cycle_info.block_id,
-            ];
-            self.execute(sql, args)?;
-        }
         Ok(())
     }
 
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -447,8 +447,6 @@ impl FromRow<StacksEpoch> for StacksEpoch {
 
 impl FromRow<BlockExitRewardCycleInfo> for BlockExitRewardCycleInfo {
     fn from_row<'a>(row: &'a Row) -> Result<BlockExitRewardCycleInfo, db_error> {
-        let block_id = StacksBlockId::from_column(row, "block_id")?;
-        let parent_block_id = StacksBlockId::from_column(row, "parent_block_id")?;
         let block_reward_cycle = u64::from_column(row, "block_reward_cycle")?;
         let curr_exit_proposal = Option::<u64>::from_column(row, "curr_exit_proposal")?;
         let curr_exit_at_reward_cycle =
@@ -458,8 +456,6 @@ impl FromRow<BlockExitRewardCycleInfo> for BlockExitRewardCycleInfo {
             .map_err(|e| db_error::SerializationError(e))?;
 
         let exit_cycle_info = BlockExitRewardCycleInfo {
-            block_id,
-            parent_block_id,
             block_reward_cycle,
             curr_exit_proposal,
             curr_exit_at_reward_cycle,
@@ -638,14 +634,12 @@ const SORTITION_DB_SCHEMA_2: &'static [&'static str] = &[r#"
 
 const SORTITION_DB_SCHEMA_3: &'static [&'static str] = &[r#"
     CREATE TABLE exit_at_reward_cycle_info (
-        block_id TEXT NOT NULL,
-        parent_block_id TEXT NOT NULL,
         block_reward_cycle INTEGER NOT NULL,
         curr_exit_proposal INTEGER,
         curr_exit_at_reward_cycle INTEGER,
         invalid_reward_cycles TEXT NOT NULL,
 
-        PRIMARY KEY(block_id)
+        PRIMARY KEY(block_reward_cycle)
     );"#];
 
 const SORTITION_DB_INDEXES: &'static [&'static str] = &[
@@ -666,7 +660,7 @@ const SORTITION_DB_INDEXES: &'static [&'static str] = &[
     "CREATE INDEX IF NOT EXISTS index_transfer_stx_burn_header_hash ON transfer_stx(burn_header_hash);",
     "CREATE INDEX IF NOT EXISTS index_missed_commits_intended_sortition_id ON missed_commits(intended_sortition_id);",
     "CREATE INDEX IF NOT EXISTS canonical_stacks_blocks ON canonical_accepted_stacks_blocks(tip_consensus_hash,stacks_block_hash);",
-    "CREATE INDEX IF NOT EXISTS exit_reward_cycle_block_ids ON exit_at_reward_cycle_info(block_id);",
+    "CREATE INDEX IF NOT EXISTS exit_reward_cycle_block_ids ON exit_at_reward_cycle_info(block_reward_cycle);",
 ];
 
 pub struct SortitionDB {
@@ -3657,10 +3651,10 @@ impl SortitionDB {
     /// Get reward cycle info for a particular block
     pub fn get_exit_at_reward_cycle_info(
         conn: &Connection,
-        stacks_block_id: &StacksBlockId,
+        reward_cycle: u64,
     ) -> Result<Option<BlockExitRewardCycleInfo>, db_error> {
-        let qry = "SELECT * FROM exit_at_reward_cycle_info WHERE block_id = ?1";
-        let result = query_row_panic(conn, qry, &[&stacks_block_id], || {
+        let qry = "SELECT * FROM exit_at_reward_cycle_info WHERE block_reward_cycle = ?1";
+        let result = query_row_panic(conn, qry, &[&u64_to_sql(reward_cycle)?], || {
             "FATAL: multiple rows for reward cycle info for one block".into()
         })?;
         Ok(result)
@@ -3799,10 +3793,8 @@ impl<'a> SortitionHandleTx<'a> {
                 None
             };
 
-        let sql = "INSERT or REPLACE INTO exit_at_reward_cycle_info (block_id, parent_block_id, block_reward_cycle, invalid_reward_cycles, curr_exit_proposal, curr_exit_at_reward_cycle) VALUES (?, ?, ?, ?, ?, ?)";
+        let sql = "INSERT or REPLACE INTO exit_at_reward_cycle_info (block_reward_cycle, invalid_reward_cycles, curr_exit_proposal, curr_exit_at_reward_cycle) VALUES (?, ?, ?, ?)";
         let args: &[&dyn ToSql] = &[
-            &exit_at_reward_cycle_info.block_id,
-            &exit_at_reward_cycle_info.parent_block_id,
             &u64_to_sql(exit_at_reward_cycle_info.block_reward_cycle)?,
             &serde_json::to_string(&exit_at_reward_cycle_info.invalid_reward_cycles).unwrap(),
             &exit_proposal,
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -2544,20 +2544,17 @@ impl SortitionDB {
 
     fn check_schema_version_and_update(&mut self, epochs: &[StacksEpoch]) -> Result<(), db_error> {
         let tx = self.tx_begin()?;
-        let mut updated = false;
-        while !updated {
+        let mut cur_version = "1".to_string();
+        while cur_version != SORTITION_DB_VERSION {
             match SortitionDB::get_schema_version(&tx) {
                 Ok(Some(version)) => {
-                    let expected_version = SORTITION_DB_VERSION.to_string();
-                    if version == expected_version {
-                        return Ok(());
-                    }
+                    cur_version = version.clone();
+
                     if version == "1" {
                         SortitionDB::apply_schema_2(&tx, epochs)?;
                     } else if version == "2" {
                         SortitionDB::apply_schema_3(&tx)?;
-                        updated = true;
-                    } else {
+                    } else if version != SORTITION_DB_VERSION {
                         panic!("The schema version of the sortition DB is invalid.")
                     }
                 }
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -450,7 +450,6 @@ impl FromRow<BlockExitRewardCycleInfo> for BlockExitRewardCycleInfo {
         let block_id = StacksBlockId::from_column(row, "block_id")?;
         let parent_block_id = StacksBlockId::from_column(row, "parent_block_id")?;
         let block_reward_cycle = u64::from_column(row, "block_reward_cycle")?;
-        let stacks_block_height_in_cycle = u64::from_column(row, "stacks_block_height_in_cycle")?;
         let curr_exit_proposal = Option::<u64>::from_column(row, "curr_exit_proposal")?;
         let curr_exit_at_reward_cycle =
             Option::<u64>::from_column(row, "curr_exit_at_reward_cycle")?;
@@ -462,7 +461,6 @@ impl FromRow<BlockExitRewardCycleInfo> for BlockExitRewardCycleInfo {
             block_id,
             parent_block_id,
             block_reward_cycle,
-            stacks_block_height_in_cycle,
             curr_exit_proposal,
             curr_exit_at_reward_cycle,
             invalid_reward_cycles,
@@ -643,7 +641,6 @@ const SORTITION_DB_SCHEMA_3: &'static [&'static str] = &[r#"
         block_id TEXT NOT NULL,
         parent_block_id TEXT NOT NULL,
         block_reward_cycle INTEGER NOT NULL,
-        stacks_block_height_in_cycle INTEGER NOT NULL,
         curr_exit_proposal INTEGER,
         curr_exit_at_reward_cycle INTEGER,
         invalid_reward_cycles TEXT NOT NULL,
@@ -3802,12 +3799,11 @@ impl<'a> SortitionHandleTx<'a> {
                 None
             };
 
-        let sql = "INSERT or REPLACE INTO exit_at_reward_cycle_info (block_id, parent_block_id, block_reward_cycle, stacks_block_height_in_cycle, invalid_reward_cycles, curr_exit_proposal, curr_exit_at_reward_cycle) VALUES (?, ?, ?, ?, ?, ?, ?)";
+        let sql = "INSERT or REPLACE INTO exit_at_reward_cycle_info (block_id, parent_block_id, block_reward_cycle, invalid_reward_cycles, curr_exit_proposal, curr_exit_at_reward_cycle) VALUES (?, ?, ?, ?, ?, ?)";
         let args: &[&dyn ToSql] = &[
             &exit_at_reward_cycle_info.block_id,
             &exit_at_reward_cycle_info.parent_block_id,
             &u64_to_sql(exit_at_reward_cycle_info.block_reward_cycle)?,
-            &u64_to_sql(exit_at_reward_cycle_info.stacks_block_height_in_cycle)?,
             &serde_json::to_string(&exit_at_reward_cycle_info.invalid_reward_cycles).unwrap(),
             &exit_proposal,
             &exit_at_reward_cycle,
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -3650,7 +3650,7 @@ impl SortitionDB {
         minimum_reward_cycle: u64,
     ) -> Result<Vec<u64>, db_error> {
         let qry =
-            "SELECT * FROM exit_at_reward_cycle_veto_info WHERE vetoed_exit_at_reward_cycle > ?";
+            "SELECT * FROM exit_at_reward_cycle_veto_info WHERE vetoed_exit_at_reward_cycle >= ?";
         query_rows(conn, qry, &[&u64_to_sql(minimum_reward_cycle)?])
     }
 
@@ -3659,7 +3659,7 @@ impl SortitionDB {
         minimum_reward_cycle: u64,
     ) -> Result<Vec<u64>, db_error> {
         let qry =
-            "SELECT * FROM exit_at_reward_cycle_proposal_info WHERE proposed_exit_at_reward_cycle > ?";
+            "SELECT * FROM exit_at_reward_cycle_proposal_info WHERE proposed_exit_at_reward_cycle >= ?";
         query_rows(conn, qry, &[&u64_to_sql(minimum_reward_cycle)?])
     }
 }
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -454,6 +454,9 @@ impl FromRow<BlockExitRewardCycleInfo> for BlockExitRewardCycleInfo {
         let curr_exit_proposal = Option::<u64>::from_column(row, "curr_exit_proposal")?;
         let curr_exit_at_reward_cycle =
             Option::<u64>::from_column(row, "curr_exit_at_reward_cycle")?;
+        let invalid_reward_cycles_json: String = row.get_unwrap("invalid_reward_cycles");
+        let invalid_reward_cycles = serde_json::from_str(&invalid_reward_cycles_json)
+            .map_err(|e| db_error::SerializationError(e))?;
 
         let exit_cycle_info = BlockExitRewardCycleInfo {
             block_id,
@@ -462,6 +465,7 @@ impl FromRow<BlockExitRewardCycleInfo> for BlockExitRewardCycleInfo {
             stacks_block_height_in_cycle,
             curr_exit_proposal,
             curr_exit_at_reward_cycle,
+            invalid_reward_cycles,
         };
         Ok(exit_cycle_info)
     }
@@ -630,6 +634,7 @@ const SORTITION_DB_INITIAL_SCHEMA: &'static [&'static str] = &[
         stacks_block_height_in_cycle INTEGER NOT NULL,
         curr_exit_proposal INTEGER,
         curr_exit_at_reward_cycle INTEGER,
+        invalid_reward_cycles TEXT NOT NULL,
 
         PRIMARY KEY(block_id)
     );"#,
@@ -3791,12 +3796,13 @@ impl<'a> SortitionHandleTx<'a> {
         &self,
         exit_at_reward_cycle_info: BlockExitRewardCycleInfo,
     ) -> Result<(), db_error> {
-        let sql = "INSERT or REPLACE INTO exit_at_reward_cycle_info (block_id, parent_block_id, block_reward_cycle, stacks_block_height_in_cycle) VALUES (?, ?, ?, ?)";
+        let sql = "INSERT or REPLACE INTO exit_at_reward_cycle_info (block_id, parent_block_id, block_reward_cycle, stacks_block_height_in_cycle, invalid_reward_cycles) VALUES (?, ?, ?, ?, ?)";
         let args: &[&dyn ToSql] = &[
             &exit_at_reward_cycle_info.block_id,
             &exit_at_reward_cycle_info.parent_block_id,
             &u64_to_sql(exit_at_reward_cycle_info.block_reward_cycle)?,
             &u64_to_sql(exit_at_reward_cycle_info.stacks_block_height_in_cycle)?,
+            &serde_json::to_string(&exit_at_reward_cycle_info.invalid_reward_cycles).unwrap(),
         ];
         self.execute(sql, args)?;
 
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -48,7 +48,9 @@ use chainstate::burn::operations::{
 };
 use chainstate::burn::Opcodes;
 use chainstate::burn::{BlockSnapshot, ConsensusHash, OpsHash, SortitionHash};
-use chainstate::coordinator::{Error as CoordinatorError, PoxAnchorBlockStatus, RewardCycleInfo};
+use chainstate::coordinator::{
+    BlockExitRewardCycleInfo, Error as CoordinatorError, PoxAnchorBlockStatus, RewardCycleInfo,
+};
 use chainstate::stacks::db::{StacksChainState, StacksHeaderInfo};
 use chainstate::stacks::index::marf::MarfConnection;
 use chainstate::stacks::index::marf::MARF;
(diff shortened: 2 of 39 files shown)


gregorycoppola:
Thanks for important change.

What about these test failures? Are they flakes or real?

https://github.com/stacks-network/stacks-blockchain/runs/5072277005?check_suite_focus=true

src/burnchains/mod.rs
```diff
@@ -320,6 +321,41 @@ pub struct PoxConstants {
     _shadow: PhantomData<()>,
 }
 
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
```
gregorycoppola, line 324: can comment explain which part of the code will use this?

src/chainstate/coordinator/mod.rs
```diff
@@ -613,7 +654,45 @@ impl<
             };
 
             // at this point, we need to figure out if the sortition we are
-            //  about to process is the first block in reward cycle.
+            //  about to process is the first block in the exit reward cycle.
```
gregorycoppola, line 657: can this be a self-contained function to keep the function length from growing too big?

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +737,240 @@ impl<
         Ok(())
     }
 
+    /// This function reads veto-related information from the exit-at-rc clarity contract.
+    /// It returns true if the veto succeeded, and false if it failed.
+    pub fn read_veto_state(
```
gregorycoppola, line 742: why is `self` mutable if this is a "read"?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -6680,3 +6688,1393 @@ fn test_flash_block_skip_tenure() {
 
     channel.stop_chains_coordinator();
 }
+
+#[test]
+#[ignore]
+fn exit_at_rc_integration_test() {
```
gregorycoppola, line 6694: is this really one indivisiable test?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -6680,3 +6688,1393 @@ fn test_flash_block_skip_tenure() {
...
+        // The run loop should terminate at RC 53
+        if !btc_regtest_controller.should_keep_running() {
+            info!("about to terminate");
+            test_observer::clear();
+            info!("stop cc");
+            return;
+        }
+    }
+
+    panic!("Should not have reached this line; node should exit once it gets to RC 53");
```
gregorycoppola, line 8079: why does this test end this way?

testnet/stacks-node/src/tests/neon_integrations.rs
gregorycoppola, line 6694: can you briefly explain the logic of the test?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -6680,3 +6688,1393 @@ fn test_flash_block_skip_tenure() {
...
 }
+
+#[test]
+#[ignore]
+fn exit_at_rc_integration_test() {
+    if env::var("BITCOIND_TEST") != Ok("1".into()) {
+        return;
+    }
+
+    let spender_sk = StacksPrivateKey::new();
```
gregorycoppola, line 6699: this gets copied around but could we make a reusable util function that would return triples of (sk, addr, stacks_addr)?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -6680,3 +6688,1393 @@ fn test_flash_block_skip_tenure() {
...
+
+    // first block will hold our VRF registration
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    // second block will be the first mined Stacks block
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    let sort_height = channel.get_sortitions_processed();
+
+    // let's query the miner's account nonce:
```
gregorycoppola, line 6809: do you really need these asserts? this isn't under test here.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -6680,3 +6688,1393 @@ fn test_flash_block_skip_tenure() {
...
+                    pox_pubkey_hash
+                ))
+                .unwrap()
+                .unwrap(),
+                Value::UInt(sort_height as u128),
+                Value::UInt(4),
+            ],
+        );
+
+        // okay, let's push that stacking transaction!
```
gregorycoppola, line 6854: I would delete this comment and comment on what the loop does before the `for` line.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -6680,3 +6688,1393 @@ fn test_flash_block_skip_tenure() {
...
+    let blocks_observed = test_observer::get_blocks();
+    assert!(
+        blocks_observed.len() >= 2,
+        "Blocks observed {} should be >= 2",
+        blocks_observed.len()
+    );
+
+    // look up the return value of our stacking operations...
+    let mut tested_stacking = 0;
+    for block in blocks_observed.iter() {
```
gregorycoppola, line 6874: i know we copy this around but this can also be factored into a function that takes a functional argument about  which tx to select and what to do with it.

i can write it if you think this is too much to ask if you ping me. or we can file a bug.

but i think we should start introducing the pattern where this uses a reusable function to cut down on code and make it easier to read.

gregorycoppola (commented):
Lots of good code! Still reading. Had some interim comments.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -626,7 +668,8 @@ const SORTITION_DB_INDEXES: &'static [&'static str] = &[
     "CREATE INDEX IF NOT EXISTS index_stack_stx_burn_header_hash ON stack_stx(burn_header_hash);",
     "CREATE INDEX IF NOT EXISTS index_transfer_stx_burn_header_hash ON transfer_stx(burn_header_hash);",
     "CREATE INDEX IF NOT EXISTS index_missed_commits_intended_sortition_id ON missed_commits(intended_sortition_id);",
-    "CREATE INDEX IF NOT EXISTS canonical_stacks_blocks ON canonical_accepted_stacks_blocks(tip_consensus_hash,stacks_block_hash);"
+    "CREATE INDEX IF NOT EXISTS canonical_stacks_blocks ON canonical_accepted_stacks_blocks(tip_consensus_hash,stacks_block_hash);",
+    "CREATE INDEX IF NOT EXISTS exit_reward_cycle_block_ids ON exit_at_reward_cycle_info(block_id);",
```
jcnelson, line 672: No need to add an automated test for this, but when adding indexes, just please be sure to verify that all queries against the table (`exit_at_reward_cycle_info`) result in query plans that use only `SEARCH`es and do _not_ use `COVERING INDEX`.  Thanks!

src/chainstate/burn/db/sortdb.rs
```diff
@@ -3699,6 +3786,52 @@ impl<'a> SortitionHandleTx<'a> {
...
+            &exit_at_reward_cycle_info.block_id,
+            &exit_at_reward_cycle_info.parent_block_id,
+            &u64_to_sql(exit_at_reward_cycle_info.block_reward_cycle)?,
+            &u64_to_sql(exit_at_reward_cycle_info.stacks_block_height_in_cycle)?,
+            &serde_json::to_string(&exit_at_reward_cycle_info.invalid_reward_cycles).unwrap(),
+        ];
+        self.execute(sql, args)?;
+
+        if let Some(exit_proposal) = exit_at_reward_cycle_info.curr_exit_proposal {
+            let sql =
```
jcnelson, line 3804: Should this verify that exactly one row was set?  Is it possible that this `UPDATE` touches zero rows?

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +737,240 @@ impl<
...
+                            tuple_data
+                                .get("vetoes")
+                                .expect("BUG: malformed cost proposal tuple")
+                                .clone()
+                                .expect_u128()
+                        }
+                        None => 0,
+                    };
+
+                    println!(
```
jcnelson, line 788: Can you change this to an `info!`?

jcnelson: Same for the other `println!`'s below.

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +737,240 @@ impl<
...
+                    }
+                })
+            })
+            .ok_or(Error::DBError(NotFoundError))?;
+
+        Ok((vote_map, total_votes))
+    }
+
+    /// At the end of each reward cycle, we tally the votes for the exit at RC contract.
+    /// We need to read PoX contract state to see how much STX is staked into the protocol - we then
```
jcnelson, line 892: "staked" --> "stacked" ;)

src/chainstate/coordinator/mod.rs
jcnelson: Seconded. This should be a separate function.

src/chainstate/coordinator/mod.rs
jcnelson: `with_read_only_clarity_tx` requires `&mut self`, since the MARF operations all require `&mut self`, since the `TrieFileStorage` keeps mutable benchmarking data on the read path.

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +737,240 @@ impl<
...
+        let veto_percent_threshold = self
+            .burnchain
+            .exit_contract_constants
+            .veto_confirmation_percent;
+        let is_mainnet = self.burnchain.is_mainnet();
+        self.chain_state_db
+            .with_read_only_clarity_tx(&self.sortition_db.index_conn(), &stacks_block_id, |conn| {
+                conn.with_clarity_db_readonly(|db| {
+                    // TODO (#3034): get correct contract ID once contract is published
+                    let exit_at_rc_contract = boot_code_id("exit-at-rc", is_mainnet);
```
jcnelson, line 758: Just to make sure we don't accidentally leave any of these calls to `boot_code_id()` around, could you rename this to `exit_at_reward_cycle_code_id()` or something like that and for now have it use the boot code address?  Thanks!

src/chainstate/coordinator/mod.rs
```diff
@@ -683,6 +996,42 @@ impl<
         )
     }
 
+    // If it is time to process the exit-at-rc contract for a specific reward cycle, this function
+    // returns that reward cycle. Otherwise, the function returns None.
```
jcnelson, line 1000: Stale comment?

src/chainstate/coordinator/mod.rs
```diff
@@ -683,6 +996,42 @@ impl<
...
+        parent_exit_info: &BlockExitRewardCycleInfo,
+    ) -> Result<Option<BlockExitRewardCycleInfo>, Error> {
+        info!(
+            "RCPR: curr height: {}, parent height: {}",
+            stacks_block_height_in_cycle, parent_exit_info.stacks_block_height_in_cycle
+        );
+        if stacks_block_height_in_cycle == 2 {
+            // We want to process the exit-at-rc contract reward cycle for the reward cycle the
+            // current block's parent's parent belongs to since the height (in number of Stacks
+            // blocks) of the current block in its reward cycle is 2.
```
jcnelson, line 1013: It's not clear to me why this is?  Why not process the votes in the first block of a reward cycle, or even the last block?

src/chainstate/coordinator/mod.rs
```diff
@@ -683,6 +996,42 @@ impl<
...
+                self.sortition_db.conn(),
+                &parent_exit_info.parent_block_id,
+            )?;
+            Ok(exit_info_opt)
+        } else if stacks_block_height_in_cycle == 1
+            && parent_exit_info.stacks_block_height_in_cycle == 1
+        {
+            // This is an edge case in which the previous reward cycle contains only a single Stacks block.
+            // It is time to process the exit-at-rc contract for the reward cycle the current block's
+            // parent's parent belongs to.
```
jcnelson, line 1024: If there's a reward cycle that contains only a single Stacks block, it might simplify things to simply declare that a vote tally will not occur.  In such conditions, the network is already well and truly dead anyway.

src/chainstate/coordinator/mod.rs
```diff
@@ -845,12 +1196,156 @@ impl<
                         );
                     }
 
+                    // compute and store information relating to exiting at a reward cycle
```
jcnelson, line 1199: Can you factor this into its own function? Thanks!

src/chainstate/coordinator/mod.rs
```diff
@@ -845,12 +1196,156 @@ impl<
...
+                            stacks_block_height_in_cycle = 1;
+                        } else if parent_exit_info.block_reward_cycle == current_reward_cycle {
+                            current_proposal = parent_exit_info.curr_exit_proposal;
+                            stacks_block_height_in_cycle =
+                                parent_exit_info.stacks_block_height_in_cycle + 1;
+                        } else {
+                            error!("A child block should never have a lower reward cycle than its parent");
+                            panic!();
+                        }
+                        invalid_reward_cycles = parent_exit_info
```
jcnelson, line 1236: Hmm, it seems that `invalid_reward_cycles` grows linearly with the chain history, unbounded?  Is there a way we can avoid this (i.e. do O(1) work per reward cycle), so it doesn't lead to a node stall over time at each reward cycle boundary?

src/chainstate/stacks/boot/exit-at-rc.clar
```diff
@@ -0,0 +1,199 @@
...
+(define-constant ERR_NOT_ALLOWED 19)
+(define-constant ERR_INVALID_PROPOSED_RC 21)
+
+;; Constants
+(define-constant ABSOLUTE_MINIMUM_EXIT_RC u33)
+(define-constant MAXIMUM_RC_BUFFER_FROM_PRESENT u25)
+(define-constant MINIMUM_RC_BUFFER_FROM_PRESENT u6)
+
+;; Data vars
+(define-data-var pox-reward-cycle-length uint POX_REWARD_CYCLE_LENGTH)
```
jcnelson, line 19: Instead of duplicating this here, would it be better if the contract simply pulled this information from `.pox` via `get-pox-info`?

jcnelson: This also applies to `first-burnchain-block-height` below.

jcnelson: It would also be more efficient to store all of these vars as a single tuple data variable, since then reading it would incur only one MARF read.  The initial values for these variables could be instantiated directly from `contract-call?`'s to `.pox`.

src/chainstate/stacks/boot/exit-at-rc.clar
```diff
@@ -0,0 +1,199 @@
...
+        amount: uint,
+        expiration-reward-cycle: uint
+    }
+)
+
+;; What's the reward cycle number of the burnchain block height?
+;; Will runtime-abort if height is less than the first burnchain block (this is intentional).
+;; Returns uint
+(define-private (burn-height-to-reward-cycle (height uint))
+    (let
```
jcnelson, line 76: So, I realize that `burn-height-to-reward-cycle` is private in `.pox`, but it's also more efficient than this since it avoids the need to create a lexical scope and declare variables.  Want to just copypasta that instead?

```
(define-private (burn-height-to-reward-cycle (height uint)) 
    (/ (- height (var-get first-burnchain-block-height)) (var-get pox-reward-cycle-length)))
```

src/chainstate/stacks/boot/exit-at-rc.clar
```diff
@@ -0,0 +1,199 @@
...
+    (burn-height-to-reward-cycle burn-block-height))
+
+
+;; For a specific reward cycle, this function tried to add "amount" number of votes for the proposed exit reward cycle.
+;; Note: votes are tied to reward cycles, since the vote's lifetime is tied to the voter's stacking duration.
+(define-private (add-to-rc-proposal-map (cycle-opt (optional uint)) (proposed-rc uint) (amount uint))
+    (begin
+        (match cycle-opt
+            cycle (match (map-get? rc-proposal-votes {proposed-rc: proposed-rc, curr-rc: cycle})
+                              existing-votes (map-set rc-proposal-votes {proposed-rc: proposed-rc, curr-rc: cycle} {votes: (+ amount (get votes existing-votes)) })
```
jcnelson, line 101: Minor style nit: you can make this more succinct with `merge`:

```
existing-votes (map-set rc-proposal-votes (merge existing-votes { votes: (+ amount (get votes existing-votes)) }))
```

src/chainstate/stacks/boot/exit-at-rc.clar
```diff
@@ -0,0 +1,199 @@
...
+            none
+        )
+    )
+)
+
+;; TODO (#3034) - fill in real address for contract-call once known
+;; A stacking voter with no outstanding vote can call this function with their proposed exit reward cycle to vote for it.
+;; This function enforces bounds on the vote (can't be above/below specific values).
+;; If a vote is accepted, the voter can only re-vote when they stack again.
+(define-public (vote-for-exit-rc (proposed-exit-rc uint))
```
jcnelson, line 127: In terms of test plans for Clarity code, one thing I've found makes code like this much more robust is to separate the validity checks into their own method, and the mutations in the success case into their own methods, and then have that validity check return all the computed state needed to carry out the mutations.  Then, you can separately and thoroughly test the validity checks without worrying about side-effects altering their behavior, and similarly, you can test the mutation code without having to mock the chainstate necessary for the validity checks to pass.

I have an example here, if you want to see how it can be done in practice: https://github.com/jcnelson/nftree/blob/main/contracts/nftree.clar#L865.  The function in question fulfills a buy order by (1) checking that the buy offer is valid (via `inner-can-fulfill-buy-offer?`), and (2) passing the result of the check into `inner-fulfill-buy-offer`.  This neatly factors testing into a side-effect-free test for the validity checks (https://github.com/jcnelson/nftree/blob/main/contracts/tests/test-nftree.clar#L1293), and a much-simpler test for the mutation (https://github.com/jcnelson/nftree/blob/main/contracts/tests/test-nftree.clar#L1380).

src/chainstate/stacks/boot/exit-at-rc.clar
```diff
@@ -0,0 +1,199 @@
...
+;; This function is used by miners to veto a proposed exit reward cycle. The veto period is active the reward cycle
+;; after a vote is confirmed.
+;; Note: a miner can send in a veto in the block after the one they mined, and they can't include multiple of these
+;; transactions in a block.
+(define-public (veto-exit-rc (proposed-exit-rc uint))
+    (let (
+        (current-reward-cycle (unwrap-panic (current-pox-reward-cycle)))
+        (curr-vetoes (default-to u0 (get vetoes (map-get? rc-proposal-vetoes { proposed-rc: proposed-exit-rc, curr-rc: current-reward-cycle } ))))
+        (last-miner (unwrap! (get-block-info? miner-address (- block-height u1))
+                    (err ERR_FETCHING_BLOCK_INFO)))
```
jcnelson, line 182: I'm not sure this can ever fail

src/chainstate/stacks/boot/exit-at-rc.clar
```diff
@@ -0,0 +1,199 @@
...
+        (vetoed (default-to false (get vetoed (map-get? exercised-veto { proposed-rc: proposed-exit-rc, veto-height: block-height }))))
+    )
+
+    ;; a miner can only veto once per block
+    (asserts! (not vetoed) (err ERR_ALREADY_VETOED))
+
+    ;; a miner can only veto if they mined the previous block
+    (asserts! (is-eq contract-caller last-miner) (err ERR_UNAUTHORIZED_CALLER))
```
jcnelson, line 190: This feels fragile.  It looks like if I'm a miner and I'm vetoing, then if the next miner is not vetoing, they could simply ignore my transaction and block my veto.  Is there a way to instead permit a miner to vote at any point in time prior to this reward cycle, and submit evidence that they were the miner for a given block (or blocks)?

src/chainstate/stacks/boot/exit-at-rc.clar
```diff
@@ -0,0 +1,199 @@
...
+    {
+        proposed-rc: uint,
+        veto-height: uint
+    }
+    { vetoed: bool }
+)
+
+;; Keeps track of voter specific information; used to ensure a voter doesn't vote twice in the same stacking period.
+(define-map voter-state
+    { address: principal }
```
jcnelson, line 64: If `voter-state`'s map key is a tuple with just one type, you can just omit the tuple and use that inner type directly:

```
(define-map voter-state
   principal
   { 
      ;; value fields
    }
 )
 ```

File: src/chainstate/stacks/boot/exit-at-rc.clar (hunk 23 above)
Comment (jcnelson) [reply src/chainstate/stacks/boot/exit-at-rc.clar:127]: Another tactic is to separate out the code from the global variables as much as possible (e.g. `block-height`, `tx-sender`, etc.) by wrapping the logic in a method that takes these values as arguments.  This makes it much easier to unit-test the logic without setting up a runtime environment for it.

File: src/chainstate/stacks/boot/exit-at-rc.clar [hunk 27]
```diff
@@ -0,0 +1,199 @@
...
+        ;; Check that the voter does not have an outstanding vote for this reward cycle
+        (match (map-get? voter-state {address: tx-sender})
+            voter-info (asserts! (>= current-reward-cycle (get expiration-reward-cycle voter-info))
+                (err ERR_PREVIOUS_VOTE_VALID))
+            ;; no existing state
+            true
+        )
+
+        ;; Check that the caller is allowed
+        (asserts! (is-eq tx-sender contract-caller) (err ERR_UNAUTHORIZED_CALLER))
```
jcnelson, line 160: This effectively prevents stacking pools from voting, which I don't think we want?  Do we instead want to check that `contract-caller` is authorized explicitly by `tx-sender` to vote on its behalf, like we do in `.pox`?  Ideally the authorization state from `.pox` would be available to this code so the authorization could just carry over, but unfortunately it's private.

src/chainstate/stacks/boot/exit-at-rc.clar
```diff
@@ -0,0 +1,199 @@
...
+;; Will runtime-abort if height is less than the first burnchain block (this is intentional).
+;; Returns uint
+(define-private (burn-height-to-reward-cycle (height uint))
+    (let
+        (
+            (local-first-burnchain-block-height (var-get first-burnchain-block-height))
+            (local-pox-reward-cycle-length (var-get pox-reward-cycle-length))
+        )
+
+        (asserts! (> height local-first-burnchain-block-height) (err ERR_BURN_BLOCK_HEIGHT_TOO_LOW))
```
jcnelson, line 82: I think these `asserts!`s belong in `vote-for-exit-rc`.

testnet/stacks-node/src/tests/neon_integrations.rs
pavitthrap: There are probably a few I could split out, but a lot of the test cases within are dependent on each other. 

Also, the advantage of testing these in one integration test is that we can see how a variety of different scenarios would inter-play with each other (giving us greater confidence for this contract being deployed on our chain)

testnet/stacks-node/src/tests/neon_integrations.rs
pavitthrap: Since the node exits, the test itself should exit (with exit code 0) at reward cycle 53. Thus, this panic should never be reached.

testnet/stacks-node/src/tests/neon_integrations.rs
pavitthrap: I think I will file a bug for this. 
I agree though, this pattern is hard to read.

pavitthrap: #3073

src/chainstate/burn/db/sortdb.rs
pavitthrap: Verified that the query plans use SEARCH

src/chainstate/coordinator/mod.rs
pavitthrap: Good idea!

src/chainstate/coordinator/mod.rs
pavitthrap: Updated it.

src/chainstate/coordinator/mod.rs
pavitthrap: Aaron felt it was a better idea to stall exit contract processing until after the first block of a reward cycle (just to avoid doing more computation in the first block of a reward cycle). 

I don't do it in the last block of the reward cycle since that would invalidate the last veto.

src/chainstate/coordinator/mod.rs
pavitthrap: The processing occurs for the parent cycle (which could be any length) so not sure if we should introduce that rule.

src/chainstate/coordinator/mod.rs
pavitthrap: It doesn't, it's bounded in size. 
In the logic here, we prune the invalid reward cycles of the parent, and only keep those that are greater than the current reward cycle. 
For a cycle to get added to this list, that cycle must have been chosen as a proposal and vetoed. Since there is a maximum allowable range for the voting, the size of the list can't exceed the size of that range (which is ~30).

src/chainstate/stacks/boot/exit-at-rc.clar
pavitthrap: It looks like `merge` just returns the value tuple. I would need to re-write the key tuple regardless, so I don't think it saves space here.

src/chainstate/stacks/boot/exit-at-rc.clar
pavitthrap: hmm.. I wonder if it's a possibility to expose this in pox v2. In that release of the blockchain, we could add the exit contract as a boot contract, and use that authorization function. 

For this next breaking change though, I can add custom authorization logic to the exit contract.

src/chainstate/coordinator/mod.rs
```diff
@@ -77,6 +94,26 @@ pub enum PoxAnchorBlockStatus {
...
 
+#[derive(Debug)]
+pub struct BlockExitRewardCycleInfo {
+    // The block id that corresponds to this exit cycle information
+    pub block_id: StacksBlockId,
+    // The block id of the parent
+    pub parent_block_id: StacksBlockId,
+    // The reward cycle of the block that corresponds to this exit cycle information
+    pub block_reward_cycle: u64,
+    // The stacks block "height" within the reward cycle that this block belongs to
```
jcnelson, line 105: Does this mean that this value varies between 0 and 2100?

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +737,240 @@ impl<
...
+                        .expect("BUG: Failed querying rc-proposal-vetoes")
+                        .expect_optional();
+
+                    let num_vetos = match entry {
+                        Some(val) => {
+                            let tuple_data = val.expect_tuple();
+                            tuple_data
+                                .get("vetoes")
+                                .expect("BUG: malformed cost proposal tuple")
+                                .clone()
```
jcnelson, line 782: Is this `.clone()` needed?

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +737,240 @@ impl<
...
+    }
+
+    /// Returns map of exit proposals to the number of votes for each, as well as the sum total of all
+    /// votes.
+    pub fn read_vote_state(
+        &mut self,
+        rc_cycle_of_vote: u64,
+        curr_exit_at_rc_opt: Option<u64>,
+        stacks_block_id: &StacksBlockId,
+        invalid_reward_cycles: Vec<u64>,
```
jcnelson, line 808: If this list is bound in length, then I think it should be a slice of `Option<u64>` (where `Some(..)` indicates that the entry should be considered).

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +737,240 @@ impl<
...
+                    let percent_veto = num_vetos * 100 / (reward_cycle_length as u128);
+
+                    Ok(percent_veto >= (veto_percent_threshold as u128))
+                })
+            })
+            .ok_or(Error::DBError(NotFoundError))?
+    }
+
+    /// Returns map of exit proposals to the number of votes for each, as well as the sum total of all
+    /// votes.
```
jcnelson, line 802: Can you update this comment to describe the range of key values for the resulting `BTreeMap`?  I.e. the keys are reward cycle IDs that fall into a particular range relative to `rc_cycle_of_vote`?

src/chainstate/coordinator/mod.rs
```diff
@@ -845,12 +1196,156 @@ impl<
...
+                    let exit_info_opt = SortitionDB::get_exit_at_reward_cycle_info(
+                        self.sortition_db.conn(),
+                        &parent_block_id,
+                    )?;
+
+                    if let Some(parent_exit_info) = exit_info_opt {
+                        info!("RCPR: cc: parent block info is non-None");
+                        // copy info from parent block
+                        current_exit_at_rc = parent_exit_info.curr_exit_at_reward_cycle;
+                        if parent_exit_info.block_reward_cycle < current_reward_cycle {
```
jcnelson, line 1226: Wouldn't this always be true?

src/chainstate/coordinator/mod.rs
```diff
@@ -845,12 +1196,156 @@ impl<
...
+                            stacks_block_height_in_cycle = 1;
+                        } else if parent_exit_info.block_reward_cycle == current_reward_cycle {
```
jcnelson, line 1228: I'm not sure how this is reached?

src/chainstate/coordinator/mod.rs
jcnelson: > Aaron felt it was a better idea to stall exit contract processing until after the first block of a reward cycle (just to avoid doing more computation in the first block of a reward cycle).

I'm not sure the trade-off is worth it here, though (cc @kantai).  If you process the exit-at-rc logic at Stacks block height mod 2, then you have to introduce and test corner cases where you have a reward cycle with only one block.  If we can instead make the PoX reward set calculation faster (something that's useful on its own anyway, and something that we could reasonably do), then the concern about doing it in the first reward cycle block goes away, and no corner cases are needed.  This is all consensus-critical code, so if we can avoid less code even if it means a modest performance hit (or a need to spend some cycles optimizing code), then I think it's worth it.  But would love to hear a strong counter-argument.

src/chainstate/coordinator/mod.rs
```diff
@@ -845,12 +1196,156 @@ impl<
...
+                    let current_block_height = block_receipt.header.burn_header_height as u64;
+                    let current_reward_cycle = self
+                        .burnchain
+                        .block_height_to_reward_cycle(current_block_height)
+                        .ok_or_else(|| DBError::NotFoundError)?;
+                    let mut current_exit_at_rc = None;
+                    let mut current_proposal = None;
+                    let mut invalid_reward_cycles = vec![];
+                    let mut stacks_block_height_in_cycle = 1;
+                    let stacks_block_id =
```
jcnelson, line 1209: Since `stacks_block_id` is derived from `block_receipt.header.consensus_hash`, might the APIs for vote-tallying and veto-tallying make more sense if they were passed `block_hash` instead of `stacks_block_id`?  Then, they can calculate the ID hash themselves, and not have to worry about defensively checking that `consensus_hash` matches `stacks_block_id`.

src/chainstate/coordinator/mod.rs
```diff
@@ -845,12 +1196,156 @@ impl<
...
+                                    &stacks_block_id,
+                                    &block_receipt.header.consensus_hash,
+                                )?;
+                                if let Some(exit_proposal) = tallied_proposal {
+                                    info!(
+                                        "RCPR: cc: storing proposal info for {:?}",
+                                        exit_proposal
+                                    );
+                                    // record the proposal in invalid_reward_cycles
+                                    let sortdb_handle = self
```
jcnelson, line 1294: Wanna combine these transactions?

jcnelson: Actually, why is this first transaction necessary?  It seems that it's not actually committing anything.

src/chainstate/coordinator/mod.rs
pavitthrap: Yes it does

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +737,240 @@ impl<
...
+            .with_read_only_clarity_tx(&self.sortition_db.index_conn(), &stacks_block_id, |conn| {
+                conn.with_clarity_db_readonly(|db| {
+                    // TODO (#3034): get correct contract ID once contract is published
+                    let exit_at_rc_contract = boot_code_id("exit-at-rc", is_mainnet);
+
+                    for proposed_exit_rc in min_rc..max_rc {
+                        if invalid_reward_cycles.contains(&proposed_exit_rc) {
+                            continue;
+                        }
+                        // from map rc-proposal-votes, use key pair (proposed_rc, curr_rc) to get the # of vetos
```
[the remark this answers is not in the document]
pavitthrap: Fixed the typo

src/chainstate/coordinator/mod.rs
pavitthrap: No, it's possible that a child and parent block are in the same reward cycle.

src/chainstate/coordinator/mod.rs
pavitthrap: It means that the parent and child block have the same reward cycle number.

src/chainstate/coordinator/mod.rs
pavitthrap: Just to confirm I know what you are suggesting, let's say the range of possible exit proposals is 15 to 30, and the invalid reward cycles are (18, 21). 
Would I have a slice of length 15, that looks like [Some(18), Some(21), None, None, None, None, ...]?

pavitthrap: @jcnelson

src/chainstate/burn/db/sortdb.rs
pavitthrap: The logic right before this inserts a row with that block_id, and it should fail if the insert failed? 

Do you think an assertion is still necessary here?

src/chainstate/coordinator/mod.rs
pavitthrap: Let's discuss this in the architecture meeting

jcnelson: I can't come this week :/  Please let me know what y'all talk about

src/chainstate/burn/db/sortdb.rs
```diff
@@ -3699,6 +3786,52 @@ impl<'a> SortitionHandleTx<'a> {
...
+        }
+        if let Some(exit_at_reward_cycle) = exit_at_reward_cycle_info.curr_exit_at_reward_cycle {
+            let sql = "UPDATE exit_at_reward_cycle_info SET curr_exit_at_reward_cycle = ? WHERE block_id = ?";
+            let args: &[&dyn ToSql] = &[
+                &u64_to_sql(exit_at_reward_cycle)?,
+                &exit_at_reward_cycle_info.block_id,
+            ];
+            self.execute(sql, args)?;
+        }
+        Ok(())
```
kantai, line 3820: I think this method could be simplified to one SQL query:

```suggestion
        let exit_proposal = exit_at_reward_cycle_info.curr_exit_proposal.map(u64_to_sql)?;
        let exit_at_reward_cycle = exit_at_reward_cycle_info.curr_exit_at_reward_cycle.map(u64_to_sql)?;
        
        let sql = "INSERT or REPLACE INTO exit_at_reward_cycle_info (block_id, parent_block_id, block_reward_cycle, stacks_block_height_in_cycle, invalid_reward_cycles, curr_exit_proposal, curr_exit_at_reward_cycle) VALUES (?, ?, ?, ?, ?, ?, ?)";
        let args: &[&dyn ToSql] = &[
            &exit_at_reward_cycle_info.block_id,
            &exit_at_reward_cycle_info.parent_block_id,
            &u64_to_sql(exit_at_reward_cycle_info.block_reward_cycle)?,
            &u64_to_sql(exit_at_reward_cycle_info.stacks_block_height_in_cycle)?,
            &serde_json::to_string(&exit_at_reward_cycle_info.invalid_reward_cycles).unwrap(),
            &exit_proposal,
            &exit_at_reward_cycle
        ];
        self.execute(sql, args)
```

src/chainstate/stacks/boot/mod.rs
```diff
@@ -68,17 +74,26 @@ lazy_static! {
...
         format!("{}\n{}", BOOT_CODE_POX_TESTNET_CONSTS, BOOT_CODE_POX_BODY);
+    // TODO (#3034) - replace when exit contract is published
+    static ref BOOT_CODE_EXIT_AT_RC_MAINNET: String =
+        format!("{}\n{}", BOOT_CODE_EXIT_AT_RC_MAINNET_CONSTS, BOOT_CODE_EXIT_AT_RC);
+    pub static ref BOOT_CODE_EXIT_AT_RC_TESTNET: String =
+        format!("{}\n{}", BOOT_CODE_EXIT_AT_RC_TESTNET_CONSTS, BOOT_CODE_EXIT_AT_RC);
+
     pub static ref BOOT_CODE_COST_VOTING_TESTNET: String = make_testnet_cost_voting();
-    pub static ref STACKS_BOOT_CODE_MAINNET: [(&'static str, &'static str); 6] = [
+    // TODO (#3034): remove exit-at-rc once deployed
```
kantai, line 84: I think that, because even once `exit-at-rc` is deployed we'll need to keep integration tests around that reference an instantiated contract, it may be better to pay the implementation cost of making this more configurable now. I'd suggest something like the following:

1. Make the `QualifiedContractIdentifier` corresponding to the `exit-at-rc` contract configurable via the `exit_at_constants` struct. The code in the coordinator can use that identifier to perform its checks (it will need to check for the existence of the contract).
2. For the "unit tests" (i.e., the tests within the `/src/` directory), I think it'll be easy to instantiate the contract synthetically (basically, the way it's done now, but without including the contract specifically in the `STACKS_BOOT_CODE_TESTNET` array).
3. For the "integration tests" (i.e., the ones in `neon_integrations.rs`), it should be possible to instantiate the contract using a normal contract publish transaction.

jcnelson:
Open design items:

* Can stacking pools vote on behalf of the users?  Must users delegate voting rights to a contract?
    * Answer: don't support this right now, but see what the SIP process tells us.  We can always change the contract to meet the SIP's needs.
    
* Do we try and do the vote tabulation in the same block as the PoX anchor block selection?  Or in a different block?
    * Yes -- do the vote tabulation in the same block as PoX anchor block selection 

* Is the veto threshold 80% of the reward cycle length in the burnchain?  Or is it fork-aware?
    * Yes -- it's X% of the reward cycle length in the burnchain (i.e. fork-ignorant), but set X to 60%.

* The contract should *not* be a boot contract; it should live at a designated address, which will be set in the SIP process.  The contract ID should be a burnchain parameter (i.e. `Option<QualifiedContractIdentifier>`).

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +708,281 @@ impl<
...
+        Ok((vote_map, total_votes))
+    }
+
+    /// At the end of each reward cycle, we tally the votes for the exit at RC contract.
+    /// We need to read PoX contract state to see how much STX is stacked into the protocol - we then
+    /// ensure that at least 50% of staked STX has a valid vote.
+    /// Regarding vote validity: we discard votes for invalid reward cycles. Examples of invalid
+    /// reward cycles include those that are the absolute minimum exit cycle, or those below a
+    /// previously confirmed exit RC.
+    pub fn tally_votes(
```
kantai, line 868: Can you add the return information to the rustdoc?

src/chainstate/stacks/boot/exit-at-rc.clar
kantai: I don't think this prevents pools from voting. A pool can still vote by using `(as-contract ..)` to invoke this contract's methods. The bigger issue for pools' participation is that they won't be able to invoke these methods "as the contract" because the pool contracts are already all published, and they don't have methods for invoking the exit at contract.

src/chainstate/stacks/boot/exit-at-rc.clar
kantai: read-only methods are also callable by other contracts, so the initial values for these variables should be settable by doing something like `(contract-call? .pox get-pox-info)`

gregorycoppola:
Dropped myself as a reviewer because jcnelson and kantai will have to approve this central PR anyway.

kantai (commented):
This overall looks great to me. I think before merging, though, a couple issues will need to be resolved.

1. As written, this will cause a testnet reset, because the chainstate setup code switches on `mainnet` to determine whether or not to instantiate a test exit at block height contract. Maybe that's okay, but we should at a minimum call that out. If not, then, I think this needs to be changed so that the exit at block height contract isn't launched as a genesis contract ever.
2. There's som `info!` statements in `coordinators/mod.rs` that need to be cleaned out.
3. We should see if the exit at block height contract can actually fetch some of its constant data from the pox contract on initialization.

src/chainstate/stacks/boot/exit-at-rc.clar
kantai: If these are settable via the pox get-pox-info call, then you should be able to switch them to constants (rather than data vars), which would simplify some of the contract code.

src/burnchains/mod.rs
```diff
@@ -320,6 +326,48 @@ pub struct PoxConstants {
...
+pub struct ExitContractConstants {
+    // The chosen exit reward cycle must have this percent of stackers that voted for it or for a lower reward cycle to be confirmed
```
jcnelson, line 331: Minor nit -- I think you confused the "veto" and "vote" docstrings here.

src/burnchains/mod.rs
```diff
@@ -320,6 +326,48 @@ pub struct PoxConstants {
...
+            veto_confirmation_percent,
+            vote_confirmation_percent,
+            percent_stacked_stx_for_valid_vote,
+            absolute_minimum_exit_rc: EXIT_RC_ABSOLUTE_MINIMUM_EXIT_RC,
+            minimum_rc_buffer_from_present: EXIT_RC_MINIMUM_RC_BUFFER_FROM_PRESENT,
+            maximum_rc_buffer_from_present: EXIT_RC_MAXIMUM_RC_BUFFER_FROM_PRESENT,
+        }
+    }
+
+    pub fn mainnet_default() -> ExitContractConstants {
```
jcnelson, line 362: I just want to confirm this, because "veto" and "vote" have a very similar spelling.  For mainnet, the _veto_ confirmation percent is 60% -- as in, 60% of stackers must reject the upgrade?  And, the _vote_ confirmation percent is 50% -- as in, 50% of the stackers must support the upgrade?

I'd recommend using different words than "veto" and "vote" so we don't accidentally introduce a consensus bug because of a typo.  Maybe "support" and "reject" instead?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -2494,23 +2550,29 @@ impl SortitionDB {
...
-                if version == expected_version {
-                    return Ok(());
-                }
-                if version == "1" {
-                    SortitionDB::apply_schema_2(&tx, epochs)?;
-                    tx.commit()?;
-                    Ok(())
-                } else {
-                    panic!("The schema version of the sortition DB is invalid.")
+        let mut updated = false;
```
jcnelson, line 2553: To future-proof this a bit better, instead of using an `updated` flag, I think it would be safer to do the following:

```rust
let mut cur_version = "1".to_string();
while cur_version != SORTITION_DB_VERSION {
   match SortitionDB::get_schema_version(&tx) {
      Ok(Some(version)) => {
         cur_version = version;
         
         // match on cur_version to decide what schema to apply
         if cur_version == "1" { /* etc. */ }
         else if _cur_version == "2" { /* etc. */ }
      }
      /* other cases */
   }
}
```

src/chainstate/coordinator/mod.rs
jcnelson: Yes, but on a subsequent pass, I'm not sure that the precondition ("this list is bound in length") is met in the first place.  If that's the case, then what you have here is fine.

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +708,281 @@ impl<
...
+        // If there is an existing exit rc, make sure the minimum rc we consider for the votes is
+        // greater than it.
+        if let Some(curr_exit_at_rc) = curr_exit_at_rc_opt {
+            min_rc = min_rc.max(curr_exit_at_rc + 1);
+        }
+
+        let mut total_votes = 0;
+        let invalid_reward_cycles: HashSet<u64> =
+            HashSet::from_iter(invalid_reward_cycles.into_iter());
+        let is_mainnet = self.burnchain.is_mainnet();
```
jcnelson, line 814: This method is broken, but we can't fix it because it's consensus-critical.  `Burnchain::is_mainnet()` always returns `false`, since it compares the Stacks network ID to the Bitcoin network ID (oops).

jcnelson: Can you add a docstring to `Burnchain::is_mainnet()` to say something to the effect of `"BROKEN! DO NOT USE IN NEW PLACES!"

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +708,281 @@ impl<
...
+                // If the accrued votes is greater than the minimum needed to achieve consensus,
+                // store this value for the upcoming veto
+                return Ok(Some(*curr_rc_proposal));
+            }
+        }
+        Ok(None)
+    }
+
+    pub fn reached_exit_reward_cycle(&self, header: &BurnchainBlockHeader) -> Result<bool, Error> {
+        if let Some(chain_tip) = self.canonical_chain_tip {
```
jcnelson, line 950: It might be safer (more defensive) to just load the canonical chain tip from the DB here, lest we accidentally not update it in some other code path in time for this call to work.

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +708,281 @@ impl<
...
+            let exit_info_opt =
```
jcnelson, line 951: This is only `Some(..)` once per reward cycle, right?

src/chainstate/coordinator/mod.rs
```diff
@@ -683,6 +1008,105 @@ impl<
...
     }
 
+    fn process_exit_reward_cycle(
+        &mut self,
+        block_receipt: &StacksEpochReceipt,
+        block_hash: &BlockHeaderHash,
+        canonical_sortition_tip: &SortitionId,
+        exit_contract_id: &QualifiedContractIdentifier,
+    ) -> Result<(), Error> {
+        let current_block_height = block_receipt.header.burn_header_height as u64;
```
jcnelson, line 1018: Specifically, current _burn_ block height.

src/chainstate/coordinator/mod.rs
```diff
@@ -683,6 +1008,105 @@ impl<
...
+            }
+            invalid_reward_cycles = parent_exit_info
+                .invalid_reward_cycles
+                .iter()
+                .filter(|rc| **rc > current_reward_cycle)
+                .map(|rc| *rc)
+                .collect();
+
+            // We process the previous reward cycle as soon as we transition to the next one.
+            if parent_exit_info.block_reward_cycle < current_reward_cycle {
```
jcnelson, line 1051: I'm not sure it matters here, but maybe the comments should speak to it.  What happens if the parent block is more than one reward cycle in the past?

src/chainstate/coordinator/mod.rs
```diff
@@ -683,6 +1008,105 @@ impl<
...
+                        parent_exit_info.curr_exit_at_reward_cycle,
+                        parent_exit_info.invalid_reward_cycles,
+                        &block_hash,
+                        &block_receipt.header.consensus_hash,
+                        exit_contract_id,
+                    )?;
+                    if let Some(exit_proposal) = current_proposal {
+                        info!("RCPR: storing proposal info for {:?}", exit_proposal);
+                        // record the proposal in invalid_reward_cycles
+                        invalid_reward_cycles.push(exit_proposal);
```
jcnelson, line 1081: I think the name is throwing my understanding off.  This path is executed if the call to `.tally_votes()` indicated that since the parent block was mined, a proposed reward cycle has been voted upon (and may now be vetoed).  But why does that make the exit proposal "invalid"?  Does it just mean that we don't have to tally it again?

src/chainstate/coordinator/mod.rs
```diff
@@ -683,6 +1008,105 @@ impl<
         )
     }
 
+    fn process_exit_reward_cycle(
```
jcnelson, line 1011: Can you add a docstring for this method?

src/chainstate/coordinator/mod.rs
```diff
@@ -683,6 +1008,105 @@ impl<
...
+                    )?;
+                    if let Some(exit_proposal) = current_proposal {
+                        info!("RCPR: storing proposal info for {:?}", exit_proposal);
+                        // record the proposal in invalid_reward_cycles
+                        invalid_reward_cycles.push(exit_proposal);
+                    }
+                }
+            }
+        } else {
+            warn!(
```
jcnelson, line 1086: This is bound to happen at least once, right?  When this PR goes live on existing chainstate, no block is going to have any reward cycle exit info.

src/chainstate/coordinator/mod.rs
```diff
@@ -683,6 +1008,105 @@ impl<
...
+                .filter(|rc| **rc > current_reward_cycle)
+                .map(|rc| *rc)
+                .collect();
+
+            // We process the previous reward cycle as soon as we transition to the next one.
+            if parent_exit_info.block_reward_cycle < current_reward_cycle {
+                info!("RCPR: cc: ancestor - rc is greater than parent rc");
+                // Check if there is some proposal. If so, need to check for a veto.
+                if let Some(curr_exit_proposal) = parent_exit_info.curr_exit_proposal {
+                    let veto_passed = self.read_veto_state(
```
jcnelson, line 1055: To be clear, we only check the veto status for a reward cycle _once_, right?  i.e. only once a block gets mined whose parent is in a subsequent reward cycle?  And, we're checking to see if the parent's proposed exit reward cycle got vetoed?

src/burnchains/mod.rs
pavitthrap: Great point - the spelling is way too similar. I'll update it now.

pavitthrap: And yes, your interpretation of the numbers is correct.

src/chainstate/coordinator/mod.rs
pavitthrap: Should I make a new is_mainnet_function (like `is_mainnet_v2`) and use that instead. In that function I would compare the Stacks network ID against `BITCOIN_NETWORK_ID_MAINNET`.

src/chainstate/coordinator/mod.rs
pavitthrap: No, I store a `BlockExitRewardCycleInfo` for every block in the Stacks chain.

src/chainstate/coordinator/mod.rs
pavitthrap: It indicates that this particular reward cycle can no longer be voted for. For example, if the proposal to exit at this reward cycle is veto'd, it is not possible to propose this same exit reward cycle again.

src/chainstate/coordinator/mod.rs
pavitthrap: Yes, this is bound to happen once. Is that acceptable? I mentioned this in the PR description as something I was unsure about.

src/chainstate/coordinator/mod.rs
pavitthrap: Updated the logic so that I only create this object once per reward cycle.

src/chainstate/coordinator/mod.rs
pavitthrap: Yes - let's discuss this in the arch meeting

jcnelson:
Todo from architecture meeting:

* Add to testnet checklist to instantiate this contract on testnet, and make sure the address is correct in the code

* Consider deleting `Burnchain::is_mainnet()` (in a separate PR)

* Need a good solution for deducing the exit reward cycle when starting a node from existing chainstate in the middle of a vote (in a separate PR)

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +704,287 @@ impl<
...
+        block_hash: &BlockHeaderHash,
+        exit_contract_id: &QualifiedContractIdentifier,
+    ) -> Result<bool, Error> {
+        let stacks_block_id = StacksBlockHeader::make_index_block_hash(consensus_hash, block_hash);
+        let reward_cycle_length = self.burnchain.pox_constants.reward_cycle_length;
+        let rejection_percent_threshold = self
+            .burnchain
+            .exit_contract_constants
+            .rejection_confirmation_percent;
+        let is_mainnet = self.burnchain.is_mainnet();
```
kantai, line 723: It doesn't look like `is_mainnet` is ever used in this method -- can be removed.

src/chainstate/coordinator/mod.rs
kantai: It doesn't look like this value is ever used in this method -- I think just remove this variable.

kantai:
This looks good to me, I think the only todos are a few superficial changes I noted, and then a larger change to make it so that the contract isn't a boot contract in testnet. That will require changing some of the tests, but that should be doable by just:

1. Adding a contract publish transaction to the tests from some address that has funds in the tests.
2. Removing the contract from the testnet burnchain config.

I don't think we need to have a contract address for the testnet burnchain config in this PR: a contract can be launched to testnet after this PR merges, and a subsequent PR would set that contract in the config.

src/chainstate/coordinator/mod.rs
```diff
@@ -658,6 +704,287 @@ impl<
...
+                            "rc-proposal-rejections",
+                            &Value::from(
+                                TupleData::from_data(vec![
+                                    ("proposed-rc".into(), Value::UInt(proposed_exit_rc as u128)),
+                                    ("curr-rc".into(), Value::UInt(rc_cycle_of_rejection as u128)),
+                                ])
+                                .expect("BUG: failed to construct simple tuple"),
+                            ),
+                        )
+                        .expect("BUG: Failed querying rc-proposal-rejections")
```
kantai, line 740: Hmm. What happens if this is evaluated before the contract is launched? Does it panic?

PR #3037, opened 2022-02-08
Comment network code + small fix

pavitthrap:
This PR does two things:
- I added comments to some modules I read over 
- Added better error handling in `BatchedDNSLookupsState`. In `try_proceed`, if a call to `poll_request` succeeds, I removed the query url from the list of requests to avoid unnecessary future queries for that same url.

Did not add tests since code change was minimal.


gregorycoppola:
The comments changes are really useful, I think, because I learned a lot by reading them.

You might try separating the comments PR from the code changes, since they're unrelated and have different testing models (comments don't need test, but the other part might).

jcnelson (approved):
LGTM.  Love the new docstrings!

kantai (approved):

Issue #3042, opened 2022-02-10
[chainstate] implement cache strategies for nodes in tries that are persisted to disk


jcnelson:
Once a trie is stored to the MARF, it will never be modified.  We should be able to fearlessly cache any/all of its nodes in RAM.  The trie storage system should be updated to support a variety of caching strategies, which the user can choose in the config file (i.e. to trade RAM for speed).  At a minimum, we should support the following:

* No-op (the default; what happens now)
* Everything (keep all nodes in RAM -- useful for miners)
* Node256 (keep `TrieNode256`'s in RAM, but nothing else)

I've tried all of the above.  `Node256` offers decent performance without eating your RAM, for example.

gregorycoppola:
Related: This issue is the line of work suggested in #3014, which suggests looking for ways to cache the MARF in RAM to improve performance.

Issue #3041, opened 2022-02-10
[chainstate] defer trie root hash calculation until no more writes will occur on the trie


jcnelson:
Right now, writing nodes to a trie in the MARF consumes *in aggregate* a lot of time, because each time a leaf is inserted, the trie root hash is recalculated.  This is unnecessary; the root hash only needs to be calculated once we're done writing leaf nodes.  To implement this, the MARF needs a `seal()` operation that (a) calculates the trie root hash and the MARF root hash in the uncommitted state and stores it there, and (b) prevents subsequent writes from happening on the uncommitted state.  Reads and transaction abort/commit will continue to be permitted.

Issue #3014, opened 2022-01-24
Profile and Optimize Mining Code


gregorycoppola:
https://github.com/stacks-network/stacks-blockchain/issues/3012 discusses how blocks often have few tx's in them.

This is a result of two things:
1) the timeout times to stop processing
2) the speed of tx processing

I have measured processing speed on my GCP box as 

```
INFO [1643037789.034557] [src/chainstate/stacks/miner.rs:1934] [main] num_considered 269 seconds_elapsed 183.046 fraction 0.6804684014869888 
```

This is about `1.47` transactions per second. This means that, to put 1k tx's in a block, it would take `680s=11minutes`. This means that it will be hard to put 1k tx's in most blocks.

Maybe we can improve processing speed, to drastically fill more blocks. Eg. we can pre-cache the MARF look-ups that will be needed for a given set of mempool tx's before mining begins.

gregorycoppola:
![profile002](https://user-images.githubusercontent.com/60008382/150841706-0c2ca38c-fbd9-4eb0-828d-1e7ab5d040b2.svg)

Here is a profile I made.

Looks like `query_row_and_then` takes up most of the time, but not sure if the graph is showing the relationships right, in that it looks like these calls are outside of `process_transaction`, but that might not be true.

jcnelson:
Closing in favor of your other issues related to improving the miner performance

PR #3044, opened 2022-02-10
MARF Performance Improvements 

### LINKED ISSUE #3059: [chainstate] store large MARF tries in a separate flat file
Once a blob gets to be bigger than the sqlite page size (which defaults to 4096), loading data from a blob starts to become expensive.  Per sqlite's own benchmarks [1], once the blob size exceeds 100kb, reading from a file becomes faster no matter what page size is used.  This suggests to me that the `TrieFileStorage` system should maintain a separate flat file for storing tries that exceed 100kb.  This is most of them for bigger blocks -- as of right now, 31,428 out of 54,400 tries in the `vm/clarity/marf.sqlite` file (almost 60%) exceed 100kb.  The sqlite db would instead store an offset and length in this file of the trie blob, and provide a `Read + Seek`-implementing struct for accessing it.

[1] https://www.sqlite.org/intern-v-extern-blob.html

jcnelson:
This PR significantly improves the read and write paths for MARFs by between a factor of 10x and 200x, depending on how many paths are inserted into each trie.  It fixes #3059, #3042, #3041, and does some work towards fixing #3014.

On the write path, this code adds the ability to _defer_ calculating a trie's node hashes until when the trie is ready to be dumped to disk.  This leads to a significant performance improvement because it means that an intermediate node's hash is only ever calculated at most once, whereas before, it could be calculated up to as many times as there were leaf inserts.  Over 90% of the write path wall-clock time was spent in calculating node hashes that would later be overwritten by subsequent hash calculations.

On the read path, this code does two things:

* It adds a caching layer for nodes, so that miners and node operators with lots of RAM can opt to cache some or all of the trie nodes in RAM.  The current cache strategies (to be expanded in later PRs) are `noop` (default; does nothing), `everything` (caches all nodes in RAM once they are loaded), and `node256` (caches all `TrieNode256` nodes in RAM once loaded).  The new cache layer also subsumes the caches for holding block hashes.

* It allows the MARF owner to opt to store tries in a separate flat file, called a `.blobs` file.  Before, the tries would be stored as sqlite blobs in the database.  However, for large tries, such as those produced by the Clarity VM, the blob paging logic in sqlite significantly degraded performance on the read path when walking tries.  Storing large tries in a flat file (i.e. tries bigger than one page size) removes sqlite from the trie node read path altogether, and yields up to a 14x improvement.  In addition, this PR sets the sqlite page size to 32k (we were using the default of 4k), which is big enough to hold other MARF tries from the sortition DB and headers DB in a single sqlite page.

In addition, this PR introduces a simple profiler for measuring how long certain hot paths take.  New measurements can be incrementally added in.

Without this PR, a node in `master` takes about 3 days to boot from genesis.  With this PR, it takes about 28 hours.

Don't let the diff line count intimidate you.  A lot of it is from indenting the unit test code needing to be indented by one tab in order to try each unit test with different `MARFOpenOpt` values.


gregorycoppola:
Nice result. Will look. Thanks for sending.

Is there a command line to run the experiment?

gregorycoppola:
@jcnelson 

I know how to run an experiment, I realized. You don't have to worry about sending it on any time frame.

Curious to know how you did it but I can also use the "mempool analyzer". (Except I think my current mempool state is not up to date.)

jcnelson:
I've made some more (uncommitted) optimizations to the read path, but I think we're at the point now where sqlite's blob storage is the limiting factor.  I'd need to test this out, but per sqlite's documentation [1], blobs bigger than 100k tend to be slower to read than the filesystem.  MARF tries in recent history are over 10x this.

jcnelson:
Move MARF tests into their own files.

src/chainstate/stacks/index/trie_sql.rs
```diff
@@ -93,6 +114,63 @@ pub fn create_tables_if_needed(conn: &mut Connection) -> Result<(), Error> {
...
+        Err(e) => {
+            debug!("Failed to get schema version: {:?}", &e);
+            1u64
+        }
+    }
+}
+
+/// Migrate the MARF database to the currently-supported schema.
+/// Returns the version of the DB prior to the migration.
+pub fn migrate_tables_if_needed<T: MarfTrieId>(
```
gregorycoppola, line 131: how long does this take?

src/chainstate/stacks/index/cache.rs
```diff
@@ -0,0 +1,880 @@
...
+    /// Load a block's hash, given its block ID.
+    pub fn load_block_hash(&mut self, block_id: u32) -> Option<T> {
+        match self {
+            TrieCache::Noop(ref mut state) => state.load_block_hash(block_id),
+            TrieCache::Everything(ref mut state) => state.load_block_hash(block_id),
+            TrieCache::Node256(ref mut state) => state.load_block_hash(block_id),
+        }
+    }
+
+    /// Store a block's ID and hash to teh cache.
```
gregorycoppola, line 314: s/teh/the

src/chainstate/stacks/index/cache.rs
```diff
@@ -0,0 +1,880 @@
...
+    pub fn store_block_hash(&mut self, block_id: u32, block_hash: T) {
+        match self {
+            TrieCache::Noop(ref mut state) => {
```
gregorycoppola, line 317: can this disjunction be collapsed? seems like the same outcome in any case

src/chainstate/stacks/index/cache.rs
```diff
@@ -0,0 +1,880 @@
...
+    pub fn load_block_id(&self, block_hash: &T) -> Option<u32> {
+        match self {
+            TrieCache::Noop(ref state) => state.load_block_id(block_hash),
+            TrieCache::Everything(ref state) => state.load_block_id(block_hash),
+            TrieCache::Node256(ref state) => state.load_block_id(block_hash),
+        }
+    }
+}
+
+#[cfg(test)]
```
gregorycoppola, line 348: thought we were moving "test" code to special `test` files?

src/chainstate/stacks/index/cache.rs
```diff
@@ -0,0 +1,880 @@
...
+    /// Get the block ID, given its hash
+    pub fn load_block_id(&self, block_hash: &T) -> Option<u32> {
+        self.block_id_cache.get(block_hash).map(|id| *id)
+    }
+}
+
+/// Trie node cache strategies
+pub enum TrieCache<T: MarfTrieId> {
+    /// Do nothing
+    Noop(TrieCacheState<T>),
```
gregorycoppola, line 176: "Nothing"?

gregorycoppola (commented):
Code makes sense. Thanks a lot!

So, for your live test mock mining against the NFT launch... if there is a problem, is it going to loudly fail or subtly fail?

src/chainstate/stacks/index/cache.rs
jcnelson: That applies to `#[test]`-decorated functions.

src/chainstate/stacks/index/cache.rs
jcnelson: Yup. Don't do anything at all.  It's a no-op strategy.

src/chainstate/stacks/index/marf.rs
```diff
@@ -68,6 +68,57 @@ struct WriteChainTip<T> {
     height: u32,
 }
 
+/// Options for opening a MARF
+#[derive(Clone, Debug)]
+pub struct MARFOpenOpts {
```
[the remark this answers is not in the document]
jcnelson: The various MARFs in the system use a pre-determined instantiation of this structure.

src/chainstate/stacks/index/trie_sql.rs
jcnelson: Just over 15 minutes for migrating all MARFs.

gregorycoppola:
@jcnelson The code and unit tests LGTM.

Curious to hear any data from your live test. But, if you think the tests were successful and it looks good I'm ready to approve.

jcnelson:
I'm gathering data with the mock miner right now during the ongoing NFT mint.  Will report back when it's done.

jcnelson:
So, I measured the number of anchored block transactions the mock-miner produces with this PR, and compared the _maximum-sized_ anchored block (by tx count) to the anchored block that was actually mined in that sortition.  I did this for Bitcoin blocks 728,420 through 729,326.  I computed the _difference_ in transaction counts, and plotted them here:

Difference between master and this PR

It's pretty noisy, but the median difference is +7 and the average is +14 (std.dev. +/- 50) -- meaning, this PR includes 7 more transactions than `master` in the best case and 14 more transactions than `master` on average.

This is an improvement overall, but it's hard to say by how much given the data.  I didn't look at how many transactions were included in the microblock streams in `master` and this PR, for example, since it was difficult to extract this from the node.  I also don't know to what extent my node was missing transactions and microblocks that a winning miner had, and vice versa.

src/chainstate/stacks/index/bits.rs
```diff
@@ -233,8 +224,9 @@ pub fn get_nodetype_hash_bytes<T: MarfTrieId, M: BlockMap>(
 
 /// Low-level method for reading a TrieHash into a byte buffer from a Read-able and Seek-able struct.
 /// The byte buffer must have sufficient space to hold the hash, or this program panics.
+#[inline]
```
kantai, line 227: Are all these `#[inline]` directives helpful? It's usually better to let the compiler make a decision on inlining (it's actually going to end up making the decision anyways), as the trade-off isn't always obvious. And unless it's really clear that these are helpful, I think it's better practice to leave these kinds of directives out.

src/chainstate/stacks/index/bits.rs
```diff
@@ -294,13 +289,26 @@ pub fn count_children(children: &[TriePtr]) -> usize {
...
+    read_nodetype_at_head(f, ptr.id(), true)
+}
+
+/// Read a node
+#[inline]
+pub fn read_nodetype_nohash<F: Read + Seek>(
+    f: &mut F,
+    ptr: &TriePtr,
+) -> Result<TrieNodeType, Error> {
+    fseek(f, ptr.ptr() as u64)?;
```
kantai, line 309: Is using custom `fseek` functions much easier than just using the native method?

```suggestion
    f.seek(SeekFrom::Start(ptr.ptr() as u64))?;
```

src/chainstate/stacks/index/bits.rs
```diff
@@ -310,12 +318,21 @@ pub fn read_nodetype<F: Read + Seek>(
 ///   node hash      id  ptrs & ptr data      path
 ///
 /// X is fixed and determined by the TrieNodeType variant.
-/// Y is variable, but no more than TriePath::len()
-pub fn read_nodetype_at_head<F: Read>(
+/// Y is variable, but no more than TriePath::len().
+///
+/// If `read_hash` is false, then the contents of the node hash are undefined.
+#[inline]
+pub fn read_nodetype_at_head<F: Read + Seek>(
```
kantai, line 325: This interface seems more dangerous than necessary.

Could we have two method variants instead:

```
fn read_nodetype_at_head<F: Read + Seek>(f: &mut F, ptr_id: u8) -> Result<(TrieNodeType, TrieHash), Error>;
fn read_nodetype_at_head_no_hash<F: Read + Seek>(f: &mut F, ptr_id: u8) -> Result<TrieNodeType, Error>;
```

And maybe these two methods are just wrappers around a (non-public) inner implementation?

That way, there's never a method returning data that shouldn't be read.

src/chainstate/stacks/index/cache.rs
```diff
@@ -0,0 +1,877 @@
...
+use util_lib::db::tx_busy_handler;
+use util_lib::db::Error as db_error;
+use util_lib::db::SQLITE_MMAP_SIZE;
+
+use stacks_common::types::chainstate::BlockHeaderHash;
+use stacks_common::types::chainstate::BLOCK_HEADER_HASH_ENCODED_SIZE;
+use stacks_common::types::chainstate::{TrieHash, TRIEHASH_ENCODED_SIZE};
+
+/// Fully-qualified address of a Trie node.  Includes both the block ID and the pointer within the
+/// block's blob as to where it is stored.
```
kantai, line 63: The Block ID is the `u32`? There's a concept of a `MarfTrieId` in the codebase, but that's a `[u8; 32]`, so what is the `block ID`?

src/chainstate/stacks/index/cache.rs
```diff
@@ -0,0 +1,877 @@
...
+    pub fn load_node_and_hash(
+        &self,
+        block_id: u32,
+        trieptr: &TriePtr,
+    ) -> Option<(TrieNodeType, TrieHash)> {
+        match (
+            self.load_node(block_id, trieptr),
+            self.load_node_hash(block_id, trieptr),
+        ) {
+            (Some(ref node), Some(ref hash)) => Some(((*node).clone(), (*hash).clone())),
```
kantai, line 103: This is a double clone, right? `self.load_node()` and `self.load_node_hash()` do one clone, and this method does another clone.

You can avoid double clones by returning references, which would also let callers that don't need ownership to avoid a clone altogether.

src/chainstate/stacks/index/cache.rs
```diff
@@ -0,0 +1,877 @@
...
+
+    /// Obtain a possibly-cached node
+    pub fn load_node(&self, block_id: u32, trieptr: &TriePtr) -> Option<TrieNodeType> {
+        match self
+            .node_cache
+            .get(&TrieNodeAddr(block_id, trieptr.clone()))
+        {
+            Some(ref node) => Some((*node).clone()),
+            None => None,
+        }
```
kantai, line 116:
```suggestion
        match self
            .node_cache
            .get(&TrieNodeAddr(block_id, trieptr.clone()))
            .cloned()
```

src/chainstate/stacks/index/cache.rs
```diff
@@ -0,0 +1,877 @@
...
+        hash: TrieHash,
+    ) {
+        self.store_node(block_id, trieptr.clone(), node);
+        self.store_node_hash(block_id, trieptr, hash)
+    }
+
+    /// Cache just a node
+    pub fn store_node(&mut self, block_id: u32, trieptr: TriePtr, node: TrieNodeType) {
+        self.node_cache
+            .insert(TrieNodeAddr(block_id, trieptr.clone()), node);
```
kantai, line 141: Seems like a spurious clone

src/chainstate/stacks/index/cache.rs
kantai: Probably more descriptive as `/// Cache nothing`

src/chainstate/stacks/index/cache.rs
kantai: I agree with moving this test code -- this file is over 50% tests. Moving large test sets away from the code will make the code easier to navigate.

src/chainstate/stacks/index/file.rs
```diff
@@ -0,0 +1,480 @@
...
+}
+
+/// Handle to a flat in-memory buffer containing Trie blobs (used for testing)
+pub struct TrieFileRAM {
+    fd: Cursor<Vec<u8>>,
+    readonly: bool,
+    trie_offsets: TrieIdOffsets,
+}
+
+pub enum TrieFile {
```
kantai, line 83: Does a TrieFile contain multiple tries? I think a rustdoc for the `TrieFile` would be useful that explained what exactly a `TrieFile` contains (does it also store metadata?), and some brief explanation of how it is used by the rest of the marf implemenation.

src/chainstate/stacks/index/file.rs
```diff
@@ -0,0 +1,480 @@
...
+                    let trie_blob = TrieFile::read_trie_blob_from_db(db, block_id)?;
+
+                    // get the block ID
+                    let bhh: T = trie_sql::get_block_hash(db, block_id)?;
+
+                    // append the blob, replacing the current trie blob
+                    info!(
+                        "Migrate block {} ({} of {}) to external blob storage",
+                        &bhh, block_id, max_block
+                    );
```
kantai, line 217: This should maybe only occur on `block_id % 1000 == 0` or something like that -- otherwise the migration will be 55k log lines.

src/chainstate/stacks/index/marf.rs
```diff
@@ -68,6 +68,57 @@ struct WriteChainTip<T> {
...
+    pub external_blobs: bool,
+    /// unconditionally do a DB migration (used for testing)
+    pub force_db_migrate: bool,
+}
+
+impl MARFOpenOpts {
+    pub fn default() -> MARFOpenOpts {
+        MARFOpenOpts {
+            hash_calculation_mode: TrieHashCalculationMode::Deferred,
+            cache_strategy: "noop".to_string(),
```
kantai, line 88: Does the `default` method in the cache strategy use a different default? That's reading an environment variable.

src/chainstate/stacks/index/marf.rs
```diff
@@ -548,7 +608,15 @@ impl<T: MarfTrieId> MARF<T> {
...
 
+    #[cfg(test)]
+    pub fn seal(&mut self) -> Result<TrieHash, Error> {
+        let mut tx = self.begin_tx()?;
+        let h = tx.seal()?;
+        Ok(h)
+    }
+
     // helper method for walking a node's backpr
+    #[inline]
```
kantai, line 619: This is a pretty large method, should we really push for the compiler to inline it?

src/chainstate/stacks/index/profile.rs
```diff
@@ -0,0 +1,409 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+use std::time::SystemTime;
+
+/// Fine-grained profiling data for Trie storage ops
+#[derive(Debug, Clone)]
+pub struct TrieBenchmark {
```
kantai, line 21: Is there a no-op version of this that is used in production, or will all this benchmarking code be invoked in all release builds?

src/chainstate/stacks/index/storage.rs
```diff
@@ -83,12 +98,34 @@ impl<T: MarfTrieId> BlockMap for TrieFileStorage<T> {
     }
 
     fn get_block_hash_caching(&mut self, id: u32) -> Result<&T, Error> {
-        if !self.data.block_hash_cache.contains_key(&id) {
-            self.data
-                .block_hash_cache
-                .insert(id, self.get_block_hash(id)?);
+        if self.cache.load_block_hash(id).is_none() {
```
kantai, line 101:
```suggestion
        if self.is_block_hash_cached(id) {
```

src/chainstate/stacks/index/storage.rs
```diff
@@ -100,12 +137,34 @@ impl<'a, T: MarfTrieId> BlockMap for TrieStorageConnection<'a, T> {
...
```
kantai, line 140:
```suggestion
        if self.is_block_hash_cached(id) {
```

src/chainstate/stacks/index/storage.rs
```diff
@@ -129,10 +200,34 @@ impl<T: MarfTrieId> BlockMap for TrieSqlHashMapCursor<'_, T> {
     }
 
     fn get_block_hash_caching(&mut self, id: u32) -> Result<&T, Error> {
-        if !self.cache.contains_key(&id) {
-            self.cache.insert(id, self.get_block_hash(id)?);
+        if self.cache.load_block_hash(id).is_none() {
```
kantai, line 203:
```suggestion
        if self.is_block_hash_cached(id) {
```

src/chainstate/stacks/index/storage.rs
```diff
@@ -321,8 +632,221 @@ impl<T: MarfTrieId> TrieRAM<T> {
...
+        storage_tx: &mut TrieStorageTransaction<T>,
+    ) -> Result<TrieHash, Error> {
+        // find trie root hash
+        debug!("Calculate trie root hash");
+        let root_trie_hash = self.calculate_node_hashes(storage_tx, 0)?;
+
+        // find marf root hash -- the hash of the trie root node hash, and the hashes of the
+        // geometric series of ancestor tries.  Because the trie is already in the process of
+        // being flushed, we have to temporarily reinstate its data into `storage_tx` so we can
+        // use it to walk down the various MARF paths needed to query ancestor tries.
```
kantai, line 669: I'm not sure I follow why the trie already being flushed makes "reinstatement" necessary for finding the marf root hash. What breaks in the normal MARF walk if you tried to `calculate_marf_root_hash` on `self` as it is?

src/chainstate/stacks/index/storage.rs
```diff
@@ -672,7 +1240,12 @@ pub struct TrieStorageTransaction<'a, T: MarfTrieId>(TrieStorageConnection<'a, T
 pub struct TrieStorageConnection<'a, T: MarfTrieId> {
     pub db_path: &'a str,
     db: SqliteConnection<'a>,
+    blobs: Option<&'a mut TrieFile>,
     data: &'a mut TrieStorageTransientData<T>,
+    cache: &'a mut TrieCache<T>,
+    bench: &'a mut TrieBenchmark,
+    pub hash_calculation_mode: TrieHashCalculationMode,
+    unconfirmed_block_id: Option<u32>,
```
kantai, line 1248: What is `unconfirmed_block_id`?

src/chainstate/stacks/index/storage.rs
```diff
@@ -685,28 +1258,38 @@ pub struct TrieStorageConnection<'a, T: MarfTrieId> {
 ///   to the underlying SQL storage. Used internally to simplify
 ///   the TrieStorageConnection/TrieFileStorage interactions
 ///
-struct TrieStorageTransientData<T: MarfTrieId> {
-    last_extended: Option<(T, TrieRAM<T>)>,
+pub struct TrieStorageTransientData<T: MarfTrieId> {
+    /// This is all the nodes written but not yet committed to disk.
```
kantai, line 1262: This is the latest trie, right?

src/chainstate/stacks/index/storage.rs
```diff
@@ -965,6 +1646,69 @@ impl<'a, T: MarfTrieId> TrieStorageTransaction<'a, T> {
...
 
+    /// Run `cls` with a mutable reference to the inner trie blobs opt.
+    fn with_trie_blobs<F, R>(&mut self, cls: F) -> R
+    where
+        F: FnOnce(&Connection, &mut Option<&mut TrieFile>) -> R,
+    {
+        let mut blobs = self.blobs.take();
+        let res = cls(&self.db, &mut blobs);
+        self.blobs = blobs;
+        res
```
kantai, line 1657: Does this need to `take`?

```suggestion
        cls(&self.db, self.blobs.as_mut())
```

kantai: It seems like any use of this method could just use `&self.db` and `self.blobs.as_mut()` directly, rather than needing a method with a function generic.

src/chainstate/stacks/index/storage.rs
```diff
@@ -965,6 +1646,69 @@ impl<'a, T: MarfTrieId> TrieStorageTransaction<'a, T> {
...
+    }
+
+    /// Store a serialized trie, given the flush options and its block hash.
+    /// Depending on how the trie storage is configured, the trie will either get dumped into a
+    /// blob in the underlying sqlite database, or it will be appended to a flat file that lives
+    /// alongside the sqlite database.
+    fn inner_store_trie_blob(
```
kantai, line 1664: It seems like this method is only ever invoked by `inner_flush`, and the structure of the method uses the same match on `FlushOptions`, would it make more sense if this was just folded into the `inner_flush` method?

src/chainstate/stacks/index/storage.rs
```diff
@@ -1363,122 +2159,191 @@ impl<'a, T: MarfTrieId> TrieStorageConnection<'a, T> {
...
+            self.unconfirmed()
+        );
+        self.bench.open_block_start();
+
         if *bhh == self.data.cur_block && self.data.cur_block_id.is_some() {
             // no-op
+            if self.unconfirmed() {
+                if self.data.cur_block_id
+                    == trie_sql::get_unconfirmed_block_identifier(&self.db, bhh)?
+                {
```
kantai, line 2220: What does it mean if this equality is false?

src/chainstate/stacks/index/storage.rs
kantai: Also, what is its relationship to `data.cur_block_id`?

src/chainstate/stacks/index/trie.rs
```diff
@@ -64,42 +75,77 @@ fn get_nodetype_hash<T: MarfTrieId>(
...
 impl Trie {
     /// Read the root node.  First try to read it as a back-pointer (since all root nodes except for
     /// the root node in the very first trie will be back-pointers), and if that fails due to a
     /// node ID mismatch (i.e. CorruptionError), then try to read it as a non-backpointer.
-    pub fn read_root<T: MarfTrieId>(
+    #[inline]
+    fn read_root_maybe_hash<T: MarfTrieId>(
         storage: &mut TrieStorageConnection<T>,
+        read_hash: bool,
     ) -> Result<(TrieNodeType, TrieHash), Error> {
```
kantai, line 94: Using `Option<TrieHash>` and unwrapping from the caller is a more explicit (and slightly safer) way to represent what's happening here than using empty buffers. It would also avoid the allocations (though that's admittedly probably a pretty small cost).

src/clarity_vm/database/marf.rs
```diff
@@ -47,11 +47,14 @@ impl MarfedKV {
             .ok_or_else(|| InterpreterError::BadFileName)?
             .to_string();
 
+        let mut marf_opts = MARFOpenOpts::default();
```
kantai, line 50: Should this take the `MARFOpenOpts` as a parameter, and default to `MARFOpenOpts::default() + external_blobs = true`?

src/chainstate/stacks/index/trie_sql.rs
```diff
@@ -62,6 +66,8 @@ static SQL_MARF_DATA_TABLE: &str = "
 CREATE TABLE IF NOT EXISTS marf_data (
    block_id INTEGER PRIMARY KEY, 
    block_hash TEXT UNIQUE NOT NULL,
+   -- the trie itself.
+   -- if not used, then set to a zero-byte entry.
```
kantai, line 70: It seems like we want a few of these fields to be NULLable, right?

If `data` isn't present, using a zero-byte entry is basically an approximation of "NULL", but with some extra weirdness involved. Similarly, `external_offset` and `external_length` using 0 rather than "NULL".

src/chainstate/stacks/index/trie_sql.rs
```diff
@@ -168,6 +235,64 @@ pub fn write_trie_blob<T: MarfTrieId>(
...
+            .insert(args)?
+            .try_into()
+            .expect("EXHAUSTION: MARF cannot track more than 2**31 - 1 blocks");
+
+        debug!(
+            "Wrote block trie {} to rowid {} offset {}",
+            block_hash, block_id, offset
+        );
+        block_id
+    };
```
kantai, line 290: Is it possible for this method to be called _without_ a `block_id` if the `block_id` already exists in the database? Should this be part of the calling contract for this method (if so, this assumption should be made explicit in a rustdoc entry for the method)?

I guess this could also be made an `INSERT OR REPLACE`.

kantai (commented):
This all looks great to me, just a few comments and questions.

One high-level comment is around cache control via environment variables. It would be much better for that to be controllable via the `stacks-node::config` module -- almost all of the configuration variables are currently controlled there, and avoiding adding more environment variable switched behaviors would be wise.

src/chainstate/stacks/index/bits.rs
jcnelson: Lol those have been there for years.  Sure, I can remove them.

src/chainstate/stacks/index/bits.rs
jcnelson: Sure, can do.

src/chainstate/stacks/index/cache.rs
jcnelson: Sorry, that's the block's local ID -- i.e. the row ID in the database.  Will update the comment.

src/chainstate/stacks/index/cache.rs
jcnelson: Why yes it is.  Will fix.

The implementation always needs an owned copy of the node when it's loaded, which is why there's no option to return a reference to it.  There is a method for getting the reference to a hash, however, since it's not always the case that the caller needs an owned copy.

src/chainstate/stacks/index/cache.rs
jcnelson: Yup; will fix.

src/chainstate/stacks/index/cache.rs
jcnelson: I believe so.

src/chainstate/stacks/index/file.rs
jcnelson: Yes -- a `TrieFile` is a flat storage space for tries (as opposed to a sqlite DB).  Will add rustdoc.

src/chainstate/stacks/index/bits.rs
kantai: I know, but these are new invocations, so I felt that now was as good a time as ever 😄

src/chainstate/stacks/index/file.rs
jcnelson: Yup, will fix.

src/chainstate/stacks/index/marf.rs
jcnelson: I'm not sure I understand the question?  The reason `default()` uses `"noop"` here, instead of reading the environment variable, is because the environment variable is the concern of the cache options, not the MARF options.

src/chainstate/stacks/index/marf.rs
jcnelson: Removed

src/chainstate/stacks/index/profile.rs
jcnelson: I can add a no-op version for production.  It's currently only used for testing.

src/chainstate/stacks/index/marf.rs
kantai: But isn't the cache option for a MARF ultimately being controlled by the `MARFOpenOpts`?

kantai: Like, if I wanted to enable caching for the Clarity MARF, how would that work? The `MARFOpenOpts` are specified by invoking `MARFOpenOpts::default()` and then setting the storage strategy to `TrieFile` storage, but the cache strategy will always be `noop`, regardless of the environment variable setting.

src/chainstate/stacks/index/storage.rs
jcnelson: I think you mean `if !self.is_block_hash_cached(id) {`, but I take your meaning.

src/chainstate/stacks/index/storage.rs
jcnelson: The call to `inner_flush` has this:

```rust
        if let Some((bhh, trie_ram)) = self.data.uncommitted_writes.take() {
            trace!("Buffering block flush started.");
            let mut buffer = Cursor::new(Vec::new());
            trie_ram.dump(self, &mut buffer, &bhh)?;
```

Because `self.data.uncommitted_writes` will be `None` at the time of the call to `trie_ram.dump()`, it will be impossible to carry out a MARF walk.  The uncommitted TrieRAM contains the root node from which the walks will begin.

src/chainstate/stacks/index/storage.rs
jcnelson: `unconfirmed_block_id` is the row ID of the trie representing unconfirmed state.  This is an odd case, because such tries can be re-loaded from disk, modified, re-written, and dropped (whereas all other tries are write-once-read-many and cannot be dropped).  This field is only set to `Some(..)` if the trie being operated upon is such a trie.

`self.cur_block_id` remains the row ID of the trie that is currently open.  It may or may not be equal to `self.unconfirmed_block_id`.

jcnelson: I'll add rustdoc for this.

src/chainstate/stacks/index/storage.rs
jcnelson: Yes; it's just renamed from `last_extended` to `uncommitted_writes`, since it's using the new struct `UncommittedState<T>`.

src/chainstate/stacks/index/storage.rs
jcnelson: It means that we're opening a block that doesn't represent unconfirmed state.  This is normal over the course of a MARF walk from an unconfirmed trie.  A MARF walk would start out from the unconfirmed trie, encounter a backptr to a confirmed trie, and then open the confirmed trie.  In that last step of resolving the backptr, this condition would be `false`.

Note that it's impossible by construction to step _to_ an unconfirmed trie, because (1) the unconfirmed trie's block ID is globally unique and derived from the confirmed trie it extends, and (2) you can't build an unconfirmed trie on top of an unconfirmed trie.

jcnelson: Recall that `self.unconfirmed()` means "this storage connection was instantiated for reading and writing unconfirmed state on top of a MARF."  This is important because unconfirmed trie blobs are treated differently -- for example, they cannot be stored to the new `TrieFile` instance, since they must later be deleted (and `TrieFile` only supports append-only writes).

src/chainstate/stacks/index/storage.rs
jcnelson: Sure, I can do that.

src/chainstate/stacks/index/storage.rs
jcnelson: N.B. The reason `self.unconfirmed_block_id` is set here, and not in the constructor, is because it follows the semantics of `self.cur_block_id` -- the row ID isn't resolved until we go to open the block.

src/chainstate/stacks/index/trie.rs
jcnelson: Sure, can do

src/clarity_vm/database/marf.rs
jcnelson: Sure, I can plumb this through.  I'll have to make it so that we can extract the open options from the `stacks-node::config` file though, so it'll add about 100-200 extra lines to the diff.

src/chainstate/stacks/index/trie_sql.rs
jcnelson: I don't believe we can alter the schema to remove the `NOT NULL` constraint without copying the entire contents of the table to a new table with the constraint removed [1], so I'm not sure how valuable this particular change would be under the circumstances.

As for `external_offset` and `external_length`, sure -- these can be nullable.

[1] https://stackoverflow.com/questions/4007014/alter-column-in-sqlite

src/chainstate/stacks/index/storage.rs
kantai: > Recall that self.unconfirmed() means "this storage connection was instantiated for reading and writing unconfirmed state on top of a MARF."

Can you add that to a rustdoc for the field `unconfirmed` and for the function `pub fn unconfirmed(&self)`?

src/chainstate/stacks/index/trie_sql.rs
jcnelson: On second thought, I'm going to push back on this.  In every case where we query `external_offset` and `external_length`, we're going to need to do `.unwrap_or(0)` anyway, as well as modify the `SELECT` queries to filter out rows where these values are `NULL`.  In order to do the latter efficiently, we're going to have to add an additional index for `external_length`, which would take more space than simply storing the default value `0` for these columns.

src/chainstate/stacks/index/trie_sql.rs
jcnelson: I'll factor this into `write_external_trie_blob()` which stores a new trie blob, and `update_external_trie_blob()` which updates an existing trie blob.  They'll just call this method (which will be made into a private "inner" method) with the appropriate arguments.

src/chainstate/stacks/index/storage.rs
jcnelson: Yes, it needs `.take()`.  You need to move `blobs` out of `self` so you can call a mutable method in `blobs` that takes an immutable reference to a member of `self`.  Otherwise you get borrow-checker errors:

```
   Compiling blockstack-core v0.0.1 (/home/jude/stacks/stacks-blockchain.marf-node-cache)
error[E0502]: cannot borrow `*self` as immutable because it is also borrowed as mutable
    --> src/chainstate/stacks/index/storage.rs:1759:63
     |
1758 |                     match self.blobs.as_mut() {
     |                           ---- mutable borrow occurs here
1759 |                         Some(blobs) => blobs.store_trie_blob(&self.db, &bhh, &buffer)?,
     |                                              ---------------  ^^^^ immutable borrow occurs here
     |                                              |
     |                                              mutable borrow later used by call

error[E0502]: cannot borrow `*self` as immutable because it is also borrowed as mutable
    --> src/chainstate/stacks/index/storage.rs:1784:63
     |
1783 |                     match self.blobs.as_mut() {
     |                           ---- mutable borrow occurs here
1784 |                         Some(blobs) => blobs.store_trie_blob(&self.db, real_bhh, &buffer)?,
     |                                              ---------------  ^^^^ immutable borrow occurs here
     |                                              |
     |                                              mutable borrow later used by call
```

src/chainstate/stacks/index/storage.rs
jcnelson: Sure!

src/chainstate/stacks/index/marf.rs
jcnelson: I instead made it so that the `MARFOpenOpts` passed to the Stacks chainstate is specified in the `stacks-node::config::Config` struct.

gregorycoppola (approved):
Seems like this is working so LGTM.

Thanks for adding all this new functionality!

src/chainstate/stacks/index/trie_sql.rs
kantai: That's fine. I don't think it's true that you'd need to filter out the NULL rows (the PR doesn't currently filter 0/0 rows, so why would you need to filter `NULL` rows?), and the `unwrap_or(0)` could be avoided by propagated this change upwards so that the structs _also_ represent lack of data as an `Option`, rather than a 0/0 (or like a storage type enum or something), but anyways, it isn't necessary, and if you want to push back on this refactor that works for me.

jcnelson: > I don't think it's true that you'd need to filter out the NULL rows (the PR doesn't currently filter 0/0 rows, so why would you need to filter NULL rows?)

The PR depends on the ordering of `external_offset`, since the code needs to know the offset at which to append the next trie blob.  I don't know how stable the rules are in sqlite3 for comparing `NULL` to `INTEGER`, since on the face of it it ought to be undefined, but I don't want to find out later that `NULL` gets (re)defined as "greater than" the maximum `INTEGER` value in a `SELECT MAX(external_offset)` statement.

src/chainstate/stacks/index/storage.rs
```diff
@@ -1655,6 +1655,7 @@ impl<'a, T: MarfTrieId> TrieStorageTransaction<'a, T> {
         res
     }
 
+    /*
```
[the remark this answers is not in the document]
jcnelson: Yup; done

kantai (approved):
LGTM! Excited to see these improvements in the node.

PR #3047, opened 2022-02-11
Fix/smooth coordinator shutdown

### LINKED ISSUE #3015: stacks-node does not broadcast new block
**Describe the bug**
The stacks-node does not attempt to broadcast a new block if shutdown is requested soon after receiving the new block. This can cause database corruption in a stacks-blockchain-api instance.

Screen Shot 2022-01-24 at 11 47 40 AM

**Steps To Reproduce**
Terminate a stacks-node right after it receives a new block.

**Expected behavior**
The stacks-node should only exit after completing any pending broadcasting steps successfully.

pavitthrap:
### Description
Updates chains coordinator to ensure that **before** a Stacks/burn block is processed, an event is emitted. 

### Applicable issues
- fixes #3015

### Checklist
- [x] Changelog is updated
- [x] New clarity functions have corresponding PR in `clarity-benchmarking` repo


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
```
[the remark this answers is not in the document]
kantai: I suspect these are just test flake -- we'll see as the tests re-run.

src/burnchains/burnchain.rs
```diff
@@ -891,8 +891,20 @@ impl Burnchain {
 
         let sortition_tip = SortitionDB::get_canonical_sortition_tip(db.conn())?;
 
-        db.evaluate_sortition(&header, blockstack_txs, burnchain, &sortition_tip, None)
-            .map(|(snapshot, transition, _)| (snapshot, transition))
+        let (parent_snapshot, parent_sort_id, reward_set_info) =
+            db.get_sortition_info(&header, burnchain, &sortition_tip, None)?;
+
+        // The block is not broadcast to the event observer here since this is a legacy function.
```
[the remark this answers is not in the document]
kantai: Updated the comment here -- this method is deprecated and invoked only in the defunct helium run-loop and some test methods. Because of that, we never broadcasted burn block info to event observers here, and that doesn't change in this PR.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -2837,10 +2827,37 @@ impl SortitionDB {
...
             )?
         };
+        sortition_db_handle.commit()?;
 
+        Ok((parent_snapshot, parent_sort_id, reward_set_info))
+    }
+
+    ///
+    /// # Arguments
+    /// * `burn_header` - the burnchain block header to process sortition for
```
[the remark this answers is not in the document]
kantai: A lot of that rustdoc (the argument docs) is existent in the base branch, but I updated the rustdoc to also explain what `evaluate_sortition` means here.

src/chainstate/coordinator/mod.rs
```diff
@@ -760,7 +770,7 @@ impl<
                             "Atlas: {} attachment instances emitted from events",
                             attachments_instances.len()
                         );
-                        match self.attachments_tx.send(attachments_instances) {
+                        match self.attachments_tx.try_send(attachments_instances) {
```
jcnelson, line 773: What will be the consequences of this failing due to a full channel?  I think it would cause the Atlas subsystem to miss attachments, right?  The `attachment_instances` here are gleaned from `name-register`, `name-update`, and `name-renewal` contract-calls, so if these get dropped, then there's no way for the Atlas subsystem to discover them.

Would it be better to just get rid of the channel, and instead just write the unfulfilled attachment instances directly to the Atlas DB?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -2742,29 +2744,27 @@ impl SortitionDB {
...
     /// # Arguments
     /// * `burn_header` - the burnchain block header to process sortition for
     /// * `ops` - the parsed blockstack operations (will be validated in this function)
     /// * `burnchain` - a reference to the burnchain information struct
     /// * `from_tip` - tip of the "sortition chain" that is being built on
     /// * `next_pox_info` - iff this sortition is the first block in a reward cycle, this should be Some
-    ///
-    pub fn evaluate_sortition(
+    /// * `announce_to` - a function that will be invoked with the calculated reward set before this method
+    ///                   commits its results.
```
jcnelson, line 2758: Can you add to this comment to describe its concrete use-case?  "For example, this function is used to post the calculated reward set to an event observer" or something like that.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -2823,9 +2823,11 @@ impl SortitionDB {
 
         sortition_db_handle.store_transition_ops(&new_snapshot.0.sortition_id, &new_snapshot.1)?;
 
+        announce_to(reward_set_info);
```
jcnelson, line 2826: There's no plausible scenario where this `announce_to()` succeeds but the subsequent commit fails, right?  Maybe the subsequent commit should `panic`, just to be sure?

src/chainstate/coordinator/mod.rs
jcnelson: If you don't want to do this in this PR, can you at least open an issue for it?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4888,9 +4935,9 @@ impl StacksChainState {
     /// block's transactions.  Finally, it returns the execution costs for the microblock stream
     /// and for the anchored block (separately).
     /// Returns None if we're out of blocks to process.
-    fn append_block(
+    fn append_block<'a>(
```
jcnelson, line 4938: Can you update the docstring to describe why returning a `PreCommitClairtyBlock<'a>` is necessary here?  Also, I think the docstring is obsolete in that it describes all the things that are now currently encapsulated in the `StacksEpochReceipt`.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5688,7 +5761,14 @@ impl StacksChainState {
             true,
         )?;
 
-        chainstate_tx.commit().map_err(Error::DBError)?;
+        // this will panic if the Clarity commit fails.
```
jcnelson, line 5764: Can you open an issue to execute the `chainstate_tx.commit()` and `clairty_commit.commit()` calls in the same sqlite transaction?  This is ultimately what we really want -- corruption due to one commit succeeding and another failing, or even stalling for a long time in-between, are both undesirable behaviors.

src/clarity_vm/clarity.rs
```diff
@@ -105,13 +105,18 @@ pub struct ClarityInstance {
     mainnet: bool,
 }
 
+pub struct PreCommitClarityBlock<'a> {
```
jcnelson, line 108: Can you document why this struct is necessary?  i.e. we need to do other blocking I/O (such as post events to an event dispatcher) before committing to the chainstate headers MARF and the Clarity DB MARF, so we need to effectively "seal" an ongoing Clarity transaction into this struct so we can do the blocking I/O first (instead of in-between these two transactions)?

testnet/stacks-node/src/main.rs
```diff
@@ -55,9 +55,9 @@ use backtrace::Backtrace;
 
 fn main() {
     panic::set_hook(Box::new(|panic_info| {
-        eprintln!("Process abort due to thread panic: {}", panic_info);
+        error!("Process abort due to thread panic: {}", panic_info);
```
jcnelson, line 58: So, the reason this was `eprintln!` was because I wanted some level of guarantee that the `panic_info` and backtrace would get logged even if the logger was somehow broken or blocked on I/O (or even memory pressure).

jcnelson (approved):
This overall LGTM, modulo the comments.  I think the biggest decision to make before merging is determining how we'll handle the case where atlas attachment events get reliably processed if the atlas channel gets full.  I think we should just store them to the DB, per @lgalabru's suggestion, but it's fine if that's handled via a separate PR.  If you choose to do it in this PR, please re-request me as a reviewer.

testnet/stacks-node/src/main.rs
kantai: Yes -- in the case of a JSON logger, any JSON-expecting consumer will lose this information without creating special case logic for panics. This is the case for Hiro's nodes, and would be the case for most people that just try to capture the logs using standard tools.

The stacks node's term logger writes to `stderr` and flushes on every macro invocation, so it would log before this panic finished (same as `eprintln!`) -- the only difference is if there is not enough memory to buffer this log line, the process would get killed before it had a chance to log the panic message, but that is exceedingly unlikely: the panic would need to occur at a time when there's just the log message (perhaps 100kb) worth of memory left to the process. This wouldn't be the case in a normal OOM situation: the panic handler won't get invoked then at all.

src/chainstate/coordinator/mod.rs
kantai: I believe that the consequences of this being dropped are low: the atlas subsystem will reread the attachments from the database anyways. This was discussed in a blockchain meeting before this PR was initially drafted -- I think @lgalabru mentioned that the Atlas subsystem was designed to allow for these events to be dropped. However, we would need to confirm that that's the case by looking at the codebase. @lgalabru, do you think that these events occassionally getting dropped is okay?

src/chainstate/burn/db/sortdb.rs
kantai: Expanded in 8608cc7fd

src/chainstate/burn/db/sortdb.rs
kantai: Updated to `expect` in 8608cc7fd

src/clarity_vm/clarity.rs
kantai: Added rustdocs for this struct in 8608cc7fd

src/chainstate/coordinator/mod.rs
gregorycoppola: What happened with this question?

gregorycoppola (commented):
LGTM, except just wanted to follow up with this question

https://github.com/stacks-network/stacks-blockchain/pull/3047#discussion_r823021404

src/chainstate/coordinator/mod.rs
lgalabru: Hmmm. So the receiver component of this channel, located in the p2p thread, is in charge of persisting these attachments instances (not resolved at this point). If some attachments instances are not resolved when a stacks-node is being shutdown, then there's a builtin mechanism that will ensure that we resolve them. If the stacks-node is being shut down while these events are still in the channel, then they only exists in memory and if we're looking at this case, I guess the only way to ensure that we resolve these attachments, is to re-process these blocks.
We could move the logic that persist the attachments in this thread, however it would mean that we now have 2 threads writing to the atlas db, which is probably something we'd like to avoid?

lgalabru: The attachement channel being filled (current capacity set to 1) would imply that the chain coordinator was able to process 2 different blocks, with a starving p2p thread, right? Is this possible by design?

lgalabru: It makes sense. In that case the method `enqueue_new_attachments` in https://github.com/stacks-network/stacks-blockchain/blob/6c5349b0422c5c88db02d578720467c16b102ab2/src/net/p2p.rs#L5271-L5287 would require some refactoring, but could be a good starting point.

kantai: Hmm. It seems like changing this to do the DB write in the coordinator thread wouldn't be entirely trivial: there's also the issue of notifying the network result of any now-resolved attachments, and it's not obvious to me if we could just write the the new attachment instances and then let the downloader thread figure out that there are new instances that can be fulfilled (I don't think so).

I think perhaps the best way forward is to deal with the Atlas synchrony issues in another PR -- the current release implementation doesn't retain all events in a shutdown, and worse, can prevent shutdown from occurring smoothly; in most cases this results in a process kill and a corrupted chainstate or API state.

The `try_send` approach resolves the second issue, but it _can_ lead to dropped events even outside of shutdown, so I think the best interim solution is to increase the size of the channel to 5 allowing the coordinator to run 5 blocks ahead of the attachments receiver, and switch this back to a synchronous `send`. In the normal execution, there won't be dropped events, and in the event of a shutdown, as long as the attachments receiver isn't already 5 blocks behind, the coordinator will be able to smoothly shutdown (albeit dropping some Atlas events, but as we already covered: Atlas events are already lost during shutdown). Even in the event that the attachments receiver _was_ behind, fixing the p2p thread panic (which this PR did) should also help here: the mpsc sync_channel implementation should give a channel closed error to the `send()` caller.

Does that make sense to you, @jcnelson and @lgalabru ?

kantai: (Also, this is what I did in 659b2c45c)

lgalabru: > I think the best interim solution is to increase the size of the channel to 5 

Sounds good to me

> it's not obvious to me if we could just write the the new attachment instances and then let the downloader thread figure out that there are new instances that can be fulfilled (I don't think so).

There could be a few ways to approach this. The coordinator could be inserting some entries in the table `attachment_instances`, setting the `created_at`, and the atlas state machine in the p2p thread could keep a value of the most recent `attachment_instances` processed, and only insert the latest ones in the state machine.

kantai:
> LGTM, except just wanted to follow up with this question

This is now addressed

gregorycoppola (approved):

src/chainstate/coordinator/mod.rs
jcnelson: Damn, wish I'd seen this before merging.  Why stop at 5?  Why not make it, I dunno, 1,000?  It's not like they take much memory.

Issue #3045, opened 2022-02-11
[chainstate] memoize block-commit descendancy information for `SortitionHandleTx::descended_from`


jcnelson:
The method `SortitionHandleTx::descended_from` is particularly slow, since it can do *lots* of MARF reads in the sortition DB.  Looking at the log files, it seems to be the biggest contributor to sortition-processing -- even more so than downloading the Bitcoin block!  Fortunately, we can easily memoize the PoX anchor block that a block-commit descends from; this should remove the need to query the MARF at all.

gregorycoppola:
Related: This issue is the line of work suggested in https://github.com/stacks-network/stacks-blockchain/issues/3014, which suggests looking for ways to cache the MARF in RAM to improve performance.

PR #3048, opened 2022-02-14
fix/memoize-block-commit-descendancy

jcnelson:
This PR addresses #3045 by adding a new field `parent_sortition_id` to the `block_commits` table in the sortition DB, which shall be equal to the sortition ID of the parent block-commit.  This field is then used in `SortitionHandleTx::descended_from` to scan the list of block-commit ancestors to see if the winning block at a given height descends from a PoX anchor block.

Using this field is much, much faster than querying the sortition DB MARF for each block height in a reward cycle.  In fact, the former behavior is the *limiting factor* in booting up a node with #3044 applied.

NOTE: this changes the sortition DB schema in an _incompatible_ way.  The release that incorporates this PR as-is will need a chainstate version number increment.


jcnelson:
Feedback: don't change the chainstate schema in an incompatible way; instead, re-calculate `parent_sortition_id` for each `block_commit` row on boot-up with a chainstate that does not yet have it.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -2362,6 +2384,17 @@ impl SortitionDB {
         query_row(conn, qry, &args)
     }
 
+    /// Get a block commit's parent commit's block's sortition ID
```
kantai, line 2387: Perhaps more clear as 

```suggestion
    /// Get the Sortition ID for the burnchain block containing `block_commit_txid`'s parent
```

src/chainstate/burn/db/sortdb.rs
```diff
@@ -3801,6 +3992,20 @@ impl<'a> SortitionHandleTx<'a> {
         let apparent_sender_str = serde_json::to_string(&block_commit.apparent_sender)
             .map_err(|e| db_error::SerializationError(e))?;
 
+        // find parent block commit's snapshot's sortition ID.
+        // If the parent_block_ptr doesn't point to a valid snapshot, then store an empty
+        // sortition.  If we're not testing, then this should never happen.
```
kantai, line 3997: This should never happen because the `parent_block_ptr` is checked for validity elsewhere? Can you expand this comment to specify where that check occurs?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -2430,6 +2463,148 @@ impl SortitionDB {
         Ok(())
     }
 
+    fn apply_schema_3(tx: &mut SortitionHandleTx) -> Result<(), db_error> {
+        // fill in block_commit_parents table
+        info!("Populating block-commit parent table (this might take a while)");
```
kantai, line 2468: Can you add something to this message that indicates this is a one-time migration? e.g., `"Migrating sortition database schema from v2 to v3: populating block-commit parents (this might take a while)"`

src/chainstate/burn/db/sortdb.rs
```diff
@@ -2430,6 +2463,148 @@ impl SortitionDB {
...
+                        let res: Result<_, rusqlite::Error> = Ok(commit);
+                        res
+                    })
+                    .expect("FATAL: failed to query block_commits table");
+                // .collect() is not implemented for AndThenRows
+                let mut ret = vec![];
+                for value in query_values {
+                    ret.push(value?);
+                }
+                ret
```
kantai, line 2509: Couldn't you use `util_lib::db::query_rows` here?

kantai (commented):
This looks good to me, just had a few superficial comments. Also, please update the `CHANGELOG.md`

src/chainstate/burn/db/sortdb.rs
jcnelson: I updated the comment.  The reason is that `insert_block_commit()` gets called only for valid block-commits.  One of the validity checks is that a block-commit's parent must have been accepted and stored already in this sortition history, or it must point to `(0,0)`.  This check happens in `LeaderBlockCommitOp::check()`.

src/chainstate/burn/db/sortdb.rs
jcnelson: Fixed

src/chainstate/burn/db/sortdb.rs
jcnelson: Fixed

src/chainstate/burn/db/sortdb.rs
jcnelson: Fixed

src/chainstate/burn/db/sortdb.rs
kantai: Maybe you missed a `git push`? I don't see any changes yet in the PR.

jcnelson: Weird. Fixed.

kantai: Same goes for the other comment threads -- I'm not seeing any of the changes from those either.

jcnelson: How bizarre.  I remember pushing them.  Lemme do them again.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -2398,8 +2435,8 @@ impl SortitionDB {
     pub fn is_db_version_supported_in_epoch(epoch: StacksEpochId, version: &str) -> bool {
         match epoch {
             StacksEpochId::Epoch10 => false,
-            StacksEpochId::Epoch20 => (version == "1" || version == "2"),
-            StacksEpochId::Epoch2_05 => version == "2",
+            StacksEpochId::Epoch20 => (version == "1" || version == "2" || version == "3"),
```
gregorycoppola, line 2438: clarification: does epoch20 support version 3?

gregorycoppola:
> > Is this really true "This PR addresses #3045 by adding a new field parent_sortition_id to the block_commits table in the sortition DB"? Seems like you are adding a new table block_commit_parents.
> 
> This is no longer true, per feedback given here: [#3048 (comment)](https://github.com/stacks-network/stacks-blockchain/pull/3048#issuecomment-1039283876). In order to keep the sortition DB compatible with existing sortition DBs, a new table was added instead.
Can you update the description before checking in so we aren't confused later?

> 
> > Would be nice if we could quantify "Using this field is much, much faster", but can skip this if it's a huge hassle for some reason.
> 
> Processing a sortition is about 10x faster now, when considering later sortitions (where the sortition DB MARF is bigger). It takes about 0.3s to process a sortition with this PR, and it took about 3s without it.

:)!

gregorycoppola (approved):
@jcnelson please update the PR description to match the change!

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ and this project adheres to the versioning scheme outlined in the [README.md](RE
 
 ### Changed
 - Sortition processing performance has been improved by about an order of
-  magnitude, by avoiding a slew of expensive database reads (#3048).
+  magnitude, by avoiding a slew of expensive database reads (#3045).
```
kantai, line 12: This should mention (and warn) about the schema migration. Restarting a node, using the an existing chainstate directory, with this change will require more time than other updates have in the past, so users should get some kind of warning in the changelog (either in this entry, or at the top level content for the release).

jcnelson: Fixed

jcnelson:
I noticed a couple bugs in the implementation when testing this on a live migration:

* The migration of the canonical sortition history takes 3-4 *hours*.  There needs to be a bigger warning about this.
* The migration logic panics for non-canonical sortition histories (i.e. ones that aren't on the canonical PoX fork).  This shouldn't happen.

jcnelson:
Don't bother doing a migration; just fall back to the old code if the `block_commit_parents` table doesn't exist or is filled with `NULL` data.

kantai (approved):
LGTM!

PR #3050, opened 2022-02-15
fix: remove blocks and microblocks from requested status if the connection times out

jcnelson:
This will prevent a block or microblock download from stalling for ~1 minute if the request times out.


kantai:
This change looks fine to me, but is this possible to test via a unit test? Also, this should get a changelog entry.

gregorycoppola (commented):
Nice idea.

I agree it would be nice if somehow a test could ensure we don't inadvertently break the new behavior later.

PR #3053, opened 2022-02-17
reduce DB contention between relayer and chains coordinator threads

jcnelson:
Just a draft for now; trying it out with a full boot-up.  Will update as I discover things.


jcnelson:
Taking notes as I go along.

Another source of lock contention during boot-up -- the chains coordinator thread and relayer thread lock the chainstate DB and block each other.  The former does so while processing blocks; the latter while pre-processing and storing them.  Maybe the relayer should just pipe validated and preprocessed blocks and microblocks to the chains coordinator via a channel insteada?  Maybe the relayer and chains coordinator should guard access to chainstate DB via a mutex to avoid the exponential back-off time penalty?  It's good to have the relayer handle validation concurrently with the chains coordinator's processing.

jcnelson:
`StacksChainState::preprocess_anchored_block()` and `StacksChainState::preprocess_streamed_microblock()` could probably be refactored to simply perform validation with a DB connection, and be given a channel `SyncSender` into which to send the validated block or microblock.  The chains coordinator would own the `Receiver` end, and would simply store the data.  We'd probably just want to make the channel argument an `Option<..>` for these two methods for compatibility, and have them fall back to using a transaction if it's `None`.

jcnelson:
Hmm, I'm less sure the channel idea is good.  Instead, we'd probably want the chains coordinator thread to assume responsibility for both block/microblock validation as well as storage.  We put the checks into the same DB transaction as the storage operation to avoid TOCTTOU problems.

jcnelson:
Watching my node pass block 20,000, it's clear that the DB contention between the relayer thread and the chains coordinator thread is slowing down boot-up.  I think the following high-level change is in order:

* The responsibility for processing the block and microblock data in a `NetworkResult` (i.e. the call to `Relayer::process_new_blocks()`) should be moved to the chains coordinator thread.  Right now, this happens in the relayer thread's main event loop in the Neon node.  Instead, the p2p thread should just pass the block data from `NetworkResults` to the chains coordinator.

* The chains coordinator thread should, in turn, pass the relevant broadcast data back to the p2p thread.  The relayer thread does this right now, via a channel opened to the `PeerNetwork` instance.  This channel should be moved into the chains coordinator.

* The relayer should remain responsible for processing unconfirmed state and transaction admission to the mempool.  The former should be structured so that it is de-prioritized relative to the chains coordinator thread's transactions.  Long-term, the real answer here is to make unconfirmed state-processing able to access the `StacksChainState` MARF without needing a transaction (but this will require a significant refactor of the MARF).

jcnelson:
This is now ready for review. Note that it is built on #3048

src/chainstate/burn/db/sortdb.rs
```diff
@@ -2661,11 +2664,17 @@ impl SortitionDB {
...
+        // do we need to instantiate indexes?
+        // only do a transaction if we need to, since this gets called each time the sortition DB
+        // is opened.
+        let exists : i64 = query_row(self.conn(), "SELECT 1 FROM sqlite_master WHERE type = 'index' AND name = ?1", &[LAST_SORTITION_DB_INDEX])?.unwrap_or(0);
+        if exists == 0 {
+            let tx = self.tx_begin()?;
+            for row_text in SORTITION_DB_INDEXES {
+                tx.execute_batch(row_text)?;
+            }
+            tx.commit()?;
```
kantai, line 2681: I think the first query should be included in the tx.

kantai: How often is this invoked? This is a database schema check

jcnelson: It's called by `SortitionDB::connect()`, which gets called via `Burnchain::sync_with_indexer()`, which gets called every time you want to download a Bitcoin block.  So, it's pretty often.

kantai: I see. I guess that's not ideal (Burnchain sync with indexer shouldn't need to reopen a connection on every invocation), but it's probably not too burdensome as long as it doesn't try to initiate a transaction.

This is fine, I suppose, the threat of `exists` being inconsistent with the index creation tx is pretty low, because the index creation is idempotent anyways.

kantai (approved):
LGTM!

src/net/download.rs
```diff
@@ -463,6 +463,9 @@ impl BlockDownloader {
                         } else {
                             self.dead_peers.push(event_id);
 
+                            // try again
+                            self.requested_blocks.remove(&block_key.index_block_hash);
```
gregorycoppola, line 467: formatting?

testnet/stacks-node/src/neon_node.rs
```diff
@@ -1083,7 +1083,8 @@ fn spawn_miner_relayer(
                         // no burnchain change, so only re-run block tenure every so often in order
                         // to give microblocks a chance to collect
                         if issue_timestamp_ms < last_tenure_issue_time + (config.node.wait_time_for_microblocks as u128) {
-                            debug!("Relayer: will NOT run tenure since issuance at {} is too fresh (wait until {} + {} = {})", issue_timestamp_ms / 1000, last_tenure_issue_time / 1000, config.node.wait_time_for_microblocks / 1000, (last_tenure_issue_time + (config.node.wait_time_for_microblocks as u128)) / 1000);
+                            debug!("Relayer: will NOT run tenure since issuance at {} is too fresh (wait until {} + {} = {})",
```
gregorycoppola, line 1086: formatting?

gregorycoppola (approved):
LGTM

Can you explain in the description why this change has the effects in the title?

Issue #3064, opened 2022-03-01
Non-critical type-checker bug related to traits


obycode:
**Describe the bug**
The type checker cannot properly handle the case where a trait defined in the same contract is used. The error that it reports is also incorrect.

**Steps To Reproduce**
Consider these two contracts:

bar.clar:
```clarity
(define-public (hello (a int))
    (ok u0)
)
```

test.clar:
```clarity
(define-trait my-trait
    ((hello (int) (response uint uint)))
)
(define-private (pass-trait (a <my-trait>))
    (print a)
)
(define-public (call-it)
    (ok (pass-trait .bar))
)
```

The type-checker will report an error on the call to `pass-trait` indicating the use of an unresolved contract "bar". However, the actual error is that it is unable to find the trait "my-trait", because the type-checker only checks in the analysis database, but since this trait is in the current contract, it is not yet in the database. It could, instead, be found in the `ContractAnalysis`.

The simple workaround is to only use a trait in a separate contract from where it is defined, but I don't think that is an intended restriction. Fixing this would be consensus-breaking, so it should be considered for 2.1, or if this is the intended behavior, it should be documented.

The problem with the error that is printed can be fixed with a patch similar to [this commit in the REPL](https://github.com/hirosystems/clarity-repl/commit/ecfad3a474f7996356eb6140f3262615557334b4).

**Expected behavior**
I expect this contract to pass type-checking.

obycode:
#3065 fixes the problem with the error message.

jcnelson:
Thanks for sending the PR! Going to go ahead and close this.

PR #3065, opened 2022-03-01
Fix trait not found error messages

obycode:
### Description

This PR fixes the bad errors described in #3064.

### Applicable issues
- See #3064 (does not fix the main problem, but fixes the error messages)

### Checklist
- [ ] Changelog is updated


jcnelson (commented):
This LGTM on the surface, but let's confirm with @kantai that no code expects this method to return `NoSuchContract` to indicate the absence of a trait.

kantai:
This looks good to me -- a different `CheckError` isn't a consensus issue here, because while the `chainstate::stacks::db::transactions` module _does_ care about `CheckError`, it doesn't actually care _which_ `CheckError` was returned: it just marks the transaction as invalid in the event of a `contract-call`, and marks the transaction a failed, but valid, in the event of a contract publish (as would be the case with this change).

This PR should include a minimal unit test, though.

obycode:
> This PR should include a minimal unit test, though.

Will do. Where is the right place to put that?

kantai:
I think a new unit test in `clarity/src/vm/analysis/trait_checker/tests.rs` would be the best spot.

obycode:
While attempting to write tests for this, I found that it might be impossible to reach the `NoSuchTrait` error from type_checker/mod.rs:427 because an earlier type check on the function definition would have reported this error. Similarly, the fixed `NoSuchContract` error on line 421 is only reachable due to the bug described in #3064. I added a test for that one.

jcnelson (approved):
LGTM!

jcnelson:
This has two approvals @obycode.  Want to merge it? ;)

obycode:
> This has two approvals @obycode. Want to merge it? ;)

It seems I don't have the power 😄

PR #3075, opened 2022-03-15
Feat/c32 optimizations

zone117x:
Various optimizations in the C32 address decoding paths. Use a decoding table (the inverse of the encoding table/alphabet) in the decoding hot path (previously an O(n) utf8 char scan was used). Reduce some allocs & copies.


stacks-common/src/address/c32.rs
```diff
@@ -69,44 +97,34 @@ fn c32_encode(input_bytes: &[u8]) -> String {
     String::from_utf8(result).unwrap()
 }
 
-fn c32_normalize(input_str: &str) -> String {
-    let norm_str: String = input_str
-        .to_uppercase()
-        .replace("O", "0")
-        .replace("L", "1")
-        .replace("I", "1");
```
zone117x, line 77: Reductions of the char scans and allocs here is one of the primary optimizations in this PR

stacks-common/src/address/c32.rs
```diff
@@ -69,44 +97,34 @@ fn c32_encode(input_bytes: &[u8]) -> String {
...
 
+fn c32_decode_ascii(input_str: &str) -> Result<Vec<u8>, Error> {
     let mut result = vec![];
     let mut carry: u16 = 0;
     let mut carry_bits = 0; // can be up to 5
 
-    let iter_c32_digits_opts: Vec<Option<usize>> = c32_normalize(input_str)
-        .chars()
-        .rev()
-        .map(|x| C32_CHARACTERS.find(x))
```
zone117x, line 94: Refactor of this `find` from O(n) to O(1) in the hot path is one of the primary optimizations in this PR

stacks-common/src/address/c32.rs
```diff
@@ -18,12 +18,40 @@ use super::Error;
...
+///     table[*x as usize] = i as isize;
+/// }
+/// let specials = [('O', '0'), ('L', '1'), ('I', '1')];
+/// for pair in specials {
+///     let i = alphabet.find(|a| a == pair.1).unwrap() as isize;
+///     table[pair.0 as usize] = i;
+///     table[pair.0.to_ascii_lowercase() as usize] = i;
+/// }
+/// ```
+const C32_CHARACTERS_MAP: [i8; 128] = [
```
zone117x, line 45: I experimented with trying both lazy_static and `const fn` for this, neither seemed right, so I took the table result and inlined it. The rust code doc has compilation validation support (at least in the tools I'm using). So seems like this is the best approach. But I'm open to alternatives!

stacks-common/src/address/c32.rs
jcnelson: Where did the normalization code go?

jcnelson (commented):
So, this is all consensus critical code you're touching.  We use the bare addresses in the chainstate MARF, which affects the block state root hashes.  At a minimum, I need to see unit tests demonstrating that this new code behaves *exactly* the same way as the old code.

zone117x:
The existing tests cover these paths pretty extensively, e.g. one for the normalization code:
https://github.com/stacks-network/stacks-blockchain/blob/03107740595e5359f2381dfb97b5ee153125c44a/stacks-common/src/address/c32.rs#L383-L404

I'm happy to add more, but not really seeing any gaps in testing, would appreciate some pointers here.

stacks-common/src/address/c32.rs
zone117x: This normalization fn was no longer called once the encoding logic was changed to use a faster C32 alphabet lookup using the `C32_CHARACTERS_MAP` table. If you take a look at the code there, you'll see where the lowercase and special chars are included. I'm happy to add it back (will be dead code though).

jcnelson:
A couple things:

* Because this PR touches consensus-critical code to implement better performance, I'd like to know what exactly the performance impact is.  If it's not very big in practice, then I'd rather not merge this PR.  It's better to have slightly-slower but known-good code in the consensus-critical paths than fast code that we're less than 100% certain of.  Like, does this PR improve the node bootup speed?  Does it improve p2p or RPC responsiveness?  Does it improve the miner speed?

* If this PR yields a significant improvement in performance, then the test I'd like to see added is to do the following:
   * Keep the old code around, but decorated as `#[cfg(test)]`.
   * Generate lots and lots of addresses with the old code, and verify that they decode via the new code *exactly the same way* as they would in the old code.
   * Generate lots and lots of addresses with the new code, and verify that they decode via the old code *exactly the same way* as they would in the new code.
   
   We do something like this with the MARF PR I have open, for example -- it keeps all the old code paths around, and verifies that the new code paths generate exactly the same trie root hashes as the old code paths given the same key/value insert schedule.

zone117x:
Sounds reasonable, keeping the old code around would also help with generating benchmarks. Btw, this code is in a library "stacks-common" which is used in projects outside the stacks-node, so I wouldn't be surprised if there was a perf gain that doesn't help with the stacks-node operations you've listed, but could be significant for other use cases. Seems like these would still be good things to upstream (assuming the consensus-critical change requirements you've outlined are met).

stacks-common/src/address/mod.rs
```diff
@@ -31,6 +31,7 @@ use std::convert::TryFrom;
 
 pub mod b58;
 pub mod c32;
+pub mod c32_old;
```
pavitthrap, line 34: I think adding `#[cfg(test]` above this mod declaration would make sure the old c32 encoding logic only gets compiled when tests are run.

zone117x: It will also prevent the benchmark from compiling, but that's probably okay?

stacks-common/src/address/c32.rs
```diff
@@ -69,44 +97,34 @@ fn c32_encode(input_bytes: &[u8]) -> String {
...
-
-    let iter_c32_digits: Vec<usize> = iter_c32_digits_opts
+    let iter_c32_digits: Vec<u8> = input_str
+        .as_bytes()
         .iter()
-        .filter_map(|x| x.as_ref())
-        .map(|ref_x| *ref_x)
+        .rev()
+        .filter_map(|x| C32_CHARACTERS_MAP.get(*x as usize))
+        .filter_map(|v| u8::try_from(*v).ok())
```
jcnelson, line 118: This code seems to depend on `u8::try_from(*v)` returning `None` if `*v` is `-1`.  Can this code be more explicit about this?  Such as:

```rust
.filter_map(|v| {
   if *v < 0 {
       None
   }
   else {
       // infallible due to the above check
       Some(*v as u8)
    }
})
```

stacks-common/src/address/c32.rs
```diff
@@ -18,12 +18,40 @@ use super::Error;
...
 use sha2::Sha256;
-
-const C32_CHARACTERS: &str = "0123456789ABCDEFGHJKMNPQRSTVWXYZ";
+use std::convert::TryFrom;
+
+const C32_CHARACTERS: &[u8; 32] = b"0123456789ABCDEFGHJKMNPQRSTVWXYZ";
+
+/// C32 chars as an array, indexed by their ASCII code for O(1) lookups.
+/// Supports lookups by uppercase, lowercase, and special (i.e. `O, L, I`) chars.
+/// Table can be generated with:
```
jcnelson, line 27: Can you add a comment to this docstring to indicate that this table will encode `O`, 'o', and `0` as `0`, and `1, `l`, `L`, 'l', `I`, and `i` as `1`?

stacks-common/src/address/c32_old.rs
```diff
@@ -0,0 +1,233 @@
...
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
```
jcnelson, line 16: Can you add a module-level docstring comment that this module is only here to test compatibility, and will be removed in the next network upgrade?

stacks-common/src/address/mod.rs
jcnelson: Could you use `#[cfg(any(test, feature = "testing"))]` or some such?  Maybe there needs to be a benchmarking build flag set?  I agree with @pavitthrap -- this shouldn't get compiled into a release build.

jcnelson (approved):
This LGTM, but please address my comments before merging.  Thanks!

stacks-common/src/address/mod.rs
kantai: Just a `#[cfg(test)]` is probably the easiest here. The `testing` feature should only really be used if the testing infrastructure needs the decorated module across build components: it is set in `debug` builds, so the code would end up getting included in those, but not `release` builds. Breaking the benchmark from compiling is fine-- this repo pretty regularly breaks benchmark code, which I'm fine with: benchmarking code isn't currently meant to be run regularly, and it's usually easy to fix back up whenever someone wants to rerun benchmarks.

kantai (approved):
This looks good to me, the only comments I think need to be addressed are the `cfg(test)` decoration for the old code, and including the normalization rules in the docstring.

stacks-common/src/address/c32.rs
zone117x: It may actually be better to use an array of `Option(u8)` rather than `-1i8` for the "null" sentinel values (this is what the b58 table does). I'll give that a try when I address the other comments and see if there's any unexpected perf issues with that approach.

zone117x: I changed the table from `[i8; 128]` to `[Option<u8>; 128]` which gets rid of this integer conversion, and the benchmark shows a slight perf increase.

kantai (approved):

PR #3080, opened 2022-03-24
Chore: Set rust edition to 2021

kantai:
### Description

This PR sets an edition for the main Cargo files to 2021, and updates the edition of the `testnet` outputs to 2021.

Previously, by leaving the edition unset, we were implicitly selecting 2015. The newer editions of rust support better lifetime inference and macro definitions. While the codebase won't immediately benefit from this change, it could let us clean up some of our lifetime specifications in the future.

The vast, vast, majority of changes here are simply to the imports. Implicit `crate` imports became illegal in edition 2018. This PR changes the imports to explicitly specify the crate. The only other change is to a macro definition: the HTTP expects macro uses a pattern specifier in a way that became illegal in Rust 2021 -- the port is simple: switch to [`pat_param`](https://doc.rust-lang.org/edition-guide/rust-2021/or-patterns-macro-rules.html).

### Checklist
- [x] Existing test coverage covers all cases.
- [x] Changelog update unneccessary: this is just a build change.
- [x] No documentation changes required


jcnelson (approved):
LGTM!

PR #3081, opened 2022-03-25
Merge `develop` to `next`

jcnelson:
This PR merges `master` and then `develop` to `next`, bringing in the major `libclarity` refactor among other things.  While nearly all of this PR is just related to merging, there are two things I had to modify to get it to work:

* I made it so that in all cases where we need to evaluate a contract AST and did not have the `ClarityVersion` to use on-hand, we could infer it from the epoch.  This is consistent with what's in `next` right now; later on, we'll make it so that there exists a pragma statement in `Clarity2` that will allow the AST parser to infer which Clarity version to use if not the one that is the default in the current epoch.  This pertains to implementations of `ClarityTransactionConnection`.

* I had to move some unit tests into `src/clarity_vm` out of `clarity`, since in `next` they depend on things that can only be found in the blockchain crate (such as MARFs).  In particular, this includes the "simple naming system" and "simple token system" tests, which are now parameterized by Clarity version via `rstest` templates.


jcnelson:
Looks like all the `CI / dist` items are broken.  Looks like a bug somewhere -- the code is trying to download crates, but via a malformed URL (that ends in what appears to be a spurious `%60`), leading to an HTTP 403.

jcnelson:
All the tests pass; it's just the `dist` CI units that are failing.

jcnelson:
Alright, cool -- looks like all `dist` units pass now too.

src/clarity_vm/clarity.rs
```diff
@@ -903,6 +800,9 @@ impl<'a> ClarityBlockConnection<'a> {
                     })
                     .unwrap();
 
+                // require 2.1 rules henceforth in this connection as well
```
gregorycoppola, line 803: So, where are you going to document this?

src/clarity_vm/clarity.rs
```diff
@@ -903,6 +800,9 @@ impl<'a> ClarityBlockConnection<'a> {
...
+                tx_conn.epoch = StacksEpochId::Epoch21;
```
kantai, line 804: This will only set the transaction's epoch, _not_ the block's epoch. `ClarityBlockConnection ::start_transaction_processing` is the common method used to start a transaction, and it sets the `epoch` to the block connection's epoch. So, I think that you need to update `self.epoch`.

jcnelson: Sure, will update both.

jcnelson:
> I don't think this is strictly necessary -- the libclarity component includes an in-memory datastore implementation for testing purposes. In develop, these tests just use with_memory_environment. It doesn't make a huge difference if these are moved, but I do think that they could remain in libclarity.

Sorry, forgot to mention this.  The tests I moved to `clarity_vm` specifically test the use of a `ClarityInstance` to drive the instantiation and execution of code in a simple token and simple name smart contract.  It's testing functionality that doesn't belong to `libclarity`.  They were the `test_simple_token_system()` and `test_simple_naming_system()` tests in `src/vm/tests/contracts.rs` in `next`, which I moved to `src/clarity_vm/tests/large_contract.rs`.  They are similar to, but _not_ identical to the tests in `clarity/src/vm/tests/assets.rs`.

src/clarity_vm/clarity.rs
Nizebudka: - [ ]

Nizebudka (changes requested):
(spam deleted by @jcnelson)

gregorycoppola (commented):
Huge PR, looks good overall.

I just had one last question about the epoch stuff

https://github.com/stacks-network/stacks-blockchain/pull/3081/files#r858098915

kantai (approved):

src/clarity_vm/clarity.rs
```diff
@@ -798,6 +722,10 @@ impl<'a> ClarityBlockConnection<'a> {
                     })
                     .unwrap();
 
+                // NOTE: we don't set tx_conn.epoch to Epoch2_05 here, even though we probably
```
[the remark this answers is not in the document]
jcnelson: We can't change it because doing so could change how a subsequent block gets evaluated, leading to a chain split.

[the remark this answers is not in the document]
gregorycoppola: So then we "shouldn't" set it, no?Either way.

gregorycoppola (approved):
Thanks for carrying 2.1!

PR #3086, opened 2022-04-08
Improved Atlas shutdown and crash tolerance (#3082)

### LINKED ISSUE #3082: Improve AtlasDB crash and shutdown tolerance
One of the issues that #3047 surfaced relates to the resiliency of AtlasDB "attachments" events. These events are pushed from the `chains-coordinator` thread over a synchronous channel to the `p2p` thread, which inserts the attachments into the AtlasDB, and includes any newly-fulfilled attachments in the network result struct. This is pushed to the `relayer` thread, which notifies the event-listener.

When the node shuts down, these events can be lost. Instead, the coordinator thread should be emitting these events in a more resilient way (by writing them to the AtlasDB themselves).

kantai:
### Description

This PR replaces the rx/tx synchronized channel used to broadcast `AttachmentInstance`s from the `chain-coordinator` thread to the `p2p` thread with direct writes to the AtlasDB. This PR preserves the `p2p` thread's behavior -- new attachment instances are written as "queued" by the coordinator, but aren't checked against extant attachment data until the `p2p` thread passes over the queued instances. After this, the behavior is exactly the same as before, with new attachment notices being passed through the network result object.

### Applicable issues
- Fixes #3082

- The amount of queued instances checked is limited by a new const

### Checklist
- Test coverage for the modified code paths should be provided by the existing atlas stress and atlas integration tests.
- Changelog is updated
- Along with the changes, I've added some rustdocs to portions of the atlas code I touched, which explain the intended life cycle of attachment instances


src/net/atlas/db.rs
```diff
@@ -73,9 +87,38 @@ const ATLASDB_INITIAL_SCHEMA: &'static [&'static str] = &[
     "CREATE TABLE db_config(version TEXT NOT NULL);",
 ];
 
+const ATLASDB_SCHEMA_2: &'static [&'static str] = &[
+    r#"
+    ALTER TABLE attachment_instances
+    ADD    status INTEGER NOT NULL
```
jcnelson, line 93: I'm surprised sqlite doesn't force you to provide a default value here.

src/chainstate/coordinator/mod.rs
```diff
@@ -769,15 +769,27 @@ impl<
...
+                            "Atlas: New attachment instances emitted by block";
+                            "attachments_count" => attachments_instances.len(),
+                            "index_block_hash" => %block_receipt.header.index_block_hash(),
+                            "stacks_height" => block_receipt.header.stacks_block_height,
                         );
-                        match self.attachments_tx.send(attachments_instances) {
-                            Ok(_) => {}
-                            Err(e) => {
-                                error!("Atlas: error dispatching attachments {}", e);
+                        if let Some(atlas_db) = self.atlas_db.as_mut() {
```
jcnelson, line 777: Correct me if I'm wrong, but doesn't this call to `queue_attachment_instance()` happen _outside_ of block-processing?  Like, if a node processed a block and then crashed before reaching this code path, would it be able to enqueue the block's attachment data on restart?

jcnelson: Would it be possible to commit to the attachment data from the event stream from within `StacksChainState::append_block()`, right before we commit to the Stacks block?  Enqueuing attachment instances is idempotent, so re-doing them in the event the node crashes in-between saving them and committing the block should be safe.

jcnelson (commented):
Did an initial pass.  My biggest question is whether or not this PR allows the node to recover if it crashes while in the middle of the block.  I can't convince myself that it does -- it seems this saves attachment events _after_ the block that produced them is committed, which I think means that there's a window of time in which the block is saved, but the attachments are not.  This would make it possible for the node to crash in such a way that it loses attachment events.

pavitthrap (approved):
lgtm except +1 on Jude's comment on `append_block`

jcnelson:
This needs to be looked at.

src/chainstate/coordinator/mod.rs
kantai: Yes, but I'd like to do that in a separate PR. It's true that this change doesn't make the atlas queue entirely crash proof, but it removes the 2-thread dependency for the atlas queue, which makes the shutdown behavior much better already.

This PR is already changing a lot of how Atlas works, so I'd prefer to not also move atlas queueing logic into `append_block` as part of the same changeset.

src/net/atlas/tests.rs
```diff
@@ -1160,7 +1178,10 @@ fn test_bit_vectors() {
     let block_id_2 = instantiated_attachment_instances[0].index_block_hash;
     for attachment_instance in instantiated_attachment_instances.iter() {
         atlas_db
-            .insert_uninstantiated_attachment_instance(attachment_instance, true)
+            .queue_attachment_instance(attachment_instance)
+            .unwrap();
+        atlas_db
+            .mark_attachment_instance_checked(attachment_instance, true)
```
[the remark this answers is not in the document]
kantai: No, but I can try to add something along those lines. It won't be a "crash", but instead a stop-and-restart.

src/net/atlas/download.rs
```diff
@@ -235,74 +230,57 @@ impl AttachmentsDownloader {
         Ok((resolved_attachments, events_to_deregister))
     }
 
-    pub fn enqueue_new_attachments(
+    fn check_attachment_instances<F, G>(
```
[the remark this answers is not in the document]
kantai: Yep, added in d9a5e1852

src/net/atlas/db.rs
```diff
@@ -73,9 +87,38 @@ const ATLASDB_INITIAL_SCHEMA: &'static [&'static str] = &[
...
+    ;"#,
+    r#"
+    UPDATE attachment_instances SET status = 2;
```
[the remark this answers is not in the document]
kantai: Sure, added in d9a5e1852

src/net/atlas/tests.rs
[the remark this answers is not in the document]
kantai: Added in d3bf2911b: `chainstate::coordinator::tests::atlas_stop_start`

kantai:
Okay -- this PR should be ready for re-review!

src/net/atlas/db.rs
```diff
@@ -462,31 +651,82 @@ impl AtlasDB {
         Ok(row)
     }
 
-    pub fn insert_uninstantiated_attachment_instance(
+    /// Queue a new attachment instance, status will be set to "queued",
+    /// and the is_available field set to false
+    pub fn queue_attachment_instance(
```
jcnelson, line 656: Can you document which thread calls this function?  Same for `insert_initial_attachment_instance`.

src/net/atlas/download.rs
```diff
@@ -303,6 +303,55 @@ impl AttachmentsDownloader {
 
         Ok(resolved_attachments)
     }
+
+    /// Check any queued attachment instances to see if we already have data for them,
+    ///  returning a vector of (instance, attachment) pairs for any of the queued attachments
+    ///  which already had the associated data
+    /// Marks any processed attachments as checked
```
jcnelson, line 310: Which thread calls this function?

testnet/stacks-node/src/node.rs
```diff
@@ -550,9 +544,8 @@ impl Node {
             }
             tx.commit().unwrap();
         }
-        let atlas_config = AtlasConfig::default(false);
-        let atlasdb =
-            AtlasDB::connect(atlas_config, &self.config.get_atlas_db_file_path(), true).unwrap();
+        let _atlas_config = Self::make_atlas_config();
```
jcnelson, line 547: Is this used somewhere?

src/net/atlas/db.rs
```diff
@@ -73,9 +87,43 @@ const ATLASDB_INITIAL_SCHEMA: &'static [&'static str] = &[
...
+/// Attachment instances pass through different states once written to the AtlasDB.
+/// These instances are initially written as a new Stacks block is processed, and marked
+/// as `Queued`. These queued instances contain all the data of the new attachment instance,
+/// but they have not yet been checked against the AtlasDB to determine if there is extant
+/// Attachment content/data associated with them. The network run loop (`p2p` thread) checks
+/// for any queued attachment instances on each pass, and performs that check. Once the check
+/// is completed, any checked instances are updated to `Checked`.
+pub enum AttachmentInstanceStatus {
+    /// This variant indicates that the attachments instance has been written,
+    /// but the downloader has not yet checked that the attachment matched
```
jcnelson, line 120: By "downloader" you mean the "downloader in the p2p thread", right?

src/net/p2p.rs
```diff
@@ -5393,7 +5392,7 @@ impl PeerNetwork {
         // enqueue them.
         PeerNetwork::with_attachments_downloader(self, |network, attachments_downloader| {
             let mut known_attachments = attachments_downloader
-                .enqueue_new_attachments(attachment_requests, &mut network.atlasdb, false)
+                .check_queued_attachment_instances(&mut network.atlasdb)
```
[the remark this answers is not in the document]
jcnelson: It's fine if you don't do this in this PR.  If not, then please open an issue and assign it to me to do this.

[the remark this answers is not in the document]
kantai: I'll open an issue for that -- it would entail moving the atlas downloader from the PeerNetwork struct into the Relayer struct.

src/net/atlas/db.rs
kantai: Yes, the `AttachmentsDownloader`, will update the comment to make more clear

src/net/atlas/download.rs
kantai: This method is invoked in the thread managing the AttachmentDownloader. This is currently the P2P thread. Will add to the comment.

src/net/atlas/db.rs
kantai: Sure, however, I'm somewhat hesitant to document _the caller_ from the called method. If atlas event processing is moved from the coordinator's purview to `db::blocks` or whatever, that move can occur without ever touching this function, which could easily lead to a stale rustdoc here.

src/net/atlas/db.rs
```diff
@@ -47,7 +52,16 @@ use crate::types::chainstate::StacksBlockId;
 
 use super::{AtlasConfig, Attachment, AttachmentInstance};
 
-pub const ATLASDB_VERSION: &'static str = "1";
+pub const ATLASDB_VERSION: &'static str = "2";
```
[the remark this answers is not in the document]
kantai: Added module level rustdocs (and more information to the other rustdocs) that talks more about this (https://github.com/stacks-network/stacks-blockchain/pull/3086/commits/cb9f8dca0f9697e7cfe04c1b2e776de4b8aec8ce)

"Availability" and "checked" have little to do with each other (except that the availability field has no meaning until an attachment has been checked). "is available" means that the atlas content is already stored in the stacks-node's atlas database. "checked" means that the atlas downloader has acknowledged that a new content hash has been added to the system (i.e., a new attachment instance). So, "(unavailable, checked)" is indeed possible and part of the normal flow of the system: it just means that the atlas downloader hasn't downloaded the corresponding content yet.

testnet/stacks-node/src/node.rs
kantai: Deleted in https://github.com/stacks-network/stacks-blockchain/pull/3086/commits/cb9f8dca0f9697e7cfe04c1b2e776de4b8aec8ce

src/net/atlas/download.rs
kantai: Added in https://github.com/stacks-network/stacks-blockchain/pull/3086/commits/cb9f8dca0f9697e7cfe04c1b2e776de4b8aec8ce

src/net/atlas/db.rs
kantai: Added in https://github.com/stacks-network/stacks-blockchain/pull/3086/commits/cb9f8dca0f9697e7cfe04c1b2e776de4b8aec8ce

src/net/atlas/db.rs
kantai: Added in https://github.com/stacks-network/stacks-blockchain/pull/3086/commits/cb9f8dca0f9697e7cfe04c1b2e776de4b8aec8ce

pavitthrap (approved):
lgtm

src/net/atlas/db.rs
```diff
@@ -73,9 +106,43 @@ const ATLASDB_INITIAL_SCHEMA: &'static [&'static str] = &[
...
+    ALTER TABLE attachment_instances
+    ADD status INTEGER
+    ;"#,
+    // All of the attachment instances that previously existed in the database
+    //  already were "checked", so set status to 2 (which corresponds to "checked").
+    r#"
+    UPDATE attachment_instances SET status = 2;
+    "#,
+    r#"
+    CREATE INDEX IF NOT EXISTS index_instance_status ON attachment_instances(status);
```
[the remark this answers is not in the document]
kantai: Added in [9b97a39](https://github.com/stacks-network/stacks-blockchain/pull/3086/commits/9b97a39191133a53cfdcb07cccceebbab99246f7)

src/net/atlas/db.rs
```diff
@@ -85,6 +152,7 @@ impl FromRow<Attachment> for Attachment {
 
 impl FromRow<AttachmentInstance> for AttachmentInstance {
     fn from_row<'a>(row: &'a Row) -> Result<AttachmentInstance, db_error> {
+        eprintln!("{:?}", row.get_raw("index_block_hash"));
```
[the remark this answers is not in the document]
kantai: Removed in [9b97a39](https://github.com/stacks-network/stacks-blockchain/pull/3086/commits/9b97a39191133a53cfdcb07cccceebbab99246f7)

src/net/atlas/db.rs
```diff
@@ -224,11 +334,65 @@ impl AtlasDB {
...
     }
 
+    fn check_schema_version_or_error(&mut self) -> Result<(), db_error> {
+        match Self::get_schema_version(self.conn()) {
+            Ok(version) => {
+                let expected_version = ATLASDB_VERSION.to_string();
+                if version == expected_version {
+                    Ok(())
+                } else {
+                    Err(db_error::Other(format!(
```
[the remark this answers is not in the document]
kantai: Updated in [9b97a39](https://github.com/stacks-network/stacks-blockchain/pull/3086/commits/9b97a39191133a53cfdcb07cccceebbab99246f7)

src/net/atlas/tests.rs
```diff
@@ -774,6 +775,89 @@ fn test_keep_uninstantiated_attachments() {
...
+    // perform the migration and unwrap() to assert that it runs okay
+    let atlas_db = AtlasDB::connect_with_sqlconn(atlas_config, conn).unwrap();
+
+    assert_eq!(
+        atlas_db
+            .find_all_attachment_instances(&Hash160([0xa0; 20]))
+            .unwrap()
+            .len(),
+        1,
+        "Should have one attachment instance marked 'checked' with hash `0xa0a0a0..`"
```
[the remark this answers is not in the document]
kantai: Added in [9b97a39](https://github.com/stacks-network/stacks-blockchain/pull/3086/commits/9b97a39191133a53cfdcb07cccceebbab99246f7)

jcnelson (approved):
Thanks @kantai! LGTM

PR #3090, opened 2022-04-13
Utility commands to analyze fees and get tenure transactions

jcnelson:
This small patch adds `get-tenure` and `analyze-fees` commands to the `blockstack-core` utility binary.  The `get-tenure` command just loads up the confirmed microblock stream and block, given an index block hash, and prints out their transactions and fees as a JSON blob.  The `analyze-fees` command looks at the last _B_ blocks on the canonical fork, and finds out which transactions in the mempool have taken _N_ sortitions to get mined.  For each set of transactions mined after _N_ blocks, it calculates a simple distribution of transaction fees (i.e. min, max, avg, median, p90, p95, p99).  This command is meant to be the beginnings of something like bitcoinfees.earn.com, whereby a user can determine what fee to set for a transaction if they want it confirmed within _N_ blocks.

No tests are added to this PR, since `blockstack-core` isn't (yet) user-facing software, but instead only serves as a Swiss army knife of low-level commands to use to diagnose problems and query chainstate.


gregorycoppola (approved):
toolz

src/main.rs
```diff
@@ -222,6 +226,257 @@ fn main() {
...
+        }
+
+        let mut deltas: Vec<_> = tx_mined_deltas.keys().collect();
+        deltas.sort();
+
+        let mut reports = vec![];
+        for delta in deltas {
+            let mut delta_tx_fees = vec![];
+            let empty_txids = vec![];
+            let txids = tx_mined_deltas.get(&delta).unwrap_or(&empty_txids);
```
pavitthrap, line 442: if there are no txids, maybe you should just `continue`. 

otherwise you might run into problems in computing stats. ex, `let median_tx_fee = delta_tx_fees[delta_tx_fees.len() / 2]` would panic

pavitthrap (commented):
lgtm, one minor comment

kantai (approved):
LGTM, once you address @pavitthrap's comment

PR #3104, opened 2022-04-18
fix: set the default subsequent-attempt time to 30s

jcnelson:
Fixes https://github.com/stacks-network/stacks-blockchain/issues/3098.  Sets the new time to 30s, based on real-world benchmarks with `develop`.


gregorycoppola (approved):
defer to you guys on what time settings to use

kantai (approved):

PR #3109, opened 2022-04-19
fix: CI error while building dist files

zone117x:
A recent cargo/rust change made the build arg `--workspace` error if a value was provided, since then all platform dist builds have been broken in CI with an error message like:
```
#9 ERROR: process "/bin/sh -c cargo build --release --workspace=./ --target x86_64-unknown-linux-gnu" did not complete successfully: exit code: 1
------
 > [build 6/7] RUN cargo build --release --workspace=./ --target x86_64-unknown-linux-gnu:
#0 0.080 error: The value '=./' was provided to '--workspace' but it wasn't expecting any more values
```
This PR should fix that.


kantai (approved):
LGTM! Thanks for catching this @zone117x

jcnelson (approved):
Thanks @zone117x !

PR #3110, opened 2022-04-21
Fix/utxo chaining

jcnelson:
This fixes miner UTXO chaining (#2358), and adds unit tests and integration tests to verify that missed block commits are now included in a miner's UTXO chain when determining its sortition weight.

Nearly all of this new code is tests and test infrastructure.  The change itself is less than 20 lines long.

### DIFF
--- a/src/chainstate/burn/operations/leader_block_commit.rs
+++ b/src/chainstate/burn/operations/leader_block_commit.rs
@@ -691,12 +691,11 @@ impl LeaderBlockCommitOp {
             let intended_sortition = match epoch.epoch_id {
                 StacksEpochId::Epoch21 => {
                     // correct behavior
-                    if self.block_height < burnchain.first_block_height {
-                        return Err(op_error::BlockCommitPredatesGenesis);
-                    }
                     let sortition_height = self
                         .block_height
-                        .saturating_sub(burnchain.first_block_height);
+                        .checked_sub(burnchain.first_block_height)
+                        .ok_or_else(|| op_error::BlockCommitPredatesGenesis)?;
+
                     if miss_distance > sortition_height {
                         return Err(op_error::BlockCommitBadModulus);
                     }
--- a/src/chainstate/burn/operations/leader_block_commit.rs
+++ b/src/chainstate/burn/operations/leader_block_commit.rs
@@ -35,8 +35,8 @@ use crate::chainstate::burn::SortitionId;
 use crate::chainstate::stacks::index::storage::TrieFileStorage;
 use crate::chainstate::stacks::{StacksPrivateKey, StacksPublicKey};
 use crate::codec::{write_next, Error as codec_error, StacksMessageCodec};
-use crate::core::STACKS_EPOCH_2_05_MARKER;
 use crate::core::{StacksEpoch, StacksEpochId};
+use crate::core::{STACKS_EPOCH_2_05_MARKER, STACKS_EPOCH_2_1_MARKER};
 use crate::net::Error as net_error;
 use crate::types::chainstate::TrieHash;
 use crate::types::chainstate::{BlockHeaderHash, BurnchainHeaderHash, StacksAddress, VRFSeed};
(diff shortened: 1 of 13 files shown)


testnet/stacks-node/src/neon_node.rs
```diff
@@ -2120,13 +2160,52 @@ impl StacksNode {
...
 
-        let res = bitcoin_controller.submit_operation(op, &mut op_signer, attempt);
-        if !res {
-            if !config.node.mock_mining {
-                warn!("Failed to submit Bitcoin transaction");
-                return None;
-            } else {
-                debug!("Mock-mining enabled; not sending Bitcoin transaction");
+        let mut send_tx = true;
+        if cfg!(test) {
```
gregorycoppola, line 2164: I think this function is already doing a lot for the reader to follow, and they will now have to keep track of a new long cryptic `cfg!(test)` block.

any way to inject from the outside without putting it into the prod code using some kind of "for test" version of an interface?

an alternative would be to make a function in this file called `maybe_test_late_block_commit`.

at the end of the day if that's not worth it and you want to check this in it's fine with me.

src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -679,17 +679,43 @@ impl LeaderBlockCommitOp {
...
             };
-            if miss_distance > self.block_height {
-                return Err(op_error::BlockCommitBadModulus);
-            }
-            let intended_sortition = tx
-                .get_ancestor_block_hash(self.block_height - miss_distance, &tx_tip)?
-                .ok_or_else(|| op_error::BlockCommitNoParent)?;
+            let intended_sortition = match epoch.epoch_id {
+                StacksEpochId::Epoch21 => {
+                    // correct behavior
```
gregorycoppola, line 693: maybe "correct behavior, fixes issues/2358"

gregorycoppola (commented):
one idea to consider but otherwise LGTM if the calculations are really correct :)

src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -679,17 +679,43 @@ impl LeaderBlockCommitOp {
...
+                    if self.block_height < burnchain.first_block_height {
+                        return Err(op_error::BlockCommitPredatesGenesis);
+                    }
+                    let sortition_height = self
+                        .block_height
+                        .saturating_sub(burnchain.first_block_height);
```
kantai, line 699:
```suggestion
                    let sortition_height = self
                        .block_height
                        .checked_sub(burnchain.first_block_height)
                        .ok_or_else(|| op_error::BlockCommitPredatesGenesis)?;
```

Using `checked_sub` is more idiomatic than a manual check + saturating sub.

testnet/stacks-node/src/neon_node.rs
kantai: I agree with this -- this function is already quite complex, and it seems like this is something that would be pretty easy to at split out to another function:

```rust
#[cfg(test)]
fn test_commit_fault_injection(op, op_signer, attempt, bitcoin_controller, block_height) -> bool {
...
}
#[cfg(not(test))]
fn test_commit_fault_injection(op, op_signer, attempt, bitcoin_controller, block_height) -> bool {
   true
}
...
let send_commit = test_commit_fault_injection(op, op_signer, attempt, bitcoin_controller, block_height);
if send_commit {
...
}
...
```

kantai (approved):
This LGTM, just some superficial comments.

testnet/stacks-node/src/neon_node.rs
```diff
@@ -70,8 +71,112 @@ use super::{BurnchainController, BurnchainTip, Config, EventDispatcher, Keychain
...
+#[cfg(not(test))]
+fn fault_injection_delay_transactions(
+    _bitcoin_controller: &mut BitcoinRegtestController,
+    _cur_burn_chain_height: u64,
+    _stacks_block_burn_height: u64,
+    _op: &BlockstackOperationType,
+    _op_signer: &mut BurnchainOpSigner,
+    _attempt: u64,
+) -> bool {
+    false
```
kantai, line 177: Shouldn't this be `true`?

testnet/stacks-node/src/operations.rs
```diff
@@ -36,12 +36,16 @@ impl BurnchainOpSigner {
 
     pub fn sign_message(&mut self, hash: &[u8]) -> Option<MessageSignature> {
         if self.is_disposed {
+            debug!("Signer is dsposed");
```
kantai, line 39: Typo

testnet/stacks-node/src/neon_node.rs
jcnelson: Yup, fixed

testnet/stacks-node/src/operations.rs
jcnelson: Fixed

kantai (approved):

gregorycoppola (approved):

PR #3112, opened 2022-04-22
Bootup fixes for 2.05.0.2.0-rc1

jcnelson:
This PR fixes a couple of boot-up issues related to schema migrations and always-allowed peers when booting a 2.05.0.2.0-rc1 peer.  In particular:

* This PR fixes a runtime panic that arises when booting from genesis due to the unavailability of the first block snapshot.

* This PR fixes a runtime panic that arises when booting from non-migrated chainstate due to needing the first block snapshot before attempting migrations to the latest schemas.

* This PR makes it so that the public key of an always-allowed peer is reset from the config file on each boot-up.

* This PR makes it so that an always-allowed peer is only considered to have timed out if it has replied to a previous message first (this is specific to always-allowed peers since they are always considered authenticated, which precludes relying on checking whether or not the public key is known to determine if a peer has been contacted).


gregorycoppola (approved):
LGTM assuming the node is really booting up :)

kantai (approved):

PR #3119, opened 2022-04-28
cleanup: added tests + updated docs for burn-related Clarity functions

pavitthrap:
### Description
As I went through to add withdraw related Clarity functions for the hyperchains, I noticed some missing test cases for burn-related Clarity functions. I also updated the docs for these functions. 

### Applicable issues
- N/A

### DIFF
--- a/clarity/src/vm/functions/assets.rs
+++ b/clarity/src/vm/functions/assets.rs
@@ -840,7 +840,7 @@ pub fn special_burn_token(
 
     if let (Value::UInt(amount), Value::Principal(ref burner)) = (amount, from) {
         if amount == 0 {
-            return clarity_ecode!(MintTokenErrorCodes::NON_POSITIVE_AMOUNT);
+            return clarity_ecode!(BurnTokenErrorCodes::NON_POSITIVE_AMOUNT);
         }
 
         let burner_bal = env.global_context.database.get_ft_balance(
--- a/clarity/src/vm/functions/assets.rs
+++ b/clarity/src/vm/functions/assets.rs
@@ -54,8 +54,7 @@ enum BurnAssetErrorCodes {
     DOES_NOT_EXIST = 3,
 }
 enum BurnTokenErrorCodes {
-    NOT_ENOUGH_BALANCE = 1,
-    NON_POSITIVE_AMOUNT = 3,
+    NOT_ENOUGH_BALANCE_OR_NON_POSITIVE = 1,
 }
 
 enum StxErrorCodes {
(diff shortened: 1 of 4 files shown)


clarity/src/vm/functions/assets.rs
```diff
@@ -840,7 +840,7 @@ pub fn special_burn_token(
 
     if let (Value::UInt(amount), Value::Principal(ref burner)) = (amount, from) {
         if amount == 0 {
-            return clarity_ecode!(MintTokenErrorCodes::NON_POSITIVE_AMOUNT);
+            return clarity_ecode!(BurnTokenErrorCodes::NON_POSITIVE_AMOUNT);
```
kantai, line 843: I think this would be a consensus-breaking change: `MintTokenErrorCodes::NON_POSITIVE_AMOUNT` is `1`, while `BurnTokenErrorCodes::NON_POSITIVE_AMOUNT` is `3`.

I think the right thing to do here is to redefine `BurnTokenErrorCodes` as:

```rust
enum BurnTokenErrorCodes {
    NOT_ENOUGH_BALANCE_OR_NON_POSITIVE = 1,
}
```

And use that variant in both places for the `BurnTokenErrorCodes`.

kantai (commented):
Nice! I think this PR discovered a difference between the originally intended error codes for "burn token" and the ones that ended up getting used. Your doc updates look good to me (except that the `BURN_TOKEN` doc should use `(err u1)` for all error cases), but the change in `functions/assets.rs` need to be reverted, because we can't change the error codes for that method without a consensus-breaking change.

kantai (approved):
LGTM!

clarity/src/vm/functions/assets.rs
```diff
@@ -840,7 +839,7 @@ pub fn special_burn_token(
 
     if let (Value::UInt(amount), Value::Principal(ref burner)) = (amount, from) {
         if amount == 0 {
-            return clarity_ecode!(MintTokenErrorCodes::NON_POSITIVE_AMOUNT);
+            return clarity_ecode!(BurnTokenErrorCodes::NOT_ENOUGH_BALANCE_OR_NON_POSITIVE);
```
jcnelson, line 842: Wait, isn't this consensus-breaking?

jcnelson (changes requested):
Pretty sure there's still a consensus break here.

jcnelson (approved):
LGTM @pavitthrap.  Thanks for clarifying @kantai!

PR #3122, opened 2022-05-01
fix: incrementally vacuum the MARF DB when migrating

### LINKED ISSUE #3121: Manage disk space in MARF sqlite blobs -> external file migration
The current MARF sqlite blobs to external file migration code effectively doubles the space usage of the MARF. This isn't for lack of trying, though. In `inner_write_external_trie_blob` (in `trie_sql.rs`):

```rust
        let empty_blob: &[u8] = &[];
        let args: &[&dyn ToSql] = &[
            block_hash,
            &empty_blob,
            &0,
            &u64_to_sql(offset)?,
            &u64_to_sql(length)?,
            &block_id,
        ];
        let mut s =
            conn.prepare("UPDATE marf_data SET block_hash = ?1, data = ?2, unconfirmed = ?3, external_offset = ?4, external_length = ?5 WHERE block_id = ?6")?;
        s.execute(args)?;
```

The data blobs get zeroed. _However_ SQLite won't actually remove the blob data from the db file: it just fills it with zeroes (in almost all cases). This means that the migration currently inflates the MARF storage by ~2x. Performing periodic [vacuums](https://www.sqlite.org/lang_vacuum.html) during the migration would help eliminate this issue.

Note: in Hiro's testing of the migration, running out of disk space caused crashes during migration (which corrupted the chain state).

jcnelson:
This fixes #3121 by making it so the MARF migration code to move trie blobs to a .blobs file will vacuum the DB periodically during the migration (i.e. every 4096 blocks).  I've confirmed this works on a live setting -- the database file now shrinks over the course of the migration.

### DIFF
--- a/src/chainstate/stacks/index/file.rs
+++ b/src/chainstate/stacks/index/file.rs
@@ -192,7 +192,8 @@ impl TrieFile {
         Ok(buf)
     }
 
-    /// Copy the trie blobs out of a sqlite3 DB into their own file
+    /// Copy the trie blobs out of a sqlite3 DB into their own file.
+    /// NOTE: this is *not* thread-safe.  Do not call while the DB is being used by another thread.
     pub fn export_trie_blobs<T: MarfTrieId>(
         &mut self,
         db: &Connection,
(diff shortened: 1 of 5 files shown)


jcnelson:
One TODO item on this PR is to also set the `SQLITE_TMPDIR` pragma on the connection so that the temporary vacuum file will be created in the _same directory_ as the database file.  Otherwise, if the user keeps their chain state on a separate disk (which we do at the Foundation) will get an unexpected "disk full" failure followed by a runtime panic.

jcnelson:
* Highlight this in the changelog -- you'll need 2x the space required to migrate

* You can't recover (yet) from a failed migration

* Add a sanity check to restart a failed migration (or abort)

* Only vacuum once

src/chainstate/stacks/index/file.rs
```diff
@@ -191,14 +192,42 @@ impl TrieFile {
...
             "Migrate {} blocks to external blob storage at {}",
             max_block,
             &self.get_path()
         );
+
+        // vacuum the DB every so often when running this, since the DB file won't shrink
+        // otherwise.
+        let vacuum_frequency = 1024;
+
+        // set SQLITE_TMPDIR if it isn't set ailready
```
pavitthrap, line 213: s/ailready/already

pavitthrap (approved):

PR #3126, opened 2022-05-05
Feat/bitcoin wire format rigidity

jcnelson:
This PR addresses #2094 by making it so that when we process a Stacks block, we search the 6 prior burnchain blocks for unprocessed `TransferStx` and `StackStx` transactions and try to apply them in this block.  This tracking is handled by mapping each Stacks block's index block hash to the list of burnchain transaction IDs it handled.  Then, to process a new Stacks block, we simply do the following:

1.  find all burnchain operations in the past 6 burnchain blocks
2.  find all burnchain operations processed in the past 6 Stacks blocks (which covers _at least_ 6 burnchain blocks in the past)
3.  remove all burnchain operations in (2) from the list in (1)
4.  process the remaining operations from (3) in blockchain order

### DIFF
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -171,6 +171,8 @@ pub struct SetupBlockResult<'a, 'b> {
         Option<(MinerReward, Vec<MinerReward>, MinerReward, MinerRewardInfo)>,
     pub evaluated_epoch: StacksEpochId,
     pub applied_epoch_transition: bool,
+    pub burn_stack_stx_ops: Vec<StackStxOp>,
+    pub burn_transfer_stx_ops: Vec<TransferStxOp>,
 }
 
 pub struct DummyEventDispatcher;
@@ -4041,15 +4043,15 @@ impl StacksChainState {
         block: &StacksBlock,
         microblocks: &Vec<StacksMicroblock>,
     ) -> Result<(), Error> {
-        let parent_sn = {
-            let db_handle = sort_ic.as_handle(&snapshot.sortition_id);
-            let sn = match db_handle.get_block_snapshot(&snapshot.parent_burn_header_hash)? {
-                Some(sn) => sn,
-                None => {
-                    return Err(Error::NoSuchBlockError);
-                }
-            };
-            sn
+        let parent_sn = match SortitionDB::get_block_snapshot_for_winning_stacks_block(
+            sort_ic,
+            &snapshot.sortition_id,
+            &block.header.parent_block,
+        )? {
+            Some(sn) => sn,
+            None => {
+                return Err(Error::NoSuchBlockError);
+            }
         };
 
         self.preprocess_anchored_block(
@@ -4599,6 +4601,8 @@ impl StacksChainState {
                                    "txid" => %txid,
                                    "burn_block" => %burn_header_hash,
                                    "contract_call_ecode" => %resp.data);
+                        } else {
+                            debug!("Processed StackStx burn op for {} uSTX for {} cycles starting at {} from {} to pay out to {}", stacked_ustx, num_cycles, block_height, &sender, &reward_addr);
                         }
                         let mut execution_cost = clarity_tx.cost_so_far();
                         execution_cost
@@ -4667,17 +4671,23 @@ impl StacksChainState {
                             )
                         });
                         match result {
-                            Ok((value, _, events)) => Some(StacksTransactionReceipt {
-                                transaction: TransactionOrigin::Burn(txid),
-                                events,
-                                result: value,
-                                post_condition_aborted: false,
-                                stx_burned: 0,
-                                contract_analysis: None,
-                                execution_cost: ExecutionCost::zero(),
-                                microblock_header: None,
-                                tx_index: 0,
-                            }),
+                            Ok((value, _, events)) => {
+                                debug!(
+                                    "Processed TransferStx {} uSTX from {} to {}",
+                                    transfered_ustx, &sender, &recipient
+                                );
+                                Some(StacksTransactionReceipt {
+                                    transaction: TransactionOrigin::Burn(txid),
+                                    events,
+                                    result: value,
+                                    post_condition_aborted: false,
+                                    stx_burned: 0,
+                                    contract_analysis: None,
+                                    execution_cost: ExecutionCost::zero(),
+                                    microblock_header: None,
+                                    tx_index: 0,
+                                })
+                            }
                             Err(e) => {
                                 info!("TransferStx burn op processing error.";
                               "error" => ?e,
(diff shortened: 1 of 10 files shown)


gregorycoppola (commented):
clarification: Why/when are `TransferSTX` and `StackSTX` written to the burnchain? I actually didn't realize that was happening.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -11080,6 +11233,557 @@ pub mod test {
...
+            1,
+            &vec![StacksPublicKey::from_private(&recipient_privk)],
+        )
+        .unwrap();
+
+        let initial_balance = 1000000000;
+        peer_config.initial_balances = vec![(addr.to_account_principal(), initial_balance)];
+        peer_config.epochs = Some(StacksEpoch::unit_test_2_1(0));
+
+        let recv_addr =
```
pavitthrap, line 11290: unused

src/chainstate/stacks/db/blocks.rs
```diff
@@ -11080,6 +11233,557 @@ pub mod test {
...
```
pavitthrap, line 11523: also unused

pavitthrap (approved):
lgtm

jcnelson:
> clarification: Why/when are TransferSTX and StackSTX written to the burnchain? I actually didn't realize that was happening.

Per SIP-007, these operations exist in order to ensure that Stacks miners will pay a high price to attempt to censor them.  If these operations did not exist, then Stacks miners could simply choose to ignore these transactions in the Stacks mempool.  But because a Stacks block is only valid if it includes the burnchain transactions in the same burnchain that sortitioned it, the cost of censoring these transactions is the cost of foregoing a Stacks block reward (which is much higher than the cost of the BTC transaction).

The reason we're worried about miner censorship is because there are scenarios where miners can be incentivized to block them.  Miners are not only potential stackers, whose income from PoX is influenced to other stackers' stacking, but also are able to take bribes from other stackers in order to keep the stacking minimum down (and thus accrue more BTC).  By making censorship very expensive for miners, we make it costly for miners to deviate from honest mining behavior in this matter.

This PR not only makes on-burnchain transfer and stacking operations less brittle and more likely to get confirmed by the network, but also significantly increases the cost of censorship by doing so.  Miners would need to forego 6x the block reward to block an on-burnchain stack-stx or transfer-stx operation.

src/chainstate/stacks/db/headers.rs
```diff
@@ -331,4 +331,33 @@ impl StacksChainState {
             .map_err(|e| Error::DBError(db_error::SqliteError(e)))?
             .is_some())
     }
+
+    /// Load up the past N ancestors' index block hashes of a given block, *including* the given
```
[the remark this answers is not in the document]
jcnelson: The comment is correct.  The given argument `index_block_hash` is included in the returned list of `StacksBlockId`s.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -11080,6 +11233,557 @@ pub mod test {
         );
     }
 
+    fn make_transfer_op(
```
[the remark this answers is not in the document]
jcnelson: It's already in a `mod test { ... }` block decorated with `#[cfg(test)]`.

src/chainstate/stacks/db/blocks.rs
jcnelson: Fixed

src/chainstate/stacks/db/blocks.rs
jcnelson: Fixed

src/chainstate/stacks/db/blocks.rs
```diff
@@ -11080,6 +11233,557 @@ pub mod test {
...
+                tenure_id as u64,
+            ),
+            vtxindex: (10 + tenure_id) as u32,
+            block_height: burn_height,
+            burn_header_hash: BurnchainHeaderHash([0x00; 32]),
+        };
+        transfer_op
+    }
+
+    #[test]
```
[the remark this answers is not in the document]
jcnelson: Fixed

testnet/stacks-node/src/tests/epoch_21.rs
```diff
@@ -456,3 +468,448 @@ fn transition_adds_burn_block_height() {
...
+            &mut miner_signer,
+            1
+        ),
+        "Pre-stx operation should submit successfully"
+    );
+
+    // mine it
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    // let's fire off our transfer op that will not land in a sortition pre-2.1
```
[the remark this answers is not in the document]
jcnelson: Fixed

jcnelson:
@kantai @gregorycoppola Need at least one of you to review please :)

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4599,6 +4601,8 @@ impl StacksChainState {
                                    "txid" => %txid,
                                    "burn_block" => %burn_header_hash,
                                    "contract_call_ecode" => %resp.data);
+                        } else {
+                            debug!("Processed StackStx burn op for {} uSTX for {} cycles starting at {} from {} to pay out to {}", stacked_ustx, num_cycles, block_height, &sender, &reward_addr);
```
kantai, line 4605: All of those fields should be k-v entries:

```suggestion
                            debug!("Processed StackStx burnchain op"; "amount_ustx" => stacked_ustx, "num_cycles" => num_cycles, "burn_block_height" => block_height, "sender" => %sender, "reward_addr" => %reward_addr);
```

kantai: Would be useful to add the `txid` as well

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4667,17 +4671,23 @@ impl StacksChainState {
...
-                                contract_analysis: None,
-                                execution_cost: ExecutionCost::zero(),
-                                microblock_header: None,
-                                tx_index: 0,
-                            }),
+                            Ok((value, _, events)) => {
+                                debug!(
+                                    "Processed TransferStx {} uSTX from {} to {}",
+                                    transfered_ustx, &sender, &recipient
+                                );
```
kantai, line 4678: Same comment as above: this should be a k-v log, with txid.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4859,6 +4869,135 @@ impl StacksChainState {
...
+        // go from oldest burn header hash to newest
+        for ancestor_bhh in ancestor_burnchain_header_hashes.iter().rev() {
+            let stacking_ops = SortitionDB::get_stack_stx_ops(sortdb_conn, ancestor_bhh)?;
+            let transfer_ops = SortitionDB::get_transfer_stx_ops(sortdb_conn, ancestor_bhh)?;
+
+            for stacking_op in stacking_ops.into_iter() {
+                if processed_burnchain_txids.contains(&stacking_op.txid) {
+                    continue;
+                }
+                all_stacking_burn_ops.push(stacking_op);
```
kantai, line 4936: The control flow is simpler if the `if` statement checks for exclusion:

```suggestion
                if !processed_burnchain_txids.contains(&stacking_op.txid) {
                    all_stacking_burn_ops.push(stacking_op);
                }
```

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4859,6 +4869,135 @@ impl StacksChainState {
...
+        sortdb_conn: &Connection,
+        burn_tip: &BurnchainHeaderHash,
+    ) -> Result<(Vec<StackStxOp>, Vec<TransferStxOp>), Error> {
+        let stacking_burn_ops = SortitionDB::get_stack_stx_ops(sortdb_conn, burn_tip)?;
+        let transfer_burn_ops = SortitionDB::get_transfer_stx_ops(sortdb_conn, burn_tip)?;
+        Ok((stacking_burn_ops, transfer_burn_ops))
+    }
+
+    fn get_stacking_and_transfer_burn_ops_v210<'b>(
+        chainstate_tx: &'b mut ChainstateTx,
```
kantai, line 4882: I don't think the lifetime generic is necessary here: rust can almost always figure out the lifetimes for references that don't land in the return type.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4859,6 +4869,135 @@ impl StacksChainState {
...
+    /// The change in Stacks 2.1+ makes it so that it's overwhelmingly likely to work
+    /// the first time -- the choice of K is significantly bigger than the length of short-lived
+    /// forks or periods of time with no sortition than have been observed in practice.
+    fn get_stacking_and_transfer_burn_ops<'b>(
+        chainstate_tx: &'b mut ChainstateTx,
+        parent_index_hash: &StacksBlockId,
+        sortdb_conn: &Connection,
+        burn_tip: &BurnchainHeaderHash,
+        burn_tip_height: u64,
+    ) -> Result<(Vec<StackStxOp>, Vec<TransferStxOp>), Error> {
```
kantai, line 4979: I don't think the lifetime generic is necessary here: rust can almost always figure out the lifetimes for references that don't land in the return type.

kantai (approved):
This all looks great to me -- just superficial comments!

PR #3129, opened 2022-05-06
feat/remove-pox-sunset

### LINKED ISSUE #2613: [Stacks-2.1] remove PoX sunset and cap unburnt PoX output to be a function of the last prepare phase
Create an extension contract that will allow network stakeholders to vote to extend the PoX sunset window by a fixed amount of blocks.  This allows PoX to keep going under "good behavior" from the system, beyond what the 2.0 implementation mandates.  I propose requiring 75% yes-votes from all liquid STX, to complement the 25% vote required to disable PoX for a reward cycle.  If 75% of the liquid STX vote to extend PoX within a year, the sunset gets bumped by a protocol-defined constant (e.g. 5 years).  Protocol constants subject to negotiation.

jcnelson:
This PR partially fixes #2613 by removing the PoX sunset altogether.  It does not attempt to gate the PoX sunset code.  If we can merge `next` to `master` before Bitcoin block 766050 (which is in just over 200 days), then no gating will be needed, since the sunset code will never be reached.

This PR does not address capping discount mining, which is the other part of #2613.  I'll work on that as a branch off of this branch. I'm just trying to get this fix into `next` ASAP.

### DIFF
--- a/src/chainstate/burn/operations/leader_block_commit.rs
+++ b/src/chainstate/burn/operations/leader_block_commit.rs
@@ -72,7 +72,6 @@ impl LeaderBlockCommitOp {
         apparent_sender: &BurnchainSigner,
     ) -> LeaderBlockCommitOp {
         LeaderBlockCommitOp {
-            sunset_burn: 0,
             block_height: block_height,
             burn_parent_modulus: if block_height > 0 {
                 ((block_height - 1) % BURN_BLOCK_MINED_AT_MODULUS) as u8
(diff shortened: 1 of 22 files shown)


src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -622,20 +591,16 @@ impl LeaderBlockCommitOp {
 
     fn check_single_burn_output(&self) -> Result<(), op_error> {
         if self.commit_outs.len() != 1 {
-            warn!("Invalid post-sunset block commit, should have 1 commit out");
+            warn!("Invalid block commit, should have 1 commit out");
```
pavitthrap, line 594: Invalid *prepare phase* block commit

pavitthrap (approved):

kantai (approved):
LGTM

Issue #2693, opened 2021-06-08
Feature request: allow hashing functions to hash other Clarity types


MarvinJanssen:
**Is your feature request related to a problem? Please describe.**
The ability to hash any Clarity value could open up a lot of DeFi/DeX potential. This suggestion is to have a counterpart to [Ethereum's EIP712](https://eips.ethereum.org/EIPS/eip-712). The hashes can be signed by the user and then passed to off-chains apps to be used in contract calls later.

Some use-cases:
- Simple address ownership proofs without having to hit the chain.
- Signing orders for off-chain order books.
- Creation of pseudo-transactions that can be stored and broadcast later without having to worry about nonces.
- Limitless signature fun.

**Describe the solution you'd like**
CVs could just be serialised to the SIP wire format and then hashed. We might also want to add the chain ID to prevent collisions across forks. A super quick example can be found in this branch: https://github.com/MarvinJanssen/clarity-repl/tree/feat/hash-structured-data.

Working examples:

Javascript:
```JS
const {
	listCV,
	uintCV,
	tupleCV,
	bufferCVFromString,
	serializeCV
	} = require('@stacks/transactions');

const {createHash} = require('crypto');

const sha256CV = input => createHash('sha256').update(serializeCV(input)).digest('hex');

let list = listCV([uintCV(3), uintCV(4), uintCV(5)]);

console.log(sha256CV(list)); // -> 6f3a852f0d216d0012c109b67fd021e34a8873d4c788c69ea79926562d38e1e0

let tuple = tupleCV({num: uintCV(5), buff: bufferCVFromString("buffer string")});

console.log(sha256CV(tuple)); // -> 9aa9471d690ba4844b1282ae8c9387b175557c71f61293660f4a72bd2361299e
```

Clarity:
```Clojure
(sha256 (list u3 u4 u5))
;; -> 0x6f3a852f0d216d0012c109b67fd021e34a8873d4c788c69ea79926562d38e1e0

(sha256 {num: u5, buff: 0x62756666657220737472696e67})
;; -> 0x9aa9471d690ba4844b1282ae8c9387b175557c71f61293660f4a72bd2361299e
```

`buff`, `uint`, and `int` types remain as is to be backwards compatible.

**Additional context**
Provides a solution to: https://github.com/blockstack/stacks-wallet-web/issues/1051

This is what signing CVs could look like in the Web Extension:
[...]

PR #3132, opened 2022-05-10
Implement consensus de/serialization in Clarity

kantai:
### Description

This PR adds `to-consensus-buff` and `from-consensus-buff` methods to the `Clarity2` native methods.

### Applicable issues
- Implements #2693 by allowing arbitrary Clarity types to be transformed into buffers
- Implements the buffer serialization aspects of the https://github.com/stacksgov/sips/pull/57 proposal

These methods are both fallible -- this may be surprising in the case of `to-consensus-buff`, but the reason is that some legal Clarity values serialize into buffers that are larger than the maximum Clarity value size; this is because of the type prefix and length encoding.

The typing of `to-consensus-buff` produces a buffer size which is the maximum possible buffer size for its input. In the case of uints, ints, and bools this is always correct. For other types, however, it will be an overestimate in many cases (e.g., for `principal`, it assumes the principal is a contract with the longest possible Clarity name).

### DIFF
--- a/clarity/src/vm/analysis/type_checker/tests/mod.rs
+++ b/clarity/src/vm/analysis/type_checker/tests/mod.rs
@@ -76,6 +76,54 @@ fn ascii_type(size: u32) -> TypeSignature {
 }
 
 #[test]
+fn test_from_consensus_buff() {
+    let good = [
+        ("(from-consensus-buff int 0x00)", "(optional int)"),
+        (
+            "(from-consensus-buff { a: uint, b: principal } 0x00)",
+            "(optional (tuple (a uint) (b principal)))",
+        ),
+    ];
+
+    let bad = [
+        (
+            "(from-consensus-buff)",
+            CheckErrors::IncorrectArgumentCount(2, 0),
+        ),
+        (
+            "(from-consensus-buff 0x00 0x00 0x00)",
+            CheckErrors::IncorrectArgumentCount(2, 3),
+        ),
+        (
+            "(from-consensus-buff 0x00 0x00)",
+            CheckErrors::InvalidTypeDescription,
+        ),
+        (
+            "(from-consensus-buff int u6)",
+            CheckErrors::TypeError(TypeSignature::max_buffer(), TypeSignature::UIntType),
+        ),
+        (
+            "(from-consensus-buff (buff 1048576) 0x00)",
+            CheckErrors::ValueTooLarge,
+        ),
+    ];
+
+    for (good_test, expected) in good.iter() {
+        let type_result = type_check_helper(good_test).unwrap();
+        assert_eq!(expected, &type_result.to_string());
+
+        assert!(
+            type_result.admits(&execute_v2(good_test).unwrap().unwrap()),
+            "The analyzed type must admit the evaluated type"
+        );
+    }
+
+    for (bad_test, expected) in bad.iter() {
+        assert_eq!(expected, &type_check_helper(&bad_test).unwrap_err().err);
+    }
+}
+
+#[test]
 fn test_to_consensus_buff() {
     let good = [
         (
@@ -106,6 +154,11 @@ fn test_to_consensus_buff() {
         ("(to-consensus-buff 0x00)", "(optional (buff 6))"),
         ("(to-consensus-buff \"a\")", "(optional (buff 6))"),
         ("(to-consensus-buff u\"ab\")", "(optional (buff 13))"),
+        ("(to-consensus-buff 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6)", "(optional (buff 151))"),
+        ("(to-consensus-buff 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6.abcdeabcdeabcdeabcdeabcdeabcdeabcdeabcde)", "(optional (buff 151))"),
+        ("(to-consensus-buff true)", "(optional (buff 1))"),
+        ("(to-consensus-buff -1)", "(optional (buff 17))"),
+        ("(to-consensus-buff u1)", "(optional (buff 17))"),
         ("(to-consensus-buff (list 1 2 3 4))", "(optional (buff 73))"),
         (
             "(to-consensus-buff { apple: u1, orange: 2, blue: true })",
--- a/clarity/src/vm/analysis/type_checker/tests/mod.rs
+++ b/clarity/src/vm/analysis/type_checker/tests/mod.rs
@@ -27,6 +27,7 @@ use crate::vm::analysis::AnalysisDatabase;
 use crate::vm::ast::errors::ParseErrors;
 use crate::vm::ast::{build_ast, parse};
 use crate::vm::contexts::OwnedEnvironment;
+use crate::vm::execute_v2;
 use crate::vm::representations::SymbolicExpression;
 use crate::vm::types::{
     BufferLength, FixedFunction, FunctionType, PrincipalData, QualifiedContractIdentifier,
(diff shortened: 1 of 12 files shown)


clarity/src/vm/analysis/type_checker/tests/mod.rs
```diff
@@ -74,6 +75,92 @@ fn ascii_type(size: u32) -> TypeSignature {
     TypeSignature::SequenceType(StringType(ASCII(size.try_into().unwrap()))).into()
 }
 
+#[test]
+fn test_to_consensus_buff() {
```
jcnelson, line 79: Do you want to add a test for standard and contract principals?

clarity/src/vm/docs/mod.rs
```diff
@@ -1978,6 +1978,51 @@ one of the following error codes:
...
+",
+    example: r#"
+(from-consensus-buff int 0x0000000000000000000000000000000001) ;; Returns (some 1)
+(from-consensus-buff uint 0x0000000000000000000000000000000001) ;; Returns none
+(from-consensus-buff uint 0x0100000000000000000000000000000001) ;; Returns (some u1)
+(from-consensus-buff bool 0x0000000000000000000000000000000001) ;; Returns none
+(from-consensus-buff bool 0x03) ;; Returns (some true)
+(from-consensus-buff bool 0x04) ;; Returns (some false)
+"#,
+};
```
jcnelson, line 2024: Can you add examples for principals and tuples for these two documentations?

clarity/src/vm/types/serialization.rs
```diff
@@ -287,6 +291,132 @@ macro_rules! check_match {
     };
 }
 
+impl TypeSignature {
+    /// Return the maximum length of the consensus serialization of a
+    /// Clarity value of this type. The returned length *may* not fit
+    /// in in a Clarity buffer! For example, the maximum serialized
```
jcnelson, line 297: typo -- "in in a"

clarity/src/vm/types/serialization.rs
```diff
@@ -287,6 +291,132 @@ macro_rules! check_match {
...
+                u32::from(length)
+                    .checked_add(str_length_encode)
+                    .ok_or_else(|| CheckErrors::ValueTooLarge)?
+            }
+            TypeSignature::SequenceType(SequenceSubtype::StringType(StringSubtype::UTF8(
+                length,
+            ))) => {
+                // u32 length as big-endian bytes
+                let str_length_encode = 4;
+                // utf-8 is maximum 4 bytes per codepoint (which is the length)
```
jcnelson, line 348: In the implementation, each unicode point is *always* represented in memory as 4 bytes.

clarity/src/vm/types/serialization.rs
```diff
@@ -287,6 +291,132 @@ macro_rules! check_match {
...
+                u32::from(length)
+                    .checked_mul(4)
+                    .and_then(|x| x.checked_add(str_length_encode))
+                    .ok_or_else(|| CheckErrors::ValueTooLarge)?
+            }
+            TypeSignature::PrincipalType => {
```
jcnelson, line 354: It's a pity that the `PrincipalType`'s maximum size calculation must assume a maximally-long contract principal, instead of distinguishing between standard and contract principals.  But given our time tables, I'm assuming that this is the right thing to do for now.

clarity/src/vm/types/serialization.rs
```diff
@@ -287,6 +291,132 @@ macro_rules! check_match {
...
+            TypeSignature::TupleType(tuple_type) => {
+                let type_map = tuple_type.get_type_map();
+                // u32 length as big-endian bytes
+                let tuple_length_encode: u32 = 4;
+                let mut total_size = tuple_length_encode;
+                for (key, value) in type_map.iter() {
+                    let value_size = value.max_serialized_size()?;
+                    total_size = total_size
+                        .checked_add(1) // length of key-name
+                        .and_then(|x| x.checked_add(key.len() as u32)) // ClarityName is ascii-only, so 1 byte per length
```
jcnelson, line 371: We can safely assume here that `key.len()` is going to be at most `MAX_STRING_LEN` or something?

clarity/src/vm/types/serialization.rs
```diff
@@ -287,6 +291,132 @@ macro_rules! check_match {
...
 }
 
+impl TypeSignature {
+    /// Return the maximum length of the consensus serialization of a
+    /// Clarity value of this type. The returned length *may* not fit
+    /// in in a Clarity buffer! For example, the maximum serialized
+    /// size of a `(buff 1024*1024)` is `1+1024*1024` because of the
+    /// type prefix byte. However, that is 1 byte larger than the maximum
+    /// buffer size in Clarity.
+    pub fn max_serialized_size(&self) -> Result<u32, CheckErrors> {
```
jcnelson, line 301: Do you wanna add a runtime `assert!()` to `Value::serialize_write()` to verify that the resulting byte buffer does indeed fit into the size reported by this function?

jcnelson (approved):
Overall LGTM.  Just had a few superficial / edification comments, with a request for a (trivial) additional bit of test coverage.  Thanks for whipping this up!

@MarvinJanssen do you want to do a pass on this before merging?  Let us know.

MarvinJanssen (approved):
Amazing, I was working on a `to-consensus-buff` for the REPL but this code is a lot more elegant. Can easily it be ported to the REPL (and thus Clarinet) before 2.1 launches?

obycode:
> Can easily it be ported to the REPL (and thus Clarinet) before 2.1 launches?

Migrating the REPL to use this canonical clarity lib is high on my todo list. Once that is done, we would enable a 2.1 option for testing before 2.1 is launched.

pavitthrap (approved):
lgtm

clarity/src/vm/types/serialization.rs
```diff
@@ -287,6 +291,132 @@ macro_rules! check_match {
...
+            }
+            TypeSignature::IntType => 16,
+            TypeSignature::UIntType => 16,
+            TypeSignature::BoolType => 0,
+            TypeSignature::SequenceType(SequenceSubtype::ListType(list_type)) => {
+                // u32 length as big-endian bytes
+                let list_length_encode = 4;
+                list_type
+                    .get_max_len()
+                    .checked_mul(list_type.get_list_item_type().max_serialized_size()?)
```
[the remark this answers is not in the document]
jcnelson: There is a limit (it's 16), but it's not clear yet if it's in a SIP yet.  It should be if it is not.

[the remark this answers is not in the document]
kantai: Limit is actually 32 (defined as `clarity::vm::types::MAX_TYPE_DEPTH`) -- SIP-002 doesn't specify either the maximum size or maximum depth of Clarity types (only that they should be set). I'll open an issue on the sips repo.

clarity/src/vm/types/serialization.rs
kantai: But not in the consensus serialization -- there, UTF-8 strings are represented as the length encoding of the total number of bytes (without padding to 4 bytes per codepoint), and then the normal utf-8 byte encoding.

clarity/src/vm/types/serialization.rs
kantai: I think this is unavoidable as long as the same type encompasses both contract and standard principals, and I'd be pretty resistant to making new types `standard-principal` and `contract-principal`, because I think we should encourage contracts to treat all principals the same.

clarity/src/vm/types/serialization.rs
kantai: Yes, it is at most `MAX_STRING_LEN`. However, the key-name is part of the type description for a tuple, so we can actually know the serialized length of the key-name component of a tuple.

clarity/src/vm/functions/conversions.rs
```diff
@@ -209,3 +211,58 @@ pub fn native_int_to_utf8(value: Value) -> Result<Value> {
...
+    let input_bytes = if let Value::Sequence(SequenceData::Buffer(buff_data)) = value {
+        Ok(buff_data.data)
+    } else {
+        Err(CheckErrors::TypeValueError(
+            TypeSignature::max_buffer(),
+            value,
+        ))
+    }?;
+
+    runtime_cost(ClarityCostFunction::Unimplemented, env, input_bytes.len())?;
```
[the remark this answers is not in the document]
kantai: Yep -- https://github.com/stacks-network/stacks-blockchain/issues/3138

clarity/src/vm/tests/simple_apply_eval.rs
```diff
@@ -211,6 +211,211 @@ fn test_keccak256() {
...
+        (
+            "(from-consensus-buff 2 0x10)",
+            "Unchecked(InvalidTypeDescription)",
+        ),
+    ];
+
+    for (input, expected) in vectors.iter() {
+        let result = vm_execute_v2(input).expect_err("Should raise an error");
+        eprintln!("{}", result.to_string());
+    }
```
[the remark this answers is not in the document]
kantai: Removed in d8f7628

clarity/src/vm/types/serialization.rs
kantai: Added in d8f7628

clarity/src/vm/types/serialization.rs
kantai: Removed in d8f7628

clarity/src/vm/docs/mod.rs
kantai: Added in d8f7628

clarity/src/vm/analysis/type_checker/tests/mod.rs
kantai: Added in d8f7628

jcnelson (approved):
Cool, looks like all my comments were addressed :+1:

Issue #3095, opened 2022-04-18
[burnchain] cap the upside of discount-mining 


jcnelson:
Discount mining is when a miner stacks their STX and leverages the fact that when their reward slots come up for PoX payout, they can pay as much BTC into them as they want (knowing they'll get it back) while also boosting their sortiton win probability.  This permits _unbound_ upside to discount-mining -- the only cost to doing so is the BTC transaction fee.

For 2.1, we can cap the upside by (a) measuring the distribution of _BTC burns_ in the prepare phase, and (b) using some threshold function of that distribution to require that all miners _burn_ BTC if they spend above the threshold on mining.  This caps the upside of discount-mining, because then the discount miner must pay a proportionate amount of BTC to considerably boost their win probability.

A simple way to do this could be to take _T = min(mean(winner_burns), median(winner_burns))_ as the threshold, where _winner_burns_ is the list of BTC burns for each winning sortition in the prepare phase, and require that if a miner spends _B > k * T_ BTC on a block-commit (for some protocol-defined _k_), then _B - k * T_ BTC must be burnt via the `OP_RETURN` output.  We'd choose _k_ so that it's highly unlikely that honest miners would be forced to burn BTC.  We can infer it from historical prepare phases -- i.e. pick _k_ so that _k * T_ is higher than any block-commit ever seen for any prepare phase.

jcnelson:
> A simple way to do this could be to take T = min(mean(winner_burns), median(winner_burns)) as the threshold,

This can't work. A discount-miner would simply run many miners paying at _T_, thereby bypassing the threshold.  Our best bet so far seems to be something around sortition weight windowing, where it doesn't matter how many UTXO chains the discount miner controls.

PR #3135, opened 2022-05-14
Feat/cap discount mining

jcnelson:
This PR addresses #3095 (and #2613) by capping the amount of BTC a PoX output can pay.  If a miner wants to spend more BTC than this amount, then they must _burn_ it via the `OP_RETURN` output.  This puts an upper bound on how much of a discount a miner who stacks can expect to receive.

The _PoX cutoff_ is calculated on a reward cycle basis within the prepare phase.  Specifically, the code examines the sequence of BTC burns _B_ by the _winning_ block-commits in the prepare phase, and sets the cutoff _C = 4 * min(mean(B), median(B))_.  If a miner wants to spend _S > C_ BTC, they must burn _S - C_ BTC.  In addition, if the miner spends _S <= C_ BTC, they _must_ spend it by sending it to PoX outputs (i.e. they may not burn it).  This only applies to reward cycles with PoX; nothing changes for PoB reward cycles, or for miners who want to mine off of a fork that does not descend from the anchor block.

The rationale for this cutoff calculation is as follows.  Consider a miner who wants to mine at a steep discount of their choosing by stacking STX and spending an unbound amount of BTC when their reward addresses come up for payment.  The cutoff calculation is designed so that the only way a miner can unilaterally choose their discount is by executing a 51% attack on the blockchain during the prepare-phase.  To set a target cutoff _C_, the miner must both ensure that the _majority_ of winning block-commits spend at least _0.25C_ BTC (i.e. they must win at least 51 blocks to guarantee this), and must ensure that the average winning block-commit spends at least _0.25C_ BTC (i.e. they have to burn enough BTC, no matter their win rate, to push the average up).  The 4x multiplier was chosen by examining the cutoff value for all of the system's reward cycles to date: in all but one reward cycle (the 3rd one in the system, when there was a large fluctuation in mining payouts), this cutoff is always higher than the network's PoX outputs.

When working on this PR, I noticed that the now-obsolete `sunset_burn` field was not being used to calculate the sortition weight of a block-commit when calculating the burn distribution.  This was likely a bug, but one that we didn't ever see in the wild because the required sunset burn has been 0.  Nevertheless, this PR (1) makes it so that the total sortition weight of a block-commit includes both the payment to PoX outputs and the burnt amount, and (2) gates this calculation so that it only applies in 2.1.  Prior to this, a block-commit that burns BTC via its OP_RETURN will be treated as if it did not do any burn at all -- i.e. the burn won't improve the block-commit's chances of winning, which is in-line with how 2.05 behaves today.

The PoX cutoff itself is stored in the sortition DB as part of its MARF, so each reward cycle has a distinct PoX cutoff value.  This value is not queried except in epoch 2.1, so it should be safe to migrate the existing chainstate to this node's new chainstate as long as its done before 2.1 activates.  To accommodate this, there's a special rule for the PoX cutoff in the event that 2.1 activates _in the middle of_ a reward cycle:  in this partially-2.1 reward cycle, the cutoff is `u64::MAX` (i.e. there's no cutoff).  The cutoff is only calculated and used in the first full 2.1 reward cycle.

As for testing, I've add the following coverage:

* The block-commit validation tests use a mocked PoX reward cycle in order to verify that a block-commit must burn the right amount of BTC if it overpays, and verifies that the block-commit cannot underpay PoX outputs by burning if it does not meet the cutoff.

* The sortition-processing validation tests verify that the epoch gating for calculating a block-commit's sortition weight applies, and that a block-commit's `destroyed` BTC gets applied only in epoch 2.1.

* The sortition DB testing covers the new PoX cutoff calculation logic, as well as the cutoff storage and retrieval

* There is a neon integration test that verifies that a miner that overpays BTC in PoX cannot win sortitions

EDIT: For clarity, this PR does _not_ make it so that a discount-miner can receive a maximum 4x discount, or anything like that. What it does is it makes it so that a discount-miner must maintain multiple independent UTXO chains (i.e. run multiple separate miner nodes) in order to avoid burning BTC.  This raises the cost of attempting to discount-mine.

Today, a discount-miner would need to spend their target BTC for the 3 blocks prior to the block in which they intend to receive a discount.  This is because the sortition weight of a miner is the minimum of the last BTC spend they made, and the median spend they made in the last 6 blocks (as tracked by the miner's UTXO chain).  With this PR, the amount of BTC the miner would be allowed to recover _per UTXO chain_ is capped by a function of the BTC burnt in the last prepare phase.  So in order to receive more BTC back than this threshold _C_, the discount-miner would be required to run multiple UTXO-chains.  If the miner wanted to spend _k * C_ BTC, they would need to maintain _k_ UTXO chains, and pay _k_ times more BTC transaction fees (whereas today, they only need to pay the fees for a single UTXO chain).

### DIFF
--- a/.github/workflows/bitcoin-tests.yml
+++ b/.github/workflows/bitcoin-tests.yml
@@ -72,6 +72,9 @@ jobs:
           - tests::epoch_205::test_cost_limit_switch_version205
           - tests::epoch_205::test_exact_block_costs
           - tests::epoch_205::bigger_microblock_streams_in_2_05
+          - tests::epoch_21::transition_fixes_utxo_chaining
+          - tests::epoch_21::transition_adds_burn_block_height
+          - tests::epoch_21::transition_caps_discount_mining_upside
     steps:
       - uses: actions/checkout@v2
       - name: Download docker image
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -120,6 +120,11 @@ impl BurnchainStateTransition {
         let mut block_commits: Vec<LeaderBlockCommitOp> = vec![];
         let mut user_burns: Vec<UserBurnSupportOp> = vec![];
         let mut accepted_ops = Vec::with_capacity(block_ops.len());
+        let epoch = SortitionDB::get_stacks_epoch(sort_tx, parent_snapshot.block_height + 1)?
+            .expect(&format!(
+                "FATAL: no epoch known at burn height {}",
+                parent_snapshot.block_height + 1
+            ));
 
         assert!(Burnchain::ops_are_sorted(block_ops));
 
@@ -240,6 +245,7 @@ impl BurnchainStateTransition {
         // calculate the burn distribution from these operations.
         // The resulting distribution will contain the user burns that match block commits
         let burn_dist = BurnSamplePoint::make_min_median_distribution(
+            epoch.epoch_id,
             windowed_block_commits,
             windowed_missed_commits,
             burn_blocks,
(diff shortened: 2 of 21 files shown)


jcnelson:
Need to re-think the cap function so that a miner can't just run a large number of small miners during the reward phase in order to give itself more room to discount mine.

jcnelson:
Spoke more with @kantai about this.  I don't think this PR does enough to really make a difference in discount-mining.  The fact that we have (1) a mining window where a miner's sortition weight is `min(last_commit, median_commit)` and (2) the `reject-pox` function are both good enough for the time being to deter discount-mining.  We can revisit this in a subsequent network upgrade.

PR #3139, opened 2022-05-16
fix: add logging for some tx outcomes that were missing

### LINKED ISSUE #3130: fix: Catch "Transaction Result" logging cases that aren't covered
**Describe the bug**
It seems possible for transactions to not have a "transaction outcome" logged for them.

**Steps To Reproduce**
Look for `30f9970f4fdb7bb837cdd4a69a0da95688979e0f5bddce85db3ccc4fa353b728` in the mock miner logs from May 5 (assuming you have access to such logs).

**Expected Result**
There should be a line like `Tx processing failed with error`, or `Tx successfully processed.`, but there is no entry of either type for this transaction.

**Additional context**
Some transaction outcome cases aren't covered in the code base.

https://github.com/stacks-network/stacks-blockchain/blob/26bfd5fcdc1f25106288f18ced05290b569f6abb/src/chainstate/stacks/miner.rs#L1997

gregorycoppola:
### Description
#2975 added "compiler-enforced logging" to the mining paths, to make sure each transaction outcome was logged. The idea with "compiler-enforced logging" is that the type system is used to make it very hard/impossible to forget to log a transaction outcome.

However, there were still some changes for outcomes to not be logged.

This PR moves the requirement to output a "transaction outcome" object up a level, to `MemPoolDB::iterate_candidates`, to which the `todo` function must now return an `Option<TransactionEvent>`, which are then collected and output by `iterate_candidates`.

Now there be no chance to forget to log an outcome, except for the "drawback" listed below.

### Applicable issues
- fixes #3130

Drawback:
* `iterate_candidate` asks that `todo` return an `Option<TransactionEvent>`, returning `None` if the outcome was an error that wasn't transaction-specific (e.g. block limit reached). This was done to avoid introducing a new event type for "non-transaction-specific error". However, the ability to return `None` does mean that the user has a chance to forget to log a meaningful outcome, by just outputting `None`, because the type system won't stop them.

### Checklist
- [x] Test coverage for new or modified code paths (using existing tests)
- [x] Changelog is updated

### DIFF
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -890,11 +890,12 @@ impl<'a> StacksMicroblockBuilder<'a> {
                                 "Microblock miner deadline exceeded ({} ms)",
                                 self.settings.max_miner_time_ms
                             );
-                            return Ok(false);
+                            return Ok(None);
                         }
 
                         if considered.contains(&mempool_tx.tx.txid()) {
-                            return Ok(true);
+                            return Ok(Some(TransactionResult::skipped(
+                                &mempool_tx.tx, "Transaction already considered.".to_string()).convert_to_event()));
                         } else {
                             considered.insert(mempool_tx.tx.txid());
                         }
@@ -907,6 +908,7 @@ impl<'a> StacksMicroblockBuilder<'a> {
                             &block_limit_hit,
                         ) {
                             Ok(tx_result) => {
+                        let result_event = tx_result.convert_to_event();
                                 tx_events.push(tx_result.convert_to_event());
                                 match tx_result {
                                     TransactionResult::Success(TransactionSuccess {
@@ -937,7 +939,7 @@ impl<'a> StacksMicroblockBuilder<'a> {
                                         num_txs += 1;
                                         num_added += 1;
                                         num_selected += 1;
-                                        Ok(true)
+                                        Ok(Some(result_event))
                                     }
                                     TransactionResult::Skipped(TransactionSkipped {
                                         error,
@@ -963,7 +965,7 @@ impl<'a> StacksMicroblockBuilder<'a> {
                                                     debug!("Block budget exceeded while mining microblock"; 
                                                         "tx" => %mempool_tx.tx.txid(), "next_behavior" => "Stop mining microblock");
                                                     block_limit_hit = BlockLimitFunction::LIMIT_REACHED;
-                                                    return Ok(false);
+                                                    return Ok(None);
                                                 }
                                             }
                                             Error::TransactionTooBigError => {
@@ -971,7 +973,7 @@ impl<'a> StacksMicroblockBuilder<'a> {
                                             }
                                             _ => {}
                                         }
-                                        return Ok(true)
+                                        return Ok(Some(result_event))
                                     }
                                 }
                             }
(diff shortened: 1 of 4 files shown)


src/core/mempool.rs
```diff
@@ -1079,9 +1087,15 @@ impl MemPoolDB {
                            "size" => consider.tx.metadata.len);
                     total_considered += 1;
 
-                    if !todo(clarity_tx, &consider, self.cost_estimator.as_mut())? {
-                        debug!("Mempool iteration early exit from iterator");
-                        break;
+                    // Run `todo` on the transaction.
+                    match todo(clarity_tx, &consider, self.cost_estimator.as_mut())? {
+                        Some(tx_event) => {
```
pavitthrap, line 1092: In this case `else if block_limit_hit
                                            == BlockLimitFunction::CONTRACT_LIMIT_HIT` in `build_anchored_block`, you return Some(event) when before Ok(false) was returned. 

This will result in different behavior

gregorycoppola: updated

jcnelson:
Can you also update the CHANGELOG to describe this fix?  Thanks!

src/core/mempool.rs
```diff
@@ -1009,17 +1009,25 @@ impl MemPoolDB {
     ///  highest-fee-first order.  This method is interruptable -- in the `settings` struct, the
     ///  caller may choose how long to spend iterating before this method stops.
     ///
-    ///  `todo` returns a boolean representing whether or not to keep iterating.
+    ///  `todo` returns an option to a `TransactionEvent` representing the outcome, or None if we
+    ///  hit an error that wasn't transaction specific.
```
kantai, line 1013: Would be better framed as:

```rust
/// `todo` controls the iteration: if it returns `None`, then the mempool iteration will stop.
/// if `todo` returns `Some`, the contents of the return are a `TransactionEvent` representing
/// the outcome of trying to include a given transaction in the block.
```

`None` doesn't always indicate an error occurred: it just indicates that the miner is done assembling the block, and wants to halt the iteration.

src/chainstate/stacks/miner.rs
```diff
@@ -1988,35 +1990,61 @@ impl StacksBlockBuilder {
...
+                                        return Ok(Some(
+                                            TransactionResult::skipped(
+                                                &txinfo.tx,
+                                                format!(
+                                                    "Bad sponsor nonce, tx nonce {} versus {}.",
+                                                    sponsor_nonce, *nonce
+                                                ),
+                                            )
+                                            .convert_to_event(),
+                                        ));
```
kantai, line 2047: Do we actually want these events included in the event observer interface? These seems like they would be basically spurious event data. I'm not opposed to logging them, but re-broadcasting them over the event observer seems excessive.

kantai (commented):
I'm not sure that the SkippedTransaction events should be sent to the event observer. I think that logging them is okay. I also think the new `todo` signature needs a better rustdoc.

src/core/mempool.rs
kantai: Can you also comment on how the new `Vec<TransactionEvent>` argument is being used?

src/core/mempool.rs
```diff
@@ -1009,17 +1009,25 @@ impl MemPoolDB {
...
+    ///
+    /// `output_events` is modified in place, adding all substantive transaction events output by `todo`.
     pub fn iterate_candidates<F, E, C>(
         &mut self,
         clarity_tx: &mut C,
+        output_events: &mut Vec<TransactionEvent>,
```
kantai, line 1019: Can you add a rustdoc for this new argument? Also, passing `&mut Vec<T>` or `&Vec<T>` should be avoided, in almost all cases, it's preferable to use `&mut [T]` or `&[T]`, because it gives you more flexibility in the caller.

gregorycoppola: 1) I added to the comments for this.

2) I think it has to be a `Vec` because of the `push`:

```
error[E0599]: no method named `push` found for mutable reference `&mut [TransactionEvent]` in the current scope
    --> src/core/mempool.rs:1093:43
     |
1093 | ...                   output_events.push(tx_event);
     |                                     ^^^^ method not found in `&mut [TransactionEvent]`
```

There actually isn't that much latitude for the caller.

gregorycoppola:
@kantai main question let is: what to do about `TransactionSkipped`.. where would we put the code to drop these? or just let the observer drop them?

src/chainstate/stacks/miner.rs
kantai: Something like option 3 -- they shouldn't be forwarded to the observer at all.

gregorycoppola: I  tried it the reverting way.. but then the logic of skipping `Skipped` events was kind of dispersed.

So I pushed a version that puts this in `iterate_candidates`, that way we just have to write the logic once...

```
                    // Run `todo` on the transaction.
                    match todo(clarity_tx, &consider, self.cost_estimator.as_mut())? {
                        Some(tx_event) => {
                            match tx_event {
                                TransactionEvent::Skipped(_) => {
                                    // don't push `Skipped` events to the observer
                                }
                                _ => {
                                    output_events.push(tx_event);
                                }
                            }
                        }
                        None => {
                            debug!("Mempool iteration early exit from iterator");
                            break;
                        }
                    }
```

Issue #3043, opened 2022-02-10
[stacks 2.1] expose BTC/STX exchange rate in Clarity 


jcnelson:
We know how much BTC was spent to mint a Stacks block, and we know how much STX was produced / acquired in it.  This information could be exposed in `get-block-info?`.

gregorycoppola:
I like this feature. I like that we can get the exchange rate without looking off chain.

Using this exchange rate, a miner could stake STX to collateralize a BTC withdrawal.

And if they try to steal the BTC, they can be penalized in STX by the Stacks chain.

MarvinJanssen:
Could be very useful. 😁

igorsyl:
I would not call this ratio an exchange rate given miners expect a premium. This ratio could differ from market-based exchange rate by more 10%.

jcnelson:
Miners spend X BTC, and get Y STX, and this information is now reported to Clarity.  `get-block-info?` now reports X and Y on request, so I closed this issue.

Also, this is _factually_ the exchange rate.  Just because an exchange says something different doesn't make it true; the only thing that matters here is what happens on-chain.

gregorycoppola:
@jcnelson What if there is an on-chain market for BTC/STX.. can we get the rate from there?

jcnelson:
Sure -- this would be the way to get that exchange rate.

gregorycoppola:
I mean where people are literally (trustlessly) trading STX for BTC.. without anyone being a miner.

Just a market.

Issue #3140, opened 2022-05-19
[chainstate] parent block miner does not receive 40% of the microblock tx fees


jcnelson:
When working on #3043, I noticed that there's a bug in the microblock fee distribution calculation.  When we generate the `MinerReward` struct for the parent that captures the streamed transaction fees that the parent produced, we accidentally use the fees that the parent confirmed.  This means that if A mines a microblock stream, B confirms it, and C mines a block off of B's block, the following will happen:

* A won't get any transaction fees from the stream
* B will get 60% of the stream's fees when B's block reward matures
* B will get 40% of the stream's fees when C's block reward matures

What should happen is:

* A gets 40% of the stream's fees when B's block reward matures
* B gets 60% of the stream's fees when B's block reward matures

Fortunately, it's easy to fix -- we can fold it into Stacks 2.1.  Also, the impact isn't too severe yet, since transaction fees are only a small portion of the overall block reward.

jcnelson:
I'll note in the comments in my PR for #3043 where this bug occurs, and I'll base another PR off of it that fixes and gates it.

PR #3142, opened 2022-05-19
Feat/btc stx exchange rate

### LINKED ISSUE #2492: [chainstate] burnchain_sortition_burn calculation is for the previous, not current sortition
The `burnchain_sortition_burn` field in the Stacks headers DB is calculated for sortition `N - 1`, whereas `burnchain_commit_burn` is calculated from sortition `N`.  I don't think this field is actually used anywhere anymore, but I'm leery to change/remove it until Stacks 2.1.

jcnelson:
This PR addresses #3043 by adding the following new fields to `get-block-info?` in Stacks 2.1:

* `block-reward`: returns the total number of uSTX awarded to the miner once the block reward matures.

* `miner-spend-total`: returns the total number of satoshis spent by all Stacks miners trying to win this block.

* `miner-spend-winner`: returns the number of satoshis spent by the winning miner on this block.

The `block-reward` value is only `(some ..)` after the queried block receives `MINER_REWARD_MATURITY + 1` confirmations.  This is because the block reward can be slashed within the first `MINER_REWARD_MATURITY` blocks by a `PoisonMicroblock` transaction, should the miner publish two or more conflicting microblock streams during their tenure.  Therefore, the system cannot reliably report the block reward until it is no longer possible to publish a `PoisonMicroblock` transaction.

The Stacks chainstate is **not backwards-compatible** with this PR.  Specifically:

* In order for this PR to faithfully report the `block-reward`, I had to make it so that the system stores the `MinerReward` structs that mature once a block is appended to the chainstate.  The actual number of tokens credited to the miner isn't recorded anywhere; we just directly pay to the miner's account when processing matured rewards.

* In order to get `miner-spend-total` to work, I had to fix #2492 to store the right sortition burn for a Stacks block.  While I don't _believe_ this is a consensus-breaking change, since we don't appear to use the value anywhere, the fact that this value is _different_ than what we have now means that we're gonna need to recalculate it.

I could go back and add schema migration code for these two things, but I'm hesitant to do so, since the time taken to write, test, and review the migration code will likely exceed the time to just boot up a 2.1 node from genesis.

### DIFF
--- a/clarity/src/vm/analysis/type_checker/natives/mod.rs
+++ b/clarity/src/vm/analysis/type_checker/natives/mod.rs
@@ -554,9 +554,10 @@ fn check_get_block_info(
         .ok_or(CheckError::new(CheckErrors::GetBlockInfoExpectPropertyName))?;
 
     let block_info_prop =
-        BlockInfoProperty::lookup_by_name(block_info_prop_str).ok_or(CheckError::new(
-            CheckErrors::NoSuchBlockInfoProperty(block_info_prop_str.to_string()),
-        ))?;
+        BlockInfoProperty::lookup_by_name_at_version(block_info_prop_str, &checker.clarity_version)
+            .ok_or(CheckError::new(CheckErrors::NoSuchBlockInfoProperty(
+                block_info_prop_str.to_string(),
+            )))?;
 
     checker.type_check_expects(&args[1], &context, &TypeSignature::UIntType)?;
 
--- a/clarity/src/vm/docs/mod.rs
+++ b/clarity/src/vm/docs/mod.rs
@@ -2244,6 +2244,20 @@ mod test {
         fn get_miner_address(&self, _id_bhh: &StacksBlockId) -> Option<StacksAddress> {
             None
         }
+        fn get_burnchain_tokens_spent_for_block(&self, id_bhh: &StacksBlockId) -> Option<u128> {
+            Some(12345)
+        }
+
+        fn get_burnchain_tokens_spent_for_winning_block(
+            &self,
+            id_bhh: &StacksBlockId,
+        ) -> Option<u128> {
+            Some(2345)
+        }
+
+        fn get_tokens_earned_for_block(&self, id_bhh: &StacksBlockId) -> Option<u128> {
+            Some(12000)
+        }
     }
 
     struct DocBurnStateDB {}
--- a/clarity/src/vm/docs/mod.rs
+++ b/clarity/src/vm/docs/mod.rs
@@ -1459,21 +1459,35 @@ const GET_BLOCK_INFO_API: SpecialAPI = SpecialAPI {
     description: "The `get-block-info?` function fetches data for a block of the given *Stacks* block height. The
 value and type returned are determined by the specified `BlockInfoPropertyName`. If the provided `block-height` does
 not correspond to an existing block prior to the current block, the function returns `none`. The currently available property names
-are `time`, `header-hash`, `burnchain-header-hash`, `id-header-hash`, `miner-address`, and `vrf-seed`.
+are as follows:
 
-The `time` property returns an integer value of the block header time field. This is a Unix epoch timestamp in seconds
+`burnchain-header-hash`: This property returns a `(buff 32)` value containing the header hash of the burnchain (Bitcoin) block that selected the 
+Stacks block at the given Stacks chain height.
+
+`id-header-hash`: This property returns a `(buff 32)` value containing the _index block hash_ of a Stacks block.   This hash is globally unique, and is derived
+from the block hash and the history of accepted PoX operations.  This is also the block hash value you would pass into `(at-block)`.
+
+`header-hash`: This property returns a `(buff 32)` value containing the header hash of a Stacks block, given a Stacks chain height.  **WARNING* this hash is
+not guaranteed to be globally unique, since the same Stacks block can be mined in different PoX forks.  If you need global uniqueness, you should use `id-header-hash`.
+
+`miner-address`: This property returns a `principal` value corresponding to the miner of the given block.
+
+`time`: This property returns a `uint` value of the block header time field. This is a Unix epoch timestamp in seconds
 which roughly corresponds to when the block was mined. **Warning**: this does not increase monotonically with each block
 and block times are accurate only to within two hours. See [BIP113](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki) for more information.
 
-The `header-hash`, `burnchain-header-hash`, `id-header-hash`, and `vrf-seed` properties return a 32-byte buffer.
+New in Stacks 2.1:
 
-`header-hash` returns the header hash of a Stacks node, given a Stacks chain height.
+`block-reward`: This property returns a `uint` value for the total block reward of the indicated Stacks block.  This value is only available once the reward for 
+the block matures.  That is, the latest `block-reward` value available is at least 101 Stacks blocks in the past (on mainnet).  The reward includes the coinbase,
+the anchored block's transaction fees, and the shares of the confirmed and produced microblock transaction fees earned by this block's miner.  Note that this value may 
+be smaller than the Stacks coinbase at this height, because the miner may have been punished with a valid `PoisonMicroblock` transaction in the event that the miner
+published two or more microblock stream forks.
 
-`burnchain-header-hash` returns header hash of the burnchain (Bitcoin) node corresponding to the given Stacks chain height.
+`miner-spend-total`: This property returns a `uint` value for the total number of burnchain tokens (i.e. satoshis) spent by all miners trying to win this block.
 
-The `miner-address` property returns a `principal` corresponding to the miner of the given block.
-
-The `id-header-hash` is the block identifier value that must be used as input to the `at-block` function.
+`miner-spend-winner`: This property returns a `uint` value for the number of burnchain tokens (i.e. satoshis) spent by the winning miner for this Stacks block.  Note that
+this value is less than or equal to the value for `miner-spend-total` at the same block height.
 ",
     example: "(get-block-info? time u0) ;; Returns (some u1557860301)
 (get-block-info? header-hash u0) ;; Returns (some 0x374708fff7719dd5979ec875d56cd2286f6d3cf7ec317a3b25632aab28ec37bb)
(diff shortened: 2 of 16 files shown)


jcnelson:
In writing this code, I discovered a consensus bug (#3140) in how we calculate microblock transaction fees.  I'm going to send another PR on top of this one that fixes it.  I documented where this bug occurs in the comments in this PR.

src/clarity_vm/tests/contracts.rs
```diff
@@ -152,3 +152,129 @@ fn test_get_burn_block_info_eval() {
...
+            Value::some(Value::UInt(0)).unwrap(),
+            tx.eval_read_only(&contract_identifier, "(test-func-2 u0)")
+                .unwrap()
+        );
+        assert_eq!(
+            Value::some(Value::UInt(0)).unwrap(),
+            tx.eval_read_only(&contract_identifier, "(test-func-3 u0)")
+                .unwrap()
+        );
+        // only works at the first block
```
gregorycoppola, line 232: does this really only work on the first block? doesn't it work on any block already in the past?

gregorycoppola, line 232: does this really only work at the "first" block? isn't it any block that has gone by? and haven't we passed more than one?

i think it would could be good to test a subsequent block anyway, so you are testing like [0,1,N] working cases.

clarity/src/vm/analysis/type_checker/natives/mod.rs
```diff
@@ -555,9 +555,10 @@ fn check_get_block_info(
         .ok_or(CheckError::new(CheckErrors::GetBlockInfoExpectPropertyName))?;
 
     let block_info_prop =
-        BlockInfoProperty::lookup_by_name(block_info_prop_str).ok_or(CheckError::new(
-            CheckErrors::NoSuchBlockInfoProperty(block_info_prop_str.to_string()),
-        ))?;
+        BlockInfoProperty::lookup_by_name_at_version(block_info_prop_str, &checker.clarity_version)
```
gregorycoppola, line 558: regarding this:

> I could go back and add schema migration code for these two things, but I'm hesitant to do so, since the time taken to write, test, and review the migration code will likely exceed the time to just boot up a 2.1 node from genesis.

Are you comparing the time of the miners to your time? I mean it's fine to make this tradeoff, but not for the reason that the users' time is 1:1 substitutable with yours.

gregorycoppola (commented):
Thanks for the change.. It's making sense to me but had a few questions.

src/chainstate/stacks/miner.rs
```diff
@@ -10146,6 +10148,357 @@ pub mod test {
...
+                        else {
+                            // genesis
+                            assert_eq!(miner_spend_winner, 0);
+                            assert_eq!(miner_spend_total, 0);
+                        }
+
+                        if i > 0 && i < num_blocks - (MINER_REWARD_MATURITY as usize) - 1 {
+                            // NOTE: the value here is wrong (see #3140), but there should be *a*
+                            // value
+                            assert!(block_reward_opt.is_some());
```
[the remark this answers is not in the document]
gregorycoppola: oh, i see, #3144

src/chainstate/stacks/db/accounts.rs
```diff
@@ -474,6 +563,198 @@ impl StacksChainState {
         Ok(())
     }
 
+    /// Store a matured miner reward for subsequent query in Clarity, without doing any validation
+    fn inner_insert_matured_miner_payment<'a>(
```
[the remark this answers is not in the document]
jcnelson: I'm deliberately distinguishing between rewards (which are _to be paid_) and payments (which are _already paid_).

src/clarity_vm/tests/contracts.rs
jcnelson: "only starts working at the first block" (as opposed to the 0th block).

clarity/src/vm/analysis/type_checker/natives/mod.rs
jcnelson: It takes less than 30 hours to boot from genesis in the latest release.  This is something you could kick off on a Friday and expect to be done on a Monday.

More generally, I think we want to get folks into the habit of spawning a node from genesis on each hard fork.  It would:

* reduce the codebase's technical debt by giving us a chance to remove the last epoch's schema migrations

* de-risk the chance that a bug in the ever-growing sequence of schema migrations leads to a chain split

* enable us to be able to make significant, principled, and even green-field changes to the database schemas without worrying about compatibility, thereby reducing development, review, and testing time for deploying them

I think it's fair to consider each new epoch of Stacks to be a wholly new blockchain, whose genesis state just happens to be derived from the last epoch's chainstate.  In fact, I'd be supportive of shipping each hard fork with all of the blocks from the last epochs as a database, so they don't even need to be downloaded on boot-up (just re-processed and re-validated against the Bitcoin blockchain).

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -1445,7 +1445,9 @@ fn bitcoind_forking_test() {
     eprintln!("Miner account: {}", miner_account);
 
     let account = get_account(&http_origin, &miner_account);
-    assert_eq!(account.balance, 0);
+
+    // N.B. rewards mature after 2 confirmations...
```
pavitthrap, line 1449: N.B.?

clarity/src/vm/docs/mod.rs
```diff
@@ -1459,21 +1459,35 @@ const GET_BLOCK_INFO_API: SpecialAPI = SpecialAPI {
...
 not correspond to an existing block prior to the current block, the function returns `none`. The currently available property names
-are `time`, `header-hash`, `burnchain-header-hash`, `id-header-hash`, `miner-address`, and `vrf-seed`.
+are as follows:
 
-The `time` property returns an integer value of the block header time field. This is a Unix epoch timestamp in seconds
+`burnchain-header-hash`: This property returns a `(buff 32)` value containing the header hash of the burnchain (Bitcoin) block that selected the 
+Stacks block at the given Stacks chain height.
+
+`id-header-hash`: This property returns a `(buff 32)` value containing the _index block hash_ of a Stacks block.   This hash is globally unique, and is derived
+from the block hash and the history of accepted PoX operations.  This is also the block hash value you would pass into `(at-block)`.
```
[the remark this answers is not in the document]
jcnelson: The alternative is to just say `consensus_hash`, which is at least as hard to explain (since it is derived from the history of PoX operations).

testnet/stacks-node/src/tests/neon_integrations.rs
jcnelson: Nota bene. https://en.wikipedia.org/wiki/Nota_bene

src/chainstate/stacks/db/mod.rs
```diff
@@ -2119,6 +2161,8 @@ impl StacksChainState {
         microblock_tail_opt: Option<StacksMicroblockHeader>,
         block_reward: &MinerPaymentSchedule,
         user_burns: &Vec<StagingUserBurnSupport>,
+        mature_miner_payouts: Option<(MinerReward, Vec<MinerReward>, MinerReward)>, // (miner, [users], parent)
+        mature_rewards_info: Option<MinerRewardInfo>,
```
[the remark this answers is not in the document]
jcnelson: Fixed

src/chainstate/stacks/miner.rs
```diff
@@ -10146,6 +10148,357 @@ pub mod test {
         }
     }
 
+    #[test]
+    fn test_get_block_info_v210() {
```
kantai, line 10152: Between this PR and #3144, there's almost 2,000 new lines of tests in this file, and these tests are really only sort of related to the miner module. Only about 2,000 lines of this 10k+ line file are implementation. Both of these things strongly suggest to me that the tests components of this file should be split out (i.e., the fact that tests are being added to this file without any changes otherwise in this module, and the fact that this file is now _huge_ and almost entirely composed of tests). It doesn't have to happen in this PR, but I definitely believe it needs to happen.

src/chainstate/stacks/db/accounts.rs
```diff
@@ -474,6 +563,198 @@ impl StacksChainState {
...
+            child_index_block_hash
+        ) VALUES (?1,?2,?3,?4,?5,?6,?7,?8)";
+
+        let args: &[&dyn ToSql] = &[
+            &reward.address.to_string(),
+            &reward.vtxindex,
+            &format!("{}", &reward.coinbase),
+            &format!("{}", &reward.tx_fees_anchored),
+            &format!("{}", &reward.tx_fees_streamed_confirmed),
+            &format!("{}", &reward.tx_fees_streamed_produced),
```
kantai, line 613: `x.to_string()` is always the same as `format!("{}", x)`

```suggestion
            &reward.coinbase.to_string(),
            &reward.tx_fees_anchored.to_string(),
            &reward.tx_fees_streamed_confirmed.to_string(),
            &reward.tx_fees_streamed_produced.to_string(),
```

kantai (commented):
This all looks good to me, just some superficial comments.

src/chainstate/stacks/miner.rs
jcnelson: Sure, I'll open an issue.

src/chainstate/stacks/db/accounts.rs
```diff
@@ -474,6 +563,198 @@ impl StacksChainState {
...
+                    .expect("FATAL: got two child rewards")
+            } else {
+                panic!("FATAL: got two parent rewards");
+            };
+            Ok(Some(reward))
+        } else if child_block_id
+            == &StacksBlockHeader::make_index_block_hash(
+                &FIRST_BURNCHAIN_CONSENSUS_HASH,
+                &FIRST_STACKS_BLOCK_HASH,
+            )
```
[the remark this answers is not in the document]
jcnelson: I kinda feel like we should just remove all instances of `StacksBlockHeader::make_index_block_hash()` in one fell swoop in a separate PR.  There's lots of them.  What do you think?

jcnelson:
I'm ready for another pass @kantai @gregorycoppola @pavitthrap.

src/chainstate/stacks/db/accounts.rs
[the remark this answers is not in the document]
kantai: Yeah, I'm fine with that, but I'd prefer not to add new ones in the mean time

kantai (approved):
LGTM

gregorycoppola (approved):
Assuming tests pass

PR #3144, opened 2022-05-19
Fix/microblock fees 

### LINKED ISSUE #3140: [chainstate] parent block miner does not receive 40% of the microblock tx fees

jcnelson:
This fixes #3140 by considering the evaluated epoch of a matured miner reward's block's parent.  If the parent block was evaluated in epoch 2.1, then the correct 40/60 fee split begins to occur between the parent and child blocks' miners (as it should have been).  Before this happens, the buggy behavior described in #3140 continues to happen.

Do not merge until #3142 is merged, since it depends on code in that branch.

### DIFF
--- a/src/chainstate/stacks/db/accounts.rs
+++ b/src/chainstate/stacks/db/accounts.rs
@@ -33,6 +33,8 @@ use clarity::vm::types::*;
 
 use crate::types::chainstate::{StacksAddress, StacksBlockId};
 
+use crate::core::StacksEpochId;
+
 #[derive(Debug, Clone, PartialEq)]
 pub struct MinerReward {
     pub address: StacksAddress,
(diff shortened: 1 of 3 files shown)


kantai:
On the topic of microblock fee rewards, I think there's another bug (though not a realizable one, because user burn support is disabled) in the handling of poisoned rewards:

https://github.com/stacks-network/stacks-blockchain/blob/master/src/chainstate/stacks/db/accounts.rs#L601

Per the comment, the supporting users should get 0 rewards, not a portion of the coinbase.

jcnelson:
Yup; might as well fix that too since it can't be hit in production anyway.

kantai (approved):
LGTM!

Issue #3128, opened 2022-05-05
Merge new parser from REPL into `next` branch


obycode:
Issues #3124 and #3123 demonstrated the need to make improvements to the Clarity parser. We already have an improved and tested parser in Hiro's https://github.com/hirosystems/clarity-repl/tree/develop/src/repl/ast/ that could be brought into this repo, with a bonus that it has a decent performance improvement as well.

Issue #3123, opened 2022-05-01
[Clarity] Escaped backslash in strings may fail parsing


njordhov:
**Describe the bug**

Parsing a string with escaped backslash `\\` sometimes fails with error.

**Steps To Reproduce**

Evaluate in the Clarity repl:
`(list "\\" "x")`
=> `error: Expected whitespace or a close parens. Found: 'x'`

A contract with the code above also fails to deploy on testnet.

**Expected behavior**

Strings with an escaped backslash should parse correctly.

Note that the expression parses if the `"x"` is removed. Evaluate in the Clarity repl:

 `(list "\\")`
=> `["\\"]`

**Environment (please complete the following information):**

Confirmed in the latest Clarity REPL and by deploying to Testnet.

obycode:
The problem is in this regular expression from [vm/ast/parser/mod.rs:140](https://github.com/stacks-network/stacks-blockchain/blob/59363a2dca5277bc01646bebcab98909c955bee1/src/vm/ast/parser/mod.rs#L139-L142):

```
        LexMatcher::new(
            r##""(?P<value>((\\")|([[ -~]&&[^"]]))*)""##,
            TokenType::StringASCIILiteral,
        ),
```

For the ascii string in the example, `(list "\\" "x")`, the first token in the list from the lever is the string `"\\\" "`.

We have the improved parser v2 in the REPL which handles this correctly. 2.1 seems like a good time to switch the canonical VM over to use this new parser. Should I begin putting together a PR for that?

jcnelson:
Closing as complete if this was fixed in the new v2 parser.  Please reopen if not.

njordhov:
Is there a unit test to validate it has been fixed for the v2 parser?

obycode:
No, I don't think there is. You're right, we should add tests for these.

njordhov:
There should also be tests for related cases that parses but shouldn't, such as:

```clarity
(list "\\" x")
```

PR #3150, opened 2022-05-25
feat: add new parser with fixes and improvements

### LINKED ISSUE #3124: Commented tuple shows up in AST
**Describe the bug**
Parsing a commented tuple results in the tuple still showing up in the AST. This was originally reported in hirosystems/clarinet#339 when a difference was detected between the v1 and v2 parsers. I was able to identify this as the core problem.

**Steps To Reproduce**

```shell
clarinet new tuple-bug
cd tuple-bug
clarinet contract new bug
echo "(define-public (execute)
        (begin
                ;;      {who: 'ST2CY5V39NHDPWSXMW9QDT3HC3GD6Q6XX4CFRK9AG, allowance: u100}
                (ok true)
        )
)" > contracts/bug.clar
clarinet check
```

That `clarinet check` will show an error indicating that the v2 parser has a different result than the canonical parser, and upon debugging this, you can see that the commented tuple shows up in the AST:

```
...
                SymbolicExpression { 
                    expr: List([
                        SymbolicExpression { 
                            expr: Atom(ClarityName("begin")), 
                            id: 6, 
                            span: Span { start_line: 2, start_column: 3, end_line: 2, end_column: 7 } 
                        }, 
                        SymbolicExpression { 
                            expr: List([
                                SymbolicExpression { 
                                    expr: Atom(ClarityName("tuple")), 
                                    id: 8, 
                                    span: Span { start_line: 0, start_column: 0, end_line: 0, end_column: 0 } 
                                }, 
                                SymbolicExpression { 
                                    expr: List([
                                        SymbolicExpression { 
                                            expr: Atom(ClarityName("who")), 
                                            id: 10, 
                                            span: Span { start_line: 3, start_column: 8, end_line: 3, end_column: 10 } 
                                        }, 
                                        SymbolicExpression { 
                                            expr: LiteralValue(Principal(Standard(StandardPrincipalData(ST2CY5V39NHDPWSXMW9QDT3HC3GD6Q6XX4CFRK9AG)))),
                                            id: 11, 
                                            span: Span { start_line: 3, start_column: 13, end_line: 3, end_column: 53 } 
                                        }
                                    ]), 
                                    id: 9, 
                                    span: Span { start_line: 0, start_column: 0, end_line: 0, end_column: 0 } 
                                }, 
                                SymbolicExpression { 
                                    expr: List([
                                        SymbolicExpression { 
                                            expr: Atom(ClarityName("allowance")), 
                                            id: 13, 
                                            span: Span { start_line: 3, start_column: 57, end_line: 3, end_column: 65 } 
                                        }, 
                                        SymbolicExpression { 
                                            expr: LiteralValue(UInt(100)), 
                                            id: 14, 
                                            span: Span { start_line: 3, start_column: 68, end_line: 3, end_column: 70 } 
                                        }
                                    ]), 
                                    id: 12, 
                                    span: Span { start_line: 0, start_column: 0, end_line: 0, end_column: 0 } 
                                }
                            ]), 
                            id: 7, 
                            span: Span { start_line: 3, start_column: 7, end_line: 3, end_column: 72 } 
                        }, 
                        SymbolicExpression { 
                            expr: List([
                                SymbolicExpression { 
                                    expr: Atom(ClarityName("ok")), 
                                    id: 16, 
                                    span: Span { start_line: 4, start_column: 4, end_line: 4, end_column: 5 } 
                                }, 
                                SymbolicExpression { 
                                    expr: Atom(ClarityName("true")), 
                                    id: 17, span: Span { start_line: 4, start_column: 7, end_line: 4, end_column: 10 } 
                                }
                            ]), 
                            id: 15, 
                            span: Span { start_line: 4, start_column: 3, end_line: 4, end_column: 11 } 
                        }
                    ]), 
                    id: 5, 
                    span: Span { start_line: 2, start_column: 2, end_line: 5, end_column: 2 } 
                }
...
```

**Expected behavior**

The commented code should not be in the AST.

**Additional context**
This may be another reason to bring the new parser from the REPL into the blockchain for 2.1 (see also #3123).

obycode:
### Description
This PR adds an updated parser which fixes several issues and improves performance. 

### Applicable issues
- fixes #3124, #3123, #3128

Use of the new parser is gated on epoch 2.1 and the old parser is used prior to then.

### Checklist
- [x] Changelog is updated


kantai (approved):
This looks great to me (and very much an improvement over the v1 parser)! Thanks @obycode!

clarity/src/vm/ast/parser_v2/lexer/error.rs
```diff
@@ -0,0 +1,79 @@
...
+            EmptyUTF8Encoding => "empty UTF8 encoding".to_string(),
+            UnterminatedUTF8Encoding => "unterminated UTF8 encoding, missing '}'".to_string(),
+            InvalidBufferLength(size) => format!("invalid buffer length, {}", size),
+            UnknownEscapeChar(c) => format!("unknown escape character, '{}'", c),
+            ExpectedClosing(c) => format!("expected closing '{}'", c),
+            ExpectedSeparator => "expected separator".to_string(),
+            SingleSemiColon => "unexpected single ';' (comments begin with \";;\"".to_string(),
+            UnknownSymbol(c) => format!("unknown symbol, '{}'", c),
+            NoteToMatchThis(c) => format!("to match this '{}'", c),
+            UnsupportedLineEnding => {
```
jcnelson, line 58: Do these errors show up in the blockchain's event stream?  These two errors might be confusing in that context.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,1957 @@
...
+    line_endings: bool,
+    // `fail_fast` mode indicates that the lexer should not report warnings
+    // and should exit on the first error. This is useful for lexing in the
+    // context of a stacks-node, while normal mode is useful for developers.
+    fail_fast: bool,
+}
+
+fn is_separator(ch: char) -> bool {
+    match ch {
+        '(' | ')' | '{' | '}' | ',' | ':' | '\0' | '.' => true,
```
jcnelson, line 36: What does `\0` separate in v2?  I don't think it's even legal in v1?

clarity/src/vm/ast/parser_v2/lexer/error.rs
obycode: No, these are `Note` level diagnostics, so they are ignored when in `fail_fast` mode.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: It should not be allowed in the contract. It's just used as a placeholder for the end of the contract string (see line 82 below):

```rust
        match self.input.next() {
             Some(ch) => {
                 if self.next == '\n' {
                     self.line = self.line + 1;
                     self.column = 0;
                 }
                 self.next = ch
             }
             None => self.next = '\0',
         }
```

jcnelson: That's the thing -- I don't think strings in rust are guaranteed to be null-terminated.  There isn't a null byte that terminates the contract payload, for example.

What are we doing to be sure that there is absolutely no way that `\0`, or any other ASCII control code besides `\t` and `\n`, can show up in the code body?  Is this handled by a regex somewhere before the contract is fed into the lexer?

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,1957 @@
...
+        }
+    }
+
+    fn read_principal(&mut self) -> String {
+        let mut principal = String::new();
+
+        loop {
+            self.read_char();
+            match self.next {
+                'A'..='Z' | '0'..='9' => principal.push(self.next),
```
jcnelson, line 234: Principals use the Crockford 32 alphabet, which is `0123456789ABCDEFGHJKMNPQRSTVWXYZ` (note that `I`, `L`, and `O` are _not_ in this alphabet).

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,1957 @@
...
+                    } else {
+                        self.proceed_through_error(LexerError::InvalidCharPrincipal(self.next));
+                        return principal;
+                    }
+                }
+            }
+        }
+    }
+
+    fn read_unsigned(&mut self) -> u128 {
```
jcnelson, line 247: I don't think this code does what it's supposed to do if there is ever a literal unsigned number that is greater than `u128::MAX`.  If this happens, then the whole contract should fail analysis.  It looks like this code instead attempts to convert the textual representation into a `u128` without handling the overflow case.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,1957 @@
...
+            num = num * 10 + digit as u128;
+            self.read_char();
+        }
+        if !is_separator(self.next) {
+            self.proceed_through_error(LexerError::InvalidCharUint(self.next));
+        }
+        num
+    }
+
+    fn read_integer(&mut self) -> i128 {
```
jcnelson, line 260: I have the same concern here as with `read_unsigned`.  What happens if there is a literal that requires more than 128 bits to represent?

clarity/src/vm/ast/parser_v2/lexer/mod.rs
jcnelson: I think such an occurence should cause the lexer to log an error at this location?

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,1957 @@
...
+                            s.push(self.next);
+                        }
+                    }
+                }
+            }
+            self.read_char();
+        }
+    }
+
+    fn read_utf8_encoding(&mut self) -> String {
```
jcnelson, line 390: I think this is meant to read `X` in `\u{X}`, right?  If so, can you document this?

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,1957 @@
...
+                    '\\' => data.push("\\".to_string().into_bytes()),
+                    '\"' => data.push("\"".to_string().into_bytes()),
+                    'n' => data.push("\n".to_string().into_bytes()),
+                    't' => data.push("\t".to_string().into_bytes()),
+                    'r' => data.push("\r".to_string().into_bytes()),
+                    '0' => data.push("\0".to_string().into_bytes()),
+                    'u' => {
+                        let encode_str = self.read_utf8_encoding();
+                        let unicode_char = {
+                            let u = u32::from_str_radix(&encode_str, 16).unwrap();
```
jcnelson, line 503: What if `encode_str` is more than 8 characters long, and thus can't fit into `u32`?

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,1957 @@
...
+                        self.proceed_through_error(LexerError::InvalidCharTraitIdent(self.next));
+                        return ident;
+                    }
+                }
+            }
+            self.read_char();
+        }
+    }
+
+    fn read_principal(&mut self) -> String {
```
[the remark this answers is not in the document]
jcnelson: Ah, I think this code just reads `X` in `'X`, right?  It doesn't really care whether or not `X` represents a valid principal?

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,1957 @@
...
+            '"' => {
+                advance = false;
+                Token::AsciiString(self.read_ascii_string())
+            }
+            '0' => {
+                advance = false;
+                self.read_char();
+                if self.next == 'x' {
+                    Token::Bytes(self.read_hex())
+                } else if self.next.is_ascii_digit() {
```
jcnelson, line 663: Wait, what about the string "0)"?  Such as in `(print (< x 0))`?

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,1957 @@
...
+
+        lexer = Lexer::new("\"quote \\\"this\\\"\"", false);
+        assert_eq!(
+            lexer.read_token().unwrap().token,
+            Token::AsciiString("quote \"this\"".to_string())
+        );
+        assert_eq!(lexer.diagnostics.len(), 0);
+
+        lexer = Lexer::new("\"\\r\\t\\0\\\\ ok\"", false);
+        assert_eq!(
```
jcnelson, line 849: I don't think this is allowed today in Clarity -- you can't have bytes`\r` or `\0` in the code body at all.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,1957 @@
...
+    }
+
+    fn read_trait_identifier(&mut self) -> String {
+        let start_line = self.last_line as u32;
+        let start_column = self.last_column as u32;
+        let mut ident = String::new();
+
+        loop {
+            match self.next {
+                'a'..='z' | 'A'..='Z' | '0'..='9' | '_' | '-' => ident.push(self.next),
```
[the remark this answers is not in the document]
jcnelson: Ah, never mind, it seems they do not.

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,2518 @@
...
+        // Early exit in fail_fast mode.
+        if !self.success && self.fail_fast {
+            return None;
+        }
+
+        if self.next_token >= self.tokens.len() {
+            return None;
+        }
+        let token = self.tokens[self.next_token].clone();
+        self.next_token = self.next_token + 1;
```
jcnelson, line 99: This is also legal rust :)

```rust
self.next_token += 1;
```

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,2518 @@
...
+        let mut whitespace = true;
+        loop {
+            if let Some(node) = self.parse_node()? {
+                if !whitespace {
+                    self.add_diagnostic(ParseErrors::ExpectedWhitespace, node.span.clone())?;
+                }
+                nodes.push(node);
+                whitespace = self.ignore_whitespace();
+            } else {
+                let token = self.tokens[self.next_token - 1].clone();
```
jcnelson, line 137: We're sure that `self.next_token > 0` here, always?

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,2518 @@
...
+            match token.token {
+                Token::Colon => {
+                    self.next_token();
+                }
+                Token::Eof => {
+                    // This indicates we have reached the end of the input.
+                    // Create a placeholder value so that parsing can continue,
+                    // then return.
+                    self.add_diagnostic(ParseErrors::TupleColonExpectedv2, token.span.clone())?;
+                    let mut placeholder =
```
jcnelson, line 257: Should there be a dedicated `PreSymbolicExpression` type variant for placeholders like this?

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,2518 @@
...
+
+    // Returns some valid expression. When None is returned, check the current token.
+    pub fn parse_node(&mut self) -> ParseResult<Option<PreSymbolicExpression>> {
+        self.ignore_whitespace();
+        let token = self.next_token();
+        if token.is_none() {
+            return Ok(None);
+        }
+        let token = token.unwrap();
+        // let mut comment = None;
```
jcnelson, line 308: dead code :)

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,2518 @@
...
+            }
+            Token::AsciiString(val) => {
+                let str_val = match Value::string_ascii_from_bytes(val.clone().into_bytes()) {
+                    Ok(s) => s,
+                    Err(_) => {
+                        self.add_diagnostic(
+                            ParseErrors::IllegalASCIIString(val),
+                            token.span.clone(),
+                        )?;
+                        Value::string_ascii_from_bytes("placeholder".as_bytes().to_vec()).unwrap()
```
jcnelson, line 330: Should there be a `Value::Placeholder` type variant instead of a bespoke value for `Value` that indicates as such?

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,2518 @@
...
+                        )?;
+                        ClarityName::try_from("placeholder").unwrap()
+                    }
+                };
+                let mut e = PreSymbolicExpression::atom(cname);
+                e.span = token.span;
+                Some(e)
+            }
+            Token::TraitIdent(mut name) => {
+                if name.len() > MAX_STRING_LEN {
```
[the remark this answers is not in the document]
jcnelson: Ah, nm, I see this only applies to the trait identifier from `(use-trait ...)`, which must be a `ClarityName`

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,2518 @@
...
+                        self.add_diagnostic(ParseErrors::InvalidBuffer, token.span.clone())?;
+                        Value::buff_from_byte(0)
+                    }
+                };
+                let mut e = PreSymbolicExpression::atom_value(value);
+                e.span = token.span;
+                Some(e)
+            }
+            Token::Principal(addr) => {
+                let principal = match PrincipalData::parse_standard_principal(&addr) {
```
jcnelson, line 403: I see that this is where the check that the `addr` is a well-formed address comes in.  However, if the user creates a principal literal that includes an invalid Crockford 32 character, they'll get `ParseErrors::InvalidPrincipalLiteral` instead of `LexerError::InvalidCharPrincipal`.  Is this the desired behavior?

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,2518 @@
...
+                start_line: 1,
+                start_column: 8,
+                end_line: 1,
+                end_column: 9
+            }
+        );
+        match &exprs[3].pre_expr {
+            PreSymbolicExpressionType::AtomValue(Value::Sequence(SequenceData::String(
+                CharType::ASCII(ASCIIData { data: s }),
+            ))) => assert_eq!(s, "hel\tlo".as_bytes()), //assert_eq!(val.expect_ascii(), "hel\tlo"),
```
jcnelson, line 1044: Dead code :)

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,2518 @@
...
+        assert_eq!(diagnostics[0].message, "illegal name (too long), 'veryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryverylong'");
+        assert_eq!(
+            diagnostics[0].spans[0],
+            Span {
+                start_line: 1,
+                start_column: 6,
+                end_line: 1,
+                end_column: 137
+            }
+        );
```
jcnelson, line 2012: Can you add a test for a principal with invalid Crockford 32 characters?

jcnelson (changes requested):
I have a few comments that need to be addressed before I'm okay with merging this.  I think there may be some errors in the implementation.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: Ah, thanks, I see the problem. This technique is not relying on a null-terminator in the rust string, but it is missing a check for `\0` showing up in the code. If it does show up, it would be treated like the end of the file and cause a problem. I'll fix that.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
jcnelson: More of a design-level question: I see that the lexer will return a `String` for every other token, and rely on the parser to try to convert the `String` into a Clarity `Value` of the corresponding type.  The parser, not the lexer, identifies malformed tokens.  Should that also be happening here, for `uint` and `int`?

clarity/src/vm/ast/parser_v2/mod.rs
obycode: Yes, there should be. When I first designed this, I was avoiding making changes there, but you're right, at this point it would be the right move. 

I also intended to create a type for comments, because there are many things that the tools could do if we maintained the comments in the AST instead of just treating them like whitespace. Are you okay with me adding that into this PR as well if I will already be making changes there?

jcnelson: I'm personally fine with it since I don't think the change will be very big, but let's also ask @kantai.  He knows this part of the codebase better than I do.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: Fixed in dbb2664fb

clarity/src/vm/ast/parser_v2/lexer/mod.rs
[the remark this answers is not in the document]
obycode: Yes, this is just reading `X` in `'X`.

Right, the parser checks that using `parse_standard_principal`, but I had not realized that principals use the Crockford 32 alphabet. Maybe the lexer should catch that, since it would be able to give a more specific error in the case that an illegal character is used. I will modify the checks there.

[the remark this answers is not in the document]
obycode: Fixed in 2d825c5ba

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: Fixed in 2d825c5ba

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: Yes, that makes sense. The lexer should just be verifying the allowed characters are used and the parser should validate the complete token. I will make that change.

clarity/src/vm/ast/parser_v2/mod.rs
obycode: Updated in 2d825c5 to now generate the `InvalidCharPrincipal`.

clarity/src/vm/ast/parser_v2/mod.rs
obycode: Added in 621f4db9d

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: Fixed in a43454bd2

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: Fixed in a43454bd2

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,1957 @@
...
+        self.read_char();
+        loop {
+            if escaped {
+                let ch = match self.next {
+                    '\\' => '\\',
+                    '\"' => '\"',
+                    'n' => '\n',
+                    't' => '\t',
+                    'r' => '\r',
+                    '0' => '\0',
```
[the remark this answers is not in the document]
obycode: Yes, this is the same set from the old parser:

https://github.com/stacks-network/stacks-blockchain/blob/38aa968d7e81d575453ba5253aa8e64a8ac7cec9/clarity/src/vm/ast/parser/mod.rs#L436-L446

I can add others if we decide we need to support others.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: Added a comment in e5d99b691a270764a7b55e09ed40825cb765d3e4

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,1957 @@
...
+                                end_line: self.line as u32,
+                                end_column: self.column as u32,
+                            },
+                            e: LexerError::EmptyUTF8Encoding,
+                        });
+                        self.read_char();
+                        return error_string.to_string();
+                    }
+                    self.read_char();
+                    return code;
```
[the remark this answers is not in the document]
obycode: Sticking with the rule of only validating characters in the lexer and checking higher-level properties in the parser, I think this can stay as-is.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: Fixed in cda9ec15c05757dc68f4ec1a2614cc53e8177817

clarity/src/vm/ast/parser_v2/lexer/mod.rs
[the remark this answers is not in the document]
obycode: This is now actually checked in the caller, `read_utf8_string`. Fixed in cda9ec15c05757dc68f4ec1a2614cc53e8177817.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: Rebased, now 639e4030e348158f63c0b8fbacbb52083b95fb74

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: Rebased, now 639e4030e348158f63c0b8fbacbb52083b95fb74

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: That would reach the next case,

```rust
                } else if is_separator(self.next) {
                    Token::Int("0".to_string())
                }
```

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: Inside a string, they are allowed in the current parser, though I am noticing now that the `\0` will later trigger an error (the `\r` is allowed) from:

https://github.com/stacks-network/stacks-blockchain/blob/38aa968d7e81d575453ba5253aa8e64a8ac7cec9/clarity/src/vm/types/mod.rs#L767-L771

So perhaps instead we should just disallow `\0` in the parser? And also remove `\r` if we don't want it to be supported.

clarity/src/vm/ast/parser_v2/mod.rs
obycode: Oops 😆. d00ff689f

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,2518 @@
...
+            | Token::Less
+            | Token::LessEqual
+            | Token::Greater
+            | Token::GreaterEqual => {
+                let name = ClarityName::try_from(format!("{}", token.token)).unwrap();
+                let mut e = PreSymbolicExpression::atom(name);
+                e.span = token.span;
+                Some(e)
+            }
+            // TODO: For now, comments are being treated as whitespace. In the future,
```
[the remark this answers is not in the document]
obycode: Based on the discussion above, I will add this feature and include it in this PR.

clarity/src/vm/ast/parser_v2/mod.rs
obycode: Removed in a8a3ff390

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,2518 @@
...
+        );
+        assert_eq!(
+            diagnostics[0].spans[0],
+            Span {
+                start_line: 1,
+                start_column: 5,
+                end_line: 1,
+                end_column: 5
+            }
+        );
```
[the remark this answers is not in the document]
obycode: Yes, all of these variants are now legal. Tests added in 29486c63d.

obycode:
Thanks for the review Jude. I've addressed most of the comments and pushed updates to the branch. The remaining items are to add `PreSymbolicExpression` type variants for comments and placeholders.

I just noticed the conflict, so I will also rebase to resolve that.

obycode:
a1a7ffe has two big improvements based on your feedback:
* Adds the placeholder type
  - Used to continue parsing through errors
  - Holds the original string, so can be used to reproduce the source
* Refactors the lexer
  - All lexer tokens are representations of the original string, leaving the further processing for the parser
  - UTF8 strings are a slight rule bender, as the utf8 code points are validated in the lexer and the decoded utf8 string is passed to the parser

I also added the comment type, but it is not used yet. I think after that change to handle comments, it will be ready for re-review. I will also review the code coverage and add new tests as needed after this last change.

clarity/src/vm/ast/parser_v2/mod.rs
obycode: Yes, added.

clarity/src/vm/ast/parser_v2/mod.rs
obycode: Yes, we know for sure that `self.next_token > 0` because the only way to get into `parse_list` is to have already processed a `(`, which would increment the count. I checked other locations where 1 is subtracted as well.

We also have a guarantee that there is always at least one token from the lexer, the `EOF` indicator, so even in the case of an empty string, one token will be processed successfully before reaching the end. This makes the `self.next_token - 1` in `parse_node_or_eof` safe.

clarity/src/vm/ast/parser_v2/mod.rs
obycode: Removed

obycode:
All items should be addressed now, other than the one question about whether we want to change the allowance of `\0` and `\r` characters.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,2207 @@
...
+fn is_separator(ch: char) -> bool {
+    match ch {
+        '(' | ')' | '{' | '}' | ',' | ':' | '.' | ';' | EOF => true,
+        _ => ch.is_ascii_whitespace(),
+    }
+}
+
+fn is_string_terminator(ch: char) -> bool {
+    match ch {
+        '"' | '\n' | EOF => true,
```
jcnelson, line 47: Does this mean that the following contract will parse?

```clarity
"hello world
```

I think you can have "bare" contracts that don't have any lists, but just raw literals.  I just want to make sure that calling `EOF` a string terminator doesn't lead to silliness like this.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
```diff
@@ -0,0 +1,2207 @@
...
+        assert_eq!(
+            lexer.read_token().unwrap().token,
+            Token::Int("123".to_string())
+        );
+        assert_eq!(lexer.diagnostics.len(), 0);
+
+        lexer = Lexer::new("0123", false).unwrap();
+        assert_eq!(
+            lexer.read_token().unwrap().token,
+            Token::Int("123".to_string())
```
jcnelson, line 947: Wait, shouldn't this be `Token::Int("123".to_string())`?  Why does the lexer strip leading 0's?  Isn't that the parser's job?

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,3227 @@
...
+            }
+            None => {
+                self.add_diagnostic(ParseErrors::ExpectedContractIdentifier, span.clone())?;
+                let mut placeholder = PreSymbolicExpression::placeholder(".".to_string());
+                placeholder.span = span;
+                return Ok(placeholder);
+            }
+        };
+
+        let contract_name = match ContractName::try_from(name.clone()) {
```
jcnelson, line 525: Should this code be checking that the contract name's length isn't too long here?  This isn't done in `ContractName::try_from()`.

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,3227 @@
...
+            level: e.e.level(),
+            message: e.e.message(),
+            spans: vec![e.span.clone()],
+            suggestion: None,
+        })
+        .collect();
+    (stmts, diagnostics, parser.success)
+}
+
+pub fn decode_hex(s: &str) -> Result<Vec<u8>, ParseIntError> {
```
jcnelson, line 842: This duplicates `stacks_common::util::hash::hex_bytes`.  Can you use that method instead?

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,3227 @@
...
+                start_line: 1,
+                start_column: 3,
+                end_line: 1,
+                end_column: 3
+            }
+        );
+
+        let (stmts, diagnostics, success) =
+            parse_collect_diagnostics("340282366920938463463374607431768211456 ");
+        assert_eq!(success, false);
```
jcnelson, line 932: This fails due to numeric overflow, right?

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,3227 @@
...
+        );
+        assert_eq!(
+            diagnostics[0].spans[0],
+            Span {
+                start_line: 1,
+                start_column: 1,
+                end_line: 1,
+                end_column: 39
+            }
+        );
```
jcnelson, line 961: Does the parser correctly handle `"0000000000123"` an `"u00000000123"`?

clarity/src/vm/ast/parser_v2/mod.rs
```diff
@@ -0,0 +1,3227 @@
...
+            Span {
+                start_line: 1,
+                start_column: 14,
+                end_line: 1,
+                end_column: 16,
+            }
+        );
+
+        let (stmts, diagnostics, success) = parse_collect_diagnostics("{1: u2, 3: u4}");
+        assert_eq!(success, true);
```
jcnelson, line 1811: Wait, shouldn't this fail? Keys of a tuple can't be numeric, can they?

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: No, this is just used to read through the "end" of a string when an error is encountered. If a `"` is not found before the end of the line or the end of the file, then the string is considered closed. This is only reached when not in "fail fast" mode, when attempting to continue parsing past an error.

clarity/src/vm/ast/parser_v2/lexer/mod.rs
obycode: Yes, good catch. This should have changed with the switch making the lexer return strings for integers. Fixed in 4851c3880.

clarity/src/vm/ast/parser_v2/mod.rs
obycode: Fixed and tested in 9d2634ef9

clarity/src/vm/ast/parser_v2/mod.rs
obycode: Great! Yes, removed in 9324482ec.

clarity/src/vm/ast/parser_v2/mod.rs
obycode: Yes, this number exceeds the range of a 128-bit integer. I'll add a comment.

clarity/src/vm/ast/parser_v2/mod.rs
obycode: The parser allows this, but the type-checker reports an error.

clarity/src/vm/ast/parser_v2/mod.rs
obycode: 5525fcead

clarity/src/vm/ast/parser_v2/mod.rs
obycode: Do you mean `"0000000000123"` with the quotes, as a string, or `0000000000123` as an integer? Both are handled correctly, but I added tests for the int/uint in a52eef3c5.

clarity/src/vm/ast/traits_resolver/mod.rs
```diff
@@ -176,24 +177,33 @@ impl TraitsResolver {
...
         let function_name = function_name.match_atom()?;
         let define_type = DefineFunctions::lookup_by_name(function_name)?;
-        Some((define_type, args))
+        Some((define_type, args.to_vec()))
     }
 
     fn probe_for_generics(
         &mut self,
-        exprs: &[PreSymbolicExpression],
+        exprs: Vec<&PreSymbolicExpression>,
```
jcnelson, line 195: Why is his now a `Vec`?

jcnelson (approved):
This LGTM now.  Thanks @obycode!

obycode:
I just completed some more targeted timing of the parsers and it showed that this v2 parser is more than 8x faster than the v1 parser!

obycode:
@jcnelson I just noticed that the unit tests failed due to a seemingly random error from codecov, so I triggered a re-run of those tests. Please help me merge after that testing is complete.

PR #3151, opened 2022-05-25
fix: use latest clarinet test lib

lgalabru:
The latest version of clarinet (that this repo is using, by using the tag `latest`) is not backward compatible with previous versions of the typescript bridge.
This PR is updating the version of the library being imported.


gregorycoppola (approved):
Thanks for fixing!

obycode (approved):

PR #3152, opened 2022-05-26
Release 2.05.0.2.1 (hotfix: check chain work when processing reorgs)

jcnelson:
Okay, this is now ready for review.

This is a hotfix to the SPV client that addresses the following bugs:

* The SPV client now tracks the best chain work for each Bitcoin difficulty adjustment interval it has ever seen.  Any attempt to store headers that would lead to a _decrease_ in total chain work will result in an error which will cause the node to re-attempt to download those headers indefinitely until they have at least the best-seen work.

* The SPV client will verify that newly-obtained headers after the initial header sync have timestamps that are within 2 hours of the current epoch time.

* The SPV client will verify that each Bitcoin header has a timestamp that is strictly greater than the median of the past 11 Bitcoin blocks.

* The SPV client will reject a block that has a strictly higher value than the target for the difficulty interval, instead of an equal or higher target.

By far the most work in this PR concerns point #1.  To build confidence in the implementation, I added a unit test that is disabled by default, but if set up with the proper environment variables, will download the mainnet Bitcoin block headers and verify that the calculated chain work in each interval is equal to values I manually extracted from `bitcoind`.  In addition, I took PoC bad block headers from the Immunefi bug report and created two unit tests whereby:
   
   * A new chain with the bad blocks is discovered and compared to the main chain with the good blocks (this new chain is verified ignored because it has lower total chain work), and
   
   * A new chain with the good blocks is discovered and compared to the main chain with the bad blocks (this new chain is verified accepted because it has a higher total chain work).
   
There are a couple of modifications to `BitcoinIndexer::find_bitcoin_reorg()` that necessitated changes to `Burnchain::sync_with_indexer()` that are difficult to test in both unit and integration tests.  I am currently in the process of filling this end-to-end test checklist:

   - [X] A node with this patch is able to _resume_ from existing mainnet chainstate and reach the chain tip
   
   - [X] A node with this patch is able to _boot from genesis_ on mainnet and reach the chain tip
   
   - [X] A node with this patch is able to _resume_ from existing testnet chainstate and reach the chain tip
   
   - [X] A node with this patch is able to _boot from genesis_ on testnet and reach the chain tip
   
Once these end-to-end tests pass, I think we can make a hotfix release.  Once the first end-to-end test passes (resuming from existing mainnet chainstate), I think we can cut a release candidate.

---
Reminder to anyone reading this: This project participates in [Immunefi](https://immunefi.com/bounty/stacks/) to pay bounties for bugs such as this one.  This bug was reported via Immunefi and is confirmed to exist, as this PR proves.


stacks-common/src/util/uint.rs
```diff
@@ -141,6 +142,68 @@ macro_rules! construct_uint {
...
             }
+
+            /// as big-endian byte array
+            pub fn to_u8_slice_be(&self) -> [u8; $n_words * 8] {
+                let mut ret = [0u8; $n_words * 8];
+                for i in 0..$n_words {
+                    let bytes = self.0[i].to_le_bytes();
+                    for j in 0..bytes.len() {
+                        ret[$n_words * 8 - 1 - (i * 8 + j)] = bytes[j];
+                    }
```
kantai, line 153: I think using `to_be_bytes()` makes this a little easier to follow

```suggestion
                    let word_end = $n_words * 8 - (i * 8);
                    let word_start = word_end - 8;
                    ret[word_start..word_end].copy_from_slice(&self.0[i].to_be_bytes());
```

src/burnchains/bitcoin/indexer.rs
```diff
@@ -385,47 +387,140 @@ impl BitcoinIndexer {
...
-                .read_block_header(start_block)?
-                .expect(&format!("BUG: missing block header for {}", start_block));
-            reorg_spv_client.insert_block_headers_before(start_block - 1, vec![start_header])?;
+            if start_block > BLOCK_DIFFICULTY_CHUNK_SIZE {
+                if remove_old {
+                    // set up a .reorg db
+                    // * needs the last difficulty interval of headers (note that the current
+                    // interval is `start_block / BLOCK_DIFFICULTY_CHUNK_SIZE - 1).
+                    // * needs the last interval's chain work calculation
+                    let interval_start_block = start_block / BLOCK_DIFFICULTY_CHUNK_SIZE - 2;
```
kantai, line 456: What is `interval_start_block`? If the interval is 500, and `start_block` is 502, this is -2? Should that panic or underflow?

src/burnchains/bitcoin/indexer.rs
```diff
@@ -600,17 +696,102 @@ impl BitcoinIndexer {
...
-                self.setup_reorg_headers(&orig_spv_client, reorg_headers_path, start_block)?;
+                self.setup_reorg_headers(&orig_spv_client, reorg_headers_path, start_block, false)?;
         }
 
-        debug!("Bitcoin headers history is consistent up to {}", new_tip);
+        if check_chain_work {
+            let reorg_total_work = reorg_spv_client.update_chain_work()?;
+            let orig_total_work = orig_spv_client.get_chain_work()?;
+
+            debug!("Bitcoin headers history is consistent up to {}. Orig chainwork: {}, reorg chainwork: {}", new_tip, orig_total_work, reorg_total_work);
```
kantai, line 706: Looks like a good case for k-v syntax

src/burnchains/bitcoin/indexer.rs
```diff
@@ -600,17 +696,102 @@ impl BitcoinIndexer {
...
+
+            if orig_total_work < reorg_total_work {
+                let reorg_tip = reorg_spv_client.get_headers_height()?;
+                let hdr_reorg = reorg_spv_client
+                    .read_block_header(reorg_tip - 1)?
+                    .expect("FATAL: no tip hash for existing chain tip");
+                info!(
+                    "New canonical Bitcoin chain found! New tip is {}",
+                    &hdr_reorg.header.bitcoin_hash()
+                );
```
kantai, line 716: Does this log line appear on every new bitcoin block, or only on reorgs?

kantai (commented):
This looks generally good to me -- I had one question about some of the potentially underflowing math in `bitcoin::indexer` which I think needs to be addressed.

On top of that, I think this needs to update the `CHANGELOG.md` with a new release section.

gregorycoppola (commented):
Many lines are showing up as "not covered by test" in the code coverage, e.g.:

* https://app.codecov.io/gh/stacks-network/stacks-blockchain/compare/3152/diff#D3L243
* https://app.codecov.io/gh/stacks-network/stacks-blockchain/compare/3152/diff#D3L375
* https://app.codecov.io/gh/stacks-network/stacks-blockchain/compare/3152/diff#D3L990

Is this an artifact of CodeCov or are they really not covered?

src/burnchains/bitcoin/indexer.rs
jcnelson: Only on reorgs.  This can only get hit if the newly-discovered headers (the "reorg" headers) have more total work than the headers currently in the headers DB (the "orig" headers).

src/burnchains/bitcoin/indexer.rs
jcnelson: This should be a saturating subtraction.

stacks-common/src/util/uint.rs
```diff
@@ -671,4 +734,21 @@ mod tests {
             Uint256([0, 0xDEADBEEFDEADBEEF, 0xDEADBEEFDEADBEEF, 0])
         );
     }
+
+    #[test]
+    pub fn hex_codec() {
```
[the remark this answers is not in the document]
jcnelson: Will do

stacks-common/src/util/uint.rs
jcnelson: Will do

src/burnchains/bitcoin/indexer.rs
jcnelson: Will fix

jcnelson:
@gregorycoppola I'm not seeing what you're seeing?

gregorycoppola:
@jcnelson

I think the linking to CodeCov system is kind of buggy/confusing, because when I click the link it doesn't bring me to the line I started with. 

I'm guessing you mean you don't know which lines I'm referring to (but maybe you mean you see the lines but you think they are actually covered).

I think the `D3` in the link revers to the file `src/burnchains/bitcoin/spv.rs`, because it's the third file on the diff list.

I'm attaching some screen shots.
image
image
image

jcnelson:
> @jcnelson i'm not trying to create busy work in the middle of a hot fix.. just checking what was going on.

Ah, sorry, I didn't realize how dismissive I sounded in that comment!  I'm glad you brought the code coverage up -- it's important that we do get this covered in CI.  I was only trying to get the reviewers' permission to do so in a separate PR :)

kantai (approved):
LGTM!

Issue #3101, opened 2022-04-18
[chainstate] Add support for pay-to-contract in coinbase transaction variant


jcnelson:
Make it so a coinbase transaction can pay out to a smart contract address, instead of just a standard address.

jcnelson:
This will go further: make it possible to supply an alternative recipient address (contract or standard) in the coinbase tx.  Then, miners can pay their block rewards to cold storage.

Issue #3159, opened 2022-06-02
[chore] refactor tests in `src/chainstate/stacks/miner.rs` into one or more files


jcnelson:
Lots of these tests aren't really related to mining, but depend on test infrastructure in this file.  They should be refactored, so that tests can live closer in the codebase to the code they test.

jcnelson:
This is completed in `next`

PR #3164, opened 2022-06-08
Feat/coinbase pay to contract

jcnelson:
EDIT: this PR now supports building coinbase transactions that pay to either a smart contract or an alternative standard principal.  The latter permits the miner to pay to a cold storage address automatically.

This PR implements #3101 by altering the `TransactionPayload::Coinbase(..)` type variant to include an optional ~~smart contract address~~ alternative recipient address, be it a contract or standard principal address.  If given, the encoded transaction payload will have ID `0x05`, and will include a serialized ~~smart contract identifier~~ principal.  When processed in epoch 2.1, the block reward will be transferred to this ~~smart contract~~ address instead of the miner.  In addition, I have updated the node software to accept a ~~`pay_to_contract` contract~~ contract or standard address under the `[miner]` heading called `block_reward_address`, which if given, will cause the node to create ~~pay-to-contract~~ alternative recipient coinbases for blocks mined after epoch 2.1 (the node ignores this field in epoch 2.05 and earlier).

I have added test coverage for:

* the new transaction encoding
* the static transaction validation tests (which are now epoch-gated to check for this new variant)
* the account crediting logic
* the miner's ability to produce a valid history of blocks that pay to a smart contract OR standard principal
* the relayer's ability to reject pay-to-contract OR pay-to-alt-principal blocks that are mined before epoch 2.1
* the node's ability to mine pay-to-contract OR pay-to-alt-principal blocks only once epoch 2.1 goes live

The code includes the relevant feature gating to ensure that a ~~pay-to-contract~~ pay-to-alt-principal block can only be successfully preprocessed in epoch 2.1.  This prevents a Stacks 2.1-series node from accepting a ~~pay-to-contract~~ pay-to-alt-principal block at all before epoch 2.1 goes live.  But, this differs slightly from how a 2.05-series node will handle such a block -- the 2.05-series node will not even be able to decode it, thereby making any attempt to push such a block to it fail with either an HTTP 400 (if sent via HTTP) or a temporary ban (if sent via p2p).  I can make the 2.1-series node just as strict by adding feature gates to the p2p and HTTP endpoints, but I'm not sure it will make an appreciable difference in safety.  Also, I'm not sure it's a good idea to make 2.1-series nodes ban each other for this, since if it turns out there's an off-by-one error somewhere in the gating logic that causes nodes to produce these blocks before they're supported, the nodes could accidentally all ban each other, leading to a catastrophic network partition.  Open to feedback on this.


pavitthrap (approved):
lgtm!

clarity/src/vm/docs/mod.rs
```diff
@@ -1470,7 +1470,9 @@ from the block hash and the history of accepted PoX operations.  This is also th
 `header-hash`: This property returns a `(buff 32)` value containing the header hash of a Stacks block, given a Stacks chain height.  **WARNING* this hash is
 not guaranteed to be globally unique, since the same Stacks block can be mined in different PoX forks.  If you need global uniqueness, you should use `id-header-hash`.
 
-`miner-address`: This property returns a `principal` value corresponding to the miner of the given block.
+`miner-address`: This property returns a `principal` value corresponding to the miner of the given block.  **WARNING** In Stacks 2.1, this is not guaranteed to 
+be the same `principal` that received the block reward, since Stacks 2.1 supports coinbase transactions that pay the reward to a contract address.  This is merely
+the address of the `principal` that produced the block.
```
kantai, line 1475: I think instead of a `**WARNING**`, this should just be a `Note:`

src/chainstate/stacks/block.rs
```diff
@@ -565,8 +565,35 @@ impl StacksBlock {
         }
     }
 
+    /// Verify that all transactions are supported in the current epoch
+    pub fn validate_transactions_static_epoch(
+        txs: &Vec<StacksTransaction>,
```
kantai, line 570: In most cases, it's better to use `&[T]` instead of `&Vec<T>` in arguments, since `Vec` will autoref to that anyways, and then you have more freedom to use slices in tests, etc.

src/chainstate/stacks/block.rs
```diff
@@ -565,8 +565,35 @@ impl StacksBlock {
...
+    ) -> bool {
+        if epoch_id < StacksEpochId::Epoch21 {
+            // nothing new since the start of the system is supported.
+            // Expand this list of things to check for as needed.
+            // * no pay-to-contract coinbases
+            for tx in txs.iter() {
+                if let TransactionPayload::Coinbase(_, ref contract_opt) = &tx.payload {
+                    if contract_opt.is_some() {
+                        // not supported
+                        error!("Coinbase pay-to-contract not supported before Stacks 2.1");
```
kantai, line 581: Can you add the txid to the log line?

testnet/stacks-node/src/tests/epoch_21.rs
```diff
@@ -469,6 +476,34 @@ fn transition_adds_burn_block_height() {
     coord_channel.stop_chains_coordinator();
 }
 
+#[test]
+#[ignore]
+fn transition_adds_pay_to_contract() {
```
kantai, line 481: Can you add the new test to the CI?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4126,14 +4154,25 @@ impl StacksChainState {
...
+            // pre-2.1, always pay to the origin
+            None
         };
 
+        // N.B. a `MinerPaymentSchedule` that pays to a contract can never be created before 2.1,
+        // per the above check (and moreover, a Stacks block with a pay-to-contract coinbase would
+        // not become valid until after 2.1 activates).
         let miner_reward = MinerPaymentSchedule {
             address: miner_addr,
+            recipient_contract: recipient_contract,
```
kantai, line 4175: Can this field instead be `recipient` and _always_ used (i.e., rather than being `Option<QualifiedContractPrincipal>`, it would be `PrincipalData`)?

That way, we don't have a conditional logic field in the database (and in this struct) which requires repeating the check in lines 4159-4167 any time we want to figure out the recipient of the scheduled payment.

src/chainstate/stacks/miner.rs
```diff
@@ -5794,19 +5794,31 @@ pub mod test {
     }
 
     pub fn make_coinbase(miner: &mut TestMiner, burnchain_height: usize) -> StacksTransaction {
```
kantai, line 5796: I think it's time for #3159 -- this is another PR that would not have any changes in `chainstate::stacks::miner` except for the fact that the test framework is there.

src/chainstate/stacks/db/accounts.rs
```diff
@@ -35,19 +35,38 @@ use crate::types::chainstate::{StacksAddress, StacksBlockId};
 
 use crate::core::StacksEpochId;
 
+/// A record of a coin reward for a miner.  There will be at most two of these for a miner: one for
+/// the coinbase + block-txs + confirmed-mblock-txs, and one for the produced-mblock-txs.  The
+/// latter reward only stores the produced-mblock-txs, and is only ever stored if the microblocks
+/// are ever confirmed.
 #[derive(Debug, Clone, PartialEq)]
 pub struct MinerReward {
```
kantai, line 43: Thanks for adding the rustdocs!

src/chainstate/stacks/db/mod.rs
```diff
@@ -132,6 +132,7 @@ pub struct StacksAccount {
 #[derive(Debug, Clone, PartialEq)]
 pub struct MinerPaymentSchedule {
     pub address: StacksAddress,
+    pub recipient_contract: Option<QualifiedContractIdentifier>,
```
kantai, line 135: Is there any use for specifying a different `StacksAddress` as the recipient, rather than a contract? As in, what if miner Alice wanted coinbase rewards to go to Bob even if Bob isn't a contract?

gregorycoppola:
Removing myself since there'll be two approvals when kantai's comments are addressed.

src/chainstate/stacks/db/blocks.rs
jcnelson: I'm not sure I understand the question?  We're going to need to gate this either way, since we cannot permit the _possibility_ that the current or future node implementation will allow a pay-to-contract coinbase transaction to be processed.  Representing the recipient contract as a `PrincipalData` instead of `Option<QualifiedContractPrincipal>` does not free you from this.

src/chainstate/stacks/db/mod.rs
jcnelson: I thought about this too, but I can't think of a compelling use-case.  If Alice wanted to pay Bob the block reward, then why can't Alice just fund Bob's block-commit transaction?

clarity/src/vm/docs/mod.rs
jcnelson: Done

src/chainstate/stacks/block.rs
jcnelson: Done

src/chainstate/stacks/block.rs
jcnelson: Done

testnet/stacks-node/src/tests/epoch_21.rs
jcnelson: Done

src/chainstate/stacks/db/mod.rs
kantai: Because they don't share keys -- like one situation could be that the coinbase rewards are going to a "cold" Stacks wallet, which you wouldn't want to use as the miner address (because the miner address must be hot).

src/chainstate/stacks/db/blocks.rs
kantai: The idea is that the `MinerPaymentSchedule` struct just specifies a recipient principal -- we construct valid MinerPaymentSchedule instances when processing coinbase transactions, and then when processing the scheduled payment, we just debit `recipient`.

This leads to a clearer database schema: to figure out the recipients from the scheduled payments table, you just look at the recipient column, instead of "look at the miner address column and the recipient_contract column, if the second is null, then use the first, otherwise use the second".

Regarding the gates: you would gate the constructor for scheduled payments. If a scheduled payment struct is constructed that is invalid, that's an implementation error, and should panic.

src/chainstate/stacks/miner.rs
kantai: Either is fine, but the follow-on PR must be merged soon after this PR.

src/chainstate/stacks/db/blocks.rs
jcnelson: Gotcha.  I've updated the code so that:

* The `recipient` column is always set on future DB inserts, but if it's not set, it defaults to `address`
* `recipient` is now a `PrincipalData` in `MinerPaymentSchedule` and `MinerReward`.
* The `TransactionPayload::Coinbase` variant takes an `Option<PrincipalData>`, and will serialize it as a `Value` via the `StacksMessageCodec` implementation for `Value`.
* Test coverage is expanded to verify that a block reward can be sent to either a standard or contract principal address

src/chainstate/stacks/db/mod.rs
jcnelson: Gotcha.  I've made it so that the coinbase can pay to either a contract or a standard principal.

src/chainstate/stacks/miner.rs
jcnelson: I've opened the PR that does this here: https://github.com/stacks-network/stacks-blockchain/pull/3180.  It's against this PR, so let's get it reviewed and merged first and then we can merge this PR to `next`.

kantai (approved):
LGTM!

jcnelson:
Thanks for the approvals!  Let's get #3180 merged to this PR, and then I'll merge this all into `next`.

Issue #2538, opened 2021-03-23
PoX Updates: `.pox-info` contract for native information getters


kantai:
As part of Stacks 2.1, we should add an extension contract for obtain PoX information that must be implemented as native functions:

* Getter for address lock information
* Getter for whether an address is able to "early unlock"
* Getter for current cycle reward set information

jcnelson:
Adding to this:

* Getter for how much BTC was committed for the winning block
* Getter for how much BTC was committed for the entire sortition (requires fixing #2492)

We need these things to figure out what the BTC-denominated price of STX is on-chain.

Drilling down into "address lock information" and "current reward cycle set information", what do you think of these?

* Getter for how many PoX addresses a Stacker clinched
* Getter for how much a Stacker has been paid in BTC so far in this cycle
* Getter for how many reward slots a Stacker will receive this cycle
* Getter for how many reward slots a Stacker has been paid out to this cycle (including instances where no BTC was paid)

aulneau:
Something that might be nice would be some getters for delegate information. Right now it's pretty hard and a manual process to get data around pools that use contracts to call the pox contract. I have to manually know of and fetch data for each contract. If there is some way to make this clearer from the pox extension contract itself, that'd be fantastic!

hstove:
Another request would be:

- Getter for how much a Stacker was paid in BTC in a particular block

hstove:
For this feature:

> Getter for how much a Stacker has been paid in BTC so far in this cycle

It would be great to also allow `cycle` to be a parameter, so you could query for past cycles

jcnelson:
Added already: `stx-get-account`

jcnelson:
Regarding the getter for address lock information, this is the interface @lgalabru proposed:

```
What's our take on adding a pox key to get-block-info?, such as (get-block-info? pox <block-id>) returns a tuple (list 2 { recipient: principal, total-transfered: uint })?
```

jcnelson:
Insofar as the minimum viable changes for `.pox-2`, simply exposing all of the contract state via getters should be sufficient.  Querying data that is not in `.pox-2` like the reward addresses should be a separate issue.

pavitthrap:
Request to make `current-pox-reward-cycle` and `burn-height-to-reward-cycle` non private for the work in https://github.com/stacks-network/stacks-blockchain/pull/3035

PR #3166, opened 2022-06-09
Feat/pox getters

jcnelson:
This addresses #2538 by adding getters in `.pox-2` for state that is currently invisible.  It also adds basic unit tests to verify that the getters work as you'd expect.


pavitthrap:
For the work in https://github.com/stacks-network/stacks-blockchain/pull/3035, could you make `current-pox-reward-cycle` and `burn-height-to-reward-cycle` non private?

gregorycoppola (approved):

PR #3178, opened 2022-06-23
chore: add documentation to the config

igorsyl:
Add documentation to config and enforce using rustdoc in GitHub Actions CI

https://github.com/stacks-network/stacks-blockchain/issues/3067


.github/workflows/ci.yml
```diff
@@ -91,6 +91,15 @@ jobs:
...
         run: docker build -f ./.github/actions/bitcoin-int-tests/Dockerfile.rustfmt .
 
+  rustdoc:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run rustdoc check
+        env:
+          DOCKER_BUILDKIT: 1
+        run: docker build -f ./.github/actions/bitcoin-int-tests/Dockerfile.rustdoc .
```
kantai, line 101: The only thing I would add here is that this use docker's `-o` option to export the data from the dockerfile. It shouldn't require much update here, but it'll require a little change in the dockerfile side as well.

kantai: We do something similar in the `docs-pr.yml`

igorsyl: The initial intent was to use rustdoc to enforce documentation but not to generate or export it. I can follow your suggestion to publish the config doc as well.

PR #3183, opened 2022-07-01
feat: Add tx/block to prometheus outputs

gregorycoppola:
### Description

Adds output to prometheus for "transactions in last block".

We don't currently have any way to get this value from existing outputs, and want to add it to the dash.

### Checklist
- [x] No tests for this!
- [x] Changelog is updated

### DIFF
--- a/src/monitoring/prometheus.rs
+++ b/src/monitoring/prometheus.rs
@@ -116,7 +116,7 @@ lazy_static! {
         "`execution_cost_runtime` for the last block observed."
     )).unwrap();
 
-    pub static ref LAST_BLOCK_TRANSACTION_COUNT: IntGauge = register_gauge!(opts!(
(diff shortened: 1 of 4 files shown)


kantai (approved):
LGTM!

src/monitoring/prometheus.rs
```diff
@@ -116,6 +116,11 @@ lazy_static! {
         "`execution_cost_runtime` for the last block observed."
     )).unwrap();
 
+    pub static ref LAST_BLOCK_TRANSACTION_COUNT: Gauge = register_gauge!(opts!(
```
pavitthrap, line 119: I think you could use an IntGauge here.

pavitthrap (approved):

src/monitoring/prometheus.rs
gregorycoppola: indeed

PR #3199, opened 2022-07-15
Update github action CI workflow

wileyj:
### Applicable issues
- related PR for clippy (@igorsyl): https://github.com/stacks-network/stacks-blockchain/pull/3186
- fixes https://github.com/stacks-network/stacks-blockchain/issues/3148

### Additional info
Lots of proposed changes here so I'll do my best to articulate them:
- migrates the existing ci.yml workflow into a more modular approach with *semi*-reusable workflow files 
- updates all Debian builds to use bullseye (latest debian)
- adds arm64 binaries/docker image
- update all Dockerfiles to use ramdisk for building
- create checksum of binary artifacts for release
- build all release images from binaries (PR's and branch builds still build from source)
- create checksum of binary artifacts for release
- adds clippy and crate advisory actions

The workflow behaves similarly to how the existing workflow runs - manual execution with a tag will create a new release. on PR's,  a simplified workflow is run. 

The `./build-scripts` Dockerfiles are all updated to be resuable but also renamed to reflect how the arches appear in the build process, i.e. `linux-glibc-arm64`. 

Clippy has been added, and right now it's set to run on PR's to develop - as of now this action is quite messy. Opinions here are welcome, it may be desired to remove entirely, or perhaps run it on a cron so the data is available. Using defaults, it's very pedantic but it can be adjusted to ignore trivial lint complaints. 

A crate advisory action is also added (I believe it's something that should remain in some form to ensure we're keeping up with dependency updates) - currently this action is to be scheduled nightly at 0330 UTC, and will create a new issue if any crate has a listed advisory. 

Last, the build process is changed slightly:
- sha1sum of all archives is stored in the file `CHECKSUMS.txt` which is included in the github release
- All release images are built using the binary archives for the specified architecture
- `./github/actions/dockerhub/**` is created to build the new images for both Alpine and Debian

#### Caveats
- the linux-musl-arm64 build is using an external musl-cross build image (for now), no other way to natively build for this arch. has been found (https://musl.cc/)
- `bitcoin-tests.yml` is set to run only when PR's are opened to the master branch (this is debatable, but in testing i felt it was running too often)
- source build will trigger on a push to the default branch automatically
- naming of the tags is slightly adjusted from something like `x.y.z-stretch` -> `x.y.z-debian`

#### Examples of builds
- Example images built during testing: https://hub.docker.com/repository/docker/wileyj/stacks-blockchain/tags?page=1&ordering=last_updated
- Issues created from crate advisories: https://github.com/wileyj/stacks-blockchain/issues
- Test build of tag `2.05.0.2.3`: https://github.com/wileyj/stacks-blockchain/actions/runs/2673483049
- Merge to master: https://github.com/wileyj/stacks-blockchain/actions/runs/2673477039
- Example clippy action: https://github.com/wileyj/stacks-blockchain/actions/runs/2673230254


.github/actions/dockerhub/Dockerfile.alpine-binary
```diff
@@ -0,0 +1,22 @@
+FROM --platform=${TARGETPLATFORM} alpine as builder
+# Use a small image to download and extract the release archive
+
+ARG TAG
+ARG BIN_ARCH
+ARG TARGETPLATFORM
+ARG BUILDPLATFORM
+ARG TARGETARCH
+ARG TARGETVARIANT
+ARG REPO=wileyj/stacks-blockchain
```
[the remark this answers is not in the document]
wileyj: it does! good catch  - my search-fu must have failed here. it's been fixed in most recent commit

.github/workflows/build-source-binary.yml
```diff
@@ -0,0 +1,65 @@
...
+
+jobs:
+  artifact:
+    if: ${{ inputs.tag != '' }}
+    name: Create Artifacts
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: ${{ inputs.parallel_jobs}}
+      matrix:
+        platform: ${{fromJson(inputs.arch)}}
```
CharlieC3, line 35: Clever!

.github/workflows/github-release.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Upload Release
+        id: upload_release
+        uses: softprops/action-gh-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        with:
+          name: Release ${{ github.event.inputs.tag || github.ref }}
+          tag_name: ${{ github.event.inputs.tag || github.ref }}
+          draft: false
+          prerelease: true
```
CharlieC3, line 56: Is this intentional? Should this be parameterized?

.github/workflows/image-build-alpine-binary.yml
```diff
@@ -0,0 +1,82 @@
...
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        id: docker_buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Docker Metadata
+        id: docker_metadata
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            ${{ env.REPO_OWNER }}/${{ github.event.repository.name }}
```
CharlieC3, line 55: I think this can be reduced to `${{ github.repository }}`, and the `REPO_OWNER` env var removed.

.github/workflows/image-build-debian-binary.yml
```diff
@@ -0,0 +1,87 @@
...
+        uses: docker/setup-buildx-action@v2
+      - name: Extract branch name
+        id: extract_branch
+        run: echo "BRANCH_NAME=$(echo ${GITHUB_REF#refs/heads/})" >> $GITHUB_ENV
+      - name: Docker Metadata
+        id: docker_metadata
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            ${{ env.REPO_OWNER }}/${{ github.event.repository.name }}
```
CharlieC3, line 61: I think this can be reduced to `${{ github.repository }}`, and the `REPO_OWNER` env var removed.

.github/workflows/image-build-debian-source.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Extract branch name (PR)
+        id: extract_branch_pr
+        if: ${{ github.event_name == 'pull_request' }}
+        run: echo "BRANCH_NAME=$(echo ${GITHUB_HEAD_REF})" >> $GITHUB_ENV
+      - name: Docker Metadata
+        id: docker_metadata
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            ${{ env.REPO_OWNER }}/${{ github.event.repository.name }}
```
CharlieC3, line 71: I think this can be reduced to `${{ github.repository }}`, and the `REPO_OWNER` env var removed.

.github/workflows/image-build-alpine-binary.yml
```diff
@@ -0,0 +1,82 @@
...
+          tags: |
+            type=raw,value=latest,enable=${{ inputs.tag != '' && (github.ref == format('refs/heads/{0}', github.event.repository.default_branch) )}}
+            type=raw,value=${{ inputs.tag }},enable=${{ inputs.tag != '' }}
+            type=ref,enable=true,priority=600,event=tag
```
CharlieC3, line 59: Not sure `priority=600` is doing anything here since the other tags don't specify their priority.

.github/workflows/image-build-debian-source.yml
```diff
@@ -0,0 +1,93 @@
...
+        uses: docker/build-push-action@v3
+        with:
+          file: ./.github/actions/dockerhub/Dockerfile.${{ inputs.linux_version }}-${{ inputs.build_type }}
+          platforms: ${{ inputs.docker_platforms }}
+          tags: ${{ steps.docker_metadata.outputs.tags }}
+          labels: ${{ steps.docker_metadata.outputs.labels }}
+          build-args: |
+            STACKS_NODE_VERSION=${{ env.GITHUB_SHA_SHORT }}
+            GIT_BRANCH=${{ env.GITHUB_REF_SHORT }}
+            GIT_COMMIT=${{ env.GITHUB_SHA_SHORT }}
```
CharlieC3, line 92: Seems `BUILDKIT_PROGRESS=plain` is missing here, but present in the other two docker build workflows. Should it be added?

.github/workflows/ci.yml
```diff
@@ -1,311 +1,177 @@
...
+  #   Only run if (a tag was NOT passed in) or (we're building a non-master branch which isn't a PR) or (this is a PR with no input tag)
+  # Only run if:
+  #    - tag is not provided
+  #    - OR
+  #        - non-default branch
+  #        - AND Not a PR
+  build-branch:
+    if: ${{ inputs.tag == '' || (github.ref != format('refs/heads/{0}', github.event.repository.default_branch) && !contains(github.ref, 'refs/pull')) }}
+    name: Docker Debian (Source)
+    uses: stacks-network/stacks-blockchain/.github/workflows/image-build-debian-source.yml@master
```
CharlieC3, line 168: Alternatively, you could make the `build-source` job run on all commits to build the binaries, [upload them to the workflow instance](https://github.com/actions/upload-artifact), then [download them](https://github.com/actions/download-artifact) to use them in the "binary" [docker images](https://github.com/stacks-network/stacks-blockchain/tree/806d99556432b9755d332d2526cf2e0d88b87847/.github/actions/dockerhub). Unless there's a reason you'd be unable to do this that I'm not seeing, you could eliminate duplicated code this way and ensure the same process is used between building PRs or a release.

[See the Clarinet CI workflow for an example if interested](https://github.com/hirosystems/clarinet/blob/develop/.github/workflows/ci.yaml)

.github/actions/dockerhub/Dockerfile.debian-source
```diff
@@ -0,0 +1,21 @@
...
+
+ARG STACKS_NODE_VERSION="No Version Info"
+ARG GIT_BRANCH='No Branch Info'
+ARG GIT_COMMIT='No Commit Info'
+ARG BUILD_DIR=/build
+ARG TARGET=x86_64-unknown-linux-gnu
+WORKDIR /src
+
+COPY . .
+
```
CharlieC3, line 11: I think this is missing:
```
RUN apt-get update && apt-get install -y git
```
since this line is in the [Dockerfile.linux-glibc-x64](https://github.com/stacks-network/stacks-blockchain/pull/3199/files#diff-a915f90fa919dda92131d8945ab78933a02ea43d7e3d373347cac02c4c045789) build script.

CharlieC3 (changes requested):
Amazing work! It's already polished, but I left some comments for minor changes, questions, and alternative ideas in some places if interested.

.github/workflows/bitcoin-tests.yml
```diff
@@ -1,36 +1,55 @@
...
-#   - PRs are opened
+#   - PRs are (re)opened against master branch
+
 on:
   pull_request:
+    branches:
+      - master
+    types:
+      - opened
+      - reopened
```
kantai, line 16: Does this mean that these tests wouldn't run for PRs against other branches? That would cut down on worker usage, but these tests are generally pretty helpful for PRs.

kantai (commented):
These changes look good to me -- I'll defer to Charlie and Jude for final approvals.

My only request is that you update the README.md (or add a RELEASING.md) with two instructions:

1. How to tag a release (I think this is already kind of mentioned in the release checklist, but it could have it's own subsubsection in the README), and what happens automatically when the release is tagged (docker images pushed? binaries included in the git release? changelog?)
2. How to trigger a "feature" build: if I have a PR, and I want to publish a docker image with a specific tag (not a version number) that doesn't create a new release, how can I do that? I think before these changes, that was done "automatically" on every `feat/...` PR, but with these changes, I think it requires a manual trigger, right?

.github/actions/dockerhub/Dockerfile.alpine-binary
```diff
@@ -0,0 +1,22 @@
...
+ARG BIN_ARCH
+ARG TARGETPLATFORM
+ARG BUILDPLATFORM
+ARG TARGETARCH
+ARG TARGETVARIANT
+ARG REPO=stacks-network/stacks-blockchain
+
+RUN case ${TARGETARCH} in \
+    "amd64")  BIN_ARCH=linux-musl-x64  ;; \
+    "arm64")  BIN_ARCH=linux-musl-arm64  ;; \
```
jcnelson, line 14: Should this error out if `${TARGETARCH}` is not matched by either of these cases?  Same question for `Dockerfile.debian-binary` below.

.github/actions/dockerhub/Dockerfile.debian-source
```diff
@@ -0,0 +1,21 @@
...
+RUN --mount=type=tmpfs,target=${BUILD_DIR} cp -R /src/. ${BUILD_DIR}/ \
```
jcnelson, line 12: You're going to need ~8GB disk space to comfortably build the release.  Is this guaranteed to be the case here?

.github/workflows/bitcoin-tests.yml
jcnelson: Yeah, these tests should run on each PR before merging to `develop`, `next`, or `master`.

.github/workflows/clippy.yml
```diff
@@ -0,0 +1,42 @@
+##
+## Perform Clippy checks - currently set to defaults
+##   https://github.com/rust-lang/rust-clippy#usage
+##   https://rust-lang.github.io/rust-clippy/master/index.html
+##
+
+name: Clippy Checks
+
+# Only run when:
+#   - PRs are (re)opened against develop branch
```
jcnelson, line 10: Should probably also run against the `next` and `master` branches

.github/workflows/ci.yml
```diff
@@ -1,311 +1,177 @@
...
+  # Build source binaries
+  # Only run if:
+  #    - tag is provided
+  #    - AND
+  #        - this is the default branch
+  #        - AND Not a PR
+  build-source:
+    if: ${{ inputs.tag != '' && (github.ref == format('refs/heads/{0}', github.event.repository.default_branch) && !contains(github.ref, 'refs/pull')) }}
+    name: Build Binaries
+    uses: stacks-network/stacks-blockchain/.github/workflows/build-source-binary.yml@master
```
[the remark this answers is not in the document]
wileyj: A fair point - I did consider that route at first, but ultimately i decided i didn't want to change the process of how we build things too drastically with this PR. most of the workflows are nearly identical to the existing workflow, just cleaned up a bit to make it easier to address future changes. 

The second reason i chose not to go that route was because then there would be another repo that would have to be maintained and I didn't want to add more work (objectively) to the changes here. 

The main focus I had was to first build the arm64 docker image, and the second was to perform a bit of a cleanup. 
I added in clippy and the dependency actions as possible nice to haves since they've come up in conversations. 

The next iteration of CI for this repo could absolutely use composite actions as you've suggested, I simply felt it was a bridge too far considering the goals I set out with. "don't rock the boat *too* hard"

.github/workflows/build-source-binary.yml
wileyj: thanks! i tried to find ways to prevent hard-coding values in the workflows or steps wherever possible.
You'll also notice I changed checking for  `master` branch to checking if the branch is the `default` - anticipating that we may move to `main` at some point, so the workflow would still be valid.  

Where i failed is that with this PR there are still 2 separate workflows for building a debian and alpine image, but the steps are nearly identical (i ran into some issues with differences in tagging the images so i moved on since the current logic was working well enough). 

So, not the ideal i set out with but with more time i'm sure it can be solved and we can remove some lines from the repo (already thinking about how it can be done, but better to get this PR merged and addressed later was the thought process).

.github/workflows/github-release.yml
wileyj: It definitely could - what would you suggest here, something like `prerelase: false` if a tag is supplied when the workflow is triggered?
for this step, it's really just a copy/paste from the current workflow in a new file: https://github.com/stacks-network/stacks-blockchain/blob/6c5349b0422c5c88db02d578720467c16b102ab2/.github/workflows/ci.yml#L257-L269

The action was updated since the current one in the existing workflow has been deprecated, and the new one has slightly different options. otherwise i kept this logic as unchanged as possible from the current workflow. 

https://github.com/actions/create-release
https://github.com/softprops/action-gh-release/blob/master/action.yml

.github/workflows/ci.yml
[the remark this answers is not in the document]
CharlieC3: @wileyj Just an FYI if considering composite actions in the future, the actions can actually be stored in this repo like the re-usable workflows are. They'd just be in a different directory. [See here for an example in the stacks-web-wallet repo.](https://github.com/hirosystems/stacks-wallet-web/blob/dev/.github/actions/provision/action.yml)

.github/workflows/image-build-alpine-binary.yml
wileyj: I'll have to double-check here, but i recall having a problem doing it that way. Possible that I made a mistake in my testing, but i'll see what happens in a small repo. 
if it works, happy to update.

.github/workflows/image-build-debian-binary.yml
wileyj: same as for the `image-build-alpine-binary.yml` comment, will test and come back

.github/workflows/image-build-debian-source.yml
wileyj: same as for the `image-build-alpine-binary.yml` comment, will test and come back

.github/workflows/image-build-debian-source.yml
wileyj: Debatable if it should be added at all to be honest. 
I left it in the others since they are simpler and i was following the output to see what would happen. 

I think it's better to remove that arg entirely, it's only really useful in my opinion for when you're having some build issues and trying to track down *where* it happened. 

I have no complaints if you think it's useful and want it included as a default though - if that's the case i'll add it here.

.github/workflows/bitcoin-tests.yml
wileyj: Got it - yeah, that was a decision I made that i wasn't positive *should* be done. i'll revert back to how the tests are triggered today.

.github/actions/dockerhub/Dockerfile.debian-source
wileyj: Technically, neither are needed but it also doesn't hurt so i'll add it

.github/workflows/ci.yml
wileyj: I really like that workflow....i'm going to have to take some time and see what i can lift from that. 
depending on timing though, it may have to be in the next iteration (alongside some composite actions perhaps?)

build-scripts/Dockerfile.macos-x64
```diff
@@ -3,22 +3,25 @@ FROM rust:bullseye as build
 ARG STACKS_NODE_VERSION="No Version Info"
 ARG GIT_BRANCH='No Branch Info'
 ARG GIT_COMMIT='No Commit Info'
-
+ARG BUILD_DIR=/build
+ARG OSXCROSS="https://github.com/hirosystems/docker-osxcross-rust/releases/download/MacOSX12.0.sdk/osxcross-d904031_MacOSX12.0.sdk.tar.zst"
+ARG TARGET=x86_64-apple-darwin
```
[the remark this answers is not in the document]
wileyj: +1 
the same issue exists for the arm64-musl build, where there's an external reliance on a muslcross image to build for the correct arch. 
I've assigned the issue to me, will take a look for the next iteration unless you think it's important to address now

.github/actions/dockerhub/Dockerfile.alpine-binary
wileyj: I hadn't considered that, since the arch is hardcoded in the workflow as a list.
I'll take a look and see what I can do about it

.github/actions/dockerhub/Dockerfile.debian-source
wileyj: i'll have to RTFM here, i don't know if that will always be the case. in testing, i haven't seen this error but i'll look into it more. At a minimum, it can be reverted to using bind-mount disk access which is how things are currently being built.

.github/workflows/clippy.yml
wileyj: Can do - would like to hear what @igorsyl thinks about this as well, and if we should be adding in some exclusions to what we want to be warned about.

.github/workflows/ci.yml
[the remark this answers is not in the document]
wileyj: Yup - I think long term though, it would make more sense to move them out of this repo since getting a workflow change added here is harder than i think it should be. 
with some external composite actions, it would be easier to make quicker iterations and get them active. we can discuss later though, just some initial thoughts.

.github/actions/dockerhub/Dockerfile.debian-source
CharlieC3: The standard GH runners have [14GB of disk space](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) so it should be fine. If the space required is expected to grow over time, we might have to switch to a hosted runner some time down the road.

build-scripts/Dockerfile.macos-x64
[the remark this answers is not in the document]
kantai: Yeah, I think it's fine to wait on another iteration for this.

.github/workflows/clippy.yml
igorsyl: We can enable the least number of categories to produce a few suggestions and later add more categories as we fix issues. Is there a way to auto-generate issues from a single clippy run?

```
Category | Description | Default level
-- | -- | --
clippy::all | all lints that are on by default (correctness, suspicious, style, complexity, perf) | warn/deny
clippy::correctness | code that is outright wrong or useless | deny
clippy::suspicious | code that is most likely wrong or useless | warn
clippy::style | code that should be written in a more idiomatic way | warn
clippy::complexity | code that does something simple but in a complex way | warn
clippy::perf | code that can be written to run faster | warn
clippy::pedantic | lints which are rather strict or have occasional false positives | allow
clippy::nursery | new lints that are still under development | allow
clippy::cargo | lints for the cargo manifest | allow
```

https://github.com/rust-lang/rust-clippy

.github/workflows/image-build-alpine-binary.yml
wileyj: https://github.com/wileyj/stacks-blockchain/commit/d8ce51e8d9a4990728bc43acf5e48c80a8079103

.github/workflows/image-build-debian-source.yml
wileyj: https://github.com/wileyj/stacks-blockchain/commit/d8ce51e8d9a4990728bc43acf5e48c80a8079103

.github/workflows/image-build-alpine-binary.yml
wileyj: https://github.com/wileyj/stacks-blockchain/commit/d8ce51e8d9a4990728bc43acf5e48c80a8079103

.github/workflows/image-build-debian-source.yml
wileyj: https://github.com/wileyj/stacks-blockchain/commit/d8ce51e8d9a4990728bc43acf5e48c80a8079103
Removed the line from all docker builds.

.github/actions/dockerhub/Dockerfile.debian-source
wileyj: https://github.com/wileyj/stacks-blockchain/commit/d8ce51e8d9a4990728bc43acf5e48c80a8079103

.github/workflows/bitcoin-tests.yml
wileyj: https://github.com/wileyj/stacks-blockchain/commit/d8ce51e8d9a4990728bc43acf5e48c80a8079103

.github/actions/dockerhub/Dockerfile.alpine-binary
wileyj: added a `*) exit` statement to the `case`
https://github.com/wileyj/stacks-blockchain/commit/d8ce51e8d9a4990728bc43acf5e48c80a8079103

.github/actions/dockerhub/Dockerfile.debian-source
wileyj: leaving this as-is for now. we can adjust in the future if we hit disk limits in the future

.github/workflows/clippy.yml
wileyj: this is now disabled from running until we can address/test how chatty this will be

.github/workflows/ci.yml
wileyj: resolving this for now -  I like the idea for the next iteration so i'm bookmarking that clarinet action

build-scripts/Dockerfile.macos-x64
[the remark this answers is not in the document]
wileyj: https://github.com/stacks-network/stacks-blockchain/issues/3208

wileyj:
> * How to tag a release (I think this is already kind of mentioned in the release checklist, but it could have it's own subsubsection in the README), and what happens automatically when the release is tagged (docker images pushed? binaries included in the git release? changelog?)
> * How to trigger a "feature" build: if I have a PR, and I want to publish a docker image with a specific tag (not a version number) that doesn't create a new release, how can I do that? I think before these changes, that was done "automatically" on every `feat/...` PR, but with these changes, I think it requires a manual trigger, right?

sorry, i missed this comment when i was addressing other changes. I'll check and reply later, optionally with a commit to address the questions

.github/workflows/github-release.yml
wileyj: @CharlieC3 ping! did you have an idea that would make sense to change here, or perhaps an idea for an issue for a future iteration?

CharlieC3: I think `prerelease` should be set to `false` by default, unless this was an intended outcome and part of the release process involves manually changing a prerelease to a release afterwards. This part should be considered for this PR I think.

Ultimately if prereleases are going to be used, I think it would be handy to add a new boolean input to this workflow called `prerelease`, and the value of the boolean would get placed here. This part doesn't have to go in this PR though.

wileyj: personally, i agree here - but it would also be a change in the current process that i wasn't planning to add in this PR: https://github.com/stacks-network/stacks-blockchain/blob/develop/.github/workflows/ci.yml#L283

@kantai @jcnelson any opinion here? the end result to changing this bool would be that when a tagged release is built, a release with binaries that is not set to "pre-release" will also be published. 

I'm in agreement with @CharlieC3 that we should change this from `prerelease: true` -> `prerelease: false` for when a github release is created (only happens on tagged workflows).

CharlieC3: > The same workflow is used for tagging release candidates (which are pre-releases) and creating feature builds (also pre-releases).

Got it. In that case defaulting on `prerelease: true` makes much more sense to me as well. Should a new input be added in the future, I would think it should continue to create a prerelease by default unless acted upon.

wileyj: I'll add to the issue for the next iteration of work here - there may be a way to get this working without having an extra input or having to manually change that pre-release label.

wileyj:
> These changes look good to me -- I'll defer to Charlie and Jude for final approvals.
> 
> My only request is that you update the README.md (or add a RELEASING.md) with two instructions:
> 
> 1. How to tag a release (I think this is already kind of mentioned in the release checklist, but it could have it's own subsubsection in the README), and what happens automatically when the release is tagged (docker images pushed? binaries included in the git release? changelog?)
> 2. How to trigger a "feature" build: if I have a PR, and I want to publish a docker image with a specific tag (not a version number) that doesn't create a new release, how can I do that? I think before these changes, that was done "automatically" on every `feat/...` PR, but with these changes, I think it requires a manual trigger, right?

1. Releases are only tagged if a tag is manually provided when the action is triggered -> this PR doesn't deviate from the current process. Updating the README.md might be better suited in https://github.com/stacks-network/stacks-blockchain/issues/3196 and https://github.com/stacks-network/stacks-blockchain/pull/3247
2. Pushing a new feature branch -> nothing is triggered automatically. PR's are required (same process as current), or the ci workflow can be triggered manually on a specific branch to build assets.


## PR a branch to develop:
- rust format is run (this happens on every run)
- docker image built from source on a debian distribution is built and pushed with the branch name and PR number as tags
  - `blockstack/stacks-blockchain:feat-112-fix-something`
  - `blockstack/stacks-blockchain:pr-112`

## Merging a PR to develop:
- rust format is run (this happens on every run)
- docker image built from source on a debian distribution is built and pushed with the branch name and PR number as tags
  - `blockstack/stacks-blockchain:develop`
  - `blockstack/stacks-blockchain:pr-113`

## Merging a PR develop to master:
- rust format is run (this happens on every run)
- docker image built from source on a debian distribution is built and pushed with the branch name as a tag
  - `blockstack/stacks-blockchain:master`

## Manually triggering workflow without tag (any branch):
- rust format is run (this happens on every run)
- no binaries are built
- no github release
- no integration tests
- docker image built from source on a debian distribution is built and pushed with the branch name as a tag
  - `blockstack/stacks-blockchain:<branch name>`

## Manually triggering workflow with tag (non-default branch):
- rust format is run (this happens on every run)
- no binaries are built
- no github release
- no integration tests
- docker image built from source on a debian distribution is built and pushed with the provided input tag and `latest`
  - `blockstack/stacks-blockchain:<branch name>`

## Manually triggering workflow with tag on default branch (i.e. tag of `2.05.0.3.0`):
- rust format is run (this happens on every run)
- integration tests
- leaked credential test
- binaries built for specified architectures
  - archives and checksum added to github release
- github release (with artifacts/checksum) is created using the manual input tag 
- docker image built from source on a debian distribution is built and pushed with the provided nput tag and `latest`
  - `blockstack/stacks-blockchain:2.05.0.3.0-debian`
  - `blockstack/stacks-blockchain:latest-debian`
  - `blockstack/stacks-blockchain:2.05.0.3.0`
  - `blockstack/stacks-blockchain:latest`

## TL;DR
Creating a release doesn't deviate from the current process. Because the builds for a release will now use a binary vs building from source for each arch, the debian-based source build docker image is **not** created when a new release is tagged. 
a release tag will produce 2 versions of the docker image (along with all binary archives):
1. Alpine image tagged with the `x.x.x.x.x` and `latest`
2. Debian image tagged with `x.x.x.x.x-debian` and `latest-debian`
The debian-based docker image *is* created when feature branches etc are PR'ed (or manually triggered via the CI workflow, choosing the appropriate branch), resulting in a docker image being pushed with a tag like `feat-fix-1234`

`clippy` workflow is being added, but it's currently disabled until we can tame it a bit so it's not overly verbose. 
`audit` workflow is being added, but it's currently disabled. this workflow is not triggered unless some conditions are met (Cargo.toml/Cargo.lock are changed); it is also configured to run on a schedule and create issues for crate advisories. I've disabled it for now so we can evaluate if it's worth keeping (i feel that it is, but in the short-term it may add extra noise to the open issues).

kantai:
Can you add that info either to the `README.md` or a new `RELEASING.md` file? That way, we'll have a place to look if we want to figure it out down the line.

wileyj:
> Can you add that info either to the `README.md` or a new `RELEASING.md` file? That way, we'll have a place to look if we want to figure it out down the line.

Can do - will add a new file and bump the issue dealing with the readme updates

wileyj:
> Can you add that info either to the `README.md` or a new `RELEASING.md` file? That way, we'll have a place to look if we want to figure it out down the line.

https://github.com/wileyj/stacks-blockchain/commit/97ae5638635cae3c8e0d61c866871954a32fa9c7

kantai (approved):

wileyj:
@CharlieC3 i think there's one more change i have to make here that i missed in my tests. 
when a release candidate is created, it should create all binaries/release/images and currently it does not - only realized this as i'm working on the current 2.05.0.3.0 release and watching the CI process

CharlieC3:
> Merging a PR to develop:
rust format is run (this happens on every run)
docker image built from source on a debian distribution is built and pushed with the branch name and PR number as tags
blockstack/stacks-blockchain:develop
blockstack/stacks-blockchain:pr-113

Why would merging a PR to develop (effectively making a commit to `develop` and closing a PR) render a new PR Docker tag? I would think this operation should only create a tag where the new commit/merge was made (e.g. `blockstack/stacks-blockchain:develop`)

> Manually triggering workflow with tag (non-default branch):
rust format is run (this happens on every run)
no binaries are built
no github release
no integration tests
docker image built from source on a debian distribution is built and pushed with the branch name
blockstack/stacks-blockchain:<branch name>

Sometimes we create release-candidate tags off the `develop` branch. With this change, I'm not sure we'll be able to do that anymore. Is that intended?

wileyj:
> > Merging a PR to develop:
> > rust format is run (this happens on every run)
> > docker image built from source on a debian distribution is built and pushed with the branch name and PR number as tags
> > blockstack/stacks-blockchain:develop
> > blockstack/stacks-blockchain:pr-113
> 
> Why would merging a PR to develop (effectively making a commit to `develop` and closing a PR) render a new PR Docker tag? I would think this operation should only create a tag where the new commit/merge was made (e.g. `blockstack/stacks-blockchain:develop`)
> 
> > Manually triggering workflow with tag (non-default branch):
> > rust format is run (this happens on every run)
> > no binaries are built
> > no github release
> > no integration tests
> > docker image built from source on a debian distribution is built and pushed with the branch name
> > blockstack/stacks-blockchain:
> 
> Sometimes we create release-candidate _Github_ tags off the `develop` branch. With this change, I'm not sure we'll be able to do that anymore. Is that intended?

there are a few typos in the comment i made updated here: <https://github.com/wileyj/stacks-blockchain/blob/feat/update-ci/RELEASING.md>

however, your second point is correct and only something i noticed myself the other day (https://github.com/stacks-network/stacks-blockchain/pull/3199#issuecomment-1219839987
) . it needs to be fixed so i'm prepping a change so that still works as exected (currently, trying to cut a release from a non-default branch wouldn't actually create the release, just the debian image). 


Once i get those changes committed, it will also update the RELEASING.md file. 

I'll ping you again when I think it's ready, it was a use case i wasn't thinking about when i sent this PR.

wileyj:
> > Merging a PR to develop:
> > rust format is run (this happens on every run)
> > docker image built from source on a debian distribution is built and pushed with the branch name and PR number as tags
> > blockstack/stacks-blockchain:develop
> > blockstack/stacks-blockchain:pr-113
> 
> Why would merging a PR to develop (effectively making a commit to `develop` and closing a PR) render a new PR Docker tag? I would think this operation should only create a tag where the new commit/merge was made (e.g. `blockstack/stacks-blockchain:develop`)
> 
> > Manually triggering workflow with tag (non-default branch):
> > rust format is run (this happens on every run)
> > no binaries are built
> > no github release
> > no integration tests
> > docker image built from source on a debian distribution is built and pushed with the branch name
> > blockstack/stacks-blockchain:
> 
> Sometimes we create release-candidate _Github_ tags off the `develop` branch. With this change, I'm not sure we'll be able to do that anymore. Is that intended?

https://github.com/stacks-network/stacks-blockchain/pull/3199/commits/663d8a71360a04b5f28c6bdec708a000f4b7a700

I've updated the RELEASING file to reflect the (tested) changes and what happens: https://github.com/wileyj/stacks-blockchain/blob/feat/update-ci/RELEASING.md

I think i have the desired behaviour, if i missed something let me know and i'll re-adjust.

jcnelson:
Note: let's wait until after 2.05.0.3.0 before merging this, since it's potentially disruptive to the release process.

jcnelson:
Waiting until after 2.1 to merge this.

wileyj:
TODO: see if we can retry the codecov upload, or time out if it's not going to work

wileyj:
Now that 2.1 is official, i'll be updating my fork for this. 
One thing i'll be adding to the build changes is the docs PR action failing (posisbly in a separate PR since i think that will be a small fix to get it working again after the docs repo refresh - appears a path the workflow was expecting no longer exists)

zone117x:
@wileyj I'm eager to use official multi-arch docker images in the API repo. 

Should this PR target master since it's a release/CI related change and not a stacks-node code change?

wileyj:
related - #3559

wileyj:
> These changes look good to me -- I'll defer to Charlie and Jude for final approvals.
> 
> My only request is that you update the README.md (or add a RELEASING.md) with two instructions:
> 
> 1. How to tag a release (I think this is already kind of mentioned in the release checklist, but it could have it's own subsubsection in the README), and what happens automatically when the release is tagged (docker images pushed? binaries included in the git release? changelog?)
> 2. How to trigger a "feature" build: if I have a PR, and I want to publish a docker image with a specific tag (not a version number) that doesn't create a new release, how can I do that? I think before these changes, that was done "automatically" on every `feat/...` PR, but with these changes, I think it requires a manual trigger, right?

moved this to `./docs/ci-release.md` in line with recent changes to the repo.

wileyj:
> Since we are seeing these codecov errors very frequently, can we just change this setting to false in the various places it is used? At least until we come up with a better solution for code coverage.
> 
> ```yaml
>       fail_ci_if_error: true
> ```

done, matching what's currently in the `master` branch

CharlieC3 (approved):
I scanned through all the changes made and comments addressed since my last review, it looks great. There's _a lot_ of changes being made here though, so catching any logical or functional errors is not easy and very time-consuming. Assuming it all works as described in the [ci-release doc](https://github.com/stacks-network/stacks-blockchain/pull/3199/files#diff-eea95e05b7582cb953189c86953db14b24aec0a3e1ba95b934aecfd72ba6a852), I think we're good to go! :shipit: 

However, we may need to be prepared to make some quick followup adjustments after merging if we find something important is not quite right with the new processes.

PR #3200, opened 2022-07-15
Move puppet-chain to contrib/tools

### LINKED ISSUE #3195: Remove puppet-chain directory and binary
The puppet-chain binary doesn't seem to be used anymore -- clarinet integrate and devnet do not use it for setting up a local regtest environment, and our integration tests do not use it either.

It could still be useful for developers trying to spin up their own bitcoin regtest environment. However, I feel pretty ambivalent about supporting that use case through the puppet-chain: we don't really cover that use case with our tests, so the current level of support is at best "waiting for an undetected regression", and we're probably better off just eliminating puppet-chain and supporting local regtest through direct bitcoin interaction (that's what our `neon_integrations` tests do).

wileyj:
- Move puppet-chain src to contrib/tools
- Update base Dockerfiles to not build/copy puppet-chain
- ignore contrib/tools/puppet-chain/target

This PR moves the puppet-chain source to `./contrib/tools` directory. 
It also removes the build from the main workspace and Dockerfiles where it's mentioned. 

What it doesn't do: net-test shell scripts still reference `puppet-chain`, but since those tests are not active it can be addressed in another issue to update net-test as a whole.

### Applicable issues
- fixes #3195 

### Checklist
- [ ] Changelog is updated


kantai (approved):
LGTM!

.vscode/launch.json
```diff
@@ -1,266 +1,208 @@
...
+          "kind": "bin"
+        }
+      },
+      "args": [],
+      "cwd": "${workspaceFolder}"
+    },
+    {
+      "type": "lldb",
+      "request": "launch",
+      "name": "executable 'blockstack-cli'",
```
jcnelson, line 35: Isn't this called `stacks-inspect` now?

jcnelson (approved):
LGTM, but let's make sure the `build-publish` and `build-publish-stretch` CI jobs continue to work before merging.

.vscode/launch.json
wileyj: Not sure - this is isn't something I changed in my branch, but I can update that file in this PR if it needs to be changed.

wileyj:
> LGTM, but let's make sure the `build-publish` and `build-publish-stretch` CI jobs continue to work before merging.

An excellent point, there are a few things I've just now noticed that shouldn't have been part of this PR so i'll address them first. 
Namely, the Dockerfiles are slightly updated with some changes in my other PR (#3199 ) that *shouldn't* be in this PR. 
as-is, merging this would break the current CI (unless that other PR is merged first which is not likely). 
i.e. `Dockerfile.stretch` being renamed to `Dockerfile.debian`  would break the CI

will address and ping back shortly

wileyj:
> LGTM, but let's make sure the `build-publish` and `build-publish-stretch` CI jobs continue to work before merging.

I've checked over the files in this PR, and I strongly suspect what's causing the failures here are that my fork of the stacks-blockchain repo doesn't have access to the secrets required for the "login to dockerhub" step. as a result, the docker build is skipped since the login step is failing. 

Merging this with upstream will resolve, since the secrets there would be correct.

That said, I'm confident this PR is ready to go (tested the dockerfile builds, and both are working as expected - along with running the binaries in both the alpine and stretch images). 

Once this is merged with the `develop` branch, i think all the failed tests that would normally pass will succeed.

jcnelson:
Alright, going ahead and merging.  Please keep an eye on CI -- these steps must pass for `develop`.

wileyj:
> Alright, going ahead and merging. Please keep an eye on CI -- these steps must pass for `develop`.

suspicions confirmed: https://github.com/stacks-network/stacks-blockchain/actions/runs/2835581029
all good, thanks!

PR #3204, opened 2022-07-18
Chore: Merge master to develop

kantai:
There have been some unmerged-to-develop changes in master (from hotfixes).

The best practice in the git-flow style flow we use in this repo is to always merge hotfixes back into develop after release, but that didn't happen in this case, so this PR is necessary.


jcnelson (approved):

jcnelson:
So, I think the mining events integration test got broken because we stopped logging `::Skipped` events (so the observer won't have them).

gregorycoppola (approved):

PR #3207, opened 2022-07-20
Feature: Make epochs configurable in mocknet and testnet modes

kantai:
### Description

This PR makes the epoch settings (already configurable at runtime) configurable via the config TOML file. This just requires adding a TOML deserializable field to the `ConfigFile` related structs.

This also patches the mocknet implementation to work with the 2.05 commitment memo.

Example config file entries:

```toml
[[burnchain.epochs]]
epoch_name = "1.0"
start_height = 0

[[burnchain.epochs]]
epoch_name = "2.0"
start_height = 0

[[burnchain.epochs]]
epoch_name = "2.05"
start_height = 1

[[burnchain.epochs]]
epoch_name = "2.1"
start_height = 2
```


testnet/stacks-node/src/config.rs
```diff
@@ -494,11 +495,44 @@ impl Config {
...
+                        epoch.end_height = configured_epoch
+                            .end_height
+                            .try_into()
+                            .expect("Epoch end height must be a positive integer");
+                    }
+
+                    result.epochs = Some(epochs);
                 }
+
+                result
```
jcnelson, line 535: Not sure how feasible this is, but in the spirit of @igorsyl's earlier work on doing config file validation earlier rather than later, would it be possible to verify here that the epochs are well-formed -- i.e. no block ranges overlap, and there are no gaps?

kantai: Hmm, maybe it'll be somewhat easy, I'll see what I can do.

jcnelson: The simplest way I can think of would be to do the following:

* Accept epochs by name, not value, so any typos or non-existent epochs would be quickly caught
* Accept only start heights, not end heights.  The end height for epoch _E_ is the block before the start height for epoch _E+1_.  Then, the user can't possibly create an overlapping range.
* Require that the start height for epoch _E_ is always less than all other epochs _E+k_.  If the user doesn't do this, then the error to report is intuitive: a later epoch cannot start before an earlier epoch.

kantai: Updated in 32d10fc92a31504e59dd2c552fc85166c57df90c -- the new format takes an `epoch_name` and a `start_height`, and it checks that epochs are in the correct order, that epoch names are as expected, and that the configured epochs are a prefix of `1.0, 2.0, 2.05, 2.1` (so you cannot specify `1.0, 2.05, 2.1`, skipping `2.0`)

testnet/stacks-node/src/config.rs
```diff
@@ -494,11 +496,97 @@ impl Config {
...
                         .rbf_fee_increment
                         .unwrap_or(default_burnchain_config.rbf_fee_increment),
-                    epochs: match burnchain.epochs {
-                        Some(epochs) => Some(epochs),
-                        None => default_burnchain_config.epochs,
-                    },
+                    epochs: default_burnchain_config.epochs,
+                };
+
+                if let Some(ref conf_epochs) = burnchain.epochs {
```
igorsyl, line 502: Could this block go inside a new function so that we get to keep the branch `Some(epochs) => Some(epochs)`?

testnet/stacks-node/src/config.rs
```diff
@@ -494,11 +496,97 @@ impl Config {
...
+                    let default_epochs = match result.get_bitcoin_network().1 {
+                        BitcoinNetworkType::Mainnet => {
+                            panic!("Cannot configure epochs in mainnet mode");
```
igorsyl, line 505: Ideally we'd return an Err here.

testnet/stacks-node/src/config.rs
```diff
@@ -494,11 +496,97 @@ impl Config {
...
+                            let epoch_id = if epoch_name == EPOCH_CONFIG_1_0_0 {
+                                StacksEpochId::Epoch10
+                            } else if epoch_name == EPOCH_CONFIG_2_0_0 {
+                                StacksEpochId::Epoch20
+                            } else if epoch_name == EPOCH_CONFIG_2_0_5 {
+                                StacksEpochId::Epoch2_05
+                            } else if epoch_name == EPOCH_CONFIG_2_1_0 {
+                                StacksEpochId::Epoch21
+                            } else {
+                                panic!("Unknown epoch name specified: {}", epoch_name);
```
igorsyl, line 523: Can we choose an Err instead of panic?

igorsyl:
Looks good, however it would be great to return Err's instead of calling panic for usability and testing purposes.

testnet/stacks-node/src/config.rs
kantai: Addressed in https://github.com/stacks-network/stacks-blockchain/pull/3207/commits/0f4ac0655a567b830b6cfe2818378e22f3b51f0e

testnet/stacks-node/src/config.rs
```diff
@@ -569,11 +571,105 @@ impl Config {
...
                         .unwrap_or(default_burnchain_config.rbf_fee_increment),
-                    epochs: match burnchain.epochs {
-                        Some(epochs) => Some(epochs),
-                        None => default_burnchain_config.epochs,
-                    },
+                    epochs: default_burnchain_config.epochs,
+                };
+
+                if let Some(ref conf_epochs) = burnchain.epochs {
+                    let default_epochs = match result.get_bitcoin_network().1 {
```
iamalwaysuncomfortable, line 578: Looking at the the get_bitcoin_network function, the string arg here might be able to be dropped ultimately since it seems to be duplicating what's in the enum. 

Barring that though, maybe let's change this to the following for clarity:

```suggestion
                    let (_, default_epochs) = match result.get_bitcoin_network()
                    match default_epochs {
```

testnet/stacks-node/src/config.rs
```diff
@@ -569,11 +571,105 @@ impl Config {
...
                         .rbf_fee_increment
                         .unwrap_or(default_burnchain_config.rbf_fee_increment),
-                    epochs: match burnchain.epochs {
-                        Some(epochs) => Some(epochs),
-                        None => default_burnchain_config.epochs,
-                    },
+                    epochs: default_burnchain_config.epochs,
+                };
+
+                if let Some(ref conf_epochs) = burnchain.epochs {
```
jcnelson, line 577: This code body is getting pretty long.  Maybe this block should be split out into its own method?

jcnelson (approved):
LGTM. Just a minor refactoring suggestion.

testnet/stacks-node/src/config.rs
kantai: Yeah, I think I agree regarding eliminating the first return element of `get_bitcoin_network()`, but I try to avoid refactoring functions that aren't directly related to the PR. Regarding your second suggestion: I refactored this code so that the enum is now a function argument to a constructor for the new configuration settings, so I think that improves the clarity of what's happening.

testnet/stacks-node/src/config.rs
kantai: Yep, good point, refactored in e852ab8

igorsyl (approved):

PR #3209, opened 2022-07-21
Fix type error in monitoring_prom build

kantai:
This fixes a build error in `develop` currently.


jcnelson (approved):

gregorycoppola (approved):
Was the build broken?

I thought it was green, that's why it could submit.

kantai:
> Was the build broken?

Yes -- it's hard to tell with the CI as flaky as it is in this repo, but the build image tasks were both broken with a type error.

PR #3212, opened 2022-07-22
Feat/drop problematic transactions

jcnelson:
This PR addresses a recurring performance degradation issue in the mempool and miner whereby a well-formed, valid transaction can trigger a `CheckError` at runtime, but the miner always re-attempts to mine it.  Such transactions can take a non-trivial amount of time and effort to evaluate before the `CheckError` is triggered.  But, transactions that trigger `CheckError`-type errors can never be included in blocks, since they are analysis-time (not runtime) errors.  This means that miners can never collect the transaction fee for these transactions, even though evaluating them took non-trivial time and effort.  So, the miner should at the very least not try to mine it again.

Ideally, no transaction should ever encounter a `CheckError` at runtime -- all such errors should be found at analysis time.  But we don't live in this ideal world -- there are a few ways to do this today that have been seen in the wild, and removing them would be a breaking change to the consensus rules.  So, until we can address them properly (e.g. in Stacks 2.1), this PR makes it so that any such transaction discovered by the miner will simply be dropped from the mempool, and never accepted via mempool sync or relay for the next 48 hours.

In the code, I call these transactions "problematic" transactions.  Strictly speaking, these transactions *could* be mined -- they are well-formed, and the code path that triggered the `CheckError` might not always be guaranteed to execute.  But the fact that a `CheckError` was found *at all* is reason enough to never consider it again.  This handling is reasonable because the only remedy for the user is to re-submit a *different* transaction that does *not* manifest the `CheckError` anyway.

The PR touches on the following areas:

* It bumps the schema of the mempool, so that the mempool can remember the txids of problematic transactions in order to never accept them again.  It also introduces logic for garbage-collecting the blacklist.

* It modifies the miner to treat the emergence of a `CheckError` as a distinct `TransactionResult` variant (`Problematic`), and handles these results by dropping and blacklisting them from the mempool.

* It adds test coverage to the mempool's new blacklist code, to the miner with example transactions that manifest `CheckError`s at runtime, and to the networking stack to verify that blacklisted transactions are never re-admitted.

### DIFF
--- a/.github/actions/bitcoin-int-tests/Dockerfile.code-cov
+++ b/.github/actions/bitcoin-int-tests/Dockerfile.code-cov
@@ -2,6 +2,8 @@ FROM rust:bullseye AS test
 
 WORKDIR /build
 
+ENV CARGO_MANIFEST_DIR="$(pwd)"
+
 RUN rustup override set nightly-2022-01-14 && \
     rustup component add llvm-tools-preview && \
     cargo install grcov
--- a/.github/workflows/bitcoin-tests.yml
+++ b/.github/workflows/bitcoin-tests.yml
@@ -59,7 +59,7 @@ jobs:
           - tests::neon_integrations::filter_low_fee_tx_integration_test
           - tests::neon_integrations::filter_long_runtime_tx_integration_test
           - tests::neon_integrations::mining_transactions_is_fair
-          - tests::neon_integrations::microblock_large_tx_integration_test
+          - tests::neon_integrations::microblock_large_tx_integration_test_FLAKY
           - tests::neon_integrations::block_large_tx_integration_test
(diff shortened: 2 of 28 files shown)


Hero-Gamer:
+1

kantai (approved):

jcnelson:
So, the mining events integration test is broken in `develop` right now because we stopped logging skipped transactions as events.  I'll wait for https://github.com/stacks-network/stacks-blockchain/pull/3204 to fix it and merge before merging this, if that works for everyone.

jcnelson:
Make it so that the deny table can only grow to a maximum size, so someone can't just ddos the node's disk.

jcnelson:
@pavitthrap @gregorycoppola @igorsyl can one of you review this please? :pray:

gregorycoppola (approved):

PR #3219, opened 2022-07-25
Adds human-readable time to logger

igorsyl:
Adds the environment variable `BLOCKSTACK_FORMAT_TIME` to use human-readable time:

```
$ cargo stacks-node check-config
INFO [1658776091.841531] [testnet/stacks-node/src/main.rs:82] [main] stacks-node 0.1.0 (human-time:8a83e72f31+, debug build, macos [aarch64])
ERRO [1658776091.841989] [testnet/stacks-node/src/main.rs:58] [main] Process abort due to thread panic: panicked at 'called `Result::unwrap()` on an `Err` value: MissingOption(Keys(["--config", ""]))', testnet/stacks-node/src/main.rs:113:71
ERRO [1658776092.441379] [testnet/stacks-node/src/main.rs:60] [main] Panic backtrace:    0: backtrace::backtrace::libunwind::trace
```

```
$ STACKS_LOG_FORMAT_TIME="%Y-%m-%d %H:%M:%S" cargo stacks-node check-config
INFO [2022-07-25 14:09:58] [testnet/stacks-node/src/main.rs:82] [main] stacks-node 0.1.0 (human-time:8a83e72f31+, debug build, macos [aarch64])
ERRO [2022-07-25 14:09:58] [testnet/stacks-node/src/main.rs:58] [main] Process abort due to thread panic: panicked at 'called `Result::unwrap()` on an `Err` value: MissingOption(Keys(["--config", ""]))', testnet/stacks-node/src/main.rs:113:71
ERRO [2022-07-25 14:09:59] [testnet/stacks-node/src/main.rs:60] [main] Panic backtrace:    0: backtrace::backtrace::libunwind::trace
```

### DIFF
--- a/stacks-common/src/util/log.rs
+++ b/stacks-common/src/util/log.rs
@@ -41,15 +41,21 @@ fn print_msg_header(mut rd: &mut dyn RecordDecorator, record: &Record) -> io::Re
     write!(rd, " ")?;
 
     rd.start_timestamp()?;
-    let elapsed = SystemTime::now()
-        .duration_since(SystemTime::UNIX_EPOCH)
-        .unwrap_or(Duration::from_secs(0));
-    write!(
-        rd,
-        "[{:5}.{:06}]",
-        elapsed.as_secs(),
-        elapsed.subsec_nanos() / 1000
-    )?;
+    let system_time = SystemTime::now();
+    if env::var("BLOCKSTACK_FORMAT_TIME") != Ok("1".into()) {
+        let elapsed = system_time
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap_or(Duration::from_secs(0));
+        write!(
+            rd,
+            "[{:5}.{:06}]",
+            elapsed.as_secs(),
+            elapsed.subsec_nanos() / 1000
+        )?;
+    } else {
+        let datetime: DateTime<Local> = system_time.into();
+        write!(rd, "[{}]", datetime.format("%Y-%m-%d %H:%M:%S"))?;
+    }
     write!(rd, " ")?;
     write!(rd, "[{}:{}]", record.file(), record.line())?;
     write!(rd, " ")?;
--- a/stacks-common/src/util/log.rs
+++ b/stacks-common/src/util/log.rs
@@ -26,8 +26,7 @@ use std::time::{Duration, SystemTime};
 
 lazy_static! {
     pub static ref LOGGER: Logger = make_logger();
-    pub static ref STACKS_LOG_FORMAT_TIME: Result<String, env::VarError> =
-        env::var("STACKS_LOG_FORMAT_TIME");
+    pub static ref STACKS_LOG_FORMAT_TIME: Option<String> = env::var("STACKS_LOG_FORMAT_TIME").ok();
 }
 struct TermFormat<D: Decorator> {
     decorator: D,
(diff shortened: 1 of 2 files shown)


diwakergupta:
Let's avoid introducing any further usage of `BLOCKSTACK` -- that's a term from another era 😅 Perhaps `STACKS_LOG_FORMAT_TIME`? Alternatively, perhaps we can generalize this to let people pass in their own format strings?

stacks-common/src/util/log.rs
```diff
@@ -41,15 +41,21 @@ fn print_msg_header(mut rd: &mut dyn RecordDecorator, record: &Record) -> io::Re
...
-        .duration_since(SystemTime::UNIX_EPOCH)
-        .unwrap_or(Duration::from_secs(0));
-    write!(
-        rd,
-        "[{:5}.{:06}]",
-        elapsed.as_secs(),
-        elapsed.subsec_nanos() / 1000
-    )?;
+    let system_time = SystemTime::now();
+    if env::var("BLOCKSTACK_FORMAT_TIME") != Ok("1".into()) {
```
jcnelson, line 45: Is there a way to avoid calling `env::var(...)` every time we log something?  This is a surprisingly time-consuming operation.  Can this be cached inside a `lazy_static! { ... }` instead?

stacks-common/src/util/log.rs
```diff
@@ -41,15 +43,24 @@ fn print_msg_header(mut rd: &mut dyn RecordDecorator, record: &Record) -> io::Re
...
-        .duration_since(SystemTime::UNIX_EPOCH)
-        .unwrap_or(Duration::from_secs(0));
-    write!(
-        rd,
-        "[{:5}.{:06}]",
-        elapsed.as_secs(),
-        elapsed.subsec_nanos() / 1000
-    )?;
+    let system_time = SystemTime::now();
+    match &*STACKS_LOG_FORMAT_TIME {
```
gregorycoppola, line 47: I think semantically it would make more sense if this was `None` vs `Some` rather than using the error.

From the perspective of the program, it's not really an error, so it's semantically confusing.

stacks-common/src/util/log.rs
```diff
@@ -26,6 +26,8 @@ use std::time::{Duration, SystemTime};
 
 lazy_static! {
     pub static ref LOGGER: Logger = make_logger();
+    pub static ref STACKS_LOG_FORMAT_TIME: Result<String, env::VarError> =
```
gregorycoppola, line 29: can you use Option here instead of result? as also discussed below.

clarification: is this a lazy assignment?

gregorycoppola (commented):
Thanks a lot for this change! One small thing..

stacks-common/src/util/log.rs
igorsyl: Makes sense. env::var returns a Result so I'll use ok() to convert into Option.

stacks-common/src/util/log.rs
igorsyl: Yes, it's a lazy assignment.

gregorycoppola (approved):
thanks for a *timely* change!

kantai (approved):
Looks great to me -- can you add an item to the changelog for this?

igorsyl:
@jcnelson Ready to merge!

PR #3222, opened 2022-07-27
Enables hot reloading of mining configuration file

igorsyl:
Enables hot reloading of mining configuration file.

Prior to every mining burnchain transaction, the config file will be reloaded if its modified time is newer and the following settings will be reloaded:

```
burnchain.burn_fee_cap
burnchain.satoshis_per_byte
burnchain.rbf_fee_increment
burnchain.max_rbf
node.microblock_frequency
node.wait_time_for_microblocks
```

The following config file items, if missing, will be read from environment variables:
```
node.seed => STACKS_NODE_SEED
burnchain.username => STACKS_BURNCHAIN_USERNAME
burnchain.password => STACKS_BURNCHAIN_PASSWORD 
```

Addresses the following issues:
- https://github.com/stacks-network/stacks-blockchain/issues/2629
- https://github.com/stacks-network/stacks-blockchain/issues/3181

### DIFF
--- a/testnet/stacks-node/src/config.rs
+++ b/testnet/stacks-node/src/config.rs
@@ -361,13 +361,13 @@ pub struct Config {
 }
 
 #[derive(Clone, Debug)]
-pub struct DynamicConfig {
+pub struct DynConfig {
     config: std::sync::Arc<std::sync::Mutex<std::cell::RefCell<Config>>>
 }
 
-impl DynamicConfig {
-    pub fn new(config: &Config) -> Self {
-        DynamicConfig {
+impl DynConfig {
+    pub fn new(config: &Config) -> DynConfig {
+        DynConfig {
             config: std::sync::Arc::new(std::sync::Mutex::new(std::cell::RefCell::new(config.clone())))
         }
     }
--- a/testnet/stacks-node/src/config.rs
+++ b/testnet/stacks-node/src/config.rs
@@ -426,12 +426,21 @@ impl Config {
         let (mut node, bootstrap_node, deny_nodes) = match config_file.node {
             Some(node) => {
                 let rpc_bind = node.rpc_bind.unwrap_or(default_node_config.rpc_bind);
+                let node_seed_bytes = match node.seed {
+                    Some(seed) => Some(seed),
+                    None => match std::env::var("STACKS_NODE_SEED") {
+                        Ok(seed) => {
+                            info!("Using STACKS_NODE_SEED for node.seed");
+                            Some(seed)
+                        },
+                        Err(_) => None
+                    },
+                };
                 let node_config = NodeConfig {
                     name: node.name.unwrap_or(default_node_config.name),
-                    seed: match node.seed {
-                        Some(seed) => hex_bytes(&seed)
-                            .map_err(|e| format!("node.seed should be a hex encoded string"))?,
-                        None => default_node_config.seed,
+                    seed: match node_seed_bytes {
+                        Some(seed) => hex_bytes(&seed).map_err(|e| format!("node.seed should be a hex encoded string"))?,
+                        None => &default_node_config.seed,
                     },
                     working_dir: node.working_dir.unwrap_or(default_node_config.working_dir),
                     rpc_bind: rpc_bind.clone(),
@@ -565,8 +574,14 @@ impl Config {
                     rpc_ssl: burnchain
                         .rpc_ssl
                         .unwrap_or(default_burnchain_config.rpc_ssl),
-                    username: burnchain.username,
-                    password: burnchain.password,
+                    username: burnchain.username.or_else(|| {
+                        info!("Using STACKS_BURNCHAIN_USERNAME");
+                        std::env::var("STACKS_BURNCHAIN_USERNAME").ok()
+                    }),
+                    password: burnchain.password.or_else(|| {
+                        info!("Using STACKS_BURNCHAIN_PASSWORD");
+                        std::env::var("STACKS_BURNCHAIN_PASSWORD").ok()
+                    }),
                     timeout: burnchain
                         .timeout
                         .unwrap_or(default_burnchain_config.timeout),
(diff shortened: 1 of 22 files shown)


jcnelson:
Why does this need to be an RPC endpoint?  Can't the node just reload its config file on e.g. `SIGUSR1` like most UNIX daemons?

igorsyl:
> 

I envision a miner bot calling this endpoint instead of a human. Giving filesystem access may be more challenging that providing an endpoint.

jcnelson:
The miner bot could just as easily modify the config file.  Then, it's the miner bot's problem to solve for providing a network-accessible c&c endpoint.

igorsyl:
The miner bot will run in a separate k8s pod and have no direct access to node filesystem.

jcnelson:
> The miner bot will run in a separate k8s pod and have no direct access to node filesystem.

That's not the blockchain node's problem.

igorsyl:
> > The miner bot will run in a separate k8s pod and have no direct access to node filesystem.
> 
> That's the miner bot's problem, not the blockchain node's.

There has been previous discussion about adding endpoints for a miner app.
What has changed since then?

https://github.com/stacks-network/stacks-blockchain/issues/1969

jcnelson:
In that conversation, the miner bot would be responsible for _everything_: assembling the Stacks block, making and sending the block-commit, and broadcasting the block.  The node would no longer be involved in mining at all.

jcnelson:
To build a separate miner, we basically need to make two major changes to the Stacks node, in this order:

1. Move all of the Stacks block-construction code and Bitcoin transaction-construction code into a separate program.
2. Add endpoints to the Stacks node so that this separate program can query the chainstate, sortition DB, and mempool for all the information it needs to construct a Stacks block.

Doing (1) is pretty straight-forward.  The code that does all of this is already somewhat self-contained; we just need to put it into a separate program that can run the block and transaction assembly logic every time a new Stacks block, Stacks microblock, or Bitcoin block arrives.  This can be achieved if the node has read-only access to the Stacks node's chainstate -- trivially, it could just poll the relevant databases to figure out when it should try to mine again, and load the relevant records to build the block and/or microblock.

Once (1) is working, the task of (2) is to make it so that instead of pulling data out of the chainstate databases directly, the miner program can just ask the Stacks node for that data via its RPC endpoint.  And, because this data is all public data, there's no safety risk to making this RPC endpoint public.

Adding an RPC endpoint to control the miner doesn't help either of these tasks.  Moreover, doing so would be a significant undertaking, because in order to control the miner _safely_, the RPC endpoint would need to be authenticated and confidential (meaning, you'd have to _also_ modify the HTTP server to support HTTP authentication, _and_ you would need to modify the socket layer to support SSL middleware).  These two sub-tasks are much, much, much larger undertakings than adding the endpoint, and would ultimately be removed once the above two tasks are complete.  Therefore, I'd much rather we didn't spend time trying to make the miner controllable via RPC.

igorsyl:
Thanks for your comments! Your proposed plan makes sense to me. I'll keep this PR to unblock mining operations for myself until a more robust solution is available and will not intend to merge it.

jcnelson:
N.B. from the meeting:  this is mainly to help @igorsyl automate mining operations for the purpose of profiling.  It will not be merged.

igorsyl:
Implemented config reload using SIGUSR1. Any concerns about race conditions when reading config values? @jcnelson

testnet/stacks-node/src/config.rs
```diff
@@ -360,6 +367,27 @@ pub struct Config {
     pub estimation: FeeEstimationConfig,
 }
 
+#[derive(Clone, Debug)]
+pub struct DynConfig {
```
kantai, line 371: Can you add a rustdoc to this struct?

kantai:
I'm sorry I missed the initial discussion here around using SIGUSR1. This is okay, but it seems like a strategy that precludes cross-platform support and adds some unnecessary UX complexity versus reloading on file changes.

It seems like it would be relatively straight-forward to use [notify](https://crates.io/crates/notify) to monitor for changes.

igorsyl:
> I'm sorry I missed the initial discussion here around using SIGUSR1. This is okay, but it seems like a strategy that precludes cross-platform support and adds some unnecessary UX complexity versus reloading on file changes.
> 
> It seems like it would be relatively straight-forward to use [notify](https://crates.io/crates/notify) to monitor for changes.
> 
> Either way, the reloading functionality needs to be documented somewhere in the miner section of the `README.md`

Agreed need to add documentation to readme.
Notify on file changes would be a better UX! Thoughts @jcnelson?

jcnelson:
> I'm sorry I missed the initial discussion here around using SIGUSR1. This is okay, but it seems like a strategy that precludes cross-platform support and adds some unnecessary UX complexity versus reloading on file changes.

Heh, I guess my age is showing :)  Sending `SIGUSR1` to a UNIX process at least is the standard way of triggering a configuration reload in a daemon without a stop/start.  But portability is more important, and `SIGUSR1` doesn't exist on Windows.

> It seems like it would be relatively straight-forward to use notify to monitor for changes.

I took a look at the `notify` crate and it's ... kinda scary.  First, the Darwin and Windows OS-specific modules use a _lot_ of `unsafe` blocks.  Second, even though the inotify and kqueue back-ends are free of `unsafe` code, they're pretty big and only lightly tested.

If all we're doing is just checking the config file to change the burn amounts, maybe it's sufficient to just have the relayer thread check the `mtime` on the config file each time it handles a `RunTenure` relayer directive, and if it has changed, it reloads the config.  Then, we don't need any new dependencies!

> Notify on file changes would be a better UX!

So, historically, the reason you'd send a daemon `SIGUSR1` is so you could edit and save the config file _without_ having to worry about the changes taking effect immediately.  Perhaps we don't really need that here, in which case, I'd be fine with dropping adherence with this convention.

jcnelson:
What's wrong with just reloading the config file on each pass of the relayer and p2p threads if the last-modified timestamp changes?

PR #3226, opened 2022-07-27
Fix in `next`: Type mismatches on new functions

kantai:
I discovered two mismatches between the typechecker and the runtime in new Clarity functions.

This PR fixes those mismatches, and adds a typing test to the documentation self-tests. This ensures that the typechecker's inferred types for all the examples admit the resulting value.


jcnelson (approved):

obycode (approved):
LGTM

gregorycoppola (approved):

Issue #3107, opened 2022-04-19
[clarity] All check errors must be handled as runtime errors


jcnelson:
Right now, `(at-block)` can emit `CheckError`s which cause the transaction to be invalid, as well as the block that includes it.  In Stacks 2.1, we should make it so that _any_ error emitted by `(at-block)` is a `RuntimeError`, so at least the transaction will be mineable and fees will be collected.

jcnelson:
N.B. this applies to all check errors.  We should not merely re-cast all check errors as runtime errors, but instead treat check errors the way we do for runtime errors -- i.e. they don't invalidate the block; they only debit the sender the tx fee (but the tx receipt should contain the original check error).

saralab:
A good one for @obycode to help with.

obycode:
👍 I just finished reviewing the PR.

jcnelson:
I already did this one @saralab.

Thank you for your review @obycode!

PR #3234, opened 2022-08-01
Feat/runtime analysis errors

jcnelson:
This PR makes all `Error::Unchecked(..)` Clarity errors encountered at runtime into runtime errors that do not invalidate the block.  This feature is gated and only takes effect in Stacks 2.1.

### DIFF
--- a/src/chainstate/stacks/db/transactions.rs
+++ b/src/chainstate/stacks/db/transactions.rs
@@ -2551,7 +2551,48 @@ pub mod test {
 
         let signed_tx = signer.get_tx().unwrap();
 
-        for (dbi, burn_db) in ALL_BURN_DBS.iter().enumerate() {
+        // invalid contract-calls
+        let contract_calls = vec![
+            (
+                addr.clone(),
+                "hello-world",
+                "set-bar-not-a-method",
+                vec![Value::Int(1), Value::Int(1)],
+            ), // call into non-existant method
+            (
+                addr.clone(),
+                "hello-world-not-a-contract",
+                "set-bar",
+                vec![Value::Int(1), Value::Int(1)],
+            ), // call into non-existant contract
+            (
+                addr_2.clone(),
+                "hello-world",
+                "set-bar",
+                vec![Value::Int(1), Value::Int(1)],
+            ), // address does not have a contract
+            (addr.clone(), "hello-world", "set-bar", vec![Value::Int(1)]), // wrong number of args (too few)
+            (
+                addr.clone(),
+                "hello-world",
+                "set-bar",
+                vec![Value::Int(1), Value::Int(1), Value::Int(1)],
+            ), // wrong number of args (too many)
+            (
+                addr.clone(),
+                "hello-world",
+                "set-bar",
+                vec![Value::buff_from([0xff, 4].to_vec()).unwrap(), Value::Int(1)],
+            ), // wrong arg type
+            (
+                addr.clone(),
+                "hello-world",
+                "set-bar",
+                vec![Value::UInt(1), Value::Int(1)],
+            ), // wrong arg type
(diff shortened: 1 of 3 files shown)


src/chainstate/stacks/db/transactions.rs
```diff
@@ -942,6 +993,35 @@ impl StacksChainState {
...
+                                let receipt =
+                                    StacksTransactionReceipt::from_runtime_failure_contract_call(
+                                        tx.clone(),
+                                        total_cost,
+                                        check_error,
+                                    );
+                                return Ok(receipt);
+                            } else {
+                                // prior to 2.1, this is not permitted in a block.
+                                error!("Unexpected analysis error invalidating transaction: if included, this will invalidate a block";
```
kantai, line 1015: I think this should be `warn!` level -- `error` level logs should mean that something is _wrong_ and that the operator of the node should be trying to take action to remedy it.

src/chainstate/stacks/db/transactions.rs
```diff
@@ -8417,4 +8530,286 @@ pub mod test {
...
+        signer.sign_origin(&privk).unwrap();
+
+        let signed_runtime_checkerror_trait_tx = signer.get_tx().unwrap();
+
+        let mut tx_runtime_checkerror_impl = StacksTransaction::new(
+            TransactionVersion::Testnet,
+            auth.clone(),
+            TransactionPayload::new_smart_contract(
+                &"foo-impl".to_string(),
+                &runtime_checkerror_trait.to_string(),
```
kantai, line 8618: Is this supposed to be `runtime_checkerror_impl`?

src/chainstate/stacks/db/transactions.rs
jcnelson: Fixed

src/chainstate/stacks/db/transactions.rs
jcnelson: Fixed

src/chainstate/stacks/db/transactions.rs
```diff
@@ -8417,4 +8530,286 @@ pub mod test {
...
+            false,
+        )
+        .unwrap();
+        assert_eq!(fee, 0);
+
+        assert!(tx_receipt.vm_error.is_some());
+        let err_str = tx_receipt.vm_error.unwrap();
+        assert!(err_str
+            .find("TypeValueError(OptionalType(TraitReferenceType(TraitIdentifier ")
+            .is_some());
```
kantai, line 8879: Can you add checks to this test for the following:

1. In 2.1, the fee (which should be positive for the test to actually assert a change) is assessed against the payer's account, and the nonce is bumped for the failed transaction.
2. In 2.05, the fee isn't assessed (again, it needs to be positive) against the payer's account, and the nonce isn't bumped for the failed transaction.
3. In 2.1, the transaction is committed, but there's no state change. To do this, I think you'd need to invoke `(flip)` at some point in the `test` method before triggering the error. Then just assert that `flip` hasn't changed after the tx.

Then, I think this needs an additional (very similar) battery of tests for the case where the "handled" error comes from a contract publish -- I think if you add a `(test .foo-impl)` invocation to the bottom of the contract publish, it'll create such a situation.

gregorycoppola (commented):
Overall, looks good to me.

1) Waiting to see the response to kantai's question about tests.
2) Seems like some context is missing from the description. Is there an issue or some higher-level plan you can link to explain why this is happening?

jcnelson:
> @gregorycoppola Seems like some context is missing from the description. Is there an issue or some higher-level plan you can link to explain why this is happening?

See #3107, which this PR addresses.

obycode (approved):
LGTM

PR #3235, opened 2022-08-01
Add mock-miner config file

igorsyl:
Add a stacks-node config file for mock mining for performance tests.


testnet/stacks-node/conf/mainnet-mockminer-conf.toml
```diff
@@ -0,0 +1,21 @@
+[node]
+# working_dir = "/dir/to/save/chainstate"
+rpc_bind = "0.0.0.0:20443"
+p2p_bind = "0.0.0.0:20444"
+mock_mining = true
```
gregorycoppola, line 5: I think what you need to add is:

```
miner = true
mock_mining = true
```

gregorycoppola (commented):
Is this running?

igorsyl:
> Is this running?

Yes

jcnelson (approved):

gregorycoppola (approved):

Issue #3155, opened 2022-05-27
[clarity] expose PoX reward recipient addresses via `get-burn-block-info?`


jcnelson:
The `get-burn-block-info?` built-in should expose the recipient addresses for a given burn block height (e.g. as a `(list 2 principal)`).

jcnelson:
Merged!

PR #3245, opened 2022-08-09
Feat/get pox addrs

jcnelson:
This PR implements #3155 by adding `get-burn-block-info? pox-addrs <burn-height>` to Clarity.  It returns an `optional` tuple with two items: 

* `addrs`: The list of up to two PoX addresses (type `(list 2 { hashbytes: (buff 20), version: (buff 1) })`)
* `payout`: The number of burnchain tokens (i.e. satoshis) paid to each PoX address across all block-commits (type `uint`).

It returns `(some ..)` when the sortition for `burn-height` is in the transaction's Stacks block's fork, and `none` if not.

~~Burn addresses are not reported.  This method will return an empty `addrs` list during the reward phase if the BTC are to be burnt, and will return an empty `addrs` list during the prepare phase.~~ EDIT: Burn addresses are reported -- up to two during the reward phase for unclaimed PoX reward slots, and one during the prepare phase.

The `payout` value is _per TX output_.  During the reward phase, there are two outputs whose values are equal to _half of_ the node's `burn_fee_cap`.  This is true even if there are no more PoX reward addresses to pay.  During the prepare phase, there is one output, and its value is equal to the node's `burn_fee_cap`.  So for example if the network is spending 10,000 sats per block-commit, then `payout` will be 5,000 during the reward phase and 10,000 during the prepare phase.

In addition to implementing this feature, this PR replaces the use of `StacksAddress` to represent Bitcoin transaction outputs with a similar `PoxAddress` structure.  This structure captures all of the fields in a `pox-addr` argument to `stack-stx` or `delegate-stack-stx`, which is not recoverable from just the Bitcoin chain state.  It is these values that are passed back to the caller from `get-burn-block-info? pox-addrs`.

The addition of `PoxAddress` is not only necessary to represent the full PoX address data, but also lays the groundwork for implementing segwit (and taproot) addresses.  These addresses _cannot_ be represented in a `StacksAddress`, but because `PoxAddress` is an enum, we can later add type variants that represent any `scriptPubKey` we want.

### DIFF
--- a/clarity/src/vm/database/clarity_db.rs
+++ b/clarity/src/vm/database/clarity_db.rs
@@ -67,195 +67,6 @@ pub const STORE_CONTRACT_SRC_INTERFACE: bool = true;
 
 pub type StacksEpoch = GenericStacksEpoch<ExecutionCost>;
 
-/// A PoX address as seen by the .pox and .pox-2 contracts.
-/// Used by the sortition DB and chains coordinator to extract addresses from the PoX contract to
-/// build the reward set and to validate block-commits.
-#[derive(Debug, PartialEq, PartialOrd, Ord, Clone, Hash, Eq, Serialize, Deserialize)]
-pub enum PoxAddress {
-    /// represents { version: (buff u1), hashbytes: (buff 20) }.
-    /// The address hash mode is optional because if we decode a legacy bitcoin address, we won't
-    /// be able to determine the hash mode since we can't distinguish segwit-p2sh from p2sh
-    Standard(StacksAddress, Option<AddressHashMode>),
-}
-
-impl std::fmt::Display for PoxAddress {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.to_db_string())
-    }
-}
-
-impl PoxAddress {
-    pub fn hashmode(&self) -> Option<AddressHashMode> {
-        match *self {
-            PoxAddress::Standard(_, hm) => hm.clone(),
-        }
-    }
-
-    #[cfg(any(test, feature = "testing"))]
-    pub fn version(&self) -> u8 {
-        self.hashmode()
-            .expect("FATAL: tried to load the hashmode of a PoxAddress which has none known")
-            as u8
-    }
-
-    pub fn bytes(&self) -> Vec<u8> {
-        match *self {
-            PoxAddress::Standard(addr, _) => addr.bytes.0.to_vec(),
-        }
-    }
-
-    #[cfg(any(test, feature = "testing"))]
-    pub fn hash160(&self) -> Hash160 {
-        match *self {
-            PoxAddress::Standard(addr, _) => addr.bytes.clone(),
-        }
-    }
-
-    /// Try to convert a Clarity value representation of the PoX address into a PoxAddress.
-    /// `value` must be `{ version: (buff 1), hashbytes: (buff 20) }`
-    pub fn try_from_pox_tuple(mainnet: bool, value: &Value) -> Option<PoxAddress> {
-        let tuple_data = match value {
-            Value::Tuple(data) => data.clone(),
-            _ => {
-                return None;
-            }
-        };
-
-        let hashmode_value = tuple_data.get("version").ok()?.to_owned();
-
-        let hashmode_u8 = match hashmode_value {
-            Value::Sequence(SequenceData::Buffer(data)) => {
-                if data.data.len() == 1 {
-                    data.data[0]
-                } else {
-                    return None;
-                }
-            }
-            _ => {
-                return None;
-            }
-        };
-
-        let hashbytes_value = tuple_data.get("hashbytes").ok()?.to_owned();
-
-        let hashbytes_vec = match hashbytes_value {
-            Value::Sequence(SequenceData::Buffer(data)) => {
-                if data.data.len() == 20 {
-                    data.data
-                } else {
-                    return None;
-                }
-            }
-            _ => {
-                return None;
-            }
-        };
-
-        let hashmode: AddressHashMode = hashmode_u8.try_into().ok()?;
-
-        let mut hashbytes_20 = [0u8; 20];
-        hashbytes_20.copy_from_slice(&hashbytes_vec[0..20]);
-        let bytes = Hash160(hashbytes_20);
-
-        let version = if mainnet {
-            hashmode.to_version_mainnet()
-        } else {
-            hashmode.to_version_testnet()
-        };
-
-        Some(PoxAddress::Standard(
-            StacksAddress { version, bytes },
-            Some(hashmode),
-        ))
-    }
-
-    /// Serialize this structure to a string that we can store in the sortition DB
-    pub fn to_db_string(&self) -> String {
-        serde_json::to_string(self).expect("FATAL: failed to serialize JSON value")
-    }
-
-    /// Decode a db string back into a PoxAddress
-    pub fn from_db_string(db_string: &str) -> Option<PoxAddress> {
-        serde_json::from_str(db_string).ok()?
-    }
-
-    /// Is this a burn address?
-    pub fn is_burn(&self) -> bool {
-        match *self {
-            PoxAddress::Standard(ref addr, _) => addr.is_burn(),
-        }
-    }
-
-    /// What is the burnchain representation of this address?
-    /// Used for comparing addresses from block-commits, where certain information (e.g. the hash
-    /// mode) can't be used since it's not stored there.  The resulting string encodes all of the
-    /// information that is present on the burnchain, and it does so in a _stable_ way.
-    pub fn to_burnchain_repr(&self) -> String {
-        match *self {
-            PoxAddress::Standard(ref addr, _) => {
-                format!("{:02x}-{}", &addr.version, &addr.bytes)
-            }
-        }
-    }
-
-    /// Make a standard burn address
-    pub fn standard_burn_address(mainnet: bool) -> PoxAddress {
-        PoxAddress::Standard(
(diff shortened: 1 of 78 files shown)


clarity/src/vm/database/clarity_db.rs
```diff
@@ -53,15 +53,209 @@ use stacks_common::consts::{
...
+/// A PoX address as seen by the .pox and .pox-2 contracts.
+/// Used by the sortition DB and chains coordinator to extract addresses from the PoX contract to
+/// build the reward set and to validate block-commits.
+#[derive(Debug, PartialEq, PartialOrd, Ord, Clone, Hash, Eq, Serialize, Deserialize)]
+pub enum PoxAddress {
+    /// represents { version: (buff u1), hashbytes: (buff 20) }.
+    /// The address hash mode is optional because if we decode a legacy bitcoin address, we won't
+    /// be able to determine the hash mode since we can't distinguish segwit-p2sh from p2sh
+    Standard(StacksAddress, Option<AddressHashMode>),
+}
```
kantai, line 79: I think this struct should be in the `chainstate` module, rather than in the Clarity module. I know there's some leakage of PoX into the Clarity module (mostly in the function signatures in the database), but most of these methods seem more important to the chainstate than it is to Clarity -- all Clarity really needs is the tuples:

```rust
/// Returns the paid out PoX addresses (as Clarity tuples of type `{ hashbytes: (buff 1), version: (buff 20) }` )
pub fn get_pox_payout_addrs_for_burnchain_height(
        &mut self,
        burnchain_block_height: u32,
    ) -> Option<(Vec<TupleData>, u128)> {
```

Whereas the chainstate module actually uses most of the methods of this struct.

jcnelson: Sure, I can do that

jcnelson: Done: https://github.com/stacks-network/stacks-blockchain/pull/3245/commits/49dea6ed734af9c43d257e0424e1dee135ccdf86

src/chainstate/burn/db/sortdb.rs
```diff
@@ -756,6 +759,19 @@ impl db_keys {
...
+    }
+
+    pub fn pox_reward_set_payouts_value(addrs: Vec<PoxAddress>, payout_per_addr: u128) -> String {
+        serde_json::to_string(&(addrs, payout_per_addr)).unwrap()
+    }
+
+    pub fn pox_reward_set_payouts_decode(addr_str: &str) -> (Vec<PoxAddress>, u128) {
+        let addrs_and_payout: (Vec<PoxAddress>, u128) = serde_json::from_str(addr_str).unwrap();
+        addrs_and_payout
+    }
```
kantai, line 773: Does this value need to be stored in the MARF, or can it just be indexed by the `SortitionID` in a sqlite table?

kantai: As in, lookups are always of the form -- `what was the payout set at <sortition-id>` (which doesn't need to traverse the sortition history), and they're never `what was the payout set at <sortition-id> or if there wasn't one, what was the last payout prior?` (which could only be answered through parent traversal).

kantai (commented):
This looks great to me, @jcnelson, but I think it may not be necessary to store the PoX payout information in the sortition MARF, but it could instead just be an entry in a sortition-id indexed table. It's up to you if you think the difference would be substantial enough to merit changes.

src/chainstate/burn/db/sortdb.rs
jcnelson: I suppose it doesn't need to be written to the MARF.  I can just write it to a DB table -- it would make the I/O faster at least.

jcnelson: Fixed

jcnelson:
I think I might want to update the API for this to return the burn addresses as well (so then it's obvious how much total BTC was spent, given the per-output spend)

jcnelson:
Need to make the return type easier to understand:

* Return burn addresses explicitly
* Return burn addresses in the prepare phase, so that the number of entries in the `list` reflects what block-commits are actually doing (so it's easier to interpret the `payout` field)

jcnelson:
Okay, the above is done.  cc @kantai @gregorycoppola @pavitthrap for review/approval.

src/chainstate/burn/operations/leader_key_register.rs
```diff
@@ -174,7 +174,14 @@ impl LeaderKeyRegisterOp {
...
 
-        let address = outputs[0].address.clone();
+        let address = outputs[0]
+            .address
+            .clone()
+            .try_into_stacks_address()
+            .ok_or_else(|| {
+                test_debug!("Invalid address: not representable as a StacksAddress");
+                op_error::InvalidInput
+            })?;
```
kantai, line 184: Is this a consensus change?

src/chainstate/stacks/address.rs
```diff
@@ -49,46 +49,246 @@ use crate::types::chainstate::STACKS_ADDRESS_ENCODED_SIZE;
...
+        self.hashmode()
+            .expect("FATAL: tried to load the hashmode of a PoxAddress which has none known")
+            as u8
+    }
+
+    pub fn bytes(&self) -> Vec<u8> {
+        match *self {
+            PoxAddress::Standard(addr, _) => addr.bytes.0.to_vec(),
+        }
     }
```
kantai, line 92: I think it's better to return a `&[u8]` here. The caller can then decide if they want to copy to a vector or not. This gets used for sorting, so requiring a bunch of copies would be inefficient there.

src/chainstate/stacks/address.rs
```diff
@@ -49,46 +49,246 @@ use crate::types::chainstate::STACKS_ADDRESS_ENCODED_SIZE;
...
 }
 
-impl StacksAddressExtensions for StacksAddress {
-    /// is this a boot code address, if the supplied address is mainnet or testnet,
-    ///  it checks against the appropriate the boot code addr
-    fn is_boot_code_addr(&self) -> bool {
-        self == &boot_code_addr(self.is_mainnet())
+/// A PoX address as seen by the .pox and .pox-2 contracts.
+/// Used by the sortition DB and chains coordinator to extract addresses from the PoX contract to
+/// build the reward set and to validate block-commits.
```
kantai, line 59: Maybe add "This comprises a larger set of possible addresses than `StacksAddress`"?

src/chainstate/stacks/address.rs
kantai, line 92: Can you add rustdocs to the public functions of this struct? I think as it abstracts multiple kinds of addresses, functions like "bytes" are less self-explanatory.

kantai (approved):
This looks good to me, I think there's just some superficial comments remaining: slight tweaks to an interface and some additional rustdocs.

src/chainstate/burn/operations/leader_key_register.rs
jcnelson: Nope. First, `.address` isn't even used in `LeaderKeyRegister`. I'm even going so far as to remove it in my segwit PR.  Second, `.try_into_stacks_address()` is infallible right now, since the only `PoxAddress` variant (`Standard(addr)`) contains a `StacksAddress`.

src/chainstate/stacks/address.rs
jcnelson: If you're referring for `StacksChainState::make_reward_set()`, the problem with doing this the way you suggest is that the reference `&addrs.bytes.0` doesn't live long enough in the closure passed to `.sort_by_key()`.  Maybe we should keep this as-is and use `.sort_by_cached_key()`?

src/chainstate/burn/operations/leader_key_register.rs
kantai: In that case, this should panic -- as it is currently written, it suggests that an error would invalidate the `LeaderKeyRegister` (similarly for all the operations that this `try_into_stacks_address()` method is invoked on).

src/chainstate/stacks/address.rs
kantai: I guess that's fine to leave as is then.

pavitthrap (approved):
lgtm

src/chainstate/stacks/address.rs
jcnelson: Sure, will do

src/chainstate/stacks/address.rs
jcnelson: Sure, will do

src/chainstate/burn/operations/leader_key_register.rs
jcnelson: The reason it's written this way is because `.try_into_stacks_address()` will soon be fallible, i.e. on the arrival of the native segwit support.  The reason it will be fallible is because there will not be a way to convert a segwit address into a `principal` type, so parsing `StackStxOp` and `TransferStxOp` must fail here if the intended recipient can't be represented as a `principal`.

src/chainstate/stacks/address.rs
jcnelson: Done

src/chainstate/stacks/address.rs
jcnelson: Done

src/chainstate/stacks/address.rs
```diff
@@ -49,46 +49,246 @@ use crate::types::chainstate::STACKS_ADDRESS_ENCODED_SIZE;
...
+    }
+
+    /// Convert this PoxAddress into a base58check string
+    pub fn to_b58(self) -> String {
+        match self {
+            PoxAddress::Standard(addr, _) => addr.to_b58(),
+        }
+    }
+
+    /*
```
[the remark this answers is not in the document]
jcnelson: Done

PR #3247, opened 2022-08-10
Simplify repo README

### LINKED ISSUE #3196: Simplify repo README.md
The current readme is very long winded with a lot of detail and explanation that could probably be moved elsewhere or linked to official docs rather than being listed in the repo readme. 

For an example of what I think it should look like, compare to bitcoin's readme: https://github.com/bitcoin/bitcoin/blob/master/README.md

or even solana's: https://github.com/solana-labs/solana/blob/master/README.md

I think we can do better here, providing some basic info on getting starting quickly and linking to more detailed docs.

criadoperez:
### Description

Simplified the README.md file.
* Updated links
* "blockstack" rebranding
* Moved some documentation to the official [Stacks docs site](docs.stacks.co)
* Moved some contribution instructions to the contributions file.
* Added references to tutorials
* ~~Fixes failing CircleCI icon~~ --> Removes failing CircleCI icon

Closes #3196


README.md
```diff
@@ -1,481 +1,48 @@
...
+Stacks 2.0 is a layer-1 blockchain that connects to Bitcoin for security and enables decentralized apps and predictable smart contracts. Stacks 2.0 implements [Proof of Transfer (PoX)](https://community.stacks.org/pox) mining that anchors to Bitcoin security. Leader election happens at the Bitcoin blockchain and Stacks (STX) miners write new blocks on the separate Stacks blockchain. With PoX there is no need to modify Bitcoin to enable smart contracts and apps around it. See [this page](https://github.com/stacks-network/stacks) for more details and resources.
 
-[![CircleCI](https://circleci.com/gh/blockstack/stacks-blockchain/tree/master.svg?style=svg)](https://circleci.com/gh/blockstack/stacks-blockchain/tree/master)
+[![CircleCI](https://circleci.com/gh/stacks-network/stacks-blockchain/tree/master.svg?style=svg)](https://circleci.com/gh/stacks-network/stacks-blockchain/tree/master)
```
wileyj, line 7: This can be removed - circleci is no longer used

criadoperez: ok, removed then.

criadoperez:
[Docs PR 1506](https://github.com/stacks-network/docs/pull/1506) has merged

wileyj:
`CODE_OF_CONDUCT.md` still mentions "blockstack" email address

wileyj:
In the readme and other markdown, links can be removed in favor of "badges" i'd argue. something along the lines here:
https://gist.github.com/tterb/982ae14a9307b80117dbf49f624ce0e8

where the license and version at least are listed. 
links to things like discord/youtube etc can also be added as badges vs text links.

criadoperez:
> In the readme and other markdown, links can be removed in favor of "badges" i'd argue. something along the lines here: https://gist.github.com/tterb/982ae14a9307b80117dbf49f624ce0e8
> 
> where the license and version at least are listed. links to things like discord/youtube etc can also be added as badges vs text links.

The scope of this PR is only for the README.md file. Other markdown files may require more changes than adding badges that will lead to further discussions, so I would rather make them on other PRs later.
I added the badges for the license, however in my opinion, if we use a double license we need to clarify more. Original text stated that 
```
This code is released under [the GPL v3 license](https://www.gnu.org/licenses/quick-guide-gplv3.en.html), and the docs are released under [the Creative Commons license](https://creativecommons.org/).
```
We don't specify what type of Creative Commons license and what exactly are the `docs`. If we only use GPLv3 license we can leave it with the latest change I made. But, if we want to keep using both licenses we need to first state what type of Creative Commons license we are using and specify what the docs are (for example, all files in the docs folder are [CC Attribution 4.0 International](http://creativecommons.org/licenses/by/4.0/).

Regarding the badges of the community section, yes I can replace them.

wileyj:
Got it. I would argue that this is the correct PR to make *all* the
changes.
There's no rush on this, and there's no impact to taking longer to merge
this -> it would actually save everyone's time to do a single PR vs many
small PR's.

criadoperez:
> I would argue that this is the correct PR to make *all* the changes

ok. If that's the case then I'll also review the other markdown files.

criadoperez:
> `CODE_OF_CONDUCT.md` still mentions "blockstack" email address

Corrected. Thanks for pointing this out.

PR #3248, opened 2022-08-12
Profiling tools

igorsyl:
- Added documentation for profiling in PROFILING.md.
- Added `stacks-event` binary to subscribe to and record `stacks-node` observer events.
- Added `tip-mine` subcommand to `stacks-inspect` to mine a block on top of a old block.
- Added feature flag `disable-costs` to disable block costs during mining for profiling purposes.
- Added sqlite profile logging using feature flag `profile-sqlite`.


kantai (approved):
This looks good to me, thanks for adding the `PROFILING.md` file! Do you think it makes sense to put that file in `docs` or leave it in the top level dir?

igorsyl:
> This looks good to me, thanks for adding the `PROFILING.md` file! Do you think it makes sense to put that file in `docs` or leave it in the top level dir?

docs would be great to keep root dir clean!

clarity/src/vm/costs/mod.rs
```diff
@@ -865,6 +865,9 @@ fn compute_cost(
 
 fn add_cost(s: &mut TrackerData, cost: ExecutionCost) -> std::result::Result<(), CostErrors> {
     s.total.add(&cost)?;
+    if cfg!(feature = "disable-costs") {
```
gregorycoppola, line 868: This is kind of a hack... maybe add a todo or an issue to remove it, longer term?

gregorycoppola (commented):
Thanks for all the changes, Igor.. just one comment.

clarity/src/vm/costs/mod.rs
igorsyl: Agreed, a comment is needed!

src/chainstate/stacks/db/blocks.rs
```diff
@@ -6047,6 +6047,19 @@ impl StacksChainState {
         query_row(&self.db(), sql, args).map_err(Error::DBError)
     }
 
+    pub fn get_stacks_chain_parent(
```
jcnelson, line 6050: Please add a docstring

src/chainstate/stacks/db/blocks.rs
```diff
@@ -6047,6 +6047,19 @@ impl StacksChainState {
...
+        &self,
+        sortdb: &SortitionDB,
+        staging_block: &StagingBlock,
+    ) -> Result<Option<StagingBlock>, Error> {
```
jcnelson, line 6054: Can you call this `get_stacks_block_parent()` instead of `get_stacks_chain_parent()`?  You're returning a block, not a chain ;)

src/stacks_events.rs
```diff
@@ -0,0 +1,86 @@
...
+use std::net::TcpStream;
+use std::time::SystemTime;
+use std::{env, io};
+
+fn main() {
+    serve_for_events();
+}
+
+fn serve_for_events() {
+    let addr = "127.0.0.1:3700";
```
jcnelson, line 21: Should this port be configurable, or at least documented?

src/stacks_events.rs
```diff
@@ -0,0 +1,86 @@
...
+            let caps = RE_POST.captures(&buf).unwrap();
+            path = Some(caps.get(1).unwrap().as_str().to_string());
+        } else if content_length.is_none() {
+            let caps = RE_CONTENT_LENGTH.captures(&buf);
+            if let Some(caps) = caps {
+                content_length = Some(caps.get(1).unwrap().as_str().parse::<u64>().unwrap());
+            }
+        } else if buf.len() == 2 {
+            buf.clear();
+            unsafe {
```
jcnelson, line 56: You don't need to do `unsafe` here.  You can easily read from the socket and buffer up bytes until the content length is reached.

src/stacks_events.rs
```diff
@@ -0,0 +1,86 @@
...
+        reader.read_line(&mut buf).unwrap();
+        if path.is_none() {
+            let caps = RE_POST.captures(&buf).unwrap();
+            path = Some(caps.get(1).unwrap().as_str().to_string());
+        } else if content_length.is_none() {
+            let caps = RE_CONTENT_LENGTH.captures(&buf);
+            if let Some(caps) = caps {
+                content_length = Some(caps.get(1).unwrap().as_str().parse::<u64>().unwrap());
+            }
+        } else if buf.len() == 2 {
```
jcnelson, line 54: Might want to explicitly check for `\r\n` since that's what it seems you're intending to do?

jcnelson (approved):
LGTM. Thanks @igorsyl! Please address all comments before merging.

src/stacks_events.rs
igorsyl: I'll add a command argument parser.

src/stacks_events.rs
igorsyl: unsafe needed due to `buf.as_mut_vec`. Looking for alternatives without unsafe.

igorsyl: Found `read_to_string` which does not need unsafe.

gregorycoppola:
@igorsyl

Overall it LGTM but I am thinking the `stacks-event` should:
1) maybe be in Javascript instead of rust
2) should write to a DB instead of a text file

I have a bunch of harness code I've written for the L2 profiling, and it is in Javascript, and that makes it a lot easier. Plus, you can use the `stacks.js` libraries if you are working in Javascript.

So, I would consider dropping that part, or else even reimplementing it in JS later.

gregorycoppola (approved):
LGTM

But, as I said in a comment, I would consider moving the `stacks-event` stuff to Javscript and writing the output to a DB.

jcnelson:
Please fix the failing unit tests and full-genesis tests before merging.

Cargo.toml
```diff
@@ -53,6 +53,7 @@ name = "c32_bench"
 harness = false
 
 [dependencies]
+pico-args = "0.3.1"
```
jcnelson, line 56: Can we please not add new dependencies?  Thanks

jcnelson (changes requested):
Please remove `pico-args`.

Cargo.toml
jcnelson: That doesn't mean we should put it into the stacks library.  Honestly, we should be _removing_ dependencies, not adding to them.

jcnelson: Thanks!

jcnelson (approved):

PR #3260, opened 2022-08-22
PoX 2 Unlock

kantai:
This PR implements automatic unlocks for PoX 2.

Stackers who do not qualify for a slot in a reward set when the reward set is selected will be automatically unlocked after the reward set is selected. The auto unlock occurs 1 bitcoin block after the auto-unlock.

This requires a _lot_ of complexity in `pox-2`. This PR adds a test for the simple use case, but probably more testing is needed.

### DIFF
--- a/clarity/src/vm/database/clarity_db.rs
+++ b/clarity/src/vm/database/clarity_db.rs
@@ -1166,6 +1166,23 @@ impl<'a> ClarityDatabase<'a> {
         )
     }
 
+    /// Like fetch_entry_unknown_descriptor, except that it expects
+    ///  to receive a value (i.e., it expects the state to exist)
+    /// This should only ever be invoked outside of Clarity, e.g. by the VM
+    ///  when loading state from a known-contract
+    pub fn expect_fetch_entry(
+        &mut self,
+        contract_identifier: &QualifiedContractIdentifier,
+        map_name: &str,
+        key_value: &Value,
+    ) -> Result<Value> {
+        self.fetch_entry_unknown_descriptor(contract_identifier, map_name, key_value)
+            .map(|v| {
+                v.expect_optional()
+                    .expect("Expected fetch_entry to return a value")
+            })
+    }
+
     pub fn fetch_entry_unknown_descriptor(
         &mut self,
         contract_identifier: &QualifiedContractIdentifier,
@@ -1176,6 +1193,7 @@ impl<'a> ClarityDatabase<'a> {
         self.fetch_entry(contract_identifier, map_name, key_value, &descriptor)
     }
 
+    /// Returns a Clarity optional type wrapping a found or not found result
     pub fn fetch_entry(
         &mut self,
         contract_identifier: &QualifiedContractIdentifier,
@@ -1358,6 +1376,16 @@ impl<'a> ClarityDatabase<'a> {
                 .checked_add(placed_size)
                 .expect("Overflowed Clarity key/value size"),
         })
+    }
+
+    pub fn delete_entry_unknown_descriptor(
+        &mut self,
+        contract_identifier: &QualifiedContractIdentifier,
+        map_name: &str,
+        key_value: &Value,
+    ) -> Result<ValueResult> {
+        let descriptor = self.load_map(contract_identifier, map_name)?;
+        self.delete_entry(contract_identifier, map_name, key_value, &descriptor)
     }
 
     pub fn delete_entry(
--- a/clarity/src/vm/database/clarity_db.rs
+++ b/clarity/src/vm/database/clarity_db.rs
@@ -1205,23 +1205,6 @@ impl<'a> ClarityDatabase<'a> {
         )
     }
 
-    /// Like fetch_entry_unknown_descriptor, except that it expects
-    ///  to receive a value (i.e., it expects the state to exist)
-    /// This should only ever be invoked outside of Clarity, e.g. by the VM
-    ///  when loading state from a known-contract
-    pub fn expect_fetch_entry(
-        &mut self,
-        contract_identifier: &QualifiedContractIdentifier,
-        map_name: &str,
-        key_value: &Value,
-    ) -> Result<Value> {
-        self.fetch_entry_unknown_descriptor(contract_identifier, map_name, key_value)
-            .map(|v| {
-                v.expect_optional()
-                    .expect("Expected fetch_entry to return a value")
-            })
-    }
-
     pub fn fetch_entry_unknown_descriptor(
         &mut self,
         contract_identifier: &QualifiedContractIdentifier,
@@ -1415,16 +1398,6 @@ impl<'a> ClarityDatabase<'a> {
                 .checked_add(placed_size)
                 .expect("Overflowed Clarity key/value size"),
         })
-    }
-
-    pub fn delete_entry_unknown_descriptor(
-        &mut self,
-        contract_identifier: &QualifiedContractIdentifier,
-        map_name: &str,
-        key_value: &Value,
-    ) -> Result<ValueResult> {
-        let descriptor = self.load_map(contract_identifier, map_name)?;
-        self.delete_entry(contract_identifier, map_name, key_value, &descriptor)
     }
 
     pub fn delete_entry(
--- a/clarity/src/vm/database/structures.rs
+++ b/clarity/src/vm/database/structures.rs
@@ -485,15 +485,9 @@ impl<'db, 'conn> STXBalanceSnapshot<'db, 'conn> {
         let new_unlock_height = self.burn_block_height + 1;
         self.balance = match self.balance {
             STXBalance::Unlocked { amount } => STXBalance::Unlocked { amount },
-            STXBalance::LockedPoxOne {
-                amount_unlocked,
-                amount_locked,
-                ..
-            } => STXBalance::LockedPoxOne {
-                amount_unlocked,
-                amount_locked,
-                unlock_height: new_unlock_height,
-            },
+            STXBalance::LockedPoxOne { .. } => {
+                unreachable!("Attempted to accelerate the unlock of a lockup created by PoX-1")
+            }
             STXBalance::LockedPoxTwo {
                 amount_unlocked,
                 amount_locked,
--- a/clarity/src/vm/database/structures.rs
+++ b/clarity/src/vm/database/structures.rs
@@ -474,6 +474,39 @@ impl<'db, 'conn> STXBalanceSnapshot<'db, 'conn> {
         };
     }
 
+    /// Lock `amount_to_lock` tokens on this account until `unlock_burn_height`.
+    /// After calling, this method will set the balance to a "LockedPoxTwo" balance,
+    ///  because this method is only invoked as a result of PoX2 interactions
+    pub fn accelerate_unlock(&mut self) {
+        let unlocked = self.unlock_available_tokens_if_any();
+        if unlocked > 0 {
+            debug!("Consolidated after account-token-lock");
+        }
+
+        let new_unlock_height = self.burn_block_height + 1;
+        self.balance = match self.balance {
+            STXBalance::Unlocked { amount } => STXBalance::Unlocked { amount },
+            STXBalance::LockedPoxOne {
+                amount_unlocked,
+                amount_locked,
+                ..
+            } => STXBalance::LockedPoxOne {
+                amount_unlocked,
+                amount_locked,
+                unlock_height: new_unlock_height,
+            },
+            STXBalance::LockedPoxTwo {
+                amount_unlocked,
+                amount_locked,
+                ..
+            } => STXBalance::LockedPoxTwo {
+                amount_unlocked,
+                amount_locked,
+                unlock_height: new_unlock_height,
+            },
+        };
+    }
+
     /// Unlock any tokens that are unlockable at the current
     ///  burn block height, and return the amount newly unlocked
     fn unlock_available_tokens_if_any(&mut self) -> u128 {
--- a/clarity/src/vm/database/structures.rs
+++ b/clarity/src/vm/database/structures.rs
@@ -474,9 +474,8 @@ impl<'db, 'conn> STXBalanceSnapshot<'db, 'conn> {
         };
     }
 
-    /// Lock `amount_to_lock` tokens on this account until `unlock_burn_height`.
-    /// After calling, this method will set the balance to a "LockedPoxTwo" balance,
-    ///  because this method is only invoked as a result of PoX2 interactions
+    /// If this snapshot is locked, then alter the lock height to be
+    /// the next burn block (i.e., `self.burn_block_height + 1`)
     pub fn accelerate_unlock(&mut self) {
         let unlocked = self.unlock_available_tokens_if_any();
         if unlocked > 0 {
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -46,6 +46,7 @@ use crate::burnchains::{
     BurnchainStateTransition, BurnchainStateTransitionOps, BurnchainTransaction,
     Error as burnchain_error, PoxConstants,
 };
+use crate::chainstate::burn::db::sortdb::SortitionHandle;
 use crate::chainstate::burn::db::sortdb::{SortitionDB, SortitionHandleConn, SortitionHandleTx};
 use crate::chainstate::burn::distribution::BurnSamplePoint;
 use crate::chainstate::burn::operations::{
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -73,11 +73,11 @@ use crate::util_lib::db::Error as db_error;
 use stacks_common::address::public_keys_to_address_hash;
 use stacks_common::address::AddressHashMode;
 use stacks_common::deps_common::bitcoin::util::hash::Sha256dHash as BitcoinSha256dHash;
-use stacks_common::util::{get_epoch_time_ms, sleep_ms};
 use stacks_common::util::get_epoch_time_secs;
 use stacks_common::util::hash::to_hex;
 use stacks_common::util::log;
 use stacks_common::util::vrf::VRFPublicKey;
+use stacks_common::util::{get_epoch_time_ms, sleep_ms};
 
 use crate::burnchains::bitcoin::indexer::BitcoinIndexer;
 use crate::chainstate::stacks::boot::POX_2_MAINNET_CODE;
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -502,7 +502,9 @@ impl Burnchain {
         burn_ht: u64,
         reward_cycle_length: u64,
     ) -> bool {
-        let effective_height = burn_ht - first_block_ht;
+        let effective_height = burn_ht
+            .checked_sub(first_block_ht)
+            .expect("FATAL: attempted to check reward cycle start before first block height");
         // first block of the new reward cycle
         (effective_height % reward_cycle_length) <= 1
     }
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -491,6 +491,18 @@ impl Burnchain {
         )
     }
 
+    /// Is this block the block in a reward cycle right before the reward phase
+    ///  starts? This is the mod 0 block.
+    pub fn is_pre_reward_cycle_start(
+        first_block_ht: u64,
+        burn_ht: u64,
+        reward_cycle_length: u64,
+    ) -> bool {
+        let effective_height = burn_ht - first_block_ht;
+        // first block of the new reward cycle
+        (effective_height % reward_cycle_length) <= 1
+    }
+
     pub fn static_is_in_prepare_phase(
         first_block_height: u64,
         reward_cycle_length: u64,
--- a/src/burnchains/burnchain.rs
+++ b/src/burnchains/burnchain.rs
@@ -492,9 +492,12 @@ impl Burnchain {
         )
     }
 
-    /// Is this block the block in a reward cycle right before the reward phase
-    ///  starts? This is the mod 0 block.
-    pub fn is_pre_reward_cycle_start(
+    /// Is this block either the first block in a reward cycle or
+    ///  right before the reward phase starts? This is the mod 0 or mod 1
+    ///  block. Reward cycle start events (like auto-unlocks) process *after*
+    ///  the first reward block, so this function is used to determine when
+    ///  that has passed.
+    pub fn is_before_reward_cycle(
         first_block_ht: u64,
         burn_ht: u64,
         reward_cycle_length: u64,
--- a/src/chainstate/burn/db/sortdb.rs
+++ b/src/chainstate/burn/db/sortdb.rs
@@ -842,6 +842,81 @@ impl db_keys {
     }
 }
 
+/// Trait for structs that provide a chaintip-indexed handle into the
+///  SortitionDB (i.e., a MARF view from a particular SortitionId)
+pub trait SortitionHandle {
+    /// Returns a connection to the SQLite db. If this handle is wrapping
+    ///  a transaction, this should point to the open transaction.
+    fn sqlite(&self) -> &Connection;
+
+    /// Returns the snapshot of the burnchain block at burnchain height `block_height`.
+    /// Returns None if there is no block at this height.
+    fn get_block_snapshot_by_height(
+        &mut self,
+        block_height: u64,
+    ) -> Result<Option<BlockSnapshot>, db_error>;
+
+    /// is the given block a descendant of `potential_ancestor`?
+    ///  * block_at_burn_height: the burn height of the sortition that chose the stacks block to check
+    ///  * potential_ancestor: the stacks block hash of the potential ancestor
+    fn descended_from(
+        &mut self,
+        block_at_burn_height: u64,
+        potential_ancestor: &BlockHeaderHash,
+    ) -> Result<bool, db_error> {
+        let earliest_block_height = self.sqlite().query_row(
+            "SELECT block_height FROM snapshots WHERE winning_stacks_block_hash = ? ORDER BY block_height ASC LIMIT 1",
+            &[potential_ancestor],
+            |row| Ok(u64::from_row(row).expect("Expected u64 in database")))?;
+
+        let mut sn = self
+            .get_block_snapshot_by_height(block_at_burn_height)?
+            .ok_or_else(|| {
+                test_debug!("No snapshot at height {}", block_at_burn_height);
+                db_error::NotFoundError
+            })?;
+
+        while sn.block_height >= earliest_block_height {
+            if !sn.sortition {
+                return Ok(false);
+            }
+            if &sn.winning_stacks_block_hash == potential_ancestor {
+                return Ok(true);
+            }
+
+            // step back to the parent
+            match SortitionDB::get_block_commit_parent_sortition_id(
+                self.sqlite(),
+                &sn.winning_block_txid,
+                &sn.sortition_id,
+            )? {
+                Some(parent_sortition_id) => {
+                    // we have the block_commit parent memoization data
+                    test_debug!(
+                        "Parent sortition of {} memoized as {}",
+                        &sn.winning_block_txid,
+                        &parent_sortition_id
+                    );
+                    sn = SortitionDB::get_block_snapshot(self.sqlite(), &parent_sortition_id)?
+                        .ok_or_else(|| db_error::NotFoundError)?;
+                }
+                None => {
+                    // we do not have the block_commit parent memoization data
+                    // step back to the parent
+                    test_debug!("No parent sortition memo for {}", &sn.winning_block_txid);
+                    let block_commit =
+                        get_block_commit_by_txid(&self.sqlite(), &sn.winning_block_txid)?
+                            .expect("CORRUPTION: winning block commit for snapshot not found");
+                    sn = self
+                        .get_block_snapshot_by_height(block_commit.parent_block_ptr as u64)?
+                        .ok_or_else(|| db_error::NotFoundError)?;
+                }
+            }
+        }
+        return Ok(false);
+    }
+}
+
 impl<'a> SortitionHandleTx<'a> {
     /// begin a MARF transaction with this connection
     ///  this is used by _writing_ contexts
@@ -1152,6 +1227,34 @@ impl<'a> SortitionHandleTx<'a> {
     }
 }
 
+impl SortitionHandle for SortitionHandleTx<'_> {
+    fn get_block_snapshot_by_height(
+        &mut self,
+        block_height: u64,
+    ) -> Result<Option<BlockSnapshot>, db_error> {
+        assert!(block_height < BLOCK_HEIGHT_MAX);
+        let chain_tip = self.context.chain_tip.clone();
+        SortitionDB::get_ancestor_snapshot_tx(self, block_height, &chain_tip)
+    }
+
+    fn sqlite(&self) -> &Connection {
+        self.tx()
+    }
+}
+
+impl SortitionHandle for SortitionHandleConn<'_> {
+    fn get_block_snapshot_by_height(
+        &mut self,
+        block_height: u64,
+    ) -> Result<Option<BlockSnapshot>, db_error> {
+        SortitionHandleConn::get_block_snapshot_by_height(self, block_height)
+    }
+
+    fn sqlite(&self) -> &Connection {
+        self.conn()
+    }
+}
+
 impl<'a> SortitionHandleTx<'a> {
     pub fn set_stacks_block_accepted(
         &mut self,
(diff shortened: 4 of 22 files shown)


clarity/src/vm/database/structures.rs
```diff
@@ -474,6 +474,39 @@ impl<'db, 'conn> STXBalanceSnapshot<'db, 'conn> {
         };
     }
 
+    /// Lock `amount_to_lock` tokens on this account until `unlock_burn_height`.
+    /// After calling, this method will set the balance to a "LockedPoxTwo" balance,
+    ///  because this method is only invoked as a result of PoX2 interactions
```
jcnelson, line 479: This doesn't appear to be happening in that `match self.balance` block?  Looks like you're merely altering an existing `LockedPoxOne` instead of converting it to a `LockedPoxTwo`.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -1585,6 +1603,66 @@ impl<'a> SortitionHandleConn<'a> {
         })
     }
 
+    /// is the given block a descendant of `potential_ancestor`?
+    ///  * block_at_burn_height: the burn height of the sortition that chose the stacks block to check
+    ///  * potential_ancestor: the stacks block hash of the potential ancestor
+    pub fn descended_from(
```
jcnelson, line 1609: Is there any way to DRY this up?  Looks like it was copy/pasted from `SortitionHandleTx<'_>`.

kantai: I can try. I didn't want to DRY it too much initially, because any approach to doing so may end obfuscating the function quite a bit.

src/chainstate/stacks/boot/mod.rs
```diff
@@ -124,7 +131,354 @@ pub fn make_contract_id(addr: &StacksAddress, name: &str) -> QualifiedContractId
...
+                Ok(())
+            })?;
+        }
+
+        Ok(())
+    }
+
+    /// After Stacks 2.1, invoke to process any Stacks chainstate operations required
+    ///  at the start of a reward cycle (i.e., qualifying unlocks)
+    pub fn process_pox_cycle_start_2_1(last_anchor_block: Option<BlockHeaderHash>) {
```
jcnelson, line 468: Is this method supposed to do anything?

src/chainstate/stacks/boot/mod.rs
```diff
@@ -249,39 +603,85 @@ impl StacksChainState {
...
+                "stacked_amt" => stacked_amt,
+                "pox_threshold" => threshold,
             );
             for _i in 0..slots_taken {
                 test_debug!("Add to PoX reward set: {:?}", &address);
                 reward_set.push(address.clone());
             }
+            // if stacker did not qualify for a slot *and* they have a stacker
+            //   pointer set by the PoX contract, then add them to auto-unlock list
+            if slots_taken == 0 && !contributed_stackers.is_empty() {
```
jcnelson, line 651: So to be clear, this means that the following behaviors are expected?

* Alice stacks enough STX for 3 reward slots for 12 cycles.  In cycle 3, the stacking minimum rises such that she only qualifies for 1.99 reward slots (so, just 1).  Alice's remaining 0.99 slots' STX remain locked, right?

* Alice stacks enough STX for 3 reward slots for 12 cycles.  In cycle 6, she does not clear the stacking minimum.  All of her tokens unlock at cycle 6 then, right?

src/chainstate/stacks/boot/mod.rs
kantai: Nope, this should be deleted.

src/chainstate/stacks/boot/mod.rs
kantai: Yep

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -269,34 +271,42 @@
...
+                        (get amount-ustx params)
+                        (get stacker params)
+                        )))
+                  ;; update running total
+                  (map-set reward-cycle-total-stacked
+                     { reward-cycle: reward-cycle }
+                     { total-ustx: (+ (get amount-ustx params) total-ustx) })
+                  (some reward-index))
+            none))
+          (next-i (if (< i num-cycles) (+ i u1) (+ i u0))))
```
jcnelson, line 299: `(+ i u0)`?  Why not just `i`?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -305,16 +315,17 @@
...
+                           reward-set-indexes: (list), amount-ustx: amount-ustx, i: u0, stacker: (some stacker) })))
     ;; For safety, add up the number of times (add-principal-to-ith-reward-cycle) returns 1.
     ;; It _should_ be equal to num-cycles.
-    (asserts! 
-     (is-eq num-cycles 
-                 (get i (fold add-pox-addr-to-ith-reward-cycle cycle-indexes 
-                         { pox-addr: pox-addr, first-reward-cycle: first-reward-cycle, num-cycles: num-cycles, amount-ustx: amount-ustx, i: u0 })))
-     (err ERR_STACKING_UNREACHABLE))
-    (ok true)))
+    (asserts! (is-eq num-cycles (get i results)) (err ERR_STACKING_UNREACHABLE))
```
jcnelson, line 326: Should we replace `(err ERR_STACKING_UNREACHABLE)` with `unwrap-panic`, now that we have the chance?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5143,6 +5147,45 @@ impl StacksChainState {
...
+                    burn_dbconn.get_burn_start_height().into(),
+                    burn_tip_height.into(),
+                    burn_dbconn.get_pox_reward_cycle_length().into(),
+                ) {
+                    pox_reward_cycle -= 1;
+                }
+                let handled = clarity_tx.with_clarity_db_readonly(|clarity_db| {
+                    Self::handled_pox_cycle_start(clarity_db, pox_reward_cycle)
+                });
+                info!("Check handled"; "pox_cycle" => pox_reward_cycle, "handled" => handled);
```
jcnelson, line 5171: Can this log line be more descriptive of what's happening?

src/chainstate/stacks/boot/pox-2.clar
kantai: I can update -- this was just reformatting of the existing pox-1 code

src/chainstate/stacks/boot/pox-2.clar
kantai: Sure

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5070,6 +5071,7 @@ impl StacksChainState {
         chainstate_tx: &'b mut ChainstateTx,
         clarity_instance: &'a mut ClarityInstance,
         burn_dbconn: &'b dyn BurnStateDB,
+        sortition_dbconn: &'b dyn SortitionDBRef,
```
jcnelson, line 5074: Is there a world in which `burn_dbconn` and `sortition_dbconn` are different objects?  Why can't we implement `SortitionDBRef` for the same things we implement `BurnStateDB` for?

src/chainstate/stacks/boot/pox-2.clar
kantai: Hmm -- Clarity doesn't have a unconditional `panic` or panicking assert method, so the way to make this panic would be something like:

```
(unwrap-panic (if (is-eq num-cycles (get i results)) (some true) none)
```

or similar. I'm not sure if that actually increases the readability here.

src/clarity_vm/database/mod.rs
```diff
@@ -224,6 +227,72 @@ fn get_matured_reward(conn: &DBConn, child_id_bhh: &StacksBlockId) -> Option<Min
...
+    handle: &mut SortitionHandleConn,
+    parent_stacks_block_burn_ht: u64,
+    cycle_index: u64,
+) -> Result<Option<PoxStartCycleInfo>, ChainstateError> {
+    let descended_from_last_pox_anchor = match handle.get_last_anchor_block_hash()? {
+        Some(pox_anchor) => handle.descended_from(parent_stacks_block_burn_ht, &pox_anchor)?,
+        None => return Ok(None),
+    };
+
+    if !descended_from_last_pox_anchor {
```
jcnelson, line 253: Wait, there's no reward addresses for PoX if miners choose an anchor block that's the nth ancestor of the last-chosen anchor block?  Is this part of SIP-007?

src/chainstate/stacks/db/blocks.rs
kantai: Not really, but the reason for `SortitionDBRef` is that I didn't want to add these functions to  a `libclarity` interface (`SortitionDBRef` is defined in the main library, not `libclarity`). Ideally we'd be able to do something like trait upcasting, but it's not a thing in rust ([yet?](https://github.com/rust-lang/rust/issues/65991))

src/clarity_vm/database/mod.rs
kantai: I'm not sure I understand this question. This method is attempting to fetch the information stored at the start of the current PoX reward cycle if and only if the current block descends from the anchor block. So, if a PoX anchor block is chosen, if this Stacks block is in a fork that doesn't build off of that anchor block, then it cannot process information for that anchor block.

src/chainstate/burn/db/sortdb.rs
jcnelson: One thing we could do is implement a trait for both `SortitionHandleTx<'_>` and `SortitionHandleConn<'_>` that exposes all the internal state we care about in order to execute this.  Like, `self.get_block_snapshot_by_height()` and `self.tx()` could be the same for both of these (but instead of `fn tx(&mut self)`, you'd have `fn conn(&self)` to return an immutable reference to the inner sqlite connection).

src/chainstate/stacks/boot/pox-2.clar
jcnelson: You could just `unwrap-panic` the `asserts!`

kantai: No, the `asserts!` exits the control-flow. It's an early return.

clarity/src/vm/database/structures.rs
kantai: Yeah, that's a copy-pasta rustdoc, will update.

src/chainstate/stacks/db/blocks.rs
kantai: I'll just remove -- this was used for test debugging, I'll add a `debug!` log to the actual reward cycle start handler.

src/chainstate/stacks/boot/pox-2.clar
jcnelson: Ah, right :facepalm: 

This is probably fine as-is then.

src/chainstate/burn/db/sortdb.rs
kantai: Makes sense. Added in 9e676926c

clarity/src/vm/database/clarity_db.rs
```diff
@@ -1205,6 +1205,23 @@ impl<'a> ClarityDatabase<'a> {
         )
     }
 
+    /// Like fetch_entry_unknown_descriptor, except that it expects
+    ///  to receive a value (i.e., it expects the state to exist)
+    /// This should only ever be invoked outside of Clarity, e.g. by the VM
+    ///  when loading state from a known-contract
```
jcnelson, line 1211: This is really only ever used with `.pox-2`, right?  Also, is there a way to programmatically enforce this property?

clarity/src/vm/database/structures.rs
```diff
@@ -474,6 +474,38 @@ impl<'db, 'conn> STXBalanceSnapshot<'db, 'conn> {
...
+            STXBalance::Unlocked { amount } => STXBalance::Unlocked { amount },
+            STXBalance::LockedPoxOne {
+                amount_unlocked,
+                amount_locked,
+                ..
+            } => STXBalance::LockedPoxOne {
+                amount_unlocked,
+                amount_locked,
+                unlock_height: new_unlock_height,
+            },
```
jcnelson, line 496: We're not going to take the opportunity to migrate an instance of `LockedPoxOne` to `LockedPoxTwo`?  Asking because I don't think `LockedPoxOne` would otherwise ever be instantiated once `.pox-2` and Stacks 2.1 are live, right?  `LockedPoxTwo` is the 2.1-specific representation of token lock state, and we even migrate prior state instances to `LockedPoxTwo` instances when we use 2.1-specific features to load balances.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -837,6 +842,81 @@ impl db_keys {
     }
 }
 
+/// Trait for structs that provide a chaintip-indexed handle into the
+///  SortitionDB (i.e., a MARF view from a particular SortitionId)
+pub trait SortitionHandle {
```
jcnelson, line 847: Thanks for adding this!

src/chainstate/burn/db/sortdb.rs
jcnelson: So is this instance of `descended_from()` redundant, now that `SortitionHandle::descended_from()` exists?

src/chainstate/stacks/boot/mod.rs
```diff
@@ -124,7 +131,340 @@ pub fn make_contract_id(addr: &StacksAddress, name: &str) -> QualifiedContractId
     )
 }
 
+/// Extract a PoX address from its tuple representation
+fn tuple_to_pox_addr(tuple_data: TupleData) -> (AddressHashMode, Hash160) {
```
jcnelson, line 135: I deleted this in my native-segwit branch because it's only used in tests below (where it is duplicated).  I think its presence here is a merge artifact?

clarity/src/vm/database/clarity_db.rs
kantai: I'm not sure if there's a way to enforce this programmatically.

Ultimately, this could also be useful for other contracts that the Stacks chainstate fetches expectant data from.

src/chainstate/stacks/boot/mod.rs
```diff
@@ -124,7 +131,340 @@ pub fn make_contract_id(addr: &StacksAddress, name: &str) -> QualifiedContractId
...
+        for (principal, amount_locked) in cycle_info.missed_reward_slots.iter() {
+            // we have to do several things for each principal
+            // 1. lookup their Stacks account and accelerate their unlock
+            // 2. remove the user's entries from every `reward-cycle-pox-address-list` they were in
+            //     (a) this can be done by moving the last entry to the now vacated spot,
+            //         and, if necessary, updating the associated `stacking-state` entry's pointer
+            //     (b) or, if they were the only entry in the list, then just deleting them from the list
+            // 3. correct the `reward-cycle-total-stacked` entry for every reward cycle they were in
+            // 4. delete the user's stacking-state entry.
+            clarity.with_clarity_db(|db| {
```
jcnelson, line 254: This closure here scares me a little bit because (a) it's pretty big and (b) it's directly writing to and deleting contract state outside of a contract-call invocation (which I don't think we do anywhere else).

Could the bulk of this closure be lifted into a set of `(define-private ...)` functions within `.pox-2` that this closure could simply invoke?  We could easily verify that these functions are unreachable code from within the VM, and writing them in Clarity would (a) shorten the implementation somewhat, (b) let us re-use all of the VM code we currently rely on for mutating contract state, and (c) permit us to easily unit-test them.

What do you think?

src/chainstate/stacks/db/blocks.rs
jcnelson: Ah, gotcha.  Can you add a comment to the docstring that these are expected to be the same object (and maybe there's a way to test that somehow at runtime)?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5143,6 +5147,45 @@ impl StacksChainState {
         let evaluated_epoch = clarity_tx.get_epoch();
         clarity_tx.reset_cost(parent_block_cost.clone());
 
+        debug!("Evaluating block with epoch = {}", evaluated_epoch);
```
jcnelson, line 5150: Can you lift this block into its own function?  This function is already pretty long.

src/clarity_vm/database/mod.rs
jcnelson: Gotcha. I was thrown off by the variable name `descended_from_last_pox_anchor`.  I thought this meant "last PoX anchor that was selected anywhere" (which is wrong) vs. "last PoX anchor selected in this sortition history".

src/chainstate/stacks/db/blocks.rs
kantai: Sure

src/chainstate/stacks/boot/mod.rs
jcnelson: > We do something similar to this for the cost voting contract.

IIRC the Clarity VM will update a couple of Clarity DB key/values directly (`vm-costs::last-processed-at-height` and the cost-contract-specific `::state-summary` key), but -- correct me if I'm wrong -- it won't mutate state within the `cost-voting` contract's data space.  The VM uses the cost contract to determine what these key/values should be, in order to determine what the right cost functions are when it goes to evaluate a block.

Unrelated, looking into this caused me to notice a bug in the cost-voting system (https://github.com/stacks-network/stacks-blockchain/issues/3268) that we can fix for 2.1.

src/chainstate/stacks/db/blocks.rs
kantai: Will add a docstring. I don't think there's a way to do a reference equality check on the trait references without some unsafe code. This is because the pointers are in fact not the same: they are two different fat pointers, so to do the reference equality check, we'd have to (using `unsafe`) grab the bits of the pointer that comprise the content address and then compare those.

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -263,6 +264,64 @@
...
+                                                      first-unlocked-cycle: uint,
+                                                      stacker: principal
+                                                    } int)))
+    (let ((data (try! data-res))
+          (cycle (get cycle data))
+          (first-unlocked-cycle (get first-unlocked-cycle data)))
+         ;; if current-cycle hasn't reached first-unlocked-cycle, just continue to next iter
+         (asserts! (>= cycle first-unlocked-cycle) (ok (merge data { cycle: (+ u1 cycle) })))
+         (let ((cycle-entry (unwrap-panic (map-get? reward-cycle-pox-address-list { reward-cycle: cycle, index: set-index })))
+               ;; todo: assert eq
```
jcnelson, line 278: :eyes:

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -263,6 +264,64 @@
...
+               (cycle-entry-u (get stacker cycle-entry))
+               (cycle-entry-total-ustx (get total-ustx cycle-entry))
+               (cycle-last-entry-ix (- (get len (unwrap-panic (map-get? reward-cycle-pox-address-list-len { reward-cycle: cycle }))) u1)))
+            (asserts! (is-eq cycle-entry-u (some (get stacker data))) (err ERR_STACKING_CORRUPTED_STATE))
```
jcnelson, line 282: How would we handle this error condition?  Should we have the chains coordinator panic?

src/chainstate/stacks/boot/pox-2.clar
kantai: Ah, this todo was actually done, see line 282. Will remove comment

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -263,6 +264,64 @@
...
+                             { reward-cycle: cycle, index: set-index }
+                             move-entry)
+                    (match (get stacker move-entry) moved-stacker
+                     ;; if the moved entry had an associated stacker, update its state
+                     (let ((moved-state (unwrap-panic (map-get? stacking-state { stacker: moved-stacker })))
+                           ;; calculate the index into the reward-set-indexes that `cycle` is at
+                           (moved-cycle-index (- cycle (get first-reward-cycle moved-state)))
+                           (moved-reward-list (get reward-set-indexes moved-state))
+                           ;; reward-set-indexes[moved-cycle-index] = set-index via slice, append, concat.
+                           (update-list-inner (print (concat (append (default-to (list) (slice moved-reward-list u0 moved-cycle-index))
```
jcnelson, line 296: Is this `print` going to stay?  Is it needed for an event?

src/chainstate/stacks/boot/pox-2.clar
kantai: Yes, the block panics off this call errors at all.

src/chainstate/stacks/boot/pox-2.clar
kantai: No, will be removed

src/chainstate/stacks/boot/mod.rs
```diff
@@ -260,206 +237,34 @@ impl StacksChainState {
...
-                            assert!(moved_reward_indexes.len() > moved_cycle_index, "FATAL: Calculated bad move index");
-                            // we moved the entry to the "target" location
-                            moved_reward_indexes[moved_cycle_index] = Value::UInt(target_entry_index);
-
-                            moved_state_entry.data_map.insert("reward-set-indexes".into(),
-                                                              Value::list_from(moved_reward_indexes)
-                                                                .expect("Failed to reconstruct Clarity list"));
-                            // store the new state back into the stacking-state map
-                            db.set_entry_unknown_descriptor(
+            let result = clarity
```
jcnelson, line 243: Just want to confirm -- because this is an "internal" process, there's no block limit here, right?

src/chainstate/stacks/boot/pox-2.clar
jcnelson: Ah, I see the `.expect()`.  Just want to confirm that the transaction erroring out _at all_ will trigger the `.expect()`?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -263,6 +264,64 @@
...
+                    (map-set reward-cycle-pox-address-list
+                             { reward-cycle: cycle, index: set-index }
+                             move-entry)
+                    (match (get stacker move-entry) moved-stacker
+                     ;; if the moved entry had an associated stacker, update its state
+                     (let ((moved-state (unwrap-panic (map-get? stacking-state { stacker: moved-stacker })))
+                           ;; calculate the index into the reward-set-indexes that `cycle` is at
+                           (moved-cycle-index (- cycle (get first-reward-cycle moved-state)))
+                           (moved-reward-list (get reward-set-indexes moved-state))
+                           ;; reward-set-indexes[moved-cycle-index] = set-index via slice, append, concat.
```
jcnelson, line 295: A pity that there isn't a `set-element-at` keyword in Clarity for sequences.  Maybe this could be a separate function that makes this clear?  Then we could at least point it out to developers if they want to do the equivalent.

src/chainstate/stacks/db/blocks.rs
jcnelson: Thanks!

src/chainstate/stacks/boot/pox-2.clar
jcnelson: Cool :+1:

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -261,6 +264,64 @@
...
+                true)
+            ;; in all cases, we now need to delete the last list entry
+            (map-delete reward-cycle-pox-address-list { reward-cycle: cycle, index: cycle-last-entry-ix })
+            (map-set reward-cycle-pox-address-list-len { reward-cycle: cycle } { len: cycle-last-entry-ix })
+            ;; finally, update `reward-cycle-total-stacked`
+            (map-set reward-cycle-total-stacked { reward-cycle: cycle }
+                { total-ustx: (- (get total-ustx (unwrap-panic (map-get? reward-cycle-total-stacked { reward-cycle: cycle })))
+                                 cycle-entry-total-ustx) })
+            (ok (merge data { cycle: (+ u1 cycle)} )))))
+
```
jcnelson, line 314: Maybe add a comment here indicating that this can only get called from an internal process -- i.e. as processing early unlocks?

src/chainstate/stacks/boot/mod.rs
jcnelson: Thanks for taking the time to do this in https://github.com/stacks-network/stacks-blockchain/pull/3260/commits/9d00a6cfd48fe557dc524a6935e7ab9c1af74d79.  I think the unlock implementation is a lot clearer and easier to understand now.

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -305,16 +315,17 @@
...
     ;; For safety, add up the number of times (add-principal-to-ith-reward-cycle) returns 1.
     ;; It _should_ be equal to num-cycles.
-    (asserts! 
-     (is-eq num-cycles 
-                 (get i (fold add-pox-addr-to-ith-reward-cycle cycle-indexes 
-                         { pox-addr: pox-addr, first-reward-cycle: first-reward-cycle, num-cycles: num-cycles, amount-ustx: amount-ustx, i: u0 })))
-     (err ERR_STACKING_UNREACHABLE))
-    (ok true)))
+    (asserts! (is-eq num-cycles (get i results)) (err ERR_STACKING_UNREACHABLE))
+    (asserts! (is-eq num-cycles (len (get reward-set-indexes results))) (err ERR_STACKING_UNREACHABLE))
```
pavitthrap, line 386: Can move `(get reward-set-indexes results)` to let statement for more readability?

pavitthrap (commented):
mostly lgtm, just some minor comments

src/chainstate/burn/db/sortdb.rs
kantai: Yep, removed in 9d00a6cfd48fe557dc524a6935e7ab9c1af74d79

clarity/src/vm/database/clarity_db.rs
kantai: Removed this method as it was only necessary in the previous implementation.

src/chainstate/stacks/boot/mod.rs
kantai: Yep, a merge artifact, removed in 9d00a6cfd48fe557dc524a6935e7ab9c1af74d79

src/chainstate/stacks/boot/mod.rs
kantai: No -- I had to update this to explicitly switch to a free tracker.

kantai: 266caf2ba

src/chainstate/stacks/boot/pox-2.clar
kantai: Removed in 266caf2ba

src/chainstate/stacks/boot/pox-2.clar
kantai: Added a private method in 266caf2ba.

It would be pretty easy to add a new Clarity built-in for this, though. I might try to include in the stacks-increase PR.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -505,6 +585,256 @@ fn test_simple_pox_lockup_transition_pox_2() {
...
+    for b in blocks.into_iter() {
+        for r in b.receipts.into_iter() {
+            if let TransactionOrigin::Stacks(ref t) = r.transaction {
+                let addr = t.auth.origin().address_testnet();
+                eprintln!("TX addr: {}", addr);
+                if addr == alice_address {
+                    alice_txs.insert(t.auth.get_origin_nonce(), r);
+                } else if addr == bob_address {
+                    bob_txs.insert(t.auth.get_origin_nonce(), r);
+                } else if addr == key_to_stacks_addr(&charlie) {
```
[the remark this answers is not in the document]
kantai: Updated in 266caf2ba

src/chainstate/stacks/boot/mod.rs
```diff
@@ -402,25 +850,70 @@ impl StacksChainState {
...
 
             let total_ustx = tuple
                 .get("total-ustx")
                 .expect(&format!("FATAL: no 'total-ustx' in return value from (get-reward-set-pox-address u{} u{})", reward_cycle, i))
                 .to_owned()
                 .expect_u128();
 
+            let stacker = tuple
+                .get("stacker")
+                .expect(&format!("FATAL: no 'total-ustx' in return value from (get-reward-set-pox-address u{} u{})", reward_cycle, i))
```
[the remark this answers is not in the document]
kantai: Updated in 266caf2ba

src/chainstate/stacks/boot/pox-2.clar
kantai: Updated in 266caf2ba

clarity/src/vm/database/structures.rs
jcnelson: Gotcha. Should this be an `unreachable!()` then?

src/burnchains/burnchain.rs
```diff
@@ -491,6 +492,21 @@ impl Burnchain {
...
+    ///  right before the reward phase starts? This is the mod 0 or mod 1
+    ///  block. Reward cycle start events (like auto-unlocks) process *after*
+    ///  the first reward block, so this function is used to determine when
+    ///  that has passed.
+    pub fn is_before_reward_cycle(
+        first_block_ht: u64,
+        burn_ht: u64,
+        reward_cycle_length: u64,
+    ) -> bool {
+        let effective_height = burn_ht - first_block_ht;
```
jcnelson, line 505: Should this be an explicit `.checked_sub(..).expect(..)`?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5061,15 +5063,73 @@ impl StacksChainState {
...
+        let mut pox_reward_cycle = Burnchain::static_block_height_to_reward_cycle(
+            burn_tip_height,
+            burn_dbconn.get_burn_start_height().into(),
+            burn_dbconn.get_pox_reward_cycle_length().into(),
+        ).expect("FATAL: Unrecoverable chainstate corruption: Epoch 2.1 code evaluated before first burn block height");
+        // Do not try to handle auto-unlocks on pox_reward_cycle 0
+        // This cannot even occur in the mainchain, because 2.1 starts much
+        //  after the 1st reward cycle, however, this could come up in mocknets or regtest.
+        if pox_reward_cycle > 1 {
+            if Burnchain::is_before_reward_cycle(
```
jcnelson, line 5087: Given the way this method is used, I just want to understand a few things:

* `burn_tip_height` always increases by 1 between subsequent calls to `check_and_handle_reward_start()`, right?  There's no way that we could process a reward cycle and somehow _skip_ this `if` statement's "true" block?  Asking because the condition `(effective_height % reward_cycle_length) <= 1` suggests that there this block will be executed more than once per reward cycle -- i.e. when the modulus is 0 and when it is 1?

* If this method is supposed to be true only when `burn_tip_height` is the burn block height refers to a block that was mined just before a new reward cycle begins -- i.e. it's the last block in the previous reward cycle, might the method name suggest this?  For example, `is_last_reward_cycle_block()` might indicate this better.  Up to you.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5143,6 +5205,18 @@ impl StacksChainState {
         let evaluated_epoch = clarity_tx.get_epoch();
         clarity_tx.reset_cost(parent_block_cost.clone());
 
+        debug!("Evaluating block with epoch = {}", evaluated_epoch);
+        if evaluated_epoch >= StacksEpochId::Epoch21 {
+            Self::check_and_handle_reward_start(
```
jcnelson, line 5210: Oh interesting -- this is happening before any transactions get processed in this block, including those in the parent microblock stream.  Is that what we want?  Asking because the parent microblock stream could have been mined at any point in the past (including an ancient reward cycle), given that the anchored block miner can confirm any microblock stream they want.  Should this be moved to after the microblock transaction processing happens, but before the anchored block transaction processing happens?

src/chainstate/stacks/db/blocks.rs
kantai: > burn_tip_height always increases by 1 between subsequent calls to check_and_handle_reward_start(), right? There's no way that we could process a reward cycle and somehow skip this if statement's "true" block? Asking because the condition (effective_height % reward_cycle_length) <= 1 suggests that there this block will be executed more than once per reward cycle -- i.e. when the modulus is 0 and when it is 1?

It depends on what your definition of time is for "subsequent calls". This call is checking whether or not the current reward cycle has been "handled" in this Stacks fork. The "current" reward cycle is either the active reward cycle if the reward set has already been stored (`> mod 1`) or the previous reward cycle if the active reward cycle hasn't stored its reward set yet. If Stacks blocks proceed without forks or missed burn blocks, then yes, the burn height increments by 1 on every invocation, and the reward cycle will be "handled" just once. However, if there's multiple Stacks forks, it may be the case that a fork needs to handle the reward cycle again.

> If this method is supposed to be true only when burn_tip_height is the burn block height refers to a block that was mined just before a new reward cycle begins -- i.e. it's the last block in the previous reward cycle, might the method name suggest this? For example, is_last_reward_cycle_block() might indicate this better. Up to you.

Because this method is checking `<= mod 1`, it returns true for two different block heights in each reward cycle. It's trying to figure out whether or not the reward set has already been written by the sortition and chainstate databases -- that occurs at modulus block 1.

src/chainstate/stacks/db/blocks.rs
kantai: No, I think this should happen after the parent microblocks.

src/chainstate/stacks/boot/mod.rs
```diff
@@ -249,39 +396,85 @@ impl StacksChainState {
...
+                "slots_taken" => slots_taken,
+                "stacked_amt" => stacked_amt,
+                "pox_threshold" => threshold,
             );
             for _i in 0..slots_taken {
                 test_debug!("Add to PoX reward set: {:?}", &address);
                 reward_set.push(address.clone());
             }
+            // if stacker did not qualify for a slot *and* they have a stacker
+            //   pointer set by the PoX contract, then add them to auto-unlock list
```
jcnelson, line 443: The case where they wouldn't have a stacker pointer is if they were in a pool, right?

src/chainstate/stacks/db/blocks.rs
jcnelson: > Because this method is checking <= mod 1, it returns true for two different block heights in each reward cycle. It's trying to figure out whether or not the reward set has already been written by the sortition and chainstate databases -- that occurs at modulus block 1.

This is where I think I'm confused.  Why two, and why these two?

Suppose I have a reward cycle in which PoX is active, but where the first Stacks block in this cycle gets mined in the middle of the cycle (so, `burn_tip_height % reward_cycle_len == 1050`).  Then, once this code is reached, the `if-true` block will not get processed, meaning that `pox_reward_cycle -= 1` will _not_ get processed, which in turn means that anyone who clinched zero reward slots for this reward cycle will _not_ unlock in this block (since `pox_reward_cycle` refers to the now-current reward cycle, instead of the previous reward cycle).  Correct me if I'm wrong, but won't this mean that users who earned zero reward slots in the previous reward cycle will remain locked in the current cycle?

More generally, what would happen if users locked their tokens for cycle `N - 1`, but a fork was constructed in cycle `N + k` for some `k >= 1` that skipped all of the blocks in cycle `N`...`N + k - 1` but mined off of a block that descended from cycle `N`'s anchor block (which would have been mined towards the end of cycle `N - 1`)?  It looks like these users would remain locked in this fork, because the code would never call `Self::handle_pox_cycle_start()` for any cycle `N` ... `N + k - 1`?  If so, then I don't think this is what we want?

jcnelson (approved):
This pretty much LGTM.  Thanks @kantai!

I just had one remaining design question about how to handle unlocks across reward cycle boundaries, which is still detailed in the comments.  Let's get that addressed before merging.  Thanks!

src/chainstate/stacks/db/blocks.rs
kantai: The answer right now is that `B3` won't process those unlocks. The current implementation will only try to process unlocks for the most recently processed reward cycle with an active anchor block in the current Stacks fork.

An alternative would be to fetch the most recently "handled" reward cycle in the current Stacks fork, and then "handle" each reward cycle since that point. This, of course, is unbounded work, but it would be a low bound: really deep forks would need to do lots of checks, but the checks would be short, because no processing occurs for blocks that don't descend from the anchor block (which deep forks necessarily wouldn't).

> This is where I think I'm confused. Why two, and why these two?

The reward set and unlock state for reward cycle `n` is computed at `1 mod R`, and is only written to the sortition database once that sortition has been processed. This branch is telling the handler what the latest reward set it could successfully fetch information for is. The reason that it cannot fetch information for `1 mod R` is that miners need to be able to figure that out without already having processed the sortition. This is _possible_ but it would require that the miners plumb the `pox_start_cycle_info` into the block processing code (this is similar to how they compute the first two addresses to send their commitments to in the reward cycle, but it occurs in different areas of the code).

jcnelson: Thanks, I understand now :+1:

jcnelson (approved):
LGTM!  Thanks for taking the time to address this @kantai!  Users are gonna love auto-unlocks :D

src/chainstate/stacks/boot/mod.rs
kantai: Yep -- if they used delegation.

src/burnchains/burnchain.rs
kantai: Sure

pavitthrap (approved):
lgtm

for the future though, I think it would be desirable for the pox contract to have some meta documentation.

kantai:
Fixed the CI tasks by switching to stable (thanks @lgalabru for pointing out that grcov works in stable now!), though it seems like the unit-tests tasks didn't successfully upload the coverage data. Will poke into that separately, but merging this PR now because the tests are passing, and its been approved.

PR #3274, opened 2022-09-02
feat: introduce puppet mode controller

bestmike007:
### Description

This is to introduce the puppet mode for the helium node, in order to better support off-chain system development.
By enabling puppet mode, a puppet controller will be started and listening on a configurable port (20445 by default).

The puppet controller will hold each run loop for 10 minutes, in order to simulate the behavior on the mainnet. This behavior can be controlled by sending control commands through http requests:

+ POST /kick: start mining the next block
+ PUT <integer> /duration: change the block generation duration
+ GET /duration: get current block generation duration
+ PUT <integer> /until: mine until the target block height

This puppet mode feature has already been very helpful in the following use cases:

+ Speed up contract bootstrapping when spinning up a new mocknet environment
+ Speed up integration tests for off-chain systems, e.g. data analysis system, wrap bridge, dex, etc.

This is different from [puppetchain](https://github.com/stacks-network/stacks-blockchain/tree/master/contrib/tools/puppet-chain). The puppet mode for the mocknet allow us to control the block generation on the fly.

### Checklist
- [ ] Changelog is updated

### DIFF
--- a/testnet/stacks-node/conf/puppetnet-conf.toml
+++ b/testnet/stacks-node/conf/puppetnet-conf.toml
@@ -1,0 +1,36 @@
+[node]
+name = "helium-node"
+rpc_bind = "127.0.0.1:20443"
+p2p_bind = "127.0.0.1:20444"
+use_test_genesis_chainstate = true
+enable_puppet_mode = true
+# puppet_bind = "127.0.0.1:20445"
+
+[connection_options]
+public_ip_address = "127.0.0.1:20444"
+
+[burnchain]
+chain = "bitcoin"
+mode = "mocknet"
+commit_anchor_block_within = 0
+
+# These are addresses from the README.md
+[[ustx_balance]]
+# Private key: b8d99fd45da58038d630d9855d3ca2466e8e0f89d3894c4724f0efc9ff4b51f001
+address = "ST2ZRX0K27GW0SP3GJCEMHD95TQGJMKB7G9Y0X1MH"
+amount = 100000000
+
+[[ustx_balance]]
+# Private key: 3a4e84abb8abe0c1ba37cef4b604e73c82b1fe8d99015cb36b029a65099d373601
+address = "ST26FVX16539KKXZKJN098Q08HRX3XBAP541MFS0P"
+amount = 100000000
+
+[[ustx_balance]]
+# Private key: 052cc5b8f25b1e44a65329244066f76c8057accd5316c889f476d0ea0329632c01
+address = "ST3CECAKJ4BH08JYY7W53MC81BYDT4YDA5M7S5F53"
+amount = 100000000
+
+[[ustx_balance]]
+# Private key: 9aef533e754663a453984b69d36f109be817e9940519cc84979419e2be00864801
+address = "ST31HHVBKYCYQQJ5AQ25ZHA6W2A548ZADDQ6S16GP"
+amount = 100000000
--- a/testnet/stacks-node/src/run_loop/puppet.rs
+++ b/testnet/stacks-node/src/run_loop/puppet.rs
@@ -1,0 +1,199 @@
+use crate::ChainTip;
+use async_std::net::TcpListener;
+use async_std::stream::StreamExt;
+use async_std::task::block_on;
+use http_types::mime::JSON;
+use http_types::{Method, Response, StatusCode};
+use std::io;
+use std::io::{Error, Write};
+use std::ops::{Add, DerefMut};
+use std::sync::{Arc, Mutex};
+use std::thread::{sleep, Builder, JoinHandle};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+struct PuppetControl {
+    pub current_block: u64,
+    pub target_block: u64,
+    pub next_block_time: SystemTime,
+    pub block_interval: Duration,
+}
+
+impl Default for PuppetControl {
+    fn default() -> Self {
+        let default_interval = Duration::from_secs(600);
+        Self {
+            current_block: 0,
+            target_block: 1,
+            next_block_time: SystemTime::now().add(default_interval),
+            block_interval: default_interval,
+        }
+    }
+}
+
+pub struct PuppetController {
+    inner: Arc<Mutex<PuppetControl>>,
+    bind_addr: String,
+    join_handle: Option<JoinHandle<Result<(), Error>>>,
+}
+
+impl PuppetController {
+    pub fn new(bind_addr: &str) -> Self {
+        Self {
+            inner: Arc::new(Mutex::new(PuppetControl::default())),
+            bind_addr: bind_addr.to_string(),
+            join_handle: None,
+        }
+    }
+
+    pub fn start(&mut self) {
+        if self.join_handle.is_some() {
+            warn!("Puppet mode control server is already started");
+            return;
+        }
+        info!("Starting puppet mode control server..");
+        let puppet_control = Arc::clone(&self.inner);
+        let puppet_bind = self.bind_addr.to_string();
+        self.join_handle = Some(
+            Builder::new()
+                .name("puppet".into())
+                .spawn(move || {
+                    block_on(async {
+                        let listener = TcpListener::bind(puppet_bind).await?;
+                        info!(
+                            "Start puppet mode control server on: {}",
+                            listener.local_addr()?
+                        );
+
+                        // For each incoming TCP connection, spawn a task and call `accept`.
+                        let mut incoming = listener.incoming();
+                        while let Some(stream) = incoming.next().await {
+                            if stream.is_err() {
+                                return Err(stream.unwrap_err());
+                            }
+                            let stream = stream?;
+                            async_h1::accept(stream.clone(), |req| async {
+                                let mut req = req;
+                                match (req.method(), req.url().path()) {
+                                    (Method::Get, "/") => Ok(Response::new(StatusCode::Ok)),
+                                    (Method::Post, "/puppet/v1/kick") => {
+                                        let mut puppet_control = puppet_control.lock().unwrap();
+                                        if puppet_control.target_block <= puppet_control.current_block {
+                                            puppet_control.target_block =
+                                                puppet_control.current_block + 1;
+                                        }
+                                        Ok(Response::new(StatusCode::Ok))
+                                    }
+                                    (Method::Put, "/puppet/v1/duration") => {
+                                        let body = req.body_string().await;
+                                        match body {
+                                            Ok(x) => {
+                                                let v = x.parse::<u64>().unwrap_or(0);
+                                                if v > 0 {
+                                                    println!("Setting duration to {}", v);
+                                                    io::stdout().flush().unwrap();
+                                                    let mut puppet_control =
+                                                        puppet_control.lock().unwrap();
+                                                    puppet_control.block_interval =
+                                                        Duration::from_secs(v);
+                                                    puppet_control.next_block_time =
+                                                        SystemTime::now()
+                                                            .add(puppet_control.block_interval);
+                                                }
+                                            }
+                                            _ => (),
+                                        }
+                                        Ok(Response::new(StatusCode::Ok))
+                                    }
+                                    (Method::Put, "/puppet/v1/until") => {
+                                        let body = req.body_string().await;
+                                        match body {
+                                            Ok(x) => {
+                                                let v = x.parse::<u64>().unwrap_or(0);
+                                                if v > 0 {
+                                                    let mut puppet_control =
+                                                        puppet_control.lock().unwrap();
+                                                    puppet_control.target_block = if puppet_control.current_block >= v {
+                                                        puppet_control.current_block
+                                                    } else {
+                                                        v
+                                                    };
+                                                    println!("Setting target block to {}", puppet_control.target_block);
+                                                    io::stdout().flush().unwrap();
+                                                }
+                                            }
+                                            _ => (),
+                                        }
+                                        Ok(Response::new(StatusCode::Ok))
+                                    }
+                                    (Method::Get, "/puppet/v1/status") => {
+                                        let mut response = Response::new(StatusCode::Ok);
+                                        let puppet_control = puppet_control.lock().unwrap();
+                                        response.set_content_type(JSON);
+                                        response.set_body(
+                                          format!(
+                                            "{{\"current_block\":{},\"target_block\":{},\"duration\":{},\"next_block_time\":{}}}",
+                                            puppet_control.current_block,
+                                            puppet_control.target_block,
+                                            puppet_control.block_interval.as_secs(),
+                                            puppet_control.next_block_time.duration_since(UNIX_EPOCH).unwrap().as_secs()));
+                                        Ok(response)
+                                    }
+                                    _ => {
+                                        let mut rs = Response::new(StatusCode::BadRequest);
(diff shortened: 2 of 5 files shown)


testnet/stacks-node/conf/puppetnet-conf.toml
```diff
@@ -0,0 +1,35 @@
+[node]
+name = "helium-node"
+rpc_bind = "127.0.0.1:20443"
+p2p_bind = "127.0.0.1:20444"
+use_test_genesis_chainstate = true
+enable_puppet_mode = true
```
jcnelson, line 6: Could you add an example of the `puppet_bind` config option in this file?  Thanks!

testnet/stacks-node/src/run_loop/puppet.rs
```diff
@@ -0,0 +1,176 @@
...
+            inner: Arc::new(Mutex::new(PuppetControl::default())),
+            bind_addr: bind_addr.to_string(),
+            join_handle: None,
+        }
+    }
+
+    pub fn start(&mut self) {
+        if self.join_handle.is_some() {
+            warn!("Puppet mode control server is already started");
+            return ();
```
jcnelson, line 51: Minor nit: `return;` will suffice here.

testnet/stacks-node/src/run_loop/puppet.rs
```diff
@@ -0,0 +1,176 @@
...
+                        let mut incoming = listener.incoming();
+                        while let Some(stream) = incoming.next().await {
+                            if stream.is_err() {
+                                return Err(stream.unwrap_err());
+                            }
+                            let stream = stream?;
+                            async_h1::accept(stream.clone(), |req| async {
+                                let mut req = req;
+                                match (req.method(), req.url().path()) {
+                                    (Method::Get, "/ping") => Ok(Response::new(StatusCode::Ok)),
```
jcnelson, line 77: Could you prefix the API with both a system identifier (e.g. `puppet`) and a version (e.g. `v1`)?  So, this would be `/puppet/v1/ping`.  Thanks!

testnet/stacks-node/src/run_loop/puppet.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    pub fn block_on(&self, chain_tip: &ChainTip) {
+        if self.join_handle.is_none() {
+            return;
+        }
+        info!(
+            "Waiting on block height {}",
+            chain_tip.metadata.stacks_block_height
+        );
+        let mut puppet_control = self.inner.lock().unwrap();
```
jcnelson, line 160: Minor nit: instead of using an explicit `drop()`, could you put this into its own method?

testnet/stacks-node/src/run_loop/puppet.rs
```diff
@@ -0,0 +1,176 @@
...
+        loop {
+            let mut puppet_control = self.inner.lock().unwrap();
+            if puppet_control.target_block > puppet_control.current_block
+                || puppet_control.next_block_time.le(&SystemTime::now())
+            {
+                puppet_control.next_block_time =
+                    SystemTime::now().add(puppet_control.block_interval);
+                break;
+            }
+            drop(puppet_control);
```
jcnelson, line 172: I'm not sure this `drop()` is necessary? Won't `puppet_control` go out of scope at the end of the loop?

jcnelson (commented):
Hey, thanks for sending this PR over!  Just need to see a few more things before I can approve, beyond the inline comments I added:

* Can you move the implementation into `contrib/tools`, and make it so it doesn't compile by default?  You can look at how `contrib/tools/puppet-chain` is set up, for example.

* Can you think of a different name for this system?  `contrib/tools/puppet-chain` already exists.  Maybe this PR could add this controller API to the existing puppet chain?

* Can you add an integration test that demonstrates that invoking the puppet APIs to drive the helium runtime will cause the helium node to advance the way one would expect?  You can use tests from `testnet/stacks-node/src/tests/integrations.rs` for inspiration, but the test should reside under `contrib/tools` as well.

* Can you add some documentation to describe how this new feature works?

testnet/stacks-node/conf/puppetnet-conf.toml
bestmike007: Added.

testnet/stacks-node/src/run_loop/puppet.rs
bestmike007: Prefix added. But I don't understand why adding a system identifier is necessary. It does make sense if we mix the API into a currently existing one, but we're not.

testnet/stacks-node/src/run_loop/puppet.rs
bestmike007: Added a new `with_lock` function to prevent these two `drop()`.

testnet/stacks-node/src/run_loop/puppet.rs
bestmike007: This is necessary because of the next line `sleep`. Refactored with the `with_lock` function mentioned above.

bestmike007:
Hey Jude,

Thanks for the reviews. Before addressing the other issues, I'd like to discuss where to put this feature first.

**First of all, why introduce this feature?** We're currently heavily using this feature in 2 scenarios: 1) Integration tests 2) running test environments and public testnet

In our integration tests, we will:

1. Spin up a mocknet stacks node, and the event observer (stacks-node-api)
2. Deploy contracts
3. Setup contracts, e.g. faucet, create pools, set authorized senders, etc.
4. Start the integration test for an offchain system, and for each step we need to:
    * Send several tx as a user, and wait them to be settled
    * Synchronize from stacks-node (by calling readonly functions, getting data-var and map-entry), and from stacks-node-api (by calling API or even directly reading from the postgres database)
    * Send arbitrage tx as a bot
    * Send triggering tx as a sender

We prefer running integration tests per commit, and also we need to be able to debug on a developer's laptop.
So we tried helium runtime with mocknet mode, by setting `commit_anchor_block_within` to a low value to avoid taking up to 30 mins to complete tests. It works with bootstrapping contracts, however, the offchain system might never catch up (calling several readonly function calls which does not support specifying the tip) since the blocks are generating too frequently.
And we tried puppet-chain, we had to run a mocknet bitcoind, and we didn't find a way to submit transactions to mempool, the rpc endpoint seems to be forwarding requests to bitcoind (correct me if wrong). Even if it works like a mocknet stacks-node, we had to carefully set the duration for each block. Unless we set it long enough, the tests will become flaky.
Both of them fail to work with debugging, which requires the block generation to be paused.

And as for a test environment, except for a quick bootstrap, we also need the block generation to be controlled by our QA or scripts any time later, e.g. skip several blocks for an AMM contract to go to the next reward cycle, faucet at the beginning of an internal dogfooding session without waiting 10 minutes for a new block.

Overall, that's the reason for building this new feature.

**Where to put it then?**

I had thought of putting it into `contrib/tools`, just like the puppet-chain. But then I realized that I needed to copy everything from the helium runtime. And if I put it in `contrib/tools` as a library, the helium runtime in `testnet/stacks-node` will have to depend on `contrib/tools`, which does not make sense and is making it worse, right?

It eventually turns out this way, the same `stacks-node` binary with a runtime configuration to turn this feature on. It's also more convenient without needing another binary to switch on/off the feature.

I'm also wondering if there are better ideas, I'd like to hear more from you and your thoughts.

jcnelson:
The reason I think this belongs in `contrib/` is because the maintainers of this repo cannot commit to supporting this feature with the same rigor as the rest of the codebase.  That's what a PR is after all -- a request for someone else to maintain code the requester wrote, regardless of the degree of support the requester commits to it.  Of course, unconditionally accepting all PRs does not scale: we have only so many person-hours available, and they're all occupied working on the rest of the codebase.  So, there needs to be explicit tiers of support for new PR-submitted features.  At a minimum, there are two feature tiers: they're either guaranteed to be maintained indefinitely by the _repo maintainers_, or they are not.

I think this PR provisionally falls into the latter tier for now.  Features in this tier live in `contrib/` to indicate this level of support: we are willing to accept PRs that add new useful tools or add bugfixes to existing ones, but that's it -- that's all we have time for.  Of course, this assessment is subject to change in the future.  If the feature becomes widely used by many projects to the point where it makes sense for this repo's maintainers to provide more rigorous levels of support, it could be migrated out of `contrib/` and integrated into the main codebase.

> But then I realized that I needed to copy everything from the helium runtime. And if I put it in contrib/tools as a library, the helium runtime in testnet/stacks-node will have to depend on contrib/tools, which does not make sense and is making it worse, right?

One way to do this is to have your PR patch the helium node to incorporate your PR's code as an opt-in compile-time flag.  There isn't very much contact between the helium node and your PR, so I think you could easily confine the points of contact to a single file or a small number of functions.  That's something we could commit to supporting indefinitely, since we already test that the helium node compiles and runs correctly with the default compiler options.  Users would be free to activate your PR's feature with an extra compile-time flag (which you would document), but doing so would correctly inform them that they're using the helium node in an way that does not receive first-tier support from us (so they'd have to inform you, not us, about bugs).

Does that work for you?

gregorycoppola:
A simple alternative that I did in my testing of the Hiro subnet L2 is as follows:

I modified the `BurnchainConfig` of the L1 to take a new scalar `wait_before_simulated_block`:

```
[burnchain]
chain = "bitcoin"
mode = "mocknet"
wait_before_simulated_block = 20000
```

Then, in the helium node `RunLoop`, I just had it sleep this long.

This way you can get a new L1 block every, N seconds (e.g. 20 seconds), and you can write your tests around that.

gregorycoppola:
But!.. there is another alternative, where you use `bitcoind` in regtest mode, and then your L1 can be configured against the `bitcoind` node, and you can send requests to the `bitcoind` node to have blocks whenever you want. That way, the entire change can be made in contrib.

bestmike007:
20 seconds should be enough for running tests, however, the facts is that we have hundreds of tx to be settled just to prepare the contracts (deploy contracts, setting initial data-var/map-entry, etc.) And every step of the tests will cost ~20 seconds. And the whole integration test will cost us more than 20 minutes per commit, which is not quite acceptable.

And for running a test environment for QA, every time we make changes to the contract, we'll have to wait for more than 30 mins to get it reset and back on.

gregorycoppola:
Yes.. tests like this inherently take a long time to run. I think it's unavoidable. The way to solve this is to get better at running long-running jobs, possibly multiple independent runs happening in parallel.

I am happy to chat offline if you want.

bestmike007:
> But!.. there is another alternative, where you use `bitcoind` in regtest mode, and then your L1 can be configured against the `bitcoind` node, and you can send requests to the `bitcoind` node to have blocks whenever you want. That way, the entire change can be made in contrib.

Yes I had also considered that. Is it possible to use `test genesis data` along with the `bitcoind` mocknet mode? And how long will it take to get it ready? bc it takes only ~1 seconds to get the mocknet helium node ready.

bestmike007:
> Yes.. tests like this inherently take a long time to run. I think it's unavoidable. The way to solve this is to get better at running long-running jobs, possibly multiple independent runs happening in parallel.
> 
> I am happy to chat offline if you want.

Sure, you can find me on Discord.

gregorycoppola:
@bestmike007 You can try using `clarinet` https://www.hiro.so/clarinet

It will run the bitcoin node and stacks node.

bestmike007:
Thanks Greg, we've tried that. And the reason we're not using it is already stated above.

And I understand the reasons you need this feature separated from the main code base, after thinking about it I figured it's the best way to keep it as it is in this fork: https://github.com/bestmike007/stacks-puppet-node/tree/puppet-node, and I'll keep it up to date with the latest release and add documents in the [wiki](https://github.com/bestmike007/stacks-puppet-node/wiki) to demonstrate how it works for off-chain system integration tests.

I'll mark this PR as draft, and I'll continue adding new features, e.g. simulate a fork and reorg. If anyone is interested, please follow up this PR and I'd really like to help.

gregorycoppola:
Here is another alternative: https://github.com/zone117x/stacks-regtest-env


This is definitely `contrib` territory, and can live in a private repo. You can use the `stacks-node` as a black box, and just manage the `bitcoind` behavior.

gregorycoppola:
To spin up a local node for "regtest" style debugging, there are two alternatives:
* https://github.com/zone117x/stacks-regtest-env/ which has docker scripts for running a `stacks-node` with a local `bitcoind` regtest node, if you want to interact primarily through the command line, or do very arbitrary things
* https://github.com/hirosystems/clarinet which is an official Hiro product has a nice visual UI, and parameters to manage some things you might want to vary

Adding regtest-style control code to the `stacks-blockchain` is something we've been moving away from.

jcnelson:
Hey @bestmike007, I see this is still a draft.  Does this supersede your other PR?

jcnelson:
Hey @bestmike007, we spoke about this some more at the blockchain meeting, and @lgalabru recommends that you consider using the latest Clarinet for this purpose.  Would that be a better fit for your needs than this PR?

bestmike007:
> Hey @bestmike007, I see this is still a draft. Does this supersede your other PR?

No, the other PR is not related.


> Hey @bestmike007, we spoke about this some more at the blockchain meeting, and @lgalabru recommends that you consider using the latest Clarinet for this purpose. Would that be a better fit for your needs than this PR?

I've just checked the latest version (clarinet-cli 1.4.2), and yes it's better now. It can serve almost the same purpose by running `bitcoin-cli -regtest generate 1`.

However, I still prefer the solution in this PR, which can deploy and setup contracts (more than 1,000 transactions) within 1 minutes, while with DevNet it's heavier and might take a few minutes to just get the stacks node ready.

Currently our integration tests costs less than 3 minutes (including setting up contracts, making trades, running bots, and synchronizing onchain data into database); I don't really want to spend time to integrate with DevNet and probably triple the time for running GitHub Actions.

I understand that you're moving away things like this, do you want me to close this PR? I can keep the fork and rebase your develop/main branch updates.

btw @lgalabru , the official build for clarinet-cli (https://github.com/hirosystems/clarinet/releases/download/v1.4.2/clarinet-linux-x64-glibc.tar.gz) does not run on Ubuntu 20.04 LTS (or below) or CentOS 7. You might need to compile it with lower GLIBC version (2.17 in CentOS 7), in order to make it compatible with most linux distributions.

lgalabru:
> However, I still prefer the solution in this PR, which can deploy and setup contracts (more than 1,000 transactions) within 1 minutes, while with DevNet it's heavier and might take a few minutes to just get the stacks node ready.

The same stacks-node software is being ran under the hood, it's mostly a matter a configuration.
By default, devnet are configuring the underlying stacks-node to produce blocks quickly, for quick feedback loops (so that block time can be dragged down to 2 to 3s).
If you're project includes a lot of contracts deployments and contract calls, you could tweak the following settings in the file `settings/Devnet.toml`:
```toml
[devnet]
stacks_node_wait_time_for_microblocks = 1_000
stacks_node_first_attempt_time_ms = 15_000
stacks_node_subsequent_attempt_time_ms = 5_000
```

> the official build for clarinet-cli (https://github.com/hirosystems/clarinet/releases/download/v1.4.2/clarinet-linux-x64-glibc.tar.gz) does not run on Ubuntu 20.04 LTS (or below) or CentOS 7. 

Thanks for reporting, we have a CI iteration in the pipe, I added your feedback there https://github.com/hirosystems/clarinet/issues/557#issuecomment-1437814427.

PR #3275, opened 2022-09-02
feat: Specify `ClarityVersion` on Docs Functions and Keywords

### LINKED ISSUE #2716: [docs] Mark new functions as requiring Stacks 2.1 in vm::docs
The problem is that in the vm::docs, we now have functions, like "buff-to-int-le", that require Stacks 2.1. But, we do not have a nice way to annotate these. Currently, I have just put in plain text:

> Note: This function is only available starting with Stacks 2.1.

However, we can consider doing something nicer, like having a little "chip" or something.

This is up for discussion. We could have a little "chip" that says "Stacks 2.1".

**Additional context**
See the discussion in https://github.com/blockstack/stacks-blockchain/pull/2691#discussion_r653085585

gregorycoppola:
### Description

Add a docs field for first Clarity version available for keywords and functions.

### Applicable issues
- towards #2716 
- Note: This PR doesn't implement "sorting by function name", which is from the discussion on 2716. I can do that in a separate PR.

### Checklist
- [x] Reuse exiting tests
- [x] Changelog is updated


clarity/src/vm/docs/mod.rs
```diff
@@ -2192,6 +2210,7 @@ fn make_for_special(api: &SpecialAPI, name: String) -> FunctionAPI {
         signature: api.signature.to_string(),
         description: api.description.to_string(),
         example: api.example.to_string(),
+        version: ClarityVersion::Clarity1,
```
kantai, line 2213: This isn't accurate for all special functions -- some of them are `Clarity2`.

I'd suggest something like:

```rust
fn make_for_special(api: &SpecialAPI, function: &NativeFunctions) -> FunctionAPI {
    FunctionAPI {
        name: function.get_name(),
        input_type: api.input_type.to_string(),
        output_type: api.output_type.to_string(),
        signature: api.signature.to_string(),
        description: api.description.to_string(),
        example: api.example.to_string(),
        version: function.get_version(),
```

clarity/src/vm/docs/mod.rs
```diff
@@ -102,20 +110,23 @@ const TX_SENDER_KEYWORD: KeywordAPI = KeywordAPI {
...
     example: "(print tx-sender) ;; Will print out a Stacks address of the transaction sender",
+    version: ClarityVersion::Clarity1,
 };
 
 const TX_SPONSOR_KEYWORD: KeywordAPI = KeywordAPI {
     name: "tx-sponsor?",
     output_type: "optional principal",
     description: "Returns the sponsor of the current transaction if there is a sponsor, otherwise returns None.",
     example: "(print tx-sponsor?) ;; Will print out an optional value containing the Stacks address of the transaction sponsor",
+    version: ClarityVersion::Clarity1,
```
kantai, line 121: `tx-sponsor?` is `Clarity2`.

Rather than re-specifying the version for all the keywords, it'd be better to use the `NativeVariables` field.

I think the way I'd refactor this is to add new internal struct `SimpleKeywordAPI` which was:

```rust
struct SimpleKeywordAPI {
    name: &'static str,
    output_type: &'static str,
    description: &'static str,
    example: &'static str,
}
```

And then change all these consts to be instances of that struct rather than `KeywordAPI`.

Then, `make_keyword_reference` would be something like:

```rust
fn make_keyword_reference(variable: &NativeVariables) -> KeywordAPI {
    let simple_api = match variable {
        NativeVariables::TxSender => TX_SENDER_KEYWORD.clone(),
        NativeVariables::ContractCaller => CONTRACT_CALLER_KEYWORD.clone(),
        NativeVariables::NativeNone => NONE_KEYWORD.clone(),
        NativeVariables::NativeTrue => TRUE_KEYWORD.clone(),
        NativeVariables::NativeFalse => FALSE_KEYWORD.clone(),
        NativeVariables::BlockHeight => BLOCK_HEIGHT.clone(),
        NativeVariables::BurnBlockHeight => BURN_BLOCK_HEIGHT.clone(),
        NativeVariables::TotalLiquidMicroSTX => TOTAL_LIQUID_USTX_KEYWORD.clone(),
        NativeVariables::Regtest => REGTEST_KEYWORD.clone(),
        NativeVariables::Mainnet => MAINNET_KEYWORD.clone(),
        NativeVariables::ChainId => CHAINID_KEYWORD.clone(),
        NativeVariables::TxSponsor => TX_SPONSOR_KEYWORD.clone(),
    };
    KeywordAPI {
        // name could even be `variable.get_name()` here, and then
        // `SimpleKeywordAPI` could drop the name field
        name: simple_api.name,
        output_type: simple_api.output_type,
        description: simple_api.description,
        example: simple_api.example,
        version: variable.get_version(),
    }
}
```

That way, we don't end up in a situation where versions need to be specified in two different places.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ This release will contain consensus-breaking changes.
 - Clarity function `stx-transfer?` now takes a 4th optional argument, which is a memo.
 - Added a new parser which will be used to parse Clarity code beginning with 2.1,
   resolving several bugs in the old parser and improving performance.
+- Documentations will indicate explicitly which Clarity version introduced each
```
jcnelson, line 17: Grammar: "Documentation" is a collective noun -- it's already plural.

gregorycoppola: lol. ok changed.

clarity/src/vm/docs/mod.rs
gregorycoppola: did this

clarity/src/vm/docs/mod.rs
gregorycoppola: did this

kantai (approved):

jcnelson (approved):

PR #3278, opened 2022-09-03
Add PoX-2 info to `/v2/pox` RPC endpoint

zone117x:
Fixes https://github.com/stacks-network/stacks-blockchain/issues/3262

Fixes the `contract_id` value in the `GET /v2/pox` response, along with the associated contract variables, so that it uses the active PoX contract.

Also adds an additional properties to the response in order to help clients perform the correct Stacking operations.
The `contract_versions` array which exposes contract ID and PoX activation information for each PoX version, and a simple `current_burnchain_block_height` property so that clients do not have to perform an additional lookup to `/v2/info` for the value while constructing a PoX transaction. 

Example:
```json
{
  "contract_id": "ST000000000000000000002AMW42H.pox-2", <-- fixed
  "first_burnchain_block_height": 0,
  "current_burnchain_block_height": 123, <-- new
  "contract_versions": [ <-- new
    {
      "contract_id": "ST000000000000000000002AMW42H.pox",
      "activation_burnchain_block_height": 0,
      "first_reward_cycle_id": 0
    },
    {
      "contract_id": "ST000000000000000000002AMW42H.pox-2",
      "activation_burnchain_block_height": 100,
      "first_reward_cycle_id": 21
    }
  ]
}
```

### Checklist
- [ ] Changelog is updated

### DIFF
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -1085,6 +1085,7 @@ pub struct RPCPoxInfoData {
     pub contract_id: String,
     pub pox_activation_threshold_ustx: u64,
     pub first_burnchain_block_height: u64,
+    pub current_burnchain_block_height: u64,
     pub prepare_phase_block_length: u64,
     pub reward_phase_block_length: u64,
     pub reward_slots: u64,
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -1072,6 +1072,13 @@ pub struct RPCPoxNextCycleInfo {
     pub ustx_until_pox_rejection: u64,
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct RPCPoxContractVersion {
+    pub contract_id: String,
+    pub activation_burnchain_block_height: u64,
+    pub first_reward_cycle_id: u64,
+}
+
 /// The data we return on GET /v2/pox
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct RPCPoxInfoData {
@@ -1094,9 +1101,8 @@ pub struct RPCPoxInfoData {
     pub rejection_votes_left_required: u64,
     pub next_reward_cycle_in: u64,
 
-    // Stacks 2.1 / PoX-2 info
-    pub pox_2_activation_burnchain_block_height: u64,
-    pub pox_2_first_cycle_id: u64,
+    // Information specific to each PoX contract version
+    pub contract_versions: Vec<RPCPoxContractVersion>,
 }
 
 /// Headers response payload
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -1093,6 +1093,10 @@ pub struct RPCPoxInfoData {
     pub reward_cycle_length: u64,
     pub rejection_votes_left_required: u64,
     pub next_reward_cycle_in: u64,
+
+    // Stacks 2.1 / PoX-2 info
+    pub pox_2_activation_burnchain_block_height: u64,
+    pub pox_2_first_cycle_id: u64,
 }
 
 /// Headers response payload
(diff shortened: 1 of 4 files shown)


src/net/rpc.rs
```diff
@@ -268,11 +268,36 @@ impl RPCPoxInfoData {
     ) -> Result<RPCPoxInfoData, net_error> {
         let mainnet = chainstate.mainnet;
         let chain_id = chainstate.chain_id;
-        let contract_identifier = boot_code_id("pox", mainnet);
+
+        let current_burn_height = chainstate
+            .maybe_read_only_clarity_tx(&sortdb.index_conn(), tip, |clarity_tx| {
```
jcnelson, line 273: Can you use `with_read_only_clarity_tx` instead?  `maybe_read_only_clarity_tx` queries the unconfirmed microblock chain, which is unnecessary here.

src/net/mod.rs
```diff
@@ -1093,6 +1093,10 @@ pub struct RPCPoxInfoData {
     pub reward_cycle_length: u64,
     pub rejection_votes_left_required: u64,
     pub next_reward_cycle_in: u64,
+
+    // Stacks 2.1 / PoX-2 info
+    pub pox_2_activation_burnchain_block_height: u64,
```
jcnelson, line 1098: Can you organize these into a separate sub-object that describes each version of PoX, such as the following?

```
"contract_versions": [
   {
     "contract_id": "SP000000000000000000002Q6VF78.pox",
     "activation_burnchain_block_height": 666050,
     "first_reward_cycle_id": 0
   },
   {
     "contract_id": "SP000000000000000000002Q6VF78.pox-2",
     "activation_burnchain_block_height": ...
     "first_reward_cycle_id": ...
   }
]
```

This would future-proof this information so if there's ever a `pox-3` or subsequent revisions, this field's structure would remain semantically unchanged (it's just that the array of contract version objects would grow by one for each additional revision).

jcnelson (commented):
Thanks for taking the initiative here @zone117x!  Just had one revision I'd like to see before approving, which is to future-proof the data returned by `/v2/pox`.  But otherwise looks good!

src/net/mod.rs
jcnelson: You'd also want to add a test for this, such as by modifying `test_rpc_getpoxinfo()`.

src/net/rpc.rs
zone117x: Done!

src/net/mod.rs
zone117x: Great idea, done except for the test, I'm not quite sure what to change for that.

jcnelson (approved):
Thanks @zone117x! LGTM

kantai (approved):

src/net/rpc.rs
```diff
@@ -420,7 +448,7 @@ impl RPCPoxInfoData {
         let cur_cycle_pox_active = sortdb.is_pox_active(burnchain, &burnchain_tip)?;
 
         Ok(RPCPoxInfoData {
-            contract_id: boot_code_id("pox", chainstate.mainnet).to_string(),
+            contract_id: boot_code_id(next_cycle_pox_contract, chainstate.mainnet).to_string(),
```
pavitthrap, line 451: Do you want to return `cur_cycle_pox_contract` or `next_cycle_pox_contract` here?

zone117x: Good catch, it should probably `cur_cycle_pox_contract`.

Although, clients will likely use the new `contract_versions` payload to determine Stacking tx construction, since blindly using the `contract_id` here can result in errors according to the PoX-2 phase transition plan.

zone117x: Fixed in [9fd5c77](https://github.com/stacks-network/stacks-blockchain/pull/3278/commits/9fd5c772c020f914e6e6cbd03ea4b62f1b01bc2a)

PR #3281, opened 2022-09-07
chore: temporary work-around for most common /v2/pox 500 error

zone117x:
Temporary fix for https://github.com/stacks-network/stacks-blockchain/issues/3261

This change simply specifies `is_pox_active: false` when encountering the `db_error:NotFoundError` rather than returning 500, because the rest of the response is still good enough for clients trying to construct a PoX transaction.

This obviously doesn't fix the root cause of whatever is going on, but it unblocks other projects for now. 

Feel free to leave unmerged and can close when a proper fix is implemented.


src/net/rpc.rs
```diff
@@ -417,7 +417,12 @@ impl RPCPoxInfoData {
             .ok_or_else(|| net_error::DBError(db_error::Overflow))?
             as u64;
 
-        let cur_cycle_pox_active = sortdb.is_pox_active(burnchain, &burnchain_tip)?;
+        let cur_cycle_pox_active = sortdb
+            .is_pox_active(burnchain, &burnchain_tip)
+            .unwrap_or_else(|e| {
+                error!("Error in /v2/pox while determining if PoX is active: {}", e);
+                false
```
pavitthrap, line 424: Do we want to make the `is_pox_active` an option here...? returning `false` seems a bit misleading to me, whereas `None` more explicitly shows the lack of data

zone117x: I imagine (after the node has boostrapped enough burnchain blocks), that this value is only ever true/false -- and that a None/error value is a bug.

pavitthrap: That is a fair point. However, the false case for `is_pox_active` as is seems overloaded - it's used both to signal the node is bootstrapping & that the network voted to reject pox for that cycle.

jcnelson:
Hey, can you post your PoX constants?  Specifically, at what height does PoX 2 activate, and what's the v1 unlock height?  I might have found something.

zone117x:
> Hey, can you post your PoX constants? Specifically, at what height does PoX 2 activate, and what's the v1 unlock height? I might have found something.

Here's the config I'm currently using: https://github.com/zone117x/stacks-regtest-env/blob/ca162a25fd49b1a0a83f385dde627fb7acc2938c/stacks-krypton-miner.toml

So something like:
```toml
[burnchain]
chain = "bitcoin"
mode = "krypton"
poll_time_secs = 1
pox_2_activation = 104

### bitcoind-regtest connection info
peer_host = "$BITCOIN_PEER_HOST"
peer_port = $BITCOIN_PEER_PORT
rpc_port = $BITCOIN_RPC_PORT
rpc_ssl = false
username = "$BITCOIN_RPC_USER"
password = "$BITCOIN_RPC_PASS"
timeout = 30

[[burnchain.epochs]]
epoch_name = "1.0"
start_height = 0

[[burnchain.epochs]]
epoch_name = "2.0"
start_height = 1

[[burnchain.epochs]]
epoch_name = "2.05"
start_height = 2

[[burnchain.epochs]]
epoch_name = "2.1"
start_height = 103
```

And the (very primitivate) bitcoind regtest mining script creates a new block as soon as the btc mempool is receives a tx from the Stacks miner (with an optional sleep/delay):
https://github.com/zone117x/stacks-regtest-env/blob/ca162a25fd49b1a0a83f385dde627fb7acc2938c/Dockerfile.e2e#L119-L131

zone117x:
@jcnelson can you confirm that you were able to reproduce this same issue and have a fix in mind? If not, I can see if anyone at Hiro might be able to help.

jcnelson:
I think it's fixed in #3341, if you want to try that.

zone117x:
> I think it's fixed in #3341, if you want to try that.

Typo in the referenced number?

zone117x:
Ahh @jcnelson it's PR https://github.com/stacks-network/stacks-blockchain/pull/3347 with the potential fix right?

jcnelson:
Closing in favor of https://github.com/stacks-network/stacks-blockchain/pull/3399.

PR #3282, opened 2022-09-07
2.1: Feature: `stack-increase` for PoX-2

kantai:
This PR adds two new methods to `pox-2`:

* `stack-increase`: this method increases the lockup for the stacker, and adds the new amount to the reward sets of the remaining cycles for the user
* `delegate-stack-increase`: this method increases the lockup for the supplied stacker, adding the new amount to the partially-stacked state

As part of this work, this also removes the `total-ustx` field from the `stacking-state` map. The rationale for this is to normalize some of the state in the PoX-2 contract: this field is not used for any checks, but it would be "replicated" in several other states: that's at least the `stx-account` lock amount and the reward-set entries for the stacker in each cycle. Because of the "increase" methods, the `total-ustx` field would either be out of sync with the increased amount or the previous cycle entries. To keep this field meaningful, it would need to be a list of `total-ustx` for each cycle, but because the field isn't used in the contract anywhere then it seemed better to just remove the state.

### DIFF
--- a/src/chainstate/stacks/boot/pox-2.clar
+++ b/src/chainstate/stacks/boot/pox-2.clar
@@ -22,6 +22,7 @@
 (define-constant ERR_INVALID_START_BURN_HEIGHT 24)
 (define-constant ERR_NOT_CURRENT_STACKER 25)
 (define-constant ERR_STACK_EXTEND_NOT_LOCKED 26)
+(define-constant ERR_STACK_INCREASE_NOT_LOCKED 27)
 
 ;; PoX disabling threshold (a percent)
 (define-constant POX_REJECTION_FRACTION u25)
@@ -63,8 +64,6 @@
 (define-map stacking-state
     { stacker: principal }
     {
-        ;; how many uSTX locked?
-        amount-ustx: uint,
         ;; Description of the underlying burnchain address that will
         ;; receive PoX'ed tokens. Translating this into an address
         ;; depends on the burnchain being used.  When Bitcoin is
@@ -560,8 +559,7 @@
           ;; add stacker record
          (map-set stacking-state
            { stacker: tx-sender }
-           { amount-ustx: amount-ustx,
-             pox-addr: pox-addr,
+           { pox-addr: pox-addr,
              reward-set-indexes: reward-set-indexes,
              first-reward-cycle: first-reward-cycle,
              lock-period: lock-period })
@@ -703,8 +701,7 @@
       ;; add stacker record
       (map-set stacking-state
         { stacker: stacker }
-        { amount-ustx: amount-ustx,
-          pox-addr: pox-addr,
+        { pox-addr: pox-addr,
           first-reward-cycle: first-reward-cycle,
           reward-set-indexes: (list),
           lock-period: lock-period })
@@ -801,8 +798,6 @@
          (stacker-state (unwrap! (map-get? stacking-state 
                                           { stacker: tx-sender })
                                           (err ERR_STACK_EXTEND_NOT_LOCKED))))
-      ;; stacker state sanity check                                         
-      (asserts! (is-eq amount-stacked (get amount-ustx stacker-state)) (err ERR_STACKING_UNREACHABLE))
       (asserts! (> amount-stacked u0)
                 (err ERR_STACK_EXTEND_NOT_LOCKED))
       (asserts! (>= increase-by u1)
@@ -819,11 +814,7 @@
                     stacker: tx-sender,
                     add-amount: increase-by })))
             (err ERR_STACKING_UNREACHABLE))
-      (map-set
-          stacking-state
-          { stacker: tx-sender }
-          (merge stacker-state 
-                 { amount-ustx: (+ amount-stacked increase-by) }))
+      ;; NOTE: stacking-state map is unchanged: it no longer tracks amount-stacked in PoX-2
       (ok { stacker: tx-sender, total-locked: (+ amount-stacked increase-by)})))
 
 ;; Extend an active stacking lock.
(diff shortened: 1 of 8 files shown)


zone117x:
Closes https://github.com/stacks-network/stacks-blockchain/issues/2533

(referencing the issue for anyone following along)

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -63,8 +64,6 @@
 (define-map stacking-state
     { stacker: principal }
     {
-        ;; how many uSTX locked?
-        amount-ustx: uint,
```
jcnelson, line 67: Do you think you could provide a read-only function that returns the number of locked tokens for a principal, via `stx-account`?  Asking because I'm pretty sure stacking.club depends on this information (and there could be others).

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -761,6 +758,65 @@
...
+    (let ((data (try! updates))
+          (first-cycle (get first-cycle data))
+          (reward-cycle (get reward-cycle data)))
+    (if (> first-cycle reward-cycle)
+        ;; not at first cycle to process yet
+        (some { first-cycle: first-cycle, reward-cycle: (+ u1 reward-cycle), stacker: (get stacker data), add-amount: (get add-amount data) })
+        (let ((existing-entry (unwrap-panic (map-get? reward-cycle-pox-address-list { reward-cycle: reward-cycle, index: reward-cycle-index })))
+              (existing-total (unwrap-panic (map-get? reward-cycle-total-stacked { reward-cycle: reward-cycle }))))
+            ;; stacker must match
+            (asserts! (is-eq (get stacker existing-entry) (some (get stacker data))) none)
```
jcnelson, line 773: If `(is-some (get stacker data))` is true, then wouldn't this assertion's failure be a Really Bad Thing, i.e. indicative that `reward-cycle-pox-address-list` was corrupted somehow, since the `reward-cycle-index` didn't refer to a PoX address from this stacker?

I can see why `(get stacker data)` could be `none` -- e.g. when the stacker had stacked through a delegate.  But, should this function even be called when this is the case (or, can we not tell prior to calling that this is the case)?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -761,6 +758,65 @@
     })
 )
 
+(define-private (increase-reward-cycle-entry
```
jcnelson, line 761: I'm not seeing how this function could ever return `none`?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -761,6 +758,65 @@
...
+      (asserts! (check-caller-allowed)
+                (err ERR_STACKING_PERMISSION_DENIED))
+      ;; update reward cycle amounts
+      (asserts! (is-some (fold increase-reward-cycle-entry 
+            (get reward-set-indexes stacker-state)
+            (some { first-cycle: first-increased-cycle,
+                    reward-cycle: (get first-reward-cycle stacker-state),
+                    stacker: tx-sender,
+                    add-amount: increase-by })))
+            (err ERR_STACKING_UNREACHABLE))
```
jcnelson, line 816: So if this `(fold increase-reward-cycle-entry)` is always supposed to be `(some ..)`, then doesn't that mean that internally, `tx-sender` must not have stacked through a delegate?  Can we check this as part of the checks above this line?

src/chainstate/stacks/db/accounts.rs
```diff
@@ -388,6 +388,59 @@ impl StacksChainState {
...
+    pub fn pox_lock_increase_v2(
+        db: &mut ClarityDatabase,
+        principal: &PrincipalData,
+        new_total_locked: u128,
+    ) -> Result<STXBalance, Error> {
+        assert!(new_total_locked > 0);
+
+        let mut snapshot = db.get_stx_balance_snapshot(principal);
+
+        if !snapshot.has_locked_tokens() {
```
jcnelson, line 407: I find myself wondering whether or not these `Err(..)` paths are ever reachable.  If the Clarity code for `pox-2` is correct, then these should never be hit, right?

jcnelson: Ah, I see we `panic()` later in the special case handler.

jcnelson (approved):
This overall LGTM, modulo the failing unit test.  Just had a few questions about how defensive we can be in the `stack-increase`.

src/chainstate/stacks/boot/pox-2.clar
kantai: Yes, this would be a Really Bad Thing, but Clarity cannot assert a node panic, so instead, we just panic the transaction.

src/chainstate/stacks/boot/pox-2.clar
kantai: Yep, will add a check.

jcnelson:
N.B from blockchain meeting:

* Increase code coverage in pox-2 to verify that the bi-directional map between stackers and the reward set state remains consistent.  Pertaining to this, it's `stack-increase`, `delegate-stack-increase`, auto-unlock, and `stack-extend`.  We should spend some time convincing ourselves that the structure of this map is preserved in all cases in which these are called.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -840,6 +873,518 @@ fn test_simple_pox_2_auto_unlock(alice_first: bool) {
...
+    for _i in 0..10 {
+        peer.tenure_with_txs(&[], &mut coinbase_nonce);
+    }
+
+    // in the next tenure, PoX 2 should now exist.
+    let tip = get_tip(peer.sortdb.as_ref());
+
+    // submit an increase: this should fail, because Alice is not yet locked
+    let fail_no_lock_tx = alice_nonce;
+    let alice_increase = make_pox_2_increase(&alice, alice_nonce, 512 * POX_THRESHOLD_STEPS_USTX);
```
pavitthrap, line 1222: could you define `512 * POX_THRESHOLD_STEPS_USTX` as a variable in this test too?

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -840,6 +873,518 @@ fn test_simple_pox_2_auto_unlock(alice_first: bool) {
...
+                    alice_txs.insert(t.auth.get_origin_nonce(), r);
+                }
+            }
+        }
+    }
+
+    assert_eq!(alice_txs.len() as u64, alice_nonce);
+
+    // transaction should fail because lock isn't applied
+    assert_eq!(&alice_txs[&fail_no_lock_tx].result.to_string(), "(err 26)");
```
pavitthrap, line 1376: could also include error "name" in the comments

pavitthrap (approved):
lgtm, just had some suggestions wrt comments & readability

PR #3288, opened 2022-09-12
Fix/3287

a3slade:
Fixed a typo in the function that converts a string to an enum where the string "clarity2" was returning Clarity1.


jcnelson (approved):

pavitthrap (approved):
LGTM!

pavitthrap:
Verified that this function is not used anywhere - no regression testing is necessary.

PR #3290, opened 2022-09-13
chore: upgrade `chronos` to address RUSTSEC-2020-0159

diwakergupta:
From `cargo audit`:

```
Crate:     chrono
Version:   0.4.19
Title:     Potential segfault in `localtime_r` invocations
Date:      2020-11-10
ID:        RUSTSEC-2020-0159
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0159
Solution:  Upgrade to >=0.4.20
```

Changelog: https://github.com/chronotope/chrono/releases/tag/v0.4.20

Note that I did not upgrade to the latest dependency -- 0.4.22 -- as it pulls in a few other depenendencies that I wasn't sure about, notably `iana-time-zone`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -493,14 +493,15 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chrono"
-version = "0.4.19"
+version = "0.4.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "670ad68c9088c2a963aaa298cb369688cf3f9465ce5e2d4ca10e6e0098a1ce73"
+checksum = "6127248204b9aba09a362f6c930ef6a78f2c1b2215f8a7b398c06e1083f17af0"
 dependencies = [
- "libc",
+ "js-sys",
  "num-integer",
  "num-traits",
  "time 0.1.44",
+ "wasm-bindgen",
  "winapi 0.3.9",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,6 +93,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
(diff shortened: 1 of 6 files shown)


Cargo.lock
```diff
@@ -493,14 +493,15 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
...
 [[package]]
 name = "chrono"
-version = "0.4.19"
+version = "0.4.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "670ad68c9088c2a963aaa298cb369688cf3f9465ce5e2d4ca10e6e0098a1ce73"
+checksum = "6127248204b9aba09a362f6c930ef6a78f2c1b2215f8a7b398c06e1083f17af0"
 dependencies = [
- "libc",
+ "js-sys",
```
jcnelson, line 500: Why is `js-sys` a dependency on what should just be a wrapper around `localtime_r`?

jcnelson: Looks like `chronos` by default includes all the dependencies needed to compile to a wasm target.  Let's turn this off:

```
chrono = { version = "0.4.22", features = ["clock"] }
```

diwakergupta: Tried this and doesn't quite work as expected -- 0.4.22 still pulls in js-sys, iana-time-zone as well as android_system_properties.

diwakergupta:
I rebased on top of Aaron's commit to fix the build, and incorporated @jcnelson's suggestion. But looks like `iana-time-zone` also pulls in `js-sys`. Sadly I'm at the limit of my cargo-fu and couldn't figure out how to get rid of the wasm related dependencies. 

If this PR is unacceptable as-is -- someone wants to commandeer this PR, go for it, otherwise feel free to close it out for now.

jcnelson:
I looked into the CVE for this.  Basically, someone needs to be able to race the call to `localtime_r` with a call to `setenv` in a different thread of the process in order to create a dangling pointer which will lead to a segfault.

This is basically unexploitable for us -- we don't set envars on-the-fly in production, and we certainly don't set the `TZ` environment variable anywhere (which is ready by `localtime_r`).

I'm fine with just closing this PR, and maybe documenting this choice somewhere.

diwakergupta:
> I'm fine with just closing this PR, and maybe documenting this choice somewhere.

Thanks for looking into this @jcnelson , closing for now. If someone's motivated enough to address all audit flags and incorporate that into CI, go for it!

PR #3293, opened 2022-09-14
Ability to compile the clarity vm to wasm

lgalabru:


stacks-common/Cargo.toml
```diff
@@ -66,9 +67,3 @@ default = ["developer-mode"]
 developer-mode = []
 slog_json = ["slog-json"]
 testing = []
-
-[target.'cfg(all(any(target_arch = "x86_64", target_arch = "x86", target_arch = "aarch64"), not(target_env = "msvc")))'.dependencies]
-sha2 = { version = "0.10", features = ["asm"] }
```
jcnelson, line 71: We need this for performance.  The MARF is noticeably slower without it, last I checked.

zone117x: Might be worth checking if the `sha2` crate is smart enough now to ignore the `asm` feature on unsupported targets -- if so then `features = ["asm"]` can be used without messy the target cfg. It was still required earlier this year though.

lgalabru: Per the `sha2` documentation, developers should avoid enabling the `asm` feature in libraries (which stacks-common is). I re-introduced the `asm` feature at the  `stacks-node` level in https://github.com/stacks-network/stacks-blockchain/pull/3293/commits/7922ec0d91e0e1a623b6e11e9e0cd2c821bb3014

jcnelson:
N.B. we'll wait on reviewing and merging this until after 2.1.

lgalabru:
Closing this PR - to be reworked in https://github.com/stacks-network/stacks-blockchain/pull/3478.

Issue #3056, opened 2022-02-19
[Clarity] `concat` should allow to concatenate more than 2 sequences at once.


LNow:
**Is your feature request related to a problem? Please describe.**
I want to concatenate more than 2 sequences at once without using `fold` because it is to expensive (execution costs wise).

**Describe the solution you'd like**
`concat` function should take 2 or more sequences of the same type and returns a concatenated sequence of the same type, with the resulting sequence_len = sequence1_len + sequence2_len ...+ ... sequenceN_len.

Instead of this: 
```clarity
(concat
  (concat
    (concat 0x00 0x01)
    (concat 0x02 0x03
  )
  (concat
    (concat 0x04 0x05)
    (concat 0x05 0x07)
  )
)
;; or
(fold my-concat (list 0x00 0x01 0x02 0x03 0x04 0x05 0x06 0x07) 0x)
(define-private (my-concat (input (buff 1)) (output (buff 8)))
  (unwrap-panic (as-max-len? (concat output input) u8))
)
```
I'd like to be able to do this:
```clarity
(concat  0x00 0x01 0x02 0x03 0x04 0x05 0x06 0x07)
```

**Describe alternatives you've considered**
It is quite easy to concatenate multiple sequences using `fold`, but this approach is quite expensive.

jcnelson:
Hey @LNow,

> It is quite easy to concatenate multiple sequences using fold, but this approach is quite expensive.

What leads you to believe that `(concat 0x00 0x01 0x02 0x03 0x04 0x05 0x06 0x07)` will be significantly cheaper than the `fold` approach?

LNow:
@jcnelson the answer to your question is quite simple. Current execution costs.

Using example I posted above.

Multiple nested `concat`:
```
+----------------------+----------+------------+
|                      | Consumed | Limit      |
+----------------------+----------+------------+
| Runtime              | 8337     | 5000000000 |
+----------------------+----------+------------+
| Read count           | 3        | 7750       |
+----------------------+----------+------------+
| Read length (bytes)  | 221      | 100000000  |
+----------------------+----------+------------+
| Write count          | 0        | 7750       |
+----------------------+----------+------------+
| Write length (bytes) | 0        | 15000000   |
+----------------------+----------+------------+
```

Fold:
```
+----------------------+----------+------------+
|                      | Consumed | Limit      |
+----------------------+----------+------------+
| Runtime              | 20130    | 5000000000 |
+----------------------+----------+------------+
[...]

PR #3294, opened 2022-09-14
Introduces Clarity3 and Epoch 2.2 and adds a variadic concat function.

a3slade:
### Description
Adds Clarity3 and Epoch 2.2.
Adds variadic concat function for Clarity3.

### Applicable issues
https://github.com/stacks-network/stacks-blockchain/issues/3056

Questions/requests for reviewers:
* I tried to make sure I updated everywhere for Clarity3 / Epoch 2.2 but please double check that I did this right. I didn't know the context for most of the stuff that I copied from 2.1 so please check that it was either carried over or updated as appropriate.
* The costs for concat in 2.0 and 2.05 were not the same. I copied the method from 2.05. Is this right?
* I allowed for one or more arg. Is this what we want?
* Should costs/mod.rs add a COSTS_3_NAME or reuse "costs-2"?
* I attempted to reproduce many of the tests from earlier Epochs and Clarity versions but I probably did not find all of them. Are there more tests that are necessary?
* I'm still learning rust and the codebase so please give me feedback if there are better ways to do things!

### Follow-up
This function needs to be benchmarked. The timing should be linear wrt the number of items. The benchmarking should also check that concat([a,b,c,d]) is at least as fast as both fold(concat, [a,b,c,d]) and concat(concat(a,b),concat(c,d)) as discussed in issue #3056.


a3slade:
Needs to include this: https://github.com/stacks-network/stacks-blockchain/pull/3310

jcnelson:
Making @obycode the owner of this PR.

kantai:
I am going to close this PR for now -- if Clarity is going to add variadic concat at some point in the future, it can always be recovered from this branch, but it doesn't seem like this PR has a path forward in any of the nakamoto related planning for now.

PR #3296, opened 2022-09-19
implement replace-at

### LINKED ISSUE #3284: [Clarity] add `set-element-at`
While implementing auto-unlock in `pox-2`, @kantai discovered the need for a way to set an element in a list.  I think this is small enough we could squeeze it into the 2.1 release.

pavitthrap:
### Description
This PR implements replace-at. The function takes in a sequence, an index, and an element, and returns a new sequence in which the element is substituted in at the provided index. 

### Applicable issues
- fixes #3284

The name was modified due to discussion in the issue.


clarity/src/vm/analysis/type_checker/natives/sequences.rs
```diff
@@ -423,3 +423,28 @@ pub fn check_special_slice(
...
+        _ => return Err(CheckErrors::ExpectedSequence(seq_type).into()),
+    };
+
+    // Check index argument
+    checker.type_check_expects(&args[1], context, &TypeSignature::UIntType)?;
+    // Check element argument
+    let elem_type = checker.type_check(&args[2], context)?;
+    TypeSignature::least_supertype(&unit_seq, &elem_type)?;
+
+    Ok(seq_type)
```
kantai, line 449: I think you want to use the `least_supertype` result to create the new sequence type in the case of a list. In the case of Buffers and Strings, I think you don't want to use `least_supertype` at all, because that will "accept" a `(ascii-string 2)` as an element. This leads to the length issues only being caught at runtime (as you show in your tests).  That's not usually how other Clarity functions work (or, if they do, like `as-max-len?`, they return an option or result type for the check).

```suggestion
    let input_type = checker.type_check(&args[0], context)?;
    let seq_type = match input_type {
        TypeSignature::SequenceType(seq) => seq,
        _ => return Err(CheckErrors::ExpectedSequence(input_type).into()),
    };
    let unit_seq = seq_type.unit_type();

    // Check index argument
    checker.type_check_expects(&args[1], context, &TypeSignature::UIntType)?;
    // Check element argument
    match &seq_type {
        ListType(list_type) => {
            let replace_type = checker.type_check(&args[2], context)?;
            let new_unit_type = TypeSignature::least_supertype(&unit_seq, &replace_type)?;
            let out_type = TypeSignature::list_of(new_unit_type, list_type.get_max_len())?;
            Ok(out_type)
        }
        BufferType(_) | StringType(_) => {
            checker.type_check_expects(&args[2], context, &unit_seq)?;
            Ok(TypeSignature::SequenceType(seq_type))
        }
    }
```

That way, length mismatches will be caught for `buff` and `string` types, and lists will get the proper supertype. For example:

```
(replace-at (list 0x00 0x0000 0x0000) 1 0x000000)
```

Will have type `(list (buff 3) 3)`. Right now, I think that would incorrectly return `(list (buff 2) 3)`.

I think this would also require an update to the runtime implementation. Alternatively, if you think that `replace-at` should always preserve the input list's type, then the type checker can do away with the call to `least_supertype`, and instead you `type_check_expects(.., .., &unit_seq)` in all cases.

clarity/src/vm/functions/sequences.rs
```diff
@@ -383,3 +383,82 @@ pub fn special_slice(
...
+) -> Result<Value> {
+    check_argument_count(3, args)?;
+
+    let seq = eval(&args[0], env, context)?;
+    let seq_type = TypeSignature::type_of(&seq);
+    let (expected_elem_type, seq_is_list_type) =
+        if let TypeSignature::SequenceType(seq_subtype) = seq_type {
+            (seq_subtype.unit_type(), seq_subtype.is_list_type())
+        } else {
+            return Err(RuntimeErrorType::BadTypeConstruction.into());
```
kantai, line 400: This should be a `CheckErrors::ExpectedSequence`

clarity/src/vm/functions/sequences.rs
```diff
@@ -383,3 +383,82 @@ pub fn special_slice(
...
+        };
+    let index = eval(&args[1], env, context)?;
+    let new_element = eval(&args[2], env, context)?;
+
+    if expected_elem_type != TypeSignature::NoType && !expected_elem_type.admits(&new_element) {
+        return Err(RuntimeErrorType::BadTypeConstruction.into());
+    }
```
kantai, line 407: This check needs to change if `list` types should expand their type to fit the replaced-in element to something like:

```rust
if seq is list {
  TypeSignature::least_supertype(&expected_elem_type, &new_element_type)?;
} else {
  if !expected_elem_type.admits(&new_element_type) {
    return Err(CheckErrors::TypeValueError(expected_elem_type, new_element).into());
  }
}
```

 If lists aren't expanding at all, I think the check still needs to be updated to use the `CheckErrors` variant.

clarity/src/vm/functions/sequences.rs
```diff
@@ -383,3 +383,82 @@ pub fn special_slice(
...
+    match (seq, index) {
+        // todo: update the ClarityCostFunction once the Clarity2 related cost functions are implemented.
+        (Value::Sequence(seq), Value::UInt(index)) => {
+            let seq_len = seq.len();
+            runtime_cost(ClarityCostFunction::Unimplemented, env, seq_len as u64)?;
+
+            let index = match u32::try_from(index) {
+                Ok(index) => index as usize,
+                _ => return Ok(Value::err_uint(1)),
+            };
```
kantai, line 418: This implementation leads to type incoherence in the return type -- the type checker is currently saying that the return type of `replace-at SEQ uint V` is `SEQ`, but here it's `(err 1)`, which isn't a value of the type `SEQ`. Instead, the return type of `replace-at SEQ uint X` should be `(result SEQ int)`.

kantai: You can also see this in the automated doc tests (once you add the new-for-clarinet `snippet` field):

```
(replace-at (list (list 1) (list 2)) u0 (list 33)) ;; returns ((33) (2))
(replace-at (list 1 2) u3 4) ;; Returns (err u1)

thread 'vm::docs::test::test_examples' panicked at 'Type checker's expected type must admit result. Expected type = (list 2 int). Result = (err u1)', clarity/./src/vm/docs/mod.rs:2715:25
```

clarity/src/vm/analysis/type_checker/natives/sequences.rs
pavitthrap: I think I was thinking the latter approach makes the most sense (preserving the input list's type).

clarity/src/vm/analysis/type_checker/tests/mod.rs
```diff
@@ -1455,6 +1455,225 @@ fn test_slice_utf8() {
...
 }
 
+#[test]
+fn test_replace_at_list() {
+    let good = [
+        "(replace-at (list 2 3 4 5 6 7 8) u0 10)",
+        "(replace-at (list u0 u1 u2 u3 u4) u3 u10)",
+        "(replace-at (list true) u0 false)",
+        "(replace-at (list 2 3 4 5 6 7 8) u6 10)",
+        "(replace-at (list (list 1) (list 2)) u0 (list 33))",
```
kantai, line 1465: Can you also add the case:

```
(replace-at (list (list 1 2) (list 3 4)) u0 (list 0))
```

kantai (approved):
LGTM! Just had a comment about one more test case

clarity/src/vm/analysis/type_checker/tests/mod.rs
jcnelson: As well as this case?

```clarity
(replace-at (list (list 1 2 3)) u0 (list 0))
```

clarity/src/vm/docs/mod.rs
```diff
@@ -2189,6 +2189,28 @@ to deserialize the type, the method returns `none`.
...
+    input_type: "sequence_A, uint, A",
+    output_type: "(response sequence_A uint)",
+    snippet: "replace-at ${1:sequence} ${2:index} ${3:element}",
+    signature: "(replace-at sequence index element)",
+    description: "The `replace-at` function takes in a sequence, an index, and an element, 
+and returns a new sequence with the data at the index position replaced with the given element. 
+The given element's type must match the type of the sequence, and must correspond to a single 
+index of the input sequence. The return type on success is the same type as the input sequence. 
+
+If the provided index is out of bounds, this functions returns `(err u1)`.
```
jcnelson, line 2202: If there's no other way for this function to fail, then would it make sense to instead have `replace-at` return `(optional sequence_A)`?

clarity/src/vm/functions/sequences.rs
```diff
@@ -383,3 +383,88 @@ pub fn special_slice(
...
+    let (expected_elem_type, seq_is_list_type) =
+        if let TypeSignature::SequenceType(seq_subtype) = seq_type.clone() {
+            (seq_subtype.unit_type(), seq_subtype.is_list_type())
+        } else {
+            return Err(CheckErrors::ExpectedSequence(seq_type).into());
+        };
+    let index_val = eval(&args[1], env, context)?;
+    let new_element = eval(&args[2], env, context)?;
+
+    if expected_elem_type != TypeSignature::NoType && !expected_elem_type.admits(&new_element) {
```
jcnelson, line 405: How would this ever be `TypeSignature::NoType`?

clarity/src/vm/functions/sequences.rs
```diff
@@ -383,3 +383,88 @@ pub fn special_slice(
...
+    }
+
+    let index = if let Value::UInt(index) = index_val {
+        index
+    } else {
+        return Err(CheckErrors::TypeValueError(TypeSignature::UIntType, index_val).into());
+    };
+
+    match seq {
+        // todo: update the ClarityCostFunction once the Clarity2 related cost functions are implemented.
```
jcnelson, line 416: Can you open an issue for this and tag it as `stacks-2.1` so we don't forget it?

clarity/src/vm/functions/sequences.rs
```diff
@@ -383,3 +383,88 @@ pub fn special_slice(
...
+        Value::Sequence(seq) => {
+            let seq_len = seq.len();
+            runtime_cost(ClarityCostFunction::Unimplemented, env, seq_len as u64)?;
```
jcnelson, line 419: Why would the runtime cost be `O(n)` for an n-item sequence?  Can't this be `O(1)`?

clarity/src/vm/docs/mod.rs
pavitthrap: hmmm I think returning an error makes more sense - @kantai do you have an opinion?

clarity/src/vm/functions/sequences.rs
pavitthrap: If the input is `(list)`, it is NoType I am pretty sure. I ran into this error when I wrote tests.

clarity/src/vm/functions/sequences.rs
pavitthrap: I mirrored the cost of the concat function - this function also has to perform an append.

clarity/src/vm/docs/mod.rs
kantai: I don't have a strong opinion either way -- there's good arguments for either option:

In favor of `err`: public methods usually already use `uint` error types, so this lets them just use `try!` if they're kind of lazy.

In favor of `optional`: because there's no real error variation, `optional` matches the true type of the result more closely.

clarity/src/vm/functions/sequences.rs
jcnelson: `concat` performs `O(n)` work because it has to copy data.  This does not.

clarity/src/vm/functions/sequences.rs
```diff
@@ -383,3 +383,88 @@ pub fn special_slice(
...
+                Ok(index) => index as usize,
+                _ => return Ok(Value::err_uint(1)),
+            };
+
+            // check that the index is in bounds
+            if index >= seq_len {
+                return Ok(Value::err_uint(1));
+            }
+
+            let left_seq_value = seq.clone().slice(0, index)?;
```
a3slade, line 431: This line is probably O(n) as well. And perhaps the next one? (Do slices copy?)

clarity/src/vm/functions/sequences.rs
a3slade: Is this supposed to be in-place or return a new seq?

jcnelson: I'd strongly prefer that this method implement an in-place item replacement, because doing so would be much more efficient (and would save multiple heap allocations).  I misspoke earlier -- I meant to say that this code should not need to do a full sequence copy.

jcnelson (changes requested):
This PR overall looks fine, but the implementation really needs to be `O(1)` in-place replacement, instead of making a copy of the sequence being modified.

gregorycoppola:
I removed myself from reviewing this PR because @jcnelson is the second reviewer and @kantai has approved.

clarity/src/vm/docs/mod.rs
```diff
@@ -2204,6 +2204,28 @@ to deserialize the type, the method returns `none`.
 "#,
 };
 
+const REPLACE_AT: SpecialAPI = SpecialAPI {
+    input_type: "sequence_A, uint, A",
+    output_type: "(response sequence_A uint)",
```
jcnelson, line 2209: I see that the type-checker reports that this returns an optional type instead of a response type.  Can you update the docs accordingly?  Thanks!

clarity/src/vm/functions/sequences.rs
```diff
@@ -383,3 +383,50 @@ pub fn special_slice(
...
 }
+
+pub fn special_replace_at(
+    args: &[SymbolicExpression],
+    env: &mut Environment,
+    context: &LocalContext,
+) -> Result<Value> {
+    check_argument_count(3, args)?;
+
+    // todo: update the ClarityCostFunction once the Clarity2 related cost functions are implemented.
```
jcnelson, line 394: I think this can be removed :)

jcnelson (approved):
LGTM!  Thanks @pavitthrap!

a3slade:
The unit tests are breaking because the cost function wasn't implemented. I think this is supposed to go in costs-3.clar?

PR #3299, opened 2022-09-21
2.1: PoX-2 Fixes and Invariant Testing

kantai:
### Description

This PR adds explicit invariant checks to the `pox_2_tests` module. These checks surfaced a couple of places where the invariants were _not_ being properly maintained in the `pox-2` functions. Namely, the `-extend` methods weren't keeping track of `first-reward-cycle`, `lock-period`, and `reward-cycle-indexes` correctly.

This PR also adds some comments to `pox-2` which list the new invariants.

~~This PR is still a bit of a draft, because the invariant checks haven't been added to all the `pox_2_tests` tests yet.~~ Updated in 9ca085b to invoke the invariant checks on every block produced in the `pox_2_tests` cases.


kantai:
Basing this off of `feat/pox-2-increase` for readability reasons for now -- I merged the `feat/pox-2-increase` PR into the wrong branch, so it's not in `next` yet.

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -866,14 +889,22 @@
...
+                                                       (old-indexes (get reward-set-indexes old-state))
+                                                       (new-list (concat (default-to (list) (slice old-indexes cur-cycle-index (len old-indexes)))
+                                                                                   extended-reward-set-indexes)))
+                                            (unwrap-panic (as-max-len? new-list u12)))
+                                       extended-reward-set-indexes)))
           ;; update stacker record
           (map-set stacking-state
             { stacker: tx-sender }
-            { pox-addr: pox-addr,
+            (print { pox-addr: pox-addr,
```
jcnelson, line 904: Is the `print` going to be dropped before this merges?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -966,21 +997,34 @@
                     (pox-addr { version: (buff 1), hashbytes: (buff 20) })
                     (extend-count uint))
     (let ((stacker-info (stx-account stacker))
+          (stacker-state (get-stacker-info stacker))
           (amount-ustx (get locked stacker-info))
-          (unlock-height (get unlock-height stacker-info)))
+          (unlock-height (get unlock-height stacker-info))
+          (unlock-in-cycle (burn-height-to-reward-cycle unlock-height))
+          ;; if the account unlocks *during* this cycle (should only occur during testing),
```
jcnelson, line 1004: "[This] should only occur during testing" sounds like someone's famous last words :)  Why would it be the case that the account unlocks during this cycle?  Wouldn't the delegate be able to extend their lock in the same cycle in which they unlock?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -966,21 +997,34 @@
...
-     (let ((unlock-in-cycle (burn-height-to-reward-cycle unlock-height))
-          ;; if the account unlocks *during* this cycle (should only occur during testing),
-          ;; set first-extend-cycle to the next cycle.
-           (cur-cycle (current-pox-reward-cycle))
-           (first-extend-cycle (if (> (+ cur-cycle u1) unlock-in-cycle)
-                                      (+ cur-cycle u1) unlock-in-cycle))
-           (last-extend-cycle  (- (+ first-extend-cycle extend-count) u1))
-           (lock-period (- last-extend-cycle cur-cycle))
+     (let ((last-extend-cycle  (- (+ first-extend-cycle extend-count) u1))
+           ;; this seems incorrect
```
jcnelson, line 1025: :eyes:

src/chainstate/stacks/boot/pox-2.clar
kantai: It's if `v1_unlock_ht` is set to the middle of a cycle rather than a boundary. It could occur if that's how the 2.1 fork specifies that value, but even in that case, this code will handle the case smoothly by setting the `first-extended-cycle` to the next cycle.

src/chainstate/stacks/boot/pox-2.clar
kantai: That comment applies to the commented out calculation.

src/chainstate/stacks/boot/pox-2.clar
kantai: Yes

src/chainstate/stacks/boot/pox-2.clar
kantai: Remove this and the commented out calculation.

src/chainstate/stacks/boot/pox-2.clar
kantai: Dropped in [9ca085b](https://github.com/stacks-network/stacks-blockchain/pull/3299/commits/9ca085bccfb1973704d50f2fd58c5a26a16bcac2)

kantai:
Updated the `TestPeer::tenure_with_txs` so that it can be configured to invoke the PoX invariants check on every produced block, and configured all the tests in pox_2_tests to do so.

src/chainstate/stacks/boot/pox-2.clar
jcnelson: If we're certain that `v1_unlock_ht` will be set on a reward cycle boundary, then should there be a runtime assert that this is the case if we're on mainnet?

jcnelson (approved):
This looks great @kantai.  Thanks for doing this!

Just had one minor comment about a potential runtime assert we could add.

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -1377,6 +1606,152 @@ fn stack_increase() {
...
+    // alice locked, so balance should be 0
+    let alice_balance = get_balance(&mut peer, &key_to_stacks_addr(&alice).into());
+    assert_eq!(alice_balance, 0);
+
+    // produce blocks until alice's first reward cycle
+    for _i in 0..4 {
+        peer.tenure_with_txs(&[], &mut coinbase_nonce);
+    }
+
+    // produce blocks until immediately after the epoch switch (8 more blocks to block height 36)
```
pavitthrap, line 1725: s/8/4

pavitthrap (approved):
lgtm!

PR #3305, opened 2022-09-27
2.1 Fix: usable version checks for DBs and epochs

kantai:
The current db version <-> epoch checks in `next` do not work -- on restarting a node, the DB version cannot run in 2.1. If restarted while in "Epoch 1.0", that also will cause a panic.

This PR changes the version/epoch checks to make it possible to restart in 2.1, and to make Epoch 1.0 supported by all DB versions.


kantai:
One potential alternative to this would be to make the DB version an enum, and then turn these checks into greater-than-or-equal comparisons.

jcnelson (approved):
We need to make this sort of thing part of the release checklist :facepalm:

zone117x (approved):
Fix works in my testing environment

PR #3308, opened 2022-09-28
Remove unused function is-principal-eq from hyperchain contract

a3slade:
This function is not used anywhere. Similar functionality (but with a different return type) exists in is-miner.


diwakergupta:
@a3slade wrong repo?

PR #3316, opened 2022-10-02
feat: Improved miner iteration for Stacks main chain

gregorycoppola:
### Description
This PR imports https://github.com/hirosystems/stacks-subnets/pull/157 from the L2 work into the main chain.

It avoids repeated sorting or joining of the mempool. Details uploaded as comments.

I will make a clean branch from master before mailing a non-draft for review.

### Status
this project is being tracked at:
* https://www.notion.so/hiropbc/Checking-in-Miner-Changes-f3584ad03c0c4430b3da53f214e472cf

### Checklist
- [ ] Changelog is updated


src/core/mempool.rs
```diff
@@ -1223,6 +1289,14 @@ impl MemPoolDB {
         Ok(rows)
     }
 
+    // Return the results in a very "minimal", or "lazy", representation.
+    // Note: Delay deserialization until we know we want to process this.
+    pub fn get_all_txs_minimal(conn: &DBConn) -> Result<Vec<MemPoolTxMinimalInfo>, db_error> {
```
jcnelson, line 1294: This approach does not scale.  It will appear to work with "small" sample sizes, but the size of the mempool (and thus the size of this dataset) is _O(number-of-addresses)_, which only grows over time.  Also, you're sorting it in-RAM, which should be unnecessary if you just index the fields and structure your queries such that `ORDER BY` statements can use the index(es).

gregorycoppola: i have measurements in the log of how long this takes.

i reiterate that this processed a 100k mint (100k objects in mempool) 

1) it does scale
2) scaling in the size of the mempool is precisely what this was meant to achieve!

i will run a new version of the stress test and upload the full logs so you can see

jcnelson (changes requested):
A few things:

* We shouldn't read whole columns into RAM. Omitting the transaction data itself does not mean we avoid an I/O-intensive full table scan, and moreover, even a reading a single column entirely into RAM can open a DDoS vector because the maximum mempool row count is a function of the number of materialized Stacks addresses (which only grows over time).

* Why is `testnet/conf/stacks-krypton-miner.toml` over 50 KB?

* Please drop `run.sh`.  It appears to be a personal diagnostic tool you accidentally committed.

src/core/mempool.rs
jcnelson: No, @gregorycoppola, you are reading an unbounded amount of data into RAM that _is remotely controlled by attackers_.  Not only does that factually not scale (since RAM is bound), but also it's _dangerous_, and sets us up for failure down the road.

Maybe you can fixed-sized read batches of rows into RAM instead?  Then at least you're bounding the amount of state you're pulling into RAM, as well as bounding the amount of I/O each pass will do.

jcnelson:
> i will mege in the changes to mater to get the benefit of the db migration

You will do no such thing.

gregorycoppola:
@jcnelson i mean merge master into this PR :)

src/core/mempool.rs
gregorycoppola: we can sort in ram by fee rate and then take a limited amount into RAM

gregorycoppola:
@jcnelson I have attached the following logs re: your questions. I agree that bounding what is in memory, or sorting in db can all make sense. Still, here is some relevant data..

1) `read-and-sort-times.txt` shows how the `SELECT *` and then sort steps you asked about are scaling. To summarize, when the mempool is at 100k, the time taken to read is negligible (less than 200ms for both read and sort):

```
[relayer] Read and sorted mempool. Total size: 100002, time to read in minimal entries: 196.873291ms,  time to sort list: 2.301625ms
```

[read-and-sort-times.txt](https://github.com/stacks-network/stacks-blockchain/files/9692940/read-and-sort-times.txt)

2) `block-sizes.txt` shows the sizes of the blocks. The limiting factor here is how fast the transactions can *get in* to the mempool. There is only 60s between simulated bitcoin ticks, and 30s budget for the L1 miner. If it had been 600s seconds between simulated bitcoin blocks, each block would be full. Also, note that the block limit here is `HELIUM_BLOCK_LIMIT_20`, which has a read count limit of 50k, versus 15k on mainnet, that's why the first block seems overly full. It's also a minimal transaction.

```
INFO [1664724830.3000] 60303 /new_block: #txs=11733 txids=
INFO [1664724899.56000] 60303 /new_block: #txs=3577 txids=
INFO [1664724953.666000] 60303 /new_block: #txs=7679 txids=
INFO [1664725007.399000] 60303 /new_block: #txs=5523 txids=
INFO [1664725077.75000] 60303 /new_block: #txs=5958 txids=
INFO [1664725131.916000] 60303 /new_block: #txs=6447 txids=
INFO [1664725188.543000] 60303 /new_block: #txs=6745 txids=
INFO [1664725258.469000] 60303 /new_block: #txs=4366 txids=
INFO [1664725323.559000] 60303 /new_block: #txs=6323 txids=
INFO [1664725370.175000] 60303 /new_block: #txs=6033 txids=
INFO [1664725431.015000] 60303 /new_block: #txs=4730 txids=
INFO [1664725503.49000] 60303 /new_block: #txs=5524 txids=
INFO [1664725559.85000] 60303 /new_block: #txs=5835 txids=
INFO [1664725619.812000] 60303 /new_block: #txs=5135 txids=
INFO [1664725691.77000] 60303 /new_block: #txs=5151 txids=
INFO [1664725758.486000] 60303 /new_block: #txs=5600 txids=
INFO [1664725787.911000] 60303 /new_block: #txs=3658 txids=
```

[block-sizes.txt](https://github.com/stacks-network/stacks-blockchain/files/9692939/block-sizes.txt)


3) you can find the flow code at https://github.com/gregorycoppola/l2-profiling/blob/main/flows/l1_flow.js .. it might have some cruft but just start at `main()` if you want to see the profiling code

I can see if I can upload the whole log

gregorycoppola:
experiment mock miner is hitting the block limit before it hits the timeout (full blocks!)

```
WARN [1664759313.547478] [src/chainstate/stacks/db/transactions.rs:926] [relayer] Block compute budget exceeded: if included, this will invalidate a block, txid: bbdd10d22e6bda7cf32dc3e024d9bce3894bf4f38489548c825365e3048d3a19, cost: {"runtime": 108718004, "write_len": 72014, "write_cnt": 2088, "read_len": 48532593, "read_cnt": 15002}, budget: {"runtime": 5000000000, "write_len": 15000000, "write_cnt": 15000, "read_len": 100000000, "read_cnt": 15000}
WARN [1664759313.547679] [src/chainstate/stacks/miner.rs:1479] [relayer] Transaction bbdd10d22e6bda7cf32dc3e024d9bce3894bf4f38489548c825365e3048d3a19 consumed over 80% of block budget, marking as invalid; budget was {"runtime": 5000000000, "write_len": 15000000, "write_cnt": 15000, "read_len": 100000000, "read_cnt": 15000}
INFO [1664759317.441664] [src/core/mempool.rs:1308] [relayer] Mempool iteration finished, num_considered_txs: 385, num_included_txs: 385, total_processing_time_ms: 12810, total_effective_processing_time_ms: 12702
INFO [1664759319.581725] [src/core/mempool.rs:1308] [relayer] Mempool iteration finished, num_considered_txs: 51, num_included_txs: 50, total_processing_time_ms: 2140, total_effective_processing_time_ms: 2102
INFO [1664759319.673746] [src/core/mempool.rs:1308] [relayer] Mempool iteration finished, num_considered_txs: 20, num_included_txs: 19, total_processing_time_ms: 92, total_effective_processing_time_ms: 63
INFO [1664759319.718892] [src/core/mempool.rs:1308] [relayer] Mempool iteration finished, num_considered_txs: 4, num_included_txs: 3, total_processing_time_ms: 45, total_effective_processing_time_ms: 18
INFO [1664759319.743904] [src/core/mempool.rs:1308] [relayer] Mempool iteration finished, num_considered_txs: 1, num_included_txs: 0, total_processing_time_ms: 24, total_effective_processing_time_ms: 0
INFO [1664759321.610586] [src/chainstate/stacks/miner.rs:2399] [relayer] Miner: mined anchored block, block_hash: f8a8a2e09c3c68d5a95c4e585b155520ee50fd9d044962194c7247239eeacf6e, height: 78051, tx_count: 457, parent_stacks_block_hash: 4f206f08ad6b18c6de2974cd6b421a08778923108f12dc9c8e8bda03330bd078, parent_stacks_microblock: 0000000000000000000000000000000000000000000000000000000000000000, parent_stacks_microblock_seq: 0, block_size: 369891, execution_consumed: {"runtime": 187697893, "write_len": 433519, "write_cnt": 2371, "read_len": 24585545, "read_cnt": 11579}, assembly_time_ms: 16994, tx_fees_microstacks: 199005316

```

gregorycoppola:
image

Here is a graph of block sizes for 1) this PR, versus 2) the actual blocks on mainnet.

gregorycoppola:
Uploading a bigger snapshot of the logs
[block-size-sample-pr3316.txt](https://github.com/stacks-network/stacks-blockchain/files/9698308/block-size-sample-pr3316.txt)

gregorycoppola:
This is still a draft but setting to "ready for review" because CI tests aren't running.

src/core/mempool.rs
```diff
@@ -1157,95 +1348,95 @@ impl MemPoolDB {
...
+                Some(tx_event) => {
+                    match tx_event {
+                        TransactionEvent::Skipped(_) => {
+                            // don't push `Skipped` events to the observer
                         }
-                        None => {
-                            debug!("Mempool iteration early exit from iterator");
-                            break;
+                        _ => {
+                            total_included += 1;
```
gregorycoppola, line 1422: this is counting success, error, problematic... break into groups?

src/core/mempool.rs
```diff
@@ -535,6 +581,7 @@ const MEMPOOL_INDEXES: &'static [&'static str] = &[
 
 pub struct MemPoolDB {
     pub db: DBConn,
+    db2: Arc<Mutex<DBConn>>,
```
gregorycoppola, line 584: note: added Arc<Mutex> to extra DbConn.. I actually think this makes sense, because it is like a tiny (size 1) db connection pool.

i had to do this to pass a DBConn into the "transactions in the mempool" iterator, without borrowing the reference to `self=MemPoolDB`. this is because the class also has to be borrowed inside the loop as mutable.

rust doesn't let you borrow as both mutable and immutable at the same time.

jcnelson: The Rust-y thing to do here would be to make `db` into `Option<DBConn>`, and then add a method like this:

```rust
fn with_db<F, R>(&mut self, func: F) -> R
where
   F: FnOnce(&mut MemPoolDB, &mut DBConn) -> R
{
   let db = self.db.take().expect("BUG: db is already taken");
   let res = func(self, db);
   self.db = Some(db);
   res
}
```

Then, you put code that needs mutable references to `self.db` and `self` into `func` like so:

```rust
let result = self.with_db(|mempool, db| {
   // this compiles since mempool is &mut MemPollDB
   mempool.path = "foo".to_string();
   // this also compiles since db is &mut DBConn
   let mut tx = tx_begin_immediate(db)?;
   /* ... etc. ... */
});
```

We do this with `PeerNetwork` in a lot of places, for example, in order to get mutable access to both the `PeerNetwork` and one of its inner state machines.

`Arc<..>` is typically only used for sharing state between threads, which is not happening here.

gregorycoppola: If we want to say that MemPoolDB doesn't spawn, that would simplify things.

I think you solution would be incorrect if there was a spawned thread, but not mutex.

jcnelson: It sounds like you need to re-think the factoring of your test.  Putting a second database connection into the mempool DB struct just to get around a compile error in your PR indicates that you're probably doing something wrong.

gregorycoppola: actually.. I added a new TestPeer test in some branch, but it's not in this branch.

Thus, we can definitely say `TestPeer` is spawning already, and if that's a problem, we should unwind it.

PR #3318, opened 2022-10-04
Events/pox 2

### LINKED ISSUE #3289: [Stacks 2.1] Emit events for PoX operations and STX locked balance updates
This is a partial blocker for PoX-2 support in the API (and clients that depend on PoX related features in the API). 

Every operation that changes PoX state should emit an event that includes both the operation-specific data, and global state for the given account.

For example, with a `stack-extend`:
```js
{
  /* The current available balance, as string quoted micro-STX */
  balance: string,
  /* [PoX] The amount of locked STX, as string quoted micro-STX. Zero if no tokens are locked. */
  locked: string,
  /* [PoX] The burnchain block height of when the tokens unlock. Zero if no tokens are locked. */
  burnchain_unlock_height: number,
  /* [PoX] Details specific to the PoX operation */
  operation: {
    name: 'stacks-extend',
    args: {
      extend_count: number
    }
  }
}
```

These events can _almost_ all be emitted through `print` statements in the PoX-2 contract, except for:
1. The auto `stack-unlock` does not have an explicit contract-call function where we can include a `print` statement. 
2. It may be expensive (from a tx fee perspective) to lookup the info required to invoke these `print` statements. An event-emitter node already expects to spend more resources processing events -- but maybe unfair if users have to pay for regular miner nodes to generate the `print` data?

The API has two primary use cases for these events:

### Determining an account's locked STX balance

See https://github.com/hirosystems/stacks-blockchain-api/issues/1277

Right now, Stacks 2.0 emits a `stx_lock_event` event which the API can store and query to determine locked STX. In PoX-2, this event is no longer enough to determine that state in given the new features in PoX-2 (e.g. `stack-extend`, auto `stack-unlock`, etc).

### Parsing "PoX operations" in the Rosetta implementation

See https://github.com/hirosystems/stacks-blockchain-api/issues/1278

The Rosetta spec (used by at least a couple major exchanges) has a [`Data API`](https://www.rosetta-api.org/docs/data_api_introduction.html) which requires the API to generate "operations" for a given block. In order to create "PoX operations", the current implementation scans contract-call transactions for the PoX contract and the `stack-stx` function name, then manually parses Clarity args to generate a Stacking operation. This approach has major limitations: 
1. It's not aware of PoX operations that happen from contract-call or contract-deploy txs that call into the PoX contract -- the contract ID for tx will not be detected.
3. For the new PoX-2 operations, we need to expand the Rosetta parsing logic. Instead of adding more hardcoded function names to parse, it would be much better to start consuming events.

pavitthrap:
### Description
This PR does the following: 
- adds a synthetic transaction type such that the auto unlock events are captured within a transaction receipt 
- adds print events to the pox-2 contract for operations that affect a user's locked token state:
    - handle-unlock (this is the auto-unlock function called by the node) 
    - stack-stx 
    - stack-extend
    - stack-increase 
    - delegate-stack-stx 
    - delegate-stack-extend
    - delegate-stack-increase 

I also cleaned up the clarity docs file a little bit. 

I tested that the functions emit the expected print event in pox_2_tests.rs.

### Applicable issues
- fixes #3289


src/chainstate/stacks/db/blocks.rs
```diff
@@ -5280,21 +5279,37 @@ impl StacksChainState {
...
+                events,
+                post_condition_aborted: false,
+                result: Value::okay_true(),
+                stx_burned: 0,
+                contract_analysis: None,
+                execution_cost: ExecutionCost::zero(),
+                microblock_header: None,
+                tx_index: 0,
+                vm_error: None,
+            };
```
rafaelcr, line 5309: This looks great. Could you share an example of how this synthetic tx will look like in node messages, though? e.g. will it have the value of `0x00` for the `raw_tx` entry?

src/chainstate/stacks/events.rs
```diff
@@ -15,6 +16,7 @@ pub use clarity::vm::events::StacksTransactionEvent;
 pub enum TransactionOrigin {
     Stacks(StacksTransaction),
     Burn(Txid),
+    NetworkProtocol,
```
gregorycoppola, line 19: can you add comment explaining where this value comes from?

src/chainstate/stacks/boot/pox_2_tests.rs
```diff
@@ -145,6 +146,31 @@ pub fn check_all_stacker_link_invariants(
     }
 }
 
+fn check_pox_print_event(event: &StacksTransactionEvent, operation_name: &str) {
```
gregorycoppola, line 149: can you explain what this does in words?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -831,7 +887,16 @@
                     add-amount: increase-by })))
             (err ERR_STACKING_UNREACHABLE))
       ;; NOTE: stacking-state map is unchanged: it no longer tracks amount-stacked in PoX-2
-      (ok { stacker: tx-sender, total-locked: (+ amount-stacked increase-by)})))
+
+      ;; emit event
+      (print (merge event-info {
+            name: "stack-increase", 
+            args: {
+                increase-by: increase-by,
```
gregorycoppola, line 895: do these args need to be accounted for the test?

gregorycoppola (commented):
Overall LGTM.. but had a couple of questions..

a3slade:
From file: src/chainstate/stacks/boot/pox-2.clar 
Function: fold-unlock-reward-cycle 
Line 315: 
 ```
 (map-set stacking-state { stacker: moved-stacker }
                                   (merge moved-state { reward-set-indexes: update-list })))
```
This function does not have a print event. Does it need one?

File: src/chainstate/stacks/boot/pox_2_tests.rs (hunk 3 above)
Comment (a3slade) [inline src/chainstate/stacks/boot/pox_2_tests.rs:149]: If I'm understanding this correctly, this will check that all the basic elements of the event-info are there but it doesn't check the extra elements that are added for some of the different types. Do those need to be tested?

**Comment by a3slade** [comment]
If I followed things correctly, then `setup_block` returns its receipts to `SetupBlockResult` on line 5512 of `blocks.rs` in the `append_block` function. But I don't see that struct get used after that. Are the tx receipts within that struct supposed to be included in the return from that function?

File: src/chainstate/stacks/boot/pox-2.clar (hunk 4 above)
Comment (jcnelson) [reply src/chainstate/stacks/boot/pox-2.clar:895]: Yes -- I think that the unit test coverage for these new events should verify that the data returned is what we expect.  There's not going to be a way to change this code once 2.1 is live.

File: src/chainstate/stacks/boot/pox_2_tests.rs (hunk 3 above)
Comment (jcnelson) [reply src/chainstate/stacks/boot/pox_2_tests.rs:149]: Seconded. I think that this test should verify _all_ data returned by the `print` event.

File: src/chainstate/stacks/boot/pox_2_tests.rs [hunk 5]
```diff
@@ -145,6 +146,31 @@ pub fn check_all_stacker_link_invariants(
...
+                .expect_ascii(),
+            operation_name
+        );
+        assert!(outer_tuple.data_map.get("stacker").is_some());
+        assert!(outer_tuple.data_map.get("balance").is_some());
+        assert!(outer_tuple.data_map.get("locked").is_some());
+        assert!(outer_tuple
+            .data_map
+            .get("burnchain-unlock-height")
+            .is_some());
```
jcnelson, line 168: Can you also verify the values of the tuple?

src/chainstate/stacks/events.rs
jcnelson: It seems to refer to an event that originates from some internal bookkeeping, instead of from a burnchain or Stacks transaction.

src/chainstate/stacks/events.rs
```diff
@@ -28,12 +30,16 @@ impl TransactionOrigin {
...
-    pub fn serialize_to_vec(&self) -> Vec<u8> {
+    /// Serialize this origin type to a string that can be stored in
+    ///  a database
+    pub fn serialize_to_dbstring(&self) -> String {
         match self {
-            TransactionOrigin::Burn(txid) => txid.as_bytes().to_vec(),
-            TransactionOrigin::Stacks(tx) => tx.txid().as_bytes().to_vec(),
+            TransactionOrigin::Burn(txid) => format!("BTC({})", txid),
+            TransactionOrigin::Stacks(tx) => to_hex(&tx.serialize_to_vec()),
+            TransactionOrigin::NetworkProtocol => "ImplicitNetworkTransaction".into(),
```
jcnelson, line 42: Technically, this event isn't a transaction -- it doesn't originate from a transaction.  What if we just call this something like "bookkeeping" or "internals" without the word "transaction" so users don't get confused?

jcnelson (commented):
Thanks for sending this in @pavitthrap!

Before approving, I think we'll want to see some unit test coverage that verifies all of the `print` arguments' values.  But other than that it looks good; I only had minor feedback.

pavitthrap:
> From file: src/chainstate/stacks/boot/pox-2.clar Function: fold-unlock-reward-cycle Line 315:
> 
> ```
> (map-set stacking-state { stacker: moved-stacker }
>                                   (merge moved-state { reward-set-indexes: update-list })))
> ```
> 
> This function does not have a print event. Does it need one?

No, it does not need its own print event, since the function `fold-unlock-reward-cycle` is a helper used by `handle-unlock` (which does have a print event).

pavitthrap:
> If I followed things correctly, then `setup_block` returns its receipts to `SetupBlockResult` on line 5512 of `blocks.rs` in the `append_block` function. But I don't see that struct get used after that. Are the tx receipts within that struct supposed to be included in the return from that function?

Let's discuss this offline - this note is a little out of scope for this specific code change.

src/chainstate/stacks/db/blocks.rs
pavitthrap: Yes, the txid field will be equal to "NetworkProtocol", and and the raw_tx field will be set to "00".

pavitthrap:
> Does this function need a print event? I don't see it changing the stacking-state map but I don't know where that happens for these commits. `stack-aggregation-commit`

Hmm, it might need a print event. That function is called by a delegatee, and calls `add-pox-addr-to-ith-reward-cycle`, which does modify pox state. @rafaelcr do you have any thoughts here?

src/chainstate/stacks/db/blocks.rs
rafaelcr: On the API, we use the `tx_id` as a unique key in combination with the `index_block_hash` and `microblock_hash`... I'm afraid that if we use the same string for all of these transactions, conflicts would arise if two addresses have an unlock tx in the same block.

Is it possible for `tx_id` to be a (random?) hex value like all others? Or what would the implications be?

rafaelcr:
> Hmm, it might need a print event. That function is called by a delegatee, and calls add-pox-addr-to-ith-reward-cycle, which does modify pox state. @rafaelcr do you have any thoughts here?

Adding a print would be very helpful. The more info we can get on the other side, the better. @pavitthrap

pavitthrap:
I discussed this feature with @jcnelson, and we decided on the following course of action: 
- Make the function modifying pox state return the event data as a tuple 
- Create a synthetic print event in the clarity vm code, and add that to the transaction receipt associated with the pox function 

By doing it this way, we can easily add more information to the event later on if anything is missing - we are essentially making tracking pox events non consensus critical by making this switch.

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -831,7 +892,15 @@
                     add-amount: increase-by })))
             (err ERR_STACKING_UNREACHABLE))
       ;; NOTE: stacking-state map is unchanged: it no longer tracks amount-stacked in PoX-2
-      (ok { stacker: tx-sender, total-locked: (+ amount-stacked increase-by)})))
+
+      (ok (merge event-info {
+            name: "stack-increase",
```
zone117x, line 897: Can`pox-addr` be added to this payload? It appears to be the only function missing this other than `handle-unlock`

zone117x: Follow up question, should `pox-addr` be added to the "base" event data (i.e. `get-event-info`)?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -720,9 +770,18 @@
...
 
       ;; return the lock-up information, so the node can actually carry out the lock. 
-      (ok { stacker: stacker,
-            lock-amount: amount-ustx,
-            unlock-burn-height: unlock-burn-height })))
+      (ok (merge event-info {
+            name: "delegate-stack-stx", 
+            data: {
+                lock-amount: amount-ustx,
+                unlock-burn-height: unlock-burn-height,
```
zone117x, line 777: @pavitthrap can you confirm that the `unlock-burn-height` specified here and in a few others will always match the value of `burnchain-unlock-height` from `event-info` data in the same payload?

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -898,6 +898,7 @@
             data: {
                 increase-by: increase-by, 
                 total-locked: new-total-locked,
```
zone117x, line 900: The `event-info` here is "stale" by the time this function returns, for example:
```
  locked: 2062698000000000n,
  balance: 7937301999980000n,
  burnchain_unlock_height: 115n,
  name: 'stack-increase',
  data: { increase_by: 123n, total_locked: 2062698000000123n }
```
The `locked` value from `event-info` shows the value from before this function was called. The accurate value is in `total_locked` but these seems like inconsistent and surprising behavior.

zone117x (changes requested):
Thanks @pavitthrap! I've left some questions and requested some minor changes

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -1077,10 +1167,18 @@
...
-
+      
       ;; return the lock-up information, so the node can actually carry out the lock. 
-      (ok { stacker: stacker,
-            unlock-burn-height: new-unlock-ht }))))
+      (ok (merge event-info {
+            name: "delegate-stack-extend", 
+            data: {
+                pox-addr: pox-addr, 
+                unlock-burn-height: new-unlock-ht,
```
zone117x, line 1176: Similar to the previous comment, the `event-info. burnchain_unlock_height` is "stale" for the few functions that update it, e.g.:
```
balance: 7937301999969877n
burnchain_unlock_height: 115n
data: { extend_count: 1n, unlock_burn_height: 120n }
locked: 2062698000000123n
name: 'stack-extend'
stacker: 'ST11NJTTKGVT6D1HY4NJRVQWMQM7TVAR091EJ8P2Y'
```

The `data.unlock_burn_height` value is correct but `burnchain_unlock_height` is old.

src/chainstate/stacks/boot/pox-2.clar
pavitthrap: I think no for the latter question, since I get the value for `pox-addr` as a function parameter (it isn't always just a property that I can query with the stacker address)

src/chainstate/stacks/boot/mod.rs
```diff
@@ -270,6 +271,10 @@ impl StacksChainState {
                 )
                 .expect("FATAL: failed to handle PoX unlock");
 
+            // Add synthetic print event for `handle-unlock`, since it alters stacking state
```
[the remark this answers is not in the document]
zone117x: @pavitthrap I can confirm these events are coming through now, thanks for the fix!

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -327,18 +327,48 @@
                                  cycle-entry-total-ustx) })
             (ok (merge data { cycle: (+ u1 cycle)} )))))
 
+(define-private (get-event-info (stacker principal))
```
jcnelson, line 330: I think you're definitely going to want to run this step outside of the contract invocation.  This takes two `read_counts`, and in some cases it's used within a `fold`.

src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -327,18 +327,48 @@
...
+          (event-info (get-event-info user)))
+        ;; Check that the expected amount locked is equal to the actual amount locked
+        (asserts! (is-eq amount-locked (get locked event-info)) (err ERR_STACKING_CORRUPTED_STATE))
         ;; iterate over each reward set the user is a member of, and remove them from the sets. only apply to reward sets after cycle-to-unlock.
         (try! (fold fold-unlock-reward-cycle reward-set-indexes (ok { cycle: first-cycle-locked, first-unlocked-cycle: cycle-to-unlock, stacker: user })))
         ;; Now that we've cleaned up all the reward set entries for the user, delete the user's stacking-state
         (map-delete stacking-state { stacker: user })
-        (ok true)))
+
+        (ok (merge event-info {
```
jcnelson, line 363: Here and elsewhere, it looks like you're returning the _prior_ state of the user?  i.e. `event-info` represents the state of the user's balances and lock-ups at the _start_ of the function call?

If so, then I'm wondering if downstream clients need this information?  IIRC @zone117x said something about this being unnecessary in the blockchain call today.  Maybe instead they can return the _current_ state of the user, after it's been modified?

The reason I'm proposing this is that if we can do this, then we can move this entire return value out of `.pox-2` and instead generate it after the contract-call finishes in the special-case contract call handler.

src/chainstate/stacks/db/blocks.rs
```diff
@@ -5280,21 +5280,24 @@ impl StacksChainState {
         // epoch defined by this miner.
         clarity_tx.reset_cost(ExecutionCost::zero());
 
-        debug!("Evaluating block with epoch = {}", evaluated_epoch);
-        if evaluated_epoch >= StacksEpochId::Epoch21 {
+        // is this stacks block the first of a new epoch?
```
jcnelson, line 5283: Just curious why we're reversing the order of processing an epoch transition and the call to `Self::check_and_handle_reward_start`?

pavitthrap:
Discussed issue of stale data offline. 

Before, the approach was to gather data relating to the stacker (locked amount, unlock height etc.), and merge that data with the operation specific data. However, the stacker related data returned data before the mutation relating to the operation at hand was applied, leading to some confusion. This approach was taken to limit computation done in the pox-2 contract. 

Now, we will remove the creation of data to be printed for each pox mutating function from the pox-2 contract, and move it to the Rust code.

zone117x:
There's one remaining locked state mutation that occurs in the stacks-node but does not emit events to make the API aware. It occurs during the PoX-2 transition period described here: https://github.com/stacksgov/sips/blob/cfcb7a38797ce27004dba3d92f92bc82b3979132/sips/sip-015/sip-015-network-upgrade.md#overview
> Every account that is locked by the original contract for cycle N and beyond is unlocked at the end of Cycle N.

Right now, no event is emitted when this happens, so the API will incorrectly report locked STX from `pox-1` after this transition. Can events similar to `handle-unlock` be emitted for when this transition happens?

src/chainstate/stacks/db/blocks.rs
pavitthrap: I did this because I wanted the evaluated epoch before running check_and_handle_reward_start

donpdonp:
> Right now, no event is emitted when this happens, so the API will incorrectly report locked STX from `pox-1` after this transition. 

@zone117x assuming the api is using /v2/accounts/<stx address> the reported values for `balance(unlocked)` and `locked` use balance functions that treat the v1 locked amount as unlocked once the v1_unlock_height is reached. Is that enough to resolve this concern?

```
src/net/rpc.rs:1246
                    let unlocked = balance
                        .get_available_balance_at_burn_block(burn_block_height, v1_unlock_height);
                    let (locked, unlock_height) = balance
                        .get_locked_balance_at_burn_block(burn_block_height, v1_unlock_height);
```

zone117x:
@donpdonp no, the API builds its own view of the blockchain which is completely dependent on the accuracy and completeness of the event stream. The RPC endpoint does not help here

donpdonp:
@zone117x the last few commits adds `pox_v1_unlock_height` to `new_block`.

jcnelson:
Keep `.pox-2` from `next`.  Looks like there's nothing new in this branch's `.pox-2` contract.

donpdonp:
@jcnelson the `pox-2.clar` merge got complicated and needs a clarity dev to resolve. I marked `pox-2.clar` with TODO in 3 places.

1) line 757: existing function stacks-aggregation-commit return value changes from true to event-info. I tried to apply pavvithra's change but it failed in a way I couldn't diagnose. should be simple to apply for a clarity dev. this is currently causing test_delegate_extend_transition_pox_2 to fail (from pox_2_tests.rs)

2) line 772: new function stacks-aggregation-commit-indexed returns an int (reward-cycle). If the event emit mechanism is based on the return value, then there is a conflict here because the return value is already used for int where as the event mechanism wants to return an event-info.

3) line 846: new function stacks-aggregation-increase needs to return an event-info and add a test to pox_2_tests.rs

jcnelson:
Okay, I guess Pavitthra didn't implement my feedback.  One of y'all is going to need to do this.

`event-info` is _NOT_ supposed to be in `.pox-2`.  Instead, the special-case contract-contract handler is supposed to synthesize the event information and add it to the transaction receipt as an event.

The reason for this is because by putting the event generation into the contract, we can't change it later without a hard fork, or without doing the above work anyway.  Also, by putting it into the contract, we're billing users for the extra runtime even though it's unnecessary.

donpdonp:
adding a couple notes for the EDT timezone folks to chime in, or if a Hiro engineer can pick this up, otherwise I'll continue on it tomorrow(Tuesday).

I've got bits and pieces of an implementation from following check_pox_print_event() in pox_2_tests.rs. The goal is to add a StacksTransactionEvent::SmartContractEvent (with a key of "print") to the events[] array in a StacksTransactionReceipt. The place to do this is in handle_pox_v2_api_contract_call() in clarity_vm/special.rs. 

Its not clear how an entry in the transaction receipt results in a json message being sent to the API, or the connection between handle_pox_v2_api_contract_call and the events[] array of a particular transaction receipt. 

The event-data structs being removed from pox-2.clar used some clarity functions to gather information to populate the event and that will have to be reproduced in rust.

src/chainstate/stacks/boot/pox-2.clar
zone117x: Chiming in here to note that Jude is correct -- downstream clients need the current state rather than the prior state.

donpdonp:
> `event-info` is _NOT_ supposed to be in `.pox-2`. Instead, the special-case contract-contract handler is supposed to synthesize the event information and add it to the transaction receipt as an event.

Most functions for pox or pox-2 functions (with exceptions*) is to return fields that help the caller know what happened. The changes to pox-2.clar in this branch generally are to convert that tuple to an event-info structure, at the same time adding a field or two.

example: pox-2.clar in `events/pox-2` diffed with pox-2.clar in `next` contains:
```
611c642,652
<           (ok { stacker: tx-sender, lock-amount: amount-ustx, unlock-burn-height: (reward-cycle-to-burn-height (+ first-reward-cycle lock-period)) }))))
---
>           (ok (merge event-info {
>                 name: "stack-stx", 
>                 data: {
>                     lock-amount: amount-ustx,
>                     unlock-burn-height: (reward-cycle-to-burn-height (+ first-reward-cycle lock-period)),
>                     pox-addr: pox-addr, 
>                     start-burn-height: start-burn-ht, 
>                     lock-period: lock-period,
>                 }
>               })
>           ))))
```

So to remove any reference of event-info out of pox-2.clar, is a matter of restoring the return values to a clarity tuple. special.rs is already modified in this branch to stick the un-modified return value into a print event. This was an ah-ha moment for me because it means no rust changes need to happen, just make sure the struct returned from the pox-2 call provides what the API needs to see in the print event.

[*] There are exceptions, notably `stack-stx` simply returns 'true'. This would be changed to a more informational tuple. @jcnelson  Is it okay to change the return value? I assume so because this PR already changed true to an event-info.

The other exception is stack-aggregation-commit which returns 'true', that would have to change to `{pox-addr: pox-addr, pox-addr-index: pox-addr-index, reward-cycle: reward-cycle, amount-ustx: amount-ustx}` The question here is if the return value is changing anyways, can `pox-addr-index` be added as well, eliminating the need for `stack-aggregation-commit-index` (and also inner-stack-aggregation-commit)

donpdonp:
The approach previously described was no different than using event-info but with a couple fewer steps. To actually remove event-info will need to implementation the constructor below in rust. 

```
(define-private (get-event-info (stacker principal))
    (let (
        (stacker-info (stx-account stacker))
        (total-balance (stx-get-balance stacker))
    )
        {
            ;; The principal of the stacker 
            stacker: stacker,
            ;; The current available balance, as string quoted micro-STX */
            balance: total-balance,
            ;; [PoX] The amount of locked STX, as string quoted micro-STX. Zero if no tokens are locked. */
            locked: (get locked stacker-info),
            ;; [PoX] The burnchain block height of when the tokens unlock. Zero if no tokens are locked. */
            burnchain-unlock-height: (get unlock-height stacker-info),
        }
    )
)
```

and also calls such as the extra fields in stack-stx that are added to event-info
```
          (ok (merge event-info {
                name: "stack-stx", 
                data: {
                    lock-amount: amount-ustx,
                    unlock-burn-height: (reward-cycle-to-burn-height (+ first-reward-cycle lock-period)),
                    pox-addr: pox-addr, 
                    start-burn-height: start-burn-ht, 
                    lock-period: lock-period,
                }
              })
```

is there an efficient way in rust to make these calls? 
* stx-account
* stx-get-balance
* get locked
* get unlock-heigh
* reward-cycle-to-burn-height
* and probably others

jcnelson:
I think in general you _don't_ want `event-info`.  It's much, much, much safer to generate the events in Rust code instead of `pox-2`, because once the system goes live, _you can't change `pox-2`_.  But, you _can_ change the Rust code, and by extension, change the data that gets returned.

This is because `.pox-2` is consensus-critical -- the chain history is built upon the hash of the contract data, so it's not possible to release a new version of the node with a different `pox-2` without invalidating all of the Stacks 2.1 blocks in the process.  The Rust code that queries the state of the `.pox-2` contract after the contract runs, however, is _not_ consensus-critical.  Nor are transaction events or receipts consensus-critical.

I think `pox-2` should stay as-is in `next` at this point, especially since we're past code-freeze and other ecosystem folks will be wanting to test with it ASAP.  Instead, the special-case contract-call handler could inspect the return value from each of the Stacking methods, and if they succeeded, it can query the Clarity database for all the information you would have gotten from `event-info` and return that as a transaction event.  Then, if we wanted to change the event structure later, we could safely do so without a hard fork.

donpdonp:
> This is because `.pox-2` is consensus-critical -- the chain history is built upon the hash of the contract data, so it's not possible to release a new version

Thanks for the detail and background info. I see why pox-2 is very difficult to change once released.

> > is there an efficient way in rust to make these calls?
> 
> Yes, absolutely. You can just have the VM evaluate the `event-info` function body as a Clarity snippet once the contract-call finishes. I can help you with this tomorrow.

Here's a sketch of the next bit to write:

Here is the available data in the special case handler
```
pub fn handle_contract_call_special_cases(
    sender: Option<&PrincipalData>,
    _sponsor: Option<&PrincipalData>,
    contract_id: &QualifiedContractIdentifier,
    function_name: &str,
    result: &Value,
```
I imagine the caller will need to pass a clarity::vm to this function, too. If there is an existing spot in the code that calls vm.eval() that I can use for reference, that'd help.

Assuming that eval is available inside handle_pox_v2_api_contract_call, then the next step would be (pseudocode):
```
fn handle_pox_v2_api_contract_call() {
    if ( function-name == "stack-stx") {
        event_info_value = vm.eval("
               (define-private (get-event-info (stacker principal))
                   ....common fields of stacker,balanced,locked,burnchain_unlock_height
               (merge (get-event-info { 
                                extra_data_specific to stacks-stx: (get-extra-data)
                           }) ")
    } else if (function-name == "delegate-stack-stx") {
        event_info_value = ...
    }

    let print_event = Environment::construct_print_transaction_event(contract_id, &event_info_value);
    receipt.events.push(print_event);
}
```
Let me know if that needs any corrections or a different direction.

donpdonp:
pushed partial progress #9b8f467 for inspection. A handful of functions exist only to return a string full of clarity code. It might make more sense to create an event-info.clar with matching function signatures of pox-2.clar, then re-run the function name (with args) on event-info.clar and use the return value in the print event.

(not sure how best to revert pox-2.clar to `next`. locally I have the version of `next` in the working tree)

jcnelson:
That's the gist of it.  I'll commit the equivalent of `vm.eval()` for you.  Stand by.

donpdonp:
> That's the gist of it. I'll commit the equivalent of `vm.eval()` for you. Stand by.

I think our github comments crossed by each other (posted within 1 min of each other), the eval in use is in src/clarity_vm/special.rs:327 

The current challenge is getting the clarity to execute without errors, and whether or not that much copy/paste clarity in a .rs file makes sense.

the current clarity error comes from running `$ RUST_BACKTRACE=1 cargo test --workspace pox_2_tests::test_delegate_extend_transition_pox_2` and reports `FATAL: no 'data': Unchecked(NoSuchTupleField("data", TupleTypeSignature { "lock-amount": uint, "stacker": principal, "unlock-burn-height": uint,}))', src/clarity_vm/special.rs:66:18`

jcnelson:
@donpdonp See the current code for `src/clarity_vm/special.rs`.  It now synthesizes the `event-info` tuple, and patches in a method-specific `data` tuple as expected by the unit tests.  It does this by using the new function `GlobalContext::special_execute_read_only()`, which lets you run a Clarity snippet in a read-only fashion off of a bare `GlobalContext` (instead of within an `OwnedEnvironment`, as is more common).

I have not finished implementing all of this, but I have implemented an example of how to synthesize the `data` tuple for `stack-stx`.  Hoping this is enough to unblock you in finishing off this PR!

Note that `pox-2.clar` has been reverted to what's in `next`.  Let's keep it that way -- you shouldn't need to modify it to synthesize the event data.

jcnelson:
@donpdonp How about I finish this PR up, and you can review it?

donpdonp (approved):
Reviewed all code changes and unit tests. Emits the same messages as before without relying on the pox-2 return value. The string substitution to build the clarity code brings up a sql-injection concern but there are multiple safeties. a) the print-event eval is in read_only mode b) the sender is the same (no new privs) c) the params have to come from a successful pox-2 call. Looks good to me.

jcnelson:
@zone117x Can you test this out when you get a chance?

jcnelson:
> The string substitution to build the clarity code brings up a sql-injection concern but there are multiple safeties

One thing I'll do in this PR before merging it is make all of the event generation fallible.  There's no reason a bug in the event-generation code should be cause for bringing down the whole network, since it's not consensus-critical.

zone117x (commented):
Thanks @jcnelson, so far the changes are working _mostly_ as expected, but with a few issues:

Here's some example event payloads I've received during integration tests:
```
# stack-stx
(ok (tuple 
  (balance u9999999999990000) 
  (burnchain-unlock-height u0) 
  (data (tuple 
    (lock-amount u937590000000000) 
    (lock-period u1) 
    (pox-addr (tuple (hashbytes 0x31ef5ee9a226a792b93f2bfbfbc54f523eba7818) (version 0x00))) 
    (start-burn-height u106) 
    (unlock-burn-height u115))) 
  (locked u0) 
  (name "stack-stx") 
  (stacker 'STRYYQQ9M8KAF4NS7WNZQYY59X93XEKR31JP64CP)))

# stack-increase
(ok (tuple 
  (balance u7937301999980000) 
  (burnchain-unlock-height u155) 
  (data (tuple 
    (increase-by u123) 
    (pox-addr (tuple (hashbytes 0x06afd46bcdfd22ef94ac122aa11f241244a37ecc) (version 0x00))) 
    (total-locked u2062698000000123))) 
  (locked u2062698000000000) 
  (name "stack-increase") 
  (stacker 'ST11NJTTKGVT6D1HY4NJRVQWMQM7TVAR091EJ8P2Y)))

# stack-extend
(ok (tuple 
  (balance u7937301999969877) 
  (burnchain-unlock-height u155) 
  (data (tuple 
    (extend-count u1) 
    (pox-addr (tuple (hashbytes 0x06afd46bcdfd22ef94ac122aa11f241244a37ecc) (version 0x00))) 
    (unlock-burn-height u160))) 
  (locked u2062698000000123) 
  (name "stack-extend") 
  (stacker 'ST11NJTTKGVT6D1HY4NJRVQWMQM7TVAR091EJ8P2Y)))

# handle-unlock
(ok (tuple 
  (balance u9062409999990000) 
  (burnchain-unlock-height u112) 
  (data (tuple 
    (first-cycle-locked u22) 
    (first-unlocked-cycle u22) 
    (pox-addr (tuple (hashbytes 0x31ef5ee9a226a792b93f2bfbfbc54f523eba7818) (version 0x00))))) 
  (locked u937590000000000) 
  (name "handle-unlock") 
  (stacker 'STRYYQQ9M8KAF4NS7WNZQYY59X93XEKR31JP64CP)))

# delegate-stack-stx
(ok (tuple 
  (balance u2062697999990000) 
  (burnchain-unlock-height u0) 
  (data (tuple 
    (delegator 'ST1Z7V02CJRY3G5R2RDG7SFAZA8VGH0Y44NC2NAJN) 
    (lock-amount u2062697999978000) 
    (lock-period u1) 
    (pox-addr (tuple (hashbytes 0xbd0393d1958f15c269649647d2e4607d9dccfc2c) (version 0x00))) 
    (stacker 'ST2YG74YHJP7HBGK9CJB4FMQ4C1YSVK7W5K0DWKHV) 
    (start-burn-height u122) 
    (unlock-burn-height u130))) 
  (locked u0) 
  (name "delegate-stack-stx") 
  (stacker 'ST2YG74YHJP7HBGK9CJB4FMQ4C1YSVK7W5K0DWKHV)))

# delegate-stack-increase
(ok (tuple 
  (balance u12000) 
  (burnchain-unlock-height u130) 
  (data (tuple 
    (delegator 'ST1Z7V02CJRY3G5R2RDG7SFAZA8VGH0Y44NC2NAJN) 
    (increase-by u2000) 
    (pox-addr (tuple (hashbytes 0xbd0393d1958f15c269649647d2e4607d9dccfc2c) (version 0x00))) 
    (stacker 'ST2YG74YHJP7HBGK9CJB4FMQ4C1YSVK7W5K0DWKHV) 
    (total-locked u2062697999980000))) 
  (locked u2062697999978000) 
  (name "delegate-stack-increase") 
  (stacker 'ST2YG74YHJP7HBGK9CJB4FMQ4C1YSVK7W5K0DWKHV)))

# delegate-stack-extend
(ok (tuple 
  (balance u10000) 
  (burnchain-unlock-height u130) 
  (data (tuple 
    (delegator 'ST1Z7V02CJRY3G5R2RDG7SFAZA8VGH0Y44NC2NAJN) 
    (extend-count u1) 
    (pox-addr (tuple (hashbytes 0xbd0393d1958f15c269649647d2e4607d9dccfc2c) (version 0x00))) 
    (stacker 'ST2YG74YHJP7HBGK9CJB4FMQ4C1YSVK7W5K0DWKHV) 
    (unlock-burn-height u135))) 
  (locked u2062697999980000) 
  (name "delegate-stack-extend") 
  (stacker 'ST2YG74YHJP7HBGK9CJB4FMQ4C1YSVK7W5K0DWKHV)))

# stack-aggregation-commit
(ok (tuple 
  (balance u99960000) 
  (burnchain-unlock-height u0) 
  (data (tuple 
    (amount-ustx u2062697999980000) 
    (delegator 'ST1Z7V02CJRY3G5R2RDG7SFAZA8VGH0Y44NC2NAJN) 
    (pox-addr (tuple (hashbytes 0xbd0393d1958f15c269649647d2e4607d9dccfc2c) (version 0x00))) 
    (reward-cycle u26))) 
  (locked u0) 
  (name "stack-aggregation-commit") 
  (stacker 'ST1Z7V02CJRY3G5R2RDG7SFAZA8VGH0Y44NC2NAJN)))
```

Many of these seem to still use the state _prior_ to the execution of the event. A simple example is the `stack-stx` event: it shows `burnchain-unlock-height: 0` and `locked: 0`.

jcnelson:
> Many of these seem to still use the state prior to the execution of the event. A simple example is the stack-stx event: it shows burnchain-unlock-height: 0 and locked: 0.

Yes; this is consistent with the unit tests @pavitthrap wrote.  I have not altered any of her tests.

My understanding was that the event info tuple was meant to reflect the state of the PoX contract _after_ the stacking operation succeeded, but _before_ the lock was actually applied.  Is this not the desired behavior?

zone117x:
> My understanding was that the event info tuple was meant to reflect the state of the PoX contract _after_ the stacking operation succeeded, but _before_ the lock was actually applied. The inner `data` tuple reflects the lock state _after_ the lock is applied. Is this not the desired behavior?

The API is using a temporary `should_patch_pox2_events: boolean` flag that when set, enables a bit of logic that is specific to each pox2-event type in order to normalize several of these top-level properties before storing them, i.e. `balance`, `locked`, and `burnchain_unlock_height`. The inner `data` tuple values are applied to those properties to make them reflect the latest / current state. This avoids some complexity and slowness in database read queries later on.

So far the `stack-stx` event, I'm using code like:
```ts
if (PATCH_EVENT_BALANCES && event.name === 'stack-stx') {
  event.burnchain_unlock_height = event.data.unlock_burn_height;
  event.balance -= event.data.lock_amount;
  event.locked = event.data.lock_amount;
}
```

I _suspect_ most event consumers would probably do something similar, so imo the events should be emitted with those top-level properties already reflecting the latest state. But I don't feel strongly, and the current code works -- we can leave as is if you feel strongly.

zone117x:
Status update from the API side: 

It's very close to being ready. So far all events that mutate locked state are processed and being used query account `balance` and `locked` state correctly, with integration tests for each of the different pox-2 operations. 

The API is using the new `pox_v1_unlock_height` event property to correctly ignore locked state from pox-v1 when querying for account balance for a given height.

A few things that aren't yet finished:

* The `handle-unlock` event _may_ have an off-by-one `burnchain_unlock_height`, at least in comparison to the rest of the event types. Or likely I'm just not querying it properly. I'll have it figured out either way soon.

* The API is not yet returning the "committed" STX amount which should now be possible to calculate using the new `stack-aggregation-*` events. However, this would be a new feature and doesn't have to make it in time for 2.1 launch. 

* The Rosetta layer is not yet generating the synthetic unlock operations for the block associated with `pox_v1_unlock_height`. This one is tricky, because for each account, I need to infer which of the pox events are from pox-v1 vs pox-v2. Right now it requires first querying state from the existing `stx_lock_event` set, but that contains events from both pox-v1 and pox-v2, so the pox-v2 events need to be excluded by querying the new `pox2_event` set. 
  * I think this could be made easier if there was a change to the `stx_lock_event` payload so that it included a property indicating if it was from pox-v1 vs pox-v2. Thoughts?

jcnelson:
Hey @zone117x, I'd like to have this merged before the holiday.

> The API is not yet returning the "committed" STX amount which should now be possible to calculate using the new stack-aggregation-* events. However, this would be a new feature and doesn't have to make it in time for 2.1 launch.

Is there anything you need to see in this PR that is needed to enable this?

> I think this could be made easier if there was a change to the stx_lock_event payload so that it included a property indicating if it was from pox-v1 vs pox-v2. Thoughts?

I think this is done already in #3415?

zone117x (approved):
The API now fully consumes these events in PR https://github.com/hirosystems/stacks-blockchain-api/pull/1403, and everything is working as expected in extensive new integration tests. 

There's the remaining oddity of the event payloads requiring normalization on the event consumer side as discussed earlier, but that's not a blocker.

Issue #3229, opened 2022-08-01
Optimize mempool walk to speed up miner loop by ~6x


gregorycoppola:
**Is your feature request related to a problem? Please describe.**
From profiling experiments, it looks like there is an opportunity to save a lot of time in the inner miner running loop.

In general, we find that over 85% of the time spent in the miner loop is spent picking the next transaction, in `MemPoolDB::get_next_tx_to_consider`, and only around 15% of time is spent actually processing transactions. This suggests that, if this time could be reduced near zero, the speed of the inner loop could be increased by around 6x.

For example, here is an example line from a mock-miner running on GCP with the special logging PR https://github.com/stacks-network/stacks-blockchain/pull/3224:

```
INFO [1659067157.220537] [src/core/mempool.rs:1148] [relayer] total_inside_time: 2.143321953s total_outside_time: 16.914410382s total_consider_next_time: 15.77858194s total_update_nonce_time: 1.062061107s total_bump_nonce_time: 71.12287ms
```

We see that, for `2.1s` processing transactions, `16.9s` were spent picking the next transaction to process, with `15.8s` spent in `consider_next_transaction` and another `1.1s` updating nonces.

I uploaded a relevant sample of the logs from one mock miner. [depict-outside-time2.txt](https://github.com/stacks-network/stacks-blockchain/files/9234525/depict-outside-time2.txt)

Statistics computed from these logs show that, for loops where the miner loop takes over `10s`, over 86% of the time is spent outside of transaction processing.

These results have been replicated using out-of-the-box settings on: 1) a Dell lap top, 2) a GCP machine, 3) a AWS machine.

**Describe the solution you'd like**
Create some optimized MemPoolDB data structure that bring to near 0 the cost of picking a new transaction.

Also, this structure should efficiently solve:
* https://github.com/stacks-network/stacks-blockchain/issues/3188
* https://github.com/stacks-network/stacks-blockchain/issues/3193

**Describe alternatives you've considered**
#3225 shows that we can actually realize the gains described by "pre-caching" the mempool walk. However, this pre-caching itself takes time, and can become invalid https://discord.com/channels/621759717756370964/730814571517968436/1002698396928311378

**Additional context**
* https://github.com/stacks-network/stacks-blockchain/issues/3188
* https://github.com/stacks-network/stacks-blockchain/issues/3193

jcnelson:
Per @igorsyl:

[...]

Issue #3193, opened 2022-07-11
Verify Mempool Candidate Iteration Algorithm


gregorycoppola:
**Is your feature request related to a problem? Please describe.**
This issue asks for better ability to verify that the mempool walk algorithm is correct.

It is motivated originally by the fact that I ran the `stacks-inspect try-mine` tool on the chain state saved at (https://storage.googleapis.com/hiro-public/mainnet.tar.gz). I got a block with only 2 transactions from 44 candidates: 

```
DEBG [1657539187.515011] [src/chainstate/stacks/miner.rs:2111] [main] Anchored block transaction selection finished (child of 05eeea820cb6283bfe5cdc34db756236eb0f47610a246fc995808a5641f3d7f4): 1 transactions selected (44 considered)
DEBG [1657539188.796492] [src/chainstate/stacks/miner.rs:2151] [main] Miner: mined anchored block, block_hash: 0ef5706030aeb4dbd12f747d556662a8a6f6a88103addf599fc155ec89cae34d, height: 65116, tx_count: 2, parent_stacks_block_hash: 05eeea820cb6283bfe5cdc34db756236eb0f47610a246fc995808a5641f3d7f4, parent_stacks_microblock: e274103c8d343cdc6f3e9dc06e91af2746495e47e1f9af08dc2da267efab577b, parent_stacks_microblock_seq: 3, block_size: 575, execution_consumed: {"runtime": 1326849, "write_len": 689, "write_cnt": 9, "read_len": 441018, "read_cnt": 79}, assembly_time_ms: 42219, tx_fees_microstacks: 180
```
This led to me to wonder if there might be a bug in the candidate selection algorithm. I don't know if there is. Maybe there is or maybe this was the size of the mempool at the relevant time. I haven't had a chance to cross-reference what the mempool might have been then.

However, this request is based on the experience I had trying to figure out if the mempool walk might have a bug.

Currently, the candidate selection algorithm is coupled with the processing of transactions. That is, they can't be separated. Processing a transaction is relatively slow (e.g., 1 transaction per second). Thus, most times there is not time to walk the entire mempool, while processing transactions along the way:
https://github.com/stacks-network/stacks-blockchain/blob/08c163bab484954b3934515ee7bf60848f52ed4a/src/core/mempool.rs#L1013

Thus:
* at no point is the entire mempool actually traversed in its entirety by anyone (either miner or follower)
* we cannot separate out the traversal of the mempool into a separate object to analyze (#3188)

[...]

PR #3326, opened 2022-10-06
feat: Improved Mempool Iteration for Miner Loop

### LINKED ISSUE #3313: Improve miner transaction selection process
@gregorycoppola has found that when the mempool has too many transactions, the miner code will spend so long deciding what are the best transactions to pick, that it runs out of time to actually execute them. Currently, we are seeing a very full mempool and blocks that are not very full, which looks like a symptom of this situation. Let's use this issue to track ideas for quick wins.

gregorycoppola:
### Description
This algorithm implements the mempool walk to consider candidates in time `O(N)` in the size of the mempool. In addition, it only does `O(M)` deserializations, where `M` is the number of transactions that 1) meet the criteria, and 2) can be processed in the available time.

This algorithm reads in from DB only pages of fixed size, and does not rely on caching transactions in RAM.

This replaces an algorithm which in these terms would be at least `O(M * NlogN)`. We want the mempool size (i.e., `N`) to be able to get quite large, and the baseline system can degrade at around 10k-100k transactions.

### Algorithm
The algorithm proceeds as:
* select a very "thin" (minimal) representation of each transaction in the entire mempool
* sort this list by fee rate
* for each `txid` in the sorted order:
  * check the nonce appropriateness, and `continue` if not appropriate
  * fetch and deserialize the associated `tx`
  * process this `tx`

### Performance
Much data can be found on #3316 

The following benchmarks were provided by @obycode :
image

This PR corresponds to the line `draft/fast-iteration--simple-limits`. We see from this mempool from Oct 1, 2022, that this system can fill a block `99.97%` in `12.6` seconds, with the ability to be reduced to `11.4` seconds, by merging some form of #3321. This is compared to current `master`, which has the block `5%` full after `31.3` seconds.

### Applicable issues
- fixes #3313 #3229 #3193

* there is currently the addition of a new field `connection_pool` to `MemPoolDB` that needs discussion
* one test is deleted because the behavior is not guaranteed: `test_build_anchored_blocks_incrementing_nonces`
* one test is proposed to be disabled, with an issue filed to reenable by adding forking to the `TestPeer` framework: `mempool_walk_over_fork`

### Checklist
- [x] Changelog is updated


src/core/mempool.rs
```diff
@@ -241,6 +245,24 @@ pub struct MemPoolTxInfo {
...
+pub struct MemPoolTxMinimalInfo {
+    pub txid: Txid,
+    pub fee_rate: Option<f64>,
+    pub origin_address: StacksAddress,
+    pub origin_nonce: u64,
+    pub sponsor_address: StacksAddress,
+    pub sponsor_nonce: u64,
+}
+
+#[derive(Debug, PartialEq, Clone)]
```
gregorycoppola, line 260: Note: I imported these changes into `develop`... wdyt?

src/core/mempool.rs
```diff
@@ -476,6 +574,9 @@ const MEMPOOL_INDEXES: &'static [&'static str] = &[
 
 pub struct MemPoolDB {
     pub db: DBConn,
+    /// A pool (of size 1) of Mutex guarded database connections. This is a hack until we decide
+    /// the most "rust"-consistent way of passing a db connection to each mempool Iterator.
+    db_connection_pool: Arc<Mutex<DBConn>>,
```
gregorycoppola, line 579: Note: This is a discussion point, I'm sure!

gregorycoppola:
Note: this is currently failing when we try to run `stacks-inspect try-mine`. I think the problem is the original changes were based on `master`, and I cherry picked the changes into `develop`, and maybe something weird is happening.

I wonder if we should merge `master` -> `develop` before we try merging this.

src/core/mempool.rs
```diff
@@ -1064,6 +1173,111 @@ impl MemPoolDB {
         Ok(updated)
     }
 
+    /// Returns an iterator over the mempool entries that do have a fee rate, sorted by fee rate.
+    /// Page size is 10_000. TODO: Make this configurable.
+    fn sorted_fee_rate_transactions(
+        connection: Arc<Mutex<Connection>>,
+    ) -> Box<dyn Iterator<Item = MemPoolTxMinimalInfo>> {
+        info!("sorted_fee_rate_transactions");
```
jcnelson, line 1181: Please remove or make `test_debug!`

src/core/mempool.rs
jcnelson: Having looked at this code, it appears that this is only needed in order to support instantiating a `TransactionPageCursor`.  You should not need to do this.  Instead, you should do one of the following:

* Create a separate, owned database connection for the iterator.

* Borrow the `db` member of `MemPoolDB` and give `TransactionPageCursor` a lifetime parameter to ensure that it can't outlive the `MemPoolDB` it was built from.

* Implement `Iterator` for `MemPoolDB` so you don't have to borrow or copy state out of it.

* Convert `db` into `Optional<DBConn>`, take the `db` member from `MemPoolDB` to instantiate the `TransactionPageCursor`, and then recover it somehow when it goes out of scope.

* Do not use the Rust iterator trait, but instead do something like `with_db<..>(..)` I described earlier to load the next "page" of transactions.

I think the first option is probably the easiest.

src/core/mempool.rs
```diff
@@ -1064,6 +1173,111 @@ impl MemPoolDB {
...
+        let sql = "
+        SELECT txid, origin_nonce, origin_address, sponsor_nonce, sponsor_address, fee_rate
+        FROM mempool
+        WHERE fee_rate IS NOT NULL
+        ORDER BY fee_rate DESC
+        LIMIT ?
+        OFFSET ?
```
jcnelson, line 1188: You probably didn't see this yet because you're working with a small database, but `OFFSET` will do a full table scan up to the given offset value.  Multiple calls to this iterator for `n` transactions will have `O(n^2)` runtime complexity.  You should avoid using `OFFSET` altogether, and instead use a `WHERE` clause on an indexed column so that database reads can begin at the right page.

Alternatively, you could issue just this query with no offset, and keep around the Rusqlite query statement for the lifetime of the iterator, and just read rows out as `.next()` is called.  You're going to do `Omega(n)` work anyway because you're doing a full scan of the DB, but we really should avoid `O(n^2)` work with multiple calls to a query with `OFFSET`.

gregorycoppola: These are all good suggestions. I like the second idea.

However, the `SELECT` part in and of itself isn't that expensive I think. It's 1) getting the data into Rust and 2) deserializing transactions (aside from 3) processing) that takes a lot of time.

In other words, this wasted `SELECT` time has a really small constant factor.

At some point, we should just merge something using this iterator paradigm and we can follow up on that with little improvements.. Like we have #3321 that builds on this, we can make and measure one adjustment at a time.

gregorycoppola: I'll try to formulate it the DB cursor way and see if it's faster/simpler.

jcnelson: @gregorycoppola The use of `OFFSET` itself is going to make iteration `O(n^2)` because it's slow in sqlite.  If you're reading a 10,000 row table, and you do a `SELECT ... FROM ... WHERE .. OFFSET 9999`, sqlite will read the first 9998 rows' contents and discard them because that's just the way the RDBMS is built.  See [here](https://stackoverflow.com/questions/12266025/sqlite-query-optimization-using-limit-and-offset) for details.

We really shouldn't be using `OFFSET` anywhere in the code if we can help it.

gregorycoppola: But, the overall cost of that scan is neglible, otherwise we wouldn't have good performance numbers on @obycode 's profiling experiments.

I was foiled on every attempt at getting `rusqlite` to create an iterator. It is so hard. Maybe someone else can get it to work but I found it insanely hard.

The `rusqlite` library uses patterns th at create downstream problems, I think, and we should use other models for how to write modern Rust code.

gregorycoppola: There is another pattern I think where you can use an index and a cursor.

The data set is around 28k in the mempoool, from #3313:

Mempool breakdown:

    Total size: 25214
    Nonce less than expected (already processed): 21308
    Nonce is expected (to be processed): 1791
    Nonce greater than expected(can not process yet): 2115

obycode:
My other concern with this technique that I'd like to get your thoughts on @jcnelson, is that it makes a tradeoff of speed versus transaction ordering. By proceeding through the list of transactions, in fee-rate order (ignoring the txs with no fee estimate for now), we will potentially process transactions with lower fee-rates before those with higher fee rates. For example, if I have 2 transactions in the mempool, nonce 2 with a fee rate of X and nonce 1 with a fee rate of Y, Y < X, and then you have 10 transactions with fee-rate Z, Z < Y, then your transactions will be processed before my nonce 2 transaction. And my nonce 2 might not be included in the block if there are enough lower fee-rate transactions to fill the block. Is that a problem?

I have another branch that solves this problem built on Greg's earlier work, and am working on one that builds off of this work.

src/core/mempool.rs
gregorycoppola: Oh yeah, you already mentioned the cursor pattern as the first alternative.

gregorycoppola:
@obycode we have to keep in mind:
1) overall code simplicity
2) use of global evals, not individual conceivable examples

Maybe you can incorporate "fee charged" as a column in your evals. Then, we can see if these strategies actually lead more fees for the miner.

src/core/mempool.rs
```diff
@@ -1064,6 +1173,111 @@ impl MemPoolDB {
...
+            current_page_remaining: vec![],
+        })
+    }
+
+    /// Take a batch of transactions *without* a fee rate estimate, in *arbitrary* order.
+    /// Page size is 10_000. TODO: Make this configurable.
+    /// Note: Nulls in the mempool are, up to a limit, over-written between mempool runs.
+    fn null_fee_rate_transactions(
+        connection: Arc<Mutex<Connection>>,
+    ) -> Box<dyn Iterator<Item = MemPoolTxMinimalInfo>> {
```
gregorycoppola, line 1204: note: iterator returned does not require have lifetime parameters.

if it did depend on the lifetime of `self:MemPoolDB`, then we would not be able to do the borrow of `self` as mutable to use the cost estimator, in calling `todo`:

```
            let inside_result = todo(clarity_tx, &consider, self.cost_estimator.as_mut());
```

src/core/mempool.rs
gregorycoppola: How would this work with the cursor pattern?

**problem**
If we sort by fee rate, then, it's possible there are multiple transactions with the same fee rate. E.g., 

```
[{fee_rate 10, id=1}, {fee_rate = 10, id = 2}, {fee_rate = 9, id=3}]
```

If the last tx processed was `id=1`, and you do the `WHERE` clause as `WHERE fee_rate < 10`, then you will miss 2. If you do `WHERE fee_rate <= 10`, then you will hit `id=1` again.

**idea**
combine `fee_rate` and `time_received` into a single field, and sort by:
* higher fee rate will win
* if tied on fee rate older `time_receivd` will win

**implementation**
Encode `fee_rate` and `time_received` into a single value, and sorting by that. We would have to either add a new field, or repurpose `fee_rate` to be this new field.


Or how would you guys do this?

src/core/mempool.rs
gregorycoppola: I was thinking to create an owned connection object, and pass it to the iterator.

But, iterators are being created and destroyed constantly. The miner more or less starts a new iteration once one is done.

The pattern I was going with is the "connection pool", which is pretty common:
* https://www.cockroachlabs.com/blog/what-is-connection-pooling/

gregorycoppola: I uploaded another alternative...

I rewrote the `TestPeer` so that `MemPoolDB` doesn't get sent across threads, and so doesn't require `Send` or `Sync` to be implemented. That test framework was the only thing requiring this, so prod doesn't need it, so let's just unwind this restriction.

Thus, we can now keep a reference to `Rc<DBConn>`, instead of `Arc<Mutex<DBConn>>`. The removal of the mutex means nothing is checked at runtime, everything is checked at compile time, and there is no new risk of deadlock.

I kept then two copies of the db (a simple kind of pool), one of which can be borrowed as mutable, and the other is a read-only `Rc`.

src/core/mempool.rs
gregorycoppola: note: I will rebase this after #3327

src/core/mempool.rs
jcnelson: You should not need to use any sort of reference counting or smart pointer to store a `DBConn`.  You should be able to either borrow the mempool's connection in the iterator implementation, or create an owned connection inside the iterator.

If you find yourself trying to use `Arc`, `Rc`, `RefCell`, etc. to get around the borrow checker, then it's a sign that your code needs to be refactored.  The Rust compiler makes a lot of very useful static guarantees about the code's correctness with respect to resource acquisition and release (be it memory, file descriptors, etc.) if you can express them in terms of lifetimes.  Using reference-counting precludes asserting these static guarantees end-to-end, so they should be avoided unless it's absolutely necessary.

src/core/mempool.rs
jcnelson: > Encode fee_rate and time_received into a single value

No need to do this.  Just have a compound `ORDER BY` and `WHERE` clauses, like `WHERE fee_rate < ? OR (fee_rate == ? AND time_received < ? ORDER BY fee_rate DESC, time_received ASC`) where `fee_rate` and `time_received` are both indexed columns.  Your cursor is then simply the last `fee_rate` and `time_received` you saw.  Whatever you do, be sure to run `EXPLAIN QUERY PLAN` on the query to make _absolutely certain_ that you are not introducing any table scans.

obycode: @gregorycoppola I have a clean solution to these issues that I was testing over the weekend but have been delayed by a family health issue. I think I should be able to chat later today if you are available and I can share that progress.

jcnelson:
> My other concern with this technique that I'd like to get your thoughts on @jcnelson, is that it makes a tradeoff of speed versus transaction ordering. By proceeding through the list of transactions, in fee-rate order (ignoring the txs with no fee estimate for now), we will potentially process transactions with lower fee-rates before those with higher fee rates. For example, if I have 2 transactions in the mempool, nonce 2 with a fee rate of X and nonce 1 with a fee rate of Y, Y < X, and then you have 10 transactions with fee-rate Z, Z < Y, then your transactions will be processed before my nonce 2 transaction. And my nonce 2 might not be included in the block if there are enough lower fee-rate transactions to fill the block. Is that a problem?

@obycode Very astute observation. Also, this is something @a3slade asked about today in the blockchain meeting.

Looking at the problem of filling a block from first principles, what we're really doing is trying to solve a 5-dimensional [knapsack problem](https://en.wikipedia.org/wiki/Knapsack_problem).  This is of course NP-complete so exact solutions are infeasible, but fortunately there are good approximation algorithms for this which run in polynomial time.  Ideally, we'd be using one of these approximation algorithms to fill blocks.

The "size" metric we're using for transactions here is really a 5-dimensional STX/work vector.  Given the runtime of the transaction and its fee, we can derive how much STX the user is paying for each resource.  From there, we can (in theory) calculate the optimal packing of a block that maximizes the total STX in transaction fees it contains.

Right now, we're missing a few things. First, we're not using the true STX/work vector; we're instead approximating it using a blind and easily-gamed learning function applied to historic data.  We can do much better than this.  Because Clarity is decidable, we could "unroll" the execution of a Clarity function body into an "execution tree," where each node in the tree represents one discrete Clarity VM computation, and parent-child links represent the next computation(s) reachable from the parent.

At each node in the tree, we could derive a function that would let us compute the 5-dimensional runtime vector for each of subtrees rooted at its children.  So each time you walk to an execution tree node while evaluating a contract-call, you could determine the lower and upper bounds of the cost incurred by the rest of the call (think of this like doing all of the `runtime_cost()` calls you'll ever call in a particular branch up-front).  Importantly, the delta between the lower and upper bounds would shrink the deeper you go in the tree (i.e. the closer you get to running the code to completion, the better your estimate of its runtime cost), because Clarity's decidability means that the execution graph cannot have cycles (i.e. the set of execution paths for every function body are guaranteed to constitute a tree).

Then, given a 5-dimensional error constant _E_, you could partially evaluate each contract-call in the mempool by walking this tree until the delta between the most-costly branch and least-costly branch is less than _E_.  You'd choose _E_ such as to find a "sweet spot" trade-off between the computation you'd have to do on each contract-call transaction and the accuracy of your estimation on what the true STX/work vector is for that contract-call (note that if _E = 0_, then you're always evaluating every contract-call to completion to learn the true STX/work vector).

Once you have the execution tree computed for each smart contract (which you could do at instantiation time), you could start attacking the block-packing problem with a knapsack approximation algorithm.  Once you computed the upper and lower bounds for each contract-call's STX/work vectors (or more generally, some distribution of STX/work vectors in-between the upper and lower bounds), you'd run the knapsack approximation algorithm on the set of transactions with their estimated costs to get your block packing.

Of course, we're nowhere near this implementation yet.

To speak to your question about nonces, you're absolutely right -- what we have now is suboptimal.  The solution would be to treat a chain of transactions that could all be mined at once as a single "transaction group" for the purpose of estimating an aggregate STX/work vector (note that a single transaction is just a transaction group with one member).  Then, we'd prioritize including transaction groups instead of bare transactions.  There are lots of corner cases to consider, though -- in particular, you'd need to make sure that the next transaction to add to a group is actually mineable.  But, I think this is doable.

src/core/mempool.rs
gregorycoppola: ok.. `Rc` is gone!... even though I miss it :(

src/core/mempool.rs
gregorycoppola: @obycode ok nice!.. this is the last issue remaining.

gregorycoppola:
@obycode For the corner case of some higher fee transaction becoming mineable in the middle of processing, I agree with @jcnelson that we are making an approximate solution to a (possibly NP) hard problem.

What if there were 1 million transactions in the mempool. Are you going to go back to the start every single time? The essence of a greedy solution is that you will get some gains and lose some others.

a3slade:
> First, we're not using the true STX/work vector; we're instead approximating it using a blind and easily-gamed learning function applied to historic data. 

Reading all of the stacks documentation, I was under the impression that the function costs were determined in a manner like Jude described with some minor exceptions. I thought the whole thing was unrolled ahead of time and the *max* for each branch was calculated and those maxes propagated back up to give the overall worse case runtime. And then the fees were set from that. Pavi clarified for me that this was not what was happening at all.

> The solution would be to treat a chain of transactions that could all be mined at once as a single "transaction group" for the purpose of estimating an aggregate STX/work vector (note that a single transaction is just a transaction group with one member).

One possibility here would be to calculate the cumulative cost/fees for each tx and select on their ratio. Like, if my txs had costs of [3, 2, 7] runtime units and my fees were [2, 1, 9] money units then the cost/fees would be [3, 5, 12] and [2, 3, 12] respectively and the ratio would be [.66, .6, 1]. So unless we were at the end of the block and attempting to fill the remainder, then we’d only want to consider the bundle of all 3.

My intuition is that we probably want two data structures. One that has one tx or one tx group from each user and is sorted by fee (or derived quantity) and another that is just a hash keyed by the user that keeps all the remaining txs. Then whenever a user’s tx or tx group is selected, it can be deleted from the first structure, the user can be looked up in the hash, their next tx or tx group can be selected and inserted into the first structure (which should probably be a tree so we can keep it sorted).

Perhaps something else to consider is that the current blocks are maxing out on the reads but not the other dimensions. We are only concerned with total runtime, right? Can there be more reads if we are not maxing out the others? Maybe we should have a weighted sum of these 5 dimensions to get an overall cost and overall limit?

Completely off-topic: since it’s primarily one contract that is using up all those reads, I assume that people have fine-combed that contract to ensure that those reads are the minimum necessary?

gregorycoppola:
@a3slade I think one problem with having these kind of work queues is that one of the design requirments we started with was **not** to put the entire mempool in RAM... if we relax that requirement, then we can try work queue stuff, afaiu.

obycode:
I’m using a very simple technique to just retry the higher fee-rate transactions in branch `simple-iterator`, that should be effective in matching what is currently happening in the mempool walk. As it processes transactions in fee-rate order, if a transaction can’t run because its nonces are too high, it gets added to a retry list. Transactions with nonces too low are just dropped (that’s the majority of transactions). Then after each transaction is processed, we first cycle through the retry list before pulling in another set of lower fee-rate transactions from the db. The remaining work for that is just to add a cap to the size of this retry list, but that will be simple. This technique has essentially no effect on the runtime in my benchmarking, but it gets better ordering.

src/core/mempool.rs
gregorycoppola: Hello everyone...

I have been doing some profiling to see how time taken to read in entries actually empirically changes as the `OFFSET` grows.

I pushed the commit that does the logging

We **do not** see evidence that the time taken to read in examples is actually even growing at all as `OFFSET` grows.

```
offset: 0, avg: 20916.010055478502, count: 2884
offset: 10000, avg: 8705.302325581395, count: 2838
offset: 20000, avg: 8698.485935984481, count: 2062
offset: 30000, avg: 15763.23076923077, count: 793
offset: 40000, avg: 17311.821292775665, count: 526
offset: 50000, avg: 16809.42105263158, count: 380
offset: 60000, avg: 13563.802973977696, count: 269
offset: 70000, avg: 7634.550898203593, count: 167
offset: 80000, avg: 2070.4074074074074, count: 81
```

Note how offset `0` has the largest average. Also, the average for `80000` offset is actually smaller than the others.

I think randomness, and IO factors are more significant.

This is important because, if we implement paging, it will:
* complicate the code
* **introduce new costs**

Thus, not only is this `OFFSET`-based solution the simplest, it will probably turn out to be the fastest.

jcnelson:
> @a3slade Reading all of the stacks documentation, I was under the impression that the function costs were determined in a manner like Jude described with some minor exceptions. I thought the whole thing was unrolled ahead of time and the *max* for each branch was calculated and those maxes propagated back up to give the overall worse case runtime. And then the fees were set from that. Pavi clarified for me that this was not what was happening at all.

This is something I hope the L1 working group has an opportunity to work on post-2.1.  Moreover, adding this isn't a consensus-breaking change, so it can be rolled out incrementally.

The reason the code works the way it does now with a blind estimator is because up until recently, transaction fees were somewhat small and the mempool was rarely congested.  So instead we implemented something that gets a lot of the benefit of having a fullly-featured estimator for comparatively little effort (also, we had fewer person-hours available at the time, and subnets and 2.1 weren't nearly as far along).  But eventually I think we should replace the current estimator with the ideal solution, especially now that transaction fees are rising and blocks are consistently full.  In the near to medium term, I suspect that the USD-denominated delta between what miners earn now with transaction fees versus what they could earn with this ideal solution will become large enough to justify taking the time to build out the ideal solution.

It sounds like you have a good idea about how to go about selecting groups of transactions :+1: 

> @a3slade Perhaps something else to consider is that the current blocks are maxing out on the reads but not the other dimensions. We are only concerned with total runtime, right? Can there be more reads if we are not maxing out the others? Maybe we should have a weighted sum of these 5 dimensions to get an overall cost and overall limit?

I think you're asking why we don't treat transactions as consuming a single-dimension runtime, but instead a 5-dimensional runtime vector?  If so, the reason is that the ability for a computer to run a Stacks node that stays in-sync with the network and boots up in a reasonable amount of time is dependent on blocks being processed relatively quickly.  But, this is dependent on the block's evaluation not consuming too much of _any_ computing resource that would slow it down too much; if the block consumes so much of one computing resource on the host computer as to make it take more than e.g. 30s to process, then we are not meeting our target performance.

The use of a 5-dimensional vector to track performance is an effort to identify and limit the consumption of each possible computing resource whose over-consumption would lead to a slow block.  We can then chose block limits based on targeted benchmark data -- if a block consumes more of a resource than the block limit, it would (likely) take too long to process.  By treating each resource separately, we can get a better understanding of how to limit block capacity so that the Stacks node can keep up with the network.

Of course, this is not the only strategy.  Other systems like Ethereum represent resource consumption in a single dimension ("gas").  A result of this design decision is that it's harder to know how much of any computing resource a maximally-packed block could use.  For example, it's harder to answer questions like "what is the maximum number of state-trie queries that can be executed with the current gas limit?" or "what is the maximum number of bytes that we can load into RAM with a transaction?".  This in turn makes it hard to reason about an Ethereum node's runtime behavior with regards to resource consumption, which in turn makes it hard to calculate the best gas cost for a given operation.  For example, if you get it too low, someone could spam the network with malicious transactions that stall the network due to underpriced disk I/O ([this happened](https://www.coindesk.com/markets/2016/10/06/so-ethereums-blockchain-is-still-under-attack)).  As another example, if you set it too high, developers will misuse the blockchain's event log facility to store data instead ([this happens to this day](https://ethereum.stackexchange.com/questions/73777/why-is-storing-via-sstore-more-expensive-than-via-events-mechanism)).  If instead Ethereum had a multi-dimensional resource accounting system, they could set more-accurate resource costs on their VM operations, and be more able to prevent resource over-consumption (or bad pricing) before it becomes a problem in production.

For estimating purposes it might make sense to calculate a weighted sum of these parameters when selecting transactions for inclusion in a non-full block.  But once at least one dimension is filled, the estimator will need to stop considering transactions that consume that dimension's resource.

> @a3slade Completely off-topic: since it’s primarily one contract that is using up all those reads, I assume that people have fine-combed that contract to ensure that those reads are the minimum necessary?

We did our best when we wrote it.  However, given that there's nothing special about BNS, it's entirely possible ([and encouraged!](https://forum.stacks.org/t/making-bns-ready-for-prime-time/13791)) to write a separate but more-efficient BNS implementation and convince the ecosystem to migrate to that contract instead.

a3slade:
Thanks so much for the detailed response! 🧡

obycode:
The `simple-iterator` branch has the following properties:
* Uses rusqlite's `Rows` iterator to read one row at a time
* Caches the nonces in memory to avoid repeated lookups
* Restarts search from the highest fee-rate transactions after every executed transaction
  * Caches potential transactions in memory to retry later

It does seem to be about 20% slower than the version using the `OFFSET` in the query, but does not have the problems associated with that when the mempool gets really big and it still fills the block within plenty of time. I'm looking into the profiling data, and it doesn't seem like there are any obvious bottlenecks left to remove with this technique.

I'd like to add a size limit to the cache of transactions to retry. Would that be worth adding a setting for, or should I just choose a reasonable value for this? I guess we may also want to add a size limit to the nonce cache as well.

jcnelson:
> @obycode Uses rusqlite's Rows iterator to read one row at a time

Hey Brice, this might be the source of the slowness.  Sqlite3 stores data in fixed-sized pages, which are potentially much larger than the size of a single row.  So, reading just a row at a time might introduce some reduction in goodput.  Does fiddling with the page size make a difference?  Does reading approximately a page-worth of data at a time (i.e. multiple rows) make a difference?

obycode:
> So, reading just a row at a time might introduce some reduction in goodput. Does fiddling with the page size make a difference? Does reading approximately a page-worth of data at a time (i.e. multiple rows) make a difference?

I didn't see a way to do this without modifying `rusqlite`. Once I have the `Rows`, I can get one item at a time using the `next` method, but I don't see a way to get _n_ rows at a time.

gregorycoppola:
Thanks for the info/change, @obycode .

gregorycoppola:
I still think the optimal for both simplicity and time is to keep the `OFFSET` for now, perhaps filing an issue to find a way to further optimize.

`OFFSET` might be a problem at web scale, but it's not showing up as a problem right now.

gregorycoppola:
I can also patch in the cursor way, though.. either way.

jcnelson:
Please add an index on `fee_rate`.  We didn't do it in 2.05.0.4.0, but we'll need to if we're going to `ORDER BY fee_rate`.

obycode:
> Please add an index on `fee_rate`. We didn't do it in 2.05.0.4.0, but we'll need to if we're going to `ORDER BY fee_rate`.

That was included in @kantai's changes for 2.05.0.4.0:
_mempool.rs:511_
```
CREATE INDEX IF NOT EXISTS by_fee_rate ON mempool(fee_rate);
```

jcnelson:
Thanks @obycode; I stand corrected.  I was playing around with some synthetic mempool data with millions of rows, and noticed that `ORDER BY fee_rate` would cause a `SELECT` statement to slow down considerably.  Looks like I was using an older version of the schema.

gregorycoppola:
Thanks for all the ideas and review, everyone.

I re-merged in `develop` and am bringing up a mock miner with this PR.

src/core/mempool.rs
gregorycoppola: rebased

src/core/mempool.rs
```diff
@@ -1129,13 +1167,110 @@ impl MemPoolDB {
...
+            current_offset: 0,
+            page_size: 10_000,
+            current_page_remaining: vec![],
+        }
+    }
+
+    /// Take a batch of transactions *without* a fee rate estimate, in *arbitrary* order.
+    /// Page size is 10_000. TODO: Make this configurable.
+    /// Note: Nulls in the mempool are, up to a limit, over-written between mempool runs.
+    fn null_fee_rate_transactions(connection: &Connection) -> TransactionPageCursor {
```
gregorycoppola, line 1193: @jcnelson @obycode 

Important question: Should we introduce a sorting order for "null fee rate transactions"? We could sort by `accept_time` or `height`. These have indices already I think:

```
    "CREATE INDEX IF NOT EXISTS by_height ON mempool(height);",
    "CREATE INDEX IF NOT EXISTS by_timestamp ON mempool(accept_time);",
```

obycode: Yeah, that's a good point. It seems like first come, first served makes sense here. What does `height` in this table actually mean?

gregorycoppola:
Here is a snapshot of the logs under filter

```
grep -E "Mempool iteration|Miner: |deadline exceeded"
```

[miner-logs-oct122022.log](https://github.com/stacks-network/stacks-blockchain/files/9767511/miner-logs-oct122022.log)

src/core/mempool.rs
jcnelson: Yes -- if you're ordering transactions by these columns, then they need to be indexed.  I would check the query with `EXPLAIN QUERY PLAN` just to be sure.  Also, depending on the ultimate structure of the `WHERE` and `ORDER BY` clauses, you may need to create a compound index, such as:

```
CREATE INDEX IF NOT EXISTS by_height_and_ts ON mempool(height,accept_time);
```

jcnelson:
> @obycode I didn't see a way to do this without modifying rusqlite. Once I have the Rows, I can get one item at a time using the next method, but I don't see a way to get n rows at a time.

That's fine.  What really matters is that we don't load data we don't need.

Thinking about it more, the fact that we're iterating through the rows in the mempool in an order that is _not_ consistent with the insertion order would hurt us if the page size is too big, because wouldn't expect to see any benefit from a page containing other row data (i.e. we wouldn't necessarily be querying that row data next).  If the transaction iteration is truly uncorrelated with the insertion order, then pre-fetching data with a large page size ought to only lead to thrashing.  So, I suspect that a page size that is about equal to the size of the row would be best -- we'd only load the data we need.

I think you'd need to fiddle with this a bit to see if it makes a difference.  The default page size is 4096 bytes, which I think is bigger than a single row by a considerable margin (e.g. a stx-transfer transaction is only 180 bytes).  If you have a moment, could either you or @gregorycoppola try using smaller page sizes?  E.g. 512 bytes? 1024 bytes? (it must be powers of 2).

gregorycoppola:
@jcnelson @obycode I uploaded a version that uses @obycode 's pattern to keep the cursor open, but doesn't mix in the other caching.

Should we just run like a 1 million transaction mempool stress test, and see which one is better?

How big of a mempool do we want to support?

If you can only fit 1000 tx's in a block, does a >1 million transaction mempool even make sense?

obycode:
> How big of a mempool do we want to support?
> If you can only fit 1000 tx's in a block, does a >1 million transaction mempool even make sense?

Our goal should be that the size of the mempool has a negligible effect on the transaction selection process. It shouldn't matter if there are a billion transactions if I can build the block by only looking at the first 5000, by fee-rate order, then this process should be unaffected by mempool size.

src/core/mempool.rs
gregorycoppola: at `5dd7fe1140893a6e26bb28dd6c2fbac98f6d61e1` the query plans are

```
sqlite> EXPLAIN QUERY PLAN SELECT txid, origin_nonce, origin_address, sponsor_nonce, sponsor_address, fee_rate FROM mempool WHERE fee_rate IS NULL ORDER BY accept_time ASC;
QUERY PLAN
|--SEARCH TABLE mempool USING INDEX by_fee_rate (fee_rate=?)
`--USE TEMP B-TREE FOR ORDER BY
sqlite> EXPLAIN QUERY PLAN SELECT txid, origin_nonce, origin_address, sponsor_nonce, sponsor_address, fee_rate FROM mempool WHERE fee_rate IS NOT NULL ORDER BY fee_rate DESC;
QUERY PLAN
`--SCAN TABLE mempool USING INDEX by_fee_rate
```

gregorycoppola: schema i have is as follows.. looks like there is an index on `accept_time`:

```
sqlite> .schema mempool
CREATE TABLE mempool(
        txid TEXT NOT NULL,
        origin_address TEXT NOT NULL,
        origin_nonce INTEGER NOT NULL,
        sponsor_address TEXT NOT NULL,
        sponsor_nonce INTEGER NOT NULL,
        tx_fee INTEGER NOT NULL,
        length INTEGER NOT NULL,
        consensus_hash TEXT NOT NULL,
        block_header_hash TEXT NOT NULL,
        height INTEGER NOT NULL,    -- stacks block height
        accept_time INTEGER NOT NULL,
        tx BLOB NOT NULL, last_known_origin_nonce INTEGER, last_known_sponsor_nonce INTEGER, fee_rate NUMBER,
        PRIMARY KEY (txid),
        UNIQUE (origin_address, origin_nonce),
        UNIQUE (sponsor_address,sponsor_nonce)
    );
CREATE INDEX by_txid ON mempool(txid);
CREATE INDEX by_height ON mempool(height);
CREATE INDEX by_txid_and_height ON mempool(txid,height);
CREATE INDEX by_sponsor ON mempool(sponsor_address, sponsor_nonce);
CREATE INDEX by_origin ON mempool(origin_address, origin_nonce);
CREATE INDEX by_timestamp ON mempool(accept_time);
CREATE INDEX by_chaintip ON mempool(consensus_hash,block_header_hash);
CREATE INDEX by_fee_rate ON mempool(fee_rate);
```

gregorycoppola: obycode says he has this for the query plan... not sure why my db is different.. I got my db by 1) start with snapshot, 2) run the `master->develop` branch 

```
sqlite> explain query plan
   ...> SELECT txid, origin_nonce, origin_address, sponsor_nonce, sponsor_address, fee_rate
   ...>             FROM mempool
   ...>             WHERE fee_rate IS NOT NULL
   ...>             ORDER BY fee_rate DESC
   ...> ;
QUERY PLAN
`--SEARCH mempool USING INDEX by_fee_rate (fee_rate>?)
```

gregorycoppola:
We can follow up on #3337, which is less commits to the same place, along with the improvements bundled.

PR #3337, opened 2022-10-12
Mempool improvements

obycode:
This is the cleaned up version of `simple-iterator` discussed in #3326 and #3313.

To-do before checking in:
* verify that a snapshot chainstate creates a DB with the right query plan
  * can do as part of bringing up mock miner from snapshot.. then can check the query plan

### DIFF
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -11256,6 +11256,190 @@ pub mod test {
         assert_eq!(stacks_block.txs.len(), 11);
     }
 
+    #[test]
+    fn mempool_walk_test_users_1_rounds_10_cache_size_2_null_prob_0() {
+        paramaterized_mempool_walk_test(1, 10, 2, 0)
+    }
+
+    #[test]
+    fn mempool_walk_test_users_10_rounds_3_cache_size_2_null_prob_0() {
+        paramaterized_mempool_walk_test(10, 3, 2, 0)
+    }
+
+    #[test]
+    fn mempool_walk_test_users_1_rounds_10_cache_size_2_null_prob_50() {
+        paramaterized_mempool_walk_test(1, 10, 2, 50)
+    }
+
+    #[test]
+    fn mempool_walk_test_users_10_rounds_3_cache_size_2_null_prob_50() {
+        paramaterized_mempool_walk_test(10, 3, 2, 50)
+    }
+
+    #[test]
+    fn mempool_walk_test_users_1_rounds_10_cache_size_2_null_prob_100() {
+        paramaterized_mempool_walk_test(1, 10, 2, 100)
+    }
+
+    #[test]
+    fn mempool_walk_test_users_10_rounds_3_cache_size_2_null_prob_100() {
+        paramaterized_mempool_walk_test(10, 3, 2, 100)
+    }
+
+    /// With the parameters given, create `num_rounds` transactions per each user in `num_users`.
+    /// `nonce_and_candidate_cache_size` is the cache size used for both of the nonce cache
+    /// and the candidate cache.
+    fn paramaterized_mempool_walk_test(
+        num_users: usize,
+        num_rounds: usize,
+        nonce_and_candidate_cache_size: u64,
+        consider_no_estimate_tx_prob: u8,
+    ) {
+        let key_address_pairs: Vec<(Secp256k1PrivateKey, StacksAddress)> = (0..num_users)
+            .map(|_user_index| {
+                let privk = StacksPrivateKey::new();
+                let addr = StacksAddress::from_public_keys(
+                    C32_ADDRESS_VERSION_TESTNET_SINGLESIG,
+                    &AddressHashMode::SerializeP2PKH,
+                    1,
+                    &vec![StacksPublicKey::from_private(&privk)],
+                )
+                .unwrap();
+                (privk, addr)
+            })
+            .collect();
+
+        let mut peer_config = TestPeerConfig::new(
+            &format!(
+                "mempool_walk_test_users_{}_rounds_{}_cache_size_{}_null_prob_{}",
+                num_users, num_rounds, nonce_and_candidate_cache_size, consider_no_estimate_tx_prob
+            ),
+            2002,
+            2003,
+        );
+
+        peer_config.initial_balances = vec![];
+        for (privk, addr) in &key_address_pairs {
+            peer_config
+                .initial_balances
+                .push((addr.to_account_principal(), 1000000000));
+        }
+
+        let mut peer = TestPeer::new(peer_config);
+
+        let chainstate_path = peer.chainstate_path.clone();
+
+        let first_stacks_block_height = {
+            let sn =
+                SortitionDB::get_canonical_burn_chain_tip(&peer.sortdb.as_ref().unwrap().conn())
+                    .unwrap();
+            sn.block_height
+        };
+
+        let recipient_addr_str = "ST1RFD5Q2QPK3E0F08HG9XDX7SSC7CNRS0QR0SGEV";
+        let recipient = StacksAddress::from_string(recipient_addr_str).unwrap();
+        let mut sender_nonce = 0;
+
+        let mut last_block = None;
+        // send transactions to the mempool
+        let tip = SortitionDB::get_canonical_burn_chain_tip(&peer.sortdb.as_ref().unwrap().conn())
+            .unwrap();
+
+        let (burn_ops, stacks_block, microblocks) = peer.make_tenure(
+            |ref mut miner,
+             ref mut sortdb,
+             ref mut chainstate,
+             vrf_proof,
+             ref parent_opt,
+             ref parent_microblock_header_opt| {
+                let parent_tip = match parent_opt {
+                    None => StacksChainState::get_genesis_header_info(chainstate.db()).unwrap(),
+                    Some(block) => {
+                        let ic = sortdb.index_conn();
+                        let snapshot = SortitionDB::get_block_snapshot_for_winning_stacks_block(
+                            &ic,
+                            &tip.sortition_id,
+                            &block.block_hash(),
+                        )
+                        .unwrap()
+                        .unwrap(); // succeeds because we don't fork
+                        StacksChainState::get_anchored_block_header_info(
+                            chainstate.db(),
+                            &snapshot.consensus_hash,
+                            &snapshot.winning_stacks_block_hash,
+                        )
+                        .unwrap()
+                        .unwrap()
+                    }
+                };
+
+                let parent_header_hash = parent_tip.anchored_header.block_hash();
+                let parent_consensus_hash = parent_tip.consensus_hash.clone();
+
+                let mut mempool =
+                    MemPoolDB::open_test(false, 0x80000000, &chainstate_path).unwrap();
+
+                let coinbase_tx = make_coinbase(miner, 0);
+
+                for round_index in 0..num_rounds {
+                    for user_index in 0..num_users {
+                        let stx_transfer = make_user_stacks_transfer(
+                            &key_address_pairs[user_index].0,
+                            round_index as u64,
+                            200,
+                            &recipient.to_account_principal(),
+                            1,
+                        );
+
+                        mempool
+                            .submit(
+                                chainstate,
+                                &parent_consensus_hash,
+                                &parent_header_hash,
+                                &stx_transfer,
+                                None,
+                                &ExecutionCost::max_value(),
+                                &StacksEpochId::Epoch20,
+                            )
+                            .unwrap();
+                    }
+                }
+
+                let mut settings = BlockBuilderSettings::max_value();
+                settings.mempool_settings.nonce_cache_size = nonce_and_candidate_cache_size;
+                settings.mempool_settings.candidate_retry_cache_size =
+                    nonce_and_candidate_cache_size;
+                settings.mempool_settings.consider_no_estimate_tx_prob =
+                    consider_no_estimate_tx_prob;
+                let anchored_block = StacksBlockBuilder::build_anchored_block(
+                    chainstate,
+                    &sortdb.index_conn(),
+                    &mut mempool,
+                    &parent_tip,
+                    tip.total_burn,
+                    vrf_proof,
+                    Hash160([0 as u8; 20]),
+                    &coinbase_tx,
+                    settings,
+                    None,
+                )
+                .unwrap();
+                (anchored_block.0, vec![])
+            },
+        );
+
+        last_block = Some(stacks_block.clone());
+
+        peer.next_burnchain_block(burn_ops.clone());
+        peer.process_stacks_epoch_at_tip(&stacks_block, &microblocks);
+
+        // Both user transactions and the coinbase should have been mined.
+        assert_eq!(
+            stacks_block.txs.len() as u64,
+            (1 + num_rounds * num_users) as u64
+        );
+    }
+
     static CONTRACT: &'static str = "
 (define-map my-map int int)
 (define-private (do (input bool))
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -10989,6 +10989,143 @@ pub mod test {
         }
     }
 
+    #[test]
+    /// Test the situation in which the nonce order of transactions from a user. That is,
+    /// nonce 1 has a higher fee than nonce 0.
+    /// Want to see that both transactions can go into the same block, because the miner
+    /// should make multiple passes.
+    fn test_fee_order_mismatch_nonce_order() {
+        let privk = StacksPrivateKey::from_hex(
+            "42faca653724860da7a41bfcef7e6ba78db55146f6900de8cb2a9f760ffac70c01",
+        )
+        .unwrap();
+        let addr = StacksAddress::from_public_keys(
+            C32_ADDRESS_VERSION_TESTNET_SINGLESIG,
+            &AddressHashMode::SerializeP2PKH,
+            1,
+            &vec![StacksPublicKey::from_private(&privk)],
+        )
+        .unwrap();
+
+        let mut peer_config = TestPeerConfig::new(
+            "test_build_anchored_blocks_stx_transfers_single",
+            2002,
+            2003,
+        );
+        peer_config.initial_balances = vec![(addr.to_account_principal(), 1000000000)];
+
+        let mut peer = TestPeer::new(peer_config);
+
+        let chainstate_path = peer.chainstate_path.clone();
+
+        let num_blocks = 10;
+        let first_stacks_block_height = {
+            let sn =
+                SortitionDB::get_canonical_burn_chain_tip(&peer.sortdb.as_ref().unwrap().conn())
+                    .unwrap();
+            sn.block_height
+        };
+
+        let recipient_addr_str = "ST1RFD5Q2QPK3E0F08HG9XDX7SSC7CNRS0QR0SGEV";
+        let recipient = StacksAddress::from_string(recipient_addr_str).unwrap();
+        let mut sender_nonce = 0;
+
+        let mut last_block = None;
+        // send transactions to the mempool
+        let tip = SortitionDB::get_canonical_burn_chain_tip(&peer.sortdb.as_ref().unwrap().conn())
+            .unwrap();
+
+        let (burn_ops, stacks_block, microblocks) = peer.make_tenure(
+            |ref mut miner,
+             ref mut sortdb,
+             ref mut chainstate,
+             vrf_proof,
+             ref parent_opt,
+             ref parent_microblock_header_opt| {
+                let parent_tip = match parent_opt {
+                    None => StacksChainState::get_genesis_header_info(chainstate.db()).unwrap(),
+                    Some(block) => {
+                        let ic = sortdb.index_conn();
+                        let snapshot = SortitionDB::get_block_snapshot_for_winning_stacks_block(
+                            &ic,
+                            &tip.sortition_id,
+                            &block.block_hash(),
+                        )
+                        .unwrap()
+                        .unwrap(); // succeeds because we don't fork
+                        StacksChainState::get_anchored_block_header_info(
+                            chainstate.db(),
+                            &snapshot.consensus_hash,
+                            &snapshot.winning_stacks_block_hash,
+                        )
+                        .unwrap()
+                        .unwrap()
+                    }
+                };
+
+                let parent_header_hash = parent_tip.anchored_header.block_hash();
+                let parent_consensus_hash = parent_tip.consensus_hash.clone();
+
+                let mut mempool =
+                    MemPoolDB::open_test(false, 0x80000000, &chainstate_path).unwrap();
+
+                let coinbase_tx = make_coinbase(miner, 0);
+
+                let stx_transfer0 =
+                    make_user_stacks_transfer(&privk, 0, 200, &recipient.to_account_principal(), 1);
+                let stx_transfer1 =
+                    make_user_stacks_transfer(&privk, 1, 400, &recipient.to_account_principal(), 1);
+
+                mempool
+                    .submit(
+                        chainstate,
+                        &parent_consensus_hash,
+                        &parent_header_hash,
+                        &stx_transfer0,
+                        None,
+                        &ExecutionCost::max_value(),
+                        &StacksEpochId::Epoch20,
+                    )
+                    .unwrap();
+
+                mempool
+                    .submit(
+                        chainstate,
+                        &parent_consensus_hash,
+                        &parent_header_hash,
+                        &stx_transfer1,
+                        None,
+                        &ExecutionCost::max_value(),
+                        &StacksEpochId::Epoch20,
+                    )
+                    .unwrap();
+
+                let anchored_block = StacksBlockBuilder::build_anchored_block(
+                    chainstate,
+                    &sortdb.index_conn(),
+                    &mut mempool,
+                    &parent_tip,
+                    tip.total_burn,
+                    vrf_proof,
+                    Hash160([0 as u8; 20]),
+                    &coinbase_tx,
+                    BlockBuilderSettings::max_value(),
+                    None,
+                )
+                .unwrap();
+                (anchored_block.0, vec![])
+            },
+        );
+
+        last_block = Some(stacks_block.clone());
+
+        peer.next_burnchain_block(burn_ops.clone());
+        peer.process_stacks_epoch_at_tip(&stacks_block, &microblocks);
+
+        // Both user transactions and the coinbase should have been mined.
+        assert_eq!(stacks_block.txs.len(), 3);
+    }
+
     static CONTRACT: &'static str = "
 (define-map my-map int int)
 (define-private (do (input bool))
--- a/src/chainstate/stacks/miner.rs
+++ b/src/chainstate/stacks/miner.rs
@@ -11129,137 +11129,6 @@ pub mod test {
     }
 
     #[test]
-    /// Test with a number of transactions (10) that is "much greater" than the cache size (2).
-    /// Test that all transactions get into the block.
-    fn test_1_user_10_transactions_with_cache_size_2() {
-        let privk = StacksPrivateKey::from_hex(
-            "42faca653724860da7a41bfcef7e6ba78db55146f6900de8cb2a9f760ffac70c01",
-        )
-        .unwrap();
-        let addr = StacksAddress::from_public_keys(
-            C32_ADDRESS_VERSION_TESTNET_SINGLESIG,
-            &AddressHashMode::SerializeP2PKH,
-            1,
-            &vec![StacksPublicKey::from_private(&privk)],
-        )
-        .unwrap();
-
-        let mut peer_config = TestPeerConfig::new(
-            "test_build_anchored_blocks_stx_transfers_single",
-            2002,
-            2003,
-        );
-        peer_config.initial_balances = vec![(addr.to_account_principal(), 1000000000)];
-
-        let mut peer = TestPeer::new(peer_config);
-
-        let chainstate_path = peer.chainstate_path.clone();
-
-        let first_stacks_block_height = {
-            let sn =
-                SortitionDB::get_canonical_burn_chain_tip(&peer.sortdb.as_ref().unwrap().conn())
-                    .unwrap();
-            sn.block_height
-        };
-
-        let recipient_addr_str = "ST1RFD5Q2QPK3E0F08HG9XDX7SSC7CNRS0QR0SGEV";
-        let recipient = StacksAddress::from_string(recipient_addr_str).unwrap();
-        let mut sender_nonce = 0;
-
-        let mut last_block = None;
-        // send transactions to the mempool
-        let tip = SortitionDB::get_canonical_burn_chain_tip(&peer.sortdb.as_ref().unwrap().conn())
-            .unwrap();
-
-        let (burn_ops, stacks_block, microblocks) = peer.make_tenure(
-            |ref mut miner,
-             ref mut sortdb,
-             ref mut chainstate,
-             vrf_proof,
-             ref parent_opt,
-             ref parent_microblock_header_opt| {
-                let parent_tip = match parent_opt {
-                    None => StacksChainState::get_genesis_header_info(chainstate.db()).unwrap(),
-                    Some(block) => {
-                        let ic = sortdb.index_conn();
-                        let snapshot = SortitionDB::get_block_snapshot_for_winning_stacks_block(
-                            &ic,
-                            &tip.sortition_id,
-                            &block.block_hash(),
-                        )
-                        .unwrap()
-                        .unwrap(); // succeeds because we don't fork
-                        StacksChainState::get_anchored_block_header_info(
-                            chainstate.db(),
-                            &snapshot.consensus_hash,
-                            &snapshot.winning_stacks_block_hash,
-                        )
-                        .unwrap()
-                        .unwrap()
-                    }
-                };
-
-                let parent_header_hash = parent_tip.anchored_header.block_hash();
-                let parent_consensus_hash = parent_tip.consensus_hash.clone();
-
-                let mut mempool =
-                    MemPoolDB::open_test(false, 0x80000000, &chainstate_path).unwrap();
-
-                let coinbase_tx = make_coinbase(miner, 0);
-
-                for i in 0..10 {
-                    let stx_transfer = make_user_stacks_transfer(
-                        &privk,
-                        i,
-                        200,
-                        &recipient.to_account_principal(),
-                        1,
-                    );
-
-                    mempool
-                        .submit(
-                            chainstate,
-                            &parent_consensus_hash,
-                            &parent_header_hash,
-                            &stx_transfer,
-                            None,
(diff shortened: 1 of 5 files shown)


obycode:
Re-running my benchmarks today, this PR now looks good when compared to #3326 in benchmarking. 

_I was previously seeing better numbers for the other implementation, so it might be good for someone else to try benchmarking these and see if it is reproducible._

| Version | Commit | Num Tx | %-Full | Fees Collected (uSTX) | Time Spent (ms) |
| --- | --- | --- | --- | --- | --- |
| master | 378fc1b80e161ce1b82cc98b3802565222100816 | 105 | 4.97% | 25175605 | 33983.1 |
| #3326 | 6052447629436343d1ccd942a6a7d4381bfc670e | 1034 | 99.97% | 144993355 | 16822.8 |
| #3337 | 9a62096894b1d8de157e811384b7f44653a46a03 | 952 | 99.96% | 155684975 | 14121.9 |

obycode:
I made a mistake when refactoring this code and the retry list was being processed in the wrong order. The updated benchmarking numbers are here:

| Version | Commit | Num Tx | %-Full | Fees Collected (uSTX) | Time Spent (ms) |
| --- | --- | --- | --- | --- | --- |
| master | 378fc1b80e161ce1b82cc98b3802565222100816 | 105 | 4.97% | 25175605 | 33983.1 |
| #3326 | 6052447629436343d1ccd942a6a7d4381bfc670e | 1034 | 99.97% | 144993355 | 16822.8 |
| #3337 (with bug) | 9a62096894b1d8de157e811384b7f44653a46a03 | 952 | 99.96% | 155684975 | 14121.9 |
| #3337 | 91900935725d7bfe757f3afc2a1e220f44bbc526 | 972 | 99.95% | 131784486 | 15563.8 |

It's surprising that the fees collected gets worse when this bug is fixed. That must indicate that the cost estimate (and thus the fee rate) is off.

obycode:
Ah, I found another bug in this refactoring. Will fix and rerun the experiment.

obycode:
Ok, after the bug fixes, the benchmark results look better:

| Version | Commit | Num Tx | %-Full | Fees Collected (uSTX) | Time Spent (ms) |
| --- | --- | --- | --- | --- | --- |
| master | 378fc1b80e161ce1b82cc98b3802565222100816 | 105 | 4.97% | 25175605 | 33983.1 |
| #3326 | 6052447629436343d1ccd942a6a7d4381bfc670e | 1034 | 99.97% | 144993355 | 16822.8 |
| #3337 (with bugs) | 9a62096894b1d8de157e811384b7f44653a46a03 | 952 | 99.96% | 155684975 | 14121.9 |
| #3337 | 1586e86d20d8931c3e51804e337539f3b3c1d33a | 1097 | 100% | 189810290 | 15624.7 |

Now we get a 100% full block (read count is 15,000), and we get 44.82 STX more fees.

obycode:
The failing unit test, `core::tests::mempool_walk_over_fork` is failing because it is relying on behavior internal to the old design, the "last known nonces" in the mempool table. This design does not use those, so the test fails. I will look into whether it is worth re-writing the test, or if it should just be skipped.

obycode:
Updated table with new numbers from the latest version from #3326:

| Version | Commit | Num Tx | %-Full | Fees Collected (uSTX) | Time Spent (ms) |
| --- | --- | --- | --- | --- | --- |
| master | 378fc1b80e161ce1b82cc98b3802565222100816 | 105 | 4.97% | 25175605 | 33983.1 |
| #3326 | faffe420d8c5dee8151a988d0f8bc1b7ebb227de | 1034 | 99.97% | 144993355 | 16604.9 |
| #3337 | 1586e86d20d8931c3e51804e337539f3b3c1d33a | 1097 | 100% | 189810290 | 15624.7 |

src/core/mempool.rs
```diff
@@ -750,6 +756,113 @@ impl MemPoolTxInfo {
...
+        C: ClarityConnection,
+    {
+        match self.cache.get(address) {
+            Some(nonce) => *nonce,
+            None => {
+                let nonce = StacksChainState::get_nonce(clarity_tx, &address.clone().into());
+                // Simple size cap to the cache -- once it's full, all nonces
+                // will be looked up every time. This is bad for performance
+                // but is unlikely to occur due to the typical number of
+                // transactions processed before filling a block.
```
jcnelson, line 786: How often is this cache cleared?  Is it once per block?

Also, it is knowable how many addresses can be loaded per block -- we could, in theory, calculate the maximum number of transactions that could be mined in a block, and use that to derive a maximum number of addresses that block could touch.

src/core/mempool.rs
```diff
@@ -750,6 +756,113 @@ impl MemPoolTxInfo {
...
+        }
+    }
+
+    fn increment(&mut self, address: StacksAddress) {
+        let nonce = self.cache.entry(address).or_insert(0);
+        *nonce += 1;
+    }
+}
+
+struct CandidateCache {
```
jcnelson, line 802: Can you add a docstring to this struct?  Thanks!

src/core/mempool.rs
```diff
@@ -750,6 +756,113 @@ impl MemPoolTxInfo {
...
+    }
+
+    fn push(&mut self, tx: MemPoolTxInfoPartial) {
+        if self.size < Self::MAX_SIZE {
+            self.next.push_back(tx);
+            self.size += 1;
+        }
+    }
+
+    fn reset(&mut self) {
```
jcnelson, line 830: Could this cause `self.next` to grow unbound?  We aren't checking its length here, or in `Self::push`.

src/core/mempool.rs
```diff
@@ -1166,7 +1212,35 @@ impl MemPoolDB {
...
+            .map_err(|err| Error::SqliteError(err))?;
+        let mut null_iterator = query_stmt
+            .query(NO_PARAMS)
+            .map_err(|err| Error::SqliteError(err))?;
+
+        let sql = "
+            SELECT txid, origin_nonce, origin_address, sponsor_nonce, sponsor_address, fee_rate
+            FROM mempool
+            WHERE fee_rate IS NOT NULL
+            ORDER BY fee_rate DESC
```
jcnelson, line 1235: I was playing around with the mempool recently, and I noticed that adding `ORDER BY fee_rate` would slow down my queries considerably.  Can you check that we aren't doing a table scan with either of these queries via `EXPLAIN QUERY PLAN` on an instantiated database?  Thanks!

src/core/mempool.rs
```diff
@@ -1175,71 +1249,137 @@ impl MemPoolDB {
...
+                    // randomly selecting from either the null fee-rate transactions
+                    // or those with fee-rate estimates.
+                    let opt_tx = if start_with_no_estimate {
+                        null_iterator
+                            .next()
+                            .map_err(|err| Error::SqliteError(err))?
+                    } else {
+                        fee_iterator.next().map_err(|err| Error::SqliteError(err))?
+                    };
+                    match opt_tx {
```
jcnelson, line 1272: Regarding this `match` case, can you add a unit test if there isn't one already to verify that all transactions will be visited once, regardless of what `consider_no_estimate_tx_prob` is?  Thanks!

src/core/mempool.rs
jcnelson: It seems that the size of `self.next` can grow to `O(size of mempool)`, given that this method gets called once per iterator pass.  Is there a way to prevent this?

src/core/mempool.rs
jcnelson: Given how this cache gets used, it might make sense to just use a simple LRU strategy for now.  This cache is meant to help minimize the number of times we have to read the MARF to load up a nonce, so eliminating the most-common cases would be a good first attempt.

One day, subsequent refinement might consider the number of nodes that must be visited in the MARF to load the nonce.  If the address's nonce was recently changed, then there are fewer tries to visit.  It would make sense then to cache a nonce with probability proportional to how long ago it was last changed.  It doesn't have to be in this PR, but I'm flagging it here for consideration.

jcnelson (commented):
Thanks for this PR @obycode!

One thing that I think we'll want to see before merging is some test coverage that verifies the following:

* All transactions are visited at least once on iteration, assuming no time-outs and no out-of-space events occur.

* We need to see what happens when there are more mempool transactions than the caches have space (maybe the cache sizes could be configurable?).  In particular, the test should verify that the caches reduce the number of I/O operations predictably, given their size.  The cache could track this information in some internal accounting state.

* We'll want to know what a good default cache size is for when the chain is under load.  I think this could be obtained with live-testing with the mock miner, but it would be ideal if there was a  unit test that could show us how to deduce this (or possibly a way for mempool to figure out for us what a good size would be).

* We'll need to verify in a unit test that the RAM usage does not increase beyond a configured constant.  I don't think this is currently happening in the code -- I think you have at least one instance of unbound memory usage.  The mempool can have an unbound number of unmineable transactions, so we'll want to make sure the cache doesn't accidentally eat up all the RAM in the process of iteration.

src/core/mempool.rs
```diff
@@ -1175,71 +1249,137 @@ impl MemPoolDB {
...
+                        Some(row) => (MemPoolTxInfoPartial::from_row(row)?, start_with_no_estimate),
+                        None => {
+                            // If the selected iterator is empty, check the other
+                            match if start_with_no_estimate {
+                                fee_iterator.next().map_err(|err| Error::SqliteError(err))?
```
gregorycoppola, line 1277: does this fail if there is no estimate?

src/core/mempool.rs
obycode: Ah, I see my logic was off here -- size shouldn't have been reset to `0`. I'll just check the size of `self.next`. Also, I checked and saw that `VecDeque` does have an efficient `len()` method, so I'll just use that and remove the `size` field.

src/core/mempool.rs
obycode: With fee rate:
```
QUERY PLAN
`--SEARCH mempool USING INDEX by_fee_rate (fee_rate>?)
```
Without fee rate:
```
QUERY PLAN
`--SEARCH mempool USING INDEX by_fee_rate (fee_rate=?)
```

src/core/mempool.rs
obycode: It only returns an error if there is a sqlite error. It returns `None` if there are no more transactions left (that would mean it's already checked both with and without fee-rates and got no results.

src/core/mempool.rs
gregorycoppola: nice

src/core/mempool.rs
jcnelson: @gregorycoppola I believe that because this cache is key to mempool iteration performance, we should get it checked in with this PR.

src/core/mempool.rs
jcnelson: @obycode I'm also saying that `.reset()` should check the size of `self.next`.  The `.append()` function will add every item in `self.cache` to `self.next`, which could push `self.next` to be well beyond `Self::MAX_SIZE`.  I think this is something we'd want to address in `.reset()`.

src/core/mempool.rs
jcnelson: thanks!

src/core/mempool.rs
```diff
@@ -1175,71 +1249,137 @@ impl MemPoolDB {
...
+                expected_origin_nonce,
+                candidate.sponsor_nonce,
+                expected_sponsor_nonce,
+            ) {
+                Ordering::Less => {
+                    debug!(
+                        "Mempool: unexecutable: drop tx ({})",
+                        candidate.fee_rate.unwrap_or_default()
+                    );
+                    // This transaction cannot execute in this pass, just drop it
```
jcnelson, line 1311: Since we're doing this check in multiple passes, and since a transaction can go from having the `::Greater` condition, to having the `::Equal` condition, to having the `::Less` condition, would it make sense here to also evict state from the nonce cache for this candidate?

src/core/mempool.rs
jcnelson: @obycode On cache miss, could you instead write the last-known nonce to the mempool DB so we don't hit the MARF more than once per candidate?

jcnelson (commented):
So, one thing about this PR that still gives me pause is that once the caches are full, there's no eviction strategy.  This means that it's possible for the miner to exhibit a pathological behavior where the first `NonceCache::MAX_SIZE` transactions are considered but are unmineable.  Then, once we find the first mineable transaction, we'll always be encountering a cache miss on `NonceCache::get()`, which incurs a MARF read or (if you agree with my comment above) a database read.

As an easy-to-implement stop-gap to avoid thrashing, could you make the cache sizes configurable, and then plumb through that configuration from the node's config file?  Then, at least miners could set higher `MAX_SIZE` values if they had enough RAM for it.

gregorycoppola:
I think this PR needs more tests.

I can take it over and write the tests that people want.

I opened a discussion https://github.com/stacks-network/stacks-blockchain/discussions/3345 that depicts the mempool walk as a pipeline of transformations as follows. These are some levels at which we can add tests, in whichever style.
dag-representation

src/core/mempool.rs
gregorycoppola: What if we just cap the cache size, and not do replacement?

That would be the simplest and provide the same benefit we have seen up to this point.

obycode: Yes, this cache is cleared once per block.

> we could, in theory, calculate the maximum number of transactions that could be mined in a block, and use that to derive a maximum number of addresses that block could touch.

I don't think we can come up with a maximum number this way, because the number of addresses that it could touch is not bound by the number of transactions in the block, but by the number of transactions that need to be considered for the block. That would include pending transactions, and the number of those is only capped by the number of possible addresses.

obycode: > On cache miss, could you instead write the last-known nonce to the mempool DB so we don't hit the MARF more than once per candidate on a call to iterate_candidates()?

Yes, we can do this. I'll benchmark again to see what kind of effect this has on performance in the normal case.

gregorycoppola: LRU or no, I think you can bound transactions at around 256k using:

```
pub const MEMPOOL_MAX_TRANSACTION_AGE: u64 = 256;
```

And a calculation that around `1000` transactions fill a block.

So, the most number of transactions you would "expect" to crawl through before hitting a full block would be `256*1000=256k`.

src/core/mempool.rs
obycode: Comment added in 27ee2f88c

src/core/mempool.rs
obycode: Added in 27ee2f88c

src/core/mempool.rs
```diff
@@ -750,6 +756,128 @@ impl MemPoolTxInfo {
     }
 }
 
+/// Used to locally cache nonces to avoid repeatedly looking them up in the nonce.
+struct NonceCache {
+    cache: HashMap<StacksAddress, u64>,
+    size: usize,
```
gregorycoppola, line 762: why is `size` separate? getting the size of a `HashMap` should be cheap

obycode: Yeah, good point. I think that is some artifact of this struct getting simplified, or my lack of trust for the implementation 😄 (I should know better with Rust).

obycode: Removed in 56d0050d2

src/core/mempool.rs
obycode: Added in fa95c06d43cec7a3c69a04df558c104756da3ac7

src/core/mempool.rs
obycode: I'm not sure what happened to that earlier commit. It is removed again in 5cd1c6194

obycode:
Latest status:
* @gregorycoppola implemented the settings for cache sizes
* @gregorycoppola and I added tests for the basic cases
* I'm going to need to think some more tomorrow about setting the cache sizes and what to do when they fill up

src/core/mempool.rs
gregorycoppola: i was assuming that, on average, 1/256 transactions would have the right nonce.. i think that in practice the "proportion ready to mine" would actually go up and down a bit.

gregorycoppola: also... the height that's used for eviction is independent of chaining, `MemPoolDB::tx_submit`... it just goes by 256 blocks since the height the tx was submitted, afaiu.

either way, this is just a model.

gregorycoppola:
I pushed a unit test that parameterizes over:
* num users
* num rounds
* cache size
* selection probability

But, the problem with this test and, I think, `test_iterate_candidates_consider_no_estimate_tx_prob` is that these do not currently toggle whether or not the estimates are null in the db (?).. I'm checking that part, but wanted to push intermediate version for visibility/feedback.

obycode:
That sounds great, thanks Greg!

For `test_iterate_candidates_consider_no_estimate_tx_prob`, it's doing just a simple back and forth between setting a fee rate or not:

```
if nonce & 1 == 0 {
            mempool_tx
                .execute(
                    "UPDATE mempool SET fee_rate = ? WHERE txid = ?",
                    rusqlite::params![Some(123.0), &txid],
                )
                .unwrap();
        } else {
            let none: Option<f64> = None;
            mempool_tx
                .execute(
                    "UPDATE mempool SET fee_rate = ? WHERE txid = ?",
                    rusqlite::params![none, &txid],
                )
                .unwrap();
        }
```

obycode:
Doing some experimentation using real world mempool and I am surprised by these results. I would expect the read count and fees collected to be the same regardless of the nonce cache size. It makes sense that these would change when the candidate cache limit is reached. I am digging in to see what is happening.

| NonceCache Size | CandidateCache Size | Runtime (ms) | Read Count | Fees Collected (STX) |
| --- | --- | --- | --- | --- |
| 0 | 0 | 19429 | 14988 | 141.553460 |
| 0 | 64*1024 | 39172 | 14998 | 146.662172 |
| 1024*1024 | 0 | 18904 | 14995 | 144.993355 |
| 1024*1024 | 64*1024 | 18286 | 15000 | 189.810290 |

jcnelson:
I'm also surprised that the runtime would bounce so high when the candidate cache is enabled but the nonce cache is disabled.  Where are those extra 10s coming from?

gregorycoppola:
@obycode  look at https://github.com/stacks-network/stacks-blockchain/pull/3337/commits/31c3ee40574b614d0935372dce73c6565aae85d3

it has a bug for the "don't overflow max cache size" thing.. but also there might have been a bug where you assume the default is "0" on `increment`, instead of what was in the marf

obycode:
Benchmarking numbers after Greg's fix:

| NonceCache Size | CandidateCache Size | Runtime (ms) | Read Count | Fees Collected (STX) |
| --- | --- | --- | --- | --- |
| 0 | 0 | 19740 | 14995 | 144.993355 |
| 0 | 64*1024 | 36682 | 15000 | 189.810290 |
| 1024*1024 | 0 | 19205 | 14995 | 144.993355 |
| 1024*1024 | 64*1024 | 18151 | 15000 | 189810290 |

Thanks for that fix @gregorycoppola ! These numbers make sense now. 

@jcnelson when the candidate cache is enabled but the nonce cache is not, it'll one hitting the MARF over and over again for the same transactions, so that additional time makes sense.

obycode:
I like the fact that in the pathological case where the number of transactions that cannot be mined  yet is large enough to overwhelm the caches, all users need to do is submit new transactions with higher fee-rates than those unmineable transactions, and the slower behavior is avoided.

obycode:
I would be happy to merge this now. Or I can first implement writing the nonces to the mempool db when the mempool cache is full. What do you think @gregorycoppola, @jcnelson?

obycode:
I don't see any better eviction strategy, because I think it makes the most sense to keep the highest fee-rate transactions in the cache. We know that those will be visited every iteration.

jcnelson:
> @obycode I like the fact that in the pathological case where the number of transactions that cannot be mined yet is large enough to overwhelm the caches, all users need to do is submit new transactions with higher fee-rates than those unmineable transactions, and the slower behavior is avoided.

> @obycode I don't see any better eviction strategy, because I think it makes the most sense to keep the highest fee-rate transactions in the cache. We know that those will be visited every iteration.

The case I'm worried about is that someone could effectively force a chronic state of mempool congestion by sending lots of high-fee unmineable transactions.  This wouldn't cost the attacker anything (since they're unmineable), so they could keep the mempool in this state forever, negating the performance boost produced by the caches.

This is why I'm advocating for using the mempool DB to store the last-known origin and sponsor nonces for an address in the `NonceCache::get()` cache-miss path.  This effectively creates a two-tier cache:  the first layer keeps nonces in RAM, and the second layer loads them from the DB.  The cache-miss path would load the nonce from the MARF, store it to the DB, and then try and store it to RAM.  The nonce cache in the DB would be preserved across calls to `iterate_candidates`, and would be cleared whenever `iterate_candidates` is called on a different parent block than it had in the last call.

jcnelson:
I can take a crack at this if you're pressed for time.

obycode:
Ok, I hadn't thought about the fact that it wouldn't cost this attacker anything to do that. Good point.

I'll implement the mempool DB cache.

obycode:
> I can take a crack at this if you're pressed for time.

No, I'll do it. This is my top priority now.

obycode:
I may need to experiment with whether we should write the nonce to all transactions in the DB that require it, or just the one transaction.

jcnelson:
> @obycode I may need to experiment with whether we should write the nonce to all transactions in the DB that require it, or just the one transaction.

I was thinking there could simply be a table of `(address, nonce)` pairs?

obycode:
Ok, I was wondering if we should cache them in the `last_known_origin_nonce`, `last_known_sponsor_nonce` fields currently in the table from the previous implementation. Then when we read in a new transaction, we can just add those two fields to the simplified transaction struct we're using.

jcnelson:
Either approach is fine with me.  However, if you reuse the last_known nonces, you'll need to clear them whenever you call `iterate_candidates()` with a new parent Stacks block.

obycode:
Right, I'll basically be adding back in some of that old code.

A little stream of consciousness comparing the strategies...
- In the method with a simple table of (address, nonce) pairs, when the in-memory cache is full, we will be pull from sqlite for every transaction twice (once to load the transaction, once to load the nonces or in the case of a miss, once to write the nonces after loading from the MARF).
- In the method with `last_known_*_nonce`, we can will pull from sqlite for every transaction once, because we can include these nonces when we load the transaction. If the `last_known_*_nonce` is not there, then we would check the in-memory cache, and if it misses there as well, then we would load from the MARF, and do a more expensive write to the mempool DB that writes to all transactions using that address.

I guess in the second case, the in-memory cache is not really doing anything for us, is it? There is no reason for a nonce to be in the in-memory cache and not in the mempool db. The in-memory nonce cache becomes part of the in-memory transaction cache in that case, because we include those fields in the candidate transactions.

jcnelson:
> @obycode I guess in the second case, the in-memory cache is not really doing anything for us, is it? There is no reason for a nonce to be in the in-memory cache and not in the mempool db. The in-memory nonce cache becomes part of the in-memory transaction cache in that case, because we include those fields in the candidate transactions.

No, the nonce will _usually_ be in memory if it's not in the mempool DB.

The in-memory cache operates on a per-_address_ basis.  The last_known nonces are cleared whenever you call `iterate_candidates()` with a different parent block from the last call.  So if you had multiple transactions that use the same address (regardless if it's a sender or sponsor), their last_known nonces would all be `None`.  Using the in-memory cache ensures that you only look up the address's nonce _once_ while the cache has space, so that when you go and load the nonces, you can often hit the RAM cache before you hit the MARF.

obycode:
>  It might be faster, especially if the mempool is really full, because you'll touch fewer rows.

Yup, I tested this, and it's too slow. I'm going to implement the simple pairs now.

gregorycoppola:
I added a parameterized test framework called `paramaterized_mempool_walk_test`, which is based on @obycode 's version that calls `iterate_candidates` directly.

At this point, some of the tests are redundant.

But, the tests should all pass no matter which cache strategies you use.

gregorycoppola:
some of the tests are failing now.. not sure if it's a bug in the test or real, but it's when the cache is tight.. when the cache is large enough to not matter, the tests are passing.

```
running 9 tests
test chainstate::stacks::miner::test::mempool_walk_test_users_1_rounds_10_cache_size_2_null_prob_100 ... FAILED
test chainstate::stacks::miner::test::mempool_walk_test_users_1_rounds_10_cache_size_2_null_prob_0 ... FAILED
test chainstate::stacks::miner::test::mempool_walk_test_users_10_rounds_3_cache_size_2_null_prob_50 ... FAILED
test chainstate::stacks::miner::test::mempool_walk_test_users_1_rounds_10_cache_size_2_null_prob_50 ... FAILED
test chainstate::stacks::miner::test::mempool_walk_test_users_10_rounds_3_cache_size_2_null_prob_0 ... FAILED
test chainstate::stacks::miner::test::mempool_walk_test_users_10_rounds_3_cache_size_2_null_prob_100 ... FAILED
test chainstate::stacks::miner::test::mempool_walk_test_users_10_rounds_3_cache_size_2000_null_prob_0 ... ok
test chainstate::stacks::miner::test::mempool_walk_test_users_10_rounds_3_cache_size_2000_null_prob_100 ... ok
test chainstate::stacks::miner::test::mempool_walk_test_users_10_rounds_3_cache_size_2000_null_prob_50 ... ok
```

obycode:
After adding the second-level cache for the nonces, I'm getting these benchmarking numbers.

_Note, I'm not sure why this is actually faster than the version without the secondary cache in the case where the cache should be plenty big enough._

| NonceCache Size | CandidateCache Size | Runtime (ms) | Read Count | Fees Collected (STX) |
| --- | --- | --- | --- | --- |
| 0 | 0 | 15203 | 14995 | 144.993355 |
| 0 | 64*1024 | 25729 | 15000 | 189.810290 |
| 1024*1024 | 0 | 14798 | 14995 | 144.993355 |
| 1024*1024 | 64*1024 | 14384 | 15000 | 189.810290 |

obycode:
The new failing unit tests are a problem with the tests. When the cache is full, a single pass of `iterate_candidates` will not capture all of the transactions that could be included in the block if they are out of order. I'll try to quickly modify the test.

src/core/mempool.rs
```diff
@@ -1129,19 +1300,30 @@ impl MemPoolDB {
...
     /// Iterate over candidates in the mempool
-    ///  `todo` will be called once for each transaction whose origin nonce is equal
-    ///  to the origin account's nonce. At most one transaction per origin will be
-    ///  considered by this method, and transactions will be considered in
-    ///  highest-fee-first order.  This method is interruptable -- in the `settings` struct, the
-    ///  caller may choose how long to spend iterating before this method stops.
+    /// `todo` will be called once for each transaction that is a valid
+    /// candidate for inclusion in the next block, meaning its origin and
+    /// sponsor nonces are equal to the nonces of the corresponding accounts.
+    /// Best effort will be made to process the transactions in fee-rate order
```
gregorycoppola, line 1307: i think it is guaranteed to work before the cache is full right?

like "Best effort will be made to process the transactions in fee-rate order. That is, transactions will be processed in fee-rate order until the candidate cache is full, ..."

obycode:
Ok, I did another pass over this PR and I think it looks good, but I have one question. We are currently bumping the nonces in the caches every time a transaction is considered. That appears to be the behavior from the existing implementation as well, but is that correct? Should the nonces get bumped when the result is `Skipped` or `Problematic`?

jcnelson:
> @obycode Should the nonces get bumped when the result is Skipped or Problematic?

I don't think so.  If a transaction from origin address A at nonce N can't be mined, then neither can any transaction from A with nonce N+k.

obycode:
> I don't think so. If a transaction from origin address A at nonce N can't be mined, then neither can any transaction from A with nonce N+k.

Thanks, that was my thought as well. What about `ProcessingError`? Does that indicate that the transaction was included in the block with an error or something else? The comment says "It may succeed later depending on the error" which makes me unsure without digging into the code.

obycode:
Looks like we should only bump the nonces on `Success`.

src/core/mempool.rs
```diff
@@ -241,10 +242,16 @@ pub struct MemPoolTxInfo {
...
 }
 
-#[derive(Debug, PartialEq, Clone)]
-pub enum MemPoolTxInfoPartial {
-    NeedsNonces { addrs_needed: Vec<StacksAddress> },
-    HasNonces(MemPoolTxInfo),
+/// This class is a minimal version of `MemPoolTxInfo`. It contains
+/// just enough information to 1) filter by nonce readiness, 2) sort by fee rate.
+#[derive(Debug, Clone)]
+pub struct MemPoolTxInfoPartial {
```
jcnelson, line 248: Are `origin_nonce` and `sponsor_nonce` still used in this struct?  Are they still used anywhere in the mempool?

If not, then can you add a comment to indicate that they'll be removed in the near future, and open an issue?  No need to block the PR on it, given that we're pressed for time.

src/core/mempool.rs
```diff
@@ -750,6 +777,198 @@ impl MemPoolTxInfo {
...
+/// In the next pass, `next` is moved to `cache` and these transactions are
+/// checked before reading more from the mempool DB.
+struct CandidateCache {
+    cache: VecDeque<MemPoolTxInfoPartial>,
+    next: VecDeque<MemPoolTxInfoPartial>,
+    /// The maximum size that this cache can be.
+    max_cache_size: usize,
+}
+
+impl CandidateCache {
```
jcnelson, line 889: Is `CandidateCache::cache` ever written to in this cache?  Does `CandidateCache::next()` ever return `Some(..)`?

obycode: > Is CandidateCache::cache ever written to in this cache? 
`CandidateCache::cache` is only written to in `reset`.

> Does CandidateCache::next() ever return Some(..)?
Yes, see previous answer

jcnelson: Ah, yes, I see that now.

src/core/mempool.rs
```diff
@@ -750,6 +777,198 @@ impl MemPoolTxInfo {
...
+
+    /// Push a candidate to the cache for the next iteration.
+    fn push(&mut self, tx: MemPoolTxInfoPartial) {
+        if self.next.len() < self.max_cache_size {
+            self.next.push_back(tx);
+        }
+
+        #[cfg(test)]
+        {
+            assert!(self.cache.len() <= self.max_cache_size + 1);
```
gregorycoppola, line 914: does this really need to be `+1`? i might have just put that when hacking.

src/chainstate/stacks/miner.rs
```diff
@@ -1010,7 +1010,7 @@ impl<'a> StacksMicroblockBuilder<'a> {
         let deadline = get_epoch_time_ms() + (self.settings.max_miner_time_ms as u128);
         let mut block_limit_hit = BlockLimitFunction::NO_LIMIT_HIT;
 
-        mem_pool.reset_last_known_nonces()?;
```
jcnelson, line 1013: We're not using any of the methods that manipulate the last_nonces columns, right?  If so, then can you delete those as well, and add a comment to the schema that added these columns that they are no longer used?

jcnelson (approved):
Overall this LGTM!  Thank you for seeing this through @obycode @gregorycoppola!

Do you have new benchmark numbers?  Can you instantiate this on a mock miner and see how it does?

src/chainstate/stacks/miner.rs
obycode: Sure. Just to be clear, you want to remove those columns from the DB in the latest schema in addition to deleting all of the related code?

jcnelson: If it's not too much trouble -- i.e. if it can be done with a `DROP COLUMN`.  Sqlite has a bunch of constraints on when you can and cannot do this, however, so don't worry about it if sqlite is preventing you.

gregorycoppola (approved):
assuming the mock miner and benchmarks are as expected, i approve.

thanks everyone!

obycode:
> Do you have new benchmark numbers?

I've run informally, but will collect new numbers now.

> Can you instantiate this on a mock miner and see how it does?

Yes, it is running now. Last block:
```
INFO [1666196520.482998] [src/chainstate/stacks/miner.rs:2413] [relayer] Miner: mined anchored block, block_hash: bdb4cae9809c7021c4bf1b6b2b7ce498543a2018
e0724dea5bb24c4e9c14caf3, height: 80211, tx_count: 72, parent_stacks_block_hash: f7d4afe1744f540899486d1c8811a3e42f22aace754dab19b59664e29722444e, parent_
stacks_microblock: 051d171d080744f083318cde80ab998774756632ee68da6d91cc86604374d3ca, parent_stacks_microblock_seq: 0, block_size: 21224, execution_consume
d: {"runtime": 64230390, "write_len": 32388, "write_cnt": 626, "read_len": 12103538, "read_cnt": 4363}, %-full: 29, assembly_time_ms: 2175, tx_fees_micros
tacks: 1054503
```

src/chainstate/stacks/miner.rs
obycode: I'm seeing:

```
thread 'main' panicked at 'Failed to open mempool db: SqliteError(SqliteFailure(Error { code: Unknown, extended_code: 1 }, Some("near \"DROP\": syntax error")))', src/main.rs:739:14
```

But strangely, I can run the same query from the command line and it works.

```
ALTER TABLE mempool DROP COLUMN last_known_origin_nonce;
```

obycode: Maybe sqlite allows it but `rusqlite` does not?

obycode:
Latest benchmarking numbers:

| Version | Commit | Num Tx | %-Full | Fees Collected (uSTX) | Time Spent (ms) |
| --- | --- | --- | --- | --- | --- |
| #3337 | 35b38cb797d78a7ee4544fa0dd16432a6cda7258 | 1097 | 100% | 189810290 | 10736.7 |

obycode:
@jcnelson those failing unit tests are due to the fact that they were depending on the fact that the old `iterate_candidates` was incrementing the nonces on a `Skipped` result. I will update the tests, but I want to make sure that the new behavior is correct. It seems correct to me.

obycode:
Tests updated in 1a65f0f0c.

obycode:
I am adding some unit tests to test the behavior with a skipped or problematic transaction.

obycode:
New unit tests added in 36690d05655181e191c9fe5dc47658a37f5d93ca. `master` fails these unit tests which I believe would've caused poor behavior -- repeatedly selecting bad transactions as candidates.

src/core/tests/mod.rs
```diff
@@ -730,6 +789,361 @@ fn test_iterate_candidates_consider_no_estimate_tx_prob() {
     );
 }
 
+#[test]
+/// This test verifies that when a transaction is skipped, other transactions
+/// from the same address with higher nonces are not included in a block.
```
jcnelson, line 794: Specifically, we want it to be the case that these higher-nonce transactions aren't even _considered_.

jcnelson: No action needed on this, btw.  I'm happy to take it once I merge this to #3335

obycode: That is what the test is checking, but you're right, the comment should be more clear. Thanks for handling that!

PR #3339, opened 2022-10-13
Btc wire format for delegate stx + use active PoX contract for btc ops

### LINKED ISSUE #3171: [chainstate] use the right PoX contract name when processing on-BTC stacking ops
Right now, `StacksChainState::process_stacking_ops()` is hardcoded to use `.pox` as the contract name for calling `stack-stx` from an on-BTC stacking transaction.  This contract name needs to switch over to `.pox-2` automatically when it goes live.

### LINKED ISSUE #3272: [2.1] Add a `delegate-stx` bitcoin wire format transaction
The new features of PoX-2 in Stacks 2.1 should be great.

_However_, people using the bitcoin wire format (which isn't a small total number: it's about 10% of the liquid supply of STX), wouldn't be able to use these features without changing their wallet setup (people using 2-of-n wallets could use https://github.com/kantai/multisig-cli as long as they are willing to use a ledger device).

Adding a `delegate-stx` bitcoin wire format transaction, though, would enable them to access `pox-2` features by delegating to a normal STX address (or another multisig address), and using the delegation support there.

pavitthrap:
### Description
This PR adds a delegate-stx BTC wire format transaction.

It also ensures that the active pox contract is correctly set for stacking operations: for the `stack-stx` wire format transaction, the node now calls the "stack-stx" function in the current active pox contract. Before, it would only call the "stack-stx" function from the "pox" contract. 

### Applicable issues
- fixes #3272 
- fixes #3171


src/chainstate/burn/db/sortdb.rs
```diff
@@ -2667,6 +2733,11 @@ impl SortitionDB {
                         let tx = self.tx_begin()?;
                         SortitionDB::apply_schema_3(&tx.deref())?;
                         tx.commit()?;
+                    } else if version == "3" {
+                        // add the tables of schema 4, but do not populate them.
+                        let tx = self.tx_begin()?;
+                        SortitionDB::apply_schema_4(&tx.deref())?;
+                        tx.commit()?;
```
jcnelson, line 2740: Don't forget to also update ` is_db_version_supported_in_epoch()` as well!

src/chainstate/burn/operations/delegate_stx.rs
```diff
@@ -0,0 +1,290 @@
...
+
+            0      2  3                19           28
+            |------|--|----------------|------------|
+             magic  op delegated ustx   until burn height
+                                                              
+
+             Note that `data` is missing the first 3 bytes -- the magic and op have been stripped
+        */
+        // magic + op are omitted
+        if data.len() < 26 {
```
jcnelson, line 69: Shouldn't this be 25?

src/chainstate/burn/operations/delegate_stx.rs
```diff
@@ -0,0 +1,290 @@
...
+                data.len()
+            );
+            return None;
+        }
+
+        let delegated_ustx = parse_u128_from_be(&data[0..16]).unwrap();
+
+        let until_burn_height = {
+            if data[16] == 1 {
+                let burn_height = parse_u64_from_be(&data[17..25]).unwrap();
```
jcnelson, line 82: You'd want to check here that `data.len() >= 26`.

src/chainstate/burn/operations/delegate_stx.rs
```diff
@@ -0,0 +1,290 @@
...
+    /*
+            Wire format:
+
+        0      2  3                     24               40 41       50
+        |------|--|---------------------|----------------|--|--------|
+         magic  op  delegate to           delegated ustx  opt       until burn height
+                                                          
+
+    */
+    fn consensus_serialize<W: Write>(&self, fd: &mut W) -> Result<(), codec_error> {
```
jcnelson, line 184: I think this is wrong.  You'll want to only encode the data in the `OP_RETURN` here (you can omit `delegate_to`).

src/chainstate/burn/operations/delegate_stx.rs
```diff
@@ -0,0 +1,290 @@
...
+        write_next(fd, &(Opcodes::DelegateStx as u8))?;
+        self.delegate_to.consensus_serialize(fd)?;
+        fd.write_all(&self.delegated_ustx.to_be_bytes())
+            .map_err(|e| codec_error::WriteError(e))?;
+        write_next(fd, &(self.reward_addr.is_some() as u8))?;
+        write_next(fd, &(self.until_burn_height.is_some() as u8))?;
+
+        // Q-JUDE
+        if let Some(height) = self.until_burn_height {
```
jcnelson, line 193: I think you're missing an `fd.write_all(&(1 as u8))` here to mark the presence of an 8-byte height to follow.

src/chainstate/burn/operations/delegate_stx.rs
```diff
@@ -0,0 +1,290 @@
...
+            fd.write_all(&height.to_be_bytes())
+                .map_err(|e| codec_error::WriteError(e))?;
+        } else {
+            fd.write_all(&(0 as u64).to_be_bytes())
```
jcnelson, line 197: Shouldn't this just be a 1-byte 0?

src/chainstate/burn/operations/delegate_stx.rs
```diff
@@ -0,0 +1,290 @@
...
+        assert_eq!(
+            &op.reward_addr,
+            &Some(PoxAddress::Standard(
+                StacksAddress::from_bitcoin_address(&tx.outputs[1].address),
+                Some(AddressHashMode::SerializeP2PKH)
+            ))
+        );
+        assert_eq!(op.delegated_ustx, u128::from_be_bytes([1; 16]));
+        assert_eq!(op.delegate_to, StacksAddress::new(22, Hash160([1u8; 20])));
+        assert_eq!(op.until_burn_height, Some(u64::from_be_bytes([1; 8])));
```
jcnelson, line 288: Can your test use different bit-patterns for the OP_RETURN fields, so we can be sure they're actually getting loaded correctly?

Also, can you add tests for malformed OP_RETURNs, as well as tests for when the `until_burn_height` is not set, as well as for both when the PoX address is and is not specified?

src/chainstate/burn/operations/delegate_stx.rs
```diff
@@ -0,0 +1,290 @@
...
+        if data.len() < 26 {
+            // too short to have required data
+            warn!(
+                "DELEGATE_STX payload is malformed ({} bytes)",
+                data.len()
+            );
+            return None;
+        }
+
+        let delegated_ustx = parse_u128_from_be(&data[0..16]).unwrap();
```
jcnelson, line 78: This value can't be 0, right?  Otherwise it's a parse error?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4628,7 +4630,7 @@ impl StacksChainState {
                 tx.run_contract_call(
                     &sender.into(),
                     None,
-                    &boot_code_id("pox", mainnet),
+                    &boot_code_id(active_pox_contract, mainnet),
```
jcnelson, line 4633: Lol this fixes #3171 :)

src/chainstate/stacks/db/blocks.rs
```diff
@@ -11875,8 +12008,8 @@ pub mod test {
                 let chainstate = peer.chainstate();
                 let (mut chainstate_tx, clarity_instance) =
                     chainstate.chainstate_tx_begin().unwrap();
-                let (stack_stx_ops, transfer_stx_ops) =
-                    StacksChainState::get_stacking_and_transfer_burn_ops_v210(
+                let (stack_stx_ops, transfer_stx_ops, delegate_stx_ops) =
+                    StacksChainState::get_stacking_and_transfer_and_delegate_burn_ops_v210(
```
jcnelson, line 12012: Can you expand this test to cover expired delegate-stx on-Bitcoin operations?  Thanks!

jcnelson:
Since you also appear to be addressing #3171, can you add an integration test to verify that the right PoX address gets used when doing a stack-stx or delegate-stx in epoch 2.1 on Bitcoin?  Thanks!

src/chainstate/burn/db/sortdb.rs
pavitthrap: Just did - does it make sense to set this to "4" for epoch 2.1?

src/chainstate/stacks/db/blocks.rs
pavitthrap: Didn't realize this was already an issue!

src/chainstate/burn/operations/delegate_stx.rs
pavitthrap: Oh, that makes sense.

src/chainstate/burn/operations/delegate_stx.rs
pavitthrap: I'm not exactly sure what you mean by malformed OP_RETURNs... Do you mean an incorrect opcode? Do you mean an malformed output?

pavitthrap: It also doesn't seem possible to construct an output that would trigger an error in the parsing code.

src/chainstate/burn/operations/delegate_stx.rs
jcnelson: Ah, yup, you're right.  It can be 0.

src/chainstate/coordinator/tests.rs
```diff
@@ -2287,6 +2289,319 @@ fn test_stx_transfer_btc_ops() {
...
+        // The chain will look something like this 
+        // B0 -> B1 -> B2 -> ... B20 -> B21 
+        //                          \ _ B22
+        //                          \ _ B23
+        //                          \ _ B24
+        //                            ....
+        //                          \ _ B30 -> B31 -> ...
+        let parent = if ix == 0 {
+            BlockHeaderHash([0; 32])
+        } else if ix >= 22 && ix <= 30 {
```
jcnelson, line 2368: I think you want `ix >= 21` here.  Otherwise, with `ix >= 22`, blocks 22 through 30 all build off of `ix == 21`, in which the delegation for `delegate_amt * 2` will have taken effect.

src/chainstate/coordinator/tests.rs
```diff
@@ -2287,6 +2289,319 @@ fn test_stx_transfer_btc_ops() {
...
+                delegate_to: recipient_addr.clone(),
+                reward_addr: None,
+                delegated_ustx: delegated_amt,
+                until_burn_height: None,
+                txid: next_txid(),
+                vtxindex: 5,
+                block_height: 0,
+                burn_header_hash: BurnchainHeaderHash([0; 32]),
+            }));
+        } else if ix == 20 {
```
jcnelson, line 2445: Note: `ix == 20` means that the on-BTC operation takes effect in the Stacks block mined at `ix == 21`.

src/chainstate/coordinator/tests.rs
```diff
@@ -2287,6 +2289,319 @@ fn test_stx_transfer_btc_ops() {
...
+        }
+
+        // check our delegated balance
+        if ix > 0 {
+            let stacks_tip =
+                SortitionDB::get_canonical_stacks_chain_tip_hash(sort_db.conn()).unwrap();
+            let mut chainstate = get_chainstate(path);
+            let parent_tip = StacksBlockId::new(&stacks_tip.0, &stacks_tip.1); 
+            
+            let computed_delegation_amt = if ix > 11 {
```
jcnelson, line 2466: I think this should be `ix >= 11`.

src/chainstate/coordinator/tests.rs
```diff
@@ -2287,6 +2289,319 @@ fn test_stx_transfer_btc_ops() {
...
+                let del_addr = if ix < 22 {
```
jcnelson, line 2467: I think this should be `ix < 21`.

src/chainstate/coordinator/tests.rs
```diff
@@ -2287,6 +2289,319 @@ fn test_stx_transfer_btc_ops() {
...
+                                    .eval_raw(&eval_str)
+                                    .unwrap();
+                                Ok(result)
+                            }
+                        )
+                        .unwrap()
+                    })
+                    .unwrap();
+
+                if ix > 31 {
```
jcnelson, line 2498: I think this should be `ix > 27`.

src/chainstate/coordinator/tests.rs
```diff
@@ -2287,6 +2289,319 @@ fn test_stx_transfer_btc_ops() {
...
+                    delegated_amt,
+                    "Lock should be active"
+                );
+            } 
+
+            // Check that the effects of the delegate stx op sent when ix==10 
+            // are materialized for ix=22..28 (n to n+6), where each of these 
+            // blocks fork off of the state from iteration ix=21. 
+            // Want to ensure that 
+            if ix >= 22 {
```
jcnelson, line 2530: I think this should be `ix >= 21 && ix < 27`.

src/chainstate/coordinator/tests.rs
```diff
@@ -2287,6 +2289,319 @@ fn test_stx_transfer_btc_ops() {
...
+                assert_eq!(
+                    computed_delegation_amt,
+                    delegated_amt*2,
+                    "Lock should be active"
+                );
```
jcnelson, line 2535: You'll want a check that `computed_delegation_amt` is 0 when `ix >= 27`.

src/chainstate/coordinator/tests.rs
jcnelson: The `delegate-stx` op materializes in `ix` blocks 21 through 27, because you send it at `ix == 20`.  So, blocks 21 through 27 must all be siblings in order to test that the burnchain operations can manifest in all Stacks blocks mined between n+1 and n+7 burnchain blocks after the burnchain operation confirms in burnchain block n.

pavitthrap: The weirdness in the indexing here is caused by the fact that the logic for the check is happening before the burnchain/stacks block is being handled. 

The logic takes the following structure, which makes it difficult to follow: 
- Create burnchain ops 
- Check the delegation amount state
- Handle the burnchain and stacks block 

ix = 20, the delegate stx burnchain op created and handled. The op is sent in the burnchain block created in this iteration.
In ix = 21, the check for delegation state occurs before the stacks block is handled. That's why I do the check in the next iteration (which I agree is confusing). I'm moving the delegation state check to after the stacks block is handled to make it easier to reason about.

pavitthrap: Update: the issue was that the stacks tip was not updating since the forks were only 1 block long. 
I had to compute the stacks tip another way for the test to work.

src/util_lib/db.rs
```diff
@@ -276,6 +276,18 @@ pub fn u64_to_sql(x: u64) -> Result<i64, Error> {
     Ok(x as i64)
 }
 
+pub fn u64_opt_to_sql(x: Option<u64>) -> Result<Option<i64>, Error> {
+    match x {
+        Some(num) => {
+            if num > (i64::MAX as u64) {
```
pavitthrap, line 282: @jcnelson should I instead be returning an `Option<u128>` from this function to get the full range of u64?

jcnelson:
Will review this on Wednesday.

src/burnchains/db.rs
```diff
@@ -777,4 +777,262 @@ mod tests {
             panic!("EXPECTED to parse a stack stx op");
         }
     }
+
+    #[test]
```
gregorycoppola, line 781: can tests go in files that have the word `test` in them? i think this is new best-practice style, even if it's not in style guide yet.

src/burnchains/db.rs
```diff
@@ -777,4 +777,262 @@ mod tests {
...
+    fn test_classify_delegate_stx() {
```
gregorycoppola, line 782: can this be broken into tests so that each case has its own `fn` and name?

src/chainstate/stacks/db/blocks.rs
```diff
@@ -4753,6 +4754,114 @@ impl StacksChainState {
         all_receipts
     }
 
+    /// Process any Delegate-related bitcoin operations
```
gregorycoppola, line 4757: what does it mean to "process"? briefly describe how the output relates to the input

src/chainstate/stacks/db/blocks.rs
```diff
@@ -11649,20 +11893,35 @@ pub mod test {
...
-    /// Verify that the stacking and transfer operations on the burnchain work as expected in
+    /// Verify that the stacking, transfer, and delegate operations on the burnchain work as expected in
     /// Stacks 2.1.  That is, they're up for consideration in the 6 subsequent sortiitons after
-    /// they are mined (including the one they are in).  This test verifies that TransferSTX
+    /// they are mined (including the one they are in).  This test verifies that TransferSTX & DelegateSTX
     /// operations are only dropped from consideration if there are more than 6 sortitions
     /// between when they are mined and when the next Stacks block is mined.
     #[test]
-    fn test_get_stacking_and_transfer_burn_ops_v210_expiration() {
+    fn test_get_stacking_and_transfer_and_delegate_burn_ops_v210_expiration() {
```
gregorycoppola, line 11918: does the old test not make sense?

are there any new tests to add?

gregorycoppola (commented):
overall, looks pretty good.. just had a few questions..

src/chainstate/stacks/db/blocks.rs
pavitthrap: The name of the function being tested changed, so the name of the test changed too.

src/burnchains/db.rs
pavitthrap: I think we can aim to make these changes after 2.1, but I agree that this is a good practice.

src/chainstate/burn/operations/delegate_stx.rs
```diff
@@ -0,0 +1,574 @@
...
+        );
+        assert_eq!(op.delegated_ustx, u128::from_be_bytes([1; 16]));
+        assert_eq!(op.delegate_to, StacksAddress::new(22, Hash160([2u8; 20])));
+        assert_eq!(op.until_burn_height, None);
+    }
+
+    // Parse a DelegateStx op in which the reward pox address is None.
+    #[test]
+    fn test_parse_delegate_stx_pox_addr_is_none() {
+        // Remove the second OP_RETURN, which is parsed as `pox_addr` if it is set
```
jcnelson, line 299: doc: there's no OP_RETURN here

src/chainstate/burn/operations/delegate_stx.rs
```diff
@@ -0,0 +1,574 @@
...
+            .clone()
+            .try_into_stacks_address()
+            .ok_or_else(|| {
+                warn!("Invalid tx: output 1 must be representable as a StacksAddress");
+                op_error::InvalidInput
+            })?;
+
+        // coerce a hash mode for this address if need be, since we'll need it when we feed this
+        // address into the .pox contract
+        let reward_addr = if outputs.len() >= 2 {
```
jcnelson, line 146: Hmm, I think the structure of the outputs here is problematic.  If you do _not_ want to set a PoX address (which is the usual behavior), then the way this is written, the Bitcoin transaction cannot have a change output.  This means that if you want to create such a transaction, all the BTC you spend from the inputs must be spent by this transaction (which is almost never what you want to do).

I think we'll need to address this in one of two ways:
* Make it so that `outputs[0]` is the `delegate-to` address; `outputs[1]` is the change address, and `outputs[2]`, if present, is the PoX address.  OR:
* Add another field to the `OP_RETURN` which indicates which output is the PoX address, if any.  You'd encode it the same way as the `until-burn-ht` value -- `0x00` for "no PoX address" and "0x01 0xXX` for "use the Bitcoin output at 0xXX as the PoX address".

Up to you how you want to proceed.  I'm more in favor of the second one since conventionally the change address is the last output in a Bitcoin transaction (and lots of wallets are written that way).

src/util_lib/db.rs
jcnelson: No, this is fine.  This and `u64_to_sql` are helper methods to just cast the u64 to an i64 if it's small enough, and store it to sqlite (which only understands signed 64-bit numbers).  It looks like it's the case already, but just be sure that whatever you're using this for will always fit into an i64 for the foreseeable future.  I think you're only using this on `until_burn_ht`, which is a burnchain block height, right?  You're not using this for token balances, which really should be u128?

src/chainstate/coordinator/tests.rs
```diff
@@ -2852,9 +3231,539 @@ fn test_epoch_switch_cost_contract_instantiation() {
...
+// This test verifies that the correct contract is used for PoX for stacking operations.
+// Need to ensure that after v1_unlock_height, stacking operations are executed in the "pox-2" contract.
+// After the transition to Epoch 2.1 but before v1_unlock_height, stacking operations that are
+// sent should occur in the "pox.clar" contract.
+#[test]
+fn test_epoch_verify_active_pox_contract() {
+    let path = "/tmp/stacks-blockchain-epoch-switch-pox-contract-instantiation";
+    let _r = std::fs::remove_dir_all(path);
+
+    let pox_v1_unlock_ht = 12;
```
jcnelson, line 3484: This is the same height as the epoch transition.  Is it the case on mainnet that the v1 unlock height and the epoch transition will be the same?

src/chainstate/burn/operations/delegate_stx.rs
pavitthrap: I'm replacing "OP_RETURN" with "output"

src/chainstate/burn/operations/delegate_stx.rs
pavitthrap: Ahh... I see. I will do it the second way.

src/chainstate/coordinator/tests.rs
pavitthrap: Hmm, it's the same height as the first 2.1 reward cycle I believe, but the actual epoch switch occurs at burn block height 8 in this test.

jcnelson (approved):
LGTM.  Can you add a note to the test plan to spend some cycles verifying that this works?  Thanks!

https://github.com/stacks-network/stacks-blockchain/discussions/3242

pavitthrap:
> LGTM. Can you add a note to the test plan to spend some cycles verifying that this works? Thanks!
> 
> #3242

Yup, added a comment.

obycode (approved):
lgtm

Issue #3188, opened 2022-07-05
Fix `stacks-node` Prometheus recording of "Mempool Size"


gregorycoppola:
**Is your feature request related to a problem? Please describe.**
We do not have an accurate count of the mempool size in the `stacks-blockchain` prometheus metrics.

This comes from the fact that we use a "lazy" way of counting the mempool size. `MEMPOOL_OUTSTANDING_TXS` is incremented in `mempool_accepted` and decremented in `log_transaction_processed`. This can be made incorrect a system re-start.

We see empirically that many followers have negative mempool sizes, indicating a bug.

![image](https://user-images.githubusercontent.com/60008382/177383234-335c07e8-971d-4a06-ba3b-1017d2956744.png)

Thus, we cannot tell how much of the variance is real, versus from the bug. And, we aren't sure the real mean mempool size between all nodes watched.

We should come up with a history-of-restart independent way, so that the counts can be correct.

**Describe the solution you'd like**
We can periodically iterate through the entire mempool, count it, and then update the prometheus statistic.

One thing that kept me from just sending a PR for this was that I wasn't sure where this would go.

**Describe alternatives you've considered**
The current way is lazy, but there is a bug.

**Additional context**
Can be considered part of creating a "mempool analyzer" https://github.com/stacks-network/stacks-blockchain/issues/3002

Related to updates on the API side https://github.com/hirosystems/stacks-blockchain-api/issues/1224

wileyj:
yea, this is something that has been known could be an issue during a restart. 
would it be sufficient to simply query the mempool during startup to "seed" this metric, so the current logic of how the metric is determined wouldn't change? or is that logic part of where you see the bug, requiring a periodic full mempool count?

gregorycoppola:
I think we should have a reference implementation that is straightforward to verify doesn't have a bug. Looping through is the only way to do this, I think.

I can get measurements about how long this loop iteration takes. It's way below 1s AFAIR.

gregorycoppola:
there should be a fast way to do this using a loop:

1) read the entire mempool from SQL, but DO NOT deserialize
2) filter according to current expected nonce
3) report the results (too low, right, too high)

this should be fast, based on results we have and are socializing

gregorycoppola:
Counting of the mempool, along with filtering by nonce appropriateness, can be done in a loop like #3316 at the end of the follower processing a block.

jcnelson:
Closing due to inactivity.  I think this is also a dup of #3342.

PR #3349, opened 2022-10-18
draft: Count and Bucket the Mempool

gregorycoppola:
Draft:  Based on #3337, solves #3188. 

Description:
* add a new `RelayerDirective` for `CountMempool`
* *followers* can set a flag `run_heavy_prometheus_analysis` to issue this directive, and periodically count their mempool and output to prometheus

Working:
* follower can issue the new `CountMempool` directive, and the count gets done
* follower outputs count results to `info!` log

Not yet working:
* use flag guarding
* actually output to prometheus


src/core/mempool.rs
```diff
@@ -1129,6 +1290,79 @@ impl MemPoolDB {
...
+                expected_origin_nonce + 1,
+                self.conn(),
+            );
+            if tx_info.sponsor_address != tx_info.origin_address {
+                nonce_cache.update(
+                    tx_info.sponsor_address,
+                    expected_sponsor_nonce + 1,
+                    self.conn(),
+                );
+            }
```
obycode, line 1360: I don't think you should bump the nonces in here. We should just be looking at transactions that are ready now.

gregorycoppola:
We definitely want to check this in. But, I'll close until progress is made.

PR #3352, opened 2022-10-20
Fix/retry store nonces

jcnelson:
This makes it so that if an attempt to store a cached nonce to the DB fails within `iterate_candidates()`, the nonce will be remembered and later stored in a mempool transaction at the end of the call (so it won't get lost).

### DIFF
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -797,7 +797,21 @@ impl NonceCache {
         }
     }
 
-    fn get<C>(&mut self, address: &StacksAddress, clarity_tx: &mut C, mempool_db: &DBConn) -> u64
+    /// Get a nonce from the cache.
+    /// First, the RAM cache will be checked for this address.
+    /// If absent, then the `nonces` table will be queried for this address.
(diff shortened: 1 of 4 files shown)


src/core/mempool.rs
```diff
@@ -1260,6 +1301,25 @@ impl MemPoolDB {
...
+    /// retry-buffer will be used to later store this data in a single transaction.
+    fn save_nonce_for_retry(
+        retry_store: &mut HashMap<StacksAddress, u64>,
+        max_size: u64,
+        addr: StacksAddress,
+        new_nonce: u64,
+    ) {
+        if (retry_store.len() as u64) < max_size {
+            if let Some(nonce) = retry_store.get_mut(&addr) {
+                *nonce = cmp::max(new_nonce, *nonce);
```
obycode, line 1316: In what scenario would this `max` be necessary? Couldn't we just simplify and overwrite what's there? I don't see any reason that a lower nonce gets passed here.

obycode (approved):
This LGTM. Thanks @jcnelson!

PR #3355, opened 2022-10-22
2.05.0.5.0 RC1

jcnelson:
This is the first release candidate for 2.05.0.5.0


obycode (approved):
Lgtm

gregorycoppola:
Nice!

src/net/p2p.rs
```diff
@@ -6391,4 +6393,191 @@ mod test {
...
+                                &mut result,
+                                sortdb,
+                                chainstate,
+                                mempool,
+                                false,
+                                None,
+                                None,
+                            )
+                        })
+                        .unwrap();
```
eleijonmarck, line 6559: hey hey, curious as to the use of `unwrap` here

since this is a test, it's all good to use `unwrap` , just curious person here  🧐

eleijonmarck: i read the test as in problematic network, was confused
Needed to read the docs string better 
// verify peer 2 stores nothing.

Thx for answering 🙏

PR #3357, opened 2022-10-26
feat/native-segwit-test

donpdonp:
Fix for failing `bitcoind_integration_test` when miner is in segwit mode. Creating early draft PR for @igorsyl to participate.

run command (optionally add BLOCKSTACK_DEBUG=1 for a lot more debug)
```
BITCOIND_TEST=1 RUST_BACKTRACE=1 cargo test -- --nocapture --test-threads 1 --ignored bitcoin_regtest::bitcoind_integration_test_segwit
```

error
```
DEBG [1666809993.325984] [testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs:1257] [main] No UTXOs for 04ee0b1602eb18fef7986887a7e8769a30c9df981d33c8380d255edef003abdcd243a0eb74afdf6740e6c423e62aec631519a24cf5b1d62bf8a3e06ddc695dcb77 (mtFzK54XtpktHj7fKonFExEPEGkUMsiXdy) in epoch 2.0
thread 'main' panicked at 'FATAL: failed to submit leader key register operation', testnet/stacks-node/src/node.rs:611:14
```

### DIFF
--- a/.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
+++ b/.github/actions/bitcoin-int-tests/Dockerfile.bitcoin-tests
@@ -6,6 +6,8 @@ ENV BITCOIND_TEST 1
 RUN cargo build && \
     cargo test -- --test-threads 1 --ignored "$test_name"
 
+RUN cat bitcoind_debug.log
+
 RUN grcov . --binary-path ../../target/debug/ -s ../.. -t lcov --branch --ignore-not-existing --ignore "/*" -o /lcov.info
 
 FROM scratch AS export-stage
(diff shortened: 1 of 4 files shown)


donpdonp:
@igorsyl two notes on BITCOIND_DOCKER
1) the BITCOIND_TEST check was removed so its not possible to have the test immediately pass as before. I'm not sure if that functionality was used or not. 
2) instead of =1 what about BITCOIND_DOCKER='image_name' so the bitcoind version can be specified from the environment.

donpdonp:
!) I see it now, bitcoind_test check is still there. thanks.

jcnelson (approved):

jcnelson:
Let's just make sure CI passes before merging.

PR #3347, opened 2022-10-18
Fix/restore and gate pox sunset

jcnelson:
This PR addresses #3343 by restoring the PoX sunset burn calculation, and gating its applicability based on the epoch.  The gating is confined to methods in `PoxConstants`; code that needs to verify whether or not a sunset is active uses these methods instead of directly reasoning about the current system epoch and block height (this "confines" the gates to one place).

In doing so, it's _possible_ that this PR also fixes #3281 by allowing calls to read-only functions in `.pox` when `.pox-2` is active.  The VM's special case handler had been too strict about this earlier.

PR #3358, opened 2022-10-27
Chore/merge master to next

jcnelson:
This PR brings everything from `master` into `next`, as well as merges #3347 (since `master` expects the PoX sunset still).

Don't merge this until #3347 merges.  I've uploaded this so I can (a) run CI and (b) merge it to my other Stacks 2.1 PRs (notably `fix/1805`) so I can unblock further work on Stacks 2.1.


clarity/src/vm/ast/mod.rs
```diff
@@ -200,6 +227,19 @@ pub fn inner_build_ast<T: CostTracker>(
...
+        // run extra stack-depth pass for tuples
+        match VaryStackDepthChecker::run_pass(&mut contract_ast, clarity_version) {
+            Err(e) if error_early => return Err(e),
+            Err(e) => {
+                diagnostics.push(e.diagnostic);
+                success = false;
+            }
+            _ => (),
+        }
+    }
```
obycode, line 241: I think this should only need to run in `Epoch205`, right? Parser v2 now catches this problem without the extra check.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -7387,3 +7894,2158 @@ fn test_chainwork_partial_interval() {
...
+
+    thread::spawn(move || run_loop.start(None, 0));
+
+    // Give the run loop some time to start up!
+    wait_for_runloop(&blocks_processed);
+
+    (conf, blocks_processed, pox_sync, channel)
+}
+
+// TODO: test in epoch 2.1 with parser_v2
```
obycode, line 8128: Should we do these todo items in this PR?

jcnelson:
All CI tests have passed.  Rejoice :tada:

testnet/stacks-node/src/tests/neon_integrations.rs
jcnelson: I have a separate issue for them here: #3368

clarity/src/vm/ast/mod.rs
jcnelson: Stacks 2.1 will need to behave identically to a Stacks 2.05 node until the activation block height.

obycode (approved):
lgtm

.github/workflows/bitcoin-tests.yml
```diff
@@ -59,7 +60,6 @@ jobs:
           - tests::neon_integrations::antientropy_integration_test
           - tests::neon_integrations::filter_low_fee_tx_integration_test
           - tests::neon_integrations::filter_long_runtime_tx_integration_test
-          - tests::neon_integrations::mining_transactions_is_fair
```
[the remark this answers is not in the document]
jcnelson: Yup; it was removed in `master` because mining isn't "fair" anymore -- if you pay higher fee rates than the next person, your transactions are always prioritized over theirs.

PR #3369, opened 2022-10-31
Replaced set-uint-at with replace-at

a3slade:
### Description
Replaces set-uint-at with replace-at in the pox-2 contract. Removes the now unused set-uint-at function.

### Applicable issues
- fixes https://github.com/stacks-network/stacks-blockchain/issues/3359


src/chainstate/stacks/boot/pox-2.clar
```diff
@@ -311,7 +304,7 @@
                            (moved-cycle-index (- cycle (get first-reward-cycle moved-state)))
                            (moved-reward-list (get reward-set-indexes moved-state))
                            ;; reward-set-indexes[moved-cycle-index] = set-index via slice, append, concat.
```
pavitthrap, line 306: you can remove this comment - or just update the latter half of it

pavitthrap (approved):
lgtm

jcnelson (approved):
LGTM.  Let's just wait for CI to finish running before merging :+1:

Issue #3364, opened 2022-10-28
[Stacks 2.1] Enforce validate `pox-addr` in `delegate-stx`


zone117x:
A two year old TODO comment/question is in both the original pox contract and pox-2 contract source:
> does the delegate _need_ to use a specific pox recipient address?
https://github.com/stacks-network/stacks-blockchain/blob/34654dee5d2f453c8ad21bbcfde94c4ec731568c/src/chainstate/stacks/boot/pox-2.clar#L96-L106

~~I _think_ the answer is: no, this is essentially scratch space that is delegate/client/pool specific which _typically_ corresponds to a valid STX and/or bitcoin address, but does not have to.~~

~~Is this correct? Either way, can we clarify this?~~

EDIT:
@jcnelson clarified that `delegate-stx` should be provided with a valid `pox-addr` (or `none`), and that if an invalid address is given, then the pool operator will not be able to perform the later call to `delegate-stack-stx`.

And also that the code comment isn't a dev question/TODO, but rather a convention of describing the values in the form of a question. That makes a lot more sense now, but the comment pair with the surprising behavior of `delegate-stx` led to confusion.

IMO the `delegate-stx` fn should be updated to require a valid `pox-addr` (it appears to be the only one with this issue?). And/or potentially updating the fn comment/doc to make users aware of this scenario.

jcnelson:
That's not a TODO.  The word "need" is speaking to whether or not the pool operator must use the stacker's given PoX address.  If the PoX address is `some`, then they "need" to use it.  If it's `none`, then they don't -- they can just use whatever PoX address they want.

zone117x:
@jcnelson sounds like 2 different subjects covered in your response:

> That's not a TODO

While "TODO" is not in the comment, the comment certainly looks like a question from the developers along the lines of "what does this do?"

> The word "need" is speaking to whether or not the pool operator must use the stacker's given PoX address. If the PoX address is some, then they "need" to use it. If it's none, then they don't

Is this convention or is this enforced? IIRC last I checked that may be convention, but is not enforced. I just tested a transaction with a definitely-incorrect (not btc/stx encodable) `pox-address`:

```ts
const poxAddr = {
  hashbytes: bufferCV(Buffer.from('send to matt@paypal!')),
  version: bufferCV(Buffer.from([0xff])),
};
const delegateStxTx = await makeContractCall({
  senderKey: delegateeAccount.secretKey,
  contractAddress,
  contractName,
[...]

PR #3371, opened 2022-11-01
Feat/3364: `delegate-stx` validates the given PoX address

jcnelson:
This PR addresses #3364.  Now, a call to `delegate-stx` will fail if the PoX address uses an unrecognized version byte.


gregorycoppola (approved):

zone117x (approved):
I have yet to test these changes in an e2e environment in the API -- but the tests you've added look good enough. 

Thanks for the quick fix for this @jcnelson!

PR #3372, opened 2022-11-02
[Stacks 2.1] fix: regression with pox_2_activation config

zone117x:
This fixes a regression from the recent merge from master into next: https://github.com/stacks-network/stacks-blockchain/pull/3358/files#diff-06a6e88675e542d59fefd0f0e8b3738028510b196074d2a7c77a07f79383a79cL317-R348. The ability to configure pox-2 activation block height in regtest mode no longer worked.

This function
```rs
    fn default_burnchain(config: &Config) -> Burnchain {
        let (network_name, _network_type) = config.burnchain.get_bitcoin_network();
        let working_dir = config.get_burn_db_path();
        let mut burnchain = Burnchain::new(&working_dir, &config.burnchain.chain, &network_name)
            .expect("Failed to instantiate burnchain");
        config.update_pox_constants(&mut burnchain.pox_constants);
        burnchain
    }
```
was refactored and the `update_pox_constants` call was dropped:
```rs
    fn default_burnchain(&self) -> Burnchain {
        let (network_name, _network_type) = self.config.burnchain.get_bitcoin_network();
        match &self.burnchain_config {
            Some(burnchain) => burnchain.clone(),
            None => {
                let working_dir = self.config.get_burn_db_path();
                match Burnchain::new(&working_dir, &self.config.burnchain.chain, &network_name) {
                    Ok(burnchain) => burnchain,
                    Err(e) => {
                        error!("Failed to instantiate burnchain: {}", e);
                        panic!()
                    }
                }
            }
        }
    }
```

Sample config that now works again:
```toml
[burnchain]
chain = "bitcoin"
mode = "krypton"
pox_2_activation = 200

[[burnchain.epochs]]
epoch_name = "1.0"
start_height = 0

[[burnchain.epochs]]
epoch_name = "2.0"
start_height = 0

[[burnchain.epochs]]
epoch_name = "2.05"
start_height = 1

[[burnchain.epochs]]
epoch_name = "2.1"
start_height = 2
```

### DIFF
--- a/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
+++ b/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
@@ -346,7 +346,8 @@ impl BitcoinRegtestController {
                 }
             }
         };
-        self.config.update_pox_constants(&mut burnchain.pox_constants);
+        self.config
+            .update_pox_constants(&mut burnchain.pox_constants);
         burnchain
     }
 
--- a/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
+++ b/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
@@ -333,7 +333,7 @@ impl BitcoinRegtestController {
     /// Get the default Burnchain instance from our config
     fn default_burnchain(&self) -> Burnchain {
         let (network_name, _network_type) = self.config.burnchain.get_bitcoin_network();
-        match &self.burnchain_config {
+        let mut burnchain = match &self.burnchain_config {
             Some(burnchain) => burnchain.clone(),
             None => {
                 let working_dir = self.config.get_burn_db_path();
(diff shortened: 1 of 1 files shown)


gregorycoppola (commented):
LGTM.. thanks!

But, I have two questions:
1) Is it possible to link to the regression so we have a record of why this PR works?
2) Is there any test that could be written that would break before this change but work after?

zone117x:
> 1. Is it possible to link to the regression so we have a record of why this PR works?

Yep, I updated the desc to link to the PR and the segment of code which introduced the regression.

> 2. Is there any test that could be written that would break before this change but work after?

We should definitely add a test that boots a neon/krypton node into epoch2.1 using _only_ a regular config.toml file. The current 2.1 tests directly access and modify variables in a way that is not representative of how a regtest node will work in practice. I have no idea how to add a test like that though :)

jcnelson (approved):

PR #3374, opened 2022-11-02
chore: consistently suffix Clarity functions with `?`

obycode:
This adds `?` to new Clarity2 functions that return a response or optional and also adds an alias for `element-at?` and `index-of?` to standardize the usage of `?`.

See discussion on SIP-015 PR: https://github.com/stacksgov/sips/pull/95#discussion_r1009393537.
cc @LNow @friedger


jcnelson:
Well, if we want to merge this, you'll also have to update `pox-2.clar`.

obycode:
> Well, if we want to merge this, you'll also have to update `pox-2.clar`.

Thanks. I made all of the changes in _clarity/_ but not in _src/_. Fixed in 534cea5f5.

jcnelson:
Wait, this is weird.  Why are we supporting two keywords that do the same thing?  E.g. `element-at` and `element-at?` are both valid names for the same native function?

Can we just pick one for Clarity 2?

obycode:
Sure, I'm fine with that option too. The thought with this was to keep both to avoid confusion that could arise by removing the old spelling, but removing it is also reasonable in my opinion.

jcnelson:
I think removing it is the better option.  It would make both developer and tool author lives easier.

jcnelson:
But that said, I'm not sure I'm 100% onboard with renaming existing functions.  Doing so would invalidate all documentation and working code examples.  If it's a choice between having a consistent naming scheme, and preserving the validity of all Clarity knowledge across the Web and human minds, I'm very biased towards the latter.  It's much harder on people to have to re-learn a language for no benefit (I'm reminded of the Python 2 to Python 3 fiasco).

obycode:
Yes, that's the kind of confusion that I was thinking about too. My preference is to leave the old one as I did in this PR, then add a warning for Clarity2 in tools (esp. Clarinet) recommending to use the `?` version and to use it in any new code snippets and tutorials.

I think that the mild inconvenience for tool developers is worth the benefit of consistent usage of the `?`. The meaning of the `?` and `!` is a common question that I see from people just learning Clarity (and also one that I I had myself), so I think it would be valuable to have a good answer to that, and not "well, sometimes it means that the function returns an optional or response, but some builtin functions don't follow that rule."

LNow:
I would recommend to do it in multiple stages:
1. Add aliases with `?`
2. Do nothing for the next 1-3 releases - during this period tools can mark old versions (without '?' as deprecated)
3. Remove version without `?`

By doing so you keep current DX and improve syntax consistency - both at the same time.
Warning about deprecated function is almost always enough annoying to convince developers to switch to new, preferable and consistent syntax.
Full removal can happen when there are no new contracts deployed in the past 2-3 months which use old syntax.

jcnelson:
I suppose it's early enough in Clarity's lifetime that renaming keywords probably isn't too big of a concern, but I do think it's healthy to get into the habit of _not_ doing this.

The alias method is fine by me, as long as it's version-gated.  Clarity 2 can use either variant, but Clarity 1 should continue to use its old keywords.

gregorycoppola (approved):

friedger:
I haven't reviewed the code, but I am happy with the conclusion in this PR.

obycode:
I just rebased on latest `next` branch. @jcnelson, yes, these changes gate the `?` variants to Clarity2 in the same way that other Clarity2-only functions are.

jcnelson (approved):
LGTM.  Thanks @obycode!

obycode:
The `rustfmt` failure must have snuck in through some other PR... this PR does not touch that file.

`tests::epoch_21::transition_empty_blocks` passes locally.

This should be good to merge.

Issue #3363, opened 2022-10-28
[Stacks-2.1] block-commit miss distance must be strictly less than `MINING_COMMITMENT_WINDOW / 2`.


jcnelson:
MattySTX raised an excellent point that PR #3110 introduces an attack vector:  an opportunistic miner who is also Stacking can simply send `MINING_COMMITMENT_WINDOW` block-commits in the Bitcoin block in which their reward slot(s) are up for PoX payouts, and in doing so, receive a non-zero chance of winning the Stacks block.  This is possible because all of their block-commits exist and are linked by their UTXOs.

To prevent this, a block-commit cannot be late by at least `MINING_COMMITMENT_WINDOW / 2` Bitcoin blocks.  If it is, then it doesn't get counted as part of the UTXO chain.  This forces the miner to have mined for at least `MINING_COMMITMENT_WINDOW / 2` Bitcoin blocks in order to have a chance of winning.

Personally, I'm in favor of setting the maximum `miss_distance` value to 1.  This covers the vast majority of missed block-commits (which are late by one block).  If a block-commit is later than this, then it's invalid and the UTXO chain breaks.  Then, the miner must have mined in at least the past `MINING_COMMITMENT_WINDOW - 1` blocks.

jcnelson:
Merged!

PR #3379, opened 2022-11-04
Feat/3363: Shorten allowed block-commit miss distance and validate PoX outputs for missed commits

jcnelson:
This addresses the Economics CAB's feedback on SIP-015 by making the following two changes (#3363):

* A block-commit cannot be late by more than 1 Bitcoin block.  If it is later than 1 Bitcoin block, it is treated as invalid and the miner's UTXO chain is broken.  This is required in order to ensure that miners _must_ be mining for at least one Bitcoin block before they can win sortition.  Without this change, a miner could just pack four block-commit transactions into the same Bitcoin block (deliberately delaying three of them in order to achieve a target median spend) and mine whenever they want.  They cannot do this today due to a bug in the way we calculate the block-commit miss distance (i.e. how far off the block-commit was from its intended Bitcoin block), but because we're fixing this bug in 2.1, we have to make this change as well.

* A late block-commit must still pay to the PoX recipients of its targeted Bitcoin block.  This is required to ensure that miners are always paying the right Stackers, even if they make a block-commit deliberately late (something they can now do).  Without this change, a miner could deliberately delay a block-commit while also paying themselves the BTC (and thereby do this for only the cost of a Bitcoin transaction).  This is not checked in 2.05, but it's a moot point there because a late block-commit in this version breaks the UTXO chain for a miner altogether.

Sorry for the delay; it was very tedious to test this.  I had to remove the original integration test for this because I couldn't get it to reliably run with the new neon node implementation, so I just ported it over to the chains coordinator test framework.


src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -905,12 +912,86 @@ impl LeaderBlockCommitOp {
...
+            return Err(op_error::BlockCommitBadInput);
+        }
         let epoch = SortitionDB::get_stacks_epoch(tx, self.block_height)?.expect(&format!(
             "FATAL: impossible block height: no epoch defined for {}",
             self.block_height
         ));
 
-        self.check_epoch_commit(epoch.epoch_id)?;
+        let intended_modulus = (self.burn_block_mined_at() + 1) % BURN_BLOCK_MINED_AT_MODULUS;
+        let actual_modulus = self.block_height % BURN_BLOCK_MINED_AT_MODULUS;
```
obycode, line 945: I think this is a dumb question, but I'm having trouble understanding how we can just use the modulus here. Where are we sure that we're in the right window of `BURN_BLOCK_MINED_AT_MODULUS`?

jcnelson: The window is always the last 5 burnchain transactions sent by this miner, plus the current one (each tx must consume a UTXO produced by the previous).  If a block-commit's modulus is valid, then at most one of the following is true:

* The block-commit landed in the intended burnchain block

* The block-commit is so late that it's no longer part of the window, and your sortition weight is now 0.

This all happens in `make_min_median_distribution()` in `src/chainstate/burn/distribution.rs`.  The call site for this function is in `BurnchainStateTransition::from_block_ops()`.

src/chainstate/coordinator/tests.rs
```diff
@@ -1108,6 +1112,681 @@ fn missed_block_commits() {
...
+                ops,
+                vec![].iter_mut(),
+            );
+        }
+        // handle the sortition
+        coord.handle_new_burnchain_block().unwrap();
+
+        let tip = SortitionDB::get_canonical_burn_chain_tip(sort_db.conn()).unwrap();
+        let burn_distribution = get_burn_distribution(sort_db.conn(), &tip.sortition_id);
+        eprintln!("{}", ix);
```
pavitthrap, line 1661: remove print?

src/chainstate/coordinator/tests.rs
```diff
@@ -1108,6 +1112,681 @@ fn missed_block_commits() {
...
+            };
+
+            if b.is_in_prepare_phase(next_mock_header.block_height) {
+                // in prepare phase -- no smoothing takes place
+                assert_eq!(
+                    burn_distribution[0].burns, last_burn,
+                    "Burn distribution should not do windowing at ix = {} block_height = {}",
+                    ix, next_mock_header.block_height
+                );
+                info!(
```
pavitthrap, line 1720: do you want to keep these as info?

obycode (approved):
LGTM

src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -925,12 +932,84 @@ impl LeaderBlockCommitOp {
...
+        if actual_modulus != intended_modulus {
```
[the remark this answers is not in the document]
obycode: This was my concern too, but I think that is handled outside of here as Jude said [here](https://github.com/stacks-network/stacks-blockchain/pull/3379#discussion_r1015965463). Definitely good to take another look at it though.

src/chainstate/coordinator/tests.rs
jcnelson: Yes - makes it easier to `grep`

src/chainstate/burn/operations/leader_block_commit.rs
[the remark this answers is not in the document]
jcnelson: > Is it possible for a tx to come in BURN_BLOCK_MINED_AT_MODULUS blocks later, such that actual_modulus == intended_modulus?

Indeed it is!  But if that happened, it must have been the case that your block-commit was at least `BURN_BLOCK_MINED_AT_MODULUS` blocks late anyway (in which case, your sortition weight is 0).

I suppose I could add this extra check, just for good measure.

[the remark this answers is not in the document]
jcnelson: Actually, we can't do this -- we don't have enough information in the block-commit to know the absolute block height at which it was intended to have landed :shrug:

pavitthrap (approved):
lgtm! could you just add a commit about the modulus thing?

PR #3382, opened 2022-11-05
Feat/3377 bitwise ops

### LINKED ISSUE #3377: [Clarity] Implement remaining common bitwise operators
Not a problem, but it would make it much easier to work with bits if the common bitwise operations were available as built-in functions.

We have `xor` today, would be nice to have bitwise `and`, `or`, `not`, `left shift` and `right shift` functions as well.  I personally don't need all of them, but if implementing one or two it would feel relevant to make sure that at least the most common bitwise functions are implemented.

Currently using good-old-fashioned math combined with list constants, however this is rather cumbersome.

**Additional context**
Having the ability to easily store information in bitwise flags would help to reduce the amount of space needed to store e.g. role-principal mappings (have seen many examples with each role-principal mapping being stored in separate maps, for example).  Could also be useful in certain parsing situations.

cylewitruk:
### Description
Implements additional native Clarity functions for the remaining common bitwise operators, namely:
- Bitwise AND (`&`)
- Bitwise OR (`|`)
- Bitwise NOT (`~`)
- Binary Left Shift (`<<`)
- Binary Right Shift (`>>`) 

In addition, adds an "alias" `~` for the existing Bitwise XOR (`xor`) function, to be consistent in offering "traditional" operator syntax.  I opted to use common symbols instead of words since the words `and`, `or` etc. are already taken.

**Note: ** There were traces of an existing `~` placeholder/function but it wasn't implemented.  I commented out the two tests using this for now, pending feedback.  Rust uses `!` for bitwise NOT, but most other languages use `~`, hence why I chose that symbol.

These additional methods are all native binary arithmetic operations, so I have followed the existing patterns of current arithmetic operators.

### Applicable issues
- fixes #3377 

Bitwise operations in Clarity today are cumbersome to implement and require multiple operations to achieve the desired result, costing more than needed.

### Checklist
- [✅] Test coverage for new or modified code paths
- [ ] Changelog is updated


cylewitruk:
As my first time contributing here, I was a little unsure of which PR criteria needed to be fulfilled for this PR,  Feedback is welcomed :)
- I'm not sure how the changelog should be updated
- I'm not sure which documentation should be updated, from what I could see it looks like Clarity docs are generated from the code?
- Do these functions need a PR in `clarity-benchmarking`?  I used the existing cost values for `xor` and other native arithmetic functions.  All added functions are native, so I assume that they will all be the same..
- Is it relevant with integration tests in bitcoin-tests.yml?

clarity/src/vm/ast/parser/v2/lexer/mod.rs
```diff
@@ -1386,13 +1428,14 @@ mod tests {
             LexerError::InvalidCharPrincipal('O')
         );
 
+        /* cylewitruk: Commenting out this reserved placeholder, assuming it might have been reserved for Bitwise NOT anyway?
```
jcnelson, line 1431: I don't think it was reserved for that.  You'll have to ask @obycode.  It might be for Clarinet or something.

src/chainstate/stacks/boot/costs.clar
```diff
@@ -232,6 +232,21 @@
 (define-read-only (cost_xor (n uint))
     (runtime u1000))
 
+(define-read-only (cost_bitwise_and (n uint))
```
jcnelson, line 235: Please remove all changes to `costs.clar`.  It's consensus-critical data that cannot be changed.

src/chainstate/stacks/boot/costs-2.clar
```diff
@@ -226,6 +226,21 @@
 (define-read-only (cost_xor (n uint))
     (runtime u170))
 
+(define-read-only (cost_bitwise_and (n uint))
```
jcnelson, line 229: Please remove all changes to `costs-2.clar`.  It's consensus-critical data that cannot be changed.

clarity/src/vm/docs/mod.rs
```diff
@@ -568,6 +568,59 @@ const XOR_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+    description: "Returns the result of bitwise not, effectively reversing the bits of `i1`.",
+    example: "(~ 3) ;; Returns -4
+"
+};
+
+const BITWISE_LEFT_SHIFT_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: Some("<< (binary left shift)"),
+    snippet: "<< ${1:expr-1} ${2:expr-2}",
+    signature: "(<< i1 i2)",
+    description: "Moves all the bits in `i1` to the left by the number of places specified in `i2`. New bits are filled with zeros. Shifting a value left by one position is equivalent to multiplying it by 2, shifting two positions is equivalent to multiplying by 4, and so on.",
```
jcnelson, line 612: Style nit: please word-wrap this.

clarity/src/vm/analysis/type_checker/tests/mod.rs
```diff
@@ -769,23 +769,25 @@ fn test_simple_arithmetic_checks() {
...
-    let expected = ["bool", "bool", "bool"];
+    let expected = ["bool", "bool", "bool", "int", "uint"];
     let bad = [
         "(+ 1 2 3 (>= 5 7))",
         "(-)",
         "(xor 1)",
         "(+ x y z)", // unbound variables.
         "(+ 1 2 3 (is-eq 1 2))",
-        "(and (or true false) (+ 1 2 3))",
+        "(and (or true false) (+ 1 2 3))"
```
jcnelson, line 782: Can you also add tests for unsigned ints?

clarity/src/vm/functions/mod.rs
```diff
@@ -517,6 +523,36 @@ pub fn lookup_reserved_functions(name: &str, version: &ClarityVersion) -> Option
...
+                NativeHandle::DoubleArg(&arithmetic::native_bitwise_and),
+                ClarityCostFunction::BitwiseAND,
+            ),
+            BitwiseOR => NativeFunction(
+                "native_bitwise_or",
+                NativeHandle::DoubleArg(&arithmetic::native_bitwise_or),
+                ClarityCostFunction::BitwiseOR,
+            ),
+            BitwiseNOT => NativeFunction(
+                "native_bitwise_or",
```
jcnelson, line 537: Typo (logic error): this should be "native_bitwise_not"

jcnelson (commented):
Hey @cylewitruk, thanks for taking the time to create this PR!  I think if we move fast, there's a chance (but not a guarantee) that this could land in Stacks 2.1 if we can address the following issues.  The Stacks 2.1 code freeze is **November 15**, so you have a little bit of time.

---
First, regarding the design:

* Can you explain the rationale as to why the `&`, `|`, and `^` operators only take two arguments?  Is there a reason they couldn't take more?  Asking because this is supported for the other arithmetic operations, so it's surprising that this PR doesn't support more than two arguments (moreover, bitwise operations are commutative -- there's no "wrong" order for the operands).

* I'm deeply uncomfortable using the C-style names of these operations at this point in the development cycle, because doing so requires changes to `CLARITY_NAME_REGEX_STRING`.  This regex is used in many places, including input validation, so a bug in the regex would be catastrophic.  I'd appreciate it if you used names that were compatible with the existing regex, so as to de-risk introducing bugs into the regex.  The regex is consensus-critical code, so once part of the codebase, it cannot be changed without a hard fork.

* I'm not really a fan of having another name for the bitwise `xor` operator that already exists.  Let's stick with one.

---
Second, regarding the implementation:

* I've done an initial pass on the implementation and left some comments.  Please be sure to address them all.  Others will do likewise.  But in general, at least two approvals will be needed, and both will need to come from separate organizations (I would represent the Stacks Foundation).

* Rust has well-defined rules for how bitwise operations happen on signed versus unsigned integers.  These behaviors need to be _rigorously tested_ in this PR, but I'm not seeing any.

* I'm also not seeing any tests to make sure that these operations can only operate on groups of signed integers at once (in which case, the return type is a signed integer), or only on groups of unsigned integers at once (in which case, the return type is an unsigned integer).  To help build confidence in the soundness of the type system, you'll want to rigorously test this as well.

---
Third, regarding the deployment:

* Adding, changing, or removing Clarity code is a breaking change, and can only be done via a hard fork.  This means that there needs to be a SIP somewhere that describes the new features, why they're necessary, and how they can be added.  See [the SIP process](https://github.com/stacksgov/sips/blob/main/sips/sip-000/sip-000-stacks-improvement-proposal-process.md) for details.  You can see [SIP-015](https://github.com/stacksgov/sips/blob/feat/sip-015/sips/sip-015/sip-015-network-upgrade.md) as a working example of a SIP that implements a breaking change going through the activation process.

* If you want these features to land in Stacks 2.1, you'll need to create a _separate_ SIP from the Stacks 2.1 SIP (SIP-015) and specify in its Activation section that it activates exactly when SIP-015 activates.  This is because SIP-015 is already Recommended, meaning that no more changes to it will be allowed.  But, that shouldn't stop you from creating a "rider SIP" for SIP-015 -- it'll be short and will be easy for the technical and governance consideration advisory boards to review (which will be necessary for getting these changes in).

* Myself and others are happy to help you create the SIP.  If you want to ask on the [Stacks forum](https://forum.stacks.org/), that would be a good place to get started.

clarity/src/vm/functions/mod.rs
cylewitruk: Good catch! Fixed.

clarity/src/vm/analysis/type_checker/tests/mod.rs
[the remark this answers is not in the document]
cylewitruk: Added some cases now.  This test was pretty thin from the beginning, so I wasn't sure of the ambition level here.

clarity/src/vm/docs/mod.rs
cylewitruk: Done :)

src/chainstate/stacks/boot/costs-2.clar
cylewitruk: Done

src/chainstate/stacks/boot/costs.clar
cylewitruk: Done

cylewitruk:
@jcnelson 

> Can you explain the rationale as to why the &, |, and ^ operators only take two arguments? Is there a reason they couldn't take more? Asking because this is supported for the other arithmetic operations, so it's surprising that this PR doesn't support more than two arguments (moreover, bitwise operations are commutative -- there's no "wrong" order for the operands).

I had thought about this as well, but opted to remain consistent with the current `xor` function which only accepts two arguments.  But I could have a look?

> I'm deeply uncomfortable using the C-style names of these operations at this point in the development cycle, because doing so requires changes to CLARITY_NAME_REGEX_STRING. This regex is used in many places, including input validation, so a bug in the regex would be catastrophic. I'd appreciate it if you used names that were compatible with the existing regex, so as to de-risk introducing bugs into the regex. The regex is consensus-critical code, so once part of the codebase, it cannot be changed without a hard fork.

Hm, I wonder if this could be a motivator for having a separate regex for Clarity tokens?  My rationale for using C-style names was to try to be consistent with existing arithmetic operators (+-/* etc.).  I would have opted for normal names if `and` and `not` weren't already taken since `xor` isn't prefixed.  One option would be to use clojure-style names (`bit-and`, `bit-not`, etc.), but then `xor` is left not following the convention 🤷 

> I'm not really a fan of having another name for the bitwise xor operator that already exists. Let's stick with one.

Maybe the above two points actually motivate having a 2nd xor operator? `bit-xor` perhaps, which both would provide naming consistency (without changing the regex), and this method could accept 2..n arguments (without risking changes to the existing function).

> Rust has well-defined rules for how bitwise operations happen on signed versus unsigned integers. These behaviors need to be rigorously tested in this PR, but I'm not seeing any.

I'll have a look at adding some additional tests 👍 I used `xor` as my template and I agree that the tests are a bit thin.

clarity/src/vm/ast/parser/v2/lexer/mod.rs
obycode: This test is just testing that `~` is detected as an unknown symbol. The `Placeholder` is used as a replacement token so that parsing can continue past the problematic symbol. With your changes in place to use `~` as a valid symbol, instead of removing this test, you could instead use a different invalid character here, `@` for example.

clarity/src/vm/ast/parser/v2/lexer/mod.rs
```diff
@@ -2093,6 +2136,7 @@ mod tests {
             }
         );
 
+        /* cylewitruk: Commenting this out as `~` isn't documented anywhere...
```
obycode, line 2139: Similarly here, replace the `~` with `@` in this test.

clarity/src/vm/analysis/type_checker/natives/mod.rs
```diff
@@ -618,8 +618,8 @@ impl TypedNativeFunction {
             CmpGeq | CmpLeq | CmpLess | CmpGreater => {
                 Simple(SimpleNativeFunction(FunctionType::ArithmeticComparison))
             }
-            Sqrti | Log2 => Simple(SimpleNativeFunction(FunctionType::ArithmeticUnary)),
-            Modulo | Power | BitwiseXOR => {
+            Sqrti | Log2 | BitwiseNOT => Simple(SimpleNativeFunction(FunctionType::ArithmeticUnary)),
+            Modulo | Power | BitwiseXOR | BitwiseAND | BitwiseOR | BitwiseLShift | BitwiseRShift | BitwiseXOR2 => {
```
obycode, line 622: Maybe the shift operations should always take a `uint` for the second parameter, regardless of the type of the first?

clarity/src/vm/representations.rs
```diff
@@ -47,7 +47,7 @@ lazy_static! {
         *STANDARD_PRINCIPAL_REGEX_STRING, *CONTRACT_PRINCIPAL_REGEX_STRING
     );
     pub static ref CLARITY_NAME_REGEX_STRING: String =
-        "^[a-zA-Z]([a-zA-Z0-9]|[-_!?+<>=/*])*$|^[-+=/*]$|^[<>]=?$".into();
+        "^[a-zA-Z]([a-zA-Z0-9]|[-_!?+<>=/*])*$|^[-+=/*^~&|]$|^[<][=<]?$|^[>][=>]?$".into();
```
obycode, line 50: `^[<][=<]?$|^[>][=>]?$` This can be simplified to: `^<[=<]?$|^>[=>]?$`.

obycode (changes requested):
Thanks for this PR! I added some minor improvement suggestions.

clarity/src/vm/analysis/type_checker/natives/mod.rs
cylewitruk: Sounds reasonable! Please have a look at my implementation, I didn't see an existing function type that covered this case, so I created a new one (but I could also be blind) - if there was an easier way to achieve this please let me know :)  Updated type checker tests to account for this.

clarity/src/vm/ast/parser/v2/lexer/token.rs
```diff
@@ -34,6 +34,12 @@ pub enum Token {
     GreaterEqual,
     Comment(String),
     Placeholder(String), // used to continue parsing after errors
+    BitwiseXOR,
```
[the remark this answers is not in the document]
cylewitruk: I had looked at the existing BitwiseXOR (which is named so) and followed that - but I have now changed them all, including the existing XOR naming.

clarity/src/vm/docs/mod.rs
```diff
@@ -568,6 +568,61 @@ const XOR_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+New bits are filled with zeros. Shifting a value left by one position is equivalent to multiplying it by 2, 
+shifting two positions is equivalent to multiplying by 4, and so on.",
+    example: "(<< 2 1) ;; Returns 4"
+};
+
+const BITWISE_RIGHT_SHIFT_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: Some(">> (binary right shift)"),
+    snippet: ">> ${1:expr-1} ${2:expr-2}",
+    signature: "(>> i1 i2)",
+    description: "Binary Right Shift Operator. The value of `i1` is moved right by the number of bits specified in `i2`.",
```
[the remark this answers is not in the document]
cylewitruk: Done 👍 Also did a little boy-scouting in that file and wrapped a few other descriptions.

clarity/src/vm/ast/parser/v2/mod.rs
```diff
@@ -3153,6 +3159,122 @@ mod tests {
...
+        let (stmts, diagnostics, success) = parse_collect_diagnostics("(^ 1 2)");
+        warn!("{:?}", success);
+        warn!("{:?}", diagnostics);
+        assert_eq!(success, true);
+        assert_eq!(stmts.len(), 1);
+        assert_eq!(diagnostics.len(), 0);
+        let exprs = stmts[0].match_list().unwrap();
+        match &exprs[0].pre_expr {
+            PreSymbolicExpressionType::Atom(cname) => assert_eq!(cname.as_str(), "^"),
+            _ => panic!("expected atom 'foo'"),
```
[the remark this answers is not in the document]
cylewitruk: Correct!  I have fixed all occurrences that I found in the file.

clarity/src/vm/ast/parser/v2/mod.rs
```diff
@@ -3153,6 +3159,122 @@ mod tests {
                 end_column: 3
             }
         );
+
+        let (stmts, diagnostics, success) = parse_collect_diagnostics("(^ 1 2)");
+        warn!("{:?}", success);
+        warn!("{:?}", diagnostics);
```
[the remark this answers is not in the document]
cylewitruk: Done!

clarity/src/vm/ast/parser/v2/lexer/mod.rs
cylewitruk: Done!

clarity/src/vm/ast/parser/v2/lexer/mod.rs
cylewitruk: Gotcha - this has been fixed now :)

clarity/src/vm/representations.rs
cylewitruk: Good catch, done!  But, please sync with @jcnelson (and any other relevant persons) on the changing of `CLARITY_NAME_REGEX_STRING` based on his feedback above. I don't feel that I have the knowledge to contribute to the discussion of whether or not this change can be considered safe or not.

clarity/src/vm/tests/simple_apply_eval.rs
```diff
@@ -1276,6 +1276,41 @@ fn test_stx_ops_errors() {
     }
 }
 
+#[test]
+fn test_bitwise() {
```
cylewitruk, line 1280: I had tried tests with mismatching types here (e.g. that the 2nd parameter for `>>` and `<<` must be a uint, but the tests passed anyway.  However the type_checker tests succeed...  Do tests in this file skip type checking?

clarity/src/vm/analysis/type_checker/mod.rs
```diff
@@ -192,6 +192,29 @@ impl FunctionType {
                 }
                 Err(CheckErrors::UnionTypeError(arg_types.clone(), found_type.clone()).into())
             }
+            FunctionType::ArithmeticBinaryUIntAsSecondArg => {
+                check_argument_count(2, args)?;
+                let (first, second) = (&args[0], &args[1]);
+                analysis_typecheck_cost(accounting, &TypeSignature::IntType, first)?;
+                analysis_typecheck_cost(accounting, &TypeSignature::IntType, second)?;
```
jcnelson, line 199: This should be `analysis_typecheck_cost(accounting, &TypeSignature::UIntType, second)?;`, right?

clarity/src/vm/analysis/type_checker/tests/mod.rs
```diff
@@ -763,6 +763,66 @@ fn test_unexpected_use_of_field_or_trait_reference() {
     }
 }
 
+#[test]
+fn test_bitwise_good_checks() {
+    let good = [
+        "(>= (+ 1 2 3) (- 1 2))",
```
jcnelson, line 769: Looks like you have a few test vectors in here that are not bitwise checks.

clarity/src/vm/ast/parser/v2/lexer/mod.rs
```diff
@@ -848,7 +860,13 @@ impl<'a> Lexer<'a> {
             | Token::Less
             | Token::LessEqual
             | Token::Greater
-            | Token::GreaterEqual => {
+            | Token::GreaterEqual
+            | Token::BitwiseXor
```
jcnelson, line 864: So, this is one of the reasons why I'm not a fan of having two tokens for xor -- you have to remember to use both in all `match` cases, which is a consensus-critical bug.  Seems you're missing `BitwiseXor2`.

clarity/src/vm/docs/mod.rs
```diff
@@ -568,11 +572,81 @@ const XOR_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+    snippet: "<< ${1:expr-1} ${2:expr-2}",
+    signature: "(<< i1 u2)",
+    description: "Moves all the bits in `i1` to the left by the number of places specified in `u2`. 
+New bits are filled with zeros. Shifting a value left by one position is equivalent to multiplying it by 2, 
+shifting two positions is equivalent to multiplying by 4, and so on. 
+
+Observe that the second parameter (number of positions to shift) must be of type uint.
+",
+    example: "(<< 2 u1) ;; Returns 4
+(<< 16 u2) ;; Returns 64
```
jcnelson, line 623: Can you provide examples of negative integers and unsigned integers?

clarity/src/vm/docs/mod.rs
```diff
@@ -568,11 +572,81 @@ const XOR_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+
+const BITWISE_RIGHT_SHIFT_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: Some(">> (binary right shift)"),
+    snippet: ">> ${1:expr-1} ${2:expr-2}",
+    signature: "(>> i1 u2)",
+    description: "Moves all the bits in `i1` to the right by the number of places specified in `u2`.
+New bits are filled with zeros.  Shifting a value right by one position is equivilent to dividing it by 2,
+shifting two positions is equivilent to dividing by 4, and so on. 
+
+Observe that the second parameter (number of positions to shift) must be of type uint.
```
jcnelson, line 635: Can you add a comment here that indicates that right-shift on a signed integer is an arithmetic right shift (e.g. sign is preserved), whereas a right-shift on an unsigned integer is a logical right shift (e.g. sign bit moves)?

clarity/src/vm/docs/mod.rs
```diff
@@ -568,11 +572,81 @@ const XOR_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+",
+    example: "(>> 2 u1) ;; Returns 1
+(>> 128 u2) ;; Returns 32
```
jcnelson, line 638: Can you add examples with signed integers?

clarity/src/vm/tests/simple_apply_eval.rs
```diff
@@ -1276,6 +1276,41 @@ fn test_stx_ops_errors() {
...
+        "(& u24 u16)",
+        "(^ 24 4)",
+        "(^ u24 u4)",
+        "(| 128 16)",
+        "(| u128 u16)",
+        "(~ 128)",
+        "(~ u128)",
+        "(~ u340282366920938463463374607431768211327)",
+        "(>> u128 u2)",
+        "(<< u4 u2)"
```
jcnelson, line 1292: Can you add test coverage for left- and right-shifting signed integers, including negative ones?

jcnelson: Actually, can you add tests for negative integers for all bitwise operators?

clarity/src/vm/tests/simple_apply_eval.rs
jcnelson: `vm_execute_v2` (really an alias to the test function `execute_v2`) does not run the type-checker.  It just builds the AST and evaluates it.  For that, you'd use `analysis::run_analysis()`.

cylewitruk: Ah okay, thanks for the explanation 👍  I'll leave type checking out of this particular test then since it doesn't seem to be the use-case in this file.

clarity/src/vm/analysis/type_checker/mod.rs
cylewitruk: Most likely :) Updated this now.  Would you be able to describe a little what the intended usage of `analysis_typecheck_cost` is? I tried to infer its usage by looking at the other function types - it looks like it's used to evaluate the type checking cost of each input parameter? Would that be a correct guess?

clarity/src/vm/analysis/type_checker/tests/mod.rs
cylewitruk: copy-pasta while extracting bitwise to its own tests, removed them now 😄

clarity/src/vm/analysis/type_checker/mod.rs
```diff
@@ -192,6 +192,28 @@ impl FunctionType {
                 }
                 Err(CheckErrors::UnionTypeError(arg_types.clone(), found_type.clone()).into())
             }
+            FunctionType::ArithmeticBinaryUIntAsSecondArg => {
```
cylewitruk, line 195: I'm not so pleased with this naming.. suggestions? Ideally I would almost like to have simply a Binary function type where allowed types for arg1 and arg2 can be explicitly specified... maybe for a future iteration.

clarity/src/vm/tests/simple_apply_eval.rs
cylewitruk: 👍

clarity/src/vm/analysis/type_checker/mod.rs
jcnelson: It assesses the runtime cost of performing the type-checking on the second argument.  Analyzing a smart contract isn't free; it consumes a block's compute budget ("gas").

clarity/src/vm/docs/mod.rs
cylewitruk: Added both signed and unsigned examples.

clarity/src/vm/docs/mod.rs
cylewitruk: Done!

clarity/src/vm/docs/mod.rs
cylewitruk: Done!

cylewitruk:
> Can you explain the rationale as to why the &, |, and ^ operators only take two arguments? Is there a reason they couldn't take more? Asking because this is supported for the other arithmetic operations, so it's surprising that this PR doesn't support more than two arguments (moreover, bitwise operations are commutative -- there's no "wrong" order for the operands).
This is implemented now.

clarity/src/vm/analysis/type_checker/natives/mod.rs
cylewitruk: Followup in my comment https://github.com/stacks-network/stacks-blockchain/pull/3382#issuecomment-1308356767 re: no existing function type.

cylewitruk:
Will re-do with a new PR and link to this PR, easiest given the small # of changes...

cylewitruk:
> > What do you guys think of something like this for standardizing native function types (instead of my `ArithmeticBinaryUIntAsSecondArg` bastardization)?
> 
> That looks reasonable to me. My only change would be to remove the `TypeOfFirstArg` and just use `TypeOfArgAtPosition`.

Updated, and managed to get rid of the need to specify the output types explicitly (which was used by docs/`get_output_type_string()`.

clarity/src/vm/docs/mod.rs
```diff
@@ -2897,4 +3056,182 @@ mod test {
...
+        function_type = FunctionType::Binary(
+            FunctionArgSignature::Union(vec![ TypeSignature::UIntType, TypeSignature::PrincipalType, TypeSignature::IntType ]),
+            FunctionArgSignature::Union(vec![ TypeSignature::UIntType, TypeSignature::IntType, TypeSignature::PrincipalType ]),
+            ret.clone());
+        result = get_input_type_string(&function_type);
+        assert_eq!(result, "uint, uint | uint, int | uint, principal | principal, uint | principal, int | principal, principal | int, uint | int, int | int, principal");
+    }
+
+    #[test]
+    fn test_get_input_type_string_unionargs() {
```
cylewitruk, line 3116: Threw this in here while I was at it

clarity/src/vm/docs/mod.rs
```diff
@@ -2897,4 +3056,182 @@ mod test {
...
+        assert_eq!(result, "uint | int");
+
+        function_type = FunctionType::UnionArgs(vec![ TypeSignature::UIntType, TypeSignature::IntType, TypeSignature::PrincipalType ],
+            ret.clone());
+        result = get_input_type_string(&function_type);
+        assert_eq!(result, "uint | int | principal");
+    }
+
+    #[test]
+    fn test_get_input_type_string_variadic() {
```
cylewitruk, line 3137: Threw this in here while I was at it

cylewitruk:
@jcnelson @obycode It would be nice to at least have a little discussion regarding the introduction of `!´, `^`, `&`, `|`, `<<`, `>>` (primarily to the regex) as @jcnelson expressed concern with this.  I haven't removed it yet as @obycode provided a simplification suggestion without expression of concern.

My answer to @jcnelson is found in my comment https://github.com/stacks-network/stacks-blockchain/pull/3382#issuecomment-1304725260 along with my rationale.

If changing the regex is indeed too risky at this stage but the rationale is acceptable, then maybe this should be pushed to the next version of clarity?  Otherwise, good to know now and I'll find a text-based naming (maybe Clojure naming, as suggested in my comment).

obycode:
I'm not so worried about changing the regex. First, I think the change required is pretty clear in what it does. 

```diff
- "^[a-zA-Z]([a-zA-Z0-9]|[-_!?+<>=/*])*$|^[-+=/*]$|^[<>]=?$"
+ "^[a-zA-Z]([a-zA-Z0-9]|[-_!?+<>=/*])*$|^[-+=/*^~&|]$|^<[=<]?$|^>[=>]?$"
```

There are four new symbols added in the pattern looking for single character operators, `^~&|`, and the handling of `<`/`>` adds the option for a second repetition of the symbol, in addition to the option for an `=` that was already there: `^<[=<]?$|^>[=>]?$`. Each of those patterns can only match 3 possible values, e.g. `<`, `<=`, or `<<`.

Second, the parser doesn't even rely on this regex, so the change does not affect the contract parsing. That regex is used to check a `ClarityName` coming in from a transaction only.

jcnelson:
> I'm not so worried about changing the regex. First, I think the change required is pretty clear in what it does.

If you're fine with it, then so am I.

> Second, the parser doesn't even rely on this regex, so the change does not affect the contract parsing. That regex is used to check a ClarityName coming in from a transaction only.

It appears that it's actually only used in `src/net/http.rs` to match paths that query Clarity state.  So that'd derisked.

> What do you guys think of something like this for standardizing native function types (instead of my ArithmeticBinaryUIntAsSecondArg bastardization)?

@cylewitruk Your PR is a consensus-critical change, the code-completion date for Stacks 2.1 is next Tuesday, and there are still `must-ship` PRs in-flight.  I strongly suggest you do _not_ change code that _does not need to be changed_.  "Bastardized" code that the reviewers already understand is better than prettier code that needs to be thoroughly vetted to ensure it doesn't accidentally break something.

obycode (commented):
Some minor issues remaining, but looks great overall. Thanks!

obycode:
@cylewitruk Looks like you've got that CLA problem on this PR too. There are also some merge conflicts that need to be resolved. After those two are resolved, I'm happy with the code changes.

What is the status of the SIP add-on to 015 that was discussed? I can help with that if neither of you have the time for that part.

clarity/src/vm/docs/mod.rs
```diff
@@ -568,11 +575,101 @@ const XOR_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+
+const BITWISE_LEFT_SHIFT_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: Some("<< (binary left shift)"),
+    snippet: "<< ${1:expr-1} ${2:expr-2}",
+    signature: "(<< i1 u2)",
+    description: "Moves all the bits in `i1` to the left by the number of places specified in `u2`. 
+New bits are filled with zeros. Shifting a value left by one position is equivalent to multiplying it by 2, 
+shifting two positions is equivalent to multiplying by 4, and so on. 
+
+Observe that the second parameter (number of positions to shift) must be of type uint.
```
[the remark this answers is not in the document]
cylewitruk: Crap, awesome catch! It causes a panic today, not okay.  Will get back with a fix (I see now that there's `checked_shr` and `checked_shl` available... new to rust 😅).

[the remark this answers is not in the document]
jcnelson: It's not clear that `checked_shr` and `checked_shl` (from `num_traits`) are desirable.  For example, `(<< u340282366920938463463374607431768211455 u1)` throwing a runtime error seems strange when the semantics of logical bit-shifting are well-understood [1] to mean "the left bits are dropped _without_ overflow".  _Arithmetic_ bit-shifting could trigger an overflow error, however, but this PR does not explicitly distinguish between logical and arithmetic shifts.

[1] https://en.wikipedia.org/wiki/Arithmetic_shift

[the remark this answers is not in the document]
cylewitruk: > 

This actually doesn't throw a runtime error with `checked_shl`, it returns `u340282366920938463463374607431768211454`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=59110f744c34be509c635fef81c5393d

clarity/src/vm/analysis/arithmetic_checker/tests.rs
```diff
@@ -324,7 +324,7 @@ fn test_functions_clarity1() {
         ("(to-consensus-buff 3)",
          Ok(())),
         ("(from-consensus-buff true 0x03)",
-         Ok(())),
+         Ok(()))
```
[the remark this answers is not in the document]
cylewitruk: Oops, I was playing around in that test 😅 Should these tests be updated with new operators? I remember I didn't quite figure out how these were intended to work.

clarity/src/vm/docs/mod.rs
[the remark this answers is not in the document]
cylewitruk: I pushed changes using `checked_shr` and `checked_shl` as well as updated tests+documentation a little to give examples of overflows etc.  Let me know what you think.

src/clarity_vm/tests/costs.rs
```diff
@@ -161,6 +161,12 @@ pub fn get_simple_test(function: &NativeFunctions) -> &'static str {
         ToConsensusBuff => "(to-consensus-buff u1)",
         FromConsensusBuff => "(from-consensus-buff bool 0x03)",
         ReplaceAt => "(replace-at list-bar u0 5)",
+        BitwiseAND => "(& 2 3)",
+        BitwiseOR => "(| 2 3)",
+        BitwiseNOT => "(~ 3)",
+        BitwiseLShift => "(<< 2 1)",
+        BitwiseRShift => "(>> 2 1)",
+        BitwiseXOR2 => "(xor 1 2)",
```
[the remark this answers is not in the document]
cylewitruk: Fixed!

cylewitruk:
> @cylewitruk Looks like you've got that CLA problem on this PR too.

Yeah as I mentioned on discord I had the wrong e-mail configured (lost access to that account years ago due to lost 2fa).  I've tried a number of different approaches/suggestions today found on google/stackoverflow on a test clone and all of them result in a mangled history 😬 It was easy with the clarinet PR to just create a new PR since the changes were minuscule, but this one's a bit more complicado.. Maybe @jcnelson has some suggestions?

> There are also some merge conflicts that need to be resolved. After those two are resolved, I'm happy with the code changes.

Resolved now, but please see the recent changes dealing with `<<` and `>>` overflows.

clarity/src/vm/functions/arithmetic.rs
```diff
@@ -84,6 +84,21 @@ macro_rules! type_force_binary_arithmetic {
     }};
 }
 
+macro_rules! type_force_binary_arithmetic_with_uint_as_2nd_arg {
```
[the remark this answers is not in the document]
cylewitruk: Gone now! 😅

clarity/src/vm/analysis/type_checker/mod.rs
```diff
@@ -452,6 +466,39 @@ impl FunctionType {
...
+                ).into());
+            }
+        },
+        FunctionArgSignature::Union(expected_types) => {
+            let mut admitted = false;
+            for expected_type in expected_types.iter() {
+                analysis_typecheck_cost(cost_tracker, expected_type, actual_type)?;
+                if expected_type.admits_type(actual_type) {
+                    admitted = true;
+                    break;
```
jcnelson, line 490: You might as well just error out right here, instead of having an extra `admitted` flag.

clarity/src/vm/docs/mod.rs
```diff
@@ -568,11 +575,105 @@ const XOR_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+(| 64 -32 -16) ;; Returns -16
+(| u2 u4 u32) ;; Returns u38
+"
+};
+
+const BITWISE_NOT_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: Some("~ (bitwise not)"),
+    snippet: "~ ${1:expr-1}",
+    signature: "(~ i1)",
+    description: "Returns the result of bitwise not, effectively reversing the bits of `i1`.",
```
jcnelson, line 620: Should mention that this is the 1's complement of `i1`.

clarity/src/vm/docs/mod.rs
```diff
@@ -568,11 +575,105 @@ const XOR_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+(~ 128) ;; Returns -129
+(~ -128) ;; Returns 127
+"
+};
+
+const BITWISE_LEFT_SHIFT_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: Some("<< (binary left shift)"),
+    snippet: "<< ${1:expr-1} ${2:expr-2}",
+    signature: "(<< i1 u2)",
+    description: "Moves all the bits in `i1` to the left by the number of places specified in `u2`.
```
jcnelson, line 632: Please use `i2` instead of `u2`.  `u2` is a Clarity literal.

"Shifts all the bits in `i1` to the left by the number of places specified in `i2`.  The rightmost `i2` places are filled with zeros."

This is _not_ the equivalent of multiplication in Clarity.  Multiplication can overflow, leading to a runtime error.  This does not.

clarity/src/vm/docs/mod.rs
```diff
@@ -568,11 +575,105 @@ const XOR_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+
+const BITWISE_LEFT_SHIFT_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: Some("<< (binary left shift)"),
+    snippet: "<< ${1:expr-1} ${2:expr-2}",
+    signature: "(<< i1 u2)",
+    description: "Moves all the bits in `i1` to the left by the number of places specified in `u2`. 
+New bits are filled with zeros. Shifting a value left by one position is equivalent to multiplying it by 2, 
+shifting two positions is equivalent to multiplying by 4, and so on. 
+
+Observe that the second parameter (number of positions to shift) must be a 32-bit unsigned integer.
```
jcnelson, line 636: > must be a 32-bit unsigned integer.

This is not reflected in the type checker.  Clarity integers are 128 bits, not 32 bits.

clarity/src/vm/docs/mod.rs
```diff
@@ -568,11 +575,105 @@ const XOR_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+(<< u240282366920938463463374607431768211327 u2402823) ;; Arithmetic overflow
+(<< u123 u24028236699) ;; Arithmetic error: `u2` is larger than a 32-bit unsigned integer
+"
+};
+
+const BITWISE_RIGHT_SHIFT_API: SimpleFunctionAPI = SimpleFunctionAPI {
+    name: Some(">> (binary right shift)"),
+    snippet: ">> ${1:expr-1} ${2:expr-2}",
+    signature: "(>> i1 u2)",
+    description: "Moves all the bits in `i1` to the right by the number of places specified in `u2`.
```
jcnelson, line 651: s/Moves/Shifts

jcnelson (changes requested):
Having slept on it, and having looked at the implementation, I strongly disagree that bit-shift operators should fail with arithmetic overflow.  If the developer wants an arithmetic overflow, then they should use `*`, `/`, and `pow` to calculate bit-shifts.

Logical and arithmetic bit shifts should never overflow or underflow, regardless of how big the operands are.  Also, there's no reason to cap the number of places by a 32-bit number.  First, if overflow/underflow don't happen, this achieves nothing but may surprise the developer.  Second, Clarity integers are 128 bits -- if a cap was even desired, then 32 bits is far too big.  You'd want 7 bits.

clarity/src/vm/analysis/type_checker/mod.rs
cylewitruk: It's not necessarily the first expected_type that matches the actual type.  Hence my earlier comment on "should a caller really have to pay extra using a method that accepts multiple possible types?" (which is a decision made by "us", not the caller).  Erroring out here would return if the first "union type" didn't match.

clarity/src/vm/docs/mod.rs
cylewitruk: No, but it will return an appropriate error:
```
Err(RuntimeErrorType::Arithmetic("The second argument to '>>' must be an unsigned 32-bit integer".to_string()).into()),
Err(RuntimeErrorType::Arithmetic("The second argument to '<<' must be an unsigned 32-bit integer".to_string()).into()),
```

`checked_shl` and `checked_shr` require a u32 as the number-of-bytes-to-shift argument.

cylewitruk:
@jcnelson 
>Having slept on it, and having looked at the implementation, I strongly disagree that bit-shift operators should fail with arithmetic overflow. If the developer wants an arithmetic overflow, then they should use *, /, and pow to calculate bit-shifts.

>Logical and arithmetic bit shifts should never overflow or underflow, regardless of how big the operands are. Also, there's no reason to cap the number of places by a 32-bit number. First, if overflow/underflow don't happen, this achieves nothing but may surprise the developer. Second, Clarity integers are 128 bits -- if a cap was even desired, then 32 bits is far too big. You'd want 7 bits.

This is a result of me moving to `checked_shr` and `checked_shl`, which take a u32 as a second operand and give the descriptions:
```Checked shift left. Computes self << rhs, returning None if rhs is larger than or equal to the number of bits in self.```
```Checked shift right. Computes self >> rhs, returning None if rhs is larger than or equal to the number of bits in self.```

We have several different behaviors available, I think at this stage we just need to decide which behavior is most desirable, document the behavior and go with it.  The native `<<` and `>>` resulted in a panic with too large of numbers, which is why I changed to `checked_shl` and `checked_shr` (similar to the other native arithmetic functions).  We can always introduce additional functions later to cover other shifting cases.

OR, we could remove left/right shift for this version and do a more focused PR for shift operators, possibly with several variations of each (like defined in `num_traits`).

obycode:
How about continuing to use the `checked_*` implementation, but in the case where it returns `None`, the result is set to whatever the expected value would be when shifting all of the bits out. That means for left shift, the result would be `0` and for right shift, the result would be `0` for positive inputs and `-1` for negative inputs.

The alternative that could also be considered "not surprising" to the developer would be the behavior of `wrapping_shl`, where the bits that would exceed the bit width of the type (128) are removed, then the shift is performed. 

My preference is the former.

cylewitruk:
@obycode there is also `overflowing_shr/shl` which would be even more explicit...

jcnelson:
> How about continuing to use the checked_* implementation, but in the case where it returns None, the result is set to whatever the expected value would be when shifting all of the bits out. That means for left shift, the result would be 0 and for right shift, the result would be 0 for positive inputs and -1 for negative inputs.

Yes, this sounds reasonable.

I briefly looked into why this `u32` constraint exists in Rust for `checked_shr` and `checked_shl`.  Turns out, different architectures have different behaviors if you specify a number of bits to shift that exceeds the word size (!!).  So, we'll definitely want to continue using `checked_*` regardless -- the shift operations _must_ evaluate the same way on _all_ architectures.  Making `<<` and `>>` evaluate to `0` or `u0` when the shift amount exceeds the word size (e.g. 128) is at least logically consistent with how bit-shifting is described in most places.

obycode:
I'm going to go ahead and close this one to avoid confusion. See #3389.

PR #3385, opened 2022-11-10
Enables hot reloading of mining configuration file

xoloki:
Enables hot reloading of mining configuration file.

Prior to every mining burnchain transaction, the config file will be reloaded if its modified time is newer and the following settings will be reloaded:

```
burnchain.burn_fee_cap
burnchain.satoshis_per_byte
burnchain.rbf_fee_increment
burnchain.max_rbf
node.microblock_frequency
node.wait_time_for_microblocks
```

The following config file items, if missing, will be read from environment variables:
```
node.seed => STACKS_NODE_SEED
burnchain.username => STACKS_BURNCHAIN_USERNAME
burnchain.password => STACKS_BURNCHAIN_PASSWORD 
```

Addresses the following issues:
- https://github.com/stacks-network/stacks-blockchain/issues/2629
- https://github.com/stacks-network/stacks-blockchain/issues/3181

### DIFF
--- a/testnet/stacks-node/src/config.rs
+++ b/testnet/stacks-node/src/config.rs
@@ -1,11 +1,11 @@
-use std::cell::{Cell, Ref, RefCell};
+use std::cell::RefCell;
 use std::convert::TryInto;
 use std::net::{SocketAddr, ToSocketAddrs};
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::sync::Mutex;
 use std::time::SystemTime;
-use std::{convert, fs, thread};
+use std::{convert, fs};
 
 use rand::RngCore;
 
@@ -1571,7 +1571,7 @@ impl FeeEstimationConfig {
                     .try_into()
                     .expect("Configured fee rate window size out of bounds."),
             )
-                .expect("Error opening fee estimator");
+            .expect("Error opening fee estimator");
             Box::new(FeeRateFuzzer::new(
                 underlying_estimator,
                 self.fee_rate_fuzzer_fraction,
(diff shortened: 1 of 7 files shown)


xoloki:
> Can you run stacks-node with a mock miner configuration to test the config reloading?

Sure, do you want me to write an integration test or just run it by hand and check?

testnet/stacks-node/src/config.rs
```diff
@@ -363,6 +370,163 @@ pub struct Config {
...
+        let field2 = extractor(struct2);
+        if field1 != field2 {
+            info!("{} changed from {} to {}", field_name, field1, field2)
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct ConfigLoaderHandle {
+    handle: std::sync::Arc<std::sync::Mutex<RefCell<ConfigLoader>>>,
```
jcnelson, line 494: You probably don't need `RefCell` here.

jcnelson:
On the whole, it's not clear to me why there needs to be a shared `ConfigHandle` structure.  The miner thread could simply load up the config file itself each time it gets spawned, which would obviate the need for any shared state in RAM.  It's not like it takes an extraordinary amount of time (especially when compared to mining).

testnet/stacks-node/src/config.rs
xoloki: RefCell removed from `ConfigHandle` and `ConfigLoaderHandle`.

pavitthrap:
Closing for now, might resume progress in future.

PR #3387, opened 2022-11-10
Feat/immutable keychain

jcnelson:
This PR reworks the `Keychain` struct in the node so that it's immutable.  It has no internal state to mutate; instead, all key data is generated from arguments.

We believe this fixes a bug we discovered today in which a node might accidentally re-use the same microblock public key hash in a block, causing it to be invalid.  This PR ensures that each microblock keypair is derived from both the miner's current burnchain block height and the parent Stacks block ID, instead of any internal state.

We will want to test this in prod before merging.

### DIFF
--- a/testnet/stacks-node/src/keychain.rs
+++ b/testnet/stacks-node/src/keychain.rs
@@ -1,7 +1,5 @@
-use std::collections::HashMap;
-
 use stacks::address::AddressHashMode;
-use stacks::burnchains::{BurnchainSigner, PrivateKey};
+use stacks::burnchains::BurnchainSigner;
 use stacks::chainstate::stacks::{
     StacksPrivateKey, StacksPublicKey, StacksTransactionSigner, TransactionAuth,
 };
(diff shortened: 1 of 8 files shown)


testnet/stacks-node/src/neon_node.rs
```diff
@@ -1406,27 +1387,34 @@ impl BlockMinerThread {
...
+    /// return None if we were unable to generate the key.
+    #[cfg(not(test))]
+    fn make_microblock_private_key(
+        &mut self,
+        _parent_stacks_hash: &StacksBlockId,
+    ) -> Secp256k1PrivateKey {
+        // Generates a new secret key for signing the trail of microblocks
+        // of the upcoming tenure.
+        self.keychain
+            .make_microblock_secret_key(self.burn_block.block_height, &[])
```
obycode, line 1417: Should the `parent_stacks_hash` be used for the salt here?

obycode (commented):
Good catch, @jcnelson! I have just a few comments.

testnet/stacks-node/src/neon_node.rs
jcnelson: In the test code, we don't do this because it's very hard to get this information from the integration test framework.

obycode: Isn't this the non-test version? `#[cfg(not(test))]`

testnet/stacks-node/src/neon_node.rs
```diff
@@ -1406,27 +1387,34 @@ impl BlockMinerThread {
...
+        self.keychain.make_microblock_secret_key(
+            self.burn_block.block_height,
+            &self.burn_block.block_height.to_be_bytes(),
+        )
+    }
 
-        Some(microblock_secret_key)
+    /// Get the microblock private key we'll be using for this tenure, should we win.
+    /// Return the private key on success
+    /// return None if we were unable to generate the key.
```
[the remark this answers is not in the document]
jcnelson: Fixed

testnet/stacks-node/src/neon_node.rs
```diff
@@ -1406,27 +1387,34 @@ impl BlockMinerThread {
     /// Get the microblock private key we'll be using for this tenure, should we win.
     /// Return the private key on success
     /// return None if we were unable to generate the key.
```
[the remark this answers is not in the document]
jcnelson: Fixed

testnet/stacks-node/src/keychain.rs
```diff
@@ -85,101 +70,78 @@ impl Keychain {
...
-    }
 
-    pub fn sign_as_origin(&self, tx_signer: &mut StacksTransactionSigner) -> () {
-        let num_keys = if self.secret_keys.len() < self.threshold as usize {
-            self.secret_keys.len()
-        } else {
-            self.threshold as usize
-        };
+        let sk_bytes = Keychain::make_secret_key_bytes(&seed.0);
+        let sk = StacksPrivateKey::from_slice(&sk_bytes[..]).expect("FATAL: Keychain::makee_secret_key_bytes() returned bytes that could not be parsed into a secp256k1 secret key!");
```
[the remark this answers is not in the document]
jcnelson: Fixed

testnet/stacks-node/src/keychain.rs
```diff
@@ -11,67 +9,54 @@ use stacks::util::vrf::{VRFPrivateKey, VRFProof, VRFPublicKey, VRF};
...
+        }
+        re_hashed_seed
+    }
 
-        let threshold = 1;
-        let hash_mode = AddressHashMode::SerializeP2PKH;
+    /// Create a secret key from our secret state
+    fn get_secret_key(&self) -> StacksPrivateKey {
+        let sk_bytes = Keychain::make_secret_key_bytes(&self.secret_state);
+        StacksPrivateKey::from_slice(&sk_bytes[..]).expect("FATAL: Keychain::makee_secret_key_bytes() returned bytes that could not be parsed into a secp256k1 secret key!")
```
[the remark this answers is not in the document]
jcnelson: Fixed

gregorycoppola (approved):

testnet/stacks-node/src/neon_node.rs
```diff
@@ -1360,73 +1361,58 @@ impl BlockMinerThread {
...
 
     /// Get the microblock private key we'll be using for this tenure, should we win.
-    /// Return the private key on success
-    /// return None if we were unable to generate the key.
-    fn make_microblock_private_key(&mut self, attempt: u64) -> Option<Secp256k1PrivateKey> {
+    /// Return the private key.
+    ///
+    /// In testing, we ignore the parent stacks block hash because we don't have an easy way to
+    /// reproduce it in integration tests.
+    #[cfg(not(test))]
```
obycode, line 1392: Sorry, I'm still stuck on this. I see you swapped the attributes, but I think the comment should now be moved to the other function. Also, the parameter `_parent_stacks_hash` is still not used in this version either.

jcnelson: Sorry, the rules should be fixed now as follows:

* In test mode, use the block height's big-endian representation.

* In prod, use the parent index block hash.

obycode (approved):
Thanks. LGTM!

PR #3389, opened 2022-11-12
Feat/clarity bitwise operators

### LINKED ISSUE #3377: [Clarity] Implement remaining common bitwise operators

cylewitruk:
**This PR supersedes https://github.com/stacks-network/stacks-blockchain/pull/3382** due to problems with an old github account and the CLA.**

### Description
Introduces additional native Clarity operators for the remaining common bitwise operators, namely:

Bitwise AND (`&`)
Bitwise OR (`|`)
Bitwise NOT (`~`)
Binary Left Shift (`<<`)
Binary Right Shift (`>>`)
In addition, adds a new implementation of Bitwise XOR `^` which accepts multiple operands.

### Applicable issues
- Fixes #3377

Please see PR https://github.com/stacks-network/stacks-blockchain/pull/3382 for more context.

### Checklist
- [ ] Changelog is updated


cylewitruk:
@jcnelson @obycode Squashed commits from previous branch and cherry-picked it here, based off a fresh `next` branch.

obycode:
Thanks for fixing those issues @cylewitruk!

Remaining work is to update the shift behavior as discussed in https://github.com/stacksgov/sips/pull/106.

clarity/src/vm/functions/arithmetic.rs
```diff
@@ -221,9 +221,44 @@ macro_rules! make_comparison_ops {
 macro_rules! make_arithmetic_ops {
     ($struct_name: ident, $type:ty) => {
         impl $struct_name {
+            fn bitwise_left_shift(x: $type, y: u32) -> InterpreterResult<Value> {
```
jcnelson, line 224: Per my comment here [1], this should simply evaluate to 0 if `y > 128`.  So, the code should read:

```rust
fn bitwise_left_shift(x: $type, y: u32) -> InterpreterResult<Value> {
   if y > 128 {
      Self::make_value(0)
    }
    else {
      Self::make_value(x
         .checked_shl(y)
         // NOTE: error path should be unreachable
         .ok_or(RuntimeErrorType::ArithmeticOverflow)?)
    }
 }
 ```
Please continue to use `checked_shl` and `checked_shr`.

Comment (cylewitruk) [reply clarity/src/vm/functions/arithmetic.rs:224]: Oops, missed this comment before pushing changes using `wrapped`-variants based on my discussion with @obycode.  Do you mind having a look at the new implementation, and seeing if it's acceptable before I go back to `checked`-variants?

Comment (jcnelson) [reply clarity/src/vm/functions/arithmetic.rs:224]: @obycode agrees with me: https://github.com/stacks-network/stacks-blockchain/pull/3382#issuecomment-1312499592

Comment (jcnelson) [reply clarity/src/vm/functions/arithmetic.rs:224]: Ah, I see you meant this: https://github.com/stacksgov/sips/pull/106#discussion_r1020840929.

Sure, `wrapping_shl` and `wrapping_shr` are fine in principle.

File: clarity/src/vm/docs/mod.rs [hunk 2]
```diff
@@ -632,39 +632,61 @@ const BITWISE_LEFT_SHIFT_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+
+Observe that the second parameter (number of positions to shift) must be a 32-bit unsigned integer (meaning
+that the maximum allowed value is 4294967295).
 ",
     example: "(<< 2 u1) ;; Returns 4
 (<< 16 u2) ;; Returns 64
 (<< -64 u1) ;; Returns -128
 (<< u4 u2) ;; Returns u16
-(<< u240282366920938463463374607431768211327 u2402823) ;; Arithmetic overflow
 (<< u123 u24028236699) ;; Arithmetic error: `u2` is larger than a 32-bit unsigned integer
```
jcnelson, line 650: This should _never_ error.

clarity/src/vm/docs/mod.rs
```diff
@@ -632,39 +632,61 @@ const BITWISE_LEFT_SHIFT_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
-Observe that the second parameter (number of positions to shift) must be a 32-bit unsigned integer.
+Observe that the second parameter (number of positions to shift) must be a 32-bit unsigned integer (meaning
+that the maximum allowed value is 4294967295).
 ",
     example: "(>> 2 u1) ;; Returns 1
 (>> 128 u2) ;; Returns 32
 (>> -64 u1) ;; Returns -32
 (>> u128 u2) ;; Returns u32
-(>> u240282366920938463463374607431768211327 u2402823) ;; Arithmetic overflow
 (>> u123 u24028236699) ;; Arithmetic error: `u2` is larger than a 32-bit unsigned integer
```
jcnelson, line 685: This should _never_ error.

clarity/src/vm/functions/arithmetic.rs
```diff
@@ -503,14 +503,12 @@ pub fn native_bitwise_left_shift(input: Value, pos: Value) -> InterpreterResult<
         match input {
             Value::Int(input) => {
                 let result = input
-                    .checked_shl(u32_val)
-                    .ok_or(RuntimeErrorType::ArithmeticOverflow)?;
+                    .wrapping_shl(u32_val);
```
jcnelson, line 506: Since you're already committing to masking the `shamt` value by the VM word size, why not just do that here?

```rust
let shamt = u32::try_from(u128_val & 0xffffffff).expect("FATAL: lower 32 bits did not convert to u32);
/* ... */
let result = input.wrapping_shl(shamt);
```

jcnelson: Then, this code path never fails.

jcnelson (changes requested):
I must insist that `<<` and `>>` be infallible.  They should _not_ fail with `ArithmeticError`.

You have until 14 November at 5pm ET to get this done if it's going to ship in 2.1.

clarity/src/vm/functions/arithmetic.rs
cylewitruk: Updated to use your implementation.  Tests and docs updated.

clarity/src/vm/docs/mod.rs
cylewitruk: Updated to use your implementation.  Tests and docs updated.

clarity/src/vm/docs/mod.rs
cylewitruk: Updated to use your implementation.  Tests and docs updated.

cylewitruk:
> I must insist that `<<` and `>>` be infallible. They should _not_ fail with `ArithmeticError`.
> 
> You have until 14 November at 5pm ET to get this done if it's going to ship in 2.1.

I have updated both left- and right-shift to use your masking example.  Please double-check the updated test cases and documentation to ensure that the behavior is as you expect.

clarity/src/vm/functions/arithmetic.rs
obycode: Can these versions be removed?

cylewitruk: @obycode `<<` and `>>` panic.  Could change to checked, but they require that we return an error or pre-determined value.. 🤷 I don't know what's best.

jcnelson:
> This in particular: (<< 123 u9999999999) ;; Returns -170141183460469231731687303715884105728 (i128::MIN, or binary 0b10000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000). The sign ends up flipped in this case. Javascript has similar behavior, just using i32::MIN. This operation using standard << panics in rust. One interesting thing in Javascript is that beyond Number.MAX_SAFE_INTEGER, the operation returns 123.

In rust:

```rust
123i128.wrapping_shl((9999999999u64 & 0xffffffffu64).try_into().unwrap())
1i128.wrapping_shl(127))
```

both evaluate to `-170141183460469231731687303715884105728`, so that looks correct.  Arithmetic left shift is the same as logical left shift in that the MSB can be set to 1 by a shift.  In the signed case, that would be interpreted as a negative number.

clarity/src/vm/docs/mod.rs
```diff
@@ -568,11 +575,133 @@ const XOR_API: SimpleFunctionAPI = SimpleFunctionAPI {
...
+    description: "Shifts all the bits in `i1` to the left by the number of places specified in `i2`. 
+Excess bits shifted off to the left are discarded, and zero-bits are shifted in from the right.
+
+Yields `i1` << mask(`i2`), where `mask` removes any high-order bits of `i2` that would cause the shift
+to exceed the bitwidth of `i1`.
+
+Note: This is not the same as a rotate-left; The `i2` operand is restricted to the bit range of type of `i1`, 
+rather than the bits shifted off to the left being returned to the other end.
+
+Observe that the second operand (number of positions to shift) must be a 32-bit unsigned integer (meaning
```
jcnelson, line 643: This is no longer true.  `i2` can be any `u128`; the implementation simply masks out the upper 96 bits.

obycode:
We will update this implementation once the back and forth on the SIP is finalized. @jcnelson

clarity/src/vm/functions/arithmetic.rs
```diff
@@ -441,6 +482,64 @@ pub fn native_mod(a: Value, b: Value) -> InterpreterResult<Value> {
     type_force_binary_arithmetic!(modulo, a, b)
 }
 
+pub fn native_bitwise_left_shift(input: Value, pos: Value) -> InterpreterResult<Value> {
+    if let Value::UInt(u128_val) = pos {
+        let shamt = u32::try_from(u128_val & 0xffffffff)
```
jcnelson, line 487: I almost wonder if this mask should really be `0x7f`, since there's only 128 bits anyway.

clarity/src/vm/tests/simple_apply_eval.rs
```diff
@@ -1276,6 +1276,96 @@ fn test_stx_ops_errors() {
...
+        "(<< u240282366920938463463374607431768211327 u2402823)", // u130729942995661611608235082407192018816
+        "(<< u340282366920938463463374607431768211455 u1)", // u340282366920938463463374607431768211454
+        "(<< -1 u7)",           // -128
+        "(<< -1 u128)",         // -1
+        "(>> -128 u7)",         // -1
+        "(>> -256 u1)",         // -128
+        "(>> 5 u2)",            // 1
+        "(>> -5 u2)",           // -2
+        "(<< 123 u9999999999)", // -170141183460469231731687303715884105728
+        "(>> 123 u9999999999)", // 0
```
jcnelson, line 1321: Can you also add `(<< -64 u121)` as a test vector?  Surprisingly, this will _panic_ in the Rust playground.  But, it works when I build this PR.

jcnelson (approved):
Looking good @cylewitruk!  Just have a couple small comments before this can be merged.

I'll put together cost functions for this later today.

clarity/src/vm/functions/arithmetic.rs
obycode: That's what's happening under the hood any way in `wrapping_shl`, but maybe it would be more clear to a reader of this code of it was explicit here.

clarity/src/vm/functions/arithmetic.rs
```diff
@@ -441,6 +482,64 @@ pub fn native_mod(a: Value, b: Value) -> InterpreterResult<Value> {
...
+                .into())
+        }
+    } else {
+        Err(CheckErrors::TypeValueError(TypeSignature::UIntType, pos).into())
+    }
+}
+
+pub fn native_bitwise_right_shift(input: Value, pos: Value) -> InterpreterResult<Value> {
+    if let Value::UInt(u128_val) = pos {
+        let shamt = u32::try_from(u128_val & 0xffffffff)
```
obycode, line 516: See comment as above about masking with `0x7f` to be more explicit about the behavior.

src/clarity_vm/tests/costs.rs
```diff
@@ -159,6 +159,12 @@ pub fn get_simple_test(function: &NativeFunctions) -> &'static str {
         StxTransferMemo => r#"(stx-transfer-memo? u1 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR 0x89995432)"#,
         StxBurn => "(stx-burn? u1 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR)",
         StxGetAccount => "(stx-account 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR)",
+        BitwiseAnd => "(bit-and 2 3)",
+        BitwiseOr => "(bit-or 2 3)",
+        BitwiseNot => "(bit-not 3)",
+        BitwiseLShift => "(bit-shift-left 2 1)",
+        BitwiseRShift => "(bit-shift-right 2 1)",
```
obycode, line 166:
```suggestion
        BitwiseLShift => "(bit-shift-left 2 u1)",
        BitwiseRShift => "(bit-shift-right 2 u1)",
```

obycode (changes requested):
Thanks Cyle! Just minor changes requested.

clarity/src/vm/tests/simple_apply_eval.rs
```diff
@@ -1347,18 +1348,19 @@ fn test_bitwise() {
         Ok(Value::Int(24)),     // (bit-and 28 24 -1)
         Ok(Value::Int(-8)),     // (bit-xor 1 2 4 -1)
         Ok(Value::UInt(0)),     // (bit-shift-right u123 u9999999999)
-        Ok(Value::UInt(170141183460469231731687303715884105728)),     // (bit-shift-left u123 u9999999999)
+        Ok(Value::UInt(u128::try_from(i128::MAX).unwrap() + 1)),//170141183460469231731687303715884105728)),     // (bit-shift-left u123 u9999999999)
```
obycode, line 1351:
```suggestion
        Ok(Value::UInt(u128::try_from(i128::MAX).unwrap() + 1)),     // (bit-shift-left u123 u9999999999)
```

obycode (approved):
🎉 Thanks Cyle!

Issue #3384, opened 2022-11-07
[Stacks 2.1] `ast::build_ast_with_rules()` should always require `ASTRules::PrecheckSize` in 2.1


jcnelson:
There's no need to support `ASTRules::Typical` after the 2.1 upgrade.  It's already no longer supported in 2.05 via a soft fork.

obycode:
Would it be correct that we can actually delete all of the code related to handling the different rulesets? After the soft-fork, I think all of this related code can be deleted, simplifying things.

jcnelson:
Yup!

jcnelson:
Ugh fat-fingered

jcnelson:
Merged!

PR #3391, opened 2022-11-14
Feat/simple-astrules

obycode:
Fixes: #3384


obycode:
`test_cost_tracking_deep_contracts_2_1` already attempts to use `ASTRules::Typical` in 2.1 and the error is caught, even before this change, since it is caught in the parser. The pre-check simply serves as a verification that the parser caught the error in this case. I don't think there is any additional test that could be added.

kusai90 (approved):

jcnelson (approved):
LGTM! Just need tests to pass

PR #3392, opened 2022-11-14
fix / pox-2 error codes

janniks:
**Stacks 2.1: `pox-2.clar`** Updates

- fixes incorrect error codes
  - i believe some `ERR_STACK_EXTEND_NOT_LOCKED` should be `ERR_STACK_INCREASE_NOT_LOCKED`
- adds comments to some uncommented error cases (based on other instances of the same error in `pox-2.clar` )
- ~removes trailing whitespace ([hide whitespace in GitHub](https://github.blog/changelog/2021-10-14-hiding-whitespace-is-now-remembered-for-each-pull-request/) to make the diff easier to read)~

⚠️ Also, `ERR_NOT_CURRENT_STACKER` is defined, but NOT used in the contract — i'm not sure if that is intended

---

> Not sure which the best current pox-2 file is, so I based this off of `next`


jcnelson:
Hi @janniks, thanks for this catch!  As a general rule, we don't mix PRs that remove whitespace with those that change code, because it makes it needlessly difficult to catch bugs.  Can you separate out the whitespace PR from the code changes?

janniks:
Definitely! Sorry, my editor config does it automatically

janniks:
Done, ✅

Should I add any changes for the unused `ERR_NOT_CURRENT_STACKER` error code?

jcnelson:
Thanks @janniks! Can you please fix the one failing unit test?  It's fine to change it to the expected error text; this was testing the (incorrect but now fixed) error return value.

zone117x (approved):
LGTM after fixing unit tests 👍

jcnelson:
> Rebased the unit test update ￼ (not sure if this repo prefers nicer commit logs or better GitHub commit reviewability)

It doesn't matter for this PR because it's so small, but in general we _don't_ rebase.  We prefer commit reviewability.

jcnelson (approved):

PR #3394, opened 2022-11-14
style: remove trailing whitespace from pox-2 contract

janniks:
### changes
- removes trailing whitespace from pox-2 contract
- removes double EOF new line

---

**tip**: the diff should be more-or-less empty when [hiding whitespace via GitHub comparison](https://github.blog/changelog/2021-10-14-hiding-whitespace-is-now-remembered-for-each-pull-request/)


jcnelson (approved):
Thanks for doing this as a separate PR!

obycode (approved):
LGTM!

Issue #3261, opened 2022-08-24
[Stacks 2.1] 500 response from `GET /v2/pox`


zone117x:
Running the latest `next` branch in mocknet mode, and the `GET /v2/pox` endpoint often returns an HTTP 500 error with the message: `Failed to query peer info`. On my machine, there's around a 2 second window of time where the 500 errors are returned.

It's easier to reproduce this issue with a short mocknet block time. So perhaps some race condition that happens around block assembly / broadcast time?

Here's a snippet of logs during a 500 request, with `STACKS_LOG_DEBUG=1`:
[stacks-node-500.log.txt](https://github.com/stacks-network/stacks-blockchain/files/9415879/stacks-node-500.log.txt)

It looks like these are the relevant lines:
```
INFO [1661342705.107448] [testnet/stacks-node/src/tenure.rs:70] [main] Node starting new tenure with VRF f1d24b909c98118e60f4396d65ab320dfa7c907c43d5de2031f8b170203441a3
DEBG [1661342705.199959] [src/net/p2p.rs:4918] [ThreadId(2)] local.80000000://(bind=0.0.0.0:20444)(pub=None): load chain view for burn block 30
DEBG [1661342705.204901] [src/net/inv.rs:2530] [ThreadId(2)] Awaken inv sync to re-scan peer block inventories at height 28
DEBG [1661342705.205719] [src/net/inv.rs:2238] [ThreadId(2)] local.80000000://(bind=0.0.0.0:20444)(pub=None): PoX bit vector is PoxId([true]) (reloaded=true)
INFO [1661342705.206280] [src/net/http.rs:1635] [ThreadId(2)] Handle HTTPRequest, verb: GET, peer_addr: 127.0.0.1:55041, path: /v2/pox, query: 
WARN [1661342705.233753] [src/net/rpc.rs:695] [ThreadId(2)] Failed to get PoX info GetPoxInfo(HttpRequestMetadata { version: Http11, peer: PeerHost::DNS(localhost,20443), keep_alive: true, canonical_stacks_tip_height: None }, UseLatestAnchoredTip): DBError(NotFoundError)
DEBG [1661342705.234666] [src/net/rpc.rs:3092] [ThreadId(2)] Processed HTTPRequest, path: /v2/pox, processing_time_ms: 28, conn_id: 8, peer_addr: 127.0.0.1:55041
[...]

PR #3399, opened 2022-11-16
Fix/3261

jcnelson:
This addresses #3261.  It fixes the way we were determining if PoX was active, and the way we were getting the next reward cycle's stacked STX.

@zone117x please give this a try when you have a moment :pray:


src/chainstate/burn/db/sortdb.rs
```diff
@@ -3794,14 +3794,44 @@ impl SortitionDB {
         burnchain: &Burnchain,
         block: &BlockSnapshot,
     ) -> Result<bool, db_error> {
-        let reward_start_height = burnchain.reward_cycle_to_block_height(
+        let mut reward_start_height = burnchain.reward_cycle_to_block_height(
```
gregorycoppola, line 3797: this can be re-written to not use a mutable variable... e.g., call this `prospective_reward_start_height`

src/chainstate/burn/db/sortdb.rs
```diff
@@ -3794,14 +3794,44 @@ impl SortitionDB {
...
+                .ok_or_else(|| {
+                    warn!(
+                        "block height {} does not have a reward cycle",
+                        block.block_height
+                    );
+                    db_error::NotFoundError
+                })?,
         );
+
+        // N.B. the reward cycle start height is 1 + (reward_cycle_number * reward_cycle_length),
```
gregorycoppola, line 3809: can we have a method level comment?

testnet/stacks-node/src/tests/epoch_21.rs
```diff
@@ -210,6 +210,12 @@ fn advance_to_2_1(
     // these should all succeed across the epoch 2.1 boundary
     for _i in 0..5 {
         let tip_info = get_chain_info(&conf);
+        let pox_info = get_pox_info(&http_origin);
+
+        eprintln!(
```
gregorycoppola, line 215: why eprintln instead of info?

gregorycoppola (commented):
looks good.. small comments

testnet/stacks-node/src/tests/epoch_21.rs
jcnelson: easier to grep

src/chainstate/burn/db/sortdb.rs
jcnelson: I don't think that's appropriate.  This is a comment about the particular implementation, not the API contract.

zone117x (approved):

PR #3406, opened 2022-11-17
Fix/3400

### LINKED ISSUE #3400: Miner accidentally stop syncing when winning a block which is not broadcasted then
**Describe the bug**
Miner accidentally stop syncing when winning a block which is not broadcasted then
node stalled
no commits
only restart helps

**Steps To Reproduce**
don't know

 - OS: Amazon Linux
 - Rust version: 1.65.0
 - Stacks node: 2.05.0.5.1

**Additional context**
[debug log](https://watcher-data.s3.amazonaws.com/freezed_miner_debug.log.gz)

p.s. is it possible to broadcast the block to be accepted by the network?

jcnelson:
This fixes #3400.

The root cause was that if the node mined an invalid block for some reason (e.g. due to a duplicate microblock public key hash, which was recently fixed by #3387), the chains coordinator will report that no block has been processed when the node proceeds to process its newly-mined block.  Specifically, the chains coordinator maintains an atomic counter of the number of blocks processed, which today is only incremented when it processes a _valid_ block (but not an _invalid_ block).  Other threads can wake up the coordinator to process a block, and use this counter to wait for the coordinator to finish processing it.

Due to this behavior, the coordinator's processing of an invalid block would _not_ bump this counter.  This, in turn, caused the relayer thread to hang on waiting for the coordinator.

The fix is as follows:

* The coordinator will always increment the blocks-processed counter, even if the block is invalid.
* The relayer will check its miner's tip against the canonical Stacks tip when it finishes processing unhandled blocks, and if they are _different_, it will conclude that there was an invalid block mined and it should not mine microblocks (meaning, it should transition back to attempting to mine Stacks blocks).

### DIFF
--- a/.github/workflows/bitcoin-tests.yml
+++ b/.github/workflows/bitcoin-tests.yml
@@ -78,6 +78,7 @@ jobs:
           - tests::neon_integrations::test_problematic_blocks_are_not_relayed_or_stored
           - tests::neon_integrations::test_problematic_microblocks_are_not_mined
           - tests::neon_integrations::test_problematic_microblocks_are_not_relayed_or_stored
+          - tests::neon_integrations::bad_microblock_pubkey
     steps:
       - uses: actions/checkout@v2
       - name: Download docker image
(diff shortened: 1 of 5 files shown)


wileyj (approved):
shipit

obycode (approved):
lgtm

Issue #3367, opened 2022-10-28
[Stacks 2.1] Epoch 2.0 boot data is not emitted to event-observers


zone117x:
The boot contracts (e.g. bns, pox, seed account balances, etc) are ignored / not emitted through the event-observer interface when bootstrapping a neon/krypton Stacks node to Epoch 2.1.

This _appears_ to be only a bug in the event-emitter code, because while boot contracts & balances are not emitted, they are still visible through RPC / `v2/<endpoint>` calls.

This behavior breaks event-observers like the API which end up with negative STX balances for config.toml seed accounts, missing contract ABIs for pox-1 and bns, etc..

Using something like the following to bootstrap a Stacks node into krypton mode (this is paired with a bitcoind instance booted into regtest mode and an regtest auto-mining script):

```toml
[node]
name = "krypton-node"
rpc_bind = "0.0.0.0:20443"
p2p_bind = "0.0.0.0:20444"
working_dir = "$DATA_DIR"

seed = "$MINER_SEED"
local_peer_seed = "$MINER_SEED"

miner = true
use_test_genesis_chainstate = true
pox_sync_sample_secs = 0
wait_time_for_microblocks = 50
microblock_frequency = 5000
# mine_microblocks = true
# max_microblocks = 10

[miner]
first_attempt_time_ms = 5000
subsequent_attempt_time_ms = 5000

[connection_options]
disable_block_download = true
disable_inbound_handshakes = true
disable_inbound_walks = true

[burnchain]
chain = "bitcoin"
mode = "krypton"
poll_time_secs = 1
pox_2_activation = 104 # note that this is based off starting with a regtest bitcoind node that mines 101 blocks for a BTC balance, where something like 104 is is stacks_block=1, 105 is stacks_block=2, etc.. this can also be set to a value lower than min bitcoin-regtest confirmation blocks, but will result in various burnblock_height values being extra buggy

### bitcoind-regtest connection info
peer_host = "$BITCOIN_PEER_HOST"
peer_port = "$BITCOIN_PEER_PORT"
rpc_port = "$BITCOIN_RPC_PORT"
rpc_ssl = false
username = "$BITCOIN_RPC_USER"
password = "$BITCOIN_RPC_PASS"
timeout = 30

[[burnchain.epochs]]
epoch_name = "1.0"
start_height = 0

[[burnchain.epochs]]
epoch_name = "2.0"
start_height = 1

[[burnchain.epochs]]
epoch_name = "2.05"
[...]

PR #3409, opened 2022-11-18
Feat/check pox 2 params

jcnelson:
This PR adds a config method that will validate the PoX 2 and epoch parameters.  Specifically:

* `v1_unlock_height` must be strictly after the epoch 2.1 start height, and cannot fall on a reward cycle start boundary if it's in the subsequent reward cycle from the epoch 2.1 start.
* If `v1_unlock_height` falls in the same reward cycle as the epoch 2.1 start height, then the epoch 2.1 start height must be in the reward phase -- ideally, the very first block.

The reasons for this are that the `v1_unlock_height` value is used to determine which PoX contract to query to determine a reward cycle's reward set.  So, `pox-2` must be instantiated before the PoX anchor block for its first reward cycle is mined.  This also means that `v1_unlock_height` must not be at the very beginning of a reward cycle -- reward cycles start at the _first_ block, _not_ the _zeroth_ block, of the reward cycle, so the reward cycle of a `v1_unlock_height` that fell on a sortition height such that `sortition_height % reward_cycle_len == 0` would be in the _preceding_ reward cycle.

This PR makes the both the mocknet and Neon nodes panic early if these conditions are not met.

I'm going to see if all integration tests continue to pass with the settings we've given them.  But in the mean time, feel free to try this out.

### DIFF
--- a/testnet/stacks-node/src/config.rs
+++ b/testnet/stacks-node/src/config.rs
@@ -8,7 +8,7 @@ use std::sync::Mutex;
 use rand::RngCore;
 
 use stacks::burnchains::bitcoin::BitcoinNetworkType;
-use stacks::burnchains::PoxConstants;
+use stacks::burnchains::{Burnchain, PoxConstants};
 use stacks::burnchains::{MagicBytes, BLOCKSTACK_MAGIC_MAINNET};
 use stacks::chainstate::stacks::index::marf::MARFOpenOpts;
 use stacks::chainstate::stacks::index::storage::TrieHashCalculationMode;
@@ -17,6 +17,7 @@ use stacks::chainstate::stacks::miner::MinerStatus;
 use stacks::chainstate::stacks::MAX_BLOCK_LEN;
 use stacks::core::mempool::MemPoolWalkSettings;
 use stacks::core::StacksEpoch;
+use stacks::core::StacksEpochExtension;
 use stacks::core::StacksEpochId;
 use stacks::core::{
     CHAIN_ID_MAINNET, CHAIN_ID_TESTNET, PEER_VERSION_MAINNET, PEER_VERSION_TESTNET,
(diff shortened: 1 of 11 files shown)


wileyj (approved):
shipit

deantchi (approved):
LGTM

jcnelson:
Alright, I have a testnet booted off of this PR.  Use this config: https://gist.github.com/jcnelson/8d86edeff62f5ab08c01ff5ca1b0e648

testnet/stacks-node/src/config.rs
```diff
@@ -398,14 +399,105 @@ lazy_static! {
...
+    /// Apply any test settings to this burnchain config struct
+    fn apply_test_settings(&self, burnchain: &mut Burnchain) {
+        if self.burnchain.get_bitcoin_network().1 == BitcoinNetworkType::Mainnet {
             return;
         }
-        if let Some(pox_2_activation_height) = self.burnchain.pox_2_activation {
-            pox_consts.v1_unlock_height = pox_2_activation_height;
+
+        if let Some(v1_unlock_height) = self.burnchain.pox_2_activation {
+            info!(
```
zone117x, line 409: In my testing environment where mining is configured for a 1 second interval, this log line is triggered around 100 times per second. Perhaps it would better as a `debug` level line?

jcnelson: Fixed

jcnelson:
This PR now addresses #3367.  It does so by pushing the boot receipts _immediately after_ instantiating the chainstate, but _before_ any bitcoin block header synchronization happens.  To do this, it pushes the events attached to the _true_ genesis block -- the one with height 0.

zone117x:
@jcnelson can you add a check to one of the TestEventObserver tests to ensure the `pox-2` tx is emitted? I'm now only seeing the `costs-3` tx in the epoch2.1 `/new_block` payload. I've tested with several different block height toml configurations, and with & without performing a stacks-node restart. 

The `pox-2` contract _does_ exist when querying the `/v2/data_var/ST000000000000000000002AMW42H/pox-2/configured` endpoint, and it can be called from contract-call txs. It's just not showing up in the event stream.

jcnelson:
@zone117x Try now.  There was a bug in the epoch initialization code in which it didn't return the tx receipts for both `costs-3` and `pox-2`.

zone117x (approved):
Tested the following changes from this PR and looks good:
* The toml config validation behavior is working as expected, e.g. early exit when invalid heights are configured, solves https://github.com/stacks-network/stacks-blockchain/issues/3407
* The API has a [PR](https://github.com/hirosystems/stacks-blockchain-api/pull/1476) that consumes the new the "block 0" changes, solves https://github.com/stacks-network/stacks-blockchain/issues/3367
* The node no longer panics when stacking to both pox-1 and pox-2 during period 2a, instead the second stacking tx is rejected, solves https://github.com/stacks-network/stacks-blockchain/issues/3423

Issue #3304, opened 2022-09-27
[sortition DB] if there's a Stacks fork tie, pick a winner at random


jcnelson:
Right now, the sortition DB tracks a pointer in each `BlockSnapshot` to the canonical Stacks chain tip at the time this snapshot was the canonical burnchain tip.  This is what makes it possible to query the canonical Stacks chain tip -- it's the highest block on the highest burnchain block.  This pointer is maintained in `SortitionHandleTx::set_stacks_block_accepted_at_tip()`, and is called whenever `StacksChainState::set_block_processed()` is called for a newly-accepted block.

The behavior of the code when a newly-accepted Stacks block is _tied with_ a different Stacks block is problematic -- right now, the code accepts the _first_ Stacks block it sees as a way to break ties.  This makes tie-breaking a function of network arrival times, which is isn't fair because it biases tip selection in favor of who has the best network.

Instead, the code should deterministically choose a winner at random.  Perhaps it could hash the burnchain block hash with the Stacks block hash, and choose the Stacks block whose resulting hash is lexicographically "less than" the other.

yknl:
Possibly related and major issue that we're seeing on the network. Lots of transactions are being rolled back to pending and never being picked up again by miners. I'm assuming these have to do with the re-orgs.

Example is the large number of pending `delegate-stx` transactions for Xverse pool:
https://explorer.stacks.co/txid/SPXVRSEH2BKSXAEJ00F1BY562P45D5ERPSKR4Q33.xverse-pool-v2?chain=mainnet

Also there are multiple reports of user's .btc name registrations being confirmed and then rolled back with the transaction going back to pending or disappearing completely.

PR #3419, opened 2022-11-22
Fix/3418

### LINKED ISSUE #3418: [chainstate] out-of-order block arrival does not update canonical Stacks chain tip
When setting up the initial Stacks 2.1 testnet, I had to stop and restart a public miner and follower while my local (NAT'ed) miner kept running.  In the mean time, it mined four blocks, and ought to have been the canonical chain.

However, this is not reflected in the rebooted public follower and miner.  The miner had mined one Stacks block, and kept believing that this shorter fork was the canonical fork.

From the public miner:
```bash
$ sqlite3 $CHAINSTATE/burnchain/sortition/marf.sqlite 'select block_height,stacks_block_accepted,stacks_block_height,arrival_index,canonical_stacks_tip_height from snapshots order by block_height asc'
2407635|0|0|0|0
2407636|1|1|2|0
2407637|1|2|3|0
2407638|1|3|4|0
2407639|1|1|1|1
2407640|1|4|5|1
2407641|0|0|0|1
```
EDIT: upon the arrival of other burnchain blocks, the canonical tip does get updated:

```
2407641|0|0|0|1
2407642|1|2|6|4
```

The logic in `SortitionDB::set_stacks_block_accepted_at_tip()` is the likely culprit.  My node had mined Stacks blocks 0,1,2,3,4, while the remote node had mined conflicting Stacks block 1.  My node shows this report for the same query:

```bash
$ sqlite3 $CHAINSTATE/burnchain/sortition/marf.sqlite 'select block_height,stacks_block_accepted,stacks_block_height,arrival_index,canonical_stacks_tip_height from snapshots order by block_height asc'
2407635|0|0|0|0
2407636|1|1|1|1
2407637|1|2|2|2
2407638|1|3|3|3
2407639|1|1|4|3
2407640|1|4|5|4
2407641|0|0|0|4
```

I suspect this may also be contributing to the orphan rates we sometimes see in the public network.

jcnelson:
This PR fixes #3418 and #3304 by altering the calculation for the pointer to the canonical Stacks chain tip in the sortition DB.  The previous method did not correctly handle the case where the blocks for the canonical Stacks fork could all arrive out-of-order in a batch.

For example, suppose miner A mines Stacks blocks at heights N and N+1.  Suppose miner B does not see these blocks when it mines, and instead mines a sibling block at height N.  Suppose that miner B then receives the blocks from A.

What should happen is miner B should process miner A's blocks, and then update its view of the canonical fork to match A's.  While miner B does do this today, it only does so once the _burnchain_ advances.  This PR changes this behavior so miner B updates the canonical chain view _immediately_, so it can RBF its in-flight block-commit.

This PR also implements deterministic tie-breaking between canonical Stacks chain tips.  If there are multiple candidates, the canonical one is chosen by hashing its tip's block ID with the current sortition hash and picking the tip with the lowest such hash.

We will want to run this as a real miner on mainnet, as well as testnet.


src/chainstate/burn/db/sortdb.rs
```diff
@@ -4380,17 +4309,79 @@ impl<'a> SortitionHandleTx<'a> {
...
-    /// given parent chain tip to have the correct memoized canonical chain tip present in the fork
-    /// it represents.
-    fn process_new_block_arrivals(
-        &mut self,
-        parent_tip: &mut BlockSnapshot,
-    ) -> Result<(Vec<String>, Vec<String>), db_error> {
-        let mut keys = vec![];
-        let mut values = vec![];
+    /// Resolve ties between blocks at the same height.
+    /// Returns the index into `tied`
```
gregorycoppola, line 4313: how does it break the tie?

CHANGELOG.md
```diff
@@ -5,6 +5,30 @@ All notable changes to this project will be documented in this file.
...
+  tip, the node deterministically chooses one _independent_ of the arrival order
+(#3419). 
+- If Stacks blocks for a different fork arrive out-of-order and, in doing so,
+  constitute a better fork than the fork the node considers canonical, the node
+will update the canonical Stacks tip pointer in the sortition DB before
+processing the next sortition (#3419).
+
+### Fixed
+
+- The node keychain no longer maintains any internal state, but instead derives
```
gregorycoppola, line 24: is this really part of this PR?

obycode (approved):
LGTM!

src/chainstate/burn/db/sortdb.rs
```diff
@@ -6909,20 +6998,14 @@ pub mod tests {
         // accept blocks 5 and 7 in one fork, and 6, 8, 9 in another.
         // Stacks fork 1,2,3,4,5,7 will be the longest fork.
         // Stacks fork 1,2,3,4 will overtake it when blocks 6,8,9 are processed.
-        let mut parent_stacks_block_hash = BlockHeaderHash([0x04; 32]);
-        for (i, height) in [5, 7].iter().zip([5, 6].iter()) {
+        for (i, height) in [7].iter().zip([6].iter()) {
```
gregorycoppola, line 7001: does this loop still match the comment just above?

gregorycoppola (approved):
please document the method contract for tie-breaking but LGTM

src/chainstate/burn/db/sortdb.rs
jcnelson: yes; it was a bug to use i=5 and height=5

CHANGELOG.md
jcnelson: no; this was supposed to be included in an earlier PR but I don't want to wait for CI to run on a PR that just adds a changelog

src/chainstate/burn/db/sortdb.rs
jcnelson: it's 20 lines down:

```rust
// break ties by hashing the index block hash with the snapshot's sortition hash, and
// picking the lexicographically smallest one
```

PR #3421, opened 2022-11-28
feat: set an appropriate span for desugared tuples

obycode:
Fixes the spans (line/column numbers) in tuple expressions.


lgalabru (approved):
Thanks @obycode!

Issue #3426, opened 2022-12-03
[Stacks 2.1] Regressions in Clarity 1 type checking


lgalabru:
In the process of releasing v1.2.0 of Clarinet - pulling the very latest next vm of this repo - a few regressions in the way Clarity v1 is supposed to handle type checking were revealed.

- Case 1

```clarity
(define-public (caller) 
    (begin 
        (asserts! (is-eq  .callee contract-caller) (err u1))
        (ok u1)))
```
fails with 
```
error: expecting expression of type '(principal ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.callee)', found 'principal'
```

- Case 2

```clarity
(define-public (caller) 
    (begin 
        (asserts! (is-eq  contract-caller .callee) (err u1))
        (ok u1)))
```
fails with 
```
error: expecting expression of type 'principal', found '(principal ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.callee)'
```

- Case 3

```clarity
(begin 
    (contract-call? .base-dao set-extensions
      (list
        {extension: .ccd001-direct-execute, enabled: true}
        {extension: .ccd002-treasury-mia, enabled: true}
        {extension: .ccd002-treasury-nyc, enabled: true}
    )
))

(define-public (set-extensions (extensionList (list 200 {extension: principal, enabled: bool})))
...
)
```

fails with
```
expecting expression of type '(list 200 (tuple (enabled bool) (extension principal)))', found '(list 3 (tuple (enabled bool) (extension principal)))'
```

This 3 issues were patch directly in https://github.com/stacks-network/stacks-blockchain/pull/3293 to unblock the Clarinet release
- Case 1 and 2 patched with https://github.com/stacks-network/stacks-blockchain/pull/3293/commits/e8ad18055de64db6fbe02914b95367c20bcb9212
- Case 3 patched with https://github.com/stacks-network/stacks-blockchain/pull/3293/commits/601ac7a0b4ed3de027382a33560d079aaa337b4d

These patches need to have their dedicated PR with unit testing, then 2 aforementioned commits will be reverted from that PR.

I think we want to spin up a Stacks 2.1 mainnet node, booting from network, that would help convince us that all the changes that were introduced in `next` are backward compatible - a stacks 2.1 node with a very distant epoch 2.1 transition should be able to produce the same chainstate than a stacks-node coming from mainline.

[...]

PR #3432, opened 2022-12-06
fix: resolve regressions in Clarity1 type-checking

obycode:
Fixes: #3426

### DIFF
--- a/clarity/src/vm/types/signatures.rs
+++ b/clarity/src/vm/types/signatures.rs
@@ -907,15 +907,31 @@ impl TypeSignature {
                     Ok(ListUnionType(HashSet::from([x.clone(), y.clone()])))
                 }
             }
-            (ListUnionType(l), CallableType(y)) => {
+            (ListUnionType(l), CallableType(c)) | (CallableType(c), ListUnionType(l)) => {
                 let mut l1 = l.clone();
-                l1.insert(y.clone());
+                l1.insert(c.clone());
                 Ok(ListUnionType(l1))
             }
-            (CallableType(x), ListUnionType(l)) => {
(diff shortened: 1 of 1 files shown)


clarity/src/vm/types/signatures.rs
```diff
@@ -907,15 +907,31 @@ impl TypeSignature {
...
-                let mut l1 = l.clone();
-                l1.insert(x.clone());
-                Ok(ListUnionType(l1))
+            (PrincipalType, CallableType(CallableSubtype::Principal(_)))
+            | (CallableType(CallableSubtype::Principal(_)), PrincipalType) => Ok(PrincipalType),
+            (PrincipalType, ListUnionType(l)) | (ListUnionType(l), PrincipalType) => {
+                let mut all_principals = true;
+                for ty in l {
+                    match ty {
+                        CallableSubtype::Trait(_) => {
```
jcnelson, line 921: While there are only two variants for `CallableSubtype` at this time, could you change this `match` so that we explicitly require that `ty` is a `CallableSubtype::Principal(..)` variant?  This way, if we add other `CallableSubtypes`, this code will continue to work the way we want -- i.e. `ListUnionType(l)` must contain only `CallableSubtype::Principal(..)` instances.

jcnelson (approved):
LGTM; just a minor comment on defensive programming.

clarity/src/vm/types/signatures.rs
obycode: Good point. Fixed in ce6151364.

lgalabru (approved):
LGTM, thanks @obycode!

PR #3435, opened 2022-12-08
Fix/ci/upgrade and lock clarinet core contract tests

netrome:


jcnelson:
Thanks @netrome! I'll approve this for now, but what was the underlying problem?

jcnelson (approved):

.github/workflows/ci.yml
```diff
@@ -86,7 +86,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: "Execute core contract unit tests in Clarinet"
-        uses: docker://hirosystems/clarinet:latest
+        uses: docker://hirosystems/clarinet:1.1.0
```
lgalabru, line 89: I think we need to use latest here.

lgalabru: in the context of this repo, we could actually pin, the contract that clarinet is testing is a boot contract based on clarity 1, and we're not actively working on it. by pinning 1.1.0, we re pulling an unstable clarity 2 vm, but that's fine the only thing we need in this context is a stable clarity 1 vm.

netrome: Great, thank you for the input! I prefer maintaining pinned versions to keep the CI more predictable.

PR #3437, opened 2022-12-09
Fuzz testing

obycode:
Note, this branch includes fix/3426-type-checker, which is not yet merged to `next` - #3432

To try it out:
```
cd clarity
cargo fuzz run fuzz_admits_type
cargo fuzz run fuzz_least_supertype
```


clarity/src/vm/types/signatures.rs
```diff
@@ -74,12 +74,36 @@ pub struct TupleTypeSignature {
...
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct BufferLength(u32);
 
+#[cfg(feature = "arbitrary")]
+impl arbitrary::Arbitrary<'_> for BufferLength {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
+        let len = u32::arbitrary(u)?;
```
jcnelson, line 91: In order to decrease the number of obviously invalid tests, would it make sense to bound the `len` here to be something smaller than 2**32 - 1?

obycode: The `try_from` on the next line will report an error if it is out of range, causing the fuzzer to pick a new number.

rust-toolchain
```diff
@@ -1 +1 @@
-stable
+nightly
```
jcnelson, line 1: What do we need from `nightly`?

obycode:
I'm still working on the part that tests the whole set of "checkers", old vs. new, which I think is the key test. The two type-checker method-specific fuzzers can be used now, but there are still some false positives with bad types getting generated that I need to work out.

rust-toolchain
obycode: That's for cargo-fuzz:

> Note: libFuzzer needs LLVM sanitizer support, so this only works on x86-64 Linux, x86-64 macOS and Apple-Silicon (aarch64) macOS for now. This also needs a nightly compiler since it uses some unstable command-line flags. You'll also need a C++ compiler with C++11 support.

pavitthrap:
out of date, closing in favor of a newer branch.

PR #3438, opened 2022-12-09
Reduce false positives and panics in fuzz-testing

fpbgg:
```
cd clarity/fuzz
./keep-fuzzing.sh
```


clarity/src/vm/types/signatures.rs
```diff
@@ -528,28 +600,124 @@ impl TypeSignature {
...
+            #[cfg(feature = "fuzzing")]
+            TypeSignature::TraitReferenceType(_) => {
+                panic!("TraitReferenceType should only be used for testing 2.05.")
+            }
+        }
+    }
+
+    #[cfg(feature = "fuzzing")]
+    pub fn old_admits_type(&self, other: &TypeSignature) -> Result<bool> {
+        let other = match other.concretize() {
```
obycode, line 627: We shouldn't need to call `concretize` here because the types that require being concretized should not show up in the "old" types (`ListUnionType` and `CallableType`).

obycode (commented):
Sorry to cause this, but can you rebase this on the latest test/fuzzing branch? I just rebased that branch on next.

PR #3443, opened 2022-12-15
Chore/merge master to next

jcnelson:
This merges `master` to `next` to prepare for 2.1.0.0.0-rc1.

@obycode Please make sure I got the Clarity changes merged in correctly!  :pray:

PR #3444, opened 2022-12-15
Chore/merge develop to next

jcnelson:
This merges `develop` to `next` by way of #3443.  Please merge this before #3443.


jcnelson:
@netrome Your patch set with `stdext::function_name!()` has broken a bunch of tests because it resulted in some smart contract names being too long.  The limit is 40 characters.  Please fix this.

netrome:
> @netrome Your patch set with `stdext::function_name!()` has broken a bunch of tests because it resulted in some smart contract names being too long. The limit is 40 characters. Please fix this.

Sure thing. I'll prepare a PR which removes the namespace information in the `stdext::function_name!()` call.

netrome:
@jcnelson here's the PR removing the namespace names. https://github.com/stacks-network/stacks-blockchain/pull/3445

PR #3446, opened 2022-12-16
Complete testing and bugfixes for PoX reorg handling

jcnelson:
This PR finishes all of the test items for PoX reorg handling.  In doing so, it uncovered a few design and implementation bugs in the current system, and adds several torture tests to ensure that we correctly handle all manor of cases where PoX anchor blocks can go missing.  In all cases, the goal is to ensure that the network converges to Stacks forks that lie on the heaviest affirmation map as determined by block-commits.

The fixes include making it so that we track both the sortition history's affirmation map, as well as the Stacks fork's affirmation map, and process a PoX reorg if _either one_ conflicts with the heaviest affirmation map.  Doing this requires a lot of bookkeeping, so that we don't reprocess an already-processed sortition and so we maintain the pointer to the canonical Stacks tip on the newly-reprocessed/revalidated sortitions after a PoX reorg.


src/chainstate/coordinator/mod.rs
```diff
@@ -797,292 +1252,304 @@ impl<
...
-                                        }
-                                    }
-                                    if !found && num_sns > 0 {
-                                        // there are snapshots, and they're all diverged
-                                        debug!("No snapshot at height {} has a PoX ID that is a prefix of {} (affirmation map {})", height, &affirmation_pox_id, &heaviest_am);
-                                        break;
-                                    }
-                                }
-                                break;
+            // let invalidation_height = revalidate_sn.block_height;
```
obycode, line 1360: There are several commented uses of `revalidate_sn` that can be deleted.

src/net/inv.rs
```diff
@@ -1775,6 +1775,26 @@ impl PeerNetwork {
 
     /// Determine at which reward cycle to begin scanning inventories
     fn get_block_scan_start(&self, sortdb: &SortitionDB, highest_remote_reward_cycle: u64) -> u64 {
+        // see if the stacks tip affirmation map and heaviest affirmation map diverge.  If so, then
+        // start scaning at the reward cycle just before that.
```
obycode, line 1779:
```suggestion
        // start scanning at the reward cycle just before that.
```

obycode (approved):
This all makes sense to me. Thanks!

PR #3449, opened 2022-12-16
Use old type-checking while in epoch 2.05

obycode:
This changeset re-adds the old type-checking behavior and uses it while epoch 2.05 is still active.


obycode:
Note to reviewers: the _type_checker/v2_05/_ directory is basically the _type_checker/_ from master, with minimal changes to pass the epoch around as needed.

obycode:
Note also that in addition to the type-checker code, the other main entry points to differing behavior are in `admits_type` and `least_supertype` from signatures.rs.

obycode:
The failing unit test is caused by a bind error -- seems to be caused by tests running in parallel. It passes locally. I'm not sure what caused this to begin failing here.

gregorycoppola (approved):
LGTM

clarity/src/vm/types/signatures.rs
```diff
@@ -1138,19 +1366,35 @@ impl TypeSignature {
...
+                if epoch < StacksEpochId::Epoch21 =>
+            {
+                match trait_definition {
+                    TraitDefinition::Defined(trait_id) => {
+                        Ok(TypeSignature::TraitReferenceType(trait_id.clone()))
+                    }
+                    TraitDefinition::Imported(trait_id) => {
+                        Ok(TypeSignature::TraitReferenceType(trait_id.clone()))
+                    }
+                }
```
jcnelson, line 1396: What happens if `epoch == StacksEpochId::Epoch21`?  Is that reachable?

src/clarity_vm/tests/forking.rs
```diff
@@ -284,8 +285,10 @@ where
 
     {
         let mut store = marf_kv.begin(&StacksBlockId([0 as u8; 32]), &StacksBlockId([1 as u8; 32]));
-        let mut owned_env =
-            OwnedEnvironment::new(store.as_clarity_db(&TEST_HEADER_DB, &TEST_BURN_STATE_DB));
+        let mut owned_env = OwnedEnvironment::new(
```
jcnelson, line 288: Can you use `rstest` or some such to make sure we test the forking tests on each epoch, and with all applicable clarity versions?  Thanks!

clarity/src/vm/callables.rs
```diff
@@ -175,7 +179,13 @@ impl DefinedFunction {
             // Clarity 1 behavior
             if *env.contract_context.get_clarity_version() < ClarityVersion::Clarity2 {
                 match (type_sig, value) {
+                    // Epoch < 2.1 uses TraitReferenceType
```
jcnelson, line 182: Should this `if`-branch also check the epoch?  There are three combinations allowed -- stacks 2.1 + clarity 2, stacks 2.1 + clarity 1, and stacks 2.05 (or lower) and clarity 1.

I think one of the invariants you're trying to preserve here is that only the 2.05-series type checker can ever instantiate a `TypeSignature::TraitReferenceType`?  Should callers simply expect this?

clarity/src/vm/tests/simple_apply_eval.rs
```diff
@@ -102,7 +102,7 @@ fn test_simple_let(#[case] version: ClarityVersion, #[case] epoch: StacksEpochId
     if let Ok(parsed_program) = parse(&contract_id, &program, version, epoch) {
         let context = LocalContext::new();
         let mut marf = MemoryBackingStore::new();
-        let mut env = OwnedEnvironment::new(marf.as_clarity_db());
+        let mut env = OwnedEnvironment::new(marf.as_clarity_db(), StacksEpochId::latest());
```
jcnelson, line 105: Should this be tested in all epochs?

clarity/src/vm/tests/traits.rs
```diff
@@ -66,7 +68,7 @@ fn test_all() {
         test_let3_trait,
     ];
     for test in to_test.iter() {
-        with_memory_environment(test, false);
+        with_memory_environment(test, StacksEpochId::latest(), false);
```
jcnelson, line 71: Should this be tested in all epochs?

clarity/src/vm/types/serialization.rs
```diff
@@ -612,7 +614,7 @@ impl Value {
                 }
 
                 if let Some(list_type) = list_type {
-                    Value::list_with_type(items, list_type.clone())
+                    Value::list_with_type(&StacksEpochId::latest(), items, list_type.clone())
```
jcnelson, line 617: I think we'll want an explicit epoch ID here.  When we're working on Stacks 3.0, we'll want the serialization code to behave the same until it's strictly necessary to change it (especially if 3.0 introduces different behavior).

clarity/src/vm/types/signatures.rs
```diff
@@ -523,20 +622,20 @@ impl TypeSignature {
...
+                            && my_inner_type.0.admits_type_v2_1(&other_inner_type.0)?)
                     }
                 } else {
                     Ok(false)
                 }
             }
             TupleType(ref tuple_sig) => {
                 if let TupleType(ref other_tuple_sig) = &other {
-                    tuple_sig.admits(other_tuple_sig)
+                    tuple_sig.admits(&StacksEpochId::Epoch21, other_tuple_sig)
```
jcnelson, line 638: Below, do we want a request to type-check an instance of `TraitReference` to be `unreachable!()`, like `CallableType` type-checking is unreachable in epoch 2.0?

clarity/src/vm/types/signatures.rs
```diff
@@ -1578,7 +1827,7 @@ mod test {
             epoch,
         )
         .unwrap()[0];
-        TypeSignature::parse_type_repr(expr, &mut ()).unwrap_err()
+        TypeSignature::parse_type_repr(StacksEpochId::latest(), expr, &mut ()).unwrap_err()
```
jcnelson, line 1830: Also flagging this to use each epoch version for testing.

jcnelson (approved):
LGTM! Nothing major; just some minor comments and questions about the gating.

Thank you for hustling on this!  After 2.1 activates, we can look at removing the 2.05 type-checker altogether.

clarity/src/vm/types/signatures.rs
obycode: Yes, it will fall through to the next case of the match, where a `CallableType` is used instead of a `TraitReferenceType`.

src/clarity_vm/tests/forking.rs
obycode: Added in 3a5d8ef83

clarity/src/vm/callables.rs
obycode: Right, when in epoch < 2.1, we should not see `CallableType` and in epoch >= 2.1, we should not see `TraitReferenceType`. Yes, I could add a check to make sure the epoch matches the correct type here.

clarity/src/vm/tests/simple_apply_eval.rs
obycode: Yup, will fix.

clarity/src/vm/tests/traits.rs
obycode: This one will be a bit trickier, because some of these tests only work in 2.1. I'll have to go through and split them up into those that can run on both 2.05 and 2.1, and those that can only run on 2.1 (because they rely on Clarity2).

clarity/src/vm/types/signatures.rs
obycode: Yes, I added this in 2c2635f80da637a2a9f87db13a7e36042ea1dc95.

clarity/src/vm/types/signatures.rs
obycode: fixed.

obycode:
Thanks for the reviews @jcnelson and @gregorycoppola. All points have been addressed.

PR #3451, opened 2022-12-20
pox-2 is default contract at v1_unlock_height+1

### LINKED ISSUE #3422: [Stacks 2.1] /v2/pox reporting incorrect .contract_id
**UPDATE:**

https://github.com/stacks-network/stacks-blockchain/pull/3451 seems to fix the issue, but due to how the pox-rpc response is set up, the resulting `.contract_id` switch to `pox-2` is delayed by a cycle (rather than just by one block). This happens because the `active_pox_contract()` method is called using the reward cycle start heights and not the exact current burn height.

https://github.com/stacks-network/stacks-blockchain/blob/e21347232f9ab2b4417a33476c8eb0131756f1a0/src/net/rpc.rs#L426-L427

---

**PREVIOUS BEHAVIOR:**

> This bug was encountered during the development of https://github.com/hirosystems/stacks-blockchain-api/pull/1454

**Describe the bug**
It would seem that the reported /v2/pox contract_id is wrong in at least one case.

The switch to .pox-2 as "active" _(being the .contract_id in /v2/pox)_ happens 1 block early (on block `v1_unlock_height`).

**Expected behavior**
According to SIP-015 and related discussions, it is my understanding that it should switch on `v1_unlock_height + 1` and NOT `v1_unlock_height` itself.

> Since the switch happens "early", it doesn't feel like a race condition — so, I don't know what else could cause this...

---

**Steps To Reproduce / Additional Context**
[Here is an API-test](https://github.com/hirosystems/stacks-blockchain-api/blob/8fdc9091d339ab1bc7a065f0f7175ea1eb369649/src/tests-2.1-transition/pox-transition.ts#L334-L342) for reproduction or as pseudo-code to follow exactly what's going on.

```ts
// We are still in Period 2a
expect(poxInfo.current_burnchain_block_height).toBeLessThan(poxInfo.contract_versions![1].activation_burnchain_block_height + 1);

expect(poxInfo.contract_id).toBe('ST000000000000000000002AMW42H.pox'); // FAILS
expect(poxInfo.contract_id).toBe(poxInfo.contract_versions![0].contract_id); // FAILS
```

> In the API-repo these `expects` are commented out to pass the tests for now...

donpdonp:
Change the reporting of the active contract to use pox_2 after v1_unlock_height instead of at v1_unlock_height.

Fixes #3422

### DIFF
--- a/src/burnchains/mod.rs
+++ b/src/burnchains/mod.rs
@@ -350,7 +350,7 @@ impl PoxConstants {
 
     /// Returns the PoX contract that is "active" at the given burn block height
     pub fn static_active_pox_contract(v1_unlock_height: u64, burn_height: u64) -> &'static str {
-        if burn_height >= v1_unlock_height {
+        if burn_height > v1_unlock_height {
             POX_2_NAME
         } else {
             POX_1_NAME
--- a/src/chainstate/coordinator/tests.rs
+++ b/src/chainstate/coordinator/tests.rs
@@ -4555,7 +4555,10 @@ fn test_epoch_verify_active_pox_contract() {
         let amount_locked_pox_1 = amount_locked_pox_1_res
             .expect("Should be able to query pox.clar for total locked ustx");
 
+        let active_pox_contract = b.pox_constants.active_pox_contract(burn_block_height);
+
         if burn_block_height <= pox_v1_unlock_ht.into() {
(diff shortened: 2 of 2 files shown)


obycode (approved):
LGTM

jcnelson (approved):
LGTM!

PR #3455, opened 2022-12-22
docs: fix index-of api reference

hugocaillard:
### Description
Update the `index-of?` documentation

It's consistent with `replace-at?`.

In the Clarity LSP, I remove the `?` for these two method in a Clarity1 context.
If future version include more aliases, we might want to think about a future proof way to handle it in the doc


obycode (approved):
Looks like I missed this when I made the changes that added the aliases. Thanks for catching it!

jcnelson (approved):

obycode:
Is it too late to slip this into rc1?

saralab:
@hugocaillard Can you merge these to feat/2.1.0.0.0-rc1 please ?

Issue #3133, opened 2022-05-11
Contract name length for Clarity 2


obycode:
It was discovered via a discussion on Discord that SIP 005 specifies that a contract name can be 128 characters

> The contract principal variant of the principal post-condition field is encoded as follows:
> A 1-byte type prefix of 0x03
> The 1-byte version of the standard principal that issued the contract
> The 20-byte Hash160 of the standard principal that issued the contract
> A 1-byte length of the contract name, up to 128
> The contract name

However the parser limits the contract name to 40 characters:
```rust
pub const CONTRACT_MIN_NAME_LENGTH: usize = 1;
pub const CONTRACT_MAX_NAME_LENGTH: usize = 40;

    pub static ref CONTRACT_NAME_REGEX: String = format!(
        r#"([a-zA-Z](([a-zA-Z0-9]|[-_])){{{},{}}})"#,
        CONTRACT_MIN_NAME_LENGTH - 1,
        CONTRACT_MAX_NAME_LENGTH - 1
    );
```

For Clarity 2, do we want to keep the 40 character limit or go back to the 128 character limit specified in the SIP?

jcnelson:
I'm in favor of keeping the 40 character limit and updating SIP 005.

jcnelson:
N.B. the transaction wire format allows 128 characters, so that should be fixed at 40.

zone117x:
Note this issue is more than just a documentation discrepancy. Contract call transactions with args containing large contract names (over 40 bytes) can be successfully included in a block. 

For an example, see `SP2C2YFP12AJZB4MABJBAJ55XECVS7E4PMMZ89YZR.aip10-arkadiko-update-tvl-liquidation-ratio` in https://explorer.stacks.co/txid/0xf402ee582892c48679e6bbfbdc1f12a1da5f1ebf705dfacd1982ad70296490f2?chain=mainnet

obycode:
This will be caught and reported as an error by the new parser.

jcnelson:
@obycode Can I confirm that this was fixed by the v2 parser?

obycode:
The new parser catches and reports an error when a contract principal in the code has a contract name longer than 40 characters - see https://github.com/stacks-network/stacks-blockchain/blob/3a24af692776c822f0ddb238c9362d09c6f6d12d/clarity/src/vm/ast/parser/v2/mod.rs#L2782.

[...]

PR #3462, opened 2023-01-02
test: add test passing long contract name

obycode:
This test verifies that the existing behavior where a contract name that
is longer than 40 characters, but less than 128 characters, can
successfully be passed to a contract call.

See: #3133


jcnelson (approved):

jcnelson:
Thanks @obycode!

PR #3463, opened 2023-01-03
fix(fuzz): Reduce false positives in fuzz-testing

fpbgg:
Supersedes #3438

### DIFF
--- a/clarity/src/vm/types/signatures.rs
+++ b/clarity/src/vm/types/signatures.rs
@@ -1216,20 +1216,20 @@ impl TypeSignature {
         }
     }
 
-    // This function checks recursively if NoType is anywhere in the type signature.
+    // Is the right hand side (rhs) an invalid type? Unlike the lhs, NoType is permitted.
     #[cfg(feature = "fuzzing")]
     pub fn contains_invalid_type_rhs(ty: &TypeSignature) -> bool {
         match ty {
             NoType => true,
             OptionalType(inner_type) => {
                 match inner_type.as_ref() {
-                    NoType => true, // NoType is a valid inner type for OptionalType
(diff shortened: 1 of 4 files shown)


clarity/src/vm/types/signatures.rs
```diff
@@ -1180,26 +1180,63 @@ impl TypeSignature {
...
+    }
+
+    // This function checks recursively if NoType is anywhere in the type signature.
+    #[cfg(feature = "fuzzing")]
+    pub fn contains_invalid_type_rhs(ty: &TypeSignature) -> bool {
+        match ty {
+            NoType => true,
+            OptionalType(inner_type) => {
+                match inner_type.as_ref() {
+                    NoType => true, // NoType is a valid inner type for OptionalType
```
obycode, line 1226: I thought that for the rhs, a `NoType` is valid here (like the comment says), so shouldn't this be `false`?

clarity/src/vm/types/signatures.rs
```diff
@@ -1180,26 +1180,63 @@ impl TypeSignature {
...
+                    inner_type => Self::contains_invalid_type_rhs(inner_type),
+                }
+            }
+            ResponseType(inner_types) => {
+                Self::contains_invalid_type_rhs(&inner_types.0)
+                    || Self::contains_invalid_type_rhs(&inner_types.1)
```
obycode, line 1232: Similarly here, I think it should only be invalid if both the ok-type and the err-type are invalid types.

clarity/src/vm/types/signatures.rs
fpbgg: Good catch. I think this was an error in recovering from the git rebase.

obycode (approved):
Thanks @fpbgg!

PR #3467, opened 2023-01-04
2.1.0.0.0-rc1

jcnelson:
This is the first release candidate of Stacks 2.1.

Note that this version does _not_ stipulate a 2.1 rules activation height!  This is effectively a 2.05-series node that uses the 2.1 chainstate schema and uses some new 2.1 code paths.

You can use it to boot up and sync with the mainnet chain tip, and then when -rc2, -rc3, etc. come out, you can reuse the chainstate this node produces (*).  Same for the stable 2.1 release.

Please test, test, test!

(*) Unless of course some truly catastrophic bug gets discovered that requires altering the schema.


wileyj (approved):
shipit

2 minor comments - changelog is missing a backtick for valid markdown and I've noticed some of the files contain different years in the copyright statement. 
i wonder if we can/should update that copyright statment so it doesn't have to be adjusted when a new year arrives, the alternative being to stop time itself - rendering blockchains ineffective.

PR #3468, opened 2023-01-05
2.1.0.0.0-rc1

jcnelson:
Creating a dedicated branch for this instead (so you can ignore #3467).  Will also rebase this to `develop` so we can go through the usual release process.

Miscellaneous fixes will be gathered here as we make them in response to testing on the mainline testnet and Stacks mainnet.


wileyj (approved):
shipit

- fixed the markdown typo in #3467 
- I've noticed some of the files contain different years in the copyright statement.
i wonder if we can/should update that copyright statment so it doesn't have to be adjusted when a new year arrives, the alternative being to stop time itself - rendering blockchains ineffective.

igorsyl (approved):
all integration tests passing.

obycode (approved):
LGTM 🚀

PR #3469, opened 2023-01-06
Fix: more efficient Merkle tree for Bitcoin

sergey-shandar:
### Applicable issues
- fixes #

### Checklist
- [ ] Changelog is updated

### DIFF
--- a/stacks-common/src/deps_common/bitcoin/util/hash.rs
+++ b/stacks-common/src/deps_common/bitcoin/util/hash.rs
@@ -463,7 +463,7 @@ impl MerkleTreeState {
         self.0.push(right);
         self
     }
-    fn collect(mut self) -> Sha256dHash {
+    fn calculate_hash(mut self) -> Sha256dHash {
         match self.0.pop() {
             None => Default::default(),
             Some(last) => {
@@ -477,14 +477,18 @@ impl MerkleTreeState {
     }
 }
 
-trait MerkleRootEx: Iterator<Item = Sha256dHash> + Sized {
+/// A mekrle hash for an iterator of SHA256.
+pub trait MerkleRootEx: Iterator<Item = Sha256dHash> + Sized {
+    /// Calculates a mekrle root hash for the given iterator of SHA256.
     fn merkle_root(self) -> Sha256dHash {
-        self.fold(MerkleTreeState::default(), MerkleTreeState::push).collect()
+        self.fold(MerkleTreeState::default(), MerkleTreeState::push)
+            .calculate_hash()
     }
 }
 
 impl<I: Iterator<Item = Sha256dHash>> MerkleRootEx for I {}
 
+/*
 /// Calculates the merkle root of a list of txids hashes directly
 pub fn bitcoin_merkle_root(data: Vec<Sha256dHash>) -> Sha256dHash {
     // Base case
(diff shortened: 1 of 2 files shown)


stacks-common/src/deps_common/bitcoin/util/hash.rs
```diff
@@ -412,31 +412,111 @@ pub trait MerkleRoot {
     fn merkle_root(&self) -> Sha256dHash;
 }
 
+fn merge(a: &Sha256dHash, b: &Sha256dHash) -> Sha256dHash {
```
jcnelson, line 415: Hey! Just an FYI, the code in `stacks-common/src/deps_common` is vendored code.  In this case, it's vendored from `rust-bitcoin`.  We'd prefer to not modify it, except by pulling in new versions of the upstream code or patching security bugs.

sergey-shandar: Good to know, thank you! What's the reason we use copy the code instead of reference to crates/GitHub?

jcnelson: Two reasons:

* We didn't want everything in the `rust-bitcoin` crate, so we stripped out what we didn't need.

* We had to fix at least on security bug in `rust-bitcoin`

xoloki: Agreed Sergey.   I have a particular dislike for vendored code.  If we want to modify a repo we should fork it, then use cargo to point to the fork.

sergey-shandar: And if we need an urgent security fix. We can fork their repo and reference directly to the repo.

```toml
[dependencies]
rust-bitcoin = { git = "https://github.com/stacks-network/rust-bitcoin" }
```

jcnelson: >  we should own forked code.

That's effectively what we're doing.  But, in this case putting it in a separate repo just creates needless organizational complexity.

You see, we only rely on the vendored code for decoding Bitcoin block headers, blocks, transactions, and scripts.  This is all consensus-critical code for Bitcoin, so it's not ever going to change.  In the past (e.g. Blockstack, Stacks 1.0) we simply wrote the code ourselves.  Cribbing the same code from `rust-bitcoin`, and leaving out all the parts we don't need, simply accelerated the development process without forcing us to take on the full maintenance burden for a `rust-bitcoin` fork.

Now, we have fixed at least one bug in this code that could lead to a runtime panic given a specially-crafted Bitcoin block.  While there is an upstream fix for this, the version of the library in which the fix landed is not API-compatible with the vendored code here.  So, the consequences of the choice between keeping the code here, versus maintaining an out-of-date fork, versus tracking the latest `rust-bitcoin` library versions, are:

* Apply a 1-line fix to repair the DoS vulnerability and ship it with a point release of the Stacks blockchain (this is what we did)
* Apply a 1-line fix to repair the DoS vulnerability in a separate repo, and ship both a point release of the forked library and a point release of the Stacks blockchain
* Refactor and test substantial portions of the Bitcoin block indexer code to use the latest `rust-bitcoin` and ship it in a point release

As you can see, the first option requires the least amount of effort.  So, I see no reason to change what we're doing right now -- there's no real difference between maintaining a few modules of cribbed code from `rust-bitcoin` related to parsing data structures whose formats are not going to change in the foreseeable future, and simply writing those modules ourselves.

PR #3473, opened 2023-01-09
Fix: more hardcoded test names

netrome:
I stumbled upon some more tests with concurrency issues, with the same old root cause of copy-pasted hard-coded test names. Took the opportunity to spread the usage of our `function_name!()` macro.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,13 @@ $ cargo build
 $ cargo test testnet  -- --test-threads=1
 ```
 
+**Run all unit tests in parallel using [nextest](https://nexte.st/):**
+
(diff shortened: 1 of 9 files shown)


netrome:
> @netrome Can you re-open this against `develop` instead? That's where the 2.1 work is happening now. Thanks!

Done.

igorsyl (approved):
LGTM. Could you add to the docs how to run tests in parallel using nextest?

netrome:
> LGTM. Could you add to the docs how to run tests in parallel using nextest?

Sure thing, updated now.

PR #3476, opened 2023-01-11
[Signer] sBTC Devnet: stacks-signer DKG phase

igorsyl:


testnet/stacks-signer/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+crate-type = ["lib"]   # The crate types to generate.
+
+[dependencies]
+sha3 = "0.10.6"
+serde = { version = "1.0", features = ["serde_derive"] }
+toml = "0.5.9"
+hashbrown = { version = "0.13", features = ["serde"] }
+frost = { git = "https://github.com/trust-machines/frost", branch = "multiple-parties-per-signer" }
+rand_core = { version  = "0.5"}
+tracing = "0.1.37"
```
jcnelson, line 24: Have you confirmed that these dependencies are free of gratuitous `unsafe` code?

* `tracing`
* `tracing-subscriber`
* `ureq`

testnet/stacks-signer/src/net.rs
```diff
@@ -0,0 +1,62 @@
...
+pub trait Net {
+    fn listen(&self);
+    fn poll(&self) -> Result<Response, ureq::Error>;
+    fn next_message(&self) -> Message;
+    fn send_message(&self, _msg: Message);
+}
+
+#[derive(Debug)]
+pub struct Message {
+    pub r#type: signer::MessageTypes,
```
jcnelson, line 20: Style nit: please use variable names that don't collide with reserved keywords.  Thanks!

testnet/stacks-signer/src/net.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+impl Debug for signer::MessageTypes {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Message Type: _")
```
jcnelson, line 25: Shouldn't this print out the message instance, instead of `_`?

jcnelson (commented):
Meta-comment: there's basically zero code comments in `stacks-signer`.  Can you at least do the following?

* Add per-member struct doc strings (one liners are fine)?
* Add descriptive doc strings for each method, so the reader knows what the API contract is?

testnet/stacks-signer/Cargo.toml
donpdonp: ureq is clean, from the ureq Readme: "Ureq is in pure Rust for safety and ease of understanding. It avoids using `unsafe` directly. It [uses blocking I/O][blocking] instead of async I/O, because that keeps the API simple and keeps dependencies to a minimum. For TLS, ureq uses [rustls or native-tls](#https--tls--ssl)."

tracing I added just to have something for logging. grepping the tracing codebase finds lots of unsafe references. I'll take tracing out of Cargo.toml.

testnet/stacks-signer/src/net.rs
donpdonp: the code was pushed over from trustmachines in an in-process state. i'll wrap up these loose ends and use a feature branch that merges into sbtc from here on.

testnet/stacks-signer/Cargo.toml
```diff
@@ -21,11 +21,13 @@ toml = "0.5.9"
 hashbrown = { version = "0.13", features = ["serde"] }
 frost = { git = "https://github.com/trust-machines/frost", branch = "multiple-parties-per-signer" }
```
xoloki, line 22: You should switch to main, v1 is merged.

xoloki (changes requested):

Cargo.toml
```diff
@@ -95,7 +95,7 @@ version = "=0.24.2"
 features = ["blob", "serde_json", "i128_blob", "bundled", "trace"]
 
 [dependencies.ed25519-dalek]
-version = "=1.0.0-pre.3"
+version = "=1.0.1"
 features = ["serde"]
 
 [dependencies.curve25519-dalek]
```
xoloki, line 101: Out of curiosity, why are we using Dalek?  My wrapped secp256k1 is significantly faster, and also maintains compatibility with bitcoin signatures.

jcnelson: We use it to implement a VRF described here: https://datatracker.ietf.org/doc/id/draft-irtf-cfrg-vrf

xoloki: Do you want me to try replacing dalek with my secp wrappers?  It's basically API compatible, by intent.

jcnelson: I don't think it's high priority.  There's only one VRF proof generated per miner per block produced, and there is only one VRF proof verification per Stacks block, so a performance speed-up won't do very much for overall block-processing performance.  Also, any new implementation needs to be "bug-for-bug" compatible with the current implementation, since VRF proof creation and verification are on consensus-critical code paths.

xoloki:
Ok @igorsyl @donpdonp @jcnelson I had to update the codecov docker file to get working clang libs, but now frost and secp-math are building correctly.  The tests are still failing but in ```stacks-signer```.

xoloki:
@jcnelson  why are we checking in Cargo.lock?  Can't we just specify appropriate versions in Cargo.toml?

testnet/stacks-signer/src/bin/coordinator.rs
```diff
@@ -0,0 +1,163 @@
...
+#[command(author, version, about, long_about = None)]
+struct Cli {
+    #[command(subcommand)]
+    command: Command,
+}
+
+#[derive(clap::Subcommand, Debug)]
+enum Command {
+    Dkg,
+    Sign { msg: String },
```
xoloki, line 42: msg should be &[u8]

xoloki: More specifically, store it as Vec<u8>, and pass as &[u8]

xoloki:
Since you're reworking the parameter names, let me give some additional context.

There are 3 parameters: signer id, party id, and key id.  In v1, there is no signer id, and party id = key id.  In v2, signer id = party id, and key id is separate.

So for maximum compatibility with v1 and v2, you should use party id when talking about signers, and key id when talking about reward slots.  That will simplify things greatly if/when we want to switch versions, or support both.

For dkg, the public shares are keyed on party id, while private shares are keyed on key id.  It won't matter for v1, but it will for v2.

For party sign, nonces and sigs are keyed on party id.  

It's probably better to get this sorted now.

xoloki:
In all cases, threshold is a function of key IDs, though for v1 it doesn't matter since key id = party id

igorsyl:
Development of this component has moved to https://github.com/Trust-Machines/core-eng/.
We will later open PR's here to integrate back code.

PR #3478, opened 2023-01-11
feat: ability to compile clarity-vm to wasm

lgalabru:


hugocaillard:
I think we can close in favor of #3748? (That confusing it's almost the same id as this one)

Issue #3431, opened 2022-12-05
Remove From<&'_ str> implementation for `guarded_string` macro


netrome:
`From` is intended for perfect/infallible conversions. The `From<&'_ str>` implementation in the `guarded_string` macro is not as it may panic.

**Why is this a problem?**
Implementing `From` also implicitly implements `TryFrom` ([link](https://doc.rust-lang.org/src/core/convert/mod.rs.html#600)). This gives us an implicit `TryFrom<&'_ str`> implementation which panics on errors that is easy to accidentally call when intending to call `TryFrom<String>`.

We already have multiple instances of this in the code, such as in `clarity/./src/vm/database/clarity_db.rs:715`:
```
ClarityName::try_from("reporter").expect("BUG: valid string representation")
```

another example is in `src/net/mod.rs:2662`
```
let smart_contract =
  TransactionPayload::SmartContract(TransactionSmartContract {
    name: ContractName::try_from(conf.test_name.as_str())
    .expect("FATAL: invalid boot-code contract name"),
```

The `expect()` statements in these examples are unreachable, even though it is evident that this was not the intention of the developer.

**Proposal**
Remove the `From<&'_ str>` implementation altogether. It provides more risks than value. The few legitimate use-cases can easily be refactored to `try_from(...).unwrap()` or `try_into(...).unwrap()`.

netrome:
> IIRC the reason `guarded_string` uses `From` is because there are lots of cases in the code where we use `.into()` to convert a `&str` literal into a `guarded_string`, or to convert a stored but already-checked string back into a `guarded_string`. _These cases should panic if they fail_, because their failure indicates database corruption.

I see. In those cases I'd argue an `.try_into().unwrap()` conveys the intention better to the reader. If this is too verbose, we could add a shorthand for this that avoids the risks of the current implementation. The ideal solution for string literals would of course be to enforce the formatting rules at compile time.

jcnelson:
Temporarily assigning to @netrome for now.  Please feel free to reassign.  If you don't think you'll get to this for some time, it might be better to just copy this issue text to a comment in #3412 and close it out here.

netrome:
Great. I'm happy to implement this. When we last discussed this you wanted to await Aaron to sanity check this idea before jumping to any implementation.

[...]

PR #3486, opened 2023-01-12
[sBTC] Feat: Peg-in operation wire format

netrome:
This PR introduces an alpha version of the wire format for the Peg-in operation, intended for early end-to-end testing of sBTC. This version do not validate the peg wallet address, which is left to an external stacker application for the alpha test phase.

I've created #3494 as a follow-up task to add the validation of the peg wallet address once Epoch 3.0 and the updated `.pox` contract is ready.


igorsyl:
Adding @donpdonp, @sergey-shandar and @xoloki as reviewers to become familiar with this code.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -699,6 +725,20 @@ const SORTITION_DB_SCHEMA_4: &'static [&'static str] = &[
...
 const LAST_SORTITION_DB_INDEX: &'static str = "index_burn_header_hash_pox_valid";
 
+const SORTITION_DB_SCHEMA_5: &'static [&'static str] = &[r#"
+    CREATE TABLE peg_in (
+        txid TEXT NOT NULL,
+        vtxindex INTEGER NOT NULL,
+        block_height INTEGER NOT NULL,
+        burn_header_hash TEXT NOT NULL,
+
+        address TEXT NOT NULL,
```
jcnelson, line 735: What's `address` here again?  Can you add a (sql) comment?  Thanks!

src/chainstate/burn/operations/mod.rs
```diff
@@ -308,6 +313,19 @@ pub struct DelegateStxOp {
     pub burn_header_hash: BurnchainHeaderHash, // hash of the burn chain block header
 }
 
+#[derive(Debug, PartialEq, Clone, Eq, Serialize, Deserialize)]
+pub struct PegInOp {
+    pub address: StacksAddress,
```
jcnelson, line 318: I know SIP-021's wire format appendix says that this is a Stacks address, but the text itself says that this can also be a smart contract address.  Can you make this the case?

Bytes 3-24 would be the address portion (1 byte version, 20 byte hash), and bytes 24-64 would be the name portion.

The name would be a smart contract name, so it would be between 5 and 40 bytes long.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -10417,3 +10420,93 @@ fn microblock_miner_multiple_attempts() {
 
     channel.stop_chains_coordinator();
 }
+
+#[test]
+#[ignore]
+fn test_submit_and_observe_peg_in_request() {
```
jcnelson, line 10426: While what you have here is great, I just wanted to let you know that there are lighter-weight operations available for testing burnchain operations.  For example, see `test_pox_btc_ops()` and `test_stx_transfer_btc_ops()` in `src/chainstate/coordinator/tests.rs`.

src/chainstate/burn/db/sortdb.rs
netrome: Sure, it's the stacks address to receive the `sBTC`. I can add a comment.

Hmm, I think `recipient` would be a better field name throughout the code. I'll rename the field and add a SQL comment for it. Let me know if you have an even better name in mind.

src/chainstate/burn/operations/mod.rs
netrome: Sure thing, I'll update.

testnet/stacks-node/src/tests/neon_integrations.rs
netrome: Thanks! I'll study those test cases and see if it's worth complementing the existing tests with something more lightweight to further future-proof the code. Since the neon integration tests are slow, the more code paths we can cover in lighter tests the better.

src/chainstate/burn/operations/mod.rs
netrome: I'll represent the smart contract name as a String for the alpha version, leaving room for a future discussion around improved type safety for contract addresses.

The reason is that I have not found any good existing types to represent smart contract addresses. The best type I know of is `clarity::vm::types::PrincipalData`, which uses `ContractName` to validate contract names. However, this would make `chainstate::burn::operations` depend on the `clarity` crate, which I would not like to do on a whim. Also, `ContractName` can introduce unintentional panics as observed in #3431, which would risk bringing down a running node on a malformed transaction.

I guess ideally it would be best to have type alongside `StacksAddress` in `stacks_common::types::chainstate` to represent contract addresses.

Let me know any thoughts on this.

testnet/stacks-node/src/tests/neon_integrations.rs
netrome: I added a test case for the peg-in operation in the `chainstate::coordinator::tests` now to complement the existing test cases.

src/chainstate/burn/db/sortdb.rs
netrome: ✅ Done.

testnet/stacks-node/src/tests/neon_integrations.rs
netrome: ✅ Done

src/chainstate/burn/operations/mod.rs
netrome: We discussed this on an sBTC working group meeting today. We're okay with depending on clarity types in `stacks_common::types::chainstate`, so I will change this ti use the `PrincipalData` type.

netrome: Done :heavy_check_mark:

netrome:
@jcnelson FYI I'm not authorized to merge this PR. Can you add me as maintainer so I can merge once it's approved and CI is green?

clarity/src/vm/types/mod.rs
```diff
@@ -1397,6 +1397,15 @@ impl From<StandardPrincipalData> for StacksAddress {
     }
 }
 
+impl From<PrincipalData> for StacksAddress {
```
jcnelson, line 1400: Please remove this conversion. It should never be possible to convert a smart contract address into a standard address.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -6127,6 +6224,50 @@ pub mod tests {
...
+            &mut db,
+            BurnchainHeaderHash([0x01; 32]),
+            &vec![BlockstackOperationType::PegIn(peg_in.clone())],
+        );
+
+        let res_peg_ins = SortitionDB::get_peg_in_ops(db.conn(), &burn_header_hash)
+            .expect("Failed to get peg-in ops from sortition DB");
+
+        assert_eq!(res_peg_ins.len(), 1);
+        assert_eq!(res_peg_ins[0].amount, 1337);
```
jcnelson, line 6268: Can you add test coverage to verify that two different burnchain blocks can contain the same transaction (i.e. with the same txid), but that `SortitionDB::get_peg_in_ops()` will return the burnchain block-specific operation metadata for a given burnchain block hash?

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -0,0 +1,451 @@
...
+        let maybe_contract_name: Option<ContractName> = data
+            .get(21..)
+            .filter(|bytes| !bytes.is_empty())
+            .map(std::str::from_utf8)
+            .transpose()?
+            .map(ToOwned::to_owned)
+            .map(TryInto::try_into)
+            .transpose()?;
+
+        println!("Contract name: {:?}", maybe_contract_name);
```
jcnelson, line 92: Please remove or use the logging facilities.

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -0,0 +1,451 @@
...
+
+        let version = *data.get(0).ok_or(ParseError::MalformedData)?;
+        let address_data: [u8; 20] = data
+            .get(1..21)
+            .ok_or(ParseError::MalformedData)?
+            .try_into()?;
+
+        let standard_principal_data = StandardPrincipalData(version, address_data);
+
+        let maybe_contract_name: Option<ContractName> = data
```
jcnelson, line 83: This code does not handle the case if there are trailing 0-bytes, or even the case where the `name` field is all 0's.  I think it should, because the space beyond the 64th byte can be used as a memo field for future extensions.  Instead, this code should just select out the leading non-zero bytes from the range 21 to 61, and try to decode _that_ into a ContractName.  If there are trailing 0's, it's fine -- they'll be ignored.  If the whole name field is 0, then there's no name.

Also, please don't use such long operator chains!  It's very hard to follow what this code is supposed to be doing.  Just do something like this:

```rust
let maybe_contract_name: Option<ContractName> = {
   // read to the first 0 byte
   let strlen = {
      let mut len = cmp::min(61, data.len());
      for i in 21..len {
         if data[i] == 0 {
            len = i;
            break;
          }
       }
       len
   };
      
   let name_bytes = data.get(21..strlen).unwrap_or(&[]);
   let name_string = std::str::from_utf8(&name_bytes).map_err(|_| ParseError::MalformedData)?.to_string();
   let name = ContractName::try_from(name_str).map_err(|_| ParseError::MalformedData)?;
   name
}
```

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -0,0 +1,451 @@
...
+        create_op(1)
+            .check()
+            .expect("Any strictly positive amounts should be ok");
+
+        create_op(u64::MAX)
+            .check()
+            .expect("Any strictly positive amounts should be ok");
+    }
+
+    // TODO: Invalid contract name test
```
jcnelson, line 376: Please complete this task before merging.

src/chainstate/coordinator/tests.rs
```diff
@@ -3236,6 +3236,181 @@ fn test_stx_transfer_btc_ops() {
...
+
+    let total_number_of_peg_in_ops = burnchain_block_hashes
+        .into_iter()
+        .flat_map(|block_hash| {
+            SortitionDB::get_peg_in_ops(&sort_db.conn(), &block_hash)
+                .expect("Failed to get peg in ops")
+        })
+        .count();
+
+    assert_eq!(total_number_of_peg_in_ops, 1);
```
jcnelson, line 3411: Please also check that it's the one you expect.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -10417,3 +10420,93 @@ fn microblock_miner_multiple_attempts() {
...
+    let receiver_stx_addr = StacksAddress::new(0, Hash160([0; 20]));
```
jcnelson, line 10427: Please also add test coverage for receiving to a contract principal.

jcnelson:
A meta-comment.  When decoding a burnchain operation that is represented in fewer than 80 OP_RETURN bytes, you should save the remaining bytes into a `memo: Vec<u8>` and store them.  We use these "memo" bytes for in-band signaling between miners.

src/chainstate/burn/operations/peg_in.rs
netrome: Ouch, this was unintentional. Thanks for spotting.

netrome:
> A meta-comment. When decoding a burnchain operation that is represented in fewer than 80 OP_RETURN bytes, you should save the remaining bytes into a `memo: Vec<u8>` and store them. We use these "memo" bytes for in-band signaling between miners.

Makes sense. I'll add it here. This memo field also seems to be missing in the `PreStxOp`, `StackStxOp` and `DelegateStxOp` operations. I created #3530 as a follow-up to add it for those operations too. @jcnelson feel free to close the issue if there's any reason these ops should not have the memo field.

src/chainstate/burn/operations/peg_in.rs
netrome: Alright. I factored out some pieces of the suggestion to a `leading_non_zero_bytes` utility function and updated the functionality according to your description. It should be more readable now without the long operator chain. Let me know if you have any further opinions.

clarity/src/vm/types/mod.rs
netrome: Done.

src/chainstate/burn/operations/peg_in.rs
netrome: Done. I had already added the test case, but forgot to clean up the stale todo.

src/chainstate/burn/db/sortdb.rs
netrome: Done. I extended the test case to submit two peg in operations with different amounts and burn header hashes.

src/chainstate/coordinator/tests.rs
netrome: Done :heavy_check_mark:

src/chainstate/burn/db/sortdb.rs
```diff
@@ -699,6 +731,23 @@ const SORTITION_DB_SCHEMA_4: &'static [&'static str] = &[
...
+        block_height INTEGER NOT NULL,
+        burn_header_hash TEXT NOT NULL,
+        sortition_id TEXT NOT NULL,
+
+        recipient TEXT NOT NULL,            -- Stacks principal to receive the sBTC, can also be a contract principal
+        peg_wallet_address TEXT NOT NULL,
+        amount TEXT NOT NULL,
+        memo TEXT,
+
+        PRIMARY KEY(txid, sortition_id)
```
jcnelson, line 747: Hey, I know I said that this needs to be `(txid, sortition_id)` earlier, but it turns out I misspoke.  It should be `(txid, burn_header_hash)`.  It turns out that my PR #3528 is wrong as well.

jcnelson: I left an explanation as to why #3528 is currently mis-designed here: https://github.com/stacks-network/stacks-blockchain/pull/3528#issuecomment-1408859705

netrome: Alright, thanks for letting me know. I've updated the PR now using `(txid, burn_header_hash)` as the new primary key.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -699,6 +731,21 @@ const SORTITION_DB_SCHEMA_4: &'static [&'static str] = &[
 // update this to add new indexes
 const LAST_SORTITION_DB_INDEX: &'static str = "index_burn_header_hash_pox_valid";
```
jcnelson, line 732: Please change this constant to `index_peg_in_burn_header_hash`.

src/chainstate/burn/operations/mod.rs
```diff
@@ -308,6 +316,20 @@ pub struct DelegateStxOp {
     pub burn_header_hash: BurnchainHeaderHash, // hash of the burn chain block header
 }
 
+#[derive(Debug, PartialEq, Clone, Eq, Serialize, Deserialize)]
+pub struct PegInOp {
+    pub recipient: PrincipalData,
+    pub peg_wallet_address: PoxAddress,
+    pub amount: u64,   // BTC amount to peg in, in satoshis
+    pub memo: Vec<u8>, // extra unused byte
```
jcnelson, line 324: This field can be more than 1 byte long. Please update the comment.

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -0,0 +1,470 @@
+use stacks_common::codec::StacksMessageCodec;
```
jcnelson, line 1: Please add the copyright statement that we use in all the other source files to this one.

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -0,0 +1,470 @@
...
+        if tx.opcode() != Opcodes::PegIn as u8 {
+            warn!("Invalid tx: invalid opcode {}", tx.opcode());
+            return Err(OpError::InvalidInput);
+        }
+
+        let (amount, peg_wallet_address) =
+            if let Some(Some(recipient)) = tx.get_recipients().first() {
+                (recipient.amount, recipient.address.clone())
+            } else {
+                warn!("Invalid tx: Output 2 not provided");
```
jcnelson, line 32: N.B. This can also fail if the output is not recognized (i.e. is non-standard).

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -0,0 +1,470 @@
...
+        if data.len() < 21 {
+            warn!(
+                "PegInOp payload is malformed ({} bytes, expected at least {})",
+                data.len(),
+                21
+            );
+            return Err(ParseError::MalformedData);
+        }
+
+        let version = *data.get(0).ok_or(ParseError::MalformedData)?;
```
jcnelson, line 77: Because we know that `data.len() >= 21`, you can instead use `.expect()` here instead of `.ok_or()`.

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -0,0 +1,470 @@
...
+
+    pub fn check(&self) -> Result<(), OpError> {
+        if self.amount == 0 {
+            warn!("PEG_IN Invalid: Peg amount must be positive");
+            return Err(OpError::PegInAmountMustBePositive);
+        }
+
+        Ok(())
+    }
+
```
jcnelson, line 108: Please add a docstring.

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -0,0 +1,470 @@
...
+    fn leading_non_zero_bytes(data: &[u8], from: usize, to: usize) -> Option<&[u8]> {
```
jcnelson, line 109: A defensive check that `from < to` might be appropriate here.

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -0,0 +1,470 @@
...
+        let end_of_non_zero_slice = {
+            let mut end = to.min(data.len());
+            for i in 21..end {
```
jcnelson, line 112: Bug -- should be `for i in from..end`, not 21.

src/chainstate/coordinator/tests.rs
```diff
@@ -3236,6 +3236,187 @@ fn test_stx_transfer_btc_ops() {
     }
 }
 
+#[test]
+fn test_sbtc_peg_in_btc_op() {
+    let path = "/tmp/stacks-blockchain-sbtc-peg-in-btc-ops";
+    let _r = std::fs::remove_dir_all(path);
```
jcnelson, line 3242: Can you just `.unwrap()` this so the test fails if the filesystem is inaccessible for this test?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -10417,3 +10422,151 @@ fn microblock_miner_multiple_attempts() {
...
+                &mut miner_signer,
+                1
+            )
+            .is_some(),
+        "Peg-in operation should submit successfully"
+    );
+
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    let parsed_peg_in_op_standard = {
```
jcnelson, line 10516: This is fine for now, but this information should also be obtainable from the event observer.  Please open an issue so we can remember to go back and add it.  Thanks!

jcnelson (commented):
This is mostly in good shape.  Just a few minor things to do and I'll approve.  Thanks @netrome!

src/chainstate/burn/db/sortdb.rs
netrome: Done ✅

src/chainstate/burn/db/sortdb.rs
```diff
@@ -2869,6 +2918,18 @@ impl SortitionDB {
         Ok(())
     }
 
+    fn apply_schema_5(tx: &DBTx) -> Result<(), db_error> {
```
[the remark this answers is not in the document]
netrome: Alright, I've updated `is_db_version_supported_in_epoch` now to only allow schema 5 in Epoch 2.1

src/chainstate/burn/operations/mod.rs
netrome: Oops, thanks for spotting. Fixed!

src/chainstate/burn/operations/peg_in.rs
netrome: Done.

src/chainstate/burn/operations/peg_in.rs
netrome: Thanks for clarifying. I've updated the error message.

src/chainstate/burn/operations/peg_in.rs
netrome: Done.

src/chainstate/burn/operations/peg_in.rs
netrome: Makes sense. I added an assert since an error would indicate a severe bug in the code at which point a panic seems warranted. Let me know if you prefer a normal error here instead.

src/chainstate/burn/operations/peg_in.rs
netrome: Thanks for spotting. Fixed.

src/chainstate/burn/operations/peg_in.rs
netrome: Done.

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -0,0 +1,470 @@
...
+
+        let op = PegInOp::from_tx(&header, &tx).expect("Failed to construct peg-in operation");
+
+        assert_eq!(op.recipient, stx_address.into());
+        assert_eq!(op.amount, amount);
+        assert_eq!(op.peg_wallet_address.bytes(), peg_wallet_address);
+    }
+
+    #[test]
+    fn test_parse_peg_in_should_succeed_given_a_contract_recipient() {
```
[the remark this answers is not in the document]
netrome: Done. I've added the memo to most of the other tests as well.

src/chainstate/coordinator/tests.rs
netrome: I can't simply unwrap, since that would cause the test to panic if the directory do not exist. I'm not a super fan of this pattern of ignoring results either but it's repeated in most test cases in this module so I'd recommend any changes to this pattern to be solved as a separate PR.

testnet/stacks-node/src/tests/neon_integrations.rs
netrome: Sure, I created #3554

netrome:
> This is mostly in good shape. Just a few minor things to do and I'll approve. Thanks @netrome!

Thank you @jcnelson , I have addressed all your comments now. Let me know if everything looks good now so we can move this forward :rocket:

igorsyl:
In `stx_transfer_btc_integration_test`:
```
 #5 159.5 thread 'tests::neon_integrations::stx_transfer_btc_integration_test' panicked at 'unexpected btc transaction type', testnet/stacks-node/src/tests/neon_integrations.rs:1709:21
```

jcnelson:
cc @pavitthrap who touched this code path last

netrome:
@jcnelson @igorsyl why was `develop` merged into this PR? This PR is based on `next`, and merging in unrelated changes makes the version history unnecessarily complex and hard to follow. I'm happy to rebase if there's any new changes in `next` that we need to get this through. Otherwise, let's try to separate refactor noise from feature additions as already agreed.

netrome:
> In `stx_transfer_btc_integration_test`:
> 
> ```
>  #5 159.5 thread 'tests::neon_integrations::stx_transfer_btc_integration_test' panicked at 'unexpected btc transaction type', testnet/stacks-node/src/tests/neon_integrations.rs:1709:21
> ```

This is unrelated to the peg-in changes, and was introduced when `develop` was merged (which I've reverted now). This test failure needs to be addressed to unblock https://github.com/stacks-network/stacks-blockchain/pull/3564. FYI @pavitthrap

netrome:
Rebased on latest `next` and CI is green :heavy_check_mark:. @jcnelson I'd love it if you could have a look if this is okay to merge.

jcnelson (approved):
LGTM!  Great job on this @netrome!

netrome:
> LGTM! Great job on this @netrome!

Thank you! I'm not currently authorized to merge, can you press the magic button for me @jcnelson?

Issue #3489, opened 2023-01-12
[Stacks 2.1] Node crashing during epoch 2.1 transition


lgalabru:
I have a node crashing during the epoch 2.1 transition - it looks like we're hitting a code path marked as `unreachable`:

```rust
    fn admits_type_v2_1(&self, other: &TypeSignature) -> Result<bool> {
            ...
            NoType => Err(CheckErrors::CouldNotDetermineType),
            TraitReferenceType(_) => {
                unreachable!("TraitReferenceType should not be used in epoch v2.1")
            }
        }
    }
```

Note: when the node is trying to assemble this block, my mempool contains both clarity 1 and clarity 2 contracts.

[Full debug log](https://github.com/stacks-network/stacks-blockchain/files/10405742/debg-trace.txt)

obycode:
I'm on it. Thanks Ludo.

jcnelson:
From the blockchain meeting:

Add an internal error type to capture all cases that are currently deemed unreachable, so at least the network doesn't crash each time we find something like this.

jcnelson:
This was fixed in #3490.  The internal error type will be a separate issue.

PR #3490, opened 2023-01-13
fix: handle `TraitReferenceType` in 2.1

obycode:
### Description

A trait processed during epoch 2.05 which contains a trait type for one
of its parameters will have a `TraitReferenceType` in its function
signature. This behavior changed when we split off the updated
type-checker and continued to run the old one in 2.05, but this one case
was not modified accordingly.

Fixes: #3489

### Applicable issues
- fixes #

### Checklist
- [ ] Changelog is updated

### DIFF
--- a/clarity/src/vm/types/signatures.rs
+++ b/clarity/src/vm/types/signatures.rs
@@ -645,6 +645,22 @@ impl TypeSignature {
         }
     }
 
+    /// Canonicalize the type. This method will replace any types from previous
+    /// epochs with the appropriate types for the current epoch.
+    pub fn canonicalize(&self, epoch: &StacksEpochId) -> TypeSignature {
+        match epoch {
+            StacksEpochId::Epoch21 => self.canonicalize_v2_1(),
+            _ => self.clone(),
+        }
+    }
+
+    pub fn canonicalize_v2_1(&self) -> TypeSignature {
+        match self {
+            TraitReferenceType(trait_id) => CallableType(CallableSubtype::Trait(trait_id.clone())),
+            _ => self.clone(),
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


jcnelson (approved):
LGTM!

obycode:
@jcnelson does your node that you are booting from genesis include this PR as well?

clarity/src/vm/types/signatures.rs
```diff
@@ -641,13 +641,26 @@ impl TypeSignature {
...
+        }
+    }
+
+    pub fn canonicalize_v2_1(&self) -> TypeSignature {
+        match self {
+            TraitReferenceType(trait_id) => CallableType(CallableSubtype::Trait(trait_id.clone())),
+            _ => self.clone(),
+        }
+    }
+
```
kantai, line 663: Does this work if the TraitReferenceType is nested in an optional?

obycode: It would get handled, but only because the function `canonicalize` is called from (`clarity2_inner_type_check_type`) is recursing into the type. Would it be more clear if this method itself recursed into the type and was called somewhere else? Or maybe if this function was just renamed to `canonicalize_simple_type` or canonicalize_base_type`?

kantai: Yeah -- I think just renaming and adding a note to the rustdocs would be sufficient here. As it's currently used, it's fine, we should just make sure that a future caller doesn't think that it will canonicalize a complex type.

obycode: Thanks. Added in f5567b216.

kantai (approved):
LGTM!

Issue #3471, opened 2023-01-08
[Stacks 2.1] Consensus bug: Stacks 2.1 reserved variables can be used in Clarity 1


jcnelson:
In testing a Stacks 2.1 node spinning up from genesis on the testnet, I noticed that gets stuck processing this transaction: https://explorer.stacks.co/txid/0x451f27420bd51bc858ff53303f4f243f2c5e943f55d4d2c684669b07ba7f0da9?chain=testnet.  It generates the wrong state root hash for this block, and specifically, it fails to calculate the correct value for `chain-nonce`.  Digging deeper into this, it appears that the use of local variable `chain-id` is being overwritten by the 2.1 VM's native variable `chain-id`, leading to the wrong value being stored.

I have identified the cause of the bug in the source code.  The native variable lookup is not gating on Clarity version.  I'm testing a locally-patched 2.1 testnet node right now; will report back in a day or so when it reaches this block height again.

jcnelson:
This is confirmed fixed in `feat/2.1.0.0.0-rc2`

PR #3492, opened 2023-01-13
Feat/2.1.0.0.0 rc2

### LINKED ISSUE #3470: [Stacks 2.1] Consensus bug: incorrect cost calculation for comparators
When comparing the calculated runtime costs for Stacks blocks between 2.05 and 2.1, I noticed that in many cases, the `runtime` field was off by a little bit.  Looking further, it appears that Stacks 2.1 does not implement cost calculations for the comparator functions (i.e. `<`, `<=`, `>`, and `>=`) the exact same way.  There could be other such instances.

This issue can only be closed once we determine that the Stacks 2.1 node calculates the same runtime costs for both mainnet and testnet for all blocks.  Additional unit tests may be needed to verify that runtime costs remain the same between versions.

### LINKED ISSUE #3480: [Stacks 2.1] Consensus bug: consider 2.05 epoch rules for selecting anchor blocks when processing a PoX reorg
When booting a 2.1 node in testnet, the 2.1-series logic for determining if an anchor block was selected returns a different result than the logic for 2.05.  Specifically, the 2.1 algorithm indicates that reward cycle 360 does _not_ have an anchor block, whereas the (correct) 2.05 logic indicates that there is indeed an anchor block.

This by itself isn't problematic -- this inconsistency is expected, because 2.1 changes the rules slightly.  The problem is that when the 2.1 epoch activates on testnet, the node will execute a deep reorg when it realizes that the heaviest affirmation map and the Stacks tip affirmation map are both incompatible with the sortition affirmation map at cycle 360.  The chains coordinator will try to unwind and rewind the chain in a bid to "correct" the sortition DB to declare that the anchor block for cycle 360 is absent, when in fact it is present under the 2.05 rules.  This is because the heaviest and Stacks affirmation map calculations exclusively rely in the 2.1 rules, and the chains coordinator requires that both the sortition and Stacks affirmation maps be compatible with the heaviest affirmation map (i.e. the one determined by affirmed block-commits).

The fix would be to specifically treat the case where the sortition affirmation map in epoch 2.05 indicates that the anchor block is `p`resent as _compatible_ with a Stacks and heaviest affirmation map that treats the anchor block as `a`bsent in this epoch.

jcnelson:
This PR fixes #3480, #3471, and #3470.

Once this merges, we can proceed to doing a "dress rehearsal" whereby we test migrating testnet and mainnet to 2.1.

Once we're confident with the dress rehearsals, we can pick the activation heights and cut `-rc3`.


jcnelson:
This branch has sync'ed with mainline testnet and has calculated all of the correct runtime costs.

obycode (approved):

src/chainstate/burn/db/sortdb.rs
```diff
@@ -3009,6 +3009,30 @@ impl SortitionDB {
...
+        let affirmation_map = match self.get_indexed(chain_tip, &db_keys::pox_affirmation_map())? {
+            Some(am_str) => {
+                AffirmationMap::decode(&am_str).expect("FATAL: corrupt affirmation map")
+            }
+            None => AffirmationMap::empty(),
+        };
+
+        // remove the first entry -- it's always `n` based on the way we construct it, while the
+        // heaviest affirmation map just has nothing.
+        if affirmation_map.len() > 0 {
```
sergey-shandar, line 3026: nit:
```rust
        Ok(match affirmation_map.as_slice() {
            [] => AffirmationMap::empty(),
            a => AffirmationMap::new(a[1..].to_vec()),
        })
```

src/chainstate/coordinator/mod.rs
```diff
@@ -682,6 +682,128 @@ fn forget_orphan_stacks_blocks(
...
+    burnchain_blocks_db: &BurnchainDB,
+    sortition_db: &SortitionDB,
+    chain_state_db: &StacksChainState,
+    sortition_tip: &SortitionId,
+) -> Result<AffirmationMap, Error> {
+    let last_2_05_rc = sortition_db.get_last_epoch_2_05_reward_cycle()? as usize;
+
+    let sort_am = sortition_db.find_sortition_tip_affirmation_map(sortition_tip)?;
+
+    let canonical_am = StacksChainState::find_canonical_affirmation_map(
```
sergey-shandar, line 732: question: can we move the statement to line # 746? It looks like we may never use `canonical_am` if `last_2_05_rc` is bigger than `sort_am.affirmation.len()`.

src/chainstate/coordinator/mod.rs
```diff
@@ -682,6 +682,128 @@ fn forget_orphan_stacks_blocks(
...
+
+fn inner_static_get_stacks_tip_affirmation_map(
+    burnchain_blocks_db: &BurnchainDB,
+    last_2_05_rc: u64,
+    sort_am: &AffirmationMap,
+    sortdb_conn: &DBConn,
+    canonical_ch: &ConsensusHash,
+    canonical_bhh: &BlockHeaderHash,
+) -> Result<AffirmationMap, Error> {
+    let last_2_05_rc = last_2_05_rc as usize;
```
sergey-shandar, line 761: the algorithm looks the same (or very similar) to `static_get_canonical_affirmation_map`. Should we make a common function?

src/chainstate/coordinator/mod.rs
```diff
@@ -682,6 +682,128 @@ fn forget_orphan_stacks_blocks(
...
+
+    let stacks_am = StacksChainState::find_stacks_tip_affirmation_map(
+        burnchain_blocks_db,
+        sortdb_conn,
+        canonical_ch,
+        canonical_bhh,
+    )?;
+
+    let mut am_entries = vec![];
```
sergey-shandar, line 770:
```suggestion
    let mut am_entries = Vec::default();
```

src/chainstate/stacks/miner.rs
```diff
@@ -140,6 +149,28 @@ pub fn signal_mining_ready(miner_status: Arc<Mutex<MinerStatus>>) {
...
+        Ok(status) => status.get_spend_amount(),
+        Err(_e) => {
+            panic!("FATAL: mutex poisoned");
+        }
+    }
+}
+
+/// set the mining amount
+pub fn set_mining_spend_amount(miner_status: Arc<Mutex<MinerStatus>>, amt: u64) {
+    match miner_status.lock() {
```
sergey-shandar, line 164: nit:

```rust
    miner_status
        .lock()
        .expect("FATAL: mutex poisoned")
        .set_spend_amount(amt);
```

sergey-shandar: See this PR: https://github.com/stacks-network/stacks-blockchain/pull/3516

src/chainstate/burn/db/sortdb.rs
jcnelson: Will accept via #3516

src/chainstate/coordinator/mod.rs
jcnelson: We use `vec![]` everywhere in the codebase, so I'd prefer to be consistent.

src/chainstate/stacks/miner.rs
jcnelson: Yup, will accept via #3516

src/chainstate/coordinator/mod.rs
jcnelson: It uses a different affirmation map (the canonical one, not the Stacks one).

I consolidated the code across all of these related methods to merge the affirmation maps with the sortition affirmation map for reward cycle affirmations in epoch 2.1.

src/chainstate/coordinator/mod.rs
jcnelson: This isn't as tenable now that the affirmation map code here has been consolidated.  But, this method is called at most three times per sortition processed (it's usually only called twice, and _that_ could be reduced if we wanted), which is pretty rare in the system's operation.

sergey-shandar (approved):
I'm still learning the codebase and code conventions so my review is covers only the parts that I already know.

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -59,6 +59,12 @@ pub type RunLoopCounter = Arc<AtomicU64>;
 #[cfg(not(test))]
 pub type RunLoopCounter = ();
 
+#[cfg(test)]
+const UNCONDITIONAL_CHAIN_LIVENESS_CHECK: u64 = 30;
+
+#[cfg(not(test))]
+const UNCONDITIONAL_CHAIN_LIVENESS_CHECK: u64 = 300;
```
kantai, line 66: What is this const controlling? Can you add a rustdoc?

kantai (approved):
This looks good to me, I just have one superficial comment

testnet/stacks-node/src/run_loop/neon.rs
jcnelson: There's a thread that periodically wakes up, loads all the affirmation maps, and if it finds a divergence between any of them, will either wake up the coordinator to announce a burn block, or a burn block plus a stacks block.

PR #3499, opened 2023-01-18
sBTC Alpha Testnet Contract

### LINKED ISSUE #3488: [sBTC] Add minimal contract to allow key administration
For sBTC we need a way to administer keys for the signers

Add a minimal clarity contract that does this

There isn't any other way of doing this that will allow the interaction we need

**Additional context**
My first clarity contract!

xoloki:
### Description
Add a simple contract to allow administering the coordinator and signer keys, and also allowing mint/burn/transfer of sBTC tokens.

### Applicable issues
- fixes #3488 

This contract is much more limited than the final one will be.  It does not allow separate admins from the contract owner, and does not allow signers to rotate their keys.  It also has neither a way to verify schnorr sigs, nor contract code to allow automatic peg handing once a threshold is reached.

### Checklist
- [ ] Changelog is updated
- [x] New clarity functions have corresponding PR in `clarity-benchmarking` repo


contrib/sbtc/key-admin/contracts/key-admin.clar
```diff
@@ -0,0 +1,68 @@
...
+
+(define-public (set-signer-key (id uint) (key principal))
+    (if (is-valid-caller)
+        (ok (map-set signers id key))
+        err-invalid-caller
+    )
+)
+
+(define-public (mint! (amount uint))
+    (if (is-valid-caller)
```
igorsyl, line 45: Instead of gating this method with `is-valid-caller`, could `mint!` accept a frost signature which the method verifies using `secp256k1-verify`? Same with burn! (which needs to be implemented as well).

contrib/sbtc/key-admin/contracts/key-admin.clar
```diff
@@ -0,0 +1,68 @@
...
+    )
+)
+
+;; read only functions
+;;
+(define-read-only (get-coordinator-key)
+  (var-get coordinator)
+)
+
+(define-read-only (get-signer-key (signer uint))
```
igorsyl, line 57: Can we add a method to return the entire map?

xoloki: @igorsyl How do you return an entire map?  var-get doesn't work, and map-get expects a key.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,115 @@
...
+;; summary:
+;; description:
+
+;; traits
+;;
+(impl-trait 'SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.sip-010-trait-ft-standard.sip-010-trait)
+
+;; token definitions
+;; 
+(define-fungible-token sbtc)
```
jcnelson, line 13: Suggestion: you can limit the circulating supply of fungible tokens when it makes sense to do so by passing the limit as a second argument to `define-fungible-token`.  Maybe that's appropriate to do here, given that we (1) want to bound the amount of BTC this contract can represent and (2) enforce the 21 million BTC rule (2.1e15 satoshis)?

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,115 @@
...
+
+;; constants
+;;
+(define-constant contract-owner tx-sender)
+(define-constant err-invalid-caller (err u1))
```
jcnelson, line 18: Minor style nit: most contracts' error constants are just `int` or `uint`.  The code body wraps this into an `(err ...)` list to make it clear in the code body when an error condition occurs.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,115 @@
...
+(define-constant err-owner-only (err u100))
+(define-constant err-not-token-owner (err u101))
+
+;; data vars
+;;
+(define-data-var coordinator (optional principal) none)
```
jcnelson, line 24: Would it make sense to just initialize `coordinator` to be the intended principal from the get-go, instead of requiring the coordinator to call `set-coordinator-key`?

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,115 @@
...
+(define-read-only (get-balance (who principal))
+	(ok (ft-get-balance sbtc who))
+)
+
+(define-read-only (get-total-supply)
+	(ok (ft-get-supply sbtc))
+)
+
+(define-read-only (get-token-uri)
+	(ok none)
```
jcnelson, line 104: Heh, if you wanted to, you could have this method simply return the URL to this github repository :)

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,115 @@
...
+;;
+(define-read-only (get-coordinator-key)
+    (var-get coordinator)
+)
+
+(define-read-only (get-signer-key (signer uint))
+    (map-get? signers signer)
+)
+
+;;(define-read-only (get-signers)
```
jcnelson, line 79: I recall you asked about why there was no native support for iteration over maps?  The reason is that Clarity offers the guarantee that the maximum runtime cost of each code path can be computed statically -- i.e. just by inspecting the contract code.  A data map can grow without bound, so the cost to iterate over all of its key/value pairs can also grow unbound.  So, native iteration over data maps is not (currently) supported.

However, there's a straightforward way to do pagination: simply `map` calls to `map-get?` over a list of `signer` values.  For example:

```clarity
(define-constant PAGE_SIZE u5)

(define-private (get-signer-key (signer uint))
   (map-get? signers signer)
)

(define-read-only (get-signer-key-page (page-id uint))
  (let (
     (page-offset (* PAGE_SIZE page-id))
  )
     (map get-signer-key
        (list page-offset 
              (+ u1 page-offset)
              (+ u2 page-offset)
              (+ u3 page-offset)
              (+ u4 page-offset)))
  )
)
```

This way, the VM can determine what the total runtime cost of a call to `get-signer-key-page` will be regardless of how big the map is.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: Done.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: We were assuming that the principal deploying the contract would be an admin, but not necessarily the coordinator.  And since we are going to need separate FROST public keys in addition to principals, I'm not sure there's a good way to set that on contract init.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: Done :)

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: As long as the FROST ```key_id```s are monotonically increasing ints from ```1..N```, then yeah it's not hard to iterate over the signers.  @igorsyl wanted to try to remove this requirement, and we totally can on the FROST side, but here it seems problematic.

contrib/sbtc/key-admin/contracts/key-admin.clar
```diff
@@ -0,0 +1,68 @@
...
+;; public functions
+;;
+(define-public (set-coordinator-key (key principal))
+    (if (is-valid-caller)
+        (ok (var-set coordinator (some key)))
+        err-invalid-caller
+    )
+)
+
+(define-public (set-signer-key (id uint) (key principal))
```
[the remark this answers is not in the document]
xoloki: Done.

contrib/sbtc/key-admin/contracts/key-admin.clar
xoloki: We discussed this in the meeting, I will try to do it in clarity but seems unlikely for short term if ever.

contrib/sbtc/key-admin/contracts/key-admin.clar
xoloki: As discussed elsewhere, this is trivial as long as ```key_id```s are ```1..N```, otherwise it will be extremely difficult.  Since we are already using such ```key_id```s it seems logical to keep doing so.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: Done.

xoloki:
@igorsyl @jcnelson are we happy with the current implementation for alpha?  If not, what other changes would you like to see?

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,126 @@
...
+(define-data-var coordinator (optional {addr: principal, key: (buff 33)}) none)
+
+;; data maps
+;;
+(define-map signers uint {addr: principal, key: (buff 33)})
+
+;; public functions
+;;
+(define-public (set-coordinator-data (data {addr: principal, key: (buff 33)}))
+    (if (is-valid-caller)
```
igorsyl, line 33: You can use `asserts!` instead of and if-then-else clause.
https://book.clarity-lang.org/ch06-01-asserts.html

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,126 @@
...
+(define-public (delete-signer-data (id uint))
+    (if (is-valid-caller)
+        (ok (map-delete signers id))
+        (err err-invalid-caller)
+    )
+)
+
+(define-public (mint! (amount uint))
+    (if (is-valid-caller)
+        (token-credit! tx-sender amount)
```
igorsyl, line 55: Why don't we call `ft-mint` directly?

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,126 @@
...
+        (err err-invalid-caller)
+    )
+)
+
+(define-public (burn! (amount uint))
+    (if (is-valid-caller)
+        (token-debit! tx-sender amount)
```
igorsyl, line 62: Why don't we call `ft-burn` directly?

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,126 @@
...
+(define-read-only (get-name)
+	(ok "sBTC")
+)
+
+(define-read-only (get-symbol)
+	(ok "sBTC")
+)
+
+(define-read-only (get-decimals)
+	(ok u0)
```
igorsyl, line 99: Should we use decimals to match Bitcoin satoshis? Then 1 BTC = 1 sBTC

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: It's copypasta from the clarity docs, I thought it was kinda weird but wanted to be properly idiomatic.  I'll change it if it's ok with @jcnelson

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: ^

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,126 @@
...
+(define-read-only (get-balance (who principal))
+	(ok (ft-get-balance sbtc who))
+)
+
+(define-read-only (get-total-supply)
+	(ok (ft-get-supply sbtc))
+)
+
+(define-read-only (get-token-uri)
+	(ok (some u"https://github.com/stacks-network/stacks-blockchain"))
```
igorsyl, line 111: Maybe this link is more appropriate?
https://assets.stacks.co/sbtc.pdf

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: From my earlier conversation with @jcnelson he seemed to want 1 sBTC = 1 sat.  I'm okay with whatever

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: Works for me, @jcnelson ?

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
igorsyl: It seems FT tokens support decimals so I'd push for parity to avoid user confusion.

igorsyl:
We need to add a data member to store frost params N and T and ability for owner to reset these params. The signer set function should check the id is in [0,N].

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: We need to wrap with (begin...) to use asserts! properly:

`
    (begin
        (asserts! (is-valid-caller) (err err-invalid-caller))
        (ok (var-set coordinator (some data)))
    )
`
So not saving any space, but happy to do it this way if it's cleaner

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,126 @@
...
+)
+
+;; private functions
+;;
+(define-private (is-valid-caller)
```
igorsyl, line 116: Could we rename to this function `assert-tx-sender-is-owner!` and add `assert-tx-sender-is-coordinator!`?

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,126 @@
...
+
+(define-public (delete-signer-data (id uint))
+    (if (is-valid-caller)
+        (ok (map-delete signers id))
+        (err err-invalid-caller)
+    )
+)
+
+(define-public (mint! (amount uint))
+    (if (is-valid-caller)
```
igorsyl, line 54: `mint` and `burn` should be gated to coordinator using `assert-tx-sender-is-coordinator!`

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
igorsyl: I think it reads cleaner.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: Done.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: Done.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: Done.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
jcnelson: :+1:

xoloki: Done.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
jcnelson: Ah, I wasn't aware.

Can you help me understand the context(s) in which we need to be able to iterate over the list of signers?  If (1) it's for the convenience of downstream clients to learn the list of all signers, and (2) the number of signers is (a) bound and (b) fixed per reward cycle, it might make sense to just store a `(list 4000 ...)` of signers as a data var, so the full list can be queried in one network RPC (along with a Merkle proof).  Growing this list will be comparatively expensive, since the full list must be loaded and stored each time a new entry is added, but it would suffice for now.

xoloki: The context was indeed to prevent multiple read-only calls to the node in order to get the signing keys for every key-id. 

I want to avoid setting the signer keys in bulk because I think people will need to rotate them at irregular intervals.

For now I'll just add a get-signers that iterates over ```[0, num-keys)``` and returns a list.  We won't be calling this frequently enough for that to be a problem.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: Done.

xoloki:
> We need to add a data member to store frost params N and T and ability for owner to reset these params. The signer set function should check the id is in [0,N].
> 
> Also, it may be useful to have functions to reset the contract-owner.
> 
> Also, we need to a member variable to store the threshold wallet bitcoin address gated by the coordinator.

Added ```num-keys``` ```num-parties``` and ```threshold``` with accessors and tests.  I'll add the rest next.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: Done.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: I named them ```is-contract-owner``` and ```is-coordinator``` since we might want to call them when we aren't doing an ```assert```.

xoloki:
BTW ```clarinet check``` is spamming ```warning: use of potentially unchecked data``` on all my setters.  What's the idiomatic way of checking data?  @igorsyl @jcnelson

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
igorsyl: Makes sense. Could we have `assert-is-contract-owner` call `is-contract-owner`?

xoloki: Will do.

xoloki: I'm getting a weird ```clarinet check``` error when I try to add these functions.

`(define-private (assert-is-contract-owner)
    (asserts! (is-eq contract-owner tx-sender) (err err-invalid-caller))
)`

```detected two execution paths, returning two different expression types (got '(response UnknownType uint)' and 'bool')```

Comment (xoloki) [reply contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar]: The existing code which asserts `is-contract-owner` seems indistinguishable and doesn't error in the check

**Comment by xoloki** [comment]
> > We need to add a data member to store frost params N and T and ability for owner to reset these params. The signer set function should check the id is in [0,N].
> > Also, it may be useful to have functions to reset the contract-owner.
> > Also, we need to a member variable to store the threshold wallet bitcoin address gated by the coordinator.
> 
> Added `num-keys` `num-parties` and `threshold` with accessors and tests. I'll add the rest next.

contract-owner is now a var and can be set

File: contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar (hunk 5 above)
Comment (igorsyl) [reply contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar]: Can we initialize coordinator to the contract admin/creator? The admin can later change it and this way the variable doesn't need to be optional.

Comment (xoloki) [reply contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar]: > Can we initialize coordinator to the contract admin/creator? The admin can later change it and this way the variable doesn't need to be optional.

No, because we need a FROST public key in addition to a principal.

File: contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar (hunk 14 above)
Comment (igorsyl) [reply contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar]: Regarding the errors, you may need to add a begin block

**Comment by igorsyl** [comment]
@lgalabru ?

> BTW `clarinet check` is spamming `warning: use of potentially unchecked data` on all my setters. What's the idiomatic way of checking data? @igorsyl @jcnelson

File: contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar (hunk 14 above)
Comment (xoloki) [reply contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar]: No, same error @igorsyl

Comment (xoloki) [reply contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar]: I think I understand; I'm guessing that ```asserts!``` in the true case returns true from the statement, and in the error case return err.  Since the compiler can't know which branch will be followed there is a conflict.

When using ```asserts!``` in the body of a function, though, the statement return from ```asserts!``` is ignored, but the function will return the error.

Comment (xoloki) [reply contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar]: Yeah that was it

Comment (xoloki) [reply contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar]: That doesn't work either:

 ;;
 (define-public (set-contract-owner (owner principal))
     (begin
-        (asserts! (is-contract-owner) (err err-invalid-caller))
+        (asserts-is-contract-owner!)
         (ok (var-set contract-owner owner))
     )
 )
@@ -173,3 +173,17 @@
 (define-private (is-valid-signer-id (id uint))
     (and (>= id u0) (< id (var-get num-keys)))
 )
+
+(define-private (asserts-is-contract-owner!)
+    (begin
+        (asserts! (is-contract-owner) (err err-invalid-caller))
+        (ok true)
+    )
+)

Still gives a clarity error:

```intermediary responses in consecutive statements must be checked```

**Comment by xoloki** [comment]
@igorsyl @jcnelson are we happy with the current code?  If not, what else do we need to do?

File: contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar [hunk 16]
```diff
@@ -0,0 +1,186 @@
...
+(define-constant err-not-token-owner u3)
+
+;; data vars
+;;
+(define-data-var contract-owner principal tx-sender)
+(define-data-var coordinator (optional {addr: principal, key: (buff 33)}) none)
+(define-data-var num-keys uint u4000)
+(define-data-var num-parties uint u4000)
+(define-data-var threshold uint u2800)
+(define-data-var bitcoin-payout-address (optional (string-ascii 72)) none)
```
igorsyl, line 28: Could we call this bitcoin-wallet-address? This address is used for receiving funds as well so payout could be confusing.

jcnelson: Either one is fine with me, as long as there's a comment somewhere that explains it.

N.B. Treating this as a `(string-ascii 72)` is fine for now, since it doesn't appear to be used to interact directly with Bitcoin state.  In the future, this would almost certainly be represented as a PoX address, which has the type `{ version: (buff 1), hashbytes: (buff 32) }`.  But there's no need to do this for this PR.

jcnelson (approved):
LGTM. Thanks @xoloki!

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,186 @@
...
+    (begin
+        (asserts! (is-contract-owner) (err err-invalid-caller))
+        (asserts! (is-valid-signer-id id) (err err-invalid-signer-id))
+        (ok (map-delete signers id))
+    )
+)
+
+(define-public (set-bitcoin-payout-address (addr (string-ascii 72)))
+    (begin
+        (asserts! (is-contract-owner) (err err-invalid-caller))
```
igorsyl, line 89: This method should allow coordinator, not contract owner.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,186 @@
...
+        (ok (var-set bitcoin-payout-address (some addr)))
+    )
+)
+
+(define-public (mint! (amount uint))
```
igorsyl, line 94: I'm thinking we should add a string memo argument to mint! to record the associated Bitcoin peg-in request txid.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,186 @@
...
+)
+
+(define-public (burn! (amount uint))
+    (begin
+        (asserts! (is-coordinator) (err err-invalid-caller))
+        (ft-burn? sbtc amount tx-sender)
+    )
+)
+
+(define-public (transfer (amount uint) (sender principal) (recipient principal) (memo (optional (buff 34))))
```
igorsyl, line 108: We should add a trading halt boolean member variable gated to contract owner if we need to disable trading for any reason.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,186 @@
...
+	(begin
+		(asserts! (is-eq tx-sender sender) (err err-not-token-owner))
```
igorsyl, line 110: At some point, we would like to let users delegate transfer rights to contracts. I'm certain we'll get this feedback when testnet goes live but we can address in the next contract iteration.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: Done.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: Done.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: I added this var with a r/w test, and another test that actually tries to transfer when halted.  The first test passes, but the second test goes into what appears to be an infinite recursion loop when we try to do the first transfer.

Can you look and see if I've done something obviously wrong @igorsyl @jcnelson ?

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: How would this work exactly?  transfer the FTs to a contract principal?  Or something internal like a delegate map?

Def not needed for this PR in any case.

igorsyl: The latter. I believe this could be addressed with a new SIP trait: https://github.com/stacksgov/sips/pull/84

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,200 @@
...
+        (ft-mint? sbtc amount tx-sender)
+    )
+)
+
+(define-public (burn! (amount uint) (memo (string-ascii 72)))
+    (begin
+        (asserts! (is-coordinator) (err err-invalid-caller))
+        (ft-burn? sbtc amount tx-sender)
+    )
+)
```
netrome, line 115: These methods should take an additional principal as an argument, since the coordinator will mint & burn tokens for arbitrary stacks principals.

igorsyl: I think `memo` is meant to encode the bitcoin ops request txid.

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
```diff
@@ -0,0 +1,200 @@
...
+
+(define-public (mint! (amount uint) (memo (string-ascii 72)))
+    (begin
+        (asserts! (is-coordinator) (err err-invalid-caller))
+        (ft-mint? sbtc amount tx-sender)
+    )
+)
+
+(define-public (burn! (amount uint) (memo (string-ascii 72)))
```
igorsyl, line 110: can we instead of calling `memo` call it `bitcoin_txid`

contrib/sbtc/token-key-admin/contracts/sbtc-token-key-admin.clar
xoloki: Done @netrome

xoloki: See below @igorsyl , I used peg-in/out-txid for mint/burn

igorsyl:
> Ok @igorsyl @netrome the contract and tests are all passing, I think I've handled all review comments. Please let me know if you need anything else.

I'll deploy this. We can keep the PR open. Thanks @xoloki

xoloki:
> > Ok @igorsyl @netrome the contract and tests are all passing, I think I've handled all review comments. Please let me know if you need anything else.
> 
> I'll deploy this. We can keep the PR open. Thanks @xoloki

Thanks @igorsyl

netrome (approved):

igorsyl:
The sbtc contract is being developed [here](https://github.com/Trust-Machines/core-eng/tree/main/sbtc-ops). We will move it to this repository once it's stable.

PR #3501, opened 2023-01-18
fix: revert all regex changes in the v1 parser

obycode:
With the addition of the v2 parser, some of the regex expressions were
moved up to representations.rs so that they could be shared and reused.
This has proven to be a problem, because the strange behavior of the v1
parser was not preserved (e.g. allowing long contract names in
contract-call parameters, but not in contracts).

Fixes: #3470

### DIFF
--- a/clarity/src/vm/ast/parser/v1.rs
+++ b/clarity/src/vm/ast/parser/v1.rs
@@ -1320,9 +1320,11 @@ mod test {
     #[test]
     fn test_long_contract_name() {
         let long_contract_name = "(define-private (transfer (id uint) (receiver principal)) (contract-call? 'SP3D6PV2ACBPEKYJTCMH7HEN02KP87QSP8KTEH335.megapont-robot-expansion-nftSPNWZ5V2TPWGQGVDR6T7B6RQ4XMGZ4PXTEE0VQ0S.guests-hosted-stacks-parrots transfer id tx-sender receiver))";
-        assert!(match ast::parser::v1::parse(long_contract_name).unwrap_err().err {
-            ParseErrors::SeparatorExpected(_) => true,
-            _ => false,
-        });
+        assert!(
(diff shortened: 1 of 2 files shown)


clarity/src/vm/ast/parser/v1.rs
```diff
@@ -1300,4 +1316,13 @@ mod test {
...
     }
+
+    #[test]
+    fn test_long_contract_name() {
+        let long_contract_name = "(define-private (transfer (id uint) (receiver principal)) (contract-call? 'SP3D6PV2ACBPEKYJTCMH7HEN02KP87QSP8KTEH335.megapont-robot-expansion-nftSPNWZ5V2TPWGQGVDR6T7B6RQ4XMGZ4PXTEE0VQ0S.guests-hosted-stacks-parrots transfer id tx-sender receiver))";
+        assert!(match ast::parser::v1::parse(long_contract_name).unwrap_err().err {
+            ParseErrors::SeparatorExpected(_) => true,
+            _ => false,
+        });
+    }
```
jcnelson, line 1327: Can you also add the test for the case seen here, if it's not in the codebase already? https://github.com/stacks-network/stacks-blockchain/issues/3133

obycode: Yes, that is here already - https://github.com/stacks-network/stacks-blockchain/blob/2894afd5767a4d32a5e4be5132abcaa04a4b47b4/src/chainstate/stacks/db/transactions.rs#L2409-L2412

obycode:
Oops, I'll resolve these failures.

clarity/src/vm/ast/parser/v1.rs
jcnelson: Thanks!

jcnelson (approved):
LGTM!

jcnelson:
Confirmed that this fixes the processing for block 76180. Will boot a node from genesis with this patch.

Issue #3481, opened 2023-01-12
[sBTC] Peg-in wire format


netrome:
Introduce an alpha version of the PegIn operation as defined in [SIP-21](https://github.com/stacksgov/sips/blob/56b73eada5ef1b72376f4a230949297b3edcc562/sips/sip-021/sip-021-trustless-two-way-peg-for-bitcoin.md#peg-in-transaction).

Preliminary exploration of the change has been done in PR #3475.

PR #3510, opened 2023-01-19
[sBTC] Feat: Peg out request & fulfillment wire formats

netrome:
Introducing wire formats for the peg out request and peg out fulfillment sBTC operations. This PR do not include any signature or sBTC validation for the peg out request. I have created #3514 for this.

I will point this PR to `next` once #3481 is merged.


src/chainstate/burn/db/sortdb.rs
```diff
@@ -3888,6 +3972,36 @@ impl SortitionDB {
...
+        burn_header_hash: &BurnchainHeaderHash,
+    ) -> Result<Vec<PegOutFulfillOp>, db_error> {
+        query_rows(
+            conn,
+            "SELECT * FROM peg_out_fulfillments WHERE burn_header_hash = ?",
+            &[burn_header_hash],
+        )
+    }
+
+    // TODO(3493): Query functions for peg out request and peg out fulfill
```
jcnelson, line 4003: Is this `TODO` satisfied by the above?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -6268,6 +6430,98 @@ pub mod tests {
...
+                peg_out_request.clone(),
+            )],
+        );
+
+        let res_peg_out_requests =
+            SortitionDB::get_peg_out_request_ops(db.conn(), &burn_header_hash)
+                .expect("Failed to get peg-in ops from sortition DB");
+
+        assert_eq!(res_peg_out_requests.len(), 1);
+        assert_eq!(res_peg_out_requests[0].amount, 1337);
```
jcnelson, line 6476: Can you test all the fields?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -6268,6 +6430,98 @@ pub mod tests {
...
+                peg_out_fulfill.clone(),
+            )],
+        );
+
+        let res_peg_out_fulfillments =
+            SortitionDB::get_peg_out_fulfill_ops(db.conn(), &burn_header_hash)
+                .expect("Failed to get peg-in ops from sortition DB");
+
+        assert_eq!(res_peg_out_fulfillments.len(), 1);
+        assert_eq!(res_peg_out_fulfillments[0].amount, 1337);
```
jcnelson, line 6522: Same here. Just to be safe.

src/chainstate/burn/operations/peg_out_fulfill.rs
```diff
@@ -0,0 +1,228 @@
...
+            amount,
+            recipient,
+            txid,
+            vtxindex,
+            block_height,
+            burn_header_hash,
+        })
+    }
+
+    fn parse_data(data: &[u8]) -> Result<BlockHeaderHash, ParseError> {
```
jcnelson, line 48: This needs to be a `StacksBlockId`, not a `BlockHeaderHash`.

A Stacks block's bare hash alone does not refer to a unique chain tip.  Instead, it's the hash of the block, hashed with the `ConsensusHash` that represents the history of PoX operations leading to its block-commit.  This hash-of-hashes is captured within a `StacksBlockId`.

jcnelson: Some elaboration.

In PoX, it is possible for the same Stacks block to be valid in two PoX forks.  This can happen if the same block is mined twice in two different PoX forks.  This can happen if their parent is in the prior reward cycle, but one child lives in a PoX fork where the anchor block was chosen and the other lives in a fork in which it was not chosen.

To address this corner case, we use `StacksBlockId` to represent the globally-unique ID for a block, since it represents the PoX history in which the block was mined.

src/chainstate/burn/db/sortdb.rs
netrome: Yes, thanks for spotting. I'll remove the comment.

src/chainstate/burn/db/sortdb.rs
netrome: Sure thing, will do.

src/chainstate/burn/operations/peg_out_fulfill.rs
netrome: Ah I see. Thank you for the explanation. I will update the code to use `StacksBlockId`.

src/chainstate/burn/db/sortdb.rs
netrome: :heavy_check_mark:

src/chainstate/burn/db/sortdb.rs
netrome: Done :heavy_check_mark:

src/chainstate/burn/db/sortdb.rs
netrome: Done :heavy_check_mark:

src/chainstate/burn/operations/peg_out_fulfill.rs
netrome: Done :heavy_check_mark:

src/chainstate/coordinator/tests.rs
```diff
@@ -3347,9 +3348,17 @@ fn test_sbtc_peg_in_btc_op() {
...
+        let recipient_btc_address = PoxAddress::Standard(stacker.into(), None);
+        let canonical_chain_tip_snapshot =
+            SortitionDB::get_canonical_burn_chain_tip(sort_db.conn()).unwrap();
+
+        let chain_tip = StacksBlockId::new(
+            &canonical_chain_tip_snapshot.consensus_hash,
+            &canonical_chain_tip_snapshot.winning_stacks_block_hash,
+        );
 
         if ix == 0 {
```
sergey-shandar, line 3360: nit: `match ix {`

netrome: Of course, thanks for spotting! That's what I get for cargo-cult-copying the structure of similar test cases in this file :sweat_smile:

netrome: Updated :heavy_check_mark:

sergey-shandar (approved):
The only thing which I don't really like is forming SQL queries as strings. Usually, it's a sign of an anti-pattern. However, it's a problem of the base code rather than this PR.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -738,7 +793,34 @@ const SORTITION_DB_SCHEMA_5: &'static [&'static str] = &[r#"
...
+        txid TEXT NOT NULL,
+        vtxindex INTEGER NOT NULL,
+        block_height INTEGER NOT NULL,
+        burn_header_hash TEXT NOT NULL,
+
+        amount TEXT NOT NULL,
+        recipient TEXT NOT NULL,
+        signature TEXT NOT NULL,
+
+        PRIMARY KEY(txid)
```
jcnelson, line 808: As with peg-ins, you're going to need to use a primary key that includes both the `txid` and the sortition ID in which it was processed (per #3527).  This applies to both of these two new tables.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -726,7 +780,8 @@ const SORTITION_DB_SCHEMA_4: &'static [&'static str] = &[
 // update this to add new indexes
 const LAST_SORTITION_DB_INDEX: &'static str = "index_burn_header_hash_pox_valid";
```
jcnelson, line 781: You'll need to change this to `index_peg_out_fulfill_burn_header_hash` so the code knows to add the new indexes.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -6265,7 +6425,99 @@ pub mod tests {
...
+            )],
+        );
+
+        let res_peg_out_requests =
+            SortitionDB::get_peg_out_request_ops(db.conn(), &burn_header_hash)
+                .expect("Failed to get peg-in ops from sortition DB");
+
+        assert_eq!(res_peg_out_requests.len(), 1);
+        assert_eq!(res_peg_out_requests[0], peg_out_request);
+    }
```
jcnelson, line 6475: Please expand both of these tests to cover the case where the same transaction gets included in two different burnchain blocks.  The operation metadata, such as the `block_height` and `vtxindex`, should be different in both cases but the operation itself should be the same.  The test should confirm that different metadata are stored for different burnchain blocks.

src/chainstate/burn/operations/mod.rs
```diff
@@ -329,6 +335,33 @@ pub struct PegInOp {
     pub burn_header_hash: BurnchainHeaderHash, // hash of the burn chain block header
 }
 
+#[derive(Debug, PartialEq, Clone, Eq, Serialize, Deserialize)]
+pub struct PegOutRequestOp {
+    pub amount: u64,                 // sBTC amount to peg out, in satoshis
+    pub recipient: PoxAddress,       // Address to receive the BTC when the request is fulfilled
+    pub signature: MessageSignature, // Signature from sBTC owner's private key(s)
```
jcnelson, line 342: To be clear, this is going to be a signature from one of the following:

* A standard principal on Stacks
* A key authorized by the owner of the sBTC (be it a contract or standard principal).

SIP-021 does not speak to the latter, but it should.  It should be possible for the sBTC owner to publish a key that will be used to sign these requests.

src/chainstate/burn/operations/peg_out_fulfill.rs
```diff
@@ -0,0 +1,228 @@
...
+        if data.len() < 32 {
+            warn!(
+                "PegInOp payload is malformed ({} bytes, expected at least {})",
+                data.len(),
+                32
+            );
+            return Err(ParseError::MalformedData);
+        }
+
+        StacksBlockId::from_bytes(data).ok_or(ParseError::MalformedData)
```
jcnelson, line 70: You should take the 32 byte slice from `data` here.  This operation should be valid even if the OP_RETURN contains data after the 35th byte.  We'd reserve bytes 35-80 for future use.

src/chainstate/burn/operations/peg_out_fulfill.rs
```diff
@@ -0,0 +1,228 @@
...
+        let tx = test_helpers::burnchain_transaction(data, output2, opcode);
+        let header = test_helpers::burnchain_block_header();
+
+        let op = PegOutFulfillOp::from_tx(&header, &tx);
+
+        match op {
+            Err(OpError::ParseError) => (),
+            result => panic!("Expected OpError::ParseError, got {:?}", result),
+        }
+    }
```
jcnelson, line 191: Please add test coverage for the case where the OP_RETURN contains more than 35 bytes.

src/chainstate/burn/operations/peg_out_request.rs
```diff
@@ -0,0 +1,256 @@
...
+            warn!(
+                "PegOutRequestOp payload is malformed ({} bytes, expected {})",
+                data.len(),
+                73
+            );
+            return Err(ParseError::MalformedPayload);
+        }
+
+        let amount = u64::from_be_bytes(data[0..8].try_into()?);
+        let signature = MessageSignature(data[8..73].try_into()?);
```
jcnelson, line 72: How could this `.try_into()` fail, and why is returning `Err(..)` acceptable if we know that `data` contains at least 73 bytes?

src/chainstate/burn/operations/peg_out_request.rs
```diff
@@ -0,0 +1,256 @@
...
+
+        Ok(ParsedData { amount, signature })
+    }
+
+    pub fn check(&self) -> Result<(), OpError> {
+        if self.amount == 0 {
+            warn!("PEG_OUT_REQUEST Invalid: Requested BTC amount must be positive");
+            return Err(OpError::AmountMustBePositive);
+        }
+
```
jcnelson, line 82: You will also need to check that the `signature` can recover to a public key.  Not all 65-byte strings can.

src/chainstate/burn/operations/peg_out_request.rs
```diff
@@ -0,0 +1,256 @@
...
+        let tx = test_helpers::burnchain_transaction(data, output2, opcode);
+        let header = test_helpers::burnchain_block_header();
+
+        let op = PegOutRequestOp::from_tx(&header, &tx);
+
+        match op {
+            Err(OpError::ParseError) => (),
+            result => panic!("Expected OpError::ParseError, got {:?}", result),
+        }
+    }
```
jcnelson, line 215: Please also add test coverage for when the OP_RETURN contains bytes beyond the signature.  This should still successfully decode; the extra bytes would be ignored.

src/chainstate/burn/operations/mod.rs
```diff
@@ -329,6 +335,33 @@ pub struct PegInOp {
...
+
+    // common to all transactions
```
jcnelson, line 344: Please also store the extra bytes that aren't used for decoding, as a `memo: Vec<u8>`.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -738,7 +793,34 @@ const SORTITION_DB_SCHEMA_5: &'static [&'static str] = &[r#"
...
 
         PRIMARY KEY(txid)
-    );"#];
+    );"#,
+    r#"
+    CREATE TABLE peg_out_requests (
+        txid TEXT NOT NULL,
+        vtxindex INTEGER NOT NULL,
+        block_height INTEGER NOT NULL,
+        burn_header_hash TEXT NOT NULL,
```
jcnelson, line 802: Please include the extra, unparsed bytes as a `memo`  column.

src/chainstate/burn/operations/test_helpers.rs
```diff
@@ -0,0 +1,89 @@
+use rand::rngs::StdRng;
```
jcnelson, line 1: Please put this into the file `src/chainstate/burn/operations/test/mod.rs`.

jcnelson: Alternatively, you can put this as a `mod test { ... }` in `src/chainstate/burn/operations/mod.rs`.

src/chainstate/coordinator/tests.rs
```diff
@@ -3401,14 +3462,32 @@ fn test_sbtc_peg_in_btc_op() {
...
+        .iter()
+        .flat_map(|block_hash| {
+            SortitionDB::get_peg_out_fulfill_ops(&sort_db.conn(), block_hash)
+                .expect("Failed to get peg out fulfillment ops")
+        })
+        .count();
+
     assert_eq!(total_number_of_peg_in_ops, 1);
+    assert_eq!(total_number_of_peg_out_request_ops, 1);
+    assert_eq!(total_number_of_peg_out_fulfill_ops, 2);
```
jcnelson, line 3490: Please also verify the contents of these operations.

src/chainstate/burn/db/sortdb.rs
netrome: Done

src/chainstate/burn/db/sortdb.rs
netrome: Fixed this while rebasing on the updated peg-in branch.

src/chainstate/burn/db/sortdb.rs
netrome: Done ✅

src/chainstate/burn/db/sortdb.rs
netrome: Done :heavy_check_mark:

src/chainstate/burn/operations/test_helpers.rs
netrome: Done :heavy_check_mark:

src/chainstate/burn/operations/peg_out_request.rs
netrome: It can't, so an unwrap would be safe here. I'll change.

src/chainstate/burn/operations/mod.rs
netrome: Alright, I've slightly changed the description and refer to SIP-021 for details. New description:

```
// Signature from sBTC owner as per SIP-021
```

src/chainstate/coordinator/tests.rs
netrome: Done.

src/chainstate/burn/operations/mod.rs
netrome: Done :heavy_check_mark:

src/chainstate/burn/operations/peg_out_request.rs
netrome: Done :heavy_check_mark:

src/chainstate/burn/operations/peg_out_fulfill.rs
netrome: Done :heavy_check_mark:

src/chainstate/burn/operations/peg_out_fulfill.rs
netrome: Done :heavy_check_mark:

src/chainstate/burn/operations/peg_out_request.rs
netrome: Done :heavy_check_mark:

netrome:
Rebased and updated to use `(txid, burn_header_hash)` as primary key for the sortdb tables, analogous to the changes commented [here](https://github.com/stacks-network/stacks-blockchain/pull/3486#discussion_r1090808273).

src/chainstate/burn/operations/mod.rs
```diff
@@ -330,6 +336,35 @@ pub struct PegInOp {
     pub burn_header_hash: BurnchainHeaderHash, // hash of the burn chain block header
 }
 
+#[derive(Debug, PartialEq, Clone, Eq, Serialize, Deserialize)]
+pub struct PegOutRequestOp {
+    pub amount: u64,                 // sBTC amount to peg out, in satoshis
+    pub recipient: PoxAddress,       // Address to receive the BTC when the request is fulfilled
+    pub signature: MessageSignature, // Signature from sBTC owner as per SIP-021
+    pub memo: Vec<u8>,               // extra unused byte
```
jcnelson, line 344: As discussed earlier this week, this op is going to need to pay a nominal amount of BTC to the peg wallet.  Eventually, this is going to be used by the requester to pay the stackers the BTC transaction fee to process the peg-out fulfillment.  For now, this is going to be used to link the peg-out request to the peg-out fulfillment.

Please update this op to include both the peg wallet address (a `PoxAddress`) and the amount paid.  The amount paid must be positive.

src/chainstate/burn/operations/mod.rs
```diff
@@ -330,6 +336,35 @@ pub struct PegInOp {
...
+    pub burn_header_hash: BurnchainHeaderHash, // hash of the burn chain block header
+}
+
+#[derive(Debug, PartialEq, Clone, Eq, Serialize, Deserialize)]
+pub struct PegOutFulfillOp {
+    pub chain_tip: StacksBlockId, // The Stacks chain tip whose state view was used to validate the peg-out request
+
+    pub amount: u64,           // Transferred BTC amount, in satoshis
+    pub recipient: PoxAddress, // Address to receive the BTC
+    pub memo: Vec<u8>,         // extra unused byte
```
jcnelson, line 359: Because we're going to have the peg-out fulfillment op consume a UTXO produced by the peg-out request, please update this struct to include a pointer to the peg-out request.  This would have type `(Txid, u32)` to encode the `vout` entry in the Bitcoin transaction.  We use a similar tactic for `PreStx`, `TransferStx`, `StackStx`, and `DelegateStx` -- the latter three operations consume a UTXO produced by `PreStx`.

src/chainstate/burn/operations/peg_out_fulfill.rs
```diff
@@ -0,0 +1,265 @@
+use stacks_common::codec::StacksMessageCodec;
```
jcnelson, line 1: Please add the copyright statement.

src/chainstate/burn/operations/peg_out_fulfill.rs
```diff
@@ -0,0 +1,265 @@
...
+    ) -> Result<Self, OpError> {
+        if tx.opcode() != Opcodes::PegOutFulfill as u8 {
+            warn!("Invalid tx: invalid opcode {}", tx.opcode());
+            return Err(OpError::InvalidInput);
+        }
+
+        let (amount, recipient) = if let Some(Some(recipient)) = tx.get_recipients().first() {
+            (recipient.amount, recipient.address.clone())
+        } else {
+            warn!("Invalid tx: Output 2 not provided");
```
jcnelson, line 26: It could also be that output 2 is present not standard, meaning we didn't decode it.

src/chainstate/burn/operations/peg_out_fulfill.rs
```diff
@@ -0,0 +1,265 @@
...
+            return Err(OpError::InvalidInput);
+        };
+
```
jcnelson, line 29: You're also going to need to look for output 3, which is a payment to the peg wallet.

src/chainstate/burn/operations/peg_out_fulfill.rs
```diff
@@ -0,0 +1,265 @@
...
+        if data.len() < 32 {
+            warn!(
+                "PegInOp payload is malformed ({} bytes, expected at least {})",
+                data.len(),
+                32
+            );
+            return Err(ParseError::MalformedData);
+        }
+
+        let chain_tip = StacksBlockId::from_bytes(&data[..32]).unwrap();
```
jcnelson, line 71: Please use `.expect()` and include an appropriate error message.

src/chainstate/burn/operations/mod.rs
netrome: How would this work for multiple fulfillment transactions given a single peg out request? Only the first fulfillment would be able to consume the UTXO.

src/chainstate/burn/operations/peg_out_fulfill.rs
```diff
@@ -0,0 +1,265 @@
...
+
+        let tx = test::burnchain_transaction(data, output2, opcode);
+        let header = test::burnchain_block_header();
+
+        let op =
+            PegOutFulfillOp::from_tx(&header, &tx).expect("Failed to construct peg-out operation");
+
+        assert_eq!(op.recipient.bytes(), recipient_address_bytes);
+        assert_eq!(op.chain_tip.as_bytes(), &chain_tip_bytes);
+        assert_eq!(op.amount, amount);
```
jcnelson, line 129: Please also test the memo.

jcnelson: Oh wait I see you did that below. Nevermind :sweat_smile:

jcnelson (changes requested):
This PR cannot merge until the peg-out fulfillment transaction consumes a UTXO from the peg-out request, and the transaction validation logic for the former checks that this happens.

src/chainstate/burn/operations/mod.rs
netrome: I've added the `peg_wallet_address` and `fulfillment_fee` fields to the op for this :heavy_check_mark:

src/chainstate/burn/operations/peg_out_fulfill.rs
netrome: I've added output 3 to the request op. There's no output 3 in the fulfill op afaik. However, I'll extend the data of this op to contain a pointer to the peg out request op.

src/chainstate/burn/operations/peg_out_fulfill.rs
netrome: Done :heavy_check_mark:

src/chainstate/burn/operations/peg_out_fulfill.rs
netrome: Updated the message :heavy_check_mark:

src/chainstate/burn/operations/peg_out_fulfill.rs
netrome: Done :heavy_check_mark:

src/chainstate/burn/operations/mod.rs
netrome: From yesterdays sBTC discussion we concluded that we're initially only going to support a 1:1 relation between fulfillment transactions and peg out requests.

netrome: Since the UTXO must be the third output of the peg-out request, I'm asserting that `vout` must be 2 and only storing the `Txid` in the `PegOutFulfillOp`.

igorsyl:
> This PR cannot merge until the peg-out fulfillment transaction consumes a UTXO from the peg-out request, and the transaction validation logic for the former checks that this happens.

@netrome gentle ping to request re-review if ready.

netrome:
> > This PR cannot merge until the peg-out fulfillment transaction consumes a UTXO from the peg-out request, and the transaction validation logic for the former checks that this happens.
> 
> @netrome gentle ping to request re-review if ready.

Yeah it's not ready yet. I have some issues with a failing UTXO signature check when updating the neon integration test to make the fulfillment op consume an UTXO from the peg out request. It seems like all UTXOs I've created for sBTC ops are failing when I attempt to spend them. I'm using `PoxAddress::to_bitcoin_tx_out` to construct the transaction output which is not widely used in the repo. I'll continue the investigation and reach out for help if I get stuck on something for too long.

And also, this PR needs to be rebased when the [peg-in PR](https://github.com/stacks-network/stacks-blockchain/pull/3486) is merged - in which I have requested re-reviews.

netrome:
I have pushed the current version of the broken testcase to this branch now, for anyone who want to reproduce it.

The test case is `test_submit_and_observe_sbtc_ops` on line 10429 in `neon_integrations.rs`. You can run it through this command

```
cargo nextest run -p stacks-node sbtc --nocapture -- --ignored
```

The error returned from bitcoind is

```
Bitcoin RPC failure: transaction submission failed - Network("Bitcoin RPC: status(500) != success, body is 'Object({\"error\": Object({\"code\": Number(-26), \"message\": String(\"mandatory-script-verify-flag-failed (Signature must be zero for failed CHECK(MULTI)SIG operation)\")}), \"id\": String(\"stacks\"), \"result\": Null})'")
```

which occurs when creating the PegOutFulfillment transaction which consumes an UTXO created by the peg out request transaction.

For reference, the `stx_transfer_btc_integration_test` test successfully consumes a UTXO created by a pre-stx op. I have yet to figure out why this scenario succeeds while the UTXO signature check fails in the sbtc test.

I've been able to verify the signature error using btcdeb with the following command (this is from a variant where I use the same key for the miner signer and peg wallet address, which causes the test to fail when constructing the PegOutRequest instead of PegOutFulfillment. It's in this scenario the first operation which consumes a UTXO from the peg-in operation)
```
btcdeb --tx=01000000027ecf083979fbe196be0f718f9f812be972f913bf54350b3baeb76136af7567f7010000008a4730440220187862e28241c1bc28cd3e13feeb0a8bdada1eae61e70f815c3e754aaab3196002202ea85ad5807dab08750357626a71a6809289261e74e0a88b0fd5d3baa214ae18014104ee0b1602eb18fef7986887a7e8769a30c9df981d33c8380d255edef003abdcd243a0eb74afdf6740e6c423e62aec631519a24cf5b1d62bf8a3e06ddc695dcb77fdffffff7ecf083979fbe196be0f718f9f812be972f913bf54350b3baeb76136af7567f7020000008b483045022100a6753dd1ed29d1021ef09fb75742e17cbf0e0eae531a87db557ccbe7bbe19f820220295feaebada6a7b028f56bebeee842a5880171f52326c90d101559c936074e58014104ee0b1602eb18fef7986887a7e8769a30c9df981d33c8380d255edef003abdcd243a0eb74afdf6740e6c423e62aec631519a24cf5b1d62bf8a3e06ddc695dcb77fdffffff0400000000000000004f6a4c4c54323e0000000000020a44000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010270000000000001976a914000000000000000000000000000000000000000088ace8030000000000001976a9148bc5147525b8f477f0bc4522a88c8339b2494db588ac1afb002a010000001976a9148bc5147525b8f477f0bc4522a88c8339b2494db588ac00000000 '[OP_DUP OP_HASH160 8bc5147525b8f477f0bc4522a88c8339b2494db5 OP_EQUALVERIFY OP_CHECKSIG]' 3045022100a6753dd1ed29d1021ef09fb75742e17cbf0e0eae531a87db557ccbe7bbe19f820220295feaebada6a7b028f56bebeee842a5880171f52326c90d101559c936074e5801 04ee0b1602eb18fef7986887a7e8769a30c9df981d33c8380d255edef003abdcd243a0eb74afdf6740e6c423e62aec631519a24cf5b1d62bf8a3e06ddc695dcb77 --verbose
```

For comparison, here's a successful signature verification in btcdeb (adopted from the transfer stx op in `stx_transfer_btc_integration_test`):

```
btcdeb --tx=0100000001ec8c4dfa60a95102997d62d96494380f6b6d6edd2a430f85b514ab300dd060ff010000006b483045022100c7b0c2c3972387eda538ad458c259687735f45b55fba939c205b1061f3adb75902203c11d08a6fbebfda588117efe2b948d1ef5962bce4cbb39d2e87c09135f3e3b7012102b37b7f1553fa17fe0ad71c57538f4a93bfada632947c04e1ca6fcdad65e8106dfdffffff020000000000000000156a13543224000000000000000000000000000186a07c150000000000001976a9142362616d49d5ab67bb1a5355f5a17d96effae4b688ac00000000 '[OP_DUP OP_HASH160 f942874ce525e87f21bbe8c121b12fac831d02f4 OP_EQUALVERIFY OP_CHECKSIG]' 3045022100c7b0c2c3972387eda538ad458c259687735f45b55fba939c205b1061f3adb75902203c11d08a6fbebfda588117efe2b948d1ef5962bce4cbb39d2e87c09135f3e3b701 02b37b7f1553fa17fe0ad71c57538f4a93bfada632947c04e1ca6fcdad65e8106d --verbose
```

FYI @igorsyl if you want to have a look while I'm off.

jcnelson:
Got this test working locally.  Let's see if it passes in CI.

There were a couple problems:
* The fee paid by the peg-out request to the peg wallet was so small that it wasn't getting found in the UTXO query
* The fee paid by the peg-out request wasn't getting accounted for in the call to `finalize_tx()`.

netrome:
> Got this test working locally. Let's see if it passes in CI.
> 
> There were a couple problems:
> 
>     * The fee paid by the peg-out request to the peg wallet was so small that it wasn't getting found in the UTXO query
> 
>     * The fee paid by the peg-out request wasn't getting accounted for in the call to `finalize_tx()`.

Great, this passes locally for me too. Thank you for helping out here :pray:

netrome:
I've removed the merge-party from the branch to keep the version history clean and PR size small. I've adopted the integration test update and the test passes locally for me. Again, huge thanks @jcnelson for the help. This would have taken long for me to figure out.

Remaining blockers to get this through:
1. Get CI fixes from `develop` to `next` #3564
2. Rebase & merge peg-in PR #3486 
3. Rebase this PR on top of `next`

FYI @igorsyl

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -1214,6 +1221,152 @@ impl BitcoinRegtestController {
...
+            value: 0,
+            script_pubkey: Builder::new()
+                .push_opcode(opcodes::All::OP_RETURN)
+                .push_slice(&op_bytes)
+                .into_script(),
+        };
+
+        tx.output = vec![amount_and_signature_output];
+        tx.output
+            .push(payload.recipient.to_bitcoin_tx_out(dust_amount));
```
jcnelson, line 1270: I think this is a bug, and an oversight in your integration test.  The peg-out amount for the first non-data output -- the payment to the peg wallet -- should be `payload.amount`, not `dust_amount`.

src/chainstate/burn/operations/peg_out_fulfill.rs
```diff
@@ -0,0 +1,306 @@
...
+use crate::burnchains::BurnchainBlockHeader;
+use crate::burnchains::BurnchainTransaction;
+use crate::burnchains::Txid;
+use crate::chainstate::burn::Opcodes;
+use crate::types::chainstate::StacksAddress;
+use crate::types::Address;
+
+use crate::chainstate::burn::operations::Error as OpError;
+use crate::chainstate::burn::operations::PegOutFulfillOp;
+
```
jcnelson, line 29: I had added some docstrings here describing the UTXOs created by this tx when I fixed the integration test.  Can you restore them?

src/chainstate/burn/operations/peg_out_request.rs
```diff
@@ -0,0 +1,378 @@
...
+use crate::burnchains::BurnchainBlockHeader;
+use crate::burnchains::BurnchainTransaction;
+use crate::chainstate::burn::Opcodes;
+use crate::types::chainstate::StacksAddress;
+use crate::types::Address;
+
+use crate::chainstate::burn::operations::Error as OpError;
+use crate::chainstate::burn::operations::PegOutRequestOp;
+
+impl PegOutRequestOp {
```
jcnelson, line 29: I had added some docstrings here from when I fixed the integration tests that described the UTXOs here.  Can you restore them?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -1214,6 +1221,152 @@ impl BitcoinRegtestController {
...
+        payload: PegOutRequestOp,
+        signer: &mut BurnchainOpSigner,
+    ) -> Option<Transaction> {
+        let public_key = signer.get_public_key();
+        let max_tx_size = 230;
+        let dust_amount = 10000;
+
+        let output_amt = DUST_UTXO_LIMIT
+            + max_tx_size * self.config.burnchain.satoshis_per_byte
+            + payload.amount
```
jcnelson, line 1247: I see now that this is the _requested_ amount to be pegged out.  This should instead be whatever the going "dust" rate is.  Using `payload.amount` is a bug here.

jcnelson (commented):
Almost there.  Just a couple of merge-blocking changes that have to happen.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
netrome: Ah I see. I'll update.

src/chainstate/burn/operations/peg_out_request.rs
netrome: Yes sir! 🫡

src/chainstate/burn/operations/peg_out_fulfill.rs
netrome: Will do 🫡

src/chainstate/burn/operations/peg_out_request.rs
netrome: Done :heavy_check_mark:

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
netrome: Done :heavy_check_mark:

src/chainstate/burn/operations/peg_out_fulfill.rs
netrome: Done :heavy_check_mark:

netrome:
> Almost there. Just a couple of merge-blocking changes that have to happen.

Thanks, I've updated the PR now. I'm just awaiting the peg-in PR to merge and I'll rebase this on `next` instead of the Peg-in branch.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -1214,6 +1221,151 @@ impl BitcoinRegtestController {
...
+        payload: PegOutRequestOp,
+        signer: &mut BurnchainOpSigner,
+    ) -> Option<Transaction> {
+        let public_key = signer.get_public_key();
+        let max_tx_size = 230;
+        let dust_amount = 10000;
+
+        let output_amt = DUST_UTXO_LIMIT
+            + max_tx_size * self.config.burnchain.satoshis_per_byte
+            + payload.fulfillment_fee;
```
jcnelson, line 1247: You should replace `DUST_UTXO_LIMIT` with `dust_amount`.

jcnelson (approved):
LGTM!  Just one last small change and you're good to merge.  Thanks for taking this on @netrome!

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
netrome: Thanks for spotting. I've added it now :heavy_check_mark:

netrome:
> LGTM! Just one last small change and you're good to merge. Thanks for taking this on @netrome!

Awesome! I've updated & rebased the PR now. I will merge this one once CI passes, it looks like I'm authorized to do so now. Thank you for all the help! :pray:

netrome:
Oh, forgot to point to next... I'll open a new PR to fix it.

PR #3511, opened 2023-01-19
sBTC: Add burn ops RPC for PegIn

kantai:
### Description

This adds an RPC endpoint for fetching burn operations:

```http
GET /v2/burn_ops/:burn_height/:op_type
```

It returns a JSON object. The only field of the JSON object is the _type_ of the operation. This type contains a list of the operations at the height (i.e., [externally tagged JSON](https://serde.rs/enum-representations.html#externally-tagged))

Example:

```json
{
  "PegIn": [
    {
      "amount": 1337,
      "block_height": 217,
      "burn_header_hash": "7ce0139a3fe356316409fe704dbdb9ac8f1847c168fe68bd913d0df75c359fe1",
      "peg_wallet_address": {
        "Addr32": [
          false,
          "P2TR",
          [ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ]
        ]
      },
      "recipient": "S0000000000000000000002AA028H",
      "recipient_contract_name": null,
      "txid": "84808f747841fb26e74f171487e56eda1a51373581eb04a8c87a421a77bd46d2",
      "vtxindex": 1
    }
  ]
}
```

### Applicable issues
- implements #3496

Right now, this only works for `peg_in` operations, but extending it to other operations is also very straightforward.

This interface uses human readable serializations for the StacksAddress and the burn header hash (using `[de]serialize_with`), but it doesn't do so for the PoxAddress field. Should this also be specifically serialized in a more human readable format?


src/chainstate/burn/operations/mod.rs
```diff
@@ -313,17 +314,32 @@ pub struct DelegateStxOp {
...
     pub amount: u64, // BTC amount to peg in, in satoshis
 
     // common to all transactions
-    pub txid: Txid,                            // transaction ID
-    pub vtxindex: u32,                         // index in the block where this tx occurs
-    pub block_height: u64,                     // block height at which this tx occurs
+    pub txid: Txid,        // transaction ID
+    pub vtxindex: u32,     // index in the block where this tx occurs
+    pub block_height: u64, // block height at which this tx occurs
+    #[serde(deserialize_with = "hex_deserialize", serialize_with = "hex_serialize")]
```
pavitthrap, line 342: what's the reason behind using custom serialization/de-serialization functions?

pavitthrap (approved):
lgtm

src/chainstate/burn/operations/mod.rs
kantai: It's so that the JSON is a little more human readable (and plays more friendly with libraries like `stacks.js`):

This makes burn header hash look like:

```json
 "burn_header_hash": "7ce0139a3fe356316409fe704dbdb9ac8f1847c168fe68bd913d0df75c359fe1"
```

Rather than

```json
"burn_header_hash": [53,203,40,245,5,29,237,163,250,153,242,20,56,92,222,163,82,59,19,248,205,175,7,57,105,252,93,116,236,86,176,9]
```

The reason to do this via custom serialization rather than redefining the `Serialization` and `Deserialization` implementation for `BurnchainHeaderHash` more generally is just defensive programming: the default serializations may be getting used elsewhere in the codebase, and I wouldn't want to end up breaking a db schema by altering those serializations.

src/chainstate/burn/operations/mod.rs
```diff
@@ -313,17 +314,32 @@ pub struct DelegateStxOp {
...
+    BurnchainHeaderHash::from_hex(&inst_str).map_err(serde::de::Error::custom)
+}
+
 #[derive(Debug, PartialEq, Clone, Eq, Serialize, Deserialize)]
 pub struct PegInOp {
+    #[serde(serialize_with = "crate::chainstate::stacks::address::DisplaySerialize")]
+    #[serde(deserialize_with = "crate::chainstate::stacks::address::AddressDeser")]
     pub recipient: StacksAddress,
     pub recipient_contract_name: Option<String>, // If set, makes the recepient a smart contract principal
     pub peg_wallet_address: PoxAddress,
```
jcnelson, line 335: Do you also want to add a serde helper to `peg_wallet_address` so it looks like a base58/bech32/bech32m address?

jcnelson (approved):
This all looks good to me. Thanks @kantai!

Just had a question about how to format `PoxAddress` instances in `PegInOp`s.  I think it might be friendlier to clients to return them as base58/base32/base32m strings, but will let @donpdonp weigh in since he has to consume this API.

src/chainstate/burn/operations/mod.rs
kantai: It would be nicer, but I was avoiding having to write a `from_b58` function -- is this already implemented somewhere?

jcnelson: It's a little janky right now, but you _could_ do this:

```rust
let addr_str = "<insert base58/bech32/bech32m string here>";
let pox_addr_opt = if let Some(btc_addr) = BitcoinAddress::from_string(addr_str) {
   PoxAddress::try_from_bitcoin_output(&BitcoinTxOutput { address: btc_addr, amount: u0 });
}
else {
   None
};

if let Some(pox_addr) = pox_addr_opt {
   /* do stuff with the PoX address */
}
else {
  /* error out */
}
```

Of course, I'd recommend packing the above into its own method in `PoxAddress` or some such, but the b58/bech32/bech32m parsing logic is already present and tested.

kantai: Added in d4a8d3e

docs/rpc/api/core-node/get-burn-ops.example.json
```diff
@@ -0,0 +1,14 @@
+{
+  "PegIn": [
+    {
+      "amount": 1337,
+      "block_height": 217,
+      "burn_header_hash": "66b2f18530ccf0e57e82dd054cca9fa41bf5f645f7757171ad9c1546b86387e1",
+      "peg_wallet_address": "tb1pqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqkgkkf5",
+      "recipient": "S0000000000000000000002AA028H",
+      "recipient_contract_name": null,
```
netrome, line 9: FYI, I changed this format today after a discussion in #3486. The `recipient` is now represented as `PrincipalData`, and `recipient_contract_name` has been removed completely.

```suggestion
      "recipient": "S0000000000000000000002AA028H",
```

src/chainstate/burn/mod.rs
```diff
@@ -188,6 +188,54 @@ impl SortitionHash {
...
+            Opcodes::PreStx
+        } else if input == Self::HTTP_TRANSFER_STX {
+            Opcodes::TransferStx
+        } else if input == Self::HTTP_DELEGATE_STX {
+            Opcodes::DelegateStx
+        } else {
+            return None;
+        };
+
+        Some(opcode)
```
netrome, line 235: Nit: This would be more readable as a match statement
```suggestion
    pub fn from_http_str(input: &str) -> Option<Opcodes> {
        Some(match input {
            Self::HTTP_PEG_IN => Opcodes::PegIn,
            Self::HTTP_BLOCK_COMMIT => Opcodes::LeaderBlockCommit,
            Self::HTTP_KEY_REGISTER => Opcodes::LeaderKeyRegister,
            Self::HTTP_BURN_SUPPORT => Opcodes::UserBurnSupport,
            Self::HTTP_STACK_STX => Opcodes::StackStx,
            Self::HTTP_PRE_STX => Opcodes::PreStx,
            Self::HTTP_TRANSFER_STX => Opcodes::TransferStx,
            Self::HTTP_DELEGATE_STX => Opcodes::DelegateStx,
            _ => return None,
        })
```

src/chainstate/burn/operations/mod.rs
```diff
@@ -42,6 +42,7 @@ use crate::chainstate::stacks::address::PoxAddress;
 use crate::util_lib::db::DBConn;
 use crate::util_lib::db::DBTx;
 use crate::util_lib::db::Error as db_error;
+use serde::Deserialize;
```
netrome, line 45: Nit: I'd prefer to put this import between the std import and the crate imports.

src/chainstate/burn/operations/mod.rs
```diff
@@ -313,17 +314,34 @@ pub struct DelegateStxOp {
...
+    s.serialize_str(inst.as_str())
+}
+
+fn hex_deserialize<'de, D: serde::Deserializer<'de>>(
+    d: D,
+) -> Result<BurnchainHeaderHash, D::Error> {
+    let inst_str = String::deserialize(d)?;
+    BurnchainHeaderHash::from_hex(&inst_str).map_err(serde::de::Error::custom)
+}
+
```
netrome, line 328: What's the reason for shuffling around these functions? I'd prefer to keep all the burnchain op structs declared together uninterrupted by serialization logic.

src/chainstate/burn/operations/mod.rs
```diff
@@ -339,6 +357,15 @@ pub enum BlockstackOperationType {
...
 }
 
+/// This enum wraps Vecs of a single kind of `BlockstackOperationType`.
+/// This allows `handle_get_burn_ops` to use an enum for the different operation
+///  types without having to buffer and re-structure a `Vec<BlockstackOperationType>`
+///  from a, e.g., `Vec<PegInOp>`
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub enum BurnchainOpsVec {
+    PegIn(Vec<PegInOp>),
+}
```
netrome, line 367: I have two suggestions here:

1. Move this to `src/net/mod`, since it's essentially a helper type for `net::HttpResponseType` and is only used in `src/net`.
2. Nit: Omit "Vec" from the name and just call this `BurnchainOps`.

src/chainstate/stacks/address.rs
```diff
@@ -92,6 +94,34 @@ pub enum PoxAddress {
...
 }
 
+pub fn display_serialize<S: Serializer, T: std::fmt::Display>(
+    input: &T,
+    ser: S,
+) -> Result<S::Ok, S::Error> {
+    ser.serialize_str(&input.to_string())
+}
+
+pub fn address_deser<'de, D: Deserializer<'de>, T: Address>(deser: D) -> Result<T, D::Error> {
```
netrome, line 104: Nit: rename to `address_deserialize` to be consistent with other deserialization function names in the codebase.

src/chainstate/burn/operations/mod.rs
kantai: What do you mean by shuffling? Where should these functions live instead?

netrome (approved):
Overall this looks good to me. Remember to rebase to get the updated `recipient` format. Thank you @kantai :pray:

src/chainstate/burn/operations/mod.rs
netrome: Oh, I read the diff wrong. My apologies! In this case, I simply would ask to move them to after the `PegInOp` declaration instead of before.

src/chainstate/stacks/address.rs
kantai: This method was removed in 92ec39e, because it uses `PrincipalData` deserialization instead. That method is called `principal_deserialize`

src/chainstate/burn/operations/mod.rs
kantai: Moved in 92ec39e4d36420fb9b5c966ac549a84ffc3417cc

src/chainstate/burn/operations/mod.rs
kantai: Moved in 92ec39e4d36420fb9b5c966ac549a84ffc3417cc

docs/rpc/api/core-node/get-burn-ops.example.json
kantai: Yep, updated in 92ec39e4d36420fb9b5c966ac549a84ffc3417cc

src/chainstate/burn/mod.rs
kantai: Definitely! Updated in 40c1208c8

src/chainstate/burn/operations/mod.rs
kantai: Done in 40c1208c8

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -10509,5 +10510,31 @@ fn test_submit_and_observe_peg_in_request() {
...
+
+    let http_origin = format!("http://{}", &conf.node.rpc_bind);
+    let path = format!("{}/v2/burn_ops/{}/peg_in", &http_origin, query_block_height);
+    let client = reqwest::blocking::Client::new();
+
+    let response: serde_json::Value = client.get(&path).send().unwrap().json().unwrap();
+
+    eprintln!("{}", response);
+
+    let parsed_resp: BurnchainOpsVec = serde_json::from_value(response).unwrap();
```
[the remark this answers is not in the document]
kantai: Updated in 40c1208c8

src/chainstate/burn/operations/mod.rs
netrome: Thanks! :pray:

jcnelson:
Hey, just want to confirm -- before this PR merges, the RPC endpoint is going to support all three peg operations that @netrome is working on, right?

kantai:
> Hey, just want to confirm -- before this PR merges, the RPC endpoint is going to support all three peg operations that @netrome is working on, right?

Not unless I rebase it off the last of those PRs, which I'd rather not do that. That would mean this PR would have to stay open until all of those merge, and I think it's probably easier for everyone to close PRs faster (especially once approved). I can do it if necessary, but in that case, someone should just ping me to update this PR once all the others have merged.

kantai:
This has now been updated with everything from `next` -- once the CI finishes, I'll merge.

netrome:
> This has now been updated with everything from `next` -- once the CI finishes, I'll merge.

Awesome. One consideration, I don't see how the serialized format described in `docs/rpc/api/core-node/get-burn-ops.example.json` is tested. Is this something we just manually verify? @kantai (also asked in #3553)

netrome:
> > Awesome. One consideration, I don't see how the serialized format described in docs/rpc/api/core-node/get-burn-ops.example.json is tested. Is this something we just manually verify? @kantai (also asked in #3553)
> 
> Good point -- right now, we just test that the deserialization works in the integration test, but I can add a unit test that uses a fixture string.

Yeah that would be a great complement. FYI @stjepangolemac.

PR #3512, opened 2023-01-20
fix: fix arg-check for passing 2.05 trait in 2.1

obycode:
This is related to #3490, but handling the type check for arguments to contract call transactions.

### DIFF
--- a/clarity/src/vm/analysis/type_checker/v2_1/mod.rs
+++ b/clarity/src/vm/analysis/type_checker/v2_1/mod.rs
@@ -423,8 +423,9 @@ impl FunctionType {
         check_argument_count(expected_args.len(), func_args)?;
 
         if clarity_version < ClarityVersion::Clarity2 {
-            for (expected_arg, arg) in expected_args.iter().zip(func_args.iter()).into_iter() {
-                match (&expected_arg.signature, arg) {
+            for (expected, arg) in expected_args.iter().zip(func_args.iter()).into_iter() {
+                let expected_arg = expected.signature.canonicalize_simple_type_v2_1();
+                match (&expected_arg, arg) {
                     (
                         TypeSignature::CallableType(CallableSubtype::Trait(trait_id)),
                         Value::Principal(PrincipalData::Contract(contract)),


obycode:
I will add a test case for this, but wanted to get the fix pushed.

@hstove Thanks for reporting this. Here is the fix.

kantai (approved):
This LGTM, but we should spend some time with test cases for this, to see if there are any other cases where 2.05 need to be canonicalized

obycode:
I agree. These last couple bugs have been caused by similar issues. We only decided at the last minute to keep the old type checker in addition to the new type checker so it wasn't in my original design to need to deal with these issues. I'll do a deep pass but would appreciate another set of eyes on it.

jcnelson (approved):

jcnelson:
> I agree. These last couple bugs have been caused by similar issues. We only decided at the last minute to keep the old type checker in addition to the new type checker so it wasn't in my original design to need to deal with these issues. I'll do a deep pass but would appreciate another set of eyes on it.

Yes, we'll want to do a deep pass on this.

@kantai Some context -- in the course of testing Stacks 2.1, we kept hitting consensus bugs in the 2.1 type-checker where it would be subtly incompatible with the 2.05 type-checker it replaces.  As a stop-gap, @obycode backported the 2.05 typechecker to 2.1 and feature-gated it.  Once we cross into epoch 2.1, we can remove this code from the 2.1 codebase once we verify that the 2.1 type-checker processes all of epoch 2.05 correctly.  But until then, we have to be read to parse _any_ valid 2.05 block.

kantai:
This seems like something we could try to get Rust's type system to help us with if the 2.05 and 2.1 `TypeSignature`s were actually two different rust types. This would probably be a lot of work (though maybe not? we already duplicate the TypeChecker, and so duplicating TypeSignature would end up similar).

I _think_ that the only way for the 2.1 type checker to ever come in contact with 2.05 types is via the `analysis_db`, and the only interaction that matters is between 2.1 contracts and 2.05 traits (because that's where the type system changed), so we could try to get coverage there by loading synthetic data into a mocked `analysis_db` in the unit tests. But it may be worthwhile to split up the `TypeSignature`  enum.

PR #3517, opened 2023-01-20
Event observer: additional info for burn operations

### LINKED ISSUE #2591: Send additional data to event observer for Stacks-over-BTC wire formats
See issue https://github.com/blockstack/stacks-blockchain-api/issues/415

> this address should be the first output address of the bitcoin transaction. I realize that the `stacks-node` doesn't currently provide enough information to obtain this, so changes would be required there as well, but we should discuss in a new issue.

_Originally posted by @kantai in https://github.com/blockstack/stacks-blockchain-api/pull/414#r564097195_

The API is unable to generate accurate data for stx-stack operations performed via btc txs.

pavitthrap:
### Description
Adds a `burnchain_op` field to transaction payload for the event observer. 

In addition, it makes the json encoding of each burnchain op type more friendly to read. Examples of the readable versions for each op can be seen in the test suite of `src/chainstate/burn/operations/mod.rs`. 
It does this through the function `blockstack_op_to_json`. 

Modified an existing test in `neon_integrations.rs` to ensure this field exists & is populated. 

### Applicable issues
- fixes #2591

- ~~Leaving the names for the BurnchainHeaderHash serialization functions the same as the ones @kantai  used in #3511- in general I used all the same names/files he did so that conflicts would occur in a future merge & could be later be easily fixed.~~ 
- ~~There's one open concern I have around the serialization for PoxAddress. Right now, the serialization does not encode the optional AddressHashMode provided in a standard PoxAddress, meaning if this field is populated with `Some`, then when deserializing, it is set to `None`. I'm waiting for feedback from @kantai on this - I moved forward with this because this is what he did in #3511, but just want to check all the same assumptions hold for the other burn ops that makes this lossiness okay.~~ 
- ~~In the future, I can consolidate some of serialization functions for the byte array types (like BurnchainHeader, BlockHeaderHash, VRFSeed). Right now the ser/deser functions are pretty much identical for a few types so it feels redundant.~~
    -  ~~I could create a hex trait associated with `impl_array_hexstring_fmt`, and move `to_hex` and `from_hex` to that trait. Then, I can create a serialization/deserialization function for a trait object. This would be done to prevent having to create serialization functions for types that are identical under the hood.~~

### Checklist
- [ ] Changelog is updated


docs/event-dispatcher.md
```diff
@@ -57,6 +57,7 @@ Example:
   "transactions": [
     {
       "contract_abi": null,
+      "burnchain_op": null,
```
zone117x, line 60: Can you include samples of the possible values here, and/or some documentation?

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -214,12 +215,14 @@ impl EventObserver {
             }
         };
 
-        let (txid, raw_tx) = match tx {
-            TransactionOrigin::Burn(txid) => (txid.to_string(), "00".to_string()),
+        let (txid, raw_tx, burnchain_op_json) = match tx {
+            TransactionOrigin::Burn(op) => {
+                (op.txid().to_string(), "00".to_string(), json!(op.clone()))
```
kantai, line 220: The default JSON encoding of a `BlockstackOperationType` is pretty unfriendly (I modified it for a specific op type in this PR: #3511 -- so you can look there for what the schema would be without any changes).

@zone117x -- would it be sufficient to just supply the whole bitcoin transaction as a hex string here, or would you prefer something like the encoding used in that PR? It looks like this:

```json
{
  "PegIn": [
    {
      "amount": 1337,
      "block_height": 217,
      "burn_header_hash": "66b2f18530ccf0e57e82dd054cca9fa41bf5f645f7757171ad9c1546b86387e1",
      "peg_wallet_address": "tb1pqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqkgkkf5",
      "recipient": "S0000000000000000000002AA028H",
      "txid": "84808f747841fb26e74f171487e56eda1a51373581eb04a8c87a421a77bd46d2",
      "vtxindex": 1
    }
  ]
}
```

kantai:
@zone117x -- in the issue #2591, you mentioned that the API needs more information from events for things like the `stack-stx` operation to be processed correctly. This PR doesn't alter the "transaction events" portion of the event notifications, does that still need to be done, or did other changes to the PoX events solve that issue?

testnet/stacks-node/src/event_dispatcher.rs
zone117x: The parsed / expanded data like the above sample looks good. It would also be helpful to include the full bitcoin tx hex string as well.

zone117x:
> @zone117x -- in the issue #2591, you mentioned that the API needs more information from events for things like the `stack-stx` operation to be processed correctly. This PR doesn't alter the "transaction events" portion of the event notifications, does that still need to be done, or did other changes to the PoX events solve that issue?

I think the new PoX-2 contract events do include the data needed, although I haven't yet looked into using it for the `stack-stx` bitcoin op parsing code. Let me give that a try and get back to you.

testnet/stacks-node/src/event_dispatcher.rs
jcnelson: If it's not too much trouble to adapt @kantai's work from #3511 to all the other burnchain ops, then we should just bite the bullet and go for it.

jcnelson (commented):
Can you add some unit tests to each of the burnchain ops to verify that the JSON representation is accurate (i.e. conforms to the desired schema)?  That's basically the only thing standing in the way of my approval right now.

jcnelson:
Also, can you rebase this on `develop`?  That's where 2.1 lives now.

jcnelson (approved):
LGTM; just need two things to be done before merging:

* This needs to be based on `develop`, not `next`.
* This should extend @kantai's work on #3511 to use the Rust-idiomatic way of serializing burnchain operations to strings.

testnet/stacks-node/src/event_dispatcher.rs
pavitthrap: done

src/chainstate/burn/operations/mod.rs
```diff
@@ -447,3 +472,398 @@ pub fn parse_u32_from_be(bytes: &[u8]) -> Option<u32> {
 pub fn parse_u16_from_be(bytes: &[u8]) -> Option<u16> {
     bytes.try_into().ok().map(u16::from_be_bytes)
 }
+
+// serialization functions
```
jcnelson, line 476: Just a thought -- it might make sense to put these methods into a file somewhere in `stacks-common` so we can use them in other contexts.  Maybe `stacks-common/src/codec/serde.rs` or something?

jcnelson (approved):
LGTM.  Just a minor code factoring nit about putting serialize/deserialize decorator methods into a common file in `stacks-common`.

docs/event-dispatcher.md
```diff
@@ -76,8 +85,46 @@ Example:
...
+            "address_hash_bytes": "0x89f5fd1f719e4449c980de38e3504be6770a2698",
+            "address_version": 22
+          },
+          "sender": {
+            "address": "ST2QKZ4FKHAH1NQKYKYAYZPY440FEPK7GZ1R5HBP2",
+            "address_hash_bytes": "0xaf3f91f38aa21ade7e9f95efdbc4201eeb4cf0f8",
+            "address_version": 26
+          },
+          "transfered_ustx": 10,
+          "txid": "85aa2106186723f3c4f1d8bb58e3a02746ca9be1be9f4be0c6557079e1f660e6",
```
zone117x, line 106: Is this burnchain/bitcoin txid or Stacks txid?

docs/event-dispatcher.md
```diff
@@ -105,6 +152,68 @@ Example:
...
+    "txid": "0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a",
+    "vtxindex": 10
+  }
+}
+```
+
+- DelegateStx
+```json
+{
+  "DelegateStx": {
```
zone117x, line 203: This `DelegateStx` payload appears to be missing a few values, e.g. `reward_addr`, `ustx`, `delegate_to`, `until-burn-ht`

kantai (changes requested):
These changes look good to me, but I think @zone117x and @jcnelson's comments should addressed. These changes impact the sortitiondb and burndb serialization and deserialization; I think this is fine, though, because 2.1 is already modifying these schemas enough that nodes need to sync from genesis anyways.

However, it does look like the serialization and deserialization changes broke some of the `stack-stx` tests, so it seems like something there needs to be fixed.

docs/event-dispatcher.md
pavitthrap: it's the burnchain txid

docs/event-dispatcher.md
```diff
@@ -76,8 +85,46 @@ Example:
       "microblock_sequence": "3",
       "microblock_hash": "0x9304fcbcc6daf5ac3f264522e0df50eddb5be85df6ee8a9fc2384c54274daaac",
       "microblock_parent_hash": "0x4893ab44636023efa08374033428e44eca490582bd39a6e61f3b6cf749b474bd"
+    },
+    {
+      "burnchain_op": {
+        "TransferStx": {
+          "block_height": 10,
```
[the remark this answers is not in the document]
pavitthrap: it's the burnchain block height

docs/event-dispatcher.md
pavitthrap: good catch! I copied over the wrong op type here.

src/chainstate/stacks/address.rs
```diff
@@ -92,6 +94,49 @@ pub enum PoxAddress {
     Addr32(bool, PoxAddressType32, [u8; 32]),
 }
 
+/// Serializes a PoxAddress as a B58 check encoded address or a bech32 address
+pub fn pox_addr_b58_serialize<S: Serializer>(
```
jcnelson, line 98: These methods don't preserve the `AddressHashMode`, which makes them unsuitable for use within the sortition DB.

I think you have two options:
* You can update the serializers for `PoxAddress` so that the relevant metadata is preserved, or
* You can create a JSON-specific serialization method that will be used for the event stream.

I'm more in favor of the latter since, selfishly, I'm in the process of using already-spun-up nodes for testing and don't want to have to sync from genesis again, but that's something I can work around on my end.  Up to you how you want to proceed.  If you take the latter, please document it.

pavitthrap: Just implemented the latter!

pavitthrap: I added docs for the serialization in the event dispatcher doc already.

docs/event-dispatcher.md
```diff
@@ -105,6 +152,76 @@ Example:
...
+```
+
+- StackStx
+```json
+{
+  "StackStx": {
+    "block_height": 10,
+    "burn_header_hash": "1010101010101010101010101010101010101010101010101010101010101010",
+    "num_cycles": 10,
+    "reward_addr": "16Jswqk47s9PUcyCc88MMVwzgvHPvtEpf",
```
jcnelson, line 187: Does this need the new JSON format?

jcnelson: Meant to say, I see that this format is different from `DelegateStx`.  Is this intentional?

src/chainstate/burn/operations/mod.rs
```diff
@@ -415,6 +430,78 @@ impl BlockstackOperationType {
...
+                "sender": stacks_addr_serialize(&op.sender),
+                "reward_addr": &op.reward_addr.as_ref().map(|(index, addr)| (index, addr.clone().to_b58())),
+                "txid": op.txid,
+                "until_burn_height": op.until_burn_height,
+                "vtxindex": op.vtxindex,
+            }
+
+        })
+    }
+
```
jcnelson, line 492: Can you add a docstring here explaining why an explicit function for JSON serialization is necessary?  Specifically, the JSON serialization is lossy, and less efficient than the serialization mode used in the DB.

docs/event-dispatcher.md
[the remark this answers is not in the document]
kantai: It might make more sense to serialize this as `burn_block_height` instead, i.e., adding the rename field attribute to the `block_height` fields:

```rust
#[serde(rename = "burn_block_height")]
```

docs/event-dispatcher.md
pavitthrap: It's different because the actual field of reward_addr has different types for StackStx (`PoxAddress`) and DelegateStx (`Option<(u32, PoxAddress)>`

kantai (approved):
Looks good to me!

src/chainstate/burn/operations/mod.rs
```diff
@@ -415,6 +430,83 @@ impl BlockstackOperationType {
             BlockstackOperationType::DelegateStx(ref mut data) => data.burn_header_hash = hash,
         };
     }
+
+    pub fn pre_stx_to_json(op: &PreStxOp) -> serde_json::Value {
+        json!({
+            "PreStx": {
```
zone117x, line 436: Once last nit here: this looks like the first time that casing conventions are breaking from `snake_case` and instead are using `PascalCase`. E.g. `PreStx` vs `pre_stx`

pavitthrap: I was mimicking what would happen if you were to call the default serialization function (it would use Pascal case because that's how the op structs are named). 

But I have no problem switching to camel case so I'll push that.

PR #3522, opened 2023-01-24
Fix bug with handling 2.05 trait in 2.1

obycode:
I've been testing this with end-to-end testing (running the node and submitting txs to it), but could use some help writing tests for this.


kantai:
I added a wip integration test: `tests::epoch_21::trait_invocation_cross_epoch`

It publishes 3 contracts in Epoch 2.05:

```rust
    let trait_contract = "(define-trait simple-method ((foo (uint) (response uint uint)) ))";
    let impl_contract = "(impl-trait .simple-trait.simple-method) (define-read-only (foo (x uint)) (ok x))";
    let use_contract = "(use-trait simple .simple-trait.simple-method)
                        (define-public (call-simple (s <simple>)) (contract-call? s foo u0))";
```

And one contract in 2.1

```rust
    let invoke_contract = "
        (use-trait simple .simple-trait.simple-method)
        (define-public (invocation-1)
          (contract-call? .use-simple call-simple .impl-simple))
        (define-public (invocation-2 (st <simple>))
          (contract-call? .use-simple call-simple st))
    ";
```

Then it tries to call `invocation-1` and `invocation-2` with `.impl-simple` supplied via a contract-call. All 5 transactions *should* succeed. The contracts all do typecheck successfully, but the `invocation-2` contract-call tx fails -- it seems like the `s` argument in `.use-simple` does not get included in the `callable_contracts` storage of the calling context (the tx is aborted with a `CheckErrors::ContractCallExpectName`).

clarity/src/vm/types/signatures.rs
```diff
@@ -645,19 +717,38 @@ impl TypeSignature {
...
+            }
+            OptionalType(ref inner_type) => OptionalType(Box::new(inner_type.canonicalize_v2_1())),
+            ResponseType(ref inner_type) => ResponseType(Box::new((
+                inner_type.0.canonicalize_v2_1(),
+                inner_type.1.canonicalize_v2_1(),
+            ))),
+            TupleType(ref tuple_sig) => {
+                let mut canonicalized_fields = BTreeMap::new();
+                for (field_name, field_type) in tuple_sig.get_type_map() {
+                    canonicalized_fields.insert(field_name.clone(), field_type.canonicalize_v2_1());
```
pavitthrap, line 746: is it possible to modify in place instead? maybe implement a `get_mut_type_map`?

obycode:
@kantai's test is passing now, but the codecov keeps failing. I am working on adding more tests.

clarity/src/vm/types/signatures.rs
obycode: The rest of the cases don't require `&mut self`, mainly because I thought it was bad practice to modify `self` in the case of the simple types. That's why this method takes an immutable `self` and returns a `TypeSignature`.

kantai:
Added one more test that I was curious about, now I'll stop pushing to this branch :)

obycode:
Thanks @kantai. Your testing is much appreciated! I just added another test as well. Looks like they are all passing.

kantai:
Yep -- there's one particularly aggressive test that I wrote that doesn't pass, but I'm not sure if we should try to address it in this PR (or address it at all). Basically, I _force publish_ a contract that supplies a dynamic argument to a trait. This would fail typechecking (like the test I added above), but if forced, the runtime will actually allow the dynamic execution. Do you think that we should be trying to more paranoid in the runtime somehow, or do you think that's going to end up adding even more complexity to this codepath?

obycode:
> Do you think that we should be trying to more paranoid in the runtime somehow, or do you think that's going to end up adding even more complexity to this codepath?

Yeah, there are several places that I am aware of that make assumptions that the type-checker has already validated some property. For example, in `execute_apply`, we don't actually check that the trait is implemented, assuming that this has already been checked. But then when the stored value gets used, in `special_contract_call`, it handles the errors with appropriate `CheckErrors` and not `unwrap`, et al.

So long as we think we can always error out appropriately at the later use sites, my preference would be to leave the checks out to avoid adding unnecessary runtime.

kantai:
Sounds good to me -- and I lied before about being done pushing to the branch: my test lacked a couple unwraps(), so I added them.

kantai (approved):
This LGTM!

src/clarity_vm/tests/contracts.rs
```diff
@@ -324,3 +326,496 @@ fn test_get_block_info_eval_v210() {
...
+#[test]
+fn trait_invocation_205_with_stored_principal() {
+    let mut sim = ClarityTestSim::new();
+    sim.epoch_bounds = vec![0, 3, 5];
+
+    // Advance two blocks so we get to Stacks 2.05.
+    sim.execute_next_block(|_env| {});
+    sim.execute_next_block(|_env| {});
+
+    let trait_contract = "(define-trait simple-method ((foo (uint) (response uint uint)) ))";
```
pavitthrap, line 356: To make test more readable, consider extending contract names (same for tests below). 
Ex. "simple-trait" -> "simple-trait-contract" 
"use-simple" -> "use-simple-contract"

src/clarity_vm/tests/contracts.rs
```diff
@@ -324,3 +326,496 @@ fn test_get_block_info_eval_v210() {
...
+                CheckErrors::TypeError(..) => (),
+                _ => panic!("Unexpected error: {:?}", error),
+            },
+            _ => panic!("Unexpected error: {:?}", error),
+        };
+    });
+}
+
+#[test]
+fn trait_invocation_cross_epoch() {
```
pavitthrap, line 400: add small comment describing purpose of each unit test in this file

testnet/stacks-node/src/tests/epoch_21.rs
```diff
@@ -4677,6 +4677,222 @@ fn test_sortition_divergence_pre_21() {
...
+                    parsed.txid(),
+                    tx.get("status").unwrap().as_str().unwrap()
+                );
+                results.push(tx.get("status").unwrap().as_str().unwrap().to_string());
+                eprintln!(
+                    "{} => {}",
+                    parsed.txid(),
+                    tx.get("raw_result").unwrap().as_str().unwrap()
+                );
+                //   eprintln!("{}", tx);
```
pavitthrap, line 4881: remove comment

clarity/src/vm/contexts.rs
```diff
@@ -1837,6 +1837,18 @@ impl ContractContext {
...
+
+    pub fn canonicalize_types(&mut self, epoch: &StacksEpochId) {
+        for (_, function) in self.functions.iter_mut() {
+            function.canonicalize_types(epoch);
+        }
+
+        for trait_def in self.defined_traits.values_mut() {
+            for (_, function) in trait_def.iter_mut() {
+                *function = function.canonicalize(epoch);
+            }
```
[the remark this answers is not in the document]
obycode: I'll add a comment. It's because only the functions and defined traits are exposed externally.

src/clarity_vm/tests/contracts.rs
obycode: I'm going to push back on this one. I don't think I see the benefit of this change. I'd prefer not to encourage naming every contract with the `-contract` suffix, since the way it is used should already indicate that it must be a contract.

clarity/src/vm/callables.rs
```diff
@@ -211,6 +211,23 @@ impl DefinedFunction {
                             },
                         );
                     }
+                    // Since this Clarity 1 contract may be called from a Clarity 2 contract,
+                    // we need to handle Clarity 2 values as well.
+                    (
+                        TypeSignature::CallableType(CallableSubtype::Trait(_)),
```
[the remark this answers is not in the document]
obycode: Addressed in be86fa7f8

clarity/src/vm/contexts.rs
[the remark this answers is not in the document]
obycode: Addressed in be86fa7f8

src/clarity_vm/tests/contracts.rs
obycode: Addressed in be86fa7f8

testnet/stacks-node/src/tests/epoch_21.rs
obycode: Addressed in be86fa7f8

lgalabru:
Note: this PR is blocking some community members in testing their migration to 2.1.

clarity/src/vm/analysis/analysis_db.rs
```diff
@@ -185,14 +221,18 @@ impl<'a> AnalysisDatabase<'a> {
...
-            .load_contract(contract_identifier)
+            .load_contract_non_canonical(contract_identifier)
             .ok_or(CheckErrors::NoSuchContract(contract_identifier.to_string()))?;
         let map_type = contract
             .get_map_type(map_name)
             .ok_or(CheckErrors::NoSuchMap(map_name.to_string()))?;
-        Ok(map_type.clone())
+        Ok((
+            map_type.0.canonicalize(epoch),
+            map_type.1.canonicalize(epoch),
```
jcnelson, line 234: Should we also be canonicalizing data-var types?

obycode: My assumption was that we only need to canonicalize the outward-facing types, which would include public and read-only functions, and defined traits. I'm not actually sure where/how this `get_map_type` method would get used.

pavitthrap: It looks like this function is not used anywhere - perhaps consider deleting

pavitthrap:
Some thoughts on how the change in types in Clarity2 affect RPC calls - potentially no action is needed, but I just wanted to flag these:
/v2/contracts/interface - in `handle_get_contract_abi`, the function `load_contract` is called, which means that types are automatically canonicalized. Maybe users should have an option to not have the types canonicalized? 

/v2/map_entry - the key is provided as a serialized Clarity value. Just to ensure we are on the same page, if the key is type principal, we would need the provided key to have type Value::Principal, and not type Value::CallableContract, right?

clarity/src/vm/analysis/analysis_db.rs
obycode: Sounds good to me. Any one have a reason to keep it?

kantai: Nope -- I think this was vestigial from when Clarity had a function to fetch data from an arbitrary contract's data map. Delete away!

obycode:
> /v2/contracts/interface - in `handle_get_contract_abi`, the function `load_contract` is called, which means that types are automatically canonicalized. Maybe users should have an option to not have the types canonicalized?

I believe the types used here are already handled to be the same in either epoch, for example, from `ContractInterfaceAtomType::from_type_signature`:
```
            PrincipalType => ContractInterfaceAtomType::principal,
            CallableType(CallableSubtype::Principal(_)) => ContractInterfaceAtomType::principal,
            CallableType(CallableSubtype::Trait(_)) | TraitReferenceType(_) => {
                ContractInterfaceAtomType::trait_reference
            }
            ListUnionType(_) => ContractInterfaceAtomType::principal,
```

> /v2/map_entry - the key is provided as a serialized Clarity value. Just to ensure we are on the same page, if the key is type principal, we would need the provided key to have type Value::Principal, and not type Value::CallableContract, right?

Yes, that's right. `CallableContract` values should only ever arise from function parameters, and not used elsewhere.

clarity/src/vm/analysis/analysis_db.rs
obycode: Awesome. Deleted in b1343ea55.

pavitthrap (approved):
lgtm!

clarity/src/vm/analysis/types.rs
```diff
@@ -187,6 +187,21 @@ impl ContractAnalysis {
...
+            *function_type = function_type.canonicalize(epoch);
+        }
+        for (_, function_type) in self.read_only_function_types.iter_mut() {
+            *function_type = function_type.canonicalize(epoch);
+        }
+        for (_, trait_definition) in self.defined_traits.iter_mut() {
+            for (_, function_signature) in trait_definition.iter_mut() {
+                *function_signature = function_signature.canonicalize(epoch);
+            }
+        }
```
jcnelson, line 202: How defensive do we want to be here?  Should we canonicalize _everything_, even if it's not outward facing?  Asking because in pox-2, for example, there's a private function that the node will call directly.

clarity/src/vm/contexts.rs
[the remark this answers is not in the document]
jcnelson: Also here -- how defensive do we want to be?  Should we just canonicalize _everything_ to be safe?

jcnelson (approved):
This overall LGTM.  Just had a question about defensive canonicalization.

We should make sure that the unit tests pass first.  Looks like there's a stack overflow?

obycode:
> We should make sure that the unit tests pass first. Looks like there's a stack overflow?

I don't believe anything in this PR should have affected that test case. What is the stack size used for the stacks node? The test case fails with the default 2MB limit, but passes with a 4MB limit (or even less). Anyone aware of anything that would've changed this stack limit for the test execution?

clarity/src/vm/analysis/types.rs
obycode: I'm not strongly opposed to canonicalizing everything to be safe. I can make that change today.

kantai:
> We should make sure that the unit tests pass first. Looks like there's a stack overflow?
> 
> I don't believe anything in this PR should have affected that test case. What is the stack size used for the stacks node? The test case fails with the default 2MB limit, but passes with a 4MB limit (or even less). Anyone aware of anything that would've changed this stack limit for the test execution?

It may be that the canonicalization changes increase stack depth just enough to trigger this issue

obycode:
> It may be that the canonicalization changes increase stack depth just enough to trigger this issue

Yes could be. This test is failing for #3536 too.

kantai:
It fails for me in `develop` as well, so it's probably not specifically related to this PR.

kantai:
I'll open a separate issue for this, and poke around at it. Hopefully an easy fix

clarity/src/vm/analysis/types.rs
obycode: Added in 70d72a992.

PR #3528, opened 2023-01-24
Fix/3527

### LINKED ISSUE #3527: [Stacks 2.1] burnchain operations must include the burnchain hash in their primary key
The underlying issue of #3520 is that a burnchain operation can be mined in two burnchain forks, but we store metadata about the operation that is specific to the fork.  What this means is that any time we store a burnchain operation in any database, the primary key must refer to both the operation itself (e.g. its txid) and the fork that includes it (e.g. the burnchain hash, the sortition ID) so that both fork metadata sets are preserved.  Failure to do this can lead to **loss of data** and **database corruption** in the event that both forks are processed without this protection.

Right now, some tables do not do this.  Specifically:

* `stack_stx` in the sortition DB
* `transfer_stx` in the sortiiton DB
* `delegate_stx` in the sortition DB

jcnelson:
This fixes #3527.  All burnchain operations are now keyed by sortition ID and txid, so if the same transaction gets mined in two burnchain blocks, we'll store both burnchain blocks' metadata.

Unfortunately, there's no way to apply this change to an already-existing sortition DB.  Once this PR lands in -rc4, you'll need to re-sync from genesis.


src/chainstate/burn/db/sortdb.rs
```diff
@@ -2224,29 +2240,30 @@ impl<'a> SortitionHandleConn<'a> {
...
-        let mut memoized_candidates: HashMap<_, (Txid, u64)> = HashMap::new();
+        let mut memoized_candidates: HashMap<_, (SortitionId, Txid, u64)> = HashMap::new();
 
         // iterate over every sortition winner in the prepare phase
         //   looking for their highest ancestor _before_ prepare_begin.
         let winners = self.get_sortition_winners_in_fork(prepare_begin, prepare_end)?;
-        for (winner_commit_txid, winner_block_height) in winners.into_iter() {
-            let mut cursor = (winner_commit_txid, winner_block_height);
+        for (winner_sort_id, winner_commit_txid, winner_block_height) in winners.into_iter() {
+            let mut cursor = (winner_sort_id, winner_commit_txid, winner_block_height);
```
pavitthrap, line 2249: I would recommend using a type for cursor (it'll make the code below more readable)

pavitthrap (approved):
lgtm, just had one style nit

kantai (commented):
This looks good to me, but there should be a test to cover this change.

jcnelson:
So, this PR is actually designed wrong.  The primary key for the burnchain operations should be `(txid, burn_header_hash)`.  There's no need to associate on-chain `stack-stx`, `delegate-stx`, or `transfer-stx` with a particular sortition history within the sortition DB, because their validity is determined when processing a Stacks block.  This is _not_ true for leader block commits, because their validity _is_ dependent on which sortition history they are processed in, and this processing check happens as part of executing a sortition.

Leader key registers could be keyed by `(txid, burn_header_hash)` as well, since they are valid in all sortitions.  The fact that they are currently keyed by `(txid, sortition_id)` is still consistent with a correct implementation, but too strict and leads to some data duplication -- the same leader key row data can be stored for multiple sortitions on the same burnchain block.  However, I don't see the need to fix this in this PR, since it's out-of-scope and the consequence of leaving this as-is is insignificant.  I'll leave a code comment instead.

netrome (approved):
Looks good to me. I would also like to see a struct for the [cursor](https://github.com/stacks-network/stacks-blockchain/pull/3528/files#r1085963703) for better readability.

PR #3535, opened 2023-01-27
Replace Vec References with slices in function inputs

### LINKED ISSUE #3484: Take slices as input instead of Vec references
Many functions currently take `&Vec<_>` as arguments. We should almost always be able to replace this with slices `&[_]`. For more info, read [this Clippy lint](https://rust-lang.github.io/rust-clippy/master/#ptr_arg)

asimm241:
### Description
This PR replaces all vector references with slices. It uses Clippy's  [ptr_arg](https://rust-lang.github.io/rust-clippy/master/#ptr_arg) lint for reference.
### Applicable issues
- fixes #3484


src/chainstate/stacks/index/node.rs
```diff
@@ -669,9 +669,9 @@ impl fmt::Debug for TrieNode4 {
 }
 
 impl TrieNode4 {
-    pub fn new(path: &Vec<u8>) -> TrieNode4 {
+    pub fn new(path: &[u8]) -> TrieNode4 {
         TrieNode4 {
-            path: path.clone(),
+            path: path.to_owned(),
```
netrome, line 674: Generally speaking in scenarios like this one, I'd argue it's better to take the owned type as input instead of cloning a reference. So I'd change the input to `path: Vec<u8>`, leaving it up to the caller to decide when to clone.

I recognize that this is beyond the scope of the current refactor, so feel free to leave that for future improvements.

netrome (changes requested):
This generally looks good to me. You also need to update the `announce_block` implementation on line 454 in `testnet/stacks-node/src/event_dispatcher.rs:454:19` to make `testnet/stacks-node` compile.

netrome (approved):

pavitthrap (approved):
lgtm, thanks for the PR!

src/chainstate/stacks/index/node.rs
jcnelson: Yeah, don't do this.

jcnelson (approved):
Thanks @asimm241!

src/chainstate/stacks/index/node.rs
asimm241: @netrome I have fixed where it could be simply fixed. Other places could produce ripples in the code and it should be done in a separate PR.

netrome: Great, thank you!

jcnelson:
Need unit tests to pass before we can merge @asimm241

PR #3536, opened 2023-01-27
feat: add epoch to abi RPC endpoint

obycode:
It can be useful for clients to know during which epoch a contract was
published. This was a specific request from the Clarinet team.

### DIFF
--- a/clarity/src/vm/analysis/arithmetic_checker/tests.rs
+++ b/clarity/src/vm/analysis/arithmetic_checker/tests.rs
@@ -59,6 +59,7 @@ fn arithmetic_check(
         contract_identifier,
         expressions,
         LimitedCostTracker::new_free(),
+        epoch,
         version,
     );
 
--- a/clarity/src/vm/analysis/contract_interface_builder/mod.rs
+++ b/clarity/src/vm/analysis/contract_interface_builder/mod.rs
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+use stacks_common::types::StacksEpochId;
+
 use crate::vm::analysis::types::ContractAnalysis;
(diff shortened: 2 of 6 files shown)


kantai (approved):
This LGTM, but we should probably note that this alters the VM database schema vs the current develop.

pavitthrap (commented):
lgtm

obycode:
> This LGTM, but we should probably note that this alters the VM database schema vs the current develop.

Good point. Any potential problems from that?

obycode:
This is causing a stack overflow on the `test_cost_tracking_deep_contracts_2_1` test. Hmmm...

jcnelson (approved):

jcnelson:
> Good point. Any potential problems from that?

This means that -rc4 will need to be spun up from genesis.  This is already anticipated for #3524 and #3528.

pavitthrap (approved):
lgtm part 2 (accidentally didn't approve it last time). 

Just flagging the failing unit test: `test_cost_tracking_deep_contracts_2_1`. Do we sometimes have flaky stack overflow errors or this a real error?

obycode:
> Just flagging the failing unit test: `test_cost_tracking_deep_contracts_2_1`. Do we sometimes have flaky stack overflow errors or this a real error?

See #3522 also -- this failure seems unrelated to this PR. @kantai opened #3541 to investigate further.

jcnelson:
@obycode when you get a moment, can you fix the unit tests so we can cut -rc4?  Thanks!

kantai:
> can you fix the unit tests so we can cut -rc4? Thanks!

Pushed a 1-line fix for the test build (was failing from the auto merge).

jcnelson:
@obycode @kantai looks like the fix might need to be a little bit more involved

obycode:
> @obycode @kantai looks like the fix might need to be a little bit more involved

I'll take a look.

obycode:
That should do it.

PR #3537, opened 2023-01-27
feat: mocknet stacks 2.1

bestmike007:
### Description

Currently when using `mocknet-miner-conf.toml` to start a stacks node, it crashes with panic: `FATAL: no epoch 2.1 defined`. This PR is to fix this by adding the 2.1 epoch to the initial state.

### Checklist
- [ ] Changelog is updated


testnet/stacks-node/src/burnchains/mocknet_controller.rs
```diff
@@ -99,13 +101,22 @@ impl BurnchainController for MocknetController {
...
+                StacksEpoch {
+                    epoch_id: StacksEpochId::Epoch20,
+                    start_height: 0,
+                    end_height: 1,
+                    block_limit: ExecutionCost::max_value(),
+                    network_epoch: PEER_VERSION_EPOCH_2_0,
+                },
+                StacksEpoch {
+                    epoch_id: StacksEpochId::Epoch21,
+                    start_height: 1,
```
lgalabru, line 114: Instead of hard coding start heights, could we read them from the config fil?
Thanks!

jcnelson: Yup, seconded.

zone117x: Seems like this would create multiple ways to specify epoch start heights? How does this new config differ and interact with the existing one? e.g.
```toml
[[burnchain.epochs]]
epoch_name = "2.1"
start_height = 3
```

bestmike007: That's what I was thinking. If we really need to customize the start height for epoch 2.1, we can use the existing way to configure epochs.

So should I keep the new configuration or revert it?

bestmike007: Reverted.

testnet/stacks-node/src/burnchains/mocknet_controller.rs
```diff
@@ -99,13 +101,22 @@ impl BurnchainController for MocknetController {
...
-                network_epoch: PEER_VERSION_EPOCH_2_0,
-            }],
+            None => vec![
+                StacksEpoch {
+                    epoch_id: StacksEpochId::Epoch20,
+                    start_height: 0,
+                    end_height: 1,
+                    block_limit: ExecutionCost::max_value(),
+                    network_epoch: PEER_VERSION_EPOCH_2_0,
+                },
```
jcnelson, line 111: Hey, you're going to need to add an epoch here to transition the system to Epoch 2.05.  After that, you can transition to 2.1.

jcnelson (commented):
This LGTM; just need to add the epoch to transition the system from epoch 2.0 to 2.05 and then to 2.1.

testnet/stacks-node/src/burnchains/mocknet_controller.rs
bestmike007: Amended.

bestmike007:
I noticed the latest 2.1 rc4 version was tagged from develop branch. Should I change this PR into merging into develop?

jcnelson:
Yeah that'd be best.

jcnelson (approved):
Thanks for making the changes!

bestmike007:
Done.

kantai (approved):

PR #3543, opened 2023-01-30
2.1: Eliminate recursion in `parse_list` and `parse_tuple`

### LINKED ISSUE #3541: Unit test triggers overflow in `develop`
`vm::ast::test::test_cost_tracking_deep_contracts_2_1` triggers a stack overflow with a default 2MB stack in `develop`.

kantai:
### Description

This addresses #3541 by eliminating the recursion in `parse_list` and `parse_tuple` (which were the only recursive calls in `parse_node()`.  It replaces that recursion with a do-while loop in `parse_node()`. The return value of `parse_node()` is unchanged: it still returns the next expression at the "current level" of the AST, however, for list and tuple expressions, it does this without invoking `parse_node()` again.

To do this, I added a new enum for open tuple expressions and open list expressions (i.e., expressions which are waiting on child nodes), and broke the logic from `parse_list` and `parse_tuple` into the handlers for those open expressions. This means that the logic is now more spread out than before, and relies on the `parse_node()` loop.

### Applicable issues
- fixes #3541

This makes the V2 AST parser non-recursive, however, it is kind of a sensitive change to the parser. We do have lots of tests cases for the parser, so I'm not too worried that this is radically altering anything, but I think that it is definitely worth careful review here.

### DIFF
--- a/clarity/src/vm/ast/parser/v2/mod.rs
+++ b/clarity/src/vm/ast/parser/v2/mod.rs
@@ -196,7 +196,11 @@ impl<'a> Parser<'a> {
         }
     }
 
-    /// Returns Some(node) if the open node is finished and should be popped from the stack
+    /// Process a new child node for an AST expression that is open and waiting for children nodes. For example,
+    ///  a list or tuple expression that is waiting for child expressions.
+    ///
+    /// Returns Some(node) if the open node is finished and should be popped from the stack.
+    /// Returns None if the open node is not finished and should remain on the parser stack.
     fn handle_open_node(
         &mut self,
         open_node: &mut ParserStackElement,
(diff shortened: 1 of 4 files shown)


obycode (approved):
This LGTM. Just noted one typo. Thanks @kantai!

clarity/src/vm/ast/parser/v2/mod.rs
```diff
@@ -164,189 +196,317 @@ impl<'a> Parser<'a> {
...
+            }
+            OpenTupleStatus::ParseValue => {
+                // expecting to parse a value
+                let node = match node_opt {
+                    Some(node) => node,
+                    None => {
+                        // mimic parse_node_or_eof() behavior
+                        //  if last token was an EOF, error out the tuple
+                        //  if the last token was something else, just yield back to the parse loop
+                        let last_token = self.tokens[self.next_token - 1].clone();
```
jcnelson, line 362: Both here and above, we are _absolutely_ cure that this will not be an out-of-bound index, right?

jcnelson (approved):
LGTM.  We'll want to test this from genesis.

clarity/src/vm/ast/parser/v2/mod.rs
obycode: I reviewed these accesses before and they should be safe.

To ensure these remain safe with future changes, we could add a method `peek_prev_token`, similar to the `peek_next_token` method.

PR #3548, opened 2023-01-31
 pox endpoint changed to use current stacks height

donpdonp:
Description

change /v2/pox report for active contract_id to use the current stacks block height instead of the next reward cycle block height to more accurately reflect when POX-2 becomes the default contract. POX-2 becomes the default at v1_unlock_height+1.
Applicable issues

    fixes 

    https://github.com/stacks-network/stacks-blockchain/issues/3422

Additional info (benefits, drawbacks, caveats)
Checklist

    [X ] Test coverage for new or modified code paths


jcnelson (approved):
LGTM!

src/net/rpc.rs
```diff
@@ -475,7 +476,7 @@ impl RPCPoxInfoData {
         let cur_cycle_pox_active = sortdb.is_pox_active(burnchain, &burnchain_tip)?;
 
         Ok(RPCPoxInfoData {
-            contract_id: boot_code_id(cur_cycle_pox_contract, chainstate.mainnet).to_string(),
+            contract_id: boot_code_id(cur_block_pox_contract, chainstate.mainnet).to_string(),
```
kantai, line 479: Should there be a new entry in the RPC info for the current cycle's PoX contract?

testnet/stacks-node/src/tests/epoch_21.rs
```diff
@@ -4677,6 +4679,131 @@ fn test_sortition_divergence_pre_21() {
...
+    if env::var("BITCOIND_TEST") != Ok("1".into()) {
+        return;
+    }
+
+    let reward_cycle_len = 5;
+    let prepare_phase_len = 3;
+    let epoch_2_05 = 205;
+    let epoch_2_1 = 212;
+    let v1_unlock_height = 213;
+
```
kantai, line 4694: This test looks like it could be a unit test rather than an integration test -- the `RPCPoxInfoData::from_data` function can be tested in a test like `rpc::test_rpc_getpoxinfo_use_latest_tip`. I think the `test_rpc()` function could be setup to process an epoch transition: @jcnelson, does that seem reasonable?

kantai (changes requested):
I think the current cycle's PoX contract should probably still be a field in the RPC response, and I believe that this test should be a unit test (or at worst, an additionaly assertion in the existing `epoch_21` tests) rather than an entirely new `epoch_21` test.

testnet/stacks-node/src/tests/epoch_21.rs
jcnelson: I had recommended to @donpdonp to implement an integration test instead of a unit test precisely because `test_rpc()` does not execute epoch transitions, and adding that ability without breaking the other tests and duplicating code would have required more time and effort.  It was a trade-off between velocity versus code hygiene -- I prioritized velocity because it's part of the 2.1 upgrade, which is already very late.

However, if we feel that code hygiene is more important here, then yes -- creating a variant of `test_rpc()` or modifying `test_rpc()` to do an epoch transition at a pre-determined height would be the "cleaner" thing to do.

kantai: That's fine -- in that case, this testing should be changed in two ways:

1. The assertion should just be added to an existing test: nothing is happening in this test that does not already occur in any of the other 2_1 tests
2. Because this is an end to end test, the assertion should use the `neon_integrations::get_pox_info` method rather than trying to directly read from the database

src/net/rpc.rs
donpdonp: I'm not sure - No mention was made in the requesting ticket for a reward-based active contract field. https://github.com/stacks-network/stacks-blockchain/issues/3422 

maybe @janniks has some input.

testnet/stacks-node/src/tests/epoch_21.rs
donpdonp: I looked through the existing integration tests and found a good spot in test `transition_removes_pox_sunset` that was already failing due to the active contract being block-based instead of reward-cycle based. So it was a matter of updating that test to look for pox-2 at v1_unlock_height+1. The new integration test has been removed completely.

kantai (approved):

PR #3549, opened 2023-01-31
[testing] Add integration test to verify that v1 unlock height and epoch 2.1 work as expected

jcnelson:
This PR doesn't add any new features.  It just increases test coverage by verifying that PoX 2 will activate if it is instantiated at the very end of a reward cycle, and the v1 unlock height passes at the very beginning of the next.


obycode (approved):
LGTM!

kantai (approved):
LGTM

PR #3534, opened 2023-01-27
Feat/stackerdb

jcnelson:
Closing in favor of:
- https://github.com/stacks-network/stacks-blockchain/pull/3560
- https://github.com/stacks-network/stacks-blockchain/pull/3558
- https://github.com/stacks-network/stacks-blockchain/pull/3552
- https://github.com/stacks-network/stacks-blockchain/pull/3551

This PR creates a **Stacker DB** subsystem in the peer network.  A stacker DB is an ephemeral database that can be created for any existing smart contract that confirms to a simple trait interface.  Nodes opt-in to hosting replicas of the DB, and users of the smart contract get to leverage the Stacks p2p network to store app-specific state for up to one reward cycle at a time.

This PR is just a draft because the system is not complete.  This PR still needs the following:

- [ ] An RPC interface for getting chunks
- [ ] An RPC interface for storing chunks
- [ ] An update to `/v2/neighbors` to include the list of stacker DBs that the neighbor claims to replicate
- [ ] The ability to query the DB's associated smart contract for a list of principals and number of DB slots they can write to
- [ ] The ability to vacuum DBs every reward cycle
- [ ] Messages for nodes to inform their neighbors when they have new chunks, as well as the means for neighbors to react by pulling the chunks if they are newer (right now, the system is only pull-oriented).
- [ ] Integration tests for all of the above.

I'm creating this draft to (1) start running CI on it and (2) start working with @donpdonp and @xoloki to get the RPC interface and contract interface nailed down.

PR #3551, opened 2023-02-04
Feat/stackerdb messages

jcnelson:
This PR implements the message codes and handshake protocol for nodes that support Stacker DBs.  This is part of an effort to split #3534 into a series of smaller PRs.


kantai:
Can you add some rustdocs that provide an overview of what this new protocol is for, how it should work, and how it fits into the rest of the networking code? Otherwise, this is pretty difficult to provide a thorough review.

jcnelson:
> Can you add some rustdocs that provide an overview of what this new protocol is for, how it should work, and how it fits into the rest of the networking code? Otherwise, this is pretty difficult to provide a thorough review.

Yes; I just added some in c2d3ebd41d1ddaceda1bab001cfdbd656e18a47d

src/burnchains/mod.rs
```diff
@@ -521,6 +521,7 @@ pub struct BurnchainView {
     pub burn_stable_block_height: u64, // latest stable block height (e.g. chain tip minus 7)
     pub burn_stable_block_hash: BurnchainHeaderHash, // latest stable burn block hash
     pub last_burn_block_hashes: HashMap<u64, BurnchainHeaderHash>, // map all block heights from burn_block_height back to the oldest one we'll take for considering the peer a neighbor
+    pub rc_consensus_hash: ConsensusHash, // consensus hash of the current reward cycle's start block
```
kantai, line 524: This comment should be a rustdoc instead.

src/net/chat.rs
```diff
@@ -139,6 +140,7 @@ pub struct NeighborStats {
     pub block_push_rx_counts: VecDeque<(u64, u64)>, // (timestamp, num bytes)
     pub microblocks_push_rx_counts: VecDeque<(u64, u64)>, // (timestamp, num bytes)
     pub transaction_push_rx_counts: VecDeque<(u64, u64)>, // (timestamp, num bytes)
+    pub stackerdb_push_rx_counts: VecDeque<(u64, u64)>, // (timestamp, num bytes)
```
kantai, line 143: This comment should be a rustdoc as well

src/net/chat.rs
```diff
@@ -201,6 +204,14 @@ impl NeighborStats {
         }
     }
 
+    pub fn add_stackerdb_push(&mut self, message_size: u64) -> () {
+        self.stackerdb_push_rx_counts
+            .push_back((get_epoch_time_secs(), message_size));
+        while self.stackerdb_push_rx_counts.len() > NUM_BLOCK_POINTS {
+            self.stackerdb_push_rx_counts.pop_front();
+        }
+    }
```
kantai, line 213: Can you add a rustdoc to this method (that also explains that this method will drop older entries)?

src/net/chat.rs
```diff
@@ -318,6 +334,9 @@ pub struct ConversationP2P {
 
     pub stats: NeighborStats,
 
+    // which stacker DBs this peer replicates
```
kantai, line 337: Should be a rustdoc, and should also explain that StackerDBs can be identified by ContractId

src/net/chat.rs
```diff
@@ -635,6 +656,21 @@ impl ConversationP2P {
         (peer_services & expected_bits) == expected_bits
     }
 
+    /// Does this remote neighbor support stacker DBs?  It will if it has the STACKERDB bit set
```
kantai, line 659: `this` normally refers to `self`, but `supports_stackerdb` is static.

src/net/chat.rs
```diff
@@ -1182,11 +1232,28 @@ impl ConversationP2P {
...
+        let stacks_message = if ConversationP2P::supports_stackerdb(local_peer.services)
+            && ConversationP2P::supports_stackerdb(self.peer_services)
+        {
+            StacksMessageType::StackerDBHandshakeAccept(
+                accept_data,
+                StackerDBHandshakeData {
+                    rc_consensus_hash: chain_view.rc_consensus_hash.clone(),
+                    // placeholder sbtc address for now
+                    smart_contracts: vec![
+                        ContractId::parse("SP000000000000000000002Q6VF78.sbtc").unwrap()
```
kantai, line 1244: Can you use a const for this placeholder?

src/net/mod.rs
```diff
@@ -904,6 +910,92 @@ pub enum MemPoolSyncData {
...
+    pub fn name(&self) -> ContractName {
+        self.0.name.clone()
+    }
+
+    pub fn parse(txt: &str) -> Option<ContractId> {
+        QualifiedContractIdentifier::parse(txt)
+            .ok()
+            .map(|qc| ContractId(qc))
+    }
+}
```
kantai, line 938: What we do elsewhere in the code is add a trait for extensions, which allows you to not need to wrap the object in a new struct. Extension traits are slightly more idiomatic, but probably more importantly, we should strive to only have one way of implementing extensions in our codebase, so I'd prefer if this were implemented via an extension trait.

src/net/stackerdb/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+/// StackerDB replication proceeds in a three-part protocol: discovery, inventory query, and
+/// chunk exchange.  The discovery protocol leverages the Stacks node's neighbor-walk algorithm to
+/// discover which StackerDBs other nodes claim to replicate.  On receipt of a `Handshake` message,
+/// a StackerDB-aware node replies with a `StackerDBHandshakeAccept` message which encodes both the
+/// contents of a `HandshakeAccept` message as well as a list of local StackerDBs (identified by
+/// their smart contracts' addresses).  Upon receipt of a `StackerDBHandshakeAccept`, the node
+/// stores the list of smart contracts in its `PeerDB` as part of the network frontier state.  In
+/// doing so, nodes eventually learn of all of the StackerDBs replicated by all other nodes.  To
+/// bound the size of this state, the protocol mandates that a node can only replicate up to 256
+/// StackerDBs.
```
kantai, line 78: Can you add a reference to where the handshake is implemented (i.e., `net::chat::handle_handshake`)?

src/net/stackerdb/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+/// The smart contract to which a StackerDB is bound controls how many chunks the DB has, who can
+/// write to which chunks (identified by public key hash), how big a chunk is, and how often a
+/// chunk can be written to (in wall-clock time).  This smart contract is queried once per reward cycle
+/// in order to configure the database.  The act of configuring the re-configuring the database
+/// is also the act of dropping and reinstantiating it.
+///
+/// Applications that employ StackerDBs would deploy one or more smart contracts that list out
+/// which users can store data to the StackerDB replica, and how much space they get.
+///
+/// ## Replication Protocol
```
kantai, line 67: This section doesn't mention `stackerdb_push` messages. What are those?

src/net/connection.rs
```diff
@@ -361,6 +361,7 @@ pub struct ConnectionOptions {
     pub max_block_push_bandwidth: u64,
     pub max_microblocks_push_bandwidth: u64,
     pub max_transaction_push_bandwidth: u64,
+    pub max_stackerdb_push_bandwidth: u64,
```
kantai, line 364: Can you add a rustdoc here?

src/net/chat.rs
```diff
@@ -2808,6 +2984,287 @@ mod test {
         })
     }
 
+    /// Inner function for testing various kinds of stackerdb/non-stackerdb peer interactions
+    fn inner_convo_handshake_accept_stackerdb(
+        peer_1_services: u16,
+        peer_1_rc_consensus_hash: ConsensusHash,
+        peer_2_services: u16,
+        peer_2_rc_consensus_hash: ConsensusHash,
+    ) {
```
kantai, line 2993: Should this go in the `net::stackerdb::tests` module instead?

src/net/chat.rs
```diff
@@ -1792,6 +1883,42 @@ impl ConversationP2P {
...
+        &mut self,
+        local_peer: &LocalPeer,
+        chain_view: &BurnchainView,
+        preamble: &Preamble,
+        relayers: Vec<RelayData>,
+    ) -> Result<Option<ReplyHandleP2P>, net_error> {
+        assert!(preamble.payload_len > 1); // don't count 1-byte type prefix
+
+        if !self.process_relayers(local_peer, preamble, &relayers) {
+            debug!(
```
kantai, line 1899: Should this be a `warn!`? Would seeing this indicate something is wrong?

src/net/chat.rs
```diff
@@ -1792,6 +1883,42 @@ impl ConversationP2P {
         Ok(None)
     }
 
+    /// Validate a pushed stackerdb chunk
+    /// Update bandwidth accounting, but forward the stackerdb chunk along.
+    /// Possibly return a reply handle for a NACK if we throttle the remote sender
```
kantai, line 1888: How is a pushed chunk validated? I think the rustdoc should have just a simple explanation of what is being checked here.

Also, what does forward the "stackerdb chunk along" mean? Is this method not only validating, but also forwarding the message to other peers?

kantai (approved):
This looks good to me -- I just had some feedback about comments, and then about the `ContractId` wrapper struct (I think should be replaced with an extension trait).

src/net/codec.rs
```diff
@@ -749,6 +752,131 @@ impl StacksMessageCodec for MemPoolSyncData {
...
+    }
+
+    fn consensus_deserialize<R: Read>(fd: &mut R) -> Result<ContractId, codec_error> {
+        let version: u8 = read_next(fd)?;
+        let bytes: [u8; 20] = read_next(fd)?;
+        let name: ContractName = read_next(fd)?;
+        let qn = QualifiedContractIdentifier::new(StandardPrincipalData(version, bytes), name);
+        Ok(ContractId(qn))
+    }
+}
```
kantai, line 770: When you use an extension trait, `impl StacksMessageCodec for QualifiedContractId` would have to be implemented in the `clarity` crate.

src/net/chat.rs
jcnelson: It's getting removed altogether in the follow-on #3552 PR.

src/net/chat.rs
jcnelson: I don't think so.  This is testing the chat behavior for handling handshakes, which is already covered here.

src/net/stackerdb/mod.rs
jcnelson: It's not implemented in this PR.

src/net/chat.rs
jcnelson: This is only verifying that the message is well-formed, as happens in all the other similarly-named methods for transactions, blocks, and microblocks.  The "forward along" comment refers to forwarding it back up to the reply handle that the (in this case) stackerdb state machine is using to retrieve the message.  Contrast this to messages that don't need to be forwarded, because no one is expecting them.

src/burnchains/mod.rs
jcnelson: Fixed

src/net/chat.rs
jcnelson: Fixed

src/net/chat.rs
jcnelson: Fixed

src/net/chat.rs
jcnelson: Fixed

src/net/chat.rs
jcnelson: Fixed

src/net/mod.rs
jcnelson: Fixed

src/net/stackerdb/mod.rs
jcnelson: Fixed

src/net/connection.rs
jcnelson: Fixed

src/net/chat.rs
jcnelson: Fixed

src/net/codec.rs
jcnelson: Instead, I just add special serialization/deserialization helpers before I saw this comment.  Hope that works!

jcnelson:
@kantai all of your comments have been addressed.  Please let me know what you think :pray:

kantai (approved):
LGTM

src/net/chat.rs
```diff
@@ -1801,6 +1933,42 @@ impl ConversationP2P {
...
+    fn validate_stackerdb_push(
+        &mut self,
+        local_peer: &LocalPeer,
+        chain_view: &BurnchainView,
+        preamble: &Preamble,
+        relayers: Vec<RelayData>,
+    ) -> Result<Option<ReplyHandleP2P>, net_error> {
+        assert!(preamble.payload_len > 1); // don't count 1-byte type prefix
+
+        if !self.process_relayers(local_peer, preamble, &relayers) {
```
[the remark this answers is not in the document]
jcnelson: The `process_relayers()` function is tested in the test `convo_process_relayers()` already.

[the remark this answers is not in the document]
jcnelson: Would that be sufficient for merge?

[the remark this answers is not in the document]
kantai: That's fine for `process_relayers()`, but there's many error cases in this changeset:

```
        if self.connection.options.max_stackerdb_push_bandwidth > 0
            && self.stats.get_stackerdb_push_bandwidth()
                > (self.connection.options.max_stackerdb_push_bandwidth as f64)
```

```
            StacksMessageType::StackerDBHandshakeAccept(ref data, ref db_data) => {
                if solicited {
                    test_debug!("{:?}: Got unauthenticated StackerDBHandshakeAccept", &self);
                    self.handle_handshake_accept(burnchain_view, &msg.preamble, data, Some(db_data))
                        .and_then(|_| Ok(None))
                } else {
                    test_debug!(
                        "{:?}: Unsolicited unauthenticated StackerDBHandshakeAccept",
                        &self
                    );

                    // don't update stats or state, and don't pass back
                    consume = true;
                    Ok(None)
                }
```

```
 match self.validate_stackerdb_push(
                    local_peer,
                    chain_view,
                    &msg.preamble,
                    msg.relayers.clone(),
                )? {
                    Some(handle) => Ok(handle),
                    None => {
                        // will forward upstream
                        return Ok(Some(msg));
                    }
                }
```

And for this branch, are both conditions tested (i.e., node supports "stacker dbs", but not the same stacker db as the handshake initializer?)

```
        let stacks_message = if ConversationP2P::supports_stackerdb(local_peer.services)
            && ConversationP2P::supports_stackerdb(self.peer_services)
        {
```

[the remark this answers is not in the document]
jcnelson: Sure; added a bunch of missing test coverage

kantai:
Okay, with the latest tests, this looks good to me. Before merging, be sure that you want this in `develop` as opposed to `next` (as in, this networking logic should ship in the next minor version release: `2.4.0.1.0`). You should also update the `CHANGELOG.md` (also please note if this is a minor or major version release).

PR #3558, opened 2023-02-08
Feat/stackerdb chunk db

jcnelson:
This implements the Stacker DB chunk storage database, as part of breaking apart #3534.


src/net/stackerdb/mod.rs
```diff
@@ -131,3 +139,66 @@ pub struct StackerDBSyncResult {
...
+/// A transaction against the Stacker DB
+pub struct StackerDBTx<'a> {
+    sql_tx: DBTx<'a>,
+    config: StackerDBConfig,
+}
+
+/// Chunk metadata from the DB.
+/// This is derived state from a StackerDBChunkData message.
+#[derive(Clone, Debug, PartialEq)]
+pub struct ChunkMetadata {
```
kantai, line 193: This struct and its fields should be referenced in the `StackerDB` rustdocs above -- where you currently provide an overview of what the StackerDB is and how it works, it should contain references to where the different items are implemented.

src/net/stackerdb/db.rs
```diff
@@ -0,0 +1,602 @@
...
+pub fn stackerdb_table(addr: &ContractId) -> String {
+    let normalized_name = format!("{}", &addr)
+        .replace("-", "\\x2d")
+        .replace("_", "\\x5f")
+        .replace(".", "_");
+
+    format!("stackerdb_{}", normalized_name)
+}
+
+/// Load up chunk metadata from the database, given the primary key.
```
kantai, line 171:
```suggestion
/// Load a chunk's metadata from the database, keyed by the chunk's smart contract, the reward cycle consensus hash, and the chunk identifier.
```

Looking at this PR after #3551 and #3552, I think there's a naming conflict with "StackerDB" -- you refer to the local database as "the `StackerDB`", but also, throughout the code, different smart contracts' associated chunk storage is each called "a `StackerDB`". This is already kind of confusing to me, and this is pretty early on in the life of this code, so I think we'd probably be better served with a different way to talk about these things.

src/net/stackerdb/db.rs
```diff
@@ -0,0 +1,602 @@
...
+
+    /// List all stacker DB smart contracts we have available
+    pub fn get_stackerdbs(&self) -> Result<Vec<ContractId>, net_error> {
+        let sql = "SELECT smart_contract_id FROM databases ORDER BY table_name";
+        query_rows(&self.conn(), sql, NO_PARAMS).map_err(|e| e.into())
+    }
+
+    /// Set up a database's storage slots.
+    /// The slots must be in a deterministic order, since they are used to determine the chunk ID
+    /// (and thus the key used to authenticate them)
```
kantai, line 272: What is a storage slot? It seems like this method is just writing empty data into the table. Why is this necessary?

src/net/stackerdb/db.rs
```diff
@@ -0,0 +1,602 @@
...
+        smart_contract: &ContractId,
+        rc_consensus_hash: &ConsensusHash,
+        chunk_id: u32,
+    ) -> Result<Option<ChunkValidation>, net_error> {
+        inner_get_chunk_validation(self.conn(), smart_contract, rc_consensus_hash, chunk_id)
+    }
+
+    /// Insert a chunk into the DB.
+    /// It must be authenticated, and its lamport clock must be higher than the one that's already
+    /// there.  These will not be checked.
```
kantai, line 347:
```suggestion
    /// The caller must check that this is authenticated and that its lamport clock is higher than the existing one. These are not checked by this method.
```

kantai (changes requested):
I have a couple of comments and requests that should be addressed:

1. Dynamic table names are pretty close to a sql anti-pattern and can lead to issues down the road (as well as requiring contract name sanitization), and so I think unless there's a very compelling reason to do this, it should be avoided.
2. `StackerDB` refers to two different things in the codebase (and the docs) and I think it's making the codebase confusing (sometimes "stackerdb" is a contract identifier, other times its the database storing multiple "stackerdb"s), and there's no better time than now to try to fix this (if it isn't fixed now, the codebase will probably have to live with this confusion forever)

src/net/stackerdb/bits.rs
```diff
@@ -0,0 +1,123 @@
...
+impl ChunkMetadata {
+    /// Get the digest to sign that authenticates this chunk data and metadata
+    fn sighash(&self) -> Sha512Trunc256Sum {
+        let mut bytes = vec![];
+        bytes.extend_from_slice(&self.rc_consensus_hash.0);
+        bytes.extend_from_slice(&self.chunk_id.to_be_bytes());
+        bytes.extend_from_slice(&self.chunk_version.to_be_bytes());
+        bytes.extend_from_slice(&self.data_hash.0);
+
+        Sha512Trunc256Sum::from_data(&bytes)
```
kantai, line 44: Avoid a copy:

```suggestion
        let mut hasher = Sha512_256::new();
        hasher.update(&self.rc_consensus_hash.0);
        hasher.update(&self.chunk_id.to_be_bytes());
        hasher.update(&self.chunk_version.to_be_bytes());
        hasher.update(&self.data_hash.0);

        Sha512Trunc256Sum::from_hasher(hasher)
```

src/net/stackerdb/bits.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /// Sign this chunk metadata, committing to rc_consensus_hash, chunk_id, chunk_version, and
+    /// data_hash.
```
kantai, line 48: Can you add a note to the rustdoc that the signature will be stored in `self.signature`?

src/net/stackerdb/bits.rs
```diff
@@ -0,0 +1,123 @@
...
+            chunk_id: self.chunk_id,
+            chunk_version: self.chunk_version,
+            data_hash: self.data_hash(),
+            signature: self.sig.clone(),
+        }
+    }
+
+    /// Sign this given chunk data message with the given private key.  Binds it to a particular
+    /// reward cycle in a particular PoX history (identified by `rc_consensus_hash`).
+    /// Fails if the underlying signing library fails.
```
kantai, line 101: Can you add a note to the rustdoc that the signature will be stored in `self.signature`?

src/net/stackerdb/tests/db.rs
```diff
@@ -0,0 +1,580 @@
...
+
+use stacks_common::address::{
+    AddressHashMode, C32_ADDRESS_VERSION_MAINNET_MULTISIG, C32_ADDRESS_VERSION_MAINNET_SINGLESIG,
+};
+use stacks_common::types::chainstate::{StacksPrivateKey, StacksPublicKey};
+
+/// Test that we can instantiate a stacker DB instance
+#[test]
+fn test_stackerdb_connect() {
+    let path = "/tmp/test_stackerdb_connect.sqlite";
```
kantai, line 43: Can you prefix these paths with `/tmp/stacks-node-tests/stackerdb/`? It makes cleaning up `/tmp` much easier after running the test suite.

kantai (commented):
Just a few more comments over my last review, which I think is still valid -- mostly, I think that programmatic table construction should be avoided unless there is a really compelling reason.

src/net/stackerdb/tests/db.rs
jcnelson: Sure

src/net/stackerdb/bits.rs
jcnelson: Yup

src/net/stackerdb/bits.rs
jcnelson: Yup

src/net/stackerdb/bits.rs
jcnelson: Yup, will do

src/net/stackerdb/bits.rs
```diff
@@ -0,0 +1,123 @@
...
+
+use crate::net::stackerdb::ChunkMetadata;
+
+use crate::chainstate::stacks::StacksPublicKey;
+
+use stacks_common::util::secp256k1::MessageSignature;
+
+impl ChunkMetadata {
+    /// Get the digest to sign that authenticates this chunk data and metadata
+    fn sighash(&self) -> Sha512Trunc256Sum {
```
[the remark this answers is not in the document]
jcnelson: Yup

src/net/stackerdb/db.rs
```diff
@@ -0,0 +1,602 @@
...
+
+/// Convert a smart contract name to a table name
+pub fn stackerdb_table(addr: &ContractId) -> String {
+    let normalized_name = format!("{}", &addr)
+        .replace("-", "\\x2d")
+        .replace("_", "\\x5f")
+        .replace(".", "_");
+
+    format!("stackerdb_{}", normalized_name)
+}
```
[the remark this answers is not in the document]
jcnelson: I removed the one-table-per-StackerDB code and now just store chunks in a single table.

src/net/stackerdb/bits.rs
```diff
@@ -0,0 +1,123 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+/// This module contains methods for interacting with the data contained within the messages
+use crate::net::{
+    Error as net_error, StackerDBChunkData, StackerDBChunkInvData, StackerDBGetChunkData,
```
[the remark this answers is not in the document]
jcnelson: This and the other `StackerDB*` structs are new protocol messages, whose definitions IMHO belong with the other message definitions.

src/net/stackerdb/db.rs
jcnelson: It's done this way because the follow-on StackerDB state sync PR needs to be able to extract the version vector for all of the DB's slots.  But in order to do that, those slots either have to be instantiated, or we need some kind of sparse version vector calculation.  The latter requires one SQL query per slot (since if slot data does not exist, we'd need to synthesize it on the fly), whereas this only requires a single query which can use an index.

src/net/stackerdb/db.rs
jcnelson: I'm having a hard time understanding what the ask here is.  The "smart contracts' associated chunk storage" _is_ a stacker DB.  The node maintains a local replica of that state for each stacker DB it subscribes to.  But, I don't see the value in calling the local replica by a different name, since I think that's already understood simply because a single process cannot be a whole p2p network.

jcnelson: I think you're referring to the fact that `StackerDB` the struct currently stores all stacker DB chunks for all subscribed stacker DBs?  If so, then I'll rename `StackerDB` to `StackerDBSet` to disambiguate it.

src/net/stackerdb/mod.rs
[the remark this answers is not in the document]
jcnelson: Done

obycode (approved):
LGTM!

jcnelson:
Hey @kantai, this is ready for re-review when you get the chance.  Thanks!

stackslib/src/net/mod.rs
```diff
@@ -256,6 +256,20 @@ pub enum Error {
     ExpectedEndOfStream,
     /// burnchain error
     BurnchainError(burnchain_error),
+    /// chunk is stale
+    StaleChunk(u32, u32),
```
kantai, line 260:
```suggestion
    /// chunk is stale
    StaleChunk { supplied_version: u32, latest_version: u32 },
```

stackslib/src/net/mod.rs
```diff
@@ -256,6 +256,20 @@ pub enum Error {
...
+    /// no such slot
+    NoSuchSlot(ContractId, u32),
+    /// no such DB
+    NoSuchStackerDB(ContractId),
+    /// stacker DB exists
+    StackerDBExists(ContractId),
+    /// slot signer is wrong
+    BadSlotSigner(StacksAddress, u32),
+    /// too many writes to a slot
+    TooManySlotWrites(u32, u32),
```
kantai, line 270:
```suggestion
    TooManySlotWrites { supplied_version: u32, max_writes: u32 },
```

stackslib/src/net/stackerdb/db.rs
```diff
@@ -0,0 +1,554 @@
...
+                                .expect(&format!("BUG: no parent of '{}'", path));
+                            fs::create_dir_all(&pparent_path).map_err(|e| db_error::IOError(e))?;
+
+                            OpenFlags::SQLITE_OPEN_READ_WRITE | OpenFlags::SQLITE_OPEN_CREATE
+                        } else {
+                            return Err(db_error::NoDBError.into());
+                        }
+                    } else {
+                        return Err(db_error::IOError(e).into());
+                    }
```
kantai, line 399:
```suggestion
                    if e.kind() != io::ErrorKind::NotFound {
                        return Err(db_error::IOError(e).into());
                    }
                    // path not found, must create db path.
                    if !readwrite {
                        return Err(db_error::NoDBError.into());
                    }
                    create_flag = true;
                    let ppath = Path::new(path);
                    let pparent_path = ppath
                                .parent()
                                .expect(&format!("BUG: no parent of '{}'", path));
                    fs::create_dir_all(&pparent_path).map_err(|e| db_error::IOError(e))?;

                    OpenFlags::SQLITE_OPEN_READ_WRITE | OpenFlags::SQLITE_OPEN_CREATE
```

stackslib/src/net/stackerdb/db.rs
```diff
@@ -0,0 +1,554 @@
...
+            .map_err(|e| net_error::DBError(e.into()))?;
+
+        let mut rows = stmt.query(args).map_err(|e| net_error::DBError(e.into()))?;
+
+        let mut data = None;
+        while let Some(row) = rows.next().map_err(|e| net_error::DBError(e.into()))? {
+            data = Some(row.get_unwrap(0));
+            break;
+        }
+        Ok(data)
```
kantai, line 538: You can get a single row result from a database, wrapped in an option by using `query_row().optional()`. That replaces lines 526-538 with just:

```suggestion
     self.conn.query_row(qry, args, |row| row.get(0)).optional().map_err(|e| e.into())
```

kantai (commented):
This looks good to me, my remaining feedback is pretty superficial.

jcnelson:
@kantai I addressed all your feedback in `feat/stackerdb-rpc` (by mistake -- was using the wrong tab).  Is that okay?  It'll find its way into this branch.

kantai (approved):

PR #3569, opened 2023-02-14
docs: normalize bitwise operations names

hugocaillard:
### Description

This PR removes the custom names in the API reference of bitwise operations.
The `name` property is used for other operators such as `+`, `<=`, `>`, etc, to add more details (eg: `name: Some("+ (add)")`) in the form `Some("<function-name> (<details)>")`.

It's used for bitwise functions in a more inconsistent way  `name: Some("Bitwise Xor")`:
- it doesn't contain the actual function name (eg `bit-xor`)
- other functions don't have capital letters in the names

By not specifying a custom name, it will default to the actual function name which is good enough for these functions (bit-and, bit-or, etc)

Benefits: the name is used to build to LSP completion items. The current names wouldn't be suggested with the input `(bit- )`

### Checklist
N/A


obycode (approved):
This change makes sense to me. Thanks @hugocaillard!

pavitthrap (approved):
lgtm

PR #3567, opened 2023-02-14
Stacks 2.1

jcnelson:
Stacks 2.1 stable release:  Our love letter to the Stacks community on Valentine's Day 2023.

PR #3472, opened 2023-01-08
docs: fix pox-2.clar typos

edu-stx:
### Description

Fixes small typos present in comments

### Checklist
- [x] Test coverage for new or modified code paths
- [ ] Changelog is updated
- [x] Required documentation changes (e.g., `docs/rpc/openapi.yaml` and `rpc-endpoints.md` for v2 endpoints, `event-dispatcher.md` for new events)
- [x] New clarity functions have corresponding PR in `clarity-benchmarking` repo
- [x] New integration test(s) added to `bitcoin-tests.yml`

I'm not sure how to update the changelog, do these changes need an updated changelog?

PR #3572, opened 2023-02-15
Feat/stacks 2.1 finishing touches

jcnelson:
This PR sets the block heights for testnet and mainnet at which 2.1 will activate:

* The mainline testnet will transition to 2.1 at height 2,422,101.
* The mainnet will transition to 2.1 at height 781,551.

This PR also merges #3472.

This unblocks #3567.


kantai (approved):

pavitthrap (approved):
lgtm

PR #3573, opened 2023-02-16
docs: fix principal -construct -destruct api reference

hugocaillard:
### Description

The api reference for `principal-construct` and `principal-destruct` was built with the wrong argument

---
Idea: An other option would be to pass `function` (and do it for every other `make_for_simple_native` calls here) to limit the risk of having the same issue in the future
```rust
PrincipalDestruct => make_for_simple_native(&PRINCPIPAL_DESTRUCT_API, &function, name),
PrincipalConstruct => make_for_simple_native(&PRINCIPAL_CONSTRUCT_API, &function, name),
```

I wanted to limit the footprint of my PR but let me know if I should perform this change 

### Applicable issues
N/A

### Additional info (benefits, drawbacks, caveats

Improve the LSP.
The output type is displayed in VSCode (and other editors) on hover 

### Checklist
N/A

### DIFF
--- a/clarity/src/vm/docs/mod.rs
+++ b/clarity/src/vm/docs/mod.rs
@@ -1767,7 +1767,7 @@ The `addrs` list contains the same PoX address values passed into the PoX smart 
 };
 
 const PRINCIPAL_CONSTRUCT_API: SpecialAPI = SpecialAPI {
-    input_type: "(buff 1) (buff 20) [(string-ascii 40)]",
+    input_type: "(buff 1), (buff 20), [(string-ascii 40)]",
     output_type: "(response principal { error_code: uint, principal: (option principal) })",
     snippet: "principal-construct? ${1:version} ${2:pub-key-hash}",
     signature: "(principal-construct? (buff 1) (buff 20) [(string-ascii 40)])",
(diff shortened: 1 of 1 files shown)


pavitthrap (approved):
this looks good! I think the change you proposed sounds good & would be appropriate to include in this pr

hugocaillard:
Alright I will 👌

obycode (approved):
Good catch!

jcnelson:
Your doc tests are failing

clarity/src/vm/docs/mod.rs
```diff
@@ -1818,6 +1766,59 @@ The `addrs` list contains the same PoX address values passed into the PoX smart
 "
 };
 
+const PRINCIPAL_CONSTRUCT_API: SpecialAPI = SpecialAPI {
+    input_type: "(buff 1) (buff 20) [(string-ascii 40)]",
```
hugocaillard, line 1770: I should probably name it right?
I don't know how to name the first arg.
```
input_type : "(buff 1), principal, [ContractName]",
```
?

hugocaillard, line 1770:
```suggestion
    input_type: "(buff 1), (buff 20), [(string-ascii 40)]",
```

hugocaillard:
Thanks for warning me @jcnelson.

PS: when a re-requested a review from Brice, it removed you as a reviewer and now I don't have the rights to add you again

obycode (approved):
Great! Thanks for the updates.

pavitthrap (approved):

PR #3585, opened 2023-02-24
Resolve `cargo check` errors in benchmarks

### LINKED ISSUE #3584: Benchmarks are not up to date
**Describe the bug**
The `benches/` directory contains nonfunctional code, there's something like 50-60 errors that my LSP is detecting.

**Steps To Reproduce**
Run `rust-analyzer` directly or through an editor on the whole repo.

**Expected behavior**
Repo is free from LSP errors and warnings.

If we want to do something about this I see two options:

- bring the benchmarks up to date and fix all problems
- delete the directory

stjepangolemac:
### Description
Prior to this PR we had 50-60 errors in the codebase mainly situated in the `bench/` directory. They haven't been updated for a year or two. Deleting them doesn't seem like a good solution as we might want to use them in the future. Also, bringing them up to date seems like it wouldn't be a quick task.

For now, I've excluded them from the compilation by doing these two things:

1. Renamed benchmark files to `*.rs.backup`
2. Removed `[[bench]]` entries from `Cargo.toml`

Unrelated, but I've also removed two unused variables so we end up with no errors.

### Applicable issues
- fixes #3584


kantai:
I'd be in favor of deleting any that don't compile, though I know that may be controversial.

The git history will keep that "deleted" data, so I'm not worried about data loss, and these files are not documentation, so I'm also not worried about breaking any internal links.

stjepangolemac:
My first thought was to delete this but wanted to first propose a less aggressive approach. @igorsyl what do you think?

igorsyl:
Not at all. Like @kantai pointed out, git doesn't forget.

stjepangolemac:
Okay, I removed benchmark files and this can now be reviewed.

kantai (approved):

jcnelson:
Consensus is that it's fine to nuke this code.

igorsyl (approved):
image

stjepangolemac:
I'm not authorized to merge this one, @igorsyl who usually does that?

igorsyl:
> I'm not authorized to merge this one, @igorsyl who usually does that?

It's got two review approvals and Jude expressed consensus support. I'm merging to unblock.
cc @wileyj for discussion later on how to handle merge access for collaborators.

PR #3586, opened 2023-02-24
[eng quality] pre-commit hook docs

### LINKED ISSUE #3570: [eng quality] implement git hook for rustfmt
Create a pre-commit hook to check the formatting of the Rust files. 
A description of how to do so can be found here: (https://eugene-babichenko.github.io/blog/2018/11/08/rustfmt-git-hook/)

Since the `.git/hooks` is not in version control, the hook should be stored in a different folder, like `.githooks`, so that all developers have the same hook. (https://www.atlassian.com/git/tutorials/git-hooks) 

This issue is especially important because of the new check for import order in the CI, implemented in #3556

pavitthrap:
### Description
Adds documentation explaining how to locally setup a pre-commit hook

### Applicable issues
- fixes #3570

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -12,6 +12,7 @@ You can find information on joining online community forums (Discord, mailing li
 [How Can I Contribute?](#how-can-i-contribute)
 * [Development Workflow](#development-workflow)
 * [Contributing Conventions](#contributing-conventions)
+* [Developer Setup](#recommended-developer-setup)
 
 [Style](#style)
 * [Git Commit Messages](#git-commit-messages)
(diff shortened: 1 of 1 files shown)


fpbgg (approved):
I followed these instructions then tried to commit an invalid change (adding whitespace to the imports).

```terminal
% git commit
rustfmt failed: run "cargo fmt --all -- --config group_imports=StdExternalCrate"
```

I ran the suggested command. I was prompted to provide a commit message as expected.

donpdonp (approved):
Approved - as is looks good to me but I'm leaving this suggested change:

The script block below is easy to read, keeps the complicated cargo fmt command in one place for future edits, and actually formats the code if the --check fails but does not check it in. 

The user is alerted that the code has been formatted and re-running the git commit -a (usually an up arrow from the terminal) will include the format changes. Its also appendable to pre-commit (only exits on failure) for when users might have an existing pre-commit hook.

```
2. Append the following:
```

```
CARGO_FMT="cargo fmt --all -- --config group_imports=StdExternalCrate"

$CARGO_FMT --check
CARGO_FMT_RESULT=$?

if [ $CARGO_FMT_RESULT -eq 1 ]
then
    $CARGO_FMT
    echo 'cargo fmt --all; applied, commit again.'
    exit $CARGO_FMT_RESULT
fi
```

pavitthrap:
> Approved - as is looks good to me but I'm leaving this suggested change:
> 
> The script block below is easy to read, keeps the complicated cargo fmt command in one place for future edits, and actually formats the code if the --check fails but does not check it in.
> 
> The user is alerted that the code has been formatted and re-running the git commit -a (usually an up arrow from the terminal) will include the format changes. Its also appendable to pre-commit (only exits on failure) for when users might have an existing pre-commit hook.
> 
> ```
> 2. Append the following:
> ```
> 
>  ```
> CARGO_FMT="cargo fmt --all -- --config group_imports=StdExternalCrate"
> 
> $CARGO_FMT --check
> CARGO_FMT_RESULT=$?
> 
> if [ $CARGO_FMT_RESULT -eq 1 ]
> then
>     $CARGO_FMT
>     echo 'cargo fmt --all; applied, commit again.'
>     exit $CARGO_FMT_RESULT
> fi
> ```

Discussed with @donpdonp and agreed there are advantages to both versions (for example, the one I proposed only runs the check command for staged files).

PR #3587, opened 2023-02-24
Fix: Transaction receipts for runtime errors in contract publish

### LINKED ISSUE #3154: Bad tx receipts when contract deployment top-level code fails
**Describe the bug**
A contract is deployed successfully, even though the top-level code in the contract contains errors.

**Steps To Reproduce**
1. Deploy a contract with a function that can fails
2. Deploy a contract that calls the first contract's function as top-level code so that the first contract fails
3. Deploy another contract that calls the first contract's function as top-level code so that the first contract fails but with the contract call wrapped in `try!`
4. Deploy another contract that calls the first contract's function as top-level code so that the first contract fails but with the contract call wrapped in `unwrap-panic`

Successful contract deployments with contract calls that fail and succeed and with wrapping in `try!` and `unwrap-panic`:
* https://explorer.stacks.co/txid/0x3ebf410f601e0a837791e1c5119a7921c5971af5be0499328da2454445f2b6db?chain=testnet
* https://explorer.stacks.co/txid/0xfd452e1507bcd5303e662e53f68698f64fb9f2e83471cf24c794e31ec0b2d611?chain=testnet
* https://explorer.stacks.co/txid/0x60cc62cfd6afb1d364efde72f254275e610d0de977142f560dbd0a3efa4954cd?chain=testnet
* https://explorer.stacks.co/txid/0x54483249f857918d59a086ba6d632142b3eb53b80e4830b252ce06fa0926f9e3?chain=testnet
* https://explorer.stacks.co/txid/0xfc88243a3ac8f97ae914e412bec91a50ce2a9276d8dd712662fdae7112377aaa?chain=testnet
* https://explorer.stacks.co/txid/0xd03884c0305cfe18cbd8ba910ad9ff49c024ec638fee127924d5ca3bb7dcbc6d?chain=testnet
* https://explorer.stacks.co/txid/0x9e7efe23e27b3c587b70550cd8e18956e44af00cf6dd91f59ca28927f5a1bc97?chain=testnet

**Expected behavior**
A contract deploy transaction should only succeed if and only if the code is correct and all top-level statements do not throw errors and do not return err responses.

**Additional context**
This was discovered during prize distribution of the 90stx raffle, e.g. https://explorer.stacks.co/txid/0x5421e0963f29b4409718b2f864262aee559f00ccc8edc302bf274d9cea98c320?chain=mainnet

kantai:
### Description

When top-level code in a contract publish causes a runtime error, the transactions should be accepted, but the contracts should not be created, and that is the existing behavior of the stacks blockchain.

However, the transaction receipts for these transactions do not indicate that a failure occurred. This leads to pretty confusing API and explorer results. This PR fixes that by returning a receipt with result set to `(err none)`

### Applicable issues
- fixes #3154 

### Checklist
- [x] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
+## [Unreleased]
+
+### Fixed
+
+- The transaction receipts for smart contract publish transactions now indicate
+  a result of `(err none)` if the top-level code of the smart contract contained
+  runtime error. Fixes issue 3154.
+
 ## [2.1]
 
 This is a **consensus-breaking** release that introduces a _lot_ of new
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to the versioning scheme outlined in the [README.md](RE
 
 - The transaction receipts for smart contract publish transactions now indicate
   a result of `(err none)` if the top-level code of the smart contract contained
-  runtime error. Fixes issue 3154.
+  runtime error. Fixes issue #3154.
 
 ## [2.1]
 
--- a/src/chainstate/stacks/db/transactions.rs
+++ b/src/chainstate/stacks/db/transactions.rs
@@ -1176,7 +1176,18 @@ impl StacksChainState {
                                       "contract" => %contract_id,
                                       "code" => %contract_code_str,
                                       "error" => ?error);
-                            (AssetMap::new(), vec![])
+                            let receipt = StacksTransactionReceipt {
+                                transaction: tx.clone().into(),
+                                events: vec![],
+                                post_condition_aborted: false,
+                                result: Value::err_none(),
+                                stx_burned: 0,
+                                contract_analysis: Some(contract_analysis),
+                                execution_cost: total_cost,
+                                microblock_header: None,
+                                tx_index: 0,
+                            };
+                            return Ok(receipt);
                         }
                         ClarityRuntimeTxError::AbortedByCallback(_, assets, events) => {
                             let receipt =
--- a/src/chainstate/stacks/db/transactions.rs
+++ b/src/chainstate/stacks/db/transactions.rs
@@ -1176,6 +1176,10 @@ impl StacksChainState {
                                       "contract" => %contract_id,
                                       "code" => %contract_code_str,
                                       "error" => ?error);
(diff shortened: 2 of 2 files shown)


src/chainstate/stacks/db/transactions.rs
```diff
@@ -1176,7 +1176,23 @@ impl StacksChainState {
...
+                                transaction: tx.clone().into(),
+                                events: vec![],
+                                post_condition_aborted: false,
+                                result: Value::err_none(),
+                                stx_burned: 0,
+                                contract_analysis: Some(contract_analysis),
+                                execution_cost: total_cost,
+                                microblock_header: None,
+                                tx_index: 0,
+                                vm_error: None,
```
obycode, line 1193: Should we record the error here?
```suggestion
                                vm_error: Some(error.to_string()),,
```

zone117x:
To clarify, the current behavior in this condition is `result: (ok true)`, and with this PR will be changed to `result: (err none)`, right?

E.g. current behavior:
```shell
$ curl -s "https://stacks-node-api.testnet.stacks.co/extended/v1/tx/0x3ebf410f601e0a837791e1c5119a7921c5971af5be0499328da2454445f2b6db" \
| jq .tx_result.repr
"(ok true)"
```

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
+## [Unreleased]
+
+### Fixed
+
+- The transaction receipts for smart contract publish transactions now indicate
+  a result of `(err none)` if the top-level code of the smart contract contained
+  runtime error. Fixes issue 3154.
```
jcnelson, line 14: s/3154/#3154 so we can see it from github markdown.

jcnelson (approved):

src/chainstate/stacks/db/transactions.rs
kantai: Good idea -- added and tested in d97111393

kantai:
> To clarify, the current behavior in this condition is `result: (ok true)`, and with this PR will be changed to `result: (err none)`, right?
> 
> E.g. current behavior:
> 
> ```shell
> $ curl -s "https://stacks-node-api.testnet.stacks.co/extended/v1/tx/0x3ebf410f601e0a837791e1c5119a7921c5971af5be0499328da2454445f2b6db" \
> | jq .tx_result.repr
> "(ok true)"
> ```

Yep, that's correct.

obycode (approved):
lgtm!

PR #3600, opened 2023-03-02
Feat/expr-comments

obycode:
### Description

Comments are only enabled with feature, `developer-mode`. This will be
useful for tools that want to manipulate Clarity source code, like a
formatter for example.

### Checklist
- [x] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to the versioning scheme outlined in the [README.md](RE
 
 ## [Unreleased]
 
+### Added
+
+- When the Clarity library is built with feature flag `developer-mode`, comments
+  from the source code are now attached to the `SymbolicExpression` nodes. This
+  will be useful for tools that use the Clarity library to analyze and
+  manipulate Clarity source code, e.g. a formatter.
+
 ### Fixed
 
 - The transaction receipts for smart contract publish transactions now indicate
--- a/clarity/src/vm/ast/sugar_expander/mod.rs
+++ b/clarity/src/vm/ast/sugar_expander/mod.rs
@@ -663,7 +663,7 @@ mod test {
             make_pre_atom("bar", 4, 1, 4, 3),
         ];
         let foo = make_atom("foo", 1, 1, 1, 3);
-        let mut bar = make_atom("bar", 3, 1, 3, 3);
+        let mut bar = make_atom("bar", 4, 1, 4, 3);
         bar.pre_comments = vec![
             (
                 "this is a comment".to_string(),
@@ -710,7 +710,7 @@ mod test {
             make_pre_atom("bar", 5, 1, 5, 3),
         ];
         let foo = make_atom("foo", 1, 1, 1, 3);
-        let mut bar = make_atom("bar", 3, 1, 3, 3);
+        let mut bar = make_atom("bar", 5, 1, 5, 3);
         bar.pre_comments = vec![
             (
                 "this is a comment".to_string(),
--- a/clarity/src/vm/ast/sugar_expander/mod.rs
+++ b/clarity/src/vm/ast/sugar_expander/mod.rs
@@ -579,7 +579,7 @@ mod test {
         expander.run(&mut contract_ast).unwrap();
         assert_eq!(
             contract_ast.expressions, ast,
-            "`foo` should have the post-comment attached"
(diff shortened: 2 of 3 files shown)


jcnelson:
Hey! This needs to be opened against `develop` at the very least.

obycode:
Oops. Yup, I mean to open against develop. Sorry about that.

clarity/src/vm/representations.rs
```diff
@@ -427,15 +427,25 @@ pub struct SymbolicExpression {
...
     #[cfg(feature = "developer-mode")]
     fn cons() -> SymbolicExpression {
         SymbolicExpression {
             id: 0,
-            span: Span::zero(),
             expr: SymbolicExpressionType::AtomValue(Value::Bool(false)),
+            span: Span::zero(),
+            pre_comments: vec![],
+            end_line_comment: None,
+            post_comments: vec![],
```
kantai, line 448: So, I'm _pretty_ sure that altering the `SymbolicExpression` serialization format won't impact a block's consensus hash (because the AST is stored as metadata), but it would be good to have some kind of test fixture-based of this.

There's a couple ways we could go about structuring a test like this, but I think the easiest way to do this would be to add a coordinator-style test. See, for example, the test I add in my atlas PR:

 https://github.com/stacks-network/stacks-blockchain/blob/9b97a39191133a53cfdcb07cccceebbab99246f7/src/chainstate/coordinator/tests.rs#L4312-L4322)

You could run the test first without your changes and get a value for the index block hash. Then, you could use that value as a hardcoded string and add it as an assertion to the test (that way, the test will assert that the commitment hash doesn't change with these changes).

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to the versioning scheme outlined in the [README.md](RE
 
 ## [Unreleased]
 
+### Added
+
+- When the Clarity library is built with feature flag `developer-mode`, comments
+  from the source code are now attached to the `SymbolicExpression` nodes. This
+  will be useful for tools that use the Clarity library to analyze and
+  manipulate Clarity source code, e.g. a formatter.
```
kantai, line 15: As implemented, I believe these changes are DB breaking changes (so it would require a resync from genesis), because the AST nodes will fail to parse database entries that existed before this patch. I think, if possible, we should avoid this, but if we can't, this needs to be stated in the CHANGELOG

obycode:
Thanks for the review, @kantai! In response to both of those comments, I believe that the feature flag means that nothing should change in the clarity library used for a stacks node, since the `developer-mode` feature is not enabled. This follows the logic previously used by the `span` field in `SymbolicExpression`. Do we need to be compatible when that feature is enabled too?

clarity/src/vm/ast/sugar_expander/mod.rs
```diff
@@ -509,4 +556,309 @@ mod test {
...
+            "S1G2081040G2081040G2081040G208105NK8PE5.test-comments",
+        )
+        .unwrap();
+        let mut contract_ast = ContractAST::new(contract_id.clone(), pre_ast);
+
+        let expander = SugarExpander::new(contract_id.issuer);
+        expander.run(&mut contract_ast).unwrap();
+        assert_eq!(
+            contract_ast.expressions, ast,
+            "`foo` should have the post-comment attached"
```
pavitthrap, line 582: should be "end line comment" here, not "post-comment"

pavitthrap (approved):
lgtm - but unclear on what kind of behavior is acceptable for developer-mode, so that requires further discussion.

clarity/src/vm/ast/sugar_expander/mod.rs
obycode: Thanks! Fixed in 1495456ee.

kantai:
> Thanks for the review, @kantai! In response to both of those comments, I believe that the feature flag means that nothing should change in the clarity library used for a stacks node, since the `developer-mode` feature is not enabled. This follows the logic previously used by the `span` field in `SymbolicExpression`. Do we need to be compatible when that feature is enabled too?

Ah, yes, you're correct!

That's fine then.

kantai (approved):

obycode:
Ok, I just fixed the line numbers that were off in two of my new unit tests. Once they pass, this should be good to merge.

PR #2707, opened 2021-06-14
Fix/1805

jcnelson:
At long last, this PR fixes #1805.

## The Problem

There are currently two related design flaws in the way the Stacks blockchain deals with PoX anchor blocks:

* If it is ever the case in which a PoX anchor block is missing, and yet somehow manages to achieve 80% or more confirmations during the prepare phase, then the subsequent arrival of that anchor block will cause a _deep_ chain reorg.  It doesn't matter how many future blocks get mined -- if the anchor block is later revealed, it will invalidate all of the blocks that did not build on it.  While mining and confirming an anchor block is very costly, it's not only possible, but profitable: anyone who manages to do this could hold the blockchain for ransom by threatening to disclose the anchor block and invaldiate all blocks after it unless they were paid not to (i.e. in perpetuity).

* If it is ever the case that not enough STX get locked for PoX to begin in reward cycle _R_, then a node that processes Stacks blocks first without the anchor block in _R_ and then with the anchor block in _R_ will crash because it will attempt to calculate the same sortition twice.  This is because the same block-commits would be processed in both cases -- they'd both be PoB commits.

This PR fixes both problems by making the _history of anchor blocks itself_ forkable, and by implementing _Nakamoto consensus_ on the anchor block history forks so that there will always be a canonical anchor block history.  In doing so, the Stacks blockchain now has _three_ levels of forks: the Bitcoin chain, the history of PoX anchor blocks, and the history of Stacks blocks.  The canonical Stacks fork is the longest history of Stacks blocks that passes through the canonical history of anchor blocks which resides on the canonical Bitcoin chain.

## Background: Sortition Histories

Recall that each Bitcoin block can contain block-commits that are valid only if certain anchor blocks are known to the node, and invalid if other anchor blocks are known.  Specifically, a block-commit can be a valid PoX block-commit _only if_ the current reward cycle has an anchor block, _and_ that anchor block is known to the node.  Otherwise, if the block-commit does not descend from the anchor block, or there is no anchor block for this reward cycle, then the block-commit can only be valid if it's a PoB block-commit.

What this means is that there is a _set_ of sortition histories on the Bitcoin chainstate that will each yield a unique history of block-commits (which in turn represent a unique set of possible Stacks forks).  This set has _O(2**n)_ members, where _n_ is the number of reward cycles that have anchor blocks.  This is because each time a new reward cycle is processed with an anchor block, there will be a sortition history that descends from it in which the anchor block is known to the node, and a sortition history in which it is _not_ known.

[...]

PR #3603, opened 2023-03-04
fix: always_use_affirmation_maps defaults to false

jcnelson:
Testnet nodes currently do not sync with the default config.  This is because at reward cycle 360, the 2.05 PoX anchor block rules differ from the 2.1 rules -- 2.05 believes that there is an anchor block, while 2.1 does not.

A while back, we added a config option `always_use_affirmation_maps` which forced the node to use the 2.1 rules all the time.  This was meant to allow us to stress-test #2707.  The value for this config option defaults to `true`, meaning that the node uses the 2.1 rules by default in epoch 2.05.  While this isn't a problem on mainnet, it prevents nodes from syncing on testnet due to the above.

This PR switches this config option's default value to `false`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to the versioning scheme outlined in the [README.md](RE
 ### Fixed
 
 - Handle the case where a bitcoin node returns zero headers (#3588)
+- The default value for `always_use_affirmation_maps` is now set to `false`,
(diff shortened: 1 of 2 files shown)


obycode (approved):
👍

igorsyl (approved):
I look forward to Hiro's Blockchain Fridays learning session on Affirmation Maps.

PR #3607, opened 2023-03-09
fix/ updated mempool sync protocol tests

pavitthrap:
### Description
When working on tests for the microblock tip sync protocol, I realized that `ibd` was set to true in the call to `TestPeer::step()`. Since `ibd=true`, the mempool (and microblock) sync protocol were effectively skipped over in the unit test. To avoid this problem, I am passing ibd as a parameter to `step_with_ibd`. 

The mempool sync tests are now fast (because the protocol now runs correctly), so I was also able to remove the timeout. I thus also removed the `#[ignore]` above the tests.


diwakergupta:
> The mempool sync tests are now fast (because the protocol now runs correctly), so I was also able to remove the timeout. I thus also removed the `#[ignore]` above the tests.

👏🏽 👏🏽 Thanks Pavi!!

jcnelson (approved):

kantai (approved):
LGTM!

PR #3609, opened 2023-03-13
Draft: re-attempt mining after blocking_signal flaps

kantai:
This PR is attempting to address an issue we're seeing in the Hiro testnet miner related to the miner thread blocking signal. In the current implementation, if a miner attempts to mine and sees the "block" signal, the "RunTenure" command is not reissued. The miner thread does not attempt to mine again, and this results in missed sortitions.

This is still a draft as we're still trying to diagnose the exact locking behavior (we're also seeing chains coordinator thread stalls with db contention, which is magnifying this issue, but it isn't clear year what's driving that).


kantai:
Closing, in lieu of #3610

PR #3612, opened 2023-03-14
chore: update atlas config

lgalabru:
In this PR, we're bumping `max_uninstantiated_attachments` to 50k and `uninstantiated_attachments_expire_after` to 1 day.

### DIFF
--- a/src/net/atlas/mod.rs
+++ b/src/net/atlas/mod.rs
@@ -71,8 +71,8 @@ impl AtlasConfig {
         AtlasConfig {
             contracts,
             attachments_max_size: 1_048_576,
-            max_uninstantiated_attachments: 10_000,
-            uninstantiated_attachments_expire_after: 3_600,
+            max_uninstantiated_attachments: 50_000,
+            uninstantiated_attachments_expire_after: 86_400,
             unresolved_attachment_instances_expire_after: 172_800,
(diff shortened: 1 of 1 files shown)


jcnelson (approved):
LGTM.  @CharlieC3 @deantchi when deploying, can you keep some nodes with the old setting, just in case this puts too much load on the p2p threads of the new nodes?

CharlieC3:
@jcnelson Are these settings configurable in the config.toml? It would be a big help if they are.
cc @lgalabru

lgalabru:
I agree that ideally, this settings would be exposed in the config.
Could someone working full-time on the blockchain team could follow up? 
It sounds like a good first issue (cc @igorsyl / @kantai  / @jcnelson for dispatch).

kantai:
Now that we have a configurable version of this PR (#3618), should we close this one?

jbencin:
> Now that we have a configurable version of this PR (#3618), should we close this one?

PR #3618 is still using the old default values. If the testing by @CharlieC3 and @zone117x show these values to be an improvement, we can merge this PR to update the defaults

pavitthrap:
Blockchain meeting: testing halfway complete, need to see how these values affect attachment performance

kantai (approved):

PR #3613, opened 2023-03-14
fix(deps): only depend on one version of secp256k1

### LINKED ISSUE #3608: Two versions of secp256k1 used
We're using two versions of the `secp256k1` dependency. I'm not sure why that is the case and if it's still needed.

#### blockstack-lib
https://github.com/stacks-network/stacks-blockchain/blob/c1a2c2e59e588fe57067c33e7aba7395cb3fa12a/Cargo.lock#L405

#### stacks-commons
https://github.com/stacks-network/stacks-blockchain/blob/c1a2c2e59e588fe57067c33e7aba7395cb3fa12a/Cargo.lock#L2652

This can be resolved by removing the explicit version selector:
https://github.com/stacks-network/stacks-blockchain/blob/c1a2c2e59e588fe57067c33e7aba7395cb3fa12a/stacks-common/Cargo.toml#L39-L41

- Who would know more about this?
- Why was this version selected in the first place?
- Can we remove the version selector?

I'll tag some people that might know something about this.

cc @sergey-shandar @kantai @zone117x

fpbgg:
Update the Cargo.toml / Cargo.lock files to avoid multiple versions from being included

### Applicable issues
- fixes #3608 

### Checklist
- [ ] Changelog is updated


Cargo.toml
```diff
@@ -71,7 +71,7 @@ version = "1.0"
 features = ["arbitrary_precision", "unbounded_depth"]
 
 [dependencies.secp256k1]
-version = "0.24.2"
+version = "0.24.3"
```
xoloki, line 74: Should we really specify a full rev, not just `0.24.`?

xoloki (approved):
LGTM

Cargo.toml
fpbgg: The only risk in doing so is a very minor nuisance: If someone runs `cargo update` in an unrelated PR, it will make their PR messy because the dependency moves.

Barring that, I'm happy to leave the patch version unpinned in our .toml files.

xoloki: > The only risk in doing so is a very minor nuisance: If someone runs `cargo update` in an unrelated PR, it will make their PR messy because the dependency moves.
> 
> Barring that, I'm happy to leave the patch version unpinned in our .toml files.

Probably safe to leave full rev for now, we dependabot should tell us if there's a problem.

jcnelson (approved):
Thanks @fpbgg!

obycode (approved):
👍

stjepangolemac (approved):

jferrant (approved):

PR #3615, opened 2023-03-14
docs: consistent markdown formatting

obycode:
The main goal of this commit is to resolve the issues where a newline was intended, but because Markdown formatting uses two newlines to separate paragraphs, these lines were being merged into one. This was especially apparent in the lists of possible error codes, which I've now reformatted to use bulleted lists. 

For example, the following string in the source:
```
"...
`(err u1)` -- `sender` does not have enough balance to transfer
`(err u2)` -- `sender` and `recipient` are the same principal
`(err u3)` -- amount to send is non-positive
"
```

Resulted in the this formatting on the docs site:
image


clarity/src/vm/docs/mod.rs
```diff
@@ -2324,12 +2331,12 @@ const STX_BURN: SimpleFunctionAPI = SimpleFunctionAPI {
...
 context's `tx-sender`.
 
-This function returns (ok true) if the transfer is successful. In the event of an unsuccessful transfer it returns
+This function returns (ok true) if the burn is successful. In the event of an unsuccessful burn it returns
 one of the following error codes:
 
-`(err u1)` -- `sender` does not have enough balance to transfer
-`(err u3)` -- amount to send is non-positive
-`(err u4)` -- the `sender` principal is not the current `tx-sender`
+* `(err u1)` -- `sender` does not have enough balance to transfer
```
pavitthrap, line 2337: enough balance to *burn*

clarity/src/vm/docs/mod.rs
```diff
@@ -2324,12 +2331,12 @@ const STX_BURN: SimpleFunctionAPI = SimpleFunctionAPI {
...
+* `(err u3)` -- amount to send is non-positive
```
pavitthrap, line 2338: amount to *burn*

pavitthrap (approved):

clarity/src/vm/docs/mod.rs
obycode: Thanks for catching these 👍

igorsyl:
Good first issue to review @soju-drinker. You should be getting contributor access later today! cc @wileyj

jcnelson (approved):

PR #3618, opened 2023-03-16
feat: Add Atlas configuration via Stacks.toml

jbencin:
### Description

The following parameters can be used in Stacks.toml to configure Atlas database. Example (using defaults):

```toml
[atlas]
attachments_max_size = 1048576
max_uninstantiated_attachments = 10000
uninstantiated_attachments_expire_after = 3600
unresolved_attachment_instances_expire_after = 172800
```

### Applicable issues
- fixes #

I considered using https://crates.io/crates/validator_derive to implement Atlas config validation, but I know this project has a preference for not adding new dependencies, so I ended up doing it manually. This package could be used in a lot of places, and would result in code that is cleaner and less error prone, so it's worth considering adding it

Made a couple minor code quality improvements also:
- Changed `AtlasConfig::default()` to `AtlasConfig::new()` since `default()` shouldn't be used with parameters
- Replaced `FeeEstimationConfigFile::default()` with `#[derive(Default)]`

### Checklist
- [ ] Changelog is updated


testnet/stacks-node/src/config.rs
```diff
@@ -1978,6 +1976,34 @@ pub struct MinerConfigFile {
...
+    pub max_uninstantiated_attachments: Option<u32>,
+    pub uninstantiated_attachments_expire_after: Option<u32>,
+    pub unresolved_attachment_instances_expire_after: Option<u32>,
+}
+
+impl AtlasConfigFile {
+    // Can't inplement `Into` trait because this takes a parameter
+    fn into_config(&self, mainnet: bool) -> AtlasConfig {
+        let mut conf = AtlasConfig::new(mainnet);
+        if let Some(val) = self.attachments_max_size {
```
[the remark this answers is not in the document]
jbencin: I'll add a constant `ATTACHMENTS_MAX_SIZE_MIN` and default to that if the user sets a lower value. Should we use the current defaults to enforce minimum values for the other parameters in the `[atlas]` config as well?

jbencin:
I have a few more ideas for Atlas improvements that I'd like to get some feedback on. I will probably make a new issue(s) for this work, but I'm going to start the conversation here, so I have some idea of what to do before creating it. 

- Add a `zonefile-size` parameter to `name-register` in the BNS contract. This could allow a fee to be charged based on size, which would reduce the opportunity for DoS attacks by flooding the network with large zonefiles
- Add a `fetch_mempool_attachments` parameter to the `[atlas]` config section. Currently, I am told, attachments are not replicated until a transaction is confirmed in a block, which creates a single point of failure until then. Even just one or two nodes with this parameter set would drastically reduce the possibility of attachments being lost.

I'd like some feedback from people more familiar with the Stacks ecosystem, and distributed systems in general, to let me know if these ideas are feasible, if my assumptions are correct, and if there are better solutions.

jcnelson:
> Add a zonefile-size parameter to name-register in the BNS contract. This could allow a fee to be charged based on size, which would reduce the opportunity for DoS attacks by flooding the network with large zonefiles

Sure; however, the BNS contract cannot be changed (generally goes for all contracts). You'll want to speak with the BNS X folks who are writing new wrappers around BNS to add new features.

> Add a fetch_mempool_attachments parameter to the [atlas] config section. Currently, I am told, attachments are not replicated until a transaction is confirmed in a block, which creates a single point of failure until then. Even just one or two nodes with this parameter set would drastically reduce the possibility of attachments being lost.

Attachments are not stored in the mempool.  They are only stored on the node that receives the transaction and attachment data from the client.  The client could post this data to a few different public nodes and achieve the same end, however.

kantai:
I think this should be opened against `develop`

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -515,7 +515,7 @@ impl RunLoop {
         let chain_state_db = self.boot_chainstate(burnchain_config);
 
         // NOTE: re-instantiate AtlasConfig so we don't have to keep the genesis attachments around
-        let moved_atlas_config = AtlasConfig::default(self.config.is_mainnet());
+        let moved_atlas_config = AtlasConfig::new(self.config.is_mainnet());
```
kantai, line 518: This will need to use the configured atlas config:

```suggestion
        let moved_atlas_config = self.config.atlas.clone();
```

kantai, line 518: This will need to use the configured atlas config:

```suggestion
        let moved_atlas_config = self.config.atlas.clone();
```

kantai (commented):
This looks good to me, but it will have to be tested on some production nodes before this can merge.

testnet/stacks-node/src/run_loop/neon.rs
jbencin: Good catch. Now that `AtlasConfig` can take user supplied values, it needs to be passed around and cloned rather than instantiated from scratch. This might not be the only place that needs to be changed, there are several other uses of `AtlasConfig::new()` in the codebase:

```
❯ grep -rIn AtlasConfig::new * | wc -l
11
```

jbencin: All other uses of `AtlasConfig::new()` are in test code, so nothing else needs to be changed

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -515,7 +515,7 @@ impl RunLoop {
         let chain_state_db = self.boot_chainstate(burnchain_config);
```
jbencin, line 515: It's not something I changed in this PR, but it looks like this section here with `atlas_config` and `genesis_attachments` does nothing and should be removed. As far as I can tell these variables are not used after this line

kantai (approved):
LGTM -- please just open an issue for the genesis attachments point you raised.

testnet/stacks-node/src/run_loop/neon.rs
jbencin: Done, issue #3635

pavitthrap (approved):

PR #3624, opened 2023-03-20
Fix: Incorporate unlocks in mempool admitter, #3623

### LINKED ISSUE #3623: Mempool admission doesn't handle unlocks correctly
This is not a consensus-critical bug, but it is something that can cause some user headaches.

The current mempool admission logic doesn't correctly handle unlocks and could tell people who have PoX-locked all of their account funds and try to transfer (or do a contract-call with a very high fee) after an unlock that they don't have enough funds for their transaction.

The consensus codebase performs "lazy unlocks" (this is true of both the PoX-2 unlock and the old way that PoX unlocked at the end of a cycle): the code stores the height at which an unlock occurs and only actually alters the account entry to "unlocked" when the account changes (via credit, debit, or applying a new lock). The mempool admission logic doesn't check the lazy unlock correctly: it will think an account is still locked up until the unlock actually materializes via one of those changes. I'm pretty sure this bug has existed since 2.0 launch (and would have come up if someone tried to transfer immediately after an unlock).

The patch for this is fairly straightforward. Fundamentally, it looks something like this:

```diff
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -6975,6 +6975,7 @@ impl StacksChainState {
     /// unconfirmed microblock stream trailing off of it.
     pub fn will_admit_mempool_tx(
         &mut self,
+        sortdb: &SortitionDB,
         current_consensus_hash: &ConsensusHash,
         current_block: &BlockHeaderHash,
         tx: &StacksTransaction,
@@ -7019,7 +7020,7 @@ impl StacksChainState {
 
         let current_tip =
             StacksChainState::get_parent_index_block(current_consensus_hash, current_block);
-        let res = match self.with_read_only_clarity_tx(&NULL_BURN_STATE_DB, &current_tip, |conn| {
+        let res = match self.with_read_only_clarity_tx(&sortdb.index_conn(), &current_tip, |conn| {
             StacksChainState::can_include_tx(conn, &conf, has_microblock_pubk, tx, tx_size)
         }) {
             Some(r) => r,
@@ -7039,7 +7040,7 @@ impl StacksChainState {
                 {
                     debug!("Transaction {} is unminable in the confirmed chain tip due to nonce {} != {}; trying the unconfirmed chain tip",
                            &tx.txid(), mismatch_error.expected, mismatch_error.actual);
-                    self.with_read_only_unconfirmed_clarity_tx(&NULL_BURN_STATE_DB, |conn| {
+                    self.with_read_only_unconfirmed_clarity_tx(&sortdb.index_conn(), |conn| {
                         StacksChainState::can_include_tx(
                             conn,
                             &conf,
```

This should be hotfixed. However, note that actually remedying the user pain here will require that a sufficient number of miners and nodes in the network upgrade to the hotfixed version. In the meantime, user workarounds are possible.

## User workarounds

Users can work around this issue before nodes upgrade by sending any transaction that performs a credit or debit action on their account. For example:

1. Send a small amount of STX from a different address to the account they want to unlock.
2. Send a very small transfer from the account they want to unlock to another account.
3. Perform a contract-call from the account they want to unlock.

kantai:
### Description

This fixes the mempool admission behavior in #3623 by supplying a pointer to the current burn state (sortition db) to the mempool admission logic.

Will update this PR with regression testing before merging.

### Applicable issues
- fixes #3623

### Checklist
- [x] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
+## [2.1.0.0.2]
+
+This software update is a hotfix to resolve improper unlock handling
+in mempool admission. This release's chainstate directory is
+compatible with chainstate directories from 2.1.0.0.2.
+
+### Fixed
+
+- Fix mempool admission logic's improper handling of PoX unlocks. This would
+  cause users to get spurious `NotEnoughFunds` rejections when trying to submit
+  their transactions (#3623)
+
 ## [2.1.0.0.1]
 
 ### Fixed
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -6975,6 +6975,7 @@ impl StacksChainState {
     /// unconfirmed microblock stream trailing off of it.
     pub fn will_admit_mempool_tx(
         &mut self,
+        burn_state_db: &dyn BurnStateDB,
         current_consensus_hash: &ConsensusHash,
         current_block: &BlockHeaderHash,
         tx: &StacksTransaction,
@@ -7019,7 +7020,7 @@ impl StacksChainState {
 
         let current_tip =
             StacksChainState::get_parent_index_block(current_consensus_hash, current_block);
-        let res = match self.with_read_only_clarity_tx(&NULL_BURN_STATE_DB, &current_tip, |conn| {
+        let res = match self.with_read_only_clarity_tx(burn_state_db, &current_tip, |conn| {
             StacksChainState::can_include_tx(conn, &conf, has_microblock_pubk, tx, tx_size)
         }) {
             Some(r) => r,
@@ -7039,7 +7040,7 @@ impl StacksChainState {
                 {
                     debug!("Transaction {} is unminable in the confirmed chain tip due to nonce {} != {}; trying the unconfirmed chain tip",
                            &tx.txid(), mismatch_error.expected, mismatch_error.actual);
-                    self.with_read_only_unconfirmed_clarity_tx(&NULL_BURN_STATE_DB, |conn| {
+                    self.with_read_only_unconfirmed_clarity_tx(burn_state_db, |conn| {
                         StacksChainState::can_include_tx(
                             conn,
                             &conf,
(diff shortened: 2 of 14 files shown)


kantai:
New regression test added:

```
chainstate::stacks::tests::block_construction::mempool_incorporate_pox_unlocks
```

You can test the regression pretty simply by applying this diff to this PR and then running that test:

```diff
diff --git a/src/core/mempool.rs b/src/core/mempool.rs
index 7f3c195d2..5d5d121dc 100644
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -180,7 +180,7 @@ impl MemPoolAdmitter {
         tx_size: u64,
     ) -> Result<(), MemPoolRejection> {
         chainstate.will_admit_mempool_tx(
-            &sortdb.index_conn(),
+            &clarity::vm::database::NULL_BURN_STATE_DB,
             &self.cur_consensus_hash,
             &self.cur_block,
             tx,
```

And then testing:

```bash
$ cargo test block_construction::mempool_incorporate
...
INFO [1679332313.674012] [src/chainstate/burn/db/sortdb.rs:4694] [chainstate::stacks::tests::block_construction::mempool_incorporate_pox_unlocks] ACCEPTED(38) leader key register 9158ed4afbd48e7c855e025384ddd1c181a8e26ff5b5083e0067f21ca0bc3b06 at 38,1
INFO [1679332313.720628] [src/chainstate/stacks/tests/block_construction.rs:4373] [chainstate::stacks::tests::block_construction::mempool_incorporate_pox_unlocks] Checking balance, v1_unlock_height: 37, burn_block_height: 37
thread 'chainstate::stacks::tests::block_construction::mempool_incorporate_pox_unlocks' panicked at 'called `Result::unwrap()` on an `Err` value: NotEnoughFunds(9999990000, 0)', src/chainstate/stacks/tests/block_construction.rs:4440:27


failures:
    chainstate::stacks::tests::block_construction::mempool_incorporate_pox_unlocks

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1176 filtered out; finished in 2.31s

error: test failed, to rerun pass `--lib`
```

jcnelson (approved):
Thanks for jumping on this @kantai!

pavitthrap (approved):
verified regression - thanks for the fix!

igorsyl:
@kantai sorry I missed this review. I'm way overloaded. Thanks for working on this!

PR #3632, opened 2023-03-23
CI/Build: Speedup unit tests 

kantai:
### Description

Our unit-test task in CI currently takes ~2 hours to run. This impacts the time-from-approval to time-of-merge for PRs. I think there's a couple potential quick wins here:

1. Use opt-level=3 for package dependencies in `dev` profiles. This shouldn't impact developer compiler times substantially, because those packages should only be rebuilt quite rarely. However, these can speed up the performance of tests.
2. Replace the dockerized unit test with raw GH workflow `run` commands. 

I'm keeping this as a draft because I want to see how the CI actually responds to these changes.


kantai:
Looks like 4cd8ef001f7b8c05b40831e00a9bb0f14f9d6418 shaved ~30 minutes off of the unit-tests runtime. Will try out the raw Github actions next.

pavitthrap (approved):

PR #3633, opened 2023-03-23
Feat: Add `peer-pub-key` command to `stacks-inspect`

kantai:
### Description

Specifying `bootstrap_node` entries in a stacks-node config file requires the peer's public key. Figuring out a peer's public key from its `local_peer_seed` isn't always trivial (you can get it from the logs during node bootup, but you don't necessarily want to have to deploy a node to figure out what its public key will be).

But, computing the public key from a peer seed value is pretty easy from the library, so I added a `peer-pub-key` command to `stacks-inspect` which can help in this case.


stacks-common/src/util/secp256k1.rs
```diff
@@ -268,6 +270,27 @@ impl Secp256k1PrivateKey {
         }
     }
 
+    /// Create a Secp256k1PrivateKey from seed bytes by repeatedly
+    ///  SHA256 hashing the seed bytes until a private key is found.
+    ///
+    /// If `seed` is a valid private key, it will be returned without hashing.
+    /// The returned private key's compress_public flag will be `true`
+    pub fn from_seed(seed: &[u8]) -> Secp256k1PrivateKey {
```
pavitthrap, line 278: perhaps you should bound the number of loops at some reasonable number so there's no chance of this running forever

pavitthrap (approved):
lgtm other than build issue

obycode (approved):
LGTM!

stacks-common/src/util/secp256k1.rs
kantai: I'd rather leave as is, because this PR is mostly just refactoring this code. The likelihood of running forever is also exponentially low, and given that a node operator is in control of this seed themselves, they can choose a seed that doesn't loop forever.

PR #3636, opened 2023-03-23
fix: can get the burn header hash of the last-validated block

### LINKED ISSUE #3616: `get-burn-block-info?` function return `none` on (burnchain_tip_height - 1) on testnet 2.1
**Describe the bug**
When calling the function `(get-burn-block-info? header-hash bc-height)`, where `bc-height` is the burnchain block height, the return value is `none` when it should be the header hash.

I deployed a contract on the live testnet network and made contract calls to make sure that it's not an issue with the API.

Deployed contract: https://explorer.stacks.co/txid/ST2VRK3C70CND13BTQ0SA9YK6V73DY5Z4WGTKCHD3.retail-lavender-vulture?chain=testnet

The code in the contract:
```
(define-public (get-block (height uint))
    (ok (get-burn-block-info? header-hash height))
)
```

I was testing calling the function while the block with Stacks height `100427` was being processed. Last-processed burnchain block height at that state was `2424493`.

Contract-call with height `2424494` argument returns `none` (as expected): https://explorer.stacks.co/txid/0x803b816c6d9ec4c9c31ad4ce75c8ed539d4bf34ba1545d5f3812361927a7bf30?chain=testnet

Contract-call with height `2424493` argument returns `none` (unexpected behavior): https://explorer.stacks.co/txid/0x24560499f0ba9c257bccd5c7e6e099bc16fa75b6bfaad2e826179a23e03f4f40?chain=testnet 
Contract-call with height `2424492` argument returns the header hash (as expected): https://explorer.stacks.co/txid/0x69088d1ba1c28757546502eea9237729d66a2b09a202e5ed4a5db93be48bc060?chain=testnet
Contract-call with height `2424491` argument returns the header hash (as expected): https://explorer.stacks.co/txid/0xddbee923f9dac206c9ed4a7630a91dc81f8dcbc2ddd88de15e165d6321c0be29?chain=testnet

**Steps To Reproduce**
This is on the live testnet, so if anyone wants to try, they can deploy a contract and make contract calls on the deployed contract. In the same way that I did in the previous section.

**Expected behavior**
Looking at the [SIP-015](https://github.com/stacksgov/sips/blob/feat/sip-015/sips/sip-015/sip-015-network-upgrade.md#new-method-get-burn-block-info), I think we should be able to get the data that we're requesting up until the last processed last-processed burnchain block. In the presented scenario, I should have been able to get the header hash from burnchain block `2424493`.

 - [Public Testnet](https://stacks-node-api.testnet.stacks.co/v2/info) : stacks-node 2.1.0.0.0 (master:f0ed1b9, release build, linux [x86_64])

FriendsFerdinand:
### Description
Breaking change to the clarity function `get-burn-block-info?` as explained in the SIP-015. This fix is so that when a clarity user requests for the burn_header_hash of the latest-validated block, they will get the hash and not the value None.

### Applicable issues
- fixes #3616 

Leaving as a draft as I'm still figuring contributions to this repo

### Checklist
- [ ] Changelog is updated


kantai (commented):
Hi @FriendsFerdinand -- thanks for opening this PR!

Before this can be merged, two changes are needed:

1. The change in behavior has to (at least) be gated by the current epoch identifier. That is, the behavior change needs to check if the epoch is greater than or equal to 2.2. For an example, see https://github.com/stacks-network/stacks-blockchain/pull/3625). This means this PR should probably be targeted against that same `stacks-2.2` branch, and perhaps wait until the new epoch ID has been added into that branch (or incorporate those changes and we can handle any merge conflicts)
2. Test cases must be added to test the behavior before and after the epoch activation. The test cases should be both unit cases, and integration cases in the the `stacks-node/src/tests/` module. These should assert the desired behavior from `get-burn-block-info?`

pavitthrap:
We believe that this has been addressed in 2.1.0.0.3 release - want to take a look and confirm @FriendsFerdinand ?

pavitthrap:
Closing PR because we think it may have been addressed. We'll continue discussion in the issue.

PR #3637, opened 2023-03-24
docs: add tag specifications

pavitthrap:
### Description
Adds docs to standardize what GitHub tags to use for PRs/issues.


CONTRIBUTORS.md
```diff
@@ -32,6 +32,48 @@ This project and everyone participating in it is governed by this [Code of Condu
 - For consensus breaking changes, branch off of the `next` branch.
 - For hotfixes, branch off of `master`.
 
+### GitHub Tags
+We use tags on PRs and issues to categorize them. You should try to select a tag from each category below if possible. 
+
+**Timeline- and consensus- based tags:**
+- Next non-breaking change => `ship now`
```
igorsyl, line 39: can rename to "next: non-breaking"

CONTRIBUTORS.md
```diff
@@ -32,6 +32,48 @@ This project and everyone participating in it is governed by this [Code of Condu
...
+- Next breaking change => `ship next`
```
igorsyl, line 40: "next: consensus"

igorsyl (changes requested):
Thank you for doing this!

CONTRIBUTORS.md
```diff
@@ -32,6 +32,48 @@ This project and everyone participating in it is governed by this [Code of Condu
...
+- Issues that are not being worked on (and there are no plans to) => `icebox`
+- PRs that are temporarily frozen => `frozen`
+- Must be in a breaking change => `consensus-critical`
+
+**Issue Type Tags:**
+- For first-time or new contributors => `good first issue`
+- Stacks lab issues (infrastructure-type changes) => `help wanted`
+- Documentation updates => `documentation`
+- Improvement to the chain (speed, reliability) => `optimization`
+- Improvements to CI/CD => `CI`
```
igorsyl, line 51: "devops"?

kantai: I think `CI` is better than `devops` here.

CONTRIBUTORS.md
kantai: I don't think we should use `next` in multiple tags, because it also refers to a branch (which is the next consensus version). `ship now` works better I think.

CONTRIBUTORS.md
kantai: Yeah, I prefer something like `next stacks` or `next consensus` to `ship next`

CONTRIBUTORS.md
```diff
@@ -32,6 +32,48 @@ This project and everyone participating in it is governed by this [Code of Condu
 - For consensus breaking changes, branch off of the `next` branch.
 - For hotfixes, branch off of `master`.
 
+### GitHub Tags
+We use tags on PRs and issues to categorize them. You should try to select a tag from each category below if possible. 
+
+**Timeline- and consensus- based tags:**
+- Next non-breaking change => `ship now`
+- Next breaking change => `ship next`
+- Future change, in the work pipeline => `ship future`
```
kantai, line 41: Like above, I'd prefer `future stacks` or `future consensus`

kantai (commented):
This looks good to me, just had some comments on the `ship *` tags.

CONTRIBUTORS.md
igorsyl: - agree on not using next name
- however, ship now doesn't communicate that change is consensus-compatible

CONTRIBUTORS.md
igorsyl: "ship consensus" to avoid using "next"?

CONTRIBUTORS.md
igorsyl: "future consensus" +1

CONTRIBUTORS.md
igorsyl: CI does not include CD?
Should we have separate CD?

igorsyl (changes requested):
What about references to L1? Should they be changed to L2?

CONTRIBUTORS.md
pavitthrap: I think by default changes tend to be consensus-compatible, so I will leave it as `ship now` (unless I get a better suggestion)

CONTRIBUTORS.md
pavitthrap: I don't think it's worth it to have a separate CD tag. I'm going to leave this as CI.

CONTRIBUTORS.md
pavitthrap: So - ship future was supposed to encapsulate both consensus and non-consensus breaking changes. 

It's why I have `consensus-critical` as a totally separate tag.

pavitthrap:
> What about references to L1? Should they be changed to L2?

What references are you referring to? I'm not seeing this ...

saralab:
Can we close this? Or do we want to implement these tag specifications?

PR #3645, opened 2023-04-03
feat: save & serve block receipts

bestmike007:
### Description

Users can query events on an EVM node by calling the `evm_getLogs` rpc method. However, on stacks, this is currently only possible by relying on an event observer.

To address this issue, this PR proposes a solution by introducing a way to serialize and store block receipts in the block chunk store. The binary data can then be accessed through a new http endpoint at `^/v2/block_receipts/([0-9a-f]{64})$`.

This PR is currently in the POC stage and will be strengthened after discussions.

### Checklist
- [ ] Changelog is updated


wileyj:
@bestmike007 2 questions here:
1. how much additional storage would be required to add this? ballpark estimate is fine
2. can this be configurable? i.e. a config switch to turn it on and have it be disabled by default?

zone117x:
A dump of events from March (saved as their original json in a new-line-delimited json file) is 11 GB.

If the goal is to have something similar to `evm_getLogs`, then shouldn't this be storing and retrieving the Stacks equivalent? i.e. contract-log-print payloads, rather than the entire event payload.

bestmike007:
> @bestmike007 2 questions here:
> 
> 1. how much additional storage would be required to add this? ballpark estimate is fine
> 2. can this be configurable? i.e. a config switch to turn it on and have it be disabled by default?

@wileyj
hey,

1. we're still running the sync with this, and will get back later
2. sure, it's on my todo list, a better idea would be a `--clean-up-antique-events` feature flag.

> A dump of events from March (saved as their original json in a new-line-delimited json file) is 11 GB.

@zone117x 
Yes this is true if we save events as jsonl files, a significant storage will be wasted because the field names are saved repeatedly. Compression helps, but not the perfect. That's why I started from implementing `StacksMessageCodec` for related structs, and then I realized I need to implement it for clarity `Value` as well. And I skipped it by serializing clarity values into json string instead of the binary format, this can be addressed later after our discussion.

I do have a question regarding `StacksMessageCodec`: why don't we use an existing serialization solution, e.g. bincode? And if it's really necessary to have it, isn't it better to implement the derive for it, rather than implementing `StacksMessageCodec` for every single struct?

> If the goal is to have something similar to `evm_getLogs`, then shouldn't this be storing and retrieving the Stacks equivalent? i.e. contract-log-print payloads, rather than the entire event payload.

@zone117x 
It'll be really hard to implement the `evm_getLogs` equivalent, because 1. evm events are indexed, it can be queried by topics; 2. evm events includes nft/ft transfers as well, not just contract-log-print payloads.

So the goal is different, in this PR we're just trying to store the events in raw binary, and then the events can be accessed through index block hash. And then I thought, why don't we save the whole block receipt? There's not much difference in terms of the size.

Once we have the API to access events for each block, it'll be much easier for a program to synchronize Stacks onchain data into an offchain analysis database.

jcnelson:
I think this is a good idea in principle, because it makes booting up an API side-car faster.  I agree with @wileyj that it should be a configurable option, and off by default.  I like how you made use of the `StreamCursor` enum for serving the block receipt data.

I'm on leave until the end of June, so I won't be able to leave a timely and detailed review.  But from a first glance, I had the following recommendations:

* I'd avoid adding new dependencies like `bincode` if at all possible.  Storing the event data as JSON is fine with me, given that the goal is just to gather the requisite data for different event observers to quickly boot up.

* This code needs unit tests.

* This code needs to be tested on mainnet before it can be merged, and every block receipt must be queried from the node and compared to the receipts obtained from a production event observer.  In all cases, the receipts must match.

* Please don't change the `Cargo.lock` file.

bestmike007:
I'll split this PR into multiple smaller ones.

bestmike007:
> @bestmike007 2 questions here:
> 
> 1. how much additional storage would be required to add this? ballpark estimate is fine
> 2. can this be configurable? i.e. a config switch to turn it on and have it be disabled by default?

@wileyj for the question 1, we've synced at block height 89555, and the disk usage is:

image

bestmike007:
@pavitthrap hey, I started to split this PR into smaller ones, within each I'll add tests and update doc if necessary. So I'm just waiting for the first one (#3663 ) to be merged before I can continue.

bestmike007:
Hey @jcnelson @pavitthrap @kantai ,

As discussed in https://github.com/stacks-network/stacks-blockchain/pull/3663, instead of implementing StacksMessageCodec, I added dedicated serialization for block receipt without adding new dependencies.

Most of the types are serialized into binary format; except for StacksTransactionEvent and clarity Value, which are serialized into JSON strings.

There's no need to implement deserialization for now.

Please take a look, and if it looks good to you, I'll add tests.

saralab:
@bestmike007  Any updates on this to move forward? Can we close this ?

bestmike007:
I made the changes according to review, stilling waiting for you to review it before I can continue adding test cases and wrap up.

kantai:
@bestmike007, we do really appreciate your contributions here, but I think that the `stacks-core` implementation cannot accept this PR: while I'm sure that storing and serving block receipts is useful for some operators, the vast majority of nodes _won't_ enable this and probably cannot: the amount of data that would be stored here is _very_ large. While this could be managed by enabling or disabling operation with compile-time flags, ultimately, if the code lands in this repo, the maintainers are responsible for ensuring it continues to function, and I don't that's feasible currently. Furthermore, there's ways to implement this kind of feature without modifying `stacks-core`: an event-observer could itself store store and serve all of these receipts. Then, because the event-observer isn't a part of the `stacks-core` codebase, it could use any http server or language it chooses for its implementation.

PR #3649, opened 2023-04-06
[sBTC] Add stx_address method to PegOutRequestOp

netrome:
This PR adds a `PegOutRequestOp::stx_address()` method to derive the stacks address of an sBTC holder requesting to peg out.

Closes https://github.com/stacks-network/stacks-blockchain/issues/3647

### DIFF
--- a/src/chainstate/burn/operations/peg_out_request.rs
+++ b/src/chainstate/burn/operations/peg_out_request.rs
@@ -162,7 +162,7 @@ enum ParseError {
     SliceConversion,
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum RecoverError {
     PubKeyRecoveryFailed(&'static str),
     AddressConstructionFailed,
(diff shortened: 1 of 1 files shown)


jferrant (approved):

igorsyl:
@kantai I would be grateful for your attention to this PR given it's one of the last remaining pieces for a working sBTC Alpha.

image

kantai (commented):
This LGTM, but it needs two more kinds of tests:

1. A test that uses a test vector (i.e., just a hex encoded bitcoin transaction) as input, with a hard coded expected `stx_address`.
2. Test coverage for the two error cases

netrome:
> This LGTM, but it needs two more kinds of tests:
> 
>     1. A test that uses a test vector (i.e., just a hex encoded bitcoin transaction) as input, with a hard coded expected `stx_address`.
> 
>     2. Test coverage for the two error cases

Thank you for the feedback. I have added a test case with the test vector and another one for one of the error cases. I discovered that the other error case was infallible, so I changed the method and return type to reflect that.

It would be great if you could have another look @kantai :pray:

kantai (approved):
LGTM!

igorsyl:
Thank you @kantai

PR #3656, opened 2023-04-10
microblock antientropy protocol

### LINKED ISSUE #2031: Microblocks: anti-entropy protocol for unconfirmed microblocks
Need to devise and implement a protocol for nodes that are missing some unconfirmed microblocks to go and figure out who else has them, and go fetch them.

The best idea I have so far is to make is so that when relaying a microblock, a node also indicates whether or not it has all of the ancestors of the microblock being relayed.  Then, have recipient nodes request unconfirmed microblocks from the RESTful API if it's missing any microblock ancestors.  This protocol is more efficient than polling neighbors for unconfirmed microblocks -- a node will only pull exactly the data it needs from exactly the neighbors that have it.  Also, this protocol works across multiple microblock forks.

pavitthrap:
### Description
This implements the unconfirmed microblock anti-entropy protocol. This protocol runs by default, and makes nodes monitor its peers unconfirmed tips through the v2/info endpoint. If it discovers at least one peer with a higher unconfirmed tip height, it queries that peer for its microblocks. 

The protocol logic is encapsulated within a state machine. Refer to the function `do_microblock_tip_sync` in `p2p.rs`. 

### Applicable issues
- fixes #2031 
#### Unrelated fixes 
- Fixes the way `num_new_confirmed_microblocks` is computed in `process_network_result`. 
- Changes logic in src/net/[dns.rs](http://dns.rs/)` to remove an `unreachable`. 
- Changes logic in `preprocess_downloaded_microblocks` such that only microblocks that were actually stored are propagated to neighbors and returned from the function. 
- Simplified logic in `mempool_sync_send_query`.

#### Feedback needed 
- I need feedback for the values for `microblock_tip_sync_interval` and `microblock_tip_sync_timeout` - they are now 10 and 60 seconds, respectively. I think it could make sense to make each of these values smaller. Might need to play around with them in live testing. 
- Need feedback on whether I should be propagating synced microblocks differently. I am currently forwarding them to peers the way other unconfirmed microblocks are forwarded, and I set the relayers data to be empty. 

#### Other considerations
- Perhaps we might want to consider toggling this protocol through a config setting (versus running it by default)
- We might want to consider making this protocol more state aware? Keeping track of which peers were already queried off of a particular canonical tip to avoiding re-querying the same peer. 
- A note about how the tip sync protocol keeps track of the number of microblocks it has received. The protocol can receive duplicate blocks, so double-counting. This could be “fixed” but not sure if it’s worth doing. Right now, this quantity is tracked by the `PeerNetwork` struct by the field `microblock_tip_sync_amount`. 

### Checklist
- [ ] Changelog is updated


src/net/relay.rs
```diff
@@ -1334,11 +1445,12 @@ impl Relayer {
...
     }
 
-    /// Process blocks and microblocks that we recieved, both downloaded (confirmed) and streamed
-    /// (unconfirmed). Returns:
+    /// Process blocks and microblocks that we recieved, downloaded (confirmed), streamed
+    /// (unconfirmed), and obtained through microblock tip sync protocol (unconfirmed). Returns:
     /// * set of consensus hashes that elected the newly-discovered blocks, and the blocks, so we can turn them into BlocksAvailable / BlocksData messages
     /// * set of confirmed microblock consensus hashes for newly-discovered microblock streams, and the streams, so we can turn them into MicroblocksAvailable / MicroblocksData messages
     /// * list of unconfirmed microblocks that got pushed to us, as well as their relayers (so we can forward them)
+    /// * an tuple of the microblocks synced through the microblock tip sync protocol & the block ID of the tip they were built off of (as an option)
```
obycode, line 1453:
```suggestion
    /// * a tuple of the microblocks synced through the microblock tip sync protocol & the block ID of the tip they were built off of (as an option)
```

src/net/p2p.rs
```diff
@@ -3613,2219 +3750,4120 @@ impl PeerNetwork {
...
+                            Err(e) => {
+                                warn!(
+                                    "{:?}: microblock_tip_sync_send_microblock_query failed: {:?}",
+                                    &self.local_peer, &e
+                                );
+                                self.microblock_tip_sync_state =
+                                    MicroblockTipSyncState::RequestMicroblocks(
+                                        next_seq_num_data,
+                                        sorted_tip_data,
+                                    );
```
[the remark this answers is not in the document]
pavitthrap: So a few lines above this there is `if let Some(tip_data) = next_seq_num_data.pop()`. Each time the state machine enters the `RequestMicroblocks` state, either next_seq_num or sorted_tip_data gets shorter, so the state machine can't loop here forever. 

Also, this is the error case, the blocking case is handled in the case `WaitForMicroblocks`. 

Technically I don't need to explicitly set `self.microblock_tip_sync_state` but I prefer to for the sake of clarity.

src/net/p2p.rs
```diff
@@ -3613,2219 +3750,4120 @@ impl PeerNetwork {
...
+        tip_data
+            .into_iter()
+            .filter(|(_, _, seq_num)| {
+                if let Some(local_seq_num) = local_seq_num_opt {
+                    seq_num > &local_seq_num
+                } else {
+                    true
+                }
+            })
+            .collect()
```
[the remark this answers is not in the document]
pavitthrap: Addressed in `763d18dadd9f36b776b1198f2622921e8e8dedbe`

obycode (approved):
LGTM!

src/net/p2p.rs
```diff
@@ -3613,2219 +3750,4119 @@ impl PeerNetwork {
...
+                        }
+                        Some(http_response) => {
+                            match http_response {
+                                HttpResponseType::Microblocks(_, data) => {
+                                    debug!("{:?}: Microblock sync received response with {} microblocks", &network.local_peer, data.len());
+                                    Ok((true, Some(data)))
+                                }
+                                _ => {
+                                    warn!(
+                                    "{:?}: Microblock tip sync request v2/microblocks/unconfirmed, but received unexpected response type {:?}",
```
jcnelson, line 4049: minor style nit: whitespace

src/net/p2p.rs
```diff
@@ -3613,2219 +3750,4119 @@ impl PeerNetwork {
...
+            None => {
+                return None;
+            }
+        };
+
+        let mut next_seq_num_data = vec![];
+        loop {
+            let should_pop_next = match sorted_tip_data.back() {
+                Some((_, _, seq_num)) => {
+                    if *seq_num == next_seq_num {
```
jcnelson, line 4109: This could just be compressed to `*seq_num == next_seq_num`

src/net/p2p.rs
```diff
@@ -3613,2219 +3750,4119 @@ impl PeerNetwork {
...
+                            &self.local_peer
                         );
-                        return None;
+                        self.microblock_tip_sync_reset();
+                        return (true, faulty_peers, None);
+                    }
+                    // gather URLs from peers
+                    let mut urls = vec![];
+                    for (_event_id, convo) in &self.peers {
+                        if !convo.is_authenticated() || !convo.is_outbound() {
```
jcnelson, line 4215: It's okay to query inbound peers for this -- you can drop the `!convo.is_outbound()`

src/net/p2p.rs
```diff
@@ -3613,2219 +3750,4119 @@ impl PeerNetwork {
...
+    /// Do the actual work in the state machine.
+    /// Return true if we need to prune connections.
+    fn do_network_work(
         &mut self,
         sortdb: &SortitionDB,
-        chainstate: &StacksChainState,
-        event_id: usize,
-        new_blocks: &BlocksAvailableData,
+        mempool: &MemPoolDB,
+        chainstate: &mut StacksChainState,
```
jcnelson, line 4493: Why is this `&mut` now?

src/net/p2p.rs
jcnelson, line 1: Any idea why the diff on this file is so utterly massive?

src/net/relay.rs
```diff
@@ -1795,6 +1932,22 @@ impl Relayer {
...
+                    if let Some((id, mbs)) = new_synced_microblocks {
+                        if mbs.len() > 0 {
+                            debug!("{:?}: Microblocks from tip sync protocol: {}", &_local_peer, mbs.len());
+                            let mblocks_msg = MicroblocksData {
+                                index_anchor_block: id,
+                                microblocks: mbs,
+                            };
+
+                            let msg = StacksMessageType::Microblocks(mblocks_msg);
+                            if let Err(e) = self.p2p.broadcast_message(vec![], msg) {
```
jcnelson, line 1946: I'm not sure we should be doing this here.  These microblocks have _already_ been broadcasted -- there's a good chance that most or all of the neighbors already have them, so re-sending them is likely to be wasteful.  The microblock anti-entropy protocol is meant for nodes that for whatever reason didn't receive the microblocks to recover from a local lack of data.

src/net/p2p.rs
pavitthrap: I'm indenting this line, I assume that's what you mean?

src/net/p2p.rs
pavitthrap: I believe this is just strange formatting on the diff. My PR does not change the type of `chainstate` for either the functions `do_network_work` or `handle_unsolicited_BlocksAvailable` (as compared to the branch `develop`)

src/net/relay.rs
pavitthrap: Removed.

src/net/p2p.rs
pavitthrap: I have no idea why the diff on this PR is so strange.

src/net/dns.rs
```diff
@@ -345,20 +345,17 @@ impl DNSClient {
...
-        }
 
         let _ = match self.requests.get(&req) {
             Some(None) => {
                 return Ok(None);
             }
             Some(Some(resp)) => resp,
             None => {
-                unreachable!();
+                return Err(net_error::LookupError(format!(
```
jcnelson, line 355: Thanks for this refactor :+1:

src/net/relay.rs
```diff
@@ -998,17 +1008,114 @@ impl Relayer {
...
                     (*consensus_hash).clone(),
-                    (index_block_hash, microblock_stream.clone()),
+                    (index_block_hash, stored_microblocks),
                 );
             }
         }
-        ret
+
+        // Now, process and store microblocks obtained through the microblock tip sync protocol.
+        if let Some(result) = &network_result.synced_microblock_result {
```
jcnelson, line 1027: Do you think this code could be in its own method, like `preprocess_synced_microblocks()`?

src/net/server.rs
```diff
@@ -593,12 +593,14 @@ impl HttpPeer {
             }
 
             let client_sock_opt = self.sockets.get_mut(&event_id);
-            if client_sock_opt.is_none() {
-                test_debug!("No such socket event {}", event_id);
-                to_remove.push(*event_id);
-                continue;
-            }
-            let client_sock = client_sock_opt.unwrap();
+            let client_sock = match client_sock_opt {
```
jcnelson, line 596: Thanks for this refactor :+1:

jcnelson (approved):
LGTM!  Thanks @pavitthrap!

saralab:
Closing, microblocks aren't enabled at the moment

PR #3663, opened 2023-04-17
feat: add default stacks message codec for bool & option 

bestmike007:
### Description

Part of #3645 

### Checklist
- [ ] Changelog is updated

### DIFF
--- a/stacks-common/src/codec/mod.rs
+++ b/stacks-common/src/codec/mod.rs
@@ -9,6 +9,9 @@ use crate::util::secp256k1::MESSAGE_SIGNATURE_ENCODED_SIZE;
 #[macro_use]
 pub mod macros;
 
+#[cfg(test)]
+pub mod tests;
+
 #[derive(Debug)]
 pub enum Error {
     /// Failed to encode
@@ -93,6 +96,7 @@ impl_stacks_message_codec_for_int!(u8; [0; 1]);
 impl_stacks_message_codec_for_int!(u16; [0; 2]);
 impl_stacks_message_codec_for_int!(u32; [0; 4]);
 impl_stacks_message_codec_for_int!(u64; [0; 8]);
+impl_stacks_message_codec_for_int!(u128; [0; 16]);
 impl_stacks_message_codec_for_int!(i64; [0; 8]);
 
 impl StacksMessageCodec for [u8; 32] {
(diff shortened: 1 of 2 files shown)


bestmike007:
@kantai please review this when you have time. This PR is part of https://github.com/stacks-network/stacks-blockchain/pull/3645

kantai:
Marking as frozen until after SIP-022 and SIP-023 are resolved -- any PRs unrelated to those issues will not be reviewed until after that.

stacks-common/src/codec/tests/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+#[test]
+fn codec_for_bool() {
+    let t = true;
+    let f = false;
+    let t_binary = t.serialize_to_vec();
+    let f_binary = f.serialize_to_vec();
+    assert_eq!(&t_binary, &vec![1u8]);
+    assert_eq!(&f_binary, &vec![0u8]);
+    assert_eq!(
+        bool::consensus_deserialize(&mut Cursor::new(&vec![1u8])).unwrap(),
```
pavitthrap, line 14: Could you add an assert where you call deserialize for t_binary and f_binary respectively?

stacks-common/src/codec/tests/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+    let n_binary = n.serialize_to_vec();
+    assert_eq!(
+        &n_binary,
+        &vec![
+            0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef, 0x12, 0x34, 0x56, 0x78, 0x90, 0xab,
+            0xcd, 0xef
+        ]
+    );
+    assert_eq!(
+        u128::consensus_deserialize(&mut Cursor::new(&vec![
```
pavitthrap, line 64: Same comment.

pavitthrap (approved):

stacks-common/src/codec/tests/mod.rs
bestmike007: Do you mean this?
```rust
    assert_eq!(
        bool::consensus_deserialize(&mut Cursor::new(&vec![1u8])),
        Ok(true)
    );
```

Can you give me an example if it's not?

bestmike007: But the only difference is that I'm using literal instead of a variable, isn't it?

stacks-common/src/codec/mod.rs
```diff
@@ -191,6 +195,58 @@ where
...
+        Ok(())
+    }
+
+    fn consensus_deserialize<R: Read>(fd: &mut R) -> Result<bool, Error> {
+        let b: u8 = read_next(fd)?;
+        match b {
+            0 => Ok(false),
+            1 => Ok(true),
+            _ => {
+                warn!("Invalid bool value");
```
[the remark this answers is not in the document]
bestmike007: Good catch!

stacks-common/src/codec/tests/mod.rs
obycode: With the literals, there is still a chance that they could get out of sync if one is changed without changing the other, while this style ensures that the serialize and deserialize are coordinated.

bestmike007: Understood. I do prefer defining inputs and expected results above all, but if this is the convention, I'd like to stick to it. Please check again.

pavitthrap (changes requested):
After discussing this PR in the blockchain weekly, we realized we want to avoid implementing the trait `StacksMessageCodec` unless it is absolutely necessary.

Looking at your larger PR, it appears you use the `consensus_serialize` function in `store_block_receipt` in `blocks.rs`. 

You can instead use `serialize` directly, instead of relying on the `StacksMessageCodec` trait. This function gets implemented for structs if you specify `#[derive(Serialize)]`.

bestmike007:
TBH, I didn't want to implement `StacksMessageCodec`; however, if we store the raw data in JSON format, it's going to require a lot of disk space, which is not acceptable.

kantai:
> TBH, I didn't want to implement StacksMessageCodec; however, if we store the raw data in JSON format, it's going to require a lot of disk space, which is not acceptable.

I'm not so worried about the additional disk space here, but even so, you don't necessarily need to use serde_json to perform the serialization. You can write your own deserialization/serialization implementations by implementing the serde traits or chose not use serde at all.

The overall point is that `StacksMessageCodec` is a trait that not only is used for serialization, but it _implies_ that the serialization is consensus-critical (hence the function name `consensus_serialize`). Adding new implementations of that trait implies that those implementations are consensus critical, which I don't think is your intention.

bestmike007:
I see what you mean, let me dig more on how to implement this in a better way.

PR #3675, opened 2023-04-20
3509 sbtc stackerdb rpc interface

### LINKED ISSUE #3509: [sBTC] StackerDB RPC interface
Stackers are expected to interact with their [StackerDB](#3504) mainly through the node's RPC interface.  This interface needs the following endpoints:

* An endpoint for pushing new chunks.  This endpoint will require the chunk, plus a signature over the chunk's ID and new Lamport clock

* An endpoint for downloading raw chunk data.

* An endpoint for querying the chunk metadata, so the client can learn what Lamport clock to use.

* An endpoint for listing which StackerDBs this node serves

* An endpoint for listing which neighbors of this node replicate a particular StackerDB.

stjepangolemac:
### Description
Attempt to add a new RPC call for StackerDB.

### DIFF
--- a/src/chainstate/stacks/db/blocks.rs
+++ b/src/chainstate/stacks/db/blocks.rs
@@ -5202,10 +5202,10 @@ impl StacksChainState {
                         // strictly speaking this check is defensive. It will never be the case
                         // that a `miner_reward` has a `recipient_contract` that is `Some(..)`
                         // unless the block was mined in Epoch 2.1.  But you can't be too
-                        // careful... 
+                        // careful...
                         if evaluated_epoch >= StacksEpochId::Epoch21 {
                             // in 2.1 or later, the coinbase may optionally specify a contract into
-                            // which the tokens get sent.  If this is not given, then they are sent 
+                            // which the tokens get sent.  If this is not given, then they are sent
                             // to the miner address.
                             miner_reward.recipient.clone()
                         }
--- a/src/core/mempool.rs
+++ b/src/core/mempool.rs
@@ -1731,7 +1731,7 @@ impl MemPoolDB {
         nonce: u64,
     ) -> Result<Option<MemPoolTxMetadata>, db_error> {
         let sql = format!(
-            "SELECT 
+            "SELECT
                           txid,
                           origin_address,
                           origin_nonce,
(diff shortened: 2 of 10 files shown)


stjepangolemac:
I've come across some ownership problems, that's why I opted for it. Let me get back to you when I'm able to confirm the spot and the error message.

jcnelson:
Then it's even more important that you drop the Mutex now. You should never
use concurrency primitives to avoid compile-time checks. It prevents you
from developing a deep enough understanding of the code to be effective at
shipping needful features.

stjepangolemac:
So if I got back to commit 12cd2386b821c7724ec3db4b965ff74986be21fd and try to compile I get this:

```
error[E0502]: cannot borrow `*self` as mutable because it is also borrowed as immutable
    --> src/net/p2p.rs:5455:9
     |
5455 |           PeerNetwork::with_network_state(self, |ref mut network, ref mut network_state| {
     |           ^------------------------------       ---------------------------------------- immutable borrow occurs here
     |           |
     |  _________immutable borrow later used by call
     | |
5456 | |             let http_stacks_msgs = PeerNetwork::with_http(network, |ref mut net, ref mut http| {
5457 | |                 http.run(
5458 | |                     network_state,
...    |
5465 | |                     &self.stacker_db,
     | |                      --------------- first borrow occurs due to use of `self.stacker_db` in closure
...    |
5469 | |             Ok(())
5470 | |         })
     | |__________^ mutable borrow occurs here

```

This is the section:

https://github.com/stacks-network/stacks-blockchain/blob/12cd2386b821c7724ec3db4b965ff74986be21fd/src/net/p2p.rs#L5455-L5470

I'll take a closer look on Monday to see if I'm missing a bigger picture and try to come up with some ideas.

stjepangolemac:
@jcnelson I've reverted the mutex changes and updated the code. I've overlooked that `PeerNetwork` is passed down to all the right places and its `stacker_db` can be used where needed.

I've realized that StackerDB PRs are based on an old master that has a bug preventing me from running and testing this locally. It is fixed in the current master.

```
Process abort due to thread panic: panicked at 'FATAL: no epoch 2.1 defined', testnet/stacks-node/src/config.rs:463:14
```

I'll check priorities with the team and might rebase StackerDB PRs on latest master.

PR #3676, opened 2023-04-20
MARF trie blob compression support

cylewitruk:
### Description
This PR aims to address the growing storage requirements of stacks nodes from a node operator's perspective by introducing configuration options to compress the disk-backed MARF blob files using either LZ4 or ZStd-0. This PR does *not* address the actual bytes stored in tandem with Clarity-related operations (the "fees"-side of things). In other words, this is for node operators to be able to decide for themselves based on their environment how to balance CPU vs disk usage.

- According to tests, these changes reduce MARF blob file on-disk size by between 40 and 70% from current levels, dependent of course on the type of data stored and compression algorithm used.  LZ4 is faster but compresses less, ZStd is marginally slower but compresses better.
- The compression works on the blob level. Each blob is independently compressed and reflected in the sqlite3 database (in table `marf_data`) together with its offset, _compressed_ length and compression type (none = 0, lz4 = 1, zstd0 = 2).
- The `TrieFile` enum (which had `Disk` and `RAM` options) is removed in favor of a more uniform `TrieFile2` (named for now) struct.
- Several "common" items have been moved to the  `stacks_common` crate (such as `MARFOpenOpts`).
- A new attribute procedural macro `#[generate_test_cases_for_marf_open_opts]` is added in the new crate `stacks-proc-macros` which makes use of the `test_cases` crate to generate separate tests for each `MARFOpenOpts` combination (instead of iterating over the possibilities within one single test).
- Migrations have been updated/added for both DB->File and File->File (with different compression options).  For DB->File the new external trie blob file will be created using the configured compression algorithm.  For File->File the file will be migrated.
- Added an additional caching layer for external trie files, using an LRU-cache for uncompressed trie blobs by block id to avoid performance degradation (the same block is often read many times within a short period of time).
- Made `TrieCachingStrategy` an enum.
- Default behavior remains the same (no compression), it must actively be enabled via config to use it.

### Todo
- Implement proper performance counters (`bench` counters)
- Make uncompressed LRU-cache size configurable (uses memory); currently hardcoded to 256 blobs.
- Following tests are failing:
  - `chainstate::stacks::db::unconfirmed::test::test_unconfirmed_refresh_10_microblocks_10_stx_transfers`
  - `chainstate::stacks::db::unconfirmed::test::test_unconfirmed_refresh_invalid_microblock`
  - `chainstate::stacks::db::unconfirmed::test::test_unconfirmed_refresh_one_microblock_stx_transfer`
- Clean up/restore logging
- Clean up/add comments

Still have a little cleaning up to do, but thought I'd create the PR as a draft anyway to get feedback :)

### Applicable issues
- fixes #

- 40-70% reduction in node storage needs; this directly translates to reduced node TCO especially in cloud environments where storage is charged per GB and month.
- Ever so slightly higher CPU usage. The compression/decompression adds marginal latency, however is partially offset by lower I/O latency due to less bytes written/read to disk.

### Checklist
- [ ] Changelog is updated


cylewitruk:
@jcnelson would be fun to hear your thoughts on this if/when you get a few minutes to have a peek :)

jcnelson:
I'm on leave until June 30.

src/chainstate/stacks/index/bits.rs
```diff
@@ -351,22 +352,27 @@ fn inner_read_nodetype_at_head<F: Read + Seek>(
         Error::CorruptionError(format!("read_node_type: Unknown trie node type {}", ptr_id))
     })? {
         TrieNodeID::Node4 => {
+            //eprintln!("Node4");
```
jcnelson, line 355: Style nit: please remove commented-out code before merging. Thanks!

src/chainstate/stacks/index/cache.rs
```diff
@@ -181,7 +181,18 @@ impl<T: MarfTrieId> TrieCache<T> {
     /// environ, or failing that, it will use a no-op strategy.
     pub fn default() -> TrieCache<T> {
         if let Ok(strategy) = std::env::var("STACKS_MARF_CACHE_STRATEGY") {
-            TrieCache::new(&strategy)
+            match strategy.as_ref() {
```
jcnelson, line 184: Thanks for this refactor :fist_oncoming:

src/chainstate/stacks/index/file.rs
```diff
@@ -139,35 +191,31 @@ impl TrieFile {
...
+        //eprintln!("Stored trie blob {} to offset {} with length {:?}", bhh, result.offset, result.storage_size);
+        let block_id = trie_sql::write_external_trie_blob(
+            db,
+            bhh,
+            result.offset,
+            result.storage_size as u64,
+            compression_type,
+        )?;
+
+        self.decompressed_lru.put(block_id, buffer.to_vec());
```
jcnelson, line 216: It seems like `self.decompressed_lru` is going to get pretty big.  Each serialized trie is between 10**5 and 10**7 bytes.

Also, it seems like the LRU cache stores up to 256 tries, right?  If so, then while this would save disk I/O accesses and decompression work for walking trie nodes in recently-accessed tries, I worry about cache thrashing.  For example, someone who wanted to DoS the network could craft a malicious transaction that deliberately loaded a large number of infrequently-accessed tries during its execution, causing mining to slow down and causing validation to slow down (possibly by a considerable margin).

src/chainstate/stacks/index/file.rs
```diff
@@ -181,15 +229,194 @@ impl TrieFile {
...
+
+        let bench_start = SystemTime::now();
+        let extern_trie = trie_sql::get_external_trie_offset_length(db, block_id)?;
+        //eprintln!("Loading trie blob with block_id={}, offset={}, length={}, compression={}",
+        //    &block_id, &extern_trie.offset, &extern_trie.length, &extern_trie.compression);
+
+        self.seek(SeekFrom::Start(extern_trie.offset))?;
+        let mut buf = vec![0u8; extern_trie.length as usize];
+        self.read_exact(&mut buf)?;
+        let result: Vec<u8>;
```
jcnelson, line 281: Minor style nit: you don't have to declare `result` in its own statement.  You could write this as

```
let result = match extern_trie.compression {
   BlobCompressionType::None => buf
   BlobCompressionType::LZ4 => lz4_decompress_size_prepended(buf.as_slice()).unwrap()
   BlobCompressionType::ZStd(..) => zstd::decode_all(buf.as_slice()).unwrap()
};
```

src/chainstate/stacks/index/file.rs
```diff
@@ -181,15 +229,194 @@ impl TrieFile {
...
+
+        match extern_trie.compression {
+            BlobCompressionType::None => {
+                result = buf;
+            }
+            BlobCompressionType::LZ4 => {
+                result = lz4_decompress_size_prepended(buf.as_slice()).unwrap();
```
jcnelson, line 288: We generally try to avoid using `.unwrap()` for fallible code.  Can you elaborate under what circumstances this function can fail?  Even if it fails via an irrecoverable error, the user of `.expect()` is preferred because then you at least get an error message.

src/chainstate/stacks/index/file.rs
```diff
@@ -329,60 +652,83 @@ impl TrieFile {
...
+        self.seek(SeekFrom::Start(offset))?;
+
+        let mut compression_elapsed: Duration = Duration::from_micros(0);
+        let compression_bench = SystemTime::now();
+
+        let storage_bytes = match compression_type {
+            BlobCompressionType::None => buf.into(),
+            _ => {
+                let compressed = Self::compress_blob(compression_type, buf)?;
+                compression_elapsed = compression_bench.elapsed().unwrap();
```
jcnelson, line 707: Benchmarking code should not lead to a node crash.  The `.unwrap()` can be reached for benign reasons -- for example, the host might not support the system clock ID that `SystemTime` passes.

src/chainstate/stacks/index/file.rs
```diff
@@ -329,60 +652,83 @@ impl TrieFile {
...
+            buf.len(),
+            storage_bytes.len(),
+            offset,
+            &compression_type,
+            compression_elapsed);
+
+        self.seek(SeekFrom::Start(offset))?;
+        self.write_all(storage_bytes_slice)?;
+        self.flush()?;
+        //self.sync_data()?;
```
jcnelson, line 724: `flush()` by itself does not persist data.  In the case of writing data to a file, you must also call `.sync_data()` on the file descriptor to ensure that all newly-written blocks are persisted.

src/chainstate/stacks/index/file.rs
```diff
@@ -392,9 +738,11 @@ impl TrieFile {
         block_id: u32,
         ptr: &TriePtr,
     ) -> Result<TrieHash, Error> {
-        let offset = self.get_trie_offset(db, block_id)?;
-        self.seek(SeekFrom::Start(offset + (ptr.ptr() as u64)))?;
-        let hash_buff = read_hash_bytes(self)?;
+        self.load_trie_blob(db, block_id)?;
```
jcnelson, line 741: This could easily lead to significant disk thrashing.  In the worst case, you'd be loading a whole trie (10^5 to 10^7 bytes) from disk simply to read a single trie node as part of following backptrs.  Moreover, this evicts other cached trie blobs.

jcnelson: This applies to several points in the code where you call `self.load_trie_blob()` as part of reading either a trie node or a trie hash.  For brevity, I won't comment on them all unless you want me to.

src/chainstate/stacks/index/proofs.rs
```diff
@@ -699,7 +699,7 @@ impl<T: MarfTrieId> TrieMerkleProof<T> {
                 hash_idx += 1;
             }
         }
-        trace!("Shunt proof node: idx={}, all_hashes={:?}", idx, all_hashes);
+        test_debug!("Shunt proof node: idx={}, all_hashes={:?}", idx, all_hashes);
```
jcnelson, line 702: Please keep the use of `trace!`. You can trigger trace debugging by setting the `STACKS_LOG_TRACE=1` environ.

stacks-proc-macros/src/lib.rs
```diff
@@ -0,0 +1,186 @@
+extern crate quote;
+extern crate syn;
+
+use proc_macro::TokenStream;
+use quote::quote;
+use stacks_common::types::chainstate::*;
+use syn::{parse_macro_input, ItemFn};
+
+#[proc_macro_attribute]
+pub fn generate_test_cases_for_marf_open_opts(
```
jcnelson, line 10: Nice refactoring!

src/chainstate/stacks/index/bits.rs
cylewitruk: Yeah there's a bit of tidying up to do before an eventual merge :)

src/chainstate/stacks/index/file.rs
cylewitruk: Yeah it's hardcoded at 256 right now but have a point above to make it configurable (cache size of zero would effectively disable the cache), and the default could be a much more conservative number.

Since this is an opt-in feature for node operators it would be impossible for a potential attacker to know the configurations of each node to plan such an attack (is compression even enabled? If so, which algorithm? Is their LRU cache enabled and if so how large is their cache? Spec of the node, etc.).

But I agree, potential attack vectors should be analyzed :)

src/chainstate/stacks/index/file.rs
cylewitruk: Thanks for the tip! 😊

stacks-proc-macros/src/lib.rs
cylewitruk: Thanks! 🙂

src/chainstate/stacks/index/cache.rs
```diff
@@ -337,537 +341,3 @@ impl<T: MarfTrieId> TrieCache<T> {
         self.state_ref().load_block_id(block_hash)
     }
 }
-
-#[cfg(test)]
```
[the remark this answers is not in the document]
cylewitruk: Moved to test/cache.rs

src/chainstate/stacks/index/proofs.rs
cylewitruk: Yeah I was gonna go and revert all of the logging back to original state before an eventual merge, still have a few failing tests I need to figure out then I'll bite that bullet :)

src/chainstate/stacks/index/file.rs
cylewitruk: Hm.. Perhaps maintining a separate index file with all (back)pointers & hashes, which would most likely easily fit in memory as well, could be a potential solution and possibly also offer a performance boost if walking can be done in-memory?

My spontaneous thought anyway.. thoughts?

cylewitruk:
@jcnelson I haven't had any more time to look into this and I'm assuming given the changes coming in Nakamoto that it's pretty much a dead-end anyway? Should I close this?

jcnelson:
Nakamoto will retain the MARF structures for dealing with Bitcoin (and thus Stacks) forks.  However, the biggest thing that stops me from approving this PR is the fact that its design will lead to _extreme_ disk thrashing.  Caching tries via LRU just isn't a good idea when an adversarial transaction can force the nodes to load and evict tries according to an attacker-controlled schedule.

cylewitruk:
> Nakamoto will retain the MARF structures for dealing with Bitcoin (and thus Stacks) forks.  However, the biggest thing that stops me from approving this PR is the fact that its design will lead to _extreme_ disk thrashing.  Caching tries via LRU just isn't a good idea when an adversarial transaction can force the nodes to load and evict tries according to an attacker-controlled schedule.

Yeah I remember the argument, but I had a counter-suggesting maintaining a separate index for pointers - anything you think could work?

Not high prio at all - I'm happy to close the PR and keep it as historical reference and continue in a GH discussion as time/energy allows?

PR #3678, opened 2023-04-21
fix: disable PoX auto-unlock 

jcnelson:
This disables the call to `handle-unlock` at the start of a reward cycle in Epoch 2.2, and adds test coverage.

### DIFF
--- a/src/chainstate/stacks/boot/mod.rs
+++ b/src/chainstate/stacks/boot/mod.rs
@@ -354,65 +354,74 @@ impl StacksChainState {
         let pox_contract = boot::boot_code_id(POX_2_NAME, clarity.is_mainnet());
 
         let mut total_events = vec![];
-        for (principal, amount_locked) in cycle_info.missed_reward_slots.iter() {
-            // we have to do several things for each principal
-            // 1. lookup their Stacks account and accelerate their unlock
-            // 2. remove the user's entries from every `reward-cycle-pox-address-list` they were in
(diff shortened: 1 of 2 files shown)


src/chainstate/stacks/boot/mod.rs
```diff
@@ -354,65 +354,74 @@ impl StacksChainState {
...
-                    },
-                    |_, _| false,
-                )
-                .expect("FATAL: failed to handle PoX unlock");
+        // short-circuit if auto-unlock no longer happens (as of epoch 2.2)
+        if clarity.get_epoch() >= StacksEpochId::Epoch22 {
+            debug!("PoX auto-unlock is disabled in this epoch"; "reward_cycle" => cycle_number);
+        } else {
+            debug!("PoX auto-unlock starting"; "reward_cycle" => cycle_number, "epoch" => %clarity.get_epoch());
+            for (principal, amount_locked) in cycle_info.missed_reward_slots.iter() {
```
kantai, line 363: Purely to reduce the line changes (I know its just indentation, but a smaller delta is probably better anyways), I'd suggest early returning rather than if-else:

```rust
       if clarity.get_epoch() >= StacksEpochId::Epoch22 {
            debug!("PoX auto-unlock is disabled in this epoch"; "reward_cycle" => cycle_number);
            return Ok(vec![])
       }
```

pavitthrap:
Closing in favor of other solution, which is to disable pox-2 entirely.

PR #3679, opened 2023-04-22
merge changes for devnet

obycode:
This is `feat/address-pox-increase` plus changes merged in from `feat/devnet-support-next` to make it work with devnet.

It is not working yet.


obycode:
Closed to follow branch `feat/sip-022-pox-disable` instead.

PR #3684, opened 2023-04-26
CI: Upgrade debian dockerfiles from stretch (eol) to bullseye (current stable)

kantai:
Stretch is EOL, so our debian dockerfiles stopped working. This PR upgrades to the current stable release (bullseye), and disables macOS builds until cross-compilation support is fixed.

### DIFF
--- a/build-scripts/Dockerfile.windows-x64
+++ b/build-scripts/Dockerfile.windows-x64
@@ -1,4 +1,4 @@
-FROM rust:stretch as build
+FROM rust:buster as build
 
 ARG STACKS_NODE_VERSION="No Version Info"
 ARG GIT_BRANCH='No Branch Info'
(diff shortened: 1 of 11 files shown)


obycode (approved):
Thanks!

build-scripts/Dockerfile.windows-x64
```diff
@@ -1,4 +1,4 @@
-FROM rust:stretch as build
+FROM rust:buster as build
```
zone117x, line 1: It looks like Windows builds succeed if this uses `rust:bullseye`. The `gcc-mingw-w64-x86-64` package version currently used by the buster image runs into an internal compiler error.

kantai: Cool, I'll switch to `bullseye` in these docker files and reenable to windows output.

PR #3685, opened 2023-04-27
add epoch gate to claritydb::get_v2_unlock_height()

obycode:


obycode:
Testing update:

We are using the repo https://github.com/hirosystems/stacks-2-1-testing to do end-to-end testing on this node using `stacks-devnet-js`. That library allows us to spin up a bitcoin node and stacks node then interact with them. The tests also interact with the stacks node using the stacks.js library to submit transactions, make read-only calls, etc.

This repo was previously used for testing 2.1, so for 2.2, we can focus on the tests in _pox/stacking/_. To run them locally, there is a [README file](https://github.com/hirosystems/stacks-2-1-testing#stacks-devnet-js) with details. The infrastructure should all be in place to make adding new tests relatively straight forward (big thanks @friedger for work on the framework and the existing tests).

Test include:
* Single user stacks and increases
  - No bug before or after 2.2
  - Stacks unlocked after 2.2
* Multiple users stack and increase
  - Exhibits bug before 2.2
  - Stacks unlocked after 2.2
* Multiple users stack and increase enough to trigger crash
  - Node crashes with 2.1 node
  - Node continues producing blocks with 2.2 node
* Various scenarios where a user stacks directly and delegates simultaneously
* Direct stacking with auto-unlock
* User stacks and extends
* Call `stx-account` before and after 2.2 activation

I will continue to add more tests throughout this afternoon.

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
+## [2.2.0.0.1]
+
+This is a **consensus-breaking** release to address a bug and DoS vector in `stacks-increase`.
```
obycode, line 10:
```suggestion
This is a **consensus-breaking** release to address a bug and DoS vector in pox-2's `stack-increase` function.
```

pavitthrap (approved):

kantai (approved):
LGTM!

This testing infrastructure should *definitely* get included in this repo's CI in the future, it makes writing integration tests a lot easier.

PR #3689, opened 2023-04-28
Add `delegated-to` field in map

obycode:
This is part of SIP-022 to fix the problems with `stack-increase`. This
field allows us to track whether the stacker is stacking directly or
through a pool, and ensure that they cannot stack more than one way at a
time. More testing is still needed.

### Applicable issues
- fixes #

### Checklist
- [ ] Changelog is updated


obycode:
On the testing side, I plan to:
1. Expand on the unit tests in https://github.com/Trust-Machines/pox-clarinet-tests
2. Fix and validate existing tests in https://github.com/hirosystems/stacks-2-1-testing
3. Add new tests to https://github.com/hirosystems/stacks-2-1-testing

src/chainstate/stacks/boot/pox-3.clar
```diff
@@ -0,0 +1,1308 @@
...
+         ;; to extend, there must already be an etry in the stacking-state
+         (stacker-state (unwrap! (get-stacker-info tx-sender) (err ERR_STACK_EXTEND_NOT_LOCKED)))
+         (amount-ustx (get locked stacker-info))
+         (unlock-height (get unlock-height stacker-info))
+         (cur-cycle (current-pox-reward-cycle))
+         (unlock-in-cycle (burn-height-to-reward-cycle unlock-height))
+         ;; if the account unlocks *during* this cycle (should only occur during testing),
+         ;; set first-extend-cycle to the next cycle.
+         (first-extend-cycle (if (> (+ cur-cycle u1) unlock-in-cycle)
+                                    (+ cur-cycle u1) unlock-in-cycle))
```
kantai, line 1028: I think this is something we may want to push a little bit on -- in `pox-3`, can we assume the `first-extend-cycle` is the next cycle? (because unlocks all occurred at epoch 2.2) If so, I think we should do this check, but yield an `err` on failing the check.

src/chainstate/stacks/boot/pox-3.clar
```diff
@@ -0,0 +1,1308 @@
...
+         (cur-first-reward-cycle (get first-reward-cycle stacker-state))
+         (first-reward-cycle (if (> cur-cycle cur-first-reward-cycle) cur-cycle cur-first-reward-cycle)))
+
+    ;; must be called with positive extend-count
+    (asserts! (>= extend-count u1)
+              (err ERR_STACKING_INVALID_LOCK_PERIOD))
+
+    ;; stacker must not be delegating
+    (asserts! (is-none (get delegated-to stacker-state))
+              (err ERR_STACKING_IS_DELEGATED))
```
kantai, line 1039: I'd also add the `(len (get reward-set-indexes ..)` check for a solo stacker that you did above in `stacks-increase`

src/chainstate/stacks/boot/pox-3.clar
```diff
@@ -0,0 +1,1308 @@
...
+
+      ;; must be called directly by the tx-sender or by an allowed contract-caller
+      (asserts! (check-caller-allowed)
+        (err ERR_STACKING_PERMISSION_DENIED))
+
+      ;; stacker must be delegated to tx-sender
+      (asserts! (is-eq (unwrap! (get delegated-to stacker-state)
+                                (err ERR_STACKING_NOT_DELEGATED))
+                       tx-sender)
+                (err ERR_STACKING_PERMISSION_DENIED))
```
kantai, line 1128: You should also add a `(asserts! (is-eq 0 (len (get reward-set-indexes ..)` check here.

src/chainstate/stacks/boot/pox-3.clar
```diff
@@ -0,0 +1,1308 @@
...
```
kantai, line 1219: You should also add a `(asserts! (is-eq 0 (len (get reward-set-indexes ..)` check here (just as a sanity check).

src/chainstate/stacks/boot/pox-3.clar
```diff
@@ -0,0 +1,1308 @@
...
+              (existing-total (unwrap-panic (map-get? reward-cycle-total-stacked { reward-cycle: reward-cycle })))
+              (add-amount (get add-amount data))
+              (total-ustx (+ (get total-ustx existing-total) add-amount)))
+            ;; stacker must match
+            (asserts! (is-eq (get stacker existing-entry) (some (get stacker data))) none)
+            ;; update the pox-address list
+            (map-set reward-cycle-pox-address-list
+                     { reward-cycle: reward-cycle, index: reward-cycle-index }
+                     { pox-addr: (get pox-addr existing-entry),
+                       total-ustx: (+ (get total-ustx existing-entry) add-amount),
```
kantai, line 957: Can you add a `;;` comment above this line that says it addresses the `stack-increase` issue in 2.1?

kantai (approved):
This looks good to me, just had a few comments about some extra guards we could add, and a comment about some sanity checks that could be applied.

src/chainstate/stacks/boot/pox-3.clar
obycode: > If so, I think we should do this check, but yield an err on failing the check.

Yes, I think that makes sense.

obycode: Looks that check already exists below:

```clarity
(asserts! (> first-extend-cycle cur-cycle) (err ERR_STACKING_INVALID_LOCK_PERIOD))
```

kantai: Yep, so line 1027 would become just:

```clarity
(first-extend-cycle unlock-in-cycle)
```

obycode: I had deleted the variable `first-extend-cycle` and updated the uses. Any problem with that?

obycode: Maybe better to rename `unlock-in-cycle` to `first-extend-cycle`.

obycode:
Ok, @kantai's comments have all been addressed in the latest.

src/chainstate/stacks/boot/pox-3.clar
kantai: > Maybe better to rename unlock-in-cycle to first-extend-cycle.

Yeah, that seems preferable, with a comment like `;; first-extend-cycle will be the cycle in which tx-sender *would have* unlocked`

zone117x:
Should this be rebased onto https://github.com/stacks-network/stacks-blockchain/pull/3688 (since that PR already includes pox-3.clar, and several other PRs are currently branched off that one)?

Issue #3690, opened 2023-05-01
Contract calls with traits as parameters are rejected


friedger:
**Describe the bug**
Tx and read-only calls to functions with traits as parameters are rejected with unchecked TypeValueError

**Steps To Reproduce**
Try to make a swap on alexgo or make a read-only call to get-staked-of SP2C2YFP12AJZB4MABJBAJ55XECVS7E4PMMZ89YZR.arkadiko-stake-pool-diko-v1-2 

**Expected behavior**
Accept and process the tx

**Data**
Example response from alex swap:
POST to `https://api.hiro.so/v2/transactions`
```
{"error":"transaction rejected",
"reason":"BadFunctionArgument",
"reason_data":{"message":"TypeError(
TraitReferenceType(TraitIdentifier { name: ClarityName(\"sip-010-trait\"), contract_identifier: QualifiedContractIdentifier { issuer: StandardPrincipalData(SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9), name: ContractName(\"trait-sip-010\") } }), 
PrincipalType)"},
"txid":"d3b187e5606fd538ec875496ee3ca757177d836f6f7a31281eafe68d94ea7ff9"}
``` 

Example response from arkadiko:

POST to `/v2/contracts/call-read/SP2C2YFP12AJZB4MABJBAJ55XECVS7E4PMMZ89YZR/arkadiko-stake-pool-diko-v1-2/get-stake-of`
```
{"okay":false,
"cause":"Unchecked(TypeValueError(
TraitReferenceType(TraitIdentifier { name: ClarityName(\"stake-registry-trait\"), 
contract_identifier: QualifiedContractIdentifier { issuer: StandardPrincipalData(SP2C2YFP12AJZB4MABJBAJ55XECVS7E4PMMZ89YZR), name: ContractName(\"arkadiko-stake-registry-trait-v1\") } }), 
Principal(Contract(QualifiedContractIdentifier { issuer: StandardPrincipalData(SP2C2YFP12AJZB4MABJBAJ55XECVS7E4PMMZ89YZR), name: ContractName(\"arkadiko-stake-registry-v1-1\") }))))"}
```

pavitthrap:
Resolved in Epoch 2.3.

PR #3691, opened 2023-05-01
Fix epoch check for type canonicalization

obycode:
Potential fix for #3690


obycode:
Closed in favor of #3693

PR #3693, opened 2023-05-01
Epoch 2.3: Fix trait invocations via canonicalization

kantai:
### Description

This PR introduces a new Stacks epoch, 2.3, fixing the canonicalization regression introduced in Epoch 2.2.

This PR is a hard-fork, but will be compatible with existing chainstate directories (as long as the node is upgraded before the fork height).

As this is an emergency hotfix, I'm not sure if this PR must go through the SIP process, or if there's an abbreviated process (@jcnelson would know better than I).

### Testing

This PR includes tests for the witnessed bad behavior in 2.2, and demonstrates that the behavior is fixed in 2.3.

This PR also includes an E2E test using bitcoin regtest that tests the epoch switch and the behavior.

### TODO

Set the testnet and mainnet activation heights.


jcnelson:
Please also bump the Stacks block version.

kantai:
> Please also bump the Stacks block version.

Bumped in c281835

clarity/src/vm/tests/traits.rs
```diff
@@ -37,6 +37,7 @@ use crate::vm::ContractContext;
 #[case(ClarityVersion::Clarity1, StacksEpochId::Epoch2_05)]
 #[case(ClarityVersion::Clarity1, StacksEpochId::Epoch21)]
 #[case(ClarityVersion::Clarity2, StacksEpochId::Epoch21)]
+#[case(ClarityVersion::Clarity2, StacksEpochId::Epoch22)]
```
jcnelson, line 40: Are you going to include epoch 2.3 here as well?

stacks-common/src/types/mod.rs
```diff
@@ -73,11 +73,12 @@ pub enum StacksEpochId {
     Epoch2_05 = 0x02005,
     Epoch21 = 0x0200a,
     Epoch22 = 0x0200f,
+    Epoch23 = 0x02014,
```
jcnelson, line 76: Curious what the bit pattern here means?  e.g why wasn't epoch 2.2 `0x0200b`?

testnet/stacks-node/src/tests/epoch_23.rs
```diff
@@ -0,0 +1,626 @@
+use std::collections::HashMap;
```
jcnelson, line 1: Missing copyright statement

testnet/stacks-node/src/tests/epoch_23.rs
```diff
@@ -0,0 +1,626 @@
...
+
+use clarity::vm::types::PrincipalData;
+
+#[test]
+#[ignore]
+/// Test the trait invocation behavior for contracts instantiated in epoch 2.05
+///  * in epoch 2.1: the trait invocation works
+///  * in epoch 2.2: trait invocation is broken, and returns a runtime error, even when wrapped
+///  * in epoch 2.3: the trait invocation works
+fn trait_invocation_behavior() {
```
jcnelson, line 32: There isn't a simple way to test this in a unit test?  I'm pretty sure this is also an option (and it would be less flaky and run faster).

clarity/src/vm/types/signatures.rs
```diff
@@ -722,8 +724,13 @@ impl TypeSignature {
...
         match epoch {
-            StacksEpochId::Epoch21 => self.canonicalize_v2_1(),
-            _ => self.clone(),
+            StacksEpochId::Epoch10
+            | StacksEpochId::Epoch20
+            | StacksEpochId::Epoch2_05
+            // Epoch-2.2 had a regression in canonicalization, so it must be preserved here.
+            | StacksEpochId::Epoch22 => self.clone(),
+            // Note for future epochs: Epochs >= 2.3 should use the canonicalize_v2_1() routine
+            StacksEpochId::Epoch21 | StacksEpochId::Epoch23 => self.canonicalize_v2_1(),
```
jcnelson, line 733: Is the new gating behavior covered by unit tests?

jcnelson (commented):
To confirm, there are no wildcard matches on the epoch anywhere else in the codebase?  Also, there are no equality checks on epochs anywhere?  We want to be sure that code introduced in 2.1 that is meant to run in epochs 2.1 _and later_ actually runs in the _later_ epochs as well.

stacks-common/src/types/mod.rs
kantai: They're just all +5.

testnet/stacks-node/src/tests/epoch_23.rs
kantai: The `clarity_vm::tests::contracts::trait_invocation_cross_epoch` unit test covers the behavior.

I think having both here is important -- because the integration test covers the miner path as well (also, using a real sortition db for the epoch switch).

clarity/src/vm/types/signatures.rs
kantai: Yes -- `clarity_vm::tests::contracts::trait_invocation_cross_epoch` covers the behavior in epochs 2.05, 2.1, 2.2, and 2.3.

clarity/src/vm/tests/traits.rs
kantai: Sure, will add.

testnet/stacks-node/src/tests/epoch_23.rs
kantai: Added

kantai:
> To confirm, there are no wildcard matches on the epoch anywhere else in the codebase?

That's correct -- I read through every instance of the Epoch identifiers.

> Also, there are no equality checks on epochs anywhere? We want to be sure that code introduced in 2.1 that is meant to run in epochs 2.1 _and later_ actually runs in the _later_ epochs as well.

That's correct -- I read through every instance of the Epoch identifiers.

obycode:
> To confirm, there are no wildcard matches on the epoch anywhere else in the codebase? Also, there are no equality checks on epochs anywhere? We want to be sure that code introduced in 2.1 that is meant to run in epochs 2.1 _and later_ actually runs in the _later_ epochs as well.

I have been looking for this and have not found any instances outside of tests.

fpbgg (approved):
These changes look good to me. I left a few inline remarks but nothing significant.

pavitthrap (approved):
mostly nits, feel free to ignore those.

clarity/src/vm/analysis/type_checker/v2_1/tests/mod.rs
```diff
@@ -58,6 +58,10 @@ pub mod contracts;
 #[case(ClarityVersion::Clarity1, StacksEpochId::Epoch2_05)]
 #[case(ClarityVersion::Clarity1, StacksEpochId::Epoch21)]
 #[case(ClarityVersion::Clarity2, StacksEpochId::Epoch21)]
+#[case(ClarityVersion::Clarity1, StacksEpochId::Epoch22)]
+#[case(ClarityVersion::Clarity2, StacksEpochId::Epoch22)]
+#[case(ClarityVersion::Clarity1, StacksEpochId::Epoch23)]
+#[case(ClarityVersion::Clarity2, StacksEpochId::Epoch23)]
 fn test_clarity_versions_type_checker(
```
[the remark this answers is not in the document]
pavitthrap: Not a pressing change though, but good practice in general when adding a new epoch to add these cases to all relevant functions.

src/clarity_vm/tests/contracts.rs
```diff
@@ -443,6 +446,72 @@ fn trait_invocation_cross_epoch() {
...
+    // now in Stacks 2.2
+    sim.execute_next_block_as_conn(|conn| {
+        let epoch = conn.get_epoch();
+        conn.as_transaction(|clarity_db| {
+            let error = clarity_db
+                .run_contract_call(
+                    &sender,
+                    None,
+                    &invoke_contract_id,
+                    "invocation-2",
```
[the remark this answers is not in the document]
kantai: These are the 2.2 blocks. It'd be worthwhile for the unit test to be updated for both of them in the 2.2 blocks. Rather than push the change to this PR, though, I'll open another one, because I don't want to have to run the CI process on this PR again (further delaying release). I have confirmed that it works as expected with the change though.

PR #3698, opened 2023-05-02
Fuzzing Branch for Epoch 2.3

fpbgg:
### Description

This implements support for fuzz-testing (see #3437) atop the Epoch 2.3 branch (see #3693). The purpose of this branch is to identify any issues or regressions introduced in the new release before they cause impact.

To get started:

```terminal
cargo build

cd clarity/fuzz
cargo build
```

To fuzz the `admits_type` function:
```terminal
cargo fuzz run fuzz_admits_type  -- -rss_limit_mb=4096
```

To fuzz the `least_supertype` function:
```terminal
car go fuzz run fuzz_least_supertype  -- -rss_limit_mb=4096
```

If the fuzzer causes a panic, you will need to manually run this command again. To keep running and generating new test cases in the event of a failure, you can use `keep-fuzzing.sh [fuzz_name_here]` to keep the loop executing.


pavitthrap:
This branch was created to support epoch 2.3 testing.

PR #3706, opened 2023-05-03
Fix: update peer versions to latest

kantai:
This updates the peer versions in the stacks-node to the latest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to the versioning scheme outlined in the [README.md](RE
 
 ## [2.3.0.0.1]
 
-This is a hotfix release to update the peer version identifier used by the stacks-node p2p network.
+This is a hotfix release to update:
+- peer version identifier used by the stacks-node p2p network.
+- yield interpreter errors in deser_hex
 
 This release is compatible with chainstate directories from 2.3.0.0.x and 2.1.0.0.x
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
+## [2.3.0.0.1]
+
+This is a hotfix release to update the peer version identifier used by the stacks-node p2p network.
+
+This release is compatible with chainstate directories from 2.3.0.0.x and 2.1.0.0.x
+
 ## [2.3.0.0.0]
 
 This is a **consensus-breaking** release to address a Clarity VM bug discovered in 2.2.0.0.1.
--- a/src/core/mod.rs
+++ b/src/core/mod.rs
@@ -50,8 +50,8 @@ pub use stacks_common::consts::{CHAIN_ID_MAINNET, CHAIN_ID_TESTNET, STACKS_EPOCH
 // fourth byte == highest epoch supported by this node
 // - 0x05 for 2.05
 // - 0x06 for 2.1
-pub const PEER_VERSION_MAINNET: u32 = 0x18000007;
-pub const PEER_VERSION_TESTNET: u32 = 0xfacade07;
+pub const PEER_VERSION_MAINNET: u32 = 0x18000008;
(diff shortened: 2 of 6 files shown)


src/core/mod.rs
```diff
@@ -50,8 +50,8 @@ pub use stacks_common::consts::{CHAIN_ID_MAINNET, CHAIN_ID_TESTNET, STACKS_EPOCH
 // fourth byte == highest epoch supported by this node
 // - 0x05 for 2.05
 // - 0x06 for 2.1
-pub const PEER_VERSION_MAINNET: u32 = 0x18000007;
-pub const PEER_VERSION_TESTNET: u32 = 0xfacade07;
+pub const PEER_VERSION_MAINNET: u32 = 0x18000008;
+pub const PEER_VERSION_TESTNET: u32 = 0xfacade08;
```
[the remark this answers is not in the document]
kantai: Let's save that for 2.4, I'd prefer to minimize the changes here as much as possible.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
+## [2.3.0.0.1]
+
+This is a hotfix release to update the peer version identifier used by the stacks-node p2p network.
+
+This release is compatible with chainstate directories from 2.3.0.0.x and 2.1.0.0.x
```
wileyj, line 12: I don't think this is necessary since this is not a hard-fork, so the chainstate being compatible is implied.

kantai: I always opt to put it in the top matter of the release sections in the changelog, just because there _have_ been releases that were consensus-compatible, but also required genesis re-syncs (earlier in the stacks-blockchain's life cycle, but I think it's still a fair question to answer here).

diwakergupta: +1 for keeping it - even when it's explicit, people ask questions; best to minimize assumptions. That said, at this point anyone not already on 2.2 is already ignored and anyone attempting to boot off a 2.1 chainstate [might run into the issue reported on Discord ](https://discord.com/channels/621759717756370964/730814571517968436/1103320645229617202), so is it worth including a note about that or dropping 2.1 altogether? And include 2.2.0.0.x

src/core/mod.rs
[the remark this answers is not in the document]
kantai: I added this commit to 2.4: https://github.com/stacks-network/stacks-blockchain/commit/3ca4b015898a82c8e960ffaedd57ac5e2c304dff

It adds a `PEER_NETWORK_EPOCH` const, and also validates it during the epochs validation, so that things like the integration tests (and also testing via genesis syncs, etc.) will panic if the network epoch isn't updated alongside the epochs list.

CHANGELOG.md
wileyj: > I always opt to put it in the top matter of the release sections in the changelog, just because there _have_ been releases that were consensus-compatible, but also required genesis re-syncs (earlier in the stacks-blockchain's life cycle, but I think it's still a fair question to answer here).

I see your point, my reasoning is based almost entirely on historically how the changelog has mentioned chainstate compatibility (seems to be only hard forks had mentioned it). 

I am in favor of *more* info though, but maybe we can adjust so that's clear what the minimum compatible chainstate is? i.e. rather than list a few versions, only list the minimum (in this case 2.1.0.0.0). 

further, i'm in favor of keeping this line for all future versions - but it does add an extra check on us to ensure we have this data.

wileyj:
With 2.4 release approaching, does it make sense to merge this and create a new release?
I'm leaning towards no, and that the changes here should be moved to the 2.4 branch.  unless we see a compelling reason to merge this now, i say we close it in favor of 2.4

CharlieC3:
@wileyj Not the most compelling reason, but I would prefer to be back to a state where [our infra shows an official release is deployed](https://api.hiro.so/v2/info). Releasing this sooner as a bug fix would allow us to do that without having to wait until 2.4 is released.

obycode (approved):
LGTM 👍

wileyj (approved):
shipit

PR #3715, opened 2023-05-12
Neon integration-style test for epoch 2.4

pavitthrap:
### Description
- Added two neon integration style tests for epoch 2.4. 
- Fixed a bug in the reward set calculation function - it would error if the epoch 2.4 boundary was crossed but if the pox-3 contract was uninitialized (which occurs if there is no Stacks block).


kantai (approved):
LGTM -- the broken unit tests are fixed upstream, so I wouldn't worry about that.

obycode (approved):
LGTM. Just one minor comment.

PR #3718, opened 2023-05-15
3717 add length prefixing to contract names in peg in requests

stjepangolemac:
### Description
Changes the way we store the contract name in the peg in wire format.

#### Before
Contract name is a 40 byte slice that needs to be zero padded.

#### After
Contract name is prefixed by a size byte and it doesn't need to be zero padded — thus saving space. Contract name is not specified if size is `0`.

### Applicable issues
- Closes https://github.com/stacks-network/stacks-blockchain/issues/3717

- Original issue in core-eng https://github.com/Trust-Machines/core-eng/issues/342

### Checklist
- ~~[ ] Changelog is updated~~
- ~~[ ] Required documentation changes (e.g., `docs/rpc/openapi.yaml` and `rpc-endpoints.md` for v2 endpoints, `event-dispatcher.md` for new events)~~

@netrome can you offer any guidance on 2 outstanding bullet points? I'm not sure if they're relevant at all. Also, I want to be extra considerate in asking for reviews. Do you know else could be the right person to review this?

### DIFF
--- a/src/chainstate/burn/operations/peg_in.rs
+++ b/src/chainstate/burn/operations/peg_in.rs
@@ -76,7 +76,7 @@ impl PegInOp {
         /*
             Wire format:
 
-            0      2  3                  24                            64       80
+            0      2  3                  24                            65       80
             |------|--|------------------|-----------------------------|--------|
              magic  op   Stacks address    Size prefixed contract name    memo
                                                (optional, see note)
@@ -108,17 +108,16 @@ impl PegInOp {
 
         let standard_principal_data = StandardPrincipalData(version, address_data);
 
-        let (recipient, bytes_read) =
-            if let Some((contract_name, bytes_read)) = Self::read_contract_name(&data[21..])? {
-                let contract_id =
-                    QualifiedContractIdentifier::new(standard_principal_data, contract_name);
+        let recipient = if let Some(contract_name) = Self::read_contract_name(&data[21..])? {
+            let contract_id =
+                QualifiedContractIdentifier::new(standard_principal_data, contract_name);
 
-                (contract_id.into(), 21 + bytes_read)
-            } else {
-                (standard_principal_data.into(), 22)
-            };
+            contract_id.into()
+        } else {
+            standard_principal_data.into()
+        };
 
-        let memo = data.get(bytes_read..).unwrap_or(&[]).to_vec();
+        let memo = data.get(62..).unwrap_or(&[]).to_vec();
 
         Ok(ParsedData { recipient, memo })
     }
@@ -137,19 +136,19 @@ impl PegInOp {
     /// # Errors
     /// - `Utf8Error` if the contract name is not valid UTF-8
     /// - `ParseError` if the contract name is not a valid `ContractName`
-    fn read_contract_name(data: &[u8]) -> Result<Option<(ContractName, usize)>, ParseError> {
+    fn read_contract_name(data: &[u8]) -> Result<Option<ContractName>, ParseError> {
         let size = data[0] as usize;
 
         if size == 0 {
             return Ok(None);
         }
 
-        let contract_name_bytes = &data[1..1 + size];
+        let contract_name_bytes = &data[1..41];
         let contract_name: ContractName = str::from_utf8(contract_name_bytes)?
             .to_string()
             .try_into()?;
 
-        Ok(Some((contract_name, 1 + size)))
+        Ok(Some(contract_name))
     }
 }
 
@@ -190,6 +189,8 @@ impl From<ClarityRuntimeError> for ParseError {
 
 #[cfg(test)]
 mod tests {
+    use std::iter::repeat;
+
     use rand::{distributions::Alphanumeric, Rng};
 
     use super::*;
@@ -235,6 +236,7 @@ mod tests {
         let stx_address = StacksAddress::new(1, addr_bytes.into());
         data.extend_from_slice(&addr_bytes);
         data.push(0); // Contract name not provided, so size is 0
+        data.extend(repeat(0).take(40)); // Contract name not provided, so zeroed out
         data.extend_from_slice(&memo);
 
         let tx = test::burnchain_transaction(data, Some(output2), opcode);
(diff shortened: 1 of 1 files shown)


netrome:
> @netrome can you offer any guidance on 2 outstanding bullet points? I'm not sure if they're relevant at all. Also, I want to be extra considerate in asking for reviews. Do you know else could be the right person to review this?

We can hold the changelog update until sBTC is ready to be released. There are no API changes from what I know here, so we should be good there too.

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -76,11 +78,17 @@ impl PegInOp {
...
+                                               (optional, see note)
+
+             # Notes
 
-             Note that `data` is missing the first 3 bytes -- the magic and op must
+             `data` is missing the first 3 bytes -- the magic and op must
              be stripped before this method is called. At the time of writing,
              this is done in `burnchains::bitcoin::blocks::BitcoinBlockParser::parse_data`.
+
+             Contract name is prefixed by a single byte size followed by the UTF-8 encoded
```
netrome, line 90: Technically speaking we're only allowing a subset of ascii in the contract name. Although ascii is a subset of UTF-8, hence why we can use utf8 to decode the string.

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -76,11 +78,17 @@ impl PegInOp {
 
             0      2  3                  24                            64       80
```
netrome, line 79: I guess we need an extra byte for the size prefix
```suggestion
            0      2  3                  24                            65       80
```

netrome (changes requested):
Nice stuff. Just need to update the size prefix. Also I would like to see a test case covering the scenario where the size prefix is shorter than the contract name. I assume we should truncate the name in this scenario.

src/chainstate/burn/operations/peg_in.rs
stjepangolemac: Just to confirm, we want to take a byte from the memo and not make the data 81 bytes long?

netrome:
@stjepangolemac I added @MarvinJanssen and @igorsyl for review. I don't know who else to add, since the foundation people are busy with the hard fork work.

src/chainstate/burn/operations/peg_in.rs
netrome: Yes. We only have 80 bytes. The memo is not important and is just a field to absorb any remaining bytes unused by the protocol.

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -100,18 +108,18 @@ impl PegInOp {
...
-
-                QualifiedContractIdentifier::new(standard_principal_data, contract_name.try_into()?)
-                    .into()
+                (contract_id.into(), 21 + bytes_read)
             } else {
-                standard_principal_data.into()
+                (standard_principal_data.into(), 22)
             };
 
+        let memo = data.get(bytes_read..).unwrap_or(&[]).to_vec();
```
netrome, line 121: I don't think we support variable-sized memos. The memo should only be the space unused by the rest of the fields.

src/chainstate/burn/operations/peg_in.rs
stjepangolemac: Confirmed.

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -124,30 +132,24 @@ impl PegInOp {
...
-            }
-            end
-        };
-
-        if from == end_of_non_zero_slice {
-            return None;
+    /// # Errors
+    /// - `Utf8Error` if the contract name is not valid UTF-8
+    /// - `ParseError` if the contract name is not a valid `ContractName`
+    fn read_contract_name(data: &[u8]) -> Result<Option<(ContractName, usize)>, ParseError> {
```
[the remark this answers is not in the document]
stjepangolemac: I will create a type for this.

netrome:
@stjepangolemac @MarvinJanssen noted that we're approaching the wire format for principals. If we just add another byte (indicating wether it's a contract principal or standard) the format will be fully compatible. What do you think about changing to use the wire format directly?

Original comment: https://github.com/Trust-Machines/core-eng/pull/339#discussion_r1194208407

MarvinJanssen (approved):
I think just having principal wire format in the payload will make the Clarity processing much more efficient. 👍

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -124,30 +131,24 @@ impl PegInOp {
...
-            end
-        };
-
-        if from == end_of_non_zero_slice {
-            return None;
+    /// # Errors
+    /// - `Utf8Error` if the contract name is not valid UTF-8
+    /// - `ParseError` if the contract name is not a valid `ContractName`
+    fn read_contract_name(data: &[u8]) -> Result<Option<ContractName>, ParseError> {
+        let size = data[0] as usize;
```
jcnelson, line 140: What's the point of having the size prefix if you're just going to inject `&data[1..41]` into a `ContractName`?  Shouldn't you instead be using `&data[1..(1 + size)]`, with a bounds-check on `size`?

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -124,30 +131,24 @@ impl PegInOp {
...
+
+        if size == 0 {
+            return Ok(None);
         }
 
-        data.get(from..end_of_non_zero_slice)
+        let contract_name_bytes = &data[1..41];
```
jcnelson, line 146: N.B. `data` might not be 41 bytes long.  It could be shorter.

src/chainstate/burn/operations/peg_in.rs
```diff
@@ -423,7 +430,7 @@ mod tests {
         let addr_bytes = test::random_bytes(&mut rng);
         let stx_address = StacksAddress::new(1, addr_bytes.into());
         data.extend_from_slice(&addr_bytes);
-        data.extend_from_slice(&[0; 40]); // Padding contract name
+        data.push(0); // Contract name not provided, so size is 0
```
jcnelson, line 433: You also need to test the following:

* a non-zero size length with no additional bytes for the name (this is invalid)
* a non-zero size length with insufficiently many bytes for the name (this is invalid)
* a non-zero size length with more than enough bytes for the name (this should be fine, but the name should be truncated to the given size, and the remaining bytes should be in the memo)
* a zero-size name with trailing bytes (this should be fine -- these bytes should all be in the memo)

src/chainstate/burn/operations/peg_in.rs
netrome: Sorry just saw this comment now. The size prefix makes it much easier to parse the data in Clarity, as per this discussion https://github.com/Trust-Machines/stacks-sbtc/pull/339#discussion_r1193026002

PR #3719, opened 2023-05-15
Feat: Add SIP-024 sanitization support to Epoch 2.4

kantai:
### Description

This PR adds sanitization support to epoch 2.4, as proposed in SIP-024. This also includes a refactoring of the deserialization routine from recursive to iterative (this improves serialization performance for deep types).


clarity/src/vm/functions/conversions.rs
```diff
@@ -260,7 +266,7 @@ pub fn from_consensus_buff(
     // Perform the deserialization and check that it deserialized to the expected
     // type. A type mismatch at this point is an error that should be surfaced in
     // Clarity (as a none return).
-    let result = match Value::try_deserialize_bytes_exact(&input_bytes, &type_arg) {
+    let result = match Value::try_deserialize_bytes_exact(&input_bytes, &type_arg, false) {
```
pavitthrap, line 269: Might be beneficial to add an explanatory comment everywhere you hard-code `sanitize=false`.

stacks-common/src/types/mod.rs
```diff
@@ -81,6 +81,12 @@ impl StacksEpochId {
     pub fn latest() -> StacksEpochId {
         StacksEpochId::Epoch24
     }
+
+    /// Returns whether or not this Epoch should perform
+    ///  Clarity value sanitization
+    pub fn value_sanitizing(&self) -> bool {
+        self >= &StacksEpochId::Epoch24
```
pavitthrap, line 88: For the sake of the style we agreed upon after the canonicalize types issue, should we do an explicit match for the epoch ID here instead of using a comparison operator?

clarity/src/vm/functions/conversions.rs
kantai: Yep, good idea -- this site should be epoch-switched anyways, so I updated it in 08fd9f5

stacks-common/src/types/mod.rs
kantai: Yep, updated.

clarity/src/vm/types/serialization.rs
```diff
@@ -452,275 +544,444 @@ impl Value {
...
             };
 
-            assert!(
-                expect_size as u64 >= bytes_read,
-                "Deserialized more bytes than expected size during deserialization. Expected size = {}, bytes read = {}, type = {}",
-                expect_size,
-                bytes_read,
-                expected_type,
-            );
+            if expect_size as u64 > bytes_read {
```
pavitthrap, line 547: Before it was an assertion, so I think here it should be bytes_read  > expect_size as u64

clarity/src/vm/types/serialization.rs
```diff
@@ -452,275 +544,444 @@ impl Value {
...
-            TypePrefix::StringUTF8 => {
-                let mut total_len = [0; 4];
-                r.read_exact(&mut total_len)?;
-                let total_len = BufferLength::try_from(u32::from_be_bytes(total_len))?;
+                    if len > 0 {
+                        let items = Vec::with_capacity(expected_len as usize);
+                        let first_key = ClarityName::deserialize_read(r)?;
+                        let next_sanitize = sanitize
+                            && tuple_type
+                                .map(|tt| tt.field_type(&first_key).is_none())
```
pavitthrap, line 777: It would be helpful to add a comment to explain this logic: 
`.map(|tt| tt.field_type(&first_key).is_none())`

clarity/src/vm/types/serialization.rs
```diff
@@ -452,275 +544,444 @@ impl Value {
...
                 }
+            }?;
 
-                value
+            let mut finished_item = Some(item);
+            while let Some(item) = finished_item.take() {
+                let stack_bottom = if let Some(stack_item) = stack.pop() {
+                    stack_item
+                } else {
+                    // this should be unreachable!
```
pavitthrap, line 866: I think it would be good to put a log statement here

clarity/src/vm/types/serialization.rs
```diff
@@ -882,18 +1145,123 @@ impl Write for WriteCounter {
...
+                let mut tuple_data_map = tuple_data.data_map;
+                let mut did_sanitize_children = false;
+                for (key, expect_key_type) in tt.get_type_map().iter() {
+                    let field_data = tuple_data_map.remove(key)?;
+                    let (sanitized_field, did_sanitize) =
+                        Self::sanitize_value(epoch, expect_key_type, field_data)?;
+                    sanitized_tuple_entries.push((key.clone(), sanitized_field));
+                    did_sanitize_children = did_sanitize_children || did_sanitize;
+                }
+                let did_sanitize_tuple = did_sanitize_children || (tt.len() != tuple_data_len);
```
pavitthrap, line 1214: For the list type, you checked the len earlier (and also returned None). 

Perhaps something analagous to 
```
if l.len() > lt.get_max_len() {
                     return None;
                 }
```

makes sense in the tuple case.

pavitthrap (approved):

clarity/src/vm/types/serialization.rs
```diff
@@ -452,275 +544,444 @@ impl Value {
...
-                };
 
-                let data = Value::inner_deserialize_read(r, expect_contained_type, depth + 1)?;
-                let value = if committed {
-                    Value::okay(data)
-                } else {
-                    Value::error(data)
+            let expected_type = stack
+                .last()
+                .expect("FATAL: stack.last() should always be some() because of loop condition")
```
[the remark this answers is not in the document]
kantai: `stack.last()`'s type is `Option<&DeserializeStackItem>` -- it will always be `Some(&DeserializeStackItem)` because the loop condition is:

```rust
while !stack.is_empty() {
...
}
```

`expected_type`, however _could_ be None, in exactly the situation (and some others) you mention.

obycode (approved):
Everything looks good to me! Let's go 🚀

PR #3721, opened 2023-05-15
Fix/increase delegation amount

friedger:
### Description
In epoch 2.4, pox-3 should allow to increase the delegation amount so that users can compound rewards.

This PR
* allows users to change all delegation data (after user revoke permission)

The required assertions to prevent swimming in two pools at the same time or to change the pool without a cool down cycle are already present in pox-3.

### DIFF
--- a/src/chainstate/stacks/boot/pox-3.clar
+++ b/src/chainstate/stacks/boot/pox-3.clar
@@ -658,6 +658,8 @@
       ;; delegate-stack-* functions assert that
       ;; 1. users can't swim in two pools at the same time.
       ;; 2. users can't switch pools without cool down cycle.
+      ;;    Other pool admins can't increase or extend.
+      ;; 3. users can't join a pool while already directly stacking.
 
       ;; pox-addr, if given, must be valid
       (match pox-addr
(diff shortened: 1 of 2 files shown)


src/chainstate/stacks/boot/pox-3.clar
```diff
@@ -653,9 +653,11 @@
...
                 (err ERR_STACKING_PERMISSION_DENIED))
 
-      ;; tx-sender principal must not be stacking
-      (asserts! (is-none (get-stacker-info tx-sender))
-        (err ERR_STACKING_ALREADY_STACKED))
+      ;; delegate-stx no longer requires the delegator to not currently
+      ;; be stacking.
+      ;; delegate-stack-* functions assert that
+      ;; 1. users can't swim in two pools at the same time.
+      ;; 2. users can't switch pools without cool down cycle.
```
obycode, line 660: Can we also add that a directly stacked user cannot join a pool without a cool down cycle and vice versa? Then we will also need to add the missing test cases around these situations.

obycode (approved):
I reviewed these cases with @friedger and I agree that removing this assertion should enable the desired behavior without adding complexity, since the other checks needed are already in place in the `delegate-stack-*` functions.

kantai (approved):

friedger:
Unit tests for pox-3 passed.
Other tests failed, maybe because it is a PR from a different repo?

obycode:
> Unit tests for pox-3 passed.
> Other tests failed, maybe because it is a PR from a different repo?

Looks like just port binding errors, should be intermittent. I just kicked off a re-run.

friedger:
CI is happy :-)

PR #3722, opened 2023-05-15
add more test coverage for epoch 2.4

pavitthrap:
### Description
Ensures that epoch 2.4 case is added to various places in Clarity VM tests as well as in trait testing module.


kantai (approved):
LGTM -- looks like it just needs to run `cargo fmt`

obycode (approved):
lgtm

PR #3726, opened 2023-05-23
chore: output logs to stdout instead of stderr.

diwakergupta:
### Description

Output logs to `stdout` instead of `stderr`.

Rationale: most tools expect programs to output to `stderr`.
`stderr` should be used when there's an actual error (e.g. the program
fails to start because of incorrect options). The current behavior
causes surprises and frustrations (e.g. trying to `tee` the output of
stacks-node to a file only to realize the file's empty).

### DIFF
--- a/stacks-common/src/util/log.rs
+++ b/stacks-common/src/util/log.rs
@@ -205,7 +205,7 @@ fn make_json_logger() -> Logger {
                       }),
     );
 
-    let drain = Mutex::new(slog_json::Json::default(std::io::stdlog())).map(slog::Fuse);
+    let drain = Mutex::new(slog_json::Json::default(std::io::stdout())).map(slog::Fuse);
     let filtered_drain = slog::LevelFilter::new(drain, get_loglevel()).fuse();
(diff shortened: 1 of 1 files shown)


jcnelson:
NACK. Stderr is for diagnostic information, which definitely includes logs.

We should not push a policy choice into the implementation.

diwakergupta:
> NACK. Stderr is for diagnostic information, which definitely includes logs. We should not push a policy choice into the implementation.

Fine, this is not my hill. I'll add a note to the README instead, because this behavior is not obvious.

I'll close this PR and leave you with some comments:

* the current implementation is already encoding a policy choice
* there's nothing absolutely better about this choice: it's merely a preference. There are other, valid, preferences (e.g. https://12factor.net/logs) in use by large platforms (like Heroku)
* tons of other programs (including `bitcoind`) emit logs to stdout by default (if writing to a term)

igorsyl:
Can we add this feature using an env var toggle? cc @jcnelson

jcnelson:
No. Use `2>&1` if you want to redirect stderr to stdout.

PR #3729, opened 2023-05-26
Implement new RPC endpoint: health check

### LINKED ISSUE #1229: RESTful API for querying blockchain status

pavitthrap:
### Description
- Implements v2/health endpoint + tests.  A node is considered healthy if it is caught up with its bootstrap peers.
- Makes a change to get_random_neighbors: it now checks that the obtained peers satisfy `(peer_version & 0x000000ff) <= network_epoch` (as opposed to `(peer_version & 0x000000ff) >= network_epoch`). 

### Applicable issues
- fixes #1229

### Feedback needed 
- In `get_valid_initial_neighbors` I am passing in the peer_version. Alternatively I could pass in the network_epoch. 
- In `handle_get_health` in rpc.rs, I need feedback on how I computed `last_processed_burn_height` and `burnchain_height`. The former should be the last processed burn height (so less than the actual burn chain tip if ibd=true. The latter should be the actual current height of the burnchain. I want to mimic how ibd is calculated in `pox_sync_wait`.

### Checklist
- [x] Changelog is updated


src/net/http.rs
obycode, line 1: Please add `/v2/health` to the various tests in this file.

obycode (commented):
Looking good, but I have a few comments/questions.

src/net/mod.rs
```diff
@@ -1654,6 +1664,8 @@ pub enum HttpResponseType {
     ServerError(HttpResponseMetadata, String),
     ServiceUnavailable(HttpResponseMetadata, String),
     Error(HttpResponseMetadata, u16, String),
+    GetHealthError(HttpResponseMetadata, serde_json::Value),
```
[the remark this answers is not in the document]
pavitthrap: Yes, I thought of doing that. 

The reason I didn't is because the function that generates the responses is generic (`parse_error`). 

```
let mut error_text = String::new();
        let mut json_val: serde_json::Value = Default::default();
        fd.read_to_string(&mut error_text)
            .map_err(net_error::ReadError)?;
        if preamble.content_type == HttpContentType::JSON {
            let json_result = serde_json::from_str(&error_text.clone());
            json_val = match json_result {
                Ok(val) => val,
                Err(serde_json::Error { .. }) => {
                        "Invalid response: unable to deserialize json response".to_string(),
                    ))
                }
            }
        }

        let md = HttpResponseMetadata::from_preamble(request_version, preamble);
        let resp = match preamble.status_code {
            400 => HttpResponseType::BadRequest(md, error_text),
            401 => HttpResponseType::Unauthorized(md, error_text),
            402 => HttpResponseType::PaymentRequired(md, error_text),
            403 => HttpResponseType::Forbidden(md, error_text),
            404 => HttpResponseType::NotFound(md, error_text),
            500 => HttpResponseType::ServerError(md, error_text),
            503 => HttpResponseType::ServiceUnavailable(md, error_text),
            512 => HttpResponseType::GetHealthError(md, json_val),
            513 => HttpResponseType::GetHealthNoDataError(md, error_text),
            _ => HttpResponseType::Error(md, preamble.status_code, error_text),
        };
        Ok(resp)
```

It seemed like better style not to match specifically on `HttpResponseType::GetHealthError`. I see the argument for doing that though, let me know what you think.

src/net/http.rs
pavitthrap: Any tests in particular? The endpoint has tests in rpc.rs. 

Also a quick search of other endpoints didn't reveal any tests that cover all the endpoints in this file.

src/net/mod.rs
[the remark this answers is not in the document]
obycode: That makes sense to me. If we end up with more json objects, then we can refactor this later.

src/net/http.rs
pavitthrap: I see. Most of those tests test the http protocol (rather than a specific endpoint) - so things like too many headers, too large of a header, etc. 

I added to one test in which errors are deserialized.

src/net/db.rs
```diff
@@ -1275,7 +1275,7 @@ impl PeerDB {
 
         if always_include_allowed {
             // always include allowed neighbors, freshness be damned
-            let allow_qry = "SELECT * FROM frontier WHERE network_id = ?1 AND denied < ?2 AND (allowed < 0 OR ?3 < allowed) AND (peer_version & 0x000000ff) >= ?4".to_string();
+            let allow_qry = "SELECT * FROM frontier WHERE network_id = ?1 AND denied < ?2 AND (allowed < 0 OR ?3 < allowed) AND (peer_version & 0x000000ff) <= ?4".to_string();
```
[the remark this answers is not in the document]
pavitthrap: I changed it and added an extra check (copying the logic in `has_acceptable_epoch`)

src/net/http.rs
obycode: Ok got it. Thanks.

src/net/db.rs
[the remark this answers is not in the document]
obycode: Thanks! That looks good to me. Would it be possible to add a test that would've caught the problem with the previous logic?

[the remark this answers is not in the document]
pavitthrap: Yup! Added in `e29874`

obycode (approved):
Thanks!

src/net/db.rs
```diff
@@ -1275,12 +1276,16 @@ impl PeerDB {
 
         if always_include_allowed {
             // always include allowed neighbors, freshness be damned
-            let allow_qry = "SELECT * FROM frontier WHERE network_id = ?1 AND denied < ?2 AND (allowed < 0 OR ?3 < allowed) AND (peer_version & 0x000000ff) >= ?4".to_string();
+            // the peer_version check mirrors the check in `has_acceptable_epoch`:
+            //    (my_epoch <= peer_epoch) OR (curr_epoch <= peer_epoch)
+            let allow_qry = "SELECT * FROM frontier WHERE network_id = ?1 AND denied < ?2 AND \
+            (allowed < 0 OR ?3 < allowed) AND (?4 <= (peer_version & 0x000000ff) OR ?5 <= (peer_version & 0x000000ff))".to_string();
```
jcnelson, line 1282: I'm not sure this is what we want here.  First, it seems logically incorrect -- I think you'd want `(?4 & 0x000000ff) <= (peer_version & 0x000000ff)`; otherwise this clause is always false because `peer_version` includes the network's major and minor version numbers in higher bits.  Second, it's always true that `(curr_network_epoch as u32) <= (peer_version & 0x000000ff)` because the node's `peer_version` reflects the highest epoch supported by the software.  The `has_acceptable_epoch` code performs both checks because the first check is a short-circuit that saves the code from performing the slightly-more-expensive second check.

src/net/rpc.rs
```diff
@@ -2384,6 +2390,110 @@ impl ConversationHttp {
...
+        }
+
+        let ibd = infer_initial_burnchain_block_download(
+            &network.burnchain,
+            last_processed_burn_height,
+            burnchain_height,
+        );
+
+        // get max block height amongst bootstrap nodes
+        let response = if let Some(block_stats) = block_stats {
```
jcnelson, line 2440: I think this code block should be part of the `InvState` state machine.  The code here should just call it in order to get the maximum block height reported by a given set of nodes.

jcnelson (commented):
I notice that the two error paths for this endpoint are using unallocated HTTP status codes in addition to an application-level error message body.  I think perhaps we should consider using a well-defined HTTP status code and perhaps make the message body more detailed (e.g. inform the caller to check `/v2/neighbors` to diagnose why the node can't report health data).

pavitthrap:
> I notice that the two error paths for this endpoint are using unallocated HTTP status codes in addition to an application-level error message body. I think perhaps we should consider using a well-defined HTTP status code and perhaps make the message body more detailed (e.g. inform the caller to check `/v2/neighbors` to diagnose why the node can't report health data).

I decided to go ahead with the unallocated HTTP status codes based on a discussion on the old PR for this same issue. @wileyj  do you have any opinions on how to proceed here?

src/net/db.rs
pavitthrap: Regarding your second comment: peer_version is a neighbor's version, so I don't see why (curr_network_epoch <= peer_epoch) would necessarily always be true. 

An example:
my_epoch = 2.05 
curr_epoch = 2.0 
peer_epoch = 2.0

Because of the second check (curr_epoch <= peer_epoch), the peer would be found to have an acceptable epoch prior to the start of Epoch 2.05, but after the node upgraded.

wileyj:
> > I notice that the two error paths for this endpoint are using unallocated HTTP status codes in addition to an application-level error message body. I think perhaps we should consider using a well-defined HTTP status code and perhaps make the message body more detailed (e.g. inform the caller to check `/v2/neighbors` to diagnose why the node can't report health data).
> 
> I decided to go ahead with the unallocated HTTP status codes based on a discussion on the old PR for this same issue. @wileyj do you have any opinions on how to proceed here?

i see what you're doing - i tend to agree with @jcnelson here in that we shouldn't define unused http status codes (even though the endpoint errors don't specifically line up to a defined status). 

I think a generic `500` error for a non-successful request would be more than sufficient here - the request either returns some data with a 200 that can be evaluated, or it has failed and throws a 500

src/net/db.rs
jcnelson: Ah, yes, you're right about my second comment.

The first part needs to be addressed though.

src/net/db.rs
```diff
@@ -1325,14 +1331,53 @@ impl PeerDB {
...
+        burn_block_height: u64,
+    ) -> Result<Vec<Neighbor>, db_error> {
+        // UTC time
+        let now_secs = util::get_epoch_time_secs();
+
+        // the peer_version check mirrors the check in `has_acceptable_epoch`:
+        //    (my_epoch <= peer_epoch) OR (curr_epoch <= peer_epoch)
+        let query = "SELECT * FROM frontier WHERE initial = 1 AND (allowed < 0 OR ?1 < allowed) \
+         AND network_id = ?2 AND denied < ?3 AND ?4 < expire_block_height \
+         AND (?5 <= (peer_version & 0x000000ff) OR ?6 <= (peer_version & 0x000000ff))"
```
jcnelson, line 1367: Just flagging this here too -- I think you need `((?5 & 0x000000ff) <= (peer_version & 0x000000ff))`.

src/net/db.rs
```diff
@@ -1772,49 +1832,73 @@ mod test {
         .unwrap();
 
         // epoch 2.0
-        let n5 =
-            PeerDB::get_random_neighbors(db.conn(), 0x9abcdef0, 0x00, 5, 23455, false).unwrap();
+        let n5 = PeerDB::get_random_neighbors(db.conn(), 0x9abcdef0, 0x00, 0x00, 5, 23455, false)
```
jcnelson, line 1835: In these tests, `peer_version` would be something like `0x18000009` (for epoch 2.4).

src/net/inv.rs
```diff
@@ -1143,6 +1143,37 @@ impl InvState {
         list
     }
 
+    pub fn get_max_height_of_neighbors(&self, neighbors: &Vec<Neighbor>, ibd: bool) -> Option<u64> {
```
jcnelson, line 1146: Nit: missing docstring

src/net/mod.rs
```diff
@@ -1356,6 +1358,12 @@ pub struct AttachmentPage {
     pub inventory: Vec<u8>,
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
```
jcnelson, line 1361: Nit: missing docstring

src/net/mod.rs
```diff
@@ -1644,14 +1653,15 @@ pub enum HttpResponseType {
...
+    GetHealth(HttpResponseMetadata, GetHealthResponse),
     // peer-given error responses
     BadRequest(HttpResponseMetadata, String),
     BadRequestJSON(HttpResponseMetadata, serde_json::Value),
     Unauthorized(HttpResponseMetadata, String),
     PaymentRequired(HttpResponseMetadata, String),
     Forbidden(HttpResponseMetadata, String),
     NotFound(HttpResponseMetadata, String),
-    ServerError(HttpResponseMetadata, String),
+    ServerError(HttpResponseMetadata, serde_json::Value),
```
jcnelson, line 1664: :+1: thanks for making this change!

src/net/mod.rs
```diff
@@ -2091,6 +2102,27 @@ pub trait Requestable: std::fmt::Display {
     fn make_request_type(&self, peer_host: PeerHost) -> HttpRequestType;
 }
 
+// TODO: DRY up from PoxSyncWatchdog
```
jcnelson, line 2105: It's fine if you want to address this in this PR :)

src/net/rpc.rs
```diff
@@ -2384,6 +2409,91 @@ impl ConversationHttp {
...
+    /// Handle a GET health.
+    /// A node is considered healthy if it is up to the max height of its bootstrap nodes.
+    /// This endpoint also returns what percent of the max height the node height is at.
+    fn handle_get_health<W: Write>(
+        http: &mut StacksHttp,
+        fd: &mut W,
+        req: &HttpRequestType,
+        network: &PeerNetwork,
+        network_id: u32,
+        peerdb: &PeerDB,
```
jcnelson, line 2421: I'm curious why these two arguments (`peerdb` and `inv_state_opt`) are explicit arguments? This data already lives within the `network`.

CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ and this project adheres to the versioning scheme outlined in the [README.md](RE
   from the source code are now attached to the `SymbolicExpression` nodes. This
   will be useful for tools that use the Clarity library to analyze and
   manipulate Clarity source code, e.g. a formatter.
+- Added a new RPC endpoint to query the node's health. A node is considered healthy if it
+  is caught up with its bootstrap peers. The endpoint returns 
+  200 if it is synced to its peers' chaintips, 513 if the query was unsuccessful (no viable
```
jcnelson, line 18: Looks like this changelog entry is out of sync with the code's current behavior.

jcnelson (approved):
LGTM!  Only minor things need to be done before this can be merged.  Thanks for getting this over the finish line, @pavitthrap!

src/net/mod.rs
pavitthrap: I think that might not be straightforward given the structure of the codebase. Let me know if it is straightforward though.

kantai:
Closing this PR for now. It's stale, and would require some effort to revive. If someone wants to take it on, they can always revive it later.

Issue #3720, opened 2023-05-15
Change Clarity case testing 


pavitthrap:
Right now, we use the `rstest` library to enable us to create templates. These templates allow us to run the same test for different clarity versions & Stacks epochs. Ex:
```
#[template]
#[rstest]
#[case(ClarityVersion::Clarity1, StacksEpochId::Epoch2_05)]
#[case(ClarityVersion::Clarity1, StacksEpochId::Epoch21)]
#[case(ClarityVersion::Clarity2, StacksEpochId::Epoch21)]
...
fn test_edge_counting_runtime_template(
    #[case] version: ClarityVersion,
    #[case] epoch: StacksEpochId,
) {
}
```

When we add new epochs, there's no verification check to ensure the new epoch cases are added to each of these `rstest` uses. We currently manually add it. We should make a change such that this is checked programatically.

kantai:
Completely agreed. I think this should be paired with a general rstest update (the version we use makes use of some deprecated rust features) and cleanup: we should not need to redefine the template continuously as we do now, and should be able to just define the template once and then import the macro.

PR #3731, opened 2023-05-30
Chore: rstest improvements

kantai:
### Description

This PR makes a few improvements related to rstest usage in the codebase.

1. It updates the `rstest` and `rstest-reuse` dependencies (these dependencies were generating a future-incompatibilities warning from rustc).
2. It defines two rstest templates which are reused throughout the tests, rather than defining templates in every module in the testing codebase.
3. It creates the rstest templates using a `macro_rules!` macro. This macro includes a test function which includes a match statement. This test function and match statement will trigger a compile time error if an epoch case (in the epochs template) or if an (epoch, clarity_version) pair are unspecified in the template.
4. It explicitly uses a `/tmp/` directory path rather than the rust `temp_dir` method. This fixes an issue running the tests concurrently in some environments (notably, my development machine!)

### Applicable issues
- Addresses #3720 

This PR only alters testing code.

It's currently based off `master`, but I can rebase on `develop` as soon as `master -> develop` merges.


pavitthrap (approved):
lgtm

clarity/src/vm/tests/contracts.rs
```diff
@@ -889,7 +926,14 @@ fn test_aborts(owned_env: &mut OwnedEnvironment) {
...
+fn test_factorial_contract(
+    version: ClarityVersion,
+    epoch: StacksEpochId,
+    mut env_factory: MemoryEnvironmentGenerator,
+) {
+    let mut owned_env = env_factory.get_env(epoch);
+
     let mut placeholder_context = ContractContext::new(
         QualifiedContractIdentifier::transient(),
         ClarityVersion::Clarity2,
```
obycode, line 939: Here and several other places in this file, there are tests where you've added `test_clarity_versions` but then didn't use `version`.

clarity/src/vm/tests/contracts.rs
obycode, line 1: Any reason `test_get_block_info_eval` didn't get updated to use `test_clarity_versions`?

obycode (commented):
This is great! Thanks @kantai. Just a couple comments.

clarity/src/vm/tests/contracts.rs
kantai: Corrected in [7c10037](https://github.com/stacks-network/stacks-blockchain/pull/3731/commits/7c10037a7d539fb9f1cc4f9da990cd74fc115d3b). I added a `warn(unused)` lint definition to the tests module, and fixed all those warnings, then removed the label again, because it interacts pretty unsatisfactorily with the workspace build (it warns on unused items when the `testing` build flag is set, which generates a ton of invalid warns -- there's probably a way to fix that, but I'll save that for another time).

obycode (approved):
LGTM!

kantai:
There looks to be two CI issues on this PR (one of which is blocking merging):

1. `create-release` is a required status check, but this PR isn't running that action.
2. It seems like the integration tests workflow only runs when the PR is opened, but not since updating the PR. That's actually not much of an issue for this PR in particular, but it's an issue that should be addressed.

I'll open an issue for this, but just highlighting here as well, @wileyj

PR #3738, opened 2023-06-06
Fix/utxo lookup

### LINKED ISSUE #3733: UTXO lookup uses parent block height, but should use current block height instead
**Describe the bug**

The bitcoin_regtest_controller used to submit block commit operations looks up UTXOs in a buggy way which can lead to the miner submitting multiple concurrent commits in certain circumstances (we're seeing this on testnet due to the deep testnet reorg).

https://github.com/stacks-network/stacks-blockchain/blob/master/testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs#L1158-L1165

This should use the current burn block height instead.

wileyj:
### Description
During the planned testnet re-org (https://forum.stacks.org/t/stacks-2-4-and-testnet-reorg/15027), it was discovered that attempting to mine a block on a very deep fork caused the miner to submit several concurrent utxo's per block to the burnchain, ultimately never mining a block. This PR changes the logic to use the current burnchain tip rather than the parent block. 

### Applicable issues
- fixes #3733


pavitthrap (approved):

CHANGELOG.md
```diff
@@ -24,6 +24,12 @@ and this project adheres to the versioning scheme outlined in the [README.md](RE
 - Added config setting `burnchain.wallet_name` which addresses blank wallets no 
   longer being created by default in recent bitcoin versions. Fixes issue #3596
 
+## [2.4.0.0.1]
```
obycode, line 27: Since this is framed as 2.4.0.0.1, should it be opened against master?

obycode (approved):
Looks good to me otherwise.

jbencin (approved):

CHANGELOG.md
wileyj: good point - i'll remove the version section here and add to the correct changelog location

obycode (approved):
👍

PR #3735, opened 2023-06-05
Refactor `u32::max_value()` with `u32::MAX`

Acaccia:
### Description
I changed all occurrences of deprecated `u32::max_value()`.
While I was at it, I changed the same deprecated method for `u64` and `i32` hiding there.

### Applicable issues
- fixes #3682 

### Checklist
- [x] Test coverage for new or modified code paths
- [ ] Changelog is updated

PR #3739, opened 2023-06-06
Chore: replace u32::max_value() with u32::MAX

pavitthrap:
### Description
This is a chore PR to merge PR #3735.

### DIFF
--- a/clarity/src/vm/docs/mod.rs
+++ b/clarity/src/vm/docs/mod.rs
@@ -2764,7 +2764,7 @@ mod test {
             Some(StacksEpoch {
                 epoch_id: StacksEpochId::Epoch21,
                 start_height: 0,
-                end_height: u64::max_value(),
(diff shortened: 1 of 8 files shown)


netrome (approved):
I'm curious, why do we need an additional PR to merge this one? I.e. why couldn't https://github.com/stacks-network/stacks-blockchain/pull/3739 be merged directly to `develop`?

obycode (approved):

obycode:
> I'm curious, why do we need an additional PR to merge this one? I.e. why couldn't #3739 be merged directly to `develop`?

It's because the two checks that build and publish docker images fail when PRs come from forks, since the keys are not available to the CI. We discussed this before and I think the consensus is that these checks should not be required, or should only run in certain scenarios, but until that is resolved, we have to do this to workaround it.

PR #3754, opened 2023-06-12
chore: added burnchain sync percentage

### LINKED ISSUE #3557: [observability] add sync percentage during boot-up
As @wileyj mentioned in the weekly blockchain sync, it would be helpful to add a sync percentage to provide some visibility of the progress of the block download (Bitcoin has something like this).

pavitthrap:
### Description
When the node is in initial block download, the percentage of the burnchain sync will now print. 

### Applicable issues
- fixes #3557

### DIFF
--- a/testnet/stacks-node/src/run_loop/neon.rs
+++ b/testnet/stacks-node/src/run_loop/neon.rs
@@ -1082,7 +1082,7 @@ impl RunLoop {
             let ibd = match self.get_pox_watchdog().pox_sync_wait(
                 &burnchain_config,
                 &burnchain_tip,
-                Some(remote_chain_height),
+                remote_chain_height,
                 num_sortitions_in_last_cycle,
             ) {
                 Ok(ibd) => ibd,
@@ -1091,6 +1091,16 @@ impl RunLoop {
                     continue;
                 }
             };
+
+            // print burnchain sync percentage
+            if ibd {
+                let percent: f64 =
+                    burnchain_tip.block_snapshot.block_height as f64 / remote_chain_height as f64;
+                debug!("Burnchain sync percentage";
+                       "percent" => %percent,
+                       "local_tip_height" => %burnchain_tip.block_snapshot.block_height,
+                       "remote_tip_height" => %remote_chain_height);
+            }
 
             // will recalculate this in the following loop
             num_sortitions_in_last_cycle = 0;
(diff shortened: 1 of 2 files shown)


obycode (approved):
LGTM!

zone117x (commented):
Would it be possible to expose this in the /v2/info endpoint and/or included in one of the emitted events? This feature would be pretty useful for applications interacting with the stacks-node.

wileyj:
> Would it be possible to expose this in the /v2/info endpoint and/or included in one of the emitted events? This feature would be pretty useful for applications interacting with the stacks-node.

@pavitthrap to confirm, but it may be a slower endpoint than `/v2/info` should be expected to be.

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -1092,6 +1092,16 @@ impl RunLoop {
...
+            // print burnchain sync percentage
+            if ibd {
+                let percent: f64 =
+                    burnchain_tip.block_snapshot.block_height as f64 / remote_chain_height as f64;
+                debug!("Burnchain sync percentage";
+                       "percent" => %percent,
+                       "local_tip_height" => %burnchain_tip.block_snapshot.block_height,
+                       "remote_tip_height" => %remote_chain_height);
+            }
+
```
kantai, line 1104: Can this be incorporated into the `debug!` message on line `1113`, e.g.:

```rust
            debug!(
                "Runloop: Downloading burnchain blocks up to reward cycle #{} (height {})",
                burnchain_config
                    .block_height_to_reward_cycle(target_burnchain_block_height)
                    .expect("FATAL: target burnchain block height does not have a reward cycle"),
                target_burnchain_block_height;
                "total_burn_sync_percent" => %percent,
                "local_burn_height" => burnchain_tip.block_snapshot.block_height,
                "remote_tip_height" => remote_chain_height
            );
```

That way, this change doesn't add any additional log lines.

kantai (changes requested):
LGTM, but I'd prefer incorporating this information in an existing log message.

kantai (approved):

pavitthrap:
> Would it be possible to expose this in the /v2/info endpoint and/or included in one of the emitted events? This feature would be pretty useful for applications interacting with the stacks-node.

I can explore this in a different PR.

PR #3773, opened 2023-06-29
Update github workflow

### LINKED ISSUE #3755: CI: Integration tests run on PR opening, but not on updates to the PR
It seems like the integration tests workflow only runs when the PR is opened, but not since updating the PR.

You can see this in action in my rstest PR (#3731)

I *think* it should run on every update (and cancel prior runs), but if that's too much, we should see if we could do something like run-on-approval or some other kind of manual process.

Tagging @wileyj for visibility.

pavitthrap:
### Description
Add tag to make sure that stacks blockchain tests workflow runs when pull requests are opened, updated, or re-opened. 

### Applicable issues
- fixes #3755

Can refer to docs [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) if more info is desired.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 ## Only run when:
 ##   - manually triggered
-##   - PR's are (re)opened
+##   - PR's are (re)opened or updated
 ##   - push to master (i.e. merge develop -> master)
 
 on:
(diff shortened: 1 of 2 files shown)


jcnelson:
@wileyj is this still relevant?

wileyj:
this can be closed - the change proposed here is not something we currently *want* to do since it would trigger a workflow on any PR update (i.e. a PR comment). 
the work i'm doing will address the concerns here.

PR #3776, opened 2023-07-03
Events/block reward

### LINKED ISSUE #3621: [Event Stream] Report matured miner disbursement and application of PoisonMicroblocks
Right now, the matured block reward disbursement process is opaque.  The only evidence that a miner gets paid their block reward is that their account balance increases, and a debug log message gets emitted.  Instead, this information should be reported on the node's event stream, so that miners can see when their block rewards are materialized.

In addition, the system should report whether or not the block reward was slashed by a `PoisonMicroblock` transaction (and if so, which one).  This information is currently only visible in debug logs, and must be reconstructed by trawling the chain state and correlating them to offending miners.

pavitthrap:
### Description
Emits a block reward event. 

### Applicable issues
- fixes #3621

### DIFF
--- a/clarity/src/vm/database/clarity_db.rs
+++ b/clarity/src/vm/database/clarity_db.rs
@@ -1051,6 +1051,8 @@ impl<'a> ClarityDatabase<'a> {
         reporter: &StandardPrincipalData,
         seq: u16,
     ) -> Result<()> {
+        // TODO(eo): snake in the Txid and emit an event here with txid, height, reporter, and seq
+
         let key = ClarityDatabase::make_microblock_poison_key(height);
         let value = Value::Tuple(
             TupleData::from_data(vec![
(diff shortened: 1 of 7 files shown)


zone117x (commented):
The `/new_block` event already contains a `matured_miner_rewards` property with the following fields:
```ts
interface matured_miner_reward {
    from_index_consensus_hash: string;
    from_stacks_block_hash: string;
    /** STX principal */
    recipient: string;
    /** STX principal (available starting in Stacks 2.1) */
    miner_address: string | null;
    /** String quoted micro-STX amount. */
    coinbase_amount: string;
    /** String quoted micro-STX amount. */
    tx_fees_anchored: string;
    /** String quoted micro-STX amount. */
    tx_fees_streamed_confirmed: string;
    /** String quoted micro-STX amount. */
    tx_fees_streamed_produced: string;
  }
```

AFAIK, the Hiro API doesn't need the new event data added in this PR. The referenced issue from Jude mentions including information about slashed rewards from poison-microblocks, which sounds helpful but we haven't seen any requests for that (and also microblocks are going away in Nakamoto anyway).

saralab:
Don't need this anymore

PR #3779, opened 2023-07-06
The new version of Regex (1.9.0) is not compatible with 1.8

sergey-shandar:
### Description

The new version of Regex (1.9.0) is not compatible with 1.8 if we use `regex::internal::*` See

https://github.com/Trust-Machines/stacks-sbtc/actions/runs/5466661304/jobs/9951913250

And the fix:

https://github.com/Trust-Machines/stacks-sbtc/pull/522

### Applicable issues
- fixes #

### Checklist
- [ ] Changelog is updated

### DIFF
--- a/clarity/Cargo.toml
+++ b/clarity/Cargo.toml
@@ -23,7 +23,7 @@ rand_chacha = "=0.2.2"
 serde = "1"
 serde_derive = "1"
 serde_stacker = "0.1"
-regex = "~1.8"
+regex = "1"
 lazy_static = "1.4.0"
 integer-sqrt = "0.1.3"
 slog = { version = "2.5.2", features = [ "max_level_trace" ] }
--- a/clarity/Cargo.toml
+++ b/clarity/Cargo.toml
@@ -23,7 +23,7 @@ rand_chacha = "=0.2.2"
 serde = "1"
 serde_derive = "1"
 serde_stacker = "0.1"
-regex = "1"
+regex = "~1.8"
 lazy_static = "1.4.0"
 integer-sqrt = "0.1.3"
 slog = { version = "2.5.2", features = [ "max_level_trace" ] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,9 +85,9 @@ checksum = "739f4a8db6605981345c5654f3a85b056ce52f37a39d34da03f25bf2151ea16e"
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
+version = "1.0.2"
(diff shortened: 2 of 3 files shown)


kantai:
Thanks for finding this @sergey-shandar --

I'd suggest just deleting that line from `costs/mod.rs` instead:

```
diff --git a/clarity/src/vm/costs/mod.rs b/clarity/src/vm/costs/mod.rs
index ee9a1a6eb..c70ab9600 100644
--- a/clarity/src/vm/costs/mod.rs
+++ b/clarity/src/vm/costs/mod.rs
@@ -18,7 +18,6 @@ use std::collections::{BTreeMap, HashMap};
 use std::convert::{TryFrom, TryInto};
 use std::{cmp, fmt};
 
-use regex::internal::Exec;
 use rusqlite::types::{FromSql, FromSqlResult, ToSql, ToSqlOutput, ValueRef};
 use serde::{Deserialize, Serialize};
```

The `Exec` method isn't used anywhere, so we can just remove that import and then we don't need to version pin `regex`.

clarity/Cargo.toml
```diff
@@ -23,7 +23,7 @@ rand_chacha = "=0.2.2"
 serde = "1"
 serde_derive = "1"
 serde_stacker = "0.1"
-regex = "1"
+regex = "~1.8"
```
jcnelson, line 26: Please avoid using tilde, and use a precise version.  We want all nodes everywhere to have the exact same code.

sergey-shandar: I though we are using [Cargo.lock](https://github.com/stacks-network/stacks-blockchain/blob/master/Cargo.lock) for this.

jcnelson:
>  I still need clarification about the versioning strategy in the repo.

Just put the full version. Humans look at the cargo.toml file to determine
dependencies, not the cargo.lock.  The full version removes ambiguity.

jcnelson:
> This PR doesn't touch the Cargo.toml at all right now

Ah I see that now. Will approve once I'm not AFK (give me an hour).

sergey-shandar:
@kantai is it ok that the PR is open against the `master` branch? Do we sync changes from the master to the `develop` and `next` branches?

kantai:
> @kantai is it ok that the PR is open against the master branch?

Yes -- for changes that don't touch any code (e.g., docs only, github workflows, etc.) it's okay to merge to master. This PR, because it just removes an unused import, is okay.

> Do we sync changes from the master to the develop and next branches?

Yes, changes to master should always be merged back into `develop` and `next`. There isn't an automated process for this in the github repository.

jcnelson (approved):

PR #3780, opened 2023-07-06
Fuzz Testing: Clarity value sanitization and serialization

kantai:
### Description

This PR adds a `cargo fuzz` target that instruments and makes some assertions about the 2.4-enabled clarity value sanitization as well as existing serialization and deserialization routines.

Unlike previous fuzzing PRs, this one does not need to make any changes in the `clarity` module: implementation of `Arbitrary` is done in the fuzzing module itself.

This PR (apart from a `warn -> debug` log downgrade) does not make any changes to the codebase, it just adds new testing coverage. This could also be used a starting point for future fuzz testing (see discussion #3732)


clarity/fuzz/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+[profile.release]
+debug = 1
+
+[[bin]]
+name = "fuzz_sanitize"
+path = "fuzz_targets/fuzz_sanitize.rs"
+test = false
+doc = false
+
+# [[bin]]
```
jcnelson, line 35: nit: Are these just TODOs, or are the artifacts of a change that is no longer meant to be included? Either way, commented-out code is confusing.

clarity/fuzz/fuzz_targets/fuzz_sanitize.rs
```diff
@@ -0,0 +1,342 @@
+#![no_main]
```
jcnelson, line 1: copyright statement is missing

clarity/fuzz/fuzz_targets/fuzz_sanitize.rs
```diff
@@ -0,0 +1,342 @@
...
+                    .into()
+            },
+            _ => return Err(arbitrary::Error::IncorrectFormat),
+        };
+
+        Ok(FuzzClarityValue(clar_value))
+    }
+}
+
+pub fn strict_admits(me: &TypeSignature, x: &ClarityValue) -> Result<bool, CheckErrors> {
```
jcnelson, line 117: I notice there's no recursion depth checking here.  Could this lead to a stack overflow, given that both `me` and `x` are generated from the fuzzer?

clarity/fuzz/fuzz_targets/fuzz_sanitize.rs
```diff
@@ -0,0 +1,342 @@
...
+
+fuzz_target!(|value: FuzzClarityValue| {
+    fuzz_sanitize(value.0);
+});
+
+
+
+/// Same as fuzz_sanitize, but does not check any serialization routines
+fn fuzz_value_sanitize(input: ClarityValue) {
+    let computed_type = TypeSignature::type_of(&input);
```
jcnelson, line 236: Can `TypeSignaure::type_of` OOM if `input` is too deep?

clarity/fuzz/fuzz_targets/fuzz_sanitize.rs
kantai: Like below, it shouldn't. `TypeSignature` construction (which uses the `TypeSignature::type_of()` method) will not construct a deep enough type signature. The tuple bug fixed in 2.4 also cannot trigger a stack overflow here -- the length is compared before descent.

If it did result in a stack overflow, the fuzzer would (correctly) raise the input case as a failure case.

clarity/fuzz/fuzz_targets/fuzz_sanitize.rs
kantai: Added

clarity/fuzz/Cargo.toml
kantai: Removed -- these were artifacts from the prior fuzzing draft PRs.

jcnelson (approved):
Thanks for addressing my feedback!

obycode (approved):
LGTM!

PR #3782, opened 2023-07-07
Add memo field to StackStx and DelegateStx

### LINKED ISSUE #3530: Add missing memo field for burnchain operations
The  `PreStxOp`, `StackStxOp` and `DelegateStxOp` burnchain operations are using less than 80 bytes of the OP_RETURN data. We would like to save the remaining bytes in these cases as a `memo: Vec<u8>` field which can be used for in-band signaling between miners.

This issue originates from a PR comment: [link](https://github.com/stacks-network/stacks-blockchain/pull/3486#issuecomment-1402705145).

pavitthrap:
### Description
Stores the memo for two burnchain operations, stackstx and delegatestx

### Applicable issues
- fixes #3530


src/chainstate/burn/operations/delegate_stx.rs
```diff
@@ -680,5 +794,68 @@ mod tests {
         assert_eq!(op.delegated_ustx, u128::from_be_bytes([1; 16]));
         assert_eq!(op.delegate_to, StacksAddress::new(22, Hash160([2u8; 20])));
         assert_eq!(op.until_burn_height, Some(u64::from_be_bytes([1; 8])));
+        assert_eq!(op.memo, vec![1; 27]);
+    }
+
+    // Parse a DelegateStx op in which data is longer than 77, so that it is truncated.
```
jcnelson, line 800: Is this desirable behavior?  IIRC OP_RETURNs can be as long as 520 bytes, but the relay policy constrains them to 80 bytes.  Miners can directly include 520-byte OP_RETURNs, though.  Do we want to support this?

src/chainstate/burn/db/sortdb.rs
```diff
@@ -858,6 +861,53 @@ const SORTITION_DB_SCHEMA_5: &'static [&'static str] = &[
     );"#,
 ];
 
+// force the node to go and store the memo for stack_stx and delegate_stx
+const SORTITION_DB_SCHEMA_6: &'static [&'static str] = &[
```
jcnelson, line 865: This looks like a breaking change -- you can't just `DROP TABLE` these two tables on a running node, because block validation depends on them being populated with the latest Bitcoin transaction data.

This branch can't be a point release without a big warning in the CHANGELOG that node operators will need to sync from genesis.  Can you instead just `ALTER TABLE` to add the `memo` column?

src/chainstate/burn/operations/delegate_stx.rs
pavitthrap: My reasoning here is as follows: 
- TransferStxOp does already parse a memo field, and the parsing essentially fails if there are more than 80 bytes: 
```
if data.len() > (61 + 16) {
            // too long
            warn!(
                "TransferStxOp payload is malformed ({} bytes, expected <= {})",
                data.len(),
                16 + 61
            );
            return None;
        }
```
- I wanted to model the parsing of StackStx and DelegateStx off of TransferStx, but I couldn't "fail" the parsing if there were more than 80 bytes (because that it not backwards compatible), so that's why I thought truncating it would mirror that behavior while maintaining backwards compatibility.

src/chainstate/burn/db/sortdb.rs
pavitthrap: I can `ALTER TABLE`, but the reason I did `DROP TABLE` is because I thought this change would require a sync from genesis in order to populate the memo fields? 

I am going to assume from your suggestion that it is okay that memo remains unpopulated for already processed transaction.

src/chainstate/burn/operations/delegate_stx.rs
```diff
@@ -32,13 +33,13 @@ impl DelegateStxOp {
         )
     }
 
-    fn parse_data(data: &Vec<u8>) -> Option<ParsedData> {
+    fn parse_data(data: &mut Vec<u8>) -> Option<ParsedData> {
```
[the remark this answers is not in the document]
pavitthrap: This is because I call the function `truncate` on it. 

I'll decide how to proceed with this based on the other discussion in this review on truncation.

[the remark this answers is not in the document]
jcnelson: Ah.  I don't think the burnchain parser allows the payload to be longer than 77 bytes as it is.  But even if it does (or will do so in a subsequent release), it's probably better to just do this instead of mutating the argument:

```
let relevant_data = &data[0, cmp::max(77, data.len())]
// do stuff with relevant_data instead of data.
```

src/chainstate/burn/operations/delegate_stx.rs
jcnelson: If we're going to support OP_RETURN payloads greater than 80 bytes, then we'll need to make sure the burnchain parser logic will accept this.

src/chainstate/burn/operations/stack_stx.rs
```diff
@@ -196,12 +199,28 @@ impl StackStxOp {
...
         }
 
+        if data.len() > 77 {
+            // too long
+            warn!(
+                "StacksStxOp payload is too long ({} bytes, expected <= {}); truncating to correct length",
+                data.len(),
+                77
+            );
+            data.truncate(77);
```
jcnelson, line 209: Same idea here -- I don't think you should modify `data` if you can just get away with taking a slice.

jcnelson (commented):
Overall LGTM.  My biggest concern is whether or not we should assume OP_RETURNs are 80 bytes or fewer, even though a Bitcoin miner can create one of up to 520 bytes.  I'm inclined to cap the OP_RETURN payload to 80 bytes (i.e. assume the `memo` does not exceed the 80th byte) because that's what the Bitcoin network relay policy requires, and there's no obvious use-case for marginally longer memos that I can think of.

src/chainstate/burn/db/sortdb.rs
```diff
@@ -858,6 +861,19 @@ const SORTITION_DB_SCHEMA_5: &'static [&'static str] = &[
...
+    r#"
+    ALTER TABLE stack_stx ADD COLUMN memo TEXT;
+    "#,
+    r#"
+    ALTER TABLE delegate_stx ADD COLUMN memo TEXT;
+    "#,
```
kantai, line 871: Per my comment above -- this alter table syntax will fill the table with `NULL` sqlite values in this column. This can be avoided with defaults:

```sql
ALTER TABLE stack_stx ADD COLUMN memo TEXT NOT NULL DEFAULT "";
```

```sql
ALTER TABLE delegate_stx ADD COLUMN memo TEXT NOT NULL DEFAULT "";
```

This would still need a test case to ensure that the pre-schema-6 values are still parseable from the DB (they should just have a zero-length memo field).

src/chainstate/burn/operations/delegate_stx.rs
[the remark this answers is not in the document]
kantai: Yeah, I agree with the above -- I'd prefer either adding a new slice variable like @jcnelson recommends or shadowing the existing one:

```rust
let data = &data[..cmp::min(77, data.len()];
```

(Note: I think you want cmp::min here, not max!)

kantai (commented):
This PR looks mostly okay, but I think it needs a couple changes -- the alter table behavior needs to be tested (and possibly fixed), and I agree with @jcnelson's comments on the vec mutations.

src/chainstate/burn/db/sortdb.rs
pavitthrap: Hmm.. Personally I'm in favor of dropping the table and re-adding it, instead of doing `ALTER TABLE`.  

Otherwise, in `insert_stack_stx`, I would need to case the db query on the version of the Sortition database (which feels clunky to me).

PR #3783, opened 2023-07-10
Parse memo field for PreStxOp

### LINKED ISSUE #3530: Add missing memo field for burnchain operations

pavitthrap:
### Description
Parsing memo field for the PreStxOp

### Applicable issues
- fixes #3530

### DIFF
--- a/src/chainstate/burn/operations/stack_stx.rs
+++ b/src/chainstate/burn/operations/stack_stx.rs
@@ -59,6 +59,7 @@ impl PreStxOp {
     pub fn new(sender: &StacksAddress) -> PreStxOp {
         PreStxOp {
             output: sender.clone(),
+            memo: vec![],
             // to be filled in
             txid: Txid([0u8; 32]),
             vtxindex: 0,
@@ -116,6 +117,9 @@ impl PreStxOp {
             return Err(op_error::InvalidInput);
         };
 
+        let mut memo = tx.data();
+        memo.truncate(77);
+
         let outputs = tx.get_recipients();
         assert!(outputs.len() > 0);
 
@@ -144,6 +148,7 @@ impl PreStxOp {
 
         Ok(PreStxOp {
             output: output,
+            memo,
             txid: tx.txid(),
             vtxindex: tx.vtxindex(),
             block_height,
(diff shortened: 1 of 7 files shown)


src/chainstate/burn/operations/stack_stx.rs
```diff
@@ -116,6 +117,9 @@ impl PreStxOp {
             return Err(op_error::InvalidInput);
         };
 
+        let mut memo = tx.data();
+        memo.truncate(77);
```
jcnelson, line 121: Is there a reason we're limiting it to 77 bytes?

jcnelson (commented):
I don't think this is backwards-compatible with the existing chainstate -- previously-stored `PreStxOp`s won't be parseable with this new structure.

Is this change intended to be part of a point release, or is it part of the sBTC hard fork?  Because if it's the former, then I would highly recommend making it backwards compatible. If not, then no need to worry.

pavitthrap:
> I don't think this is backwards-compatible with the existing chainstate -- previously-stored `PreStxOp`s won't be parseable with this new structure.
> 
> Is this change intended to be part of a point release, or is it part of the sBTC hard fork? Because if it's the former, then I would highly recommend making it backwards compatible. If not, then no need to worry.

Hmm.. PreStxOp's are actually not stored anywhere, so I'm not sure what exactly you're pointing to. 

Also, this is intended to be rolled into a hard fork, not a point release.

jcnelson:
> Hmm.. PreStxOp's are actually not stored anywhere, so I'm not sure what exactly you're pointing to.

Ah, yes, you're right.  Was thinking of the other burnchain ops when I wrote that.

jcnelson (approved):

wileyj:
ping @kantai @obycode

src/chainstate/burn/operations/stack_stx.rs
obycode: Yes, why 77? Should it be 34, like the `stx-transfer-memo?`?

obycode: Oh, I see the stx transfer burnchain op allows up to 61 bytes, so the size is not consistent.

obycode: Ah, it's 77 because it's just 2 magic bytes, 1 byte for the op, and the rest are memo.

obycode (approved):
LGTM

PR #3784, opened 2023-07-10
Feat: ignore logging errors rather than panicking

kantai:
### Description

The log macros in the `stacks-blockchain` use a [`Fuse`](https://docs.rs/slog/latest/slog/struct.Fuse.html) drain. This panics if there's a drain error -- like the logging FS is busy and would block.

This patch uses an `IgnoreResult` drain instead.


jcnelson (approved):

obycode (approved):
LGTM

PR #3785, opened 2023-07-10
workaround to failing actions for "no space left"

wileyj:
ex: https://github.com/stacks-network/stacks-blockchain/actions/runs/5510601197

not sure of the root cause here, but the last i had checked this docker image was around 8GB in size. 
there *should* be plenty of space available on the runner (historically this has been true) - so either the free space is less now for some reason, or we're using more space than we have in the past. 

this is a test workaround that does 3 things:
1. prints`dh -f` output
2. runs [this action](https://github.com/jlumbroso/free-disk-space/releases/tag/v1.1.0) to reclaim space
2. prints `df -h` output again

### DIFF
--- a/.github/workflows/bitcoin-tests.yml
+++ b/.github/workflows/bitcoin-tests.yml
@@ -44,6 +44,10 @@ jobs:
         with:
           tool-cache: true
 
+      - name: docker prune
+        run: |
+          docker system prune --force
+
       - name: Post clean diskfree
         run: | 
           df -h
--- a/.github/workflows/bitcoin-tests.yml
+++ b/.github/workflows/bitcoin-tests.yml
@@ -27,6 +27,9 @@ jobs:
       - name: Checkout the latest code
         id: git_checkout
         uses: actions/checkout@v3
+
+      # Could run a `docker system prune here to clean up a little extra space, ~4-5GB
+
       - name: Build bitcoin integration testing image
         id: build_docker_image
         env:
@@ -35,23 +38,12 @@ jobs:
         run: |
           rm .dockerignore
           docker build -f ./.github/actions/bitcoin-int-tests/Dockerfile.generic.bitcoin-tests -t stacks-blockchain:integrations .
-      - name: Pre clean diskfree
-        run: |
-          df -h
 
       - name: Free Disk Space (Ubuntu)
         uses: jlumbroso/free-disk-space@main
         with:
           tool-cache: true
           docker-images: false
-
-      - name: Post clean diskfree
-        run: |
-          df -h
-
-      - name: Docker image sizes
-        run: |
-          docker images
 
       - name: Export docker image as tarball
         id: export_docker_image
--- a/.github/workflows/bitcoin-tests.yml
+++ b/.github/workflows/bitcoin-tests.yml
@@ -44,10 +44,6 @@ jobs:
         with:
           tool-cache: true
 
-      - name: docker prune
-        run: |
-          docker system prune --force
-
       - name: Post clean diskfree
         run: | 
           df -h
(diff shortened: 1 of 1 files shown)


wileyj:
trigger bump

wileyj:
update:
1. ~~prints `dh -f output`~~
2. runs [this action](https://github.com/jlumbroso/free-disk-space/releases/tag/v1.1.0) to reclaim space
3. ~~prints `df -h` output again~~
4. moves the docker image storage from `tar` -> `tar.gz`.

.github/workflows/bitcoin-tests.yml
```diff
@@ -35,15 +38,24 @@ jobs:
...
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: true
+          docker-images: false
+
       - name: Export docker image as tarball
         id: export_docker_image
-        run: docker save -o integration-image.tar stacks-blockchain:integrations
+        # run: docker save -o integration-image.tar stacks-blockchain:integrations
```
obycode, line 50: Should this comment be removed?

obycode (approved):
LGTM. Thanks @wileyj!

.github/workflows/bitcoin-tests.yml
wileyj: yes, i can remove them

.github/workflows/bitcoin-tests.yml
```diff
@@ -35,15 +38,23 @@ jobs:
         run: |
           rm .dockerignore
           docker build -f ./.github/actions/bitcoin-int-tests/Dockerfile.generic.bitcoin-tests -t stacks-blockchain:integrations .
+
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: true
+          docker-images: false
+
```
kantai, line 47: Is this step necessary with the gzip change?

I'm pretty wary of using third-party actions.

wileyj: without that step, it's *really* close. 
without cleaning up packages on the host, we may hit this issue again. 

if this is a blocker, the only other thing we could (outside of manually removing packages, which is what this action does) is as j2p2 suggested: `docker system prune --force` before our image is built. 
there's about 3-4GB of images loaded by default on the runner that we don't use/need.

the free-disk-space workflow is pretty simple and straightforward: https://github.com/jlumbroso/free-disk-space/blob/main/action.yml  but you're correct in that it is a 3rd party action and they could add some logic there to do something we don't like (the same can be said for other 3rd party actions we use though). 

if you feel strongly about not using it,  there should be just enough space by a combination of removing unnecessary docker images and using gzip vs tar. the caveat is that if the base packages change to use more space, we'll have this issue again (however, hopefully by then we won't even need this compressed docker image).

wileyj: ping @kantai - any thoughts on my reply?

jcnelson: It's simple enough that we could just vendor it if we're worried.  I don't think Docker has the equivalent of Cargo.lock; if they truly don't, then that'd at least de-risk supply chain poisoning.

wileyj: Unless there is a strong reason to **not** use this action, i'd say leave it in and let it be the temporary workaround it was meant to be. 
I'm also working on some methods to make this change moot; it is only meant to fix a very current and specific issue,  and should not be something we use long term.

@jcnelson you do raise an interesting idea along with @kantai though - if there are concerns about using 3rd party workflows, we should have a discussion about recreating them ourselves or vendoring them.

wileyj: Understood - I do want to highlight that it feels like a "not built here" condition. But, I can refactor this PR to address your concerns and we can discuss what to do about 3rd party actions at some other time. 

I'll have a commit doing what i suggested in the first reply to your comment later today. It should be enough for the short term as a workaround to this issue.

kantai (approved):

PR #3793, opened 2023-07-14
Refactor: PoX lock handling into separate workspace member

kantai:
### Description

This is a refactor that moves the PoX lock handling into a separate workspace member.

This was a pretty simple set of operations to lift: they did not depend on any non-public members and the only circular dependency possibility was the use of ChainstateError in some of the methods. ChainstateError wasn't really necessary (and in some cases, was doing like a `ClarityError(ChainstateError(ClarityError)))` error wrapping), so this PR uses its own LockingError enum instead in those places.

The reason for this refactoring (beyond the fact that it's easy to do, and separates some complexity from the StacksChainstate struct) is to support something like my PR on the clarinet repo (https://github.com/hirosystems/clarinet/pull/1074) without code duplication (i.e., the ability to evaluate the PoX locks in clarinet).


jcnelson (approved):

obycode (approved):
LGTM! Great idea to separate this out and use it in clarinet!

PR #3798, opened 2023-07-19
Clean unused imports

obycode:
### Description

```
use regex::internal::Exec;
```
causes an error with newer versions of the `regex` crate, and it is not
used anyway, so remove it.

Also re-format using the recommended:
```
rustfmt clarity/src/vm/costs/mod.rs --config group_imports=StdExternalCrate
```

### DIFF
--- a/clarity/src/vm/costs/mod.rs
+++ b/clarity/src/vm/costs/mod.rs
@@ -18,9 +18,9 @@ use std::collections::{BTreeMap, HashMap};
 use std::convert::{TryFrom, TryInto};
 use std::{cmp, fmt};
 
-use regex::internal::Exec;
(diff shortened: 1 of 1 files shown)


kantai (approved):
LGTM -- note that `Exec` was removed in `master` as well.

obycode:
> note that `Exec` was removed in `master` as well.

Oh, good. Then maybe we should not merge this and instead merge master back to develop?

kantai:
Yeah, `master` should be merged to `develop` -- I was going to open a PR to do so, but am waiting until the github action fixes are merged.

obycode:
Ok, cool. I'll close this one then.

PR #3813, opened 2023-07-28
[stacks-core] Extract fundamental types and logic from blockstack lib

stjepangolemac:
- hashing primitives
- uint256 primitives


Cargo.lock
```diff
@@ -85,18 +85,33 @@ checksum = "739f4a8db6605981345c5654f3a85b056ce52f37a39d34da03f25bf2151ea16e"
...
-version = "0.7.18"
+version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
+checksum = "43f6cb1bf222025340178f382c426f13757b2960e89779dfcb319c32542a5a41"
 dependencies = [
  "memchr",
 ]
 
+[[package]]
```
jcnelson, line 95: Why are these new dependencies?  Please remove them -- there's no reason to depend on Android-specific things because we don't support Android.

Cargo.lock
```diff
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.17.0"
+version = "0.20.0"
```
jcnelson, line 7: In general, why are there any revisions at all to Cargo.lock?  This should not change at all.

stacks-common/src/util/hash.rs
```diff
@@ -109,23 +113,23 @@ impl_array_newtype!(Keccak256Hash, u8, 32);
...
-impl_array_newtype!(Sha256Sum, u8, 32);
-impl_array_hexstring_fmt!(Sha256Sum);
-impl_byte_array_newtype!(Sha256Sum, u8, 32);
-
-impl Default for Sha256Sum {
-    fn default() -> Self {
-        Sha256Sum::zero()
-    }
-}
+// #[derive(Serialize, Deserialize)]
```
jcnelson, line 116: Please drop commented-out code

stjepangolemac:
I'm a bit stuck with the Uint256 logic. I've attemped to use a 3rd party lib that handles bignums and managed to port everything except [`low_u32()` and `low_u64()` functions](https://github.com/stacks-network/stacks-blockchain/blob/master/stacks-common/src/util/uint.rs#L59-L71).

- 3rd party lib https://docs.rs/ruint/latest/ruint/index.html
- blockstack_lib exposes two similar methods to create uints from hashes `into_le()` and `into_be()` but only the first one is used https://github.com/stacks-network/stacks-blockchain/blob/master/stacks-common/src/util/hash.rs#L330-L339
- I _think_ I have [an equivalent ](https://github.com/stacks-network/stacks-blockchain/blob/3812-stacks-core-extract-fundamental-types-and-logic-from-blockstack_lib/stacks-core/src/hash/sha256.rs#L11-L24)to `into_le()` and a [test that is green](https://github.com/stacks-network/stacks-blockchain/blob/master/stacks-common/src/util/uint.rs#L59-L71)
- the problem is that ruint doesn't expose something like `low_u32()` and I don't really know how to get it, looking into it

Here is the type I'm trying to use for bignums https://docs.rs/ruint/latest/ruint/struct.Uint.html#method.most_significant_bits

@jcnelson / @kantai do you have any ideas on how to implement `low_u32()` and similar? Alternatively, would you prefer to keep using our implementation and discard the ruint crate?

stacks-core/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "stacks-core"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+hex = "0.4.3"
+sha2 = "^0.10.7"
+ripemd = "^0.1.3"
+ruint = "1.9.0"
```
jcnelson, line 10: What's this?

stacks-core/src/hash/sha256.rs
```diff
@@ -0,0 +1,135 @@
+use ruint::aliases::U256;
```
jcnelson, line 1: We already have a `Uint256` struct we use pervasively throughout the code. Please use that instead.

jcnelson:
> @jcnelson / @kantai do you have any ideas on how to implement low_u32() and similar? Alternatively, would you prefer to keep using our implementation and discard the ruint crate?

Just keep using the `Uint256` struct that already exists.  Discard `ruint`.

Cargo.lock
stjepangolemac: A few dependencies I've added to stacks-core are using their latest version and that might be causing this. When / if we reorganize our dependencies to be defined in the workspace level this should happen less often.

That said, with this background do you still consider it a problem?

stacks-common/src/util/hash.rs
stjepangolemac: I've now deleted all commented code I could find and remember. Let me know if you find anything else.

stacks-core/Cargo.toml
stjepangolemac: I've removed ruint as suggested.

stacks-core/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "stacks-core"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+hex = "0.4.3"
```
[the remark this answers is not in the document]
stjepangolemac: We've had a from scratch implementation of hex encoding in our codebase. Considering how ubiquitous hex is, and how much this crate is used, I believe there's no reason to have a custom implementation. This way we don't need to maintain it and probably won't need to think about it for a long time (if ever).

Let me know if you disagree.

stacks-core/src/hash/sha256.rs
stjepangolemac: I've now done that.

stjepangolemac:
After your feedback here are the things I've done / changed:

- Removed `ruint` usage and ported `uint` code from `stacks-common`
  - Replaced macro usage with const generics
  - Rewrote small bits of code where it was easy and where it looked like a net value gain
  - Changed `Uint` API slightly to make intent and failure modes more explicit
- Moved subset of SHA256 related types from `stacks-common`
  - Also replaced macro usage with const generics
  - Changed API to allow for more argument types that can easily be converted to byte slices
- Replaced all usages of `Uint` and SHA256 hashing in the codebase
- Fixed all tests that were failing for me locally
- There's a single tests that runs "indefinitely" and that is `chainstate::stacks::db::unconfirmed::test::test_unconfirmed_refresh_one_microblock_stx_transfer` - I left it for 20 minutes and it was still running

Cargo.lock
jcnelson: I think you can just remove `chrono`.  It's only needed by `stacks-node`; the `chrono` dependency really ought to be present only in `testnet/stacks-node/Cargo.toml`.

Cargo.lock
jcnelson: Yes. Please revert them.  I audit packages that we add as dependencies.

stacks-core/Cargo.toml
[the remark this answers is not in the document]
jcnelson: Disagree.  Please remove it.  I'd rather have the devil I know running than a devil I don't know, especially in a system that manages close to $1 billion of other peoples' money.

> Considering how ubiquitous hex is, and how much this crate is used, I believe there's no reason to have a custom implementation. This way we don't need to maintain it and probably won't need to think about it for a long time (if ever).

I strongly disagree with this mentality in the context of blockchain development when it comes to "small" dependencies like `hex`.  Code we add as a dependency is still code we're responsible for maintaining.  If it misbehaves, it hardly matters in the eyes of our users as to whether or not we wrote it, because their funds are at risk either way.  It's still our problem to fix, even if it means forking the broken package and patching it.  As such, it's better to have to debug code that we intimately understand than code that we do not.

Our hex implementation is very small compared to `hex` -- so small that it's very hard to get wrong.  It doesn't need the extra logic for `no_std` or `serde`, and we currently don't need the custom iterators that `hex` offers.

clarity/src/vm/functions/crypto.rs
```diff
@@ -62,8 +63,30 @@ macro_rules! native_hash_func {
     };
 }
 
+macro_rules! native_hash_func_new {
```
jcnelson, line 66: This is consensus-critical code.  I'd strongly prefer if you did not add this, given that the existing code already works (and if it has bugs, it's better to be stuck with them until the next hard fork than risk a chain split).

src/burnchains/bitcoin/indexer.rs
```diff
@@ -3083,7 +3083,7 @@ mod test {
             SpvClient::new(db_path, 0, None, BitcoinNetworkType::Mainnet, false, false).unwrap();
         for (interval, work_str) in chain_work.iter() {
             let calculated_work = spv_client.find_interval_work(*interval).unwrap().unwrap();
-            let expected_work = Uint256::from_hex_be(work_str).unwrap();
+            let expected_work = Uint256::from_be_hex(work_str).unwrap();
```
jcnelson, line 3086: Generally not a fan of renaming things for the sake of renaming them.

stacks-core/src/uint.rs
```diff
@@ -0,0 +1,891 @@
+use std::{
```
jcnelson, line 1: Missing license comment (both here and elsewhere)

jcnelson: Actually, please drop this file entirely.  There is nothing wrong with `stacks-common/src/util/uint.rs`.

stacks-core/src/lib.rs
```diff
@@ -0,0 +1,23 @@
+use std::array::TryFromSliceError;
```
jcnelson, line 1: Missing license comment.

jcnelson (changes requested):
This PR is unacceptably risky.  The files in `stacks-common` that implement bigints, hashing, and so on are all on many, many consensus-critical code paths, and they must remain **bug-for-bug compatible** with the existing chainstate.

I thought this PR was only going to move a few files around from `stacks-common` into `stacks-core`.  Please do that instead, but don't change the code itself.  Don't worry about the licensing; once the crate exists, we can change it.

stjepangolemac:
Closed in favor of the reimplementation happening in the repo below.

https://github.com/stacks-network/sbtc

PR #3814, opened 2023-07-28
Move blockstack-core into a separate folder (develop branch)

### LINKED ISSUE #3809: Top level Cargo.toml file shouldn't have package properties.
**Describe the bug**

The Rust cargo tool supports two types of cargo.toml files:
- for a package
- workspace

Our top level [Cargo.toml](https://github.com/stacks-network/stacks-blockchain/blob/master/Cargo.toml) contains both sets of properties, from a package and from a workspace.

**Steps To Reproduce**

- Some IDEs and plugins (for example VSCode) can't properly read and understand the file.
- Run `cargo build`
  ```
  warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
  package:   /.../stacks-network/stacks-blockchain/clarity/Cargo.toml
  workspace: /.../stacks-network/stacks-blockchain/Cargo.toml
  ```

**Expected behavior**

A simple command `cargo build` should build all packages in the repo without ignoring some of them.

A top level Cargo.toml should contain only workspace properties. For example https://github.com/Trust-Machines/stacks-sbtc/blob/main/Cargo.toml 

**Additional context**

We can use the `default-run` option if we need a default binary to run by `cargo run` command. See https://stackoverflow.com/questions/51785457/how-can-i-specify-which-crate-cargo-run-runs-by-default-in-the-root-of-a-cargo

sergey-shandar:
### Applicable issues
- fixes #3809

### Checklist
- [ ] Changelog is updated


stjepangolemac (approved):
Looks good. Do you think we could fit changing dependencies to workspace ones in this PR? Maybe it's better to leave that for another day.

jcnelson:
`blockstack` is a name we shouldn't be using anymore. Please rename to
`stackslib` or something like that to reflect the system's current name.

jcnelson (changes requested):
See earlier emailed comment

sergey-shandar:
> The only changes to `Cargo.lock` should be changing `blockstack-core` to `stackslib`.

How often do we update `Cargo.lock`? Do we have separate PRs for these type of changes? Or are we stuck with the old versions of packages? 

@AshtonStephens here's another example of `Cargo.lock` in the repo.

jcnelson:
> How often do we update Cargo.lock? Do we have separate PRs for this type of changes? Or are we stuck with the old versions of packages?

We change versions when it's necessary to make progress or to fix disclosed vulnerabilities.  I make it a point to look through the code for each dependency so I know what we're responsible for maintaining.

stackslib/Cargo.toml
```diff
@@ -0,0 +1,128 @@
...
+default = ["developer-mode"]
+profile-sqlite = []
+disable-costs = []
+developer-mode = []
+monitoring_prom = ["prometheus"]
+slog_json = ["slog-json", "stacks-common/slog_json", "clarity/slog_json", "pox-locking/slog_json"]
+testing = []
+
+# Use a bit more than default optimization for
+#  dev builds to speed up test execution
```
jcnelson, line 114: Remove this.  If anything, we want build times to be fast when developing.

jcnelson (approved):
Overall LGTM now, but please make that one change about `opt-level` before merging.  Thanks!

stackslib/Cargo.toml
sergey-shandar: Can we make it in the next PR, I would like to make sure that it would be granular changes and easy to revert if something happen.

kantai: This is present in the current `Cargo.toml` and should IMO be replicated here as @sergey-shandar did.

Opt level 1 has pretty small impacts on compile times but pretty big impacts on test execution. This and the setting below it yielded shaved like 25%-30% of the execution time off of unit tests when it was added to the toml file. Build times, on the other hand, are generally pretty quick if you don't need to rebuild dependencies (as would be the case for local development) and can take advantage of caching on github actions. Not to mention that build times for local developers should also improve dramatically as more pieces get refactored into other workspace members so that you need to rebuild less of the project to test any changes.

jcnelson: Okay, fine to leave it.

sergey-shandar: > This is present in the current `Cargo.toml` and should IMO be replicated here as @sergey-shandar did.
> 
> Opt level 1 has pretty small impacts on compile times but pretty big impacts on test execution. This and the setting below it shaved like 25%-30% of the execution time off of unit tests when it was added to the toml file. Build times, on the other hand, are generally pretty quick if you don't need to rebuild dependencies (as would be the case for local development) and can take advantage of caching on github actions. Not to mention that build times for local developers should also improve dramatically as more pieces get refactored into other workspace members so that you need to rebuild less of the project to test any changes.

IMHO, we should do all optimization it on workspace level.

kantai:
Hmm.

This PR radically slowed down the unit test time -- went from ~40 minutes to now about 100 minutes.

sergey-shandar:
> Hmm.
> 
> This PR radically slowed down the unit test time -- went from ~40 minutes to now about 100 minutes.

@kantai check this PR #3814 if it fixes the issue.

PR #3838, opened 2023-08-04
Workspace level profiles

### LINKED ISSUE #3835: Set `opt-level` on workspace level
> Opt level 1 has pretty small impacts on compile times but pretty big impacts on test execution. This and the setting below it shaved like 25%-30% of the execution time off of unit tests when it was added to the toml file. Build times, on the other hand, are generally pretty quick if you don't need to rebuild dependencies (as would be the case for local development) and can take advantage of caching on github actions. Not to mention that build times for local developers should also improve dramatically as more pieces get refactored into other workspace members so that you need to rebuild less of the project to test any changes.

IMHO, we should do all optimization it on workspace level.

_Originally posted by @sergey-shandar in https://github.com/stacks-network/stacks-blockchain/pull/3814#discussion_r1283734094_

sergey-shandar:
### Applicable issues
- fixes #3835 

### Checklist
- [ ] Changelog is updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,3 +7,19 @@ members = [
     "clarity",
     "stx-genesis",
     "testnet/stacks-node"]
+# Use a bit more than default optimization for
+#  dev builds to speed up test execution
+[profile.dev]
+opt-level = 1
+
+# Use release-level optimization for dependencies
+# This slows down "first" builds on development environments,
+#  but won't impact subsequent builds.
+[profile.dev.package."*"]
+opt-level = 3
+
+[profile.dev.package.regex]
+opt-level = 2
+
+[profile.release]
+debug = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "clarity",
     "stx-genesis",
     "testnet/stacks-node"]
+
 # Use a bit more than default optimization for
 #  dev builds to speed up test execution
 [profile.dev]
@@ -17,9 +18,6 @@ opt-level = 1
 #  but won't impact subsequent builds.
 [profile.dev.package."*"]
 opt-level = 3
-
-[profile.dev.package.regex]
-opt-level = 2
 
 [profile.release]
 debug = true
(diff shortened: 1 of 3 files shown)


kantai:
@sergey-shandar -- thanks for opening this. I also opened a similar PR #3836.

I don't think we should actually remove the profile settings from the workspace members yet, though. `clarity`, in particular, can be published by itself, and in that case, the profile settings from its Cargo file _are_ actually used. There's a long-standing open issue about that cargo warning here (https://github.com/rust-lang/cargo/issues/8264), and I'm inclined to just let that warning continue to echo in our case, because we do often need to define workspace member profiles

sergey-shandar:
> @sergey-shandar -- thanks for opening this. I also opened a similar PR #3836.
> 
> I don't think we should actually remove the profile settings from the workspace members yet, though. `clarity`, in particular, can be published by itself, and in that case, the profile settings from its Cargo file _are_ actually used. There's a long-standing open issue about that cargo warning here ([rust-lang/cargo#8264](https://github.com/rust-lang/cargo/issues/8264)), and I'm inclined to just let that warning continue to echo in our case, because we do often need to define workspace member profiles

It depends how do you build `clarity` or any other package. For example, if you build and publish it from a repository root, then it will always use workspace settings. Also, if you reference `clarity` using git URL `clarity = { git = "https://github.com/stacks-network/stacks-blockchain/", branch = "develop" }` it will also use workspace settings.

kantai:
> @sergey-shandar -- thanks for opening this. I also opened a similar PR https://github.com/stacks-network/stacks-blockchain/pull/3836.
>
> I don't think we should actually remove the profile settings from the workspace members yet, though. clarity, in particular, can be published by itself, and in that case, the profile settings from its Cargo file are actually used. There's a long-standing open issue about that cargo warning here (https://github.com/rust-lang/cargo/issues/8264), and I'm inclined to just let that warning continue to echo in our case, because we do often need to define workspace member profiles
>
> It depends how do you build clarity or any other package. For example, if you build and publish it from a repository root, then it will always use workspace settings. Also, if you reference clarity using git URL clarity = { git = "https://github.com/stacks-network/stacks-blockchain/", branch = "develop" } it will also use workspace settings.

Ah, in that case, I think that's fine.

kantai (commented):
This LGTM, but please drop the regex specific opt-level setting

stjepangolemac (approved):

PR #3847, opened 2023-08-14
CI changes to improve build/testing times

wileyj:
This PR reworks the entire CI workflow process in use today. 

The two main issues it deals with are: test speed/reliability, with a secondary focus on making it easier to modify in the future. 

## Reliability
some tests/operations that need to run are always going to be flaky. to address this, the PR moves all tests that are currently defined in a matrix and separates them into specific workflows based on *what* they are testing. i.e. "Atlas Tests" runs all the the tests relating to Atlas. There is also workflow for "Slow" tests - 2 specific tests have been identified that take longer than most others (roughly 20 minutes each), and they fail to complete more than others. For these, a 30 minute timeout has been added, and these tests are only being required when a release is created. 

## Speed:
The largest amount of work was done around speed. the current workflows contain several bottlenecks that have been removed - mainly, running jobs natively as opposed to the current method of using Docker. 
The first change was moving from `cargo test` to using [nextest](https://nexte.st/) which has noticeable speed improvements in how long a test takes to run. 
The second change is removing the docker steps where a large image is created as an artifact, then shared amongst many differnt matrix jobs. To counter this, we are creating a cache of [nextest test archives](https://nexte.st/book/reusing-builds.html) which is roughly 1/10th the size of the previous docker container used to run tests (this does add some addtional complexity, however). 
The third change is modifying the tests themselves. Two approaches were taken here - separating the tests themselves into different resuable workflows, and for the integration tests, using [nextest partitioning](https://nexte.st/book/partitioning.html) to run several tests concurrently (vs running them serially, as is the default).

## Other improvments this PR brings:
- Easier to read conditionals for jobs/steps. This was an easy win, but it should be much more clear about what will trigger a workflow. 
- uploading to codecov.io is now being retried on failure, this should help with workflow errors where the upload fails without a clear reason (i.e. a 500 response)
- locking each action to a specific version/commit so we know what we're running when using external actions. 
    - this PR defaults to using a commit for *all* actions, even historically "trusted" ones like github owned actions
- DRY the docker build workflow
    - single workflow will now build both alpine/debian images for release using a single workflow via a matrix of defined OS'es. there's room for improvement here, but it reduces the amount of yml needed. 
- cleanup: one aspect of the current workflow methods is that there is no cleanup performed. a configurable `cleanup.yml` is being proposed here to do a few things: 
    - delete caches after 1 day
    - delete failed actions after 1 day
    - delete *completed* actions (this would apply to any workflow that has completed successfully or not) after 2 days
    - this workflow should be run via manual trigger, as well as via nightly cron
- manual test runs: in the current workflow, to run some tests it is required to trigger a full workflow run. with the `standalone.yml` workflow, we can now:
    - run individual test suites when not triggered by a ci action
    - run all tests that would run during a tagged release
    - run all tests that would run during a PR action (opened, re-opened, synchronized)
- removed some unnecessary/outdated jobs
- if a PR is only changing yml or markdown, no CI is run (this could possibly be extended to only trigger if `*.rs, *.clar` is modified)
    - if a PR contains code changes, then a later commit changes markdown, CI is triggered again due to the first commit
- jobs are now triggered on PR changes:
    - opened/re-opened PR
    - synchronized PR (new commits after PR is opened)
    - PR approvals (will be triggered once per approval)
- enables an option to build a docker image from source on demand, vs running the `ci.yml` workflow which will take longer due to tests

## Ongoing issues:
As with any tool, there will always be some things we simply cannot fix. Using github's cache mostly improves the workflows, but it can happen where a cache does not restore onto the runner for unknown reasons. This is the reason the caches will persist (unti the cleanup workflow runs), so they are available for re-running the failed jobs. 

## Timing:
A lot of the timing hinges on how long it takes to create the cache (this is CPU bound).
It's also hard to definitively say a workflow will take x mins, due to how runners are allocated out of our control. 
However, it will be possible to open a PR and have the workflows complete in roughly 30 minutes. In testing, I've seen times from 15-35 min on successful runs (with an existing cache, so it's safe to add another 10-20 minutes if a cache needs to be created). 

with an existing cache, a small sample of the timing:
Atlas Tests: 12 mins
Bitcoin Tests: 23 mins
Epoch Tests: 16 mins
Slow Tests: 23 mins
Stacks Blockchain Tests (integration tests/full genesis): 21 mins
manual ci workflow: 23 mins (this includes PR approvals, opening a PR)
Release workflow (ci with an input tag): 44 mins

## Future work:
- fork action repos from the https://github.com/actions-rs org: there are some good repos here that could be very useful, but they've not actively maintained
- ~~build and provide bitcoin binaries required by tests vs pulling from a throttled public resource (this is in progress)~~
- it may be possible to simplify the tests where a matrix is used for a defined set of tests using nextest's [filtering](https://nexte.st/book/filter-expressions.html). a POC of this worked well, but it can be tricky due to how the tests are partitioned (it may make the workflows harder to read)


wileyj:
Note: the workflows still contain some comments that will not be needed after review. 
they will be removed before approval/merge (for example, there is a comment on some `uses:` lines to use buildjet runners, which may still be a change we want to make before merging this. 

Secondly, the `cleanup.yml` file will need to have the scheduling uncommented before merge. generically, i propose using `0 0 * * *` so it runs nightly (as well as keeping the option to run manually.

kantai:
@wileyj -- I don't see any action runs for this PR, should more actions have run on this?

wileyj:
> @wileyj -- I don't see any action runs for this PR, should more actions have run on this?

since most of the work was unknown and started as a POC (as well as the need to hardcode some values), all the actions i've run to test this have been in a fork: https://github.com/wileyj/stacks-blockchain

https://github.com/wileyj/stacks-blockchain/actions

the commit history is also quite ugly in this fork due to what's required to test a github action (in particular some extra files and lots of one line PR's to test trigger conditionals). 

I can invite anyone to my fork if it would be helpful in evaluating though

.github/workflows/atlas-tests.yml
```diff
@@ -0,0 +1,134 @@
+## Github workflow to run atlas tests
```
AshtonStephens, line 1: Yes, making these tests separate makes a lot of sense to me! Especially since they're so slow.

What do you think of adding more of a description to either the top of this file or the `Run Tests` job on line 101 that'll justify the history of these tests. Something like 

```yaml
## Github workflow for atlas tests. 
## Historically these tests take far longer than the others in the CI so we've separated them out
## For speed of testing.
```

.github/workflows/atlas-tests.yml
```diff
@@ -0,0 +1,134 @@
...
+
+jobs:
+  # atlas integration tests with code coverage
+  atlas-tests:
+    name: Atlas Test
+    runs-on: ubuntu-latest
+    ## run a maximum of 2 concurrent tests from the test matrix
+    strategy:
+      fail-fast: false
+      max-parallel: 2
```
AshtonStephens, line 30: Clarifying - is this "max-parallel" required when the matrix only has two options?

.github/workflows/atlas-tests.yml
```diff
@@ -0,0 +1,134 @@
...
+      - name: Setup Rust Toolchain
+        id: setup_rust_toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@f3c84ee10bf5a86e7a5d607d487bf17d57670965 # v1.5.0
+        with:
+          toolchain: stable
+          components: llvm-tools-preview
+          cache: false
+
+      ## Restore cargo cache data
+      - name: Restore Cargo Cache
```
AshtonStephens, line 51: Clarifying - why do we need to do anything with the cargo cache?

.github/workflows/atlas-tests.yml
```diff
@@ -0,0 +1,134 @@
...
+        uses: Wandalen/wretry.action@a163f62ae554a8f3cbe27b23db15b60c0ae2e93c # v1.3.0
+        with:
+          action: codecov/codecov-action@eaaf4bedf32dbdc6b720b63067d99c4d77d6047d # v3.1.4
+          with: |
+            files: ./lcov.info
+            name: ${{ matrix.test-name }}
+            fail_ci_if_error: false
+            verbose: true
+          attempt_limit: 3
+          attempt_delay: 10000
```
AshtonStephens, line 134: Do we think it could be clearer if we merged the two ./lcov.info files generated from these tests and sent them up to codecov in one go? In my mind we don't care where the coverage came from but I don't know if codecov merges the two coverage results if we upload two for the same commit under different names.

.github/workflows/bitcoin-tests.yml
```diff
@@ -1,205 +1,173 @@
-##
-## Bitcoin Integration Tests
-##
+## Github workflow to run bitcoin tests
```
AshtonStephens, line 1: This new workflow is far simpler and easier to read than what was here before - 10/10

.github/workflows/bitcoin-tests.yml
```diff
@@ -1,205 +1,173 @@
...
-          - tests::neon_integrations::microblock_large_tx_integration_test_FLAKY
-          - tests::neon_integrations::block_large_tx_integration_test
-          - tests::neon_integrations::microblock_limit_hit_integration_test
-          - tests::neon_integrations::block_limit_hit_integration_test
-          - tests::neon_integrations::fuzzed_median_fee_rate_estimation_test_window5
-          - tests::neon_integrations::fuzzed_median_fee_rate_estimation_test_window10
-          - tests::neon_integrations::use_latest_tip_integration_test
-          - tests::neon_integrations::test_flash_block_skip_tenure
+          - tests::neon_integrations::stx_delegate_btc_integration_test
+          - tests::neon_integrations::stx_transfer_btc_integration_test
```
AshtonStephens, line 62: Is there any way to autogenerate this so we don't need to update it manually?

.github/workflows/ci.yml
```diff
@@ -1,180 +1,228 @@
...
+      - "**.md"
+      - "**.yml"
+    ## might be better to use inclusive v exclusive paths here, ex:
+    # paths:
+    #   - "**.rs"
+    #   - "**.clar"
+  pull_request_review:
+    types:
+      - submitted
+    paths-ignore:
```
AshtonStephens, line 31: I don't think this is a valid member of `pull_request_review`

AshtonStephens:
> Note: the workflows still contain some comments that will not be needed after review. they will be removed before approval/merge (for example, there is a comment on some `uses:` lines to use buildjet runners, which may still be a change we want to make before merging this.

I have only seen comments that I would prefer stay within the workflow. Which comments would you remove?

AshtonStephens:
> > Note: the workflows still contain some comments that will not be needed after review. they will be removed before approval/merge (for example, there is a comment on some `uses:` lines to use buildjet runners, which may still be a change we want to make before merging this.
> 
> I have only seen comments that I would prefer stay within the workflow. Which comments would you remove?

Oh, actually I'm seeing some of them now. But the plain english comment's I'd keep in

.github/workflows/cleanup.yml
```diff
@@ -0,0 +1,140 @@
...
+            '
+          `; do
+            echo "*** Deleting cache: $ID"
+            gh api \
+              -H "Accept: application/vnd.github.v3+json" \
+              -X DELETE \
+              repos/${{ github.repository }}/actions/caches/$ID \
+            | echo;
+          done
+          echo "Done"
```
AshtonStephens, line 139: Personally, I'd love for there to be comments within this section - I'm having a hard time uderstanding the individual commands even with the "echos"

.github/workflows/ci.yml
```diff
@@ -1,180 +1,228 @@
...
-  ##        - Not a PR
-  build-source:
-    if: ${{ inputs.tag != '' || (github.ref != format('refs/heads/{0}', github.event.repository.default_branch) && !contains(github.ref, 'refs/pull')) }}
-    name: Build Binaries
-    uses: stacks-network/stacks-blockchain/.github/workflows/build-source-binary.yml@master
+  ######################################################################################
+  ## Create a tagged github release
+  ##
+  ## Runs when the following is true:
+  ##  - tag is provided
```
AshtonStephens, line 86: There's a lot of logic around tags, but I don't understand the significance of the tags (this could be a me issue). In addition to describing these conditions as dependent on `tag`s being provided what do you think of describing these conditions with what the tag's significance is to the job?

.github/workflows/clarity-js-sdk-pr.yml
```diff
@@ -20,26 +24,30 @@ jobs:
     runs-on: ubuntu-latest
     # This condition can be removed once the main `stacks-blockchain` workflow creates pre-releases
```
AshtonStephens, line 25: Thoughts on creating a gh issue to do this and linking it to this comment?

.github/workflows/cleanup.yml
```diff
@@ -0,0 +1,140 @@
+## Github workflow to clean up old caches and workflow runs
+##   TODO: look into using  https://github.com/actions/github-script
```
AshtonStephens, line 2: Would we do this to break the scripts up into smaller script steps?

.github/workflows/create-cache.yml
```diff
@@ -0,0 +1,291 @@
+## Github workflow to create reusable caches
```
AshtonStephens, line 1: I'd like an explanation of the cache here if possible, because I don't understand fully. It's to prevent building the same sources repeatedly?

.github/workflows/create-source-binary.yml
```diff
@@ -0,0 +1,81 @@
...
+            OS_ARCH=${{ matrix.platform }}
+            GIT_BRANCH=${{ env.GITHUB_REF_SHORT }}
+            GIT_COMMIT=${{ env.GITHUB_SHA_SHORT }}
+
+      ## Compress the artifact
+      - name: Compress artifact
+        id: compress_artifact
+        run: zip --junk-paths ${{ matrix.platform }} ./release/${{ matrix.platform }}/*
+
+      ## Upload the artifact to the github action (used in `github-release.yml` to create a release)
```
AshtonStephens, line 76: Where do we use these artifacts?

.github/workflows/epoch-tests.yml
```diff
@@ -0,0 +1,157 @@
...
+      - name: Upload Code Coverage
+        id: code_coverage
+        uses: Wandalen/wretry.action@a163f62ae554a8f3cbe27b23db15b60c0ae2e93c # v1.3.0
+        with:
+          action: codecov/codecov-action@eaaf4bedf32dbdc6b720b63067d99c4d77d6047d # v3.1.4
+          with: |
+            files: ./lcov.info
+            name: ${{ matrix.test-name }}
+            fail_ci_if_error: false
+            verbose: true
```
AshtonStephens, line 155: Just curious, why are we doing

```yaml
with: | # <--- this line
  files: ...
  name: ...
```

instead of 

```yaml
with:
  files: ...
  name: ...
```

.github/workflows/image-build-binary.yml
```diff
@@ -0,0 +1,135 @@
+name: Docker Image (Binary)
```
AshtonStephens, line 1: I don't have a good sense of the immediate purpose for some of these workflows. I don't doubt there's a good reason so I think these files could benefit from a little "workflow justification" at the top so that if one of us gets hit by a bus the workflow is self justifying.

.github/workflows/bitcoin-tests.yml
wileyj: sort of...i tested using nextest filters https://nexte.st/book/filter-expressions.html

ex (from my notes, have to adjust to run the tests):
```
cargo nextest list \
   --run-ignored all \
   --archive-file /tmp/test_archive.tar.zst \
   -E '(
      test(/^tests::neon_integrations::*/) &
      test(/^tests::bitcoin_regtest::*/) &
      not test(/^tests::neon_integrations::atlas*/)
   )'
```

and the resultant list of tests matched what i expected the runner to run. 
however, doing it this way causes the tests to run serially, which increases the runtime of the test(s). 


using a matrix with a nextest partition (https://nexte.st/book/partitioning.html) is an option, but then we're getting into a scenario where we have to decide how many concurrent tests to run. 8? 12? 32? etc.

an argument can be made to use these filters though, the result will be slower tests and more time running the CI. as well as a little more care needed in deciding how many separate jobs to run - i found it useful to know exactly how many tests were being run, and seeing what test had failed in the UI vs a generic (partition 1 failed) and having to consult the logs to see what test(s) had failed. 

i do not have a strong opinion here - i can see the benefit of both approaches and will agree with the consensus, whatever it may be.

.github/workflows/atlas-tests.yml
```diff
@@ -0,0 +1,134 @@
+## Github workflow to run atlas tests
+
+name: Tests::Atlas
+
+on:
+  workflow_call:
```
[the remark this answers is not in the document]
wileyj: yes, this was a choice i made in this PR. i felt there were some tests that weren't needed for most CI runs (i.e. when a PR is synchronized) deferring to running these tests when a PR is approved or when a release is tagged. 

I did add the option of running any tests manually via the `standalone test` workflow, which has a dropdown to select which test workflow to run. i felt this was a good compromise vs running every test every time.

.github/workflows/atlas-tests.yml
wileyj: it is - without that directive, the tests will run serially essentially doubling the test time (give or take)

.github/workflows/atlas-tests.yml
wileyj: we don't - but having the cargo cache makes executing the tests much faster to start (since all the cargo dependencies are restored via that cache).

the caching is not foolproof though - i have seen the occasional instance where a cache fails to restore, resulting in any test relying on that cache also failing. it's rare, but it does happen and there's no clear reason as to why (the logs only show a download pausing at %x until the job times out).

wileyj: in lieu of using the cache, we'd have to install the cargo dependencies for each cargo job

.github/workflows/atlas-tests.yml
wileyj: that i'm unsure of - this is the way the CI currently works, the only changes i've added here (outside of  moving the atlas tests to a new workflow) are retrying the upload (we've seen spurious 500 errors from codecov), and locking the external workflow to a specific commit has (vs using a version/master)

wileyj: i suppose it's possible to upload them all at once - off the top of my head, we could save the output from each matrix job as an artifact, then a new job to upload the artifacts all at once

.github/workflows/ci.yml
wileyj: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_review

it's in the docs and works as expected where the workflow triggers when a review is submitted (conditionals prevent running unless it's an approved request).

wileyj:
> > Note: the workflows still contain some comments that will not be needed after review. they will be removed before approval/merge (for example, there is a comment on some `uses:` lines to use buildjet runners, which may still be a change we want to make before merging this.
> 
> I have only seen comments that I would prefer stay within the workflow. Which comments would you remove?

in other PR's, there's a tendency to not have any comments unrelated to the code/text.
ex:
```
uses: some-workflow.yml
# uses some-test-workflow.yml
```
i have a few instances like this in my branch, which i'll clean up soon as the changes are closer to final. nothing to worry about, just noting it.

.github/workflows/cleanup.yml
wileyj: will do!

.github/workflows/ci.yml
wileyj: this is done for historic reasons mainly - when we have a release, it's tagged with a version. 
in our repo/CI - we have an input for the ci.yml workflow that creates the tag vs tagging a commit and pushing it (which would also work, but historically we've used the input tag). 

the tag is created via the `github-releese.yml` workflow (implicitly) creating the tag when a github release is created. 

it may be tribal knowledge that i can comment better, but if you don't input a tag...no release is created and subset of the CI tests are run. if you do add an input tag, all the tests are run and release is made (with the binary archives/docker images)

.github/workflows/cleanup.yml
wileyj: possibly - it may make the logic a lot easier to follow as well (as evidenced by your other comment). 
using the shell/gh cli it works...but as you noticed it's pretty ugly

.github/workflows/create-cache.yml
wileyj: correct. i chose using the cache very specifically for these workflows (vs using an artifact like we do currently). 

the main reason is that the caches are *fast* to restore from, and will persist across workflow runs (as long as the cleanup workflow hasn't deleted it) - making it possible to easily re-run a failed workflow without having to recreate the cache. 

secondly, with a cache - it's available to every workflow that runs vs only the local workflow, so we can split apart different jobs (i.e. atlas tests/bitcoin tests) to better parallelize them. 

as noted above though, i have seen a few cases where the cache fails to restore (which is why i added a timeout), resulting in the workflow failing due to the cache, and not a failed test etc. it's rare, but it does happen.

.github/workflows/create-source-binary.yml
wileyj: these are all the compressed binaries that we upload when creating a release, i.e. https://github.com/stacks-network/stacks-blockchain/releases/tag/2.4.0.0.0

.github/workflows/image-build-binary.yml
wileyj: yea, some context may be helpful - this workflow is 100% required though, it creates the tagged docker images that are published to dockerhub (these are all bulilt from the binary artifacts you've asked about above). 

some better comments (as you have noted) might be a good idea

.github/workflows/clarity-js-sdk-pr.yml
wileyj: i stand corrected - this is still needed, it works different than i had thought. 
the premise here is that we can create a release tag for a release candidate, i.e. `2.4.0.0.1-rc1`, and this workflow should not be run (even though it's set to be triggered since it's a valid tag. 

one more reason it may be useful to move to a pushing  tag vs using an input tag.
an issue should be created

.github/workflows/create-cache.yml
wileyj: the caveat with re-using the cache obviously is the key. for the bitcoin binary cache, it's static (although i may see about using a version in the cache name). 
as long as the commit sha matches for a PR's cache, you can reuse an existing cache (if there has been a commit, however, a new cache is needed)

.github/workflows/bitcoin-tests.yml
wileyj: you can see an example of this partitioning here: https://github.com/stacks-network/stacks-blockchain/blob/feat/faster-ci/.github/workflows/stacks-blockchain-tests.yml#L140

i felt using a filter as well as a partition was a bit too messy - but it's totally doable if others feel strongly about going this way.

.github/workflows/create-source-binary.yml
wileyj: incidentally, these binaries are also used when creating the tagged docker images for multiple platforms

.github/workflows/cleanup.yml
wileyj: comments to the shell being run here have been added

.github/workflows/cleanup.yml
CharlieC3, line 1: What's the impetus for having this workflow? While I can imagine it would be useful in certain scenarios, having the history persist for as long as Github allows it to has saved or helped me countless times.
I agree sometimes finding the exact instance run may be challenging at times, just like digging around for a commit made a while ago, but having the record is important for error analysis, direction, education, and guidance.

.github/workflows/atlas-tests.yml
wileyj: added some top-level file comments

wileyj: > Yes, making these tests separate makes a lot of sense to me! Especially since they're so slow.
> 
> What do you think of adding more of a description to either the top of this file or the `Run Tests` job on line 101 that'll justify the history of these tests. Something like
> 
> ```yaml
> ## Github workflow for atlas tests. 
> ## Historically these tests take far longer than the others in the CI so we've separated them out
> ## For speed of testing.
> ```

I'm not sure there will be a benefit to adding comments in the `Run Tests` step - the implication of separating the tests from a single workflow is we can now parallelize *more* tests. 
I went for a logical separation based on what type of test is running (i.e. atlast, epoch, etc). 

i've added some top-level comments so it should be more clear what a `*-test.yml` yaml is doing, but i don't think there would be a benefit of adding more comments to the step itself.

.github/workflows/ci.yml
```diff
@@ -1,180 +1,233 @@
...
+  ## Runs when:
+  ##   - tag is provided
+  ##   or:
+  ##     - no tag provided
+  ##       and any of:
+  ##         - PR is approved (any approval will trigger)
+  ##         - this workflow is called manually
+  ##         - PR is opened
+  ##         - commit to either (development, master) branch
+  create-cache:
```
CharlieC3, line 144: I noticed the caches being created (save for the bitcoin cache) is using the commit sha as part of the restore key, which as you know is pretty standard. However, if the cache is only being created at the end of a PR's life, when it gets approved, I wouldn't imagine it would get used very much. Perhaps only if there are more than 1 approval.

On a similar note, why have anything trigger on approvals? Maybe there are some critical workflows that must only run after approvals which I'm missing, but my initial thoughts are this wouldn't be needed as long as the right workflows run every time a PR is synchronized.

CharlieC3 (commented):
Another PR, another series of great updates! Only a couple questions/comments for now, may leave more if I think of any.

.github/workflows/ci.yml
AshtonStephens: Yeah, do you think adding a comment would be okay? Tribal knowledge isn't very bus proof.

.github/workflows/ci.yml
wileyj: the cache is actually created anytime the `ci.yml` (or any of the test workflows, if triggered manually) is run. 
so, on PR creation, synchronization (new commit), or a PR approval - after rustfmt, the next thing that happens is the cache is either confirmed to exist or it is created for the tests that need to be run. 

Ideally, at the end of PR's approval - the cache may still exist so that action will save ~15 minutes of runtime since the commit sha will match the key (as you noted) of  an existing cache (this is also dependent on the timing of the `cleanup.yml` workflow, which is a bit aggressive currently). 

the reason for triggering workflows on PR approval is simple - i'm proposing that for opening a PR, PR synchronization, we run a subset of tests (bitcoin tests, unit tests) vs a PR approval which would run *all* tests we require for a release.  The reasoning behind this was simply to save time on running the same tests multiple time over the course of a PR.  There is an argument to be made that perhaps the only tests that should not be run as part of a PR action is what's defined in `slow-tests.yml`, this was simply the choice I've made here. 

I felt that running a limited number of tests for opening/synchronizing a PR was a good compromise as opposed to running all the tests based on any PR change (ex: if you forget to add a changelog to your PR - the workflow still needs to run currently, even though no code has changed). 

the second compromise was adding a manual way to run the tests in lieu of running them all every action run (`standalone-tests.yml`)

.github/workflows/ci.yml
wileyj: I'm not sure if adding more comments would be useful outside of what already exists. 
there already exists some docs covering how tags are used, and the comments in `ci.yml` and `github-release.yml` should be enough to fill in the gaps without being overly verbose. if you think there's something extra that can be done here, can you provide an example? i'm simply not seeing how it would improve the understanding. 

https://github.com/stacks-network/stacks-blockchain/blob/master/docs/ci-release.md
https://github.com/stacks-network/stacks-blockchain/blob/master/docs/release-process.md

.github/workflows/cleanup.yml
CharlieC3: > caches do not clean up by themselves... there is a limited amount of space available for caches

[They'll clean up automatically](https://github.com/actions/cache#cache-limits) once the 10GB limit is reached per repo, only deleting the least-used cache(s) until enough room is freed up. So we could potentially get more use out of caches if we opt to _not_ clean them up and let Github do that for us when necessary.

> i felt it important that we keep our cache usage manageable

It's not clear to me why this would need to be something managed by humans. As long as caching is being used correctly in the workflows, the caching page shouldn't be something people would need to manage or visit.

> but i've also run into this a few times where i've had to delete hundreds of older and failed jobs

I could be missing some perspective, but it's also not clear to me why this would be necessary in a real world scenario. Correct me if I'm wrong, I'm assuming you had to do this because you were testing new workflow runs in your fork, which could quickly generate hundreds of old jobs you'd care little about.
But when we have an operational set of workflows being used for different features and bug fixes, each past run, failed or successful, may hold crucial information to different devs. Even with thousands of GH actions runs, I've always been able to find the run I'm looking for through the workflow filter on the left and the run filters on the top right.

Applying a blanket cleanup automatically seems like it would do more harm than good here, like it's assuming every GH action run does not hold information anyone needs. Perhaps this can be on a manual-only trigger so we have it if people want to use it, but I don't think this should be running automatically unless this is what the core devs want.

Screenshot 2023-08-24 at 9 42 01 AM

.github/workflows/ci.yml
CharlieC3: Makes sense to me! I missed that caches were being created on every run, thanks for clarifying.

.github/workflows/cleanup.yml
wileyj: You make some valid points in regards to the cache limit (buildjet has a similar limit at 20GB before it will delete the oldest and last-accessed cache).  
for the actions, it's a different problem i was aiming to solve than the one you suggested - artifacts and logs count against the usage for actions: i've had to perform cleanups in the past and it gets quite messy. 

You're correct in that we don't **need** to delete them, but I feel that keeping our workflows in a *clean* state is beneficial to us. 

I'm not saying older actions do not hold value either - with this workflow, i'm saying that we can specify when they stop holding value and remove them so what we do care about is easier to find and won't count against our usage. 

running this workflow manually is also an option, and although i'd prefer to keep this as something that runs nightly (or even weekly) - i agree that it's better to let the core-devs make the final call on this *very opinionated* workflow. 

the reasoning i felt this workflow was necessary was quite simply that i've spent too much time on random occasions cleaning up old/failed workflows. 

as an aside - for my fork, i actually had this disabled for the reasons you've outlined, only enabling it once the other workflows were working as expected. so, i definitely see where you're coming from with this comment. 

i don't think we should delete/disable this workflow - but after hearing what core-devs would prefer, perhaps converting it to a manual trigger is the right way to go. 

great comment by the way - this is precisely what i was looking for with this PR. i've made lots of choices here - the more feedback on those choices the better!

CharlieC3: > artifacts and logs count against the usage for actions

I hadn't considered this; I can see how this could be pretty impactful if logs and artifacts are costing us in some adverse way.

And thanks for the discussion and perspective! Excited to see the impact of these awesome changes.

.github/workflows/atlas-tests.yml
wileyj: one important piece of info i forgot to add: the cargo cache actually has the binaries we need to run (`nextest`). keeping them in a cache improves the runtime of later jobs (but it's not technically required, we could always re-install and sacrifice a little time)

.github/workflows/stacks-blockchain-tests.yml
```diff
@@ -1,117 +1,321 @@
...
+          fail-on-cache-miss: true
+          path: ./target
+          key: ${{ github.event.repository.name }}-${{ github.sha }}-target
+
+      ## Symlink bitoind to $PATH
+      - name: Link Bitcoin Binary
+        id: bitcoin_binary
+        run: |
+          sudo ln -s /tmp/bitcoin/bin/bitcoind /bin/
+          ls -al /bin/bitcoind
```
jcnelson, line 99: Seems like these blocks are duplicated in a few places.  Is there an opportunity to DRY this up?

jcnelson (dismissed):
Going to approve this, but is there a way to test out these various new test workflows?

wileyj:
> Going to approve this, but is there a way to test out these various new test workflows?

not directly, no. because of the way the repo is currently configured and because of the standard we've historically had of re-using a local workflow on the master branch only, it makes fully testing the changes here impossible without a hardcoded fork. 


however, @jbencin raised the idea of using directives like `uses: ./.github/workflows/....` vs tying a job step to `repo@branch` for worfklows in the stacks-blockchain repo: https://github.com/stacks-network/stacks-blockchain/pull/3888

this might be a good place to have that discussion of how we re-use workflows vs always using `repo@branch` when calling a local workflow, and then document when and how we would allow using a local workflow and when to use `repo@branch`.

wileyj:
> 

Bringing in changes from @jbencin PR #3895 - this should allow us to test the changes directly. 
But, it does change the standard we've used in the past where we call most re-usable workflows using the repo name, i.e. `uses: stacks-network/stacks-blockchain/.github....@master`
and replaces it with `uses: ././github/...yml`

.github/workflows/stacks-blockchain-tests.yml
wileyj: with the current workflow, i don't think it's possible to DRY this. since the bitcoin binary needs to be present for multiple jobs in the workflow. 

I have an idea on how to DRY this, but i'm not sure if:
a) it'll even work
b) it won't make things messier

i can create a test for this and see what happens. 

as to your and @kantai ask about workflows that have run - i think i can get a few execs in the the stacks-blockchain repo so you an see how it works in practice vs my personal fork where i've been testing these changes. 

in light of https://github.com/stacks-network/stacks-blockchain/actions/runs/6511171596/job/17688907514?pr=3950 as well, i think i'm going to add some logic to perform a consistent cleanup of the VM across several workflows (vs targeting them when we hit an error like @xoloki noticed in the linked action)

wileyj:
closing in favor of https://github.com/stacks-network/stacks-core/pull/4079

PR #3850, opened 2023-08-18
Feat/stacks signer

jcnelson:
This PR adds in the event observer system for StackerDB chunk arrivals, and adds a `libsigner` crate which consumes them and feeds them into a signer-specific runloop.  StackerDB events contain the chunk data, so downstream consumers do not need to go and pull them.

`libsigner` is a basic crate that just implements an event observer client for `StackerDB` messages, and a means of passing them into an implementation-specific runloop body.  See the unit test in `libsigner/src/tests/mod.rs` for example usage.


kantai (commented):
The stacks signer seems like it could use a request library like `ureq` or `reqwest` rather than implementing a new HTTP client.

jcnelson:
> The stacks signer seems like it could use a request library like ureq or reqwest rather than implementing a new HTTP client.

Sure; I just added a really simple one for testing / debugging purposes. The stacker-signer will code to the trait, not this particular implementation.

libsigner/src/events.rs
```diff
@@ -0,0 +1,268 @@
...
+use libstackerdb::StackerDBChunkData;
+
+use serde::{Deserialize, Serialize};
+
+use crate::http::{decode_http_body, decode_http_request};
+use crate::EventError;
+
+/// Event structure for newly-arrived StackerDB data
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct StackerDBChunksEvent {
```
jferrant, line 38: Would this code be eligible to be moved to the stacks-core SDK that @stjepangolemac is writing?

libsigner/src/events.rs
```diff
@@ -0,0 +1,268 @@
+// Copyright (C) 2013-2020 Blockstack PBC, a public benefit corporation
```
jferrant, line 1: Would this stacker db communication code be eligible to be moved to the stacks-core SDK that @stjepangolemac is writing?

libsigner/src/error.rs
```diff
@@ -0,0 +1,131 @@
...
+pub enum RPCError {
+    IO(io::Error),
+    Deserialize(String),
+    NotConnected,
+    MalformedRequest(String),
+    MalformedResponse(String),
+    HttpError(u32),
+}
+
+impl fmt::Display for RPCError {
```
jferrant, line 32: I think a lot of this can be simplified if you use thiserror.

libsigner/src/events.rs
stjepangolemac: I skimmed through the changes and it seems to me that the way it works is:

1. You run a HTTP server
2. Node sends events to it through a predetermined set of HTTP routes

I'm not sure how you "register" or let the node know where to send the events. Still, the communication looks fairly simple. I suppose we could start with any kind of HTTP server implementation in `stacks-core`.

libsigner/src/events.rs
```diff
@@ -0,0 +1,268 @@
...
+    type ST: EventStopSignaler + Send + Sync;
+
+    /// Open a server socket to the given socket address.
+    fn bind(&mut self, listener: SocketAddr) -> Result<SocketAddr, EventError>;
+    /// Return the next event
+    fn next_event(&mut self) -> Result<StackerDBChunksEvent, EventError>;
+    /// Add a downstream event consumer
+    fn add_consumer(&mut self, event_out: Sender<StackerDBChunksEvent>);
+    /// Forward the event to downstream consumers
+    fn forward_event(&mut self, ev: StackerDBChunksEvent) -> bool;
```
netrome, line 63: What's the intention of the returned bool? It seems to me like we'd want to have a result returned here to convey more information as I assume forwarding an event is a fallible operation.

libsigner/src/events.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub fn new(sig: Arc<AtomicBool>, local_addr: SocketAddr) -> StackerDBStopSignaler {
+        StackerDBStopSignaler {
+            stop_signal: sig,
+            local_addr,
+        }
+    }
+}
+
+impl EventStopSignaler for StackerDBStopSignaler {
+    fn send(&mut self) {
```
netrome, line 164: Nit: I sugesst using a more explicit name here.
```suggestion
    fn send_stop_signal(&mut self) {
```

libsigner/src/events.rs
```diff
@@ -0,0 +1,268 @@
...
+    fn forward_event(&mut self, ev: StackerDBChunksEvent) -> bool;
+    /// Determine if the receiver should hang up
+    fn is_stopped(&self) -> bool;
+    /// Get a stop signal instance that, when sent, will cause this receiver to stop accepting new
+    /// events.  Called after `bind()`.
+    fn get_stop_signaler(&mut self) -> Result<Self::ST, EventError>;
+
+    /// Main loop for the receiver.
+    /// Typically, this is started in a separate thread.
+    fn main_loop(&mut self) {
```
netrome, line 72: Is there any point in keeping the EventReceiver once the main loop has terminated? Otherwise I suggest having this loop take ownership of the event receiver.
```suggestion
    fn main_loop(mut self) {
```

libsigner/src/http.rs
```diff
@@ -0,0 +1,241 @@
...
+use crate::error::RPCError;
+
+pub const MAX_HTTP_HEADERS: usize = 32;
+pub const MAX_HTTP_HEADER_LEN: usize = 4096;
+
+/// Decode the HTTP request payload into its headers and body.
+/// Returns (verb, path, table of headers, body_offset) on success
+pub fn decode_http_request(
+    payload: &[u8],
+) -> Result<(String, String, HashMap<String, String>, usize), EventError> {
```
netrome, line 37: Could you make the result tuple an explicit type or at least a type alias for readability?

libsigner/src/runloop.rs
```diff
@@ -0,0 +1,244 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#![allow(unused_imports)]
+#![allow(dead_code)]
+
```
netrome, line 19: Can we clean up any unused imports or dead code before merging this?

libsigner/src/runloop.rs
```diff
@@ -0,0 +1,244 @@
...
+/// Implement this, and you get all the multithreaded setup for free.
+pub trait SignerRunLoop<R> {
+    /// Hint to set how long to wait for new events
+    fn set_event_timeout(&mut self, timeout: Duration);
+    /// Getter for the event poll timeout
+    fn get_event_timeout(&self) -> Duration;
+    /// Run one pass of the event loop, given new StackerDB events discovered since the last pass.
+    /// Returns Some(R) if this is the final pass -- the runloop evaluated to R
+    /// Returns None to keep running.
+    fn run_one_pass(&mut self, events: Option<StackerDBChunksEvent>) -> Option<R>;
```
netrome, line 54: "events" suggests there are more than one event. Should this be a collection/iterator instead of an option?

libsigner/src/session.rs
```diff
@@ -0,0 +1,216 @@
...
+        &mut self,
+        host: SocketAddr,
+        stackerdb_contract_id: QualifiedContractIdentifier,
+    ) -> Result<(), RPCError>;
+    fn list_chunks(&mut self) -> Result<Vec<SlotMetadata>, RPCError>;
+    fn get_chunks(
+        &mut self,
+        slots_and_versions: &[(u32, u32)],
+    ) -> Result<Vec<Option<Vec<u8>>>, RPCError>;
+    fn get_latest_chunks(&mut self, slot_ids: &[u32]) -> Result<Vec<Option<Vec<u8>>>, RPCError>;
```
netrome, line 44: Why do we need to wrap the inner Vec with an Option?

libsigner/src/session.rs
```diff
@@ -0,0 +1,216 @@
...
+                    return Err(e);
+                }
+            };
+            payloads.push(chunk);
+        }
+        Ok(payloads)
+    }
+
+    /// query the replica for zero or more latest chunks
+    fn get_latest_chunks(&mut self, slot_ids: &[u32]) -> Result<Vec<Option<Vec<u8>>>, RPCError> {
```
netrome, line 185: What does "latest" mean here?

stacks-signer/src/main.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    let mut session = StackerDBSession::new(host.clone(), contract.clone());
+    session.connect(host, contract).unwrap();
+    let chunk_ack = session.put_chunk(chunk).unwrap();
+    println!("{}", serde_json::to_string(&chunk_ack).unwrap());
+    process::exit(0);
+}
+
+fn main() {
+    let argv: Vec<String> = env::args().collect();
```
netrome, line 271: I would prefer to parse the arguments to a concrete type instead of passing a vector of strings around to each sub-command.

libsigner/src/events.rs
```diff
@@ -0,0 +1,268 @@
...
+                }
+            }
+            return true;
+        }
+    }
+
+    /// Add an event consumer.  A received event will be forwarded to this Sender.
+    fn add_consumer(&mut self, out_channel: Sender<StackerDBChunksEvent>) {
+        self.out_channels.push(out_channel);
+    }
```
netrome, line 254: Is there any plan to include a mechanism for new consumers to sync up with old events? Or do we assume consumers have to separately query the node RPC endpoint to get the latest StackerDB state when starting to consume events?

netrome (commented):
A few thoughts, questions and comments from me.

kantai (commented):
I disagree with implementing a new HTTP server in this PR. I have a lot of reasons for this. 

Implementing an HTTP server does not seem like a chief concern for the signer binary, and there are no unique requirements for the signer that seem to necessitate a custom-written HTTP server. My thoughts on an HTTP server here are basically identical to the thoughts discussed here (https://github.com/stacks-network/stacks-blockchain/discussions/3761#discussioncomment-6377420) about the HTTP server in the stacks-blockchain. The desire to implement _a second_ HTTP server is another datapoint in favor of replacing this PR's server and the existing HTTP server with a library. As mentioned in the linked discussion, the existing HTTP server is painful to debug, difficult to monitor the performance of, requires working around its deficiencies (see: https://github.com/stacks-network/stacks-blockchain/issues/3864), and is difficult to implement new endpoints for. A new HTTP server implementation is that and more: it doesn't have the advantage of being previously deployed, and no attempts have been made yet to make it extensible (and it will need to handle other kinds of events).

jcnelson:
> Implementing an HTTP server does not seem like a chief concern for the signer binary, and there are no unique requirements for the signer that seem to necessitate a custom-written HTTP server.

The choice here is between this small custom HTTP server (a few hundred lines of code) and a 3rd party library that is between one and three orders of magnitude larger.  Considering that the signer process is a piece of high-assurance software that must be live for the Stacks blockchain to make progress, it seems to me that it's overall a better decision to keep the total TCB small, even if it means applying a little bit of elbow grease.

It's way easier to audit this.

> My thoughts on an HTTP server here are basically identical to the thoughts discussed here (#3761 (reply in thread)) about the HTTP server in the stacks-blockchain. The desire to implement a second HTTP server is another datapoint in favor of replacing this PR's server and the existing HTTP server with a library.

The key difference between the HTTP server in the Stacks node and the HTTP server here is that the latter has exactly one client: the Stacks node event dispatcher.  It's far simpler a task to code to that one client's needs than the multitude of clients that could talk to the Stacks node.

At a higher level, I'd like to call into question why the Stacks node event dispatcher needs to use HTTP _at all_ to send events to downstream consumers.  Considering the event consumer is a colocated process, a named pipe would have probably been better.  At least with a named pipe, we could avoid the overhead of using three separate network protocols (IP, TCP, HTTP) to marshal and unmarshal data, and just sent the JSON directly.

> As mentioned in the linked discussion, the existing HTTP server is painful to debug, difficult to monitor the performance of, requires working around its deficiencies (see: #3864), and is difficult to implement new endpoints for. A new HTTP server implementation is that and more: it doesn't have the advantage of being previously deployed, and no attempts have been made yet to make it extensible (and it will need to handle other kinds of events).

I think this is apples to oranges.  I'm already in the process of addressing all of those concerns regarding the Stacks HTTP server via #3884, and the HTTP server here is only ever meant to receive signing data from the Stacks node's event dispatcher (which always arrives via StackerDB events).  Arguably, the Stacks node event dispatcher is over-engineered as it is.

kantai:
> The choice here is between this small custom HTTP server (a few hundred lines of code) and a 3rd party library that is between one and three orders of magnitude larger. Considering that the signer process is a piece of high-assurance software that must be live for the Stacks blockchain to make progress, it seems to me that it's overall a better decision to keep the total TCB small, even if it means applying a little bit of elbow grease.

This obfuscates the actual tradeoff being made here. A third party library is more complex, yes, but that complexity helps it support all sorts of things. But, in any event, I do not accept the premise of this question: the complexity of a third party library does not matter in the same way that the complexity of written, and now re-written, code in the stacks-blockchain matters. A third party library comes with its own maintainers and its own support systems. Writing an HTTP in this repo is instead a commitment to maintain, update, and piecemeal reimplement standard HTTP server features entirely by maintainers of this repo. I will re-quote my feelings from the linked discussion: 

> [Existing HTTP libraries in rust] have rigorous security policies, taking full advantage of RustSec, CVE databases, and Github's security reporting. They're widely used and depended on by multiple other blockchain projects. Polkadot, near, solana, and diem all use tokio and hyper--- nearly all rust blockchains I've examined do. I do not believe that the stacks-blockchain has wildly different trust requirements for dependencies than other blockchains, and I am very skeptical of the idea that we are getting any safety benefit at all from this kind of policy.

I believe all of those points continue to apply. I do not believe there is any safety benefit to be gained from implementing multiple custom-written HTTP servers (I suspect that any safety impact of such would in fact be negative).

> At a higher level, I'd like to call into question why the Stacks node event dispatcher needs to use HTTP at all to send events to downstream consumers. Considering the event consumer is a colocated process, a named pipe would have probably been better. At least with a named pipe, we could avoid the overhead of using three separate network protocols (IP, TCP, HTTP) to marshal and unmarshal data, and just sent the JSON directly.

JSON over HTTP is perhaps the most widely supported protocol in existence. It allows downstream consumers to write a a capable receiver in any language in minutes. Abandoning this to avoid the overhead of HTTP doesn't seem worthwhile. In any event, to the best of my knowledge, the assumption that event consumers are always "colocated processes" isn't correct in the way that you mean. I believe that a typical deployed instance of https://github.com/hirosystems/stacks-blockchain-api in kubernetes would use two different pods.

> In other words, we're only having this conversation because the Stacks node event dispatcher implementation was itself over-engineered. I'd prefer to address that problem first.

Replacing a new custom-written HTTP server with a new custom-written server for a new custom protocol wouldn't address my concerns.

jcnelson:
> This obfuscates the actual tradeoff being made here. A third party library is more complex, yes, but that complexity helps it support all sorts of things. But, in any event, I do not accept the premise of this question: the complexity of a third party library does not matter in the same way that the complexity of written, and now re-written, code in the stacks-blockchain matters. A third party library comes with its own maintainers and its own support systems. Writing an HTTP server in this repo is instead a commitment to maintain, update, and piecemeal reimplement standard HTTP server features entirely by maintainers of this repo. I will re-quote my feelings from the linked discussion:

Okay, a few things to unpack here.

First, what are the things you believe the HTTP server _in this specific use-case_ will need to support?  The signer daemon's HTTP server is only going to receive JSON blobs from the event dispatcher, right?  What else could we possibly be sending here?  I'm asking because I prefer to use the least-complex code to address the task at hand.  If we can't think of anything else, then perhaps it's too early to be committing to a 3rd party library with lots of extra complexity that we won't need.

Second, your argument that the complexity of the 3rd party library does not matter the same way as a bespoke HTTP server assumes that we somehow are off the hook for maintaining it.  This isn't as true as you seem to think.  While it is the case that a 3rd party library spares us the initial effort of writing the thing, it does not free us from having to deal with it if it breaks.  While dealing with breakage could be as simple as bumping the version of the package we use, it could also be as complex as patching the implementation ourselves.

Third, your perspective on this seems to be tainted from your experience with dealing with the node's HTTP server.  While I agree that there is a lot of room for improvement there (and it is being addressed), we're not talking about that code here.  We're trying to figure out the best way for the signer to receive StackerDB events from the Stacks node.

> [Existing HTTP librares in rust] have rigorous security policies, taking full advantage of RustSec, CVE databases, and Github's security reporting

Great!  But, these are also _reactive_ security policies.  Where are the _proactive_ security policies that prevent CVEs from arising in the first place?

Speaking of, have you ever delved into the code of `tokio` and `hyper`?  It's a trip.  There's exactly 1.013 metric crap-tons of `unsafe` code in there.  And you know what else?  A _disturbing_ amount of it has to do with consuming data from the network.  And that's just those two packages.  I haven't yet looked at all their dependencies, but I do know that `httparse`, one of `hyper`'s dependencies, is also full of `unsafe` code (so much that I decided it was safer to vendor it and replace the `unsafe` code with safe equivalents).

Now, that by itself is not direct evidence of security bugs, but ask yourself this:  why should we trust these `unsafe` blocks to be 100% free of memory safety errors?  I don't know about you, but I haven't seen anything to convince me that the human capacity for making these errors has decreased in the last 10 years.  You don't have to answer that, though; some researchers at Georgia Tech [already did](https://dl.acm.org/doi/pdf/10.1145/3477132.3483570), and found tons of new memory-related CVEs in `unsafe` code (over 50% of all RustSec CVEs in 2016), including a couple in the standard library (!!) and in the official `futures` crate (!!).  And that's just from an automated analyzer they wrote; their numbers are an undercount.  So given that, what makes these packages so special that we can assume they have no such bugs?  We can do so for my code at least -- no `unsafe`s at all!

I'd rather deal with code that is proactively safe than code that is not proven to be safe today (but at least I'll get a nice, timely email about it once an exploit is discovered).

> Replacing a new custom-written HTTP server with a new custom-written server for a new custom protocol wouldn't address my concerns.

Can you find me a suitable HTTP server (better than any server we could write) that isn't full of `unsafe` code, and does not pull in a crap-load of `unsafe` code as dependencies?  I've looked, and have yet to find one.  But, I'm afraid I must insist on this.  The reason Rust exists is because it prevents fallible humans from introducing memory safety bugs (and data races between threads), and the history of CVEs in system software is mostly comprised of dangling pointers, double-frees, invalid pointer dereferences, out-of-bounds reads and writes, and so on.

Btw, if you know of anything better than `mio` I'm also all ears.

jcnelson:
> Btw, if you know of anything better than mio I'm also all ears. Lots of unsafe code there, but some of it (namely the FFI stuff) is unavoidable.

Answering my own question here -- it looks like [polling](https://docs.rs/polling/latest/polling/) is what I'm looking for.  Yes, it uses `unsafe`, but only in an FFI capacity.  It uses [rustix](https://docs.rs/rustix/latest/rustix/) for UNIX compatibility, which is maintained by the same team (Bytecode Alliance) that ships the WASM interpreter the Clarity team is using.

libsigner/src/events.rs
netrome: Sure, personally I'd prefer to convey that information more explicitly as
```
fn forward_event(&mut self, ev: ...) -> Result<(), ForwardError>;

enum ForwardError{
  ConsumerChannelBroken
}
```

Although this is more verbose, it's clearer to read and conveys more information about the system. Regarding complexity, the `Result<(), ForwardError>` has cardinality of two so it is as simple as a boolean.

libsigner/src/events.rs
netrome: Not necessarily for all structs implementing the trait, but sure - for the StackerDBStopSignaler I wouldn't imagine anything else being sent through this method.

libsigner/src/events.rs
netrome: Fair point :ok_hand:

libsigner/src/http.rs
netrome: Thank you :pray:

stacks-signer/src/main.rs
netrome: Yeah and I see clap is coming here soon https://github.com/stacks-network/stacks-blockchain/pull/3885

libsigner/src/events.rs
netrome: Not sure, but I would anticipate there being some initial state per signing round the signer should catch up on when connecting - such as knowing which the other signers are. I don't have the full picture in my head though so you know this better than I.

jcnelson:
> That's correct. I disagree with custom-writing an HTTP server in projects whose main concern is not HTTP. It violates several well-trod principals of software engineering practices. The state of the HTTP server in the stacks node proves this out to me for reasons I've discussed elsewhere.

I agree in principle.  But, how well does this principle hold up when the project needs an HTTP server when there are no good 3rd party options to be had?  That's the position I find myself in.

I've already conceded that the Stacks HTTP server could use some (major) refactoring, and as I've said earlier, I'm working on this as we speak.  But that's neither here nor there when it comes to this PR.

> I disagree with repeating this process again with yet another HTTP server.

Look, there are basically four options:

1. Use a 3rd party HTTP server of sufficient quality 
2. Put the Stacks HTTP server and client (once it's been refactored) into its own repository, and use that instead
3. Accept the HTTP server in this PR.
4. Don't use HTTP

The jury is still out as to whether or not option 1 even exists.  I've not been impressed by the ecosystem offerings so far, but I would like to be proven wrong.

I'm willing to execute option 2 both here and on the Stacks node, if you're happy with the refactored HTTP server.  We can put the HTTP server implementation into its own repository, and link it into both projects as an external crate.  This at least ought to satisfy your desire that we avoid "custom-writing an HTTP server in projects whose main concern is not HTTP."

Option 3 is what this PR does, but it sounds like you're steadfastly opposed to taking it. It also sounds like you're not a fan of option 4 either.

stjepangolemac:
I'd like to chime in on the discussion here.

> But, how well does this principle hold up when the project needs an HTTP server when there are no good 3rd party options to be had?

What's the criteria for HTTP servers, or any crates we might use? Not having unsafe code in its source, and sources of all direct and indirect dependencies, is a requirement that's theoretically admirable but in practice has absolutely terrible tradeoffs.

image

@kantai mentioned a few notable projects that use crates widely regarded as standards in the Rust community. Some of them are:

- Solana, market cap of 8B (ATH 78B)
- Polkadot, market cap of 3B (ATH 53B)

I'm sure there's a lot of other very important codebases using crates like tokio, hyper, etc. [Here](https://github.com/paradigmxyz/reth) is an Ethereum node implementation in Rust using all of that.

Most of the time if you want to produce something useful and have limited time and resources the only path forward is to make tradeoffs.

https://xkcd.com/844/

jcnelson:
> What's the criteria for HTTP servers, or any crates we might use? Not having unsafe code in its source, and sources of all direct and indirect dependencies, is a requirement that's theoretically admirable but in practice has absolutely terrible tradeoffs.

The HTTP server in this PR is a few hundred lines of 100% safe Rust code, so this is definitely doable.

The Stacks HTTP server, for all its API warts, is also 100% safe Rust code.  And, the warts are straightforward to remove (being done now).

> Solana, market cap of 8B (ATH 78B)
> Polkadot, market cap of 3B (ATH 53B)

Do market caps prove memory safety?

> I'm sure there's a lot of other very important codebases using crates like tokio, hyper, etc. Here is an Ethereum node implementation in Rust using all of that.

Does importance prove memory safety?

Did the combined importance and economic might of the Fortune 500, the world's stock exchanges, the world's leading computer companies, and everything that depends on them, prevent OpenSSL from shipping with Heartbleed?

stjepangolemac:
No but the world is still in a better spot with Heartbleed than without OpenSSL.

libsigner/src/http.rs
xoloki, line 1: With the exception of the User-Agent, nothing in this file seems specific to libsigner.  Why is this here?

jcnelson: It has nowhere else to go.  But, I can put it into a separate but very small crate.

jcnelson:
> No but the world is still in a better spot with Heartbleed than without OpenSSL.

Except that's not what's happening here.  We have a very simple HTTP server that's just functional enough to handle events passed to it from the Stacks node (because that's all it's going to be used for), and moreover, it's 100% memory-safe.  This is more akin choosing between a memory-safe ECDSA signer and verifier versus linking in the whole of OpenSSL.

jcnelson:
@kantai @stjepangolemac @xoloki @jferrant There are a few PRs that are blocked on this PR.  Can we merge this to `develop`?  It's becoming a staging ground for a bunch of signer PRs that really should be hitting `develop`.  Thanks!

libsigner/src/http.rs
```diff
@@ -0,0 +1,269 @@
...
+
+use crate::error::EventError;
+use crate::error::RPCError;
+
+pub const MAX_HTTP_HEADERS: usize = 32;
+pub const MAX_HTTP_HEADER_LEN: usize = 4096;
+
+/// Decoding of the relevant parts of a signer-directed HTTP request from the Stacks node
+#[derive(Debug)]
+pub struct SignerHttpRequest {
```
xoloki, line 35: Since we're using `tiny_http` now, probably all of this can go away.

libsigner/src/tests/http.rs
```diff
@@ -0,0 +1,347 @@
...
+use std::io::Write;
+use std::str;
+
+use crate::error::EventError;
+use crate::error::RPCError;
+use crate::http::{decode_http_body, decode_http_request, decode_http_response, run_http_request};
+
+use stacks_common::util::chunked_encoding::*;
+
+#[test]
```
xoloki, line 29: This can also probably all go away since we're using `tiny_http`

xoloki (approved):
Looks good!  My only real comment is that it would be great to remove as much custom `HTTP` code as possible, since we're using `tiny_http` internally now in `libsigner`.  But I'd be happy if this all went in as-is.

libsigner/src/http.rs
xoloki: This happened in the first `Signer` PR:

https://github.com/stacks-network/stacks-blockchain/blob/feat/stacks-signer/libsigner/Cargo.toml#L29
https://github.com/stacks-network/stacks-blockchain/blob/feat/stacks-signer/libsigner/src/events.rs#L33
https://github.com/stacks-network/stacks-blockchain/blob/feat/stacks-signer/libsigner/src/events.rs#L201

We're currently using it as an `HTTP` server for `StackerDBEventReceiver`.  So all of the server side `HTTP` parsing can go away in `libsigner`, since `tiny_http` comes with its own parser and request/response objects, along with all the tests.

In a later PR we can explore getting rid of `async_h1`, which is used for the client side of the stacks event observer code.

xoloki: So AFAICT all of the `libsigner` custom `HTTP` code is now dead.

stjepangolemac (approved):
I haven't seen anything obviously wrong or weird, but the PR is large so take this with a grain of salt.

libsigner/src/tests/http.rs
jferrant: Lines 29 through 101 can but until we update SignerSession, we can't get rid of the rest.

jferrant (approved):
Aside from potentially deleting the portions of dead HTTP code, might want to make a ticket to fully remove it from SignerSession and to use tiny http instead. Aside from that would just merge and build smaller PRs off of Develop. 

So TLDR. LGTM

Issue #3856, opened 2023-08-22
Nakamoto: Event interface for StackerDB notifications


kantai:
To avoid any need to perform a polling loop, binaries like the block
producer and stacker signer need to be able to receive push
notifications when relevant StackerDB data arrives in their paired
`stacks-node`.

Unless it becomes an issue for stalling (which I don't expect), this
should just use the existing event observer interface. This means
adding new events for StackerDB notifications.

jcnelson:
This is now merged!

PR #3870, opened 2023-08-22
Feat/stackerdb event observer

jcnelson:
Address #3856.  Event observers can subscribe to stacker DB events by listing the "stackerdb" event concern in their configs, or the "any" concern.  `StackerDBChunkEvent` JSON objects will get pushed to `/stackerdb_chunks` for each successful chunk write (described in `docs/event_observer.md`).

### DIFF
--- a/stackslib/src/net/relay.rs
+++ b/stackslib/src/net/relay.rs
@@ -44,7 +44,9 @@ use crate::net::http::*;
 use crate::net::p2p::*;
 use crate::net::poll::*;
 use crate::net::rpc::*;
-use crate::net::stackerdb::{StackerDBConfig, StackerDBSyncResult, StackerDBs};
+use crate::net::stackerdb::{
+    StackerDBConfig, StackerDBEventDispatcher, StackerDBSyncResult, StackerDBs,
+};
 use crate::net::Error as net_error;
 use crate::net::*;
 use crate::types::chainstate::StacksBlockId;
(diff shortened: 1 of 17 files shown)


stackslib/src/net/relay.rs
```diff
@@ -1701,11 +1736,38 @@ impl Relayer {
         }
     }
 
+    /// Process HTTP-uploaded stackerdb chunks.
+    /// They're already stored by the RPC handler, so just forward events for them.
+    pub fn process_uploaded_stackerdb_chunks(
+        uploaded_chunks: &[StackerDBPushChunkData],
+        event_observer: Option<&dyn StackerDBEventDispatcher>,
```
jferrant, line 1743: Seems a bit weird for this to take an optional when it is a no-op when it is None. Is this just to maintain consistency in function params to other functions?

jcnelson: The reason this code and other methods in this module take an `Option<..>` for an event dispatcher is because the dispatcher implementation that gets used in production lives in the Stacks node proper.  By keeping it as an `Option<..>`, we can both write unit tests for this code without introducing a dependency on the node, and we can mock the dispatcher when appropriate.

kantai (approved):
LGTM -- just had a couple pretty superficial comments.

jferrant (approved):

stackslib/src/net/stackerdb/mod.rs
```diff
@@ -291,6 +291,16 @@ impl StackerDBSyncResult {
...
 }
 
+/// Event dispatcher trait for pushing out new chunk arrival info
+pub trait StackerDBEventDispatcher {
+    /// A set of one or more chunks has been obtained by this replica
+    fn new_stackerdb_chunks(
+        &self,
+        contract_id: &QualifiedContractIdentifier,
+        chunk_info: &[StackerDBChunkData],
+    );
```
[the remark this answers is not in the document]
jcnelson: Fixed

PR #3872, opened 2023-08-23
docs: Fix pre-commit hook, and simplify/parallelize using `xargs`

jbencin:
Current pre-commit hook in docs pipes `stderr` to /dev/null. This is how it ignores non-Rust files, but it also allows broken code in Rust files. I changed to use `grep` to filter out files that don't end in `.rs` instead

Also, I changed it to use `xargs`, which results in less code and allows it to check multiple files in parallel


jcnelson (approved):
LGTM

obycode (approved):
lgtm!

PR #3873, opened 2023-08-23
updating startup scripts

wileyj:
With the `stacks.system` unitfile, stopping the service during when not at chain tip was resulting in a panic (thread stopped). 
The methods used in the current systemd unit no longer appear to be needed where a PID is recorded via forking the process using `sh`. 

Using a standard execstart line of `ExecStart=/usr/local/bin/stacks-node start --config=/etc/stacks-blockchain/Config.toml`
seems more than sufficient now - start/stop/restart all work as expected during a sync as well as when the chain has reached tip. 

Minor correction was made in the `stacks.init` script to use the default logfile mentioned in the comment vs a non-existent `/stacks-blockchain` directory (this is configurable by the user).


contrib/init/stacks.service
```diff
@@ -38,6 +35,10 @@ Group=stacks
 RuntimeDirectory=stacks-blockchain
 RuntimeDirectoryMode=0710
 
+# /etc/bitcoin
```
CharlieC3, line 38: Small clarification
```suggestion
# /etc/stacks-blockchain
```

contrib/init/stacks.service
```diff
@@ -58,5 +59,9 @@ NoNewPrivileges=true
 # such as /dev/null, /dev/zero and /dev/random.
 PrivateDevices=true
 
+# Deny the creation of writable and executable memory mappings.
+MemoryDenyWriteExecute=true
```
CharlieC3, line 63: Since you're adding this, can we also add `SystemCallArchitectures=native`?
> [Specifically, it is recommended to combine this option with SystemCallArchitectures=native or similar](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#MemoryDenyWriteExecute=)

CharlieC3 (approved):
Just a couple minor comments, but otherwise LGTM! :shipit:

contrib/init/stacks.service
wileyj: will do!

contrib/init/stacks.service
wileyj: hmm, i'll run some tests with this change but i'm not opposed to adding it.

jcnelson (approved):
Blindly approving this.  I famously don't use `systemd` ;)  But if it works on your end and Charlie's end, I'm sure it's fine.

CharlieC3 (approved):
Send it! :shipit:

PR #3879, opened 2023-08-28
Feat: Add `/v2/block_proposal` endpoint

jbencin:
### Description

Add `/v2/block_proposal` endpoint which will be used by future miner to validate proposed blocks

Note that this PR also includes some minor refactoring and style changes

To run integration tests for new endpoint:
```sh
BITCOIND_TEST=1 cargo test block_proposal -- --include-ignored
```

### Checklist
- [x] Changelog is updated


kantai:
This is great, @jbencin!

My first note is that I think that this PR requires a high-level change before it can move to ready from draft status.

Since block validation is pretty slow, it shouldn't block the network stack by executing the validation in the RPC handler directly (see issue: https://github.com/stacks-network/stacks-blockchain/issues/3864)

I think there's two good options for implementing this:

1. Use the current `NetworkResult` scheme for message passing.
2. Introduce a new comms channel using [`channel`](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html#)

Using `NetworkResult` has the advantage of already being in place, so there's no need to make sure a new `Sender` struct is passed to the RPC handler. However, it has a major disadvantage when it comes to handling an already-in-progress validation: if a block proposal has already been submitted and has not yet been processed, the RPC handler doesn't have an easy way to return an error to the requester (e.g., through a 429 HTTP response). If a single-proposal length sync channel is used, then the RPC handler knows immediately if the channel is full and can return a 429 quickly.

In either case, the actual block proposal validation should occur in a thread spawned by the `neon_node`. The result of the validation should then be sent over the event observer interface.

jbencin:
A couple questions:
- Are there existing RPC endpoints that already use the desired threading model?
- If the answer to the above is "no", then is this worth doing? I ask because I've heard we are currently working on replacing the HTTP stack. I don't know how far along that work is and how it will handle threading though...

jcnelson:
> Work on improving (or replacing) the current HTTP stack is at least a ways off (and there's still plenty contentious about it), so its important that the endpoints required for supporting Nakamoto play nicely with the current stack.

FWIW I'm currently refactoring the HTTP stack to cleanly separate transport concerns from business concerns, so all API endpoints' business logic can just be put into their own files.

jbencin:
> FWIW I'm currently refactoring the HTTP stack to cleanly separate transport concerns from business concerns

If you think it will be production ready before we ship Nakamoto, maybe we can wait

> I think there's two good options for implementing this:

I might be able to do something simpler and just pass a `StacksHttp` struct and the file descriptor (which is all we need to send a response) to a new thread using `std::thread::spawn()`, send a response, and then terminate the thread. This should work as long as it's safe to have multiple threads using the Chainstate and Burnchain databases

Or maybe we do some benchmarking first, and if it's not too bad, we leave the optimizations as a TODO for later?

kantai:
>> FWIW I'm currently refactoring the HTTP stack to cleanly separate transport concerns from business concerns
> If you think it will be production ready before we ship Nakamoto, maybe we can wait

I don't think that refactoring would impact this decision. The issue is that the network stack is single-threaded, and so long RPC calls can stall it. The discussed refactoring there may or may not change that.

> I might be able to do something simpler and just pass a StacksHttp struct and the file descriptor (which is all we need to send a response) to a new thread using std::thread::spawn(), send a response, and then terminate the thread. This should work as long as it's safe to have multiple threads using the Chainstate and Burnchain databases

The rationale for passing the request on to the `neon_node` is to try to centralize the thread management of the system so that only the main thread is really responsible for spawning other threads.

> Or maybe we do some benchmarking first, and if it's not too bad, we leave the optimizations as a TODO for later?

StacksBlock processing (which the validation does) can take up to a minute depending on the Stacks block size -- that's pretty much a non-starter for the network thread.

Another option when it comes to this PR specifically is to accept this PR with the synchronous RPC endpoint, and introduce the asynchronous system in a subsequent PR. I'd be okay with this -- I already have some work for the communication channel approach.

jbencin:
> The rationale for passing the request on to the neon_node is to try to centralize the thread management of the system so that only the main thread is really responsible for spawning other threads

So what's the planned threading model here...
- Pass the request to an existing thread?
- Create a single new thread to handle all long-running HTTP requests?
- Create a new thread for each request?

> Another option when it comes to this PR specifically is to accept this PR with the synchronous RPC endpoint

I'd prefer this. I think smaller PRs focused around a single topic are preferable and easier to review. Right now my main concern is correctness of the `/v2/block_proposal` endpoint.

However, since this operation can be slow and I don't want to introduce an easy method for DOS attacks, I think that before merging, I should mitigate that risk by rejecting any request to this endpoint that is not from `localhost`

jbencin:
I made a few more improvements here, including rejecting requests that are not from a local loopback address. I'm now going to...

- Merge @kantai's changes from #3889 and resolve conflicts
- Merge `develop` and resolve conflicts
- Update docs (changelog, openapi.yaml, etc.)

Then this PR should be ready for review and merge

jbencin:
The code here is ready for review (still work to do on docs). Note that `BITCOID_TEST` tests fail, they just hang and time out after like 20 minutes. I think RPC communication to `bitcoind` is broken, but I see the same behavior in `develop` as of f4808469e, so it's not this PR that broke it.

jbencin:
All docs are updated, ready for review

jbencin:
@kantai @jcnelson I've merged the latest `develop` into this PR, it's ready for review

stackslib/src/chainstate/stacks/miner.rs
```diff
@@ -2661,3 +2661,244 @@ impl StacksBlockBuilder {
...
+pub enum BlockValidateResponse {
+    Ok(BlockValidateOk),
+    Reject(BlockValidateReject),
+}
+
+/// A response for block proposal validation
+///  that the stacks-node thinks is acceptable.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct BlockValidateOk {
+    pub block: StacksBlock,
```
jcnelson, line 2719: Just flagging here that this will eventually need to be ported to `NakamotoBlock`.

stackslib/src/net/connection.rs
```diff
@@ -364,6 +372,9 @@ pub struct ConnectionOptions {
     /// how long a mempool sync is allowed to take, in total, before timing out
     pub mempool_sync_timeout: u64,
 
+    /// a channel for submitting long-running tasks to the stacks-node
+    pub async_rpc_channel: Option<SyncSender<AsyncRequests>>,
```
jcnelson, line 376: Hmm, I don't think this is the right place for this.  I think this belongs instead in `ConversationHttp`, which is the struct concerned with managing the lifecycle of an HTTP connection.  `ConnectionOptions` is not meant to contain any information that gets initialized at runtime.

stackslib/src/net/http.rs
```diff
@@ -155,6 +158,7 @@ lazy_static! {
             *STANDARD_PRINCIPAL_REGEX_STRING, *CONTRACT_NAME_REGEX_STRING
         )).unwrap();
     static ref PATH_OPTIONS_WILDCARD: Regex = Regex::new("^/v2/.{0,4096}$").unwrap();
+    static ref PATH_POST_BLOCK_PROPOSAL: Regex = Regex::new(&format!("^{PATH_STR_POST_BLOCK_PROPOSAL}$")).unwrap();
```
jcnelson, line 161: Small nit: can you just get rid of `PATH_STR_POST_BLOCK_PROPOSAL` and put it here directly?  We've been bitten before by not having the complete regex in one place in the code.

stackslib/src/net/http.rs
```diff
@@ -725,13 +729,18 @@ impl HttpResponsePreamble {
         }
     }
 
-    pub fn ok_JSON_from_md<W: Write>(
+    fn ok_JSON_from_md<W: Write>(fd: &mut W, md: &HttpResponseMetadata) -> Result<(), codec_error> {
+        Self::ok_JSON_with_status(fd, md, 200)
+    }
+
+    fn ok_JSON_with_status<W: Write>(
```
jcnelson, line 736: Strictly speaking, this isn't an "OK" if `status` is not 200.  Can you refactor this method to be called `JSON_with_status` and add `accepted_JSON_with_status()` as the helper method for rendering a 202 response?

stackslib/src/net/http.rs
```diff
@@ -3978,48 +4043,50 @@ impl HttpResponseType {
     }
 
     pub fn metadata(&self) -> &HttpResponseMetadata {
```
jcnelson, line 4045: Can you undo the whole `*self` --> `self` transformation here and elsewhere?  It's gonna make the merge conflict that much worse when we merge this into the HTTP RPC refactor PR.  That goes for any beatification of this file -- it's going to disappear in `develop` so there's no use in trying to make it look better ;)

stackslib/src/net/http.rs
```diff
@@ -5047,6 +5143,321 @@ mod test {
     use crate::net::test::*;
     use crate::net::{RPCNeighbor, RPCNeighborsInfo};
 
+    /// Simulate reading variable-length segments
```
jcnelson, line 5146: Why are all these chunked HTTP transfer tests moved into this file?  Please put them back.

stackslib/src/net/rpc.rs
```diff
@@ -1223,6 +1225,42 @@ impl ConversationHttp {
         response.send(http, fd).map(|_| ())
     }
 
+    fn handle_validate_block_proposal(
+        metadata: HttpResponseMetadata,
+        block_proposal: BlockProposal,
+        options: &ConnectionOptions,
+    ) -> HttpResponseType {
+        // TODO: Reject if proposal is from invalid sender (allow localhost only?)
```
jcnelson, line 1233: Yes, you should do this before the PR merges.  It's `localhost` or bust.  If you're feeling ambitious, you can make it configurable in `ConnectionOptions`, but `localhost` will suffice.

stackslib/src/net/connection.rs
jcnelson: Seeing that you're setting up the `Receiver` in `neon_node.rs`, maybe the initial `Sender` should go into the `HttpPeer` instance within `PeerNetwork`.  Then, it can give any `ConversationHttp` instances a copy whenever a new client is accepted.

jcnelson (approved):
Overall this LGTM, but please address the feedback before merging.  Thanks @jbencin!

stackslib/src/net/connection.rs
jbencin: @kantai added this part, maybe he can explain

stackslib/src/net/rpc.rs
jbencin: Pretty sure I did this already, I just need to remove the comment

stackslib/src/net/http.rs
jbencin: I don't know what these are. There was a merge conflict here and maybe I didn't resolve it right. Where are these supposed to be?

stackslib/src/net/rpc.rs
jbencin: Already implemented, see https://github.com/jbencin/stacks-blockchain/blob/b1cd0afac08112bf6fdc79fc9fc50999710c1ee8/stackslib/src/net/rpc.rs#L3199

stackslib/src/net/http.rs
jbencin: I see these were moved to `chunked_encoding.rs` and I ended up duplicating them here

stackslib/src/chainstate/stacks/miner.rs
jbencin: Is there any reason this needs to work for non-Nakamoto blocks?

stackslib/src/net/connection.rs
kantai: Adding it as an argument to `HttpPeer::new` is fine with me -- I added it to `ConnectionOptions` because my understanding was that struct was meant to encapsulate all of the configuration arguments to network peer structs, etc to avoid cluttering those interfaces. I will update.

stackslib/src/chainstate/stacks/miner.rs
kantai: No -- it doesn't need to work for non-Nakamoto blocks. Updating this to be `NakamotoBlock` would require re-basing this off of `next`, though.

stackslib/src/net/connection.rs
jbencin: I moved it to `HttpPeer::new()` in latest commit. Let me know if that looks right to you.

jbencin:
@jcnelson I'm guessing there's no need to have this endpoint for non-Nakamoto blocks or to even have it on `develop`. If that's the case, I think it's best to abandon this and redo the work on a fresh branch off `next`, copying from here as needed.

saralab:
Jeffg has a new pR

PR #3880, opened 2023-08-29
Clarity wasm runtime

obycode:
This is a WIP and not intended to be merged.

### DIFF
--- a/clarity/src/vm/clarity_wasm.rs
+++ b/clarity/src/vm/clarity_wasm.rs
@@ -180,19 +180,21 @@ pub fn initialize_contract(
     let mut linker = Linker::new(&engine);
 
     // Link in the host interface functions.
-    link_define_function_fn(&mut linker);
-    link_define_variable_fn(&mut linker);
-    link_get_variable_fn(&mut linker);
-    link_set_variable_fn(&mut linker);
+    link_define_function_fn(&mut linker)?;
+    link_define_variable_fn(&mut linker)?;
+    link_get_variable_fn(&mut linker)?;
+    link_set_variable_fn(&mut linker)?;
     link_log(&mut linker);
 
-    let instance = linker.instantiate(store.as_context_mut(), &module).unwrap();
+    let instance = linker
+        .instantiate(store.as_context_mut(), &module)
+        .map_err(|e| Error::Wasm(WasmError::UnableToLoadModule(e)))?;
 
     // Call the `.top-level` function, which contains all top-level expressions
     // from the contract.
     let func = instance
         .get_func(store.as_context_mut(), ".top-level")
-        .expect(".top-level function was not found in the generated WASM binary.");
+        .ok_or(Error::Wasm(WasmError::TopLevelNotFound))?;
     let mut results = [];
 
     func.call(store.as_context_mut(), &[], &mut results)
@@ -218,13 +220,15 @@ pub fn call_function(
     let mut linker = Linker::new(&engine);
 
     // Link in the host interface functions.
-    link_define_function_fn_error(&mut linker);
-    link_define_variable_fn_error(&mut linker);
-    link_get_variable_fn(&mut linker);
-    link_set_variable_fn(&mut linker);
-    link_log(&mut linker);
+    link_define_function_fn_error(&mut linker)?;
+    link_define_variable_fn_error(&mut linker)?;
+    link_get_variable_fn(&mut linker)?;
+    link_set_variable_fn(&mut linker)?;
+    link_log(&mut linker)?;
 
-    let instance = linker.instantiate(store.as_context_mut(), &module).unwrap();
+    let instance = linker
+        .instantiate(store.as_context_mut(), &module)
+        .map_err(|e| Error::Wasm(WasmError::UnableToLoadModule(e)))?;
 
     // Call the specified function
     let func = instance
@@ -235,7 +239,10 @@ pub fn call_function(
     let stack_pointer = instance
         .get_global(store.as_context_mut(), "stack-pointer")
         .ok_or(Error::Wasm(WasmError::StackPointerNotFound))?;
-    let mut offset = stack_pointer.get(store.as_context_mut()).unwrap_i32();
+    let mut offset = stack_pointer
+        .get(store.as_context_mut())
+        .i32()
+        .ok_or(Error::Wasm(WasmError::ValueTypeMismatch))?;
 
     let memory = instance
         .get_memory(store.as_context_mut(), "memory")
@@ -287,7 +294,7 @@ pub fn call_function(
 
 /// Link host interface function, `define_function`, into the Wasm module.
 /// This function is called for all function definitions.
-fn link_define_function_fn(linker: &mut Linker<ClarityWasmInitContext>) {
+fn link_define_function_fn(linker: &mut Linker<ClarityWasmInitContext>) -> Result<(), Error> {
     linker
         .func_wrap(
             "clarity",
@@ -373,11 +380,17 @@ fn link_define_function_fn(linker: &mut Linker<ClarityWasmInitContext>) {
                 Ok(())
             },
         )
-        .unwrap();
+        .map(|_| ())
+        .map_err(|e| {
+            Error::Wasm(WasmError::UnableToLinkHostFunction(
+                "define_function".to_string(),
+                e,
+            ))
+        })
 }
 
 /// When in run-mode (not initialize-mode), this should never be called.
-fn link_define_function_fn_error<T>(linker: &mut Linker<T>)
+fn link_define_function_fn_error<T>(linker: &mut Linker<T>) -> Result<(), Error>
 where
     T: ClarityWasmContext,
 {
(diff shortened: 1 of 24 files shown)


clarity/src/vm/contexts.rs
```diff
@@ -1263,20 +1267,36 @@ impl<'a, 'b, 'hooks> Environment<'a, 'b, 'hooks> {
...
+        let mut contract_ast = ast::build_ast_with_rules(
             &contract_identifier,
             contract_content,
             self,
             clarity_version,
             self.global_context.epoch_id,
             ast_rules,
         )?;
+
+        let mut store = MemoryBackingStore::new();
```
cylewitruk, line 1283: Is the idea that this will go to a real store later on? Feels like a pretty heavy operation to just throw away in a memory store?

clarity/src/vm/contracts.rs
```diff
@@ -35,19 +38,28 @@ pub struct Contract {
...
-            &contract.expressions,
-            &mut contract_context,
-            global_context,
-            sponsor,
-        )?;
+        if let Some(wasm_module) = contract.wasm_module.take() {
+            contract_context.set_wasm_module(wasm_module);
+
+            // Execute the contract via the compiled Wasm module
+            initialize_contract(global_context, &mut contract_context, contract_analysis)?;
```
cylewitruk, line 53: Hm, in this context `initialize_contract` doesn't really feel like the right naming, since it's actually executing (according to the comment)?

clarity/src/vm/functions/define.rs
```diff
@@ -152,6 +152,7 @@ fn handle_define_function(
         define_type,
         function_name,
         &env.contract_context.contract_identifier.to_string(),
+        TypeSignature::NoType, // FIXME: this is a placeholder
```
cylewitruk, line 155: Is this fixable before merge, or will this be remembered? :)

src/chainstate/stacks/db/transactions.rs
```diff
@@ -1153,12 +1155,18 @@ impl StacksChainState {
                     .expect("BUG: total block cost decreased");
                 let sponsor = tx.sponsor_address().map(|a| a.to_account_principal());
 
+                // Compile the contract to Wasm
+                let mut module =
+                    compile_contract(&mut contract_analysis).expect("Failed to compile contract");
```
cylewitruk, line 1160: We're probably going to want an actual error here if the contract fails to compile to WASM instead of panicing

clarity/src/vm/clarity_wasm.rs
```diff
@@ -0,0 +1,1035 @@
...
+    link_set_variable_fn(&mut linker);
+    link_log(&mut linker);
+
+    let instance = linker.instantiate(store.as_context_mut(), &module).unwrap();
+
+    // Call the `.top-level` function, which contains all top-level expressions
+    // from the contract.
+    let func = instance
+        .get_func(store.as_context_mut(), ".top-level")
+        .expect(".top-level function was not found in the generated WASM binary.");
```
cylewitruk, line 193: Will likely want proper error handling instead of panic

cylewitruk (commented):
Biggest issues are error handling, but I guess we can write that off for now since we've still got implementations to do? Would be nice to have the correct error variants being returned anyway, I think, instead of panicing (so we don't forget any).

clarity/src/vm/contracts.rs
obycode: It's "initializing the contract" in that it's running the top-level code, that only gets run once, during deployment, so it's initialization code. I can improve the comment to make that more clear.

clarity/src/vm/contexts.rs
obycode: This should actually be using the `contract_analysis` that is already done. I forgot I put this here to test it out before changing the various function signatures to pass that one in.

obycode: Oh, no... I didn't forget about this, I just forgot about why! This code is not reached during normal execution. This path is only hit from tests and from the clarity CLI, so it's fine. I'll add a comment.

obycode: Added in f945ea56d.

clarity/src/vm/contracts.rs
obycode: Added comments in ce8488c55

clarity/src/vm/functions/define.rs
obycode: This new return type field is only used in the new code paths, for Wasm. So the question is... should we just leave it as a `NoType` in this interpreter path, or should we make it an `Option` and set it to `None`, or should we go through the trouble of setting it in the interpreter path? Any thoughts?

src/chainstate/stacks/db/transactions.rs
obycode: Addressed in 291c4a555

clarity/src/vm/clarity_wasm.rs
obycode: 4565abf70 replaces all panics with errors.

obycode: Please re-review to make sure I didn't miss any.

clarity/src/vm/functions/define.rs
cylewitruk: I'm personally not a huge fan of having different meanings for fields dependent on code path... so if it were me I would probably set it in all paths if it would make sense... but if that's a bunch of work, the next best option imo would be to use the `Option+None` variant and a comment on the field indicating when it can be `Some`.

clarity/src/vm/clarity_wasm.rs
cylewitruk, line 1: I updated all of the `match` blocks to be exhaustive with `todo!()`'s, so we don't accidentally forget something :)

src/chainstate/stacks/db/transactions.rs
```diff
@@ -1090,7 +1092,7 @@ impl StacksChainState {
                     &contract_code_str,
                     ast_rules,
                 );
-                let (contract_ast, contract_analysis) = match analysis_resp {
+                let (mut contract_ast, mut contract_analysis) = match analysis_resp {
                     Ok(x) => x,
                     Err(e) => {
                         match e {
```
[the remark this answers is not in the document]
obycode: Those are existing `expect`s in the code, not new ones from us. Those are the types of things that we discussed before, where if something should never happen, should it return an error or panic. In these cases, they chose panic. I don't think we should change them in this PR, regardless of our thoughts on that.

clarity/src/vm/functions/define.rs
obycode: I think setting it here is a bunch of work that doesn't give us any benefits. I'll update this to use an `Option<TypeSignature>`.

obycode: Updated in 164b9299e145d5bceafcbc993095b3552069c661

clarity/src/vm/clarity_wasm.rs
```diff
@@ -0,0 +1,1038 @@
...
+fn link_define_function_fn_error<T>(linker: &mut Linker<T>) -> Result<(), Error>
+where
+    T: ClarityWasmContext,
+{
+    linker
+        .func_wrap(
+            "clarity",
+            "define_function",
+            |mut _caller: Caller<'_, T>, _kind: i32, _name_offset: i32, _name_length: i32| {
+                // FIXME: I don't understand why I have to write this like this
```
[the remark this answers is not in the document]
Acaccia: Yes I do, it's a bit tricky.
The closure here should implement `IntoFunc`. For the arguments, no issue, but for the return type, here you are returning a `Result`, which does not implement `WasmRet`.
What does implement `WasmRet` however is `WasmTime::Result`, which they say is a wrapper for `Result<T, anyhow::Error`. 
So if I add `anyhow` as dep and rewrite the closure like this, it works:
```rust
|mut _caller: Caller<'_, T>, _kind: i32, _name_offset: i32, _name_length: i32| {
    Err::<(), _>(anyhow!(Error::Wasm(WasmError::DefineFunctionCalledInRunMode)))
}
```

Why it worked with a `?` is because this calls an implicit "into" on the error, converting it automatically to a `anyhow::Error`.

clarity/src/vm/clarity_wasm.rs
```diff
@@ -0,0 +1,1038 @@
...
+                // instead of just:
+                //   Err(Error::Wasm(WasmError::DefineFunctionCalledInRunMode))
+                let _ = Err(Error::Wasm(WasmError::DefineFunctionCalledInRunMode))?;
+                Ok(())
+            },
```
Acaccia, line 422:
```suggestion
            |mut _caller: Caller<'_, T>,
             _kind: i32,
             _name_offset: i32,
             _name_length: i32|
             -> Result<(), _> {
                // This should be a Anyhow error, but we don't have it as a dependency.
                // `?` does the trick.
                let _ = Err(Error::Wasm(WasmError::DefineFunctionCalledInRunMode))?;
                unreachable!();
            },
```

obycode:
Replaced by #4093

PR #3885, opened 2023-08-30
Chore/stacks signer cli cleanup

jferrant:
Changes:

Added clap to cleanup the arguments
Added config file option to the CLI commands (making sure to make host, contract, and private key conditional on it NOT being specified)
Cleaned up error formatting using thiserror
Added stacks private key to config file
Added README
 
 Part of https://github.com/stacks-network/stacks-blockchain/issues/3882
 Related to https://github.com/stacks-network/stacks-blockchain/pull/3883

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -102,6 +102,54 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f58811cfac344940f1a400b6e6231ce35171f614f26439e80f8c1465c5cc0c"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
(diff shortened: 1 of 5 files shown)


jcnelson (approved):
LGTM!  Thanks for cleaning this up!

Btw, for anyone else reading this, this PR adds a dependency on `clap` and `thiserror` to `libsigner`.  While `thiserror` does not have any `unsafe` code, `clap_lex` does.

I'm more okay with such dependencies in non-blockchain code, because while the consequence of a signer crashing is bad, it's not as bad as the whole network crashing.  Also, `unsafe` code in `clap` can be worked around at runtime -- we `fork(2)` the daemon, have the child drop privileges and close everything but a "results" pipe created prior to the fork, use `clap` to parse the CLI args in the child process, and feed back the CLI args through the "results" pipe before exiting.  Not saying that we have to do this right now; I'm just noting it here as a possible future remediation in the event that `clap` has a memory safety CVE.

netrome (approved):

kantai (approved):
LGTM!

PR #3487, opened 2023-01-12
Rust version of `data relay`.

sergey-shandar:
### Applicable issues
- fixes #3477 

### Checklist
- [ ] Test coverage for new or modified code paths
- [ ] Changelog is updated
- [ ] Required documentation changes (e.g., `docs/rpc/openapi.yaml` and `rpc-endpoints.md` for v2 endpoints, `event-dispatcher.md` for new events)
- [ ] New clarity functions have corresponding PR in `clarity-benchmarking` repo
- [ ] New integration test(s) added to `bitcoin-tests.yml`

PR #3888, opened 2023-09-01
ci: Fix CI failing when we don't have DockerHub credentials

jbencin:
### Description

Currently, CI always fails when I commit to my personal fork of `stacks-blockchain`. The same behavior would also occur in any other fork of this repo. This is because in forks, the secrets `DOCKERHUB_USERNAME` and `DOCKERHUB_PASSWORD` are not accessible.

This PR fixes the failure by checking whether these credentials are empty. If they are, it will still attempt to build the Docker images, but will not attempt to login and push them to DockerHub.

### Additional Notes

**The changes to `.github/workflows/ci.yml` should not be merged.** Those only to enable running the tests without pushing to `master`

As an alternative, I could have modified `.github/workflows/ci.yml` to skip the Docker image builds all together. I don't know if building the image itself is a useful test, maybe not?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,7 +128,7 @@ jobs:
   docker-alpine:
     if: ${{ inputs.tag != '' || (github.ref != format('refs/heads/{0}', github.event.repository.default_branch) && !contains(github.ref, 'refs/pull')) }}
     name: Docker Alpine (Binary)
-    uses: stacks-network/stacks-blockchain/.github/workflows/image-build-alpine-binary.yml@master
+    uses: jbencin/stacks-blockchain/.github/workflows/image-build-alpine-binary.yml@ci/pass-without-dockerhub-credentials
     needs: github-release
     with:
       tag: ${{ inputs.tag }}
@@ -147,7 +147,7 @@ jobs:
   docker-debian:
     if: ${{ inputs.tag != '' || (github.ref != format('refs/heads/{0}', github.event.repository.default_branch) && !contains(github.ref, 'refs/pull')) }}
     name: Docker Debian (Binary)
-    uses: stacks-network/stacks-blockchain/.github/workflows/image-build-debian-binary.yml@master
+    uses: jbencin/stacks-blockchain/.github/workflows/image-build-debian-binary.yml@ci/pass-without-dockerhub-credentials
     needs: github-release
     with:
       tag: ${{ inputs.tag }}
@@ -167,7 +167,7 @@ jobs:
   build-branch:
     if: ${{ inputs.tag == '' }}
     name: Docker Debian (Source)
-    uses: stacks-network/stacks-blockchain/.github/workflows/image-build-debian-source.yml@master
+    uses: jbencin/stacks-blockchain/.github/workflows/image-build-debian-source.yml@ci/pass-without-dockerhub-credentials
     needs:
       - rustfmt
       - leaked-cred-test
--- a/.github/workflows/image-build-alpine-binary.yml
+++ b/.github/workflows/image-build-alpine-binary.yml
@@ -39,6 +39,7 @@ jobs:
         run: |
           echo "GITHUB_SHA_SHORT=${GITHUB_SHA::7}" >> $GITHUB_ENV
           echo "GITHUB_REF_SHORT=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+          echo "DOCKER_PUSH=${{ (secrets.DOCKERHUB_USERNAME != '') && (secrets.DOCKERHUB_PASSWORD != '') }}" >> $GITHUB_ENV
       - name: Set up QEMU
         id: docker_qemu
         uses: docker/setup-qemu-action@v2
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -128,7 +128,7 @@ jobs:
   docker-alpine:
     if: ${{ inputs.tag != '' || (github.ref != format('refs/heads/{0}', github.event.repository.default_branch) && !contains(github.ref, 'refs/pull')) }}
     name: Docker Alpine (Binary)
-    uses: stacks-network/stacks-blockchain/.github/workflows/image-build-alpine-binary.yml@master
+    uses: jbencin/stacks-blockchain/.github/workflows/image-build-alpine-binary.yml@ci/pass-without-dockerhub-credentials
```
kantai, line 131: Should this just be:

```suggestion
    uses: ./.github/workflows/image-build-alpine-binary.yml
```

That way, it always uses the workflow from the current branch? @wileyj

jbencin: I agree with this and will modify the PR to include it. I didn't understand the reason for `@master` (it complicates CI testing IMO), but I didn't want to change too much and complicate the PR.

wileyj: hmm, this is actually something i've run into in the past. 
for testing, i definitely use the filepath `./.github/...` but there was a decision made long ago to only use the action in the master branch :
```
    uses: stacks-network/stacks-blockchain/.github/workflows/image-build-alpine-binary.yml@master
```
that way, the action that runs will always be a PR approved workflow vs some workflow in a branch. 

unless there's a reason to keep the filepath that i'm not seeing, i'd prefer to keep the standard we've been using for a while of requiring actions to run off of the master branch.

.github/workflows/image-build-alpine-binary.yml
```diff
@@ -39,6 +39,7 @@ jobs:
         run: |
           echo "GITHUB_SHA_SHORT=${GITHUB_SHA::7}" >> $GITHUB_ENV
           echo "GITHUB_REF_SHORT=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+          echo "DOCKER_PUSH=${{ (secrets.DOCKERHUB_USERNAME != '') && (secrets.DOCKERHUB_PASSWORD != '') }}" >> $GITHUB_ENV
```
wileyj, line 42: nice, i like what you're doing here. simple and it'll work

wileyj:
Overall, i like the proposed changes here - simple but effective and it'll resolve some ongoing questions about why a fork has problems building (i think we can take this further as well, like for the circleci steps, but that can be another PR). 

Only change I'd like to see is moving the `uses:` directive back from a filepath to the repository master branch, which has been our standard for a while now. if we want to change that, i don't think this repo should be where it happens (since there would be several other places we could also change).

.github/workflows/ci.yml
jbencin: > that way, the action that runs will always be a PR approved workflow vs some workflow in a branch.

That seems pretty easy to get around though, since anyone that wants to run their own unapproved workflow can do the `./.github...` trick also, right?

I think a better solution for preventing unapproved workflows being run (which does matter here because CI has DockerHub secrets) is to restrict write access to the repo. Instead, contributors could fork it to their own namespace and issue PRs from there (which is what I do and why I'm making this change)

jbencin:
> Only change I'd like to see is moving the `uses:` directive back from a filepath to the repository master branch, which has been our standard for a while now

My intent was to discard the changes to `uses:` before merging. I only pushed these changes because they were necessary to run the files  I modified and demonstrate that CI passes

I think it's worth having further discussion on the relative path issue, but it should be done in a separate issue/PR

.github/workflows/ci.yml
wileyj: yea, i see  your point and i don't entirely disagree with you. in my fork, i also have to make similar changes to test things, so i'm acutely aware of the problem you're trying to solve. 
i just think this PR is the wrong place to push for that change. 

i think it would be better suited to have another PR where we can answer that question of what standard(s) do we want to use for localized workflows. 
I know @kantai has opinions on workflow versioning (in context, he had raised concerns about using un-versioned external workflows so it's not entirely the same with a localized workflow), and if we're going to change which workflows we allow to run, i'd like to have more input from others. 

incidentally, i do have an open PR to modify a lot of the workflows. perhaps that's a good place to have this discussion about resuable workflows? https://github.com/stacks-network/stacks-blockchain/pull/3847

wileyj: @jbencin would you also be able to port your changes to the branch i'm trying to PR in #3487 ? if not, i can copypasta

jbencin: @wileyj Sure I'll take a look at that today. You're referring to the changes with `DOCKER_PUSH` I assume, and not `./.github...`?

jbencin: I made both changes in separate commits. See PR #3895

jbencin:
@wileyj  or @kantai: Can someone with admin privileges please cherry-pick [79d3760](https://github.com/stacks-network/stacks-blockchain/pull/3888/commits/79d376057d4cea082ba2d24f20df7b84072ad079) onto `develop`? I have no way to merge this because CI will fail when I revert the [9a34933](https://github.com/stacks-network/stacks-blockchain/pull/3888/commits/9a349337e67457f568288a14b0fef773a3ef5cfd)

As it is now I can't merge my other PRs because of the CI failures from not having DockerHub credentials. Once this is merged into `develop`, I can merge/rebase it into other PRs to fix CI there

wileyj:
> 79d3760

https://github.com/stacks-network/stacks-blockchain/pull/3905

jbencin:
#3905 has been merged, so this can be closed

PR #3891, opened 2023-09-02
Epoch 3.0

cylewitruk:
### Description
Figured I'd take a stab at adding the 3.0 epoch for Nakamoto since over in Clarity-WASM we're starting discussing epoch gating, and it would be beneficial if the enum variant and related were already now available in `next` for all WG's who need it.

The code compiles and, unless I missed any, all failing tests are due to `todo!()` statements which I added in match statements where it felt like the logic will be handled by @jcnelson + @kantai in their WGs?

Also added boilerplate for costs v4 (and added a placeholder `costs-4.clar` for tests) as I'm assuming there will be a new costs version given the scope of changes?

### DIFF
--- a/clarity/src/vm/analysis/mod.rs
+++ b/clarity/src/vm/analysis/mod.rs
@@ -133,7 +133,7 @@ pub fn run_analysis(
             StacksEpochId::Epoch20 | StacksEpochId::Epoch2_05 => {
                 TypeChecker2_05::run_pass(&epoch, &mut contract_analysis, db)
             }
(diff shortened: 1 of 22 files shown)


will-at-stacks:
yooooo!

cylewitruk:
Noticed that `next` is way behind `master` after creating this, though...

obycode:
`master` has been merged into `develop` has been merged into `next` now, if you'd like to try to re-apply these changes on top of the latest.

cylewitruk:
Not even going to try to rebase on that haha.. will open a new PR :)

PR #3898, opened 2023-09-07
Chore/port dkg from test

jferrant:
Want eyes on this though since its been sitting in pretty much the same state the last week with me just trying to fix the integration test.

`cargo test --package stacks-node --bin stacks-node -- tests::signer --test-threads 1 --nocapture`

TODO:
- Update frost coordinator/signer to not have the copy pasted code.
- Fix integration test


testnet/stacks-node/src/tests/signer.rs
```diff
@@ -0,0 +1,249 @@
...
+            r#"
+            {{public_key = "{ecdsa_public_key}", key_ids = [{ids}]}}
+        "#
+        );
+        if i != signer_ecdsa_private_keys.len() - 1 {
+            signers_array += ",";
+        }
+    }
+    signers_array += "]";
+    let mut port = 30000;
```
jcnelson, line 98: I think this needs to be the port from `config.node.rpc_bind`, since `node_host` gets passed to `libstackerdb` the RPC endpoint for the node.

jferrant: ah so this means I need to spin up num_signers neon nodes. They cannot just use the same one as we will have port conflicts if I try to bind multiple times to the same port.

xoloki: If you look at the current code, I'm using the rpc_bind from stacks node config as node_host, and then have a separate endpoint config key that uses the 3000x ports for the stackerdb receivers

jferrant:
Replaced by https://github.com/stacks-network/stacks-blockchain/pull/3911

PR #3903, opened 2023-09-08
Update rusqlite to 0.29.0 [NEXT]

tbraun96:
In order for this repo to be usable by other up to date repos, we must update rusqlite

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,7 +95,7 @@ version = "0.24.2"
 features = ["serde", "recovery"]
 
 [dependencies.rusqlite]
-version = "=0.24.2"
+version = "0.29.0"
 features = ["blob", "serde_json", "i128_blob", "bundled", "trace"]
 
 [dependencies.ed25519-dalek]


jcnelson:
Closing as spam

diwakergupta:
> Closing as spam

Why? AFAICT this is a legit contributor (e.g. see https://github.com/Trust-Machines/stacks-sbtc/pull/697). Closing a PR as spam without any explanations feels incredibly hostile to new contributors.

tbraun96:
Agreed, this was very rude.

jcnelson:
Hi, every so often we get one-line PRs like this with minimal explanation.  In the past, they had originated from code bootcamps -- opening a PR is a homework assignment.  That's one reason why it looked spammy.

Perhaps @tbraun96 can explain the context and need for this PR?  The current version of rusqlite serves the project adequately.  In general, simply bumping the version of a package is not enough for an approval.  At the very least, the `Cargo.lock` file would need to be updated as well.  The fact that the new `Cargo.lock` isn't part of this PR suggests that there was no attempt on the submitter's part to check that the project even builds with this change, let alone that tests pass (which contributed to my initial assessment that this is PR spam).

If this is part of adding `async/await` to TM's sBTC tooling, then perhaps the bigger question that needs answering is why the sBTC tooling is dependent on the version of `rusqlite` in this repository?  Ideally there would be no dependency at all.

PR #3905, opened 2023-09-08
cherry-picking docker login env var commit

wileyj:
ref: https://github.com/stacks-network/stacks-blockchain/pull/3888

This is the develop branch with https://github.com/stacks-network/stacks-blockchain/pull/3888/commits/79d376057d4cea082ba2d24f20df7b84072ad079 cherry picked.

### DIFF
--- a/.github/workflows/image-build-alpine-binary.yml
+++ b/.github/workflows/image-build-alpine-binary.yml
@@ -39,6 +39,7 @@ jobs:
         run: |
           echo "GITHUB_SHA_SHORT=${GITHUB_SHA::7}" >> $GITHUB_ENV
           echo "GITHUB_REF_SHORT=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


jbencin (approved):
Thanks! LGTM

kantai (approved):

jbencin:
Well, I guess this didn't actually fix my specific issue. It was merged to `develop`, but the CI references `master`. So to actually get CI working on forks, we could...
- Merge this single commit into `master`
- Implement the `./.github/...` change

jbencin:
We couldn't merge this branch to `master` as-is because that would introduce a lot of unwanted changes from `develop`. We'd have to cherry-pick this commit again onto the tip of `master` and make a new PR.

I will do that now

Also, I don't see any need to undo the change on `develop`. I don't think it matters one way or the other

wileyj:
> We couldn't merge this branch to `master` as-is because that would introduce a lot of unwanted changes from `develop`. We'd have to cherry-pick this commit again onto the tip of `master` and make a new PR.
> 
> I will do that now
> 
> Also, I don't see any need to undo the change on `develop`. I don't think it matters one way or the other

true. only downside is it may create a conflict when develop is merged to master next

PR #3907, opened 2023-09-08
chore: Run `cargo fmt` with `imports_granularity="Module"`

jbencin:
### Description

Run `cargo fmt` with `imports_granularity="Module"` in order to merge `use` statements. For example:

```rust
use a::b::c
use a::b::d
use a::b::e
```

would be merged to

```rust
use a::b::{c, d, e}
```

My reasons for doing this:
 - It makes the `use` statements look nicer and easier to read
 - It reduces the size of the code, removing a total of nearly 2000 lines from the codebase
 - It gives `use` statements a canonical form, reducing the chance of merge conflicts in the future

I would have liked to make a `rustfmt.toml` file with the following config:

```toml
group_imports="StdExternalCrate"
imports_granularity="Module"
```

So that we could guarantee these settings would be used any time `cargo fmt` or `rustfmt` was executed. Unfortunately, `rustfmt` only accepts these arguments on the command line for now, so I added the command with CLI args in a couple places:
 - `CONTRIBUTING.md`: Under the "Formatting" section and in the pre-commit hook
 - `.github/actions/bitcoin-int-tests/Dockerfile.rustfmt`: Added this config to CI check
 -  I also created a `Justfile` (which is kind of like a `Makefile`) with the command, so it can be invoked by `just fmt`

These are the only changes that I made myself and that need to be reviewed. Everything else was done by `cargo fmt`


Justfile
```diff
@@ -0,0 +1,32 @@
+# Print help message
```
jcnelson, line 1: What's this?

jcnelson (commented):
Before approving:

* What's a `JustFile`?

* Can you also update the CI process's `rust fmt` to require that `imports_granularity="Module"` is applied as a pre-condition to merging the PR?

Justfile
jbencin: The output of `just help` is auto-generated from comments in the file

jbencin:
> What's a `JustFile`?

I wanted to add a way to call `cargo fmt` without requiring the user to type out the full command (remember, these particular config options **don't work** in `.rustfmt`!)

A `Justfile` is kind of like a `Makefile`, but specifically designed for running commands, and without some of the drawbacks of `make`: It runs on Windows, no `.PHONY` targets, etc.

I figured this was a better choice than a `Makefile` or shell script since it's cross-platform and trivial for a Rust dev to install. Just run `cargo install just`

It's important to note that this is **not** adding a dependency, and using it is entirely optional. It's just a short-cut for certain "verbose" commands. If the user has `just` installed, they just have to run:

```sh
just fmt
```

instead of:

```sh
cargo fmt --all -- --config group_imports=StdExternalCrate,imports_granularity=Module
```

> Can you also update the CI process's `rust fmt` to require that `imports_granularity="Module"` is applied as a pre-condition to merging the PR?

Will do!

jbencin:
Applied the new config to CI and refactored this PR into two commits:
 - [89f8d66](https://github.com/stacks-network/stacks-blockchain/pull/3907/commits/89f8d666219a183c11cbefb8869987707a6c4f5f): Changes made manually
 - [c5d2356](https://github.com/stacks-network/stacks-blockchain/pull/3907/commits/c5d2356ae57bb7e17eaab671d06dc74c5a396227): Changes made by `cargo fmt`

This should make the PR easier to review, and to cherry-pick/rebase if necessary

jcnelson:
> A Justfile is kind of like a Makefile, but specifically designed for running commands, and without some of the drawbacks of make: It runs on Windows, no .PHONY targets, etc.

I think there was a conversation about this earlier, related to CI.  The TL;DR was that we would use Makefiles instead of Justfiles.  The rationale was that GNU Make has been around for longer than anyone of us have been alive (so it's not going anywhere, whereas Justfiles are relatively new), it's a well-documented and well-understood program, and it runs everywhere (including Windows, modulo some slightly different quoting rules that I don't think apply here).  `.PHONY` is a one-line decorator for a target, so it's not a concern.  Is there a way to use a Makefile here instead?

jbencin:
> Is there a way to use a Makefile here instead?

I could, and for the simple use case of calling `cargo fmt` with some hard-coded parameters they are equivalent

However, since `just` is designed to be a task runner (rather than a build system for C) it does have some advantages here. One thing you can do in `just` that you can't in `make` is to forward CLI arguments. For example, I could modify `just fmt` to take additional args:

```just
fmt *args:
    cargo fmt -- --config group_imports=StdExternalCrate,imports_granularity=Module {{args}}
```

Then if the user wanted to do something slightly different, like they just wanted to check the formatting without making changes, they could run:

```sh
just fmt --check
```

If you wanted to do that in `make` you'd be stuck doing:

```sh
make fmt ARGS="--check"
```

Which would be totally fine when used from scripts or a CI system, but is not something I'd want to be typing out constantly

I agree with you that `make` is the right choice for CI due to how established it is, but a user-facing CLI utility has different requirements, so it's worth considering alternatives

obycode (approved):
This looks good to me!

jcnelson:
@jbencin I'd prefer a Makefile because `make` is already installed on basically every system on earth, whereas `just` is still pretty new.  If users find themselves typing `make fmt ARGS="..."` too annoying, they can always alias it or put it into a shell script, just as they would with anything else that needs a localized shorthand.

jbencin:
> I'd prefer a Makefile because make is already installed

No problem, I can make this change

jcnelson (approved):
Thanks @jbencin!  Once this merges, can you also merge it to `next`?

jbencin:
> can you also merge it to `next`?

Sure

I suppose I should also post on Discord to notify people of this change. I don't want people to be confused about why all of a sudden CI is failing on all their PRs

jcnelson:
Oh, sorry -- meant to say, please open a PR against `next` :pray:

jbencin:
> please open a PR against `next`

@jcnelson Sorry forgot about this. By this did you mean merge `develop` into `next`, or just create a PR with these commits `cherry-pick`ed onto `next`?

jcnelson:
It can happen as part of another PR.  It doesn't have to be its own PR.  Thanks!

Issue #11, opened 2014-12-05
move from text file DB to more manageable data store


shea256:

jcnelson:
My stress tests indicate that we're probably going to be fine until we reach ~1 million names.  Then, we need to seriously look at a database.  Sqlite would probably work until we reach into the billions.

shea256:
OK how about we plan on switching to sqlite next (whenever necessary)?

muneeb-ali:
I think we can add SQLite support for the next version and flag this issue with the next release, which reminds me that what release numbers are we using? The last released version was blockstore-0.0.3 and @jcnelson and I have also been calling this release "1.0". It might make sense to call it v0.1.0 since this is the first production release and we will not be changing the genesis block after this release.

jcnelson:
This is present in the `develop` branch, but will need to be sync'ed with the new stable release.

jcnelson:
This will be available in 0.14.

shea256:
Ha awesome this issue is from Dec 5, 2014. Excited for SQL :D

jcnelson:
Been working on it off and on since I joined :)

muneeb-ali:
@jcnelson next release meaning v0.14 correct? As in this is already done?

jcnelson:
Yup! Closing this issue.

PR #3911, opened 2023-09-13
Implement full DkgSign over StackerDB

xoloki:
### Description

This PR builds on the `test_stackerdb_dkg` integration test, but now it uses  `libstackerdb` and `libsigner`.  It implements a full state-machine based `stacks-signer`, with all of the crypto and networking necessary to run a full `DkgSign` command.

### Applicable issues
- Closes https://github.com/stacks-network/stacks-blockchain/issues/3881

### Checklist
- [ ] Changelog is updated


libsigner/src/runloop.rs
```diff
@@ -75,7 +78,17 @@ pub trait SignerRunLoop<R> {
...
+            //info!("SignerRunLoop::main_loop: command_recv");
+            let next_command_opt = match command_recv.recv_timeout(poll_timeout) {
+                Ok(cmd) => Some(cmd),
+                Err(RecvTimeoutError::Timeout) => None,
+                Err(RecvTimeoutError::Disconnected) => {
+                    info!("Command receiver disconnected");
+                    return None;
+                }
+            };
+            //info!("SignerRunLoop::main_loop: run_one_pass");
```
jcnelson, line 90: Before merging, can you remove the commented-out debug statements?

stacks-signer/src/config.rs
```diff
@@ -38,25 +43,68 @@ pub enum ConfigError {
...
+    /// endpoint to stackerdb receiver
+    pub endpoint: String,
     /// contract identifier
     pub stackerdb_contract_id: String,
-    /// the private key used to sign blocks, chunks, and transactions in hexademical format
-    pub private_key: String,
+    /// the 32 byte ECDSA private key used to sign blocks, chunks, and transactions
+    pub message_private_key: String,
+    /// The hex representation of the signer's Stacks private key
+    pub stacks_private_key: String,
```
jcnelson, line 99: I'm having a hard time understanding the difference between `message_private_key` and `stacks_private_key`.  Can you help me understand why there need to be two private keys?

stacks-signer/src/crypto/frost.rs
```diff
@@ -0,0 +1,780 @@
...
+        };
+        self.ids_to_await = (0..self.total_signers).collect();
+        self.move_to(State::NonceGather)?;
+        Ok(nonce_request_msg)
+    }
+
+    fn gather_nonces(&mut self, message: &Message) -> Result<(), Error> {
+        match &message.msg {
+            MessageTypes::NonceResponse(nonce_response) => {
+                // TODO: check sign_id and sign_nonce_id
```
jcnelson: Does this need to be addressed in order for the code to be robust against a malicious signer?

stacks-signer/src/crypto/frost.rs
```diff
@@ -0,0 +1,780 @@
...
+            let sig = aggregator.sign(&self.message, &nonces, shares)?;
+
+            info!("Signature ({}, {})", sig.R, sig.z);
+
+            let proof = SchnorrProof::new(&sig)?;
+
+            info!("SchnorrProof ({}, {})", proof.r, proof.s);
+
+            if !proof.verify(&self.aggregate_public_key.x(), &self.message) {
+                warn!("SchnorrProof failed to verify!");
```
jcnelson, line 429: From the demo today, it seemed that this happened?  Can this happen as part of the signer's normal operation?  If so, what would the signer need to do here?

stacks-signer/src/crypto/frost.rs
```diff
@@ -0,0 +1,780 @@
...
+pub trait StateMachine {
+    /// Attempt to move the state machine to a new state
+    fn move_to(&mut self, state: State) -> Result<(), Error>;
+    /// Check if the state machine can move to a new state
+    fn can_move_to(&self, state: &State) -> Result<(), Error>;
+}
+
+impl StateMachine for Coordinator {
+    fn move_to(&mut self, state: State) -> Result<(), Error> {
+        self.can_move_to(&state)?;
```
jcnelson, line 472: I love how this system checks for a valid state transition!  Do you think it makes sense to outright `panic!()` if we encounter an invalid state transaction? Because, that would indicate a logic bug in the signer, right?

stacks-signer/src/crypto/frost.rs
```diff
@@ -0,0 +1,780 @@
...
+    pub fn new(total_signers: u32, total_keys: u32, threshold: u32, message_private_key: Scalar) -> Self {
+        Self {
+            current_dkg_id: 0,
+            current_dkg_public_id: 0,
+            current_sign_id: 0,
+            current_sign_nonce_id: 0,
+            total_signers,
+            total_keys,
+            threshold,
+            dkg_public_shares: Default::default(),
```
jcnelson, line 87: It seems these fields (`dkg_public_shares`, `public_nonces`, `signature_shares`, `aggregate_public_key`, `signature`, and `schnorr_proof`) are generated as part of either DKG or a signing round.  If so, then would it make sense to treat them as `Option<..>`, so it's unambiguous as to whether or not they have been calculated as part of this DKG or signing round?

stacks-signer/src/crypto/frost.rs
```diff
@@ -0,0 +1,780 @@
...
+        let sig_share_request = SignatureShareRequest {
+            dkg_id: self.current_dkg_id,
+            sign_id: self.current_sign_id,
+            correlation_id: 0, // TODO: what is this?
+            nonce_responses,
+            message: self.message.clone(),
+        };
+        let sig_share_request_msg = Message {
+            sig: sig_share_request.sign(&self.message_private_key).expect(""),
+            msg: MessageTypes::SignShareRequest(sig_share_request), // TODO: either SigShare or SignatureShare, SignShare is right out
```
jcnelson, line 367: Will this TODO be addressed before merge?

stacks-signer/src/runloop.rs
```diff
@@ -0,0 +1,439 @@
...
+    /// Wait for an initial command
+    Wait,
+    /// Continaully poll for commands to execute
+    Run,
+    /// Generate a DKG aggregate public key
+    Dkg,
+    /// Generate a DKG aggregate public key and then sign the message
+    DkgSign {
+        /// The bytes to sign
+        message: Vec<u8>,
```
jcnelson, line 31: Presumably, the `message` will be decoded and validated by the business logic before signing, right?  For example, the signer might want to check that the data to sign is a partially-signed Bitcoin transaction, or a well-formed Stacks block, but nothing else? If so, then would it make sense to treat this field as an `enum` of the possible types of data that could be signed, so only valid inputs can be represented?

stacks-signer/src/runloop.rs
```diff
@@ -0,0 +1,439 @@
...
+        (outbound_messages, results)
+    }
+}
+
+impl RunLoop<FrostCoordinator> {
+    /// Creates new runloop from a config and command
+    pub fn new(config: &Config, command: RunLoopCommand) -> Self {
+        // TODO: should this be a config option?
+        // The threshold is 70% of the shares
+        let threshold = (config.signer_ids_public_keys.key_ids.len() / 10 * 7)
```
jcnelson, line 177: Should this be `(config.signer_ids_public_keys.key_ids.len() * 7) / 10`?

stacks-signer/src/runloop.rs
```diff
@@ -0,0 +1,439 @@
...
+            .key_ids
+            .len()
+            .try_into()
+            .unwrap();
+        let key_ids = config
+            .signer_key_ids
+            .get(&config.signer_id)
+            .unwrap()
+            .iter()
+            .map(|i| i - 1) // SigningRound::new (unlike SigningRound::from) doesn't do this
```
jcnelson, line 197: This `- 1` can never underflow, right?

stacks-signer/src/runloop.rs
```diff
@@ -0,0 +1,439 @@
...
+}
+
+/// Process inbound messages using the frost_signer signing round mechanism
+pub fn process_inbound_messages(
+    signing_round: &mut SigningRound,
+    messages: Vec<Message>,
+) -> Result<Vec<Message>, frost_signer::signing_round::Error> {
+    let mut responses = vec![];
+    for message in messages {
+        // TODO: this code was swiped from frost-signer. Expose it there so we don't have duplicate code
```
jcnelson, line 230: Flagging this TODO for completion before merge.

stacks-signer/src/runloop.rs
```diff
@@ -0,0 +1,439 @@
...
+
+/// Helper function for determining the coordinator public key given the the public keys
+fn calculate_coordinator(public_keys: &PublicKeys) -> (u32, &ecdsa::PublicKey) {
+    // TODO do some sort of VRF here to calculate the public key
+    // Mockamato just uses the first signer_id as the coordinator for now
+    (0, public_keys.signers.get(&0).unwrap())
+}
+
+/// TODO: this should not be here.
+/// Temporary copy paste from frost-signer
```
jcnelson, line 346: Flagging this for completion

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -90,13 +92,6 @@ pub struct MinedMicroblockEvent {
     pub anchor_block: BlockHeaderHash,
 }
 
-/// Event structure for newly-arrived StackerDB data
-#[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct StackerDBChunksEvent {
```
jcnelson, line 95: Ha, thanks for DRY'ing this up!

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -0,0 +1,315 @@
...
+// Helper struct for holding the btc and stx neon nodes
+#[allow(dead_code)]
+struct RunningNodes {
+    pub btc_regtest_controller: BitcoinRegtestController,
+    pub btcd_controller: BitcoinCoreController,
+    pub join_handle: thread::JoinHandle<()>,
+    pub conf: NeonConfig,
+}
+
+fn build_contract(num_signers: u32, signer_stacks_private_keys: &[StacksPrivateKey]) -> String {
```
jcnelson, line 37: Could `build_contract` and `build_signer_config_tomls` be dropped here, in favor of the ones in the `stacks-signer` crate?

jcnelson: Ah, I see that already happened in @jferrant's follow-on PR.

libsigner/src/runloop.rs
xoloki: Done.

stacks-signer/src/crypto/frost.rs
xoloki: It needs to be addressed for general robustness, and specifically we will need it for the meta-protocols FIRE/ROAST (where we handle byzantine signers).  It's fine as is for now, but if you want I can do an initial robustness pass before merging this PR.

stacks-signer/src/crypto/frost.rs
xoloki: We didn't see the actual warning, but we did see multiple DkgPublic and Nonce rounds.  So I think we just ran out of time for the 30 seconds the integration test allocates before stopping the Coordinator.

In a follow on PR I will add another channel so that the Signer can tell us results of operations, like the channel I added so we can give it commands async.  I can do that now if you'd like.

stacks-signer/src/crypto/frost.rs
xoloki: I have a moderate distaste for `panic!` in anything that runs in a network context.  Having said that, we `panic!` if we fail to sign messages, since that means our private key is completely broken and we can't reasonably continue.  

For state machine errors, I can see a possibility of getting an unexpected message that triggers a bad state machine change.  If we can handle that kind of thing without `panic!` I think we should.

stacks-signer/src/crypto/frost.rs
xoloki: I think that's a very good idea.  Let me know if you want me to do it as part of this PR or if it can wait for the follow on.

stacks-signer/src/crypto/frost.rs
xoloki: Can't do it as part of this PR because the relevant code is in `stacks-sbtc`.  We are planning on number of `stacks-sbtc` changes before the follow on PR.

stacks-signer/src/runloop.rs
xoloki: Presumably there will be a small set of things we are willing to sign, and they will definitely need to be validated before doing so.  There's lots of weird crypto things that can happen if you can trick people into signing arbitrary messages.

It's definitely something to think about but seems very orthogonal to this PR itself.

stacks-signer/src/runloop.rs
xoloki: I certainly think so.  Hopfully clippy won't hate the extra parens.

Fixed.

stacks-signer/src/runloop.rs
xoloki: It can absolutely underflow if we don't start our `key_ids` at `1`.  But we are starting at `1`, explicitly.

After I fix `wsts` to eliminate the whole off-by-one nastiness this will go away.

stacks-signer/src/runloop.rs
xoloki: `frost-signer` is part of `stacks-sbtc`, so IMO these TODOs should be part of follow on PR after we've made a number of changes to `wsts` and `stacks-sbtc`.

stacks-signer/src/runloop.rs
xoloki: More `stacks-sbtc` changes.

stacks-signer/src/stacks_client.rs
```diff
@@ -0,0 +1,111 @@
...
+            debug!("Sending a chunk to stackerdb!\n{:?}", chunk.clone());
+            let chunk_ack = self.stackerdb_session.put_chunk(chunk)?;
+            self.slot_versions.insert(slot_id, slot_version);
+            info!("ACK: {:?}", &chunk_ack);
+            if chunk_ack.accepted {
+                debug!("Chunk accepted by stackerdb! ACK: {:?}", chunk_ack);
+                return Ok(chunk_ack);
+            }
+            if let Some(reason) = chunk_ack.reason {
+                // TODO: fix this jankiness. Update stackerdb to use an error code mapping instead of just a string
```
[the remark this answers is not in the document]
xoloki: @jferrant do you want to add an issue for this, and the other TODOs and `wsts`/`stacks-sbtc` changes?

stacks-signer/src/stacks_client.rs
```diff
@@ -0,0 +1,111 @@
...
+                    return Err(ClientError::PutChunkRejected(reason));
+                }
+            }
+        }
+    }
+
+    /// Retrieve the total number of slots allocated to a stacker-db writer
+    #[allow(dead_code)]
+    pub fn slots_per_user(&self) -> u32 {
+        // TODO: retrieve this from the stackerdb instance?
```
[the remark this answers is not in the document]
xoloki: Do you want that as part of this PR, or is it okay if @jferrant makes and issue and we deal with it later?

stacks-signer/src/crypto/frost.rs
xoloki: Lol actually I had bad copypasta, the aggregate public key was actually being computed not the nonce XD

Fixed now, will test repeatedly until I see this nonce repeat and it still works.

xoloki: Yup, just watched current code do 3 nonce rounds before getting an even Y, and SchnorrProof worked.

Fixed for realz.

stacks-signer/src/crypto/frost.rs
xoloki: I went ahead and implemented this, which should make it less work for when I implement `FIRE`.

@jferrant we got another TODO: incrementing `dkg_public_id` if public distribution fails (due to odd y coord on agg public key) doesn't work because we don't have a `DkgPublicBegin` message; we just re-send `DkgBegin`, so we're always sending `dkg_public_id: 0` from the `SigningRound`.

stacks-signer/src/stacks_client.rs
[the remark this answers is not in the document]
jferrant: Already done. Thought I had pushed code to include the github issue? https://github.com/stacks-network/stacks-blockchain/issues/3917

stacks-signer/src/stacks_client.rs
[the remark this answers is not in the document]
jferrant: I will make an issue and link it in here. i think the sooner we get this in the better and then we can all work on the various todos without clobbering each other.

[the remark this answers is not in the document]
xoloki: Yeah let's try to get this merged ASAP.  I have the change ready that adds a `Sender` to everything so we can get notified whenever of all `OperationResult`s.  But it needs a tiny `wsts` fix and I don't want to hold this PR up.

stacks-signer/src/crypto/frost.rs
xoloki: Follow on branch is ready with extra channel for `OperationResult`s.  Will wait until this is merged and try to batch a bunch of upstream changes.

stacks-signer/src/config.rs
jferrant: I can update the comment on the stacks_private_key as well to indicate what it is used for if that helps (communicating with the stacks node/stacker db contract)?

stacks-signer/src/crypto/frost.rs
jferrant: Yeah adding this channel should also make the return time of the run_one_pass irrelevant and is closely related to this issue. https://github.com/stacks-network/stacks-blockchain/issues/3916

xoloki: @jferrant here's a draft of the follow on PR with a results channel: 

https://github.com/stacks-network/stacks-blockchain/pull/3923

jferrant (approved):
Excited to see this merged. :)

stacks-signer/src/config.rs
jcnelson: Thanks for updating the comment @jferrant!

@xoloki I agree in principal that the keys ought to be separate and have different storage strategies, but I'm curious why the signer needs to have a key that controls large amounts of STX in the first place?  When the user stacks their STX in `pox-4`, they'd additionally set a dedicated signing key for the purposes of signing blocks, chunks, and transactions (i.e. the counterpart to `message_private_key`).  The STX would presumably be managed by the user's wallet, not the signer.  Additionally, the PoX rewards would be sent to a (separate) address of the user's choosing, which almost certainly would not be under the control of the signer.  Please help me understand what I'm missing?

stacks-signer/src/crypto/frost.rs
jferrant: Oh missed this. Adding https://github.com/stacks-network/stacks-blockchain/issues/3924

libsigner/src/runloop.rs
```diff
@@ -63,10 +63,11 @@ pub trait SignerRunLoop<R> {
     fn main_loop<EVST: EventStopSignaler>(
         &mut self,
         event_recv: Receiver<StackerDBChunksEvent>,
+        command_recv: Receiver<CMD>,
         mut event_stop_signaler: EVST,
     ) -> Option<R> {
         loop {
-            let poll_timeout = self.get_event_timeout();
+            let poll_timeout = Duration::from_millis(128); //self.get_event_timeout();
```
jferrant, line 70: I think this actually can be fixed in this PR. If you just update the duration set in the timeout in the config file to be 128 millisecs, it should have the same effect. It is an optional that defaults to 5 seconds right now. however, you could add it as an option in the build_singer_config_tomls function or just chane the EVENT_TIMEOUT_SECS default.

stacks-signer/src/crypto/frost.rs
jcnelson: Either way is fine.  If you do it in a subsequent PR, then can you create an issue for it?  Thanks!

stacks-signer/src/crypto/frost.rs
```diff
@@ -0,0 +1,888 @@
...
+                .iter()
+                .fold(Point::default(), |s, (_, dps)| s + dps.public_share.A[0]);
+            // check to see if aggregate public key has even y
+            if key.has_even_y() {
+                info!("Aggregate public key has even y coord!");
+                info!("Aggregate public key: {}", key);
+                self.aggregate_public_key = key;
+                self.move_to(State::DkgPrivateDistribute)?;
+            } else {
+                warn!("DKG Round #{} Failed: Aggregate public key does not have even y coord, re-running dkg.", self.current_dkg_id);
```
jcnelson, line 299: I remember a while back we were discussing this possible outcome.  I don't remember this point, though: can this happen by accident, or does this condition only arise from malicious behavior on the coordinator's part?

Either way the signers need to re-run DKG, but if this only happens due to malicious behavior, then should the logic here identify the malicious actor and exclude them from the next DKG round?

Also, regarding the comment below -- does _not_ incrementing the `dkg_public_id` (or having some way of tracking the distribute/gather instance) mean that some signers' gathered public shares could be from a "stale" distribute/gather cycle?  If so, is that a problem?

stacks-signer/src/crypto/frost.rs
jcnelson: Gotcha, thanks!

stacks-signer/src/crypto/frost.rs
```diff
@@ -0,0 +1,888 @@
...
+                sig_share_response.signature_shares.clone(),
+            );
+            self.ids_to_await.remove(&sig_share_response.signer_id);
+            info!(
+                "Sign round #{} SignShareResponse from signer #{}. Waiting on {:?}",
+                sig_share_response.sign_id, sig_share_response.signer_id, self.ids_to_await
+            );
+        }
+        if self.ids_to_await.is_empty() {
+            // Calculate the aggregate signature
```
jcnelson, line 447: Would it make sense to use a draining iterator for the following?  This state won't be used again in this signing round, right?  It would save you some `clone()`s and ensure this signature generation step only happens at most once.

stacks-signer/src/runloop.rs
jcnelson: > There's lots of weird crypto things that can happen if you can trick people into signing arbitrary messages.

That's what I'm worried about.

> It's definitely something to think about but seems very orthogonal to this PR itself.

Either way, it seems important that we don't forget to do this.  Can you put this into an issue before merging?

stacks-signer/src/runloop.rs
jcnelson: Gotcha.  Just trying to make sure that they don't get dropped on the floor :pray:

stacks-signer/src/stacks_client.rs
```diff
@@ -0,0 +1,115 @@
...
+    #[allow(dead_code)]
+    pub fn slots_per_user(&self) -> u32 {
+        // TODO: retrieve this from the stackerdb instance and make it a function of a given signer public key
+        // See: https://github.com/stacks-network/stacks-blockchain/issues/3921
+        SLOTS_PER_USER
+    }
+}
+
+/// Helper function to determine the slot ID for the provided stacker-db writer id and the message type
+fn slot_id(id: u32, message: &MessageTypes) -> u32 {
```
jcnelson, line 101: Linking the comment I had on the other PR here for visibility: https://github.com/stacks-network/stacks-blockchain/pull/3923#discussion_r1326362766

jcnelson:
> BTW CI is failing with: @xoloki 

I get something similar when building on Alpine.  It seems that CI needs to be updated to install `libLLVM-16` development libraries and headers (in particular, the static libraries).  It comes from `p256k1`.

stacks-signer/src/stacks_client.rs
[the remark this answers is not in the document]
jcnelson: Sounds good; thanks for adding the issue @jferrant!

stacks-signer/src/crypto/frost.rs
jcnelson: Ah nm, I see this was discussed in another comment thread :pray:

jcnelson:
I only have three remaining concerns before approving:

* Getting CI to pass (probably just need to install LLVM dev packages)
* Trying to figure out why `stacks_private_key` is necessary
* Trying to figure out if we need more slots in the StackerDB, or if we can get away with giving the signer only one slot (instead of N slots for each of the N signing states).

stacks-signer/src/crypto/frost.rs
jferrant: Issue made :D https://github.com/stacks-network/stacks-blockchain/issues/3925

stacks-signer/src/runloop.rs
jferrant: Issue created https://github.com/stacks-network/stacks-blockchain/issues/3926

stacks-signer/src/config.rs
xoloki: Totally possible that I'm making bad assumptions here.  I though the way PoX worked was that users stacked significant (~100k) STX for each reward slot, and that the STX address associated with the stacking wallets would be the same STX address that Signers will use to talk to sbtc contracts etc.

If this is a bad assumption, and the STX address the Signer will use with stackerDB is not the address which controls the large STX stake, then yeah we can just use the same private key for both.

xoloki:
> * Getting CI to pass (probably just need to install LLVM dev packages)

I'll fix this today.

> * Trying to figure out why `stacks_private_key` is necessary

See my comment above.

> * Trying to figure out if we need more slots in the StackerDB, or if we can get away with giving the signer only one slot (instead of N slots for each of the N signing states).

We don't need a set of slots per key_id/reward slot, we just need a set of slots per stacking entity/Signer.  This will be more like 100, so with like 8 slots per `Signer` we only need like 800 slots total (currently we're using 16 slots per `Signer` but we don't need anywhere near that many).  I think we'll be okay.

stacks-signer/src/config.rs
jcnelson: The users don't stack to specific reward slots per se; they just stack a large quantity of STX, and depending on how much STX they stacked relative to everyone else, they are allotted zero or more reward slots by the system at the start of the next reward cycle.

That said, the key that holds the STX and performs the stacking operation would _not_ need to be known to the signer.  Instead, the act of stacking from this high-value key would assign a _different_ block-signing key for the signer to use.  Depending on what you think about this, it may even be desirable to have the user delegate _two_ keys to the signer: one for signing Stacks blocks, and one for signing Bitcoin sBTC transactions.  But in no case would the signer ever possess the key that controls the user's STX.

jcnelson (approved):
LGTM. Thank you for your clarifications @xoloki!  Once CI passes, feel free to merge :pray:

xoloki:
Ugh, CI just keeps throwing more errors:

#11 598.3    Compiling stacks-node v0.1.0 (/build/testnet/stacks-node)
#11 653.7 LLVM ERROR: IO failure on output stream: No space left on device

PR #3920, opened 2023-09-13
[Signer] add generate-files cli option and update README accordingly

### LINKED ISSUE #3918: [Signer] Add means to generate testing toml, stacker-db contracts, and config files
Should be able to generate all necessary files for testing a collection of signers with some variable number of signers and key ids per signer.

jferrant:
Closes https://github.com/stacks-network/stacks-blockchain/issues/3918


stacks-signer/src/main.rs
```diff
@@ -188,83 +96,160 @@ fn spawn_running_signer(
...
+            .collect::<Vec<StacksPrivateKey>>()
     };
+    let signer_stacks_addresses = signer_stacks_private_keys
+        .iter()
+        .map(|key| to_addr(key, &args.network))
+        .collect::<Vec<StacksAddress>>();
+    // Build the stackerdb contract
+    let stackerdb_contract = build_stackerdb_contract(&signer_stacks_addresses);
+    debug!("Stacker DB Contract: {}", &stackerdb_contract);
+    let contract_id = QualifiedContractIdentifier::new(
```
jcnelson, line 161: This implies that the first signer must be the one to submit the contract.  Is this intentional?  Would it make sense to just write the contract code to a file containing only the contract name?

stacks-signer/src/tests/conf/signer-0.toml
```diff
@@ -0,0 +1,19 @@
+
+message_private_key = "CosqumxnRfykP24DdUCDEEMZ2qCwHPUyBy21Qf55EPKJ"
+stacks_private_key = "8a99f02ddede302a0e9fbb4eb4ef1bd587a9c076f67230c3e7a62a9d78d34d0401"
+node_host = "127.0.0.1:20444"
```
jcnelson, line 4: Port 20444 is usually the p2p interface.  The RPC interface is usually on 20443.  Is this intentional here?

jferrant: Naw this was not intentional. Was just the first port I thought of when I called the cli command to generate the files.

xoloki (approved):
LGTM

jcnelson (approved):
LGTM

PR #3923, opened 2023-09-14
Send Signer results over a channel

xoloki:
### Description

Right now the only way to find out what a `Signer` has done is to stop it and get the cached results.  

Add another channel which lets `Signer` send results as they come in.

### Applicable issues
- Closes https://github.com/stacks-network/stacks-blockchain/issues/3916

### Checklist
- [ ] Changelog is updated

### DIFF
--- a/stacks-signer/src/main.rs
+++ b/stacks-signer/src/main.rs
@@ -30,7 +30,7 @@ use clap::Parser;
 use clarity::vm::types::QualifiedContractIdentifier;
 use libsigner::{RunningSigner, Signer, SignerSession, StackerDBEventReceiver, StackerDBSession};
 use libstackerdb::StackerDBChunkData;
-use slog::slog_debug;
+use slog::{slog_debug, slog_warn};
 use stacks_common::{
     address::{
         AddressHashMode, C32_ADDRESS_VERSION_MAINNET_SINGLESIG,
@@ -38,11 +38,12 @@ use stacks_common::{
     },
     debug,
     types::chainstate::{StacksAddress, StacksPrivateKey, StacksPublicKey},
+    warn,
 };
 use stacks_signer::{
     cli::{
         Cli, Command, GenerateFilesArgs, GetChunkArgs, GetLatestChunkArgs, PutChunkArgs,
-        StackerDBArgs,
+        RunDkgArgs, SignArgs, StackerDBArgs,
     },
     config::{Config, Network},
     crypto::{frost::Coordinator as FrostCoordinator, OperationResult},
@@ -57,6 +58,12 @@ use std::{
     sync::mpsc::{channel, Receiver, Sender},
     time::Duration,
 };
+
+struct SpawnedSigner {
+    running_signer: RunningSigner<StackerDBEventReceiver, Vec<OperationResult>>,
+    cmd_send: Sender<RunLoopCommand>,
+    res_recv: Receiver<Vec<OperationResult>>,
+}
 
 /// Create a new stacker db session
 fn stackerdb_session(host: SocketAddr, contract: QualifiedContractIdentifier) -> StackerDBSession {
@@ -78,15 +85,13 @@ fn write_chunk_to_stdout(chunk_opt: Option<Vec<u8>>) {
     }
 }
 
-fn spawn_running_signer(
-    path: &PathBuf,
-    command: RunLoopCommand,
-    cmd_recv: Receiver<RunLoopCommand>,
-    res_send: Sender<Vec<OperationResult>>,
-) -> RunningSigner<StackerDBEventReceiver, Vec<OperationResult>> {
+// Spawn a running signer and return its handle, command sender, and result receiver
+fn spawn_running_signer(path: &PathBuf) -> SpawnedSigner {
     let config = Config::try_from(path).unwrap();
+    let (cmd_send, cmd_recv) = channel();
+    let (res_send, res_recv) = channel();
     let ev = StackerDBEventReceiver::new(vec![config.stackerdb_contract_id.clone()]);
-    let runloop: RunLoop<FrostCoordinator> = RunLoop::new(&config, command);
+    let runloop: RunLoop<FrostCoordinator> = RunLoop::from(&config);
     let mut signer: Signer<
         RunLoopCommand,
         Vec<OperationResult>,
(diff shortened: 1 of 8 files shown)


libsigner/src/session.rs
```diff
@@ -96,9 +96,9 @@ impl StackerDBSession {
     where
         F: FnOnce(&mut StackerDBSession, &mut TcpStream) -> R,
     {
-        if self.sock.is_none() {
-            self.connect_or_reconnect()?;
-        }
+        // TODO: fix this so we can use persistent connection
```
jcnelson, line 99: I strongly suspect that the node is closing the socket after it dumps the event, but not on a predictable schedule (see https://github.com/stacks-network/stacks-blockchain/blob/master/testnet/stacks-node/src/event_dispatcher.rs#L126).  The socket gets consumed by `async_h1`, which presumably drops it and closes it at some future date.  A next-step on this issue would be to update the node to either (1) always close the socket _explicitly_ once the payload has been sent, or (2) require `async_h1` to keep it open so the TCP window can at least be reused.  I don't care which one happens, but it would at least let us address this comment.

stacks-signer/src/config.rs
```diff
@@ -38,25 +43,70 @@ pub enum ConfigError {
...
     pub stackerdb_contract_id: String,
-    /// the private key used to sign blocks, chunks, and transactions in hexademical format
-    pub private_key: String,
+    /// the 32 byte ECDSA private key used to sign blocks, chunks, and transactions
+    pub message_private_key: String,
+    /// The hex representation of the signer's Stacks private key
+    pub stacks_private_key: String,
+    /// The network to use. One of "mainnet" or "testnet".
+    pub network: Network,
+    // TODO: Optionally retrieve the signers from the pox contract
```
jcnelson, line 102: Doesn't have to be done in this PR ofc, but it's not optional ;)  The default behavior will be to obtain the signers' public keys from `pox-4` (or, at the user's discretion, a conforming smart contract that could be used to mock `pox-4` for testing).

jcnelson: Just commenting here so it gets on y'all's radar.

stacks-signer/src/stacks_client.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Retrieve the total number of slots allocated to a stacker-db writer
+    #[allow(dead_code)]
+    pub fn slots_per_user(&self) -> u32 {
+        // TODO: retrieve this from the stackerdb instance?
+        SLOTS_PER_USER
+    }
+}
+
+/// Helper function to determine the slot ID for the provided stacker-db writer id and the message type
+fn slot_id(id: u32, message: &MessageTypes) -> u32 {
```
jcnelson, line 100: Right now, stackerdb only permits up to 4096 slots per replica.  If we have 2000 signers (one signer per reward slot), this would mean each signer gets two slots.

The reason for the 4096 cap is because nodes must exchange slot version vectors with one another to detect the availability of new chunks, and the size of the message is basically 4x the number of slots (each version is 4 bytes).  The cap could be raised, but I'm not sure yet that its wise to make it an order of magnitude bigger.  I'd want to do some performance tests to see if nodes regularly exchanging 160kb chunks of data with up to 32 neighbors on each write is sustainable (that's just over 5 MB write amplification per slot push).

jcnelson: If it turns out that the 4096 cap is the upper bound, would it be feasible to adapt this code later on so that each signer gets a single slot?  If we can swing this, then the task of setting up the next set of stacker signers at a reward cycle boundary is made easier:  the first 2048 slots would belong to the current signers, and the next 2048 slots would belong to the next signers coming online.

stacks-signer/src/tests/contracts/signers-dkg.clar
```diff
@@ -0,0 +1,37 @@
+;; title: signers-dkg
+;; version:
+;; summary: Signer StackerDB test contract
+;; description:
+
+;; traits
+;;
+(impl-trait 'ST1NXBK3K5YYMD6FD41MVNP3JS1GABZ8TRVX023PT.sip-010-trait-ft-standard.sip-010-trait)
```
jcnelson, line 8: Is there a reason this code needs to implement SIP-010?  If so, then it's missing the SIP-010 trait methods and deployment will fail.

jferrant: This is removed in my other PR. It is not required as far as I can tell.

libsigner/src/session.rs
jferrant: I have created a ticket already for this https://github.com/stacks-network/stacks-blockchain/issues/3922 . I will link to this comment for additional context :D

stacks-signer/src/config.rs
jferrant: Only reason I have this marked optional is because I could see for testing purposes that we should be able to provide this information in a config rather than just always from the pox contract.

stacks-signer/src/main.rs
```diff
@@ -169,6 +209,8 @@ fn main() {
         )
     };
 
+    let (_cmd_send, cmd_recv) = channel();
+    let (res_send, _res_recv) = channel();
```
jferrant, line 213: You should update the DKG and SIGN CLI commands to use this res_recv so that they terminate upon completion.

jferrant (changes requested):
I updated the state machine substantially and fixed the test to actually tear down the signers at the end. My initial check on the meaty portion (following node and signer setup), it takes ~15 seconds to run DKG and Sign a message.

stacks-signer/src/stacks_client.rs
xoloki: As we discussed in the first Signer PR, we only need one set of slots per Signer, not per reward slot.  So 2048 is more than enough for our use case with ~100 Signers.

xoloki:
I fixed the build so we don't expect any additional types to `impl Clone`, and rebased on the merged `signer-first` PR.  This PR should be ready to go now.

stacks-signer/src/main.rs
```diff
@@ -79,22 +85,66 @@ fn write_chunk_to_stdout(chunk_opt: Option<Vec<u8>>) {
...
+    SpawnedSigner {
+        running_signer,
+        cmd_send,
+        res_recv,
+    }
+}
+
+// Process a DKG result
+fn process_dkg_result(dkg_res: &[OperationResult]) {
+    if dkg_res.len() != 1 {
```
jcnelson, line 112: Why doesn't this method just take a single `&OperationResult` instead of a slice reference?

jcnelson: Same question for `process_sign_result()` below

stacks-signer/src/runloop.rs
```diff
@@ -68,60 +50,47 @@ pub struct RunLoop<C> {
...
-                        .stacks_client
-                        .send_message(self.signing_round.signer.signer_id, msg);
+    fn execute_command(&mut self) {
+        match self.state {
+            State::Idle => {
+                if let Some(command) = self.commands.pop_front() {
+                    match command {
+                        RunLoopCommand::Dkg => {
+                            info!("Starting DKG");
+                            if let Ok(msg) = self.coordinator.start_distributed_key_generation() {
```
jcnelson, line 69: What happens if this fails? What's the `Err(..)` case?  Same question for `RunLoopCommand::Sign` below with `start_signing_message()`.

stacks-signer/src/runloop.rs
```diff
@@ -68,60 +50,47 @@ pub struct RunLoop<C> {
...
-                self.state = State::Sign
             }
-            (RunLoopCommand::DkgSign { message }, State::SignDone)
-            | (RunLoopCommand::Sign { message }, State::SignDone) => {
-                info!("Done signing message: {:?}", message);
-                self.state = State::Exit;
+            State::Dkg | State::Sign => {
+                // We cannot execute the next command until the current one is finished...
+                // Do nothing
+                debug!("Waiting for operation to finish");
```
jcnelson, line 92: Does this command just get dropped then?  Asking because the signer daemon will receive a list of things it needs to sign (sBTC transactions, block headers), and it can't just drop them.  Will these requests be buffered up somewhere?

jferrant: It gets stored in the command queue until it is able to process the next one. This will have to be reviewed down the line to enable processing multiple things at once, but currently it does it sequentially, buffering commands as they are received until it can continue processing them.

stacks-signer/src/main.rs
jferrant: Because the run one pass can in theory return multiple items (it doesn't right now because of how we do things sequentially).  I added this check because we would not expect more than one DKG result or more than one sign. However, handling it internally to this function call made more sense to me/removed the most amount of duplicate code. In theory of course I can assert the size of the list outside of this function call and access the first element in the list, but then we will have more duplicate code in handle_dkg, handle_sign, and handle_dkg_sign.

stacks-signer/src/runloop.rs
jferrant: Good point. If it fails, we should not update the state to State::Dkg because it will become stuck. We should move the state update internal if let Ok() check. Currently if start_distributed_key_generation fails. Beyond this, it would imply that there is something substantially wrong with the coordinator state. Not sure if we should panic here, just warn. Attempt to push back the command and reprocess it? Not sure how it should fail here. @xoloki

jferrant: Done :)

jferrant (approved):

jcnelson (approved):
LGTM. Thanks for addressing my comments!

stacks-signer/src/main.rs
jcnelson: I think it's fine as long as it's just part of `main`'s CLI handling.  I was mainly curious as to why the runloop itself doesn't just send a single `OperationResult` back instead of `Vec<OperationResult>`, because it seemed to be the case that the code always expected a `Vec` with exactly one element.  But if in the future there will be multiple items, then this is fine.

PR #3933, opened 2023-09-18
Nakamoto: Nakamoto blocks and block headers, append_block

kantai:
This PR adds Nakamoto block and block header structs, as well as a `NakamotoChainstate::append_block` method.

This mocks the "tenure change" transaction check -- all nakamoto blocks are assumed to be tenure changes in this append_block implementation. This PR also highlights the need for some thinking around the "chain height" in Nakamoto -- miner rewards in pre-nakamoto mature after a certain number of stacks blocks, but in Nakamoto, stacks blocks are no longer limited by the bitcoin block count. So miner rewards must either mature at a given burnchain height or after a number of tenure changes.

This includes a very simple unit test -- it just invokes `append_block` on a new NakamotoBlock that builds off of the genesis (its parent is a non-NakamotoBlock).


stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1056 @@
+use std::ops::DerefMut;
```
jcnelson, line 1: Minor nit: needs copyright comment block

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1056 @@
...
+          -- The following fields are not part of either the StacksHeaderInfo struct
+          --   or its contained NakamotoBlockHeader struct, but are used for querying
+                     header_type TEXT NOT NULL,
+                     block_hash TEXT NOT NULL,
+                     -- index_block_hash is the hash of the block hash and consensus hash of the burn block that selected it, 
+                     -- and is guaranteed to be globally unique (across all Stacks forks and across all PoX forks).
+                     -- index_block_hash is the block hash fed into the MARF index.
+                     index_block_hash TEXT NOT NULL,
+                     -- the total cost of the block
+                     cost TEXT NOT NULL,
```
jcnelson, line 108: Can you explain more what `cost` means here?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1056 @@
...
+                     burn_header_height INT NOT NULL,
+                     -- timestamp from burnchain block header that generated this consensus hash
+                     burn_header_timestamp INT NOT NULL,
+                     block_size TEXT NOT NULL,
+          -- The following fields all correspond to entries in the NakamotoBlockHeader struct
+                     version INTEGER NOT NULL,
+                     -- this field is the total number of blocks in the chain history (including this block)
+                     chain_length INTEGER NOT NULL,
+                     -- this field is the total amount of BTC spent in the chain history (including this block)
+                     btc_spent INTEGER NOT NULL,
```
jcnelson, line 89: Can this be renamed to `burn_spent` to keep it consistent with the burnchain nomenclature in the code?

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: Ah, I see -- this is the serialized `ExecutionCost` that represents how much runtime this block consumed, right?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1056 @@
...
+    }
+
+    /// Called in both follower and miner block assembly paths.
+    ///
+    /// Returns clarity_tx, list of receipts, microblock execution cost,
+    /// microblock fees, microblock burns, list of microblock tx receipts,
+    /// miner rewards tuples, the stacks epoch id, and a boolean that
+    /// represents whether the epoch transition has been applied.
+    ///
+    /// The `burn_dbconn`, `sortition_dbconn`, and `conn` arguments
```
jcnelson, line 516: I think trait upcasting can still happen though.  You could do something like what I've done for the StackerDB event observer, where you have a "super-trait" that combines `BurnStateDB` and `SortitionDBRef`, as well as a third trait that provided `fn conn(&self) -> &Connection` to recover the inner DB connection.  Then, you could make traits that convert a struct which implements the super-trait into a struct which implements one of the above traits, and create blanket implementations for each of these traits.  For example:

```rust
/// A trait for implementing both mempool event observer methods and stackerdb methods.
/// This is required for event observers to fully report on newly-relayed data.
pub trait RelayEventDispatcher:
    MemPoolEventDispatcher
    + StackerDBEventDispatcher
    + AsMemPoolEventDispatcher
    + AsStackerDBEventDispatcher
{
}
impl<T: MemPoolEventDispatcher + StackerDBEventDispatcher> RelayEventDispatcher for T {}

/// Trait for upcasting to MemPoolEventDispatcher
pub trait AsMemPoolEventDispatcher {
    fn as_mempool_event_dispatcher(&self) -> &dyn MemPoolEventDispatcher;
}

/// Trait for upcasting to StackerDBEventDispatcher
pub trait AsStackerDBEventDispatcher {
    fn as_stackerdb_event_dispatcher(&self) -> &dyn StackerDBEventDispatcher;
}

impl<T: RelayEventDispatcher> AsMemPoolEventDispatcher for T {
    fn as_mempool_event_dispatcher(&self) -> &dyn MemPoolEventDispatcher {
        self
    }
}

impl<T: RelayEventDispatcher> AsStackerDBEventDispatcher for T {
    fn as_stackerdb_event_dispatcher(&self) -> &dyn StackerDBEventDispatcher {
        self
    }
}
```

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1056 @@
...
+    ) -> Result<SetupBlockResult<'a, 'b>, ChainstateError> {
+        let parent_index_hash = StacksBlockId::new(&parent_consensus_hash, &parent_header_hash);
+        let parent_sortition_id = burn_dbconn
+            .get_sortition_id_from_consensus_hash(&parent_consensus_hash)
+            .expect("Failed to get parent SortitionID from ConsensusHash");
+
+        // find matured miner rewards, so we can grant them within the Clarity DB tx.
+        // TODO: this must be updated to either:
+        //   (A) use the TENURE HEIGHT -- i.e., count of tenures, rather than count of stacks blocks
+        //   (B) use BURNCHAIN HEIGHT
```
jcnelson, line 547: In the Nakamoto v1 design doc, these are on 1-to-1 correspondence.  Stackers still create a `TenureChange` transaction for each Bitcoin block, even if there was no miner for the tenure.  A subsequent miner just packages the `TenureChange` transactions in order (see Figure 3 here: https://github.com/stacks-network/stacks-blockchain/discussions/3892).

jcnelson: I would use tenure height, since then you can correlate it to one of these `TenureChange` transactions and the block that contained it (and then recover the burnchain block in which the Stacks block was mined)

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1056 @@
...
+            &block.header.block_hash().to_hex(),
+            block.txs.len()
+        );
+
+        let ast_rules = ASTRules::PrecheckSize;
+
+        let mainnet = chainstate_tx.get_config().mainnet;
+        let next_block_height = block.header.chain_length;
+
+        let (parent_ch, parent_block_hash) = if block.is_first_mined() {
```
jcnelson, line 733: I don't think this will ever be true?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1056 @@
...
+                        warn!(
+                            "CORRUPTION: block {}/{} does not correspond to a burn block",
+                            &parent_ch, &parent_block_hash
+                        );
+                        (BurnchainHeaderHash([0; 32]), 0, 0)
+                    }
+                }
+            };
+
+        let epoch_receipt = StacksEpochReceipt {
```
jcnelson, line 1002: Now that we use the word "epoch" to refer to a set of consensus rules, can we call this something else? e.g. "StacksBlockReceipt?"

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1056 @@
...
+impl StacksMessageCodec for NakamotoBlock {
+    fn consensus_serialize<W: std::io::Write>(&self, fd: &mut W) -> Result<(), CodecError> {
+        write_next(fd, &self.header)?;
+        write_next(fd, &self.txs)
+    }
+
+    fn consensus_deserialize<R: std::io::Read>(fd: &mut R) -> Result<Self, CodecError> {
+        let header: NakamotoBlockHeader = read_next(fd)?;
+        let txs: Vec<_> = {
+            let mut bound_read = BoundReader::from_reader(fd, MAX_MESSAGE_LEN as u64);
```
jcnelson, line 1029: Nit -- you should use a `bound_read` for decoding the `NakamotoBlockHeader` as well as the txs.  The header plus the txs should not exceed `MAX_MESSAGE_LEN` bytes.

jcnelson:
Thanks for cleaning up all of those imports!

Also, should this be against `next` or `develop`? Right now, it's against `develop`.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Yep, exactly -- this is the same as the current `cost` field in the existing headers table.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Yep -- I agree, tenure height is the thing to use here.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Maybe in `helium`?

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Sure -- should I do this rename in this PR? (I can make sure it's in a single commit of its own)

stackslib/src/chainstate/nakamoto/mod.rs
kantai: I think it may be more straightforward to just expand the `SortitionDBRef` trait:

```rust
pub trait SortitionDBRef: BurnStateDB {
    fn get_pox_start_cycle_info(
        &self,
        sortition_id: &SortitionId,
        parent_stacks_block_burn_ht: u64,
        cycle_index: u64,
    ) -> Result<Option<PoxStartCycleInfo>, ChainstateError>;

    /// Return an upcasted dynamic reference for the sortition DB
    fn as_burn_state_db(&self) -> &dyn BurnStateDB;

    /// Return a pointer to the underlying sqlite connection or transaction for
    /// this DB reference
    fn sqlite_conn(&self) -> &Connection;
}
```

Because `SortitionDBRef` already requires any implementor also implement `BurnStateDB`, this doesn't really alter the requirements for the trait.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Updated

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Implemented the above (added `as_burn_state_db` to `SortitionDBRef` trait)

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Implemented miner reward schedules in this PR (with assertions in the unit tests) that uses tenure-height to track maturity. It also computes tx fees for the prior tenure at tenure changes.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Updated

kantai:
> Also, should this be against next or develop? Right now, it's against develop.

Rebased against `next` now.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1375 @@
...
+    pub txs: Vec<StacksTransaction>,
+}
+
+pub struct NakamotoChainState;
+
+impl FromRow<NakamotoBlockHeader> for NakamotoBlockHeader {
+    fn from_row(row: &rusqlite::Row) -> Result<NakamotoBlockHeader, DBError> {
+        let version = row.get("version")?;
+        let chain_length_i64: i64 = row.get("chain_length")?;
+        let chain_length = chain_length_i64.try_into().map_err(|_| DBError::Overflow)?;
```
jbencin, line 203: It's cleaner to use shadowing than create intermediate variables:

```suggestion
        let chain_length: i64 = row.get("chain_length")?;
        let chain_length = chain_length.try_into().map_err(|_| DBError::Overflow)?;
```

Also, if a `i64` fails to convert to a `u64`, shouldn't that be a `TypeError` (because it's a negative number), not an `Overflow` error

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1375 @@
...
+                    VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10, ?11, ?12, ?13, ?14, ?15, ?16, ?17, ?18, ?19, ?20, ?21, ?22, ?23, ?24, ?25)",
+            args
+        )?;
+
+        Ok(())
+    }
+
+    /// Append a Stacks block to an existing Stacks block, and grant the miner the block reward.
+    /// Return the new Stacks header info.
+    pub fn advance_tip(
```
jbencin, line 547: There are a few functions here (like this one) that seem to have the same names/args as those in `StacksChainState`. Maybe we should define a `ChainState` trait?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1375 @@
...
+
+        let new_block_hash = new_tip.block_hash();
+        let index_block_hash = StacksBlockId::new(&new_consensus_hash, &new_block_hash);
+
+        // store each indexed field
+        test_debug!(
+            "Headers index_put_begin {}-{}",
+            &parent_hash,
+            &index_block_hash,
+        );
```
jbencin, line 594: I think we should prefer inline variables in format strings, it's easier to read:

```suggestion
        test_debug!(
            "Headers index_put_begin {parent_hash}-{index_block_hash}"
        );
```

jbencin (commented):
Leaving some comments, going to take a break now and add some more later...

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: IIRC we've used `ParseError` in the past, but at the end of the day, the outcome here would be that the node would panic since it's operating on corrupt chainstate.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1375 @@
...
+    /// The bitcoin block whose data has been handled most recently by
+    /// the Stacks chain as of this block.
+    pub burn_view: BurnchainHeaderHash,
+    /// The root of a SHA512/256 merkle tree over all this block's
+    /// contained transactions
+    pub tx_merkle_root: Sha512Trunc256Sum,
+    /// The MARF trie root hash after this block has been processed
+    pub state_index_root: TrieHash,
+    /// Recoverable ECDSA signature from the tenure's miner.
+    pub signature: MessageSignature,
```
jcnelson, line 188: There are two signatures we need:  the signature from the miner (this field), and the signature from the stackers (see https://github.com/stacks-network/stacks-blockchain/pull/3974/files#r1337466506)

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1375 @@
...
+    /// Return the total ExecutionCost consumed during the tenure up to and including
+    ///  `block`
+    pub fn get_total_tenure_cost_at(
+        conn: &Connection,
+        block: &StacksBlockId,
+    ) -> Result<Option<ExecutionCost>, ChainstateError> {
+        let qry = "SELECT total_tenure_cost FROM nakamoto_block_headers WHERE index_block_hash = ?";
+        conn.query_row(qry, &[block], |row| row.get(0))
+            .optional()
+            .map_err(|e| ChainstateError::DBError(e.into()))
```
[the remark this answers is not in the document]
jcnelson: I think he's copy/pasting my old not-so-Rusty code ;)

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1375 @@
...
+            if ancestor_at_height == candidate.index_block_hash() {
+                return Ok(Some(candidate));
+            }
+        }
+        Ok(None)
+    }
+
+    /// Return the tenure height of `block` if it was a nakamoto block, or the
+    ///  Stacks block height of `block` if it was an epoch-2 block
+    pub fn get_tenure_height(
```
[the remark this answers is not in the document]
jcnelson: In Stacks 2.x, the tenure height and block height are the same.  A miner's tenure in Stacks 2.x is entirely encompassed in the single Bitcoin-anchored Stacks block they produce, as well as the microblock stream they append to it.

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: Perhaps, but can we wait to do this until we're certain about the interfaces to these two structs?  It might be that they look the same right now simply because the Nakamoto ones are derived from the 2.x ones (but may change substantially later).

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1375 @@
...
+            //   block_height = tenure_height
+            let ancestor_at_height = match tx
+                .get_ancestor_block_hash(tenure_height, tip_index_hash)?
+                .map(|ancestor| Self::get_block_header(tx.tx(), &ancestor))
+                .transpose()?
+                .flatten()
+            {
+                Some(x) => x,
+                None => return Ok(None),
+            };
```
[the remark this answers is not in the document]
kantai: I did not know about this construct!

stackslib/src/chainstate/nakamoto/mod.rs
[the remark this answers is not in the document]
kantai: Yep -- as Jude said, `tenure-height` is the block height in Stacks 2.x. I'll expand the rustdoc comment here.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Yep, I agree -- I'd rather wait to define an interface until its clear something wants to consume it.

stackslib/src/chainstate/nakamoto/mod.rs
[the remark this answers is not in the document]
jbencin: The `let-else` construct was introduced almost a year ago in Rust 1.65. It can be used almost anywhere you have a `match` with `Some(x) => x` or `Ok(x) => x`. Unless we are trying to maintain compatibility with old versions of `rustc` (I don't know why we would...) we should prefer this for readability and smaller code

jbencin (commented):
A few more minor things...

stackslib/src/chainstate/stacks/db/headers.rs
```diff
@@ -111,19 +112,20 @@ impl FromRow<StacksMicroblockHeader> for StacksMicroblockHeader {
...
         affirmation_weight: u64,
     ) -> Result<(), Error> {
-        assert_eq!(
-            tip_info.stacks_block_height,
-            tip_info.anchored_header.total_work.work
-        );
+        let header = match tip_info.anchored_header {
+            StacksBlockHeaderTypes::Epoch2(ref x) => x,
+            _ => return Err(Error::InvalidChildOfNakomotoBlock),
+        };
```
jbencin, line 124: Could be simplified with `let-else`:

```suggestion
        let StacksBlockHeaderTypes::Epoch2(header) = &tip_info.anchored_header else {
            return Err(Error::InvalidChildOfNakomotoBlock)
        };
```

jbencin (commented):
Here's the last batch of comments...

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+                     block_size TEXT NOT NULL,
+          -- The following fields all correspond to entries in the NakamotoBlockHeader struct
+                     version INTEGER NOT NULL,
+                     -- this field is the total number of blocks in the chain history (including this block)
+                     chain_length INTEGER NOT NULL,
+                     -- this field is the total amount of BTC spent in the chain history (including this block)
+                     burn_spent INTEGER NOT NULL,
+                     -- the parent BlockHeaderHash
+                     parent TEXT NOT NULL,
+                     -- the latest bitcoin block whose data is viewable from this stacks block
```
jcnelson, line 113: Is this a `BurnHeaderHash`?  How is this different from `burn_header_hash` above?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+                     -- and is guaranteed to be globally unique (across all Stacks forks and across all PoX forks).
+                     -- index_block_hash is the block hash fed into the MARF index.
+                     index_block_hash TEXT NOT NULL,
+                     -- the total cost of the block
+                     cost TEXT NOT NULL,
+                     -- the total cost up to and including this block in the current tenure
+                     total_tenure_cost TEXT NOT NULL,
+                     -- the parent index_block_hash
+                     parent_block_id TEXT NOT NULL,
+                     affirmation_weight INTEGER NOT NULL,
```
jcnelson, line 136: This can get dropped post-Nakamoto

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+impl NakamotoBlockHeader {
+    pub fn block_hash(&self) -> BlockHeaderHash {
+        BlockHeaderHash::from_serializer(self)
+            .expect("BUG: failed to serialize block header hash struct")
+    }
+}
+
+impl NakamotoBlock {
+    /// Did the stacks tenure change on this nakamoto block? i.e., does this block
+    ///  include a TenureChange transaction?
```
jcnelson, line 272: More specifically, does it include a `TenureChange` transaction for the current tenure?  A block can have multiple `TenureChange` transactions -- for example, it must include them for miners who never produced a block during their tenures, per the SIP.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+            .map(|x| x.parse())
+            .transpose()
+            .map_err(|_| ChainstateError::DBError(DBError::ParseError))
+    }
+
+    /// Return a Nakamoto StacksHeaderInfo at a given tenure height in the fork identified by `tip_index_hash`
+    pub fn get_header_by_tenure_height(
+        tx: &mut StacksDBTx,
+        tip_index_hash: &StacksBlockId,
+        tenure_height: u64,
```
jcnelson, line 377: Might be worth a comment to say that `tenure_height` is equivalent to the Stacks block height pre-Nakamoto.

jcnelson: Might also be worth a comment to explain that in Nakamoto, this gets the `StacksHeaderInfo` for the first block in a Nakamoto tenure.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+            .query_row(epoch_2_qry, &[block], |row| row.get(0))
+            .optional()?;
+        opt_height
+            .map(u64::try_from)
+            .transpose()
+            .map_err(|_| ChainstateError::DBError(DBError::ParseError))
+    }
+
+    /// Load block header (either Epoch-2 rules or Nakamoto) by `index_block_hash`
+    pub fn get_block_header(
```
jcnelson, line 452: Would it make sense to try the nakamoto block headers table first?  That might be more likely to return data.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: One is the bitcoin block for which data has been processed and can be queried by clarity. The other is the bitcoin block in which the tenure was elected. They do not necessarily have to be the same (consider the case of a missed sortition).

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+            affirmation_weight,
+            tenure_height,
+            tenure_changed,
+            block_fees,
+        )
+        .expect("FATAL: failed to advance chain tip");
+
+        chainstate_tx.log_transactions_processed(&new_tip.index_block_hash(), &tx_receipts);
+
+        monitoring::set_last_block_transaction_count(block.txs.len() as u64);
```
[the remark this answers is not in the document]
jcnelson: Does `usize` always convert to `u64`?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -0,0 +1,1385 @@
...
+
+        // get previous burn block stats
+        let (parent_burn_block_hash, parent_burn_block_height, parent_burn_block_timestamp) =
+            if block.is_first_mined() {
+                (BurnchainHeaderHash([0; 32]), 0, 0)
+            } else {
+                match SortitionDB::get_block_snapshot_consensus(burn_dbconn, &parent_ch)? {
+                    Some(sn) => (
+                        sn.burn_header_hash,
+                        sn.block_height as u32,
```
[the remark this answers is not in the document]
jcnelson: Yes; we can worry about that once there are 4.1 billion blocks XD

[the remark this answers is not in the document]
jcnelson: We can have this panic of course, with a `try_into().unwrap()`.

stackslib/src/chainstate/stacks/db/accounts.rs
```diff
@@ -204,13 +212,21 @@ impl MinerPaymentSchedule {
...
 
     /// If this is a MinerPaymentSchedule for a miner who _produced_ a microblock stream, then
     /// this calculates the percentage of that stream this miner is entitled to
     pub fn streamed_tx_fees_produced(&self) -> u128 {
-        (self.tx_fees_streamed * 2) / 5
+        let tx_fees_streamed = match self.tx_fees {
+            MinerPaymentTxFees::Epoch2 { streamed, .. } => streamed,
+            MinerPaymentTxFees::Nakamoto { .. } => 0,
+        };
+        (tx_fees_streamed * 2) / 5
```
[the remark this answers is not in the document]
jcnelson: Yes, a small amount can get lost (up to 0.000004 STX).  However, this isn't worthy of a consensus-breaking change given that Nakamoto is expected to be live within the next year.

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: I'm still not seeing this.  A Nakamoto block header contains the hash of the Bitcoin block in which its tenure began.  If a sortition is missed, then Stackers grant the currently-active miner a new tenure (a new `TenureChange` transaction), which ought to cause the `burn_view` field to change to the next Bitcoin block (i.e. the one without the sortition).

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: It doesn't have to happen here

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: > I'd argue that Nakamoto Block C's "consensus_hash" should still correspond to Bitcoin Block 10.

That's where we differ.  I'd argue that Nakamoto block C's consensus hash should be the same as Bitcoin block 11's consensus hash.  This is because it vastly simplifies evaluating `(get-burn-block-info?)`:  to query the burnchain state in the `BurnStateDB` implementation, we simply query the sortition history tipped by the consensus hash in the transaction's block.  This is only possible if Nakamoto block C's consensus hash is derived from tenure in which it is mined (i.e. Bitcoin block 11), instead of the miner that mined it.

We don't do this yet, but I think we'll want to have a `tenures` table in the `NakamotoChainState` that keeps track of all tenures that have been executed (even if there was no corresponding sortition, and even if there were no Stacks blocks mined in the tenure).  That is, we need to treat tenures as first-class objects in Stacks.  If we tracked them independent of block headers, then we could do things like query the current tenure's miner (and thus its public key) as well as the tenure's corresponding sortition (so we can get the PoX payouts).  In addition, it would make it easy to select all Stacks blocks by tenure, so we can serve them via the RPC endpoint on boot-up.  We have to do all of this anyway because the design calls for one `TenureChange` transaction per Bitcoin block, and having a separate table describing them all would greatly simplify this.

jcnelson: I can add a `tenures` table in my coordinator PR

jbencin (approved):
Sorry should have checked and approved this sooner

stackslib/src/chainstate/nakamoto/mod.rs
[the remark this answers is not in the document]
jbencin: `usize` is based on pointer size. If there's some platform with 128-bit addressing (I doubt it), I guess `u64::from()` would fail to compile on it

I know these `as` conversions probably won't ever truncate in real world use, but these sorts of silent errors are a real pain to debug and it's trivial to completely avoid them, so it's better just not to `as` at all

PR #3946, opened 2023-09-19
Chore/clarity fix clippy

jferrant:
Closes https://github.com/stacks-network/stacks-blockchain/issues/3935

I had time to kill while waiting for my node to spin up...

To run clippy on clarity ONLY:

`cargo clippy -p clarity --no-deps --tests`

### DIFF
--- a/clarity/src/libclarity.rs
+++ b/clarity/src/libclarity.rs
@@ -101,9 +101,9 @@ const GIT_COMMIT: Option<&'static str> = option_env!("GIT_COMMIT");
 const GIT_TREE_CLEAN: Option<&'static str> = option_env!("GIT_TREE_CLEAN");
 
 #[cfg(debug_assertions)]
-const BUILD_TYPE: &'static str = "debug";
+const BUILD_TYPE: &str = "debug";
 #[cfg(not(debug_assertions))]
-const BUILD_TYPE: &'static str = "release";
+const BUILD_TYPE: &str = "release";
 
 pub fn version_string(pkg_name: &str, pkg_version: &str) -> String {
     let git_branch = GIT_BRANCH.unwrap_or("");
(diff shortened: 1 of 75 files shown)


jcnelson (approved):
This LGTM, but I just want to make sure @obycode and @kantai are aware of this. They may have opinions on when the appropriate time to merge it will be.

obycode (approved):
This looks great! Thanks for taking this on. I'll happily rebase on top of this and clippy my changes going forward.

kantai (approved):
LGTM, thanks!

obycode:
@jferrant when you need to merge in upstream changes, it's better to use a merge commit, rather than rebasing and force-pushing, since the latter makes it a lot harder for reviewers to see what changed.

jferrant:
> @jferrant when you need to merge in upstream changes, it's better to use a merge commit, rather than rebasing and force-pushing, since the latter makes it a lot harder for reviewers to see what changed.

I am much more of a fan of linear history and avoid merge commits as best as possible, but understand it can make it difficult to see which commit changes have been made since last review without manually inspecting.

obycode:
> I am much more of a fan of linear history and avoid merge commits as best as possible, but understand it can make it difficult to see which commit changes have been made since last review without manually inspecting.

I agree 💯 about linear history, and that's always been my preference too, but I'm not sure how to do that while making our other processes work smoothly. It's no good if GitHub can't show reviewers "changes since my last review". I don't think the tooling is there to make it work right.

So I gave up on the linear history and have embraced the merge commits 😰.

jferrant:
> > I am much more of a fan of linear history and avoid merge commits as best as possible, but understand it can make it difficult to see which commit changes have been made since last review without manually inspecting.
> 
> I agree 💯 about linear history, and that's always been my preference too, but I'm not sure how to do that while making our other processes work smoothly. It's no good if GitHub can't show reviewers "changes since my last review". I don't think the tooling is there to make it work right.
> 
> So I gave up on the linear history and have embraced the merge commits 😰.

Oh no! Something tells me this may happen to me too XD I do remember the nightmare of a time I had trying to rebase next onto a sep branch I had in order to extract out hte sbtc changes...But I did manage to get it done! and it was so easy to find the buggy code that it ....ALMOST....made the 2 days I spent rebasing worth it XD

PR #3947, opened 2023-09-19
Chore/stacks common fix clippy

jferrant:
Closes https://github.com/stacks-network/stacks-blockchain/issues/3938

To run clippy ONLY on stacks-common:
`cargo clippy -p stacks-common --no-deps --tests --all-features`

### DIFF
--- a/stacks-common/src/deps_common/httparse/mod.rs
+++ b/stacks-common/src/deps_common/httparse/mod.rs
@@ -101,7 +101,7 @@ impl<'a> Bytes<'a> {
     #[inline]
     pub fn new(slice: &'a [u8]) -> Bytes<'a> {
         Bytes {
-            slice: slice,
+            slice,
             pos: 0,
             skipped_pos: 0,
         }
@@ -119,7 +119,7 @@ impl<'a> Bytes<'a> {
 
     #[inline]
     pub fn bump(&mut self) {
-        assert!(self.pos + 1 <= self.slice_peek().len(), "overflow");
+        assert!(self.pos < self.slice_peek().len(), "overflow");
         self.pos += 1;
     }
 
@@ -133,6 +133,11 @@ impl<'a> Bytes<'a> {
     #[inline]
     pub fn len(&self) -> usize {
         self.slice_peek().len()
+    }
+
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.slice_peek().is_empty()
     }
 
     #[inline]
@@ -226,10 +231,7 @@ impl<'a, 'b: 'a> Bytes8<'a, 'b> {
 
     #[inline]
     fn new(bytes: &'a mut Bytes<'b>) -> Bytes8<'a, 'b> {
-        Bytes8 {
-            bytes: bytes,
-            pos: 0,
-        }
+        Bytes8 { bytes, pos: 0 }
     }
 
     #[inline]
@@ -242,7 +244,7 @@ impl<'a, 'b: 'a> Bytes8<'a, 'b> {
 #[inline]
 fn shrink<T>(slice: &mut &mut [T], len: usize) {
     assert!(slice.len() >= len);
-    let full = mem::replace(slice, &mut []);
+    let full = mem::take(slice);
     *slice = &mut full[..len];
 }
 
@@ -449,7 +451,7 @@ impl<'h, 'b> Request<'h, 'b> {
             method: None,
             path: None,
             version: None,
-            headers: headers,
+            headers,
         }
     }
 
@@ -516,7 +518,7 @@ impl<'h, 'b> Response<'h, 'b> {
             version: None,
             code: None,
             reason: None,
-            headers: headers,
+            headers,
         }
     }
 
@@ -641,7 +643,7 @@ fn parse_reason<'a>(bytes: &mut Bytes<'a>) -> Result<&'a str> {
         } else if b == b'\n' {
             let res = str::from_utf8(bytes.slice_skip(1)).map_err(|_e| Error::Status)?;
             return Ok(Status::Complete(res));
-        } else if !((b >= 0x20 && b <= 0x7E) || b == b'\t') {
+        } else if !((0x20..=0x7E).contains(&b) || b == b'\t') {
             return Err(Error::Status);
         }
     }
@@ -700,10 +702,7 @@ pub fn parse_headers<'b: 'h, 'h>(
 }
 
 #[inline]
-fn parse_headers_iter<'a, 'b>(
-    headers: &mut &mut [Header<'a>],
-    bytes: &'b mut Bytes<'a>,
-) -> Result<usize> {
+fn parse_headers_iter<'a>(headers: &mut &mut [Header<'a>], bytes: &mut Bytes<'a>) -> Result<usize> {
     let mut num_headers: usize = 0;
     let mut count: usize = 0;
     let mut result = Err(Error::TooManyHeaders);
@@ -1204,7 +1203,7 @@ mod tests {
         }
     }
 
-    static RESPONSE_REASON_WITH_OBS_TEXT_BYTE: &'static [u8] = b"HTTP/1.1 200 X\xFFZ\r\n\r\n";
+    static RESPONSE_REASON_WITH_OBS_TEXT_BYTE: &[u8] = b"HTTP/1.1 200 X\xFFZ\r\n\r\n";
     res! {
         test_response_reason_with_obsolete_text_byte,
         RESPONSE_REASON_WITH_OBS_TEXT_BYTE,
--- a/stacks-common/src/types/chainstate.rs
+++ b/stacks-common/src/types/chainstate.rs
@@ -21,17 +21,12 @@ pub type StacksPublicKey = Secp256k1PublicKey;
 pub type StacksPrivateKey = Secp256k1PrivateKey;
 
 /// Hash of a Trie node.  This is a SHA2-512/256.
+#[derive(Default)]
 pub struct TrieHash(pub [u8; 32]);
 impl_array_newtype!(TrieHash, u8, 32);
 impl_array_hexstring_fmt!(TrieHash);
 impl_byte_array_newtype!(TrieHash, u8, 32);
 impl_byte_array_serde!(TrieHash);
-
-impl Default for TrieHash {
-    fn default() -> TrieHash {
-        TrieHash([0x00; 32])
-    }
-}
 
 pub const TRIEHASH_ENCODED_SIZE: usize = 32;
 
@@ -83,7 +78,7 @@ impl SortitionId {
 
     pub fn new(bhh: &BurnchainHeaderHash, pox: &PoxId) -> SortitionId {
         if pox == &PoxId::stubbed() {
-            SortitionId(bhh.0.clone())
+            SortitionId(bhh.0)
         } else {
             let mut hasher = Sha512_256::new();
             hasher.update(bhh);
@@ -130,6 +125,10 @@ impl PoxId {
 
     pub fn len(&self) -> usize {
         self.0.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
     }
 
     pub fn bit_slice(&self, start: usize, len: usize) -> (Vec<u8>, u64) {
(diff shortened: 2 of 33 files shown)


stacks-common/src/types/chainstate.rs
```diff
@@ -34,18 +34,13 @@ pub type StacksPublicKey = Secp256k1PublicKey;
...
 
 /// Hash of a Trie node.  This is a SHA2-512/256.
+#[derive(Default)]
 pub struct TrieHash(pub [u8; 32]);
 impl_array_newtype!(TrieHash, u8, 32);
 impl_array_hexstring_fmt!(TrieHash);
 impl_byte_array_newtype!(TrieHash, u8, 32);
 impl_byte_array_serde!(TrieHash);
 
-impl Default for TrieHash {
```
jcnelson, line 43: I'm curious why Clippy deleted this?

stacks-common/src/util/hash.rs
```diff
@@ -533,8 +527,8 @@ where
     /// Get the Merkle root hash.
     /// will be all 0's if the tree is empty.
     pub fn root(&self) -> H {
-        if self.nodes.len() > 0 {
-            if self.nodes[self.nodes.len() - 1].len() > 0 {
+        if !self.nodes.is_empty() {
```
jcnelson, line 530: Is there a way to get Clippy to skip this transformation?  Semantically speaking, I find it easier to reason about a an array bounds-check in terms of the number of items it contains (i.e. the domain of valid indexes), instead of a container-generic `is_empty()` function.

stacks-common/src/util/pipe.rs
```diff
@@ -364,7 +353,7 @@ mod test {
             let expected_recv_buf = send_bytes.clone();
 
             pipe_read.set_nonblocking(true);
-            pipe_write.write(&send_bytes[..]).unwrap();
```
jcnelson, line 367: The use of `write` in this test module is deliberate.

stacks-common/src/util/secp256k1.rs
```diff
@@ -551,7 +560,7 @@ mod tests {
                 (_, _) => {
                     // either got a key when we didn't expect one, or didn't get a key when we did
                     // expect one.
-                    assert!(false);
+                    panic!("Unexpected result");
```
jcnelson, line 563: Can this be filled in with useful information?

stacks-common/src/util/secp256k1.rs
```diff
@@ -621,18 +630,16 @@ mod tests {
                 (Err(e1), Err(e2)) => assert_eq!(e1, e2),
                 (Err(e1), _) => {
                     test_debug!("Failed to verify signature: {}", e1);
-                    eprintln!(
```
jcnelson, line 624: The reason this is `eprintln!()` is because we install a custom panic handler that, when run, can cause the message within the `panic!(..)` to get lost.  These changes will need to be reverted.

stacks-common/src/util/vrf.rs
```diff
@@ -290,12 +296,12 @@ impl VRFProof {
         let c_bytes = c.reduce().to_bytes();
 
         // upper 16 bytes of c must be 0's
-        for i in 16..32 {
-            if c_bytes[i] != 0 {
```
jcnelson, line 294: I think this loses semantic information.  The algorithm specifically calls for checking the upper 16 bytes of  `c`, and the original `for`-loop makes this clearer.

stacks-common/src/util/vrf.rs
```diff
@@ -533,16 +529,16 @@ impl VRF {
         let (Y_point, x_scalar, trunc_hash) = VRF::expand_privkey(secret);
         let H_point = VRF::hash_to_curve(&Y_point, alpha);
 
-        let Gamma_point = &x_scalar * &H_point;
+        let Gamma_point = x_scalar * H_point;
```
jcnelson, line 532: We'll want to check the code, but the reason these are `&`-referenced might be because `ed25519-dalek` implements `Deref` for these variables, and we're really interested in getting the inner type's behavior.  I'm leery of changing this module to begin with.

stacks-common/src/types/chainstate.rs
jferrant: It gets it for free using the above `#[derive(Default)]`. Clippy recognized that it could easily be derived and suggested using the keyword instead.

stacks-common/src/util/secp256k1.rs
jferrant: Ah good to know. I will revert.

stacks-common/src/util/vrf.rs
jferrant: Certainly not a hill I will die on, but I think its best to avoid index accesses if possible and generally find it easier to reason about bytes in an array using the "byte in bytes.iter()" syntax. However, this is not my codebase and I will revert as desired :D

stacks-common/src/util/vrf.rs
jferrant: Checking these types, it does not implement Deref for these variables. This most likely is why clippy reports a needless ref

stacks-common/src/util/pipe.rs
jferrant: ah kk will update :) seems strange in a test to only want to write up to some number of bytes and not ensure that number of bytes and panic otherwise.

stacks-common/src/util/hash.rs
jferrant: There definitely is a way. But this is one that I would probably push back on as is_empty is exactly this check (len() > 0) and it makes sense to me to check if its empty especially given the comment "will be all 0's if the tree is empty".

kantai: I agree with @jferrant -- in most cases, the goal is to check that a container is empty and handle it as such. In this case, for example, the code is explicitly handling the situation in which the tree is empty (hence, the invocations of `H::empty()`. Thinking about this in terms of bounds checks is confusing: this code isn't really concerned with "array bounds checks", it should be concerned with checking the whole container for emptiness. Further, usually boundary checking suggests that something non-rustic is happening anyways (as in this function). This would be better represented as:

```rust
    pub fn root(&self) -> H {
        let Some(top_level_nodes) = self.nodes.last() else { return H::empty() };
        let Some(root) = top_level_nodes.first() else { return H::empty() };
        root.clone()
    }
```

stacks-common/src/util/secp256k1.rs
kantai: Is that true in tests? I don't recall missing panic messages in test executions. I suspect that these were added when creating the fixtures -- the asserts were probably commented out so that a single execution would always test every fixture even if one failed.

jcnelson: We have a custom panic hook to (1) force the node to print a backtrace (this doesn't always happen by default, it turns out), and (2) to force a core dump on unix-like systems by sending `SIGQUIT` to itself on panic.  However, for reasons I haven't figured out yet, the custom panic hook stops the panic message from being written.

stacks-common/src/util/vrf.rs
kantai: I agree about avoiding looping with an index access in general, but I also think there's a way to get more semantic information from the iteration:

```rust
        for c_byte in c_bytes[16..32].iter() {
            if *c_byte != 0 {
                return false;
            }
        }
```

jcnelson: Thanks :pray:

kantai (approved):
This LGTM, I suggested a couple of alternative "fixes" for clippy's suggestions. For the testing of the write() in pipe.rs:356, you should add a clippy ignore for that function.

stacks-common/src/util/secp256k1.rs
kantai: Is that custom hook active in unit tests? When I altered one of the fixtures here and run `cargo test util::secp256k1` in `stacks-common`, I see:

```
$ cargo test util::secp256k1
    Blocking waiting for file lock on build directory
   Compiling stacks-common v0.0.1 (/Users/aaron/devel/stacks-blockchain/stacks-common)
    Finished test [optimized + debuginfo] target(s) in 7.63s
warning: the following packages contain code that will be rejected by a future version of Rust: rstest v0.11.0
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running unittests src/libcommon.rs (/Users/aaron/devel/stacks-blockchain/target/debug/deps/stacks_common-50c7257d9fc9067d)

running 5 tests
test util::secp256k1::tests::test_verify_benchmark_roundtrip ... ignored
test util::secp256k1::tests::test_parse_serialize_compressed ... ok
test util::secp256k1::tests::test_parse_serialize ... ok
test util::secp256k1::tests::sk_from_seed ... ok
test util::secp256k1::tests::test_verify ... FAILED

failures:

---- util::secp256k1::tests::test_verify stdout ----
thread 'util::secp256k1::tests::test_verify' panicked at 'failed fixture (verification: Ok(false)): VerifyFixture {
    public_key: "0385f2e2867524289d6047d0d9c5e764c5d413729fc32291ad2c353fbc396a4219",
    data: "b84d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9",
    signature: "00354445a1dc98a1bd27984dbe69979a5cd77886b4d9134af5c40e634d96e1cb445b97de5b632582d31704f86706a780886e6e381bfed65228267358262d203fe6",
    result: Ok(
        true,
    ),
}', stacks-common/./src/util/secp256k1.rs:639:21


failures:
    util::secp256k1::tests::test_verify

test result: FAILED. 3 passed; 1 failed; 1 ignored; 0 measured; 105 filtered out; finished in 0.00s

error: test failed, to rerun pass `--lib`
```

jferrant: ah so revert in integration tests? Leave in unit? Or revert in all cases? Cause there seems to be inconsitency throughout the code base as i already does panic!("MESSAGE") in a number of places but eprintln and unreachable or assert in others.

jferrant:
> This LGTM, I suggested a couple of alternative "fixes" for clippy's suggestions. For the testing of the write() in pipe.rs:356, you should add a clippy ignore for that function.

Reverted the changes as requested. Just waiting on confirmation concerning logging changes.

stacks-common/src/util/vrf.rs
jferrant: Rethinking this because I was talking with Joey and he pointed out that this will introduce a copy that could have non neligable perofmrance impacts. I am going to silence this warning out of an abundance of caution as I have not bench marked this change.

jferrant: Joey reminded me that these structs may be actually quite large and that removing these references may have non negligable performance impacts (for the mathy bits). Since I have not benchmarked this change, i have reverted and silenced the warning with a clippy allow out of an abundance of caution.

stacks-common/src/util/secp256k1.rs
jcnelson: It should revert in all cases.  Panics generally need to be accompanied by an `eprintln()`.

jferrant: Reverted `panic!(MESSAGE)` changes back to `assert!(false, MESSAGE)` and I silenced the warnings. I did not update any existing `panic!(MESSAGE)`'s though.

stacks-common/src/deps_common/httparse/mod.rs
```diff
@@ -645,7 +647,7 @@ fn parse_reason<'a>(bytes: &mut Bytes<'a>) -> Result<&'a str> {
         } else if b == b'\n' {
             let res = str::from_utf8(bytes.slice_skip(1)).map_err(|_e| Error::Status)?;
             return Ok(Status::Complete(res));
-        } else if !((b >= 0x20 && b <= 0x7E) || b == b'\t') {
+        } else if !((0x20..=0x7E).contains(&b) || b == b'\t') {
```
jcnelson, line 650: Can you check that this doesn't do an _O(n)_ search through `0x20..=0x7E`?

jferrant: This is just a  check if a byte b is within the RANGE of ASCII characters from 0x20 to 0x7E. This operation is O(1).

xoloki (approved):
LGTM

PR #3950, opened 2023-09-19
Address clippy complaints in pox-locking repo

jferrant:
Closes https://github.com/stacks-network/stacks-blockchain/issues/3937

To run clippy ONLY in pox-locking:
`cargo clippy -p pox-locking --no-deps --tests --all-features `

### DIFF
--- a/pox-locking/src/pox_2.rs
+++ b/pox-locking/src/pox_2.rs
@@ -272,6 +272,7 @@ fn pox_lock_v2(
 }
 
 /// Handle responses from stack-stx and delegate-stack-stx -- functions that *lock up* STX
+#[allow(clippy::needless_return)]
 fn handle_stack_lockup_pox_v2(
     global_context: &mut GlobalContext,
     function_name: &str,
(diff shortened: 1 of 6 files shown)


pox-locking/src/pox_2.rs
```diff
@@ -370,7 +368,7 @@ fn handle_stack_lockup_extension_pox_v2(
         }
     };
 
-    match pox_lock_extend_v2(&mut global_context.database, &stacker, unlock_height as u64) {
+    match pox_lock_extend_v2(&mut global_context.database, &stacker, unlock_height) {
```
jcnelson, line 371: I'm not sure how I feel about dropping the `return` statements in these files.  I know it's what clippy wants, but because these return values are present in the middle of the code body, it's easier on the reader to see an explicit `return` to know that this is where flow-control stops.

jferrant: I can certainly silence the warnings instead. Let me go through and find ones that happen in what appears to be middle of code paths/where its non obvious and revert.

kantai (approved):

jcnelson (approved):

Issue #3943, opened 2023-09-18
`block-height` in Nakamoto


obycode:
This question arose on Discord:
> What happens to BNS name renewals with the Nakamoto upgrade? Will the dramatic increase in block time mean that BNS renewals will be due much sooner?

To clarify that a bit more, in Nakamoto, the block height will be incrementing on the order of seconds, rather than 10 minutes. This would cause BNS, and potentially other software that uses the `block-height` to have unexpected behavior (e.g. much sooner expiration than expected).

One proposal is to make `block-height` in Clarity 1 or Clarity 2 contracts continue to increment at a similar pace by incrementing on tenure changes, instead of at each new block. In Nakamoto, the tenure changes once per Bitcoin block, so it would be similar to the current block-height pace. In Clarity 3, `block-height`would still indicate the height.

Two problems with this proposal are:
1. It could be confusing for `block-height` to have different meanings in different versions of Clarity
2. Functions that take in a block height, such as `at-block` or `get-block-info?` would not work as expected if passed a tenure height instead of a block height.

jcnelson:
> It could be confusing for block-height to have different meanings in different versions of Clarity

We could create a different variable in Clarity 3 called `tenure-height`, which captures the same data that `block-height` does today.

> Functions that take in a block height, such as at-block or get-block-info? would not work as expected if passed a tenure height instead of a block height.

`at-block` takes a block hash, so I don't think that's affected.  `get-block-info?` would still need to take a Nakamoto block height, because it queries information about that specific block (not the tenure).  Perhaps there ought to be a separate `get-tenure-info?` function that indicates tenure-specific information?

Specific to BNS, what probably ought to happen is we should create a "wrapper" contract around BNS for Nakamoto.  The new BNS can use `tenure-height` to track name expirations and timeouts for newly-registered names in the Nakamoto epoch, and would return old name information written in the Stacks 2.x epoch via `at-block` for everything else.

This would require a PM to coordinate work on the contract and all BNS-aware wallets, though, so they'd know to migrate over to the new BNS contract right at the epoch boundary.  Moreover, we'd want to do some testing on the Atlas system to verify that it can track both the old and new BNS contracts' transactions' attachments (I'm pretty sure it can, but we haven't done this in prod yet).

jcnelson:
[...]

PR #3954, opened 2023-09-20
Add epoch 3.0

cylewitruk:
This PR aims to add the _boilerplate_ code around Epoch 3.0 and Clarity 3.0 so that teams can start building feature gating as needed based on these.

I've made some assumptions, and I've added some TODO's and even `todo!()`'s in areas where it was either unclear, or obvious that something would need to be done (e.g. in chainstate where I know there's big changes coming).

If we can get this PR in an "OK" state for `next`, and make sure we have Github issues for each `todo` (ideally that we even add the issue number in a comment next to the `todo`, give me the issue number and I can do this), then I hope the chances of forgetting about any of those will be slim :)

So, please have a look through the `todo`s and comment those which you can/will take and a link to the Github issue to track it. If I've missed any, please also comment where I've missed (or if you would like me to _add_ a `todo` somewhere for you) and I'll add them :)

Keeping this as a draft for now.


cylewitruk:
After some thought, I wonder if we shouldn't go for Clarity 2.1 instead of 3 here? From a _language_ perspective, there isn't all that much that's changing (with the Wasm runtime), that change will mostly be an implementation detail.
Thoughts?

obycode:
You're right that the runtime change should be an implementation detail, and can be epoch-gated rather than clarity version gated. We have discussed before that we'd like to push any clarity changes out after Nakamoto, unless absolutely necessary, just to avoid adding more complexity into an already very large change. So I'd say, if we can get away with it, we don't bump the clarity version at all yet.

cylewitruk:
> You're right that the runtime change should be an implementation detail, and can be epoch-gated rather than clarity version gated. We have discussed before that we'd like to push any clarity changes out after Nakamoto, unless absolutely necessary, just to avoid adding more complexity into an already very large change. So I'd say, if we can get away with it, we don't bump the clarity version at all yet.

Won't we have to bump the version due to changes in `block-height` and addition of `tenure-height` ?

obycode:
> Won't we have to bump the version due to changes in `block-height` and addition of `tenure-height` ?

I guess that's still an open question - #3943

clarity/src/vm/types/signatures.rs
```diff
@@ -1059,7 +1062,9 @@ impl TypeSignature {
             StacksEpochId::Epoch21
             | StacksEpochId::Epoch22
             | StacksEpochId::Epoch23
-            | StacksEpochId::Epoch24 => Self::least_supertype_v2_1(a, b),
+            | StacksEpochId::Epoch24
+            // TODO: afaik we're not changing anything that would affect this behavior?
```
kantai, line 1066: Yep, that's correct.

stacks-common/src/types/mod.rs
```diff
@@ -92,6 +93,7 @@ impl StacksEpochId {
             | StacksEpochId::Epoch22
             | StacksEpochId::Epoch23 => false,
             StacksEpochId::Epoch24 => true,
+            StacksEpochId::Epoch30 => true, //TODO: Is this correct, or should it only be the latest epoch which is true?
```
kantai, line 96: That's correct, 2.4 and 3.0 should both value sanitize.

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -3059,7 +3059,14 @@ impl SortitionDB {
                     || version == "5"
                     || version == "6"
                     || version == "7"
-                    // TODO: This should move to Epoch 30 once it is added
+            }
+            // TODO: Please double-check this
```
kantai, line 3063: Yep, this is fine.

stackslib/src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -774,6 +775,7 @@ impl LeaderBlockCommitOp {
     ) -> Result<SortitionId, op_error> {
         let tx_tip = tx.context.chain_tip.clone();
         let intended_sortition = match epoch_id {
+            StacksEpochId::Epoch30 => todo!(), // Need input here
```
kantai, line 778: This should have the same behavior as Epochs 2.1-2.4.

stackslib/src/chainstate/coordinator/mod.rs
```diff
@@ -292,6 +292,7 @@ impl RewardSetProvider for OnChainRewardSetProvider {
                     return Ok(RewardSet::empty());
                 }
             }
+            StacksEpochId::Epoch30 => todo!(), // Need input here
```
kantai, line 295: This should have the same behavior as Epoch24

stackslib/src/chainstate/coordinator/mod.rs
```diff
@@ -3052,6 +3053,7 @@ impl<
                                     return Ok(Some(pox_anchor));
                                 }
                             }
+                            StacksEpochId::Epoch30 => todo!(), // Need input here
```
kantai, line 3056: This should have the same behavior as Epochs 2.1 and 2.4 (for now: this will likely be changed in nakamoto, but in this PR, it should just share that behavior with 2.1-2.4).

stackslib/src/chainstate/stacks/db/blocks.rs
```diff
@@ -5291,6 +5335,7 @@ impl StacksChainState {
                     cur_epoch.start_height,
                 )
             }
+            StacksEpochId::Epoch30 => todo!(), // Need input here
```
kantai, line 5338: This should share the behavior of 2.1-2.4.

stackslib/src/chainstate/stacks/db/blocks.rs
```diff
@@ -5371,6 +5416,7 @@ impl StacksChainState {
                     pox_reward_cycle,
                     pox_start_cycle_info,
                 ),
+                StacksEpochId::Epoch30 => todo!(), // Need input here
```
kantai, line 5419: This should share the behavior of 2.4.

stackslib/src/chainstate/stacks/db/mod.rs
```diff
@@ -225,6 +225,7 @@ impl DBConfig {
             StacksEpochId::Epoch22 => self.version == "3" || self.version == "4",
             StacksEpochId::Epoch23 => self.version == "3" || self.version == "4",
             StacksEpochId::Epoch24 => self.version == "3" || self.version == "4",
+            StacksEpochId::Epoch30 => todo!(), // Need input from Jude here
```
kantai, line 228: This should be the same as 2.1-2.4 for now.

stackslib/src/clarity_vm/clarity.rs
```diff
@@ -1276,6 +1276,10 @@ impl<'a, 'b> ClarityBlockConnection<'a, 'b> {
         })
     }
 
+    pub fn initialize_epoch_3_0(&mut self) -> Result<Vec<StacksTransactionReceipt>, Error> {
+        todo!()
```
kantai, line 1280: This should just be `Ok(vec![])` for now: once there's actual changes necessary for epoch initialization, they can be added.

kantai (commented):
Thanks for this! This looks good to me, left a couple of comments.

clarity/src/vm/functions/mod.rs
```diff
@@ -61,6 +61,9 @@ macro_rules! switch_on_global_epoch {
                 StacksEpochId::Epoch23 => $Epoch205Version(args, env, context),
                 // Note: We reuse 2.05 for 2.4.
                 StacksEpochId::Epoch24 => $Epoch205Version(args, env, context),
+                // TODO: Is this correct?
+                // Note: We reuse 2.05 for 3.0.
+                StacksEpochId::Epoch30 => $Epoch205Version(args, env, context),
```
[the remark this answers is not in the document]
cylewitruk: Removed comment and resolving

clarity/src/vm/costs/mod.rs
```diff
@@ -723,6 +724,7 @@ impl LimitedCostTracker {
             | StacksEpochId::Epoch22
             | StacksEpochId::Epoch23
             | StacksEpochId::Epoch24 => COSTS_3_NAME.to_string(),
+            StacksEpochId::Epoch30 => COSTS_4_NAME.to_string(),
```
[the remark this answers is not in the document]
cylewitruk: Ok, removed COSTS_4_NAME

clarity/src/vm/types/signatures.rs
```diff
@@ -534,7 +534,10 @@ impl TypeSignature {
             StacksEpochId::Epoch21
             | StacksEpochId::Epoch22
             | StacksEpochId::Epoch23
-            | StacksEpochId::Epoch24 => self.admits_type_v2_1(other),
+            | StacksEpochId::Epoch24
+            // TODO: Afaik we're not making any changes which should change this behavior?
+            // Maybe add a "Note for future epochs, Epochs >= 2.1 should use `admits_type_v2_1` function."
+            | StacksEpochId::Epoch30 => self.admits_type_v2_1(other),
```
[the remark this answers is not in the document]
cylewitruk: Removed todo and added note

clarity/src/vm/types/signatures.rs
cylewitruk: Ok, comment removed

stacks-common/src/types/mod.rs
cylewitruk: Okay, comment removed

stackslib/src/chainstate/burn/db/sortdb.rs
cylewitruk: Ok, comment removed

stackslib/src/chainstate/burn/operations/leader_block_commit.rs
cylewitruk: Okay, added 3.0 to their match arm.

stackslib/src/chainstate/coordinator/mod.rs
cylewitruk: Okay, added 3.0 to the same match arm as 2.4

stackslib/src/chainstate/coordinator/mod.rs
cylewitruk: Okay, moved into their match arm

stackslib/src/chainstate/stacks/db/blocks.rs
cylewitruk: Okay, moved into their match arm

stackslib/src/chainstate/stacks/db/blocks.rs
cylewitruk: Okay, moved 3.0 into 2.4's match arm.

stackslib/src/chainstate/stacks/db/mod.rs
cylewitruk: Oki, copied

stackslib/src/clarity_vm/clarity.rs
cylewitruk: Okay, updated!

jcnelson:
`next` now has both `Epoch25` and `Epoch30`.  Can I close this?

cylewitruk:
> `next` now has both `Epoch25` and `Epoch30`.  Can I close this?

Yup 🙂

jcnelson:
Thanks for understanding :pray:

PR #3974, opened 2023-09-21
Feat: Add `TenureChange` transaction type

### LINKED ISSUE #3944: Nakamoto: New `TenureChange` Stacks transaction
In Nakamoto, each bitcoin block may perform a sortition -- much like the Stacks 2.0 protocol performs sortition. However, because the bitcoin commitment no longer contains the current stacks tip (see: #3892), the actual handoff from one miner to the next requires a `TenureChange` stacks transaction.

This is a new `TransactionPayload` variant which indicates that the prior tenure has ended (and what the last Nakamoto block of that tenure was) and that the next tenure should begin.

It should contain:

* The StacksBlockId of the last block from the previous tenure
* The number of blocks produced in the previous tenure
* A flag to indicate which of the following triggered the tenure change
      * A valid winning block-commit
      * No winning block-commits
      * A “null miner” won the block-commit (see the MEV solution below)
* The ECDSA public key hash of the current tenure
* A Schnorr signature from at least 70% of the Stackers
* A bitmap of which Stackers signed

Implementing this transaction payload for mockamoto involves:

1. adding the new transaction variant
2. adding a handler in `db/transactions.rs` which should verify that the current block is a Nakamoto block and that the `StacksBlockId` of the last block from the previous tenure is the current block's parent
3. adding a step in `NakamotoChainState` that checks for the presence of a `TenureChange` transaction, confirming that the current block was signed by the appropriate hash.

jbencin:
### Description

Add a new Stacks transaction type, `TenureChange`, that Stackers will use to control mining tenure in Nakamoto release.

This PR adds the enum variant `TransactionPayload::TenureChange` and establishes the wire format for a tenure change. It does *not* do much in the way of checking whether the trenure change is valid. That will come in a separate PR.

### Applicable issues
- #3944

### Checklist
- [x] Add `TenureChange` variant to `TransactionPayload` enum
- [ ]  In `db/transactions.rs`,  verify that the current block is a Nakamoto block and that the `StacksBlockId` of the last block from the previous tenure is the current block's parent 
- [ ] Add step in `NakamotoChainState` that checks for the presence of a `TenureChange` transaction, confirming that the current block was signed by the appropriate hash
- [ ] Changelog is updated


stackslib/src/chainstate/stacks/mod.rs
```diff
@@ -621,13 +619,57 @@ impl_array_hexstring_fmt!(TokenTransferMemo);
...
+    /// The StacksBlockId of the last block from the previous tenure
+    pub previous_tenure_end: StacksBlockId,
+    /// The number of blocks produced in the previous tenure
+    pub previous_tenure_blocks: u16,
+    /// A flag to indicate which of the following triggered the tenure change
+    pub cause: TenureChangeCause,
+    /// The ECDSA public key hash of the current tenure
+    pub pubkey_hash: Hash160,
+    /// A Schnorr signature from at least 70% of the Stackers
+    //pub signature: ???
```
jcnelson, line 660: CC @xoloki

xoloki: A taproot schnorr signature is 64 bytes, using the x-only curve point format.  A regular schnorr proof using secp256k1 would be 65 bytes, with a 33 byte compressed curve point, or 96 bytes with an uncompressed curve point.

xoloki: This is the standard `R,s` schnorr proof where `R` is a curve point and `s` is a scalar.

stackslib/src/chainstate/stacks/transaction.rs
```diff
@@ -138,52 +138,89 @@ fn ClarityVersion_consensus_deserialize<R: Read>(
...
 impl StacksMessageCodec for TransactionPayload {
     fn consensus_serialize<W: Write>(&self, fd: &mut W) -> Result<(), codec_error> {
-        match *self {
-            TransactionPayload::TokenTransfer(ref address, ref amount, ref memo) => {
+        match self {
+            TransactionPayload::TokenTransfer(address, amount, memo) => {
                 write_next(fd, &(TransactionPayloadID::TokenTransfer as u8))?;
                 write_next(fd, address)?;
                 write_next(fd, amount)?;
-                write_next(fd, memo)?;
```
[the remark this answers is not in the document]
jbencin: I wouldn't consider these additional exit points, those are created with `return`. The only exit point here is when the output of the `match` expression is returned (unless you count each `?`, in which case there are now fewer)

I can change it back if you want, it's not that important, but IMO it's cleaner and less code to return `Result` types directly when possible, like:

```rust
fn a_function() -> Result<(), Error> {
    // ...
    another_function()
}
```

instead of

```rust
fn a_function() -> Result<(), Error> {
    // ...
    another_function()?;
    Ok(())
}
```

stackslib/src/chainstate/stacks/mod.rs
xoloki: No reason to allow different formats, I'm just letting y'all know what the options are.  If we're using `wsts` to make a distributed threshold signature, then we get either the 65 or 64 byte formats.  If we're using a libsecp256k1 wrapper for a non distributed threshold signature, then we'll only get the 64 byte format.

xoloki: Which we should use depends on what tradeoffs we care about.  If we're going to use `wsts` to make a distributed signature, and also verify it, then the 65 byte format is the cleanest, since it doesn't require jumping through all the hoops necessary for the 64 byte format.  If we care about saving the extra byte, or plan on using a `libsecp256k1` wrapper to verify, then we need to use the 64 byte format.

jcnelson: I'm in favor of the 65 byte format in `wsts`.

xoloki: Agreed, I think that's the best choice for block signing for `Nakamoto`.  `sBTC` will want to use the 64 byte taproot format, since they need to sign `BTC` tx.

jbencin: So a signature in `TenureChange` should consist of both a point and scalar from `wsts`, like this?

```rust
struct SchnorrThresholdSignature {
    pub point: wsts::Point,
    pub scalar: wsts::Scalar,
}
```

stackslib/src/chainstate/stacks/transaction.rs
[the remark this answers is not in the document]
jcnelson: The issue here is that each arm of the `match` expression is now an exit point, so reasoning about where execution returned is a bit harder.  Before, there was only a single exit point.

stackslib/src/chainstate/stacks/mod.rs
xoloki: Yes, that works perfectly.  If you want a byte representation, you can call

`point.compress().as_bytes()`
`scalar.to_bytes()`

xoloki: There's also a `wsts::common::Signature` struct that you can use instead of defining your own.

jbencin: Thanks, that's exactly what I need

stackslib/src/chainstate/stacks/transaction.rs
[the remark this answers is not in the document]
jbencin: No problem, I can change it back

stackslib/src/chainstate/stacks/db/transactions.rs
```diff
@@ -1334,6 +1351,21 @@ impl StacksChainState {
...
+                // post-conditions are not allowed for this variant, since they're non-sensical.
+                // Their presence in this variant makes the transaction invalid.
+                if tx.post_conditions.len() > 0 {
+                    let msg = format!("Invalid Stacks transaction: TenureChange transactions do not support post-conditions");
+                    warn!("{msg}");
+
+                    return Err(Error::InvalidStacksTransaction(msg, false));
+                }
+
+                // TODO: More checks before adding to block?
```
jcnelson, line 1364: A few:

* This must be the first block in the miner's tenure.  It cannot be in a subsequent block
* This `TenureChange` must be for `last_tenure + 1`.  No `TenureChange`s can be skipped.
* It must point to the last active miner's block-commit's `StacksBlockId`
* It must count up the number of blocks since the last block with a `TenureChange` (it may be zero; a block can have multiple `TenureChange` transactions per the design doc)
* It must be signed by the current Stacker set (this could be checked on `consensus_deserialize`)
* It must precede any other kinds of transactions (this could be checked in `StacksBlock::consensus_deserialize`)

stackslib/src/chainstate/stacks/mod.rs
```diff
@@ -621,13 +619,74 @@ impl_array_hexstring_fmt!(TokenTransferMemo);
 impl_byte_array_newtype!(TokenTransferMemo, u8, 34);
 impl_byte_array_serde!(TokenTransferMemo);
 
+/// Cause of change in mining tenure
+/// Depending on cause, tenure can be ended or extended
```
jcnelson, line 623: There is supposed to be a separate `TenureExtend` transaction for this.

stackslib/src/chainstate/stacks/mod.rs
```diff
@@ -621,13 +619,74 @@ impl_array_hexstring_fmt!(TokenTransferMemo);
...
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct SchnorrThresholdSignature {
+    //pub point: wsts::Point,
+    //pub scalar: wsts::Scalar,
+}
+
+/// Reasons why a `TenureChange` transaction can be de
+pub enum TenureChangeError {
+    SignatureInvalid,
+    /// Not signed by required threshold (>70%)
```
jcnelson, line 657: This is indistinguishable from `SignatureInvalid`.  If less than 70% of signers sign, then they cannot produce a valid signature.

stackslib/src/chainstate/stacks/transaction.rs
```diff
@@ -182,6 +234,10 @@ impl StacksMessageCodec for TransactionPayload {
                     }
                 }
             }
+            TransactionPayload::TenureChange(tc) => {
+                write_next(fd, &(TransactionPayloadID::TenureChange as u8))?;
+                tc.consensus_serialize(fd)?;
+            }
```
jcnelson, line 240: You're going to somehow need to epoch-gate `consensus_deserialize` below so that this new transaction variant does not even decode until epoch 3.0.  I think we're going to need a separate trait for this (see here: https://github.com/stacks-network/stacks-blockchain/pull/3710#issuecomment-1747137322)

jbencin: Yeah this makes sense. I'll check with Vlad to see if he's working on this for his PR

Cargo.lock
```diff
@@ -2031,7 +2484,16 @@ version = "0.2.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fb5a58c1855b4b6819d59012155603f0b22ad30cad752600aadfcb695265519a"
 dependencies = [
- "bitflags",
+ "bitflags 1.3.2",
+]
+
+[[package]]
```
jcnelson, line 2490: This is a package for RedoxOS, which is not even supported by the blockchain.  Why is this a dependency?

stackslib/src/chainstate/stacks/transaction.rs
```diff
@@ -138,20 +137,94 @@ fn ClarityVersion_consensus_deserialize<R: Read>(
...
+        fd.read_exact(&mut buf)
+            .map_err(crate::codec::Error::ReadError)?;
+        let z = Scalar::from(buf);
+
+        Ok(Self { R, z })
+    }
+}
+
+impl TenureChangePayload {
+    pub fn validate(&self) -> Result<(), TenureChangeError> {
```
jcnelson, line 187: This is going to require access to a _lot_ more context, such as access to the chainstate databases.  In fact, the code for this might instead belong in `NakamotoChainState`.

stackslib/src/chainstate/stacks/transaction.rs
jcnelson: Right now, there are at least two PRs in flight (Vlad's and Cyle's) that do this.  We should just bite the bullet and get one of them shipped, and then base this PR off of it.

jcnelson: Actually, @kantai has already done this in https://github.com/stacks-network/stacks-blockchain/pull/3992.  Can you rebase off of this?

Cargo.lock
jbencin: I added `wsts` and `p256k1` as deps under `./stackslib` but I'm not sure how these got here. I will investigate...

stackslib/src/chainstate/stacks/transaction.rs
fess-v: Hey @jbencin @jcnelson ! Can't find in #3992 the logic for `DeserializeWithEpoch` traits for StacksTransaction and StacksBlock, I'm currently working on it and will finish soon, do you think I should wait for that PR to implement this logic and drop mine?

kantai: No -- #3992 isn't going to implement that logic -- so you should continue with your work.

fess-v: Got it, thanks @kantai

Cargo.lock
jbencin: The `Cargo.lock` file on `master` also contains `redox_syscall` and `redox_users`. See [here](https://github.com/stacks-network/stacks-blockchain/blob/e53ba10e75dc26f3dfe92226df2862f1167f806b/Cargo.lock#L2102). Here's the full dependency chain:

```
slog-term -> term -> dirs-next -> dirs-sys-next -> redox_users -> redox_syscall
```

In this PR, whatever I added needs an *additional* version of `redox_syscall` to be specified. This adds the new dependency chain:
```
pk256k1 -> rustfmt-wrapper -> tempfile -> redox_syscall
```

IMO it doesn't matter because as far as I can tell it's never downloaded and built. I don't see it using `cargo tree` or anywhere under `./target`. I guess it's disabled by a feature flag but the dependency gets resolved and pinned anyway?

jcnelson: Huh, what an odd behavior on Cargo's part.

As long as it's not getting linked into the binary, then I guess this is fine. I wish Cargo would only list packages that actually get built in the `.lock` file.

stackslib/src/chainstate/stacks/transaction.rs
kantai: Just to close the loop on this PR: I don't think it's strictly necessary to epoch gate *this* deserialization. All that matters is whether or not the transaction is valid for inclusion in a block. During block processing, the transaction processor should just yield an invalid block error if this transaction payload is matched.

stackslib/src/chainstate/stacks/db/transactions.rs
kantai: I think this should live as a separate issue from this PR and let this PR function mostly just as a wire format implementation (https://github.com/stacks-network/stacks-blockchain/issues/4014)

stackslib/src/chainstate/stacks/transaction.rs
jcnelson: Yeah that's fine for now.  Furthermore, the block validation code reached from `accept_block()` can reject blocks that include them too early.  I'm already doing this in my coordinator branch.

stackslib/src/chainstate/stacks/db/transactions.rs
jcnelson: Agreed

jcnelson:
Before this comes out of draft, we need unit tests for the new functionality.

kantai:
> @jcnelson or @kantai, I'm trying to finish up this PR now, and I've got a couple minor questions regarding how to sign. Right now I have the following struct:
> 
> ```rust
> pub struct TenureChangePayload {
>     pub previous_tenure_end: StacksBlockId,
>     pub previous_tenure_blocks: u16,
>     pub cause: TenureChangeCause,
>     pub pubkey_hash: Hash160,
>     pub signature: ThresholdSignature,
>     pub signers: Vec<u8>,
> }
> ```
> 
> What I'm wondering is:
> 
> * Does the `signature` field need to cover all other fields, including `signers`?

Yes, it needs to cover `signers`.

> * How to do the signature? My guess is that it should be done while the struct is encoded with `StacksMessageCodec` and treat that as a `u8` slice. So then I'd need to pull out `signature` like this:

I would recommend just adding a function:

```rust
impl TenureChangePayload {
    /// Serialize the parts of this payload to `out` required for signing and validating
    /// stacker signatures.
    pub fn serialize_for_sign<W: Write>(&self, out: &mut W) -> Result<(), CodecError> {
          s.previous_tenure_end.consensus_serialize(out)?;
          ...
    }
}
```

jbencin:
> I would recommend just adding a function:

Cool, this works too. The only other change then I'll need is to move `signature` as the last field in the struct, so `serialize_for_sign()` can be called in `TenureChangePayload::consensus_serialize()`.

jbencin:
I've finished up with the structure and wire format for a `TenureChange` transaction. I'm limiting the scope of this PR and pushing most of the validation work off to a later one. With the change in scope this PR is ready for review.

There's one thing that needs to be fixed before merging: There are still 2 unit tests failing. I've been looking into this since yesterday but haven't made any progress:

```
thread 'chainstate::nakamoto::tests::nakamoto_advance_tip_simple' panicked at stackslib/src/chainstate/nakamoto/tests.rs:167:6:
called `Result::unwrap()` on an `Err` value: InvalidStacksBlock("Failed to load total tenure cost from parent. parent_stacks_block_id = 55c9861be5cff984a20ce6d99d4aa65941412889bdc665094136429b84f8c2ee")


failures:
    chainstate::nakamoto::tests::nakamoto_advance_tip_multiple
    chainstate::nakamoto::tests::nakamoto_advance_tip_simple

test result: FAILED. 1241 passed; 2 failed; 125 ignored; 0 measured; 0 filtered out; finished in 416.59s
```

Any ideas what in here might have broken these tests @kantai?

jcnelson:
I've scrapped those two tests in my coordinator PR and replaced them into the `TestPeer`-derived tests (but it's not committed yet).  Going forward, there's going to be a lot of things that these tests will need to do in order to produce validate-able Nakamoto blocks, which `TestPeer` already does for free.

kantai:
> Any ideas what in here might have broken these tests @kantai?

I'm not sure -- I know that I needed to fix those tests when I merged some of this work into my previous PR.

I'd recommend rebasing this PR off of `next` -- that might fix those tests for you. If not, as @jcnelson mentions, those tests are getting supplanted in the coordinator PR anyways, so it may be easier to simply disable them.

jbencin:
> so it may be easier to simply disable them.

Okay, I'm just adding `#[ignore]` to these tests for now

kantai:
Can you rebase this to `next`?

jbencin:
Rebased, force pushed, and changed target to `next`

kantai (approved):
LGTM!

jcnelson (approved):
A good chunk of this got merged to `next` as part of the coordinator PR.  If you rebase to `next` then the rest of it is fine with me.  Thanks!

jbencin:
There's one more thing I want to do with `TenureChange`: I want it to use the `wsts::common::Signature` struct so that we can use it's `verify()` method. The only problem is that `Signature` does not implement the necessary derives like `Debug` and `Serialize` (even though `Point` and `Scalar` do), so I can't put it in `TenureChange`

I'll make a PR to WSTS adding the `derive`s, and I'll make a separate PR for this to stacks-core

cc @xoloki

jbencin:
Hey @jcnelson I've resolved the conflicts with `next` and this is ready to merge. Would you like me to wait until you've merged #4074, so we don't risk breaking anything there?

PR #3982, opened 2023-09-22
Bugfixes for signer when testing CLI through mainnet develop node

jferrant:
This PR is ongoing, fixing bugs are required (found through testing stacker db on mainnet)
- [x] Fix spawn_running_signer to use endpoint not node_host for creating the event listeners.
- [x] Loop the run thread forever so that we always wait for messages
- [x] Do not unwrap the results of the stop() in main as we expect these to return None. Nothing to actually unwrap.
- [x] Fixed clap to use String instead of Vec<u8>. This may need to be updated to take a hex string explicitly and dehexify that string to a vec<u8>? EDIT: updated to use stacks_common's copy of b58 decoder

### DIFF
--- a/stacks-signer/src/cli.rs
+++ b/stacks-signer/src/cli.rs
@@ -94,6 +94,8 @@ pub struct PutChunkArgs {
     pub slot_version: u32,
     /// The data to upload
     #[arg(required = false, value_parser = parse_data)]
+    // Note this weirdness is due to https://github.com/clap-rs/clap/discussions/4695
+    // Need to specify the long name here due to invalid parsing in Clap which looks at the NAME rather than the TYPE which causes issues in how it handles Vec's.
     pub data: alloc::vec::Vec<u8>,
 }
 
@@ -105,6 +107,8 @@ pub struct SignArgs {
     pub config: PathBuf,
     /// The data to sign
     #[arg(required = false, value_parser = parse_data)]
+    // Note this weirdness is due to https://github.com/clap-rs/clap/discussions/4695
+    // Need to specify the long name here due to invalid parsing in Clap which looks at the NAME rather than the TYPE which causes issues in how it handles Vec's.
     pub data: alloc::vec::Vec<u8>,
 }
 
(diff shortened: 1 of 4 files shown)


jferrant:
> Is a busy loop the best way to keep the process alive? We might want to rethink that

I could add a "join" function to the libsigner but it would inevitably hang forever as the event receiver will not ever end unless you explicitly send a "stop" signal. Which seems a bit dangerous to implement this function as some external user may try to use it with the idea of that the thread will eventually join when it in theory should hang forever (unless an error occurs)

I can certainly do it though and add a very explicit comment to the function description.

stacks-signer/src/cli.rs
```diff
@@ -106,7 +106,7 @@ pub struct SignArgs {
     pub config: PathBuf,
     /// The data to sign
     #[arg(required = false, value_parser = parse_data)]
-    pub data: Vec<u8>,
+    pub data: String,
```
xoloki, line 109: If we're gonna take this arg in as a `String`, we need to assume that it's `b58` encoded, and turn it back into a `Vec<u8>` rather than passing the string bytes around.

xoloki (changes requested):

stacks-signer/src/cli.rs
jferrant: Yes you are def right. Forgot about needing to add this change.

xoloki (approved):
LGTM after getting rid of `BS58Data`

stacks-signer/src/cli.rs
```diff
@@ -95,7 +96,7 @@ pub struct PutChunkArgs {
     pub slot_version: u32,
     /// The data to upload
     #[arg(required = false, value_parser = parse_data)]
-    pub data: Vec<u8>,
+    pub data: alloc::vec::Vec<u8>,
```
jferrant, line 99: Note this weirdness is due to the following: 
https://github.com/clap-rs/clap/discussions/4695

jcnelson (approved):

PR #3986, opened 2023-10-03
Enable signer tracing

xoloki:
### Description
`wsts`, like many rust libraries, uses the `tracing` crate for logging.  This PR subscribes to the `tracing` logs using the default `RUST_LOG` environmental variable, which allows setting log levels on a per-module basis, in both the `stacks-signer` binary and the `signer::test_stackerdb_dkg` integration test.

### Applicable issues
- fixes #

### Checklist
- [ ] Changelog is updated

### DIFF
--- a/testnet/stacks-node/src/tests/signer.rs
+++ b/testnet/stacks-node/src/tests/signer.rs
@@ -24,6 +24,7 @@ use stacks_signer::{
     runloop::RunLoopCommand,
     utils::{build_signer_config_tomls, build_stackerdb_contract},
 };
+use tracing_subscriber::{fmt, prelude::*, EnvFilter};
 use wsts::{
     state_machine::{coordinator::Coordinator as FrostCoordinator, OperationResult},
     v2,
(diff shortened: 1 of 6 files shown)


jcnelson (approved):

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -156,9 +157,15 @@ fn test_stackerdb_dkg() {
...
     }
+
+    tracing_subscriber::registry()
+        .with(fmt::layer())
+        .with(EnvFilter::from_default_env())
+        .init();
+
     // Generate Signer Data
     let num_signers: u32 = 16;
-    let num_keys: u32 = 4000;
```
jferrant, line 161: We don't want to keep the 4000 keys?

jferrant (approved):
Just the one question about the number of keys in the test, but other than that LGTM. :)

testnet/stacks-node/src/tests/signer.rs
xoloki: I lowered the number so the test would go faster, but happy to change it back.  It takes ~60 seconds to bootstrap the node, so I guess it doesn't matter if it takes another 290 to finish?

PR #3987, opened 2023-10-03
[Refactor] HTTP RPC System Refactor

jcnelson:
Alright folks, here it is:  **The Big HTTP RPC Refactor**.

This PR achieves two things:  it separates HTTP transport concerns from RPC business logic, and it refactors the RPC system so that each RPC method's business logic can live entirely in its own file.  You can find them all under `stackslib/src/net/api`.

The bulk of the delta here is composed:

* moving all of the code from `stackslib/src/net/http.rs` into the new module `stackslib/src/net/http`
* moving most of the code from `stackslib/src/net/rpc.rs` into `stackslib/src/net/api`
* moving most of the HTTP-specific code from `stackslib/src/net/mod.rs` into `stackslib/src/net/httpcore.rs`.

The first change sets the stage for fully removing HTTP transport concerns from `stackslib`, and putting them into their own top-level crate.  The code in this module does not depend on any Stacks-specific code or data.  The second change effectively isolates the RPC functions from the rest of the system, and exposes various Stacks node state under a new `StacksNodeState` struct (which can gain new members and functions as we need them).  The third change binds the newly-isolated HTTP transport code in `stackslib/src/net/http` to the Stacks blockchain, and implements a high-level HTTP state machine atop `net::connection::ProtocolFamily<P>` to be used by `net::server::ConversationHttp` to manage client connection lifecycles.

As you can see from the individual RPC methods, one only needs to implement the `net::http::HttpRequest`, `net::http::HttpResponse`, and `net::httpcore::RPCRequestHandler` traits for a given struct in order to implement a new RPC method.  Once this is done, it can be registered with the `StacksHttp` state machine with a call to `register_rpc_endpoint()` in `src/net/api/mod.rs`.

In a hopeless bid to keep this PR smaller than it could have been, I have opted to keep `net::http` and `net::connection.rs` within `stackslib` for now.  However, I fully expect to move these both into top-level crates, and possibly move the chunked transfer encoding and `PeerAddress`/`PeerHost` types out of `stacks_common` and into one of these crates.  But that is for another day.


stacks-common/src/types/net.rs
```diff
@@ -0,0 +1,389 @@
+// Copyright (C) 2013-2020 Blockstack PBC, a public benefit corporation
```
jcnelson, line 1: This is all just copy/pasted from `src/net/mod.rs`

stackslib/src/chainstate/stacks/db/mod.rs
```diff
@@ -202,6 +206,57 @@ pub struct StacksEpochReceipt {
     pub epoch_transition: bool,
 }
 
+/// Headers we serve over the network
```
jcnelson, line 209: This used to live in `src/net/mod.rs`

stackslib/src/core/mempool.rs
```diff
@@ -150,6 +152,167 @@ impl StacksMessageCodec for TxTag {
     }
 }
 
+define_u8_enum!(MemPoolSyncDataID {
```
jcnelson, line 155: This all used to live in `src/net/mod.rs`

stacks-common/src/util/mod.rs
```diff
@@ -66,8 +66,8 @@ pub enum HexError {
 impl fmt::Display for HexError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
-            HexError::BadLength(n) => write!(f, "bad length {} for sha256d hex string", n),
-            HexError::BadCharacter(c) => write!(f, "bad character {} in sha256d hex string", c),
```
jferrant, line 70: made more sense when this said "in hex string"

jcnelson: That's weird, my code had the opposite delta.  Will revert.

jcnelson:
@wileyj I'm not sure what to do about the `CI / Docker Debian (Source) / Build Image (pull_request)` error.  Do you have any insights?

wileyj:
> @wileyj I'm not sure what to do about the `CI / Docker Debian (Source) / Build Image (pull_request)` error. Do you have any insights?

per the call, i've seen errors compiling stackslib when `clang` isn't present. however, i don't think that's the reason here:
```
#11 338.0    Compiling stackslib v0.0.1 (/build/stackslib)
#11 349.1 error[E0599]: no method named `get_path` found for struct `StacksHttpRequest` in the current scope
#11 349.1    --> stackslib/src/monitoring/mod.rs:62:52
#11 349.1     |
#11 349.1 62  |     let timer = prometheus::new_rpc_call_timer(req.get_path());
#11 349.1     |                                                    ^^^^^^^^ method not found in `StacksHttpRequest`
#11 349.1     |
#11 349.1    ::: stackslib/src/net/httpcore.rs:452:1
#11 349.1     |
#11 349.1 452 | pub struct StacksHttpRequest {
#11 349.1     | ---------------------------- method `get_path` not found for this struct
#11 349.1 
#11 349.3    Compiling fixed-hash v0.8.0
#11 349.3 For more information about this error, try `rustc --explain E0599`.
#11 349.3 error: could not compile `stackslib` (lib) due to previous error
#11 349.3 warning: build failed, waiting for other jobs to finish...
```

stacks-common/src/util/mod.rs
jferrant: Is this a change you want to revert still?

stackslib/src/net/api/callreadonly.rs
```diff
@@ -0,0 +1,390 @@
...
+        let map_entry: CallReadOnlyResponse = parse_json(preamble, body)?;
+        Ok(HttpResponsePayload::try_from_json(map_entry)?)
+    }
+}
+
+impl StacksHttpRequest {
+    /// Make a new request to run a read-only function
+    pub fn new_callreadonlyfunction(
+        host: PeerHost,
+        contract_addr: StacksAddress,
```
jferrant, line 354: I think you could make a lot of these arguments references since they are only ever used as references within this function. (e.g. contract_addr, contract_name, function_name)

stackslib/src/net/api/callreadonly.rs
```diff
@@ -0,0 +1,390 @@
...
+                })
+                .expect("FATAL: failed to encode infallible data"),
+            ),
+        )
+        .expect("FATAL: failed to construct request from infallible data")
+    }
+}
+
+impl StacksHttpResponse {
+    pub fn decode_call_readonly_response(self) -> Result<CallReadOnlyResponse, NetError> {
```
jferrant, line 383: Should this take a reference to self? Or does it need to consume self?

stackslib/src/net/api/getaccount.rs
```diff
@@ -0,0 +1,270 @@
...
+impl RPCRequestHandler for RPCGetAccountRequestHandler {
+    /// Reset internal state
+    fn restart(&mut self) {
+        self.account = None;
+    }
+
+    /// Make the response
+    fn try_handle_request(
+        &mut self,
+        preamble: HttpRequestPreamble,
```
jferrant, line 136: These could also be references. `preamble` is only ever used as a reference it seems and I think `content` as well.

stackslib/src/net/api/callreadonly.rs
```diff
@@ -0,0 +1,390 @@
...
+        body: &[u8],
+    ) -> Result<HttpResponsePayload, Error> {
+        let map_entry: CallReadOnlyResponse = parse_json(preamble, body)?;
+        Ok(HttpResponsePayload::try_from_json(map_entry)?)
+    }
+}
+
+impl StacksHttpRequest {
+    /// Make a new request to run a read-only function
+    pub fn new_callreadonlyfunction(
```
jferrant, line 352: It will be nice to take advantage of this in my DKG pr! :D

stackslib/src/net/api/getaccount.rs
```diff
@@ -0,0 +1,270 @@
...
+    ) -> Result<HttpResponsePayload, Error> {
+        let account: AccountEntryResponse = parse_json(preamble, body)?;
+        Ok(HttpResponsePayload::try_from_json(account)?)
+    }
+}
+
+impl StacksHttpRequest {
+    /// Make a new request for an account
+    pub fn new_getaccount(
+        host: PeerHost,
```
jferrant, line 245: I think a lot of these can be references too.

stackslib/src/net/api/getaccount.rs
```diff
@@ -0,0 +1,270 @@
...
+                .for_tip(tip_req)
+                .query_arg("proof".into(), if with_proof { "1" } else { "0" }.into()),
+        )
+        .expect("FATAL: failed to construct request from infallible data")
+    }
+}
+
+impl StacksHttpResponse {
+    pub fn decode_account_entry_response(self) -> Result<AccountEntryResponse, NetError> {
+        let contents = self.get_http_payload_ok()?;
```
jferrant, line 264: I assume this is intentional. But does it need to consume self?

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+
+/// Trait that every HTTP round-trip request type must implement.
+pub trait RPCRequestHandler: HttpRequest + HttpResponse + RPCRequestHandlerClone {
+    /// Reset the RPC handler.  This clears any internal state this handler stored between calls to
+    /// `try_handle_request()`
+    fn restart(&mut self);
+    /// Instantiate the HTTP response headers and body from a request
+    fn try_handle_request(
+        &mut self,
+        request_preamble: HttpRequestPreamble,
```
jferrant, line 386: I should really comment on the trait...less of the functions... :P So I think you may be able to take references for the request_preamble and request_body....Though I didn't go through every single implementation...

stackslib/src/net/api/getattachmentsinv.rs
```diff
@@ -0,0 +1,269 @@
...
+            let msg = format!(
+                "Number of attachment inv pages is limited by {} per request",
+                MAX_ATTACHMENT_INV_PAGES_PER_REQUEST
+            );
+            warn!("{}", msg);
+            return StacksHttpResponse::new_error(&preamble, &HttpBadRequest::new(msg))
+                .try_into_contents()
+                .map_err(NetError::from);
+        }
+        if page_indexes.len() == 0 {
```
jferrant, line 174: Should generally do is_empty() instead of len() == 0. Clippy will complain :P

stackslib/src/net/api/getattachmentsinv.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// Make a new request for attachment inventory page
+    pub fn new_getattachmentsinv(
+        host: PeerHost,
+        index_block_hash: StacksBlockId,
+        page_indexes: HashSet<u32>,
+    ) -> StacksHttpRequest {
+        let page_list: Vec<String> = page_indexes.into_iter().map(|i| format!("{}", i)).collect();
+        StacksHttpRequest::new_for_peer(
+            host,
+            "GET".into(),
```
jferrant, line 249: We should use the self.verb() function I think instead of the hardcoded "GET" string for maintainability.

stackslib/src/net/api/getconstantval.rs
```diff
@@ -0,0 +1,242 @@
...
+        let constant_val: ConstantValResponse = parse_json(preamble, body)?;
+        Ok(HttpResponsePayload::try_from_json(constant_val)?)
+    }
+}
+
+impl StacksHttpRequest {
+    /// Make a new request for a constant val
+    pub fn new_getconstantval(
+        host: PeerHost,
+        contract_addr: StacksAddress,
```
jferrant, line 216: The `contract_*` function args can be refs too.

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+    /// Instantiate a request to a remote Stacks peer
+    /// `path` is just the request path.  Query arguments are added via `contents`.  Any query
+    /// component to `path` will be silently dropped.
+    ///
+    /// In Stacks, all requests must have a known content length.  If it cannot be calculated, then
+    /// this method will fail.
+    pub fn new_for_peer(
+        peerhost: PeerHost,
+        verb: String,
+        path: String,
```
jferrant, line 471: I wonder if this path should take something more like a URL (or equivalent) https://github.com/stacks-network/stacks-blockchain/pull/3947

It will make sure we don't inadvertantly have extra slashes where not wanted. Not by any means a blocking change. but might be worth thinking about.

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+                Ok((0, num_consumed)) => {
+                    trace!(
+                        "consume_data blocked on 0 decoded bytes ({} consumed)",
+                        num_consumed
+                    );
+                    blocked = true;
+                    (0, num_consumed)
+                }
+                Ok((num_read, num_consumed)) => (num_read, num_consumed),
+                Err(e) => match e.kind() {
```
jferrant, line 795: I think instead of doing a second match you can do 
```rust
    Ok((0, num_consumed)) => {..},
    Err(io::ErrorKind::WouldBlock) | Err(io::ErrorKind::TimedOut) => {
        trace!("consume_data blocked on read error");
        blocked = true;
        (0, 0)
    }
    _ => {
        return Err(NetError::ReadError(e));
    }
}
```

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+                warn!("Irrecoverable error when handling request"; "path" => %request_preamble.path_and_query_str, "error" => %e);
+                return Err(e);
+            }
+        };
+        Ok((response_preamble, response_contents))
+    }
+
+    #[cfg(test)]
+    pub fn num_pending(&self) -> usize {
+        if self.reply.is_some() {
```
jferrant, line 1124: If you wanted to remove the branches, you could change this to `u8::from(self.reply.is_some())` or `self.reply.is_some() as u8` since Rust now defines the memory representation for bool as a 1-byte value equal to either 0 or 1

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+        &mut self,
+        preamble: &HttpResponsePreamble,
+        fd: &mut R,
+    ) -> Result<(Option<(Vec<u8>, usize)>, usize), NetError> {
+        assert!(preamble.is_chunked());
+        assert!(self.reply.is_some());
+
+        if let Some(reply) = self.reply.as_mut() {
+            match reply.stream.consume_data(fd) {
+                Ok(res) => match res {
```
jferrant, line 1193: You can similarly collapse this double match into a single one

```rust
            match reply.stream.consume_data(fd) {
                Ok((None, sz)) => Ok((None, sz)),
                Ok(Some((byte_vec, bytes_total)), sz) => {
                        // done receiving
                        self.reply = None;
                        Ok((Some((byte_vec, bytes_total)), sz))
                 },
                Err(e) => {
                    // broken stream
                    self.reset();
                    Err(e)
                }
            }

File: stackslib/src/net/httpcore.rs [hunk 19]
```diff
@@ -0,0 +1,1625 @@
...
+                    Err(e)
+                }
+            }
+        } else {
+            unreachable!();
+        }
+    }
+
+    /// Calculate the search window for \r\n\r\n
+    fn body_start_search_window(&self, i: usize, buf: &[u8]) -> [u8; 4] {
```
jferrant, line 1213: Not sure this is better as its also confusing to me but you could simplify this code to 

```rust
fn body_start_search_window(&self, i: usize, buf: &[u8]) -> [u8; 4] {
    let mut window = [0; 4];
    for j in 0..4 {
        window[j] = if i < 4 - j {
            self.last_four_preamble_bytes[j]
        } else {
            buf[i - 4 + j]
        };
    }
    window
}
```

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+                    "Invalid HTTP message: did not get a Response preamble".to_string(),
+                ));
+            }
+        };
+
+        let mut message_bytes = &response_buf[message_offset..];
+
+        if is_chunked {
+            match http.stream_payload(&preamble, &mut message_bytes) {
+                Ok((Some((message, _)), _)) => Ok(message),
```
jferrant, line 1273: You could simplify this if you wanted
```rust
            match http.stream_payload(&preamble, &mut message_bytes)? {
                (Some((message, _)), _) => Ok(message),
                (None, _) => Err(NetError::UnderflowError(
                    "Not enough bytes to form a streamed HTTP response".to_string(),
                )),
            }
```

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+    }
+
+    /// how big is this message?  Might not know if we're dealing with chunked encoding.
+    fn payload_len(&mut self, preamble: &StacksHttpPreamble) -> Option<usize> {
+        match *preamble {
+            StacksHttpPreamble::Request(ref http_request_preamble) => {
+                Some(http_request_preamble.get_content_length() as usize)
+            }
+            StacksHttpPreamble::Response(ref http_response_preamble) => {
+                match http_response_preamble.content_length {
```
jferrant, line 1302: Can remove this match and just do a map:
`http_response_preamble.content_length.map(|len| len as usize)`

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+
+    /// Read the next HTTP preamble (be it a request or a response), and return the preamble and
+    /// the number of bytes consumed while reading it.
+    fn read_preamble(&mut self, buf: &[u8]) -> Result<(StacksHttpPreamble, usize), NetError> {
+        // does this contain end-of-headers marker, including the last four bytes of preamble we
+        // saw?
+        if self.body_start.is_none() {
+            for i in 0..=buf.len() {
+                let window = self.body_start_search_window(i, buf);
+                if window == [13, 10, 13, 10] {
```
jferrant, line 1318: This would be clearer as 
`if window == [b'\r', b'\n', b'\r', b'\n']`

stackslib/src/net/httpcore.rs
jferrant: Also, if preferred, you could simplify even further to 
```rust
            match reply.stream.consume_data(fd).map_err(|e| {
                // broken stream
                self.reset();
                Err(e)
            })? {
               (Some((byte_vec, bytes_total)), sz) => {
                        // done receiving
                        self.reply = None;
                        Ok((Some((byte_vec, bytes_total)), sz))
                 },
                (res) => Ok(res),
            }
```

There might be some better map() that could get rid of the match altogether but not sure off the top of my head.

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+    /// Returns Ok((None, number of bytes gotten in this call)) if there's more to decode.
+    pub fn consume_data<R: Read>(
+        &mut self,
+        preamble: &HttpResponsePreamble,
+        fd: &mut R,
+    ) -> Result<(Option<(Vec<u8>, usize)>, usize), NetError> {
+        assert!(preamble.is_chunked());
+        assert!(self.reply.is_some());
+
+        if let Some(reply) = self.reply.as_mut() {
```
jferrant, line 1191: Just a preference, but I rather be explicit with something like this (remove the unreachable branch)
```rust 
       let reply = self.reply.as_mut().ok_or_else(|| NetError::CustomError(
            "Empty reply. Nothing to consume".into(),
        ))?;
```

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+    fn stream_payload<R: Read>(
+        &mut self,
+        preamble: &StacksHttpPreamble,
+        fd: &mut R,
+    ) -> Result<(Option<(StacksHttpMessage, usize)>, usize), NetError> {
+        assert!(self.payload_len(preamble).is_none());
+        match preamble {
+            StacksHttpPreamble::Request(_) => {
+                // HTTP requests can't be chunk-encoded, so this should never be reached
+                unreachable!()
```
jferrant, line 1369: Similarly to the other comment, I would do an explicit error. However, I see unreachable is used a lot throughout so imagine this is the preference to prevent a plethora of error types?

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+            StacksHttpPreamble::Request(ref http_request_preamble) => {
+                // all requests have a known length
+                let len = http_request_preamble.get_content_length() as usize;
+                assert!(len <= buf.len(), "{} > {}", len, buf.len());
+
+                trace!("read http request payload of {} bytes", len);
+
+                match self.try_parse_request(http_request_preamble, &buf[0..len]) {
+                    Ok(data_request) => Ok((StacksHttpMessage::Request(data_request), len)),
+                    Err(e) => {
```
jferrant, line 1464: Simlarly to other error match comment, this double match can be reduced to a single match.

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+                match http.connect_http(
+                    network_state,
+                    network,
+                    data_url.clone(),
+                    addr.clone(),
+                    Some(request.clone()),
+                ) {
+                    Ok(event_id) => Ok(event_id),
+                    Err(NetError::AlreadyConnected(event_id, _)) => {
+                        match http.get_conversation_and_socket(event_id) {
```
jferrant, line 1587: Because this is a match on a singular match, you could conver this to an if let (Some(ref mut convo), Some(ref mut socket)) = BLAH if you want to. I think clippy might actually complain about this, but regardless...is just a syntactical preference.

jferrant:
First pass done. All my comments have to do with just readability/syntax. Didn't see anything functionally wrong, but I will do another pass on monday!

CAGS295:
> > @wileyj I'm not sure what to do about the `CI / Docker Debian (Source) / Build Image (pull_request)` error. Do you have any insights?
> 
> per the call, i've seen errors compiling stackslib when `clang` isn't present. however, i don't think that's the reason here:
> 
> ```
> #11 338.0    Compiling stackslib v0.0.1 (/build/stackslib)
> #11 349.1 error[E0599]: no method named `get_path` found for struct `StacksHttpRequest` in the current scope
> #11 349.1    --> stackslib/src/monitoring/mod.rs:62:52
> #11 349.1     |
> #11 349.1 62  |     let timer = prometheus::new_rpc_call_timer(req.get_path());
> #11 349.1     |                                                    ^^^^^^^^ method not found in `StacksHttpRequest`
> #11 349.1     |
> #11 349.1    ::: stackslib/src/net/httpcore.rs:452:1
> #11 349.1     |
> #11 349.1 452 | pub struct StacksHttpRequest {
> #11 349.1     | ---------------------------- method `get_path` not found for this struct
> #11 349.1 
> #11 349.3    Compiling fixed-hash v0.8.0
> #11 349.3 For more information about this error, try `rustc --explain E0599`.
> #11 349.3 error: could not compile `stackslib` (lib) due to previous error
> #11 349.3 warning: build failed, waiting for other jobs to finish...
> ```

`RequestType` was replaced by `StacksHttpRequest` in 0a97fac68725ffed43982bb109c8306d48b7d4fa.
`RequestType::get_path()` was deleted in 4e41c2311eb87dcd481a3ff5310efdb2530b6d08.

@jcnelson I assume you want something like:
`let timer = prometheus::new_rpc_call_timer(&decode_request_path(req.request_path())?.0);`

It compiles for me.

stackslib/src/net/httpcore.rs
jcnelson: The reason the `RPCRequestHandler::try_handle_request()` function implementation calls for a pass-by-value instead of a pass-by-reference is to avoid requiring the caller to perform extra allocations to get owned copies of parts of the request headers and body.  I don't think it hurts anything to keep it as pass-by-value, since the caller of `.try_handle_request()` does not need to make copies either.

stackslib/src/net/api/callreadonly.rs
jcnelson: I try to use pass-by-value so the callee does not need to perform any extra allocations just to construct a request.  The caller almost always needs to perform them, however, but it does so solely to build a request.  I'm not sure pass-by-reference gains anything here?

stackslib/src/net/api/getattachmentsinv.rs
jcnelson: We can't -- `StacksHttpRequest` does not have `self.verb()`.  The `RPCGetAttachmentsInvRequestHandler` struct does, as part of implementing `HttpRequestHandler` (and `RPCRequestHandler`).  However, this has to be an instance method so the correct function body can get called.  This is required in `StacksHttp::try_handle_request()`, which calls the `.verb()` and `.regex()` functions on a `Box<dyn RPCRequestHandler>` in order to dispatch to the correct implementations' methods.  If we didn't do this, then we'd need to somehow identify the trait's implementation type at runtime to handle the dispatch.

stackslib/src/net/httpcore.rs
jcnelson: We sanitize `path` in the inner call to `decode_request_path`, which extracts both the normalized path and the query string. The `path` here just captures everything that follows the host.

stackslib/src/net/httpcore.rs
jcnelson: Sure, will add comments.

stackslib/src/net/httpcore.rs
jcnelson: I can go back and convert these to an error.

stackslib/src/net/api/getattachmentsinv.rs
jcnelson: Fixed

stackslib/src/net/httpcore.rs
jcnelson: Fixed

stackslib/src/net/httpcore.rs
jcnelson: Fixed

stackslib/src/net/httpcore.rs
jcnelson: Fixed

stackslib/src/net/httpcore.rs
jcnelson: Fixed

stackslib/src/net/httpcore.rs
jcnelson: Fixed

stackslib/src/net/httpcore.rs
jcnelson: Fixed

stackslib/src/net/httpcore.rs
jcnelson: Fixed

stackslib/src/net/httpcore.rs
jcnelson: Fixed

stackslib/src/net/httpcore.rs
```diff
@@ -0,0 +1,1625 @@
...
+            StacksHttpPreamble::Response(ref http_response_preamble) => {
+                assert!(!http_response_preamble.is_chunked());
+
+                // message of known length
+                test_debug!("read http response payload of {} bytes", buf.len(),);
+
+                // sanity check -- if we're receiving a response, then we must have earlier issued
+                // a request. Thus, we must already know which response handler to use.
+                // Otherwise, someone sent us malformed data.
+                let handler_index = if let Some(i) = self.request_handler_index.as_ref() {
```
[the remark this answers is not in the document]
jcnelson: Fixed

stackslib/src/net/httpcore.rs
jcnelson: Fixed

xoloki (changes requested):
When I try to run `tests::signer::test_stackerdb_dkg`, it never completes.  I have no idea what's happening but the logs are spamming very fast.

Get this test running, then try it with `num_keys = 4000` (`testnet/stacks-node/src/tests/signer.rs:161`).  If DKG takes longer than ~90sec, or either of the sign operations takes longer than ~100sec, something is badly wrong with the HTTP layer and it's effectively breaking stackerDB under load.

jcnelson:
Thanks @xoloki!  The bug was that the new handler uploading StackerDB chunks was not forwarding the chunk data to the relayer.  This, in turn, prevented any StackerDB events from propagating to the signer.

I noticed that `test_stackerdb_dkg` isn't part of the integration test battery.  I went ahead and added it.

xoloki:
Thanks @jcnelson!  Glad it was a quick fix, and that we're putting the dkg test into CI.

> I noticed that `test_stackerdb_dkg` isn't part of the integration test battery. I went ahead and added it.

Can you change the number of signers/keys at `testnet/stacks-node/src/tests/signer.rs:161`?  It's currently at `16` signers and `40` keys, which is weird since it means we can't evenly divide those keys among the signers.  I'd suggest something like `100` keys and `4` signers.  This should add less than a minute to CI.  

Get that done and I'll approve this PR.

xoloki (approved):
LGTM, thanks so much for the quick fix and CI inclusion!

xoloki:
> The DKG test would only run on release CI, so added time isn't too much of a concern for day-to-day tests. I went ahead and used 100 signers and 4000 keys. If this somehow fails release CI due to a timeout, I can bump it back to 4 signers and 100 keys. Does that work for you?

Keep in mind that there will be multiple threads per signer, and depending on the OS and hardware hundreds of threads might run poorly.  But either way I'm happy.

jferrant (approved):

PR #3988, opened 2023-10-04
WIP: Add placeholder functions to trigger DKG from POX contract state

jferrant:
This is a WIP PR. 

TODO: 
- [ ] Address all TODO comments by resolving in this PR or creating tickets for them 
- [ ] Update wsts once WSTS release
- [ ] Test the changes with dummy responses


stacks-signer/src/stacks_client.rs
```diff
@@ -91,6 +94,28 @@ impl StacksClient {
         }
     }
 
+    /// Retrieve the current DKG aggregate public key
+    pub fn get_aggregate_public_key(&self) -> Result<Option<Point>, RPCError> {
```
jferrant, line 98: Clarity call should be a function of burn block height only and return an Optional Point (Public Key)

stacks-signer/src/stacks_client.rs
```diff
@@ -91,6 +94,28 @@ impl StacksClient {
...
+        let _burn_block_height = self.get_burn_block_height()?;
+        todo!("Make the read only contract call to retrieve the aggregate public key for the given block height")
+    }
+
+    /// Retreive the DKG aggregate public key vote cast by the signer
+    pub fn get_aggregate_public_key_vote(&self) -> Result<Option<Point>, RPCError> {
```
jferrant, line 104: Clarity function should take a burn block height and a signer public key.

stacks-signer/src/stacks_client.rs
```diff
@@ -91,6 +94,28 @@ impl StacksClient {
...
+        todo!("Make the read only contract call to retrieve the aggregate public key vote cast by the sender for a given block height")
+    }
+
+    /// Retreive the current burn block height
+    fn get_burn_block_height(&self) -> Result<u64, RPCError> {
+        todo!("Get the current burn block height from the stacks node")
+    }
+
+    /// Cast the DKG aggregate public key vote
+    pub fn cast_aggregate_public_key_vote(&self, _vote: Point) -> Result<(), RPCError> {
```
jferrant, line 115: Clarity contract should expect a public key and verify that the sender is a valid signer who has NOT yet cast their vote.

stacks-signer/src/stacks_client.rs
```diff
@@ -158,143 +188,189 @@ impl StacksClient {
...
+            } else {
+                Ok(None)
+            }
+        } else {
+            Err(ClientError::MalformedClarityValue(public_key_clarity_value))
+        }
+    }
+
+    /// Sends a transaction to the stacks node for a modifying contract call
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
jferrant: Yes this is still required because we are not yet actually submitting a modifying contract call anymore (originally we were because of the cast_aggregate_public_key_vote() function that I made, but we may no longer need this. This function will then only be used when making the sBTC transactions.

stacks-signer/src/stacks_client.rs
```diff
@@ -158,143 +188,189 @@ impl StacksClient {
...
+    /// Helper function to retrieve the pox contract address and name from the stacks node
+    fn get_pox_contract(&self) -> Result<(StacksAddress, ContractName), ClientError> {
+        // Check if we have overwritten the pox contract ID in the config
+        if let Some(pox_contract) = self.pox_contract_id.clone() {
+            return Ok((pox_contract.issuer.into(), pox_contract.name));
         }
-
-        let Some(tx) = tx_signer.get_tx() else {
-            return Err(ClientError::SignatureGenerationFailure);
+        // TODO: we may want to cache the pox contract inside the client itself (calling this function once on init)
```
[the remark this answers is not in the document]
jferrant: Will add to the code https://github.com/stacks-network/stacks-blockchain/issues/4005

stacks-signer/src/stacks_client.rs
```diff
@@ -158,143 +188,189 @@ impl StacksClient {
...
-        let tx_fee = 0;
+        let mut unsigned_tx = StacksTransaction::new(self.tx_version, tx_auth, tx_payload);
 
-        self.serialize_sign_sig_tx_anchor_mode_version(
-            payload.into(),
-            nonce,
-            tx_fee,
-            TransactionAnchorMode::OnChainOnly,
-        )
+        // FIXME: Because signers are given priority, we can put down a tx fee of 0
```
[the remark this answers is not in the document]
jferrant: Will add to the code https://github.com/stacks-network/stacks-blockchain/issues/4006

xoloki (approved):
LGTM thanks!

stacks-signer/src/stacks_client.rs
```diff
@@ -158,143 +193,191 @@ impl StacksClient {
...
+    }
+
+    /// Helper function that attempts to deserialize a clarity hex string as the aggregate public key
+    fn parse_aggregate_public_key(&self, hex: &str) -> Result<Option<Point>, ClientError> {
+        let public_key_clarity_value = ClarityValue::try_deserialize_hex_untyped(hex)?;
+        if let ClarityValue::Optional(optional_data) = public_key_clarity_value.clone() {
+            if let Some(ClarityValue::Sequence(SequenceData::Buffer(public_key))) =
+                optional_data.data.map(|boxed| *boxed)
+            {
+                let mut bytes = [0_u8; 32];
```
jcnelson, line 259: You're going to want to check the length of `public_key`.  It might not be 32 bytes; a `(buff 32)` is _at most_ 32 bytes.

stacks-signer/src/stacks_client.rs
```diff
@@ -320,6 +403,46 @@ impl StacksClient {
...
+    ) -> String {
+        format!(
+            "{}/v2/contracts/call-read/{contract_addr}/{contract_name}/{function_name}",
+            self.http_origin
+        )
+    }
+}
+
+/// Retry a function F with an exponential backoff and notification on transient failure
+pub fn retry_with_exponential_backoff<F, E, T>(request_fn: F) -> Result<T, ClientError>
```
jcnelson, line 429: I'd be curious as to how often you need to retry requests?  The blockchain should strive to get this down to zero.

jcnelson (approved):
This LGTM, modulo the small bug in parsing the aggregate public key.  Thanks!

stacks-signer/src/stacks_client.rs
xoloki: We ran into a problem where hammering `stacks-node` during startup caused it to go into a bad state.  We didn't try to diagnose the problem, though we should have filed a ticket.  Whatever the state was it prevented the node from bootstrapping.  Adding large retry timeouts (O(seconds)) fixed it.   There were retry timeouts already in some places in the code, but now they cover all cases not just connection failures.

During normal operation we didn't usually see failures.

jferrant: Unfortunately...often . Bceause in the integration test. The node will not start up if the signer isn't up. But the signer cannot be up without trying to contact the node. :P So we initialize the signer and open its listener port. The node starts up and connects to it. But immediately the run_one_loop starts running and starts hammering the node which is still starting up...and this causes it to hang. So the backoff keeps it from destroying it as it tries to connect for the necessary info (DKG key for example)

stacks-signer/src/stacks_client.rs
jferrant: Oh good catch! Fixed :D

PR #3989, opened 2023-10-05
Feat/dkg contract interaction

soju-drinker:
### Description

This PR adds the ability for the Signer to interact with Clarity contracts

### Applicable issues
- needed for https://github.com/stacks-network/stacks-blockchain/issues/3971

### Checklist
- [x] Read only contract calls
- [x] Submit tx contract calls
- [x] Simulate contract call responses in tests

### DIFF
--- a/stacks-signer/src/stacks_client.rs
+++ b/stacks-signer/src/stacks_client.rs
@@ -21,7 +21,7 @@ use stacks_common::{
 };
 use wsts::net::{Message, Packet};
 
-use crate::config::{Config, Network};
+use crate::config::Config;
 
 /// Temporary placeholder for the number of slots allocated to a stacker-db writer. This will be retrieved from the stacker-db instance in the future
 /// See: https://github.com/stacks-network/stacks-blockchain/issues/3921
@@ -83,7 +83,7 @@ pub struct StacksClient {
     stacks_address: StacksAddress,
     /// The private key used in all stacks node communications
     stacks_private_key: StacksPrivateKey,
-    /// A map of a slot ID to last chunk version   
+    /// A map of a slot ID to last chunk version
     slot_versions: HashMap<u32, u32>,
     /// The RPC endpoint used to communicate HTTP endpoints with
     http_origin: String,
@@ -106,8 +106,8 @@ impl From<&Config> for StacksClient {
             stacks_address: config.stacks_address,
             slot_versions: HashMap::new(),
             http_origin: format!("http://{}", config.node_host),
-            tx_version: TransactionVersion::Testnet,
-            chain_id: Network::Testnet.to_chain_id(),
+            tx_version: config.network.to_transaction_version(),
+            chain_id: config.network.to_chain_id(),
             stacks_node_client: reqwest::blocking::Client::new(),
         }
     }
--- a/stacks-signer/src/config.rs
+++ b/stacks-signer/src/config.rs
@@ -14,6 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+use blockstack_lib::chainstate::stacks::TransactionVersion;
 use clarity::vm::types::QualifiedContractIdentifier;
 use hashbrown::HashMap;
 use p256k1::ecdsa;
@@ -82,6 +83,14 @@ impl Network {
         match self {
             Self::Mainnet => C32_ADDRESS_VERSION_MAINNET_SINGLESIG,
             Self::Testnet => C32_ADDRESS_VERSION_TESTNET_SINGLESIG,
+        }
+    }
+
+    /// Convert a Network enum variant to a Transaction Version
+    pub fn to_transaction_version(&self) -> TransactionVersion {
+        match self {
+            Self::Mainnet => TransactionVersion::Mainnet,
+            Self::Testnet => TransactionVersion::Testnet,
         }
     }
 }


stacks-signer/src/stacks_client.rs
```diff
@@ -54,6 +71,23 @@ impl From<&Config> for StacksClient {
     }
 }
 
+/// Trait used to make interact with Clarity contracts for use in the signing process
+pub trait InteractWithStacksContracts {
```
jferrant, line 75: Just nitpicky but I prefet this being called StacksContractable :P

stacks-signer/src/stacks_client.rs
```diff
@@ -98,8 +132,122 @@ impl StacksClient {
...
+                res,
+                StacksTransaction::consensus_deserialize(&mut &tx[..])
+                    .unwrap()
+                    .txid()
+                    .to_string()
+            );
+            return res;
+        } else {
+            eprintln!("Submit tx error: {}", res.text().unwrap());
+            panic!("");
```
jferrant, line 247: we will have to turn this probably into some sort of retry with backoff logic and never panic.

stacks-signer/src/stacks_client.rs
```diff
@@ -54,6 +70,34 @@ impl From<&Config> for StacksClient {
     }
 }
 
+/// Trait used to make interact with Clarity contracts for use in the signing process
+pub trait InteractWithStacksContracts {
```
jferrant, line 74: Really nitpicky. Can ignore if you like, but think maybe something like "StacksContractCallable" is better. I can be easily convinced otherwise though :P

stacks-signer/src/stacks_client.rs
```diff
@@ -98,8 +142,157 @@ impl StacksClient {
...
+        })
+        .to_string();
+
+        let client = reqwest::blocking::Client::new();
+        let path = format!(
+            "{http_origin}/v2/contracts/call-read/{contract_addr}/{contract_name}/{function_name}"
+        );
+        let res = client
+            .post(&path)
+            .header("Content-TYpe", "application/json")
```
jferrant, line 282: Typo. Should be Content-Type

stacks-signer/src/stacks_client.rs
```diff
@@ -98,8 +142,157 @@ impl StacksClient {
...
+            .body(body)
+            .send()
+            .unwrap();
+        if res.status().is_success() {
+            let res: String = res.json().unwrap();
+            return res;
```
jferrant, line 288: Mind running clippy on this PR? 
`cargo clippy -p stacks-signer --tests --no-deps --all-features`

stacks-signer/src/stacks_client.rs
jferrant: Would you mind updating this to return a result. And then move the "FIXME" to the panic to the external calling of this function? Think its better for the traits to never panic and leave it to the caller to panic.

stacks-signer/src/stacks_client.rs
```diff
@@ -98,8 +142,157 @@ impl StacksClient {
...
+            None,
+            tx_fee,
+            TransactionAnchorMode::OnChainOnly,
+            tx_version,
+            chain_id,
+        )
+    }
+
+    fn submit_tx(http_origin: &str, tx: &Vec<u8>) -> String {
+        let client = reqwest::blocking::Client::new();
```
jferrant, line 237: I would maybe make client part of StacksClient struct and pass it in here. This way we can just keep reusing the same one/not create a new connection every time.

stacks-signer/Cargo.toml
```diff
@@ -35,11 +35,14 @@ slog = { version = "2.5.2", features = [ "max_level_trace" ] }
...
 slog-term = "2.6.0"
 stacks-common = { path = "../stacks-common" }
+stackslib = { path = "../stackslib"}
 thiserror = "1.0"
 toml = "0.5.6"
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 wsts = { git = "https://github.com/Trust-Machines/wsts", tag = "4.0.0rc2" }
+tinyhttp = "0.3.8"
+reqwest = { version = "0.11.22", features = ["blocking", "json"] }
```
jferrant, line 45: I do believe Jude has concerns about using reqwest because of unsafe code in the blocking call. We may need to hunt around for a different crate :(

jferrant: I would also be happy with allowing this through for now and making a ticket to repalce it with some alternative. I think the HTTP code throughout stacks is being pulled into a lib and maybe we could reuse that?

jcnelson: > I do believe Jude has concerns about using reqwest because of unsafe code in the blocking call. We may need to hunt around for a different crate :(

I have many, many concerns about `reqwest` -- in particular, the god-awful dependencies it pulls in.

If you're only doing blocking calls, I'd recommend `ureq`.  No need to import the whole and run the whole of `tokio` just to read and write data on a socket.

> I really really really do not want to have to implement a custom http client.

Oh but it's so much fun! ;)

stacks-signer/src/stacks_client.rs
jcnelson: Can I ask what the trait definition here is for?  Are there two or more implementations?  If not, then perhaps the trait can be dropped altogether.

stacks-signer/src/stacks_client.rs
```diff
@@ -98,8 +142,157 @@ impl StacksClient {
...
+                TransactionAuth::Standard(sender_spending_condition)
+            }
+        };
+        let mut unsigned_tx = StacksTransaction::new(version, auth, payload);
+        unsigned_tx.anchor_mode = anchor_mode;
+        unsigned_tx.post_condition_mode = TransactionPostConditionMode::Allow;
+        unsigned_tx.chain_id = chain_id;
+
+        let mut tx_signer = StacksTransactionSigner::new(&unsigned_tx);
+        tx_signer.sign_origin(sender).unwrap();
```
jcnelson, line 185: Can this just return an `Err(..)` instead of panicking?

stacks-signer/src/stacks_client.rs
```diff
@@ -98,8 +142,157 @@ impl StacksClient {
...
+        payload: TransactionPayload,
+        sender: &StacksPrivateKey,
+        payer: Option<&StacksPrivateKey>,
+        sender_nonce: u64,
+        payer_nonce: Option<u64>,
+        tx_fee: u64,
+        anchor_mode: TransactionAnchorMode,
+        version: TransactionVersion,
+        chain_id: u32,
+    ) -> Vec<u8> {
```
jcnelson, line 156: Let's not `panic` if we don't need to :)  This should be a `Result`.

stacks-signer/src/stacks_client.rs
```diff
@@ -98,8 +142,157 @@ impl StacksClient {
...
+        unsigned_tx.chain_id = chain_id;
+
+        let mut tx_signer = StacksTransactionSigner::new(&unsigned_tx);
+        tx_signer.sign_origin(sender).unwrap();
+        if let (Some(payer), Some(_)) = (payer, payer_nonce) {
+            tx_signer.sign_sponsor(payer).unwrap();
+        }
+
+        let mut buf = vec![];
+        tx_signer
```
jcnelson, line 191: There's `StacksMessageCodec::serialize_to_vec()` which converts `T: StacksMessageCodec` into `Vec<u8>`:

```rust
// N.B. this error variant needs to exist
let tx = tx_signer.get_x().ok_or(Error::TransactionSignatureFailure)?;
let buf = tx.serialize_to_vec();
```

stacks-signer/src/stacks_client.rs
jcnelson: A lot of the trait's functions take the same arguments over and over (e.g. `http_origin`, `chain_id`, `tx_version`).  I think these are values that the trait implementation can just supply.  Why would a caller ever need to supply different values for the same trait implementation instantiation?

stacks-signer/src/stacks_client.rs
```diff
@@ -98,8 +142,157 @@ impl StacksClient {
         // See: https://github.com/stacks-network/stacks-blockchain/issues/3921
         SLOTS_PER_USER
     }
+
+    fn seralize_sign_tx_anchor_mode_version(
```
jcnelson, line 146: It might make more sense to have one function for non-sponsored transactions, and one for sponsored transactions.  Then, the caller won't have to supply `None` gratuitously.

stacks-signer/src/stacks_client.rs
```diff
@@ -98,8 +142,157 @@ impl StacksClient {
...
+            nonce,
+            None,
+            tx_fee,
+            TransactionAnchorMode::OnChainOnly,
+            tx_version,
+            chain_id,
+        )
+    }
+
+    fn submit_tx(http_origin: &str, tx: &Vec<u8>) -> String {
```
jcnelson, line 236: I don't know if this is ever going to be a concern, but if you're registering a BNS name, then you'd also need to supply an `Attachment` which contains the name's zonefile data.

stacks-signer/src/stacks_client.rs
jcnelson: Seconded.  Let's not panic if we can help it.  Panicking is the library consumer's job :)

Cargo.lock
```diff
@@ -1679,6 +1737,19 @@ dependencies = [
  "tokio-rustls",
 ]
 
+[[package]]
```
jcnelson, line 1740: When you switch to `ureq`, please be sure to only use the minimum features required to solve the problem (e.g. JSON support and that's it).  No need for cookies or TLS -- the Stacks node supports neither anyway.

stacks-signer/src/stacks_client.rs
jferrant: I am not entirely sure what this means XD Forgive me...

stacks-signer/Cargo.toml
jferrant: @soju-drinker pulled out the ureq suggestions into a separate issue - https://github.com/stacks-network/stacks-blockchain/issues/3995

Hopefully we can get this merged asap to build off of (prevent massive PR)

jcnelson: Just to be clear, this is going to need to be addressed before a merge to `master`.

stacks-signer/src/stacks_client.rs
```diff
@@ -28,17 +42,47 @@ pub enum ClientError {
...
+    #[error("Failure to submit tx")]
+    TransactionSubmissionFailure,
+    /// Failed to sign with the provided private key
+    #[error("Failed to sign with the given private key")]
+    SignatureGenerationFailure,
+    /// Failed to sign with the provided private key
+    #[error("Failed to sign with the sponsor private key")]
+    SponsorSignatureGenerationFailure,
+    /// Failed to sign with the provided private key
+    #[error("Failed to ")]
```
jcnelson, line 64: Can serialization fail?  Or is this just for signature failure?

stacks-signer/src/stacks_client.rs
```diff
@@ -49,11 +93,41 @@ impl From<&Config> for StacksClient {
...
+/// Trait used to make interact with Clarity contracts for use in the signing process
+pub trait StacksContractCallable {
+    /// Submits a transaction to a node RPC server
+    fn submit_tx(&self, tx: Vec<u8>) -> Result<String, ClientError>;
+
+    /// Call read only tx
+    fn read_only_contract_call(
+        &self,
+        contract_addr: &StacksAddress,
+        contract_name: &str,
```
jcnelson, line 115: We have specific containers for contract names, function names, and function arguments, since they have to use limited character sets and since not all strings are valid.  You should use `ContractName` and `ClarityName` here instead of a bare `&str`.  Also, `function_args` should be `&[clarity::vm::Value]` in both cases.

stacks-signer/src/stacks_client.rs
```diff
@@ -98,6 +172,173 @@ impl StacksClient {
...
+impl StacksContractCallable for StacksClient {
+    fn transaction_contract_call(
+        &self,
+        nonce: u64,
+        contract_addr: &StacksAddress,
+        contract_name: &str,
+        function_name: &str,
+        function_args: &[ClarityValue],
+    ) -> Result<Vec<u8>, ClientError> {
+        let contract_name = ContractName::from(contract_name);
```
jcnelson, line 256: These can panic.  Yes, it's bad rust practice, but it is what it is today.  `::from(x)` just does `::try_from(x).unwrap()` for `ClarityName` and `ContractName`.  This is why this trait interface should use `ContractName` and `ClarityName`.

stacks-signer/src/stacks_client.rs
```diff
@@ -98,6 +172,173 @@ impl StacksClient {
...
+
+    fn submit_tx(&self, tx: Vec<u8>) -> Result<String, ClientError> {
+        let path = format!("{}/v2/transactions", self.http_origin);
+        let res = self
+            .stacks_node_client
+            .post(path)
+            .header("Content-Type", "application/octet-stream")
+            .body(tx.clone())
+            .send()
+            .unwrap();
```
jcnelson, line 284: This should not panic.

stacks-signer/src/stacks_client.rs
```diff
@@ -98,6 +172,173 @@ impl StacksClient {
...
+        if res.status().is_success() {
+            let res: String = res.json().unwrap();
```
jcnelson, line 286: Please remove all panicking unwraps and expects.  This method is already fallible, so it should just return an `Err(..)` if the network request fails for some reason.

stacks-signer/src/stacks_client.rs
```diff
@@ -98,6 +172,173 @@ impl StacksClient {
...
+
+    fn read_only_contract_call(
+        &self,
+        contract_addr: &StacksAddress,
+        contract_name: &str,
+        function_name: &str,
+        function_args: &[&str],
+    ) -> Result<String, ClientError> {
+        debug!("Calling read-only function {}...", function_name);
+        let body = json!({"sender": self.stacks_address.to_string(), "arguments": function_args})
```
jcnelson, line 308: This can fail if `function_args` does not contain serialized `clarity::vm::Value`s.  This interface should instead take `&[clarity::vm::Value]` for `function_args`.

stacks-signer/src/stacks_client.rs
```diff
@@ -98,6 +172,173 @@ impl StacksClient {
...
+            .to_string();
+        let path = format!(
+            "{}/v2/contracts/call-read/{contract_addr}/{contract_name}/{function_name}",
```
jcnelson, line 311: Same problem here -- `contract_name` should be a `ContractName` and `function_name` should be a `ClarityName` in the interface, because this RPC will fail if you pass an invalid contract name or function name string.

stacks-signer/src/stacks_client.rs
```diff
@@ -115,3 +356,88 @@ fn slot_id(id: u32, message: &Message) -> u32 {
...
+                &config.client.stacks_address,
+                "contract-name",
+                "function-name",
+                &[],
+            )
+        });
+        write_response(
+            config.mock_server,
+            b"HTTP/1.1 200 OK\n\n{\"okay\":false,\"cause\":\"Some reason\"}",
+        );
```
jcnelson, line 439: You might get a 400 or a 404 as well.  Can you test these conditions too?

jcnelson (changes requested):
This PR is on the right track, but it needs a few changes:

* It needs to use `ContractName`, `ClarityName`, and `clarity::vm::Value` instead of bare `&str`s
* The `StacksContractCallable` trait may not need to exist

stacks-signer/src/stacks_client.rs
jferrant: Done.

jferrant:
> This PR is on the right track, but it needs a few changes:
> 
>     * It needs to use `ContractName`, `ClarityName`, and `clarity::vm::Value` instead of bare `&str`s
> 
>     * The `StacksContractCallable` trait may not need to exist

Done :)

stacks-signer/src/stacks_client.rs
```diff
@@ -50,7 +103,12 @@ impl From<&Config> for StacksClient {
                 config.stackerdb_contract_id.clone(),
             ),
             stacks_private_key: config.stacks_private_key,
+            stacks_address: config.stacks_address,
             slot_versions: HashMap::new(),
+            http_origin: format!("http://{}", config.node_host),
+            tx_version: TransactionVersion::Testnet,
```
jferrant, line 109: Gotta fix.

jferrant: Fixed.

jferrant (approved):
LGTM

stacks-signer/src/stacks_client.rs
jferrant: Quick note, can you create a valid "StacksAddress" from the pox contract address? My initial attempts have failed. The reason I changed this to String was to accommodate this. However, perhaps I am just doing it wrong?

jcnelson: Sure -- `QualifiedContractIdentifier::parse("ST000000000000000000002AMW42H.pox-3").unwrap()` should work.

stacks-signer/src/stacks_client.rs
jcnelson: Ah, so there's this storage system in the Stacks node called Atlas  It's a hold-over from the Blockstack / Stacks 1.0 days.  Basically, a smart contract can emit specially-crafted Atlas events which contain the hash of a chunk of data.  The Atlas system tracks these hash announcements, and then communicates with other Stacks nodes to see if they have a chunk of data which hashes to it.  Eventually, all nodes get a copy of this "attached" data.  We use it in BNS to tie a BNS name to a zone file, as well as other off-chain data like Nostr public keys and such.

I don't think the signer is ever going to use Atlas, so it's not a big deal if you're not going to support attachments.  However, if this module becomes more generic and gets spun off into its own crate, it might become important later to add support.

stacks-signer/src/stacks_client.rs
jferrant: Yes sorry I wasn't clear/ I am saying that does not work 
 'called `Result::unwrap()` on an `Err` value: InvalidContractAddress("\"ST000000000000000000002AMW42H")',

jferrant: Sorry, I was not clear. This is what I already tried and it results in 
```thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Runtime(ParseError("Invalid principal literal: base58ck checksum 0x51104e95 does not match expected 0x31287a31"), None)', stacks-signer/src/stacks_client.rs:315:83```

This is the same result as trying to parse ST000000000000000000002AMW42H as a StacksAddress.

xoloki (approved):
LGTM.  More tests would be nice, especially if they weren't mocked, but this is fine for now.

jcnelson (approved):

PR #3992, opened 2023-10-06
Nakamoto staging blocks

kantai:
This PR has the initial work for handling nakamoto staging blocks.

Currently, the staging blocks table is part of the same chainstate sqlite db as the old staging blocks table. This simplifies things initially, but part of the nakamoto workstream will be putting that in its own sqlite db.

This extends the `TenureChange` PR with a few more items:

1. Adds Epoch30. This patch will need to be reconciled with #3954, but that should be relatively easy.
2. It adds miner pubkey hash tracking to the sortition db. This isn't discussed in the nakamoto SIP, but I think this is a useful thing for the implementation: the miner's VRF key register transaction should include the Hash160 of the miner's pubkey to be used in its tenures. Otherwise, stacks nodes would have to wait until they've processed the tenure change transaction to validate a signature.

It also adds three methods for the staging blocks: getting the next ready block, marking a block as processed, and marking a burn block as processed. Because a block's "burn view" is somewhat independent of the burn block that selected its tenure, this should be tracked separately, so the nakamoto staging blocks table tracks "burn_attachable" and "stacks_attachable".


stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -1830,6 +1842,39 @@ impl<'a> SortitionHandleConn<'a> {
...
     }
 
+    /// Does the sortition db expect to receive unknown blocks from
+    /// this tenure?
+    ///
+    /// This is used by nakamoto nodes while they are at or near the
+    /// current chain tip: only recent tenures can receive blocks this
+    /// way. Otherwise, the `BlockHeaderHash` must have been
+    /// explicitly confirmed by a block commit.
+    pub fn expects_blocks_from_tenure(
```
jcnelson, line 1852: I'm having a hard time reconciling the docstring with the code body.  The code body returns the _latest_ `BlockSnapshot` whose block-signing key hash matches the given public key, if there any such snapshots within the last `NAKAMOTO_TENURE_BLOCK_ACCEPTANCE_PERIOD` sortitions.  What does this have to do with "expect[ing] to receive unknown blocks?"  Also, it seems miners can reuse the same signing key in multiple tenures, so if there were two snapshots from the same miner in this window, this method would not return the earlier one.  Is this desired?

stackslib/src/chainstate/burn/operations/leader_key_register.rs
```diff
@@ -77,6 +77,12 @@ impl LeaderKeyRegisterOp {
         Some(LeaderKeyRegisterOp::new(&prover_pubk))
     }
 
+    /// Interpret the first 20 bytes of the key registration's memo field as the Hash160 of
+    ///  of the public key that will sign this miner's nakamoto blocks.
+    pub fn interpret_nakamoto_signing_key(&self) -> Option<Hash160> {
```
jcnelson, line 82: Ah, interesting idea!  This should get added to the SIP and design doc.  Can you add a quick one-sentence comment in both places so we remember to merge it in?  Thanks!

jcnelson: Also, this field should probably be explicit in `LeaderKeyRegisterOp` as an `Option<..>`.  The field would be required to be `Some(..)` in epoch 3.0, but it's not considered in earlier epochs.

stackslib/src/chainstate/coordinator/mod.rs
```diff
@@ -2141,6 +2141,16 @@ impl<
         self.inner_handle_new_burnchain_block(&mut HashSet::new())
     }
 
+    /// Are affirmation maps active during the epoch?
```
jcnelson, line 2144: Good news -- they're going away in Epoch 3.0.  There's no longer a need for them because the chain can't fork.

stackslib/src/chainstate/burn/db/sortdb.rs
kantai: Good point. I think this should be a bool instead.

At first, I intended to use this result to figure out the consensus hash that started the miner's tenure. However, for exactly the reason you point out, it cannot do that (because the miner's key could be reused). Instead, I think that the Nakamoto block headers should contain the consensus hash of their tenure. There's other ways to figure out their associated tenure, I'm sure, but it dramatically simplifies things if the block's include it. I know it's a somewhat counter-intuitive change from Stacks 2.0's perspective: in 2.0, a miner cannot know their consensus hash when they produce their block, but because block production happens _after_ their tenure starts in 3.0, they actually **do** know their consensus hash.

stackslib/src/chainstate/coordinator/mod.rs
kantai: Yeah, I agree. But I think the code may have to stay in place for a while, and the intention here is to have a function that encapsulates the decision on whether an affirmation map check should occur.

stackslib/src/chainstate/burn/operations/leader_key_register.rs
kantai: I think there is a little bit of code clarity tradeoff with adding a new field to the struct, and I'm genuinely not sure which would work out more elegantly. If there's a new field in the struct, there's two kind of immediate impacts:

1. The db schema of the Sortition DB needs to change and FromRow/ToSql need to handle the new field. The db schema of the burnchain DB doesn't change, but the deserialization of the opcode would need to make sure that it handles the absence of the field correctly. *Or* the schema of the SortitionDB doesn't change and the FromRow / Insert method handles shuttling the new field via the existing memo column.
2. The consensus deserialize/serialize needs to be sure to be compatible. This is pretty straight forward and I'm not worried about its impact.

In light of (1), do you still think there should be a new field for this struct?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -270,9 +340,39 @@ impl NakamotoBlockHeader {
 impl NakamotoBlock {
     /// Did the stacks tenure change on this nakamoto block? i.e., does this block
     ///  include a TenureChange transaction?
-    pub fn tenure_changed(&self) -> bool {
-        // TODO: when tenure change txs are implemented, this must be updated
-        true
+    pub fn tenure_changed(&self, parent: &StacksBlockId) -> bool {
```
jcnelson, line 343: I think it might make more sense to return the _list_ of `TenureChange` transactions here.  The tenure can change multiple times in a block, since it has to contain all _missing_ `TenureChange` transactions as well.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -288,6 +388,150 @@ impl NakamotoBlock {
...
+    ) -> Result<(), ChainstateError> {
+        let recovered_miner_pk = block.header.recover_miner_pk().ok_or_else(|| {
+            warn!(
+                "Stacks block downloaded with unrecoverable miner public key";
+                "block_hash" => %block.header.block_hash(),
+            );
+            return ChainstateError::InvalidStacksBlock("Unrecoverable miner public key".into());
+        })?;
+
+        if sortdb
```
jcnelson, line 466: I think this whole `if`-block needs to be dropped.  A node can accept any block back to the last PoX anchor block, especially while booting up.

stackslib/src/chainstate/burn/db/sortdb.rs
jcnelson: > Instead, I think that the Nakamoto block headers should contain the consensus hash of their tenure.

I agree -- we should use this instead of `BurnchainHeaderHash` for `burn_view`.

stackslib/src/chainstate/burn/operations/leader_key_register.rs
jcnelson: I think that in each breaking change, we've required users to spin up from genesis anyway, so introducing a new field which renders the Stacks 2.x chainstate incompatible I think is fine (especially if it simplifies the consensus code).

stackslib/src/chainstate/coordinator/mod.rs
jcnelson: Ah, I meant to say that the check could return `false` for epoch 3.0.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: I think there needs to be two different acceptance paths for blocks:

1. Any block that has been committed to by a leader block commit -- leader block commits still commit to block hashes, but those are not the chain tip. These are used to accept blocks while booting up, catching up to the chain tip.
2. The if block in the code above: while operating at the chain tip, the only blocks that should be accepted are ones in recent tenures.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: I think that will be necessary once the `TenureChange` transactions are actually getting validated, but right now, all the stacks block processor cares about is whether or not a tenure change occurred.

stackslib/src/chainstate/burn/db/sortdb.rs
jcnelson: I'll do this when working on the coordinator

stackslib/src/chainstate/burn/operations/leader_key_register.rs
jcnelson: I can take this on

stackslib/src/chainstate/nakamoto/tests.rs
```diff
@@ -229,11 +387,11 @@ pub fn nakamoto_advance_tip_multiple() {
 
     let mut last_block: Option<NakamotoBlock> = None;
     let index_roots = [
-        "a8fa50e42ee48d393ac86036d8eec1bf6ada344de4100478fed37f1106281033",
-        "c431be733f37dee9dce70e6de51f616c2ed8b23583af8f7c4a8bce86c5f0a183",
-        "0c5e890e95e2f92ef36934bc0e5d71a6715974593f7d952b07ee6f959dae3f1c",
-        "0cd7696c4920ea5bc498ea46ee1df8566e06ea0bc8fd16a1e0ffd292d55f746e",
-        "84177188b1c02af772d2442b760fd9215b9dfadeed5723504acda2c94b068d15",
+        "c76d48e971b2ea3c78c476486455090da37df260a41eef355d4e9330faf166c0",
```
jcnelson, line 390: I ended up changing these hashes when I was working on the coordinator (since I had to change some of the header structs, which changes the state root hash).  That's okay, right?

jcnelson (approved):
Thanks @kantai!

stackslib/src/chainstate/nakamoto/tests.rs
kantai: Yeah, it's fine -- perhaps after factoring the mining routine in the mockamoto node, we should update this test anyways so that its able to figure out its own hashes.

saralab:
@obycode @jbencin : We need you attention on this PR, can you start reviewing this? This is critical for Mockamoto readiness

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -118,7 +148,10 @@ lazy_static! {
                      stacker_signature TEXT NOT NULL,
                      tx_merkle_root TEXT NOT NULL,
                      state_index_root TEXT NOT NULL,
-
+                     -- the consensus hash of the burnchain block that selected this block's tenure
+                     consensus_hash TEXT NOT NULL,
+                     -- the consensus hash of the burnchain block that selected this block's tenure
```
obycode, line 153: Is this comment intended to be identical to the above field's comment?

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -3062,6 +3157,15 @@ impl SortitionDB {
                     // TODO: This should move to Epoch 30 once it is added
                     || version == "8"
             }
+            StacksEpochId::Epoch30 => {
+                version == "3"
+                    || version == "4"
+                    || version == "5"
+                    || version == "6"
+                    || version == "7"
+                    // TODO: This should move to Epoch 30 once it is added
```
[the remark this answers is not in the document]
kantai: Yeah, also needed to update the version checks in the other epochs -- higher versions are also valid in earlier epochs (so `"8"` is valid in Epoch 2.0 - Epoch 3.0)

stackslib/src/chainstate/nakamoto/mod.rs
kantai: No -- will update

muneeb-ali:
Awesome to see this being reviewed!

> "miner pubkey hash tracking to the sortition db. This isn't discussed in the nakamoto SIP"
If this is not in the SIP already then probably worth adding. The SIPs will be the first place a lot of devs will look for information and can help their understanding of how the code works.

obycode (approved):
LGTM! 🚀

kantai:
> Awesome to see this being reviewed!
> 
> > "miner pubkey hash tracking to the sortition db. This isn't discussed in the nakamoto SIP"
> 
> If this is not in the SIP already then probably worth adding. The SIPs will be the first place a lot of devs will look for information and can help their understanding of how the code works.

Yep -- opened as a PR against the current SIP (https://github.com/stacksgov/sips/pull/158)

PR #3997, opened 2023-10-09
Update wsts and p256k1 libs in stacks-signer

jferrant:
WSTS has been released with the necessary changes we need. Have upated the dependencies.


jcnelson (approved):

xoloki (approved):
Looks great, thanks!

PR #4001, opened 2023-10-17
Initial implementation of mockamoto stack-node mode

kantai:
This PR implements a first version of mockamoto. It has several deficiencies at the moment, but those could be sorted in separate PRs (which is, perhaps, an argument for leaving those to a separate PR):

1. It does not use the chain coordinator. This is pretty intentional, because I didn't want this entry-point to be dependent on the ongoing coordinator work. An update to use the chain coordinator (once that supports nakamoto) would be to simply insert mock data into the burnchain db rather than the sortition db (as this PR does).
2. It does not mine transactions from the mempool: it just produces empty-ish blocks. There isn't really a hard problem to solve here, it's just a matter of doing the work of factoring the mempool iteration logic in the miner module into something that could be invoked by the mockamoto mining method.
3. It only produces blocks that undergo tenure changes. 
4. The p2p stack doesn't use the new canonical chain tip lookup methods, so they will return incorrect stack tip information.

If those don't get covered by this PR, I'll open issues for them.

This is still a draft because right now, there's no associated unit tests -- it may also be worthwhile to just include some of the above features in this PR.


zone117x:
> It only produces blocks that undergo tenure changes.

Does this mean the blocks include the new `TenureChange` transaction type? If so then would you recommend using this PR as a mocknet node in the API to start integrating the new tx type(s)?

kantai:
> Does this mean the blocks include the new TenureChange transaction type? If so then would you recommend using this PR as a mocknet node in the API to start integrating the new tx type(s)?

Yes, it does include the new transaction type. Once the mocknet node integrates with the chain coordinator work (which this PR doesn't yet), then yeah, the mocknet node is a good place to start integrating the new tx type. Until then, it won't emit events (events are emitted by the chain coordinator).

saralab:
@kantai : Is this still a draft? Can I request reviews?

kantai:
Yes its ready for reviews. This PR is ultimately pretty minor and will just be a good place for merging in the work from Jude's coordinator PR.

muneeb-ali:
Would having more reviewers help or is that not a bottleneck right now? I understand that only people currently on the Nakamoto WG are probably best for reviewing but just wondering if reviewing is something that needs more attention or not.

testnet/stacks-node/src/mockamoto.rs
```diff
@@ -0,0 +1,565 @@
...
+            coinbase_tx_payload,
+        );
+        coinbase_tx.chain_id = chain_id;
+        coinbase_tx.set_origin_nonce(miner_nonce);
+        let mut coinbase_tx_signer = StacksTransactionSigner::new(&coinbase_tx);
+        coinbase_tx_signer.sign_origin(&self.miner_key).unwrap();
+        let coinbase_tx = coinbase_tx_signer.get_tx().unwrap();
+
+        let parent_block_id = StacksBlockId::new(&chain_tip_ch, &chain_tip_bh);
+        let tenure_change_tx_payload = TransactionPayload::TenureChange(TenureChangePayload {
```
obycode, line 377: Can you add a comment here? I guess you're just including a tenure change in every block to test some more paths?

testnet/stacks-node/src/mockamoto.rs
```diff
@@ -0,0 +1,565 @@
...
+        sortition_id: SortitionId::new(&new_bhh.clone(), &PoxId::new(vec![true])),
+        parent_sortition_id: parent_snapshot.sortition_id.clone(),
+        pox_valid: true,
+        accumulated_coinbase_ustx: 0,
+        miner_pk_hash: Some(miner_pkh.clone()),
+    };
+    Ok(new_snapshot)
+}
+
+pub struct MockamotoNode {
```
obycode, line 187: Can you add a comment on this struct, or a top-level commenting explaining what Mockamoto does and why it is useful? It doesn't need to be anything to complicated, since this is just temporary, but I think this will be helpful for others using it during development.

obycode (approved):
Just had a couple of requests for added comments for clarity, but it looks great!

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1000,11 +1049,12 @@ impl NakamotoChainState {
         clarity_instance: &'a mut ClarityInstance,
         sortition_dbconn: &'b dyn SortitionDBRef,
         pox_constants: &PoxConstants,
-        chain_tip: &StacksHeaderInfo,
```
jcnelson, line 1003: Can you help me understand why replacing `chain_tip` with `parent_stacks_height` and `parent_burn_height` is required?  This seems needless to me.

jcnelson (approved):
Approving this to move things along, but please re-consider the change to `setup_block`'s arguments.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: It reduces the amount that needs to be mocked: the `chain_tip` argument is _only_ used for obtaining the parent stacks height and parent burn height (for epoch checks, mostly). It also makes writing unit tests easier. An argument like `chain_tip` which includes a bunch of information that the function _doesn't_ use confuses the interface.

kantai:
Merging -- if we want to change the `setup_block` interface back to accepting a `StacksHeaderInfo` struct, we can always do that later.

PR #4009, opened 2023-10-24
Feat/nakamoto coordinator

jcnelson:
This is a PR which implements the full Nakamoto coordinator.  It makes the following notable changes:

* Nakamoto blocks now contain only the consensus hash and parent StacksBlockId to identify their placement in the block history.  No more parent consensus hash, and no more `burn_view`.

* It reuses the canonical Stacks tip memoization logic from the 2.x days to enable the Sortition DB to track the canonical Stacks tip as blocks get processed

* It splits the coordinator's main loop into a 2.x and a Nakamoto subroutine.  It will use the 2.x coordinator loop up until the PoX anchor block for the first Nakamoto reward cycle is processed.  At this point, the Nakamoto coordinator loop will be used.

* It prevents the 2.x coordinator logic from processing Stacks and Bitcoin blocks beyond the start of the Nakamoto epoch.

* It requires that the Nakamoto PoX anchor block be the parent of the first Nakamoto block mined in the prepare phase.  In doing so, it requires that the node calculate the reward set at the _beginning_ of the prepare phase, instead of at the end.

The last bullet point imposes three additional design requirements:  the Nakamoto epoch must start ~100 blocks into a reward phase (so Stackers can come online and DKG can happen), the last 2.x reward cycle _must_ be a PoX reward cycle, and there needs to be a "Nakamoto bootstrap" contract deployed to mainnet in which Stackers can register their signing keys prior to the first Nakamoto reward cycle.  This contract would be used just for this purpose, and it cannot be a boot contract.

---

I still need to do a lot of testing in this PR, but I'm posting it as a draft for now to get early feedback in the design.


testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -354,7 +354,7 @@ impl EventObserver {
     fn make_new_block_processed_payload(
         &self,
         filtered_events: Vec<(usize, &(bool, Txid, &StacksTransactionEvent))>,
-        block: &StacksBlock,
+        block: StacksBlockEventData,
```
kantai, line 357: Why not pass `block` as a reference? It doesn't seem like this will avoid copying data, because the `block` fields are all getting string serialized anyways.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+            SortitionDB::get_stacks_epoch(sortdb.conn(), current_burn_height)?.expect(&format!(
+                "FATAL: no epoch defined for burn height {}",
+                current_burn_height
+            ));
+
+        if participation == 0 {
+            // no one is stacking.  This is a fatal error.
+            error!("No PoX participation. Aborting.");
+            panic!();
+        }
```
kantai, line 96: Should this actually panic, or should the node just stall?

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+        else {
+            // no header for this snapshot (possibly invalid)
+            continue;
+        };
+
+        let parent_block_id = &prepare_start_block_header
+            .anchored_header
+            .as_stacks_nakamoto()
+            .expect("FATAL: queried non-Nakamoto tenure start header")
+            .parent_block_id;
```
kantai, line 235: This imposes a requirement on the Nakamoto start -- the prepare phase before the first reward cycle must _also_ be evaluated with nakamoto rules. I think this is important to highlight in the comments.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+
+        let stacks_block_hash = anchor_block_header.anchored_header.block_hash();
+        let txid = anchor_block_sn.winning_block_txid;
+
+        info!(
+            "Anchor block selected for cycle {}: {}/{}",
+            reward_cycle, &anchor_block_header.consensus_hash, &stacks_block_hash
+        );
+
+        let block_id = StacksBlockId::new(&anchor_block_header.consensus_hash, &stacks_block_hash);
```
kantai, line 256: Is this not just `parent_block_id`?

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+                &mut self.chain_state_db,
+                sortdb_handle,
+                1,
+                self.dispatcher,
+            )?;
+
+            if processed_blocks.len() == 0 {
+                // out of blocks
+                break;
+            }
```
kantai, line 415: Given that we only process nakamoto blocks 1 at a time, does it makes sense to keep the `process_nakamoto_blocks` function which is just a looping wrapper around `process_next_nakamoto_block`?

kantai (commented):
This looks great to me, just have some review comments.

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -1842,36 +1861,23 @@ impl<'a> SortitionHandleConn<'a> {
...
+        let Some(bhh) = SortitionDB::get_burnchain_header_hash_by_consensus(self, consensus_hash)?
+        else {
+            return Ok(false);
+        };
+        let Some(_sortition_id) = self.get_sortition_id_for_bhh(&bhh)? else {
+            return Ok(false);
+        };
+
+        // TODO: query set of stacker signers in order to get the aggregate public key
         Ok(true)
```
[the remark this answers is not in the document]
jcnelson: Yes -- the TODO ensures that the only way this method returns `true` is if the signature came from the _current_ stackers, identified by the `consensus_hash` (which in turn identifies the reward cycle its MARF'ed PoX info, which would include the aggregate public key).

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -409,22 +423,47 @@ impl NakamotoChainState {
...
+                                 WHERE consensus_hash = ?";
+        staging_db_tx.execute(&update_dependents, &[consensus_hash])?;
 
         Ok(())
     }
 
-    pub fn next_ready_block(
+    /// Find the next ready-to-process Nakamoto block, given a connection to the staging blocks DB.
+    /// Returns (the block, the size of the block)
+    pub fn next_ready_nakamoto_block(
```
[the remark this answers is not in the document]
jcnelson: I'm just trying to disambiguate Stacks 2.x methods from Nakamoto methods whenever possible, so the reader won't have to figure out which implementation is used.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -449,6 +488,268 @@ impl NakamotoChainState {
...
+                rusqlite::params![consensus_hash, block_hash],
+                |row| {
+                    let data: Vec<u8> = row.get("data")?;
+                    let block = NakamotoBlock::consensus_deserialize(&mut data.as_slice())?;
+                    if &block.header.block_hash() != block_hash {
+                        panic!(
+                            "Staging DB corruption: expected {}, got {}",
+                            &block_hash,
+                            &block.header.block_hash()
+                        );
```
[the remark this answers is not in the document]
jcnelson: I'm not sure how this is recoverable in Nakamoto?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -462,42 +763,31 @@ impl NakamotoChainState {
...
         })?;
 
-        if sortdb
-            .expects_blocks_from_tenure(&recovered_miner_pk)?
-            .is_none()
-        {
-            let msg = format!("Received block, signed by {recovered_miner_pk:?}, but this pubkey was not associated with recent tenures");
-            warn!("{}", msg);
-            return Err(ChainstateError::InvalidStacksBlock(msg));
-        };
```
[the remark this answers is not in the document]
jcnelson: Ah, that's a bug.  Yes, the miner signature should be checked.  Will fix.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -828,10 +1149,36 @@ impl NakamotoChainState {
...
+                &FIRST_STACKS_BLOCK_HASH,
+            )
+        {
             // not the first-ever block, so linkage must occur
-            assert_eq!(new_tip.parent, parent_tip.block_hash());
-            assert_eq!(&new_tip.parent_consensus_hash, parent_consensus_hash);
+            match parent_tip {
+                StacksBlockHeaderTypes::Epoch2(stacks_header) => {
+                    // this is the first nakamoto block
+                    assert_eq!(parent_tip.block_hash(), stacks_header.block_hash());
```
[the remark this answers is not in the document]
jcnelson: Yup, will delete

testnet/stacks-node/src/event_dispatcher.rs
jcnelson: By passing by value, we ensure that the caller can't accidentally send the same events twice.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
jcnelson: I don't think there's a functional difference?  The absence of PoX necessarily means the node can't do anything more.  I can make it just return an error instead, and have the coordinator runloop handle it by stalling if you prefer.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
jcnelson: Ah, that's an oversight.  Elsewhere, the code assumes that the Nakamoto epoch kicks off early in the reward phase of a reward cycle (e.g. 100 blocks in), so the last Stacks 2.x prepare phase can choose a PoX anchor and give the signers some time to do DKG and prepare to sign Nakamoto blocks.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
jcnelson: I can do that.

stackslib/src/chainstate/burn/db/sortdb.rs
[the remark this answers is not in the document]
kantai: I think it cannot use the `consensus_hash` argument, but instead something like a new `tip: &ConsensusHash` argument. For example, if a node's current view is like Bitcoin block `10,000`, it should _never_ use this method for accepting blocks from Bitcoin block `6,000`, because then it is a DoS vector: a prior stacking set could sign off on blocks and broadcast them and the node would have no way to reject them. For "old" blocks, the node must use the block hashes committed to during leader block commits. This method must only accept current stacker set signatures (_not_ "current at the time the block claims to have been produced"), otherwise the trust relationship for nodes and stackers is that nodes trust prior stacker sets _indefinitely_ to not DoS them.

stackslib/src/chainstate/nakamoto/mod.rs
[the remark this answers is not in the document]
kantai: But doesn't the name of the struct already disambiguate them?

stackslib/src/chainstate/nakamoto/mod.rs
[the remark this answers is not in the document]
kantai: I suppose its fine to `panic!` here, but this scenario seems basically equivalent to DBErrors like `parse` errors (i.e., if `data` isn't a `Vec<u8>` or deserialize to a block, why does that error rather that `panic`?) where we typically opt to yield an error rather than immediately panicking.

testnet/stacks-node/src/event_dispatcher.rs
kantai: But the caller is calling `block.clone()`, and it definitely _is_ invoking this method multiple times with the same events.

muneeb-ali:
Great to see this PR! 

> I still need to do a lot of testing in this PR, but I'm posting it as a draft for now to get early feedback in the design.

@jcnelson really appreciate the effort to get early feedback. A lot of devs are really interested in learning/knowing more Nakamoto implementation and sharing things early even when they don't feel fully ready helps a ton with that 🙏 🙌

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -1122,6 +1124,22 @@ impl db_keys {
...
+
+    pub fn last_reward_cycle_from_string(rc_str: &str) -> u64 {
+        let bytes = hex_bytes(rc_str).expect("CORRUPTION: bad format written for reward cycle ID");
+        assert_eq!(
+            bytes.len(),
+            8,
+            "CORRUPTION: expected 8 bytes for reward cycle"
+        );
+        let mut rc_buff = [0; 8];
+        rc_buff.copy_from_slice(&bytes[0..8]);
```
jferrant, line 1140: NIT: you have already asserted the length here so you should be able to just do copy_from_slice(&bytes)

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -1991,12 +2006,13 @@ impl<'a> SortitionHandleConn<'a> {
...
-    ) -> Result<bool, db_error> {
-        self.get_sortition_id_for_bhh(burn_header_hash)
+    /// Has `consensus_hash` been processed in the current fork?
+    pub fn processed_block(&self, consensus_hash: &ConsensusHash) -> Result<bool, db_error> {
+        let Some(bhh) = SortitionDB::get_burnchain_header_hash_by_consensus(self, consensus_hash)?
+        else {
+            return Ok(false);
+        };
+        self.get_sortition_id_for_bhh(&bhh)
             .map(|result| result.is_some())
```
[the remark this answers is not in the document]
jferrant: If we end up implementing recovery mode to revert to pre-Nakamoto block production rules, we definitely will have forks in this instance.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
jferrant: I think I would stall. I would expect the node to crash when there is some unrecoverable error. If this can be fixed by waiting and then someone stacking...then I would stall. If it is an irrecoverable error...panic.

jcnelson:
Just a heads-up, this PR will be kept in draft status until I can finish upgrading `TestBurnchainBlock`, `TestStacksNode`, and `TestPeer` to produce Nakamoto epochs.  Then, I can add tests for the Nakamoto chains coordinator, and proceed to work on the new network state machines (whose testing depends on a working `TestPeer` implementation).  I've also added a Nakamoto miner implementation as part of the `TestPeer` work (it's pretty small -- it's just boilerplate and a few wrappers around the existing miner's transaction-selection code).

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+    pub fn get_nakamoto_reward_cycle_info(
+        &mut self,
+        burn_header: &BurnchainBlockHeader,
+    ) -> Result<Option<RewardCycleInfo>, Error> {
+        let sortition_tip_id = self
+            .canonical_sortition_tip
+            .as_ref()
+            .expect("FATAL: Processing anchor block, but no known sortition tip");
+
+        get_nakamoto_reward_cycle_info(
```
jferrant, line 548: Seems a bit strange to keep this separated into two functions like this with the same name? Is this to accommodate easier testing?

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+        debug!("Handle new canonical burnchain tip";
+               "height" => %canonical_burnchain_tip.block_height,
+               "block_hash" => %canonical_burnchain_tip.block_hash.to_string());
+
+        // Retrieve all the direct ancestors of this block with an unprocessed sortition
+        let mut cursor = canonical_burnchain_tip.block_hash.clone();
+        let mut sortitions_to_process = VecDeque::new();
+
+        // We halt the ancestry research as soon as we find a processed parent
+        let mut last_processed_ancestor = loop {
```
jferrant, line 583: NIT: This is by no means important, but it might be nice to pull this loop into a function called "get_last_processed_ancestor" as this function is quite long and am having a hard time following it. Am sure there are other logical blocks that might be better in a sep function, but just commenting on this one to spur thought about it.

jcnelson:
@kantai Bunch of commits just landed.  This adds the PoC for an end-to-end test of a Nakamoto node bootup, complete with Epoch 2.5, the new Nakamoto Coinbase variant (which must contain a VRF proof), the Nakamoto miner, the stubbed `.pox-4` contract, and the fleshed out `TestPeer` and supporting infrastructure.  Will continue to add tests, but feel free to crib from this to expand mockamoto to mine blocks :+1:

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -876,6 +876,16 @@ const SORTITION_DB_SCHEMA_8: &'static [&'static str] = &[
     r#"ALTER TABLE snapshots ADD miner_pk_hash TEXT DEFAULT NULL"#,
 ];
 
+const SORTITION_DB_SCHEMA_9: &'static [&'static str] = &[
```
kantai, line 879: SortitionDB version 8 is already new in `next`.

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -4852,6 +4902,16 @@ impl SortitionDB {
             None => Ok(None),
         }
     }
+    
+    /// Get a block snapshot for a winning Nakamoto tenure in a given burn chain fork.
```
kantai, line 4906: What does the function really do -- from this description it isn't clear to me. It seems like its querying by the committed-to block id? Maybe

```
/// Get the block snapshot of a tenure in the burn chain fork specified by `tip` where the winning block commit specified `last_tenure_id` as the most recent stacks block.
```

stackslib/src/chainstate/coordinator/mod.rs
```diff
@@ -723,7 +739,30 @@ pub fn get_reward_cycle_info<U: RewardSetProvider>(
...
+    if let Ok(Some(reward_cycle_info)) = reward_cycle_info.as_ref() {
+        // cache the reward cycle info as of the first sortition in the prepare phase, so that
+        // the Nakamoto epoch can go find it later
+        let ic = sort_db.index_handle(sortition_tip);
+        let prev_reward_cycle = burnchain
+            .block_height_to_reward_cycle(burn_height)
+            .expect("FATAL: no reward cycle for burn height");
+
+        if prev_reward_cycle > 1 {
+            let prev_reward_cycle_start = burnchain.reward_cycle_to_block_height(prev_reward_cycle - 1);
```
kantai, line 753: I think the prepare phase start math here should go in the `burnchain/mod.rs` `PoxConstants` implementation. It would also be good to dedupe with the `static_is_in_prepare_phase`.

stackslib/src/chainstate/coordinator/mod.rs
```diff
@@ -2212,9 +2251,40 @@ impl<
...
+    /// that matters to you, then you should call them directly.
     pub fn handle_new_burnchain_block(&mut self) -> Result<Option<BlockHeaderHash>, Error> {
-        self.inner_handle_new_burnchain_block(&mut HashSet::new())
+        let canonical_burnchain_tip = self.burnchain_blocks_db.get_canonical_chain_tip()?;
+        let epochs = SortitionDB::get_stacks_epochs(self.sortition_db.conn())?;
+        let target_epoch_index = StacksEpoch::find_epoch(&epochs, canonical_burnchain_tip.block_height).expect("FATAL: epoch not defined for burnchain height");
+        let target_epoch = epochs.get(target_epoch_index).expect("FATAL: StacksEpoch::find_epoch() returned an invalid index");
+        if target_epoch.epoch_id < StacksEpochId::Epoch30 {
+            // burnchain has not yet advanced to epoch 3.0
+            self.handle_new_epoch2_burnchain_block(&mut HashSet::new())
```
kantai, line 2265: Rather than a nesting `if-else` here, can you just `return self.handle_new_epoch2_burnchain_block()` at this branch? That way, the function can just continue without an else.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -77,6 +80,8 @@ impl OnChainRewardSetProvider {
 
         let liquid_ustx = chainstate.get_liquid_ustx(block_id);
 
+        debug!("PoX addrs at {} ({}): {:?}", block_id, registered_addrs.len(), &registered_addrs);
```
kantai, line 83: This is already logged via `StacksChainState::make_reward_set`

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -390,36 +415,54 @@ impl<
...
-
-            let Some(Some(block_receipt)) = processed_blocks.pop() else {
-                // this block was invalid
-                debug!("Bump blocks processed (invalid)");
-                self.notifier.notify_stacks_block_processed();
-                increment_stx_blocks_processed_counter();
-                continue;
             };
 
             // only bump the coordinator's state if the processed block
```
kantai, line 468: I think this check needs to be dropped -- the block_hash may not show up in the sortition-fork yet.

jcnelson:
Okay, I think this PR is now functionally complete.  I'll go back-fill with unit test coverage before marking it as ready.

The good news is that this PR sufficiently expands `TestPeer` and its supporting test infrastructure to the point where we can now use it to verify that the Nakamoto chains coordinator correctly processes new burnchain and Stacks blocks as they arrive, and correctly chooses the PoX anchor block and PoX recipients :partying_face:. Moreover, we can verify that the system gracefully transitions from epoch 2.4 to epoch 2.5 (which instantiates `.pox-4` and unlocks `.pox-3`), and from epoch 2.5 to epoch 3.0.  All the while, we can verify that the Nakamoto coordinator chooses the correct PoX anchor block and reward recipients, regardless of whether or not the previous reward cycle's anchor bock was an epoch 2.x Stacks block or a Nakamoto block (turns out there's a subtle difference in handling these two cases!)

The less-good news is that it took a lot of work to get there because I found a few of surprising architectural constraints that we didn't originally foresee in SIP-021.  In order to discover these, it was necessary to get the Nakamoto VRF to work as well as parts of the Nakamoto miner, since their correct functioning is on the critical path to being able to test the above.  However, this is necessary work, since it would need to be done one way or another.

That said, I discovered a few subtle architectural differences between epoch 2.x and Nakamoto I'd like to draw your attention to.  These will need to find their way into the SIP at some point.  They're not earth-shaking, but they do supersede SIP-007.

* In Nakamoto, the miner generates the VRF proof on the sortition hash of the sortition which contains the block-commit of the _current_ tenure.  In epoch 2.x, the miner used the sortition hash of the penultimate sortition (i.e. the sortition that was processed prior to the one containing the block-commit).  This is possible in Nakamoto, but impossible in epoch 2.x, because Nakamoto produces blocks _after_ the block-commit is confirmed.

* In Nakamoto, the reward set is calculated as part of processing the first prepare-phase Nakamoto block.  The reward set is then cached to disk, and later loaded in order to (1) find the first PoX recipients of the next reward cycle and (2) instantiate the PoX reward set into the sortition DB MARF as part of processing the first reward-phase sortition.  Stacks epoch 2.x calculates the reward set when the first Stacks block of the reward phase is mined, and then re-calculates and stores the reward set into the sortition DB when the first reward-phase sortition is processed.  The reason for this difference is two-fold:  (1) there's no need to vote on an anchor block in Nakamoto since it can't fork (so it's possible to do eagerly), and (2) we need to calculate the reward set info right away so that Stacker signers can access it in order to carry out DKG.  In the future, the preprocessed reward set info will be served via JSON RPC.

* In Nakamoto, the anchor block is always the _tenure-start_ block of the _last_ reward-phase tenure.  The reason it must be this block is because of how block-commit processing works.  To validate a block-commit, it is necessary to prove that the block-commit descends from the PoX anchor block.  This is achieved today by querying the sortition DB for the ID of the sortition which contains the anchor block ID, and performing an ancestor walk from the block-commit's block hash back to this anchor block ID.  The anchor block ID itself _must_ be in a prior block-commit.  So, in order to avoid changing the block-commit validation logic, I opted to simply require that the PoX anchor block be a Nakamoto block whose hash is in a block-commit -- namely, the Nakamoto block whose index block hash is in the block-commit produced by the miner of the first prepare-phase block (which is the index block hash of the tenure-start block of the parent tenure, which happens to be the last tenure-start block in the reward phase).

stackslib/src/chainstate/burn/db/sortdb.rs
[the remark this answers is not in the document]
jcnelson: I've updated this method to require that `consensus_hash` is within one reward cycle of the `SortitionHandleConn`'s tip.  During initial block download, it's possible that there will still be some blocks from the previous reward cycle in-flight while the next reward cycle is processed, so the limit must be permitted to straddle the reward cycle boundary.

stackslib/src/chainstate/nakamoto/mod.rs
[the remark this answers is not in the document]
jcnelson: For now, assuming the reader doesn't rename it on `use` and assuming that it never takes `self` as an argument.

stackslib/src/chainstate/nakamoto/mod.rs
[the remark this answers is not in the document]
jcnelson: Changed to DBError

stackslib/src/chainstate/nakamoto/mod.rs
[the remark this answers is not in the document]
jcnelson: This is now fixed in the code

testnet/stacks-node/src/event_dispatcher.rs
jcnelson: I can change to a reference

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
jcnelson: Okay, as of right now (e6df731) the Nakamoto epoch would start in the reward phase of an epoch 2.x reward cycle.

stackslib/src/chainstate/burn/db/sortdb.rs
jcnelson: I know; this is just me being defensive

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
jcnelson: At one point, I had `get_nakamoto_next_recipients` call it (which doesn't have a handle to the coordinator instance).  I'll just move the code body back.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
jcnelson: Ha, this is copypasta from the epoch 2.x coordinator.  I'll consolidate.

stackslib/src/chainstate/burn/db/sortdb.rs
jcnelson: Yup, fixed

stackslib/src/chainstate/coordinator/mod.rs
jcnelson: Happy to put this into `PoxConstants` but the act of deduping with `static_is_in_prepare_phase` will entail writing even more code than leaving it as-is (we'd need a `prepare_phase_start`, `static_reward_cycle_to_block_height`, and `static_prepare_phase_start`)

stackslib/src/chainstate/coordinator/mod.rs
jcnelson: sure

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -0,0 +1,1320 @@
+;; The .pox-3 contract
```
jferrant, line 1: Nit: Update the comment

jcnelson:
Okay, I've transitioned this from draft to ready-for-review.  I've added lots of test coverage to `NakamotoChainState`, the static block-validation logic in `NakamotoBlock`, and the coordinator via `TestPeer`.  In particular, the coordinator tests will now replay Stacks blocks _in random order_ against a blank `TestPeer` in order to verify that the coordinator correctly reacts to their arrival and processes the next reward cycle's sortitions only once the PoX anchor block can be identified.

stackslib/src/chainstate/stacks/boot/pox-4.clar
jcnelson: Thanks! fixed

clarity/src/vm/docs/mod.rs
```diff
@@ -2773,10 +2773,18 @@ mod test {
...
 
+        fn get_v3_unlock_height(&self) -> u32 {
+            u32::MAX
+        }
+
         fn get_pox_3_activation_height(&self) -> u32 {
             u32::MAX
         }
 
+        fn get_pox_4_activation_height(&self) -> u32 {
```
jferrant, line 2784: Is there a purpose to having different "get_*_unlock_height" ? they all return the same u32::MAX? Should this be reduced to just one function "get_unlock_height" ? Similarly wondering about activation height.

stackslib/src/burnchains/mod.rs
```diff
@@ -473,6 +510,15 @@ impl PoxConstants {
         }
     }
 
+    /// What's the first block in the prepare phase
+    pub fn prepare_phase_start(&self, first_block_height: u64, reward_cycle: u64) -> u64 {
+        let reward_cycle_start =
+            self.reward_cycle_to_block_height(first_block_height, reward_cycle);
```
jferrant, line 516: I know that this should never underflow as the length of the reward cycle and prepare length are known. Do we still rather have an explicit check with a panic if self.prepare_length > reward_cycle_start + self.reward_cycle_length since this does introduce a potential underflow ?

stackslib/src/burnchains/tests/mod.rs
```diff
@@ -468,15 +479,37 @@ impl TestBurnchainBlock {
...
+        &mut self,
+        ic: &SortitionDBConn,
+        miner: &mut TestMiner,
+        block_hash: &BlockHeaderHash,
+        burn_fee: u64,
+        leader_key: &LeaderKeyRegisterOp,
+        fork_snapshot: Option<&BlockSnapshot>,
+        parent_block_snapshot: Option<&BlockSnapshot>,
+    ) -> LeaderBlockCommitOp {
+        self.inner_add_block_commit(
```
jferrant, line 501: This is mostly out of curiosity than a complaint...why did you chose to separate this into a function directly wrapping another function?

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -3447,6 +3597,42 @@ impl SortitionDB {
...
+        rc_info: &RewardCycleInfo,
+    ) -> Result<(), db_error> {
+        let sql = "INSERT INTO preprocessed_reward_sets (sortition_id,reward_set) VALUES (?1,?2)";
+        let rc_json = serde_json::to_string(rc_info).map_err(db_error::SerializationError)?;
+        let args: &[&dyn ToSql] = &[sortition_id, &rc_json];
+        sort_tx.execute(sql, args)?;
+        Ok(())
+    }
+
+    /// Get a pre-processed reawrd set.
```
jferrant, line 3615: nit: Typo

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -0,0 +1,844 @@
...
+    // Note that we may not have processed it yet.  But, if we do find it, then it's
+    // unique (and since Nakamoto Stacks blocks are processed in order, the anchor block
+    // cannot change later).
+    let prepare_phase_sortitions =
+        find_prepare_phase_sortitions(sort_db, burnchain, sortition_tip)?;
+
+    // did we already calculate the reward cycle info?  If so, then return it.
+    let first_sortition_id = if let Some(first_sn) = prepare_phase_sortitions.first() {
+        if let Some(persisted_reward_cycle_info) =
+            SortitionDB::get_preprocessed_reward_set(sort_db.conn(), &first_sn.sortition_id)?
```
jferrant, line 219: Is it possible that a reward set is preprocessed BEFORE DKG is set and then this call occurs AFTER DKG is set, but the preprocessed reward set is not updated correctly to include this? (commenting mostly for my PR built ontop of this one)

stackslib/src/burnchains/mod.rs
jcnelson: This is checked in `PoxConstants::new()` :+1:

stackslib/src/burnchains/tests/mod.rs
jcnelson: Because `TestBurnchainBlock::add_nakamoto_tenure_commit()` also calls this function.  The difference between the two is that `TestBurnchainBlock::add_nakamoto_tenure_commit()` converts the `StacksBlockId` into a `BlockHeaderHash`, which is what the `LeaderBlockCommitOp` ultimately expects.  The algorithm is otherwise the same for both functions; it's just that the argument types are different.

stackslib/src/chainstate/burn/operations/leader_key_register.rs
```diff
@@ -83,16 +83,27 @@ impl LeaderKeyRegisterOp {
         self.memo.get(0..20).map(Hash160::from_bytes).flatten()
     }
 
+    /// Set the miner public key hash160 for block-signing
+    pub fn set_nakamoto_signing_key(&mut self, pubkey_hash160: &Hash160) {
+        if self.memo.len() < 20 {
+            let mut new_memo = vec![0; 20];
+            new_memo[0..self.memo.len()].copy_from_slice(&self.memo);
```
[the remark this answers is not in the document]
jcnelson: In Nakamoto, the `memo` field has been repurposed to contain the hash160 of the miner's public key (which is 20 bytes).  I don't think this is enforced by this PR, but I just opened an issue for it.  It'll ship with Neon.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
jcnelson: Yes, that will need to be addressed in your PR.  The reward set will need to be updated with the aggregate public key if it has since become available.  Perhaps the aggregate public key could be stored in a separate column in the same table? Then you don't have to load the entire reward set to check each time.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -0,0 +1,844 @@
...
+                    &block.header.block_hash.to_string(),
+                    block.header.block_height
+                )
+            })
+            .collect();
+
+        debug!(
+            "Unprocessed burn chain blocks [{}]",
+            dbg_burn_header_hashes.join(", ")
+        );
```
jbencin, line 721: It's probably more efficient debug print a string vector:

```suggestion
        debug!(
            "Unprocessed burn chain blocks {dbg_burn_header_hashes:?}"
        );
```

stackslib/src/chainstate/nakamoto/miner.rs
```diff
@@ -0,0 +1,812 @@
...
+        &self,
+        chainstate: &'a mut StacksChainState,
+        burn_dbconn: &'a SortitionDBConn,
+        tenure_start: bool,
+    ) -> Result<MinerTenureInfo<'a>, Error> {
+        debug!("Nakamoto miner tenure begin");
+
+        let burn_tip = SortitionDB::get_canonical_chain_tip_bhh(burn_dbconn.conn())?;
+        let burn_tip_height =
+            SortitionDB::get_canonical_burn_chain_tip(burn_dbconn.conn())?.block_height as u32;
```
jbencin, line 312: `u64 as u32` can truncate. Yes it will take thousands of years at current Bitcoin block rate, but suppose one day Bitcoin decides on faster blocks, or this code gets used elsewhere (like Subnets, or another chain). Or suppose we add integration tests with unrealistic values for `burn_height`. This sort of bug is unlikely to occur, but very difficult to track down, so it's better to make it impossible

```suggestion            
 u32::try_from(SortitionDB::get_canonical_burn_chain_tip(burn_dbconn.conn())?.block_height).expect("block_height overflow");
```

jbencin (commented):
I'm going to submit what I have and save the rest of this PR for tomorrow. Most of my comments are pretty minor and about Rust style/safety/optimization. I don't understand the codebase well enough yet to find the sort of logic errors that you and Aaron can.

clarity/src/vm/analysis/mod.rs
```diff
@@ -141,6 +141,7 @@ pub fn run_analysis(
             | StacksEpochId::Epoch22
             | StacksEpochId::Epoch23
             | StacksEpochId::Epoch24
+            | StacksEpochId::Epoch25
```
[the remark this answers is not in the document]
jcnelson: Epoch 2.5 is only used to deploy `pox-4`, which must happen one reward cycle prior to epoch 3.0.  This is because Nakamoto's PoX contract needs different functions than those in `pox-3`, and there need to be stackers ready to go when the first epoch 3.0 Bitcoin block gets mined (meaning, there must be stackers who stacked in `pox-4`).

stackslib/src/chainstate/nakamoto/miner.rs
```diff
@@ -0,0 +1,812 @@
...
+        // save the block so we can build microblocks off of it
+        let block = builder.mine_nakamoto_block(&mut tenure_tx);
+        let size = builder.bytes_so_far;
+        let consumed = builder.tenure_finish(tenure_tx);
+
+        let ts_end = get_epoch_time_ms();
+
+        if let Some(observer) = event_observer {
+            observer.mined_nakamoto_block_event(
+                SortitionDB::get_canonical_burn_chain_tip(burn_dbconn.conn())?.block_height + 1,
```
jferrant, line 590: Should this be a wrapping_add ?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -334,108 +365,514 @@ impl NakamotoBlockHeader {
...
+    ///    * all TenureChange transactions are present and in the right order, starting with
+    ///    `stacks_tip` and leading up to this block
+    /// * that only epoch-permitted transactions are present
+    pub fn validate_transactions_static(
+        &self,
+        mainnet: bool,
+        chain_id: u32,
+        epoch_id: StacksEpochId,
+    ) -> bool {
+        if self.txs.len() == 0 {
```
jferrant, line 821: This is a nit that will prob get fixed in a clippy PR, but in general when doing len() checks against 0, should use is_empty() (otherwise silence the clippy warning)

stackslib/src/clarity_vm/clarity.rs
```diff
@@ -2471,6 +2610,10 @@ mod tests {
                 u32::MAX
             }
 
+            fn get_v3_unlock_height(&self) -> u32 {
```
jferrant, line 2613: Is there a purpose to keeping these unlock height and activation height functions separated into v3/v1 and pox_3/pox_4? they always seem to be the same in which case it makes more sense to keep just a single "get_unlock_height" and a "get_pox_activation_height" ?

stackslib/src/chainstate/stacks/mod.rs
```diff
@@ -658,7 +658,13 @@ pub struct SchnorrThresholdSignature {
     //pub scalar: wsts::Scalar,
 }
 
-/// Reasons why a `TenureChange` transaction can be de
+impl SchnorrThresholdSignature {
```
jferrant, line 661: Oh I didn't see this when i was building off this branch. Perhaps I should be using this then instead of my SchnorrSignature in my own PR.

stackslib/src/chainstate/stacks/db/transactions.rs
```diff
@@ -8544,9 +8546,15 @@ pub mod test {
             fn get_v2_unlock_height(&self) -> u32 {
                 u32::MAX
             }
+            fn get_v3_unlock_height(&self) -> u32 {
```
jferrant, line 8549: Similar comment as before. Not sure why these are separated by version and pox since they are all identical

jferrant (approved):
All my comments are nits or about typos really. Don't think I know enough to find logic errors, so take this review with a grain of salt XD

Sooooo
 
LGTM

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -4887,6 +5024,17 @@ impl SortitionDB {
         })
     }
 
+    /// Get the block-commit for a Nakamoto block, given the block-commit's sortition's consensus
+    /// hash and its given last_tenure_id
+    pub fn get_block_commit_for_nakamoto_block(
```
kantai, line 5029: Can you expand on the rustdocs for this function with something like:

"If the winning block commit of the sortition identified by `consensus_hash` committed to `last_tenure_id`, then return `Some(winning_block_commit)`, otherwise, returns `None`."

Otherwise, it's not obvious that this returns the _winning block commit_ (and only under certain conditions).

stackslib/src/chainstate/coordinator/mod.rs
```diff
@@ -2251,40 +2260,54 @@ impl<
...
         }
+
+        // proceed to process sortitions in epoch 3.0
+        self.handle_new_nakamoto_burnchain_block()
+            .map(|can_proceed| {
+                if can_proceed {
+                    None
+                } else {
+                    // missing PoX anchor block, but unlike in 2.x, we don't know what it is!
+                    Some(BlockHeaderHash([0x00; 32]))
```
kantai, line 2308: I think it'd be better to just expand the return type of this function:

```rust
pub enum NewBurnchainBlock {
    ReadyForStacksBlocks,
    WaitForPox2x(BlockHeaderHash),
    WaitForPoxNakamoto,
}

 pub fn handle_new_burnchain_block(&mut self) -> Result<NewBurnchainBlock, Error>
```

That way there's no possibility of confusing the return behavior difference between 2.x and 3.x.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -362,7 +371,7 @@ impl NakamotoBlockHeader {
     }
 
     pub fn is_first_mined(&self) -> bool {
-        StacksBlockHeader::is_first_index_block_hash(&self.parent_block_id)
+        self.block_id() == StacksBlockId::first_mined()
```
kantai, line 374: This is a change of this method's semantics, right?

That `first_mined()` method returns `StacksBlockId::new(&FIRST_BURNCHAIN_CONSENSUS_HASH, &FIRST_STACKS_BLOCK_HASH)`, which is the _parent_ of the first mined block, right? The first actually mined block should have a real block hash and a real consensus hash.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -422,47 +431,84 @@ impl NakamotoBlockHeader {
...
             );
         }
 
-        let validate = |tc: &TenureChangePayload| -> Result<(), TenureChangeError> {
-            if tc.previous_tenure_end != *parent {
-                return Err(TenureChangeError::PreviousTenureInvalid);
-            }
+        let validate = |tc: &StacksTransaction| -> Result<(), TenureChangeError> {
+            if let TransactionPayload::TenureChange(tc) = &tc.payload {
+                if tc.previous_tenure_end != self.header.parent_block_id {
```
kantai, line 489: What does it mean if there are multiple tenure change transactions in a block? Would all of those tenure change transactions have the same `previous_tenure_end` field?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1137,34 +1577,131 @@ impl NakamotoChainState {
...
             SortitionDB::get_canonical_stacks_chain_tip_hash(sortdb.conn())?;
-        let index_block_hash = StacksBlockId::new(&consensus_hash, &block_bhh);
-        Self::get_block_header(conn, &index_block_hash)
+        Self::get_block_header(
+            chainstate_conn,
+            &StacksBlockId::new(&consensus_hash, &block_hash),
+        )
+    }
+
+    /// Get the parent header of a Nakamoto block.
```
kantai, line 1591: Is this the `parent header` of a nakamoto block or is it returning the block header of the first block in the tenure identified by `consensus_hash`?

kantai (approved):
LGTM -- just had a few comments, and one or two questions about behavior.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -334,108 +365,514 @@ impl NakamotoBlockHeader {
...
+        let wellformed = self.is_wellformed_first_tenure_block();
+        if wellformed.is_none() {
+            // block isn't a first-tenure block, so no valid tenure changes
+            return None;
+        } else if let Some(false) = wellformed {
+            // this block is malformed
+            info!("Block is malformed";
+                  "block_id" => %self.block_id());
+            return Some(false);
+        }
```
jbencin, line 469: If a function returns an `Option`, you can use the `?` operator with `Option` types:

```suggestion
        if !self.is_wellformed_first_tenure_block()? {
            // this block is malformed
            info!("Block is malformed";
                  "block_id" => %self.block_id());
            return Some(false);
        }
```

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -334,108 +365,514 @@ impl NakamotoBlockHeader {
...
     pub fn get_coinbase_tx(&self) -> Option<&StacksTransaction> {
-        match self.txs.get(0).map(|x| &x.payload) {
-            Some(TransactionPayload::Coinbase(..)) => Some(&self.txs[0]),
-            _ => None,
+        let wellformed = self.is_wellformed_first_tenure_block();
+        if wellformed.is_none() {
+            // block isn't a first-tenure block, so no coinbase
+            return None;
+        }
+        if let Some(false) = wellformed {
```
jbencin, line 527: Simplify:

```suggestion
        if !self.is_wellformed_first_tenure_block()? {
```

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -334,108 +365,514 @@ impl NakamotoBlockHeader {
...
+
+        // there is one coinbase.
+        // go find it.
+        self.txs.iter().find(|tx| {
+            if let TransactionPayload::Coinbase(..) = &tx.payload {
+                true
+            } else {
+                false
+            }
+        })
```
jbencin, line 540: Simplify:

```suggestion
        self.txs.iter().find(|tx| mathes!(&tx.payload, TransactionPayload::Coinbase(..)))
```

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -334,108 +365,514 @@ impl NakamotoBlockHeader {
...
+        self.get_coinbase_tx()
+            .map(|coinbase_tx| {
+                if let TransactionPayload::Coinbase(_, _, vrf_proof) = &coinbase_tx.payload {
+                    vrf_proof.as_ref()
+                } else {
+                    // actually unreachable
+                    None
+                }
+            })
+            .flatten()
```
jbencin, line 555: Simplify:

```suggestion
            .and_then(|coinbase_tx| {
                if let TransactionPayload::Coinbase(_, _, vrf_proof) = &coinbase_tx.payload {
                    vrf_proof.as_ref()
                } else {
                    // actually unreachable
                    None
                }
            })
```

jbencin (commented):
More comments...

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -904,6 +938,40 @@ impl EventDispatcher {
...
+        let payload = serde_json::to_value(MinedNakamotoBlockEvent {
+            target_burn_height,
+            block_hash: block.header.block_hash().to_string(),
+            block_id: block.header.block_id().to_string(),
+            stacks_height: block.header.chain_length,
+            block_size: block_size_bytes,
+            cost: consumed.clone(),
+            tx_events,
+        })
+        .unwrap();
```
jbencin, line 968: This probably should never fail, but it's still preferable to use `expect()` in non-test code

```suggestion
        .expect("Desarializing MinedNakamotoBlockEvent failed");
```

testnet/stacks-node/src/neon_node.rs
```diff
@@ -3526,8 +3542,8 @@ impl ParentStacksBlockInfo {
         Ok(ParentStacksBlockInfo {
             stacks_parent_header: stacks_tip_header,
             parent_consensus_hash: mine_tip_ch.clone(),
-            parent_block_burn_height: parent_block.block_height,
-            parent_block_total_burn: parent_block.total_burn,
+            parent_block_burn_height: parent_block_height,
+            parent_block_total_burn: parent_block_total_burn,
```
jbencin, line 3546:
```suggestion
            parent_block_total_burn,
```

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -310,6 +314,43 @@ pub mod test_observer {
...
+                    inner_obj
+                } else if let Some(inner_obj) = txevent_obj.get("ProcessingError") {
+                    inner_obj
+                } else if let Some(inner_obj) = txevent_obj.get("Skipped") {
+                    inner_obj
+                } else {
+                    panic!("TransactionEvent object should have one of Success, ProcessingError, or Skipped")
+                };
+                inner_obj
+                    .as_object()
```
jbencin, line 342: Can be simplified:
```suggestion
                txevent_obj.get("Success")
                    .or_else(|| txevent_obj.get("ProcessingError"))
                    .or_else(|| txevent_obj.get("Skipped"))
                    .expect("TransactionEvent object should have one of Success, ProcessingError, or Skipped")
                    .as_object()
```

jbencin (approved):
Approving... I left plenty of Rust suggestions but nothing critical. Only thing you should definitely fix are truncating `as` conversions

stackslib/src/core/mod.rs
```diff
@@ -432,9 +483,13 @@ pub static STACKS_EPOCH_2_3_MARKER: u8 = 0x08;
 /// *or greater*.
 pub static STACKS_EPOCH_2_4_MARKER: u8 = 0x09;
 
-/// Stacks 3.0 epoch marker.  All block-commits in 3.0 must have a memo bitfield with this value
+/// Stacks 2.5 epoch marker.  All block-commits in 2.5 must have a memo bitfield with this value
 /// *or greater*.
-pub static STACKS_EPOCH_3_0_MARKER: u8 = 0x0a;
+pub static STACKS_EPOCH_2_5_MARKER: u8 = 0x0a;
+
+/// Stacks 3.0 epoch marker.  All block-commits in 2.4 must have a memo bitfield with this value
```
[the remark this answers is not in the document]
jcnelson: Yes

stackslib/src/chainstate/stacks/mod.rs
jcnelson: I think this might be a @jbencin and @xoloki question. IIRC @xoloki told him to use `wsts::Point` and `wsts::Scalar` to encode the signature, but that was a while ago. It's possible that `wsts::common::Signature` is more appropriate now.

jcnelson: You should do whatever makes the most sense in your PR :)  This PR only needs this struct to encode/decode; it doesn't care about validating the contents.

stackslib/src/chainstate/stacks/db/transactions.rs
jcnelson: They're not identical on mainnet.  This trait impl is just for testing transaction-handling.

stackslib/src/chainstate/burn/db/sortdb.rs
jcnelson: This code is dead, actually.  I'll just remove it.

stackslib/src/chainstate/coordinator/mod.rs
jcnelson: Sure

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: That was a bug, but it was never tripped in my code because my tests only concern themselves with Nakamoto blocks that build atop epoch 2.x blocks (so, code branches that depend on `is_first_mined()` are never used in the first place).

stackslib/src/chainstate/stacks/mod.rs
jbencin: `ThresholdSignature` is the same as `wsts::common::Signature`, I defined a struct in this repo so that I could implement `StacksMessageCodec` for it (can't do that with types defined outside this crate)

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: There can be N > 1 tenure changes in block if and only if the past N - 1 tenures did not contain any blocks.  These are empty tenures, so their parent blocks are necessarily the same as the chain tip atop which the Nth tenure change builds if this block is valid.

jcnelson: Put another way, the `previous_tenure_end` points to the chain tip at which this tenure begins.  This is true even if the tenure is empty.  If N tenures in a row are empty, then their N tenure change transactions would have the same value for `parent_tenure_end` (since the chain tip is the same in all cases).  The N+1'st tenure's tenure change transaction would also have the same `parent_tenure_end` value.

jcnelson: Should point out -- this PR is not trying to handle the case where we need to issue a tenure change to deal with an empty sortition or a tenure budget extension.  I didn't think that was in-scope.  Right now, this PR only considers a tenure change to be valid if it is signaling the beginning of a new miner tenure triggered by a winning sortition.

jcnelson: I suspect this code, as well as `is_wellformed_first_tenure_block()`, is going to need to be modified in the follow-on PR that fully validates tenure-changes.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -473,38 +519,357 @@ impl NakamotoBlock {
...
-                // (a coinbase in any other position is invalid anyway).
-                return None;
+        let wellformed = self.is_wellformed_first_tenure_block();
+        if wellformed.is_none() {
+            // block isn't a first-tenure block, so no coinbase
+            return None;
+        }
+        if let Some(false) = wellformed {
+            // block isn't well-formed
+            return None;
```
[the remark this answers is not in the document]
jcnelson: It does, by way of `validate_transactions_static()`, which gets called by `validate_nakamoto_block_burnchain()`, which in turn is called by `accept_block()`.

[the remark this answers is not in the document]
jcnelson: So, an invalid Nakamoto block (i.e. one that's missing a coinbase) will never get stored to the staging DB.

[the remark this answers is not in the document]
jcnelson: As an added safety check, the case where the call to `get_coinbase_tx()` returns `None` causes `append_block()` to fail with an invalid-block error.

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: The latter.  The comment is outdated.

jcnelson:
Hi @jbencin, I appreciate your feedback, but I'm going to defer on most of the ergonomics changes for now.  I'm already too anxious about this PR blocking other work, and these changes don't affect the behavior or even the legibility of the code.  I did take the time to go through all the files I touched and use explicit integer conversions, however.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Okay, that makes sense to me then.

Issue #4030, opened 2023-11-02
Update core contract tests to use the Clarinet SDK


wileyj:
With the recent changes outlined here: https://www.hiro.so/blog/announcing-the-clarinet-sdk-a-javascript-programming-model-for-easy-smart-contract-testing

The core contract tests should be updated to use this SDK rather than relying on a soon to be outdated docker image. 
Currently, the only test being run is for the `bns.clar` contract using this step:
```yml
      - name: Execute core contract unit tests in Clarinet
        id: clarinet_unit_test
        uses: docker://hirosystems/clarinet:1.1.0
        with:
          args: test --coverage --manifest-path=./contrib/core-contract-tests/Clarinet.toml
 ```
ref: https://github.com/stacks-network/stacks-blockchain/blob/master/contrib/core-contract-tests/Clarinet.toml#L9

Future work should also be done to include other active boot contracts like `pox-3.clar` 
https://github.com/stacks-network/stacks-blockchain/tree/master/src/chainstate/stacks/boot

tagging @hugocaillard

hugocaillard:
I wonder what would be the purpose of these tests.
Since we are testing code that is already deployed on mainnet, it's not going to change or to break for no reason.
It could be interesting to have some sort of regtests, in a sense that if the tests break for some reason , it would mean that the testing env is broken.
But here the testing env is clarinet, and it does make much sense to me to run regtests on clarinet on this repo.

What do you think? cc @kantai?

I still migrated the bns test, because it was alre

kantai:
I think it's useful to have these tests in place even if the contract shouldn't change. Having them in place makes it easy to find tests to use as starting point for any future boot contracts, or for creating tests for bug reports in the boot contract. Running them as part of CI is useful precisely because they _shouldn't_ fail: if they fail, they mean that a PR unexpectedly altered a boot contract. While something like that should be very easy to spot during code review, the fact that a simple CI check would fail is helpful. This CI task is also very lightweight -- it's one of the fastest CI tasks in the repo currently.

hugocaillard:
Sounds good @kantai 👍 

For future reference, I'm linking here @moodmosaic's suggestion to introduce tests with fast-check:
https://github.com/stacks-network/stacks-core/pull/4031#discussion_r1382087703
https://github.com/stacks-network/stacks-core/pull/4031#discussion_r1382090406

moodmosaic:
> Sounds good @kantai 👍
> 
[...]

PR #4031, opened 2023-11-03
Test/migrate to sdk

hugocaillard:
#### Context

With the release of Clarinet 2.0.0, `clarinet test` has been deprecated. This PR uses the preferred clarinet-sdk to write unit-tests for the BNS contract.

### Description

I tried to stick to the initial test implementation as much as possible, but instead of having 1 big linear test, I split it in multiple `describe`/`it` blocks. Leveraging the `beforeEach` to set up the namespace/name registration for each test.

The code coverage is almost exactly the same, expect that the previous CI runs on Clarinet v1.1, which is a year old and we made a few fixes on the code coverage since then.
- The biggest change is that we now have branching data (since v1.7.0) in unit test, so a line like `(if (<= a b) a b))` (L22 in bns.clar) can be only partially hit but we wouldn't know
- Before v1.7, the `define-<function>` line we not considered as executable lines. (This is what explains the decreasing of 0.32% on bns.clar)

### Implementation

To give a bit of context on how to migrate the Clarinet v1 tests (with Deno) to the Clarinet JS SDK tests (with Node), here are I think the biggest changes:

#### - Switching to Node
The clarinet-sdk could basically run in any WASM environment (Node, Deno, web browsers etc) but is really tested and optimized for Node only. Also, the test runner we recommend is Vitest (instead of the Deno one before).
Hence why this PR creates some config files (package.json, vitest.config, tsconfig).
This is all done with the command `npx @hirosystems/clarinet-sdk@latest`.

#### - Clarity Values

Instead of `types.<type>` such as `types.uint` that was used in the Deno lib, we now rely on Stacks.js. So `types.uint` becomes `Cl.uint`.

#### - Assertions

Assertions on Clarity values are different as well. The Deno lib would take a string such as `"(some u1)"` and allow to chain assertions functions that would also deserialise the result (`"(some u1)".expectSome().expectUint(1)`).
We now use custom Vitest matchers, that can take Stacks.js Clarity values in case of composite types (`expect(<result>).toBeSome(Cl.uint(1))`).

#### - Public function call

In the old tests, we had to call `chain.mineBlock([])` to call 1 public function.
While `simnet.mineBlock([])` is still available in the clarinet-sdk and useful to make multiple transaction in 1 block. We can also use `simnet.callPublicFunction()` to mine a block with 1 public function call transaction.

### Applicable issues
- #4030 (only partially)

I also updated the CI.

#### Test output

```
 ✓ tests/bns/name_register.test.ts (15) 1514ms
   ✓ test bns contract namespace errors (2)
     ✓ should throw ERR_NAMESPACE_BLANK
     ✓ should throw ERR_NAMESPACE_NOT_FOUND
   ✓ preorder namespace (1)
     ✓ should preorder a namespace
   ✓ namespace reveal workflow (3)
     ✓ should reveal a namespace
     ✓ fails if the namespace is not revealed
     ✓ can launch a namespace
   ✓ name revealing workflow (5) 526ms
     ✓ should fail if no preorder
     ✓ should fail if stx burnt is too low
     ✓ should fail if existing pre-order
     ✓ should successfully register
     ✓ should fail registering twice
   ✓ register a name again before and after expiration (4) 523ms
     ✓ fails if someone else tries to register it
     ✓ should fail to register to register two names
     ✓ should allow registering a new name after first name expiration
     ✓ should allow someone else to register after expiration
```

### Checklist
N/A


hugocaillard:
cc @wileyj

contrib/core-contract-tests/tests/bns/name_register.test.ts
```diff
@@ -0,0 +1,597 @@
...
+import { beforeEach, describe, expect, it } from "vitest";
+import { createHash } from "node:crypto";
+
+const accounts = simnet.getAccounts();
+const alice = accounts.get("wallet_1")!;
+const bob = accounts.get("wallet_2")!;
+const charlie = accounts.get("wallet_3")!;
+
+const cases = [
+  {
```
moodmosaic, line 11: @hugocaillard, this may be a good candidate for a fast-check generator (in a future PR) 🤔

contrib/core-contract-tests/tests/bns/name_register.test.ts
```diff
@@ -0,0 +1,597 @@
...
+        Cl.bufferFromUtf8(cases[0].salt),
+        Cl.bufferFromUtf8(cases[0].zonefile),
+      ],
+      bob
+    );
+  });
+
+  it("fails if someone else tries to register it", () => {
+    const name = "bob";
+    let salt = "1111";
```
moodmosaic, line 464: This as well may be a good candidate for a fast-check generator (in a future PR).

moodmosaic (commented):
This looks good! If possible, it'd be nice if you can keep the current clarinet v1/deno tests and have them run side-by-side, for a few iterations at least, just in case the new tests accidentally introduce any false [positives](http://xunitpatterns.com/false%20positive.html)/[negatives](http://xunitpatterns.com/false%20negative.html) (since it's a fairly sized PR).

hugocaillard:
@moodmosaic 
Since we are testing a boot contract, which code isn't going to change, do you still think that we should keep both test files side-by-side?

moodmosaic:
>[..] code isn't going to change, do you still think that we should keep both test files side-by-side?

- If all the test cases from the old suite are covered *and*
- the test coverage report is the same *and*
- we have cross-referenced the old and new test outcomes to ensure consistency

then we can go ahead and merge 🎉

moodmosaic:
> see my comment here: #4030

Perhaps in the future, when needed, we can mix and match deno/clarinet v1 tests with clarinet-sdk.

hugocaillard:
> - If all the test cases from the old suite are covered and
> - the test coverage report is the same and
> - we have cross-referenced the old and new test outcomes to ensure consistency

I can't say that the last point has be strictly respected. I'm confident about the consistency of the new sdk, but can't prove it. I guess I should keep the old tests then.
For how long do you think we should keep both side by side?

Looks like I need to update the PR anyway, the CI doesn't pass

moodmosaic:
You can have them in for as long as clarinet/v1 can run as part of the CI build. This way you maintain an [append-only](https://blog.ploeh.dk/2013/04/02/why-trust-tests/#b23efe23c1574f2b8ff85fb3c529ec9d) test-suite keeping its trustworthiness level higher (and higher).

contrib/core-contract-tests/tests/bns/name_register.test.ts
hugocaillard: Maybe in a future PR. We could consider it as part of https://github.com/stacks-network/stacks-blockchain/issues/4030

contrib/core-contract-tests/tests/bns/name_register.test.ts
hugocaillard: Maybe in a future PR. We could consider it as part of https://github.com/stacks-network/stacks-blockchain/issues/4030

hugocaillard:
Changes done @moodmosaic 
I had to fight a bit with the CI to get both working properly but we're good now 👍 

> You can have them in for as long as clarinet/v1 can run as part of the CI build.

I think it's confusing to have this to similar but different files sitting next to each other.
I don't want an external contributor to be wondering what to do with this clarinet v1 test file in 6 or 12 months from now.

contrib/core-contract-tests/tests/bns/name_register_test.ts
```diff
@@ -1,3 +1,7 @@
+/* This Clarinet v1 tests have been migrated to the clarinet-sdk */
+/* This file is still executed in the CI in order to avoi false positives or negatives in the CI */
```
moodmosaic, line 2:
```suggestion
/* This file is still executed in the CI in order to avoid false positives or negatives in the CI */
```

contrib/core-contract-tests/vitest.config.js
```diff
@@ -0,0 +1,36 @@
...
+import { vitestSetupFilePath, getClarinetVitestsArgv } from "@hirosystems/clarinet-sdk/vitest";
+
+/*
+  In this file, Vitest is configured so that it works seamlessly with Clarinet and the Simnet.
+
+  The `vitest-environment-clarinet` will initialise the clarinet-sdk
+  and make the `simnet` object available globally in the test files.
+
+  `vitestSetupFilePath` points to a file in the `@hirosystems/clarinet-sdk` package that does two things:
+    - run `before` hooks to initialize the simnet and `after` hooks to collect costs and coverage reports.
```
moodmosaic, line 13:
```suggestion
    - run `before` hooks to initialize the Simnet and `after` hooks to collect costs and coverage reports.
```

contrib/core-contract-tests/vitest.config.js
```diff
@@ -0,0 +1,36 @@
...
+    - load custom vitest matchers to work with Clarity values (such as `expect(...).toBeUint()`)
```
moodmosaic, line 14:
```suggestion
    - load custom Vitest matchers to work with Clarity values (such as `expect(...).toBeUint()`).
```

moodmosaic (approved):
LGTM 💯 

---

Left a few nitpicks/comments. ✍️ 👀

contrib/core-contract-tests/vitest.config.js
hugocaillard: We usually refer to mainnet, testnet, devnet, and simnet with lower case in docs, blog posts and such

moodmosaic (approved):
LGTM 💯

contrib/core-contract-tests/vitest.config.js
moodmosaic: True, thanks 👍

wileyj (approved):
shipit

obycode (approved):
LGTM! Thanks @hugocaillard

hugocaillard:
The for the reviews @moodmosaic, @wileyj, @obycode!

PR #4039, opened 2023-11-07
Nakamoto: Validate stacker signature of a Nakamoto Block

jferrant:
### Description

Add functionality to check if a Nakamoto block has been correctly signed with the Signers (Stackers) WSTS signature associated with the aggregate public key. 

### Applicable issues
- fixes https://github.com/stacks-network/stacks-blockchain/issues/3927

### DIFF
--- a/stacks-common/src/util/secp256k1.rs
+++ b/stacks-common/src/util/secp256k1.rs
@@ -744,4 +744,79 @@ mod tests {
             runtime_verify - runtime_recover
         );
     }
+
+    #[test]
+    fn test_schnorr_signature_serde() {
+        use wsts::traits::Aggregator;
+
+        // Test that an empty conversion fails.
+        let empty_signature = SchnorrSignature::default();
+        assert!(empty_signature.to_wsts_signature().is_none());
+
+        // Generate a random Signature and ensure it successfully converts
+        let mut rng = rand_core::OsRng::default();
+        let msg =
+            "You Idiots! These Are Not Them! You\'ve Captured Their Stunt Doubles!".as_bytes();
+
+        let num_keys = 10;
+        let threshold = 7;
+        let party_key_ids: Vec<Vec<u32>> =
+            vec![vec![0, 1, 2], vec![3, 4], vec![5, 6, 7], vec![8, 9]];
+        let num_parties = party_key_ids.len().try_into().unwrap();
+
+        // Create the parties
+        let mut signers: Vec<wsts::v2::Party> = party_key_ids
+            .iter()
+            .enumerate()
+            .map(|(pid, pkids)| {
+                wsts::v2::Party::new(
+                    pid.try_into().unwrap(),
+                    pkids,
+                    num_parties,
+                    num_keys,
+                    threshold,
+                    &mut rng,
+                )
+            })
+            .collect();
+
+        // Generate an aggregate public key
+        let comms = match wsts::v2::test_helpers::dkg(&mut signers, &mut rng) {
+            Ok(comms) => comms,
+            Err(secret_errors) => {
+                panic!("Got secret errors from DKG: {:?}", secret_errors);
+            }
+        };
+        let aggregate_public_key = comms
+            .iter()
+            .fold(Point::default(), |s, comm| s + comm.poly[0]);
+
+        // signers [0,1,3] have "threshold" keys
+        {
+            let mut signers = [signers[0].clone(), signers[1].clone(), signers[3].clone()].to_vec();
+            let mut sig_agg = wsts::v2::Aggregator::new(num_keys, threshold);
+
+            sig_agg.init(comms.clone()).expect("aggregator init failed");
+
+            let (nonces, sig_shares, key_ids) =
+                wsts::v2::test_helpers::sign(msg, &mut signers, &mut rng);
+            let original_signature = sig_agg
+                .sign(msg, &nonces, &sig_shares, &key_ids)
+                .expect("aggregator sig failed");
+            // Serialize the signature and verify the results
+            let schnorr_signature = SchnorrSignature::from(&original_signature);
+            assert_eq!(
+                schnorr_signature[..33],
+                original_signature.R.compress().data[..]
+            );
+            assert_eq!(schnorr_signature[33..], original_signature.z.to_bytes());
+
+            // Deserialize the signature and verify the results
+            let reverted_signature = schnorr_signature
+                .to_wsts_signature()
+                .expect("Failed to convert schnorr signature to wsts signature");
+            assert_eq!(reverted_signature.R, original_signature.R);
+            assert_eq!(reverted_signature.z, original_signature.z);
+        }
+    }
 }
--- a/stacks-common/src/util/secp256k1.rs
+++ b/stacks-common/src/util/secp256k1.rs
@@ -13,8 +13,6 @@
 //
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
-use p256k1::point::Compressed;
 use rand::{thread_rng, RngCore};
 use secp256k1;
 use secp256k1::ecdsa::{
@@ -29,7 +27,8 @@ use serde::de::{Deserialize, Error as de_Error};
 use serde::ser::Error as ser_Error;
 use serde::Serialize;
 use wsts::common::Signature as WSTSSignature;
-use wsts::{Point, Scalar};
+use wsts::curve::point::{Compressed, Point};
+use wsts::curve::scalar::Scalar;
 
 use super::hash::Sha256Sum;
 use crate::impl_byte_array_message_codec;
@@ -817,6 +816,8 @@ mod tests {
                 .expect("Failed to convert schnorr signature to wsts signature");
             assert_eq!(reverted_signature.R, original_signature.R);
             assert_eq!(reverted_signature.z, original_signature.z);
+            assert!(original_signature.verify(&aggregate_public_key, msg));
+            assert!(reverted_signature.verify(&aggregate_public_key, msg));
         }
     }
 }
--- a/stacks-common/src/util/secp256k1.rs
+++ b/stacks-common/src/util/secp256k1.rs
@@ -14,6 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+use p256k1::point::Compressed;
 use rand::{thread_rng, RngCore};
 use secp256k1;
 use secp256k1::ecdsa::{
@@ -27,8 +28,11 @@ use secp256k1::{
 use serde::de::{Deserialize, Error as de_Error};
 use serde::ser::Error as ser_Error;
 use serde::Serialize;
+use wsts::common::Signature as WSTSSignature;
+use wsts::{Point, Scalar};
 
 use super::hash::Sha256Sum;
+use crate::impl_byte_array_message_codec;
 use crate::types::{PrivateKey, PublicKey};
 use crate::util::hash::{hex_bytes, to_hex};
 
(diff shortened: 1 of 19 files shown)


stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -1931,19 +1938,40 @@ impl<'a> SortitionHandleConn<'a> {
...
+        sortition_id: &SortitionId,
+    ) -> Result<Option<Point>, db_error> {
+        self.get_indexed(sortition_id, &db_keys::pox_stackers_aggregate_public_key())
+            .map(|x| {
+                x.map(|x| {
+                    let x = hex_bytes(&x).expect("CORRUPTION: bad stackers aggregate public key. Invalid hex string.");
+                    let mut x_bytes: [u8; 33] = [0; 33];
+                    if x_bytes.len() != x.len() {
+                        panic!("CORRUPTION: bad stackers aggregate public key. Expected {} bytes, got {} bytes", x_bytes.len(), x.len());
+                    }
```
obycode, line 1970: Could this be an `assert_eq!` instead?

obycode (approved):
LGTM! I just had one minor suggestion. 
May the Schwartz be with you.

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -1931,19 +1938,40 @@ impl<'a> SortitionHandleConn<'a> {
...
+                    x_bytes.copy_from_slice(&x[0..33]);
+                    Point::try_from(&Compressed::from(x_bytes)).expect("CORRUPTION: bad stackers aggregate public key")
+                })
+            })
```
kantai, line 1974: Not a huge issue, but I think in a case with nested `map()` invocations, it's probably more clear to flatten the nesting use the try syntax `?` instead:

```suggestion
        let Some(pk_string) = self.get_indexed(sortition_id, &db_keys::pox_stackers_aggregate_public_key())?
        else {
              return Ok(None)
        };
        let pk_byte_vec = hex_bytes(&pk_string).expect("CORRUPTION: bad stackers aggregate public key. Invalid hex string.");
        let mut pk_bytes: [u8; 33] = [0; 33];
        if pk_bytes.len() != pk_byte_vec.len() {
             panic!("CORRUPTION: bad stackers aggregate public key. Expected {} bytes, got {} bytes", x_bytes.len(), x.len());
        }
        pk_bytes.copy_from_slice(&pk_byte_vec[0..33]);
        Ok(Point::try_from(&Compressed::from(pk_bytes)).expect("CORRUPTION: bad stackers aggregate public key"))
```

kantai (approved):
LGTM -- only had a minor comment

jcnelson (changes requested):
I don't think we should be using `MessageSignature` to encapsulate a Schnorr signature. I was using that as a placeholder for a Schnorr-specific signature type.

If you can instead create a Schnorr signature container, and use that in place of `MessageSignature` in the block header, then I'll approve :+1:

stacks-common/src/util/secp256k1.rs
```diff
@@ -698,4 +725,79 @@ mod tests {
...
+        use wsts::traits::Aggregator;
+
+        // Test that an empty conversion fails.
+        let empty_signature = MessageSignature::empty();
+        assert!(empty_signature.to_schnorr_signature().is_none());
+
+        // Generate a random Signature and ensure it successfully converts
+        let mut rng = rand_core::OsRng::default();
+        let msg =
+            "You Idiots! These Are Not Them! You\'ve Captured Their Stunt Doubles!".as_bytes();
```
jcnelson, line 740: lol

stackslib/src/chainstate/burn/db/sortdb.rs
jcnelson: Yeah, it's generally a good idea to avoid excessive indentations

jferrant:
> I don't think we should be using `MessageSignature` to encapsulate a Schnorr signature. I was using that as a placeholder for a Schnorr-specific signature type.
> 
> If you can instead create a Schnorr signature container, and use that in place of `MessageSignature` in the block header, then I'll approve 👍

Done. Just working on adding "aggregate_public_key" to the RewardCycleInfo

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -1126,6 +1133,38 @@ impl StacksChainState {
...
+            );
+            return Ok(None);
+        }
+
+        let aggregate_public_key = self
+            .eval_boot_code_read_only(
+                sortdb,
+                block_id,
+                POX_4_NAME,
+                &format!("(get-aggregate-pubilc-key u{})", reward_cycle),
```
jferrant, line 1157: Fix typo in call

jcnelson: Is there a `get-aggregate-public-key` function in `pox-4`?  If not, can you add it?

Also, I don't think this is ever going to be optional in PoX-4.  Instead, this method should just return `Result<Point, Error>`.

jferrant: This could be optional for a given reward cycle as signers may not have set it yet no? In theory they should have set it BEFORE the start of their reward cycle, but they may fail to do so before the start.

jferrant:
I am also inclined in this PR to update stacker_signature and all references to it to signers_signature as technically stackers may have offloaded their responsibilities to some other signer operator. Does anyone have objections to this? @jcnelson @kantai

stackslib/src/chainstate/stacks/boot/mod.rs
jcnelson: > In theory they should have set it BEFORE the start of their reward cycle, but they may fail to do so before the start and be still executing a DKG round AFTER the start of a reward cycle.

Ah, yes, you're right. This has to be optional. The signers can't compute the aggregate public key until after DKG, so there's going to be a delay between calculating the reward set and setting the aggregate public key in `pox-4`.

However, I think you'll need to update `get_nakamoto_reward_cycle_info()` to return `None` if the aggregate public key is not known.  We should keep trying to compute the reward set until we have an aggregate public key.

jcnelson: > but they may fail to do so before the start and be still executing a DKG round AFTER the start of a reward cycle.

As of right now, if this happens, then the chain will freeze.  There's no way for the Stacks chain state to advance if the next reward cycle begins without the new aggregate public key. 

This could be addressed if there was a Stacks-on-Bitcoin way to set the aggregate public key, however.  I'll open a discussion for it in the SIP repo.

jcnelson: https://github.com/stacksgov/sips/discussions/162

jcnelson:
Sounds good to me :+1:

jferrant:
Waiting on https://github.com/stacks-network/stacks-core/pull/4009 to be merged
Waiting on the merge of Developer into Next
Will rebase once that is done

Adding testing now.

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -374,7 +386,8 @@ fn test_simple_nakamoto_coordinator_1_tenure_10_blocks() {
 /// Mine a 10 Nakamoto tenures with 10 Nakamoto blocks
 #[test]
 fn test_simple_nakamoto_coordinator_10_tenures_10_blocks() {
```
jferrant, line 388: This test is currently failing. Seems to be failing on third iteration of the loop. 

```
WARN [1700578532.327215] [stackslib/src/chainstate/burn/db/sortdb.rs:2038] [chainstate::nakamoto::coordinator::tests::test_simple_nakamoto_coordinator_10_tenures_10_blocks] reward start height 41 for reward cycle 8 does not have a sorition from 54538e85cabb2462772bb668b6a8d1e967d837c1060788fd289a295c93424c49
```

Jude has discovered that it is off by one, looking at start height 41 when it should be at height 40. Why this off by one only happens in this test on the third iteration is unknown still.

jferrant: Found part of the problem: find_prepare_phase_sortitions was incorrectly adding the sortition ID of one BEFORE the prepare phase. Still failing at replay (I think because it tries to access the aggregate public key when it is not set)

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -409,7 +434,13 @@ fn test_simple_nakamoto_coordinator_10_tenures_10_blocks() {
                         &recipient_addr,
                     );
 
-                    vec![stx_transfer]
+                    let aggregate_tx = make_pox_4_aggregate_key(
```
jcnelson, line 437: Why do we need to set the aggregate public key multiple times (e.g. in each block)?

stackslib/src/chainstate/nakamoto/tests/node.rs
```diff
@@ -63,6 +65,88 @@ use crate::net::test::{TestPeer, TestPeerConfig, *};
 use crate::util_lib::boot::boot_code_addr;
 use crate::util_lib::db::Error as db_error;
 
+#[derive(Debug, Clone)]
+pub struct TestSigners {
+    /// The parties that will sign the blocks
+    pub signer_parties: Vec<wsts::v2::Party>,
+    /// The commitments to the polynomials for the aggregate public key
+    pub comms: Vec<wsts::common::PolyCommitment>,
```
jcnelson, line 73: We use the word `comms` elsewhere to refer to "communication channels".  Can this member be renamed to something like `signer_commits` to remove ambiguity?

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -1125,6 +1127,39 @@ impl StacksChainState {
...
     }
+
+    /// Get the aggregate public key for a given reward cycle from pox 4
+    pub fn get_aggregate_public_key_pox_4(
+        &mut self,
+        sortdb: &SortitionDB,
+        block_id: &StacksBlockId,
+        reward_cycle: u64,
+    ) -> Result<Option<Point>, Error> {
+        if !self.is_pox_active(sortdb, block_id, reward_cycle as u128, POX_4_NAME)? {
```
jcnelson, line 1138: Nit: `u128::from(reward_cycle)`

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1318,3 +1322,18 @@
...
     )
 )
+
+;; What is the given reward cycle's stackers' aggregate public key?
+;; *New in Stacks 3.0*
+(define-read-only (get-aggregate-public-key (reward-cycle uint))
+    (map-get? aggregate-public-keys reward-cycle)
+)
+
+;; Set the aggregate public key to the provided value
```
jcnelson, line 1332: You don't need to remove this _per se_; you should instead just make it a private function, and have the node itself call it at the start of each PoX-4 reward cycle. You don't have to do this in this PR, but it does need to happen prior to Neon because this function should not be publicly reachable.

jcnelson: Please open a Neon-tagged issue for the above :pray:

jcnelson (approved):
LGTM; just minor nits to address before merging.  Thanks @jferrant for powering through this!

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
jcnelson: Oh I see, we still need to add a way for the node itself to do this as part of processing a reward cycle.

jferrant: It probably does not need to be EVERY block. I just decided lets do overkill to avoid missing processing it during replay. In theory this should not be needed as we simulate signer's submitting their votes across a reward cycle rather than a straight "set-aggregate-public-key" call. (this call should be removed once we have the ability to vote correctly)

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1385,6 +1418,59 @@ impl NakamotoChainState {
...
+        // Get the current reward cycle
+        let Some(reward_cycle) = sort_handle
+            .context
+            .pox_constants
+            .block_height_to_reward_cycle(
+                sort_handle.context.first_block_height,
+                ch_sn.block_height,
+            )
+        else {
+            // can't do anything
```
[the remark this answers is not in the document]
jferrant: Will do!

stackslib/src/chainstate/nakamoto/tests/node.rs
jferrant: Will do!

stackslib/src/chainstate/stacks/boot/pox-4.clar
jferrant: I have created and also adding as comment in code: https://github.com/stacks-network/stacks-core/issues/4101

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1385,6 +1418,59 @@ impl NakamotoChainState {
         Ok(true)
     }
 
+    /// Get the aggregate public key for the given block.
+    pub fn get_aggregate_public_key(
+        sortdb: &SortitionDB,
+        sort_handle: &SortitionHandleConn,
+        chainstate: &mut StacksChainState,
+        header: &NakamotoBlockHeader,
+        canonical_block_header: &StacksHeaderInfo,
```
kantai, line 1427: `canonical_block_header` is the block that the `aggregate_public_key` will be fetched at, which isn't necessary the "canonical chain tip" -- can you rename this variable to something like `fetch_at_block`?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1385,6 +1418,59 @@ impl NakamotoChainState {
         Ok(true)
     }
 
+    /// Get the aggregate public key for the given block.
+    pub fn get_aggregate_public_key(
+        sortdb: &SortitionDB,
+        sort_handle: &SortitionHandleConn,
+        chainstate: &mut StacksChainState,
+        header: &NakamotoBlockHeader,
```
kantai, line 1426: `header` doesn't really tell me much other than the type of the variable, perhaps, `for_block` ?

stackslib/src/net/relay.rs
```diff
@@ -711,10 +713,43 @@ impl Relayer {
             &block.header.block_hash()
         );
 
+        let Some(canonical_block_header) =
+            NakamotoChainState::get_canonical_block_header(chainstate.db(), &sortdb)?
```
kantai, line 717: I think this shouldn't be the `canonical_block_header`. It should be the last stacks block of the prepare phase or the parent of the stacks block being considered.

I don't think this needs to block this PR, though, so I opened an issue (#4109) -- can you just add a todo comment, with the issue number in it?

kantai (approved):
LGTM -- I think we should just flag a relevant issue for the block that the agg key is fetched from, I don't think it needs to block this PR.

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: I think that, specifically, this needs to be at least as high as the last block in the last reward cycle.

stackslib/src/net/relay.rs
jferrant: Will do!

Issue #4047, opened 2023-11-10
[Node] Reload config file every time it tries to mine a block


jcnelson:
The miner thread should load the miner's `satoshis_per_byte` field from the config file every time it builds a block.  Then, miners can adjust their fee rates on-the-fly without restarting their nodes.  They can even write a separate script that polls a 3rd party fee-rate service and writes this value to the config file.

AcrossfireX:
In order to make best use of the PR that would mitigate this issue, we need a system of determining what the appropriate fee is and also when to make an update to the config file so this does not need to be manually done by miners each step of the way as fees change. This can be done in a few steps that can be done by a script run periodically within cron.

1. Make an API request to a destination to get the latest fee estimation, this may be an electrum server or alternative endpoint that has the appropriate data.
2. Store that result as a variable
3. Use sed in bash to update the miner config file to have the appropriate value.

Once those are done, the manual process of updating is complete and the miner should not have to do anything to keep up to date with fee market dynamics.

Here is an example script that does this (this may need to be modified for your particular scenario).
```
#!/bin/bash
#REPLACE THIS PATH WITH YOUR MINER CONFIG FILE
MINER_CONFIG=/var/tmp/config


fee_estimate=`curl https://api.blockcypher.com/v1/btc/main | grep high_fee_per_kb | awk -F ':' '{print $2}' | cut -d',' -f 1 | xargs`

sed -i "s/satoshis_per_byte =.*/satoshis_per_byte = ${fee_estimate}/g" $MINER_CONFIG
echo "fee estimate updated"
```

Users may want to modify the endpoint used for fee estimates to use their own node and also may want to add additional logic to maybe only update the fees if its a significant enough change from what they already had set, but this may be helpful nonetheless

jcnelson:
This issue was closed by #4061.  I've moved @AcrossfireX's comment to a separate issue.

PR #4061, opened 2023-11-14
Feat: reload from config for up-to-date `satoshis_per_byte`

tippenein:
feat #4047 

introduces `get_satoshis_per_byte` fn for fetching most up to date value from config


tippenein:
I'm not sure if it's desirable to do file reads every time a function needs `spb` but maybe it's nbd?

wileyj:
@tippenein can you change the target branch to `develop`?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -162,6 +162,13 @@ pub fn make_bitcoin_indexer(config: &Config) -> BitcoinIndexer {
     burnchain_indexer
 }
 
+pub fn get_satoshis_per_byte(config: Config) -> u64 {
+    match config.get_burnchain_config() {
+        Ok(s) => s.satoshis_per_byte,
+        Err(_) => config.burnchain.satoshis_per_byte,
```
jcnelson, line 168: Might be worth logging an `info!()` error here

testnet/stacks-node/src/config.rs
```diff
@@ -48,6 +48,7 @@ const INV_REWARD_CYCLES_TESTNET: u64 = 6;
 
 #[derive(Clone, Deserialize, Default, Debug)]
 pub struct ConfigFile {
+    pub path: Option<String>,
```
jcnelson, line 51: Because this field is set at runtime, instead of via `toml::from_str()`, could you call it something that is unlikely to ever be set in the config file?  Something like `__path` instead of just `path` would work.

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -162,6 +162,13 @@ pub fn make_bitcoin_indexer(config: &Config) -> BitcoinIndexer {
     burnchain_indexer
 }
 
+pub fn get_satoshis_per_byte(config: Config) -> u64 {
```
jcnelson, line 165: Do you think this could be an instance method of `Config`, and take `&self`?

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
```diff
@@ -190,7 +197,7 @@ impl LeaderBlockCommitFees {
         let value_per_transfer = payload.burn_fee / number_of_transfers;
         let sortition_fee = value_per_transfer * number_of_transfers;
         let spent_in_attempts = 0;
-        let fee_rate = config.burnchain.satoshis_per_byte;
+        let fee_rate = get_satoshis_per_byte(config);
```
jcnelson, line 200: Because you're removing all instances of `config.burnchain.satoshis_per_byte`, can you take some preventative steps to make sure that `config.burnchain.satoshis_per_byte` cannot be accidentally accessed?  For example, is it possible to make `satoshis_per_byte` private?

jcnelson (commented):
Thank you for sending this PR @tippenein!  It looks pretty good to me; just had a few of small comments that are easily addressed.  EDIT: Also, I agree with @wileyj that this will need to be rebased on `develop`, since that's where all new non-consensus-breaking features land (i.e. we follow the gitflow branch model).

Do you think you could add some test coverage?  A unit test that verifies that `get_satoshis_per_byte()` works as intended would suffice.  Thanks!

jcnelson:
> @tippenein I'm not sure if it's desirable to do file reads every time a function needs spb but maybe it's nbd?

Given how small the config file is, and how often it'll be loaded, it's probably going to be resident in at least the kernel's block cache in perpetuity.  It's not a big deal at all :)

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
tippenein: There's one usage on the `LeaderBlockCommitFees` impl that prevented me from doing this

testnet/stacks-node/src/burnchains/bitcoin_regtest_controller.rs
tippenein: It's used as the backup in get_satoshis_per_byte, but there might be some way around it

tippenein: you _could_ load a default BurnchainConfig in the `Err` case of `get_satoshis_per_byte` instead of accessing the one in config.burnchain but it's not the same. I think it would be pretty weird to introduce a different config in the case it can't find the original, but it seems like such an edge case it might not matter?

I'm just thinking of a scenario where they run the miner with a spb of 100, then the config gets moved or deleted, get_satoshis_per_byte would look for it, then default to BurnchainConfig::default() and use 50, which is definitely not what they want

wileyj: > you _could_ load a default BurnchainConfig in the `Err` case of `get_satoshis_per_byte` instead of accessing the one in config.burnchain but it's not the same. I think it would be pretty weird to introduce a different config in the case it can't find the original, but it seems like such an edge case it might not matter?
> 
> I'm just thinking of a scenario where they run the miner with a spb of 100, then the config gets moved or deleted, get_satoshis_per_byte would look for it, then default to BurnchainConfig::default() and use 50, which is definitely not what they want

I don't think we need to worry about using a default burnchain config for a miner, since the miner will need to have a private key in order to broadcast btc txs. 
I would prefer an `Err` here if the config is moved or deleted.

jcnelson: Yeah same

jcnelson: Regarding my original question, I think what you have is fine.

jcnelson (approved):
LGTM!  Thanks for this PR!

kantai (approved):
This LGTM, but can you add a `CHANGELOG.md` entry?

jcnelson:
Congratulations on your first accepted PR @tippenein! :partying_face:

PR #4068, opened 2023-11-16
chore: Add alias for `cargo fmt` with custom config

jbencin:
### Description

As suggested by @cylewitruk, add `cargo fmt-stacks` alias to call `rustfmt` with our custom config. This PR also removes the Makefile with `make fmt` target, as having a Makefile is now unnecessary and misleading

### Applicable issues

Improves on #3907

Using a Cargo alias is necessary because these options can't be set in `.rustfmt.toml`. If this ever changes, we should add a `.rustfmt.toml` file and remove the alias.


CONTRIBUTING.md
```diff
@@ -386,22 +386,18 @@ A test should be marked `#[ignore]` if:
 
 PRs will be checked against `rustfmt` and will _fail_ if not properly formatted.
 Unfortunately, some config options that we require cannot currently be set in `.rustfmt` files currently, so arguments must be passed via the command line.
```
cylewitruk, line 388: double "currently"

cylewitruk (commented):
Will the change to the Makefile cause any problems for CI? Want to be sure of that before clicking the approve button :)

jbencin:
@cylewitruk Force pushed grammar fix and change to CI. See changes in `Dockerfile.rustfmt`. CI never used Makefile, so it won't break anything

cylewitruk (approved):
lgtm :)

kantai (approved):

Issue #2810, opened 2021-08-11
Duplicate events broadcasted to same API server


CharlieC3:
**Describe the bug**
When an API server is configured in both a `events_observer` block in the stacks-node's config TOML, and the `STACKS_EVENT_OBSERVER` environment variable, stacks-node will broadcast transactions to the target API server twice, causing it to fail over time.

**Steps To Reproduce**
1. Start a stacks-blockchain-api server
2. Export the `STACKS_EVENT_OBSERVER` env variable pointing to the API server.
3. Run stacks-node with a config TOML which has a `events_observer` block pointing to the API server

**Expected behavior**
The stacks-node should detect if there are duplicate event server entries between the `events_observer` block in the TOML and the `STACKS_EVENT_OBSERVER` env variable. If duplicate entries are found, print a log message saying so and exit.

**Environment (please complete the following information):**
stacks-node version `2.0.11.2.0`

[More info can be found here](https://github.com/blockstack/stacks-blockchain-api/issues/695#issuecomment-896344705).

gregorycoppola:
Sounds like something we should be doing!

pseudozach:
is there any way to get out of this? And does this lead to node sync to fail? I'm having this issue on my testnet node.

I've unset the env var but behavior is the same.

>ERRO [1631209353.116014] [testnet/stacks-node/src/event_dispatcher.rs:124] [chains-coordinator] Event dispatcher: POST 127.0.0.1:3700/http://127.0.0.1:3700/new_block failed with error Response { status: InternalServerError, headers: Headers { headers: {HeaderName("content-length"): [HeaderValue { inner: "12" }], HeaderName("date"): [HeaderValue { inner: "Thu, 09 Sep 2021 17:42:33 GMT" }], HeaderName("content-type"): [HeaderValue { inner: "application/json; charset=utf-8" }], HeaderName("x-powered-by"): [HeaderValue { inner: "Express" }], HeaderName("connection"): [HeaderValue { inner: "keep-alive" }], HeaderName("etag"): [HeaderValue { inner: "W/\"c-3Rk1bge0s6VuJbi+S2m2iU0UGdY\"" }], HeaderName("keep-alive"): [HeaderValue { inner: "timeout=5" }]} }, version: None, sender: Some(Sender { .. }), receiver: Receiver { .. }, body: Body { reader: "<hidden>", length: Some(12) }, local: TypeMap }

CharlieC3:
[...]

PR #4076, opened 2023-11-19
Fix: duplicate events

tippenein:
fix: #2810 

This simply uses a HashSet for `event_observers` which means it might not be exactly what the issue asked for but I'm putting it here for discussion. It wouldn't be particularly hard to WARN in the case of inserting an event observer which is already specified, but maybe it's sufficient to just disallow it?

(didn't add any tests because it ends up just testing HashSet's insert..)

### DIFF
--- a/clarity/src/vm/types/signatures.rs
+++ b/clarity/src/vm/types/signatures.rs
@@ -39,7 +39,7 @@ use crate::vm::types::{
 
 type Result<R> = std::result::Result<R, CheckErrors>;
 
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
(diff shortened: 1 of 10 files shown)


obycode (commented):
This change will just silently ignore a duplicate, but the issue requests that it errors and exits. This should be an easy addition. Can you add that @tippenein?

CharlieC3:
@obycode Re-thinking my original issue request, this may actually be a better solution since it still guards against user error while preserving flexibility to run the binary.
Unless you think otherwise, I'd be happy with this as-is!

obycode:
> @obycode Re-thinking my original issue request, this may actually be a better solution since it still guards against user error while preserving flexibility to run the binary. Unless you think otherwise, I'd be happy with this as-is!

Works for me!

obycode (approved):

jcnelson (approved):

PR #4079, opened 2023-11-20
Ci/faster tests

wileyj:
This PR reworks the entire CI workflow process in use today, and builds off of un-merged https://github.com/stacks-network/stacks-blockchain/pull/3847.

The two main issues it deals with are: test speed/reliability, with a secondary focus on making it easier to modify by moving a lot of the steps to a [composite actions](https://github.com/stacks-network/actions) monorepo.

Sample action (not all workflows are triggered in a feature branch): https://github.com/stacks-network/stacks-core/actions/runs/6936506358

## Reliability
Some tests/operations that need to run are always going to be flaky. To address, this PR moves all tests that are currently defined in a matrix and separates them into specific workflows based on *what* they are testing. i.e. "Atlas Tests" runs all the the tests related to Atlas. There is also a workflow for "Slow" tests - 2 specific tests which have been identified as taking longer than most others (roughly 20 minutes each for a successful test), and they fail to complete more than others. Additionally, a timeout of 30 minutes is being set for all tests (set via environment variable in the workflows).

## Speed:
The largest amount of work was done around speed. The current workflows contain several bottlenecks that have been removed - mainly, running jobs natively as opposed to the current method of using Docker.
The first change was moving from `cargo test` to using [nextest](https://nexte.st/), which has noticeable speed improvements in how long a test takes to run. 
The second change is removing the docker steps where a large image is created as an artifact, then shared amongst the many matrix jobs. To counter this, we are creating a cache of [nextest test archives](https://nexte.st/book/reusing-builds.html), as well as a cache of `cargo` and `target` which results in a size roughly 1/10th of the previous docker container used to run tests. 
The third change is modifying the tests themselves. Two approaches were taken here - separating the tests themselves into different re-suable workflows, and for the integration tests, using [nextest partitioning](https://nexte.st/book/partitioning.html) to run several tests concurrently (vs running them serially, as is the default).

## Easily Modifiable
Previous PR [3847](https://github.com/stacks-network/stacks-blockchain/pull/3847) only targeted how the builds and tests were run - there were a lot of workflows that had steps that were copy/pasted (requiring a small change to be copied to many workflows, i.e. update a version). 
This PR takes what was in 3847, and extends it further by moving a lot of the repeatable steps to an external monorepo of [composite actions](https://github.com/stacks-network/actions). This allows the workflows to become **much** simpler, and most CI related changes can now be done *outside of the blockchain repo*.

## Other improvements this PR brings:

- **Much less yml** since most of the work is now defined in the monorepo of [composite actions](https://github.com/stacks-network/actions)
- Easier to read conditionals for jobs/steps. This was an easy win, but it should be much more clear about what triggers a workflow. 
- Uploading to codecov.io is now being retried on failure, this should help with workflow errors where the upload fails without a clear reason (i.e. a 500 response)
- DRY the docker build workflow
    - single workflow will now build both alpine/debian images for release using a single workflow via a matrix of defined OS'es. there's room for improvement here, but it reduces the amount of yml needed. 
- Cleanup: one downside of the current workflow methods is that there is no cleanup performed, resulting in a large list of previous workflow runs. a configurable `cleanup.yml` is being proposed here to do a few things: 
    - Delete caches after 7 day
    - Delete failed workflows after 14 day
    - Delete *successful* workflows after 30 days
    - This workflow can be run manually, as well as via nightly schedule
- Manual test runs: in the current workflow, to run some tests it is required to trigger a full workflow run. with the `standalone.yml` workflow, we can now:
    - Run individual test suites when not triggered by a ci action
    - Run all tests that would run during a tagged release
    - Run all tests that would run during a PR action (opened, re-opened, synchronized)
- Removed some unnecessary/outdated jobs
- If a PR is **only** changing yml or markdown, no CI is run (this could possibly be extended to only trigger if `*.rs, *.clar` is modified)
    - Rf a PR contains code changes, then a later commit changes markdown, CI is triggered again due to the first commit
- Jobs are now triggered on PR changes:
    - Opened/re-opened PR
    - Synchronized PR (new commits after PR is opened)
    - PR approvals (will be triggered once per approval)
- Enables an option to build a docker image from source on demand, vs running the `ci.yml` workflow which will take longer due to tests

## Timing:
A lot of the timing hinges on how long it takes to create the cache (this is CPU bound).
It's also hard to definitively say a workflow will take x mins, due to how runners are allocated is out of our control. 
However, it will be possible to open a PR and have the workflows complete in roughly 30-60 minutes. For a tagged release, the timing is generally around 60 minutes, whereas a PR is in the neighborhood of 45 minutes. 

## Future work:
- Fork action repos from the https://github.com/actions-rs org: there are some good repos here that could be very useful, but they're not actively maintained
- It may be possible to simplify the tests where a matrix is used for a defined set of tests using [nextest filtering](https://nexte.st/book/filter-expressions.html). a POC of this worked well, but it can be tricky due to how the tests are partitioned (it may make the workflows harder to read)

## Ongoing issues

As with any tool, there will always be some things we simply cannot fix. Using github's cache drastically improves the workflow times, but it can happen where a cache does not restore onto the runner for several reasons (i.e. a cache download fails/stops). This is the reason the caches will persist (until the cleanup workflow runs), so they are available for re-running the failed jobs.


wileyj:
@hugocaillard since I see a few PR's related to CI that this may supersede or would benefit merging into my branch

.github/workflows/ci.yml
```diff
@@ -1,180 +1,223 @@
...
+  ##
+  ## Runs when:
+  ##   - tag is provided
+  ##   or:
+  ##     - no tag provided
+  ##       and any of:
+  ##         - PR is approved (any approval will trigger)
+  ##         - this workflow is called manually
+  ##         - PR is opened
+  ##         - commit to either (development, master) branch
```
jcnelson, line 134: Can you also add the `next` branch?

.github/workflows/ci.yml
```diff
@@ -1,180 +1,223 @@
...
+          (
+            github.event_name == 'pull_request_review' &&
+            github.event.action == 'submitted' &&
+            github.event.review.state == 'approved'
+          ) ||
+          github.event_name == 'workflow_dispatch' ||
+          github.event_name == 'pull_request_target' ||
+          (
+            contains('
+              refs/heads/develoment
```
jcnelson, line 148: I think this should be `refs/heads/develop`?

docs/ci-release.md
```diff
@@ -3,148 +3,184 @@
...
-   2. Docker image pushed with tags of the **input version** and **latest**
+   2. All release tests are run
+   3. Docker image pushed with tags of the **input version** and **latest**
+   - ex:
+   - `stacks-blockchain:2.1.0.0.0-debian`
+   - `stacks-blockchain:latest-debian`
+   - `stacks-blockchain:2.1.0.0.0`
+   - `stacks-blockchain:latest`
+
+## Tests
```
jcnelson, line 70: Can you add some guidance on where we should put new integration tests?  Perhaps a table that lists out all the test batches, and what kinds of tests would need to go into them?

jcnelson (dismissed):
Overall LGTM, but this isn't my wheelhouse.  Just a few minor concerns.

AshtonStephens (dismissed):
Made some comments initially that seem to be addressed. LGTM, lets monitor this for a bit to make sure that we didn't introduce anything.

.github/workflows/ci.yml
wileyj: fixed, good catch!

docs/ci-release.md
wileyj: This will take some time to document, but it's a good idea. i'll make it so.

.github/workflows/bitcoin-tests.yml
```diff
@@ -1,205 +1,100 @@
...
-        id: git_checkout
-        uses: actions/checkout@v3
-      - name: Download docker image
-        id: download_docker_image
-        uses: actions/download-artifact@v3
+      ## Run test matrix using restored cache of archive file
+      ##   - Test will timeout after env.TEST_TIMEOUT minutes
+      - name: Run Tests
+        id: run_tests
+        timeout-minutes: ${{ fromJSON(env.TEST_TIMEOUT) }}
```
CharlieC3, line 89: Curious, does `${{ env.TEST_TIMEOUT }}` not work here? `TEST_TIMEOUT` [does not appear to be a JSON object](https://github.com/stacks-network/stacks-core/pull/4079/files#diff-0e6368aea5b3b1595c45133e133576c67450c749bd978b7b35b284c94a2746e9R19).

.github/workflows/bitcoin-tests.yml
```diff
@@ -1,205 +1,100 @@
...
 on:
-  pull_request:
-    types:
-      - opened
-      - reopened
+  workflow_call:
+
+defaults:
+  run:
+    shell: bash
```
CharlieC3, line 10: Is this necessary or just for clarity? Generally I tend to favor excluding explicit default settings to reduce LoC and perceived complexity (I know this isn't _technically_ the default, but it's very close to it). It's the ol' convention over configuration adage.

.github/workflows/github-release.yml
```diff
@@ -1,50 +1,75 @@
...
-        description: "Stringified JSON object listing of platform matrix"
+        description: "Release Tag"
         required: true
         type: string
     secrets:
       GH_TOKEN:
         required: true
 
+concurrency:
+  group: github-release-${{ github.head_ref || github.ref }}
```
CharlieC3, line 17: Is this intended to be:
```suggestion
  group: github-release-${{ github.head_ref || github.ref || github.run_id}}
```

 following the concurrency pattern set in your other workflows?

.github/workflows/workflow-cleanup.yml
```diff
@@ -0,0 +1,47 @@
...
+    - cron: "0 0 * * *"
+
+## env vars are transferred to composite action steps
+env:
+  CACHE_TTL: 7 ## number of days to keep a cache
+  WORKFLOW_TTL: 30 ## number of days to keep a successful workflow
+  FAILED_WORKFLOW_TTL: 15 ## number of days to keep a failed workflow
+
+concurrency:
+  group: cleanup-${{ github.head_ref || github.ref }}
```
CharlieC3, line 31: Is this intended to be:
```suggestion
  group: cleanup-${{ github.head_ref || github.ref || github.run_id}}
```
 following the concurrency pattern set in your other workflows?

.github/workflows/ci.yml
```diff
@@ -1,180 +1,225 @@
...
+    types:
+      - submitted
+
+defaults:
+  run:
+    shell: bash
 
 concurrency:
-  group: ${{ github.head_ref || github.run_id }}
+  group: ci-${{ github.head_ref || github.ref }}
```
CharlieC3, line 39: Is this intended to be:
```suggestion
  group: ci-${{ github.head_ref || github.ref || github.run_id}}
```
 following the concurrency pattern set in your other workflows?

.github/workflows/image-build-binary.yml
```diff
@@ -0,0 +1,93 @@
...
+      ## Set docker metatdata
+      ## - depending on the matrix.dist, different tags will be enabled
+      ## ex. alpine will have this tag: `type=ref,event=tag,enable=${{ matrix.dist == 'alpine' }}`
+      - name: Docker Metadata ( ${{matrix.dist}} )
+        id: docker_metadata
+        uses: docker/metadata-action@96383f45573cb7f253c731d3b3ab81c87ef81934 #v5.0.0
+        with:
+          images: |
+            ${{env.docker-org}}/${{ github.event.repository.name }}
+            ${{env.docker-org}}/stacks-core
```
CharlieC3, line 69: Correct me if I'm wrong, but I believe this will result in a list of:
```
stacks-network/stacks-core
stacks-network/stacks-core
```

Instead, did you intend this to continue pushing to the old Docker repo?

```suggestion
            ${{env.docker-org}}/${{ github.event.repository.name }}
            ${{env.docker-org}}/stacks-blockchain
```

.github/workflows/image-build-source.yml
```diff
@@ -0,0 +1,69 @@
...
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+
+      ## Set docker metatdata
+      - name: Docker Metadata ( ${{matrix.dist}} )
+        id: docker_metadata
+        uses: docker/metadata-action@96383f45573cb7f253c731d3b3ab81c87ef81934 #v5.0.0
+        with:
+          images: |
+            ${{env.docker-org}}/${{ github.event.repository.name }}
+            ${{env.docker-org}}/stacks-core
```
CharlieC3, line 50: Same here, I believe this will result in a list of:
```
stacks-network/stacks-core
stacks-network/stacks-core
```

Instead, did you intend this to continue pushing to the old Docker repo?

```suggestion
            ${{env.docker-org}}/${{ github.event.repository.name }}
            ${{env.docker-org}}/stacks-blockchain
```

.github/workflows/atlas-tests.yml
```diff
@@ -0,0 +1,61 @@
...
+
+## env vars are transferred to composite action steps
+env:
+  BITCOIND_TEST: 1
+  RUST_BACKTRACE: full
+  SEGMENT_DOWNLOAD_TIMEOUT_MINS: 15
+  RETRIES: 3
+  RETRY_DELAY: 10000
+  TEST_TIMEOUT: 30
+  TEST_RETRIES: 2
```
CharlieC3, line 20: What's making use of the `RETRIES`, `RETRY_DELAY`, and `TEST_RETRIES` vars? I see [some of your composite actions](https://github.com/stacks-network/actions/blob/main/stacks-core/run-tests/action.yml) have the ability to accept these are parameters, but they don't seem to be getting passed along. Any insight here is appreciated!

.github/workflows/standalone-tests.yml
CharlieC3, line 1: just wanted to comment this is very cool and a great add!

.github/workflows/workflow-cleanup.yml
```diff
@@ -0,0 +1,47 @@
...
+        description: "How many days to keep a successful workflow (default: 30)"
+        required: false
+        default: "30"
+      failed-workflow-ttl:
+        description: "How many days to keep failed workflows (default: 15)"
+        required: false
+        default: "15"
+  schedule:
+    ## Run every day at 00:00:00
+    - cron: "0 0 * * *"
```
CharlieC3, line 22: IIRC, the cron respects UTC. So this will run at 7PM ET and 4PM PT. Should this be pushed back a bit to account for that considering most core engineers are around PT/ET?

build-scripts/Dockerfile.linux-glibc-arm64
```diff
@@ -24,3 +24,4 @@ RUN --mount=type=tmpfs,target=${BUILD_DIR} cp -R /src/. ${BUILD_DIR}/ \
 
 FROM scratch AS export-stage
 COPY --from=build /out/stacks-inspect /out/blockstack-cli /out/clarity-cli /out/stacks-node /
+## comment
```
CharlieC3, line 27: Should this be removed?

build-scripts/Dockerfile.linux-glibc-x64
```diff
@@ -21,3 +21,4 @@ RUN --mount=type=tmpfs,target=${BUILD_DIR} cp -R /src/. ${BUILD_DIR}/ \
 
 FROM scratch AS export-stage
 COPY --from=build /out/stacks-inspect /out/blockstack-cli /out/clarity-cli /out/stacks-node /
+## comment
```
CharlieC3, line 24: Same here, unintended?

docs/ci-release.md
```diff
@@ -3,148 +3,184 @@
...
+
+1. [Github Actions matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs)
+2. [nextest partitioning](https://nexte.st/book/partitioning.html)
 
-## PR a branch to develop:
+A matrix is used when there are several known tests that need to be run. Partitions (shards) are used when there is a large and unknown number of tests to run (ex: `cargo test` to run all tests).
 
-ex: Branch is named `feat/112-fix-something` and the PR is numbered `112`
+There is also a workflow designed to run tests that are manually triggered: [Standalone Tests](../.github/workflows/standalone-tests.yml).
+This workflow requires you to select which test(s) you want to run, which then triggers a reusbale workflow via conditional. For example, selecting "Epoch Tests" will run the tests defined in [Epoch Tests](../.github/workflows/epoch-tests.yml). Likewise, selecting `Release Tests` will run the same tests as a release workflow.
```
CharlieC3, line 82:
```suggestion
This workflow requires you to select which test(s) you want to run, which then triggers a reusable workflow via conditional. For example, selecting "Epoch Tests" will run the tests defined in [Epoch Tests](../.github/workflows/epoch-tests.yml). Likewise, selecting `Release Tests` will run the same tests as a release workflow.
```

CharlieC3 (commented):
Great PR! Excited to see these new and improved workflows being used daily. I left a handful of comments/questions/suggestions. Nothing major.

.github/workflows/bitcoin-tests.yml
wileyj: correct! it's a quirk of yaml and how github actions works. 
since the env var value is being sent to a composite job, it's sending the value as a string (with `"` wrapped around the int). 
since an int with quotes is treated as a string, the [composite aciton](https://github.com/stacks-network/actions/blob/main/stacks-core/run-tests/action.yml) will fail (i suppose an alternative is to have it defined as an input, but it seems unnecessary to add that in the composite action)

the only way i found to not hardcode that timeout in the composite action was to treat it as a json object. 
i could not tell you where i found the documentation on that either - i found it, tried it and it worked so i moved on 😄

.github/workflows/bitcoin-tests.yml
wileyj: it's technically not needed - but i went for verbosity here since github may change the runner environment at any time (today that `run:` would also work on ksh/zsh etc). 
i have no strong opinion here, the shell needs to be defined in the composite action, but since we're not calling `run:` here anymore, i'll remove it.

.github/workflows/github-release.yml
wileyj: i'll have to check - it may be a typo

.github/workflows/image-build-binary.yml
wileyj: i believe you are correct - the repo name change threw a wrench in things. thanks!

.github/workflows/workflow-cleanup.yml
wileyj: it is UTC, correct. 
It should not matter *when* it runs though, the outcome is still the same and i feel the defaults defined in https://github.com/stacks-network/actions/blob/main/cleanup/workflows/action.yml#L12-L23  are generous enough to not interfere with any current work. 

I expect this workflow will have some of the strongest opinions, so i'm happy to adjust/disable, or even change it to a manual trigger only. 

to be clear, this workflow only runs to clean up old actions/failed actions/caches:
- caches (excepting bitcoin binary are kept for 7 days ) *in the previous PR, you noted these will be rotated as space is needed as well. i saw no reason to remove this step, but it's debatable if it's needed*
- failed workflows are kept for 15 days
- successful workflows are kept for 30 days

build-scripts/Dockerfile.linux-glibc-arm64
wileyj: 💯 this was an artifact from my fork testing a non-text update (dockerfile counts as code since it's not .md/.yml)

build-scripts/Dockerfile.linux-glibc-x64
wileyj: indeed so, will address

.github/workflows/bitcoin-tests.yml
CharlieC3: Interesting! TIL, thanks!

.github/workflows/atlas-tests.yml
wileyj: leftover from previous PR - i'll remove them. they're still *valid* but they should be inputs
ex: 
```yml
      ## Create and upload code coverage file
      - name: Code Coverage
        id: codecov
        uses: stacks-network/actions/codecov@main
        with:
          test-name: ${{ matrix.test-name }}
          retries: 5 # default is 3
          retry-delay: 50000 # default is 10000
  ```

kantai (dismissed):
This is awesome @wileyj !

This LGTM, and I'm _very_ excited that this separates a ton of the CI logic into a different repo. The branching structure of the `stacks-core` repo makes changes to the CI pretty painful.

docs/ci-release.md
wileyj: added to the `./docs/ci-release.md` file - i don't think i can answer *where* a dev may want to put a test, but i tried to explain why the current tests are are in each file. 
https://github.com/stacks-network/stacks-core/blob/ci/faster-tests/docs/ci-release.md?plain=1#L74-L136

wileyj:
> This is awesome @wileyj !
> 
> This LGTM, and I'm _very_ excited that this separates a ton of the CI logic into a different repo. The branching structure of the `stacks-core` repo makes changes to the CI pretty painful.

thanks! 
i'm about 99% ready to merge; there has been a new development in my testing fork that i feel is important to understand first though. 
tl;dr - 
for some reason, the test i have defined in the `./github/workflows/slow-tests.yml` workflow: `tests::epoch_21::test_pox_reorg_flap_reward_cycles` has been consistently failing (timing out) in my fork and it's not clear why this happening. working to understand it before a merge can happen.

kantai (approved):

jcnelson (approved):

PR #4083, opened 2023-11-22
fix: Address warning messages and errors in `cargo build` and `cargo build`

jbencin:
### Description

Fix compiler warnings while running `cargo build` and `cargo test`

Also fix doc comment in `stackslib/src/net/stackerdb/config.rs` that was incorrectly interpreted as a doc test and failing, causing an error


diwakergupta (approved):
changes are pretty small, so I took the liberty of reviewing

obycode (approved):
lgtm!

PR #4084, opened 2023-11-22
feat: Add file for `/v2/block_proposal` endpoint

jbencin:
### Description

Add `/v2/block_proposal` endpoint which will be used by miners to validate proposed Nakamoto blocks

This replaces #3879, which has diverged too far from `next` to be merged/rebased (old HTTP stack, didn't handle `NakamotBlock`, etc)

### Applicable issues
- fixes #

### Checklist
- [x] Changelog is updated


stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,281 @@
...
+            chain_id: read_next(fd)?,
+        })
+    }
+}
+
+/// This enum is used to supply a `reason_code` for validation
+///  rejection responses. This is serialized as an enum with string
+///  type (in jsonschema terminology).
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum ValidateRejectCode {
```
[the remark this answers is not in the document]
jbencin: The way it works now is that we only respond to requests from `localhost`, so we know the miner is trusted. This also means that to run a Nakamoto miner, you have to run stacks-node

Maybe later we'll add support for signing to remove this restriction

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,281 @@
...
+};
+
+/// Represents a block proposed to the `v2/block_proposal` endpoint for validation
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct NakamotoBlockProposal {
+    /// Proposed block
+    pub block: NakamotoBlock,
+    /// Identify the stacks/burnchain fork we are on
+    pub parent_consensus_hash: ConsensusHash,
+    /// These are all the microblocks that the proposed block
```
jcnelson, line 65: I don't think we're doing microblocks anymore

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,281 @@
...
+pub struct BlockValidateReject {
+    pub reason: String,
+    pub reason_code: ValidateRejectCode,
+}
+
+/// A response for block proposal validation
+///  that the stacks-node thinks is acceptable.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct BlockValidateOk {
+    pub block: StacksBlock,
```
jcnelson, line 121: Should this be NakamotoBlock?

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub block_proposal: Option<NakamotoBlockProposal>,
+}
+
+impl RPCBlockProposalRequestHandler {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Decode a bare transaction from the body
+    fn parse_posttransaction_octets(mut body: &[u8]) -> Result<NakamotoBlockProposal, Error> {
```
jcnelson, line 146: "posttransaction"?

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,281 @@
...
+            ));
+        }
+
+        match preamble.content_type {
+            None => {
+                return Err(Error::DecodeError(
+                    "Missing Content-Type for transaction".to_string(),
+                ));
+            }
+            Some(HttpContentType::Bytes) => {
```
jcnelson, line 198: Copy/paste job? ;)

stackslib/src/net/api/postblock_proposal.rs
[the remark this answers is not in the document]
jferrant: I am more from the perspective of a caller to this endpoint could provide a block with an invalid signature. A stacks signer could be provided with a BlockProposal from an invalid miner and wish to validate it. I am writing the signer end and can add this signature validation check locally if its better there? I guess this endpoint is meant more from a "syntactically" correct BlockProposal rather than any sort of signature validation? @jcnelson ?

stackslib/src/net/api/postblock_proposal.rs
jbencin: Yeah, I removed that but forgot the comment

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// This refers to the burn block that was the current tip
+    ///  at the time this proposal was constructed. In most cases,
+    ///  if this proposal is accepted, it will be "mined" in the next
+    ///  burn block.
+    pub burn_tip_height: u32,
+    /// Mainnet, Testnet, etc.
+    pub chain_id: u32,
+}
+
+impl StacksMessageCodec for NakamotoBlockProposal {
```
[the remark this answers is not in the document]
jbencin: I don't know, but I added support because it was trivial to do. Probably slightly faster to use the binary format

stackslib/src/net/api/postblock_proposal.rs
jbencin: Yes

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,281 @@
...
+        )
+        .expect("FATAL: failed to construct request from infallible data")
+    }
+}
+
+impl StacksHttpResponse {
+    pub fn decode_stacks_block_proposal_accepted(
+        self,
+    ) -> Result<NakamotoBlockProposalResponse, NetError> {
+        todo!()
```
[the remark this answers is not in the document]
jbencin: Probably, I'm just getting back into this today, I have to figure out what I was doing again. I'll let you know

stackslib/src/net/api/postblock_proposal.rs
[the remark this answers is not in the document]
jbencin: It does this now in the latest commit. Basically, this endpoint will do two things with a block proposal:
- Call `validate_nakamoto_block_burnchain()` to check that the block is well-formed and a valid Nakamoto block (this is where the miner signature is validate). This function already exists in `chainstate/nakamoto/mod.rs`
- Validate the blocks transactions against the current chainstate to make sure they are valid. I still have to do this...

jbencin:
Hey @jcnelson and @kantai, I have the code for `NakamotoBlockProposal::validate()` written, I would appreciate a quick look to see if you think this the right approach to validate a block

I'm pretty confident I can write the tests for this on my own, but I may need help making the response async (though this could be done in another PR)

jbencin:
Marking this as ready for review even though I'm sure it needs a few more test cases

docs/rpc/openapi.yaml
```diff
@@ -551,3 +551,41 @@ paths:
...
+      description: |
+        Used by miner to validate a proposed Stacks block.
+        Can accept either JSON or binary encoding.
+        
+        **This endpoint will only accept requests over the local loopback network interface.**
+      responses:
+        202:
+          description: Block proposal is valid
+        400:
+          description: Bad request
```
jcnelson, line 570: Can you enumerate the reasons why this error will be reached?

docs/rpc/openapi.yaml
```diff
@@ -551,3 +551,41 @@ paths:
...
+        403:
+          description: Request not over loopback interface
+        429:
+          description: Too many requests
```
jcnelson, line 574: Same here -- how many is too many, and what should the caller do if they hit this?

stackslib/src/chainstate/nakamoto/miner.rs
```diff
@@ -833,3 +835,252 @@ impl BlockBuilder for NakamotoBlockBuilder {
...
+    }
+}
+
+impl From<DBError> for BlockValidateReject {
+    fn from(value: DBError) -> Self {
+        Error::from(value).into()
+    }
+}
+
+/// Represents a block proposed to the `v2/block_proposal` endpoint for validation
```
jcnelson, line 901: Is there a reason this isn't JSON?  Seems like that would be a lot friendlier for 3rd party mining software

jcnelson: Not saying you have to change it; just curious why only the `StacksMessageCodec`-specific representation is valid.

stackslib/src/chainstate/nakamoto/miner.rs
```diff
@@ -833,3 +835,252 @@ impl BlockBuilder for NakamotoBlockBuilder {
...
+    /// either accept or reject the proposal
+    ///
+    /// This is done in 2 steps:
+    /// - Static validation of the block, which checks the following:
+    ///   - Block is well-formed
+    ///   - Transactions are well-formed
+    ///   - Miner signature is valid
+    /// - Validation of transactions by executing them agains current chainstate.
+    ///   This is resource intensive, and therefore done only if previous checks pass
+    pub fn validate(
```
jcnelson, line 941: This is duplicating lots of logic in `NakamotoChainState::process_next_nakamoto_block()`. Can you instead just make it so that method can optionally (1) take a particular block to process as an argument, and (2) just return the receipts _without_ committing the new chainstate?  Then, you can just call that directly instead of duplicating the same checks.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -503,7 +503,7 @@ fn pox_extend_transition() {
     for cycle_number in first_v4_cycle..(first_v4_cycle + 4) {
         let cycle_start = burnchain.reward_cycle_to_block_height(cycle_number);
         let reward_set_entries = get_reward_set_entries_at(&mut peer, &latest_block, cycle_start);
-        info!("----- {cycle_number} -----");
+        //info!("----- {cycle_number} -----");
```
jcnelson, line 506: Please don't commit commented-out code ;)

stackslib/src/net/api/postblock_proposal.rs
jcnelson: Reupping this.  This method name is wrong.  You're not posting a transaction.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -18,34 +18,40 @@ use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 use std::{env, thread};
 
+use clarity::vm::ast::ASTRules;
 use clarity::vm::costs::ExecutionCost;
 use clarity::vm::types::PrincipalData;
 use lazy_static::lazy_static;
+//use stacks::burnchains::tests::TestMinerFactory;
```
jcnelson, line 25: Commented-out code

testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -490,3 +514,250 @@ fn simple_neon_integration() {
...
+
+    // ----- Setup boilerplate finished, test block proposal API endpoint -----
+
+    let tip = NakamotoChainState::get_canonical_block_header(chainstate.db(), &sortdb)
+        .unwrap()
+        .unwrap();
+
+    let privk = conf.miner.mining_key.unwrap().clone();
+    let parent_block_id = tip.index_block_hash();
+    // TODO: Get current `total_burn` from somewhere
```
jcnelson, line 629: Is this still a todo?

testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -490,3 +514,250 @@ fn simple_neon_integration() {
...
+    // Construct a valid proposal. Make alterations to this to test failure cases
+    let proposal = NakamotoBlockProposal {
+        block,
+        tenure_start_block: parent_block_id,
+        chain_id: chainstate.chain_id,
+    };
+
+    const HTTP_ACCEPTED: u16 = 202;
+    const HTTP_BADREQUEST: u16 = 400;
+    // TODO: Check error codes?
```
jcnelson, line 702: Yes, please do :pray:

testnet/stacks-node/src/tests/nakamoto_integrations.rs
jbencin: Yeah, I just pulled this number from an error message

stackslib/src/net/api/postblock_proposal.rs
jbencin: Will fix

stackslib/src/chainstate/nakamoto/miner.rs
jbencin: It derives Serde `Serialize` and `Deserialize` for JSON as well. The see the integration tests, they submit `NakamotoBlockProposal`s as JSON.

stackslib/src/chainstate/nakamoto/miner.rs
jbencin: Sure

docs/rpc/openapi.yaml
jbencin: I copied this from the previous PR, I don't think it makes sense to implement now, but when we implement async request handling we could add a max request queue depth that is set in the config file

stackslib/src/net/api/postblock_proposal.rs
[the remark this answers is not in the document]
kantai: I'd just suggest deleting this function. The stacks-node itself isn't handling these responses, so I don't see any need to implement this.

stacks-common/src/types/net.rs
```diff
@@ -218,6 +218,21 @@ impl PeerAddress {
...
+            // 127.0.0.0/8
+            self.0[12] == 127
+        } else {
+            // ::1/128
+            *self
+                == PeerAddress([
+                    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+                    0x00, 0x00, 0x01,
+                ])
+        }
```
kantai, line 233: I realize its less efficient to do:

```rust
pub fn is_loopback(&self) -> bool {
     self.to_socketaddr(0).ip().is_loopback()
}
```

But I think its preferable to not implement our own IPv4 and IPv6 spec. In general, I think `PeerAddress` should just be replaced by `std::net::IpAddr`

stackslib/src/chainstate/nakamoto/miner.rs
```diff
@@ -833,3 +835,252 @@ impl BlockBuilder for NakamotoBlockBuilder {
...
+    pub block: NakamotoBlock,
+    // tenure ID -- this is the index block hash of the start block of the last tenure (i.e.
+    // the data we committed to in the block-commit).  If this is an epoch 2.x parent, then
+    // this is just the index block hash of the parent Stacks block.
+    pub tenure_start_block: StacksBlockId,
+    /// Identifies which chain block is for (Mainnet, Testnet, etc.)
+    pub chain_id: u32,
+}
+
+impl StacksMessageCodec for NakamotoBlockProposal {
```
kantai, line 914: I'd recommend just deleting the `StacksMessageCodec` implementation here. Having two representations of the serialization for the proposal seems unnecessary, and JSON via serde is preferable.

docs/rpc/openapi.yaml
kantai: I think the answer here is "There is an ongoing request being evaluated. The `stacks-node` will only validate one block proposal at a time and will not queue such requests. The caller must wait until the ongoing request is completed."

kantai: I think this isn't a reachable code until the async implementation is in place.

stackslib/src/chainstate/nakamoto/miner.rs
jbencin: Pushed my work to `feat/nakamoto-block-proposal-refactor-validate` if you want to check it out (doesn't compile)

stackslib/src/chainstate/nakamoto/miner.rs
```diff
@@ -833,3 +835,252 @@ impl BlockBuilder for NakamotoBlockBuilder {
...
+    fn consensus_deserialize<R: Read>(fd: &mut R) -> Result<Self, CodecError> {
+        Ok(Self {
+            block: read_next(fd)?,
+            tenure_start_block: read_next(fd)?,
+            chain_id: read_next(fd)?,
+        })
+    }
+}
+
+impl NakamotoBlockProposal {
```
jcnelson, line 930: Please move this impl to your block proposal request handler

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,220 @@
...
+        preamble: HttpRequestPreamble,
+        _contents: HttpRequestContents,
+        node: &mut StacksNodeState,
+    ) -> Result<(HttpResponsePreamble, HttpResponseContents), NetError> {
+        let block_proposal = self
+            .block_proposal
+            .take()
+            .ok_or(NetError::SendError("`block_proposal` not set".into()))?;
+
+        let res = node.with_node_state(|_network, sortdb, chainstate, _mempool, _rpc_args| {
```
jcnelson, line 165: You may need to modify this function `with_node_state()` to pass `&mut SortitionDB`.  Alternatively, you can add a new function `with_node_state_mut()` which passes `&mut` references for all of these things.

jcnelson: You may need to tweak `PeerNetwork::run()` to take a `&mut SortitionDB` as well

jcnelson: Actually, forget I said this.  Let's try something else.

stackslib/src/chainstate/nakamoto/miner.rs
jcnelson: Actually, thinking about this a bit more, I think what you have here is fine for now.  We'll refactor `NakamotoChainState::process_next_nakamoto_block()` so this code can crib from it without needing a `&mut SortitionDB`.  `NakamotoChainState::process_next_nakamoto_block()` has lots of places where it can write data, so it doesn't make sense to use it here as-is since this function is not meant to ever write data to the chainstate.  Also, this function shouldn't open a transaction against any DBs if we can help it.

stacks-common/src/types/net.rs
jbencin: I agree, I was just following the patterns. Is there some reason we're not using Rust's standard lib here?

stackslib/src/chainstate/nakamoto/miner.rs
jbencin: I'm not sure I even need the `NakamotoBlockProposal` struct anyway. I've already removed most of the fields from the original `BlockProposal` struct in Subnets. I suppose I could just look up the most recent `tenure_start_block`, and a block proposal for the wrong `chain_id` would be invalid anyways. Would be simpler for miners if the endpoint just takes `NakamotoBlock`. What do you think?

stackslib/src/net/api/postblock_proposal.rs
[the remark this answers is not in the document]
jbencin: Deleted. Not sure what that was supposed to be, was copying patterns from other files, but it was never called, so probably don't need it

docs/rpc/openapi.yaml
jbencin: This specific error just means that what was submitted wasn't a valid block proposal (would not be accepted be a Stacks node). I changed it to make that more clear, but there are too many things that could make a block invalid to enumerate them individually

stacks-common/src/types/net.rs
jbencin: @kantai Do you want this change now, or as part of a larger refactor to remove `PeerAddress`? Kinda looks weird having one function convert to `IpAddr` while similar ones don't

kantai: Yes, I'd prefer the change now -- no reason to add our own implementation of a loopback check.

stackslib/src/chainstate/nakamoto/miner.rs
kantai: I think it's best to keep a `NakamotoBlockProposal` struct even if it just contains `block` for a couple reasons:

1. The JSON interface will make it less painful in the future if we decide there's some other information that's important to supply to the proposal validation.
2. Having the `validate()` method be a member of a new proposal struct avoids continuing the expansion of some already fairly unwieldy interfaces.

jbencin: I'm removing `tenure_start_block` because that isn't used by `NakamotBlockBuilder` anymore. So it'll just have `block` and `chain_id` for now

> The JSON interface will make it less painful in the future if we decide there's some other information that's important to supply to the proposal validation

I'm having a hard time thinking of anything that might be useful for block validation but not already contained in the block. That's the whole point of a blockchain, isn't it, to be able to validate the entire history from just the blocks?

jbencin:
Merged as part of #4228

Issue #4059, opened 2023-11-14
[stacks-core] Update .pox-4 to calculate signer key vote shares


jferrant:
Given the map of signer-key (public key) to an amount of stacked stx, calculate the number of vote shares allocated to that signer. Note that the vote power should be capped at 25% and the remaining percentage normalized over the other signers.

kantai:
Something related to this that occurs to me is the consequences of participation in the DKG process. The DKG process _cannot_ require 100% participation. Otherwise, someone qualifying for a single stacker slot could simply stall the process.

If the DKG has 100% participation, then the threshold for signing should be 70%, and the vote threshold should be 100% -- _not_ 70%. Because otherwise, a 70% group could construct an aggregate key including only shares for themselves, and then successfully elect that aggregate key.

If the DKG has _less_ than 100% participation, then the threshold for signing must be higher (i.e, if only 70% of stackers participate in DKG, then the threshold signature must be 100%), and the voting process must enforce this.

kantai:
Tagging @jcnelson @xoloki for some more thoughts on the above.

jcnelson:
I don't think there was ever a world where the system could only work with 100% Stacker participation in DKG.  It needs a way to operate in a degraded mode of operation when less than 100% of Stackers are online.

However, I also believe that there is no world in which the system allows blocks to be produced with fewer than 70% of the locked STX signing them.  So, that leaves us in a position where if fewer than 100% Stackers participate in DKG, then a greater fraction of the participating Stackers will need to participate in signing rounds so that the absolute 70% minimum is cleared.

A naive solution of running DKG once per reward cycle and just doing our best afterwards could put us in a precarious position where there's essentially no room for signers to drop off.  For example, suppose only 71% of the Stackers (by stacked STX) participated in DKG -- then, >99% of these Stackers would be required to sign blocks, and the remaining 29% of Stackers don't have anything to do.

So, I'd propose two mitigations:

* **Allow DKG to happen at any point in the reward cycle.**  but such that the number of stacked STX the aggregate public key represents can only increase.  Then, the implementation begin the reward cycle with a smaller-than-desired number of Stackers signing blocks, but signers could re-request DKG until enough Stackers have participated (e.g. over 95%).

[...]

Issue #4058, opened 2023-11-14
[stacks-core] Update the stacker-db contract from the list of registered signers


jferrant:
There should be a map of signer keys (public keys) to stacked stx in pox-4. Utilize this map to update the stacker-db config of the stacker-db instance to enable these signers to write to the appropriate slots. Note that we will need to sets of stacker-db slots. One for signers for cycle N, one for signers of cycle N+1.

setzeus:
Pasting design doc that mentions this issue for reference: https://docs.google.com/document/d/1IFTaHjEGHJkdFiEfKNayfmD0UlseoXpVcoMgJv0VW1g/edit?usp=sharing

setzeus:
Does #4125 change the requirements here?

Issue #4057, opened 2023-11-14
[stacks-core] Update .pox-4 stacker registration functions to include a signer key


jferrant:
Currently there are two stacking fucntions: stack-stx and stack-aggregation-commit (I think) that should be updated to include a signer-key. This key should be tracked in a map of signer key to stacked amount for a given reward cycle.

saralab:
Will be done in aggregate-commit & also necessary in stack-extend. 

When delegating provide a public key.

setzeus:
Pasting design doc that mentions this issue for reference: https://docs.google.com/document/d/1IFTaHjEGHJkdFiEfKNayfmD0UlseoXpVcoMgJv0VW1g/edit?usp=sharing

setzeus:
Closed / addressed by #4209

Issue #3970, opened 2023-09-21
[stacks-core] Update .pox-4 to properly handle stacker's signing key registration


jferrant:
All stacking registration functions within pox-4 must include a public key to be used for signing messages. This public key is also used to determine the stacker-db configuration of the signers' stacker-db instance.

This ticket can be broken into 3 tasks:

- [x] #4057
- [x] #4059
- [x] #4058
- [x] #4147 

Context:

Signers need to be able to query the node for the following up to date info: 

```
signers = [
            {public_key = "swBaKxfzs4pQne7spxhrkF6AtB34WEcreAkJ8mPcqx3t", key_ids = [1, 2, 3]}
        ,
            {public_key = "yDJhntuJczbss1XGDmyWtG9Wpw5NDqoBBnedxmyhKiFN", key_ids = [4, 5, 6, 7, 8]}
        ,
            {public_key = "xNVCph6zd7HLLJcuwrWz1gNbFoPHjXxn7cyRvvTYhP3U", key_ids = [9, 10]}
        ,
            {public_key = "p2wFfLEbwGCmxCR5eGa46Ct6i3BVjFrvBixRn7FnCQjA", key_ids = [11, 12, 13, 14, 15, 16]}
        ,
            {public_key = "26jpUNnJPvzDJRJg3hfBn5s5MR4eQ4LLTokjrSDzByh4i", key_ids = [17, 18, 19, 20]}
        ]
```

As long as there is a list of public keys mapped to vote shares that is consistent between calls, should work

NOTE: reward_slots == vote_shares == key_ids (each key id represents ONE vote sh

saralab:
@jferrant TO breakdown these into tasks and we'll revisit the estimations

jferrant:
As signers' keys are registered via stack-stx and similar calls, the node can read these registered signers' from the pox contract at T-99th block of Reward cycle T. It can then calculate the vote shares associated with each signing key and in turn should update the contract state via a call to some function similar to "register-signer-set(list(tuple(buff [33], list(4000 u32))" where the list is of a tuple of full public key to a list of key ids (the vote shares). There is at most 4000 signers. Also each signer has at most 4000 vote shares (if one signer...then it will own 4000 vote shares).

setzeus:
Working on a draft for setting & getting signer-set in #4147 & I believe I'm running into run-cost limits while creating the data structure.

[...]

PR #4092, opened 2023-11-25
PoX-4 Signing Keys Stacking-State

setzeus:
# PoX-4 Signing Key Stacking-State Update
## Summary of Changes

This PR is meant to introduce, not complete, the core logic around signing keys in PoX-4 & address #4057 . The focus is specifically on updating the 'stacking-state' map & all the functions that read & write from it. This includes updates to the following public functions:

1. (stack-stx)
2. (delegate-stack-stx)
3. (stack-extend)
4. (delegate-stack-extend)

_what this does not include_
Getters & setters to the new map 'reward-cycle-signing-keys,' that'll be in another PR. This PR also does not include any logic that'll be used in relevant issues #4058, #4059, & #3970.

### How were these changes tested?

This was only tested locally for syntax errors / clarinet check.

### What future testing should occur?
As soon as this is merged I'll open an issue with the user stories that require testing for @moodmosaic to work into pox_4_tests.rs.

## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/stackslib/src/chainstate/nakamoto/coordinator/tests.rs
+++ b/stackslib/src/chainstate/nakamoto/coordinator/tests.rs
@@ -25,7 +25,7 @@ use stacks_common::types::chainstate::{
 };
 use stacks_common::types::{Address, StacksEpoch};
 use stacks_common::util::vrf::VRFProof;
-use wsts::curve::point::Point;
+use wsts::curve::point::{Point, Compressed};
 
 use crate::chainstate::burn::db::sortdb::{SortitionDB, SortitionHandle};
 use crate::chainstate::burn::operations::BlockstackOperationType;
@@ -50,6 +50,9 @@ use crate::net::test::{TestPeer, TestPeerConfig};
 fn advance_to_nakamoto(peer: &mut TestPeer) {
     let mut peer_nonce = 0;
     let private_key = peer.config.private_key.clone();
+    let public_key = StacksPublicKey::from_private(&private_key);
+    let compressed_key = Compressed::try_from(public_key.to_bytes_compressed().as_slice()).expect("Failed to convert public key to compressed struct");
+    let public_key_point = Point::try_from(&compressed_key).expect("Failed to convert public key to point");
     let addr = StacksAddress::from_public_keys(
         C32_ADDRESS_VERSION_TESTNET_SINGLESIG,
         &AddressHashMode::SerializeP2PKH,
@@ -69,6 +72,7 @@ fn advance_to_nakamoto(peer: &mut TestPeer) {
                 PoxAddress::from_legacy(AddressHashMode::SerializeP2PKH, addr.bytes.clone()),
                 12,
                 34,
+                &public_key_point
             );
             vec![stack_tx]
         } else {
--- a/stackslib/src/chainstate/nakamoto/coordinator/tests.rs
+++ b/stackslib/src/chainstate/nakamoto/coordinator/tests.rs
@@ -25,7 +25,7 @@ use stacks_common::types::chainstate::{
 };
 use stacks_common::types::{Address, StacksEpoch};
 use stacks_common::util::vrf::VRFProof;
-use wsts::curve::point::{Point, Compressed};
+use wsts::curve::point::{Compressed, Point};
 
 use crate::chainstate::burn::db::sortdb::{SortitionDB, SortitionHandle};
 use crate::chainstate::burn::operations::BlockstackOperationType;
@@ -51,8 +51,10 @@ fn advance_to_nakamoto(peer: &mut TestPeer) {
     let mut peer_nonce = 0;
     let private_key = peer.config.private_key.clone();
     let public_key = StacksPublicKey::from_private(&private_key);
-    let compressed_key = Compressed::try_from(public_key.to_bytes_compressed().as_slice()).expect("Failed to convert public key to compressed struct");
-    let public_key_point = Point::try_from(&compressed_key).expect("Failed to convert public key to point");
+    let compressed_key = Compressed::try_from(public_key.to_bytes_compressed().as_slice())
+        .expect("Failed to convert public key to compressed struct");
+    let public_key_point =
+        Point::try_from(&compressed_key).expect("Failed to convert public key to point");
     let addr = StacksAddress::from_public_keys(
         C32_ADDRESS_VERSION_TESTNET_SINGLESIG,
         &AddressHashMode::SerializeP2PKH,
@@ -72,7 +74,7 @@ fn advance_to_nakamoto(peer: &mut TestPeer) {
                 PoxAddress::from_legacy(AddressHashMode::SerializeP2PKH, addr.bytes.clone()),
                 12,
                 34,
-                &public_key_point
+                &public_key_point,
             );
             vec![stack_tx]
         } else {
--- a/stackslib/src/chainstate/stacks/boot/contract_tests.rs
+++ b/stackslib/src/chainstate/stacks/boot/contract_tests.rs
@@ -62,6 +62,8 @@ lazy_static! {
         boot_code_id("pox-2", false);
     pub static ref COST_VOTING_CONTRACT_TESTNET: QualifiedContractIdentifier =
         boot_code_id("cost-voting", false);
+    pub static ref POX_4_CONTRACT_TESTNET: QualifiedContractIdentifier =
+        boot_code_id("pox-4", false);
     pub static ref USER_KEYS: Vec<StacksPrivateKey> =
         (0..50).map(|_| StacksPrivateKey::new()).collect();
     pub static ref POX_ADDRS: Vec<Value> = (0..50u64)
@@ -557,6 +559,31 @@ impl HeadersDB for TestSimHeadersDB {
         // if the block is defined at all, then return a constant
         self.get_burn_block_height_for_block(id_bhh).map(|_| 3000)
     }
+}
+
+#[test]
+fn pox_4_instantiates() {
+    let mut sim = ClarityTestSim::new();
+    sim.epoch_bounds = vec![0, 1, 2];
+    let delegator = StacksPrivateKey::new();
+
+    let expected_unlock_height = POX_TESTNET_CYCLE_LENGTH * 4;
+
+    // execute past 2.1 epoch initialization
+    sim.execute_next_block(|_env| {});
+    sim.execute_next_block(|_env| {});
+    sim.execute_next_block(|_env| {});
+
+    sim.execute_next_block(|env| {
+        env.initialize_versioned_contract(
+            POX_4_CONTRACT_TESTNET.clone(),
+            ClarityVersion::Clarity2,
+            &boot::POX_4_TESTNET_CODE,
+            None,
+            ASTRules::PrecheckSize,
+        )
+        .unwrap()
+    });
 }
 
 #[test]
--- a/stackslib/src/chainstate/stacks/boot/mod.rs
+++ b/stackslib/src/chainstate/stacks/boot/mod.rs
@@ -1666,11 +1666,6 @@ pub mod test {
         burn_ht: u64,
         public_key: &Point,
     ) -> StacksTransaction {
-        // ;; TODO: add signer key
-        // (define-public (stack-stx (amount-ustx uint)
-        //                           (pox-addr (tuple (version (buff 1)) (hashbytes (buff 32))))
-        //                           (burn-height uint)
-        //                           (lock-period uint))
         let addr_tuple = Value::Tuple(addr.as_clarity_tuple().unwrap());
         let payload = TransactionPayload::new_contract_call(
             boot_code_test_addr(),
--- a/stackslib/src/chainstate/stacks/boot/mod.rs
+++ b/stackslib/src/chainstate/stacks/boot/mod.rs
@@ -1664,6 +1664,7 @@ pub mod test {
         addr: PoxAddress,
         lock_period: u128,
         burn_ht: u64,
+        public_key: &Point,
     ) -> StacksTransaction {
         // ;; TODO: add signer key
         // (define-public (stack-stx (amount-ustx uint)
@@ -1680,6 +1681,8 @@ pub mod test {
                 addr_tuple,
                 Value::UInt(burn_ht as u128),
                 Value::UInt(lock_period),
+                Value::buff_from(public_key.compress().data.to_vec())
+            .expect("Failed to serialize aggregate public key")
             ],
         )
         .unwrap();
@@ -1792,13 +1795,17 @@ pub mod test {
         nonce: u64,
         addr: PoxAddress,
         lock_period: u128,
+        signing_key: Option<&Point>,
     ) -> StacksTransaction {
+        let signing_key = Value::Optional(OptionalData {
+            data: signing_key.map(|key| Box::new(Value::buff_from(key.compress().data.to_vec()).unwrap()))
+        });
         let addr_tuple = Value::Tuple(addr.as_clarity_tuple().unwrap());
         let payload = TransactionPayload::new_contract_call(
             boot_code_test_addr(),
             POX_4_NAME,
             "stack-extend",
-            vec![Value::UInt(lock_period), addr_tuple],
+            vec![Value::UInt(lock_period), addr_tuple, signing_key],
         )
         .unwrap();
 
--- a/stackslib/src/chainstate/stacks/boot/mod.rs
+++ b/stackslib/src/chainstate/stacks/boot/mod.rs
@@ -1677,7 +1677,7 @@ pub mod test {
                 Value::UInt(burn_ht as u128),
                 Value::UInt(lock_period),
                 Value::buff_from(public_key.compress().data.to_vec())
-            .expect("Failed to serialize aggregate public key")
+                    .expect("Failed to serialize aggregate public key"),
             ],
         )
         .unwrap();
@@ -1793,7 +1793,8 @@ pub mod test {
         signing_key: Option<&Point>,
     ) -> StacksTransaction {
         let signing_key = Value::Optional(OptionalData {
-            data: signing_key.map(|key| Box::new(Value::buff_from(key.compress().data.to_vec()).unwrap()))
+            data: signing_key
+                .map(|key| Box::new(Value::buff_from(key.compress().data.to_vec()).unwrap())),
         });
         let addr_tuple = Value::Tuple(addr.as_clarity_tuple().unwrap());
         let payload = TransactionPayload::new_contract_call(
--- a/stackslib/src/chainstate/stacks/boot/pox-4.clar
+++ b/stackslib/src/chainstate/stacks/boot/pox-4.clar
@@ -140,13 +140,6 @@
     { sender: principal, contract-caller: principal }
     { until-burn-ht: (optional uint) })
 
-;; The signing key slots for a reward cycle
-;; Written by a Stacks node as part of processing the first tenure-start-block after last tenure-start-block in previous reward cycle
-(define-map reward-cycle-signing-keys
-    { reward-cycle: uint, signer-address: principal }
-    { num-slots: uint, stx-amount: uint }
-)
-
 ;; How many uSTX are stacked in a given reward cycle.
 ;; Updated when a new PoX address is registered, or when more STX are granted
 ;; to it.
@@ -258,10 +251,6 @@
         ;; no state at all
         none
     ))
-
-;; Get the number of slots & amount of STX for a given reward cycle signer
-(define-read-only (get-signer-info-for-cycle (signer-principal principal) (reward-cycle uint))
-    (map-get? reward-cycle-signing-keys { reward-cycle: reward-cycle, signer-address: signer-principal }))
 
 (define-read-only (check-caller-allowed)
     (or (is-eq tx-sender contract-caller)
--- a/stackslib/src/chainstate/stacks/boot/pox-4.clar
+++ b/stackslib/src/chainstate/stacks/boot/pox-4.clar
@@ -27,21 +27,20 @@
 (define-constant ERR_DELEGATION_WRONG_REWARD_SLOT 29)
 (define-constant ERR_STACKING_IS_DELEGATED 30)
 (define-constant ERR_STACKING_NOT_DELEGATED 31)
+(define-constant ERR_STACKING_DURING_PREPARE_PHASE 32)
+(define-constant ERR_STACK_EXTEND_NOT_SIGNED 33)
 
 ;; PoX disabling threshold (a percent)
 (define-constant POX_REJECTION_FRACTION u25)
 
 ;; Valid values for burnchain address versions.
-;; These first four correspond to address hash modes in Stacks 2.1,
-;; and are defined in pox-mainnet.clar and pox-testnet.clar (so they
-;; cannot be defined here again).
 ;; (define-constant ADDRESS_VERSION_P2PKH 0x00)
 ;; (define-constant ADDRESS_VERSION_P2SH 0x01)
 ;; (define-constant ADDRESS_VERSION_P2WPKH 0x02)
 ;; (define-constant ADDRESS_VERSION_P2WSH 0x03)
-(define-constant ADDRESS_VERSION_NATIVE_P2WPKH 0x04)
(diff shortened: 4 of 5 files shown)


setzeus:
Additional context re: using 'delegate-stack-stx' vs. 'stack-aggregation-commit-index' to submit a pool's signing-key for delegates that registered without a signing-key here:

https://github.com/stacks-network/stacks-core/discussions/4082

setzeus:
Note: per Clarity WG morning discussion it's been noted that our preferred workflow here is to actively work on a version of PoX-4 that exists in /contrib per #4094 & then periodically copy that over to the more hardened PoX-4 that'll live in /boot. 

Once #4094 is merged in I'll make these adjust *only* to the PoX-4 version in /contrib for now.

(unless there's reasonable push back on this workflow)

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -586,8 +596,9 @@
 ;; The tokens will unlock and be returned to the Stacker (tx-sender) automatically.
 (define-public (stack-stx (amount-ustx uint)
                           (pox-addr (tuple (version (buff 1)) (hashbytes (buff 32))))
+                          (signing-key (buff 33))
                           (start-burn-ht uint)
-                          (lock-period uint))
+                          (lock-period uint)))
```
jcnelson, line 601: Typo -- extra parenthesis.

Please make sure that this contract instantiates.  The reason your tests are failing is because there are syntax errors in `pox-4`, which prevents it from instantiating.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -643,6 +655,7 @@
 ;;   * amount-ustx: the total amount of ustx the delegate may be allowed to lock
 ;;   * until-burn-ht: an optional burn height at which this delegation expires
 ;;   * pox-addr: an optional address to which any rewards *must* be sent
+;; A signing-key is *not* provided because the delegate assigns this during 'stack-aggregation-commit-index'
```
jcnelson, line 658: Please delete this comment

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1018,12 +1033,17 @@
 ;; This method extends the `tx-sender`'s current lockup for an additional `extend-count`
 ;;    and associates `pox-addr` with the rewards
 (define-public (stack-extend (extend-count uint)
-                             (pox-addr { version: (buff 1), hashbytes: (buff 32) }))
+                             (pox-addr { version: (buff 1), hashbytes: (buff 32) })
+                             (updated-signing-key (optional (buff 33))))
    (let ((stacker-info (stx-account tx-sender))
          ;; to extend, there must already be an etry in the stacking-state
```
jcnelson, line 1039: can you fix this typo? `etry`

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1018,12 +1033,17 @@
...
          (stacker-state (unwrap! (get-stacker-info tx-sender) (err ERR_STACK_EXTEND_NOT_LOCKED)))
          (amount-ustx (get locked stacker-info))
          (unlock-height (get unlock-height stacker-info))
+         (current-signing-key (unwrap! (get signing-key stacker-state) (err ERR_STACK_EXTEND_NOT_SIGNED)))
```
jcnelson, line 1043: This error case shouldn't be reachable because `stack-stx` requires a signing key.  Maybe `unwrap-panic` would be better here?

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1190,12 +1211,17 @@
...
                     (pox-addr { version: (buff 1), hashbytes: (buff 32) })
-                    (extend-count uint))
+                    (extend-count uint)
+                    (updated-signing-key (optional (buff 33))))
     (let ((stacker-info (stx-account stacker))
           ;; to extend, there must already be an entry in the stacking-state
           (stacker-state (unwrap! (get-stacker-info stacker) (err ERR_STACK_EXTEND_NOT_LOCKED)))
           (amount-ustx (get locked stacker-info))
           (unlock-height (get unlock-height stacker-info))
+          (current-signing-key (unwrap! (get signing-key stacker-state) (err ERR_STACK_EXTEND_NOT_SIGNED)))
```
jcnelson, line 1221: Same -- `unwrap-panic` might be a better choice here.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -117,7 +118,9 @@
         ;;  previous stack-stx calls, or prior to an extend)
         reward-set-indexes: (list 12 uint),
         ;; principal of the delegate, if stacker has delegated
-        delegated-to: (optional principal)
+        delegated-to: (optional principal),
+        ;; signing key for Nakamoto+, only 'none' when delegated & before delegate calls 'stack-aggregation-commit-indexed
+        signing-key: (optional (buff 33))
```
kantai, line 123: It seems like this could just be `(buff 33)` rather than an `(optional (buff 33))`: the delegate sets this value in `delegate-stack-stx`, which is also the first time the stacker-state is instantiated for the delegator.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -139,6 +142,13 @@
     { sender: principal, contract-caller: principal }
     { until-burn-ht: (optional uint) })
 
+;; The signing key slots for a reward cycle
+;; Written by a Stacks node as part of processing the first tenure-start-block after last tenure-start-block in previous reward cycle
+(define-map reward-cycle-signing-keys
+    { reward-cycle: uint, signer-address: principal }
+    { num-slots: uint, stx-amount: uint }
+)
```
kantai, line 150: I'm curious about this map: what is it used for?

kantai (commented):
Echoing Jude's comments, and just had a few more

stackslib/src/chainstate/stacks/boot/pox-4.clar
friedger: Does it make sense to allow delegators enforce a signing key? If yes, should it be stored in a separate map?

stackslib/src/chainstate/stacks/boot/pox-4.clar
kantai: Never mind -- I see it's discussed here #4059

stackslib/src/chainstate/stacks/boot/pox-4.clar
friedger: Sorry, I meant the user/pool member. My question does not make sense here.

friedger: I agree, that stack-aggregation-commit can't change the signing key.

friedger (changes requested):
I try to understand whether the same logic for the pox-address should be applied for the signing key for pools: If the pool operator decides to use a different signer key, the pool members rewards are not really at risk. However, would pool members like to decide whether the pool operator can delegate to one or the other signer? I think yes, therefore `delegate-stx` should also contain a signing key in the delegation-state

friedger:
@setzeus There at least one rust functions that create stack-stx transactions with 4 instead of 5 arguments that makes the unit tests fail.
I found one in blocks.rs: https://github.com/stacks-network/stacks-blockchain/blob/cef8f87c59d326b5d30d9f2e6baf258d820da6b1/stackslib/src/chainstate/stacks/db/blocks.rs#L4222

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -139,6 +142,13 @@
     { sender: principal, contract-caller: principal }
     { until-burn-ht: (optional uint) })
 
+;; The signing key slots for a reward cycle
+;; Written by a Stacks node as part of processing the first tenure-start-block after last tenure-start-block in previous reward cycle
+(define-map reward-cycle-signing-keys
```
friedger, line 147: Could you please add a getter?

moodmosaic: @setzeus what `nextest` command you're running? Can you paste it here so I can look into it?

kantai (approved):

friedger (approved):

muneeb-ali:
Great to see the changes approved here! 🚀

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -283,8 +283,9 @@ fn test_simple_nakamoto_coordinator_bootup() {
 #[test]
 fn test_simple_nakamoto_coordinator_1_tenure_10_blocks() {
     let mut test_signers = TestSigners::default();
-    let mut peer = boot_nakamoto(function_name!(), vec![], test_signers.aggregate_public_key);
-    let private_key = peer.config.private_key.clone();
+    //let mut peer = boot_nakamoto(function_name!(), vec![], test_signers.aggregate_public_key);
```
jcnelson, line 286: Please remove dead code

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -283,8 +283,9 @@ fn test_simple_nakamoto_coordinator_bootup() {
...
+    //let private_key = peer.config.private_key.clone();
+    let private_key = StacksPrivateKey::new();
```
jcnelson, line 288: Why is this a new key?  Can you continue to please use the same key so we get consistent output?

stackslib/src/chainstate/stacks/boot/contract_tests.rs
```diff
@@ -559,6 +561,31 @@ impl HeadersDB for TestSimHeadersDB {
...
 
+#[test]
+fn pox_4_instantiates() {
+    let mut sim = ClarityTestSim::new();
+    sim.epoch_bounds = vec![0, 1, 2];
+    let delegator = StacksPrivateKey::new();
+
+    let expected_unlock_height = POX_TESTNET_CYCLE_LENGTH * 4;
+
+    // execute past 2.1 epoch initialization
```
jcnelson, line 572: What is this test supposed to check?  If all it's checking is that pox-4 is syntax-free, then these steps aren't needed -- you can go straight to instantiating pox-4.  If it's instead checking that pox-4 instantiates at the right time, then this test needs to verify that instead.

jcnelson (changes requested):
Thank you for updating the stacking interfaces with these new methods!  Before merging, I must insist that you add unit tests for them.  It's not enough to see that PoX-4 instantiates; you must also verify that the public keys get added to the stacking state.  Thanks!

stackslib/src/chainstate/stacks/boot/contract_tests.rs
jcnelson: Yes; however, this is not being tested here

moodmosaic: Then would a pox lockup transition from pox-3 to pox-4 be enough? (As a test.)

jcnelson: Yes; you'd need to verify that the `instantiate_epoch_2_5()` and `initialize_epoch_3_0()` functions in `stackslib/src/clarity_vm/clarity.rs` are getting called.

moodmosaic: OK, and then this test shouldn’t be in this file but instead it should be in `pox_4_tests.rs` yes?

jcnelson: Either way is fine with me. If @jbencin already tests this in his PR, then it's fine to omit it here.  However, it's really important that you test the new stacking API.

jcnelson:
An update from the testing WG.  This PR is blocked on getting the `pox_4.rs` tests in place, which is being handled in https://github.com/stacks-network/stacks-core/pull/4106.  Once this is in place, it should be merged into this PR, so then it will become possible to test that the signing keys were indeed set.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1194,12 +1218,17 @@
...
                     (pox-addr { version: (buff 1), hashbytes: (buff 32) })
-                    (extend-count uint))
+                    (extend-count uint)
+                    (updated-signing-key (optional (buff 33))))
     (let ((stacker-info (stx-account stacker))
           ;; to extend, there must already be an entry in the stacking-state
           (stacker-state (unwrap! (get-stacker-info stacker) (err ERR_STACK_EXTEND_NOT_LOCKED)))
           (amount-ustx (get locked stacker-info))
           (unlock-height (get unlock-height stacker-info))
+          (current-signing-key (get signing-key stacker-state))
```
friedger, line 1228: maybe simplify to
`(next-signing-key (defaults-to (get signing-key stacker-state) updated-signing-key))`

friedger (commented):
nit on pox-4.clar

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -1660,11 +1660,8 @@ pub mod test {
         lock_period: u128,
         burn_ht: u64,
     ) -> StacksTransaction {
-        // ;; TODO: add signer key
-        // (define-public (stack-stx (amount-ustx uint)
-        //                           (pox-addr (tuple (version (buff 1)) (hashbytes (buff 32))))
-        //                           (burn-height uint)
-        //                           (lock-period uint))
+        let public_key = StacksPublicKey::from_private(key);
```
jcnelson, line 1663: This `key` is the key used to sign the transaction, not the DKG signing key.  There needs to be a separate argument to `make_pox_4_lockup` for that signing key.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -117,7 +118,9 @@
         ;;  previous stack-stx calls, or prior to an extend)
         reward-set-indexes: (list 12 uint),
         ;; principal of the delegate, if stacker has delegated
-        delegated-to: (optional principal)
+        delegated-to: (optional principal),
+        ;; signing key for Nakamoto, only 'none' when delegated & before delegate calls stack-aggregation-commit-indexed
```
jcnelson, line 122: I don't think this is ever `none` if we're setting the signing key on `delegate-stack-stx` (which IIRC is the plan now)

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -591,7 +605,8 @@
 (define-public (stack-stx (amount-ustx uint)
                           (pox-addr (tuple (version (buff 1)) (hashbytes (buff 32))))
                           (start-burn-ht uint)
-                          (lock-period uint))
+                          (lock-period uint)
+                          (signing-key (buff 33)))
```
jcnelson, line 609: Can you add a comment about `signing-key` to the method doc?

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -829,7 +845,8 @@
                                    (amount-ustx uint)
                                    (pox-addr { version: (buff 1), hashbytes: (buff 32) })
                                    (start-burn-ht uint)
-                                   (lock-period uint))
+                                   (lock-period uint)
```
jcnelson, line 848: Same here -- can you add a comment to the method doc about the signing key?

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1022,12 +1040,17 @@
...
-                             (pox-addr { version: (buff 1), hashbytes: (buff 32) }))
+                             (pox-addr { version: (buff 1), hashbytes: (buff 32) })
+                             (updated-signing-key (optional (buff 33))))
    (let ((stacker-info (stx-account tx-sender))
-         ;; to extend, there must already be an etry in the stacking-state
+         ;; to extend, there must already be an entry in the stacking-state
          (stacker-state (unwrap! (get-stacker-info tx-sender) (err ERR_STACK_EXTEND_NOT_LOCKED)))
          (amount-ustx (get locked stacker-info))
          (unlock-height (get unlock-height stacker-info))
+         (current-signing-key (get signing-key stacker-state))
```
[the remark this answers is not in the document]
jcnelson: Agreed

stackslib/src/chainstate/stacks/boot/pox-4.clar
jcnelson: Agreed

testnet/stacks-node/src/mockamoto.rs
```diff
@@ -824,6 +824,9 @@ impl MockamotoNode {
             Some(AddressHashMode::SerializeP2PKH),
         );
 
+        let public_key = StacksPublicKey::from_private(&self.miner_key);
```
jcnelson, line 827: I think there instead needs to be a separate `signing_key` field in `MockamotoNode`.  Same for the Nakamoto node.  Perhaps it can be part of the `Keychain`?

stackslib/src/chainstate/stacks/boot/pox-4.clar
setzeus: Updated a sentence & added the following to comment doc above function:

`;; * You need to provide a signing-key used to allocate signature-weight to a valid signer`

stackslib/src/chainstate/stacks/boot/pox-4.clar
setzeus: Updated a sentence & added the following to comment doc above function:

`;; * You need to provide a signing-key used to allocate signature-weight to a valid signer`

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -27,6 +27,7 @@
 (define-constant ERR_DELEGATION_WRONG_REWARD_SLOT 29)
 (define-constant ERR_STACKING_IS_DELEGATED 30)
 (define-constant ERR_STACKING_NOT_DELEGATED 31)
+(define-constant ERR_STACK_EXTEND_NO_SIGNING_KEY 32)
```
moodmosaic, line 30: Where is this constant used?

moodmosaic (commented):
It would be nice with some tests to ensure everything works as intended:

1. **`signing-key` in stacking-state:**
   - Verify that the `signing-key` is correctly set.

2. **Modified Functions (`stack-stx`, `delegate-stack-stx`, etc.):**
   - Ensure all functions modified to include the `signing-key` handle it correctly.

3. **`stack-extend`, `delegate-stack-extend`:**
   - Test handling of `updated-signing-key`, ensuring proper updates in the stacking state.

4. **`get-signer-info-for-cycle`:**
   - Verify retrieval of correct signer info for various inputs.

@jcnelson, @setzeus, could you share your thoughts on these test cases or suggest additional ones that might be relevant?

jcnelson:
@moodmosaic Yes, please test all of those in `pox_4_tests.rs`.

contrib/core-contract-tests/tests/bns/name_register_test.ts
```diff
@@ -98,7 +98,7 @@ Clarinet.test({
                 ], 
                 cases[0].nameOwner.address),
         ]);
-        assertEquals(block.height, 2);
+        assertEquals(block.height, 3);
```
jcnelson, line 101: This looks like there was an off-by-one error somewhere, with all these changes. Can you elaborate more on why all of these block heights are now +1?

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -1798,7 +1794,7 @@ pub mod test {
             boot_code_test_addr(),
             POX_4_NAME,
             "stack-extend",
-            vec![Value::UInt(lock_period), addr_tuple],
+            vec![Value::UInt(lock_period), addr_tuple, Value::Optional(OptionalData {data: None})],
```
jcnelson, line 1797: Don't pass `None`; instead, make this an argument in Rust which gets passed to `stack-extend`.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -27,21 +27,20 @@
...
-;; and are defined in pox-mainnet.clar and pox-testnet.clar (so they
-;; cannot be defined here again).
 ;; (define-constant ADDRESS_VERSION_P2PKH 0x00)
 ;; (define-constant ADDRESS_VERSION_P2SH 0x01)
 ;; (define-constant ADDRESS_VERSION_P2WPKH 0x02)
 ;; (define-constant ADDRESS_VERSION_P2WSH 0x03)
-(define-constant ADDRESS_VERSION_NATIVE_P2WPKH 0x04)
-(define-constant ADDRESS_VERSION_NATIVE_P2WSH 0x05)
-(define-constant ADDRESS_VERSION_NATIVE_P2TR 0x06)
+;; (define-constant ADDRESS_VERSION_NATIVE_P2WPKH 0x04)
```
jcnelson, line 41: Please don't commit commented-out code

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -27,21 +27,20 @@
...
+(define-constant ERR_STACK_EXTEND_NOT_SIGNED 33)
 
 ;; PoX disabling threshold (a percent)
 (define-constant POX_REJECTION_FRACTION u25)
 
 ;; Valid values for burnchain address versions.
-;; These first four correspond to address hash modes in Stacks 2.1,
-;; and are defined in pox-mainnet.clar and pox-testnet.clar (so they
-;; cannot be defined here again).
 ;; (define-constant ADDRESS_VERSION_P2PKH 0x00)
```
jcnelson, line 37: If you're now using `is-in-mainnet` instead of a separate `pox-4-mainnet.clar`, then where are these constants defined now?

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -508,6 +535,14 @@
     (and (>= lock-period MIN_POX_REWARD_CYCLES)
          (<= lock-period MAX_POX_REWARD_CYCLES)))
 
+;; Is the given burn block height in the prepare phase?
+;; This computes `((height - first-burnchain-block-height) + pox-prepare-cycle-length) % pox-reward-cycle-length) < pox-prepare-cycle-length`.
+(define-read-only (check-prepare-phase (height uint))
```
jcnelson, line 540: @setzeus This needs to be in #4164

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -545,6 +580,10 @@
     (asserts! (check-pox-lock-period num-cycles)
               (err ERR_STACKING_INVALID_LOCK_PERIOD))
 
+    ;; stacking must not happen during prepare phase
+    (asserts! (not (check-prepare-phase burn-block-height))
```
jcnelson, line 584: Please revert this -- it has not been decided yet that stacking will be disabled in the prepare phase

setzeus: Yes. Reverting all the work here to un-do rebasing from #4094.

contrib/core-contract-tests/tests/bns/name_register_test.ts
friedger: The pox-4 is in a different epoch than bns, therefore booting needs one more block

setzeus:
Work picked up by @MarvinJanssen in #4209 to add required tests in addition to base tests - changing to draft until #4209 is merged in then will close.

setzeus:
Closing since #4209 is now merged in.

PR #3503, opened 2023-01-18
Feat/synthetic print delegate stx

pavitthrap:
### Description
This PR adds a synthetic print event to the `delegate-stx` function in the `pox-2` contract. It does so by means of the special contract handler (defined in `special.rs`). 

This PR adds 2 tests:
- a test in `pox_2_tests.rs` that called `delegate-stx` through a normal stacks transaction
- a test in `neon_integrations.rs` that sends a btc wire DelegateStx tx, and verifies (a) that the delegation was successful, and (b) that a synthetic print event was emitted. 

### Applicable issues
- fixes #3465

### Checklist
- [x] Test coverage for new or modified code paths
- [ ] Changelog is updated
- [ ] Required documentation changes (e.g., `docs/rpc/openapi.yaml` and `rpc-endpoints.md` for v2 endpoints, `event-dispatcher.md` for new events)
- [ ] New clarity functions have corresponding PR in `clarity-benchmarking` repo
- [x] New integration test(s) added to `bitcoin-tests.yml`

PR #4094, opened 2023-11-28
pox-4: disallow stacking during prepare phase

### LINKED ISSUE #3775: [pox-4] Improve stacking deadline handling
As a stacker, I lock my stacks before the start of the cycle but after the start of the prepare phase and my stx do not earn any rewards 

Currently, there is no way to learn about the cycle anchor block. It would be nice to include that information in an api. 
Furthermore, the stacking transactions with (delegate-) stack-stx, (delegate-) stack-extend/increase or fail if called during the prepare phase. 

**additional context**
See "Not earning locked stx" table on https://stacksonchain.com/dashboards/Fast%20Pool/276

friedger:
### Description
This PR update pox-4 to disallow stacking during the prepare phase

### Applicable issues
- fixes #3775 

When user stack stx during the prepare phase, their stx is locked, but they do not earn rewards. In addition, pool operators have extra work to remove these pool members.

### Checklist
- [ ] Changelog is updated

### DIFF
--- a/stackslib/src/chainstate/stacks/boot/pox-4.clar
+++ b/stackslib/src/chainstate/stacks/boot/pox-4.clar
@@ -27,6 +27,7 @@
 (define-constant ERR_DELEGATION_WRONG_REWARD_SLOT 29)
 (define-constant ERR_STACKING_IS_DELEGATED 30)
 (define-constant ERR_STACKING_NOT_DELEGATED 31)
+(define-constant ERR_STACKING_DURING_PREPARE_PHASE 32)
 
 ;; PoX disabling threshold (a percent)
 (define-constant POX_REJECTION_FRACTION u25)
@@ -508,6 +509,12 @@
     (and (>= lock-period MIN_POX_REWARD_CYCLES)
          (<= lock-period MAX_POX_REWARD_CYCLES)))
 
+;; Is the current block height outside of the prepare phase?
+(define-read-only (check-not-prepare-phase)
+    (let ((cycle-length (var-get pox-reward-cycle-length))))
+    (< (mod (- burn-block-height (var-get first-burnchain-block-height)) cycle-length)
+      (- cycle-length (var-get pox-prepare-cycle-length))))
+
 ;; Evaluate if a participant can stack an amount of STX for a given period.
 ;; This method is designed as a read-only method so that it can be used as
 ;; a set of guard conditions and also as a read-only RPC call that can be
@@ -544,6 +551,10 @@
     ;; lock period must be in acceptable range.
     (asserts! (check-pox-lock-period num-cycles)
               (err ERR_STACKING_INVALID_LOCK_PERIOD))
+
+    ;; stacking must not happen during prepare phase
+    (asserts! (check-not-prepare-phase)
+              (err ERR_STACKING_DURING_PREPARE_PHASE))
 
     ;; address version must be valid
     (asserts! (check-pox-addr-version (get version pox-addr))
@@ -1014,7 +1025,7 @@
                     stacker: tx-sender,
                     add-amount: increase-by })))
             (err ERR_STACKING_UNREACHABLE))
-      ;; NOTE: stacking-state map is unchanged: it does not track amount-stacked in PoX-4
+      ;; NOTE: stacking-state map is unchanged: it does not track amount-stacked in pox-4
       (ok { stacker: tx-sender, total-locked: (+ amount-stacked increase-by)})))
 
 ;; Extend an active Stacking lock.
--- a/stackslib/src/chainstate/stacks/boot/pox-4.clar
+++ b/stackslib/src/chainstate/stacks/boot/pox-4.clar
@@ -33,16 +33,13 @@
 (define-constant POX_REJECTION_FRACTION u25)
 
 ;; Valid values for burnchain address versions.
-;; These first four correspond to address hash modes in Stacks 2.1,
-;; and are defined in pox-mainnet.clar and pox-testnet.clar (so they
-;; cannot be defined here again).
 ;; (define-constant ADDRESS_VERSION_P2PKH 0x00)
 ;; (define-constant ADDRESS_VERSION_P2SH 0x01)
 ;; (define-constant ADDRESS_VERSION_P2WPKH 0x02)
 ;; (define-constant ADDRESS_VERSION_P2WSH 0x03)
-(define-constant ADDRESS_VERSION_NATIVE_P2WPKH 0x04)
-(define-constant ADDRESS_VERSION_NATIVE_P2WSH 0x05)
-(define-constant ADDRESS_VERSION_NATIVE_P2TR 0x06)
+;; (define-constant ADDRESS_VERSION_NATIVE_P2WPKH 0x04)
+;; (define-constant ADDRESS_VERSION_NATIVE_P2WSH 0x05)
+;; (define-constant ADDRESS_VERSION_NATIVE_P2TR 0x06)
 ;; Keep these constants in lock-step with the address version buffs above
 ;; Maximum value of an address version as a uint
 (define-constant MAX_ADDRESS_VERSION u6)
@@ -52,6 +49,21 @@
 ;; Maximum value of an address version that has a 32-byte hashbytes
 ;; (0x05 and 0x06 have 32-byte hashbytes)
 (define-constant MAX_ADDRESS_VERSION_BUFF_32 u6)
+
+;; PoX mainnet constants
+;; Min/max number of reward cycles uSTX can be locked for
+(define-constant MIN_POX_REWARD_CYCLES u1)
+(define-constant MAX_POX_REWARD_CYCLES u12)
+
+;; Default length of the PoX registration window, in burnchain blocks.
+(define-constant PREPARE_CYCLE_LENGTH (if is-in-mainnet u100 u50))
+
+;; Default length of the PoX reward cycle, in burnchain blocks.
+(define-constant REWARD_CYCLE_LENGTH (if is-in-mainnet u2100 u1050))
+
+;; Stacking thresholds
+(define-constant STACKING_THRESHOLD_25 (if is-in-mainnet u20000 u8000))
+(define-constant STACKING_THRESHOLD_100 (if is-in-mainnet u5000 u2000))
 
 ;; Data vars that store a copy of the burnchain configuration.
 ;; Implemented as data-vars, so that different configurations can be
@@ -509,11 +521,13 @@
     (and (>= lock-period MIN_POX_REWARD_CYCLES)
          (<= lock-period MAX_POX_REWARD_CYCLES)))
 
-;; Is the current block height outside of the prepare phase?
-(define-read-only (check-not-prepare-phase)
-    (let ((cycle-length (var-get pox-reward-cycle-length))))
-    (< (mod (- burn-block-height (var-get first-burnchain-block-height)) cycle-length)
-      (- cycle-length (var-get pox-prepare-cycle-length))))
+;; Is the given burn block height in the prepare phase?
+;; This computes `((height - first-burnchain-block-height) + pox-prepare-cycle-length) % pox-reward-cycle-length) < pox-prepare-cycle-length`.
+(define-read-only (check-prepare-phase (height uint))
+    (let ((prepare-cycle-length (var-get pox-prepare-cycle-length)))
+        (< (mod (+ (- height (var-get first-burnchain-block-height)) prepare-cycle-length)
+            (var-get pox-reward-cycle-length))
+        prepare-cycle-length)))
 
 ;; Evaluate if a participant can stack an amount of STX for a given period.
 ;; This method is designed as a read-only method so that it can be used as
@@ -553,7 +567,7 @@
               (err ERR_STACKING_INVALID_LOCK_PERIOD))
 
     ;; stacking must not happen during prepare phase
-    (asserts! (check-not-prepare-phase)
+    (asserts! (not (check-prepare-phase burn-block-height))
               (err ERR_STACKING_DURING_PREPARE_PHASE))
 
     ;; address version must be valid
@@ -1025,7 +1039,7 @@
                     stacker: tx-sender,
                     add-amount: increase-by })))
             (err ERR_STACKING_UNREACHABLE))
-      ;; NOTE: stacking-state map is unchanged: it does not track amount-stacked in pox-4
+      ;; NOTE: stacking-state map is unchanged: it does not track amount-stacked in PoX-4
       (ok { stacker: tx-sender, total-locked: (+ amount-stacked increase-by)})))
 
 ;; Extend an active Stacking lock.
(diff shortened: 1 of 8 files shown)


stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -634,6 +645,7 @@
     ;; must be called directly by the tx-sender or by an allowed contract-caller
     (asserts! (check-caller-allowed)
               (err ERR_STACKING_PERMISSION_DENIED))
+    (print {notification: "revoke-delegate-stx", payload: {stacker: tx-sender}})
```
zone117x, line 648: Is there a reason to implement this event differently from the others? See here for how the other 10 or so pox events are implements: https://github.com/stacks-network/stacks-core/blob/6d0effc0d1135546196af05c0067a7fd2275fc63/src/clarity_vm/special.rs#L570

friedger: I thought there was a reason why is wasn't synthesized. If we can do that, I am happy to change the PR. @zone117x

zone117x: I think it was just never implemented. If you can add it to the synthesized events that would help with integration in the Hiro API as well 👍

jcnelson:
Hi @friedger, thanks for sending this over!

Regarding the PR organization, we don't use Clarinet to test `pox-4` (or any PoX contract) because PoX is not a smart contract.  It's a part of the Clarity VM that just happens to be written in Clarity.  Its behavior and implementation is deeply dependent on the ways in which the node interacts with it (e.g. in order to lock/unlock STX, calculate reward sets, and so on).  We test PoX in Rust via tests in `pox-locking/src` and in `stackslib/src/chainstate/stacks/boot/`.

contrib/core-contract-tests/contracts/pox/pox-4.clar
```diff
@@ -0,0 +1,1355 @@
+;; PoX testnet constants
```
jcnelson, line 1: This belongs in `stackslib/src/chainstate/stacks/boot/pox-4.clar`

contrib/core-contract-tests/contracts/pox/pox-4.clar
```diff
@@ -0,0 +1,1355 @@
...
+
+    ;; sender principal must not have rejected in this upcoming reward cycle
+    (asserts! (is-none (get-pox-rejection tx-sender first-reward-cycle))
+              (err ERR_STACKING_ALREADY_REJECTED))
+
+    ;; lock period must be in acceptable range.
+    (asserts! (check-pox-lock-period num-cycles)
+              (err ERR_STACKING_INVALID_LOCK_PERIOD))
+
+    ;; stacking must not happen during prepare phase
```
jcnelson, line 574: Why is this no longer allowed?

contrib/core-contract-tests/contracts/pox/pox-4.clar
```diff
@@ -0,0 +1,1355 @@
...
+            (is-eq (len hashbytes) u32)
+            false)))
+
+;; Is the given lock period valid?
+(define-read-only (check-pox-lock-period (lock-period uint))
+    (and (>= lock-period MIN_POX_REWARD_CYCLES)
+         (<= lock-period MAX_POX_REWARD_CYCLES)))
+
+;; Is the current burn block height in the prepare phase?
+(define-read-only (check-not-prepare-phase)
```
jcnelson, line 532: A few things:

* Can this be an affirmative test?  e.g. `check-prepare-phase`?  Then, the caller performs the negation.
* Instead of relying on globals (e.g. `burn-block-height`), can this instead be an argument?  The caller would supply the global, and the test code would supply arbitrary values without having to set up a whole blockchain to test it.

contrib/core-contract-tests/contracts/pox/pox-4.clar
friedger: When stacking during prepare phase the stx are locked but are not earning any rewards. It confuses users and it is extra work for pool operators.

friedger: Only stx locked before the anchor block are considered for the reward slots.

contrib/core-contract-tests/contracts/pox/pox-4.clar
friedger: It is part of `stackslib/src/chainstate/stacks/boot/pox-testnet.clar`

stackslib/src/chainstate/stacks/boot/pox-4.clar
friedger: added to `pox-locking/events.rs`

jcnelson:
Yes

contrib/core-contract-tests/contracts/pox/pox-4.clar
jcnelson: What I mean is, this file should not exist.  It's a duplicate of code in `stackslib/src/chainstate/stacks/boot/pox-4.clar`.  There should only be one PoX-4 code body in the repository.

friedger:
checks are green now :-)

contrib/core-contract-tests/contracts/pox/pox-4.clar
friedger: I have added a test that ensure that the duplicated code is always the same. Hopefully, we can remove it once clarity can handle this better.

jcnelson: I'm sorry, but I must insist that we only have one copy of the PoX-4 code body in the repo.

zone117x (commented):
The new event for `revoke-delegate-stx` looks good. I'm not sure about the implications of the stacking during prepare phase change, so my approval for this PR wouldn't mean much. @janniks any thoughts?

contrib/core-contract-tests/contracts/pox/pox-4.clar
wileyj: @friedger I concur with Jude here - keeping two copies of a boot contract adds an extra burden to keep them both up to date at a minimum. 
could you use the contract(s) in `./stackslib/src/chainstate/stacks/boot`, or possibly create a symlink in the `./contrib` dir?

wileyj: > I have added a test that ensure that the duplicated code is always the same. Hopefully, we can remove it once clarity can handle this better.

another option rather than a symilnk - wouldn't copying the existing contract to an ephemeral location also work (provided the test removes that ephemeral contract upon conclusion)?

friedger: @jcnelson The two files are different, therefore the two different names.

friedger: I created an issue on clarinet: https://github.com/hirosystems/clarinet/issues/1267 Do we want to have this blocked PR blocked until it is released? Can we afford that?

Alternatively, we can remove clarinet tests and commit a change to a boot contract without tests. Does not look good to me. 

I take ownership of the situation with clarinet testing of pox-4 and try to improve the situation as soon as possible. To me, it is not a blocker.

contrib/core-contract-tests/contracts/pox/pox-4.clar
friedger: @jcnelson Does it answer you question?

friedger:
I have added a link to the discussion on discord before this feature request (https://discord.com/channels/621759717756370964/730814571517968436/1112787020117463100) to the corresponding github issue

contrib/core-contract-tests/contracts/pox/pox-4.clar
friedger: `pox-testnet.clar + pox-4.clar = pox-4-testnet.clar`
Clarinet unit tests can't run on pox-4.clar because that file does not define a valid contract. Instead we create pox-4-testnet.clar and run the unit tests against the new, valid contract file. (pox-testnet/pox-mainnet.clar only consists of constant definitions.)

Hugo suggested on [a different issue](https://github.com/stacks-network/stacks-core/pull/4106#issuecomment-1836655726) to make pox-4.clar a valid contract and use `is-in-mainnet` in that file. If @jcnelson  agrees, I am happy to update this PR and remove the current situation.

contrib/core-contract-tests/contracts/pox/pox-4.clar
CAGS295: I'm sorry for butting in. Even though stackers don't earn rewards, does stacking during the preparation phase confer voting power?

friedger: My understanding is that they don't get any reward slots. 1 slot, 1 vote on pox. Therfore, no voting right.

jcnelson: Yup, that's right. Yes @friedger this answers my question. I didn't know that this was a pain point for users, so I'm glad to see that it's getting fixed :+1:

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1,4 +1,26 @@
...
+
+;; Default length of the PoX reward cycle, in burnchain blocks.
+(define-constant REWARD_CYCLE_LENGTH (if is-in-mainnet u2100 u1050))
+
+;; Valid values for burnchain address versions.
+;; These correspond to address hash modes in Stacks 2.0.
+(define-constant ADDRESS_VERSION_P2PKH 0x00)
+(define-constant ADDRESS_VERSION_P2SH 0x01)
+(define-constant ADDRESS_VERSION_P2WPKH 0x02)
+(define-constant ADDRESS_VERSION_P2WSH 0x03)
```
jcnelson, line 18: Per https://github.com/stacks-network/stacks-core/pull/4117#pullrequestreview-1763898214, can you consolidate these `ADDRESS_VERSION_*` constants with the ones below, at line 65?  Also, can you move the remaining constants _beneath_ the existing constants (after line 75) so as to keep the layout of this file consistent with the other `pox`-es?  Thanks!

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -508,6 +531,12 @@
     (and (>= lock-period MIN_POX_REWARD_CYCLES)
          (<= lock-period MAX_POX_REWARD_CYCLES)))
 
+;; Is the given burn block height in the prepare phase?
```
jcnelson, line 534: Can you add this comment, so other folks reading this have an easier time following it?

```
;; This computes `((height - first-burnchain-block-height) + pox-prepare-cycle-length) % pox-reward-cycle-length) < pox-prepare-cycle-length`.
```

jcnelson: Also, you can save one read-count if you store `pox-prepare-cycle-length` with a `let`-binding :)

stackslib/src/chainstate/stacks/boot/pox-4.clar
friedger: @jcnelson Addressed in https://github.com/stacks-network/stacks-core/pull/4094/commits/174b6b0b6ff311b627281af68d1eba64c2fd6a90

pox-locking/src/events.rs
```diff
@@ -362,7 +364,8 @@ pub fn synthesize_pox_2_or_3_event_info(
         | "delegate-stack-extend"
         | "stack-increase"
         | "delegate-stack-increase"
-        | "delegate-stx" => Some(create_event_info_stack_or_delegate_code(
+        | "delegate-stx"
+        | "revoke-delegate-stx" => Some(create_event_info_stack_or_delegate_code(
```
zone117x, line 368: @friedger I'm looking into ingesting this new event, and it seems to have issues. It looks like this event is missing code in the above `create_event_info_data_code` function. Can you fix that?

And can you please add tests for this event similar to how the others are tested? Here's a PR showing how tests for the `delegate-stx` event were implemented: https://github.com/stacks-network/stacks-core/pull/3503/files

zone117x (changes requested):

pox-locking/src/events.rs
friedger: @zone117x I have added the missing code. However, the test does not pass. The event is not created. Do you have any hints what is missing?

pox-locking/src/events.rs
```diff
@@ -337,6 +337,15 @@ fn create_event_info_data_code(function_name: &str, args: &[Value]) -> String {
                 pox_addr = &args[3],
             )
         }
+        "revoke-delegate-stx" => {
+            format!(
+                r#"
+                {{
+                    data: {{ }}
```
zone117x, line 344: Can this include the same set of properties as the corresponding `delegate-stx` event? At a minimum the `delegate-to` property which would make it easier to expose these events to the associated pool operators

pox-locking/src/events.rs
zone117x: It might be related to https://github.com/stacks-network/stacks-core/issues/4143

pox-locking/src/events.rs
friedger: That data is not available during the call. It can be called while delegating or while non-delegating

zone117x: Seems like arguably a bug or at least unexpected behavior in pox-4. Consumers of this event will have to do extra work to determine if the revocation call/event was actually meaningful. 

The `(map-delete delegation-state ..)` here should not return success if no change occurred (i.e. check if the map has something to delete first) https://github.com/stacks-network/stacks-core/blob/2c1c777b9ed0124a65e9cd2c019bd85f0735a5e0/stackslib/src/chainstate/stacks/boot/pox-4.clar#L636-L641

friedger: I haven't seen a requirement for meaningful or non-meaningful revoke txs. My experience says it is ok like this. We only want to have the tx-sender.

zone117x: I created an issue for discussion https://github.com/stacks-network/stacks-core/issues/4146

friedger:
Changes related to revoke-delegate-stx have been removed.
This PR should be now much easier to merge.

pox-locking/src/events.rs
friedger: Moved to https://github.com/stacks-network/stacks-core/pull/4157

pox-locking/src/events.rs
friedger: moved to https://github.com/stacks-network/stacks-core/pull/4157

setzeus (approved):

setzeus (approved):
Reviewed & all LGTM. Only thoughts are maybe the name registration tests should be extracted / part of another PR?

Edit* Especially since I see more BNS tests in #4100 & #4116 I think it might make sense to make a separate PR strictly for BNS items.

friedger:
@zone117x Could you please review this again?

obycode:
Feedback from PR meeting - let's open a discussion.

jcnelson:
Discussion created: https://github.com/stacks-network/stacks-core/discussions/4193

friedger:
Converted to draft until discussion is concluded

saralab:
Risky , will reconsider for next HF

PR #4095, opened 2023-11-28
Bugfix/nakamoto coordinator reward set prepare phase calcs

jferrant:
### Description
This pulls some bugfixes out of other branches that have discovered them:
- The first commit was cherry picked from feat/nakamoto-validate-signer 
- The second commit was taken from feat/mockamoto-mining-redux
###  Fixes
- off by one bug in find_prepare_phase_sortitions which incorrectly included one BEFORE prepare phase in the list of sortitions
- off by one bug in get_reward_set_nakamoto which was incorrectly using the current reward cycle rather than the NEXT reward cycle for getting reward addresses. This bug was due to change in assumptions in Nakamoto vs pre-Nakamoto: Nakamoto calculates reward set DURING the prepare phase of the PRIOR reward cycle whereas pre-Nakamoto calculated reward set at the START of the CURRENT reward cycle.

### DIFF
--- a/.github/workflows/bitcoin-tests.yml
+++ b/.github/workflows/bitcoin-tests.yml
@@ -28,19 +28,10 @@ jobs:
         id: git_checkout
         uses: actions/checkout@v3
 
-      - name: Reclaim disk space
-        id: cleanup
-        run: |
-          sudo apt-get update
-          sudo apt-get remove -y '^dotnet-.*'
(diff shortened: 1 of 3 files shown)


xoloki (approved):
LGTM, but I'm not at all familiar with the code that's changed, so someone who is needs to review it before merge

PR #4098, opened 2023-11-28
Feat: Mockamoto + coordinator integration

kantai:
### Description

This PR integrates the mockamoto node into the latest nakamoto coordinator work. It includes a test case that ensures that the event observer sees 100 nakamoto blocks.

This operation simulates bitcoin blocks, storing them directly into the burndb, and then notifying the chains coordinator.


stackslib/src/chainstate/burn/operations/leader_block_commit.rs
```diff
@@ -884,7 +885,8 @@ impl LeaderBlockCommitOp {
 
         let is_already_committed = tx.expects_stacks_block_in_fork(&self.block_header_hash)?;
 
-        if is_already_committed {
+        // in Epoch3.0+, block commits can include Stacks blocks already accepted in the fork.
```
jcnelson, line 888: I'm not sure how this is possible?  All this is checking is that there are no duplicate block-commit block hashes (i.e. StacksBlockIds) in this Bitcoin fork.

testnet/stacks-node/src/config.rs
```diff
@@ -452,6 +526,19 @@ impl Config {
...
+                .iter()
+                .find(|epoch| epoch.epoch_id == StacksEpochId::Epoch21)
+            {
+                // Override v1_unlock_height to the start_height of epoch2.1
+                debug!(
+                    "Override v2_unlock_height from {} to {}",
+                    burnchain.pox_constants.v1_unlock_height,
+                    epoch.start_height + 1
+                );
+                burnchain.pox_constants.v1_unlock_height = epoch.start_height as u32 + 1;
```
jcnelson, line 539: Are you going to do this for `v2_unlock_height` as well?

testnet/stacks-node/src/mockamoto.rs
```diff
@@ -247,61 +558,177 @@ impl MockamotoNode {
...
+                self.chainstate.db(),
+                &parent_snapshot.consensus_hash,
+            )
+            .map_err(|_e| BurnchainError::MissingParentBlock)?
+            .unwrap_or_else(|| VRFProof::empty());
+
+            let vrf_seed = VRFSeed::from_proof(&parent_vrf_proof);
+
+            let block_commit = LeaderBlockCommitOp {
+                block_header_hash: BlockHeaderHash([0; 32]),
```
jcnelson, line 685: Pretty sure this field is getting validated now.  This should be the index block hash of the last tenure's tenure-start block.

testnet/stacks-node/src/mockamoto.rs
```diff
@@ -339,18 +766,20 @@ impl MockamotoNode {
             (tip_info.stacks_block_height, tip_info.burn_header_height)
         };
 
-        let miner_nonce = 2 * parent_chain_length;
+        info!("Mining block"; "parent_chain_length" => parent_chain_length, "chain_tip_bh" => %chain_tip_bh, "chain_tip_ch" => %chain_tip_ch);
+        let miner_nonce = 3 * parent_chain_length;
```
jcnelson, line 770: Would it be less brittle to just get the miner's account from the Stacks chain tip?

testnet/stacks-node/src/mockamoto.rs
```diff
@@ -377,23 +806,57 @@ impl MockamotoNode {
...
+                function_name: "stack-stx".try_into().unwrap(),
+                function_args: vec![
+                    ClarityValue::UInt(99_000_000_000_000),
+                    pox_address.as_clarity_tuple().unwrap().into(),
+                    ClarityValue::UInt(u128::from(parent_burn_height)),
+                    ClarityValue::UInt(12),
+                ],
+            })
+        } else {
+            // NOTE: stack-extend doesn't currently work, because the PoX-4 lockup
```
jcnelson, line 832: Doesn't this block this PR, then?

testnet/stacks-node/src/mockamoto.rs
```diff
@@ -475,75 +964,14 @@ impl MockamotoNode {
         Ok(block)
     }
 
-    fn mine_and_stage_block(&mut self) -> Result<(), ChainstateError> {
+    fn mine_and_stage_block(&mut self) -> Result<u64, ChainstateError> {
         let block = self.mine_stacks_block()?;
         let config = self.chainstate.config();
         let chainstate_tx = self.chainstate.db_tx_begin()?;
```
jcnelson, line 970: This should be `self.chainstate.staging_db_tx_begin()`.  We'll ultimately be putting these blocks into a separate DB, so we should use the appropriate transaction initializer for it.

testnet/stacks-node/src/mockamoto/tests.rs
```diff
@@ -0,0 +1,60 @@
...
+        let Some(ref latest_block) = latest_block else {
+            info!("No block observed yet!");
+            continue;
+        };
+        let stacks_block_height = latest_block.get("block_height").unwrap().as_u64().unwrap();
+        info!("Block height observed: {stacks_block_height}");
+        if stacks_block_height >= 100 {
+            break true;
+        }
+    };
```
jcnelson, line 50: Should you also check to see that transactions are getting pulled from the mempool to produce blocks?

stackslib/src/chainstate/burn/operations/leader_block_commit.rs
kantai: There's at least two cases in which this would be possible:

1. A miner wins sortition, but doesn't produce any blocks.
2. During the transition from 2.5 to 3.0, the first block will fail this test.

testnet/stacks-node/src/mockamoto.rs
kantai: Not in my opinion -- the mockamoto node runs fine until it reaches 12 reward cycles, and there's very little overlap between the PoX-4 lockup work and the work in this PR, so I think they should be considered separately.

testnet/stacks-node/src/config.rs
```diff
@@ -671,7 +775,39 @@ impl Config {
...
+        if config_file.burnchain.as_ref().map(|b| b.mode.clone()) == Some(Some("mockamoto".into()))
+        {
+            let default = Self::from_config_default(ConfigFile::mockamoto(), None)?;
+            Self::from_config_default(config_file, Some(default))
+        } else {
+            Self::from_config_default(config_file, None)
+        }
+    }
+
+    fn from_config_default(
```
jferrant, line 787: This is just a personal preference. But I would make from_config_default always accept a Config (nonoptional) and then call it with Config::default() at line 781.

testnet/stacks-node/src/mockamoto/tests.rs
kantai: Yep -- expanded the test to check this in 05274b6.

testnet/stacks-node/src/mockamoto.rs
kantai: Sure -- updated in 05274b6.

testnet/stacks-node/src/mockamoto.rs
kantai: Definitely -- updated in 05274b6.

testnet/stacks-node/src/mockamoto.rs
kantai: I updated this to be the parent block ID, but I would suggest that this is _not_ being validated currently: the coordinator absolutely accepted blocks with null block header hashes.

zone117x (commented):
@kantai events for the initial boot contracts and seed balance are missing:
* The initial synthetic `token_transfer` tx containing `stx_mint` events for the seed account(s)
* These contracts: `bns`, `pox`, `costs`, cost-voting`, `genesis`.

The API tests don't work very well when the stx_mints and bns boot contract in particular are missing. Anyway to get these working?

zone117x (commented):
The `/new_block` event has the properties `pox_v1_unlock_height` and `pox_v2_unlock_height` -- can `pox_v3_unlock_height` be added?

kantai:
@zone117x -- both should be added in my last commit.

testnet/stacks-node/src/config.rs
kantai: Yep, agreed. Updated in b447984

stackslib/src/chainstate/burn/operations/leader_block_commit.rs
jcnelson: Gotcha

testnet/stacks-node/src/mockamoto.rs
jcnelson: > the coordinator absolutely accepted blocks with null block header hashes.

Oh interesting!  I'll see if I can fix this in my tenure-change validation PR.

testnet/stacks-node/src/mockamoto.rs
jcnelson: kk

zone117x (approved):
This works well for initial API integration, the events were usable to ingest TenureChange txs and NakamotoBlocks. We're still working on pox-3 force unlocks & pox-4 integration and testing. I'll open issues for any bugs found if necessary.

jcnelson (approved):
LGTM

jferrant (approved):

PR #4102, opened 2023-11-29
chore: Update `wsts` to `v5.0.0`

jbencin:
### Description

Update `wsts` to `v5.0.0`

This commit also contains the following changes:
  - Make `ThresholdSignature` wrapper around `wsts::common::signature` in order to use `verify()`
  - Eliminate `p256k1` from `Cargo.toml`
  - Use a common version of `wsts` for all workspace members

I had to update all workspace members, including `stacks-signer`, because it wouldn't build if `stackslib` was using a different version of `wsts`

### DIFF
--- a/stackslib/src/chainstate/stacks/mod.rs
+++ b/stackslib/src/chainstate/stacks/mod.rs
@@ -658,10 +658,7 @@ pub enum TenureChangeError {
 
 /// Schnorr threshold signature using types from `wsts`
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub struct ThresholdSignature {
-    R: wsts::Point,
-    z: wsts::Scalar,
-}
+pub struct ThresholdSignature(pub wsts::common::Signature);
 
 /// A transaction from Stackers to signal new mining tenure
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
--- a/stackslib/src/chainstate/stacks/transaction.rs
+++ b/stackslib/src/chainstate/stacks/transaction.rs
@@ -29,7 +29,9 @@ use stacks_common::types::StacksPublicKeyBuffer;
 use stacks_common::util::hash::{to_hex, MerkleHashFunc, MerkleTree, Sha512Trunc256Sum};
 use stacks_common::util::retry::BoundReader;
 use stacks_common::util::secp256k1::MessageSignature;
-use wsts::curve as p256k1;
+use wsts::common::Signature as Secp256k1Signature;
+use wsts::curve::point::{Compressed as Secp256k1Compressed, Point as Secp256k1Point};
+use wsts::curve::scalar::Scalar as Secp256k1Scalar;
 
 use crate::burnchains::Txid;
 use crate::chainstate::stacks::{TransactionPayloadID, *};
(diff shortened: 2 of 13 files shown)


stackslib/src/chainstate/stacks/mod.rs
```diff
@@ -658,10 +658,7 @@ pub enum TenureChangeError {
 
 /// Schnorr threshold signature using types from `wsts`
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub struct ThresholdSignature {
-    R: wsts::Point,
-    z: wsts::Scalar,
-}
+pub struct ThresholdSignature(pub wsts::common::Signature);
```
jcnelson, line 661: I think we have two structs for this in-flight -- we have `ThresholdSignature` and `SchnorrSignature`.  Can we pick one? cc @jferrant

stackslib/src/chainstate/stacks/transaction.rs
```diff
@@ -151,17 +152,18 @@ impl StacksMessageCodec for TenureChangeCause {
...
             .map_err(crate::codec::Error::WriteError)?;
-        write_next(fd, &self.z.to_bytes())?;
+        write_next(fd, &self.0.z.to_bytes())?;
         Ok(())
     }
 
     fn consensus_deserialize<R: Read>(fd: &mut R) -> Result<Self, codec_error> {
-        use p256k1::point::Compressed;
-        use wsts::{Point, Scalar};
+        use p256k1::point::{Compressed, Point};
```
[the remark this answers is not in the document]
jbencin: I can change that

stackslib/src/chainstate/stacks/mod.rs
jbencin: I don't see `SchnorrSignature` in the codebase. Is it just a type alias for `common::Signature` or is it a wrapper like `ThresholdSignature`? I implemented it this way so I could `impl StacksMessageCodec` for it, but I have no preference for what we should name it.

stackslib/src/chainstate/stacks/transaction.rs
```diff
@@ -29,6 +29,7 @@ use stacks_common::types::StacksPublicKeyBuffer;
 use stacks_common::util::hash::{to_hex, MerkleHashFunc, MerkleTree, Sha512Trunc256Sum};
 use stacks_common::util::retry::BoundReader;
 use stacks_common::util::secp256k1::MessageSignature;
+use wsts::curve as p256k1;
```
xoloki, line 32: This statement will likely make the diff rather smaller, but ideally we'd go ahead and remove any mention of `p256k1`.

Not worth holding this up, but eventually it would be good to fix this

stackslib/src/chainstate/stacks/mod.rs
jcnelson: `SchnorrSignature` is in https://github.com/stacks-network/stacks-core/pull/4039.  Maybe this PR can be rebased on it?

xoloki (approved):
`transaction.rs` contains several back references to `p256k`.  Ideally these should be removed, plus as @jcnelson noted that file has weird `fn` level `use` statements.

Overall though this PR is quite useful even with these caveats, so I'll go ahead and +1.

stackslib/src/chainstate/stacks/transaction.rs
jbencin: Moved the `use` statements to top-level and removed the `p256k1` alias

stackslib/src/chainstate/stacks/mod.rs
xoloki: @jcnelson this PR is tiny by comparison with #4039 I think this should go in first then that can be rebased on this

jcnelson: Very well

jcnelson (approved):

stackslib/src/chainstate/stacks/mod.rs
jbencin: I see... `SchnorrSignature` is implemented as a single-element tuple with a byte array, and `ThresholdSignature` is a single-element tuple with `common::Signature`. I think `ThresholdSignature` is a little simpler and allows for zero-cost, zero-risk conversions to and from `common::Signature'.

Probably best to ask @xoloki which implementation and naming convention we should go with, as there are all sorts of subtleties in cryptographic code that I don't fully understand

xoloki: Now that we have proper `serde` support I recommend the `ThresholdSignature` approach where we just wrap `common::Signature`.  When `serde` support was shakey I used to recommend byte arrays to make sure people did explicit conversions.

PR #4106, opened 2023-11-29
Implement PoX-4 Locking via Special Contract-Call Handler

### LINKED ISSUE #4090: [Nakamoto] Add `pox_locking::pox_4` lockup methods
The special-case handlers for the PoX-4 lockups need to be implemented.  This is basically a copy/paste of pox_3.rs

### LINKED ISSUE #4091: [Nakamoto] Update `handle_contract_call_special_cases()` in `stackslib/src/clarity_vm/special.rs` to handle PoX-4
The special-case handlers for PoX lockups need to be updated to handle PoX 4, and to handle a read-only PoX-3

jbencin:
### Description

Copy/paste `pox_3.rs` to `pox_4.rs` and rename things as needed. Also, fix `clippy::needless_return` lint

### Applicable issues
- fixes #4090
- fixes #4091

### Checklist
- [ ] Port the following pox3 tests to pox4
  - [ ] `delegate_extend_pox_3`
  - [ ] `delegate_stack_increase`
  - [x] `get_pox_addrs` -> `pox_lock_unlock`
  - [ ] `pox_3_delegate_stx_addr_validation`
  - [ ] `pox_3_getters`
  - [ ] `pox_auto_unlock_ab`
  - [ ] `pox_auto_unlock_ba`
  - [x] `pox_extend_transition`
  - [ ] `simple_pox_lockup_transition_pox_2`
  - [ ] `stack_aggregation_increase`
  - [ ] `stack_increase`
  - [ ] `stack_with_segwit`
- [x] Test that pox3 methods fail (`pox_3_fails`)
- [x] Test that STX locked in pox3 unlocks


pox-locking/src/pox_4.rs
```diff
@@ -0,0 +1,395 @@
...
+        "available_ustx" => out_balance.amount_unlocked(),
+        "unlock_burn_height" => out_balance.unlock_height(),
+        "account" => %principal,
+    );
+
+    snapshot.save();
+    Ok(out_balance)
+}
+
+/////////////// PoX-4 //////////////////////////////////////////
```
jcnelson, line 153: duplicate comment?

pox-locking/src/pox_4.rs
```diff
@@ -0,0 +1,395 @@
...
+    value: &Value,
+) -> Result<(), ClarityError> {
+    // Generate a synthetic print event for all functions that alter stacking state
+    let print_event_opt = if let Value::Response(response) = value {
+        if response.committed {
+            // method succeeded.  Synthesize event info, but default to no event report if we fail
+            // for some reason.
+            // Failure to synthesize an event due to a bug is *NOT* an excuse to crash the whole
+            // network!  Event capture is not consensus-critical.
+            let event_info_opt = match synthesize_pox_2_or_3_event_info(
```
jcnelson, line 343: This function `synthesize_pox_2_or_3_event_info()` should be named something more informative, now that pox-4 exists.  Perhaps `synthesize_pox_event_info()`?

friedger: I like the name. If we agree on it, I'll update the PR #4094 that changes the pox-lockup::events

friedger: Done in https://github.com/stacks-network/stacks-core/pull/4094/commits/915fba782da0009cdef509a0522bdc0476c11265

hugocaillard:
Following this issue on Clarinet: https://github.com/hirosystems/clarinet/issues/1267, we brainstormed with @lgalabru for a clean way to handle pox-4 on Clarinet (regarding the concatenation with `pox-<netowrk>.clar`).

While we all agree that it would be nice to have a standard way to handle such use cases at the Clarity or Clarinet level, it may take a bit of time to get there.

In the meantime, regarding the pox-4 contract specifically, what about leveraging **`is-in-mainnet`** that was introduced in Clarity 2? With the constants defined based on that, it would completely remove the need for contract concatenation. 
```clar
(define-constant PREPARE_CYCLE_LENGTH (if is-in-mainnet u100 u50))
```

jbencin:
@moodmosaic I managed to fix one test `get_pox_addrs()` by doing the following:
- Disable epoch 3.0 and nakamoto blocks. We can test pox4 in Epoch 2.5
- Experiment with stacking start height. We submit the stacking txs at `first_v4_cycle - 3`
- Experiment with when we start checking things, at `first_v4_cycle + 5`

I still don't understand why these block heights work, but the tests clearly show STX are being locked for a couple cycles, then unlocked, so it looks okay to me

Also, going to give you write access to this branch so you can push (since it's in my repo)

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -282,9 +282,7 @@ fn test_simple_nakamoto_coordinator_bootup() {
 /// Mine a single Nakamoto tenure with 10 Nakamoto blocks
 #[test]
 fn test_simple_nakamoto_coordinator_1_tenure_10_blocks() {
-    let mut test_signers = TestSigners::default();
-    let mut peer = boot_nakamoto(function_name!(), vec![], test_signers.aggregate_public_key);
-    let private_key = peer.config.private_key.clone();
+    let private_key = StacksPrivateKey::new();
```
jcnelson, line 285: Why did you change this?

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -0,0 +1,4399 @@
+use std::collections::{HashMap, HashSet, VecDeque};
```
jcnelson, line 1: Per our meeting earlier, I think the only things that this file needs to test are:

* PoX 4 instantiates 
* STX locked in PoX 3 unlock automatically
* Once PoX 4 is active, only PoX 4 stacking methods work.  PoX 3 ones fail.

There's no reason to test PoX 2 and PoX behavior, since both these contracts have been defunct for some time.

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
jbencin: `test_simple_nakamoto_coordinator_1_tenure_10_blocks()` and `test_simple_nakamoto_coordinator_10_tenures_10_blocks()` started failing as soon as I added support for pox4 in `pox-locking/src/lib.rs`. Aaron suggested this as a way to get the coordinator tests working again

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
moodmosaic: @jcnelson
>* PoX 4 instantiates

does [`get_pox_addrs`](https://github.com/stacks-network/stacks-core/pull/4106/files#diff-1b7e3da4025244bdde5f977c36554b94ab89e844c05e4c3c0a106dd9da9c7636R3369) captures this? If it does, I would like to start the current `pox_4_tests.rs` (in this PR) from a clean state and start by just adding this test, and then keep adding to it.

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
jcnelson: Ah, okay.  I can see why that is -- now that the special-case handlers are in place, stacking would have locked up all of the STX tied to `peer.config.private_key`, which would prevent this test from running (since it sends STX from this key).  Can you at least use a constant private key so the addresses in this test are deterministic?

kantai, line 285: Yep -- I'd recommend using a constant private key, e.g.:

```suggestion
    let private_key = StacksPrivateKey::from_seed(&[2]);
```

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -383,9 +388,7 @@ fn test_simple_nakamoto_coordinator_1_tenure_10_blocks() {
 /// Mine a 10 Nakamoto tenures with 10 Nakamoto blocks
 #[test]
 fn test_simple_nakamoto_coordinator_10_tenures_10_blocks() {
-    let mut test_signers = TestSigners::default();
-    let mut peer = boot_nakamoto(function_name!(), vec![], test_signers.aggregate_public_key);
-    let private_key = peer.config.private_key.clone();
+    let private_key = StacksPrivateKey::new();
```
kantai, line 391: I would recommend using a constant private key here as well.

kantai (approved):
LGTM -- I would just recommend using `from_seed` to initialize the necessary new StacksPrivateKeys in `nakamoto::coordinator::tests`

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -0,0 +1,1150 @@
...
```
jcnelson, line 1: Missing copyright comment block

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -0,0 +1,1150 @@
...
+        BOB_LOCKUP,
+        PoxAddress::from_legacy(
+            AddressHashMode::SerializeP2PKH,
+            key_to_stacks_addr(&bob).bytes,
+        ),
+        3,
+        tip.block_height,
+    );
+
+    // Alice _will_ auto-unlock: she can stack-extend in PoX v2
```
jcnelson, line 400: Shouldn't this be `Alice _will not_ auto-unlock`?  Also, wouldn't Alice have unlocked from pox-2 activating?

kantai: No -- the comment is confusing, but "correct". Alice stacked in pox v1 and so would have experienced auto-unlock but instead she stack-extends in PoX v2.

The v2 unlock height configuration setting _doesn't_ need to match epoch2.1 start (and it doesn't in these tests) -- the auto-unlock occurs after pox-2 is activated, which is why Alice can extend: she's currently locked, and she's extending so she doesn't miss a cycle.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -0,0 +1,1166 @@
...
+    info!("Checking STX unlocked after {lock_period} cycles");
+    for _ in 0..burnchain.pox_constants.reward_cycle_length {
+        latest_block = peer.tenure_with_txs(&[], &mut coinbase_nonce);
+        assert_latest_was_burn(&mut peer);
+    }
+}
+
+/// Test that pox3 methods fail once pox4 is activated
+#[test]
+fn pox_3_fails() {
```
jcnelson, line 878: Can you call this something more descriptive, like "test_pox_3_defunct()`?

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -0,0 +1,1166 @@
...
+
+    // Check that STX are locked for 2 reward cycles
+    for _ in 0..2 {
+        let tip = get_tip(peer.sortdb.as_ref());
+        let cycle = burnchain
+            .block_height_to_reward_cycle(tip.block_height)
+            .unwrap();
+        info!("Checking STX locked for cycle {cycle}");
+        for i in 0..reward_blocks {
+            latest_block = peer.tenure_with_txs(&[], &mut coinbase_nonce);
```
jcnelson, line 1053: Can you actually check the reward set itself?

jbencin: This line inserts the address into `rewarded` and checks it a few lines later. Is that sufficient or did you mean something else?

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
jbencin: Will do

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
jcnelson: That works!  Didn't catch that

jcnelson (approved):
LGTM; just please address my remaining concerns before merging.  Thanks!

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
jbencin: Used Aaron's suggestion of `StacksPrivateKey::from_seed()`

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -0,0 +1,1166 @@
...
+    for _ in 0..=lock_period {
+        let tip = get_tip(peer.sortdb.as_ref());
+        let cycle = burnchain
+            .block_height_to_reward_cycle(tip.block_height)
+            .unwrap();
+        info!("Checking no stackers for cycle {cycle}");
+        for _ in 0..burnchain.pox_constants.reward_cycle_length {
+            latest_block = peer.tenure_with_txs(&[], &mut coinbase_nonce);
+            // Should all be burn because no stackers
+            assert_latest_was_burn(&mut peer);
```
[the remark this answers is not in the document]
jbencin: Trying to figure out how to do this, I think I understand, but want to check:

- Get `StacksTransactionReceipt`s from latest block after executing transactions
- Filter the ones that contain a `StacksTransactionEvent` with a `pox-3` call
- Check that `StacksTransactionReceipt.result` is an error value

Does this sound right?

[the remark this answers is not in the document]
jcnelson: Yes; if possible, also test that we specifically fail because contract was defunct (as opposed to some other error that might arise due to a regression in the test).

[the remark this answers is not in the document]
kantai: I'm not sure that the `StacksTransactionReceipt`s will be very discerning for these checks -- runtime errors (like these should be) just return a receipt with a result of `err_none`, an empty assetmap, and an empty event list. But, you could usefully assert that a runtime error occurs by checking the receipt.

This is how `pox_3_tests` does this: https://github.com/stacks-network/stacks-core/blob/develop/stackslib/src/chainstate/stacks/boot/pox_3_tests.rs#L512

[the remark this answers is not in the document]
kantai: Ah, that's because `pox-3` doesn't seem to actually be marked defunct in these changes. You need to do something like:

```diff
diff --git a/pox-locking/src/lib.rs b/pox-locking/src/lib.rs
index 05303a4d9..057667341 100644
--- a/pox-locking/src/lib.rs
+++ b/pox-locking/src/lib.rs
@@ -107,6 +107,19 @@ pub fn handle_contract_call_special_cases(
             result,
         );
     } else if *contract_id == boot_code_id(POX_3_NAME, global_context.mainnet) {
+        if !pox_3::is_read_only(function_name) && global_context.epoch_id >= StacksEpochId::Epoch25
+        {
+            warn!("PoX-3 function call attempted on an account after Epoch 2.5";
+                  "v3_unlock_ht" => global_context.database.get_v3_unlock_height(),
+                  "current_burn_ht" => global_context.database.get_current_burnchain_block_height(),
+                  "function_name" => function_name,
+                  "contract_id" => %contract_id
+            );
+            return Err(ClarityError::Runtime(
+                RuntimeErrorType::DefunctPoxContract,
+                None,
+            ));
+        }
         return pox_3::handle_contract_call(
             global_context,
             sender,
diff --git a/pox-locking/src/pox_3.rs b/pox-locking/src/pox_3.rs
index cccfbb264..e0110987e 100644
--- a/pox-locking/src/pox_3.rs
+++ b/pox-locking/src/pox_3.rs
@@ -32,6 +32,33 @@ use crate::pox_2::{parse_pox_extend_result, parse_pox_increase, parse_pox_stacki
 use crate::{LockingError, POX_3_NAME};
 
 /////////////////////// PoX-3 /////////////////////////////////
+/// is a PoX-3 function call read only?
+pub(crate) fn is_read_only(func_name: &str) -> bool {
+    "get-pox-rejection" == func_name
+        || "is-pox-active" == func_name
+        || "burn-height-to-reward-cycle" == func_name
+        || "reward-cycle-to-burn-height" == func_name
+        || "current-pox-reward-cycle" == func_name
+        || "get-stacker-info" == func_name
+        || "get-check-delegation" == func_name
+        || "get-reward-set-size" == func_name
+        || "next-cycle-rejection-votes" == func_name
+        || "get-total-ustx-stacked" == func_name
+        || "get-reward-set-pox-address" == func_name
+        || "get-stacking-minimum" == func_name
+        || "check-pox-addr-version" == func_name
+        || "check-pox-addr-hashbytes" == func_name
+        || "check-pox-lock-period" == func_name
+        || "can-stack-stx" == func_name
+        || "minimal-can-stack-stx" == func_name
+        || "get-pox-info" == func_name
+        || "get-delegation-info" == func_name
+        || "get-allowance-contract-callers" == func_name
+        || "get-num-reward-set-pox-addresses" == func_name
+        || "get-partial-stacked-by-cycle" == func_name
+        || "get-total-pox-rejection" == func_name
+}
+
 
 /// Lock up STX for PoX for a time.  Does NOT touch the account nonce.
 pub fn pox_lock_v3(
```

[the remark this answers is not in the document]
jbencin: Thanks @kantai, that fixed it! Will check in and merge this ASAP

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -949,6 +955,28 @@ fn pox_3_defunct() {
...
+
+    let receipts = last_observer_block.receipts
+        .iter()
+        .filter(|receipt| match &receipt.result {
+            Value::Response(r) => !r.committed,
+            _ => false,
+        })
+        .collect::<Vec<_>>();
+
+    assert_eq!(receipts.len(), 4);
```
kantai, line 978: Can you make sure that every transaction failed, i.e. `assert_eq!(receipts.len(), txs.len()`?

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -949,6 +955,28 @@ fn pox_3_defunct() {
...
+        .last()
+        .unwrap()
+        .clone();
+
+    let receipts = last_observer_block.receipts
+        .iter()
+        .filter(|receipt| match &receipt.result {
+            Value::Response(r) => !r.committed,
+            _ => false,
+        })
```
kantai, line 975: Can you check that the `result` is specifically `(err none)`?

kantai (approved):
LGTM, just a couple tweaks to the test case.

rondales:
> ### Description
> Copy/paste `pox_3.rs` to `pox_4.rs` and rename things as needed. Also, fix `clippy::needless_return` lint
> 
> ### Applicable issues
> * fixes [[Nakamoto] Add `pox_locking::pox_4` lockup methods #4090](https://github.com/stacks-network/stacks-core/issues/4090)
> * fixes [[Nakamoto] Update `handle_contract_call_special_cases()` in `stackslib/src/clarity_vm/special.rs` to handle PoX-4 #4091](https://github.com/stacks-network/stacks-core/issues/4091)
> 
> ### Additional info (benefits, drawbacks, caveats)
> ### Checklist
> * [ ]  Port the following pox3 tests to pox4
>   
>   * [ ]  `delegate_extend_pox_3`
>   * [ ]  `delegate_stack_increase`
>   * [x]  `get_pox_addrs` -> `pox_lock_unlock`
>   * [ ]  `pox_3_delegate_stx_addr_validation`
>   * [ ]  `pox_3_getters`
>   * [ ]  `pox_auto_unlock_ab`
>   * [ ]  `pox_auto_unlock_ba`
>   * [x]  `pox_extend_transition`
>   * [ ]  `simple_pox_lockup_transition_pox_2`
>   * [ ]  `stack_aggregation_increase`
>   * [ ]  `stack_increase`
>   * [ ]  `stack_with_segwit`
> * [x]  Test that pox3 methods fail (`pox_3_fails`)
> * [x]  Test that STX locked in pox3 unlocks

PR #4110, opened 2023-11-30
feat(wasm): include proper rollback handling for contract-calls

obycode:
This code can still use some improvements and DRYing up.


hugocaillard (approved):
LGTM!

PR #4112, opened 2023-12-01
Feat/remove sbtc ops

### LINKED ISSUE #4072: [Nakamoto] Drop old peg-in/peg-out burnchain ops
They're not going to be used in the current sBTC design, and sBTC is going to ship later.  If we want them back, they're in the git history.

soju-drinker:
### Description

Remove SBTC operations

Closes #4072


stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -403,112 +402,6 @@ impl FromRow<DelegateStxOp> for DelegateStxOp {
     }
 }
 
-impl FromRow<PegInOp> for PegInOp {
```
jcnelson, line 406: Can you also remote the sqlite tables in the sortition DB schema?

jcnelson (approved):
LGTM. Just make sure all the tests pass before merging.  Thanks!

jferrant (approved):

PR #4113, opened 2023-12-01
Feat/mockamoto genesis block routine

### LINKED ISSUE #4028: Nakamoto: Implement a neon genesis block generation routine
In Neon operation, a single miner produces nakamoto blocks using bitcoin regtest. This single miner _does not_ communicate with stackers in order to obtain valid TenureChange transactions. Instead, the neon miner produces and signs `TenureChange` transactions by themselves. In order for those transactions to validate, the miner's public key must be the PoX-4 contract's active aggregate public key.

In order to do this, during the "genesis" block of the neon miner, the pox-4 contract should be explicitly instantiated with the aggregate public key.

The neon node initiates the chainstate and executes genesis instructions here:

https://github.com/stacks-network/stacks-core/blob/feat/nakamoto-coordinator/testnet/stacks-node/src/run_loop/neon.rs#L444

jferrant:
### Description
This does the following:
- It adds a bootup callback to instantiation to set the aggregate public key (required as pox-4 is not initialized until after genesis block).
- It adds a routine to the pox-4 instantiation to check if its in testnet, and if so, it checks if an initial agg key has been written.
- It makes sure to set al pre-pox-4 reward cycles aggregate public keys
- Every mockamoto block sets the aggregate public key for the following reward cycle
- It pulls the transaction creation into sep functions to cleanup the super long mine_stacks_block

Note every commit is a separation of concerns so feel free to review commit at a time if that is easier to grok.

### Applicable issues
- Closes https://github.com/stacks-network/stacks-core/issues/4028

### DIFF
--- a/testnet/stacks-node/src/mockamoto.rs
+++ b/testnet/stacks-node/src/mockamoto.rs
@@ -5,8 +5,10 @@ use std::thread;
 use std::thread::{sleep, JoinHandle};
 use std::time::Duration;
 
+use clarity::boot_util::boot_code_id;
 use clarity::vm::ast::ASTRules;
-use clarity::vm::Value as ClarityValue;
+use clarity::vm::clarity::TransactionConnection;
+use clarity::vm::{ClarityVersion, Value as ClarityValue};
 use lazy_static::lazy_static;
 use stacks::burnchains::bitcoin::address::{
     BitcoinAddress, LegacyBitcoinAddress, LegacyBitcoinAddressType,
@@ -33,6 +35,9 @@ use stacks::chainstate::nakamoto::{
     NakamotoBlock, NakamotoBlockHeader, NakamotoChainState, SetupBlockResult,
 };
 use stacks::chainstate::stacks::address::PoxAddress;
+use stacks::chainstate::stacks::boot::{
+    BOOT_TEST_POX_4_AGG_KEY_CONTRACT, BOOT_TEST_POX_4_AGG_KEY_FNAME,
+};
 use stacks::chainstate::stacks::db::{ChainStateBootData, ClarityTx, StacksChainState};
 use stacks::chainstate::stacks::miner::{
     BlockBuilder, BlockBuilderSettings, BlockLimitFunction, MinerStatus, TransactionResult,
@@ -64,7 +69,7 @@ use stacks_common::types::chainstate::{
     StacksPrivateKey, VRFSeed,
 };
 use stacks_common::types::{PrivateKey, StacksEpochId};
-use stacks_common::util::hash::{Hash160, MerkleTree, Sha512Trunc256Sum};
+use stacks_common::util::hash::{to_hex, Hash160, MerkleTree, Sha512Trunc256Sum};
 use stacks_common::util::secp256k1::{MessageSignature, Secp256k1PublicKey};
 use stacks_common::util::vrf::{VRFPrivateKey, VRFProof, VRFPublicKey, VRF};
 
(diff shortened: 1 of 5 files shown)


testnet/stacks-node/src/mockamoto.rs
```diff
@@ -975,3 +992,117 @@ impl MockamotoNode {
         Ok(chain_length)
     }
 }
+
+// Helper function to make a signed tenure change transaction
```
jcnelson, line 996: Thank you for separating these out :pray:

testnet/stacks-node/src/mockamoto/tests.rs
```diff
@@ -1,25 +1,19 @@
...
 use crate::neon_node::PeerThread;
-use crate::tests::make_stacks_transfer;
 use crate::tests::neon_integrations::test_observer;
-use crate::tests::to_addr;
-use crate::Config;
-use crate::ConfigFile;
-
-use super::MockamotoNode;
+use crate::tests::{make_stacks_transfer, to_addr};
+use crate::{Config, ConfigFile};
```
jcnelson, line 16: Is there a way you can add a simple test to verify that the aggregate public key is getting set, and that `get_aggregate_public_key()` works?

testnet/stacks-node/src/mockamoto/tests.rs
```diff
@@ -126,3 +127,156 @@ fn observe_100_blocks() {
...
+    let aggregate_tx = make_contract_call(
+        &submitter_sk,
+        0,
+        10,
+        &boot_code_addr(false),
+        POX_4_NAME,
+        "set-aggregate-public-key",
+        &[Value::UInt(u128::from(reward_cycle)), aggregate_public_key],
+    );
+    let aggregate_tx_hex = format!("0x{}", to_hex(&aggregate_tx));
```
kantai, line 218: Can this test run without this `contract-call`, and instead line 254 asserting that `aggregate_key == node.self_signer.aggregate_public_key`? That way, it's checking that the mockamoto setup code itself is successfully setting the aggregate key.

kantai (approved):
LGTM! Just had one comment on testing the aggregate key setup.

testnet/stacks-node/src/mockamoto/tests.rs
jferrant: I will update to do both! Cause then it will test but contract call to set as well as code logic to init it correctly

testnet/stacks-node/src/mockamoto.rs
```diff
@@ -961,7 +957,28 @@ impl MockamotoNode {
...
-        let aggregate_public_key = self.self_signer.aggregate_public_key;
+        let aggregate_public_key = if chain_length <= 1 {
+            self.self_signer.aggregate_public_key
+        } else {
+            let block_sn = SortitionDB::get_block_snapshot_consensus(
+                sortition_handle.conn(),
+                &block.header.consensus_hash,
+            )?
+            .ok_or(ChainstateError::DBError(DBError::NotFoundError))?;
+            // TODO: https://github.com/stacks-network/stacks-core/issues/4109
```
jcnelson, line 968: Is this meant to be addressed in this PR?

jcnelson (approved):
LGTM; just one question

testnet/stacks-node/src/mockamoto.rs
jferrant: No it is not meant to be addressed in this PR as far as I am aware. @kantai ?

kantai: No, I don't think so.

jferrant:
Replaced by https://github.com/stacks-network/stacks-core/pull/4165#pullrequestreview-1780071969

Issue #4053, opened 2023-11-13
[Nakamoto] Tenures table


jcnelson:
We need to treat tenures (and tenure-changes) as first-class objects in the Nakamoto chain state in order to (1) validate tenure changes and (2) deal with missed or invalid Stacks block-commits.

Issue #4014, opened 2023-10-27
Nakamoto: Tenure Change transaction Validation


kantai:
In order to support nakamoto in neon and beyond, the tenure change transaction must be validated further than in mockamoto.

* The block must be the first block in the miner's tenure. It cannot be in a subsequent block
* TenureChange must be for last_tenure + 1. No TenureChanges can be skipped.
* It must point to the last active miner's block-commit's StacksBlockId
* It must count up the number of blocks since the last block with a TenureChange (it may be zero; a block can have multiple TenureChange transactions per the design doc)
* It must be signed by the current Stacker set
* It must precede any other kinds of transactions (except coinbases?)

jcnelson:
I'm already working on this as part of the coordinator PR, since I need to be able to generate Nakamoto block streams as part of the test infrastructure (`TestMiner`, `TestBurnchainBlock`, `TestPeer`).

kantai:
Which of these are necessary for the coordinator PR?

jcnelson:
Not all of them.  Just the ability to ensure that there is one `TenureChange` per tenure, and it's in the right place in the block stream.  I'm just trying to upgrade `TestPeer` to produce well-formed Nakamoto block-commits and block streams so I can verify that the coordinator will correctly detect the presence of and correctly handle new burnchain blocks and new Nakamoto blocks.  I'm focusing on getting `TestPeer` upgraded so I can switch over to implementing the network state machines.

jcnelson:
I'll add a follow-on PR that performs all of the necessary validations.  I just wanted to point out that I'm already well-positioned to take this task, since I had to do a few of them already.

saralab:
@jcnelson : To add more details on tasks needed for tenurechange validation

PR #4114, opened 2023-12-01
Feat/tenure change validation

jcnelson:
This PR addresses #4014 by way of #4053.  Tenure-change transactions are now fully validated.  In particular:

* Tenures can now span multiple sortitions if there was no new miner chosen
* Nakamoto can keep making blocks even if a miner does not produce any blocks in its sortition
* Tenures can have their runtime budgets extended

Leaving as a draft for now so I can add more unit tests and let CI run on the existing ones.


stacks-common/src/types/chainstate.rs
```diff
@@ -15,7 +15,7 @@ use crate::consts::{FIRST_BURNCHAIN_CONSENSUS_HASH, FIRST_STACKS_BLOCK_HASH};
 use crate::deps_common::bitcoin::util::hash::Sha256dHash;
 use crate::util::hash::{to_hex, DoubleSha256, Hash160, Sha512Trunc256Sum, HASH160_ENCODED_SIZE};
 use crate::util::secp256k1::{
-    MessageSignature, SchnorrSignature, Secp256k1PrivateKey, Secp256k1PublicKey,
+    MessageSignature, /* SchnorrSignature, */ Secp256k1PrivateKey, Secp256k1PublicKey,
```
xoloki, line 18: Why comment this out instead of removing it?  I thought we'd basically agreed to just use the `ThresholdSignature` wrapper around `wsts::common::Signature`?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -605,44 +673,73 @@ impl NakamotoBlock {
...
+                // wrong position
+                return Some(false);
+            }
+
+            // must be a non-sortition-triggered tenure change
+            let TransactionPayload::TenureChange(tc_payload) = &self.txs[0].payload else {
+                // this transaction is not a tenure change
+                // (should be unreachable)
+                return Some(false);
+            };
```
kantai, line 699: Can you DRY some of this checking up with what's going on in the next check as well? It seems like in all cases, the tenure_change_transaction _must_ be at position 0, and you want to access the payload in any event so you can check `.expects_sortition`:

stacks-common/src/util/secp256k1.rs
```diff
@@ -118,6 +118,7 @@ impl Default for Secp256k1PublicKey {
     }
 }
 
+/*
```
xoloki, line 121: Same.  Unless you really want to use byte arrays explicitly?  The current `serde` impl in `p256k1` will give you bytes when you need them and fast math objects otherwise.

stacks-common/src/util/secp256k1.rs
```diff
@@ -156,6 +157,7 @@ impl From<&WSTSSignature> for SchnorrSignature {
         SchnorrSignature(buf)
     }
 }
+*/
 
 impl Secp256k1PublicKey {
```
xoloki, line 162: Ideally we'd also move away from other `libsecp256k1` wrappers.  `p256k1` wraps everything, including public keys, ecdsa/schnorr sigs, etc.

stackslib/src/chainstate/nakamoto/tests/node.rs
```diff
@@ -373,93 +370,90 @@ impl TestStacksNode {
...
+                .unwrap();
 
-            test_debug!(
-                "Work in {} {} for Nakamoto parent: {},{}",
-                burn_block.block_height,
-                burn_block.parent_snapshot.burn_header_hash,
-                parent_sortition.total_burn,
-                last_parent.header.chain_length + 1,
-            );
+                test_debug!(
```
xoloki, line 389: The diff seems to be implying the indentation is weird here, was this properly `cargo fmt`'d?

xoloki (approved):
Minor comments about crypto libs and little things, but LGTM.  However I'm not well versed in this code.

stacks-common/src/types/chainstate.rs
jcnelson: Because I forgot to delete it.  Will remove

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1921,52 +2102,362 @@ impl NakamotoChainState {
...
+                  "tenure_payload.previous_tenure_blocks" => tenure_payload.previous_tenure_blocks
+            );
+            return Ok(false);
+        }
+
+        Ok(true)
+    }
+
+    /// Advance the tenures table with a validated block's tenure data.
+    /// Only stores tenures that are paired with sortitions
```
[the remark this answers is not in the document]
jcnelson: Yup, that's a bug -- it shouldn't store tenure extensions

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -2486,6 +3174,25 @@ impl NakamotoChainState {
             None
         };
 
+        // process the tenure-change if it happened, so that when block-processing begins, it happens in whatever the
+        // current tenure is
```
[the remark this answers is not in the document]
jcnelson: I'm not sure it's significant, however, because once the miner produces the tenure-start block, and it gets signed, they'll immediately apply it to their chainstate before building the next block.  This will cause `advance_nakamoto_tenure` to get called.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -605,44 +673,73 @@ impl NakamotoBlock {
...
-        for (i, pos) in tenure_change_positions.iter().enumerate() {
-            if &i != pos {
-                // tenure-change is out of place
+            if tc_payload.cause.expects_sortition() {
+                // not valid
                 return Some(false);
             }
+
+            // not a tenure-start block, but syntactically valid w.r.t. tenure changes
+            return None;
```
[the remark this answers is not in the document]
jcnelson: Re (2), the consensus hash of a tenure-extension _must be the same as_ the last winning block-commit's sortition's consensus hash.  This is true for all tenure-changes -- we identify tenures by their block-commit's sortition's consensus hashes, and the `consensus_hash` field in a tenure change binds it to that tenure.

The reason for this is because calculating block rewards and maturity becomes really tricky otherwise, and moreover, does not provide the same incentives.  Namely, we want miners to get rewarded only when the _tenure_ gets sufficiently confirmed by subsequent tenures -- we want them to keep working on the chain in order to get paid.  But to do this, we'd only count a reward as matured if there have been enough sortition-induced tenure changes.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1921,52 +2102,362 @@ impl NakamotoChainState {
...
+    pub(crate) fn check_nakamoto_tenure(
+        headers_conn: &Connection,
+        sort_tx: &mut SortitionHandleTx,
+        block_header: &NakamotoBlockHeader,
+        tenure_payload: &TenureChangePayload,
+    ) -> Result<bool, ChainstateError> {
+        if !tenure_payload.cause.expects_sortition() {
+            // not paired with a sortition
+            return Ok(true);
+        }
```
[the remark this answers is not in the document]
jcnelson: Yes.  Also, its consensus hash and previous consensus hash must both match the ongoing tenure's consensus hash.

stackslib/src/clarity_vm/clarity.rs
```diff
@@ -1651,6 +1651,14 @@ impl<'a, 'b> ClarityTransactionConnection<'a, 'b> {
...
     }
 
+    /// Reset the cost tracker to the given cost
+    pub fn reset_cost(&mut self, new_cost: ExecutionCost) {
+        match self.cost_track {
+            Some(ref mut track) => track.set_total(new_cost),
+            None => { /* no-op */ }
+        }
+    }
+
```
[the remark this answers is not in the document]
jcnelson: Sure

stackslib/src/chainstate/nakamoto/mod.rs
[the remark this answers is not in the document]
kantai: > the consensus hash of a tenure-extension _must be the same as_ the last winning block-commit's sortition's consensus hash. This is true for all tenure-changes -- we identify tenures by their block-commit's sortition's consensus hashes, and the `consensus_hash` field in a tenure change binds it to that tenure.

There are two consensus hash fields in the tenure changes, right? The “previous consensus hash” must match the prior tenure, but the new consensus hash does not always — in the case of a sortition, it would never, but in extensions due to a missed sortition as well. In that last case, referencing such as a single tenure (i.e., burn block 1 _and_ the extended to burn block 2) becomes tricky, because all the blocks of the tenure do not have the same consensus hash.

> The reason for this is because calculating block rewards and maturity becomes really tricky otherwise, and moreover, does not provide the same incentives.

I mostly agree with this: miner rewards and maturity shouldn’t “tick” on non-sortition tenure change txs. However, I think that if the implementation treats every tenure change transaction as a new tenure, this property can still be implemented (my suggestion was to essentially count coinbase transactions, which are only allowed when a tenure change occurs due to sortition).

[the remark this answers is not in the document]
jcnelson: > because all the blocks of the tenure do not have the same consensus hash.

This is no longer true.  The `consensus_hash` field in `NakamotoBlockHeader` identifies the tenure, and the tenure is defined as the sequence of blocks produced by the last sortition-selected miner.  Subsequent Bitcoin blocks and failed sortitions can happen, but the consensus hash of the ongoing stream of Nakamoto blocks will remain the same as that of the sortition which selected the currently-active miner.

This differs from the SIP, btw.  I need to update the SIP.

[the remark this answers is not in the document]
kantai: How do extended blocks pick up the new burn chain changes, then?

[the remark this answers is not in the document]
jcnelson: I'm in the process of tweaking the `TenureChangePayload` and `tenure_headers` table to track three consensus hashes:

* `tenure_consensus_hash` -- the consensus hash that identifies the winning miner (i.e. it's the consensus hash of the sortition in which the winning miner was selected)
* `prev_tenure_consensus_hash` -- the consensus hash of the previous winning miner.  It's not guaranteed to be the parent sortition consensus hash of `tenure_consensus_hash`, since the miner's tenure may span multiple sortitions
* `sortition_consensus_hash` -- the consensus hash of the sortition which triggered Stackers to produce this `TenureChange` transaction

The `sortition_consensus_hash` will identify new burnchain changes.  However, it only does so when there is an active miner.  As such, the `sortition_consensus_hash` values in `nakamoto_tenures` don't correspond to sortitions 1-to-1 -- there can be sortitions that do not have a corresponding `nakamoto_tenures` row, because it's possible that there were no active miners during a sortition.

What I expect the Clarity VM to do (specifically, the `BurnStateDB` implementation) to handle `get-burn-block-info?` is to perform an ancestor query from the highest-seen `sortition_consensus_hash`'s sortition back to the block hash given in `get-burn-block-info?` in order to (1) determine that the block exists and (2) identify where that block falls in the burnchain history.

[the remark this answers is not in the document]
kantai: So you are saying that if a sortition doesn’t occur, then new burn chain operations will not be processed until the next sortition occurs.

What is the advantage of treating tenures this way?

stackslib/src/chainstate/stacks/mod.rs
```diff
@@ -627,10 +631,8 @@ impl_byte_array_serde!(TokenTransferMemo);
 pub enum TenureChangeCause {
     /// A valid winning block-commit
     BlockFound = 0,
-    /// No winning block-commits
-    NoBlockFound = 1,
-    /// A null miner won the block-commit
-    NullMiner = 2,
+    /// The next burnchain block is taking too long, so extend the runtime budget
+    Extended = 1,
 }
```
zone117x, line 636: It looks like this implementation differs from what is currently described in [SIP-021](https://github.com/stacksgov/sips/blob/feat/sip-021-nakamoto/sips/sip-021/sip-021-nakamoto.md) -- rather than two distinct tx types `TenureChange` and `TenureExtension`, this PR bundles the functionality of both into one tx type. Is this an intended change?

jcnelson: Yes, it's intentional.  `TenureExtend` is no longer a transaction type; instead there are two different variants of `TenureChange`.

jbencin (commented):
Made some comments, minor stuff. I'll let Aaron give the second approval though because I'm not confident that I understand all the details here

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -605,44 +673,73 @@ impl NakamotoBlock {
             return None;
         }
 
-        if coinbase_positions.len() > 1 {
-            // has more than one coinbase
+        if coinbase_positions.len() > 1 || tenure_change_positions.len() > 1 {
```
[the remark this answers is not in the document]
jcnelson: That's no longer the case in this PR.  It's easier to simply capture all of that information within a single TenureChange.

stackslib/src/chainstate/stacks/transaction.rs
```diff
@@ -195,40 +192,27 @@ impl ThresholdSignature {
...
-
 impl StacksMessageCodec for TenureChangePayload {
     fn consensus_serialize<W: Write>(&self, fd: &mut W) -> Result<(), codec_error> {
+        write_next(fd, &self.consensus_hash)?;
+        write_next(fd, &self.prev_consensus_hash)?;
         write_next(fd, &self.previous_tenure_end)?;
         write_next(fd, &self.previous_tenure_blocks)?;
         write_next(fd, &self.cause)?;
         write_next(fd, &self.pubkey_hash)?;
+        write_next(fd, &self.signature)?;
```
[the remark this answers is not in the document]
jcnelson: No; signature only includes the data above it. The signers sign `signature`.

stackslib/src/chainstate/nakamoto/mod.rs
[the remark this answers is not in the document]
jcnelson: Take a look at what I have now in `src/chainstate/nakamoto/tenure.rs`

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -2729,41 +3371,18 @@ impl NakamotoChainState {
...
             header: new_tip,
             tx_receipts,
             matured_rewards,
-            matured_rewards_info,
+            matured_rewards_info: matured_rewards_info_opt,
             parent_microblocks_cost: ExecutionCost::zero(),
             anchored_block_cost: block_execution_cost,
             parent_burn_block_hash,
-            parent_burn_block_height,
+            parent_burn_block_height: u32::try_from(parent_burn_block_height).unwrap_or(0), // shouldn't be fatal
```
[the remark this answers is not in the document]
jcnelson: I wanted the error to be very obvious

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -563,16 +608,127 @@ impl NakamotoBlock {
...
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
+
+        if coinbase_positions.len() > 0 {
+            // can't be
+            return None;
+        }
```
kantai, line 639:
```suggestion
        let contains_coinbase = self
            .txs
            .iter()
            .find(|(tx)| matches!(&tx.payload, TransactionPayload::Coinbase(_)))
            .is_some();

        if contains_coinbase {
            // can't be a tenure-extend, would have to be a tenure-begin
            return None;
        }
```

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -563,16 +608,127 @@ impl NakamotoBlock {
...
+        if tenure_change_positions.len() == 0 {
+            return None;
+        }
+
+        if tenure_change_positions.len() > 1 {
+            // invalid
+            debug!(
+                "Invalid block -- {} tenure txs",
+                tenure_change_positions.len()
+            );
```
kantai, line 664: Should be a warn, and more informative:

```rust
warn!("Invalid block: tenure-extension block contains multiple TenureChange transactions"; "tenure_change_txs" => tenure_change_positions.len());
```

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -563,16 +608,127 @@ impl NakamotoBlock {
...
+            || tc_payload.prev_tenure_consensus_hash != self.header.consensus_hash
+        {
+            // tenure-extends don't change the current miner
+            debug!(
+                "Invalid block -- expected {} = {} && {} = {}",
+                &tc_payload.tenure_consensus_hash,
+                &self.header.consensus_hash,
+                &tc_payload.prev_tenure_consensus_hash,
+                &self.header.consensus_hash
+            );
```
kantai, line 714: Warn and more descriptive:

```
warn!("Invalid block: tenure extended block must have the same consensus hash as the tenure change transaction"; ...);
```

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -563,16 +608,127 @@ impl NakamotoBlock {
...
+    /// this block (this checks `cause` and `previous_tenure_end`)
+    /// * It then has a Nakamoto coinbase
     /// * Coinbases and TenureChanges do not occur anywhere else
     ///
     /// Returns Some(true) if the above are true
     /// Returns Some(false) if this block has at least one coinbase or TenureChange tx, but one of
     /// the above checks are false
     /// Returns None if this block has no coinbase or TenureChange txs
-    pub fn is_wellformed_first_tenure_block(&self) -> Option<bool> {
+    pub fn is_wellformed_tenure_start_block(&self) -> Option<bool> {
```
kantai, line 731: As in the other `is_wellformed_` method, anything returning `Some(false)` needs to emit a `warn!` (rather than `debug!`) and should be descriptive enough that someone can understand why the block was invalid without looking at this code.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -563,16 +608,127 @@ impl NakamotoBlock {
...
+    /// Determine if this is a well-formed tenure-extend block.
+    /// * It has exactly one TenureChange, and it does _not_ require a sortiton (it's `cause` is
+    /// `Extended`)
+    /// * Its consensus hash and previous consensus hash values point to this block.
+    /// * There is no coinbase
+    /// * There are no other TenureChange transactions
+    ///
+    /// Returns Some(true) if the above are true
+    /// Returns Some(false) if at least one of the above is false
+    /// Returns None if this block is not a tenure-extend block
```
kantai, line 620: These `is_wellformed` functions seem like pretty un-ergonomic return types. `Some(false)` indicates a block is invalid -- which should almost always manifest in an InvalidBlock. It seems more logical to have something like:

```
Result<bool, ChainstateError>
```

Where the `Ok(true)` indicates that the block is an extend block, `Ok(false)` indicates that it is not, and `Err` is an invalid block.

stackslib/src/chainstate/stacks/mod.rs
```diff
@@ -659,22 +675,73 @@ pub enum TenureChangeError {
...
+    /// block was chosen.  It may be the case that this miner's tenure gets _extended_ across
+    /// subsequent sortitions; if this happens, then this `consensus_hash` value _remains the same_
+    /// as the sortition in which the winning block-commit was mined.
+    pub tenure_consensus_hash: ConsensusHash,
+    /// Consensus hash of the previous tenure.  Corresponds to the sortition of the previous
+    /// winning block-commit.
+    pub prev_tenure_consensus_hash: ConsensusHash,
+    /// Current consensus hash on the underlying burnchain.  Corresponds to the last-seen
+    /// sortition.
+    pub sortition_consensus_hash: ConsensusHash,
```
kantai, line 709: Can this be called `burn_view` (or something like that) instead?

`sortition` implies that a sortition  occurred(i.e., cryptographic random-weighted leader selection), but unless I'm confused, this field is used specifically to deal with the case that sortition _didn't_ occur: a burn block was processed which didn't select a new miner. This is how the sortition_db uses the term, and I think it would be very confusing to use it differently.

stackslib/src/chainstate/nakamoto/tenure.rs
```diff
@@ -0,0 +1,1107 @@
...
+            }
+            None => Ok(None),
+        }
+    }
+
+    /// Get the highest processed tenure on the canonical sortition history.
+    pub fn get_highest_nakamoto_tenure<SH: SortitionHandle>(
+        conn: &Connection,
+        sort_handle: &mut SH,
+    ) -> Result<Option<NakamotoTenure>, ChainstateError> {
```
kantai, line 517: Would `SortitionDB::get_canonical_block_header()` be useful here?

It seems something like this would be equivalent:

```rust
let stacks_tip = SortitionDB::get_canonical_block_header();
let nakamoto_tenure_at_tip = query("SELECT * FROM nakamoto_tenures WHERE tenure_id_consensus_hash = ?", stacks_tip.consensus_hash);
return nakamoto_tenure_at_tip
```

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -2729,41 +2901,18 @@ impl NakamotoChainState {
...
-                        warn!(
-                            "CORRUPTION: block {}/{} does not correspond to a burn block",
-                            &parent_ch, &parent_block_hash
-                        );
-                        (BurnchainHeaderHash([0; 32]), 0, 0)
-                    }
-                }
-            };
-
+            Self::get_tenure_burn_block_info(burn_dbconn, block.is_first_mined(), &parent_ch)?;
```
kantai, line 2906: The refactored interface of this function is very confusing to me and seems unecessary. `is_first_mined()` is something that only makes sense in the context of `block`, but its passed as a bool to the refactored method, which also returns two unlabeled u64s, and the only thing that this code block is doing is looking up a snapshot and pulling out three fields. Like, we could just make this less verbose:

```rust
            if block.is_first_mined() {
                (BurnchainHeaderHash([0; 32]), 0, 0)
            } else {
                let snapshot = SortitionDB::get_block_snapshot_consensus(burn_dbconn, &parent_ch)?
                    .unwrap_or_else(||
                          warn!("CORRUPTION: {parent_ch} does not correspond to a burn block");
                          ChainstateError::InvalidStacksBlock("No such parent consensus hash")
                    )?;
                (sn.burn_header_hash, sn.block_height, sn.burn_header_timestamp)
            };
```

But I think adding a once-used function (in a separate file) that just shuttles fields out of another function call obfuscates what's happening.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1192,40 +1476,101 @@ impl NakamotoChainState {
...
+
+        let Some(sn) = SortitionDB::get_block_snapshot_consensus(sort_handle.sqlite(), &target_ch)?
+        else {
+            warn!("Unacceptable Nakamoto block -- no sortition for tenure";
+                  "sortition_consensus_hash" => %target_ch
+            );
+            return Ok(None);
+        };
+        Ok(Some(sn.total_burn))
+    }
```
kantai, line 1513: This function seems strange to me -- the expected burns for a block are intrinsic to that block and its view of the burnchain. Querying the "highest known tenure" for that information may be correct as well, but that's only context dependent, and it requires multiple steps to reason about.

For example, if a block is mined in Tenure 1, and has a consensus_hash view corresponding to bitcoin block 2, then the total burn amount for that block should *always and forever* be the burn block total at bitcoin block 2. However, as this function is currently written, it would return the burn block total at bitcoin block 2 when first evaluated, but if someone called this function after subsequent tenure's are processed on the _same block_, they would get a different burn amount.

I understand that this function is only to be called while processing a newly arrived block, but this seems like very avoidable behavior:

```rust
let burn_view_ch = if let Some(tenure_payload) = block.get_tenure_tx_payload() {
      tenure_payload.sortition_consensus_hash
} else {
      // if there's no new tenure for this block, the burn total should be the same as its parent
     let parent = Self::get_block_header(chainstate_conn, &block.parent_block_id)?
          .unwrap_or_else(|| ChainstateError::NoSuchParent)?;
    // I think we should add a `.total_burn()` method to StacksHeaderInfo: both epoch2 and epoch3 headers have that information.
     return Ok(parent.total_burn()) 
};
let burn_view_sn = SortitionDB::get_block_snapshot_consensus(sort_handle.sqlite(), &target_ch)?
          .unwrap_or_else(|| ChainstateError::NoSuchParent)?;
Ok(burn_view_sn.total_burn)
```

kantai (commented):
This looks good to me, I just had a few comments which are mostly superficial. This PR also currently breaks the `stacks-node` target, so that must be fixed before it can be approved.

``` $ cargo check
$ cargo check
...
error: could not compile `stacks-node` (bin "stacks-node") due to 7 previous errors; 2 warnings emitted
```

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: Sure; will use `Result<bool, ()>`

stackslib/src/chainstate/stacks/mod.rs
jcnelson: Sure, we can call it `burn_view`

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: The goal here was to cut down on the size of `append_block()`, so different self-contained bits of functionality within it could be unit-tested.

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: Sure, I can do that

stackslib/src/chainstate/nakamoto/tenure.rs
jcnelson: It would, but then the resulting function would be harder to test.  Can this wait until after this PR?

stacks-common/src/util/secp256k1.rs
```diff
@@ -118,45 +118,6 @@ impl Default for Secp256k1PublicKey {
     }
 }
 
-pub struct SchnorrSignature(pub [u8; 65]);
```
jferrant, line 121: Really nice to see this go XD Didn't like it in there to begin with.

stackslib/src/chainstate/nakamoto/tenure.rs
kantai: I'd strongly prefer that this happen in this PR. Otherwise, this method contains what is essentially a parallel implementation of "get the canonical stacks tip", and we should prevent duplication of code not only for its own sake, but because parallel implementations not only confuse the codebase (what's the canonical way to obtain the chain tip?), but also introduce the possibility of difference between the implementations, which would be a problem.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: I think that this particular case is too aggressive of a separation. All this is doing is invoking a single, already-existent function, and pulling three fields out of the result.

jcnelson: I put it back

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -486,6 +973,691 @@ fn test_simple_nakamoto_coordinator_10_tenures_10_blocks() {
...
+        let matured_reward = (*matured_reward_opt).clone().unwrap();
+        debug!("{}: {:?}", i, &matured_reward);
+
+        if i < 10 {
+            assert_eq!(matured_reward.parent_miner.coinbase, 3600_000_000);
+        } else {
+            assert_eq!(matured_reward.parent_miner.coinbase, 1000_000_000);
+        }
+
+        if i < 11 {
```
jferrant, line 1082: Super nit especially since in a test, ut this would be clearer to me with a match on a std::cmp
Like so 
```
    match i.cmp(&11) {
        Ordering::Less => {...},
        Ordering::Greater => {...},
        Ordering::Equal => {...},
    }
    ```

File: stackslib/src/chainstate/nakamoto/coordinator/tests.rs [hunk 26]
```diff
@@ -486,6 +973,691 @@ fn test_simple_nakamoto_coordinator_10_tenures_10_blocks() {
...
+        assert_eq!(matured_reward.latest_miners.len(), 1);
+
+        let miner_reward = &matured_reward.latest_miners[0];
+
+        if i < 9 {
+            assert_eq!(miner_reward.coinbase, 3600_000_000);
+        } else {
+            assert_eq!(miner_reward.coinbase, 1000_000_000);
+        }
+        if i < 10 {
```
jferrant, line 1116: Similar to above comment.

stackslib/src/chainstate/nakamoto/tenure.rs
jcnelson: Okay, this is done.

kantai (approved):
LGTM, thanks for the changes!

PR #4119, opened 2023-12-04
chore: load clar2wasm from github instead of local

hugocaillard:
To be merged on clarity-wasm-next

Link clar2wasm from github instead of local path


obycode (approved):
👍 Thanks!

PR #4124, opened 2023-12-04
Merge master into develop

wileyj:
Merge master into develop using intermediary branch to resolve conflicts


kantai (approved):

stacks-common/src/util/log.rs
```diff
@@ -238,7 +238,8 @@ fn make_logger() -> Logger {
         let plain = slog_term::PlainSyncDecorator::new(slog_term::TestStdoutWriter);
         let isatty = isatty(Stream::Stdout);
         let drain = TermFormat::new(plain, false, debug, isatty);
-        Logger::root(drain.ignore_res(), o!())
+        let logger = Logger::root(drain.ignore_res(), o!());
+        logger
```
obycode, line 242: I think for this one, you can take the other version. It probably avoids a clippy warning.

obycode (commented):
I think there's just one place I would choose the other option.

jcnelson (approved):

stacks-common/src/util/log.rs
wileyj: reverted

wileyj: on second thought, we may want to keep this and adjust clippy - i believe this is part of a hotfix: https://github.com/stacks-network/stacks-core/commit/34703d7b63135987c296014bff7377b4115ff57f

obycode: The change from `fuse` to `ignore_res` is the important part for the hotfix. The simplification from defining a variable and then returning it is what clippy complains about.

PR #4126, opened 2023-12-05
Make `set-aggregate-public-key` private

### LINKED ISSUE #4101: Nakamoto: make set-aggregate-public-key private in pox-4
Inside pox-4, there is a public function "set-aggregate-public-key". You don't need to remove this per se; but it should not be publicly reachable. As signers vote for their aggregate public key for a given reward cycle, after the 70% threshold has been reached, then this function can be called internally. However, it can also be made private prior to this signer change and the node call it at the start of every pox-4 reward cycle instead. This MUST happen in Neon because this function should not be publicly reachable.

xoloki:
### Description
Currently, the `set-aggregate-public-key` function in `pox-4` is public, but it needs to be private.  This will require that the key is set when a new reward cycle begins.

In the future this will only happen after `DKG` completes, but for now just set the key to that of the previous cycle.

### Applicable issues

- fixes #4101 

### Checklist

- [ ] Changelog is updated

### DIFF
--- a/stackslib/src/chainstate/nakamoto/mod.rs
+++ b/stackslib/src/chainstate/nakamoto/mod.rs
@@ -18,10 +18,11 @@ use std::collections::HashSet;
 use std::ops::DerefMut;
 
 use clarity::vm::ast::ASTRules;
-use clarity::vm::costs::ExecutionCost;
+use clarity::vm::costs::{ExecutionCost, LimitedCostTracker};
 use clarity::vm::database::BurnStateDB;
 use clarity::vm::events::StacksTransactionEvent;
 use clarity::vm::types::StacksAddressExtensions;
+use clarity::vm::{ClarityVersion, SymbolicExpression, Value};
 use lazy_static::{__Deref, lazy_static};
 use rusqlite::types::{FromSql, FromSqlError};
 use rusqlite::{params, Connection, OptionalExtension, ToSql, NO_PARAMS};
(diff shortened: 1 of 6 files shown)


stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -2514,6 +2519,78 @@ impl NakamotoChainState {
...
+                        StacksAddress::burn_address(false).into(),
+                        None,
+                        LimitedCostTracker::Free,
+                        |vm_env| {
+                            vm_env.execute_contract_allow_private(
+                                &boot_code_id(POX_4_NAME, false),
+                                "set-aggregate-public-key",
+                                &vec![
+                                    SymbolicExpression::atom_value(Value::UInt(
+                                        my_reward_cycle as u128,
```
jferrant, line 2582: Might want to change this to u128::from(my_reward_cycle)

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1332,7 +1332,7 @@
 ;; Set the aggregate public key to the provided value
 ;; TODO: https://github.com/stacks-network/stacks-core/issues/4101
```
jferrant, line 1333: Don't forget to remove the TODO if its done!

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -2514,6 +2519,78 @@ impl NakamotoChainState {
             tenure_height,
         )?;
 
+        if !block.is_first_mined() {
```
jcnelson, line 2522: What you have here is fine, but can you put it into its own method?  It'll make testing it easier.  Thanks! :pray:

kantai: Yep -- I think this code block should be moved into its own function.

stackslib/src/chainstate/stacks/boot/pox-4.clar
xoloki: Fixed

stackslib/src/chainstate/nakamoto/mod.rs
xoloki: Fixed

testnet/stacks-node/src/mockamoto/tests.rs
```diff
@@ -129,7 +129,7 @@ fn observe_100_blocks() {
 }
 
 #[test]
```
wileyj, line 131:
```suggestion
#[test]
#[ignore]
```

jferrant:
Replaced by https://github.com/stacks-network/stacks-core/pull/4165#pullrequestreview-1780071969

Issue #3929, opened 2023-09-15
Nakamoto[Block Proposal]: Signer binary submits block proposal to stacks-node for validation


kantai:
To validate a BlockProposal from the current tenure, the signer binary needs to:

1. Check that the block proposal was produced by the active miner.
2. Check that the block proposal's parent was the last block signed by the signer.
3. Submit the proposal to a `stacks-node` to validate the block for correctness.

Step 3 is handled via the block proposal endpoint (implemented in #3889)

saralab:
Related to https://github.com/stacks-network/sbtc/issues/367

Issue #3928, opened 2023-09-15
Nokamoto[Block Proposal]: Signer binary handles BlockProposal event stored to StackerDB


kantai:
To support the argon milestone, the signer binary must receive an event from the stacks-node when the current tenure's miner submits a block for the signers to approve.

The signer binary will need to do some processing on this block proposal, but the first step is to handle the event at all.

This requires registering for the correct StackerDB event, fetching the stored data, and deserializing it into a BlockProposal.

saralab:
Related to **https://github.com/stacks-network/sbtc/issues/368**

saralab:
Prereq:  Retrieve the winning miner from .miners contract

kantai:
This requires figuring out who the current miner is -- this can be queried from the `.miners` contract using the RPC endpoint. This also is relayed in burnchain block events. I wouldn't say this is a pre-req, it's just a component of implementing this.

Issue #3930, opened 2023-09-15
Nokamoto[Block Proposal]: Signer binary issues signature and broadcasts the block


kantai:
After the signer binary has validated a block proposal, it needs to sign the block.

In the Neon milestone, this should be a simple "single signature".

Once the signature has been applied to the block, the signer binary passes the now-finished block back to the `stacks-node` for broadcasting to the rest of the network.

**### Copied from the sBTC breakdown:** 

Description
Update signers to perform signing rounds Nakamoto block stacker signature hash (the entire block header including miner signature but NOT including stacker signature)

Context
Signers are responsible for validating/signing Nakamoto blocks and returning them to the miner. This ticket enables us to perform the actual signing side of things.

Expected Approach
The message bytes being sent via the signing rounds need to include probably some message type to indicate what type of thing is being signed so that signers can deserialize the contents into a nakamoto block vs btc transaction to inspect. Signers should be able to decide whether they want to sign or not so being able to inspect the contents is required as part of this ticket.

Acceptance Criteria
 PR adds the feature to sign blocks
 PR adds an integration test for signing a Nakamoto block that includes both a failure case (no quorum), no, and yes result

saralab:
Related to https://github.com/stacks-network/sbtc/issues/380

PR #4130, opened 2023-12-06
Feat/stacks signer block proposal

jferrant:
### Description

This is a rather large PR...please forgive.

major changes include:
- Complete reorg of the stacker db and stacks clients
- Complete reorg of the signer integration tests for easier adding of tests
- Changing libsigner to take SignerEvents rather than StackerDBEvents as we now need to subscribe to blocks
- Handling the new block proposal events within Stacks signer
- Added a test to handle a validated block (against the RPC endpoint)
- Enabled stackerdb_dkg_sign test as well since I made substantial changes to the test layout to ensure it still works

### Applicable issues

- Closes https://github.com/stacks-network/stacks-core/issues/3928
- Closes https://github.com/stacks-network/stacks-core/issues/3929

Major items left to do that are NOT included in this PR
- State needs to be added to the signers to allow them to filter out already seen blocks requests/block validation responses/block sign requests. See https://github.com/stacks-network/stacks-core/issues/4242
- Block is not properly deserialized during signing rounds to enable verification of a sign requests contents. See https://github.com/stacks-network/stacks-core/issues/3930
- State needs to be added to verify a block contains the expected transactions https://github.com/stacks-network/stacks-core/issues/4243

### Checklist

- [ ] Changelog is updated


stacks-signer/src/client/stackerdb.rs
```diff
@@ -0,0 +1,312 @@
...
+        message: StackerDBMessage,
+    ) -> Result<StackerDBChunkAckData, ClientError> {
+        let message_bytes = message.serialize_to_vec();
+        let slot_id = message.slot_id(id);
+
+        loop {
+            let slot_version = *self.slot_versions.entry(slot_id).or_insert(0) + 1;
+            let mut chunk = StackerDBChunkData::new(slot_id, slot_version, message_bytes.clone());
+            chunk.sign(&self.stacks_private_key)?;
+            debug!("Sending a chunk to stackerdb!\n{:?}", chunk.clone());
```
jcnelson, line 180: Nit: You probably don't need the clone here; you could just pass `&chunk`.

stacks-signer/src/client/stacks_client.rs
```diff
@@ -120,45 +51,26 @@ impl From<&Config> for StacksClient {
...
-                // See: https://github.com/stacks-network/stacks-blockchain/issues/3917
-                if reason == "Data for this slot and version already exist" {
-                    warn!("Failed to send message to stackerdb due to wrong version number {}. Incrementing and retrying...", slot_version);
-                } else {
-                    warn!("Failed to send message to stackerdb: {}", reason);
-                    return Err(ClientError::PutChunkRejected(reason));
-                }
-            }
+    /// Retrieve the current miner public key
+    pub fn get_miner_public_key(&self) -> Result<StacksPublicKey, ClientError> {
```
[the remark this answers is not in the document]
jferrant: I am not entirely sure. I only really added this per the comment https://github.com/stacks-network/stacks-core/issues/3928#issuecomment-1812947134

I think the main thing is so long as I can verify the signature of the block is valid/belongs to the appropriate miner, it should be fine.

stacks-signer/src/client/stackerdb.rs
jferrant: Oh yes definitely here :)

stacks-signer/src/client/stackerdb.rs
```diff
@@ -0,0 +1,312 @@
...
+            let send_request = || {
+                self.signers_stackerdb_session
+                    .put_chunk(chunk.clone())
```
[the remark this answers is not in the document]
jferrant: I updated put_chunk to take a reference cause it didn't seem like it actually required ownership so now I can get rid of the clone :D

stacks-signer/src/client/mod.rs
```diff
@@ -0,0 +1,85 @@
+/// The stacker db module for communicating with the stackerdb contract
```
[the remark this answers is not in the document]
jferrant: Done :D I also added it to lib.rs

stacks-signer/src/client/stacks_client.rs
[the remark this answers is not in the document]
jferrant: Signature validation happens via the endpoint call. No need for this.

Cargo.toml
```diff
@@ -14,7 +14,7 @@ members = [
 
 # Dependencies we want to keep the same between workspace members
 [workspace.dependencies]  
-wsts = "6.0"
+wsts = { path = "../wsts" }
```
jferrant, line 17: TODO: remove this change. Just temp to access Packet changes I made..

libsigner/src/events.rs
```diff
@@ -20,23 +20,60 @@ use std::sync::atomic::{AtomicBool, Ordering};
...
 
-/// Event structure for newly-arrived StackerDB data
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub struct StackerDBChunksEvent {
-    /// The contract ID for the StackerDB instance
-    pub contract_id: QualifiedContractIdentifier,
-    /// The chunk data for newly-modified slots
-    pub modified_slots: Vec<StackerDBChunkData>,
+#[repr(u8)]
+enum EventPrefix {
```
jcnelson, line 43: Do you want to use `define_u8_enum!` here?

stacks-signer/src/lib.rs
```diff
@@ -3,13 +3,30 @@
 # stacks-signer: a libary for creating a Stacks compliant signer. A default implementation binary is also provided.
 Usage documentation can be found in the [README](https://github.com/Trust-Machines/core-eng/stacks-signer-api/README.md).
 */
+
+// Copyright (C) 2013-2020 Blockstack PBC, a public benefit corporation
+// Copyright (C) 2020-2023 Stacks Open Internet Foundation
```
jcnelson, line 8: 2020-2024

stacks-signer/src/runloop.rs
```diff
@@ -164,45 +178,159 @@ impl<C: Coordinator> RunLoop<C> {
...
+                        merkle_root: None,
+                    });
+                }
+            }
+            BlockValidateResponse::Reject(block_validate_reject) => {
+                warn!(
+                    "Received a block proposal that was rejected by the stacks node: {:?}",
+                    block_validate_reject
+                );
+                // TODO: submit a rejection response to the .signers contract for miners
```
jcnelson, line 202: Will this be done in this PR?

stacks-signer/src/runloop.rs
```diff
@@ -164,45 +178,159 @@ impl<C: Coordinator> RunLoop<C> {
...
                     Some(packet)
                 } else {
                     None
                 }
             })
             .collect();
+
         // First process all messages as a signer
-        let mut outbound_messages = self
+        // TODO: deserialize the packet into a block and verify its contents
```
jcnelson, line 253: Presumably, there's also a dependency between validating the block and signing it, right?  What happens if your signer gets a "sign block" request before it has even obtained the block to validate?  Where is this captured in the run loop?

stacks-signer/src/lib.rs
jferrant: Oh right. XD

stacks-signer/src/runloop.rs
jferrant: Oh sorry I had actually done this...just hadn't pushed the code LOL

stacks-signer/src/runloop.rs
jferrant: I will change it around. i.e. as soon as a block arrives from the miner. Trigger a signing round. and then have every signer receiving the sign request pass it to the endpoint to be validated before continuing the signing round. however, this will be heavily dependent on Joey's changes to the sign logic to enable me to deserialize it. I would leave it to a sep PR.

stacks-signer/src/lib.rs
jferrant: Done

stacks-signer/src/runloop.rs
jferrant: Done.

libsigner/src/events.rs
jferrant: Actually this was dead code from a prior version where I serialized things differently. Deleted!

stacks-signer/src/runloop.rs
```diff
@@ -164,45 +176,211 @@ impl<C: Coordinator> RunLoop<C> {
         }
     }
 
-    /// Process the event as both a signer and a coordinator
-    fn process_event(
+    /// Handle the block validate response returned from our prior calls to submit a block for validation
+    fn handle_block_validate_response(&mut self, block_validate_response: BlockValidateResponse) {
+        match block_validate_response {
```
kantai, line 181: Should the signer binary track block validation responses directly? This seems like state that will be important for the signer binary to have: that way if a sign request comes in, the signer knows whether or not it has already validated (or rejected) the requested proposal.

jferrant: We do have a ticket in plcae to add state to the signers. I do believe we will need it for not only that but to know what transactions we expect to see in a block. I think this should remain a second ticket though considering the size of this change/because I believe we could work on it asyncronously at the same time as the serialization of the block ticket that Joey is assigned to.

jferrant: I have made two tickets that track the need for local state for both block requests/responses/sign requests and the need to track transactions to verify a block against

https://github.com/stacks-network/stacks-core/issues/4242
https://github.com/stacks-network/stacks-core/issues/4243

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -182,16 +171,28 @@ impl NakamotoBlockProposal {
         let mainnet = self.chain_id == CHAIN_ID_MAINNET;
         if self.chain_id != chainstate.chain_id || mainnet != chainstate.mainnet {
             return Err(BlockValidateReject {
+                block: self.block.clone(),
```
kantai, line 174: I think it would simplify this function a bit to just have a new type:

```rust
#[derive(Debug, Clone, PartialEq)]
pub struct BlockValidateRejectReason {
    pub reason: String,
    pub reason_code: ValidateRejectCode,
}
```

and change the return type of `validate` to `Result<BlockValidateOk, BlockValidateReject>`

with the previous `From<ChainstateError>` implementation.

Then, the caller of `validate()` could do:

```rust
proposal.validate(..).map_err(|reason| BlockValidateReject { block: proposal.block.clone(), reason_code: reason.reason_code, reason: reason.reason })
```

That way, this `validate()` function doesn't have to repeatedly perform that step.

stacks-signer/src/runloop.rs
kantai: Sounds good to me

libsigner/src/events.rs
```diff
@@ -209,27 +217,45 @@ impl EventReceiver for StackerDBEventReceiver {
...
             }
-            if request.url() != "/stackerdb_chunks" {
-                let url = request.url().to_string();
+            if request.url() == "/stackerdb_chunks" {
+                debug!("Got stackerdb_chunks event");
+                let mut body = String::new();
+                request
+                    .as_reader()
+                    .read_to_string(&mut body)
+                    .expect("failed to read body");
```
kantai, line 227: I think you should respond `200` as soon as you've read body into the string. The `stacks-node` will stall progress until it gets a 200 response, and if you exit this function early due to error handling, the `stacks-node` may end up stalling entirely.

libsigner/src/events.rs
```diff
@@ -209,27 +217,45 @@ impl EventReceiver for StackerDBEventReceiver {
...
+                }
 
                 request
                     .respond(HttpResponse::empty(200u16))
                     .expect("response failed");
-                Err(EventError::UnrecognizedEvent(url))
-            } else {
+
+                Ok(SignerEvent::StackerDB(event))
+            } else if request.url() == "/proposal_response" {
```
kantai, line 250: Same thing here -- as soon as the body is read into the string, return 200.

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -200,142 +335,183 @@ fn test_stackerdb_dkg() {
...
+        &mut signer_test.running_nodes.btc_regtest_controller,
+        60,
+        &signer_test.running_nodes.coord_channel,
+        &commits_submitted,
+    )
+    .unwrap();
+
+    info!("------------------------- Test Block Processed -------------------------");
+    //Wait for the block to show up in the test observer
+    let validate_responses = test_observer::get_proposal_responses();
```
kantai, line 484: Should this loop and timeout, e.g.,

```rust
let t_start = Instant::now();
while test_observer::get_proposal_responses().is_empty() {
     assert!(t_start.elapsed() < TIMEOUT, "Timed out while waiting for block proposal event");
     thread::sleep(Duration::from_secs(1));
}
```

Otherwise, the test might have a race condition.

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -200,142 +335,183 @@ fn test_stackerdb_dkg() {
...
-    );
+    let elapsed = now.elapsed();
+    info!("DKG and Sign Time Elapsed: {:.2?}", elapsed);
+    signer_test.shutdown();
+}
 
-    res_receivers.push(coordinator_res_recv);
+#[test]
+#[ignore]
+fn stackerdb_block_proposal() {
```
kantai, line 404: Can you add a rustdoc comment to this test method that just describes at a high level, what behavior this function is testing and what it does to do so?

kantai (commented):
This looks great to me, just had a few comments

libsigner/src/events.rs
jferrant: Done!

libsigner/src/events.rs
jferrant: Done!

testnet/stacks-node/src/tests/signer.rs
jferrant: Done! :)

testnet/stacks-node/src/tests/signer.rs
jferrant: Done!

stackslib/src/net/api/postblock_proposal.rs
jferrant: Done!

kantai (approved):
LGTM!

stacks-signer/src/client/stackerdb.rs
```diff
@@ -0,0 +1,197 @@
+use blockstack_lib::burnchains::Txid;
```
jcnelson, line 1: Copyright comment block please

stacks-signer/src/client/stackerdb.rs
```diff
@@ -0,0 +1,197 @@
...
+use blockstack_lib::chainstate::nakamoto::NakamotoBlock;
+use blockstack_lib::net::api::postblock_proposal::{BlockValidateReject, ValidateRejectCode};
```
jcnelson, line 3: Isn't this `stackslib::` now, instead of `blockstack_lib`?

stacks-signer/src/client/stackerdb.rs
```diff
@@ -0,0 +1,197 @@
...
+
+/// This enum is used to supply a `reason_code` for block rejections
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[repr(u8)]
+pub enum RejectCode {
+    /// RPC endpoint Validation failed
+    ValidationFailed(ValidateRejectCode),
+    /// Missing expected transactions
+    MissingTransactions(Vec<Txid>),
+    // No Consensus Reached
```
jcnelson, line 83: Commented-out code

stacks-signer/src/runloop.rs
jcnelson: Can you add a ticket for this before we merge this?

stacks-signer/src/runloop.rs
```diff
@@ -164,45 +176,211 @@ impl<C: Coordinator> RunLoop<C> {
...
+                    warn!("Failed to submit block for validation: {:?}", e);
+                });
+        }
+    }
+
+    /// Helper function to extract block proposals from signature results and braodcast them to the stackerdb slot
+    fn send_block_response_messages(&mut self, _operation_results: &[OperationResult]) {
+        //TODO: Deserialize the signature result and broadcast an appropriate Reject or Approval message to stackerdb
+        // https://github.com/stacks-network/stacks-core/issues/3930
+        // for result in operation_results {
```
jcnelson, line 292: Please don't commit commented-out code

stackslib/src/net/api/poststackerdbchunk.rs
```diff
@@ -54,6 +54,15 @@ use crate::net::{
 };
 use crate::util_lib::db::{DBConn, Error as DBError};
 
+/// Event structure for newly-arrived StackerDB data
```
jcnelson, line 57: Can you instead put this into `stackslib/src/chainstate/stacks/events.rs`?

jcnelson (approved):
Just a couple comments. Thanks!

stacks-signer/src/client/stackerdb.rs
jferrant: I don't find it that way. as far as I am aware its still blockstack_lib. Can find no instance of stackslib::* throughout the code base.

stackslib/src/net/api/poststackerdbchunk.rs
jferrant: Done :)

stacks-signer/src/runloop.rs
jferrant: Done.

stacks-signer/src/client/stackerdb.rs
jferrant: Done :)

stacks-signer/src/client/stackerdb.rs
jferrant: Done. Added to ever file in stacks-signer now.

PR #4132, opened 2023-12-07
Fix wrong return value in `wasm_to_clarity_value`

Acaccia:
### Description

We return the `index` instead of the increment in the Tuple's match branch in `wasm_to_clarity_value` .

Added tests will be in the [`clarity-wasm` repo](https://github.com/stacks-network/clarity-wasm).


obycode (approved):
Thanks @Acaccia! That's my bad.

PR #4133, opened 2023-12-07
Nakamoto-Neon: Add nakamoto_node

kantai:
This PR implements a first version of a "nakamoto neon" node and operation mode.

As part of this implementation, this PR does several things:

* Refactor some of the reused structs from `neon_node` (most notably `globals`).
* Fix a logic-bug in `nakamoto::coordinator`: the first prepare phase information will be a Epoch2x block, so the reward set calculation has to handle that.
* Add `nakamoto_node` module based on `neon_node`. It uses a new nakamoto-specifix `RelayerDirective` enum.
* Add `boot_nakamoto` module which wraps and orchestrates the 2.x <-> 3.x node handoff. It wraps a `neon` or `nakamoto` runloop, and when the Epoch-3.0 transition is reached, stops the `neon` loop and starts the `nakamoto` loop. 
* Add "simple" integration test for `nakamoto_node`: this integration test actually spins up a neon node to mine through 2.0-2.5, stack enough to active pox-4, then shuts down the neon node and spins up the nakamoto node. This is effectively a first test of node/miner operation during the transition from 2.x to 3.x

### Things that are still todo for this PR:
14fc7fee7 - Update: these are now all done.

**Tests**
1. More advanced test that pushes more nakamoto blocks (updated in 14fc7fee7)
2. Test that accepts transactions and confirms that they get mined (updated in 14fc7fee7)

**Implementation**

Block commits need to support RBF _and_ check if the tip tenure ID changed. The reason for this is that miners should normally expect to RBF their block commit at least once:

1. Before the first block is announced for Tenure 2, the commit for Tenure 3 will point at Tenure 1.
2. Once the first block is announced for Tenure 2, the commit for Tenure 3 should be RBFed to Tenure 2.

(updated in 14fc7fee7)

**Upstream PRs**

There are multiple PRs that will impact this one:

1. #4114 -- this shouldn't be a huge impact, but will require updating the self-signing mode in the `nakamoto_node` so that it generates a `TenureChange` tx correctly.
2. #4126 and #4113 -- the aggregate public key setting should eventually impact the `nakamoto_node`. Right now, the `nakamoto_node` avoids the issue by explicitly passing its self-signing aggregate key to the block acceptance routine, but there probably needs to be a new issue for loading that aggregate key automatically (because such circumvention shouldn't be possible).


testnet/stacks-node/src/globals.rs
```diff
@@ -0,0 +1,289 @@
...
+        *key_state = LeaderKeyRegistrationState::Pending(target_block_height, txid);
+    }
+
+    /// Advance the leader key registration state to active, given the VRF key registration ops
+    /// we've discovered in a given snapshot.
+    /// The runloop thread calls this whenever it processes a sortition.
+    pub fn try_activate_leader_key_registration(
+        &self,
+        burn_block_height: u64,
+        key_registers: Vec<LeaderKeyRegisterOp>,
```
jferrant, line 248: Does this need to be an owned vector rather than a slice?

testnet/stacks-node/src/run_loop/boot_nakamoto.rs
```diff
@@ -0,0 +1,220 @@
...
+            config.burnchain.get_bitcoin_network().1,
+            config.burnchain.epochs.as_ref(),
+        );
+        let epoch_3 = &epochs[StacksEpoch::find_epoch_by_id(&epochs, StacksEpochId::Epoch30)
+            .ok_or("No Epoch-3.0 defined")?];
+
+        Ok(u64::from(burn_height) >= epoch_3.start_height - 1)
+    }
+
+    fn get_burn_height(config: &Config) -> Result<u32, String> {
```
jferrant, line 197: Does this have to return a result? seems like it always returns Ok?

testnet/stacks-node/src/run_loop/nakamoto.rs
```diff
@@ -0,0 +1,1023 @@
...
+
+        // prepare to fetch the first reward cycle!
+        let mut target_burnchain_block_height = cmp::min(
+            burnchain_config.reward_cycle_to_block_height(
+                burnchain_config
+                    .block_height_to_reward_cycle(burnchain_height)
+                    .expect("BUG: block height is not in a reward cycle")
+                    + 1,
+            ),
+            burnchain.get_headers_height() - 1,
```
jferrant, line 821: I assume this can never underflow? I generally prefer to use saturating_sub just to be absolutely certain.

testnet/stacks-node/src/run_loop/boot_nakamoto.rs
kantai: Yeah, it's just to make it easier to return an error in the future if we want to error for any reason in this function (rather than defaulting to 0). One of the lessons for me from a lot of the interfaces in the `testnet/stacks-node` is that most functions should have been returning `Result`s.

stackslib/src/chainstate/nakamoto/miner.rs
```diff
@@ -76,6 +76,8 @@ use crate::net::Error as net_error;
 
 /// New tenure information
 pub struct NakamotoTenureStart {
+    /// tenure change transaction
+    pub tenure_change_tx: StacksTransaction,
```
jcnelson, line 80: You may want to take a look at what I have in the tenure validation PR

kantai: Yes, I'm familiar with the changes, and I expect to have to resolve conflicts on merge/rebase. It should be pretty easy to resolve. In a subsequent PR, I'd prefer to refactor what's happening with `chainstate::nakamoto::miner` -- I believe that the miner's instantiation routines (i.e., the `from_*`s) should be substantially refactored. Most of that tenure information is not actually important to the BlockBuilder struct-- really, it's just about supplying initial transactions to the builder (which the builder doesn't assemble, they're passed in).

testnet/stacks-node/src/globals.rs
```diff
@@ -0,0 +1,289 @@
...
+/// neon nodes use different relayer directives.
+pub struct Globals<T> {
+    /// Last sortition processed
+    last_sortition: Arc<Mutex<Option<BlockSnapshot>>>,
+    /// Status of the miner
+    miner_status: Arc<Mutex<MinerStatus>>,
+    /// Communication link to the coordinator thread
+    pub(crate) coord_comms: CoordinatorChannels,
+    /// Unconfirmed transactions (shared between the relayer and p2p threads)
+    unconfirmed_txs: Arc<Mutex<UnconfirmedTxMap>>,
```
jcnelson, line 49: This goes away in Nakamoto

testnet/stacks-node/src/globals.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// Writer endpoint to the relayer thread
+    pub relay_send: SyncSender<T>,
+    /// Cointer state in the main thread
+    pub counters: Counters,
+    /// Connection to the PoX sync watchdog
+    pub sync_comms: PoxSyncWatchdogComms,
```
jcnelson, line 55: This can be dropped for Nakamoto as well -- the system _must_ wait for the missing PoX anchor block in all cases.

testnet/stacks-node/src/nakamoto_node.rs
```diff
@@ -0,0 +1,305 @@
...
+        let _ = monitoring::set_burnchain_signer(BurnchainSigner(miner_addr_str)).map_err(|e| {
+            warn!("Failed to set global burnchain signer: {:?}", &e);
+            e
+        });
+    }
+
+    pub fn spawn(
+        runloop: &RunLoop,
+        globals: Globals,
+        // relay receiver endpoint for the p2p thread, so the relayer can feed it data to push
```
jcnelson, line 121: Nit: Argument comments go in method blocks

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -0,0 +1,621 @@
...
+        globals.block_miner();
+        prior_miner
+            .join()
+            .expect("FATAL: IO failure joining prior mining thread");
+        globals.unblock_miner();
+    }
+
+    pub fn run_miner(mut self, prior_miner: Option<JoinHandle<()>>) {
+        // when starting a new tenure, block the mining thread if its currently running.
+        // the new mining thread will join it (so that the new mining thread stalls, not the relayer)
```
jcnelson, line 130: :+1:

testnet/stacks-node/src/globals.rs
kantai: This struct is shared with the `neon_node`. I could make this field generic as well (and use a void type for it in nakamoto), but that didn't seem like a clear improvement to me.

testnet/stacks-node/src/globals.rs
kantai: Same as above-- this struct is shared with the `neon_node`. This field eventually will need to be generic. I think if we don't need the `pox_watchdog`, this would just be a struct monitoring the `mine_after` setting (where miners can set when they want to begin issuing block commits).

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -0,0 +1,621 @@
...
+                    "Miner should be starting a new tenure, but failed to load parent tenure info"
+                );
+                return None;
+            }
+        }
+
+        // create our coinbase if this is the first block we've mined this tenure
+        let tenure_start_info = if let Some(ref par_tenure_info) = parent_block_info.parent_tenure {
+            let parent_block_id = parent_block_info.stacks_parent_header.index_block_hash();
+            let current_miner_nonce = parent_block_info.coinbase_nonce;
```
jcnelson, line 403: Would it make more sense / be more robust to just query the miner's account here to get the nonce?

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -0,0 +1,621 @@
...
+            let coinbase_tx = self.generate_coinbase_tx(
+                current_miner_nonce + 1,
+                target_epoch_id,
+                vrf_proof.clone(),
+            );
+            Some(NakamotoTenureStart {
+                coinbase_tx,
+                // TODO (refactor): the nakamoto block builder doesn't use this VRF proof,
+                //  it has to be included in the coinbase tx, which is an arg to the builder.
+                //  we should probably just remove this from the nakamoto block builder.
```
jcnelson, line 419: Yup; this is removed by #4114

testnet/stacks-node/src/nakamoto_node/miner.rs
kantai: That is what is queried in `MinerThread::load_block_parent_info` to get that field.

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -0,0 +1,621 @@
...
+            block.txs.len(),
+        );
+
+        // last chance -- confirm that the stacks tip is unchanged (since it could have taken long
+        // enough to build this block that another block could have arrived), and confirm that all
+        // Stacks blocks with heights higher than the canoincal tip are processed.
+        let cur_burn_chain_tip = SortitionDB::get_canonical_burn_chain_tip(burn_db.conn())
+            .expect("FATAL: failed to query sortition DB for canonical burn chain tip");
+
+        if cur_burn_chain_tip.consensus_hash != block.header.consensus_hash {
```
jcnelson, line 482: You wouldn't want to compare to the block's consensus hash, because the sortition may have been empty and the current miner's tenure may still be ongoing.  Would it make more sense to compare the sortition tips before and after building the block?

testnet/stacks-node/src/nakamoto_node/miner.rs
jcnelson: Oh I see, you're doing this on the call to `load_block_parent_info()`, right?

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -0,0 +1,621 @@
...
+                //  it has to be included in the coinbase tx, which is an arg to the builder.
+                //  we should probably just remove this from the nakamoto block builder.
+                vrf_proof: vrf_proof.clone(),
+                tenure_change_tx,
+            })
+        } else {
+            None
+        };
+
+        parent_block_info.stacks_parent_header.microblock_tail = None;
```
jcnelson, line 427: Might also want to set the seq to 0

testnet/stacks-node/src/nakamoto_node/relayer.rs
```diff
@@ -0,0 +1,907 @@
+use core::fmt;
```
jcnelson, line 1: Needs to be beneath comment block

testnet/stacks-node/src/nakamoto_node/miner.rs
kantai: I'm not sure yet how the nakamoto MinerThread should handle the sortition changing. My current plan is to treat it like a tenure change and do a thread stop and start. That way, the `MinerThread` really should stop mining if it detects a consensus hash change (which it will sort of need to do in any event: it will have to get a new tenure change tx).

testnet/stacks-node/src/nakamoto_node/relayer.rs
```diff
@@ -0,0 +1,907 @@
...
+                &mut self.sortdb,
+                &mut self.chainstate,
+                &mut self.mempool,
+                self.globals.sync_comms.get_ibd(),
+                Some(&self.globals.coord_comms),
+                Some(&self.event_dispatcher),
+            )
+            .expect("BUG: failure processing network results");
+
+        if net_receipts.num_new_blocks > 0 || net_receipts.num_new_confirmed_microblocks > 0 {
```
jcnelson, line 264: Does this also check Nakamoto blocks?

testnet/stacks-node/src/nakamoto_node/relayer.rs
```diff
@@ -0,0 +1,907 @@
...
+            signal_mining_blocked(self.globals.get_miner_status());
+        }
+
+        let mempool_txs_added = net_receipts.mempool_txs_added.len();
+        if mempool_txs_added > 0 {
+            self.event_dispatcher
+                .process_new_mempool_txs(net_receipts.mempool_txs_added);
+        }
+
+        let num_unconfirmed_microblock_tx_receipts =
```
jcnelson, line 277: No longer necessary in Nakamoto

testnet/stacks-node/src/nakamoto_node/relayer.rs
```diff
@@ -0,0 +1,907 @@
...
+        }
+
+        // Dispatch retrieved attachments, if any.
+        if net_result.has_attachments() {
+            self.event_dispatcher
+                .process_new_attachments(&net_result.attachments);
+        }
+
+        // synchronize unconfirmed tx index to p2p thread
+        self.globals.send_unconfirmed_txs(&self.chainstate);
```
jcnelson, line 298: No longer necessary in Nakamoto

testnet/stacks-node/src/nakamoto_node.rs
kantai: Will add a rustdoc for these arguments.

testnet/stacks-node/src/nakamoto_node/relayer.rs
kantai: This method is unchanged from the neon_node's `process_net_result` method.

testnet/stacks-node/src/run_loop/nakamoto.rs
```diff
@@ -0,0 +1,1023 @@
...
+    globals: Option<Globals>,
+    counters: Counters,
+    coordinator_channels: Option<(CoordinatorReceivers, CoordinatorChannels)>,
+    should_keep_running: Arc<AtomicBool>,
+    event_dispatcher: EventDispatcher,
+    #[allow(dead_code)]
+    pox_watchdog: Option<PoxSyncWatchdog>, // can't be instantiated until .start() is called
+    is_miner: Option<bool>,       // not known until .start() is called
+    burnchain: Option<Burnchain>, // not known until .start() is called
+    pox_watchdog_comms: PoxSyncWatchdogComms,
```
jcnelson, line 77: Not necessary anymore

testnet/stacks-node/src/run_loop/nakamoto.rs
```diff
@@ -0,0 +1,1023 @@
...
+            ),
+            sn,
+        )
+    }
+
+    /// Wake up and drive stacks block processing if there's been a PoX reorg.
+    /// Be careful not to saturate calls to announce new stacks blocks, because that will disable
+    /// mining (which would prevent a miner attempting to fix a hidden PoX anchor block from making
+    /// progress).
+    fn drive_pox_reorg_stacks_block_processing(
```
jcnelson, line 396: This can be vastly simplified now that affirmation maps don't omit anchor blocks.

testnet/stacks-node/src/run_loop/nakamoto.rs
```diff
@@ -0,0 +1,1023 @@
...
+        *last_stacks_pox_reorg_recover_time = get_epoch_time_secs().into();
+    }
+
+    /// Wake up and drive sortition processing if there's been a PoX reorg.
+    /// Be careful not to saturate calls to announce new burn blocks, because that will disable
+    /// mining (which would prevent a miner attempting to fix a hidden PoX anchor block from making
+    /// progress).
+    ///
+    /// only call if no in ibd
+    fn drive_pox_reorg_burn_block_processing(
```
jcnelson, line 514: Same; this can be simplified

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -63,6 +64,10 @@ pub struct Counters {
     pub missed_tenures: RunLoopCounter,
     pub missed_microblock_tenures: RunLoopCounter,
     pub cancelled_commits: RunLoopCounter,
+
+    pub naka_submitted_vrfs: RunLoopCounter,
```
jcnelson, line 68: Nakanaka

zone117x (approved):
PR is working for me to spin up a node to epoch2.5 with a fully functional pox-4. I've been unable to get it to boot into epoch3.0 (trying to replicate your integration test). It's likely just a block height config and pox cycle config issue on my end, but would be really helpful if:
1. Default pox cycle config for neon works to boot epoch3.0 (maybe it already is, but not sure yet)
2. While loading the config, if the given epoch block heights and pox config do not allow epoch3.0 to activate, then exit with error (e.g. invalid cycle boundaries)

testnet/stacks-node/src/nakamoto_node/miner.rs
kantai: Yep, exactly

testnet/stacks-node/src/run_loop/nakamoto.rs
kantai: I removed the chain liveness thread (and its functions) for now.

testnet/stacks-node/src/run_loop/nakamoto.rs
kantai: I removed the chain liveness thread (and its functions) for now.

testnet/stacks-node/src/nakamoto_node/miner.rs
kantai: The sequence number is part of that `StacksMicroblockHeader` struct, which is being set to `None` here, right?

testnet/stacks-node/src/nakamoto_node/relayer.rs
kantai: Removed in fc147df85

testnet/stacks-node/src/nakamoto_node/relayer.rs
kantai: Fixed in fc147df85

testnet/stacks-node/src/run_loop/nakamoto.rs
kantai: Right now, I'm just using it as a placeholder for an `ibd` global, I think as we move towards `Argon`, there will need to be a new struct for this, but I'd prefer to hold off on making a decision on that for now.

testnet/stacks-node/src/nakamoto_node.rs
```diff
@@ -0,0 +1,315 @@
...
+
+use self::peer::PeerThread;
+use self::relayer::{RelayerDirective, RelayerThread};
+
+pub const RELAYER_MAX_BUFFER: usize = 100;
+const VRF_MOCK_MINER_KEY: u64 = 1;
+
+pub const BLOCK_PROCESSOR_STACK_SIZE: usize = 32 * 1024 * 1024; // 32 MB
+
+pub type BlockCommits = HashMap<Txid, ()>;
```
jcnelson, line 51: Why isn't this just a `HashSet`?

kantai: Switched to `HashSet` in 02c645748

jcnelson (approved):
LGTM!  Played around with this last night in the integration test framework

PR #4139, opened 2023-12-08
Fix: RPC endpoints should default to proof=1

kantai:
### Description

HTTP changes in `develop` changed the default behavior of the `proof` query string. In `master`, it defaults to `1`, but in `develop` it defaults to `0`.

I'm open to changing this default, but this is why `tests::integrations::integration_test_get_info` broke, and the change would need to be mentioned specifically in the changelog. I think the right thing to do here is to return to the prior behavior, and if we want to change the default, we can do that in another PR that also updates the tests.

### DIFF
--- a/stackslib/src/net/http/request.rs
+++ b/stackslib/src/net/http/request.rs
@@ -612,7 +612,7 @@ impl HttpRequestContents {
     }
 
     /// Get a query argument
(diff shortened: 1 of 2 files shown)


jcnelson (approved):

wileyj (approved):
confirmed test is now passing.
shipit

PR #4140, opened 2023-12-08
Test: fix httpcore test for defaulting in proof query string

kantai:
### Description

PR #4139 updated the defaulting proof query string behavior, but didn't update the unit test for that behavior in `net::tests`


wileyj (approved):
confirmed to work. shipit

jferrant (approved):

jcnelson (approved):

PR #4144, opened 2023-12-11
change the wording around the signature description

tippenein:
> Traits are defined with a name, and a list functions, defined with a name, a list of argument types, and return type.

I had to read this line a couple times to understand what it was trying to say. It might actually be better to delete it entirely since it's trying to describe the `signature` which is directly above it. Not sure.. Thought I'd open this at least to a discussion. The opening sentence feels a bit redundant too. "`define-trait` is used to define a new trait"


tippenein:
context from discord: https://discord.com/channels/621759717756370964/623217767356694547/1183566565199843378

clarity/src/vm/docs/mod.rs
```diff
@@ -2001,7 +2001,7 @@ const DEFINE_TRAIT_API: DefineAPI = DefineAPI {
 can implement a given trait and then have their contract identifier being passed as a function argument in order to be called
 dynamically with `contract-call?`.
 
-Traits are defined with a name, and a list functions, defined with a name, a list of argument types, and return type.
+Traits are defined with a name, a list of functions, and a return type. The list of functions here is defined with a name and the list of argument types.
```
obycode, line 2004: Hi @tippenein. Thanks for trying to make this more clear! The return type should go with the functions. Would something like this be more clear?

```
Traits are defined with a name, and a list functions, where each function is defined with a name, a list of argument types, and a return type.
```

tippenein: Yeah that's better imo

clarity/src/vm/docs/mod.rs
```diff
@@ -2001,7 +2001,7 @@ const DEFINE_TRAIT_API: DefineAPI = DefineAPI {
 can implement a given trait and then have their contract identifier being passed as a function argument in order to be called
 dynamically with `contract-call?`.
 
-Traits are defined with a name, and a list functions, defined with a name, a list of argument types, and return type.
+Traits are defined with a name, and a list functions, where each function is defined with a name, a list of argument types, and a return type.
```
obycode, line 2004: Thanks! While you're here, I just realized we're missing an "of":
"... and a list **of** functions,..."

obycode (dismissed):
🎉 Thanks for that back and forth!

obycode:
Let's rebase this to merge into master. @tippenein Thanks!

kantai (commented):
LGTM, just rebase to `master` so that it's not pulling in the `develop` changes as well.

obycode (approved):
Thanks!

Issue #4148, opened 2023-12-11
Nakamoto: Mempool Admission must be updated for Nakamoto


kantai:
Mempool admission has to be updated for nakamoto (and this is necessary for the neon milestone).

This has so far been avoided in tests by using the `mempool.submit_raw` function: this skips the admission checks.

To see an error for this support, you can apply this patch:

```diff
diff --git a/testnet/stacks-node/src/mockamoto/tests.rs b/testnet/stacks-node/src/mockamoto/tests.rs
index d1f369641..8fcedca83 100644
--- a/testnet/stacks-node/src/mockamoto/tests.rs
+++ b/testnet/stacks-node/src/mockamoto/tests.rs
@@ -11,7 +11,7 @@ use stacks_common::util::hash::to_hex;
 use super::MockamotoNode;
 use crate::config::{EventKeyType, EventObserverConfig};
 use crate::neon_node::PeerThread;
-use crate::tests::neon_integrations::test_observer;
+use crate::tests::neon_integrations::{test_observer, submit_tx};
 use crate::tests::{make_stacks_transfer, to_addr};
 use crate::{Config, ConfigFile};
 
@@ -45,6 +45,7 @@ fn observe_100_blocks() {
     .unwrap();
     let burnchain = conf.get_burnchain();
     let sortdb = burnchain.open_sortition_db(true).unwrap();
+    let http_origin = format!("http://{}", &conf.node.rpc_bind);
 
     let start = Instant::now();
 
@@ -72,20 +73,7 @@ fn observe_100_blocks() {
         info!("Block height observed: {stacks_block_height}");
 
         if stacks_block_height == 1 {
-            let tip = NakamotoChainState::get_canonical_block_header(chainstate.db(), &sortdb)
-                .unwrap()
-                .unwrap();
-            mempool
-                .submit_raw(
-                    &mut chainstate,
-                    &sortdb,
-                    &tip.consensus_hash,
-                    &tip.anchored_header.block_hash(),
-                    transfer_tx.clone(),
-                    &ExecutionCost::max_value(),
-                    &StacksEpochId::Epoch30,
-                )
-                .unwrap();
+            submit_tx(&http_origin, &transfer_tx);
         }
 
         if stacks_block_height >= 100 {
```

and run:
```
$ cargo test -p stacks-node observe_100_blocks -- --nocapture
```

You'll see the transfer tx fails to submit with an error (and the test will fail).

[...]

PR #4149, opened 2023-12-11
Fix mempool admission for Nakamoto Blocks

jferrant:
Changes
- added get_nakamoto_block_height to check against nakamoto_staging_blocks in chainstate
- first attempt to check against nakamoto before checking against regular stacks when retrieving the block height in admission check
- updates test to use the submit_tx rather than bypassing the admission checks. 
- increase the tx fee so it doesn't fail to submit the tx because of fee too low

Fixes https://github.com/stacks-network/stacks-core/issues/4148


testnet/stacks-node/src/mockamoto/tests.rs
```diff
@@ -72,20 +59,7 @@ fn observe_100_blocks() {
...
-                    &mut chainstate,
-                    &sortdb,
-                    &tip.consensus_hash,
-                    &tip.anchored_header.block_hash(),
-                    transfer_tx.clone(),
-                    &ExecutionCost::max_value(),
-                    &StacksEpochId::Epoch30,
-                )
-                .unwrap();
+            submit_tx(&http_origin, &transfer_tx);
```
kantai, line 62: Can you make this a new test, like `mempool_rpc_submit()`, and leave the old test unchanged?

stackslib/src/chainstate/stacks/db/blocks.rs
```diff
@@ -6330,23 +6330,32 @@ impl StacksChainState {
...
-                }
-                Err(_e) => {
-                    panic!("DB CORRUPTION: failed to query block height");
+        // First check if this transaction is valid against the Nakamoto chain tip
+        let mut staging_height = NakamotoChainState::get_nakamoto_block_height(
+            self.db(),
+            current_consensus_hash,
+            current_block,
+        )
+        .expect("DB CORRUPTION: failed to query block height");
```
kantai, line 6339: I think it's better to not panic here:

```suggestion
        .map_err(|e| {
            warn!("DB Error during mempool admission: {e:?}");
            MemPoolRejection::NoSuchChainTip(
                        current_consensus_hash.clone(),
                        current_block.clone(),
             )
        })?; 
```

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -939,6 +939,17 @@ impl NakamotoChainState {
         Ok(())
     }
 
+    /// Get the height of the nakamoto staging block
+    pub fn get_nakamoto_block_height(
```
jcnelson, line 943: Staging blocks aren't guaranteed to be processed.  I think you want to drop this method altogether, and use `NakamotoChainState::get_canonical_stacks_header()` to get the highest _processed_ block.

stackslib/src/chainstate/stacks/db/blocks.rs
```diff
@@ -6330,23 +6330,32 @@ impl StacksChainState {
...
+        if staging_height.is_none() {
+            // If we fail to find the transaction against Nakamoto, check pre Nakamoto
+            staging_height = self
+                .get_stacks_block_height(current_consensus_hash, current_block)
+                .expect("DB CORRUPTION: failed to query block height");
```
kantai, line 6344: I think it's better to not panic here:

```suggestion
        .map_err(|e| {
            warn!("DB Error during mempool admission: {e:?}");
            MemPoolRejection::NoSuchChainTip(
                        current_consensus_hash.clone(),
                        current_block.clone(),
             )
        })?; 
```

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: EDIT: this might not be necessary at all; see below.

kantai (approved):
LGTM -- just some small feedback.

stackslib/src/chainstate/stacks/db/blocks.rs
jferrant: Can do :)

stackslib/src/chainstate/stacks/db/blocks.rs
```diff
@@ -6330,23 +6330,32 @@ impl StacksChainState {
         StacksChainState::can_admit_mempool_semantic(tx, is_mainnet)?;
 
         let conf = self.config();
-        let staging_height =
```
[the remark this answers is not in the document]
kantai: But what is the alternative, in that case?

Adding a nakamoto-switch here for validation, where in one path, the `block_height` must be passed in, and in nakamoto it isn't (and poison microblocks are just always rejected)?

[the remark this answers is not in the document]
jferrant: Ohhhhh really? well then this PR becomes substantially smaller XD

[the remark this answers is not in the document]
jcnelson: Also, the `BadNonces` error case below can be dropped, since there's no longer an unconfirmed microblock chain tip to check.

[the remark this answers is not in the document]
jferrant: I mean we still need to keep the check for PoisonMicroblok for Pre-nakamoto right? so will need a check to see what epoch we are at?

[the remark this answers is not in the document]
jcnelson: That, and removing the code that queries the `staging_block` and updating the test to use `submit_tx()`

testnet/stacks-node/src/mockamoto/tests.rs
jferrant: Will do!

jcnelson (approved):

kantai (approved):
LGTM!

Issue #4152, opened 2023-12-11
[stacks-signer] Make Signers use only two stacker-db slots each


jferrant:
Update stacks signer's to only utilize one stacker db slot each. This means also updating the generate files CLI command.

xoloki:
I know this is possible, but it means that there is no ability for a signer, or anyone else, to use stackerdb to recreate state.  If we do this we might as well just pass around p2p messages.

In the current design, we have one message of every type for every signer, so we can always see the current dog and signing state for every signer.

jcnelson:
@xoloki How big is the per-slot state? Would it all fit into a single stackerdb chunk?  Because if so, you could achieve the same end simply by appending the new DKG message to the chunk instead of overwriting it.

xoloki:
@jcnelson dkg messages vary in size based on the total number of reward slots, threshold needed to sign, and number of slots controlled by a signer. It's doubtful you'd fit even one, much less multiple, into a single chunk in a production scenario.

jcnelson:
@xoloki The trade-off is as follows:

* We can use stackerdb as a persistence mechanism for DKG state, but that would mean that we need three separate stackerdb instances -- one for the current signers, one for the new signers coming online at the reward cycle boundary, and one for the miners.  This is certainly doable, but the cost is that each node would need to do bookkeeping for three separate stackerdb instances.  This entails synchronizing full DB inventory messages on each StackerDB replica state-machine pass (every `write-freq` seconds as obtained by the contract), which at 20kb each would be around 60kb of data exchanged with each neighbor.  If `write-freq` is 30s, then this would amount to about 1.2mb data exchanged every 10 minutes (by contrast, a Stacks block is about 0.2-0.5mb today). 

* To cut down on the data exchanged, we can use a single stackerdb instance, which means 0.4mb exchanged every 10 minutes.  But, it would mean that we can't use stackerdb for persistence.

We can vary `write-freq` above to be longer, if benchmarking shows us that the stackerdb message broadcast code reliably gets updates sent to each replica (state-machine passes only do inventory syncs as a fall-back in case they miss a pushed message).

jcnelson:
Summarizing a huddle we had:

[...]

PR #4153, opened 2023-12-11
Only use a single slot for DKG and Signing and add second slot for fu…

jferrant:
Fixes https://github.com/stacks-network/stacks-core/issues/4152

We only need one slot for signer communication and a second slot for the miner to observe. note that we probably could get away with a single slot, but for now lets leave it like this so a miner can query the miner slot if it happens to miss an event. This enables the Clarity folks to use a single stacker db instance for both current and onboarding signers and the miners rather than having three sep stacker db instances (There is a limit of 4096 slots per stacker db instance and this ensures we are well under it for up to 2047 signers and two miners)

### DIFF
--- a/stacks-signer/src/stacks_client.rs
+++ b/stacks-signer/src/stacks_client.rs
@@ -32,7 +32,7 @@ const BACKOFF_MAX_INTERVAL: u64 = 16384;
 /// Temporary placeholder for the number of slots allocated to a stacker-db writer. This will be retrieved from the stacker-db instance in the future
 /// See: https://github.com/stacks-network/stacks-blockchain/issues/3921
 /// Is equal to the number of message types
-pub const SLOTS_PER_USER: u32 = 10;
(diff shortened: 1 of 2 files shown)


jcnelson (approved):
LGTM as long as the stackerdb tests pass

xoloki (approved):
As I explained in #4152 I think this change is a bad idea; if we're just overwriting the slot with every message, there's no point to having a stackerdb at all; we should just pass p2p messages and skip this, there's no benefit and it's a big chunk of code with uncertain performance.

But this PR does do what #4152 is asking so I'll approve as is.

jferrant:
@setzeus Heads up. We are closing this ticket for now with a won't do. There are some PROs to keeping 10 slots per signer specifically around signers thta crash in the middle of a round or come online in the middle of one as they can then sync up more easily and either join or continue the round.

PR #4156, opened 2023-12-12
Fix broken mempool_setup_chainstate test

jferrant:
Not sure how this PR passed CI: https://github.com/stacks-network/stacks-core/pull/4149 because when I investigated closer...it caused a failed test (mempool_setup_chainstate). This PR fixes the failing test.

### DIFF
--- a/testnet/stacks-node/src/tests/mempool.rs
+++ b/testnet/stacks-node/src/tests/mempool.rs
@@ -699,13 +699,7 @@ fn mempool_setup_chainstate() {
                     )
                     .unwrap_err();
                 eprintln!("Err: {:?}", e);
-                assert!(
-                    if let MemPoolRejection::PoisonMicroblocksDoNotConflict = e {
-                        true
-                    } else {
-                        false
-                    }
-                );
+                assert!(matches!(e, MemPoolRejection::Other(_)));
 
                 let microblock_1 = StacksMicroblockHeader {
                     version: 0,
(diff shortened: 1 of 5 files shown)


testnet/stacks-node/src/tests/mempool.rs
```diff
@@ -843,15 +827,17 @@ fn mempool_setup_chainstate() {
...
                     .will_admit_mempool_tx(
                         &NULL_BURN_STATE_DB,
                         consensus_hash,
                         block_hash,
                         &tx,
                         tx_bytes.len() as u64,
                     )
-                    .unwrap();
+                    .unwrap_err();
+                eprintln!("Err: {:?}", e);
```
xoloki, line 839: This prinkln can go away

xoloki (approved):
Looks reasonable modulo the extraneous eprintln, but the test is still falling

testnet/stacks-node/src/tests/mempool.rs
jferrant: Its there for consistency with the rest of the test.

jferrant:
> Looks reasonable modulo the extraneous eprintln, but the test is still falling

I think this is a caching problem. Local runs show it passes no problem. And the error it returns is not making sense. Hopefully retriggering will fix it.

jcnelson (approved):

saralab (approved):

jcnelson:
My PR ate this PR

PR #4157, opened 2023-12-12
Improve revoke delegate stx

### LINKED ISSUE #4143: PoX-4 events are missing
In the latest `next` branch, the neon node is not emitting contract events for `pox-4`. It's missing both `stx_lock_event` and the synthetic pox events.

These are critical for the API, exchanges, and pool operators.

### LINKED ISSUE #4146: PoX-4 `revoke-delegate-stx` calls should fail if no state is changed
There's a discussion in PR https://github.com/stacks-network/stacks-core/pull/4094 around the behavior of `revoke-delegate-stx`. Right now the function always returns `(ok)` even if no state was changed:

https://github.com/stacks-network/stacks-core/blob/2c1c777b9ed0124a65e9cd2c019bd85f0735a5e0/stackslib/src/chainstate/stacks/boot/pox-4.clar#L636-L641

This means A) the new synthetic event for `revoke-delegate-stx` can be triggered repeatedly for no valid/useful reason, and B) the event has empty data: https://github.com/stacks-network/stacks-core/blob/789c23544e95058baa8610f3b928c238747e12a1/pox-locking/src/events.rs#L340-L346

Ideally the event would at least include the `delegate-to` address so that the event can be associated with the prior delegation. And ideally the call would return an error and not emit repeated events if called repeatedly.

friedger:
### Description
This PR improves the function `revoke-delegate-stx` so that the information which pool the user just left is transparent. The function call would fail if the user is not part of a pool.

Furthermore, an event for revoking is synthesized similar to other pox events.

### Applicable issues

- fixes #4146 
- fixes #4143 

### Checklist

- [ ] Changelog is updated


stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -658,12 +659,17 @@
           ;; return the lock-up information, so the node can actually carry out the lock.
           (ok { stacker: tx-sender, lock-amount: amount-ustx, unlock-burn-height: (reward-cycle-to-burn-height (+ first-reward-cycle lock-period)) }))))
 
+;; Revokes the delegation to the current stacking pool.
+;; New in pox-4: Fails if the delegation was already revoked.
+;; Returns the last delegation state even if the delegation is already expired.
 (define-public (revoke-delegate-stx)
-  (begin
+  (let ((stacker { stacker: tx-sender })
+        (last-delegation-state (map-get? delegation-state stacker)))
```
friedger, line 667: Should we use `get-check-delegation` instead of map-get and fail also if the delegation already expired?

zone117x: That makes sense to me, one less odd condition that event-observers have to deal with

zone117x (commented):
Thanks @friedger! Should this PR be against the `next` branch?

friedger:
@zone117x This PR uses infrastructure code (yarn test) added in #4094

stackslib/src/chainstate/stacks/boot/pox-4.clar
friedger: However, that implies that the map entry for expired delegation might never be deleted.

zone117x: Did they get deleted prior to this PR if users never performed a `revoke-delegate-stx`?

friedger: Applied the change so that revoke-delegate-stx also fails after expiry

friedger: Today, they are not deleted if the user does not perform a revoke. However, with this change, the user does not have a chance to remove the map entry once it is expired. 

I think it does not matter.

pox-locking/src/events.rs
```diff
@@ -335,11 +341,31 @@ fn create_event_info_data_code(function_name: &str, args: &[Value]) -> String {
...
+                    delegate_to = opt
+                        .data
+                        .map(|boxed_value| *boxed_value)
+                        .unwrap()
+                        .expect_tuple()
+                        .get("delegated-to")
+                        .unwrap()
+                )
+            } else {
+                "{data: {unimplemented: true}}".into()
```
jcnelson, line 361: Is this reachable?  Same for below.

jcnelson (approved):
LGTM!  Thanks @friedger

pox-locking/src/events.rs
friedger: Yes, for older pox contracts, because the older contracts return no data.

setzeus (approved):

kantai (approved):

zone117x (approved):
I have yet to test these changes in the API, but so far LGTM. This should be merged, and I'll open an issue if any bugs are uncovered during testing.

jcnelson (approved):

pox-locking/src/events.rs
```diff
@@ -335,11 +341,31 @@ fn create_event_info_data_code(function_name: &str, args: &[Value]) -> String {
                 pox_addr = &args[3],
             )
         }
-        _ => "{{ data: {{ unimplemented: true }} }}".into(),
+        "revoke-delegate-stx" => {
+            if let Value::Optional(opt) = *response.data.clone() {
+                format!(
+                    r#"
+                    {{
+                        data: {{ delegate-to: '{delegate_to} }}
```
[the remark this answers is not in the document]
friedger: @janniks I was inspired by the delegate-stx parameter name and the database column of the pox3_events table

But your suggestion makes sense.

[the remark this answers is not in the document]
friedger: Created new issue https://github.com/stacks-network/stacks-core/issues/4246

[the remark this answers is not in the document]
janniks: Yeah, good point. Might make sense in that case to keep named the same. Not sure 🤔

PR #4158, opened 2023-12-12
Experimental Refactor: Traits, composability and DI

cylewitruk:
Creating a draft PR of this **experimental** branch simply to be able to see changes more easily - this isn't meant to be merged, and will likely result in a number of smaller, more area-focused PRs.

Please see this PR as purely suggestive changes/ideas, and I welcome input/discussion as such, in hopes that we may be able to implement some of the ideas born here to make for a cleaner and more understandable codebase :)

This refactoring was born out of frustrations with the current ownership model in the code being quite inflexible. The idea here is to move toward more traits and composability w/static over dynamic dispatch, and also introducing a dependency injection model (i.e. the binary/caller should define the implementations to use, not the implementations themselves).

I call it experimental because a lot of this is trial and error trying to find the right abstractions, and the chances that you'll check out this branch in a compileable-state are slim ;)


stackslib/src/chainstate/burn/db/v2/test_helpers.rs
```diff
@@ -0,0 +1,1011 @@
...
+
+use crate::chainstate::stacks::index::trie_db::TrieDb;
+
+use super::{SortitionDb, SortitionDbTransaction};
+
+pub struct InMemorySortitionDb {
+
+}
+
+impl SortitionDbTransaction for InMemorySortitionDb {
```
jcnelson, line 11: Wow, this trait is _huge_!  This, to me, suggests that this is the wrong place to add a layer of indirection -- it looks like the trait is really just coding to a particular implementation.  Also, a lot of these methods are specific to Stacks 2.x, and are gone in Nakamoto.  Perhaps there's a way to decompose this trait into a few smaller, well-defined traits?

stackslib/src/chainstate/burn/db/v2/test_helpers.rs
```diff
@@ -0,0 +1,1011 @@
...
+    #[cfg(test)]
+    fn read_all_block_hashes_and_roots<T: crate::chainstate::stacks::index::MarfTrieId>(
+        &self,
+    ) -> Result<Vec<(stacks_common::types::chainstate::TrieHash, T)>, crate::chainstate::stacks::index::Error> {
+        todo!()
+    }
+}
+
+
+impl SortitionDb for InMemorySortitionDb {
```
jcnelson, line 469: Same here -- this trait is way too big and encompasses several orthogonal concerns (some of which I don't think belong in the trait interface at all).

stackslib/src/core/mempool.rs
```diff
@@ -316,15 +318,16 @@ impl MemPoolAdmitter {
         self.cur_consensus_hash = cur_consensus_hash.clone();
         self.cur_block = cur_block.clone();
     }
-    pub fn will_admit_tx<Conn>(
```
jcnelson: So, one wrinkle with adding all these type parameters is that a copy of this code will be generated for each type concretization. This would bloat the binary considerably and would slow down compile times.

cylewitruk: Yeah, that's the downside, but compile-time safety and a little performance would be to gain. I'm sort-of counting on that there are only three implementations (null, memory and sqlite), and we could gate null + memory to only compile for `test`, so my hope is that it shouldn't bloat _too_ much.

stackslib/src/chainstate/burn/db/v2/test_helpers.rs
cylewitruk: Haha yes :) I definitely think that it could be broken up (like I did for `ClarityDb*` - but since I'm less proficient in these areas I opted to simply take them "as-is" from their struct impl blocks.

stackslib/src/chainstate/burn/db/v2/test_helpers.rs
```diff
@@ -0,0 +1,1011 @@
...
+        next_pox_info: Option<crate::chainstate::coordinator::RewardCycleInfo>,
+        parent_pox: stacks_common::types::chainstate::PoxId,
+        reward_set_info: Option<&crate::chainstate::burn::operations::leader_block_commit::RewardSetInfo>,
+        initial_mining_bonus_ustx: u128,
+    ) -> super::Result<(crate::chainstate::burn::BlockSnapshot, crate::burnchains::BurnchainStateTransition)> {
+        todo!()
+    }
+}
+
+impl TrieDb for InMemorySortitionDb {
```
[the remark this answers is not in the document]
cylewitruk: Completely agree :) Would need some help there though in sorting out which methods should belong to which trait.

stackslib/src/chainstate/burn/db/v2/test_helpers.rs
cylewitruk: Yeah, same answer here, I simply copy-pasted from impl blocks - any guidance here is welcomed :)

stackslib/src/chainstate/burn/db/v2/test_helpers.rs
cylewitruk: An afterthought is that we should _probably_ keep it as a big trait _to begin with_, and split it up in a later iteration - just to try to minimize the changes in a single PR as much as possible.

PR #4162, opened 2023-12-12
use pull_request to trigger ci runs

wileyj:
this is a minor fix in how the CI triggers due to issues noticed downstream in next.
pull_request_target is not working the way that was expected, in that it is building against the target branch, and not the base branch.
this commit fixes that.

these changes are coming down from develop in https://github.com/stacks-network/stacks-core/pull/4161


jcnelson (approved):

saralab (approved):
Let's Go! 🔥 🚀

Issue #4109, opened 2023-11-29
Nakamoto: Determine which stacks block to fetch an aggregate public key from


kantai:
Related to #4039 and somewhat downstream of #4107: the `stacks-node`, given a Stacks block `new_block`, needs to figure out the stacker aggregate public key. To do this, it queries `pox-4` state, but it needs two critical pieces of information:

1. The reward cycle number for the aggregate public key. I think this should be the reward cycle of the bitcoin block that elected `new_block`'s tenure.
2. The Stacks Block ID at which to invoke `get-aggregate-public-key`. This is important because the aggregate key is stored in stacks chain state, meaning that it will have a different value at different points in the chain, so there must be _one_ value that is agreed upon here.

In order to support #4039, the answer to (2) would need to be something like the parent of `new_block`. This complicates some matters though, because it means signature validation (and therefore block acceptance/download) cannot proceed until the parent block is processed. Alternatively, the aggregate public key could be computed much like Stacks-2.0 computes the reward-set: at one specific block for the duration of the reward cycle. To implement this, the check would look up the last stacks block of the prepare phase.

jcnelson:
> In order to support #4039, the answer to (2) would need to be something like the parent of new_block

It needs to be the parent of the first block in the reward cycle; the last block in the previous reward cycle (e.g. the last block in the current reward cycle's prepare phase).  The block downloader and IBD behavior will need to guarantee that this block will be available before the node will download and process blocks the subsequent reward cycle (much like how it does for the anchor block)

kantai:
> It needs to be the parent of the first block in the reward cycle; the last block in the previous reward cycle (e.g. the last block in the current reward cycle's prepare phase). The block downloader and IBD behavior will need to guarantee that this block will be available before the node will download and process blocks the subsequent reward cycle (much like how it does for the anchor block)

Ah, yes, I agree. I linked the wrong issue in that first sentence: "parent of `new_block`" would be the block to evaluate if supporting agg key change during the reward cycle.

jcnelson:
An open question here is whether or not we allow DKG to happen mid-cycle, such as to include more stackers who were offline in the prepare phase. If we do that in any form, then this solution changes to some form of "check the chain tip for the aggregate public key" instead of "check the last block in the last reward cycle".

kantai:
> whether or not we allow DKG to happen mid-cycle, such as to include more stackers who were offline in the prepare phase. "check the chain tip for the aggregate public key"

[...]

PR #4165, opened 2023-12-13
Get/Set aggregate public key from Clarity

jcnelson:
This merges #4113 and #4126


jcnelson:
Actually, let me rebase this onto `feat/naka-neon`

xoloki (approved):
There's a stale comment in `observe_set_aggregate_key`, but ow LGTM.  Thanks so much for untangling this @jcnelson

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -2533,6 +2552,103 @@ impl NakamotoChainState {
...
+            )
+            .ok()
+            .map(|agg_key_value| {
+                let agg_key_opt = agg_key_value.expect_optional();
+                let agg_key_buff =
+                    agg_key_opt.expect("FATAL: aggregate public key not set in boot code");
+                Value::buff_from(agg_key_buff.expect_buff(33))
+                    .expect("failed to reconstruct buffer")
+            })
+            .expect("get-aggregate-public-key returned None");
```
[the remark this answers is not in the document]
jcnelson: This whole function is going away once we have aggregate public key voting

testnet/stacks-node/src/run_loop/nakamoto.rs
```diff
@@ -218,13 +225,49 @@ impl RunLoop {
             .map(|e| (e.address.clone(), e.amount))
             .collect();
 
-        // TODO (nakamoto-neon): check if we're trying to setup a self-signing network
-        //  and set the right genesis data
+        let agg_pubkey_boot_callback = if let Some(self_signer) = self.config.self_signing() {
```
jferrant, line 228: Ah this was the key I was needing. Was wondering how to set boot contract state, but couldn't seem to figure it out!

jferrant (approved):

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -2533,6 +2566,113 @@ impl NakamotoChainState {
...
+                    agg_key_opt
+                })
+                .flatten()
+            else {
+                debug!(
+                    "No aggregate public key in parent cycle {}",
+                    parent_reward_cycle
+                );
+                continue;
+            };
```
kantai, line 2636: Looping over reward cycles should be unnecessary. If an integration test requires it, the integration test itself should be altered, rather than this code.

jcnelson: The problem that this fixes is that integration tests do the following:

1. Install the aggregate public key in `pox-4` for reward cycle 0
1. Boot up 200-ish Bitcoin blocks (which is many reward cycles)
1. Mine some epoch 2.x blocks and instantiate `pox-4` (more reward cycles)
1. Start epoch 3, and try to build a block

Without searching back to reward cycle 0, this last step would fail with a panic because (1) there's no way to "pull forward" the aggregate public key from reward cycle 0 when the first Stacks block is mined without this search code, and (2) epoch 2.x block-production doesn't use this code.

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -471,10 +478,50 @@ impl RunLoop {
             .map(|e| (e.address.clone(), e.amount))
             .collect();
 
+        // TODO: delete this once aggregate public key voting is working
+        let agg_pubkey_boot_callback = if let Some(self_signer) = self.config.self_signing() {
+            let agg_pub_key = to_hex(&self_signer.aggregate_public_key.compress().data);
+            info!("Mockamoto node setting agg public key"; "agg_pub_key" => &agg_pub_key);
```
kantai, line 484: This is now the neon/nakamoto node.

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: Remember, this whole code path goes away once aggregate public key voting comes online.  Then, we'd have stackers vote for (and thus set) the aggregate public key in `pox-4` in the penultimate reward cycle prior to epoch 3.0.  This removes the need for this entire method.

testnet/stacks-node/src/run_loop/neon.rs
```diff
@@ -471,10 +478,50 @@ impl RunLoop {
...
+                            ClarityVersion::Clarity2,
+                            &ast,
+                            &contract_content,
+                            None,
+                            |_, _| false,
+                        )
+                        .unwrap();
+                    clarity.save_analysis(&contract_id, &analysis).unwrap();
+                })
+            }) as Box<dyn FnOnce(&mut ClarityTx)>;
```
kantai, line 514: This should be de-duped -- there's three instances of this code block.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1750,34 +1754,53 @@ impl NakamotoChainState {
...
+                None => {
+                    debug!(
+                        "No aggregate public key set; trying in a lower cycle";
+                        "block_id" => %at_block_id,
+                        "reward_cycle" => rc,
+                    );
+                    continue;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
jcnelson: We call `NakamotoChainState::get_aggregate_public_key()` prior to processing the first Nakamoto block (which contains the `set-aggregate-public-key` call), since the relayer needs the aggregate public key to validate the block.

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -170,13 +171,35 @@ impl BlockMinerThread {
...
+
+            let aggregate_public_key = NakamotoChainState::get_aggregate_public_key(
+                &sort_db,
+                &sortition_handle,
+                &mut chain_state,
+                block_sn.block_height,
+                &aggregate_key_block_header.index_block_hash(),
+            )?;
+            aggregate_public_key
+        };
```
kantai, line 194: This should be de-duped with the invocation in the mockamoto node, perhaps a

`NakamotoChainState::get_stacker_key_for_block(&mut chain_state, &sortdb, &block.header)`

Or, perhaps this logic should be part of the invocation of `get_aggregate_public_key`, which should take `&block.header` instead of `block_sn.block_height` and `at_block: StacksBlockId`. Otherwise, we're going to have the logic of how to find an aggregate key for a given block repeated many times in the codebase (if we haven't already?)

jcnelson: Yup, can do

stackslib/src/chainstate/nakamoto/mod.rs
kantai: What you described sounds incorrect to me— the epoch 3.0 instantiation code sets all pre-pox-4 cycles’ agg key. Then the set-agg-key method is invoked in every block after that. This loop is masking an issue.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1803,6 +1800,29 @@ impl NakamotoChainState {
...
+        sortdb: &SortitionDB,
+        sort_handle: &SH,
+        block: &NakamotoBlock,
+    ) -> Result<Point, ChainstateError> {
+        let block_sn =
+            SortitionDB::get_block_snapshot_consensus(sortdb.conn(), &block.header.consensus_hash)?
+                .ok_or(ChainstateError::DBError(DBError::NotFoundError))?;
+        let aggregate_key_block_header =
+            Self::get_canonical_block_header(chainstate.db(), sortdb)?.unwrap();
+
```
kantai, line 1815: Can you add a todo comment here with a reference to #4109 ? As currently implemented here (which I think is fine for now) the `at_block` is variable: when a new stacks block is received, `get_canonical_block_header()` returns a different Stacks block, meaning this `get_aggregate_public_key()` will return different values for the same `block` at different times.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1803,6 +1800,29 @@ impl NakamotoChainState {
         ))
     }
 
+    /// Get the aggregate public key for a block
```
kantai, line 1803: Maybe just expand this comment that this is fetching the _expected_ stacker aggregate key for a given block.

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: I'm actually no longer sure that #4109 is desirable, if we're going to allow stackers to re-try DKG in the middle of the cycle.  If that were the case, then we'd want the current behavior.

stackslib/src/chainstate/nakamoto/mod.rs
jcnelson: Actually, it's the epoch 2.5 instantiation logic that populates the pre-pox-4 reward cycles' aggregate public key.  And in the integration tests, epoch 2.5 starts at Bitcoin block height 6, which is well before the first Stacks block is mined.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Maybe so, but that merits discussion and that discussion shouldn't stall this PR. Leaving a pointer to that issue makes it so that it's easy to find the code that assumes one approach or the other.

As written, this function may be somewhat dangerous -- I don't think it's strange for a developer to believe that the expected aggregate key for a given block is always the same, but this function will return different results at different times.

Issue #4120, opened 2023-12-04
Stackerdb Benchmark testing


wileyj:
Stackerdb needs a benchmark function to determine how long it takes all nodes to see an update. 
*very* similar to how ping/traceroute works- the end result is we have an idea of how long it takes an object to be shared across a network. 

stackerdb instance (publish contract with 2 slots) 
  - 1 node posts to their slot a random value (waits for everyone to see that value)
    - measure time it takes to receive the response (when locally posted and locally see the value)
  - separate process to measure (like a ping tool)

jferrant:
Before I get deep in the weeds, this seems to be a request to create a new binary that implements stacker db soley for the prupose of measuring its time to disseminate messages, yes? I..e a binary that writes a message to stacker db and waits for every other binary that is subscribed/part of the ystem to receive it and respond back to the originator?

CAGS295:
@wileyj I still need to be assigned this task. My involvement in this is not clear to me. Are we both working on this ticket together?

My line of thought follows what @jferrant said. My a priori guess is to implement pinging and a broadcast scan. I need to take a look at libstackerdb first to get more context.

wileyj:
> Before I get deep in the weeds, this seems to be a request to create a new testing binary that implements stacker db solely for the purpose of measuring its time to disseminate messages, yes? I..e a binary that writes a message to stacker db and waits for every other binary that is subscribed/part of the system to receive it and respond back to the originator?
> 
> I will see if I can make an even more generic solution that adds a trait to libstackerdb to implement a ping function.

i think the implementation is up to you  - the goal is that it's something that we can invoke as needed. 
what you proposed as a question is the gist of it - a ping tool for stackerdb to see how long it takes for each node to see a change. 
@CAGS295 pinged you since this might be something you can dig into as part of sending you more blockchain work. 
If you think you're up to it, i would encourage you to get in touch with @AshtonStephens and @jferrant, but you're more than welcome to take ownership of this if you'd like. 

ideally, this is something we want to have functional in the next few weeks (i.e. prior to Jan 2024).

jcnelson:
We need to alter the signer binary to have a "benchmark" mode.  We want to measure the distribution of round-trip times (RTTs) for signers.  To do this, we'll need the following:

* A "benchmark-runner" mode for the signer.  This mode drives the benchmark by sending a "ping" to a designated slot (e.g. slot 0), and measuring the RTTs for all other signers to respond to the ping.
[...]

PR #4167, opened 2023-12-13
WIP: StackerDB benchmark

CAGS295:
### Description

Early-stage development for #4120 

### Applicable issues

- fixes #

### Checklist

- [ ] Changelog is updated


stacks-signer/src/cli.rs
```diff
@@ -152,6 +154,17 @@ pub struct GenerateFilesArgs {
     pub timeout: Option<u64>,
 }
 
+#[derive(Parser, Debug, Clone)]
+/// Arguments for the generate-files command
```
jcnelson, line 158: What `generate-files command`?

stacks-signer/src/main.rs
```diff
@@ -294,6 +296,57 @@ fn handle_generate_files(args: GenerateFilesArgs) {
...
 
+fn handle_ping(args: PingArgs) {
+    debug!("Pinging Signers...");
+    let config = Config::try_from(&args.config).unwrap();
+
+    // Ping::RunLoop doesnt use the command channel.
+    let (_cmd_send, cmd_recv) = channel();
+    let (res_send, res_recv) = channel();
+    let ev = StackerDBEventReceiver::new(vec![config.stackerdb_contract_id.clone()]);
+    //let stop_signal = ev.get_stop_signaler().unwrap();
```
jcnelson, line 307: Nit: please remove commented-out code

stacks-signer/src/main.rs
```diff
@@ -294,6 +296,57 @@ fn handle_generate_files(args: GenerateFilesArgs) {
...
+    );
+    let running_signer = Signer::new(runloop, ev, cmd_recv, res_send)
+        .spawn(config.endpoint)
+        .unwrap();
+
+    // hacky, single use-client to send the request.
+    let chunk_ack = {
+        let mut client = StacksClient::from(&config);
+
+        // `send_message_with_retry` will increment the counter prior use; POLA, use the value provided by user.
```
jcnelson, line 328: What's `POLA`?

stacks-signer/src/ping/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+pub enum Packet {
+    /// Outgoing
+    Ping(Ping),
+    /// Incoming
+    Pong(Pong),
+}
+
+#[derive(Serialize, PartialEq, Clone, Deserialize, Debug)]
+/// A ping in a slot means someone has requested you to push a Pong into your slot.
+pub struct Ping {
```
jcnelson, line 30: Can you expand this so that the `Ping` can have an arbitrarily-large payload?  We'll want to see what the RTT looks like when the message must be sent as a flow of multiple TCP segments.

stacks-signer/src/ping/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+    id: u64,
+}
+#[derive(Serialize, Deserialize, Debug, Clone)]
+/// A pong in a slot means someone has responded to an RTT request.
+pub struct Pong {
+    id: u64,
```
jcnelson, line 36: Same with the pong -- it should return the arbitrarily-large payload sent in the Ping.

stacks-signer/src/ping/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+    fn from(value: Ping) -> Self {
+        Self::Ping(value)
+    }
+}
+
+impl Ping {
+    /// Uniquely identify the RTT request
+    pub fn new() -> Self {
+        Ping {
+            id: Uuid::new_v4().as_u64_pair().0,
```
jcnelson, line 55: This is overkill.  Can you just generate a random number here?  And for the payload, just fill it with random numbers.

stacks-signer/src/ping/processor.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+///Holds the RTT distribution.
+struct EntryData {
+    tick: Instant,
+    average: RTTAverage,
+    variates: Vec<Duration>,
+}
+
+/// Calculate the average of an RTT ping-pong.
```
jcnelson, line 30: The mean ping time isn't likely to be a meaningful statistical summary of the distribution of ping times, since there will likely be a long tail of response times.  Can you instead simply capture the RTTs and just dump them?  Let's leave statistical analysis to the caller.

stacks-signer/src/ping/runloop.rs
```diff
@@ -0,0 +1,130 @@
+use stacks_common::{debug, trace};
```
jcnelson, line 1: Why is there a wholly separate runloop?  Can't the ping/pong just be part of the main runloop?

stacks-signer/src/ping/mod.rs
CAGS295: ack

stacks-signer/src/ping/mod.rs
CAGS295: ack

stacks-signer/src/ping/processor.rs
CAGS295: ack

stacks-signer/src/ping/runloop.rs
CAGS295: Acknowledged, I'll shift efforts into a unified RunLoop.
On an earlier attempt, I had a single run loop. I noticed high variance (I had seen 200ms vs 15s frequently). My unbacked PoV is that there are many points in the RunLoop where it could be stalled. Waiting on commands, retry loops with exponential back-off requests to the stacks-node, expensive crypto computation, etc. And it could add a lot of variance. But what I had implemented wasn't RTT. It was one-way latency. On my second attempt, I right away scraped the single Runloop in favor of a dedicated one.

stacks-signer/Cargo.toml
```diff
@@ -33,16 +42,17 @@ reqwest = { version = "0.11.22", features = ["blocking", "json"] }
...
 slog-term = "2.6.0"
 stacks-common = { path = "../stacks-common" }
-stackslib = { path = "../stackslib"}
+stackslib = { path = "../stackslib" }
 thiserror = "1.0"
 toml = "0.5.6"
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 wsts = "4.0.0"
+uuid = { version = "1.6.1", features = ["v4"] }
```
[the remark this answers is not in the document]
CAGS295: ack

stacks-signer/src/cli.rs
CAGS295: ack

stacks-signer/src/main.rs
CAGS295: ack

stacks-signer/src/main.rs
CAGS295: Ack. Principle of least astonishment. It was a reminder to myself why I subtract one from the slot_version.

wileyj:
@CAGS295 can you share the minimum required framework needed to test this in a real-world environment?
i.e. 2 vm's with public IP's running your fork in this PR?

CAGS295:
>2 vm's with public IP's running your fork in this PR

- Each VM runs as a seed node with a signer bin.
-  At least one signer must be run with pinging options `cargo run -p stacks-signer -- run --config=/path/to/config.toml --ping-in-millis 10000 --ping-payload-size=4096`
- The payload size option should be the max msg size allowed in the pox contract.
- The slots per signer in the contract must also be increased by one.
- The slots per signer in the contract should match [this variable](https://github.com/stacks-network/stacks-core/blob/develop/stacks-signer/src/stacks_client.rs#L34).
- If everything goes well and once the signers start processing events, you should expect lines like "New RTT for id 1234: 7.7s".

[This is the time](https://github.com/CAGS295/stacks-core/blob/dev/cags295/4120-1/stacks-signer/src/runloop.rs#L216) it took from sending a command to the run loop until each incoming Pong is processed by the Ping requester.

stacks-signer/src/cli.rs
```diff
@@ -152,6 +152,20 @@ pub struct GenerateFilesArgs {
     pub timeout: Option<u64>,
 }
 
+#[derive(Parser, Debug, Clone)]
+/// Arguments for the Run ommands
```
jcnelson, line 156: This should be an enum.

stacks-signer/src/ping/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+                f.debug_struct(stringify!($names))
+                    .field("id", &self.id)
+                    .finish()
+            }
+        }
+        )*
+    };
+}
+
+debug_skip_payload_field!(Ping, Pong);
```
jcnelson, line 108: This is overkill. Can you just implement `Debug` for `Ping` and `Pong` directly?

stacks-signer/src/runloop.rs
```diff
@@ -31,6 +32,11 @@ pub enum RunLoopCommand {
         /// Taproot merkle root
         merkle_root: Option<MerkleRoot>,
     },
+    /// Send this Command to the runlop to start an RTT broadcast.
+    Ping {
+        /// Attach a payload of random bytes to the Ping/Pong messages.
+        payload_size: u16,
```
jcnelson, line 38: The payload can be considerably bigger than 2^16 bytes.  The maximum payload size is 16 MB.

stacks-signer/src/runloop.rs
```diff
@@ -175,8 +196,50 @@ impl<C: Coordinatable> RunLoop<C> {
         let inbound_messages: Vec<Packet> = event
             .modified_slots
             .iter()
+            // intercept and consume ping packets
+            .filter_map(|chunk| {
```
jcnelson, line 200: I'm generally not a fan of deeply-nested code like this.  Can you refactor this so that (1) this behavior is encapsulated in its own method and (2) you keep the nesting / indentation level to a minimum?  Like, there's no reason to put the whole of the code body inside an `if is_ping_slot(..) { ... }` block.

stackslib/src/net/stackerdb/mod.rs
```diff
@@ -14,102 +14,102 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-/// # The StackerDB System
```
jcnelson, line 17: Please revert.

stacks-signer/src/cli.rs
CAGS295: For what? There is a single Run command at the moment. What am I missing?

stacks-signer/src/runloop.rs
CAGS295: How do you find something in between?
```rust
            // intercept and consume ping packets
            .filter_map(|chunk| {
                let msg = LatencyPacket::verify_packet(chunk, self.signing_round.signer_id)?;

                match msg {
                    LatencyPacket::Pong(pong) => {
                        let id = pong.id();
                        // Signer won't react to Pongs from Pings not initiated by it.
                        self.ping_entries.get(&id).map(|tick| {
                            let variate = tick.elapsed();
                            info!("New RTT for id {id}: {:?}", variate);
                        });
                    }
                    LatencyPacket::Ping(ping) => {
                        let msg: LatencyPacket = ping.pong().into();
                        let _ = self
                            .stacks_client
                            .send_message_with_retry(signer_id, &msg)
                            .map(|ack| debug!("ACK: {:?}", ack))
                            .map_err(|e| warn!("Sending RTT probe failed! noop with error: {e}"));
                    }
                }
                None
            })
```
I can't find a zero-cost alternative to circumvent the fact that abstracting the last match into a method requires a `&mut self` such that stacks_client can increment the slot_version. The code above works because `self` is captured by the closure, and the Iterator is collected in the enclosing scope. As far as I know, it does not work with methods the same way as with closures. Taking that much effort to put the last match behind a method is not worth it. This logic can't be used anywhere else, and the match looks irreducible beyond moving it to its own method.
```rust
impl StacksClient {
    /// Sends messages to the stacker-db with an exponential backoff retry
    pub fn send_message_with_retry<'a, Packet>(
        &mut self,
        signer_id: u32,
        message: &'a Packet,
    ) -> Result<StackerDBChunkAckData, ClientError>
    where
        Packet: Serialize,
        &'a Packet: ToSlotId,
//snip ...
```

libsigner/Cargo.toml
```diff
@@ -22,12 +31,13 @@ libstackerdb = { path = "../libstackerdb" }
...
 serde_derive = "1"
 serde_stacker = "0.1"
-slog = { version = "2.5.2", features = [ "max_level_trace" ] }
+slog = { version = "2.5.2", features = ["max_level_trace"] }
 slog-term = "2.6.0"
 slog-json = { version = "2.3.0", optional = true }
 stacks-common = { path = "../stacks-common" }
 thiserror = "1.0"
 tiny_http = "0.12"
+tracing = { version = "0.1.40", optional = true }
```
jcnelson, line 40: What's this?

stacks-signer/src/ping/periodic_pinger.rs
```diff
@@ -0,0 +1,102 @@
...
+    },
+    thread::{self, JoinHandle},
+    time::Duration,
+};
+
+use crate::runloop::RunLoopCommand;
+use tracing::{info, trace, Span};
+
+/// Stop the pinger thread
+pub struct ThreadStopHandle {
```
jcnelson, line 15: Nit: can you name this something specific to the ping thread?  Like `PingStopHandle`?

testnet/stacks-node/Cargo.toml
```diff
@@ -33,15 +40,27 @@ libsigner = { path = "../../libsigner" }
 ring = "0.16.19"
 warp = "0.3.5"
 tokio = "1.15"
-reqwest = { version = "0.11", default_features = false, features = ["blocking", "json", "rustls", "rustls-tls"] }
-clarity = { path = "../../clarity", features = ["default", "testing"]}
-stacks-common = { path = "../../stacks-common", features = ["default", "testing"] }
-stacks = { package = "stackslib", path = "../../stackslib", features = ["default", "testing"] }
-stacks-signer = { path = "../../stacks-signer" }
+reqwest = { version = "0.11", default_features = false, features = [
```
jcnelson, line 43: Generally not a fan of shipping a reformatting of unrelated files with a feature PR, since it distracts the reviewer (like, I have to go and manually check that all of the settings are the same before and after the reformat).  Can you just revert this for now?

stacks-signer/src/ping/mod.rs
```diff
@@ -0,0 +1,192 @@
+use std::{fmt, fmt::Debug};
```
jcnelson, line 1: Missing copyright statement

stacks-signer/src/ping/periodic_pinger.rs
```diff
@@ -0,0 +1,102 @@
+use std::{
```
jcnelson, line 1: Missing copyright statement

stacks-signer/src/ping/periodic_pinger.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    fn keep_running(&self) -> bool {
+        self.keep_running.load(Ordering::Relaxed)
+    }
+}
+
+impl ThreadStopHandle {
+    /// Stop the open loop and join the thread handle
+    pub fn stop(self) {
+        self.keep_running.store(false, Ordering::Relaxed);
```
jcnelson, line 77: Why `Ordering::Relaxed` and not `Ordering::SeqCst`?

stacks-signer/src/runloop.rs
```diff
@@ -1,11 +1,13 @@
-use std::collections::VecDeque;
+use std::collections::{HashMap, VecDeque};
```
jcnelson, line 1: Can you also add a copyright statement to this file?

stacks-signer/src/runloop.rs
```diff
@@ -175,8 +194,40 @@ impl<C: Coordinatable> RunLoop<C> {
         let inbound_messages: Vec<Packet> = event
             .modified_slots
             .iter()
+            // intercept and consume ping packets
+            .filter(|chunk| {
```
jcnelson, line 198: Thanks for cleaning this up a bit.  Can you also put this `filter()` and `filter_map()` logic into its own method?

stacks-signer/src/utils.rs
```diff
@@ -2,9 +2,9 @@ use std::time::Duration;
 
 use p256k1::ecdsa;
 use rand_core::OsRng;
-use slog::slog_debug;
```
jcnelson, line 5: Can you please add a copyright statement to this file?

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -329,3 +329,112 @@ fn test_stackerdb_dkg() {
...
+        signer_cmd_senders.push(cmd_send);
+        signer_res_receivers.push(res_recv);
+    }
+
+    let coordinator_cmd_send = signer_cmd_senders.last().unwrap().clone();
+    let coordinator_res_recv = signer_res_receivers.last().unwrap();
+
+    // Let's scope the node to ensure stopping the signers doesn't cause issues.
+    {
+        // Build our simulated pox-4 stacks contract TODO: replace this with the real deal?
```
jcnelson, line 383: Is this TODO resolvable in this PR?

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -329,3 +329,112 @@ fn test_stackerdb_dkg() {
...
+                };
+            }
+        }
+    }
+
+    for signer in running_signers {
+        assert!(signer.stop().is_none());
+    }
+
+    logs_assert(|lines| {
```
jcnelson, line 427: Instead of scanning log files here and elsewhere, can you instead structure the pinger thread to just keep track of RTTs it has observed, and have its join handle destruct into that data so the tests can inspect them?  It would be a lot more maintainable.

stacks-signer/src/runloop.rs
jcnelson: > I can't find a zero-cost alternative to circumvent the fact that abstracting the last match into a method requires a &mut self such that stacks_client can increment the slot_version. The code above works because self is captured by the closure, and the Iterator is collected in the enclosing scope. As far as I know, it does not work with methods the same way as with closures. Taking that much effort to put the last match behind a method is not worth it. This logic can't be used anywhere else, and the match looks irreducible beyond moving it to its own method.

It is absolutely worth the time to make the code maintainable.  It is very straightforward to create a function that consumes the vector of messages, performs a "pass" on them to eliminate the pings, and returns the filtered vector.

testnet/stacks-node/src/tests/signer.rs
CAGS295: I'm carrying it over from `test_stackerdb_dkg.` I'll give it a try.

stacks-signer/src/ping/periodic_pinger.rs
CAGS295: In my experience, simple toggles don't require extra consistency guarantees. I'll change it anyway.

testnet/stacks-node/src/tests/signer.rs
CAGS295: I asked around, and the 'pox-4' file is nowhere to be found in `develop.` That may be a no atm.

jcnelson: Correct; it's in `next`

stacks-signer/src/stacks_client.rs
```diff
@@ -31,7 +48,9 @@ const BACKOFF_MAX_INTERVAL: u64 = 16384;
 /// Temporary placeholder for the number of slots allocated to a stacker-db writer. This will be retrieved from the stacker-db instance in the future
 /// See: https://github.com/stacks-network/stacks-blockchain/issues/3921
 /// Is equal to the number of message types
-pub const SLOTS_PER_USER: u32 = 10;
+pub const SLOTS_PER_USER: u32 = 11;
```
jcnelson, line 51: Please keep this at 10.  Find a slot to re-use for pinging.  Perhaps slot 3?

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -329,3 +333,105 @@ fn test_stackerdb_dkg() {
...
+        }
+    }
+
+    for signer in running_signers {
+        assert!(signer.stop().is_none());
+    }
+
+    for rx in ping_rxs {
+        rx.recv().unwrap();
+        assert_eq!(rx.recv(), Err(RecvError));
```
jcnelson, line 435: I'm not seeing in this test where the ping/pong happens?

CAGS295: [Here](https://github.com/stacks-network/stacks-core/pull/4167/files#diff-a2df5fcad27701f7a726d41b6fa85dfaa3382377e1a1fbb6f03e58f11047dbb3R378) a Ping command is sent to each signer's runloop. This means there will be a single RTT per signer in the Receiver<(u64,Duration)> end. We assert that there is exactly one RTT per signer as expected. The Receiver would normally be consumed upon instantiating a PingerThread, but that is covered already in another test.

jcnelson:
One last thing - I just saw that this is based against `develop`.  It should be based against `next` since that's where Nakamoto development is happening.

testnet/stacks-node/src/tests/signer.rs
jcnelson: Thanks! :pray:

jcnelson:
Otherwise LGTM.  I'll approve once it's based on `next`.

wileyj:
@CAGS295 can you resolve the conflicts here?

PR #4174, opened 2023-12-14
Enable `FIRE` coordinator for `stacks-signer`

### LINKED ISSUE #4169: Enable `FIRE` coordinator for `signer`
Currently, we use the original `FROST` coordinator from `wsts`.  This is a very basic coordinator implementation, it requires that everyone takes part in `dkg` and `sign` rounds, and will simply hang if anyone doesn't respond.  

Current `wsts` also has an advanced `FIRE` coordinator, which will timeout and retry with various strategies for the various operations.  It will also return errors if it can no longer make progress, and these errors will specify which signers were responsible for the inability to continue.

We should enable the `FIRE` coordinator for `stacks-signer`, configure it as desired, and handle the errors it returns.

There are other meta protocols which are even more advanced than `FIRE`, specifically `ROAST` which does the same things but operates in parallel with much shorter timeouts.  But it would require a significant rewrite of the state machine, and probably stackerdb as well.  The `FIRE` coordinator is already implemented and well tested, and did not require significant alteration of the state machine.

xoloki:
### Description
`stacks-signer` is currently using a very basic `wsts` `Coordinator` implementation that doesn't handle non-responsive or malicious signers.  This has been fine for testing so far, but for Nakamoto we will need to sign all blocks using `wsts`, so we have to handle these and other errors.

`wsts` has a better `Coordinator` which supports the `FIRE` meta-protocol; it uses timeouts to handle non-responsive signers, returns errors identifying malicious signers, and correctly identifies when the number of malicious and/or non-responsive signers is large enough to make progress on any round impossible.

This PR therefore enables the `FIRE` coordinator.

### Applicable issues

- fixes #4169 

### Checklist

- [ ] Changelog is updated

### DIFF
--- a/stacks-signer/src/config.rs
+++ b/stacks-signer/src/config.rs
@@ -119,6 +119,14 @@ pub struct Config {
     pub signer_id: u32,
     /// The time to wait for a response from the stacker-db instance
     pub event_timeout: Duration,
+    /// timeout to gather DkgPublicShares messages
+    pub dkg_public_timeout: Option<Duration>,
+    /// timeout to gather DkgEnd messages
+    pub dkg_end_timeout: Option<Duration>,
+    /// timeout to gather nonces
+    pub nonce_timeout: Option<Duration>,
+    /// timeout to gather signature shares
+    pub sign_timeout: Option<Duration>,
 }
 
 /// Internal struct for loading up the config file signer data
@@ -290,6 +298,10 @@ impl TryFrom<RawConfigFile> for Config {
             signer_id: raw_data.signer_id,
             signer_key_ids,
             event_timeout,
+            dkg_end_timeout: None,
+            dkg_public_timeout: None,
+            nonce_timeout: None,
+            sign_timeout: None,
         })
     }
 }
--- a/stacks-signer/src/main.rs
+++ b/stacks-signer/src/main.rs
@@ -52,7 +52,7 @@ use stacks_signer::runloop::{RunLoop, RunLoopCommand};
 use stacks_signer::utils::{build_signer_config_tomls, build_stackerdb_contract};
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::{fmt, EnvFilter};
-use wsts::state_machine::coordinator::fire::Coordinator as FrostCoordinator;
+use wsts::state_machine::coordinator::fire::Coordinator as FireCoordinator;
 use wsts::state_machine::OperationResult;
 use wsts::v2;
 
@@ -88,11 +88,11 @@ fn spawn_running_signer(path: &PathBuf) -> SpawnedSigner {
     let (cmd_send, cmd_recv) = channel();
     let (res_send, res_recv) = channel();
     let ev = StackerDBEventReceiver::new(vec![config.stackerdb_contract_id.clone()]);
-    let runloop: RunLoop<FrostCoordinator<v2::Aggregator>> = RunLoop::from(&config);
+    let runloop: RunLoop<FireCoordinator<v2::Aggregator>> = RunLoop::from(&config);
     let mut signer: Signer<
         RunLoopCommand,
         Vec<OperationResult>,
-        RunLoop<FrostCoordinator<v2::Aggregator>>,
+        RunLoop<FireCoordinator<v2::Aggregator>>,
         StackerDBEventReceiver,
     > = Signer::new(runloop, ev, cmd_recv, res_send);
     let endpoint = config.endpoint;
--- a/stacks-signer/src/main.rs
+++ b/stacks-signer/src/main.rs
@@ -52,7 +52,7 @@ use stacks_signer::runloop::{RunLoop, RunLoopCommand};
 use stacks_signer::utils::{build_signer_config_tomls, build_stackerdb_contract};
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::{fmt, EnvFilter};
-use wsts::state_machine::coordinator::frost::Coordinator as FrostCoordinator;
+use wsts::state_machine::coordinator::fire::Coordinator as FrostCoordinator;
 use wsts::state_machine::OperationResult;
 use wsts::v2;
 
--- a/stacks-signer/src/main.rs
+++ b/stacks-signer/src/main.rs
@@ -124,8 +124,11 @@ fn process_dkg_result(dkg_res: &[OperationResult]) {
                 &schnorr_proof.r, &schnorr_proof.s,
             );
         }
-        OperationResult::DkgError(..) | OperationResult::SignError(..) => {
-            todo!()
+        OperationResult::DkgError(dkg_error) => {
+            panic!("Received DkgError {}", dkg_error);
+        }
+        OperationResult::SignError(sign_error) => {
+            panic!("Received SignError {}", sign_error);
         }
     }
 }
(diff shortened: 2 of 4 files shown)


stacks-signer/src/main.rs
```diff
@@ -52,7 +52,7 @@ use stacks_signer::runloop::{RunLoop, RunLoopCommand};
 use stacks_signer::utils::{build_signer_config_tomls, build_stackerdb_contract};
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::{fmt, EnvFilter};
-use wsts::state_machine::coordinator::frost::Coordinator as FrostCoordinator;
+use wsts::state_machine::coordinator::fire::Coordinator as FrostCoordinator;
```
jferrant, line 55: NIT: Should probably call this FireCoordinator instead of FrostCoordinator

stacks-signer/src/runloop.rs
```diff
@@ -1,15 +1,15 @@
...
 
-use backoff::default;
 use libsigner::{SignerRunLoop, StackerDBChunksEvent};
 use slog::{slog_debug, slog_error, slog_info, slog_warn};
 use stacks_common::{debug, error, info, warn};
 use wsts::common::MerkleRoot;
 use wsts::curve::ecdsa;
 use wsts::net::{Message, Packet, Signable};
-use wsts::state_machine::coordinator::frost::Coordinator as FrostCoordinator;
+use wsts::state_machine::coordinator::fire::Coordinator as FrostCoordinator;
```
jferrant, line 12: Similar NIT: should rename this as FireCoordinator

stacks-signer/src/main.rs
xoloki: Yeah, I was trying to keep the changes minimal, but it's def confusing this way.

I think I'll call it CoordinatorStateMachine or something else neutral, so we can just import whatever with that name.

stacks-signer/src/runloop.rs
```diff
@@ -234,12 +234,23 @@ impl From<&Config> for RunLoop<FrostCoordinator<v2::Aggregator>> {
             .iter()
             .map(|i| i - 1) // Signer::new (unlike Signer::from) doesn't do this
             .collect::<Vec<u32>>();
+        // signer uses a Vec<u32> for its key_ids, but coordinator uses a HashSet for each signer since it needs to do lots of lookups
+        let mut signer_key_ids = HashMap::new();
+        for (signer_id, key_ids) in &config.signer_key_ids {
+            let ids = key_ids.iter().map(|i| *i - 1).collect::<HashSet<u32>>();
```
jferrant, line 240: Won't this be off by one? We already do -1 above in key_ids. This essentially will be doing signer_key_ids - 2 now, no? Might be misunderstanding the logic.

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -11,7 +11,7 @@ use stacks_signer::runloop::RunLoopCommand;
 use stacks_signer::utils::{build_signer_config_tomls, build_stackerdb_contract};
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::{fmt, EnvFilter};
-use wsts::state_machine::coordinator::frost::Coordinator as FrostCoordinator;
+use wsts::state_machine::coordinator::fire::Coordinator as FrostCoordinator;
```
jferrant, line 14: Same nit as before

stacks-signer/src/runloop.rs
xoloki: No, because `key_ids` here is an ephemeral iterator for this loop.  It's not the `key_ids` variable declared above.  Yay scoping!

I'll change the loop iterator name so it's not confusing.

stacks-signer/src/main.rs
xoloki: I went ahead and called it `FireCoordinator`

stacks-signer/src/runloop.rs
xoloki: Done

testnet/stacks-node/src/tests/signer.rs
xoloki: Fixed.

stacks-signer/src/runloop.rs
xoloki: I ended up doing a nested map, it's cleaner and that's the style the `fn` uses anyway.

jferrant (approved):

jcnelson (approved):

stacks-signer/src/config.rs
```diff
@@ -290,6 +298,10 @@ impl TryFrom<RawConfigFile> for Config {
             signer_id: raw_data.signer_id,
             signer_key_ids,
             event_timeout,
+            dkg_end_timeout: None,
```
wileyj, line 301: would this be better suited to have some default timeouts set?
what would happen if these config settings are not defined - i see them referenced in the runloop.rs file

stacks-signer/src/main.rs
```diff
@@ -124,8 +124,11 @@ fn process_dkg_result(dkg_res: &[OperationResult]) {
                 &schnorr_proof.r, &schnorr_proof.s,
             );
         }
-        OperationResult::DkgError(..) | OperationResult::SignError(..) => {
-            todo!()
+        OperationResult::DkgError(dkg_error) => {
+            panic!("Received DkgError {}", dkg_error);
```
wileyj, line 128: could this be a DOS vector? is it possible to craft a scenario to trigger this panic?

stacks-signer/src/main.rs
```diff
@@ -124,8 +124,11 @@ fn process_dkg_result(dkg_res: &[OperationResult]) {
...
+        }
+        OperationResult::SignError(sign_error) => {
+            panic!("Received SignError {}", sign_error);
```
wileyj, line 131: same comment as line 128

stacks-signer/src/main.rs
```diff
@@ -150,8 +153,11 @@ fn process_sign_result(sign_res: &[OperationResult]) {
                 &schnorr_proof.r, &schnorr_proof.s,
             );
         }
-        OperationResult::DkgError(..) | OperationResult::SignError(..) => {
-            todo!()
+        OperationResult::DkgError(dkg_error) => {
+            panic!("Received DkgError {}", dkg_error);
```
wileyj, line 157: same as line 128

stacks-signer/src/main.rs
```diff
@@ -150,8 +153,11 @@ fn process_sign_result(sign_res: &[OperationResult]) {
...
+        }
+        OperationResult::SignError(sign_error) => {
+            panic!("Received SignError {}", sign_error);
```
wileyj, line 160: same as line 128

stacks-signer/src/config.rs
xoloki: I think the only reasonable default is to not have timeouts.  We will likely have a lively debate over what timeouts we should use in prod as we approach release.  

Without timeouts, everything works properly, but users can block dkg or sign rounds forever if they are malicious.  So it's fine for testing to not have timeouts, but we need them for real world usage.

stacks-signer/src/main.rs
xoloki: All unexpected responses `panic` the current `stacks-signer` binary.  This will need to be addressed at some point, but is orthogonal to this PR.

stacks-signer/src/main.rs
xoloki: Yeah current `stacks-signer` doesn't really handle errors, or even keep running indefinitely.

stacks-signer/src/main.rs
wileyj: got, thanks!

stacks-signer/src/config.rs
wileyj: understood, thanks!

wileyj (approved):
shipit

PR #4175, opened 2023-12-14
Merged in latest next

obycode:
This is not tested with clarinet yet. @hugocaillard if it works, then we can go ahead and merge this.


hugocaillard (approved):
Thanks @obycode!

PR #3806, opened 2023-07-25
feat: Mutation Testing

fpbgg:
### Description

This is my first swing at introducing mutation tests to our GitHub Actions workflows.

Mutation testing is a methodology where you mutate your source code and measure how many mutants "survive" your test suites. This provides a useful measurement for the quality of your test suite.

To run these tests locally:

1. `cargo install cargo-mutagen`
2. `cargo mutagen`

(Currently there's a missing file error, but the test cases are run and evaluated in `target/mutagen/mutations` so I'm assuming it's a coverage file issue.)

### Checklist
- [x] Test coverage for new or modified code paths
- [ ] Changelog is updated
- [ ] Required documentation changes (e.g., `docs/rpc/openapi.yaml` and `rpc-endpoints.md` for v2 endpoints, `event-dispatcher.md` for new events)
- [ ] New clarity functions have corresponding PR in `clarity-benchmarking` repo
- [ ] New integration test(s) added to `bitcoin-tests.yml`

Issue #4176, opened 2023-12-15
Mutation testing - Check on PR tests passing


ASuciuX:
`filter-pr` which has the role on the `**PR**` to run the mutants for the functions highlighted by the `git diff` ( functions modified or created ). 
- If the functions don't have missing, timeout or unviable tests, the workflow will pass. 
- Otherwise it will fail and display which are the missing, timeout and unviable functions.  In the case that a specific function shouldn't be tested, it can be highlighted so by specifying in its header #[mutants::skip] or other attributes containing mutants::skip (e.g. #[cfg_attr(test, mutants::skip)]).

PR #4178, opened 2023-12-15
Mutation testing - pr differences

ASuciuX:
## Applicable issues
- fixes #3806 
- https://github.com/stacks-network/stacks-core/issues/4176

## This has to be merged before in order to have the references right. Please review this as well
- https://github.com/stacks-network/actions/pull/4

### Solutions and Recommended Usage

[Cargo-mutant does an incremental build for each mutation](https://github.com/sourcefrog/cargo-mutants/wiki/Compared#:~:text=Mutagen%20builds%20the%20tree%20only%20once%3B%20cargo%2Dmutants%20does%20an%20incremental%20build%20for%20each%20mutation.), that’s why this solution comes with the implementation of tracking the output and only updating it with the differences, to run 10-20, maybe a few 100s mutants instead of 20,000.

**CI YML File:** Are linked with the actions in the actions' repo https://github.com/stacks-network/actions/pull/3.

Should be executed with every PR, as it tests only the updated or newly added functions in the PR's commits.

### How to read the status from the `pr-differences` workflow
If it is not failing, it means all the functions are tested properly.
If it is failing it will display what type of tests are the problem: missing tests, timeout tests, or unviable tests. For each section, it will specify the functions, the files they are in and their line numbers.

### Handling Exit Codes ([docs](https://github.com/sourcefrog/cargo-mutants/blob/main/book/src/exit-codes.md))
Mutation testing produces exit codes post-completion. 
In the `pr-differences` workflow they are caught and specified with extra details about what should be done based on them.


wileyj:
moved to draft so @ASuciuX et al can work through some issues discovered in initial tests

ASuciuX:
The `Tracking PR Mutants / Incremental Mutants Testing (pull_request)` is failing on this because the reference is not valid ( the actions PR has to be merged first)

ASuciuX:
Overall Flow Diagram https://github.com/stacks-network/stacks-core/discussions/4210#discussioncomment-8041025

.github/workflows/pr-differences-mutants.yml
```diff
@@ -0,0 +1,363 @@
...
+        with:
+          fetch-depth: 0
+
+      - run: cargo install --version 23.12.2 cargo-mutants
+
+      - name: Relative diff
+        run: |
+          git diff origin/${{ github.base_ref }}.. > git.diff
+
+      - name: Remove deleted file's lines from git.diff file
```
wileyj, line 42: For the names of the steps, i'd recommend keeping them short and generic since the UI is harder to read with longer names. 
ex: 
```
- name: Update git diff
```

wileyj: the same comment applies to other steps - KISS

wileyj:
I'm also considering that you may want to run this workflow *only* after an approval - otherwise, it will currently trigger on any commit to the branch being PR'ed. 
I'm ambivalent if that's the right way to go here - but most tests currently do not run until a PR is approved (only a small sample of workflows are run on commits, with goal being faster iteration. if mutant tests are run on every commit, it may slow down that objective).

ASuciuX:
> I'm also considering that you may want to run this workflow _only_ after an approval - otherwise, it will currently trigger on any commit to the branch being PR'ed. I'm ambivalent if that's the right way to go here - but most tests currently do not run until a PR is approved (only a small sample of workflows are run on commits, with goal being faster iteration. if mutant tests are run on every commit, it may slow down that objective).

The goal here was to run the mutants before the PR's approval, since the workflow is a reference to which of the added/modified functions fail when running mutants, so the developers working on the PR can check and fix them and the others can approve the PR based on the fact that there are no mutants unchecked.

wileyj:
> > I'm also considering that you may want to run this workflow _only_ after an approval - otherwise, it will currently trigger on any commit to the branch being PR'ed. I'm ambivalent if that's the right way to go here - but most tests currently do not run until a PR is approved (only a small sample of workflows are run on commits, with goal being faster iteration. if mutant tests are run on every commit, it may slow down that objective).
> 
> The goal here was to run the mutants before the PR's approval, since the workflow is a reference to which of the added/modified functions fail when running mutants, so the developers working on the PR can check and fix them and the others can approve the PR based on the fact that there are no mutants unchecked.

i see. i guess this is probably fine, as long as the diff step works as intended so it's not re-running this workflow unnecessarily. 



the other comment i have here is that the workflow is quite verbose - it would be ideal to move most of the workflow into a composite action.

ASuciuX:
> > > I'm also considering that you may want to run this workflow _only_ after an approval - otherwise, it will currently trigger on any commit to the branch being PR'ed. I'm ambivalent if that's the right way to go here - but most tests currently do not run until a PR is approved (only a small sample of workflows are run on commits, with goal being faster iteration. if mutant tests are run on every commit, it may slow down that objective).
> > 
> > 
> > The goal here was to run the mutants before the PR's approval, since the workflow is a reference to which of the added/modified functions fail when running mutants, so the developers working on the PR can check and fix them and the others can approve the PR based on the fact that there are no mutants unchecked.
> 
> i see. i guess this is probably fine, as long as the diff step works as intended so it's not re-running this workflow unnecessarily.

I'm not certain about what you mean by that, currently it checks the differences between the base branch and the latest commit from the source branch and runs mutants for all the differences It runs it every time, as if a new commit would be pushed to the source branch before the latest run finishes, it would stop it and no output would be available for what was till that point. If the workflow is finalised and there are missed mutants, it would flag them for the current commit, but if they are fixed in a future commit it marks them as caught then. Also, in this manner, there is only one standard final output that can be checked by anyone to see which mutants were created for that specific PR.

> the other comment i have here is that the workflow is quite verbose - it would be ideal to move most of the workflow into a composite action.

I've moved pretty much everything that could be moved to composite, what is left is necessary for the way it currently runs. Maybe the output step can be moved in a composite action, in that case, would printing to "$GITHUB_STEP_SUMMARY" still work the way it currently does?

wileyj:
> > > > I'm also considering that you may want to run this workflow _only_ after an approval - otherwise, it will currently trigger on any commit to the branch being PR'ed. I'm ambivalent if that's the right way to go here - but most tests currently do not run until a PR is approved (only a small sample of workflows are run on commits, with goal being faster iteration. if mutant tests are run on every commit, it may slow down that objective).
> > > 
> > > 
> > > The goal here was to run the mutants before the PR's approval, since the workflow is a reference to which of the added/modified functions fail when running mutants, so the developers working on the PR can check and fix them and the others can approve the PR based on the fact that there are no mutants unchecked.
> > 
> > 
> > i see. i guess this is probably fine, as long as the diff step works as intended so it's not re-running this workflow unnecessarily.
> 
> I'm not certain about what you mean by that, currently it checks the differences between the base branch and the latest commit from the source branch and runs mutants for all the differences It runs it every time, as if a new commit would be pushed to the source branch before the latest run finishes, it would stop it and no output would be available for what was till that point. If the workflow is finalised and there are missed mutants, it would flag them for the current commit, but if they are fixed in a future commit it marks them as caught then. Also, in this manner, there is only one standard final output that can be checked by anyone to see which mutants were created for that specific PR.
> 
> > the other comment i have here is that the workflow is quite verbose - it would be ideal to move most of the workflow into a composite action.
> 
> I've moved pretty much everything that could be moved to composite, what is left is necessary for the way it currently runs. Maybe the output step can be moved in a composite action, in that case, would printing to "$GITHUB_STEP_SUMMARY" still work the way it currently does?

yes, `GITHUB_STEP_SUMMARY` would still work  - what i'm interested in is if the step logic you have here can all be moved to a composite action, and the workflow here would simply run a git checkout and then run the composite action. 
keeping it as-is isn't incorrect, but creates a larger barrier to resolve issues or make updates.

ASuciuX:
> > > > > I'm also considering that you may want to run this workflow _only_ after an approval - otherwise, it will currently trigger on any commit to the branch being PR'ed. I'm ambivalent if that's the right way to go here - but most tests currently do not run until a PR is approved (only a small sample of workflows are run on commits, with goal being faster iteration. if mutant tests are run on every commit, it may slow down that objective).
> > > > 
> > > > 
> > > > The goal here was to run the mutants before the PR's approval, since the workflow is a reference to which of the added/modified functions fail when running mutants, so the developers working on the PR can check and fix them and the others can approve the PR based on the fact that there are no mutants unchecked.
> > > 
> > > 
> > > i see. i guess this is probably fine, as long as the diff step works as intended so it's not re-running this workflow unnecessarily.
> > 
> > 
> > I'm not certain about what you mean by that, currently it checks the differences between the base branch and the latest commit from the source branch and runs mutants for all the differences It runs it every time, as if a new commit would be pushed to the source branch before the latest run finishes, it would stop it and no output would be available for what was till that point. If the workflow is finalised and there are missed mutants, it would flag them for the current commit, but if they are fixed in a future commit it marks them as caught then. Also, in this manner, there is only one standard final output that can be checked by anyone to see which mutants were created for that specific PR.
> > > the other comment i have here is that the workflow is quite verbose - it would be ideal to move most of the workflow into a composite action.
> > 
> > 
> > I've moved pretty much everything that could be moved to composite, what is left is necessary for the way it currently runs. Maybe the output step can be moved in a composite action, in that case, would printing to "$GITHUB_STEP_SUMMARY" still work the way it currently does?
> 
> yes, `GITHUB_STEP_SUMMARY` would still work - what i'm interested in is if the step logic you have here can all be moved to a composite action, and the workflow here would simply run a git checkout and then run the composite action. keeping it as-is isn't incorrect, but creates a larger barrier to resolve issues or make updates.

Technically only a job out of the 4 calling the composite action could be moved - one without strategy matrix - but it would require additional checks in the composite. The other 2 jobs have different matrix dimensions, so there's really no way to get rid of them, since at least 1 job would run without matrix, one would run with 4 and the other with 8 shards. The job with the verification that's done with bash scripts before them is made in order to know which one of the jobs to run, and the only thing that is left to be moved to composite is the output job, which would decrease the lines count by ~35%, but we would still have over 200. If that's the case, do you think I should still move the output job to a composite action, or leave it here?

wileyj:
@ASuciuX it's entirely doable - 
the conditionals can remain (mostly, maybe small changes would be needed) - you'd simply have to have an output in the composite action to report a status.  

there is a LOT of shell in this workflow. moving that to a different repo would make fixing/updating it much easier. 

looking it through, i don't think it's an easy task (mainly due to the checks/outputs you have) but i think it's worthwhile nonetheless for the reason i stated above. it would require some care in the composite action (mainly around inputs/outputs), but again - i feel it's worth the effort and will make maintaining this workflow much easier.

wileyj:
this is interesting @ASuciuX : https://github.com/stacks-network/stacks-core/actions/runs/7547829177/job/20548598881?pr=4178#step:2:1323


even more peculiar is that it tried to download a musl package to satisfy:
```
info: downloading https://github.com/cargo-bins/cargo-binstall/releases/download/v1.4.4/cargo-binstall-x86_64-unknown-linux-musl.tgz
info: verifying sha256 checksum for cargo-binstall-x86_64-unknown-linux-musl.tgz
info: cargo-binstall installed at /home/runner/.cargo/bin/cargo-binstall
```

ASuciuX:
> this is interesting @ASuciuX : https://github.com/stacks-network/stacks-core/actions/runs/7547829177/job/20548598881?pr=4178#step:2:1323
> 
> even more peculiar is that it tried to download a musl package to satisfy:
> 
> ```
> info: downloading https://github.com/cargo-bins/cargo-binstall/releases/download/v1.4.4/cargo-binstall-x86_64-unknown-linux-musl.tgz
> info: verifying sha256 checksum for cargo-binstall-x86_64-unknown-linux-musl.tgz
> info: cargo-binstall installed at /home/runner/.cargo/bin/cargo-binstall
> ```

I've reviewed past worflow executions and noticed that this issue has been consistently present in all of them. I believe it will work alright, but i could change it to https://github.com/cargo-bins/cargo-binstall?tab=readme-ov-file#in-github-actions. The interesting part is that binstall and also the [mutants docs](https://mutants.rs/ci.html)  recommend using the one I've added.

wileyj:
> > this is interesting @ASuciuX : https://github.com/stacks-network/stacks-core/actions/runs/7547829177/job/20548598881?pr=4178#step:2:1323
> > even more peculiar is that it tried to download a musl package to satisfy:
> > ```
> > info: downloading https://github.com/cargo-bins/cargo-binstall/releases/download/v1.4.4/cargo-binstall-x86_64-unknown-linux-musl.tgz
> > info: verifying sha256 checksum for cargo-binstall-x86_64-unknown-linux-musl.tgz
> > info: cargo-binstall installed at /home/runner/.cargo/bin/cargo-binstall
> > ```
> 
> I've reviewed past worflow executions and noticed that this issue has been consistently present in all of them. I believe it will work alright, but i could change it to https://github.com/cargo-bins/cargo-binstall?tab=readme-ov-file#in-github-actions. The interesting part is that binstall and also the [mutants docs](https://mutants.rs/ci.html) recommend using the one I've added.

my concern is mostly that the workflow *failed* and nothing else ran as a result:
```
thread 'main' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/patch-0.7.0/src/parser.rs:84:5:
bug: failed to parse entire input. Remaining: 'diff --git a/stacks-common/src/libcommon.rs b/stacks-common/src/lib.rs
```

do you know why that error was triggered?

ASuciuX:
@wileyj thanks for pointing it. just fixed it https://github.com/stacks-network/actions/pull/14

it was related to: renaming `.rs` files & having no mutants found

wileyj:
@ASuciuX the last question i have here is timing. if i understand correctly, the idea is that this workflow will run on every PR. 
The current action has been running for over 2 hours: https://github.com/stacks-network/stacks-core/actions/runs/7547829177?pr=4178

can you share some timings that we can expect based on different types of PR's?
for example, when we merge to master for a release - the CI in total takes around 45-60 minutes. adding another 2+ hours to that time may not be desirable.

ASuciuX:
I've identified and resolved an issue with the git diff command used in our workflow, which you can review [here](https://github.com/stacks-network/actions/commit/29c7b860bcca66d68acf7486ed8ad5b033ac0e47). The workflow run details are available [here](https://github.com/stacks-network/stacks-core/actions/runs/7547829177/job/20645462735).

Previously, the command was erroneously capturing changes not only from the source branch but also from the base branch, starting from the initiation of the PR up to the present. This resulted in an extended runtime of about [3 hours](https://github.com/stacks-network/stacks-core/actions/runs/7547829177/attempts/3) for the mutants process, encompassing all differences on the `develop` branch from December 15 to January 18.
image
All the shards took less than 1 hour and 10 minutes, except 2 shards:
- shard-big-0 took 2h 44m because 1h 50m was [the timeout mutant](https://github.com/stacks-network/stacks-core/actions/runs/7547829177/job/20619985424#step:2:1739) 
- shard-big-2 took 3h 4m because 1h 48m was [the timeout mutant](https://github.com/stacks-network/stacks-core/actions/runs/7547829177/job/20619986800#step:2:1738) 

An enhancement that could significantly optimize our runtime involves dynamically terminating mutant processes based on their initial test durations. For instance, if the initial test for stacks-node takes X minutes, we could set a threshold to automatically stop the process after 1.5X minutes and call that a `timeout mutant`. Similarly, for stackslib, if the initial test takes Y minutes, the process would halt after 1.5Y minutes. Unfortunately, such a dynamic stopping mechanism isn't currently implemented in cargo mutants.



---

For context, in previous runs:

- Small packages typically completed in under 30 minutes, aided by the use of shards.
- Large packages like stackslib and stacks-node initially required about 20-25 minutes for build and test processes.
   - Each "missed" and "caught" mutant took approximately 15 minutes. Using shards, this meant about 50-55 minutes for processing around 32 mutants (10-16 functions modified). Every additional 8 mutants added another 15 minutes to the runtime.
   - "Unviable" mutants, which are functions lacking a `Default` implementation for their returned struct type, took less than a minute each.
   - "Timeout" mutants typically required more time. However, these should be marked to be skipped (by adding a `skip` flag to their header) since they indicate functions unable to proceed in their test workflow with mutated values, as opposed to the original implementations.
   
---

With the fix applied, the workflow should now correctly process only the relevant changes, leading to more efficient runtimes.

ASuciuX:
On top of the above specified run times, the integration of `nextest` in `cargo-mutants` is expected to significantly reduce the time needed for processing caught mutants. For more information, see [this discussion](https://github.com/sourcefrog/cargo-mutants/pull/223#issuecomment-1883997255).

wileyj:
> For context, in previous runs:
> 
> * Small packages typically completed in under 30 minutes, aided by the use of shards.
> * Large packages like stackslib and stacks-node initially required about 20-25 minutes for build and test processes.
>   
>   * Each "missed" and "caught" mutant took approximately 15 minutes. Using shards, this meant about 50-55 minutes for processing around 32 mutants (10-16 functions modified). Every additional 8 mutants added another 15 minutes to the runtime.
>   * "Unviable" mutants, which are functions lacking a `Default` implementation for their returned struct type, took less than a minute each.
>   * "Timeout" mutants typically required more time. However, these should be marked to be skipped (by adding a `skip` flag to their header) since they indicate functions unable to proceed in their test workflow with mutated values, as opposed to the original implementations.

this is great! 
can you add this comment :
```
- Small packages typically completed in under 30 minutes, aided by the use of shards.
- Large packages like stackslib and stacks-node initially required about 20-25 minutes for build and test processes.
    - Each "missed" and "caught" mutant took approximately 15 minutes. Using shards, this meant about 50-55 minutes for processing around 32 mutants (10-16 functions modified). Every additional 8 mutants added another 15 minutes to the runtime.
    - "Unviable" mutants, which are functions lacking a Default implementation for their returned struct type, took less than a minute each.
    - "Timeout" mutants typically required more time. However, these should be marked to be skipped (by adding a skip flag to their header) since they indicate functions unable to proceed in their test workflow with mutated values, as opposed to the original implementations.
```
to the file [./docs/ci-release.md](./docs/ci-release.md)? ideally under a section regarding the mutant workflow

wileyj (changes requested):
https://github.com/stacks-network/stacks-core/pull/4178#issuecomment-1900926390

i think one this documentation comment is addressed, everything else looks good

wileyj (approved):
shipit

docs/ci-release.md
```diff
@@ -228,4 +228,81 @@ ex: Branch is named `develop` and the PR is numbered `113`
...
+- The function has side effects or performance characteristics that are hard to test.
+- You've decided that the function is not important to test.
+
+To mark functions as skipped, so they are not mutated:
+
+- Add a Cargo dependency of the [mutants](https://crates.io/crates/mutants) crate, version `0.0.3` or later (this must be a regular `dependency`, not a `dev-dependency`, because the annotation will be on non-test code) and mark functions with `#[mutants::skip]`, or
+
+- You can avoid adding the dependency by using the slightly longer `#[cfg_attr(test, mutants::skip)]`.
+
+**Example:**
```
jbencin, line 276: Minor style nit: Use Markdown headers (`###...`) instead of bolding

docs/ci-release.md
```diff
@@ -228,4 +228,81 @@ ex: Branch is named `develop` and the PR is numbered `113`
   - `stacks-core:2.1.0.0.0`
   - `stacks-core:latest`
 
+## Mutation Testing
+
+When a new Pull Request (PR) is submitted, this feature evaluates the quality of the tests added or modified in the PR. It checks the new and altered functions through mutation testing. Mutation testing involves making small changes (mutations) to the code to check if the tests can detect these changes. The mutations are run with or without a [Github Actions matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs). The matrix is used when there is a large number of mutations to run.
```
jbencin, line 233: > It checks the new and altered functions through mutation testing

Could you explain the basics of *how* it does this, or link to a doc in the actions repo?

docs/ci-release.md
```diff
@@ -228,4 +228,81 @@ ex: Branch is named `develop` and the PR is numbered `113`
...
+}
+
+mod test {
+    #[test]
+    fn controlled_loop_terminates() {
+        super::controlled_loop()
+    }
+}
+```
+
```
jbencin, line 307: Minor style nit: I recommend writing Markdown using one sentence per line. It will be rendered the same (you need two newlines for a paragraph break), but this makes diffs and merges easier

.github/workflows/pr-differences-mutants.yml
```diff
@@ -0,0 +1,140 @@
...
+          shard: ${{ matrix.shard }}
+          package-dimension: "big"
+
+  # Output the mutants and fail the workflow if there are missed/timeout/unviable mutants
+  output-mutants:
+    name: Output Mutants
+
+    runs-on: ubuntu-latest
+
+    if: always()
```
jbencin, line 123: Does this line do anything?

ASuciuX: removed it

docs/ci-release.md
ASuciuX: updated with this format

jbencin (approved):

Issue #4177, opened 2023-12-15
Mutation testing - Logger for missing/timeout/unviable tests


ASuciuX:
`logger` which has the role of keeping track of all the functions that weren't tested before implementing this. Its final role is to be removed by having complete completion, but till that point, it makes it a lot easier to see which functions need extra tests and add them. On `**merge**`, it will run cargo mutants in the same way as the filter-pr and will modify the logged output with the information from the new generated output. (eg. fn A had missing tests before, now it has the tests. it will be removed from missed.txt and appended to caught.txt)

PR #4179, opened 2023-12-15
Test/mutants logger

ASuciuX:
## Applicable issues
- fixes #3806 
- #4177 

### Solutions and Recommended Usage

[Cargo-mutant does an incremental build for each mutation](https://github.com/sourcefrog/cargo-mutants/wiki/Compared#:~:text=Mutagen%20builds%20the%20tree%20only%20once%3B%20cargo%2Dmutants%20does%20an%20incremental%20build%20for%20each%20mutation.), that’s why this solution comes with the implementation of tracking the output and only updating it with the differences, to run 10-20 mutants instead of 4000.

**CI YML File:** Are linked with the actions in the actions' repo https://github.com/stacks-network/actions/pull/3.

Should be executed with every PR, as it tests only the updated or newly added functions in the PR's commits.

### How to read the status from the `filter-pr` workflow
If it is not failing, it means all the functions are tested properly.
If it is failing it will display what type of tests are the problem: missing tests, timeout tests, or unviable tests. For each section, it will specify the functions, the files they are in and their line numbers.

### Handling Exit Codes ([docs](https://github.com/sourcefrog/cargo-mutants/blob/main/book/src/exit-codes.md))
Mutation testing produces exit codes post-completion. 
In the `filter-pr` workflow they are caught and specified with extra details about what should be done based on them. 
In the `logger` workflow they are updating the place they belong to.

### Checklist

- [x] Fix mutation testing on packages to run on the supposed functions, not just on some
- [x] PR CI checker to have all functions tested 
  - missed/unviable/timeout outputs result in an error and stop the PR from being merged
  - outputs the functions that have to be tested
- [x] On merge on main/develop/next branch override (by appending) current output with the merged one
- [x] Mutation Testing Coverage for the already existing functions
  - [ ] Mutation testing triggering on git diff that didn't affect the functions, but affected the tests. (if the function already exists and is not modified and the tests that use it are created/updated, the functions will not be included in the cargo mutants)
- [ ] Mutants initial sync completed for `develop` and `next` branches


obycode:
Feedback from PR meeting - let's move to draft for now.

obycode:
Related discussion in Nakamoto slack channel.

.github/workflows/logger-mutants.yml
```diff
@@ -0,0 +1,30 @@
...
+      - next
+
+jobs:
+  save_cache:
+    runs-on: ubuntu-latest
+
+    # test
+    steps:
+      - name: Run logging mutants from actions
+        uses: stacks-network/actions/mutation-testing/logger@feat/mutation-testing
```
wileyj, line 28: this workflow doesn't seem to exist on this branch (or in main). is that something that needs to be re-added?

jbencin: I'd recommend merging the `feat/mutation-testing` branch into stacks-network/actions first, so you don't have to make another PR changing/removing the branch name later

jbencin: Also, I can't find the `logger` action either

wileyj: > Also, I can't find the `logger` action either

i expect this PR will be changing quite a bit as we've just started testing this. i'm converting this to draft for now

ASuciuX: yes, i will be updating it these days, wanted to have the `PR-differences` fully finished asap. this one will have to wait till we have all the mutants from the general sync before being merged

PR #4181, opened 2023-12-15
Update CODE_OF_CONDUCT.md

cuevasm:
Adding secret code to the code of conduct (simple text change only) so users in our community onboarding flows and contests can find it. This will bring some additional attention the repo 🙏

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -71,6 +71,9 @@ Community leaders will follow these Community Impact Guidelines in determining t
 
 **Consequence**: A permanent ban from any sort of public interaction within the community.
 
+### Secret Code: 
+The code to the contest is: BITCOINL2
+
 ## Attribution
 
 This Code of Conduct is adapted from the [Contributor Covenant][homepage], version 2.0,


CODE_OF_CONDUCT.md
```diff
@@ -71,6 +71,9 @@ Community leaders will follow these Community Impact Guidelines in determining t
 
 **Consequence**: A permanent ban from any sort of public interaction within the community.
 
+### Secret Code: 
+The code to the contest is: BITCOINL2
```
wileyj, line 75: it's a little unclear what `the contest` is. can you add some more text so it's clear what this is for?

cuevasm:
Some additional context. It's a text change to add a small reference in the Code of Conduct, such that folks registering for our events can find this code to enter into a contest we're running perpetually. The code proves they at least came to the repo since Github doesn't have a direct way to verify this. They get points in this contest for 'starring' the repo. More stars is good as VCs watch this and it matters in some developer communities. You can see the contest page here: https://bitcoinunleashed.org/community-prize

CODE_OF_CONDUCT.md
cuevasm: https://github.com/stacks-network/stacks-core/pull/4181#issuecomment-1858334139

diwakergupta:
@cuevasm is this a time boxed change that will be reverted later, or something you anticipate keeping for some time?

Either way, rather than stuff this in the code of conduct (where it feels a bit out of place IMO), I'd recommend adding it as a comment at the end of something like `.gitignore` instead.

cuevasm:
The code of conduct is preferred, the point of this campaign is to guide and walk people through joining the community - an ideal outcome is them becoming a contributor. I intentionally chose the Code of Conduct because it is relevant to new people _and_ relatively easy to find, even for non-technical people. I plan to leave this for as long as these campaigns to drive attention and stars to the repo are effective. It could be a month, it could stay there indefinitely. 

I'll just say this with love as well, I'm feeling a bit like there are too many smart people spending too much time worrying about a very small text change that has little to no downside and good upside. I came in thinking this should be a pretty simple yes so we can add this to the experiments we're running trying to build up more stacks visibility ahead of launches.

wileyj:
> @cuevasm is this a time boxed change that will be reverted later, or something you anticipate keeping for some time?
> 
> Either way, rather than stuff this in the code of conduct (where it feels a bit out of place IMO), I'd recommend adding it as a comment at the end of something like `.gitignore` instead.

@cuevasm following @diwakergupta thinking, would be it more appropriate to add your change to this file vs code of conduct?
https://github.com/stacks-network/stacks-core/blob/cuevasm-patch-2/docs/community.md

the change seems community related

diwakergupta:
> I'll just say this with love as well, I'm feeling a bit like there are too many smart people spending too much time worrying about a very small text change

Hahha, yes that's fair 😅 All I can say is, I didn't spend too much time on this :) I'll see myself out, onwards!

saralab (approved):

jcnelson (approved):

wileyj (approved):
lgtm

AshtonStephens (approved):

PR #4186, opened 2023-12-18
Process coordinator messages to duplicate state between multiple coor…

jferrant:
### Description

Coordinator state should be replicated between signers so that every signer can submit its aggregate key and also so it can pick up where other coordinators leave off if said coordinators go down.

### Applicable issues

- Closes #https://github.com/stacks-network/sbtc/issues/393

### DIFF
--- a/stacks-signer/src/runloop.rs
+++ b/stacks-signer/src/runloop.rs
@@ -170,7 +170,7 @@ impl<C: Coordinator> RunLoop<C> {
         event: &StackerDBChunksEvent,
     ) -> (Vec<Packet>, Vec<OperationResult>) {
         // Determine the current coordinator id and public key for verification
-        let (coordinator_id, coordinator_public_key) =
+        let (_coordinator_id, coordinator_public_key) =
             calculate_coordinator(&self.signing_round.public_keys);
         // Filter out invalid messages
         let inbound_messages: Vec<Packet> = event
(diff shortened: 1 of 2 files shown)


stacks-signer/src/runloop.rs
```diff
@@ -190,14 +190,11 @@ impl<C: Coordinator> RunLoop<C> {
...
-                .process_inbound_messages(&inbound_messages)
-                .unwrap_or_default()
-        } else {
-            (vec![], vec![])
-        };
+        // Next process the message as the coordinator
+        let (messages, results) = self
+            .coordinator
+            .process_inbound_messages(&inbound_messages)
+            .unwrap_or_default();
```
kantai, line 197: What does `unwrap_or_default()` do here? Does this mean if there's an error processing the inbound messages, the runloop just ignores the inbound messages? Should this be logged?

kantai (commented):
This LGTM, just a question about the error handling of `process_inbound_messages`

stacks-signer/src/runloop.rs
jferrant: It will ignore the error and return an empty vector. In general we never want the coordinator failing to process a message to crash and should continue processing the next event. We could later try to special case handle errors or I could add a print statement, but so far there has been no desire/need. If you want it logged, I would open another PR for this just to get this through. XD

kantai: Yeah -- I think there's logging in most places that would yield an error here, but I think it'd be good to add a catch all log there.

jferrant: Done :)

kantai (approved):

xoloki (approved):
LGTM, just one question about inbound processing but this PR isn't responsible for the behavior in question so it's not a blocker.

stacks-signer/src/runloop.rs
```diff
@@ -189,15 +189,18 @@ impl<C: Coordinator> RunLoop<C> {
...
-        };
+            .unwrap_or_else(|e| {
+                error!("Failed to process inbound messages as a signer: {e}");
+                vec![]
+            });
+        // Next process the message as the coordinator
+        let (messages, results) = self
+            .coordinator
+            .process_inbound_messages(&inbound_messages)
+            .unwrap_or_else(|e| {
```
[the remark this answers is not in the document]
jferrant: I think its the currently best behaviour. For now we don't want messages received out of order for example to crash the system. I.e. if a rogue signer starts sending valid but inappropriate responses, the coordinator state could error. I think we might need to introduce some sort of recovery. But for now, I think its better than crashing outright.

PR #4187, opened 2023-12-18
Nakamoto-Node: Support for interim block mining

kantai:
### Description

This PR adds support to the nakamoto node for interim block mining (i.e., blocks within the current tenure) with a configurable wait time between such blocks. This PR also refactors the `chainstate::nakamoto::miner` module to match the actual usage of the miner interfaces, and moving the `TestPeer` specific method out into `TestPeer` itself.

This adds a new `nakamoto_integration` test which mines 10 nakamoto tenures, each with 2 blocks in the tenure.

### DIFF
--- a/stackslib/src/chainstate/nakamoto/miner.rs
+++ b/stackslib/src/chainstate/nakamoto/miner.rs
@@ -53,8 +53,8 @@ use crate::chainstate::stacks::db::transactions::{
     handle_clarity_runtime_error, ClarityRuntimeTxError,
 };
 use crate::chainstate::stacks::db::{
-    ChainstateTx, ClarityTx, MinerRewardInfo, StacksChainState, StacksHeaderInfo,
-    MINER_REWARD_MATURITY,
+    ChainstateTx, ClarityTx, MinerRewardInfo, StacksBlockHeaderTypes, StacksChainState,
+    StacksHeaderInfo, MINER_REWARD_MATURITY,
 };
 use crate::chainstate::stacks::events::{StacksTransactionEvent, StacksTransactionReceipt};
 use crate::chainstate::stacks::miner::{
@@ -99,22 +99,18 @@ impl NakamotoTenureInfo {
 }
 
 pub struct NakamotoBlockBuilder {
-    /// if this is building atop an epoch 2 block, then this is that block's header
-    epoch2_parent_header: Option<(StacksBlockHeader, ConsensusHash)>,
-    /// if this is building atop an epoch 3 block, then this is that block's header
-    nakamoto_parent_header: Option<NakamotoBlockHeader>,
+    /// If there's a parent (i.e., not a genesis), this is Some(parent_header)    
+    parent_header: Option<StacksHeaderInfo>,
     /// Signed coinbase tx, if starting a new tenure
     coinbase_tx: Option<StacksTransaction>,
     /// Tenure change tx, if starting or extending a tenure
     tenure_tx: Option<StacksTransaction>,
     /// Total burn this block represents
     total_burn: u64,
-    /// parent block-commit hash value
-    parent_commit_hash_value: BlockHeaderHash,
     /// Matured miner rewards to process, if any.
     matured_miner_rewards_opt: Option<MaturedMinerRewards>,
     /// bytes of space consumed so far
-    bytes_so_far: u64,
+    pub bytes_so_far: u64,
     /// transactions selected
     txs: Vec<StacksTransaction>,
     /// header we're filling in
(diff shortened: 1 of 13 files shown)


testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -490,3 +495,175 @@ fn simple_neon_integration() {
...
+
+#[test]
+#[ignore]
+/// This test spins up a nakamoto-neon node.
+/// It starts in Epoch 2.0, mines with `neon_node` to Epoch 3.0, and then switches
+///  to Nakamoto operation (activating pox-4 by submitting a stack-stx tx). The BootLoop
+///  struct handles the epoch-2/3 tear-down and spin-up.
+/// This test makes three assertions:
+///  * 10 tenures are mined after 3.0 starts
+///  * Each tenure has 2 blocks
```
jcnelson, line 507: Can you boost this to something like 10 blocks?

testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -490,3 +495,175 @@ fn simple_neon_integration() {
...
+        let tip = NakamotoChainState::get_canonical_block_header(chainstate.db(), &sortdb)
+            .unwrap()
+            .unwrap();
+
+        let mut mempool = naka_conf
+            .connect_mempool_db()
+            .expect("Database failure opening mempool");
+
+        mempool
+            .submit_raw(
```
jcnelson, line 626: Can we avoid `submit_raw` now that mempool admission is working?

jcnelson (approved):
LGTM, provided that all tests pass.  Thanks for the `NakamotoBlockBuilder` cleanup!

kantai:
The code coverage drop seems to be due to tests which aren't run on PRs, but are run on the base branch (e.g., the epoch tests). The patch coverage is 90%.

stackslib/src/chainstate/nakamoto/miner.rs
```diff
@@ -330,74 +227,36 @@ impl NakamotoBlockBuilder {
...
-                &StacksBlockId::new(&parent_tenure_id_consensus_hash, &parent_header_hash),
-            ) {
+        let parent_block_id = StacksBlockId::new(&parent_consensus_hash, &parent_header_hash);
+        let parent_coinbase_height =
+            NakamotoChainState::get_coinbase_height(chainstate.db(), &parent_block_id)
+                .ok()
+                .flatten()
+                .unwrap_or(0);
+
+        let new_tenure = cause == Some(TenureChangeCause::BlockFound);
```
jferrant, line 253: This is a total nitpick, but is there a reason to assign cause == Some(Tenure..) to a var?

kantai: I think it makes the if statement below easier to read. Probably just a personal preference for me, but I find binary comparisons inside of rust's ternary-like structures kind of hard to read:

```rust
let coinbase_height = if cause == Some(TenureChangeCause::BlockFound) {
   ...
}
```

jferrant: I usually just rely on comments in this instance but it was more to make sure I wasn't missing something. So sounds good :D

testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -490,3 +495,166 @@ fn simple_neon_integration() {
...
+        "Latest tip";
+        "height" => tip.stacks_block_height,
+        "is_nakamoto" => tip.anchored_header.as_stacks_nakamoto().is_some(),
+    );
+
+    assert!(tip.anchored_header.as_stacks_nakamoto().is_some());
+    assert_eq!(
+        tip.stacks_block_height,
+        block_height_pre_3_0 + ((inter_blocks_per_tenure + 1) * tenure_count),
+        "Should have mined (1 + interim_blocks_per_tenur) * tenure_count nakamoto blocks"
```
jferrant, line 650: nit. Typo. "interim_blocks_per_tenur"

jferrant (approved):

testnet/stacks-node/src/tests/nakamoto_integrations.rs
kantai: Yep, updated in 3cef3b4.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
kantai: Fixed.

PR #4188, opened 2023-12-18
Feat/stackerdb miners contract

### LINKED ISSUE #3931: Nakamoto[Block Proposal]: Miner constructs BlockProposals and stores them in StackerDB
In mockamoto, the `stacks-node` miner produces blocks directly, however, in the argon milestone, the miner needs to submit those blocks for consideration by the signer.

In order to do this, the miner must produce `BlockProposal` structs and store them to the StackerDB.

The miner also needs to watch the network for their prior `BlockProposal`'s acceptance before producing another proposal.

### LINKED ISSUE #4018: [Nakamoto] `.miners` boot contract for StackerDB-assisted block propagation to signers
Miners will need to propagate blocks to Stackers for signing.  To do this, there would be a dedicated boot contract called `.miners` which implements the StackerDB trait.  The miners would write their blocks to this StackerDB, and the Stacks node itself would update the contract state as part of processing a new burnchain block.

The StackerDB in `.miners` would have two slots: the current miner, and the parent (previous) miner.  The principals for these miners are derived from their `LeaderKeyRegisterOp`s' public key hashes (see SIP-021).  When the node processes a sortition, it allocates the new miner's `principal` a StackerDB slot by evicting the slot of the now-grandparent miner.  The parent miner retains their slot as-is.  In the event of no sortition, then the new miner's `principal` would simply be set to the parent miner's `principal`.

The parent miner and new miner write their blocks to their respective slots, and Stackers receive them by subscribing to the `.miners` StackerDB via their nodes' event listeners.  Stackers continue to sign the parent miner's blocks until they can produce a `TenureChange` transaction, at which point they cease to sign the parent miner's blocks and will only sign the new miner's blocks.

jcnelson:
Implements #4018, with the caveats listed in #4018 (namely, the `.miners` `StackerDBConfig` structure must be synthesized directly from chainstate due to a chicken-and-egg problem of needing `.miners` to reflect the latest sortition state before any blocks have been processed).

The code here is mostly complete and has some minimal unit tests.  The following features need to be added:

- [x] The Nakamoto node needs to be upgraded to write its proposed block to the `.miners` contract.  There's already a helper method in `NakamotoBlockBuilder` that will generate the chunk.

The following test coverage needs to be added:

- [x] A `TestPeer`-based unit test to verify that the `StackerDBConfig` for `.miners` is correctly synthesized after processing each Bitcoin block
* See test_make_miners_stackerdb_config

- [x] A `TestPeer`-based unit test to verify that two nodes can be instantiated, which both track `.miners`, and a `StackerDBChunkData` post to one will show up on the other (i.e. this shows that StackerDB can propagate `NakamotoBlock`s for miners)
* See miner_writes_proposed_block_to_stackerdb. Note this test does not have two nodes tracking miners but simply uses a stackerdb session to check that the block is found as expected in the appropriate slot. Does not bother to test stacker db propogation which is tested in other tests already. It also it not TestPeer based as the propose nakamoto block logic is found only within the Relayer for the neon node itself so really should be an integration test.

- [x] A Nakamoto integration test to ensure that the miner propagates the (unsigned) `NakamotoBlock` to the `.miners` contract
* See above point. 
   * Optional (can be a separate PR): test that the stacks signer receives the `NakamotoBlock`: will be done in another PR once stacks signer begins to process proposed blocks
   
As discussed earlier, I'm handing this off to @jferrant so I can focus on building out the Nakamoto p2p layer.

Closes https://github.com/stacks-network/stacks-core/issues/4018
Closes https://github.com/stacks-network/stacks-core/issues/3931


stackslib/src/chainstate/nakamoto/miner.rs
```diff
@@ -498,6 +499,58 @@ impl NakamotoBlockBuilder {
...
         Ok((block, consumed, size))
     }
+
+    /// Make a StackerDB chunk message containing a proposed block.
+    /// Sign it with the miner's private key.
+    /// Automatically determine which StackerDB slot to use.
+    /// Returns Some(chunk) if the given key corresponds to one of the expected miner slots
+    /// Returns None if not
+    /// Returns an error on signing or DB error
+    pub fn make_stackerdb_block_proposal(
```
jferrant, line 509: Should this be creating a serialized NakamotoBlockProposal rather than just a serialized block?

jferrant: Got my answer! No, the miners should just propose a NakamotoBlock and the Stacks-Signer should fill in the remaining info for the "NakamotoBlockProposal" struct in the endpoint PR as their view of the chain is the "most valid" one (70% need to agree on their view of the chain state).

stackslib/src/net/chat.rs
```diff
@@ -1195,6 +1195,7 @@ impl ConversationP2P {
         network: &mut PeerNetwork,
         message: &mut StacksMessage,
         authenticated: bool,
+        ibd: bool,
```
kantai, line 1198: This seems like it should be either (1) some kind of global, or (2) a state variable of `PeerNetwork` or `ConversationP2P`, rather than a function argument.

stackslib/src/net/p2p.rs
```diff
@@ -5301,20 +5305,35 @@ impl PeerNetwork {
...
+                                warn!(
+                                    "Failed to load StackerDB config for {}: {:?}",
+                                    &stackerdb_contract_id, &e
+                                );
+                                StackerDBConfig::noop()
+                            }
+                        }
+                    };
                 if new_config != stackerdb_config && new_config.signers.len() > 0 {
                     if let Err(e) =
```
kantai, line 5338: In this error case, should the new config still be stored to `new_stackerdb_configs`?

stackslib/src/net/p2p.rs
```diff
@@ -5301,20 +5305,35 @@ impl PeerNetwork {
...
+                            Ok(config) => config,
+                            Err(e) => {
+                                warn!(
+                                    "Failed to load StackerDB config for {}: {:?}",
+                                    &stackerdb_contract_id, &e
+                                );
+                                StackerDBConfig::noop()
+                            }
+                        }
+                    };
```
kantai, line 5336:
```suggestion
                let new_config_res =
                    if stackerdb_contract_id == boot_code_id(MINERS_NAME, chainstate.mainnet) {
                        // .miners contract -- directly generate the config
                        NakamotoChainState::make_miners_stackerdb_config(sortdb)
                    } else {
                        // normal stackerdb contract
                        StackerDBConfig::from_smart_contract(
                            chainstate,
                            sortdb,
                            &stackerdb_contract_id,
                        )
                     };
                let new_config = new_config_res
                    .unwrap_or_else(|e| {
                                warn!(
                                    "Failed to load StackerDB config";
                                    "contract" => %stackerdb_contract_id,
                                    "err" => ?e,
                                );
                                StackerDBConfig::noop()
                     });
```

This suggestion changes the error behavior for the miners contract to use the noop db config. Is it necessary for `continue` to be used here? If so, why is it different in this regard than other stackerdb configs?

stackslib/src/chainstate/stacks/boot/miners.clar
```diff
@@ -0,0 +1,21 @@
...
+;; StackerDB-required method to get the DB configuration.
+;; The values here are ignored.
+(define-public (stackerdb-get-config)
+    (ok {
+        chunk-size: u0,
+        write-freq: u0,
+        max-writes: u0,
+        max-neighbors: u0,
+        hint-replicas: (list )
+    }))
```
kantai, line 21: Why is it necessary for this contract to exist at all?

stackslib/src/chainstate/nakamoto/tests/mod.rs
```diff
@@ -1636,3 +1643,272 @@ pub fn test_get_highest_nakamoto_tenure() {
...
+    debug!("miners = {:#?}", &miner_hash160s);
+
+    // extract chainstate and sortdb -- we don't need the peer anymore
+    let chainstate = &mut peer.stacks_node.as_mut().unwrap().chainstate;
+    let sort_db = peer.sortdb.as_mut().unwrap();
+    let mut last_snapshot = SortitionDB::get_canonical_burn_chain_tip(sort_db.conn()).unwrap();
+
+    // make two leader keys (miner 1 and miner 2)
+    let mut miners = vec![];
+    for i in 3..13 {
```
kantai, line 1685: This seems to create 10 miners, not 2

stackslib/src/chainstate/nakamoto/tests/mod.rs
```diff
@@ -1636,3 +1643,272 @@ pub fn test_get_highest_nakamoto_tenure() {
...
+#[test]
+fn test_make_miners_stackerdb_config() {
+    let test_signers = TestSigners::default();
+    let mut peer = boot_nakamoto(
+        function_name!(),
+        vec![],
+        test_signers.aggregate_public_key.clone(),
+    );
+
+    let miner_keys: Vec<_> = (3..13).map(|_| StacksPrivateKey::new()).collect();
```
kantai, line 1659: Why is `3..13` being used? That seems like a very specific range, but I don't understand why.

stackslib/src/chainstate/nakamoto/tests/mod.rs
```diff
@@ -1636,3 +1643,272 @@ pub fn test_get_highest_nakamoto_tenure() {
...
+        miners.push(miner);
+    }
+
+    let mut stackerdb_configs = vec![];
+    let mut stackerdb_chunks = vec![];
+
+    // synthesize some sortitions and corresponding winning block-commits
+    for i in 3..13 {
+        // no winner every 3rd sortition
+        let sortition = i > 3 && i % 3 != 0;
```
kantai, line 1706: I don't understand these indexes. Why not `0..10`?

stackslib/src/chainstate/nakamoto/tests/mod.rs
```diff
@@ -1636,3 +1643,272 @@ pub fn test_get_highest_nakamoto_tenure() {
...
+            )
+            .unwrap();
+            tx.commit().unwrap();
+        }
+
+        last_snapshot = SortitionDB::get_block_snapshot(sort_db.conn(), &snapshot.sortition_id)
+            .unwrap()
+            .unwrap();
+
+        if i > 3 {
```
kantai, line 1814: What happens if `make_miners_stackerdb_config()` is invoked at `i == 3` ?

stackslib/src/chainstate/nakamoto/tests/mod.rs
```diff
@@ -1636,3 +1643,272 @@ pub fn test_get_highest_nakamoto_tenure() {
...
+                assert_eq!(chunk.slot_version, (i + 1) as u32);
+                assert_eq!(chunk.data, block.serialize_to_vec());
+                stackerdb_chunks.push(chunk);
+            } else {
+                assert!(NakamotoBlockBuilder::make_stackerdb_block_proposal(
+                    &sort_db,
+                    i as u32,
+                    &block,
+                    &miner_keys[i - 3]
+                )
```
kantai, line 1860: Should the prior two miners still be able to create block proposals?

stackslib/src/chainstate/nakamoto/tests/mod.rs
```diff
@@ -1636,3 +1643,272 @@ pub fn test_get_highest_nakamoto_tenure() {
...
+                signer_signature: ThresholdSignature::mock(),
+            };
+            let block = NakamotoBlock {
+                header,
+                txs: vec![],
+            };
+            if sortition {
+                let chunk = NakamotoBlockBuilder::make_stackerdb_block_proposal(
+                    &sort_db,
+                    i as u32,
```
kantai, line 1845: Why is `write_count = i`?

stackslib/src/chainstate/nakamoto/tests/mod.rs
```diff
@@ -1636,3 +1643,272 @@ pub fn test_get_highest_nakamoto_tenure() {
...
+            // have block-commit
+            // check the stackerdb config as of this chain tip
+            let stackerdb_config =
+                NakamotoChainState::make_miners_stackerdb_config(sort_db).unwrap();
+            eprintln!(
+                "stackerdb_config at i = {} (sorition? {}): {:?}",
+                &i, sortition, &stackerdb_config
+            );
+
+            stackerdb_configs.push(stackerdb_config);
```
kantai, line 1824: Should these configs be checked at the end of this test?

kantai (commented):
This mostly LGTM, just a few comments and questions.

stackslib/src/chainstate/nakamoto/miner.rs
```diff
@@ -498,6 +499,58 @@ impl NakamotoBlockBuilder {
...
+        sortdb: &SortitionDB,
+        write_count: u32,
```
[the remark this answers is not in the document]
jferrant: Write count is the version of the chunk being written (must sequentially increase). i have updated this though in my commit such that it gets it from the database itself.

stackslib/src/chainstate/stacks/boot/miners.clar
jferrant: I think it will werror out when "subscribing" to the stacker db instance if there is no contract found, but I can see if that is actually the case.

stackslib/src/net/p2p.rs
jferrant: I actually wonder if this should error if it fails to make a MINERS stackerdb config...If we continue, we will have a more nebulous error down the line due to the missing .miners stacker db no? Otherwise, I think using a noop is a better option as well to just continuing.

stackslib/src/chainstate/stacks/boot/miners.clar
jcnelson: The contract has to exist so the RPC endpoints will work.  Normally (i.e. for every other stacker DB), these methods are only evaluated every Bitcoin block in order to generate a (node-side) StackerDB configuration.  In the specific case of `.miners`, the node generates that configuration on its own.

kantai: Does it make more sense to special case the RPC endpoints for `.miners`? The config lookups are already treated as special cases, so its not like this really avoids the need for special casing, and having a contract around that returns data that doesn't correspond to the actual behavior of the StackerDB seems confusing if it can be avoided.

stackslib/src/net/chat.rs
jcnelson: It gets passed in from the node implementation, since it is able to determine whether or not there are more blocks to process based on burnchain state.  I'd prefer to leave it this way.

stackslib/src/chainstate/stacks/boot/miners.clar
jcnelson: We already expect users to understand that any contracts instantiated with the boot address behave differently than expected, so I don't think it's unreasonable to expect the same understanding for `.miners`.  Also, there are currently four StackerDB-specific endpoints, and some others (e.g. /v2/neighbors) that convey StackerDB data.  I'd rather not have to duplicate them all just for `.miners`.

stackslib/src/net/p2p.rs
jcnelson: Ah, this is an oversight. If we fail to generate the `.miners` StackerDB config, then we should use the previous one instead of simply refusing to synthesize it.  Then at least the chain can continue.

This error path is reachable only from database errors in the SortitionDB, however.  If it fails, it may be best to just shut down the node since it's unclear how those errors could ever be recoverable.

stackslib/src/chainstate/nakamoto/tests/mod.rs
jcnelson: The comment is the bug.  There ought to be 10 miners.

stackslib/src/chainstate/nakamoto/tests/mod.rs
jcnelson: At some point there was special behavior for `i < 3`, but it no longer appears to be there.  It's fine to change this all to `0..10`, and remove any `i - 3`'s in the loops that follow (as well as any other place where `i` is compared to `3`).

stackslib/src/chainstate/stacks/boot/miners.clar
kantai: Why would you need to duplicate the whole endpoints? What fails in those endpoints if the contract doesn't exist? It's true that boot address contracts do behave somewhat differently than normal contracts, but none of them actually return incorrect data (as this one would).

stackslib/src/chainstate/nakamoto/tests/mod.rs
jcnelson: It would panic because there would be no block-commit associated with the first-ever snapshot (which is what the code would return when asked for the last snapshot with a sortition, when there are no associated block-commits).

stackslib/src/chainstate/nakamoto/tests/mod.rs
jcnelson: Why prior two?  I think we just want the prior

stackslib/src/chainstate/nakamoto/tests/mod.rs
jcnelson: Yes

stackslib/src/chainstate/nakamoto/tests/mod.rs
kantai: It seems like it would return an error of some kind from `get_last_snapshot_with_sortition` rather than panic. If that's the case, the test should check that error case.

stackslib/src/chainstate/stacks/boot/miners.clar
jcnelson: > Why would you need to duplicate the whole endpoints?

It sounded like that's what you're asking for?  A set of API endpoints specific to querying, storing, and getting the metadata for chunks of data written for `.miners`?

> What fails in those endpoints if the contract doesn't exist?

The StackerDB API endpoints query the `stackerdbs.sqlite` database directly; they don't query the Clarity VM.  Instead, the `PeerNetwork` will do so whenever it detects that the burnchain has advanced -- it will re-evaluate these two Clarity functions to synthesize a `StackerDBConfig`, and will then update the `stackerdbs.sqlite` database with it (this is documented in `src/stackslib/net/stackerdb/mod.rs`).

If the contract doesn't exist -- i.e. if `PeerNetwork` doesn't synthesize a `StackerDBConfig` struct for its StackerDB replica -- then there won't be any configuration data for them in `stackerdbs.sqlite`, and thus the endpoint will fail.

> but none of them actually return incorrect data (as this one would).

The RPC endpoints don't query the smart contract for any state (as described above, they're only involved in the `StackerDBConfig` generation), so this shouldn't be an issue.

stackslib/src/net/p2p.rs
jferrant: I chose to have it error as I think this would be due to sortition DB errors which would indicate a much bigger problem

stackslib/src/chainstate/nakamoto/tests/mod.rs
jferrant: I d obelieve this was to test the functionality in make_stackerdb_block_proposal was actually using a different version number each time based on the past in value. Its no longer relevant as I determine the version number on the fly now.

stackslib/src/clarity_vm/clarity.rs
```diff
@@ -1277,8 +1286,67 @@ impl<'a, 'b> ClarityBlockConnection<'a, 'b> {
...
+            let boot_code_auth = TransactionAuth::Standard(
+                TransactionSpendingCondition::Singlesig(SinglesigSpendingCondition {
+                    signer: boot_code_address.bytes.clone(),
+                    hash_mode: SinglesigHashMode::P2PKH,
+                    key_encoding: TransactionPublicKeyEncoding::Uncompressed,
+                    nonce: 0,
+                    tx_fee: 0,
+                    signature: MessageSignature::empty(),
+                }),
+            );
```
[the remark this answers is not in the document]
jferrant: yes it is! Will use boot_code_tx_auth()

stackslib/src/net/chat.rs
jferrant: Do we have a consensus? Is PeerNetwork created from the node implementation? Cause in that case, making if a state variable of PeerNetwork would be what I would lean towards, but I am happy to do whatever.

stackslib/src/chainstate/nakamoto/tests/mod.rs
jferrant: Will do!

stackslib/src/chainstate/nakamoto/tests/mod.rs
jferrant: Will fix!

stackslib/src/chainstate/nakamoto/tests/mod.rs
jferrant: Will update to 0..10 :)

stackslib/src/chainstate/nakamoto/tests/mod.rs
jferrant: Will do!

stackslib/src/chainstate/nakamoto/tests/mod.rs
jferrant: It doesn't panic. But it does generate a result that I am having trouble figuring out what it means XD the assigned miner private key for the stacker dbs writers is not one I can seem to figure out where it comes from

stackslib/src/net/chat.rs
jcnelson: It needs to be an argument to `PeerNetwork::run()`, since `ibd` is computed from state that lives exclusively in the node implementation.  Please leave as-is.

jferrant (approved):

testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -142,6 +146,9 @@ pub fn naka_neon_integration_conf(seed: Option<&[u8]>) -> (Config, StacksAddress
 
     conf.node.miner = true;
     conf.node.wait_time_for_microblocks = 500;
+    conf.node
+        .stacker_dbs
+        .push(boot_code_id(MINERS_NAME, conf.is_mainnet()));
```
jcnelson, line 151: The constructor for `Config` needs to ensure that the `MINERS_NAME` contract is present in the list of StackerDBs if the node is mining.  It can't be opted out.  Same will go for `.stackers`.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -914,3 +921,118 @@ fn correct_burn_outs() {
...
+        miners_stackerdb
+            .get_latest_chunk(0)
+            .expect("Failed to get latest chunk from the miner slot ID")
+            .expect("No chunk found")
+    })
+    .join()
+    .expect("Failed to join chunk handle");
+    // We should now successfully deserialize a chunk
+    let _block = NakamotoBlock::consensus_deserialize(&mut &chunk[..])
+        .expect("Failed to deserialize chunk into block");
```
jcnelson, line 1028: Is there any way you can verify that it's the same block mined?  E.g. does it have the expected transactions and so on?

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -115,6 +129,7 @@ impl BlockMinerThread {
             burn_block,
             event_dispatcher: rt.event_dispatcher.clone(),
             parent_tenure_id,
+            miners_stackerdb,
```
jcnelson, line 132: I'm not sure this needs to be a member of `BlockMinerThread`.  You're basically connecting to yourself -- you can afford to instantiate a `StackerDBSession` each time you mine a block, because there's no gain to be had by preserving the TCP window across multiple block productions.

stackslib/src/net/p2p.rs
jcnelson: Agreed.  Keep the old config on error.

stackslib/src/net/mod.rs
```diff
@@ -2135,6 +2135,25 @@ pub mod test {
             test_path
         }
 
+        /// Initialize the .miners StackerDB instance
+        pub fn init_stacker_db_miners(&mut self) {
```
jcnelson, line 2139: This shouldn't be necessary.  Instead, you can just call `peer.step_with_ibd(true)` in-between calls to `peer.next_burnchain_block()`. Then, the `.miners` contract DB state should initialize on its own.

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -84,8 +87,8 @@ fn advance_to_nakamoto(peer: &mut TestPeer) {
 
         peer.tenure_with_txs(&txs, &mut peer_nonce);
     }
-
     // peer is at the start of cycle 8
+    peer.init_stacker_db_miners();
```
jcnelson, line 91: Here, just call `peer.step_with_ibd(true)` after each call to `peer.tenure_with_txs(..)`.  That should initialize the `.miners` StackerDB.

jcnelson (commented):
This overall LGTM. Thanks for taking it over @jferrant!  Just a couple comments need to be addressed and then I'll approve.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
jferrant: kk will add!

testnet/stacks-node/src/tests/nakamoto_integrations.rs
jferrant: I am sure there is. Will see if I can add it in :D

testnet/stacks-node/src/nakamoto_node/miner.rs
jferrant: Alright will do :D

stackslib/src/net/mod.rs
jferrant: Ah kk ! I was wondering how ot enforce the stacker db miners being instantiated and this seemed the least janky option (though still janky). Will try with peer.step_with_ibd and see what happens!

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
jferrant: Hm it still breaks the make_stackerdb_config test. But perhaps I should just change that test D:

stackslib/src/net/mod.rs
jferrant: Actually I remember why I did this. Because it breaks test_make_miners_stackerdb_config as it attempts to build a block proposal which makes the config using the stacker db .miners contract which doesn't exist. Not sure how to fix this.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
jferrant: Done!

testnet/stacks-node/src/nakamoto_node/miner.rs
jferrant: Done.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
jferrant: I am not sure which contructor you are referencing? do you mean the default one or the one which is created from a ConfigFile? I feel this might be better ina. sep PR.

stackslib/src/chainstate/stacks/boot/miners.clar
kantai: > It sounded like that's what you're asking for? A set of API endpoints specific to querying, storing, and getting the metadata for chunks of data written for .miners?

I see -- no, that's not what I was suggesting. I was suggesting adding a check anywhere that would fail without the presence of a contract for `.miners`.

> The RPC endpoints don't query the smart contract for any state (as described above, they're only involved in the StackerDBConfig generation), so this shouldn't be an issue.

Yep, I see that now, which still leaves me wondering why the `.miners` contract needs to exist.

> If the contract doesn't exist -- i.e. if PeerNetwork doesn't synthesize a StackerDBConfig struct for its StackerDB replica -- then there won't be any configuration data for them in stackerdbs.sqlite, and thus the endpoint will fail.

But this is exactly what this PR _avoids_ for `.miners`, right? `create_or_reconfigure_stackerdbs` has a special case for `.miners`, and never actually invokes the Clarity VM. So, it doesn't seem to need the contract to exist either.

Applying this diff to the PR, for example, still leaves the end-to-end test passing -- meaning that the stackerdbs instance still initializes the miners stackerdb, still accepts the block proposal, and still returns it, all without `.miners` ever existing. If its actually the case that the contract doesn't need to exist for the stackerdb to work, I'd much prefer not instantiating it at all. As you say, the contract doesn't respond on those public functions with data that actually corresponds to anything.

```diff
diff --git a/stackslib/src/clarity_vm/clarity.rs b/stackslib/src/clarity_vm/clarity.rs
index 9a18decfd..8bbd1cb95 100644
--- a/stackslib/src/clarity_vm/clarity.rs
+++ b/stackslib/src/clarity_vm/clarity.rs
@@ -1288,28 +1288,6 @@ impl<'a, 'b> ClarityBlockConnection<'a, 'b> {
             let miners_contract_tx =
                 StacksTransaction::new(tx_version.clone(), boot_code_auth.clone(), payload);
 
-            let miners_initialization_receipt = self.as_transaction(|tx_conn| {
-                // initialize with a synthetic transaction
-                debug!("Instantiate {} contract", &miners_contract_id);
-                let receipt = StacksChainState::process_transaction_payload(
-                    tx_conn,
-                    &miners_contract_tx,
-                    &boot_code_account,
-                    ASTRules::PrecheckSize,
-                )
-                .expect("FATAL: Failed to process .miners contract initialization");
-                receipt
-            });
-
-            if miners_initialization_receipt.result != Value::okay_true()
-                || miners_initialization_receipt.post_condition_aborted
-            {
-                panic!(
-                    "FATAL: Failure processing .miners contract initialization: {:#?}",
-                    &miners_initialization_receipt
-                );
-            }
-
             /////////////////// .pox-4 ////////////////////////
             let first_block_height = self.burn_state_db.get_burn_start_height();
             let pox_prepare_length = self.burn_state_db.get_pox_prepare_length();
```

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -3047,6 +3050,122 @@ impl NakamotoChainState {
         Ok((epoch_receipt, clarity_commit))
     }
 
+    /// Create a StackerDB config for the .miners contract.
+    /// It has two slots -- one for the past two sortition winners.
+    pub fn make_miners_stackerdb_config(
+        sortdb: &SortitionDB,
+    ) -> Result<StackerDBConfig, ChainstateError> {
+        let tip = SortitionDB::get_canonical_burn_chain_tip(sortdb.conn())?;
```
jcnelson, line 3058: Can this be an argument, instead? This is called from the network main loop, so ideally we don't do any gratuitous I/O.

jferrant: Can do!

testnet/stacks-node/src/config.rs
```diff
@@ -2294,6 +2162,85 @@ pub struct NodeConfigFile {
...
+            pox_sync_sample_secs: self
+                .pox_sync_sample_secs
+                .unwrap_or(default_node_config.pox_sync_sample_secs),
+            use_test_genesis_chainstate: self.use_test_genesis_chainstate,
+            always_use_affirmation_maps: self
+                .always_use_affirmation_maps
+                .unwrap_or(default_node_config.always_use_affirmation_maps),
+            // miners should always try to mine, even if they don't have the anchored
+            // blocks in the canonical affirmation map. Followers, however, can stall.
+            require_affirmed_anchor_blocks: self.require_affirmed_anchor_blocks.unwrap_or(!miner),
```
jferrant, line 2223: This was actually a bugfix. It used to have a double negation.

testnet/stacks-node/src/config.rs
```diff
@@ -1209,20 +920,56 @@ impl Config {
...
+            }
+        } else if node.require_affirmed_anchor_blocks {
+            // testnet requires that we use the 2.05 rules for anchor block affirmations,
+            // because reward cycle 360 (and possibly future ones) has a different anchor
+            // block choice in 2.05 rules than in 2.1 rules.
+            debug!("Set `require_affirmed_anchor_blocks` to `false` for non-mainnet config");
+            node.require_affirmed_anchor_blocks = false;
+        }
+
+        let miners_contract_id = boot_code_id(MINERS_NAME, is_mainnet);
```
jferrant, line 938: I assumed it was better to forcibly add subscription to the stacker db miners contract rather than to error out. But let me know if this is not the case.

kantai (commented):
Overall, LGTM, I just have one unresolved comment about the `.miners` contract. It seems like it doesn't actually need to exist, and would prefer to just remove it in that case.

stackslib/src/chainstate/stacks/boot/miners.clar
jferrant: @jcnelson if nothing breaks to remove the .miners contract, I am happy to do so. I thought you mentioned that the RPC calls would fail though D:

jcnelson:
Well, since I'm the one who opened it, I can't approve it, but LGTM.  Thanks for carrying this over the finish line @jferrant!  Just a couple comments to address before merging :pray:

stackslib/src/chainstate/stacks/boot/miners.clar
jcnelson: I thought they would, but it turns out I was wrong -- the RPC endpoint only looks at the `stackerdbs.sqlite` DB to service queries, not the smart contract in the chainstate.  So as long as the `StackerDBConfig` is getting synthesized and applied, then the RPC endpoints should work.

kantai (approved):
LGTM, appreciate you for getting this across the line @jferrant!

PR #4205, opened 2023-12-20
Nakamoto Fixes: IssueBlockCommit and Reward Maturity Lookups

kantai:
### Description

This fixes two issues surfaced by the interim blocks mining in nakamoto-node.

1. The IssueBlock directive should use the tenure block _not_ the chain tip block for the directive. The interim-mining integration test has been updated to deal with this.
2. `calculate_matured_miner_rewards` should use the `coinbase_height`, not the block height. This was caught by running a long-running node with interim mining turned on, rather than an integration test. There's no integration test for this new behavior, but there probably should be. The issue surface from a panicking assertion because the 150th Stacks block had a coinbase height < 100, so the maturity lookup returned an empty list, triggering a panic. In testing mode, the maturity height is _2_, which means that it would be hard to surface this exact panic in a test.

### DIFF
--- a/stacks-signer/src/runloop.rs
+++ b/stacks-signer/src/runloop.rs
@@ -1,8 +1,8 @@
-use hashbrown::{HashMap, HashSet};
 use std::collections::VecDeque;
 use std::sync::mpsc::Sender;
 use std::time::Duration;
 
+use hashbrown::{HashMap, HashSet};
 use libsigner::{SignerRunLoop, StackerDBChunksEvent};
 use slog::{slog_debug, slog_error, slog_info, slog_warn};
 use stacks_common::{debug, error, info, warn};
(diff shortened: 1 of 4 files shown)


stacks-signer/src/runloop.rs
```diff
@@ -1,8 +1,8 @@
-use hashbrown::{HashMap, HashSet};
 use std::collections::VecDeque;
 use std::sync::mpsc::Sender;
 use std::time::Duration;
 
+use hashbrown::{HashMap, HashSet};
```
jcnelson, line 5: Is there a reason we're still using `hashbrown`?

jcnelson (approved):
Great catch @zone117x @kantai!

stacks-signer/src/runloop.rs
kantai: Unclear to me. Maybe open an issue for the signer? This changeline just came from running `cargo fmt-stacks`.

zone117x (approved):
I'm able to get a deployment running with this PR, but still running into a couple other block processing issues (details shared on Slack).

PR #4206, opened 2023-12-21
Fix: RBF-mechanism should check bitcoin RPC

kantai:
### Description

The stacks miner's bitcoin operation handling tracks submitted transactions via the `ongoing_op` mechanism. When deciding whether or not to attempt an RBF, it checks if the `ongoing_op` has been confirmed. Currently, it does this via the burnchain_db, however, this will not work if the transaction is confirmed, but it was unparseable. In this case, the miner will continuously attempt to RBF an already confirmed transaction, leading to bitcoin mempool rejections.

### Relevant Scenario

This scenario surfaced during the nakamoto "controlled testnet" testing through a pretty convoluted series of events:

1. The neon_node (required for booting through epochs 2.0-2.5) is set to mine very quickly (i.e., very low `wait_on_blocks` time)
2. The regtest bitcoind network is set to mine a new block whenever it sees a bitcoin tx in the mempool.
3. Not infrequently, the stacks miner will assemble and attempt to submit a block during bitcoin block 1 (usually this is a "resubmission" of an identical block commit), but when actually submitted, bitcoin block 2 has already arrived. This requires some pretty specific interlacing of events to occur. The `bitcoin_regtest_controller` normally performs a check to see if the `ongoing_op` matches the about-to-submit op, but since bitcoin block 2 has now been mined, ongoing_op is confirmed, so the 'resubmission' is submitted.
4. When the above scenario happens at a _reward cycle boundary_, the block commit _fails to parse_: instead of having PoX outputs, its a prepare phase single-burn-output commit.
5. The ongoing_op never appears in the burnchaindb even though it has been confirmed, so the miner tries to RBF it, but it gets rejected.

Actually fixing the timing issue in step 3 above isn't possible: no matter how late the `stacks-node` checks for a new bitcoin block before submitting the block commit, it's always possible that the bitcoin block arrives right "before" the commit is submitted (or equivalently, the bitcoin block just doesn't include the commit). But fixing the tracking of `ongoing_op` is possible, and that's what this PR does.

In theory, this could be opened against `develop` rather than `next`. I don't get a sense that this is a huge issue in practice, otherwise I'd expect there'd be a bug report already. In which case, this seems mostly relevant for the testing infrastructures for `next`, so I just want to simplify the workstream by opening this against `next` directly.

### DIFF
--- a/.github/workflows/bitcoin-tests.yml
+++ b/.github/workflows/bitcoin-tests.yml
@@ -68,6 +68,7 @@ jobs:
           - tests::neon_integrations::test_problematic_microblocks_are_not_relayed_or_stored
           - tests::neon_integrations::test_problematic_txs_are_not_stored
           - tests::neon_integrations::use_latest_tip_integration_test
+          - tests::neon_integrations::confirm_unparsed_ongoing_ops
           - tests::should_succeed_handling_malformed_and_valid_txs
           - tests::nakamoto_integrations::simple_neon_integration
           - tests::nakamoto_integrations::mine_multiple_per_tenure_integration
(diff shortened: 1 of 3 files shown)


jcnelson (approved):

obycode (approved):
This is a beautiful thing - finding this issue, fixing it, and coming up with a good way to test for it, all of it 🙌

PR #4212, opened 2023-12-26
Fix: commit burn out selection in nakamoto-node

kantai:
### Description

This PR fixes the "burn output" selection in the nakamoto-node to use `get_nakamoto_next_recipients()`. It also includes a refactor of the `get_reward_set()` invocations in nakamoto to pass in the cycle start height (as is done in 2.x) rather than "current burn height". This eliminates the need for the `+ 1` in the reward cycle number (commit eee3f20e78842abe0bb68eaf8fe60f70a3f49917)

* Includes a test with alternating stackers, this ensures that the reward set changes between each cycle. Improper handling of `get_next_recipients()` would cause a block-commit to be rejected (and the test to fail).


jcnelson (approved):

obycode (approved):
Looks great!

PR #4216, opened 2023-12-26
min tx count

jcnelson:
Make it so that the miner will only produce a block if it can meet a minimum number of txs on the _subsequent_ attempt to mine a block.

### DIFF
--- a/testnet/stacks-node/src/config.rs
+++ b/testnet/stacks-node/src/config.rs
@@ -900,6 +900,7 @@ impl Config {
                     .unprocessed_block_deadline_secs
                     .unwrap_or(miner_default_config.unprocessed_block_deadline_secs),
                 min_tx_count: miner.min_tx_count.unwrap_or(0),
+                only_increase_tx_count: miner.only_increase_tx_count.unwrap_or(false),
             },
             None => miner_default_config,
         };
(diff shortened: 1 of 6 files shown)


testnet/stacks-node/src/config.rs
```diff
@@ -1906,6 +1898,7 @@ impl MinerConfig {
             nonce_cache_size: 10_000,
             candidate_retry_cache_size: 10_000,
             unprocessed_block_deadline_secs: 30,
+            min_tx_count: 0,
```
AshtonStephens, line 1901: Let's do a > 0 default? It's not forcing behavior but highly encouraging it ~

jcnelson: No; let's not surprise users by causing their miners to stop working altogether if this value is too high

jcnelson: Broadly speaking, systemically trying to force miners to include more than X transactions in a block simply ensures that miners fill in their blocks with junk transactions by default to ensure that they can always mine (folks had the same proposition to "force" Bitcoin miners to mine non-empty blocks back in the mid-2010s, but it was never implemented for the same reason).

jcnelson:
This is superseded by #4262

PR #4223, opened 2023-12-31
Make all miner settings configurable at runtime

jcnelson:
This PR backports #4061 from `develop` to `master` and further makes every setting in `[miner]` and most miner-related settings in `[burnchain]` configurable at runtime. The config file will be reloaded whenever the relevant fields are accessed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
+## [Unreleased]
+
+- All miner-related settings can be changed in the config file and take effect
+  without a node restart (#4223)
+
 ## [2.4.0.0.4]
 
 This is a high-priority hotfix that addresses a bug in transaction processing which
--- a/testnet/stacks-node/src/config.rs
+++ b/testnet/stacks-node/src/config.rs
@@ -410,14 +410,24 @@ impl Config {
     }
 
     /// get the up-to-date miner options from the config
-    pub fn get_miner_config(&self) -> Result<MinerConfig, String> {
-        if let Some(path) = &self.config_path {
-            let config_file = ConfigFile::from_path(path.as_str())?;
-            let config = Config::from_config_file(config_file)?;
-            Ok(config.miner)
-        } else {
-            Ok(self.miner.clone())
-        }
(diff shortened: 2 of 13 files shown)


testnet/stacks-node/src/config.rs
```diff
@@ -394,6 +398,28 @@ lazy_static! {
...
+    /// get the up-to-date miner options from the config
+    pub fn get_miner_config(&self) -> Result<MinerConfig, String> {
+        if let Some(path) = &self.config_path {
+            let config_file = ConfigFile::from_path(path.as_str())?;
+            let config = Config::from_config_file(config_file)?;
+            Ok(config.miner)
+        } else {
+            Ok(self.miner.clone())
+        }
+    }
```
kantai, line 421: Looking at the invocations of `get_miner_config` (and, for that matter `get_burnchain_config`), it seems like every caller does the same defaulting behavior, so I think it makes sense to just do:

```suggestion
    /// re-read an up-to-date config from this config file's path, or if there is no path,
    /// just return a clone of the current config.
    fn reload_config(&self) -> Result<Config, String> {
        let Some(path) = &self.config_path else {
            return self.clone()
        };
        let config_file = ConfigFile::from_path(path.as_str())?;
        Config::from_config_file(config_file)
    }
    
    /// re-read the up-to-date miner options from the config file's path, or if there is 
    /// no path or it is unreadable, just return the current miner config.
    pub fn get_miner_config(&self) -> MinerConfig {
        let config = self.reload_config().unwrap_or_else(|e| {
             warn!("Failed to reload miner config: {e}");
             self.clone()
        });
        config.miner
    }
```

kantai (commented):
LGTM, just had a comment about some opportunities to reduce repetition.

But before merging into `master`, this needs a changelog entry.

obycode (approved):
lgtm!

jcnelson:
Oh fuck me `cargo fmt` just blew this diff up.  Lemme revert

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to the versioning scheme outlined in the [README.md](README.md).
 
+## [Unreleased]
```
kantai, line 8: This should be `[2.4.0.0.5]` on merging to `master`

jcnelson:
This is superseded by #4262

PR #4224, opened 2024-01-01
Fix fee estimate

obycode:


jcnelson (approved):

AshtonStephens:
While you're in here can you change the fee estimates to be for the fastest version and not the 30 minute version?

```python
# Fee estimation API URLS and their corresponding fee extraction functions.
FEE_ESTIMATIONS = [
    # Bitcoiner Live API
    (
        'https://bitcoiner.live/api/fees/estimates/latest',
        lambda response_json: response_json["estimates"]["30"]["sat_per_vbyte"],
    ),

    # Mempool Space API
    (
        'https://mempool.space/api/v1/fees/recommended',
        lambda response_json: response_json["fastestFee"],
    ),

    # Blockchain.info API
    (
        'https://api.blockchain.info/mempool/fees',
        lambda response_json: response_json["priority"],
    ),
]
```

obycode:
> While you're in here can you change the fee estimates to be for the fastest version and not the 30 minute version?

You're talking about the bitcoiner.live API, right? Looking at the output, it looks like 30 is the fastest option. It returns 30, 60, 120, 180, 360, 720, and 1440.

obycode:
> > While you're in here can you change the fee estimates to be for the fastest version and not the 30 minute version?
> 
> You're talking about the bitcoiner.live API, right? Looking at the output, it looks like 30 is the fastest option. It returns 30, 60, 120, 180, 360, 720, and 1440.

Oh, I see I can bump the confidence level up to 90% with `?confidence=0.9`. The value that returns for 30 minutes is the best we'll get I think.

AshtonStephens (approved):

PR #4226, opened 2024-01-02
Fix: nakamoto-node shouldn't respawn prometheus thread

### LINKED ISSUE #4214: Prometheus error: Address already in use
**Describe the bug**
When running [this commit](https://github.com/stacks-network/stacks-core/commit/c8db64c59ad9d95b53dbc37d529da4b43043fac3) from the `clarity-wasm-next` branch, the stacks node emits an error when booting using a Bitcoin node in regtest mode.

```
2023-12-26T10:20:50-05:00	Process abort due to thread panic: panicked at testnet/stacks-node/src/monitoring/prometheus.rs:13:14:
Prometheus monitoring: unable to bind address: Os { code: 98, kind: AddrInUse, message: "Address already in use" }
2023-12-26T10:20:50-05:00	Start serving prometheus metrics
```

The node doesn't crash from this error, and the Prometheus metrics still seem to be getting served on the configured port. So perhaps the Prometheus server is trying to start twice?

**Steps To Reproduce**
Run the `clarity-wasm-next` branch in a regtest environment. You can do so by cloning the [regtest project](https://github.com/hirosystems/stacks-regtest-env/tree/epoch3-interim-blocks) (epoch3-interim-blocks branch) and modifying [this variable](https://github.com/hirosystems/stacks-regtest-env/blob/epoch3-interim-blocks/docker-compose.yml#L4C26-L4C26) to the `clarity-wasm-next` branch.

kantai:
### Description

This PR:
1. Updates the prometheus thread to return errors to the JoinHandle rather than panicking in an unchecked async task. The PR doesn't change the behavior to _handle_ or _exit_ on the error, it retains the previous behavior of "basically just issue a warning but otherwise execute the node". 
2. Stores the thread's `JoinHandle` to the runloop struct.
3. Moves the `JoinHandle` from the neon runloop to the nakamoto runloop (if the `boot_nakamoto` loop needed to start the neon runloop). If the nakamoto runloop has a thread handle, it doesn't attempt to spawn a new prometheus thread.
4. Extends the `nakamoto_integrations::simple_neon` test to check the prometheus output before and after nakamoto block processing.

### Applicable issues

- fixes #4214

### DIFF
--- a/testnet/stacks-node/src/tests/nakamoto_integrations.rs
+++ b/testnet/stacks-node/src/tests/nakamoto_integrations.rs
@@ -370,6 +370,8 @@ fn simple_neon_integration() {
     }
 
     let (mut naka_conf, _miner_account) = naka_neon_integration_conf(None);
+    let prom_bind = format!("{}:{}", "127.0.0.1", 6000);
+    naka_conf.node.prometheus_bind = Some(prom_bind.clone());
     naka_conf.miner.wait_on_interim_blocks = Duration::from_secs(1000);
     let sender_sk = Secp256k1PrivateKey::new();
     // setup sender + recipient for a test stx transfer
@@ -436,6 +438,21 @@ fn simple_neon_integration() {
             .unwrap()
             .unwrap()
             .stacks_block_height;
+
+    // query for prometheus metrics
+    #[cfg(feature = "monitoring_prom")]
+    {
+        let prom_http_origin = format!("http://{}", prom_bind);
+        let client = reqwest::blocking::Client::new();
+        let res = client
+            .get(&prom_http_origin)
+            .send()
+            .unwrap()
+            .text()
+            .unwrap();
+        let expected_result = format!("stacks_node_stacks_tip_height {block_height_pre_3_0}");
+        assert!(res.contains(&expected_result));
+    }
 
     info!("Nakamoto miner started...");
     // first block wakes up the run loop, wait until a key registration has been submitted.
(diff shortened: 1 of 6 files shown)


testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -437,6 +439,21 @@ fn simple_neon_integration() {
             .unwrap()
             .stacks_block_height;
 
+    // query for prometheus metrics
+    #[cfg(feature = "monitoring_prom")]
```
jcnelson, line 443: For testing purposes, should we just always enable this flag?

kantai: You mean in the testing CI, or via the Cargo.toml?

The last time I looked into this (which admittedly, was over a year ago, and Cargo development can be comparatively speedy), you couldn't set default features for testing (as opposed to `dev` profile). So, my understanding is that we'd have to make `monitoring_prom` a default feature in `dev` if we wanted it enabled in tests.

jcnelson: > So, my understanding is that we'd have to make monitoring_prom a default feature in dev if we wanted it enabled in tests.

I think that's probably fine.  Could an alternative here be to use `#[cfg(any(feature = "monitoring_prom", feature = "testing"))]`?

kantai: > Could an alternative here be to use #[cfg(any(feature = "monitoring_prom", feature = "testing"))]?

I think that would need to be placed in the `stackslib` dependency, but it would still be a problem getting it to default "on" when run as a test: the `testing` feature would only default as on for the `stackslib` dependency, not the `stacks-node` binary.

Realistically, it looks like the only way to really get this to work without changing the feature flag set during testing is to add `monitoring_prom` to the `defaults`.

jcnelson: Okay, is there any way at all we can test this in CI that doesn't also require `monitoring_prom` to be set by default?

kantai: Yes, definitely, we just need to add the feature flag to this command: 

https://github.com/stacks-network/actions/blob/129b799256eb97eef7fc15f5679c1295bec4279a/stacks-core/cache/build-cache/action.yml#L71

jcnelson: Okay, can you do that as part of this PR?

kantai: Sure -- I opened a PR for it. https://github.com/stacks-network/actions/pull/6

jcnelson: Great, thanks!

jcnelson (approved):

Issue #4111, opened 2023-11-30
[Nakamoto] Stacker-Signer binary must handle non-participation


kantai:
Related to #4059 (and especially discussion there), the stacks network _must_ be capable of operating even with non-participation of some stackers. While the threshold for participation must clear the 70% requirement outlined in the SIPs, non-participating stackers cannot prevent DKG from completing. This requires updates to the stacker-signer binary such that it supports a protocol that accommodates this.

I **think** that solving this problem is a form of distributed census, meaning that it will require a BFT protocol. My view on this is that a solution should incorporate a standard BFT protocol, rather than trying to do something ad hoc. But I'm open to being convinced otherwise (if there's a clear argument about why this isn't a byzantine census).

jcnelson:
The nice things about DKG here are that we already have an _a priori_ signer census from the `.pox-4` contract.  Specifically, that means that going into DKG:

* We already know all of their signing keys with which they'll sign DKG messages.
* Each signer already has a one-to-many authenticated broadcast primitive (i.e. StackerDB slots).

These two aspects of the DKG protocol turn detecting Byzantine signers into a trivial, localized decision:  if you see equivocation from signer _S_, you (1) locally declare _S_ to be faulty and (2) post your proof-of-equivocation to your StackerDB slot so _everyone else_ sees the evidence that _S_ is faulty.  You can't forge _S_'s proof-of-equivocation without _S_'s signing key, so everyone else will reach the same localized decision that _S_ is faulty.  So, all honest nodes eventually detect that _S_ is faulty if _S_ equivocates to any honest node.

I think that means that it's sufficient to just perform multiple rounds of DKG if there are any faulty signers.  If a faulty signer is detected, honest nodes restart DKG with the faulty signer excluded from the census.  They continue to do so until DKG completes, or fewer than 70% of the signers are honest.

jcnelson:
Hoping to hear more from @xoloki on this.  Did I get the above description right?  Also, I elided details on how the system deals with a faulty DKG coordinator, but I'm assuming it's basically the same as dealing with a faulty signer.

xoloki:
@jcnelson you basically have it right.  As we've discussed, though, if we allow DKG to complete with only 70% participation, that means that signing will require 100% participation of those who completed DKG.

[...]

Issue #3972, opened 2023-09-21
[stacks-signer] All signers should calculate the DKG key simulateously 


jferrant:
To prevent malicious actors, we should have all signers publish the result of the DKG round. This means we need to update stacks-signer code to have every signer run DKG simultaneously. A single coordinator should still trigger the round, but every signer should aggregate the results.

jferrant:
1. https://github.com/stacks-network/sbtc/issues/392
2. https://github.com/stacks-network/sbtc/issues/393
3. https://github.com/stacks-network/sbtc/issues/394 Depends on 1 and 2

jcnelson:
Task breakdown
- need to have a way to pick a coordinator, in order to start the protocol or resume from a coordinator crash
   - need to deal with network partitions / dueling coordinators
- need to replicate coordinator state to all nodes so they can monitor the coordinator's behavior
   - need to locally decide to ignore faulty (equivocating) or fail-slow/fail-stop coordinators
   - need to be able to detect and resume/restart DKG from a failed coordinator
   - nodes need to individually vote that nodes are faulty (66% threshold required to overcome Byzantine voters)
- need to individually vote for aggregate public key, taproot address, and tapscripts via `.pox-4`

N.B. DKG requires 100% participation from all signers -- all of them must complete their DKG steps.  Whoever completes DKG becomes the set of signers for the reward cycle (meaning, the more signers that must be kicked out, the lower our margin of error for clearing the 70% signing threshold)

PR #4227, opened 2024-01-03
Update wsts version to 6.0 and use Packet::verify function

jferrant:
### Description
Need the newest wsts version changes for handling <100% participation in DKG
Also required for processing coordinator messages while in the IDLE state which enables DKG to be calculated by all signers simultaneously

### Applicable issues

- Part of https://github.com/stacks-network/stacks-core/issues/4111
- Part of https://github.com/stacks-network/stacks-core/issues/3972

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4712,9 +4712,9 @@ dependencies = [
 
 [[package]]
 name = "wsts"
-version = "5.0.0"
+version = "6.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c250118354755b4abb091a83cb8d659b511c0ae211ccdb3b1254e3db199cb86"
+checksum = "1b2cb1ef1b26d526daae40c1ee657c83bbedaeefd7196f827b40ca79d13f0f34"
 dependencies = [
  "aes-gcm 0.10.2",
  "bs58 0.5.0",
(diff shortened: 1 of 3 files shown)


xoloki (approved):
Lgtm as long as CI (test_stackerdb_dkg specifically) passes

jcnelson:
I don't think the stackerdb tests are running here?

kantai (approved):

xoloki:
@jferrant I changed the issue description to specify that this does not close #4111, but it is a part of it.  I don't think we should close #4111 until I do that last piece of robustness, i.e. allow partial DkgPrivateShares participation as long as a DKG threshold subset of the DkgPublicShares responders also send private shares.

xoloki:
> I don't think the stackerdb tests are running here?

I ran `test_stackerdb_dkg` locally, and it passed.  But yeah I don't see an explicit CI action  that calls it @jcnelson

xoloki:
Ok @jcnelson it looks like `test_stackerdb_dkg` and the other `StackerDB` tests are part of `sbtc-tests.yml`, not `bitcoin-tests.yml`, and they aren't running currently.  @wileyj is there some reason the `sBTC`/`StackerDB` tests aren't running currently?  I know they take a while, but are they flakey in CI?

Issue #3864, opened 2023-08-22
Nakamoto[Block Validation]: Comms channels for RPC handler and block template/validation


kantai:
RPC methods for assembling a block template and validating a block must
be asynchronous: block assembly and validation are slow operations, and
the network stack should not stall while processing them. In order to
facilitate this, the RPC handler will send a request for template
generation or block validation over a comms channel. A different
thread will handle the work and then pass the result via the event
interface.

jcnelson:
I think this can be dropped

kantai:
I’m not sure that this can be dropped. Stackers still need to use a block validation endpoints, right?

PR #4228, opened 2024-01-04
Feat/async block proposal

kantai:
### Description

Adds the `/v2/block_proposal` endpoint with async processing and callback over the event observer interface.

It does this by spawning a thread (if no such thread is currently running) and passing it a handle to an event dispatcher.

### Applicable issues

- #3864 (it doesn't need a "comms channel" anymore, it just uses a thread with its own dispatcher).
- Supercedes #4084

### Todo

- Update the API/RPC docs with the synchronous -> asynchronous changes.


docs/rpc/openapi.yaml
```diff
@@ -551,3 +551,45 @@ paths:
             type: string
           description: The Stacks chain tip to query from. If tip == latest, the query will be run from the latest
             known tip (includes unconfirmed state).
+
+  /v2/block_proposal:
+    post:
+      summary: Validate a proposed Stacks block
+      tags:
+        - Mining
+      #operationId: ???
```
jcnelson, line 560: ???

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,549 @@
...
+    /// either accept or reject the proposal
+    ///
+    /// This is done in 3 stages:
+    /// - Static validation of the block, which checks the following:
+    ///   - Block header is well-formed
+    ///   - Transactions are well-formed
+    ///   - Miner signature is valid
+    /// - Validate threshold signature of stackers
+    /// - Validation of transactions by executing them agains current chainstate.
+    ///   This is resource intensive, and therefore done only if previous checks pass
```
jcnelson, line 173: It's really a shame that we can't easily overlap this with the similar code in `NakamotoChainState` without substantial refactoring.  Perhaps this is just going to have to be a back-burner issue.

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,549 @@
...
+    fn parse_octets(mut body: &[u8]) -> Result<NakamotoBlockProposal, Error> {
+        NakamotoBlockProposal::consensus_deserialize(&mut body).map_err(|e| match e {
+            CodecError::DeserializeError(msg) => {
+                Error::DecodeError(format!("Failed to deserialize posted transaction: {msg}"))
+            }
+            _ => e.into(),
+        })
+    }
+
+    /// Decode a JSON-encoded transaction
```
jcnelson, line 372: "transaction"?

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,549 @@
...
+impl RPCBlockProposalRequestHandler {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Decode a bare transaction from the body
+    fn parse_octets(mut body: &[u8]) -> Result<NakamotoBlockProposal, Error> {
+        NakamotoBlockProposal::consensus_deserialize(&mut body).map_err(|e| match e {
+            CodecError::DeserializeError(msg) => {
+                Error::DecodeError(format!("Failed to deserialize posted transaction: {msg}"))
```
jcnelson, line 366: "transaction"?

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,549 @@
...
+            PeerHost::IP(addr, ..) => addr.is_loopback(),
+        };
+
+        if !is_loopback {
+            return Err(Error::Http(403, "Forbidden".into()));
+        }
+
+        if preamble.get_content_length() == 0 {
+            return Err(Error::DecodeError(
+                "Invalid Http request: expected non-zero-length body for PostBlock".to_string(),
```
jcnelson, line 411: "PostBlock"?

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,549 @@
...
+                "Invalid Http request: BlockProposal body is too big".to_string(),
+            ));
+        }
+
+        let block_proposal = match preamble.content_type {
+            Some(HttpContentType::Bytes) => Self::parse_octets(body)?,
+            Some(HttpContentType::JSON) => Self::parse_json(body)?,
+            None => {
+                return Err(Error::DecodeError(
+                    "Missing Content-Type for transaction".to_string(),
```
jcnelson, line 426: "transaction"?

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,549 @@
...
+                ))
+            }
+            _ => {
+                return Err(Error::DecodeError(
+                    "Wrong Content-Type for transaction; expected application/json or application/octet-stream".to_string(),
```
jcnelson, line 431: "transaction"?

stackslib/src/net/api/postblock_proposal.rs
kantai: Yeah, I agree. The code overlap isn't _so_ excessive, though, because most of the work here is already taking place through invocations of `begin_tenure`, etc. There's also a repeated code block here that needs to be removed: the stacker signature check. That cannot be part of the proposal validation (because the proposal validation is occurring to decide whether or not to sign in the first place).

stackslib/src/net/api/postblock_proposal.rs
kantai: I'm removing `parse_octets` entirely, but yes, I'll clean up the "transaction" copypasta while working on the docs updates.

stackslib/src/net/api/postblock_proposal.rs
kantai: Fixed in aad22e9

stackslib/src/net/api/postblock_proposal.rs
kantai: Fixed in aad22e9

stackslib/src/net/api/postblock_proposal.rs
kantai: Fixed in aad22e9

stackslib/src/net/api/postblock_proposal.rs
kantai: Removed in aad22e9

stackslib/src/net/api/postblock_proposal.rs
kantai: Fixed in aad22e9

docs/rpc/openapi.yaml
kantai: Added an operationId in aad22e9

jcnelson (approved):
LGTM. Thanks!

jcnelson:
Seems your OpenAPI validation is failing

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -0,0 +1,489 @@
...
+        }
+
+        let block_proposal = match preamble.content_type {
+            Some(HttpContentType::JSON) => Self::parse_json(body)?,
+            None => {
+                return Err(Error::DecodeError(
+                    "Missing Content-Type for block proposal".to_string(),
+                ))
+            }
+            _ => {
```
jferrant, line 386: This is just a side note, but I had a brain fart right here reading this. Would be clearer if you reorged it to 

```
...
            Some(HttpContentType::JSON) => Self::parse_json(body)?,
            Some(_) => {
                return Err(Error::DecodeError(
                    "Wrong Content-Type for block proposal; expected application/json".to_string(),
                ))
            }
            None => {
                return Err(Error::DecodeError(
                    "Missing Content-Type for block proposal".to_string(),
                ))
            }

File: testnet/stacks-node/src/tests/nakamoto_integrations.rs [hunk 9]
```diff
@@ -946,3 +972,336 @@ fn correct_burn_outs() {
...
+    let mut sign = |mut p: NakamotoBlockProposal| {
+        p.block
+            .header
+            .sign_miner(&privk)
+            .expect("Miner failed to sign");
+        signer.sign_nakamoto_block(&mut p.block);
+        p
+    };
+
+    let block = {
```
jferrant, line 1116: I wonder if I should use something like this rather than the test observer to validate the block written to stacker db in my .miners PR.

jferrant (approved):

stackslib/src/net/api/posttransaction.rs
```diff
@@ -162,7 +162,7 @@ impl HttpRequest for RPCPostTransactionRequestHandler {
             }
             _ => {
                 return Err(Error::DecodeError(
-                    "Wrong Content-Type for transaction; expected application/json".to_string(),
+                    "Wrong Content-Type for transaction; expected application/json or application/octet-stream".to_string(),
```
jbencin, line 165: You removed the `application/octet-stream` option

docs/rpc/openapi.yaml
```diff
@@ -551,3 +551,35 @@ paths:
...
+    post:
+      summary: Validate a proposed Stacks block
+      tags:
+        - Mining
+      operationId: post_block_proposal
+      description: |
+        Used by stackers to validate a proposed Stacks block from a miner.
+        
+        **This endpoint will only accept requests over the local loopback network interface.**
+      responses:
```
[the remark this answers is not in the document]
jbencin: Okay I see that on a `202` results are returned via event observer. There are still situations where you can get a `400` response, like a `NetError`

docs/rpc/openapi.yaml
```diff
@@ -551,3 +551,35 @@ paths:
...
+        202:
+          description: Block proposal is valid
```
jbencin, line 567: Now `202` just means "Accepted for processing", not necessarily "valid"

stackslib/src/net/p2p.rs
```diff
@@ -500,6 +506,23 @@ impl PeerNetwork {
         network
     }
 
+    pub fn set_proposal_thread(&mut self, thread: JoinHandle<()>) {
```
jbencin, line 509: Should this check if `self.block_proposal_thread` is `None` first?

jbencin (commented):
Looks good, I think there's a couple issues in the docs that need attention before merging

stackslib/src/net/api/posttransaction.rs
kantai: That's not true for `posttransaction`, though -- I think that still accepts an `octet-stream`

stackslib/src/net/p2p.rs
kantai: I don't think so -- the caller needs to have already checked this, because it shouldn't even get a `JoinHandle` if the `block_proposal_thread` is still running.

docs/rpc/openapi.yaml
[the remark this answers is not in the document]
kantai: Right, those 400 responses are common to all API endpoints though.

stackslib/src/chainstate/stacks/mod.rs
```diff
@@ -693,6 +693,22 @@ impl ToSql for ThresholdSignature {
     }
 }
 
+impl serde::Serialize for ThresholdSignature {
```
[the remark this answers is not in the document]
kantai: Yep, and this is already done for the SQL implementations, so it seemed like a straightforward improvement.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
kantai: I think the approach you use in your PR is better for the `.miners` check: you want to confirm that the miner thread's actual block matches the data read out of the stackerdb, not that an independently assembled block matches either of those.

stackslib/src/net/api/posttransaction.rs
jbencin: Oh yeah I didn't check the filename there... that's a message I had fixed for a *different* endpoint

jbencin (approved):

PR #4232, opened 2024-01-05
Nakamoto Tenure Inventories

jcnelson:
This PR adds the ability for the networking stack to query and generate Nakamoto inventory messages for Nakamoto tenures.  It adds `GetNakamotoInv` to request an inventory for a reward cycle (identified by its start-block consensus hash), and `NakamotoInv` to reply the bit vector of available inventories.  The ith bit is set if the node has the ith _processed_ tenure in the identified reward cycle.

This PR appears big only because it splits up `stackslib/src/net/inv.rs` into separate epoch 2.x and nakamoto files, and further splits out the unit tests in `inv.rs` into a separate test file.  The following files are unchanged from `stackslib/src/net/inv.rs`:

* `stackslib/src/net/inv/epoch2x.rs` -- the epoch 2.x inventory state machine (unchanged)
* `stackslib/src/net/inv/tests/epoch2x.rs` -- the epoch 2.x inventory state machine tests (unchanged, except to bind on any available ports instead of hard-coded ones).


stackslib/src/net/inv/nakamoto.rs
```diff
@@ -0,0 +1,215 @@
...
+
+            if let Some(cur_tenure_info) = cur_tenure_opt.as_ref() {
+                // a tenure was active when this sortition happened...
+                if cur_tenure_info.tenure_id_consensus_hash == cur_consensus_hash {
+                    // ...and this tenure started in this sortition
+                    tenure_status.push(true);
+                    cur_tenure_opt = self.get_processed_tenure(
+                        chainstate,
+                        &cur_tenure_info.parent_tenure_id_consensus_hash,
+                    )?;
```
kantai, line 191: is `cur_tenure_info.parent_tenure_id_consensus_hash` _always_ equal to `parent_sortition_consensus_hash`? I don't think it is... for example, if the prior bitcoin block didn't produce a sortition winner. 

In which case, I think this loop's variable maintenance would be broken: `cur_height` would decrement by 1, `cur_consensus_hash` would point at the parent bitcoin block, and `cur_tenure_opt` would be a tenure starting at a different bitcoin block.

My suggestion would be to move `cur_tenure_opt`'s binding into the loop's scope, and then the start of the loop initializes  cur_tenure_opt:

```rust
loop {
    ...
    if cur_reward_cycle < reward_cycle {
        break;
    }
    let cur_tenure_opt = self.get_processed_tenure(chainstate, &cur_consensus_hash);
```

I think this makes the loop easier to understand, and eliminates code. Because the tenure_status could just be:

```rust
// we've processed the tenure if we have information for it, and the tenure began in this sortition.
let tenure_status = matches!(
    cur_tenure_opt.as_ref(), 
    Some(cur_tenure_info) if cur_tenure_info.tenure_id_consensus_hash == cur_consensus_hash
);
tenure_statuses.push(tenure_status);
```

Or something along those lines if you don't prefer to use `matches!`

stackslib/src/net/mod.rs
```diff
@@ -836,19 +836,47 @@ pub struct Preamble {
...
+/// Aligned to a PoX reward cycle.
+/// This struct is used only in Nakamoto, for Nakamoto inventories
+#[derive(Debug, Clone, PartialEq)]
+pub struct GetNakamotoInvData {
+    /// Consensus hash at the start of the reward cycle
+    pub consensus_hash: ConsensusHash,
+}
+
+/// A bit vector that describes Nakamoto tenure availability.  Sent in reply for GetBlocksInv for
+/// Nakamoto block data.
```
kantai, line 872: My understanding from reading the generator code is:

`inv[i] = true` means that the i-th bitcoin block in a given reward cycle has a tenure start in the given block _and_ that this node has processed all the blocks of that sortition.

`inv[i] = false` is the inverse of that (i.e., either no tenure started at the block or this node has not processed all the blocks of that sortition).

Is that understanding correct? If so, can you add it to this rustdoc?

stackslib/src/chainstate/nakamoto/tenure.rs
```diff
@@ -509,6 +509,21 @@ impl NakamotoChainState {
         }
     }
 
+    /// Get a nakamoto tenure-change by its tenure ID consensus hash.
+    /// Get the highest such record.
```
kantai, line 513: This also looks like its filtering by tenure changes caused by sortitions -- can you add that to the rustdoc?

kantai (commented):
This looks mostly good to me -- I had a couple questions and suggestions. There's only one point that I think may require more investigation (the iteration on generating the bit vector) and testing if so.

stackslib/src/net/inv/nakamoto.rs
```diff
@@ -0,0 +1,215 @@
...
+        &mut self,
+        tip: &BlockSnapshot,
+        sortdb: &SortitionDB,
+        chainstate: &StacksChainState,
+        reward_cycle: u64,
+    ) -> Result<Vec<bool>, NetError> {
+        let ih = sortdb.index_handle(&tip.sortition_id);
+        let reward_cycle_end_height = sortdb
+            .pox_constants
+            .reward_cycle_to_block_height(sortdb.first_block_height, reward_cycle + 1)
```
jferrant, line 143: Wouldn't the end height of the reward cycle be -1? D:

stackslib/src/net/inv/nakamoto.rs
```diff
@@ -0,0 +1,215 @@
...
+            }
+            let cur_sortition_info = if let Some(info) = self.sortitions.get(&cur_consensus_hash) {
+                info
+            } else {
+                let loaded_info = InvSortitionInfo::load(sortdb, &cur_consensus_hash)?;
+                self.sortitions
+                    .insert(cur_consensus_hash.clone(), loaded_info);
+                self.sortitions
+                    .get(&cur_consensus_hash)
+                    .expect("infallbile: just inserted this data".into())
```
jferrant, line 177: Nit: typo "infallible"

stackslib/src/net/inv/nakamoto.rs
jcnelson: > is cur_tenure_info.parent_tenure_id_consensus_hash always equal to parent_sortition_consensus_hash? I don't think it is... for example, if the prior bitcoin block didn't produce a sortition winner.

Yes, correct. That's accounted for in the `else`-branch of this `if/else` block.  This is also tested via the `test_nakamoto_inv_2_tenures_3_sortitions` unit tests.  You can see that we get a `false` for one of the sortitions, since it didn't have a corresponding tenure.

stackslib/src/net/mod.rs
jcnelson: Correct -- if the node didn't process a tenure's data, then the bit for it is 0 (note that this is different from epoch 2, where you'd get a 1 as long as the node had a copy of the block, even if it was not processed).  I'll update the doc.

stackslib/src/chainstate/nakamoto/tenure.rs
jcnelson: Sure

stackslib/src/net/inv/nakamoto.rs
jcnelson: Sure -- `reward_cycle_to_block_height()` returns a value that is `mod 1` of the reward cycle length.  The `- 2` gets us the last block in the last reward cycle. So if `reward_cycle_to_block_height` returns (for example) 823551, then this would evaluate to 8235549.  This is the bitcoin block height we actually want to start with, because the loop invariant will scan down to block 821450 (inclusive) due to the fact that `block_height_to_reward_cycle` will return the same reward cycle for 823450 as 823551.

stackslib/src/net/inv/nakamoto.rs
jcnelson: Fixed

stackslib/src/net/codec.rs
```diff
@@ -316,6 +296,66 @@ impl BlocksInvData {
...
+impl NakamotoInvData {
+    pub fn empty() -> Self {
+        Self {
+            bitlen: 0,
+            tenures: vec![],
+        }
+    }
+
+    pub fn bools_to_bitvec(bits: &[bool]) -> Vec<u8> {
+        let mut bitvec = vec![0u8; (bits.len() / 8) + (if bits.len() % 8 != 0 { 1 } else { 0 })];
```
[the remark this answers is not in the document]
jcnelson: Sure; added

stackslib/src/net/inv/nakamoto.rs
kantai: The problem is this part of the code:

```rust
            if let Some(cur_tenure_info) = cur_tenure_opt.as_ref() {
                // a tenure was active when this sortition happened...
                if cur_tenure_info.tenure_id_consensus_hash == cur_consensus_hash {
                    // ...and this tenure started in this sortition
                    tenure_status.push(true);
                    cur_tenure_opt = self.get_processed_tenure(
                        chainstate,
                        &cur_tenure_info.parent_tenure_id_consensus_hash,
                    )?;
  ...
```

Combined with this part of the code:

```rust
            cur_consensus_hash = parent_sortition_consensus_hash;
            if cur_height == 0 {
                break;
            }
            cur_height = cur_height.saturating_sub(1);
```

So, for example, if the bitcoin blocks 1 & 3 contained successful sortitions but 2 didn't, after evaluating block 3, the loop would set:

`cur_consensus_hash = bitcoin block 2` (because `parent_consensus_hash` of 3 is 2).
`cur_tenure_opt = Some(tenure of bitcoin block 1)` (because the `parent_tenure_id_consensus_hash` of 3 would be 1)
`cur_height = 2`

So, the three loop variables would be out of sync. Maybe this doesn't create an issue in tests because `cur_tenure_info.consensus_hash != consensus_hash`, so it pushes `tenure_status = false` anyways, but it should be observable at some point, because the loop variables do appear to get unsynced.

stackslib/src/net/inv/nakamoto.rs
jferrant: ahhh okay this makes sense.

stackslib/src/net/inv/nakamoto.rs
jcnelson: > So, the three loop variables would be out of sync.

No, that's deliberate.  `cur_height` tracks the position in the inventory vector, which gets set to `true` or `false` based on whether or not `cur_tenure_opt` matches the consensus hash of the sortition corresponding to `cur_height`.

Basically, this loop checks each sortition at every height in the reward cycle, and determines from there whether or not the sortition was the start of a tenure. If so, then `true` gets pushed to the inventory vector. Otherwise, `false` will be pushed.

I've significantly increased test coverage to synthesize multi-reward-cycle chain histories from inventory vectors, and then verified that the inventory generation logic reproduces the inventory vectors from the synthesized chain history.

jcnelson:
@kantai @jferrant @xoloki ping :pray:

kantai (approved):
This LGTM, just add a comment to the loop at `inv/nakamoto.rs:158` describing the invariants of the loop variables (`cur_tenure_opt`, `cur_height`, and `cur_consensus_hash`, where `cur_tenure_opt` if set refers to the active tenure, even if that tenure started before the current height.)

jferrant (approved):
LGTM

PR #4239, opened 2024-01-10
[clarity] Cast aggregated public key vote

### LINKED ISSUE #3968: [Clarity] DKG vote registration
Signers should be able to submit their DKG vote to the clarity smart contract

May be able to reuse vote DKG registration code within https://github.com/Trust-Machines/stacks-sbtc/blob/main/sbtc-mini/contracts/sbtc-registry.clar

### LINKED ISSUE #4060: [Clarity] fetch vote status per signer per cycle
Needed by signers, will expand on after initial research w/ @jferrant

friedger:
### Description
This PR adds clarity code to tally a vote for the aggregated key for each cycle.

### Applicable issues

- fixes #3968 
- fixes #4060

### Checklist

- [ ] Changelog is updated


stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
```diff
@@ -0,0 +1,56 @@
...
+
+(define-public (vote-for-aggregate-public-key (key (buff 33)) (reward-cycle uint) (round uint) (tapleaves (list 4001 (buff 33))))
+    (let ((signer-public-key (unwrap! (get-signer-public-key tx-sender reward-cycle) err-not-allowed))
+            ;; one slot, one vote
+            (num-slots (get-signer-slots signer-public-key reward-cycle))
+            (tally-key {reward-cycle: reward-cycle, round: round, aggregate-public-key: key})
+            (new-total (+ num-slots (default-to u0 (map-get? tally tally-key))))
+            (current-round (var-get last-round)))
+        (asserts! (is-eq reward-cycle (current-reward-cycle)) err-incorrect-reward-cycle)
+        (asserts! (is-eq round current-round) err-incorrect-round)
```
setzeus, line 52: We can add an assert that the aggregate-public-key isn't being re-used no? Even if the signer-set doesn't change the aggregate-public-key should be different per round I believe.

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
```diff
@@ -0,0 +1,56 @@
...
+;; get voting info by burn block height
+(define-read-only (get-info (height uint))
+    (ok (at-block (unwrap! (get-block-info? id-header-hash height) err-invalid-burn-block-height) (get-current-info))))
+
+;; get current voting info
+(define-read-only (get-current-info)
+    (if (var-get is-state-1-active) (var-get state-1) (var-get state-2)))
+
+(define-read-only (get-signer-public-key (signer principal) (reward-cycle uint))
+    ;; TODO replace with contract-call to pox-4::get-signer-public-key
```
jcnelson, line 33: This might actually be in a `.signers` contract, which the node will populate with the reward set information when it begins a new reward cycle.  Then, it'll be cheaper to do this contract call since the `.signers` code body will be substantially shorter.

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
```diff
@@ -0,0 +1,56 @@
...
+    ;; defined in PR https://github.com/stacks-network/stacks-core/pull/4092
+    ;; (contract-call? .pox-4 get-signer-public-key reward-cycle signer)
+    (some 0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20))
+
+(define-read-only (get-signer-slots (signer-public-key (buff 33)) (reward-cycle uint))
+    u1000000000000)
```
jcnelson, line 39: Presumably this is mocked data?

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
```diff
@@ -0,0 +1,56 @@
...
+
+(define-read-only (current-reward-cycle)
+    u0)
```
jcnelson, line 42: Please implement this.

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
jcnelson: Yes, it should be.

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
```diff
@@ -0,0 +1,56 @@
...
+
+(define-public (vote-for-aggregate-public-key (key (buff 33)) (reward-cycle uint) (round uint) (tapleaves (list 4001 (buff 33))))
+    (let ((signer-public-key (unwrap! (get-signer-public-key tx-sender reward-cycle) err-not-allowed))
+            ;; one slot, one vote
+            (num-slots (get-signer-slots signer-public-key reward-cycle))
+            (tally-key {reward-cycle: reward-cycle, round: round, aggregate-public-key: key})
+            (new-total (+ num-slots (default-to u0 (map-get? tally tally-key))))
+            (current-round (var-get last-round)))
+        (asserts! (is-eq reward-cycle (current-reward-cycle)) err-incorrect-reward-cycle)
```
jcnelson, line 51: Why have `(reward-cycle uint)` as an argument if it's going to be computed and checked against the real reward cycle?

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
```diff
@@ -0,0 +1,56 @@
...
+(define-map tally {reward-cycle: uint, round: uint, aggregate-public-key: (buff 33)} uint)
+
+(define-constant err-not-allowed (err u10000))
+(define-constant err-incorrect-reward-cycle (err u10001))
+(define-constant err-incorrect-round (err u10002))
+(define-constant err-invalid-aggregate-public-key (err u10003))
+(define-constant err-duplicate-vote (err u10004))
+(define-constant err-invalid-burn-block-height (err u10005))
+
+(define-data-var last-round uint u0)
```
jcnelson, line 17: This is never set.

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
```diff
@@ -0,0 +1,56 @@
+;;
```
jcnelson, line 1: A couple of overall things about this contract:

* There doesn't need to be a `last-round` data var.  The signers themselves decide what the round number is.  For now, the node would accept the first aggregate public key that receives 70% of the votes in a given reward cycle.  In the future, the round would be used to allow DKG to happen mid-cycle, but it's currently unknown if that will ship with Nakamoto or sBTC.

* The node will track aggregate public key votes via transaction event receipts, which it will inspect in the chains coordinator.  Can you add `(print )` statements for when a vote happens?  Then, the node can just inspect the `votes` and `tally` maps at the end of the block.

jferrant:
Is this meant ot be built off of next? I don't see a corresponding PR for chore/pox-4-single-file

Also make sure it is rebased on next and does not contain merge commits

friedger:
It is meant to be off the code that contains `.signer`

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
jferrant: Tracking the last-round does actually help. 

Stacks-signers do not currently track a round. They track rounds internally in wsts but I could see it being useful to have this globally stored somewhere.

Suppose Alice wants to do a DKG round but she just came online after Bob triggered DKG round 1. She did not witness this round complete. She will need to know what the last round number used to vote was so she can trigger a new DKG round with last_round + 1. This will result in her triggering a DKG round for an outdated number that will subsequently get ignored by the contract as it will simply see signers attempting to resubmit votes when they have already voted.

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
friedger: done

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
friedger: this replaced with a map of last rounds

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
friedger: last round map has been added

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
friedger: removed by now

stackslib/src/chainstate/stacks/boot/signers.clar
```diff
@@ -13,6 +13,10 @@
 (define-read-only (stackerdb-get-signer-slots)
 	(ok (var-get stackerdb-signer-slots)))
 
+(define-read-only (get-signer-slots (signer principal) (reward-cycle uint))
+	(ok u1)
+)
```
MarvinJanssen, line 18: For now, the function `get-signer-slots` can look like this:

```clarity
(define-read-only stackerdb-get-signer-slots (signer-index uint)
	(ok (element-at (var-get stackerdb-signer-slots) signer-index))
)
```

Signers need to know their own index so it can be passed to the vote function. You can call it wrapped in the proper `at-block` from `pox-4-vote` to get the slot you want.

It is not the most efficient way but it will unblock the feature. We can later do a runtime cost assessment and decide to change it.

For example, it can also be an indexed map much like the PoX reward set info, but then the signer binary would have to do multiple RPC calls to fetch all the slots (instead of one).

If we don't mind duplicating the data then we can have both the list and the indexed map.

cc @jferrant

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
```diff
@@ -0,0 +1,77 @@
...
+    (map-get? votes {reward-cycle: reward-cycle, round: round, signer: signer}))
+    
+(define-read-only (get-signer-slots (signer principal) (reward-cycle uint))
+    (contract-call? .signers get-signer-slots signer reward-cycle))
+
+;; aggregate public key must be unique and can be used only in a single cycle-round pair
+(define-read-only (is-valid-aggregated-public-key (key (buff 33)) (dkg-id {reward-cycle: uint, round: uint}))
+    (is-eq (default-to dkg-id (map-get? used-aggregate-public-keys key)) dkg-id))
+
+(define-public (vote-for-aggregate-public-key (key (buff 33)) (reward-cycle uint) (round uint))
```
MarvinJanssen, line 58: See comment about `stackerdb-get-signer-slots` in `.signers`. If we use that pattern then instead of a `(buff 33)` key you can accept a signer index `uint` here.

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
MarvinJanssen, line 1: Does it make sense to just call this `.voting`?

friedger: :100: 
I was thinking of `signers-voting`

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
```diff
@@ -0,0 +1,77 @@
...
+(define-data-var state-2 {reward-cycle: uint, round: uint, aggregate-public-key: (optional (buff 33)),
+    total-votes: uint}  {reward-cycle: u0, round: u0, aggregate-public-key: none, total-votes: u0})
+
+;; get voting info by burn block height
+(define-read-only (get-info (height uint))
+    (ok (at-block (unwrap! (get-block-info? id-header-hash height) err-invalid-burn-block-height) (get-current-info))))
+
+;; get current voting info
+(define-read-only (get-current-info)
+    (if (is-eq (/ (current-reward-cycle) u2) u0) (var-get state-1) (var-get state-2)))
```
MarvinJanssen, line 37: This should be:

```clarity
(is-eq (mod (current-reward-cycle) u2))
```

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
MarvinJanssen: That sounds even better.

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
friedger: I don't understand. Is the `key` related to the signer index?

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
friedger: done

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
friedger: replaced

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
MarvinJanssen: The signers know their own index so when they cast a vote by calling into the contract they can provide that index. You can then retrieve the tuple from `.signers` using that index and do your checks.

stackslib/src/chainstate/stacks/boot/signers.clar
friedger: Should this be part of a different PR?

friedger: added in this PR

stackslib/src/chainstate/stacks/boot/pox-4-vote.clar
friedger: using index now

hstove:
This PR should be updated to merge into `next`

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -0,0 +1,102 @@
...
+(define-data-var state-2 {reward-cycle: uint, round: uint, aggregate-public-key: (optional (buff 33)),
+    total-votes: uint}  {reward-cycle: u0, round: u0, aggregate-public-key: none, total-votes: u0})
+
+;; get voting info by burn block height
+(define-read-only (get-info (height uint))
+    (ok (at-block (unwrap! (get-block-info? id-header-hash height) err-invalid-burn-block-height) (get-current-info))))
+
+;; get current voting info
+(define-read-only (get-current-info)
+    (if (is-eq (mod (current-reward-cycle) u2)) (var-get state-1) (var-get state-2)))
```
kantai, line 38:
```suggestion
    (if (is-eq u0 (mod (current-reward-cycle) u2)) (var-get state-1) (var-get state-2)))
```
`(is-eq <single-arg>)` would always return `true`, right?

kantai (commented):
Just had one comment in the clarity code.

Otherwise, LGTM -- this contract will need to be updated as the stackerdb/signer work changes to support more signers, but that can be handled in a  follow-up PR.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -0,0 +1,102 @@
...
+            ;; one slot, one vote
+            (num-slots (try! (get-current-signer-slots signer-index)))
+            (new-total (+ num-slots (default-to u0 (map-get? tally tally-key)))))
+        (asserts! (is-in-voting-window burn-block-height reward-cycle) err-out-of-voting-window)
+        (asserts! (>= round (default-to u0 (map-get? rounds reward-cycle))) err-old-round)
+        (asserts! (is-eq (len key) u33) err-invalid-aggregate-public-key)
+        (asserts! (is-valid-aggregated-public-key key {reward-cycle: reward-cycle, round: round}) err-duplicate-aggregate-public-key)
+        (asserts! (map-insert votes {reward-cycle: reward-cycle, round: round, signer: tx-sender} {aggregate-public-key: key, reward-slots: num-slots}) err-duplicate-vote)
+        (map-set tally tally-key new-total)
+        (map-set used-aggregate-public-keys key {reward-cycle: reward-cycle, round: round})
```
setzeus, line 94: Shouldn't this be set only once consensus is reached / the key is set? Currently, I think no two signers would be able to vote for the same key.

friedger: They should always vote for the same key in the same round.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
friedger: Fixed

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -0,0 +1,102 @@
...
+                (unwrap! (get-block-info? id-header-hash height) err-invalid-burn-block-height)
+                    (get-current-signer-slots signer-index)))))
+
+(define-read-only (get-current-signer-slots (signer-index uint))
+    (let ((details (unwrap! (unwrap-panic (contract-call? .signers stackerdb-get-signer-by-index signer-index)) err-invalid-signer-index)))
+        (asserts! (is-eq (get signer details) tx-sender) err-signer-index-mismatch)
+        (ok (get num-slots details))))
+
+;; aggregate public key must be unique and can be used only in a single cycle-round pair
+(define-read-only (is-valid-aggregated-public-key (key (buff 33)) (dkg-id {reward-cycle: uint, round: uint}))
```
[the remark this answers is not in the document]
friedger: Renamed the error to "err-ill-formed-aggregate-public-key"

friedger:
Comments addressed

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
```diff
@@ -0,0 +1,433 @@
...
+            signer_key,
+            signer_nonce + 1,
+            signer_index,
+            &aggregated_public_key,
+            0,
+        ),
+    ];
+
+    let txids: Vec<Txid> = txs.clone().iter().map(|t| t.txid()).collect();
+    dbg!(txids);
```
MarvinJanssen, line 185: Can remove these lines.

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -151,7 +152,7 @@ pub fn boot_nakamoto<'a>(
         .map(|test_stacker| {
             (
                 PrincipalData::from(key_to_stacks_addr(&test_stacker.stacker_private_key)),
-                u64::try_from(test_stacker.amount).expect("Stacking amount too large"),
+                u64::try_from(test_stacker.amount + 10000).expect("Stacking amount too large"),
```
MarvinJanssen, line 155: I think we should do this in a nicer way. We had to explicitly add this for our reasons but it is not clear to anyone using the test helpers.

I'm not sure but I think you can add the test stackers to initial balances. So from your test you will pass an initial balance for Stacker 1, and revert the change on this line.

If that does not work because the initial balances list does not allow duplicates, then we have to modify the code here to sum duplicates.

contrib/core-contract-tests/tests/pox-4/signers-voting.test.ts
MarvinJanssen, line 1: Will all of this be turned into `clarunit` tests or will we have both TS and Clarity unit tests?

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -2186,6 +2217,38 @@ pub mod test {
         make_tx(key, nonce, 0, payload)
     }
 
+    pub fn readonly_call(
```
MarvinJanssen, line 2220: This is a test helper, it should not have been moved to this location. It takes a `TestPeer` and is hardcoded to testnet. 

You can put it back in `signer_tests.rs` or some other test file.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -0,0 +1,102 @@
...
+(define-data-var state-2 {reward-cycle: uint, round: uint, aggregate-public-key: (optional (buff 33)),
+    total-votes: uint}  {reward-cycle: u0, round: u0, aggregate-public-key: none, total-votes: u0})
+
+;; get voting info by burn block height
+(define-read-only (get-info (height uint))
+    (ok (at-block (unwrap! (get-block-info? id-header-hash height) err-invalid-burn-block-height) (get-current-info))))
+
+;; get current voting info
+(define-read-only (get-current-info)
+    (if (is-eq u0 (mod (current-reward-cycle) u2)) (var-get state-1) (var-get state-2)))
```
MarvinJanssen, line 38: Per discussion with @kantai. Since the slot switching is currently not worked out he said to only implement against the current state of affairs; i.e., a single list of signers.

Issue: #4307

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
```diff
@@ -0,0 +1,433 @@
...
+    //
+    // vote in the last burn block of prepare phase
+    //
+
+    nakamoto_tenure(
+        &mut peer,
+        &mut test_signers,
+        vec![vec![dummy_tx_1]],
+        signer_1_key,
+    );
```
MarvinJanssen, line 314: Can be worth it adding a helper to advance _N_ burn blocks. (And construct the dummy tx inside.)

stackslib/src/chainstate/stacks/boot/signers-voting.clar
setzeus: Correct. I'm saying the say this (map-set ...) is set up *every* vote adds a key to the "is-valid-aggregated-public-key" no?

So let's say Alice votes for the correct aggregate public key X, with the current setup X is automatically added as a used-aggregate-public-key. 

So when Bob comes along to also vote for the correct aggregate public key X, he **won't** be able to & will get an err-duplicate-aggregate-public-key error.

friedger: is duplicate pub key function accepts the same key for the same cycle and round.

contrib/core-contract-tests/tests/pox-4/signers-voting.test.ts
friedger: When PR is merged we can convert

stackslib/src/chainstate/stacks/boot/signers-voting.clar
friedger: Info about the state is not used anywhere.

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
friedger: Let's do sponsored txs.

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
friedger: :100: 

However, tooling for rust testing pox is out of scope of this PR.

Currently, I like the explicit advancement.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
friedger: now only state-1 is used

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
friedger: Apparently, we don't need it

stackslib/src/chainstate/stacks/boot/mod.rs
friedger: done

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
friedger: done

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
MarvinJanssen: Don't need what?

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
MarvinJanssen: Fair enough!

setzeus (approved):

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
```diff
@@ -0,0 +1,435 @@
...
+        })
+    );
+}
+
+/// In this test case, Alice votes in the first block of the last tenure of the prepare phase.
+/// Bob votes in the second block of that tenure.
+/// Alice can vote successfully.
+/// Bob is out of the voting window.
+#[test]
+fn vote_for_aggregate_public_key_in_last_block() {
```
jferrant, line 228: @kantai aren't we able to vote for DKG at any point in a reward cycle ? @friedger

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -0,0 +1,105 @@
...
+
+(define-constant pox-info
+    (unwrap-panic (contract-call? .pox-4 get-pox-info)))
+
+;; maps reward-cycle ids to last round
+(define-map rounds uint uint)
+
+(define-data-var state-1 {reward-cycle: uint, round: uint, aggregate-public-key: (optional (buff 33)),
+    total-votes: uint}  {reward-cycle: u0, round: u0, aggregate-public-key: none, total-votes: u0})
+(define-data-var state-2 {reward-cycle: uint, round: uint, aggregate-public-key: (optional (buff 33)),
```
jferrant, line 29: Is state-2 for when we have hand off basically?

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
friedger: My understanding was that we need to know the number of slots. This can happen only in the prepare phase. We check that the cycle was set as and registered in .signers.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
friedger: Yes, when we alternate

MarvinJanssen (approved):

Issue #4242, opened 2024-01-12
[Nakamoto] stacks-signer needs track processed blocks responses,  block requests, and block sign requests


jferrant:
Stacks-signer currently will just process blocks as they are received and trigger signing rounds accordingly. it does not currently check if it has already processed a particular block response (received from the RPC endpoint) or block request (received from the miners) or block sign request (received from other signers). 

Add persistent state to the signers and update it to no longer unnecessarily trigger sign requests/block responses/ block validation requests based on prior received information.

Depends on https://github.com/stacks-network/stacks-core/pull/4130

hstove:
@kantai is the assumption that the signer will rely on disk (ie SQLite) for state?

jferrant:
> @kantai is the assumption that the signer will rely on disk (ie SQLite) for state?

That is kind of what I was thinking, but you can choose what you like! Voluntold!

hstove:
Ok! Sounds good. I'll definitely use SQLite.

> I think we have the logic in place for what to store on the side

Did you mean to say "on the signer"?

At first thought, I can imagine two ways of doing this:

- Track event observer events and simply skip one if it's already been seen / handled. The Stacks API has basically the same functionality which I'm familiar with. We'd likely only want to store the events that are relevant to us, otherwise that'd get very large.
- Track `SignerEvent`s and have some mechanism for identifying a specific one, and skip based on that

My hunch would be that the first mechanism is more robust, but I can imagine there would be instances where more specific / queryable information could be useful.

A different question is: why is this necessary? I'm not doubting the ticket, I'm just not sure I understand the problem fully yet.

kantai:
Ah -- yes, sorry, SQLite was exactly what I was thinking here.

jferrant:
So right now we just store block info in a Hashmap and it grows almost indefinitely (only being removed following a successful broadcast of the signature) This garbage collection needs to be done. 

We already store I think everything we need on the side inside the stacks signer runloop struct (inside hashsets and hashmaps. Specifically blocks and txs). But this needs to be garbage collected for sure. 

We may not need to track transactions (as they are already written to stackerdb every time they are created following my current open PR getting merged), but we do need the garbage collection logic to clean up transactions as they get confirmed in blocks. 

I would expect that all block requests are stored and persisted between crashes so that if a miner reqeusts a block be signed and we crash before we can respond to it, we know to process it on reboot. 

[...]

PR #4249, opened 2024-01-16
Enable approval or rejection of a block via signature

jferrant:
### Description
This uses the results of a verification to determine whether to vote no or approve a block proposal.
It also broadcasts either a rejection or an approval back to the miners to observe in the .signers contract.

### Applicable issues

- Closes https://github.com/stacks-network/stacks-core/issues/3930
- Relates to https://github.com/stacks-network/stacks-core/issues/4242
 Depends on https://github.com/Trust-Machines/wsts/pull/46

### DIFF
--- a/stacks-signer/src/config.rs
+++ b/stacks-signer/src/config.rs
@@ -36,7 +36,7 @@ use wsts::state_machine::PublicKeys;
 /// List of key_ids for each signer_id
 pub type SignerKeyIds = HashMap<u32, Vec<u32>>;
 
-const EVENT_TIMEOUT_MS: u64 = 50;
+const EVENT_TIMEOUT_MS: u64 = 5000;
 
 #[derive(thiserror::Error, Debug)]
 /// An error occurred parsing the provided configuration
--- a/stacks-signer/src/runloop.rs
+++ b/stacks-signer/src/runloop.rs
@@ -84,7 +84,18 @@ pub struct BlockInfo {
     /// Our vote on the block if we have one yet
     vote: Option<Vec<u8>>,
     /// Whether the block contents are valid
-    valid: bool,
+    valid: Option<bool>,
+}
+
+impl BlockInfo {
+    /// Create a new BlockInfo
+    pub fn new(block: NakamotoBlock) -> Self {
+        Self {
+            block,
+            vote: None,
+            valid: None,
+        }
+    }
 }
 
 /// The runloop for the stacks signer
@@ -214,17 +225,23 @@ impl<C: Coordinator> RunLoop<C> {
     fn handle_block_validate_response(&mut self, block_validate_response: BlockValidateResponse) {
         match block_validate_response {
             BlockValidateResponse::Ok(block_validate_ok) => {
-                self.blocks
-                    .entry(
-                        block_validate_ok
-                            .block
-                            .header
-                            .signature_hash()
-                            .unwrap_or(Sha512Trunc256Sum::from_data(&[])),
-                    )
-                    .and_modify(|block_info| {
-                        block_info.valid = true;
-                    });
+                let Ok(hash) = block_validate_ok.block.header.signature_hash() else {
+                    warn!("Received a block proposal with an invalid signature hash. Broadcasting a block rejection...");
+                    let block_rejection = BlockRejection::new(block_validate_ok.block, RejectCode::InvalidSignatureHash);
+                    // Submit signature result to miners to observe
+                    if let Err(e) = self
+                        .stackerdb
+                        .send_message_with_retry(self.signing_round.signer_id, block_rejection.into())
+                    {
+                        warn!("Failed to send block submission to stacker-db: {:?}", e);
+                    }
+                    return;
+                };
+                let block = self
+                    .blocks
+                    .entry(hash)
+                    .or_insert(BlockInfo::new(block_validate_ok.block.clone()));
+                block.valid = Some(true);
                 // This is a valid block proposal from the miner. Trigger a signing round for it if we are the coordinator
                 let (coordinator_id, _) = calculate_coordinator(&self.signing_round.public_keys);
                 if coordinator_id == self.signing_round.signer_id {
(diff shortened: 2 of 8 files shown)


stacks-signer/src/runloop.rs
```diff
@@ -72,6 +77,16 @@ pub enum State {
     Sign,
 }
 
+/// Additional Info about a proposed block
+pub struct BlockInfo {
+    /// The block we are considering
+    block: NakamotoBlock,
+    /// Our vote on the block if we have one yet
+    vote: Option<Vec<u8>>,
```
jcnelson, line 85: What's the `Vec<u8>` here?

stacks-signer/src/runloop.rs
```diff
@@ -90,6 +105,10 @@ pub struct RunLoop<C> {
     pub state: State,
     /// Wether mainnet or not
     pub mainnet: bool,
+    /// Observed blocks that we have seen so far
+    pub blocks: HashMap<Vec<u8>, BlockInfo>,
+    /// Transactions that we expect to see in the next block
```
jcnelson, line 110: Does the order matter?  If not, then perhaps a `HashSet` might be better (especially since there can't be duplicates anyway)?

stacks-signer/src/runloop.rs
```diff
@@ -90,6 +105,10 @@ pub struct RunLoop<C> {
     pub state: State,
     /// Wether mainnet or not
     pub mainnet: bool,
+    /// Observed blocks that we have seen so far
+    pub blocks: HashMap<Vec<u8>, BlockInfo>,
```
jcnelson, line 109: I think this `Vec<u8>` is a `Sha512Trunc256Sum`?

jcnelson: Also, does `blocks` ever get garbage-collected, or does it grow forever?

stacks-signer/src/runloop.rs
```diff
@@ -300,10 +345,177 @@ impl<C: Coordinator> RunLoop<C> {
...
+                true
+            }
+            Some(None) => {
+                // We have seen this block before, but we have not agreed to sign it. TODO: ignore it
+                debug!("Received a signature share request for a block we have not validated yet.");
+                false
+            }
+            None => {
+                // We have not seen this block before.
+                // TODO: should probably ignore any messages that are not either sBTC transactons or Nakamoto blocks. Leave now for abitrary message signing
```
jcnelson, line 370: Isn't this a problem though?  We don't want the system to offer a signing oracle

stacks-signer/src/runloop.rs
```diff
@@ -300,10 +345,177 @@ impl<C: Coordinator> RunLoop<C> {
...
+        }
+    }
+
+    /// Helper function to validate a nonce request, updating its message appropriately.
+    /// Note that if the request is for a block, we will update the request message
+    /// as either a hash indicating a vote no or the signature hash indicating a vote yes
+    /// Returns whether the request is valid or not
+    fn validate_nonce_request(&mut self, request: &mut NonceRequest) -> bool {
+        let mut ptr = &request.message[..];
+        let Some(block) = read_next::<NakamotoBlock, _>(&mut ptr).ok() else {
```
jcnelson, line 383: Strictly speaking, you don't need the `ptr` binding.  This also works:

```
let Some(block): Option<NakamotoBlock> = read_next(&mut &request.message[..]).ok() else { /* ... */ }`

File: stacks-signer/src/runloop.rs [hunk 6]
```diff
@@ -300,10 +345,177 @@ impl<C: Coordinator> RunLoop<C> {
...
+            debug!("The block passed validation. Update the request with the signature hash.");
+        }
+        // Cache our vote
+        block_info.vote = Some(hash_bytes.clone());
+        request.message = hash_bytes;
+        true
+    }
+
+    /// Helper function to validate a block's contents
+    fn validate_block(block_info: &BlockInfo, transactions: &[Txid]) -> bool {
```
jcnelson, line 415: Strictly speaking, we're not validating the block here; we're just checking that a set of required transactions is present.

stacks-signer/src/runloop.rs
```diff
@@ -300,10 +345,177 @@ impl<C: Coordinator> RunLoop<C> {
...
+        };
+        let mut hash_bytes = hash.0.to_vec();
+        let transactions = &self.transactions;
+        let block_info = self.blocks.entry(hash_bytes.clone()).or_insert(BlockInfo {
+            vote: None,
+            valid: false,
+            block: block.clone(),
+        });
+        // Validate the block contents
+        block_info.valid = Self::validate_block(block_info, transactions);
```
jcnelson, line 400: Don't we actually want to validate the block via the RPC endpoint here?  Or, is this already done by the time this code is reached?

jcnelson: Assuming this is reached by `verify_chunk()`, I'm assuming that you still need to do the RPC call to the node?

jferrant: The current flow is this:

miner proposes block, all signers see it and immediately do a call to the rpc endpoint.

Signers receive their node endpoiint rpc responses as events and they update their seen block with this info.

Coordinator gets the node endpoint rpc response and triggers a signing round

Signers get the block sign request from coordinators and check against their map of blocks and see if they have validated it against the endpoint or not (will be false if they have not) -> signers will approve or deny accordingly

Signers and coordinators all broadcast the result of the signing round back to the miners.

stacks-signer/src/runloop.rs
jferrant: We are validating it as it checks if we received the endpoint response. If not, it will return false.

stacks-signer/src/runloop.rs
jferrant: I have a remove block on each sign completion. However, if we never complete a sign, it will grow indefinitely. I have left this as is because the state ticket is on Aaron's todo list and didn't want to get into it yet.

stacks-signer/src/runloop.rs
jferrant: Good point. Will update.

stacks-signer/src/runloop.rs
jferrant: Thjis is the signature hash or the signature hash + "n" to indicate a no vote. We must store this to the side as a cooridnator could be malicious and overwrite with a fake message in the signature request OR because we lost our vote but want to remain consistent in our voting. Also to enable a signer nonce request to be over an entire serialized block while the response be over its hash (or of course the hash + b'n' to indicate rejection)

stacks-signer/src/runloop.rs
jferrant: And you are right it is a Sha512Trunc256Sum. I know I had it as this before, but it broke when I did this so I reverted it. I will try again, but it has to do with the serialization/deserialization or something of the message to sign.

stacks-signer/src/runloop.rs
jferrant: I might be able to update these tests now...will see...

stacks-signer/src/runloop.rs
jferrant: Done :D

stacks-signer/src/runloop.rs
jferrant: So I have updated the code to be a bit more resilient to out of order messages.

It will currently cache nonce requests on the side for blocks it has not yet validated. Responding to these nonce requests once it has the validated info returned. It is slightly janky how I currently store blocks on the side, but am hoping with @kantai 's ticket to properly handle state, this will go away

stacks-signer/src/config.rs
```diff
@@ -36,7 +36,7 @@ use wsts::state_machine::PublicKeys;
 /// List of key_ids for each signer_id
 pub type SignerKeyIds = HashMap<u32, Vec<u32>>;
 
-const EVENT_TIMEOUT_MS: u64 = 50;
+const EVENT_TIMEOUT_MS: u64 = 5000;
```
xoloki, line 39: One question is how this interacts with timeouts.  If our timeouts are on the order of minutes, then an extra 5sec wouldn't really affect things.  But if our timeouts are in single digit seconds, the 5sec blocking would alter the timing significantly, and not necessarily in the same way on different signers.

xoloki (approved):
LGTM.  We can worry about the granularity of the event timeout when we know what our dkg/nonce/sign timeouts will be.

kantai (approved):
This looks good to me!

Issue #3932, opened 2023-09-15
Nokamoto[Block Proposal]: Signer binary must produce TenureChange transaction


kantai:
In addition to signing blocks, in Argon, the stacker signer binary is
responsible for creating `TenureChange` transactions. These transactions
should be issued whenever the signer's associated `stacks-node` processes
a new sortition. When a new sortition is processed, the signer should receive
an event, and then sign a new transaction declaring the new Tenure.

The tenure change transaction is composed of:

* The StacksBlockId of the last block from the previous miner that they processed.
* The number of blocks produced since the last non-empty sortition
* A flag to indicate which of the following triggered the tenure change
   - A valid winning block-commit
   - No winning block-commits
   - A “null miner” won the block-commit (in Neon, this won't be an option)
* The ECDSA public key hash of the next tenure's block-signing key
* The stacker-signer's signature over the transaction

saralab:
@soju-drinker Closing this issue. Context: 
Miner can produce the `tenurechange` transaction and the block validation logic must ensure that we cover the edge scenarios

PR #4260, opened 2024-01-18
Nakamoto: remove stacker signature fields from tenure change payload

### LINKED ISSUE #4258: [Nakamoto]: Miner must produce the tenurechange transcation
This is related to https://github.com/stacks-network/stacks-core/issues/3932 

Miner must produce the `tenurechange` transaction , block validation logic must validate this transaction

kantai:
### Description

Related to closing #3932 -- TenureChange payloads don't need stacker signatures, because stackers have to sign the block header, which incorporates the TenureChange payload in its tx hash.

The signer binary doesn't need to do anything special to validate these payloads: they have to sign off on blocks anyways, and the content of tenure change transactions is validated by the stacks-node's block proposal validation endpoint already.

Conveniently, this also doesn't place much of a burden on the stacks-node miner either: the nakamoto-neon miner was already constructing tenure change transactions in self-signing-mode , so this PR just makes it construct those transactions all the time.

Closes #4258

### DIFF
--- a/stackslib/src/chainstate/stacks/mod.rs
+++ b/stackslib/src/chainstate/stacks/mod.rs
@@ -731,10 +731,6 @@ pub struct TenureChangePayload {
     pub cause: TenureChangeCause,
     /// The ECDSA public key hash of the current tenure
     pub pubkey_hash: Hash160,
-    /// The Stacker signature
-    pub signature: ThresholdSignature,
-    /// A bitmap of which Stackers signed
-    pub signers: Vec<u8>,
 }
 
 impl TenureChangePayload {
@@ -752,8 +748,6 @@ impl TenureChangePayload {
             previous_tenure_blocks: num_blocks_so_far,
             cause: TenureChangeCause::Extended,
             pubkey_hash: self.pubkey_hash.clone(),
-            signature: ThresholdSignature::mock(),
-            signers: vec![],
         }
     }
 }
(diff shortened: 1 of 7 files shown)


stackslib/src/chainstate/stacks/mod.rs
```diff
@@ -731,10 +731,6 @@ pub struct TenureChangePayload {
     pub cause: TenureChangeCause,
     /// The ECDSA public key hash of the current tenure
     pub pubkey_hash: Hash160,
-    /// The Stacker signature
-    pub signature: ThresholdSignature,
-    /// A bitmap of which Stackers signed
-    pub signers: Vec<u8>,
```
jferrant, line 737: Do we want this bitmap for blocks?

jferrant (approved):

stackslib/src/chainstate/stacks/mod.rs
jcnelson: Yes, but we can add that back in via a separate PR.

jcnelson (approved):
LGTM as long as CI passes

jcnelson:
Why is codecov misbehaving?

kantai:
> Why is codecov misbehaving?

I think it's just that codecov returns a report before all of the tests have run (it returns after the first test runs), but in order to get enough coverage, all the tests have to run, so while tests are still running, it seems like codecov is failing. Once everything has run (as it has now), codecov should pass (as it is now).

Issue #4247, opened 2024-01-16
Signer should control which stackers can delegate to the signer 


friedger:
**Is your feature request related to a problem? Please describe.**
For legal or other reasons, signers might not want to accept signing duties for some stackers who want to delegate signing to them.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
@MarvinJanssen mentioned the issue first I think

jcnelson:
Stacking transactions would require a signature over the stacking function's arguments from the signing private key.  The stacking function would authenticate the signature in order to validate the signing public key.

kantai:
There's two possible ways to handle this:

1. Require a signature from the public key. The public key would sign something like a serialization of the `tx-sender` and the current reward cycle.

2. Require a public key allowance -- i.e., track which public key can be used by a `tx-sender` in a Clarity Map.

The advantage of (1) is that it simplifies the PoX contract. The advantage of (2) is that it doesn't require the signer to actually _use_ their signing key before the stacking cycle starts.

jcnelson:
Another advantage of (1) is that you can change your key in the same transaction as you stack, so you don't have to worry about mempool congestion preventing you from rotating your signing key.

hstove:
@saralab to assign me (hoping I can get added to the repo/org so I can add myself)

diwakergupta:
> @saralab to assign me (hoping I can get added to the repo/org so I can add myself)

igotchu

friedger:
Currently, Fast pool has implemented a self-service pool. This requirement will break the pool.

friedger:
Currently, Fast pool accepts any stacker as pool member. It is a self-service pool.
With the requirement to provide a signature the pool's self-service feature will break.

hstove:
@friedger yes, but you should be able to build tooling to automatically generate signatures for anyone who wants to join.

hstove:
@MarvinJanssen, now that the signer key is only provided at `stack-aggregation-commit`, is this still really an issue? Stackers don't need the key or signature during their transactions, nor do pools need the key/sig for `delegate-stack-stx` etc. And because the signature is only needed on `stack-agg-commit`, it can be generated only occasionally and from a secure setup.

MarvinJanssen:
That's true. It would only remain for solo stacking which is a specific use-case. Nonetheless, the problem would remain there.

friedger:
[...]

Issue #3980, opened 2023-09-22
stacks-node must write the new set of stackers to the .pox-4 contract


jferrant:
At T-99 block, Miner must write the new reward cycle set of signers and reward shares to the pox-4 contract.
The clarity stacker-db contract requires this information to update the config with the appropriate stacker addresses

Depends on https://github.com/stacks-network/stacks-blockchain/issues/3970
Related to (Duplicates?): https://github.com/stacks-network/stacks-core/issues/4058

jcnelson:
Context: once the PoX anchor block is chosen in Nakamoto (something that we'll know 99 blocks prior to the start of the next reward cycle), the node will need to (a) calculate the reward set right then and there, and (b) write this information into `.pox-4` somehow _in addition_ to into the sortition DB as it does today.  This is because the StackerDB trait implementation for `.pox-4` needs to know the addresses and number of slots for the next set of Stackers _before_ the next reward cycle begins (i.e. so the new Stackers can do DKG and vote on the next reward cycle's aggregate public key).

jcnelson:
The reward set is calculated in the first prepare-phase block, so the Stacks node will simply call into `.pox-4` to write the new signing key addresses to the StackerDB configuration for the new signers.  This would be part of the consensus rules, via `setup_block()` or some such.

setzeus:
Pasting design doc that mentions this issue for reference: https://docs.google.com/document/d/1IFTaHjEGHJkdFiEfKNayfmD0UlseoXpVcoMgJv0VW1g/edit?usp=sharing

MarvinJanssen:
This was addressed in #4263 and #4269. Signer sets are now written to a dedicated boot contact `.signers` instead of the PoX-4 contract.

Issue #3957, opened 2023-09-20
[clarity] signer StackerDB contract for pox-4


jferrant:
For signers to communicate with each other, they need to have access to a deployed stacker db contract. 

A stackerdb contract must contain the following two functions (update accordingly):
```
        ;; stacker DB
        (define-read-only (stackerdb-get-signer-slots)
            (ok (list
                {
                    signer: 'ST24GDPTR7D9G3GFRR233JMWSD9HA296EXXG5XVGA,
                    num-slots: u10
                }
                {
                    signer: 'ST1MR26HR7MMDE847BE2QC1CTNQY4WKN9XDKNPEP3,
                    num-slots: u10
                }
                {
                    signer: 'ST110M4DRDXX2RF3W8EY1HCRQ25CS24PGY22DZ004,
                    num-slots: u10
                }
                {
                    signer: 'ST69990VH3BVCV39QWT6CJAVVA9QPB1715HTSN75,
                    num-slots: u10
                }
                {
                    signer: 'STCZSBZJK6C3MMAAW9N9RHSDKRKB9AKGJ2JMVDKN,
                    num-slots: u10
                }
                )))

        (define-read-only (stackerdb-get-config)
            (ok {
                chunk-size: u4096,
                write-freq: u0,
                max-writes: u4096,
                max-neighbors: u32,
                hint-replicas: (list )
            }))
            ```
The stackerdb contract is read and loaded to each stacks node once at the start of a reward cycle.
 
The signer's stackerdb would be partitioned into two sets of slots:  one for the current stackers, and a corresponding one for the new stackers.  Which partition a stacker gets assigned to would be a function of reward cycle count.  For example, if the reward cycle is even, the current stackers use the first partition and the new stackers use the second partition.  But if it's odd, then the current stackers would use the second partition and the new stackers would use the first.

As the signer stackerdb contract config is only read and loaded into a stacks node once per reward cycle, the new set of stackers for reward cycle N + 1 MUST overwrite the stackers N - 1  PRIOR to the end of reward cycle N so that at the START of reward cycle N+1, the contract contains the stackers N and stackers N+1.

Related to https://github.com/stacks-network/stacks-blockchain/issues/3948

setzeus:
Reviewing here for my own sake, "The StackerDB contract is read and loaded to each stacks node once at the start of a reward cycle" reads to me that each node loads a new StackerDB boot contract near the end of ever cycle.

[...]

PR #4263, opened 2024-01-18
.signers StackerDB contract and PoX 4 signer key changes

MarvinJanssen:
### Description

Now rebased onto #4269.

Work in progress that adds:
- PoX4 changes, takes signer key `(buff 33)` parameter where needed, converts and stores them as `principal`.
- The `.signers` StackerDB contract.
- Uses #4269 Nakamoto `make_reward_set()` to calculate the signer slots.
- Fills in the mocks/todos in #4269 pertaining to the signers.
- Writes the signer set to `.signers`.
- Basic unit tests that Stacks using two different signer keys and checks they exist in the signers StackerDB contract.

As for the coupling of PoX reward slots and signer keys we have some open questions. (#4247 and others.)

This PR should unblock @friedger's work on the `.voting` boot contract.

### Applicable issues

Addresses:
- #4058
- #3980
- #3957

### Checklist

- [ ] Changelog is updated

### DIFF
--- a/stackslib/src/burnchains/mod.rs
+++ b/stackslib/src/burnchains/mod.rs
@@ -520,17 +520,6 @@ impl PoxConstants {
         (effective_height % u64::from(self.reward_cycle_length)) == 1
     }
 
-    pub fn is_prepare_phase_start(&self, first_block_height: u64, burn_height: u64) -> bool {
-        if burn_height < first_block_height {
-            false
-        } else {
-            let effective_height = burn_height - first_block_height;
-            (effective_height + u64::from(self.prepare_length))
-                % u64::from(self.reward_cycle_length)
-                == 0
-        }
-    }
-
     pub fn reward_cycle_to_block_height(&self, first_block_height: u64, reward_cycle: u64) -> u64 {
         // NOTE: the `+ 1` is because the height of the first block of a reward cycle is mod 1, not
         // mod 0.
@@ -547,6 +536,29 @@ impl PoxConstants {
             first_block_height,
             u64::from(self.reward_cycle_length),
         )
+    }
+
+    /// Return the reward cycle that the current prepare phase corresponds to if `block_height` is _in_ a prepare
+    /// phase. If it is not in a prepare phase, return None.
+    pub fn reward_cycle_of_prepare_phase(
+        &self,
+        first_block_height: u64,
+        block_height: u64,
+    ) -> Option<u64> {
+        if !self.is_in_prepare_phase(first_block_height, block_height) {
+            return None;
+        }
+        // the None branches here should be unreachable, because if `first_block_height > block_height`,
+        //   `is_in_prepare_phase` would have returned false, but no need to be unsafe anyways.
+        let effective_height = block_height.checked_sub(first_block_height)?;
+        let current_cycle = self.block_height_to_reward_cycle(first_block_height, block_height)?;
+        if effective_height % u64::from(self.reward_cycle_length) == 0 {
+            // if this is the "mod 0" block of a prepare phase, its corresponding reward cycle is the current one
+            Some(current_cycle)
+        } else {
+            // otherwise, the corresponding reward cycle is actually the _next_ reward cycle
+            Some(current_cycle + 1)
+        }
     }
 
     pub fn is_in_prepare_phase(&self, first_block_height: u64, block_height: u64) -> bool {
--- a/stackslib/src/burnchains/mod.rs
+++ b/stackslib/src/burnchains/mod.rs
@@ -520,6 +520,17 @@ impl PoxConstants {
         (effective_height % u64::from(self.reward_cycle_length)) == 1
     }
 
+    pub fn is_prepare_phase_start(&self, first_block_height: u64, burn_height: u64) -> bool {
+        if burn_height < first_block_height {
+            false
+        } else {
+            let effective_height = burn_height - first_block_height;
+            (effective_height + u64::from(self.prepare_length))
+                % u64::from(self.reward_cycle_length)
+                == 0
+        }
+    }
+
     pub fn reward_cycle_to_block_height(&self, first_block_height: u64, reward_cycle: u64) -> u64 {
         // NOTE: the `+ 1` is because the height of the first block of a reward cycle is mod 1, not
         // mod 0.
--- a/stackslib/src/chainstate/nakamoto/coordinator/mod.rs
+++ b/stackslib/src/chainstate/nakamoto/coordinator/mod.rs
@@ -61,11 +61,13 @@ impl<'a, T: BlockEventDispatcher> OnChainRewardSetProvider<'a, T> {
         sortdb: &SortitionDB,
         block_id: &StacksBlockId,
     ) -> Result<RewardSet, Error> {
+        // TODO: this method should read the .signers contract to get the reward set entries.
+        //   they will have been set via `NakamotoChainState::check_and_handle_prepare_phase_start()`.
         let cycle = burnchain
             .block_height_to_reward_cycle(cycle_start_burn_height)
             .expect("FATAL: no reward cycle for burn height");
 
-        let mut registered_addrs =
+        let registered_addrs =
             chainstate.get_reward_addresses_in_cycle(burnchain, sortdb, cycle, block_id)?;
 
         let liquid_ustx = chainstate.get_liquid_ustx(block_id);
@@ -194,9 +196,9 @@ pub fn get_nakamoto_reward_cycle_info<U: RewardSetProvider>(
           "reward_cycle_length" => burnchain.pox_constants.reward_cycle_length,
           "prepare_phase_length" => burnchain.pox_constants.prepare_length);
 
-    // find the last tenure-start Stacks block processed in the preceeding prepare phase
-    // (i.e. the first block in the tenure of the parent of the first Stacks block processed in the prepare phase).
-    // Note that we may not have processed it yet.  But, if we do find it, then it's
+    // Find the first Stacks block in this reward cycle's preceding prepare phase.
+    // This block will have invoked `.signers.stackerdb-set-signer-slots()` with the reward set.
+    // Note that we may not have processed it yet. But, if we do find it, then it's
     // unique (and since Nakamoto Stacks blocks are processed in order, the anchor block
     // cannot change later).
     let prepare_phase_sortitions =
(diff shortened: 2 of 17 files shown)


kantai:
With just a quick skim, I have some high-level feedback about the approach in this PR.

This PR sets the signing-key via a separate list (actually data-map index) from the reward set list. But unless I'm misunderstanding how signing keys and the PoX reward set should operate, these should be the same list: any entry in the PoX reward set should have a signing-key and any signing-key should be associated with a reward-set entry.

This also impacts the signing-set calculation method, because really this should be calculated via the same `get_reward_set` method that the PoX reward set construction uses (because the signer set and the reward set are more than related: they should have an exact correspondence). This is top of mind right now to me because the signer bitvec depends on this work, and so I implemented an initial version using mocked data inputs:

https://github.com/stacks-network/stacks-core/blob/feat/stacker-bitvec/stackslib/src/chainstate/stacks/boot/mod.rs#L612

See the note here:

https://github.com/stacks-network/stacks-core/blob/feat/stacker-bitvec/stackslib/src/chainstate/nakamoto/coordinator/mod.rs#L69

MarvinJanssen:
Thanks @kantai I appreciate you giving the background and linking those files. For this PR the scope was to introduce `.signers` and to be able to do a round trip of Stacking -> signer keys to PoX -> have the node write signer keys to `.signers` -> ability to pull and check it in unit tests.

As for the coupling of PoX reward slots and signer keys we have some open questions. (#4247 and others.) With the state of the PR it will be easy to move the signer keys elsewhere / handle them differently in the PoX contract. 

I'll update the OP to reflect the above.

MarvinJanssen:
Should I rename `.signers` to `.stackers`?

stackslib/src/burnchains/mod.rs
```diff
@@ -520,6 +520,17 @@ impl PoxConstants {
         (effective_height % u64::from(self.reward_cycle_length)) == 1
     }
 
+    pub fn is_prepare_phase_start(&self, first_block_height: u64, burn_height: u64) -> bool {
```
jcnelson, line 523: Pretty sure this is off-by-one.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1795,6 +1805,152 @@ impl NakamotoChainState {
...
 
+    pub fn calculate_signer_slots(
+        clarity: &mut ClarityTransactionConnection,
+        pox_constants: &PoxConstants,
+    ) -> Result<HashMap<Value, u128>, clarity::vm::errors::Error> {
+        let is_mainnet = clarity.is_mainnet();
+        let pox4_contract = &boot_code_id(POX_4_NAME, is_mainnet);
+        let reward_cycle = clarity
+            .eval_read_only(pox4_contract, &"(current-pox-reward-cycle)")
+            .unwrap()
```
jcnelson, line 1816: Nit: we shouldn't `unwrap()` here; we should return an error.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1795,6 +1805,152 @@ impl NakamotoChainState {
...
+            if let Ok(Value::Optional(entry)) = clarity.eval_read_only(
+                pox4_contract,
+                &format!("(get-signer-key u{} u{})", reward_cycle, index),
+            ) {
+                if let Some(data) = entry.data {
+                    let data = data.expect_tuple();
+                    let key = data.get("signer-key")?.to_owned();
+                    let amount = data.get("ustx")?.to_owned().expect_u128();
+                    let sum = signers.get(&key).cloned().unwrap_or_default();
+                    //TODO HashMap insert order is not guaranteed
```
jcnelson, line 1838: Consider using a `BTreeMap`

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1795,6 +1805,152 @@ impl NakamotoChainState {
...
     }
 
+    pub fn calculate_signer_slots(
+        clarity: &mut ClarityTransactionConnection,
+        pox_constants: &PoxConstants,
+    ) -> Result<HashMap<Value, u128>, clarity::vm::errors::Error> {
+        let is_mainnet = clarity.is_mainnet();
+        let pox4_contract = &boot_code_id(POX_4_NAME, is_mainnet);
+        let reward_cycle = clarity
+            .eval_read_only(pox4_contract, &"(current-pox-reward-cycle)")
```
jcnelson, line 1815: I'd prefer if you just passed `reward_cycle` as an argument.  The code that calls this -- the block-processing logic -- ought to be able to calculate the reward cycle from `pox_constants`.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1795,6 +1805,152 @@ impl NakamotoChainState {
...
+            .unwrap()
+            .expect_u128();
+        let list_length = clarity
+            .eval_read_only(
+                pox4_contract,
+                &format!("(get-signer-key-list-length u{})", reward_cycle),
+            )
+            .unwrap()
```
jcnelson, line 1823: Same -- please no `unwrap()`s.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1795,6 +1805,152 @@ impl NakamotoChainState {
         }
     }
 
+    pub fn calculate_signer_slots(
```
jcnelson, line 1808: Thinking about this a bit more, I don't think this method needs to exist.  Instead, the function `StacksChainState::make_reward_set()` (stackslib/src/chainstate/stacks/boot/mod.rs) should be updated to assign each reward set entry a signing key.  Furthermore, you'd want to add the signing key to `RawRewardSetEntry` included the signing key, such that `get_reward_addresses_pox_4()` loaded it out of the stacking record along with every other field.  From there, once you have the reward set, you'd simply copy the signing keys into the `.signers` contract.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -583,7 +606,10 @@
       (try! (can-stack-stx pox-addr amount-ustx first-reward-cycle lock-period))
 
       ;; ensure the signer key can be used
-      (try! (insert-signer-key signer-key))
+      (try! (insert-signer-key signer-key first-reward-cycle))
+
+      ;; update the total ustx for the signer key
+      (increment-signer-key-total-ustx first-reward-cycle signer-key amount-ustx)
```
jcnelson, line 612: You're going to have to do this for all of the reward cycles in which the user has stacked

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Yes, I agree, and I implemented a version of this (before this PR opened, otherwise, I probably would have waited) in my branch while working on the bitvector stuff:

https://github.com/stacks-network/stacks-core/blob/feat/stacker-bitvec/stackslib/src/chainstate/stacks/boot/mod.rs#L612

See the note here:

https://github.com/stacks-network/stacks-core/blob/feat/stacker-bitvec/stackslib/src/chainstate/nakamoto/coordinator/mod.rs#L69

kantai: `make_signer_set` is invoked by `make_reward_set`, and the result is included in the `RewardSet` struct (so that it can be stored in the SortitionDB and emitted to event observers).

MarvinJanssen:
@kantai I rebased this onto your branch and filled in the missing pieces and finished the PoX changes. My basic unit test works. There is a bit of code repetition. Let me know if this is the direction you are thinking of and I can clean it up and add some more tests to get it merged. For now I'm also targeting your branch, if yours gets merged first then I will rebase off `next.`

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -444,12 +450,13 @@
 ;; A PoX address can be added to at most 12 consecutive cycles.
 ;; No checking is done.
 (define-private (add-pox-partial-stacked (pox-addr (tuple (version (buff 1)) (hashbytes (buff 32))))
+                                         (signer principal)
```
friedger, line 453: Why does the partial stacked state need to know about the signer?

Thinking about pooled stacking, it makes sense to provide the key only on aggregate contract calls. The voting slots are based on aggregate values as well.

kantai: Yeah -- I agree with this. Partial stacking state shouldn't have the signer, the signer should be set on aggregate-commit. The `stacking-state` map also should not have a `signer-key` field.

So I'd recommend:

1. Removing `signer-key` from `stacking-state`
2. Removing `signer` from the `partial-stacked-by-cycle`
3. Add `signer-key` argument to the `aggregate-commit` method.

Otherwise, the changes in pox-4 look good to me.

setzeus: Copy. Wrote similar thoughts here #4247 just a few mins ago. Assigning to here / @MarvinJanssen.

MarvinJanssen: Yep we were going back and forth on it. Will change it soon. Thanks @friedger

MarvinJanssen: Edit: is having the signer key in stacking state not useful to have?

hstove: So is the plan to use `(signer principal)` or `(signer-key (buff 33))`?

jcnelson: The arguments to stacking are `(signer-key (buff 33))`, but the internal data maps store `principal`. The stacks functions will need to compute the principal from the signer key.

kantai:
Okay, my last push contained a handful of things that I'd appreciate some feedback on (some of these need @jcnelson's eyes):

1. The choice of when to apply the `.signers` change is really important. This has to be strictly defined, and it can't be a single burn height. My recent commit sets it as "if we're in a prepare phase, and the update hasn't been applied for the target reward cycle, apply it". That means it is the "first block of the prepare phase", but it might be the second burn block of the prepare phase if the first burn block didn't contain a sortition.
2. Because of that constraint, the PoX anchor block selection in Nakamoto needs to be the same block or later. My commit updates the PoX anchor block selection to be the first block evaluated in a prepare phase.
3. In testing this PR, it surfaced a bug in the nakamoto miner: it was loading `miner_tenure_info` from the SortitionDB's canonical chain tip, but setting `header.consensus_hash` to be equal to a passed-in value. This PR changes that behavior so that `miner_tenure_info` is loading information based on that `consensus_hash`.

Once I've got an okay on the above, I'll wrap this PR up by applying the last pox-4 changes discussed above (i.e., removing the `signing-key` from the `stacker-state` because its duplicated information from the reward set list). I'd like to leave out a `TODO` item from this PR:

```rust
    pub fn get_reward_set_nakamoto(
        &self,
        cycle_start_burn_height: u64,
        burnchain: &Burnchain,
        sortdb: &SortitionDB,
        block_id: &StacksBlockId,
    ) -> Result<RewardSet, Error> {
        // TODO: this method should read the .signers contract to get the reward set entries.
        //   they will have been set via `NakamotoChainState::check_and_handle_prepare_phase_start()`.
        let cycle = burnchain
```

Because implementing that will take a bit of time, and its not needed for unblocking the signer-miner block proposal work. So, I'd rather take care of it in a separate PR.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -227,125 +217,99 @@ pub fn get_nakamoto_reward_cycle_info<U: RewardSetProvider>(
...
+    let rc_info = RewardCycleInfo {
+        reward_cycle,
+        anchor_status,
+    };
 
-        return Ok(Some(rc_info));
-    }
+    // persist this
+    let mut tx = sort_db.tx_begin()?;
+    SortitionDB::store_preprocessed_reward_set(&mut tx, &first_sortition_id, &rc_info)?;
```
jcnelson, line 309: Don't we want to persist this first and foremost in the Stacks chainstate?  If so, then would it make more sense to just load the persisted reward set out of the chainstate, instead of recalculating it?

stackslib/src/chainstate/nakamoto/miner.rs
```diff
@@ -224,11 +224,17 @@ impl NakamotoBlockBuilder {
     ) -> Result<MinerTenureInfo<'a>, Error> {
         debug!("Nakamoto miner tenure begin");
 
-        let burn_tip = SortitionDB::get_canonical_chain_tip_bhh(burn_dbconn.conn())?;
-        let burn_tip_height = u32::try_from(
-            SortitionDB::get_canonical_burn_chain_tip(burn_dbconn.conn())?.block_height,
-        )
-        .expect("block height overflow");
+        // must build off of the header's consensus hash as the burnchain view, not the canonical_tip_bhh:
```
jcnelson, line 227: Good catch!

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1806,6 +1813,261 @@ impl NakamotoChainState {
...
+        first_block_height: u64,
+        pox_constants: &PoxConstants,
+        burn_tip_height: u64,
+    ) -> Result<Vec<StacksTransactionEvent>, ChainstateError> {
+        let current_epoch = clarity_tx.get_epoch();
+        if current_epoch < StacksEpochId::Epoch25 {
+            // before Epoch-2.5, no need for special handling
+            return Ok(vec![]);
+        }
+        // now, determine if we are in a prepare phase, and we are the first
```
jcnelson, line 2002: Just thinking a little bit ahead here -- would it be acceptable to relax this constraint to "in or _after_ this last prepare phase?"  Asking because in the (unlikely but possible) event that the prepare phase is devoid of blocks, I think we'd want the signer set calculation to still go through during the subsequent reward phase.  Perhaps the code here could check the last-processed reward cycle against the current reward cycle, and proceed if either we're in the prepare phase for the next cycle or we're in a subsequent cycle altogether?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1806,6 +1813,261 @@ impl NakamotoChainState {
...
+
+        info!(
+            "Performing .signers state update";
+            "burn_height" => burn_tip_height,
+            "for_cycle" => cycle_of_prepare_phase,
+            "signers_contract" => %signers_contract,
+        );
+
+        clarity_tx.connection().as_free_transaction(|clarity| {
+            Self::handle_signer_stackerdb_update(
```
jcnelson, line 2062: Don't we want to store the whole reward set?  I think the signer needs it (cc @jferrant)

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -155,9 +164,23 @@ pub struct RawRewardSetEntry {
     pub reward_address: PoxAddress,
     pub amount_stacked: u128,
     pub stacker: Option<PrincipalData>,
-    pub signing_key: Option<PrincipalData>,
+    pub signer: Option<PrincipalData>,
 }
 
+/// This enum captures the names of the PoX contracts by version.
+// This should deprecate the const values `POX_version_NAME`, but
```
jcnelson, line 171: Can you open a ticket for this?

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -624,7 +647,7 @@ impl StacksChainState {
 
         let mut signer_set = BTreeMap::new();
         for entry in entries.iter() {
-            let signing_key = if let Some(PrincipalData::Standard(s)) = entry.signing_key.clone() {
+            let signing_key = if let Some(PrincipalData::Standard(s)) = entry.signer.clone() {
```
jcnelson, line 650: This wouldn't be a `signing_key` anymore ;)  Maybe a better variable name like `signing_addr`?

stackslib/src/chainstate/stacks/boot/pox-4.clar
jcnelson: > Yeah -- I agree with this. Partial stacking state shouldn't have the signer, the signer should be set on aggregate-commit. The `stacking-state` map also should not have a `signer-key` field.
> 
> So I'd recommend:
> 
> 1. Removing `signer-key` from `stacking-state`
> 2. Removing `signer` from the `partial-stacked-by-cycle`
> 3. Add `signer-key` argument to the `aggregate-commit` method.
> 
> Otherwise, the changes in pox-4 look good to me.

@kantai Are you going to make this happen?

stackslib/src/chainstate/stacks/boot/signers.clar
```diff
@@ -0,0 +1,22 @@
...
+		(print signer-slots)
+        (var-set last-set-cycle reward-cycle)
+		(ok (var-set stackerdb-signer-slots signer-slots))))
+
+(define-read-only (stackerdb-get-signer-slots)
+	(ok (var-get stackerdb-signer-slots)))
+
+(define-read-only (stackerdb-get-config)
+	(ok
+		{ chunk-size: u4096,
```
jcnelson, line 17: `chunk-size` should be about 1 MB (cc @xoloki -- how big can DKG / signing round messages be?  We can go up to 16 MB).

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
kantai: Yeah -- that's more or less the `TODO` item I've left open. My current understanding of how things should work:

1. `.signers` update occurs in the setup step of the first block in a prepare phase. This update writes the signers stackerdb data (currently implemented), _and_ the reward set state to `.signers` (not implemented).
2. `get_reward_set_nakamoto()` reads the reward set state directly from `.signers`, performing no calculations: its just essentially loading a vec.

This code you commented on would stay the same: the "reward set provider" is still invoked, just what it is doing is different. And the first time its invoked, it stashes its results in the SortitionDB as well so that it can be read directly from there.

The way it works now is that (1) and (2) independently calculate the reward set. That's not tenable because it introduces the possibility of different results even without a bug: one calculates at the start of the block and the other calculates at the end of the same block. But, I do want to leave it as a TODO, because I can handle it in a follow-up PR while unblocking the signer/miner communication work.

jcnelson: Ah I see you have that commented above.  Can you please open a ticket for it?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1806,6 +1813,261 @@ impl NakamotoChainState {
...
+        let signers_contract = &boot_code_id(SIGNERS_NAME, is_mainnet);
+
+        let liquid_ustx = clarity.with_clarity_db_readonly(|db| db.get_total_liquid_ustx());
+        let reward_slots = Self::get_reward_slots(clarity, reward_cycle, pox_contract)?;
+        let (threshold, _participation) = StacksChainState::get_reward_threshold_and_participation(
+            &pox_constants,
+            &reward_slots[..],
+            liquid_ustx,
+        );
+        let reward_set =
```
jcnelson, line 1923: Don't we want to store this in the contract?

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Replied above -- the signer will get the reward set from the event observer (as implemented in the PR this builds on). The stackerdb is updated here, which should be sufficient for unblock the miner/signer comms.

I agree, that eventually, the whole thing needs to be written, but I think that should be handled in a separate PR so that this can merge more quickly, and we're not excessively stacking PRs.

stackslib/src/chainstate/stacks/boot/mod.rs
kantai: Sure

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
jcnelson: Lol we both commented at the same time.

Agreed with the above.  Yeah, a follow-on PR is fine.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Replied elsewhere

stackslib/src/chainstate/nakamoto/miner.rs
MarvinJanssen: This was probably what caused my test to be flaky.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -2548,6 +2810,16 @@ impl NakamotoChainState {
             );
         }
 
+        // Handle signer stackerdb updates
+        if evaluated_epoch >= StacksEpochId::Epoch25 {
+            let _events = Self::check_and_handle_prepare_phase_start(
```
MarvinJanssen, line 2815: I wasn't sure if we needed to propagate these events and if so, where?

stackslib/src/chainstate/stacks/boot/signers.clar
```diff
@@ -0,0 +1,22 @@
+(define-data-var last-set-cycle uint u0)
+(define-data-var stackerdb-signer-slots (list 4000 { signer: principal, num-slots: uint }) (list))
+
+(define-private (stackerdb-set-signer-slots 
+                   (signer-slots (list 4000 { signer: principal, num-slots: uint }))
+                   (reward-cycle uint))
+	(begin
+		(print signer-slots)
```
MarvinJanssen, line 8: Just FYI that as far as I know, no (print) events were established. I added this print myself so it might not be necessary.

stackslib/src/chainstate/stacks/boot/pox-4.clar
jcnelson: @kantai @MarvinJanssen @hstove per @jferrant, we actually need to store the signing keys to the stacker map after all.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: Yeah, I think so, but I'll start an issue for it -- I would want to separate it from the upstream PR here.

stackslib/src/chainstate/stacks/boot/signers.clar
```diff
@@ -0,0 +1,22 @@
...
+		(ok (var-set stackerdb-signer-slots signer-slots))))
+
+(define-read-only (stackerdb-get-signer-slots)
+	(ok (var-get stackerdb-signer-slots)))
+
+(define-read-only (stackerdb-get-config)
+	(ok
+		{ chunk-size: u4096,
+		  write-freq: u0,
+		  max-writes: u4096,
```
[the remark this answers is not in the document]
kantai: Updated in 5510b1769 in the upstream branch.

stackslib/src/chainstate/stacks/boot/signers.clar
kantai: Updated in 5510b1769 (set to 2MB) in the upstream branch.

stackslib/src/chainstate/stacks/boot/signers.clar
kantai: Removed for now

stackslib/src/chainstate/nakamoto/mod.rs
kantai: For now, I think it's fine to not propagate any events. If we want events here for the dispatcher, we can figure that out later.

PR #4266, opened 2024-01-19
feat: default to `.signers` boot contract in signer config

hstove:
This small change makes the `stackerdb_contract_id` config variable for the signer an optional field. When it isn't provided, it defaults to `BOOT_CODE_ADDR.signers`.

Although all testing (unit and functional) is currently done with "mock" contracts, it might make sense to wait for #4263 before merging. We can also choose to merge this into that PR.

### DIFF
--- a/stacks-signer/src/config.rs
+++ b/stacks-signer/src/config.rs
@@ -356,6 +356,11 @@ mod tests {
             signers: vec![],
             signer_id: 0,
             event_timeout_ms: None,
+            dkg_end_timeout_ms: None,
+            dkg_public_timeout_ms: None,
+            dkg_private_timeout_ms: None,
+            nonce_timeout_ms: None,
+            sign_timeout_ms: None,
         };
         overrides(&mut config);
         config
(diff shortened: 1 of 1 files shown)


jferrant (approved):

kantai (approved):

stacks-signer/src/config.rs
```diff
@@ -228,13 +228,12 @@ impl TryFrom<RawConfigFile> for Config {
...
-                ConfigError::BadField(
-                    "stackerdb_contract_id".to_string(),
-                    raw_data.stackerdb_contract_id,
-                )
-            })?;
+        let stackerdb_contract_id = match raw_data.stackerdb_contract_id {
+            Some(id) => QualifiedContractIdentifier::parse(&id).map_err(|_| {
+                ConfigError::BadField("stackerdb_contract_id".to_string(), id.clone())
+            })?,
+            None => boot_code_id("signers", raw_data.network == Network::Mainnet),
```
friedger, line 235: Is it an issue to run the signer against testnet?

hstove: Nope, this is just setting the boot code address depending on the signer's config

PR #4267, opened 2024-01-19
Chore/update wsts version 7

jferrant:
### Description
Update wsts version to 7 to enable DKG robustness and voting no on a block

NOTE: also fixes slowdown due to double wait for command events using the poll_timeout

### DIFF
--- a/stacks-signer/src/config.rs
+++ b/stacks-signer/src/config.rs
@@ -36,7 +36,7 @@ use wsts::state_machine::PublicKeys;
 /// List of key_ids for each signer_id
 pub type SignerKeyIds = HashMap<u32, Vec<u32>>;
 
-const EVENT_TIMEOUT_MS: u64 = 5000;
+const EVENT_TIMEOUT_MS: u64 = 50;
 
 #[derive(thiserror::Error, Debug)]
 /// An error occurred parsing the provided configuration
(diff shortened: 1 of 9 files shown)


xoloki (approved):
LGTM thanks!

kantai (approved):
LGTM!

stacks-signer/src/config.rs
```diff
@@ -36,7 +36,7 @@ use wsts::state_machine::PublicKeys;
 /// List of key_ids for each signer_id
 pub type SignerKeyIds = HashMap<u32, Vec<u32>>;
 
-const EVENT_TIMEOUT_MS: u64 = 5000;
+const EVENT_TIMEOUT_MS: u64 = 50;
```
jcnelson, line 39: That's a little small, don't you think?

jcnelson (approved):

jcnelson:
Just had one question about the timeout -- 50ms is a little small for network latency across the broad internet.

stacks-signer/src/config.rs
xoloki: The issue here is that the libsigner runloop has to multiplex two event receivers: one with stackerdb events, the other with main loop commands.  So every time we would get an event without a command, we'd wait 5 seconds for a command that never came.

I'm still not exactly sure *why* this was suddenly a problem.  `wsts v7` added a `DkgEndBegin` message, which is being amplified by all of the inactive coordinators, so presumably that was it.  But I have no idea why we weren't often hitting this issue, since we don't send many commands but received lots of messages even before this change.

Regardless, this was always broken before, so the fix seems appropriate.

jferrant: I think once we remove the command receiver, we can update it back to 5000 ms. However, for now I think this fix is okay. If we find we need to KEEP the command receiver we should have a sep timeout for it :)

PR #4268, opened 2024-01-19
feat: include stacks-signer binary in CI images

hstove:
This PR adds the `stacks-signer` binary to Docker images that are built through CI. In each of these cases, the binary was already being built, but the `stacks-signer` binary wasn't copied into the final image. There are other Dockerfiles (like `./Dockerfile`) that only build the `stacks-node` binary, and not the whole workspace, so I didn't modify those files.

### DIFF
--- a/.github/actions/dockerfiles/Dockerfile.debian-source
+++ b/.github/actions/dockerfiles/Dockerfile.debian-source
@@ -12,7 +12,7 @@ COPY . .
 RUN apt-get update && apt-get install -y git libclang-dev
 
 # Run all the build steps in ramdisk in an attempt to speed things up
-RUN target=${BUILD_DIR} cp -R /src/. ${BUILD_DIR}/ \
+RUN --mount=type=tmpfs,target=${BUILD_DIR} cp -R /src/. ${BUILD_DIR}/ \
     && cd ${BUILD_DIR} \
     && rustup target add ${TARGET} \
     && rustup component add rustfmt \
@@ -21,5 +21,5 @@ RUN target=${BUILD_DIR} cp -R /src/. ${BUILD_DIR}/ \
     && cp -R ${BUILD_DIR}/target/${TARGET}/release/. /out
 
 FROM --platform=${TARGETPLATFORM} debian:bookworm
-COPY --from=build /out/stacks-node /bin/
+COPY --from=build /out/stacks-node /out/stacks-signer /bin/
 CMD ["stacks-node", "mainnet"]
--- a/build-scripts/Dockerfile.linux-glibc-arm64
+++ b/build-scripts/Dockerfile.linux-glibc-arm64
@@ -23,4 +23,4 @@ RUN --mount=type=tmpfs,target=${BUILD_DIR} cp -R /src/. ${BUILD_DIR}/ \
     && cp -R ${BUILD_DIR}/target/${TARGET}/release/. /out
 
 FROM scratch AS export-stage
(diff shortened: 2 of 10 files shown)


kantai (approved):

wileyj:
looks good at a glance, i'm testing the build locally to be sure it's working as expected (noticed the docker login error in the upstream action). 
one question @hstove  - if the build here is changing anyways on source builds, is there any reason to not include the other binaries that are being produced?

with the current build steps, adding those extra binaries would have required more build time and several extra steps. but if we're moving to that anyways to get the stacks-signer binary, the only reason i can think to not include the other binaries (i.e. `stacks-inspect`) is the resulting image size.

wileyj:
> looks good at a glance, i'm testing the build locally to be sure it's working as expected (noticed the docker login error in the upstream action). one question @hstove - if the build here is changing anyways on source builds, is there any reason to not include the other binaries that are being produced?
> 
> with the current build steps, adding those extra binaries would have required more build time and several extra steps. but if we're moving to that anyways to get the stacks-signer binary, the only reason i can think to not include the other binaries (i.e. `stacks-inspect`) is the resulting image size.
> 
> https://github.com/stacks-network/stacks-core/actions/runs/7590281693/job/20751727276?pr=4268

on second thought, the binaries all are quite large. so the debian image nearly doubles by adding the additional binaries:
image size from PR: `880MB`
image size copying all binares: `1.69GB`

wileyj (approved):
shipit

hstove:
Interesting, that's a sizable increase in image size. I can look into seeing how much the signer itself adds to the image, because that's the main thing we're trying to get here. It's definitely worth considering how we build release images, and I could see it making more sense to do separate images for that.

wileyj:
> Interesting, that's a sizable increase in image size. I can look into seeing how much the signer itself adds to the image, because that's the main thing we're trying to get here. It's definitely worth considering how we build release images, and I could see it making more sense to do separate images for that.

when i looked into the image i built from your dockerfile with all the binaries added - each one ranged from around 200-350MB. so, the size increase definitely makes sense. 
however, the sizes themselves on this specific image definitely seem really large. will have to dig into this a bit to see if there are any things we can do to reduce (for comparison, building the binaries for release produces much smaller binaries). 
i'm inclined to think this is still ok though - the purpose of this image is to confirm the build works, and provide a docker image to test PR changes on the network. 
i think for now -  create an issue so we can look into this later (assign to me is fine) and merge this  pr so it unblocks others

hstove:
Sounds good, I opened that ticket here: #4278

hstove:
FYI CI is not pushing Docker images from this PR because it's from my fork. Now that I can push branches I can open a new PR if it's desired to test the pushed images. JWiley and I have manually built and inspected the Docker images, so that might not be necessary, but I'm happy to open a new PR if we want to.

wileyj:
> @hstove I'd prefer to keep this PR scoped to only the debian image built from source for now until #4278 is resolved.
> 
> is there a reason those files were changed, or just trying to get ahead of things?

to be clear, the changes added in [38e58aea53d40ffa03e3b97f59e43c28d27b1d58 ](https://github.com/stacks-network/stacks-core/commit/38e58aea53d40ffa03e3b97f59e43c28d27b1d58) only affect release builds, which won't be triggered in the target branch of next for this PR.

hstove:
@wileyj no problem, makes sense! The PR is now only modifying the `debian-source` image.

wileyj (approved):

Issue #4256, opened 2024-01-18
[Nakamoto] Emit list of signers and voting power when PoX anchor block is set 


zone117x:
Related to signer support in the API & Explorer https://github.com/hirosystems/explorer/issues/1248

The API and Explorer need to know the list of signers and voting power, including the percentage and cumulative stake by Signer.  Ideally, this would be calculated and emitted when the PoX anchor block is set. Ideally it would be part of the existing `/new_block` payload which is currently the only event that doesn't have race-conditions / out-of-order threading issues.

zone117x:
@kantai (same with other issue) does this sound about right, and is it okay if I assign this to you?

Issue #4255, opened 2024-01-18
[Nakamoto] Emit amount stacked for the subsequent cycle when the PoX anchor block is set


zone117x:
Related to signer support in the API & Explorer https://github.com/hirosystems/explorer/issues/1248

The API and Explorer need to know the total STX amount for PoX cycles. Ideally, this would be calculated and emitted when the PoX anchor block is set. Ideally it would be part of the existing `/new_block` payload which is currently the only event that doesn't have race-conditions / out-of-order threading issues.

zone117x:
@kantai does this sound about right, and is it okay if I assign this to you?

kantai:
Yeah -- that's fine. What you're looking for is basically the complete stacker set, right? This is event data that is needed by the signer binary anyways. I'm not sure that it'll be possible to include in `/new_block`, because the selection of the reward set doesn't occur during the processing of a block.

zone117x:
> What you're looking for is basically the complete stacker set, right?

I think this issue is probably just for aggregate STX amount, and more complete data is covered by https://github.com/stacks-network/stacks-core/issues/4256 and https://github.com/stacks-network/stacks-core/issues/4257 -- does that make sense?

> I'm not sure that it'll be possible to include in /new_block

Ah okay, is there a way to ensure this new event doesn't have race-conditions against the `/new_block`? Almost all other events can come in out-of-order (in relation to each and/or `/new_block`) -- the API has different approaches to handle different events like A) complex detection and reconciliation code, B) presenting potentially inconsistent data to clients, C) dropping the event because it's too complex/expensive/damaging to do otherwise.

Issue #4241, opened 2024-01-12
Tracking stacker-signer participation in block signing


kantai:
Per [SIP-021: Signer Liveness Enforcement](https://github.com/stacksgov/sips/blob/feat/sip-021-nakamoto/sips/sip-021/sip-021-nakamoto.md#signer-liveness-enforcement), the `stacks-node` must have a way to validate which stackers signed a block.

The rough design discussed for this involved a bitvector in the block header which would be included in the hash that stackers sign (but not the miner, because the miner cannot know which stackers will participate when the miner signs the block).

But there's several elements that need to implemented (beyond figuring out who to include or not include in the bitvector, and what the punishment for non-participation is):

1. Construction of the bitvector itself: the order and length of this bitvector is consensus-critical, and so it must be a deterministic function of the stacker set (e.g., the lexicographically sorted list of stacker keys).
2. Inclusion of the bitvector in the signer's signature
3. Inclusion of the bitvector in the Nakamoto block header and the stacker signature verification routine
4. Code in the signer-binary operating as the coordinator which constructs the bitvector at the start of a signing round
5. Code in the signer-binary operating as a signer which only signs if the bitvector includes themselves (note: what if a coordinator consistently doesn't include a signer -- does the signer have any recourse?)

kantai:
Tagging @jcnelson, @xoloki, and @jferrant because I think this is pretty relevant to the workstreams you are on.

PR #4269, opened 2024-01-21
Nakamoto: Add stacker/signer bit vector and event

kantai:
### Description

This PR implements a handful of things:

1. It adds a `BitVec` class to stacks-common. This PR is at least the third implementation of a bit vector in the codebase, so it seemed like adding a simple data struct would be useful here.
2. Adds a `BitVec` field to the `NakamotoBlockHeader`, this is included in the consensus serialization and the stacker/signer signature hash. It is _not_ included in the miner's signature hash.
3. Adds a `new_pox_set` endpoint for the event dispatcher. This endpoint announces when a reward set is calculated in a PoX anchor block. It relays information about the reward set:
      * The cycle number of the set
      * The StacksBlockId of the anchor block
      * The canonically ordered list of reward addresses (with repetitions for addresses with multiple slots)
      * The canonically ordered list of signers, each entry containing a signing address (C32-encoded), the number of signing slots, and the amount stacked. This canonical ordering is what will be used for entries in the bit vector.
4. It implements `make_signer_set`, which is invoked by `make_reward_set` to translate the list of `RawRewardEntry`s into signing set entries. Currently, it does this by summing all entries with the same public key and assigning slots based on the slot thresholds.
5. It mocks data for the signing key in raw reward entries in `nakamoto_get_reward_set()`
6. It alters the logic deciding between using `nakamoto` reward set fetching and `2.x`: any reward set that is used by Nakamoto should be fetched via the `nakamoto` reward set endpoint. The first nakamoto reward set _will_ occur in epoch 2.5.
7. It fixes a broken integration test `correct_burn_outs` -- this integration test wasn't asserting that the sortition db advanced, so it missed the fact that the recent update in `next` to require signing keys (and unique ones) broke the test. The test now asserts the sortition db advances _and_ it works.
8. It update the `correct_burn_outs` test to check the results of the `new_pox_set` event observer API are correct.

### Still needs test coverage

- `make_signer_set` needs unit testing. This is easy enough to add and I will add it before this merges, but before I did that, I wanted to get feedback on this PR because if `make_signer_set` needs to change, it seems like a waste to write the unit tests now.

### Relevant issues

#4241 - this is the `stacks-node` side of tracking. this PR doesn't implement the punishment for non-participation, though.
#4256 - this PR emits this list
#4255 - summing over the signer set entries in this PR gives the amount of active lockup. the amount of total lockup (which could include people who miss the threshold) would need to be added as another field in the return. This should be simple to add and I can do it if its useful (@zone117x).

### DIFF
--- a/stackslib/src/chainstate/coordinator/mod.rs
+++ b/stackslib/src/chainstate/coordinator/mod.rs
@@ -307,16 +307,25 @@ impl<'a, T: BlockEventDispatcher> RewardSetProvider for OnChainRewardSetProvider
     ) -> Result<RewardSet, Error> {
         let cur_epoch = SortitionDB::get_stacks_epoch(sortdb.conn(), cycle_start_burn_height)?
             .expect(&format!(
-                "FATAL: no epoch for burn height {}",
-                cycle_start_burn_height
+                "FATAL: no epoch for burn height {cycle_start_burn_height}",
             ));
-        // TODO: should Epoch-2.5 be included in `get_reward_set_nakamoto()`?
-        // The differences are:
-        //     (a) no minimum participation threshold (I think this *is* important)
-        //     (b) panicking assertion if there are no signing-keys set
-        // Apart from (a), this shouldn't matter: the signing-keys are always set whenever
-        //  the reward set is loaded from pox-4.
-        let reward_set = if cur_epoch.epoch_id < StacksEpochId::Epoch30 {
+        let cycle = burnchain
+            .block_height_to_reward_cycle(cycle_start_burn_height)
+            .expect("FATAL: no reward cycle for burn height");
+        let is_nakamoto_reward_set = match SortitionDB::get_stacks_epoch_by_epoch_id(
+            sortdb.conn(),
+            &StacksEpochId::Epoch30,
+        )? {
+            Some(epoch_30) => {
+                let first_nakamoto_cycle = burnchain
+                    .block_height_to_reward_cycle(epoch_30.start_height)
+                    .expect("FATAL: no reward cycle for burn height");
+                first_nakamoto_cycle <= cycle
+            }
+            // if epoch-3.0 isn't defined, then never use a nakamoto reward set.
+            None => false,
+        };
+        let reward_set = if !is_nakamoto_reward_set {
             // Stacks 2.x epoch
             self.get_reward_set_epoch2(
                 cycle_start_burn_height,
@@ -338,9 +347,6 @@ impl<'a, T: BlockEventDispatcher> RewardSetProvider for OnChainRewardSetProvider
         };
 
         if let Some(dispatcher) = self.0 {
-            let cycle = burnchain
-                .block_height_to_reward_cycle(cycle_start_burn_height)
-                .expect("FATAL: no reward cycle for burn height");
             dispatcher.announce_reward_set(&reward_set, block_id, cycle);
         }
 
--- a/stackslib/src/chainstate/coordinator/mod.rs
+++ b/stackslib/src/chainstate/coordinator/mod.rs
@@ -33,6 +33,7 @@ use stacks_common::util::get_epoch_time_secs;
 
 pub use self::comm::CoordinatorCommunication;
 use super::stacks::boot::RewardSet;
+use super::stacks::db::blocks::DummyEventDispatcher;
 use crate::burnchains::affirmation::{AffirmationMap, AffirmationMapEntry};
 use crate::burnchains::bitcoin::indexer::BitcoinIndexer;
 use crate::burnchains::db::{
@@ -189,6 +190,13 @@ pub trait BlockEventDispatcher {
         burns: u64,
         reward_recipients: Vec<PoxAddress>,
     );
+
+    fn announce_reward_set(
+        &self,
+        reward_set: &RewardSet,
+        block_id: &StacksBlockId,
+        cycle_number: u64,
+    );
 }
 
 pub struct ChainsCoordinatorConfig {
@@ -280,9 +288,15 @@ pub trait RewardSetProvider {
     ) -> Result<RewardSet, Error>;
 }
 
-pub struct OnChainRewardSetProvider();
+pub struct OnChainRewardSetProvider<'a, T: BlockEventDispatcher>(pub Option<&'a T>);
 
-impl RewardSetProvider for OnChainRewardSetProvider {
+impl OnChainRewardSetProvider<'static, DummyEventDispatcher> {
+    pub fn new() -> Self {
+        Self(None)
+    }
+}
+
+impl<'a, T: BlockEventDispatcher> RewardSetProvider for OnChainRewardSetProvider<'a, T> {
     fn get_reward_set(
         &self,
         cycle_start_burn_height: u64,
@@ -296,30 +310,45 @@ impl RewardSetProvider for OnChainRewardSetProvider {
                 "FATAL: no epoch for burn height {}",
                 cycle_start_burn_height
             ));
-        if cur_epoch.epoch_id < StacksEpochId::Epoch30 {
+        // TODO: should Epoch-2.5 be included in `get_reward_set_nakamoto()`?
+        // The differences are:
+        //     (a) no minimum participation threshold (I think this *is* important)
+        //     (b) panicking assertion if there are no signing-keys set
+        // Apart from (a), this shouldn't matter: the signing-keys are always set whenever
+        //  the reward set is loaded from pox-4.
+        let reward_set = if cur_epoch.epoch_id < StacksEpochId::Epoch30 {
             // Stacks 2.x epoch
-            return self.get_reward_set_epoch2(
+            self.get_reward_set_epoch2(
                 cycle_start_burn_height,
                 chainstate,
                 burnchain,
                 sortdb,
                 block_id,
                 cur_epoch,
-            );
+            )?
         } else {
             // Nakamoto epoch
-            return self.get_reward_set_nakamoto(
+            self.get_reward_set_nakamoto(
                 cycle_start_burn_height,
                 chainstate,
                 burnchain,
                 sortdb,
                 block_id,
-            );
+            )?
+        };
+
+        if let Some(dispatcher) = self.0 {
+            let cycle = burnchain
+                .block_height_to_reward_cycle(cycle_start_burn_height)
+                .expect("FATAL: no reward cycle for burn height");
+            dispatcher.announce_reward_set(&reward_set, block_id, cycle);
         }
+
+        Ok(reward_set)
     }
 }
 
-impl OnChainRewardSetProvider {
+impl<'a, T: BlockEventDispatcher> OnChainRewardSetProvider<'a, T> {
     fn get_reward_set_epoch2(
         &self,
         // Todo: `current_burn_height` is a misleading name: should be the `cycle_start_burn_height`
@@ -417,13 +446,22 @@ impl<
         CE: CostEstimator + ?Sized,
         FE: FeeEstimator + ?Sized,
         B: BurnchainHeaderReader,
-    > ChainsCoordinator<'a, T, ArcCounterCoordinatorNotices, OnChainRewardSetProvider, CE, FE, B>
+    >
+    ChainsCoordinator<
+        'a,
+        T,
+        ArcCounterCoordinatorNotices,
+        OnChainRewardSetProvider<'a, T>,
+        CE,
+        FE,
+        B,
+    >
 {
     pub fn run(
         config: ChainsCoordinatorConfig,
         chain_state_db: StacksChainState,
         burnchain: Burnchain,
-        dispatcher: &'a mut T,
+        dispatcher: &'a T,
         comms: CoordinatorReceivers,
         atlas_config: AtlasConfig,
         cost_estimator: Option<&'a mut CE>,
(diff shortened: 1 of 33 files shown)


stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -63,9 +65,21 @@ impl OnChainRewardSetProvider {
             .block_height_to_reward_cycle(cycle_start_burn_height)
             .expect("FATAL: no reward cycle for burn height");
 
-        let registered_addrs =
+        let mut registered_addrs =
             chainstate.get_reward_addresses_in_cycle(burnchain, sortdb, cycle, block_id)?;
 
+        // TODO (pox-4-workstream): the pox-4 contract must be able to return signing keys
```
jcnelson, line 71: Is there an issue for this?  Because this seems to be pretty high-priority!

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -93,11 +110,13 @@ impl OnChainRewardSetProvider {
...
 
-        Ok(StacksChainState::make_reward_set(
-            threshold,
-            registered_addrs,
-            cur_epoch.epoch_id,
-        ))
+        let reward_set =
+            StacksChainState::make_reward_set(threshold, registered_addrs, cur_epoch.epoch_id);
+        if reward_set.signers.is_none() {
+            error!("FATAL: PoX reward set did not specify signer set in Nakamoto");
```
jcnelson, line 116: Is this truly fatal?  If so, then should it panic?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -313,6 +316,8 @@ pub struct NakamotoBlockHeader {
     pub miner_signature: MessageSignature,
     /// Schnorr signature over the block header from the signer set active during the tenure.
     pub signer_signature: ThresholdSignature,
+    /// A bitvec which represents the signers that participated in this block signature.
+    pub signer_bitvec: BitVec,
```
jcnelson, line 320: I'm a little skeptical that we should be using the generic `BitVec` here.  Yes, we need a bitmap of some kind, but I think we also need a hard bound on how big this is allowed to be (e.g. there can be no more than 4000 bits).  And, I think that bound needs to be expressed in the data type itself, so it's impossible to represent a block header that would exceed this bound (and thus it's impossible to both serialize and deserialize such a block).

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -575,6 +603,63 @@ impl StacksChainState {
...
+        if !expects_signing_keys {
+            return None;
+        }
+
+        let mut signer_set = BTreeMap::new();
+        for entry in entries.iter() {
+            let signing_key = if let Some(PrincipalData::Standard(s)) = entry.signing_key.clone() {
+                StacksAddress::from(s)
+            } else {
+                // TODO: should figure out if in mainnet?
```
jcnelson, line 630: Yes

testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -976,6 +992,34 @@ fn correct_burn_outs() {
...
+    sorted_stacker_sets.sort_by_key(|(_block_id, cycle_num, _reward_set)| *cycle_num);
+    assert_eq!(
+        sorted_stacker_sets, stacker_sets,
+        "Stacker set should be sorted by cycle number already"
+    );
+
+    let first_epoch_3_cycle = burnchain
+        .block_height_to_reward_cycle(epoch_3.start_height)
+        .unwrap();
+    for (_, cycle_number, reward_set) in stacker_sets.iter() {
```
jcnelson, line 1007: Any way to expand this to e.g. 10 stackers, some of which don't make the stacking minimum?

jcnelson (commented):
I think this is on the right track.  Just a few comments.

stackslib/src/chainstate/coordinator/mod.rs
```diff
@@ -293,33 +307,54 @@ impl RewardSetProvider for OnChainRewardSetProvider {
...
             .expect(&format!(
-                "FATAL: no epoch for burn height {}",
-                cycle_start_burn_height
+                "FATAL: no epoch for burn height {cycle_start_burn_height}",
             ));
-        if cur_epoch.epoch_id < StacksEpochId::Epoch30 {
+        let cycle = burnchain
+            .block_height_to_reward_cycle(cycle_start_burn_height)
+            .expect("FATAL: no reward cycle for burn height");
+        let is_nakamoto_reward_set = match SortitionDB::get_stacks_epoch_by_epoch_id(
```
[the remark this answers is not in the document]
zone117x: Yeah it seems like this code contradicts a line from the PR description:
> The first nakamoto reward set will occur in epoch 2.5.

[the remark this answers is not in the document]
kantai: > Should we be checking at epoch 2.5? Are all epoch 2.5 reward sets Nakamoto reward sets?

I don't think that they must be or will be. There's a couple reasons for this. First, there's basically two differences between nakamoto reward sets and 2.x reward sets: (1) they _must_ contain signing keys, and so the code asserts as much, and (2) if there's no participation, an error is yielded and burnchain processing halts.

For the first constraint, you can see that there will always be at least one reward cycle in 2.5 which didn't use pox-4. Imagine reward cycle 10 starts at block height 300, and epoch 2.5 starts at block height 300, then reward cycle 10 would be in epoch-2.5, but there's no way it could be a nakamoto reward set: it cannot load signing keys because it could not have been constructed in pox-4.

The second difference is even more important, though, because non-participation during 2.5 _doesn't_ mean a chain halts. It only means a chain halts if the reward-cycle is also part of epoch-3.0. This could be true of the last cycle in 2.5 (but no other ones).

> Yeah it seems like this code contradicts a line from the PR description:
>> The first nakamoto reward set will occur in epoch 2.5.

I think what's happening here is a little confusing -- let's say Epoch-3.0 starts at height 305 and each reward cycle is length 10. Reward cycle 30 is the first nakamoto reward set, even though its start height is in Epoch 2.5. What's important is that every reward set which is active at any time in Epoch 3.0 be a nakamoto reward set. So that's what this calculation is doing: figure out what cycle is active at epoch-3's start height and then find the cycle start height of that cycle.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
kantai: There's a lot of ongoing discussion of this in the open .signers PR, I'm not sure that it has its own issue yet, but its definitely top of mind there.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
kantai: I think its as fatal as non-participation, so it should be treated the same.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: I added a generic const to the `BitVec` struct for `MAX_SIZE`. The field in the block header sets this to 4000.

stackslib/src/chainstate/stacks/boot/mod.rs
kantai: I'd argue this is purely a code-aesthetic choice, really. This branch is only reached if the `signing_principal` set in the pox-4 contract is a contract principal, and the important thing is just that the address is unusable, which the burn address will be regardless of version bit.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
kantai: Yes, but I think that kind of thing is better handled by a unit test in the `make_signer_set` code. Integration tests are very cumbersome for that (not just to write, but to run and maintain) because it will definitely flap if its making very tight assertions (and it should). This is exactly the unit tests that I was waiting to add until I'd gotten some initial feedback on the approach.

stackslib/src/chainstate/coordinator/mod.rs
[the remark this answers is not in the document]
jcnelson: Great, thanks!

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
jcnelson: Opened an issue: https://github.com/stacks-network/stacks-core/issues/4290

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -166,10 +169,34 @@ pub struct PoxStartCycleInfo {
...
+        .ok_or_else(|| serde::de::Error::custom("Address must be a C32 encoded StacksAddress"))
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub struct NakamotoSignerEntry {
+    #[serde(
+        serialize_with = "addr_serialize",
+        deserialize_with = "addr_deserialize"
+    )]
+    pub signing_address: StacksAddress,
```
jcnelson, line 188: Per @jferrant, it turns out that this actually needs to be a public key -- DKG needs the public keys before it can begin.

stackslib/src/chainstate/stacks/boot/mod.rs
jcnelson: This is actually moot now that we know from @jferrant that we need the keys themselves.

jcnelson (commented):
My subsequent review is blocked on the need for the signer set to track public keys instead of principals.

kantai:
Okay -- this PR has now been updated with the work from #4295: the reward set is written to `.signers` _and_ the reward set entries contain the signer keys.

The signer keys are stored in the reward set (_not the stacking-state map_) as `(buff 33)`s, and checked for length. The signer keys are sent over the event dispatcher interface, in the consensus order used for the bit vec.

These signer keys have a many-to-many relationship with PoX addresses. Each reward set entry points to one signing key, but a PoX address can appear multiple times in the reward set entries, and different entries can point to the same signing key.

The `make_signer_set` sums over the signing keys, and this PR includes unit tests for its behavior.

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -71,6 +72,7 @@ pub const PATH_BURN_BLOCK_SUBMIT: &str = "new_burn_block";
 pub const PATH_BLOCK_PROCESSED: &str = "new_block";
 pub const PATH_ATTACHMENT_PROCESSED: &str = "attachments/new";
 pub const PATH_PROPOSAL_RESPONSE: &str = "proposal_response";
+pub const PATH_POX_ANCHOR: &str = "new_pox_set";
```
jferrant, line 75: I see you can subscribe to new stacker signer sets (and not saying in this PR as this should get merged and this is orthogonal) but do ou think we should add an RPC endpoint for the current pox_set? We will need to retrieve them from the node at signer spin up so we can determine the signers and their vote shares.

testnet/stacks-node/src/config.rs
```diff
@@ -2408,6 +2409,10 @@ impl EventKeyType {
             return Some(EventKeyType::BlockProposal);
         }
 
+        if raw_key == "stacker_set" {
```
jferrant, line 2412: I see we have the means to subscribe to stacker sets, but this doesn't help a signer that spins up after stackers have registered. Should I create a ticket to add this as an rpc endpoint as we will need to configure the signers with it?

jferrant:
Aside from the failed build due to a missing bitvec in the one file, LGTM

jferrant (approved):

testnet/stacks-node/src/config.rs
kantai: Yes, please open a ticket -- it should be simple enough to add in a follow-up PR.

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -103,6 +105,15 @@ pub struct MinedNakamotoBlockEvent {
     pub tx_events: Vec<TransactionEvent>,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct PoxAnchorBlockEvent {
```
zone117x, line 109: @kantai did you intend to include this in the event payload? It looks to be unused right now

kantai: Ah -- no, at one point I was going to, but it would have required some cloning of sizable structs (the reward set). Removed in 99730e0c6.

zone117x: That data looked really helpful though! Right now the stacker list data in the event is just this this right?
```rs
pub struct RewardSet {
    pub rewarded_addresses: Vec<PoxAddress>,
    pub start_cycle_state: PoxStartCycleInfo,
}
```

testnet/stacks-node/src/config.rs
jferrant: Done https://github.com/stacks-network/stacks-core/issues/4298

testnet/stacks-node/src/event_dispatcher.rs
kantai: It'll also contain:

```rust
   // only generated for nakamoto reward sets
   pub signers: Option<Vec<NakamotoSignerEntry>>,
```

Which will contain signing keys, stacked amount for each signing key, and the number of signing slots.

jcnelson (approved):

PR #4259, opened 2024-01-18
Clarity-Wasm: fix inconsistencies between reading and writing values in memory

Acaccia:
This PR fixes a few inconsistencies between reading and writing of values in Wasm memory.

The tests for this PR will exist in the [clarity-wasm repo](https://github.com/stacks-network/clarity-wasm), since those issues are linked to property testing in this repo.

PR #4275, opened 2024-01-22
Fix clarity-wasm linked function for `var-set`

Acaccia:
This PR fixes the reading of in-memory values in `link_set_variable_fn` for clarity-wasm.
This is the exact same issue as in `define-data-var` in #4259.

As in the other PR, the tests will happen in the clarity-wasm repo.


obycode (approved):
👍

Issue #3915, opened 2023-09-13
[Signer] Determine VRF algorithm to be used to select a coordinator from a list of signers


jferrant:
Currently the signer is hardcoded to take the very first signer as the coordinator. however, it should be doing some sort of VRF to determine the signer. This may be per reward cycle? Perhaps per transaction? per block height? This ticket includes the design but implementation should be done as part of a separate ticket.

jferrant:
This is low priority. Not scheduled for any time soon.

jferrant:
See https://github.com/stacks-network/stacks-core/blob/33c64f600a31e7967b6d40a02a3fcacc27ccd4f9/stacks-signer/src/runloop.rs#L328

8marz8:
Could we use the consensus_hash fetched from SortitionDB as the alpha (private message) and the signer’s stacks_private_key for generating a VRFProof in calculate_coordinator?
The proof can then be hashed and converted to a number to select a public key from the list.

8marz8:
@jferrant found this relevant discussion from before: https://github.com/stacks-network/sbtc/discussions/335
It resolves my concern that we could just use the public keys.
For the additional data part, retrieving parent_vrf_proof, consensus_hash, sortition_hash or any final combination that we choose to use, requires db connections and I am wondering if it's okay for signers to make db connections, especially if it requires additional configs from them.

8marz8:
@AshtonStephens 
Yep, I think this can be closed as I created another issue for the follow up changes:
https://github.com/stacks-network/stacks-core/issues/4292

PR #4276, opened 2024-01-22
Select stacks-signer coordinator dynamically

8marz8:
### Description
Introducing a deterministic but dynamic signer coordinator selection by hashing signer public keys and stacks tip consensus hash. 

### Applicable issues

- Relates to #3915

- Not exactly a VRF selection so still open if a more complex selection process is needed.
- Selection is similar to the approach discussed [here](https://github.com/stacks-network/sbtc/discussions/335); but stacks_tip_consensus_hash is used instead of latest sortition VRF seed
- Signer Integ. tests are updated to use the calculate_coordinator instead of picking the coordinator statically .

### Checklist

- [ ] Changelog is updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,6 @@ members = [
 wsts = "7.0"
 rand_core = "0.6"
 rand = "0.8"
-sha2 = "0.10"
 
 # Use a bit more than default optimization for
 #  dev builds to speed up test execution
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ members = [
 wsts = "7.0"
 rand_core = "0.6"
 rand = "0.8"
+sha2 = "0.10"
 
 # Use a bit more than default optimization for
 #  dev builds to speed up test execution
--- a/stacks-signer/Cargo.toml
+++ b/stacks-signer/Cargo.toml
@@ -42,7 +42,6 @@ toml = "0.5.6"
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 wsts = { workspace = true }
-sha2 = { workspace = true }
 rand = { workspace = true }
 
 [dependencies.serde_json]
--- a/stacks-signer/Cargo.toml
+++ b/stacks-signer/Cargo.toml
@@ -42,6 +42,7 @@ toml = "0.5.6"
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 wsts = { workspace = true }
+sha2 = "0.10.6"
 
 [dependencies.serde_json]
 version = "1.0"
--- a/stacks-signer/Cargo.toml
+++ b/stacks-signer/Cargo.toml
@@ -42,7 +42,8 @@ toml = "0.5.6"
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 wsts = { workspace = true }
-sha2 = "0.10.6"
+sha2 = { workspace = true }
+rand = { workspace = true }
 
 [dependencies.serde_json]
 version = "1.0"
--- a/stacks-signer/src/client/stacks_client.rs
+++ b/stacks-signer/src/client/stacks_client.rs
@@ -85,14 +85,17 @@ impl StacksClient {
             .json::<serde_json::Value>()
             .map_err(ClientError::ReqwestError)?;
 
-        json_response["stacks_tip_consensus_hash"]
-            .as_str()
+        let stacks_tip_consensus_hash = json_response
+            .get("stacks_tip_consensus_hash")
+            .and_then(|v| v.as_str())
+            .map(String::from)
             .ok_or_else(|| {
                 ClientError::UnexpectedResponseFormat(
-                    "Missing 'stacks_tip_consensus_hash' field".to_string(),
+                    "Missing or invalid 'stacks_tip_consensus_hash' field".to_string(),
                 )
-            })
-            .map(|s| s.to_string())
+            })?;
+
+        Ok(stacks_tip_consensus_hash)
     }
 
     /// Submit the block proposal to the stacks node. The block will be validated and returned via the HTTP endpoint for Block events.
@@ -344,7 +347,7 @@ impl StacksClient {
 }
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
     use std::io::{BufWriter, Read, Write};
     use std::net::{SocketAddr, TcpListener};
     use std::thread::spawn;
@@ -352,13 +355,13 @@ mod tests {
     use super::*;
     use crate::client::ClientError;
 
-    struct TestConfig {
-        mock_server: TcpListener,
-        client: StacksClient,
+    pub(crate) struct TestConfig {
+        pub(crate) mock_server: TcpListener,
+        pub(crate) client: StacksClient,
     }
 
     impl TestConfig {
-        pub fn new() -> Self {
+        pub(crate) fn new() -> Self {
             let mut config = Config::load_from_file("./src/tests/conf/signer-0.toml").unwrap();
 
             let mut mock_server_addr = SocketAddr::from(([127, 0, 0, 1], 0));
@@ -377,7 +380,7 @@ mod tests {
         }
     }
 
-    fn write_response(mock_server: TcpListener, bytes: &[u8]) -> [u8; 1024] {
+    pub(crate) fn write_response(mock_server: TcpListener, bytes: &[u8]) -> [u8; 1024] {
         debug!("Writing a response...");
         let mut request_bytes = [0u8; 1024];
         {
@@ -633,5 +636,16 @@ mod tests {
             b"HTTP/1.1 200 OK\n\n{\"stacks_tip_consensus_hash\": \"4e99f99bc4a05437abb8c7d0c306618f45b203196498e2ebe287f10497124958\"}",
         );
         assert!(h.join().unwrap().is_ok());
+    }
+
+    #[test]
+    fn core_info_call_with_invalid_response_should_fail() {
+        let config = TestConfig::new();
+        let h = spawn(move || config.client.get_stacks_tip_consensus_hash());
+        write_response(
+            config.mock_server,
+            b"HTTP/1.1 200 OK\n\n4e99f99bc4a05437abb8c7d0c306618f45b203196498e2ebe287f10497124958",
+        );
+        assert!(h.join().unwrap().is_err());
     }
 }
--- a/stacks-signer/src/client/stacks_client.rs
+++ b/stacks-signer/src/client/stacks_client.rs
@@ -633,7 +633,7 @@ pub(crate) mod tests {
         let h = spawn(move || config.client.get_stacks_tip_consensus_hash());
         write_response(
             config.mock_server,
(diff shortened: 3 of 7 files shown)


stacks-signer/src/client/stacks_client.rs
```diff
@@ -67,6 +67,34 @@ impl From<&Config> for StacksClient {
...
+        let response = retry_with_exponential_backoff(send_request)?;
+        if !response.status().is_success() {
+            return Err(ClientError::RequestFailure(response.status()));
+        }
+
+        let json_response = response
+            .json::<serde_json::Value>()
+            .map_err(ClientError::ReqwestError)?;
+
+        json_response["stacks_tip_consensus_hash"]
```
jferrant, line 88: I would deserialize the response into RPCPeerInfoData as the json response could be incorrect here and this hard array access would cause the signers to crash. It might be overkill to deserialize the entire response, but would at least convert the array access to a json_response.get("stacks_tip_consensus_hash") and map the result :)

stacks-signer/Cargo.toml
```diff
@@ -42,6 +42,7 @@ toml = "0.5.6"
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 wsts = { workspace = true }
+sha2 = "0.10.6"
```
jferrant, line 45: We should probably try to use the same version sha2 as used throughout the rest of the stacks core.

I see quite a few use this

```
[target.'cfg(all(any(target_arch = "x86_64", target_arch = "x86", target_arch = "aarch64"), not(target_env = "msvc")))'.dependencies]
sha2 = { version = "0.10", features = ["asm"] }

[target.'cfg(any(not(any(target_arch = "x86_64", target_arch = "x86", target_arch = "aarch64")), target_env = "msvc"))'.dependencies]
sha2 = { version = "0.10" }
````

I think maybe we should add this to the outer Cargo.toml file and then put sha2 = { workspace = true } in all other sub repos that require it. Not 100% sure if this is required but when Jude or Aaron gives a review they can confirm or reject it

File: stacks-signer/src/runloop.rs [hunk 3]
```diff
@@ -486,9 +493,34 @@ impl<C: Coordinator> SignerRunLoop<Vec<OperationResult>, RunLoopCommand> for Run
 }
 
 /// Helper function for determining the coordinator public key given the the public keys
-fn calculate_coordinator(public_keys: &PublicKeys) -> (u32, ecdsa::PublicKey) {
-    // TODO: do some sort of VRF here to calculate the public key
-    // See: https://github.com/stacks-network/stacks-blockchain/issues/3915
-    // Mockamato just uses the first signer_id as the coordinator for now
-    (0, public_keys.signers.get(&0).cloned().unwrap())
+fn calculate_coordinator(
```
jferrant, line 496: I don't see anything wrong with this myself. :D You will def have to update the tests to not assume that the coordinator is the first signer in the list! That is why it is hanging

jferrant:
Aside from needing to fix the signer test (which is a bit of a mess so let me know if you need help) , mind adding a local unit test to ensure that two random consensus hashes have two unique results. and that two calls to calculate coordinator with the same consensus hash returns the same result. I know its rather pedantic but I am all for excessive tests :P

stacks-signer/src/client/stacks_client.rs
8marz8: whoops. Updated to use the .get for now

stacks-signer/Cargo.toml
8marz8: Makes sense. Moved it to outer Cargo.toml.

stacks-signer/src/runloop.rs
8marz8: Added unit tests but will need to fix the signers integ tests so the coordinator is not set statically.

Cargo.toml
```diff
@@ -17,6 +17,7 @@ members = [
 wsts = "7.0"
 rand_core = "0.6"
 rand = "0.8"
+sha2 = "0.10"
```
jferrant, line 20: I think you need to change this to 
```
[target.'cfg(all(any(target_arch = "x86_64", target_arch = "x86", target_arch = "aarch64"), not(target_env = "msvc")))'.dependencies]
sha2 = { version = "0.10", features = ["asm"] }

[target.'cfg(any(not(any(target_arch = "x86_64", target_arch = "x86", target_arch = "aarch64")), target_env = "msvc"))'.dependencies]
sha2 = { version = "0.10" }
```

And then update all other instances in stacks core where the cargo tomls reference sha2, to reference the workspace one instead

8marz8: I tried this but got an error. Think the target specification can only be placed under individual crates. If that's the case, I will remove sha2 from here and add the target one to stacker-signer's Cargo.toml

jferrant: ah that might be the case. i have never tried using workspace when accessing target specific shas...yes you can ignore my intial comment and just use it in the stacks-signer sub dir as you had originally :)

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -103,30 +104,21 @@ impl SignerTest {
             Some(Duration::from_millis(128)), // Timeout defaults to 5 seconds. Let's override it to 128 milliseconds.
         );
 
-        let mut running_signers = vec![];
-        let mut _signer_cmd_senders = vec![];
+        let mut running_signers = HashMap::new();
+        let mut _signer_cmd_senders = HashMap::new();
         // Spawn all the signers first to listen to the coordinator request for dkg
```
jferrant, line 109: Nit this comment is no longer relevant and can be either removed or changed to "// Spawn all signers before the node to ensure their listening ports are open for the node event observer to bind to"

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -62,13 +63,13 @@ struct SignerTest {
     // The channel for sending commands to the coordinator
     pub coordinator_cmd_sender: Sender<RunLoopCommand>,
     // The channels for sending commands to the signers
-    pub _signer_cmd_senders: Vec<Sender<RunLoopCommand>>,
+    pub _signer_cmd_senders: HashMap<u32, Sender<RunLoopCommand>>,
```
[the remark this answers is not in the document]
8marz8: Good to know. I didn't know about that convention. Tnx!

jferrant:
LGTM! Just my final nit picks!

jferrant (approved):

stacks-signer/Cargo.toml
jferrant: Sorry. My bad making you reverse this change :(

Cargo.toml
jcnelson: I don't think you need this dependency at all.  You should be able to use one of the hash functions in `stacks-common/src/util/hash.rs`.

stacks-signer/src/client/stacks_client.rs
```diff
@@ -591,4 +626,26 @@ mod tests {
...
         assert!(h.join().unwrap().is_ok());
     }
+
+    #[test]
+    fn core_info_call_for_consensus_hash_should_succeed() {
+        let config = TestConfig::new();
+        let h = spawn(move || config.client.get_stacks_tip_consensus_hash());
+        write_response(
+            config.mock_server,
+            b"HTTP/1.1 200 OK\n\n{\"stacks_tip_consensus_hash\": \"4e99f99bc4a05437abb8c7d0c306618f45b203196498e2ebe287f10497124958\"}",
```
jcnelson, line 636: The consensus hash is a 20-byte hash, not a 32-byte hash.

stacks-signer/src/runloop.rs
```diff
@@ -797,9 +800,151 @@ impl<C: Coordinator> SignerRunLoop<Vec<OperationResult>, RunLoopCommand> for Run
...
-fn calculate_coordinator(public_keys: &PublicKeys) -> (u32, ecdsa::PublicKey) {
-    // TODO: do some sort of VRF here to calculate the public key
-    // See: https://github.com/stacks-network/stacks-blockchain/issues/3915
-    // Mockamato just uses the first signer_id as the coordinator for now
-    (0, public_keys.signers.get(&0).cloned().unwrap())
+pub fn calculate_coordinator(
+    public_keys: &PublicKeys,
+    stacks_client: &StacksClient,
+) -> (u32, ecdsa::PublicKey) {
+    let stacks_tip_consensus_hash = match stacks_client.get_stacks_tip_consensus_hash() {
```
jcnelson, line 807: There are a few of interesting consequences to using the Stacks chain tip consensus hash as the "seed" for choosing a coordinator at random.  They all stem from the fact that the consensus hash only changes whenever the node discovers and processes a new Bitcoin block.

* Stacks nodes A and B both process the same history Bitcoin blocks, and both happen to call `calculate_coordinator` _after_ they have processed the same history (this is the happy path)
* Stacks node A has not processed the full history of Bitcoin blocks (e.g. it's behind by one block), while B has. In this case, A will initially choose the wrong coordinator, but it will eventually choose the right coordinator once its node has caught up.
* Stacks nodes A and B each process divergent histories of Bitcoin blocks (i.e. different forks of Bitcoin), which cause A and B to calculate different coordinators for the same Bitcoin block.

The latter two cases are rare but they do happen in practice (the 2nd case is far more common than the 3rd case).  Regardless of which happens, node A's signer will choose the wrong coordinator.  How does node A recover?

jferrant: As far as I am aware, the system will currently just ignore that signer. it will not delcare it byzantine. We filter out "non coordinator' coordinator messages from the signer runloop with zero consequence to that signer.

jcnelson: >  I think modifying the messages sent through stacker db to include additional info such as the consensus tip and the signers responding with the NACK with hints that someone has an incorrect view of chain tip would be a good improvement ticket. Want me to copy paste your notes into a new issue?

Yes, I think we should do this, even if the only thing we do with this data is log it.  I suspect that this is going to be one of the more common and banal reasons why DKG or a signing round will fail on the first try.

jcnelson: And yeah this can be a separate PR :+1: Just want to make sure it's something the system can account for.

kantai (approved):
This looks good to me, once the outstanding comments are addressed (open an issue for adding each signer's consensus-hash-view to their messages and use `Sha256Sum::from_data` or `from_hasher`)

stacks-signer/src/client/stacks_client.rs
8marz8: Fixed.

stacks-signer/src/runloop.rs
```diff
@@ -797,9 +800,151 @@ impl<C: Coordinator> SignerRunLoop<Vec<OperationResult>, RunLoopCommand> for Run
...
+        "Using stacks_tip_consensus_hash {:?} for selecting coordinator",
+        &stacks_tip_consensus_hash
+    );
+
+    // Create combined hash of each signer's public key with stacks_tip_consensus_hash
+    let mut selection_ids = public_keys
+        .signers
+        .iter()
+        .map(|(&id, pk)| {
+            let mut hasher = Sha256::new();
```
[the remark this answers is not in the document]
8marz8: Tnx, I replaced with above.

8marz8:
Created below issue related to above discussions:
https://github.com/stacks-network/stacks-core/issues/4292

Issue #4274, opened 2024-01-22
[Nakamoto] PoX reward set state should include the signing key, not the stacker state


jcnelson:
The PoX reward set maps need to track the signing key.

setzeus:
Confirming that this means we should *remove* the existing field from 'stacking-state' which actually has a neat secondary effect: instead of setting the signing-key per-member using 'delegate-stack-stx' the pool operator can now provide a single signing-key w/ 'aggregate-commit-index.'

This is useful since currently a pool operator rotating signing-keys (for some grief reason) midway through stacker sign-ups would cause members already stacked w/ the old key to miss the next cycle.

setzeus:
Discussed & assigned to WIP in #4263.

MarvinJanssen:
This was addressed in #4263 and #4269.

PR #4277, opened 2024-01-23
feat: require signatures to prove control of `signer-key` in pox

### LINKED ISSUE #4247: Signer should control which stackers can delegate to the signer

hstove:
This closes #4247, which is to ensure that all Signers are using signing keys that they either control, or are controlled by the account they're delegating to.

- closes #4247 

This is done by adding a `signer-sig` param to:

- stack-stx
- delegate-stack-stx
- stack-aggregation-commit

The signature follows [SIP018](https://github.com/stacksgov/sips/blob/main/sips/sip-018/sip-018-signed-structured-data.md) for signed structured data. Following SIP018 has the benefit of built-in wallet support, so signers can use hardware wallets and apps to generate these signatures. 

Following SIP018, the structured data is `{ reward-cycle, reward-cycle }`, where `reward-cycle` is the _current_ stacking cycle where the transaction is made. The `domain` is `{ app: "pox-4-signer", version: "1.0.0", chain-id }`.

Because we'll almost certainly be building the ability for `stacks-signer` to generate these signatures (ie via CLI), I added a SIP18 util library at `/stacksib/src/lib_util/signed_structured_data`. That library has tests that validate the functionality based on the test vectors in SIP018.

I added a function `to_rsv` to `MessageSignature` to match the byte order that Clarity uses for signatures.

Note that I added the new `signer-sig` param as the second-to-last argument because of other conversations around expected argument order, but I'm not really familiar with the requirements there. I also am not sure what other touch points use these functions where the arguments need to be updated. cc @zone117x regarding param order

### DIFF
--- a/stacks-common/src/util/secp256k1.rs
+++ b/stacks-common/src/util/secp256k1.rs
@@ -109,6 +109,11 @@ impl MessageSignature {
             Err(_) => None,
         }
     }
+
+    /// Convert from VRS to RSV
+    pub fn to_rsv(&self) -> Vec<u8> {
+        [&self.0[1..], &self.0[0..1]].concat()
+    }
 }
 
 #[cfg(any(test, feature = "testing"))]
--- a/stackslib/src/chainstate/stacks/boot/pox-4.clar
+++ b/stackslib/src/chainstate/stacks/boot/pox-4.clar
@@ -30,6 +30,7 @@
 (define-constant ERR_INVALID_SIGNER_KEY 32)
 (define-constant ERR_REUSED_SIGNER_KEY 33)
 (define-constant ERR_DELEGATION_ALREADY_REVOKED 34)
+(define-constant ERR_DELEGATION_INVALID_SIGNATURE (err 35))
 
 ;; Valid values for burnchain address versions.
 ;; These first four correspond to address hash modes in Stacks 2.1,
@@ -636,6 +637,7 @@
                              (delegate-to principal)
                              (until-burn-ht (optional uint))
                              (pox-addr (optional { version: (buff 1), hashbytes: (buff 32) })))
+                             
     (begin
       ;; must be called directly by the tx-sender or by an allowed contract-caller
       (asserts! (check-caller-allowed)
@@ -669,6 +671,43 @@
           pox-addr: pox-addr })
 
       (ok true)))
+
+(define-read-only (get-signer-key-message-hash (signer-key (buff 33)) (stacker principal))
+  (let
+    (
+      (domain { name: "pox-4-signer", version: "1.0.0", chain-id: chain-id })
+      (data-hash (sha256 (unwrap-panic 
+        (to-consensus-buff? { stacker: stacker, reward-cycle: (current-pox-reward-cycle) }))))
+      (domain-hash (sha256 (unwrap-panic (to-consensus-buff? domain))))
+    )
+    (sha256 (concat
+      0x534950303138
+      (concat domain-hash
+      data-hash)))
+  )
+)
+
+;; Verify a signature from the signing key for this specific stacker.
+;; The message hash is the sha256 of the consensus hash of the tuple 
+;; `{ stacker, reward-cycle }`. Note that `reward-cycle` corresponds to the
+;; _current_ reward cycle, not the reward cycle at which the delegation will start.
+;; The public key is recovered from the signature and compared to the pubkey hash
+;; of the delegator.
+(define-read-only (verify-signing-key-signature (stacker principal)
+                                                (signing-key (buff 33))
+                                                (signer-sig (buff 65)))
+  (let
+    (
+      ;; (msg { stacker: stacker, reward-cycle: (current-pox-reward-cycle) })
+      ;; (msg-bytes (unwrap! (to-consensus-buff? msg) ERR_DELEGATION_INVALID_SIGNATURE)) ;;TODO
+      ;; ;; (msg-hash (sha256 msg-bytes))
+      (msg-hash (get-signer-key-message-hash signing-key stacker))
+      (pubkey (unwrap! (secp256k1-recover? msg-hash signer-sig) ERR_DELEGATION_INVALID_SIGNATURE)) ;; TODO
+    )
+    (asserts! (is-eq pubkey signing-key) ERR_DELEGATION_INVALID_SIGNATURE)
+    (ok true)
+  )                                            
+)
 
 ;; Commit partially stacked STX and allocate a new PoX reward address slot.
 ;;   This allows a stacker/delegate to lock fewer STX than the minimal threshold in multiple transactions,
--- a/stackslib/src/chainstate/stacks/boot/pox-4.clar
+++ b/stackslib/src/chainstate/stacks/boot/pox-4.clar
@@ -672,6 +672,7 @@
 
       (ok true)))
 
+;; Generate a message hash following the SIP018 standard.
 (define-read-only (get-signer-key-message-hash (signer-key (buff 33)) (stacker principal))
   (let
     (
@@ -698,9 +699,6 @@
                                                 (signer-sig (buff 65)))
   (let
     (
-      ;; (msg { stacker: stacker, reward-cycle: (current-pox-reward-cycle) })
-      ;; (msg-bytes (unwrap! (to-consensus-buff? msg) ERR_DELEGATION_INVALID_SIGNATURE)) ;;TODO
-      ;; ;; (msg-hash (sha256 msg-bytes))
       (msg-hash (get-signer-key-message-hash signing-key stacker))
       (pubkey (unwrap! (secp256k1-recover? msg-hash signer-sig) ERR_DELEGATION_INVALID_SIGNATURE)) ;; TODO
     )
--- a/stackslib/src/chainstate/stacks/boot/pox-4.clar
+++ b/stackslib/src/chainstate/stacks/boot/pox-4.clar
@@ -603,7 +603,7 @@
         (err ERR_STACKING_INSUFFICIENT_FUNDS))
 
       ;; Validate ownership of the given signer key
-      (try! (verify-signer-key-sig tx-sender signer-sig signer-key))
+      (try! (verify-signer-key-sig pox-addr signer-sig signer-key))
 
       ;; ensure that stacking can be performed
       (try! (can-stack-stx pox-addr amount-ustx first-reward-cycle lock-period))
@@ -682,14 +682,14 @@
 
 ;; Generate a message hash for validating a signer key.
 ;; The message hash follows SIP018 for signing structured data. The structured data
-;; is the tuple `{ stacker, reward-cycle }`. The domain is
+;; is the tuple `{ pox-addr: { version, hashbytes }, reward-cycle }`. The domain is
 ;; `{ name: "pox-4-signer", version: "1.0.0", chain-id: chain-id }`.
-(define-read-only (get-signer-key-message-hash (stacker principal))
+(define-read-only (get-signer-key-message-hash (pox-addr { version: (buff 1), hashbytes: (buff 32) }))
   (let
     (
       (domain { name: "pox-4-signer", version: "1.0.0", chain-id: chain-id })
       (data-hash (sha256 (unwrap-panic
-        (to-consensus-buff? { stacker: stacker, reward-cycle: (current-pox-reward-cycle) }))))
+        (to-consensus-buff? { pox-addr: pox-addr, reward-cycle: (current-pox-reward-cycle) }))))
       (domain-hash (sha256 (unwrap-panic (to-consensus-buff? domain))))
     )
     (sha256 (concat
@@ -705,12 +705,12 @@
 ;; Note that `reward-cycle` corresponds to the _current_ reward cycle,
 ;; not the reward cycle at which the delegation will start.
 ;; The public key is recovered from the signature and compared to `signer-key`.
-(define-read-only (verify-signer-key-sig (stacker principal)
+(define-read-only (verify-signer-key-sig (pox-addr { version: (buff 1), hashbytes: (buff 32) })
                                          (signer-sig (buff 65))
                                          (signer-key (buff 33)))
   (let
     (
-      (msg-hash (get-signer-key-message-hash stacker))
+      (msg-hash (get-signer-key-message-hash pox-addr))
       (pubkey (unwrap! (secp256k1-recover? msg-hash signer-sig) (err ERR_INVALID_SIGNATURE_RECOVER)))
     )
     (asserts! (is-eq pubkey signer-key) (err ERR_INVALID_SIGNATURE_PUBKEY))
@@ -741,7 +741,7 @@
     ;; must be called directly by the tx-sender or by an allowed contract-caller
     (asserts! (check-caller-allowed)
               (err ERR_STACKING_PERMISSION_DENIED))
-    (try! (verify-signer-key-sig tx-sender signer-sig signer-key))
+    (try! (verify-signer-key-sig pox-addr signer-sig signer-key))
     (let ((amount-ustx (get stacked-amount partial-stacked)))
       (try! (can-stack-stx pox-addr amount-ustx reward-cycle u1))
       ;; Add the pox addr to the reward cycle, and extract the index of the PoX address
@@ -1061,7 +1061,7 @@
               (err ERR_STACKING_IS_DELEGATED))
 
     ;; Verify signature from delegate that allows this sender for this cycle
-    (try! (verify-signer-key-sig tx-sender signer-sig signer-key))
+    (try! (verify-signer-key-sig pox-addr signer-sig signer-key))
 
     ;; TODO: add more assertions to sanity check the `stacker-info` values with
     ;;       the `stacker-state` values
--- a/stackslib/src/chainstate/stacks/boot/pox-4.clar
+++ b/stackslib/src/chainstate/stacks/boot/pox-4.clar
@@ -732,6 +732,7 @@
 ;; *New in Stacks 2.1.*
 (define-private (inner-stack-aggregation-commit (pox-addr { version: (buff 1), hashbytes: (buff 32) })
                                                 (reward-cycle uint)
+                                                (signer-sig (buff 65))
                                                 (signer-key (buff 33)))
   (let ((partial-stacked
          ;; fetch the partial commitments
@@ -740,6 +741,7 @@
     ;; must be called directly by the tx-sender or by an allowed contract-caller
     (asserts! (check-caller-allowed)
               (err ERR_STACKING_PERMISSION_DENIED))
+    (try! (verify-signer-key-sig tx-sender signer-sig signer-key))
     (let ((amount-ustx (get stacked-amount partial-stacked)))
       (try! (can-stack-stx pox-addr amount-ustx reward-cycle u1))
       ;; Add the pox addr to the reward cycle, and extract the index of the PoX address
@@ -773,8 +775,9 @@
 ;; Returns (err ...) on failure.
 (define-public (stack-aggregation-commit (pox-addr { version: (buff 1), hashbytes: (buff 32) })
                                          (reward-cycle uint)
+                                         (signer-sig (buff 65))
                                          (signer-key (buff 33)))
-    (match (inner-stack-aggregation-commit pox-addr reward-cycle signer-key)
+    (match (inner-stack-aggregation-commit pox-addr reward-cycle signer-sig signer-key)
         pox-addr-index (ok true)
         commit-err (err commit-err)))
 
@@ -782,8 +785,9 @@
 ;; *New in Stacks 2.1.*
 (define-public (stack-aggregation-commit-indexed (pox-addr { version: (buff 1), hashbytes: (buff 32) })
                                                  (reward-cycle uint)
+                                                 (signer-sig (buff 65))
                                                  (signer-key (buff 33)))
-    (inner-stack-aggregation-commit pox-addr reward-cycle signer-key))
+    (inner-stack-aggregation-commit pox-addr reward-cycle signer-sig signer-key))
 
 ;; Commit partially stacked STX to a PoX address which has already received some STX (more than the Stacking min).
 ;; This allows a delegator to lock up marginally more STX from new delegates, even if they collectively do not
--- a/stackslib/src/chainstate/stacks/boot/pox-4.clar
+++ b/stackslib/src/chainstate/stacks/boot/pox-4.clar
@@ -30,7 +30,8 @@
 (define-constant ERR_INVALID_SIGNER_KEY 32)
 (define-constant ERR_REUSED_SIGNER_KEY 33)
 (define-constant ERR_DELEGATION_ALREADY_REVOKED 34)
-(define-constant ERR_DELEGATION_INVALID_SIGNATURE (err 35))
+(define-constant ERR_INVALID_SIGNATURE_PUBKEY 35)
+(define-constant ERR_INVALID_SIGNATURE_RECOVER 36)
 
 ;; Valid values for burnchain address versions.
 ;; These first four correspond to address hash modes in Stacks 2.1,
@@ -66,6 +67,9 @@
 
 ;; Stacking thresholds
 (define-constant STACKING_THRESHOLD_25 (if is-in-mainnet u20000 u8000))
+
+;; SIP18 message prefix
+(define-constant SIP018_MSG_PREFIX 0x534950303138)
 
 ;; Data vars that store a copy of the burnchain configuration.
 ;; Implemented as data-vars, so that different configurations can be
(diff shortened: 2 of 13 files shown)


stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -658,6 +664,28 @@
...
+;; The public key is recovered from the signature and compared to the pubkey hash
+;; of the delegator.
+(define-read-only (verify-delegator-signature (stacker principal)
+                                             (delegator principal)
+                                             (delegator-sig (buff 65)))
+  (let
+    (
+      (msg { stacker: stacker, reward-cycle: (current-pox-reward-cycle) })
+      (msg-bytes (unwrap! (to-consensus-buff? msg) ERR_DELEGATION_INVALID_SIGNATURE)) ;;TODO
+      (msg-hash (sha256 msg-bytes))
```
jcnelson, line 680: We use sha512/256 for our message digests

hstove: As in sha512(sha256(message))? No problem, I'll change that.

hstove:
Do we need to add the signature check to `delegate-stack-increase` and `delegate-stack-extend`?

hstove:
After further thought, I think we should follow SIP018 (signed structured data) for this. That also unlocks the ability for signers to use a web app to request signatures from a wallet.

friedger:
#4247 is about delegating signer duties, not delegating stx. As far as I understand. 

Pool operators can already refuse to `delegate-stack-stx` for a pool member.

hstove:
@friedger @kantai is it better to describe this ticket as "delegates and delegators need to prove ownership of the signing key"? For delegates, this is proving that the signer they're delegating to has control over the signing key.

stackslib/src/chainstate/stacks/boot/pox-4.clar
hstove: I also notice that in [SIP018](https://github.com/stacksgov/sips/blob/main/sips/sip-018/sip-018-signed-structured-data.md#formal-specification) (signed structured data) only `sha256` is used, and I'd like to build this off of SIP018.

jcnelson:
_Everyone_ needs to prove ownership of their signing key.  This goes for solo stackers as well, via `stack-stx` and `stack-extend`.  Basically, anywhere you see a signing key as an argument, you need to also require a signature argument and a signature verification (that should be `stack-stx`, `delegate-stack-stx`, `stack-extend`, and `delegate-stack-extend`).

friedger:
Instead of delegate stack stx, pool operators now provide the key in stack aggregation-* function calls.

See #4274

hstove:
> Instead of delegate stack stx, pool operators now provide the key in stack aggregation-* function calls.
> 
> See #4274

Good to know - and I'll have to change this to build off the change from `signer-key` to `signer` too.

hstove:
> _Everyone_ needs to prove ownership of their signing key. This goes for solo stackers as well, via `stack-stx` and `stack-extend`. Basically, anywhere you see a signing key as an argument, you need to also require a signature argument and a signature verification (that should be `stack-stx`, `delegate-stack-stx`, `stack-extend`, and `delegate-stack-extend`).

👍 that's reflected in the current state of the PR

hstove:
There are a number of tests outside of the pox-4 tests related to nakamoto that are failing because of the change to `stack-stx`, I'm working on those still but the pox-4 code is ready for review.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -569,6 +570,7 @@
                           (pox-addr (tuple (version (buff 1)) (hashbytes (buff 32))))
                           (start-burn-ht uint)
                           (lock-period uint)
+                          (signer-sig (buff 65))
                           (signer-key (buff 33)))
```
zone117x, line 574: Thoughts on removing the `signer-key` arg? Seems redundant now that `signer-sig` is passed and the key can be derived from it? I'm guessing the burn-op versions of these functions will do that to save OP_RETURN space.

hstove: Interesting thought. If it weren't for the burn op issue I'd probably push back, because it's too easy for a mis-created signature to cause an invalid signing key to be used. If the message hash differs, then the signer-key will end up as something else.

stacks-common/src/util/secp256k1.rs
```diff
@@ -109,6 +109,11 @@ impl MessageSignature {
             Err(_) => None,
         }
     }
+
+    /// Convert from VRS to RSV
+    pub fn to_rsv(&self) -> Vec<u8> {
```
jcnelson, line 114: Why is this conversion necessary?

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -672,6 +678,44 @@
...
+(define-read-only (get-signer-key-message-hash (stacker principal))
+  (let
+    (
+      (domain { name: "pox-4-signer", version: "1.0.0", chain-id: chain-id })
+      (data-hash (sha256 (unwrap-panic 
+        (to-consensus-buff? { stacker: stacker, reward-cycle: (current-pox-reward-cycle) }))))
+      (domain-hash (sha256 (unwrap-panic (to-consensus-buff? domain))))
+    )
+    (sha256 (concat
+      0x534950303138
```
jcnelson, line 694: This should be a constant somewhere

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -1235,13 +1263,15 @@ fn pox_4_revoke_delegate_stx_events() {
...
     let alice_delegation_amount = 100_000_000;
+    let alice_delegate_nonce = alice_nonce;
     let alice_delegate = make_pox_4_delegate_stx(
         &alice,
         alice_nonce,
         alice_delegation_amount,
         bob_principal,
         None,
         None,
+        // &signature,
```
jcnelson, line 1274: Should this be commented?

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -1421,33 +1633,34 @@ fn stack_stx_signer_key() {
...
-    );
-    let signer_key_val = Value::buff_from(vec![
-        0x03, 0xa0, 0xf9, 0x81, 0x8e, 0xa8, 0xc1, 0x4a, 0x82, 0x7b, 0xb1, 0x44, 0xae, 0xc9, 0xcf,
-        0xba, 0xeb, 0xa2, 0x25, 0xaf, 0x22, 0xbe, 0x18, 0xed, 0x78, 0xa2, 0xf2, 0x98, 0x10, 0x6f,
-        0x4e, 0x28, 0x1b,
-    ])
-    .unwrap();
-    let txs = vec![make_pox_4_contract_call(
-        stacker_key,
+    // let pox_addr = make_pox_addr(
```
jcnelson, line 1650: Please drop commented code

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -1910,14 +2141,28 @@ fn delegate_stack_increase() {
...
     let delegate_stack_stx = make_pox_4_delegate_stack_stx(
         bob_delegate_key,
         bob_nonce,
-        PrincipalData::from(key_to_stacks_addr(alice_key)).into(),
+        alice_principal,
         min_ustx,
         pox_addr.clone(),
         block_height as u128,
         lock_period,
+        signature,
```
jcnelson, line 2165: Can you add some test coverage to this file for all four affected stacking functions to verify that they fail with the expected error code if the signature was bad?  Not just unit-testing the signature verification function; verifying end-to-end that the stacking function fails as expected.  Thanks!

stackslib/src/util_lib/signed_structured_data.rs
```diff
@@ -0,0 +1,122 @@
+use clarity::vm::Value;
```
jcnelson, line 1: Missing copyright comment block

stackslib/src/util_lib/signed_structured_data.rs
```diff
@@ -0,0 +1,122 @@
...
+pub const STRUCTURED_DATA_PREFIX: [u8; 6] = [0x53, 0x49, 0x50, 0x30, 0x31, 0x38];
+
+pub fn structured_data_hash(value: Value) -> Sha256Sum {
+    let bytes = value.serialize_to_vec();
+    Sha256Sum::from_data(&bytes)
+}
+
+/// Generate a message hash for signing structured Clarity data.
+/// Reference [SIP018](https://github.com/stacksgov/sips/blob/main/sips/sip-018/sip-018-signed-structured-data.md) for more information.
+pub fn structured_data_message_hash(structured_data: Value, domain: Value) -> Sha256Sum {
```
jcnelson, line 21: Why does this belong in `util_lib`?  Isn't this only used in testing?  If so, then please remove this file altogether and put this into `stackslib/src/chainstate/stacks/boot/mod.rs`.

stackslib/src/chainstate/stacks/boot/pox-4.clar
hstove: The latest in this PR does follow SIP18, the v1 definitely didn't

stacks-common/src/util/secp256k1.rs
hstove: Because when recovering or verifying signatures in Clarity, the signature needs to be in RSV order. Whether this helper function is necessary or not is a fine question, it was just helpful to re-use

stackslib/src/util_lib/signed_structured_data.rs
hstove: At the moment it's only used in tests, but we'll add it (in a separate PR) to `stacks-signer` so we can expose a CLI function like:

```
stacks-signer create-stacking-signature --stacker S --reward-cycle R
```

I don't have a strong opinion whether it should live here or not, and I'm not the best person to know. But I think it is helpful to expose somewhere that isn't just for tests.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
hstove: Added 4 new tests that cover all the stacking transactions

setzeus (changes requested):
Generally pox_4.clar updates LGTM minus a few syntax issues & one ERR constant inconsistency.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -30,6 +30,7 @@
 (define-constant ERR_INVALID_SIGNER_KEY 32)
 (define-constant ERR_REUSED_SIGNER_KEY 33)
 (define-constant ERR_DELEGATION_ALREADY_REVOKED 34)
+(define-constant ERR_DELEGATION_INVALID_SIGNATURE (err 35))
```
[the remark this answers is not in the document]
hstove: ah, my habits slipped in here 😬

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -672,6 +681,44 @@
...
+;; The public key is recovered from the signature and compared to `signer-key`.
+(define-read-only (verify-signer-key-sig (stacker principal)
+                                                (signer-key (buff 33))
+                                                (signer-sig (buff 65)))
+  (let
+    (
+      (msg-hash (get-signer-key-message-hash stacker))
+      (pubkey (unwrap! (secp256k1-recover? msg-hash signer-sig) ERR_DELEGATION_INVALID_SIGNATURE)) ;; TODO
+    )
+    (asserts! (is-eq pubkey signer-key) ERR_DELEGATION_INVALID_SIGNATURE)
```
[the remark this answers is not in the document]
hstove: Good call, I also renamed the error message to be more clear (and removed `delegation` as its inaccurate)

setzeus (approved):

hstove:
FYI I'm going to rebase/build this off of #4269, because that changes the function signatures in pox-4. Shouldn't be too much work and then I'll move the ticket back to review.

setzeus:
@hstove copy, will re-review once re-based.

stackslib/src/util_lib/signed_structured_data.rs
jcnelson: Sounds like this belongs in `libsigner` then.  `libsigner` can be a dev-dependency for `stackslib`.

stacks-common/src/util/secp256k1.rs
jcnelson: Ugh I just noticed that `secp256k1_recover()` and `secp256k_verify()` don't behave the same as the corresponding methods in `Secp256k1PublicKey` and `MessageSignature`.  The former two methods, which Clarity uses, were added by an itinerant 3rd party contributor who couldn't be bothered to figure out how the existing code worked.

hstove:
Ready for review. This was rebased off of #4269 and now off of `next`, which changed a few things from last week's version:

- the `signer-key` (and thus `signer-sig`) args were removed from `delegate-stack-*`
- `signer-sig` was added to `stack-aggregation-commit`
- Now the structured data tuple is `{ pox-addr, reward-cycle }`. Previously it was `{ stacker, reward-cycle }`

stacks-common/src/util/secp256k1.rs
hstove: Yeah, that's unfortunate because almost every `sign()` implementation returns the signature in RSV, but Clarity uses VRS, so you always have to build a helper method for it. Easy to get tripped up on.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -725,17 +775,19 @@
 ;; Returns (err ...) on failure.
 (define-public (stack-aggregation-commit (pox-addr { version: (buff 1), hashbytes: (buff 32) })
```
friedger, line 776: This method should be removed. It was kept for backwards compatibility. Now, it is not compatible anymore, therefore, we only should use `stack-aggregation-commit-indexed`

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -1805,11 +2322,20 @@ fn delegate_stack_stx_extend_signer_key() {
     )
     .expect("No stacking state, stack-stx failed")
     .expect_tuple();
+    let delegation_state = get_delegation_state_pox_4(&mut peer, &latest_block, &alice_principal)
+        .expect("No delegation state, delegate-stx failed")
+        .expect_tuple();
 
-    let next_reward_cycle = 1 + burnchain
+    let stacking_state = get_stacking_state_pox_4(&mut peer, &latest_block, &alice_principal)
+        .expect("No stacking state, stack-stx failed")
```
friedger, line 2330: bob called delegate-stack-stx that failed here.

Should be changed in line 2323 as well.

friedger (dismissed):
The burn chain tx for stack-stx and stack-extend need to be updated.

Otherwise, just small nits. This looks good if all tests pass.

stackslib/src/chainstate/stacks/boot/pox-4.clar
hstove: I'm curious how many apps use this function vs `-indexed`. My guess is that the easiest path forward is to remove `-indexed` but change the function's return signature here.

For example, looks like the stacks.js package uses `stack-aggregation-commit`.

friedger: Stacks.js has not yet been updated to stacks 2.4 🙃

Either is fine to me. Jude might have a stronger opinion.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
hstove: Can you explain more? When I run this test every contract call returns `ok`, including `delegate-stack-stx`

hstove:
> The burn chain tx for stack-stx and stack-extend need to be updated.

We have a ticket for that at #4285, should we wait for that decision to be made before merging this? I'm not sure if we have tests for those burn ops in pox-4. I think it's already broken from just `signer-key` being added, though I might be mistaken.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -571,6 +576,7 @@
                           (pox-addr (tuple (version (buff 1)) (hashbytes (buff 32))))
                           (start-burn-ht uint)
                           (lock-period uint)
+                          (signer-sig (buff 65))
```
zone117x, line 579: Synthetic pox events are generated using argument indexes rather than named values, so it's generally easier to add new args to the end of the list. Not a deal breaker in this PR because `signer-sig` isn't yet added to the pox events, but I'm working on adding that and our PRs will conflict due to the ordering.

zone117x: Conflict PR: https://github.com/stacks-network/stacks-core/pull/4329

I can modify my PR and change the event arg indexes, no problem. Just an FYI

hstove: No problem, let's merge your PR and I'll update mine.

hstove: I have no real preference on arg order. Is your PR the only real place we need to worry about arg index? I'd probably just update that synthetic events code to change the signer_key index, but if there are other downstream issues then it's probably better for me to change the argument order on pox-4.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -670,6 +680,44 @@
...
+  )
+)
+
+;; Verify a signature from the signing key for this specific stacker.
+;; See `get-signer-key-message-hash` for details on the message hash.
+;;
+;; Note that `reward-cycle` corresponds to the _current_ reward cycle,
+;; not the reward cycle at which the delegation will start.
+;; The public key is recovered from the signature and compared to `signer-key`.
+(define-read-only (verify-signer-key-sig (pox-addr { version: (buff 1), hashbytes: (buff 32) })
```
jcnelson, line 708: Similarly, can this take the reward cycle as an argument, and pass it into `get-signer-key-message-hash`?

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -692,6 +741,7 @@
     ;; must be called directly by the tx-sender or by an allowed contract-caller
     (asserts! (check-caller-allowed)
               (err ERR_STACKING_PERMISSION_DENIED))
+    (try! (verify-signer-key-sig pox-addr signer-sig signer-key))
```
jcnelson, line 744: You can use `(current-reward-cycle)` here as the reward cycle arg.

jcnelson, line 744: You'd want to pass in the given `reward-cycle` here

stackslib/src/chainstate/stacks/boot/pox-4.clar
jcnelson: Also, given the way this is used in `stack-aggregation-commit-indexed`, you'd actually need a caller-supplied reward cycle, since it'll be different than the current reward cycle.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -670,6 +680,44 @@
 
       (ok true)))
 
+;; Generate a message hash for validating a signer key.
+;; The message hash follows SIP018 for signing structured data. The structured data
+;; is the tuple `{ pox-addr: { version, hashbytes }, reward-cycle }`. The domain is
+;; `{ name: "pox-4-signer", version: "1.0.0", chain-id: chain-id }`.
+(define-read-only (get-signer-key-message-hash (pox-addr { version: (buff 1), hashbytes: (buff 32) }))
```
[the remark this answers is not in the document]
jcnelson: Given how this is reached from `stack-aggregation-commit-indexed`, you'll want the reward cycle to be an argument anyway -- the reward cycle won't be equal to `current-pox-reward-cycle`.

stackslib/src/chainstate/stacks/boot/pox-4.clar
jcnelson: It's fine to remove it for pox-4, but please add a changelog entry.

stackslib/src/chainstate/stacks/boot/pox-4.clar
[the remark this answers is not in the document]
jcnelson: Actually, I think we're missing some data here.  The signature should really be over the PoX address, the first reward cycle for which the signer will sign, and the number of reward cycles for which the signer will sign.  Without that last parameter, someone could use a signer's key for 12 cycles while the signer might have only wanted to sign for 1 cycle.

* For `stack-stx`, the signature would include `lock-period`
* For `stack-extend`, the signature would include `extend-count`
* For `stack-aggregation-commit-indexed`, the lockup period would just be 1.

jcnelson (dismissed):
I think we want the signature to cover some more data than it does right now.

stackslib/src/chainstate/stacks/boot/pox-4.clar
zone117x: Thanks!

zone117x (approved):
I'm approving the pox4 function signature and synthetic event changes, thanks @hstove!

At least 2 others should approve the pox4 behavioral changes because I'm not in a position to validate those changes.

stackslib/src/chainstate/stacks/boot/pox-4.clar
hstove: To maintain the behavior of "signer key signatures must use _current_ reward cycle", I've passed in `(- reward-cycle u1)` into the signature validation function

hstove:
I've updated this PR with changes since receiving Jude's feedback. Tagging @jcnelson @setzeus @MarvinJanssen @friedger for reviews.

- The signature message hash now includes two new fields
  - `topic`: one of `stack-stx`, `stack-extend`, or `agg-commit`. This prevents re-using signatures across methods
  - `period`: a `uint` reflecting `lock-period` for stacking or `extend-count` for extending. For `stack-aggregation-commit`, this value is `u1`.
- The `get-signer-key-message-hash` now takes all hash fields as args, instead of dynamically calling `get-current-reward-cycle`. This allows signers to easily call this read-only method to generate a message hash.
- In `stack-aggregation-commit`, the message hash uses the reward cycle provided as user input instead of the current reward cycle.

I've also added some code in `util_lib::signed_structured_data::pox4` to allow re-using the Rust code for generating these signatures instead of re-writing a helper function in different places.

kantai (commented):
This LGTM -- I'd just suggest using the `define_named_enum!` macro for the topics, and you should add another unit test that asserts equality between the rust message hash constructor and the contract's read-only fn.

stackslib/src/util_lib/signed_structured_data.rs
```diff
@@ -0,0 +1,235 @@
...
+                Pox4SignatureTopic::StackExtend => "stack-extend",
+            }
+        }
+    }
+
+    pub fn make_pox_4_signed_data_domain(chain_id: u32) -> Value {
+        make_structured_data_domain("pox-4-signer", "1.0.0", chain_id)
+    }
+
+    pub fn make_pox_4_signer_key_message_hash(
```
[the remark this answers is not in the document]
hstove: :+1: Added in https://github.com/stacks-network/stacks-core/pull/4277/commits/dbacfd06c2c19f91ed531aeac0723cdb02aec248

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -670,6 +680,57 @@
...
 
+;; Generate a message hash for validating a signer key.
+;; The message hash follows SIP018 for signing structured data. The structured data
+;; is the tuple `{ pox-addr: { version, hashbytes }, reward-cycle }`. The domain is
+;; `{ name: "pox-4-signer", version: "1.0.0", chain-id: chain-id }`.
+(define-read-only (get-signer-key-message-hash (pox-addr { version: (buff 1), hashbytes: (buff 32) }) 
+                                               (reward-cycle uint)
+                                               (topic (string-ascii 12))
+                                               (period uint))
+  (let
```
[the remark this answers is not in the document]
hstove: Yep good call. I actually just stripped out the entire `let` and have one statement for both methods. The `let` made writing the code easier, but there was no variable re-use so it wasn't necessary. Plus we can save a teeny bit of cost without it!

stackslib/src/util_lib/signed_structured_data.rs
```diff
@@ -0,0 +1,407 @@
...
+        StackStx("stack-stx"),
+        AggregationCommit("agg-commit"),
+        StackExtend("stack-extend"),
+    });
+    pub enum Pox4SignatureTopicOld {
+        StackStx,
+        AggregationCommit,
+        StackExtend,
+    }
+
```
kantai, line 95: Can `Pox4SignatureTopicOld` be removed?

kantai (approved):
LGTM -- just see if the `...Old` enum can be removed.

hstove:
> LGTM -- just see if the ...Old enum can be removed.

Ah crap, thanks for catching that. CI should pass and then I think this is good to merge!

PR #4280, opened 2024-01-23
Add `/status` endpoint to signer

hstove:
I've been documenting the workflow for running the signer, and I think it would be helpful to expose a `GET /status` endpoint to help setup and monitor the signer. This PR responds with `200 OK` for that endpoint.

Note that this could potentially conflict with an event observer path in the future. I'm not sure if the Stacks Blockchain API exposes a similar endpoint on the event receiver, but if so, we should follow use same route. @zone117x does that exist?

I could also imagine having an endpoint similar to `/v2/status` on the node, where more helpful state is returned, but I'm not sure how much we want to expose the fact that a specific IP is running a specific signer.

### DIFF
--- a/libsigner/src/events.rs
+++ b/libsigner/src/events.rs
@@ -211,6 +211,8 @@ pub enum SignerEvent {
     SignerMessages(Vec<SignerMessage>),
     /// A new block proposal validation response from the node
     BlockValidationResponse(BlockValidateResponse),
+    /// Status endpoint request
+    StatusCheck,
 }
 
 /// Trait to implement a stop-signaler for the event receiver thread.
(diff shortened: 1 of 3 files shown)


libsigner/src/events.rs
```diff
@@ -211,6 +213,19 @@ impl EventReceiver for SignerEventReceiver {
                 return Err(EventError::Terminated);
             }
 
+            if request.method() == &HttpMethod::Get {
```
jferrant, line 216: not sure we have to check if it is a Get method. we could just put this request.url() check below like how "/stackerdb_chunks" and "/block_proposal" are done.

jferrant:
If you see value in having this, I have no problem with it. It can always be removed later if we change our mind.

libsigner/src/events.rs
hstove: Yep good call

jcnelson (approved):

libsigner/src/events.rs
```diff
@@ -211,6 +213,13 @@ impl EventReceiver for SignerEventReceiver {
                 return Err(EventError::Terminated);
             }
 
+            if request.url() == "/status" {
```
jcnelson, line 216: To be clear, this is a `POST` not a `GET`, right?  These are events that are `POST`ed by the node to the signer's event observer client.

hstove: In this code it would respond to both GET and POST, though I have no problem keeping it to just GET. This is for the signer's event receiver, but the goal is to have a status endpoint for monitoring, which is why GET made sense.

This does "pollute" the event receiver, but the alternative is to have a separate endpoint/port to listen on for monitoring, if we want to have some kind of automated status check.

jferrant (approved):

Issue #4265, opened 2024-01-19
[Nakamoto] Miner handles block responses from signers stackerdb instance


jferrant:
The nakamoto miner currently self signs a block before appending it to the chain. It should be retrieving the signed block from the signers instead.

Update the nakamoto miner node to handle stackerdb events from the .signers contract and attempt to append the block if it is signed to the chain. It should also handle a block response that rejects the block and attempt to submit a new one instead.

This depends on https://github.com/stacks-network/stacks-core/pull/4249

PR #4281, opened 2024-01-24
Miners handle signers block responses from the stacker db instance

jferrant:
### Description
This is the first stab at miners handling stacker db events from the signers. Progress beyond this point is blocked by the .signers pox 4 contract combined with the ability for signers to broadcast their aggregate keys.

### Applicable issues

- PARTIALLY fixes https://github.com/stacks-network/stacks-core/issues/4265

A more complete solution would directly use the event observer trait (avoiding the use of the http server), but for an initial attempt, I just directly access the stacker db instance of the miner. As I am semi blocked, I might try making this change now. However review would be good anyway as this logic will not change much.

Next steps would be to actually append the block when we have .signers / aggregate key correctly set

### DIFF
--- a/libsigner/src/events.rs
+++ b/libsigner/src/events.rs
@@ -28,179 +28,26 @@ use blockstack_lib::net::api::postblock_proposal::{
     BlockValidateReject, BlockValidateResponse, ValidateRejectCode,
 };
 use blockstack_lib::util_lib::boot::boot_code_id;
+use clarity::vm::types::serialization::SerializationError;
 use clarity::vm::types::QualifiedContractIdentifier;
 use serde::{Deserialize, Serialize};
 use stacks_common::codec::{
-    read_next, read_next_at_most, write_next, Error as CodecError, StacksMessageCodec,
+    read_next, read_next_at_most, read_next_exact, write_next, Error as CodecError,
+    StacksMessageCodec,
 };
 use stacks_common::util::hash::Sha512Trunc256Sum;
 use tiny_http::{
     Method as HttpMethod, Request as HttpRequest, Response as HttpResponse, Server as HttpServer,
 };
 use wsts::common::Signature;
-use wsts::net::{Message, Packet};
+use wsts::net::{
+    DkgBegin, DkgEnd, DkgEndBegin, DkgPrivateBegin, DkgPrivateShares, DkgPublicShares, DkgStatus,
+    Message, NonceRequest, NonceResponse, Packet, SignatureShareRequest, SignatureShareResponse,
+};
+use wsts::state_machine::signer;
 
 use crate::http::{decode_http_body, decode_http_request};
-use crate::EventError;
-
-/// Temporary placeholder for the number of slots allocated to a stacker-db writer. This will be retrieved from the stacker-db instance in the future
-/// See: https://github.com/stacks-network/stacks-blockchain/issues/3921
-/// Is equal to the number of message types
-pub const SIGNER_SLOTS_PER_USER: u32 = 11;
-
-// The slot IDS for each message type
-const DKG_BEGIN_SLOT_ID: u32 = 0;
-const DKG_PRIVATE_BEGIN_SLOT_ID: u32 = 1;
-const DKG_END_BEGIN_SLOT_ID: u32 = 2;
-const DKG_END_SLOT_ID: u32 = 3;
-const DKG_PUBLIC_SHARES_SLOT_ID: u32 = 4;
-const DKG_PRIVATE_SHARES_SLOT_ID: u32 = 5;
-const NONCE_REQUEST_SLOT_ID: u32 = 6;
-const NONCE_RESPONSE_SLOT_ID: u32 = 7;
-const SIGNATURE_SHARE_REQUEST_SLOT_ID: u32 = 8;
-const SIGNATURE_SHARE_RESPONSE_SLOT_ID: u32 = 9;
-/// The slot ID for the block response for miners to observe
-pub const BLOCK_SLOT_ID: u32 = 10;
-
-/// The messages being sent through the stacker db contracts
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub enum SignerMessage {
-    /// The signed/validated Nakamoto block for miners to observe
-    BlockResponse(BlockResponse),
-    /// DKG and Signing round data for other signers to observe
-    Packet(Packet),
-}
-
-/// The response that a signer sends back to observing miners
-/// either accepting or rejecting a Nakamoto block with the corresponding reason
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub enum BlockResponse {
-    /// The Nakamoto block was accepted and therefore signed
-    Accepted((Sha512Trunc256Sum, ThresholdSignature)),
-    /// The Nakamoto block was rejected and therefore not signed
-    Rejected(BlockRejection),
-}
-
-impl BlockResponse {
-    /// Create a new accepted BlockResponse for the provided block signer signature hash and signature
-    pub fn accepted(hash: Sha512Trunc256Sum, sig: Signature) -> Self {
-        Self::Accepted((hash, ThresholdSignature(sig)))
-    }
-
-    /// Create a new rejected BlockResponse for the provided block signer signature hash and signature
-    pub fn rejected(hash: Sha512Trunc256Sum, sig: Signature) -> Self {
-        Self::Rejected(BlockRejection::new(
-            hash,
-            RejectCode::SignedRejection(ThresholdSignature(sig)),
-        ))
-    }
-}
-
-/// A rejection response from a signer for a proposed block
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub struct BlockRejection {
-    /// The reason for the rejection
-    pub reason: String,
-    /// The reason code for the rejection
-    pub reason_code: RejectCode,
-    /// The signer signature hash of the block that was rejected
-    pub signer_signature_hash: Sha512Trunc256Sum,
-}
-
-impl BlockRejection {
-    /// Create a new BlockRejection for the provided block and reason code
-    pub fn new(signer_signature_hash: Sha512Trunc256Sum, reason_code: RejectCode) -> Self {
-        Self {
-            reason: reason_code.to_string(),
-            reason_code,
-            signer_signature_hash,
-        }
-    }
-}
-
-impl From<BlockValidateReject> for BlockRejection {
-    fn from(reject: BlockValidateReject) -> Self {
-        Self {
-            reason: reject.reason,
-            reason_code: RejectCode::ValidationFailed(reject.reason_code),
-            signer_signature_hash: reject.signer_signature_hash,
-        }
-    }
-}
-
-/// This enum is used to supply a `reason_code` for block rejections
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-#[repr(u8)]
-pub enum RejectCode {
-    /// RPC endpoint Validation failed
-    ValidationFailed(ValidateRejectCode),
-    /// Signers signed a block rejection
-    SignedRejection(ThresholdSignature),
-    /// Insufficient signers agreed to sign the block
-    InsufficientSigners(Vec<u32>),
-}
-
-impl std::fmt::Display for RejectCode {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        match self {
-            RejectCode::ValidationFailed(code) => write!(f, "Validation failed: {:?}", code),
-            RejectCode::SignedRejection(sig) => {
-                write!(f, "A threshold number of signers rejected the block with the following signature: {:?}.", sig)
-            }
-            RejectCode::InsufficientSigners(malicious_signers) => write!(
-                f,
-                "Insufficient signers agreed to sign the block. The following signers are malicious: {:?}",
-                malicious_signers
-            ),
-        }
-    }
-}
-
-impl From<Packet> for SignerMessage {
-    fn from(packet: Packet) -> Self {
-        Self::Packet(packet)
-    }
-}
-
-impl From<BlockResponse> for SignerMessage {
-    fn from(block_response: BlockResponse) -> Self {
-        Self::BlockResponse(block_response)
-    }
-}
-
-impl From<BlockRejection> for SignerMessage {
-    fn from(block_rejection: BlockRejection) -> Self {
-        Self::BlockResponse(BlockResponse::Rejected(block_rejection))
-    }
-}
-
-impl From<BlockValidateReject> for SignerMessage {
-    fn from(rejection: BlockValidateReject) -> Self {
-        Self::BlockResponse(BlockResponse::Rejected(rejection.into()))
-    }
-}
-
-impl SignerMessage {
-    /// Helper function to determine the slot ID for the provided stacker-db writer id
-    pub fn slot_id(&self, id: u32) -> u32 {
-        let slot_id = match self {
-            Self::Packet(packet) => match packet.msg {
-                Message::DkgBegin(_) => DKG_BEGIN_SLOT_ID,
-                Message::DkgPrivateBegin(_) => DKG_PRIVATE_BEGIN_SLOT_ID,
-                Message::DkgEndBegin(_) => DKG_END_BEGIN_SLOT_ID,
-                Message::DkgEnd(_) => DKG_END_SLOT_ID,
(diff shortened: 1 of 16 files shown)


libsigner/src/events.rs
```diff
@@ -349,3 +435,86 @@ impl EventReceiver for SignerEventReceiver {
...
+    is_mainnet: bool,
+) -> Result<SignerEvent, EventError> {
+    debug!("Got stackerdb_chunks event");
+    let mut body = String::new();
+    if let Err(e) = request.as_reader().read_to_string(&mut body) {
+        error!("Failed to read body: {:?}", &e);
+
+        request
+            .respond(HttpResponse::empty(200u16))
+            .expect("response failed");
```
xoloki, line 452: I know this is copy pasta from above, but can we get rid of the expect?  Presumably network errors could cause the response to fall, so we shouldn't crash, plus we're returning an error anyway so might as well just pass on the network error?

xoloki (approved):
It would be nice to get rid of some of the expect calls, but they were already present so not a blocker for this PR

libsigner/src/events.rs
jferrant: I can do that now :)

jferrant: Done!

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -241,6 +259,84 @@ impl BlockMinerThread {
...
+        // Get the block slot for every signer
+        let slot_ids = stackerdbs
+            .get_signers(signers_contract_id)
+            .expect("FATAL: could not get signers from stacker DB")
+            .iter()
+            .enumerate()
+            .map(|(id, _)| id as u32 * SIGNER_SLOTS_PER_USER + BLOCK_SLOT_ID)
+            .collect::<Vec<u32>>();
+        // If more than 30% of the signers reject the block, we should not wait any further as we will never get a valid signed block
+        let rejection_threshold = slot_ids.len() / 10 * 3;
```
jferrant, line 287: Gonna make this configurable.

jferrant:
> It would be nice to get rid of some of the expect calls, but they were already present so not a blocker for this PR

Fixed in https://github.com/stacks-network/stacks-core/pull/4294

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -241,6 +270,90 @@ impl BlockMinerThread {
...
+        // Get the block slot for every signer
+        let slot_ids = stackerdbs
+            .get_signers(&signers_contract_id)
+            .expect("FATAL: could not get signers from stacker DB")
+            .iter()
+            .enumerate()
+            .map(|(id, _)| id as u32 * SIGNER_SLOTS_PER_USER + BLOCK_SLOT_ID)
+            .collect::<Vec<u32>>();
+        // If more than a threshold percentage of the signers reject the block, we should not wait any further
+        let rejection_threshold =
```
jferrant, line 297: This rejection logic needs to be revamped. A single signer could spam the miner with rejections and it cause the miner to falsely drop the block when the other signers actually are signing it. Therefore, track who sent the block rejections and ensure they are from a unique signer that we haven't already processed.

testnet/stacks-node/Cargo.toml
```diff
@@ -31,6 +31,7 @@ wsts = { workspace = true }
 rand = { workspace = true }
 rand_core = { workspace = true }
 hashbrown = "0.14"
+bincode = "1.3.3"
```
jcnelson, line 34: Why is this now a depedency?

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -241,6 +270,90 @@ impl BlockMinerThread {
         }
     }
 
+    fn wait_for_signed_block(
```
jcnelson, line 273: Architecturally speaking, wouldn't it make more sense to have the node react to push notifications from arriving StackerDB chunks, instead of polling every so often?

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -241,6 +270,90 @@ impl BlockMinerThread {
...
+        &self,
+        stackerdbs: &StackerDBs,
+        aggregate_public_key: &Point,
+        signer_signature_hash: &Sha512Trunc256Sum,
+    ) -> Option<NakamotoBlock> {
+        let stackerdb_contracts = stackerdbs
+            .get_stackerdb_contract_ids()
+            .expect("FATAL: could not get the stacker DB contract ids");
+        // TODO: get this directly instead of this jankiness when .signers is a boot contract
```
jcnelson, line 282: Not sure what this comment means?

testnet/stacks-node/Cargo.toml
jcnelson: I'd very strongly prefer that `bincode` was kept only within the signer.  If you want the node proper to consume signer messages, then they need to be structs that implement `StacksMessageCodec`.  In particular, `bincode` does not by default guard against memory exhaustion errors that result from deserializing malicious inputs (whereas `StacksMessageCodec` does).

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -241,6 +270,90 @@ impl BlockMinerThread {
...
+                        if block_rejection.block.header.signer_signature_hash().ok()?
+                            == *signer_signature_hash
+                        {
+                            let signature = block_rejection.block.header.signer_signature;
+                            let mut message = signer_signature_hash.0.to_vec();
+                            message.push(b'n');
+                            if signature.verify(aggregate_public_key, &message) {
+                                // We received a verified rejection. We will NEVER get a signed block from the signers for this particular block
+                                return None;
+                            } else if signature
```
jcnelson, line 336: Why even count unverified rejections? If signers can't get an aggregate signature for either accept or reject, then no blocks could be produced anyway.

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -241,6 +270,90 @@ impl BlockMinerThread {
...
+                        if block.header.signer_signature_hash().ok()? == *signer_signature_hash
+                            && block
+                                .header
+                                .signer_signature
+                                .verify(aggregate_public_key, &signer_signature_hash.0)
+                        {
+                            // We verified that the returned block contains a valid signature across the signer signature hash of our original block request
+                            // Let's just immediately attempt to append the block without waiting for more signer messages
+                            return Some(block);
+                        }
```
jcnelson, line 323: What happens if verification fails?

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -241,6 +270,90 @@ impl BlockMinerThread {
...
+        let slot_ids = stackerdbs
+            .get_signers(&signers_contract_id)
+            .expect("FATAL: could not get signers from stacker DB")
+            .iter()
+            .enumerate()
+            .map(|(id, _)| id as u32 * SIGNER_SLOTS_PER_USER + BLOCK_SLOT_ID)
+            .collect::<Vec<u32>>();
+        // If more than a threshold percentage of the signers reject the block, we should not wait any further
+        let rejection_threshold =
+            slot_ids.len() / 100 * self.config.miner.signer_rejection_threshold;
```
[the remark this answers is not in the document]
jferrant: Because it is configurable by signers what the dkg threshold is/sign threshold is. If it is configurable for them, it must be for us as well.

testnet/stacks-node/src/nakamoto_node/miner.rs
jferrant: I am okay with reacting to push notifications, but this will be a more involved change as I think I will need to do some revamping to prevent having to spin a listener as a node to itself. (so I can directly access the callbacks.) Following convos with Aaron, he suggested this approach for first pass. But don't mind making the change now.

testnet/stacks-node/src/nakamoto_node/miner.rs
jferrant: I shouldn't have to verify that the miner contains the signers_contract_id. I wrote this before it even existed. This enabled the tests to pass for both my own where I manually deployed signers vs ones that did not.

testnet/stacks-node/src/nakamoto_node/miner.rs
jferrant: If verification fails, it will attempt to mine a new block

testnet/stacks-node/src/nakamoto_node/miner.rs
jferrant: We need to verify the rejection as a malicious signer could spam the miner with false rejections forcing them to remine a new block when they should not have to.

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -241,6 +270,90 @@ impl BlockMinerThread {
...
+                                .verify(aggregate_public_key, &signer_signature_hash.0)
+                        {
+                            // We verified that the returned block contains a valid signature across the signer signature hash of our original block request
+                            // Let's just immediately attempt to append the block without waiting for more signer messages
+                            return Some(block);
+                        }
+                    }
+                    SignerMessage::BlockResponse(BlockResponse::Rejected(block_rejection)) => {
+                        // First check that this block rejection is for the block we proposed
+                        if block_rejection.block.header.signer_signature_hash().ok()?
```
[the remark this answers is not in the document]
jferrant: It will continue to wait for verification and not prematurely mine a new block (not until enough rejections appear that are valid indicating that the block is not gonna get accepted.) They will then mine a new block.

stackslib/src/net/stackerdb/db.rs
```diff
@@ -537,6 +537,19 @@ impl StackerDBs {
...
 
+    /// Get all principals who can write to a particular stacker DB.
+    /// Returns Ok(list of addr) if this contract exists in the DB
+    /// Returns Err(..) if the DB doesn't exist of some other DB error happens
+    pub fn get_signers(
+        &self,
+        smart_contract: &QualifiedContractIdentifier,
+    ) -> Result<Vec<StacksAddress>, net_error> {
+        let stackerdb_id = self.get_stackerdb_id(smart_contract)?;
+        let sql = "SELECT signer FROM chunks WHERE stackerdb_id = ?1 GROUP BY signer";
```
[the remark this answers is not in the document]
jferrant: Oh maybe I am wrong, but I thought GROUP BY would remove duplicates? or do you mean do both to prevent the need to compress so many duplicates?

testnet/stacks-node/Cargo.toml
jferrant: This is a depenency because this is how the messages are serialized in libsigner. if we do not deserialize it the same way, then it will fail. I can add StacksMessageCodec to all the types but that adds a lot of what I would consider unnecessary code. However if the desire to remove the bincode dependency outweighs the con of a ton of extra code, can do.

jferrant: Done! :D

jferrant: Done!

libsigner/src/messages.rs
```diff
@@ -0,0 +1,1213 @@
...
+            11 => Ok(TypePrefix::SignatureShareResponse),
+            12 => Ok(TypePrefix::DkgStatusSuccess),
+            13 => Ok(TypePrefix::DkgStatusFailure),
+            _ => Err(CodecError::DeserializeError(format!(
+                "Unknown type prefix: {}",
+                value
+            ))),
+        }
+    }
+}
```
kantai, line 116: `define_u8_enum!()` provides `from_u8(v: u8) -> Option<Self>`, so this code could just be:

```suggestion
impl TryFrom<u8> for TypePrefix {
    type Error = CodecError;
    fn try_from(value: u8) -> Result<Self, Self::Error> {
           Self::from_u8(value).ok_or_else(|| {
                CodecError::DeserializeError(format!("Unknown type prefix: {value}"))
           })
    }
}
```

stackslib/src/net/api/postblock_proposal.rs
```diff
@@ -75,6 +76,21 @@ pub enum ValidateRejectCode {
...
+        match value {
+            0 => Ok(ValidateRejectCode::BadBlockHash),
+            1 => Ok(ValidateRejectCode::BadTransaction),
+            2 => Ok(ValidateRejectCode::InvalidBlock),
+            3 => Ok(ValidateRejectCode::ChainstateError),
+            4 => Ok(ValidateRejectCode::UnknownParent),
+            _ => Err(format!("Invalid value for ValidateRejectCode: {value}")),
+        }
+    }
+}
```
kantai, line 92: Related comment about `define_u8_enum!` here -- if you use that to define the enum, you could implement this without a potentially dangerous match.

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -213,6 +223,25 @@ impl BlockMinerThread {
...
+                        &stackerdbs,
+                        &aggregate_public_key,
+                        &signer_signature_hash,
+                    ) {
+                        // TODO remove self signing once casting aggregate vote is done
+                        new_block.header.signer_signature = signature;
+                    } else {
+                        debug!("Failed to get a signed block from signers");
+                    }
+                };
```
kantai, line 244: I think this behavior should be mutually exclusive with `self.config.self_signing()` --

If `self_signing()` is enabled, then it shouldn't use (or wait for) the signer-set signature.
If `self_signing()` is disabled, then it should use the signer-set signature and then broadcast the block.

I'd recommend moving this functionality into a new function like `.wait_for_signer_signature_and_broadcast`, and then that could be invoked at line 252, replacing the current warning when self-signing is disabled.

kantai (commented):
This looks mostly good to me -- I just have a couple comments about u8 enums, and a request on the changes in the miner thread.

libsigner/src/messages.rs
```diff
@@ -0,0 +1,1195 @@
...
+                    "Unknown signer message type prefix: {}",
+                    type_prefix_byte
+                )))
+            }
+        };
+        Ok(message)
+    }
+}
+
+fn consensus_serialize_scalar<W: Write>(fd: &mut W, scalar: &Scalar) -> Result<(), CodecError> {
```
jcnelson, line 172: Each of these `consensus_[de]serialize_TYPE` functions should really be trait implementations of `StacksMessageCodec`.  You would create a trait extension for each `TYPE`, and then implement that extension trait for the `TYPE` as well as `StacksMessageCodec` for the extension.

For example, you'd have `trait ScalarExtensions` here, and you'd implement it for `Scalar`, and you'd implement `StacksMessageCodec` for `T: ScalarExtensions`.

stackslib/src/net/stackerdb/db.rs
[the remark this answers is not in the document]
jcnelson: Ah, you're right.  `GROUP BY` and `SELECT DISTINCT` behave the same when there is no aggregate function, which applies here.

testnet/stacks-node/src/nakamoto_node/miner.rs
jcnelson: Ah gotcha

testnet/stacks-node/src/nakamoto_node/miner.rs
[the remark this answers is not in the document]
jcnelson: That's what I'm confused about. I thought the rejection threshold was always 30%?  Like, what happens if two signers have different rejection thresholds, and one signer believes a block is rejected while the other does not?

jcnelson (approved):
Just one question about whether or not rejection thresholds should be configurable, but otherwise LGTM.

Before merging, please implement extension traits for `wsts` types and then implement `StacksMessageCodec` for those.

testnet/stacks-node/src/nakamoto_node/miner.rs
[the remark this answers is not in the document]
jferrant: I can rip it out throughout the code base as part of this PR so it is not configurable ?

libsigner/src/messages.rs
jferrant: Ahhh I didn't think of this. I was like "Oh can't implement a trait for a object outside of its definition folder". Will try this work around!

jferrant: I cannot get the trait to work for StacksMessageCodec. I ended up using just a local trait definition StacksMessageCodecExtensions and implementing it for all the wsts objects. Attempts to then use `impl<T: StacksMessageCodecExtensions> StacksMessageCodec for T` all failed with complaints about generics not being allowed for external traits

PR #4300, opened 2024-01-26
Make signer_signature_hash infallible and use it instead of the entire block when passing around data in stackerdb

jferrant:
### Description
I realized I was unnecessarily sending around the entire block when the only thing that a miner needs is the signature corresponding to a signer signature hash of his proposed block. This change also enables signers who were not online during the nonce request portion of a signing round to broadcast signing round results as the signature cannot be spoofed enabling them to better contribute to the livliness of the chain.

NOTE: I have made the signer_signature_hash and miner_signature_hash unwrap the underlying hash functions similar to how we do block_hash. As far as I am aware...we cannot produce a NakamotoBlock that does not produce a valid signer and miner signature hash....But please correct me if I am wrong.

### DIFF
--- a/libsigner/src/events.rs
+++ b/libsigner/src/events.rs
@@ -23,6 +23,7 @@ use std::sync::Arc;
 use blockstack_lib::chainstate::nakamoto::NakamotoBlock;
 use blockstack_lib::chainstate::stacks::boot::{MINERS_NAME, SIGNERS_NAME};
 use blockstack_lib::chainstate::stacks::events::StackerDBChunksEvent;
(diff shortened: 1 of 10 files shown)


kantai (approved):
LGTM!

> NOTE: I have made the signer_signature_hash and miner_signature_hash unwrap the underlying hash functions similar to how we do block_hash. As far as I am aware...we cannot produce a NakamotoBlock that does not produce a valid signer and miner signature hash....But please correct me if I am wrong.

Yes -- if the stacks-node can't serialize the signature hash, it's in an irrecoverable state (OOM) and should panic.

xoloki (approved):
LGTM

jcnelson (approved):

Issue #4006, opened 2023-10-20
[Nakamoto] Stacks-signer should write a list of all their transactions to stackerdb rather than to the mempool


jferrant:
The signer should not have to pay transaction fees for special cased control plane  transactions. However, if the fee is set too low the mempool responds with MemPoolRejection::FeeTooLow . Instead of broadcasting these transactions to the mempool, after epoch 3 is in place, signers should broadcast these transactions as a list to stacker db for other signers and the miner to observe.
Signers can then verify that proposed blocks from the miner contain all transactions that they and the other signesr care about and the miner can in turn query the stackerdb slots for the list of all transactions that each stacks signer cares about.

Note that pre epoch 3, they should try broadcasting directly to the mempool 

related to https://github.com/stacks-network/stacks-core/pull/4239
Related to https://github.com/stacks-network/stacks-core/issues/3981

PR #4301, opened 2024-01-27
Feat/signers write transactions to stackerdb

jferrant:
### Description
Signers must broadcast their cast aggregate public key transaction to the mempool if in pre nakamoto rules, and always broadcast the transaction to stackerdb for signers and miners to observe when validating and building the block, respectively. 

### Applicable issues

- fixes #https://github.com/stacks-network/stacks-core/issues/4006

For this to be fully tested, we need to have the cast aggregate public key transaction happening

### DIFF
--- a/stacks-signer/src/client/stackerdb.rs
+++ b/stacks-signer/src/client/stackerdb.rs
@@ -21,7 +21,7 @@ use libsigner::{
 };
 use libstackerdb::{StackerDBChunkAckData, StackerDBChunkData};
 use slog::{slog_debug, slog_warn};
-use stacks_common::codec::StacksMessageCodec;
+use stacks_common::codec::{read_next, StacksMessageCodec};
 use stacks_common::types::chainstate::StacksPrivateKey;
 use stacks_common::{debug, warn};
 
@@ -37,6 +37,8 @@ pub struct StackerDB {
     stacks_private_key: StacksPrivateKey,
     /// A map of a slot ID to last chunk version
     slot_versions: HashMap<u32, u32>,
+    /// The signer ID
+    signer_id: u32,
 }
 
 impl From<&Config> for StackerDB {
@@ -48,6 +50,7 @@ impl From<&Config> for StackerDB {
             ),
             stacks_private_key: config.stacks_private_key,
             slot_versions: HashMap::new(),
+            signer_id: config.signer_id,
         }
     }
 }
@@ -56,11 +59,10 @@ impl StackerDB {
     /// Sends messages to the .signers stacker-db with an exponential backoff retry
     pub fn send_message_with_retry(
         &mut self,
-        id: u32,
         message: SignerMessage,
     ) -> Result<StackerDBChunkAckData, ClientError> {
         let message_bytes = message.serialize_to_vec();
-        let slot_id = message.slot_id(id);
+        let slot_id = message.slot_id(self.signer_id);
 
         loop {
             let slot_version = *self.slot_versions.entry(slot_id).or_insert(0) + 1;
@@ -116,7 +118,7 @@ impl StackerDB {
             if !chunk_ack.is_empty() {
                 for chunk in chunk_ack {
                     if let Some(data) = chunk {
-                        if let Ok(message) = bincode::deserialize::<SignerMessage>(&data) {
+                        if let Ok(message) = read_next::<SignerMessage, _>(&mut &data[..]) {
                             if let SignerMessage::Transactions(chunk_transactions) = message {
                                 transactions.extend(chunk_transactions);
                             } else {
--- a/stacks-signer/src/client/stackerdb.rs
+++ b/stacks-signer/src/client/stackerdb.rs
@@ -116,11 +116,14 @@ impl StackerDB {
             if !chunk_ack.is_empty() {
                 for chunk in chunk_ack {
                     if let Some(data) = chunk {
-                        let message: SignerMessage = bincode::deserialize(&data).unwrap();
-                        if let SignerMessage::Transactions(chunk_transactions) = message {
-                            transactions.extend(chunk_transactions);
+                        if let Ok(message) = bincode::deserialize::<SignerMessage>(&data) {
+                            if let SignerMessage::Transactions(chunk_transactions) = message {
+                                transactions.extend(chunk_transactions);
+                            } else {
+                                warn!("Signer wrote an unexpected type to the transactions slot");
+                            }
                         } else {
-                            warn!("Signer wrote an unexpected type to the transactions slot");
+                            warn!("Failed to deserialize chunk data into a SignerMessage");
                         }
                     }
                 }
--- a/stacks-signer/src/client/stackerdb.rs
+++ b/stacks-signer/src/client/stackerdb.rs
@@ -1,3 +1,5 @@
+use std::net::SocketAddr;
+
 use blockstack_lib::chainstate::stacks::StacksTransaction;
 // Copyright (C) 2013-2020 Blockstack PBC, a public benefit corporation
 // Copyright (C) 2020-2024 Stacks Open Internet Foundation
@@ -56,6 +58,21 @@ impl From<&Config> for StackerDB {
 }
 
 impl StackerDB {
+    /// Create a new StackerDB client
+    pub fn new(
+        host: SocketAddr,
+        stackerdb_contract_id: QualifiedContractIdentifier,
+        stacks_private_key: StacksPrivateKey,
+        signer_id: u32,
+    ) -> Self {
+        Self {
+            signers_stackerdb_session: StackerDBSession::new(host, stackerdb_contract_id),
+            stacks_private_key,
+            slot_versions: HashMap::new(),
+            signer_id,
+        }
+    }
+
     /// Sends messages to the .signers stacker-db with an exponential backoff retry
     pub fn send_message_with_retry(
         &mut self,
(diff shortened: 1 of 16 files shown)


stacks-signer/src/client/stackerdb.rs
```diff
@@ -90,8 +115,103 @@ impl StackerDB {
...
+                .get_latest_chunks(&slot_ids)
+                .map_err(backoff::Error::transient)
+        };
+        let chunk_ack = retry_with_exponential_backoff(send_request)?;
+        let mut transactions = Vec::new();
+        for (i, chunk) in chunk_ack.iter().enumerate() {
+            if let Some(data) = chunk {
+                if let Ok(message) = bincode::deserialize::<SignerMessage>(data) {
+                    if let SignerMessage::Transactions(chunk_transactions) = message {
+                        let signer_id = *signer_ids.get(i).expect(
```
jcnelson, line 142: What happens if `chunk_ack.len() >= signer_ids.len()`?

stacks-signer/src/client/stacks_client.rs
```diff
@@ -167,6 +206,26 @@ impl StacksClient {
...
 
+    /// Helper function to retrieve the account info from the stacks node for a specific address
+    fn get_account_entry(
+        &self,
+        address: &StacksAddress,
+    ) -> Result<AccountEntryResponse, ClientError> {
+        debug!("Getting account info...");
+        let send_request = || {
+            self.stacks_node_client
+                .get(self.accounts_path(address))
```
jcnelson, line 217: You're passing `?proof=0` as the query string, right?  Otherwise, you're going to get a large amount of unneeded data back.

stacks-signer/src/client/stacks_client.rs
```diff
@@ -188,6 +247,40 @@ impl StacksClient {
...
     }
 
+    /// Helper function  that attempts to deserialize a clarity hext string as a list of signer slots and their associated number of signer slots
+    fn parse_signer_slots(&self, hex: &str) -> Result<Vec<(StacksAddress, u128)>, ClientError> {
+        debug!("Parsing signer slots: {hex}...");
+        // Due to .signers definition, the  signer slots is always a result of a list of tuples of signer addresses and the number of slots they have
+        // If this fails, we have bigger problems than the signer crashing...
+        let value = ClarityValue::try_deserialize_hex_untyped(hex)?
+            .expect_result()
+            .map_err(|e| ClientError::ReadOnlyFailure(e.to_string()))?;
```
jcnelson, line 257: Instead of `ReadOnlyFailure`, do you think you could have an error type specific to Clarity deserialization failure?

stacks-signer/src/runloop.rs
```diff
@@ -459,29 +480,109 @@ impl<C: Coordinator> RunLoop<C> {
...
     }
 
+    /// Verify the transactions in a block are as expected
+    fn verify_transactions(&mut self, block: &NakamotoBlock) -> bool {
+        if let Ok(expected_transactions) = self.get_expected_transactions() {
+            // Ensure the block contains the transactions we expect
+            let missing_transactions = expected_transactions
+                .into_iter()
+                .filter_map(|tx| {
+                    if !block.txs.contains(&tx) {
```
jcnelson, line 502: Blocks can contain north of 10,000 transactions. It might be worth building a hashset of the blocks' txids and checking that against the expected transaction's txid.

stacks-signer/src/runloop.rs
```diff
@@ -459,29 +480,109 @@ impl<C: Coordinator> RunLoop<C> {
...
             return false;
         }
-        Self::determine_vote(block_info, request, transactions);
+
+        Self::determine_vote(block_info, nonce_request);
         true
     }
 
+    /// Verify the transactions in a block are as expected
+    fn verify_transactions(&mut self, block: &NakamotoBlock) -> bool {
```
jcnelson, line 496: I'm not sure if there's test coverage for this?

stacks-signer/src/runloop.rs
jferrant: That won't allow us to give a hint back to the miner, are we okay with that?

jcnelson: Ah, I'm speaking specifically about the call to `block.txs.contains(&tx)`.  `block.txs` is a `Vec`, so `.contains()` runs in _O(n)_ time.  I'm just asking about speeding this up.

stacks-signer/src/client/stacks_client.rs
jferrant: No I did not, will update to do!

stacks-signer/src/client/stackerdb.rs
jferrant: This cannot happen. The call get_latest_chunks will just return None for slots that don't exist so we will always return as many chunks as we queried. and we query an equal number of slots to signer_ids.If this was to happen, it would be a BUG in the system hence the expect.

stacks-signer/src/client/stacks_client.rs
jferrant: Can do!

stacks-signer/src/runloop.rs
```diff
@@ -459,29 +480,109 @@ impl<C: Coordinator> RunLoop<C> {
...
+                .send_message_with_retry(block_rejection.into())
+            {
+                warn!("Failed to send block submission to stacker-db: {:?}", e);
+            }
+            false
+        }
+    }
+
+    /// Get the transactions we expect to see in the next block
+    fn get_expected_transactions(&mut self) -> Result<Vec<StacksTransaction>, ClientError> {
```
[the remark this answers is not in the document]
jferrant: It is tested by the integration test, but will add a smaller unit test :D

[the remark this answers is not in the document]
jferrant: Done!

stacks-signer/src/runloop.rs
jferrant: Not sure why i couldn't get the invalid case to not hang in my simulations so have removed that test. I know I check the invalid case in my integration test. But I at least added a unit test for the valid case.

stacks-signer/src/client/stackerdb.rs
```diff
@@ -90,8 +115,146 @@ impl StackerDB {
...
+            self.signers_stackerdb_session
+                .get_latest_chunks(&slot_ids)
+                .map_err(backoff::Error::transient)
+        };
+        let chunk_ack = retry_with_exponential_backoff(send_request)?;
+        let mut transactions = Vec::new();
+        for (i, chunk) in chunk_ack.iter().enumerate() {
+            if let Some(data) = chunk {
+                if let Ok(message) = bincode::deserialize::<SignerMessage>(data) {
+                    if let SignerMessage::Transactions(chunk_transactions) = message {
```
kantai, line 141: Kind of a nit, but I think it's better to try to avoid deep nesting like this by inverting the checks or using `let...else`:

```rust
let Some(data) = chunk else {
    continue;
}
let Ok(message) = bincode::deserialize::<SignerMessage>(data) else {
    warn!("Failed to deserialize chunk data into a SignerMessage");
    continue;
}
...
```

stacks-signer/src/utils.rs
```diff
@@ -138,3 +143,18 @@ pub fn build_stackerdb_contract(
...
+        Network::Testnet | Network::Mocknet => C32_ADDRESS_VERSION_TESTNET_SINGLESIG,
+    };
+    StacksAddress::from_public_keys(
+        version,
+        &AddressHashMode::SerializeP2PKH,
+        1,
+        &vec![StacksPublicKey::from_private(stacks_private_key)],
+    )
+    .expect("BUG: failed to generate address from private key")
+}
```
kantai, line 160: I'd suggest using a `mainnet` bool and the `p2pkh()` method in `StacksAddress`:

```suggestion
pub fn to_addr(stacks_private_key: &StacksPrivateKey, network: &Network) -> StacksAddress {
    // this could be a function of Network, so you would call `network.is_mainnet()`
    let is_mainnet = match network {
        Network::Mainnet => true,
        Network::Testnet | Network::Mocknet => false,
    };
    StacksAddress::p2pkh(is_mainnet, &StacksPublicKey::from_private(stacks_private_key))
}
```

That way, we avoid a `vec![]`, an `.expect()`, and also this code doesn't have to worry about getting the version flags right.

stacks-signer/src/runloop.rs
```diff
@@ -251,69 +263,80 @@ impl<C: Coordinator> RunLoop<C> {
...
                     return;
                 };
                 block_info.valid = Some(false);
                 // Submit a rejection response to the .signers contract for miners
                 // to observe so they know to send another block and to prove signers are doing work);
-                if let Err(e) = self.stackerdb.send_message_with_retry(
-                    self.signing_round.signer_id,
-                    block_validate_reject.into(),
-                ) {
+                debug!("Broadcasting a block rejection due to stacks node validation failure...");
```
kantai, line 292: Should this be a `warn!`?

kantai (approved):
This LGTM, just a few superficial comments.

stacks-signer/src/runloop.rs
jcnelson: Yeah I agree

jcnelson (approved):
LGTM; just address Aaron's comments before merging and make sure CI is passing.

Issue #4298, opened 2024-01-26
[Nakamoto] Create an RPC endpoint to retrieve the list of signer public keys and their total number of reward slots (vote shares)


jferrant:
In order for signers to be configured, there needs to be an endpoint available to call to retrieve the list of registered signer public keys and their corresponding number of vote shares / reward slots (key_ids)

xoloki:
> In order for signers to be configured, there needs to be an endpoint available to call to retrieve the list of registered signer public keys and their corresponding number of vote shares / reward slots (key_ids)

To be clear, we need the signers' actual key_ids, not just the number of them.

jferrant:
> > In order for signers to be configured, there needs to be an endpoint available to call to retrieve the list of registered signer public keys and their corresponding number of vote shares / reward slots (key_ids)
> 
> To be clear, we need the signers' actual key_ids, not just the number of them.

It can just be the number. We would then create the key ids manually. I.e. if alice's key has 100 shares and is the first in the returned list, she becomes signer_id 0 with key ids 1 through 100

Issue #4295, opened 2024-01-25
Reward set must be written to and read from .signers


kantai:
In my PR #4269, I left as future work:

```rust
    pub fn get_reward_set_nakamoto(
        &self,
        cycle_start_burn_height: u64,
        chainstate: &mut StacksChainState,
        burnchain: &Burnchain,
        sortdb: &SortitionDB,
        block_id: &StacksBlockId,
    ) -> Result<RewardSet, Error> {
        // TODO: this method should read the .signers contract to get the reward set entries.
        //   they will have been set via `NakamotoChainState::check_and_handle_prepare_phase_start()`.
        let cycle = burnchain
            .block_height_to_reward_cycle(cycle_start_burn_height)
            .expect("FATAL: no reward cycle for burn height");
```

Roughly, my understanding of how this would need to work is as follows:

1. .signers update occurs in the setup step of the first block in a prepare phase. This update writes the signers stackerdb data (currently implemented), and the reward set state to .signers (not implemented).
2. get_reward_set_nakamoto() reads the reward set state directly from .signers, performing no calculations: its just essentially loading a vec.

The "reward set provider" is still invoked, just what it is doing is different. And the first time its invoked, it stashes its results in the SortitionDB as well so that it can be read directly from there.

The way it works now is that (1) and (2) independently calculate the reward set. That's not tenable because it introduces the possibility of different results even without a bug: one calculates at the start of the block and the other calculates at the end of the same block. This issue would resolve that.

PR #4311, opened 2024-01-30
Reward set calculation and RPC endpoint

kantai:
### Description

Implement #4295 and #4298

### 4295: Reading the reward set data from `.signers`

Implementation of #4295 is kind of subtle, and will have to impose additional constraints on Epoch-2.5 (one of which isn't implemented in this PR, but could be added, I suppose). 

Reading the reward set state from the calculation performed for `.signers` is only possible _after_ that calculation is performed (due to the general nature of causality and the linear passage of time). This calculation is performed during the first block of a prepare phase. This does not pose a challenge for nakamoto, because that's the block it chooses as the PoX anchor block anyways, so, of course it can read that data. For Epoch 2.5, however, this _is_ a challenge: the PoX anchor block always comes _before_ the prepare phase.

To deal with this, this PR splits the reward set loading by epoch -- in Nakamoto, it reads from the `.signers`, but in Epoch 2.5, it continues to use the 2.x reward set routine. However, we still want to ensure that the calculation done by the 2.x reward set routine at the PoX anchor block and the calculation performed by `.signers` work out to the same result. The PoX anchor block (if one is chosen) in any given fork is the _last block before the prepare phase_. This means that the `.signers`
calculation will always occur with the PoX anchor block as its anchor block parent. Because the 2.x reward set routine uses the state at the end of block processing, we can ensure that the reward set calculations are the same if:

1. The `.signers` calculation occurs before any burn operation processing.
2. There is no microblock.

This PR implements the first step there. For the second step, I'd propose to disable microblocks in Epoch 2.5.

### 4298: New RPC endpoint `/v2/stacker_set/{cycle_number}

This endpoint returns the same `RewardSet` JSON data propagated over the event observer interface. Callers must supply a cycle number to query for. It will return 400 if the set hasn't been calculated yet or if the requested cycle wasn't a PoX-4 cycle.

### DIFF
--- a/clarity/src/vm/database/clarity_db.rs
+++ b/clarity/src/vm/database/clarity_db.rs
@@ -619,6 +619,22 @@ impl<'a> ClarityDatabase<'a> {
             .map_err(|e| e.into())
     }
 
+    /// Set a metadata entry if it hasn't already been set, yielding
+    ///  a runtime error if it was. This should only be called by post-nakamoto
+    ///  contexts.
+    pub fn try_set_metadata(
+        &mut self,
+        contract_identifier: &QualifiedContractIdentifier,
+        key: &str,
+        data: &str,
+    ) -> Result<()> {
+        if self.store.has_metadata_entry(contract_identifier, key) {
+            Err(Error::Runtime(RuntimeErrorType::MetadataAlreadySet, None))
+        } else {
+            Ok(self.store.insert_metadata(contract_identifier, key, data))
+        }
+    }
+
     fn insert_metadata<T: ClaritySerializable>(
         &mut self,
         contract_identifier: &QualifiedContractIdentifier,
--- a/clarity/src/vm/database/clarity_db.rs
+++ b/clarity/src/vm/database/clarity_db.rs
@@ -619,22 +619,6 @@ impl<'a> ClarityDatabase<'a> {
             .map_err(|e| e.into())
     }
 
-    /// Set a metadata entry if it hasn't already been set, yielding
-    ///  a runtime error if it was. This should only be called by post-nakamoto
-    ///  contexts.
-    pub fn try_set_metadata(
-        &mut self,
-        contract_identifier: &QualifiedContractIdentifier,
-        key: &str,
-        data: &str,
-    ) -> Result<()> {
-        if self.store.has_metadata_entry(contract_identifier, key) {
-            Err(Error::Runtime(RuntimeErrorType::MetadataAlreadySet, None))
-        } else {
-            Ok(self.store.insert_metadata(contract_identifier, key, data))
-        }
-    }
-
     fn insert_metadata<T: ClaritySerializable>(
         &mut self,
         contract_identifier: &QualifiedContractIdentifier,
--- a/docs/rpc/api/core-node/get_stacker_set.400.example.json
+++ b/docs/rpc/api/core-node/get_stacker_set.400.example.json
@@ -1,4 +1,4 @@
 {
   "response": "error",
-  "err_msg": "Could not read reward set. Prepare phase may not have started for this cycle yet. Cycle = 22, Err= PoXAnchorBlockRequired"
+  "err_msg": "Could not read reward set. Prepare phase may not have started for this cycle yet. Cycle = 22, Err = PoXAnchorBlockRequired"
 }
--- a/docs/rpc/api/core-node/get_stacker_set.400.example.json
+++ b/docs/rpc/api/core-node/get_stacker_set.400.example.json
@@ -1,0 +1,4 @@
+{
+  "response": "error",
+  "err_msg": "Could not read reward set. Prepare phase may not have started for this cycle yet. Cycle = 22, Err= PoXAnchorBlockRequired"
+}
--- a/docs/rpc/openapi.yaml
+++ b/docs/rpc/openapi.yaml
@@ -593,7 +593,7 @@ paths:
       description: |
         Used to get stacker and signer set information for a given cycle.
 
-        This will only return information for cycles started in Epoch-2.5 where PoX-4 was active.
+        This will only return information for cycles started in Epoch-2.5 where PoX-4 was active and subsequent cycles.
       responses:
         200:
           description: Information for the given reward cycle
--- a/docs/rpc/openapi.yaml
+++ b/docs/rpc/openapi.yaml
@@ -583,3 +583,27 @@ paths:
           application/json:
             example:
               $ref: ./api/core-node/post-block-proposal-req.example.json
+
+  /v2/stacker_set/{cycle_number}:
+    get:
+      summary: Fetch the stacker and signer set information for a given cycle.
+      tags:
+        - Mining
+      operationId: get_stacker_set
+      description: |
+        Used to get stacker and signer set information for a given cycle.
+
+        This will only return information for cycles started in Epoch-2.5 where PoX-4 was active.
+      responses:
+        200:
+          description: Information for the given reward cycle
+          content:
+            application/json:
+              example:
+                $ref: ./api/core-node/get_stacker_set.example.json
+        400:
+          description: Could not fetch the given reward set
+          content:
+            application/json:
+              example:
+                $ref: ./api/core-node/get_stacker_set.400.example.json
--- a/stackslib/src/chainstate/coordinator/mod.rs
+++ b/stackslib/src/chainstate/coordinator/mod.rs
@@ -327,7 +327,9 @@ impl<'a, T: BlockEventDispatcher> RewardSetProvider for OnChainRewardSetProvider
         //
         //  Data **cannot** be read from `.signers` in epoch 2.5 because the write occurs
         //   in the first block of the prepare phase, but the PoX anchor block is *before*
-        //   the prepare phase. Therefore
+        //   the prepare phase. Therefore, we fetch the reward set in the 2.x style, and then
+        //   apply the necessary nakamoto assertions if the reward set is going to be
+        //   active in Nakamoto (i.e., check for signer set existence).
 
         let is_nakamoto_reward_set = match SortitionDB::get_stacks_epoch_by_epoch_id(
             sortdb.conn(),
--- a/stackslib/src/chainstate/coordinator/mod.rs
+++ b/stackslib/src/chainstate/coordinator/mod.rs
@@ -380,6 +380,7 @@ impl<'a, T: BlockEventDispatcher> RewardSetProvider for OnChainRewardSetProvider
             burnchain,
             sortdb,
             block_id,
+            false,
         )
     }
 }
--- a/stackslib/src/chainstate/coordinator/mod.rs
+++ b/stackslib/src/chainstate/coordinator/mod.rs
@@ -286,6 +286,15 @@ pub trait RewardSetProvider {
         sortdb: &SortitionDB,
         block_id: &StacksBlockId,
     ) -> Result<RewardSet, Error>;
+
+    fn get_reward_set_nakamoto(
+        &self,
+        cycle_start_burn_height: u64,
+        chainstate: &mut StacksChainState,
+        burnchain: &Burnchain,
+        sortdb: &SortitionDB,
+        block_id: &StacksBlockId,
+    ) -> Result<RewardSet, Error>;
 }
 
 pub struct OnChainRewardSetProvider<'a, T: BlockEventDispatcher>(pub Option<&'a T>);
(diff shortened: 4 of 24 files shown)


clarity/src/vm/database/clarity_db.rs
```diff
@@ -611,6 +611,22 @@ impl<'a> ClarityDatabase<'a> {
         self.store.insert_metadata(contract_identifier, key, data);
     }
 
+    /// Set a metadata entry if it hasn't already been set, yielding
+    ///  a runtime error if it was. This should only be called by post-nakamoto
+    ///  contexts.
```
jcnelson, line 616: Can you help me understand how this error could be triggered?

stackslib/src/chainstate/coordinator/mod.rs
```diff
@@ -312,6 +321,14 @@ impl<'a, T: BlockEventDispatcher> RewardSetProvider for OnChainRewardSetProvider
         let cycle = burnchain
             .block_height_to_reward_cycle(cycle_start_burn_height)
             .expect("FATAL: no reward cycle for burn height");
+        // `self.get_reward_set_nakamoto` reads the reward set from data written during
+        //   updates to .signers
+        // `self.get_reward_set_epoch2` reads the reward set from the `.pox-*` contract
+        //
+        //  Data **cannot** be read from `.signers` in epoch 2.5 because the write occurs
+        //   in the first block of the prepare phase, but the PoX anchor block is *before*
+        //   the prepare phase. Therefore
```
jcnelson, line 330: unfinished thought?

stackslib/src/chainstate/coordinator/tests.rs
```diff
@@ -520,6 +520,17 @@ impl RewardSetProvider for StubbedRewardSetProvider {
...
+
+    fn get_reward_set_nakamoto(
+        &self,
+        cycle_start_burn_height: u64,
+        chainstate: &mut StacksChainState,
+        burnchain: &Burnchain,
+        sortdb: &SortitionDB,
+        block_id: &StacksBlockId,
+    ) -> Result<RewardSet, CoordError> {
+        panic!("Stubbed reward set provider cannot be invoked in nakamoto")
```
jcnelson, line 532: Is there a corresponding ticket for this?

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -53,59 +54,115 @@ use crate::util_lib::db::Error as DBError;
...
+            .eval_boot_code_read_only(
+                sortdb,
+                block_id,
+                SIGNERS_NAME,
+                &format!("(map-get? cycle-set-height u{})", cycle),
+            )?
+            .expect_optional()
+            .map(|x| u64::try_from(x.expect_u128()).expect("FATAL: block height exceeded u64"))
+        else {
+            if debug_log {
```
jcnelson, line 84: Is there a way to DRY this up?  Perhaps a custom logging function?

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -53,59 +54,115 @@ use crate::util_lib::db::Error as DBError;
...
+        debug_log: bool,
     ) -> Result<RewardSet, Error> {
-        // TODO: this method should read the .signers contract to get the reward set entries.
-        //   they will have been set via `NakamotoChainState::check_and_handle_prepare_phase_start()`.
         let cycle = burnchain
             .block_height_to_reward_cycle(cycle_start_burn_height)
             .expect("FATAL: no reward cycle for burn height");
+        // figure out the block ID
+        let Some(coinbase_height_of_calculation) = chainstate
+            .eval_boot_code_read_only(
```
jcnelson, line 75: Doesn't have to be done now, but this is going to be relatively slow while being on a network hot-path.  Is there a way this could be cached?

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
```diff
@@ -53,59 +54,115 @@ use crate::util_lib::db::Error as DBError;
...
+            &reward_set_block.index_block_hash(),
+        )?
+        else {
+            if debug_log {
+                debug!(
+                    "No reward set stored at the block in which .signers was written";
+                    "checked_block" => %reward_set_block.index_block_hash()
+                );
+            } else {
+                error!(
```
jcnelson, line 123: Got a log discrepancy (needs to be DRY'ed up)

stackslib/src/chainstate/nakamoto/signer_set.rs
```diff
@@ -0,0 +1,387 @@
+// Copyright (C) 2013-2020 Blockstack PBC, a public benefit corporation
+// Copyright (C) 2020-2023 Stacks Open Internet Foundation
```
jcnelson, line 2: 2024

stackslib/src/net/api/getstackers.rs
```diff
@@ -0,0 +1,225 @@
...
+            .ok_or("Failed to lookup PoX contract version at tip")?;
+        if !matches!(pox_version, PoxVersions::Pox4) {
+            return Err(
+                "Active PoX contract version at tip is Pre-PoX-4, the signer set is not fetchable"
+                    .into(),
+            );
+        }
+
+        let provider = OnChainRewardSetProvider::new();
+        let stacker_set = provider.read_reward_set_nakamoto(
```
jcnelson, line 77: This is something you could potentially cache in RAM in perpetuity.

stackslib/src/net/api/mod.rs
```diff
@@ -114,6 +115,7 @@ impl StacksHttp {
         self.register_rpc_endpoint(postmicroblock::RPCPostMicroblockRequestHandler::new());
         self.register_rpc_endpoint(poststackerdbchunk::RPCPostStackerDBChunkRequestHandler::new());
         self.register_rpc_endpoint(posttransaction::RPCPostTransactionRequestHandler::new());
+        self.register_rpc_endpoint(getstackers::GetStackersRequestHandler::default());
```
jcnelson, line 118: Miner code aesthetic: can you place the call to `self.register_rpc_endpoint()` in alphabetical order with the other RPC endpoints?  So, after the one for `getstackerdbmetadata`?

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -68,13 +68,19 @@ fn advance_to_nakamoto(
     )
     .unwrap();
 
+    // use the signing key of addr, otherwise the test stackers
+    //  will not stack enough for any single signing key
+    // let signing_key = StacksPublicKey::from_private(&private_key);
```
jcnelson, line 73: Commented code

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -56,7 +56,7 @@ use crate::util_lib::boot::boot_code_id;
 fn advance_to_nakamoto(
     peer: &mut TestPeer,
     test_signers: &TestSigners,
-    test_stackers: Vec<TestStacker>,
+    test_stackers: &[TestStacker],
```
jcnelson, line 59: Somewhere in this file, can you add a basic unit test (or extend the existing ones) so they load and validate the reward set from the chainstate?  Just so we have some targeted test coverage on these new methods, in addition to the indirect ones we get from querying it via RPC? Thanks!

clarity/src/vm/database/clarity_db.rs
kantai: Actually, this function should just be removed -- at first I was going to try to store the computed reward set as metadata, in which case, this function might have been needed. But metadata storage in Clarity is very closely tied to contract instantiation, so it would be really cumbersome to have done things this way.

stackslib/src/chainstate/coordinator/tests.rs
kantai: No, and there shouldn't be -- this is a struct only ever used in testing. If it would be helpful to test something in Nakamoto with a stubbed reward set provider, we can update this then.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
kantai: Yeah, a new macro definition would be necessary. I can add.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
kantai: Yeah -- if we want to cache this in the network path, we could certainly do so there (or cache the whole struct).

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
kantai: We get them both from the RPC, but also all of the existing nakamoto coordinator tests -- the nakamoto coordinator always uses the same path as the RPC endpoint.

stackslib/src/net/api/mod.rs
kantai: Sure, but are we sure it shouldn't be after `getstackerdbchunk` and before `getpoxinfo` ;) ?

stackslib/src/chainstate/coordinator/mod.rs
kantai: Yes, I think I completed the thought in my latest commit, but it's hard to know exactly where I was planning to go with that sentence.

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
kantai: Just trying to understand what you're asking for here-- the nakamoto coordinator loads its reward sets from the Chainstate DB whenever it uses the `OnChainProvider`, which is true in all of the non-unit coordinator tests. So this is covered in several different integration tests. The functionality of actually loading the reward set state in the coordinator is inherently an integration. In 2.x, its an integration between the coordinator, pox-3, the on chain provider, and the SortitionDB. In Nakamoto, its an integration between the coordinator, pox-4, the on chain provider, the SortitionDB, and the block processing methods. So, this functionality at least has to be tested via integration tests (which it is). The functionality of computing the reward set is just a unit (it's just a function applied to the PoX-4 state), and that _is_ tested via unit tests (implemented in the PR that this one built off of).

stackslib/src/net/api/getstackers.rs
kantai: Sure -- but I'd prefer to leave that optimization for elsewhere.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
kantai: Added a local macro for this.

stackslib/src/chainstate/nakamoto/coordinator/mod.rs
jcnelson: Yeah, was thinking we'd cache the whole struct.  Can do this in a separate PR though.

jcnelson (approved):

jcnelson:
LGTM; just fix merge conflicts and make sure CI passes.  Thanks!

docs/rpc/api/core-node/get_stacker_set.400.example.json
```diff
@@ -0,0 +1,4 @@
+{
+  "response": "error",
+  "err_msg": "Could not read reward set. Prepare phase may not have started for this cycle yet. Cycle = 22, Err= PoXAnchorBlockRequired"
```
obycode, line 3:
```suggestion
  "err_msg": "Could not read reward set. Prepare phase may not have started for this cycle yet. Cycle = 22, Err = PoXAnchorBlockRequired"
```

docs/rpc/openapi.yaml
```diff
@@ -583,3 +583,27 @@ paths:
...
+  /v2/stacker_set/{cycle_number}:
+    get:
+      summary: Fetch the stacker and signer set information for a given cycle.
+      tags:
+        - Mining
+      operationId: get_stacker_set
+      description: |
+        Used to get stacker and signer set information for a given cycle.
+
+        This will only return information for cycles started in Epoch-2.5 where PoX-4 was active.
```
obycode, line 596: This will not continue to work in epoch 3.0?

jferrant (approved):
LGTM :D

obycode (commented):
This looks good. I just had two minor comments.

obycode (approved):
👍

PR #4318, opened 2024-01-31
Signers include coordinator metadata in messages and define NACK policy

### LINKED ISSUE #4292: Signers should include their stacks_tip_consensus_hash view in their StackerDB messages
Stacks-signer uses the stacks_tip_consensus_hash to introduce randomness in selecting a coordinator (part of #4276). This issue is related to the feedback on the implications of this selection mechanism and the necessary steps to distinguish slower or divergent signers from Byzantine ones. To address this, signers should include their view of the stacks_tip_consensus_hash in StackerDB messages.

Below is a detailed specification:
https://github.com/stacks-network/stacks-core/pull/4276#discussion_r1465399027

8marz8:
### Description
Enhancements are being made to signer messages by including coordinator metadata that reflects each signer's view of the stacks chain. This allows signers to compare the received metadata with their own chain view, determining if they are lagging behind. In cases where a signer identifies that they possess a more current view of the chain, they can issue a NACK message.

When configured to acknowledge NACK messages, the local signer accumulates and counts these messages. Upon reaching a predetermined threshold, it initiates a back-off delay. Both the threshold for triggering this delay and its duration are configurable settings for the signer.

### Applicable issues

- fixes #4292


libsigner/src/events.rs
```diff
@@ -61,14 +63,106 @@ const SIGNATURE_SHARE_REQUEST_SLOT_ID: u32 = 8;
...
+        Self {
+            coordinator_id,
+            coordinator_public_key,
+            stacks_consensus_hash,
+            stacks_block_height,
+        }
+    }
+}
+
+impl PartialEq for CoordinatorInfo {
```
jferrant, line 152: I wonder can this be auto inferred or does PublicKey not use PartialEq?

8marz8: yep, PublicKey doesn't

xoloki: Yeah I ran into this in my PR.  As a workaround I passed around Points instead of PublicKeys.

libsigner/src/messages.rs
```diff
@@ -927,11 +998,70 @@ impl SignerMessage {
...
+            Self::Nack(_) => NACK_SLOT_ID,
         };
         SIGNER_SLOTS_PER_USER * id + slot_id
     }
 }
 
+/// Signed network packets sent by signers along with their view
+/// of stacks tip consensus hash and block height for
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct PacketInfo {
```
xoloki, line 1010: Why is the packet struct PacketInfo, while the coordinator struct is CoordinatorMetadata?  It seems on first blush that they could have the same suffix.

8marz8: I can rename it to PacketMetadata instead of PacketInfo. 
Initially I only had one CoordinatorInfo that included CoordinatorMetadata fields in this file but I ran into issues for serialization/deserialization of PublicKey based on the StacksMessageCodec requirements so I decided to break it apart since I didn't really need the publicKey/coordinator_id as part of messages.

xoloki: I'd prefer the same suffix, but it's not a huge deal.

stacks-signer/src/runloop.rs
```diff
@@ -176,12 +212,24 @@ impl<C: Coordinator> RunLoop<C> {
     /// Execute the given command and update state accordingly
     /// Returns true when it is successfully executed, else false
     fn execute_command(&mut self, command: &RunLoopCommand) -> bool {
+        let coordinator_info = calculate_coordinator(
```
hstove, line 215: Right here we only need `coordinator_metadata`, so I think it would be better to have a function like `calculate_coordinator_metadata(...)`. And then at the top of `calculate_coordinator` we can call that function.

stacks-signer/src/runloop.rs
```diff
@@ -937,21 +1071,60 @@ impl<C: Coordinator> SignerRunLoop<Vec<OperationResult>, RunLoopCommand> for Run
...
+        .find_map(|(coordinator_metadata, signers)| {
+            if signers.len() >= *nack_threshold as usize {
+                Some(coordinator_metadata.clone())
+            } else {
+                None
+            }
+        })
+}
+
+fn wrap_packet_in_packet_metadata(
```
hstove, line 1089: IMO this function isn't necessary, just use `PacketMetadata::new`

stacks-signer/src/runloop.rs
```diff
@@ -788,15 +861,73 @@ impl<C: Coordinator> RunLoop<C> {
...
+        if current_signer_consensus_hash != packet_sender_consensus_hash {
+            if current_signer_block_height < packet_sender_block_height {
+                warn!(
+                    "Signer {} has an outdated view of stacks chain",
+                    self.signing_round.signer_id
+                );
+            } else {
+                self.broadcast_nack_message(current_signer_coordinator_metadata.clone());
+            }
+        }
```
hstove, line 902: I'm not exactly sure what the spec is here, but my assumption would be that the signer should broadcast a `nack` for any messages with mismatched coordinator metadata. Is that not true, and it's only when heights are equal?

stacks-signer/src/runloop.rs
```diff
@@ -788,15 +861,73 @@ impl<C: Coordinator> RunLoop<C> {
...
+        let packet_sender_consensus_hash = packet_sender_coordinator_metadata.stacks_consensus_hash;
+        let packet_sender_block_height = packet_sender_coordinator_metadata.stacks_block_height;
+        if current_signer_consensus_hash != packet_sender_consensus_hash {
+            if current_signer_block_height < packet_sender_block_height {
+                warn!(
+                    "Signer {} has an outdated view of stacks chain",
+                    self.signing_round.signer_id
+                );
+            } else {
+                self.broadcast_nack_message(current_signer_coordinator_metadata.clone());
```
hstove, line 900: A `warn!` would also probably be good here

stacks-signer/src/runloop.rs
```diff
@@ -873,6 +1004,9 @@ impl From<&Config> for RunLoop<FireCoordinator<v2::Aggregator>> {
             signer_ids: config.signer_ids.clone(),
             signer_id: config.signer_id,
             signer_addresses: vec![],
+            nack_messages_received: HashMap::new(),
+            nack_messages_sent: HashSet::new(),
+            nack_threshold: (total_signers + 1) / 2,
```
hstove, line 1009: This should be a `get_nack_threshold` function, which returns the threshold based off of `coordinator.total_signers`. That way when we later have code to update `coordinator` for new reward cycles, we don't have to manually update the nack threshold too.

stacks-signer/src/runloop.rs
```diff
@@ -937,21 +1071,60 @@ impl<C: Coordinator> SignerRunLoop<Vec<OperationResult>, RunLoopCommand> for Run
     }
 }
 
+fn check_nack_threshold(
```
hstove, line 1074: Add comments here and maybe rename it. It took me a minute to understand this was a function to "find the first of coordinator metadata where the nack threshold is reached"

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -7,6 +7,7 @@ use std::time::{Duration, Instant};
 use std::{env, thread};
 
 use clarity::vm::types::QualifiedContractIdentifier;
+use hashbrown::HashMap as hash_map;
```
hstove, line 10: unused

hstove (changes requested):
Overall this is pretty good, but I have some comments about how we're calculating whether a nack threshold is reached. I'm not 100% sure in my comments about how the logic should work, so please let me know if I have some mistaken assumptions.

testnet/stacks-node/src/tests/signer.rs
8marz8: I use this to pass a hashbrown hashmap to calculate_coordinator

stacks-signer/src/runloop.rs
8marz8: Sure, I can remove it. I added it to keep the .into() conversion pattern used before the same

stacks-signer/src/runloop.rs
8marz8: I could be wrong here but I when the current signer is behind, I don't send a Nack. I only do so when the current signer is more updated than what was stamped by other signers....Assuming that when this signer sends a packet with its view, other signers will eventually send a Nack with their more updated view.
I need to add an else if case for when the consensus_hash is not the same but the heights are the same....I am guessing that could mean there's a fork.

stacks-signer/src/runloop.rs
8marz8: Thanks for pointing that out. Changing that function to return the coordinator_metadata that has the most broadcast _if_  it also meets the threshold.
I'm also adding a cleanup method that gets invoked whenever the current signer has a more updated view of the chain.

libsigner/src/messages.rs
```diff
@@ -927,11 +998,70 @@ impl SignerMessage {
...
+            Self::Nack(_) => NACK_SLOT_ID,
         };
         SIGNER_SLOTS_PER_USER * id + slot_id
     }
 }
 
+/// Signed network packets sent by signers along with their view
+/// of stacks tip consensus hash and block height for
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct PacketMetadata {
```
jcnelson, line 1010: Nit: calling this `PacketMetadata` is confusing, because it contains both metadata and a payload.  Maybe this could be called something more like `PacketMessage` in order to convey that it's a container for these two things?

stacks-signer/src/runloop.rs
```diff
@@ -937,21 +1069,98 @@ impl<C: Coordinator> SignerRunLoop<Vec<OperationResult>, RunLoopCommand> for Run
...
+    nack_threshold: &u32,
+) -> Option<CoordinatorMetadata> {
+    nack_messages_received
+        .iter()
+        .filter(|(_, signers)| signers.len() >= *nack_threshold as usize)
+        .max_by_key(|(_, signers)| signers.len())
+        .map(|(coordinator_metadata, _)| *coordinator_metadata)
+}
+
+/// Create coordinator Info with fake consensus hash and block height to not block signers if they can't calculate
```
jcnelson, line 1108: Typo "to not block signers"

stacks-signer/src/runloop.rs
```diff
@@ -937,21 +1069,98 @@ impl<C: Coordinator> SignerRunLoop<Vec<OperationResult>, RunLoopCommand> for Run
...
+/// coordinator properly even after retries. This will result to NACKs and will eventually get ignored/overthrown
+fn get_default_coordinator_info(signer_id: u32, public_key: PublicKey) -> CoordinatorInfo {
+    let default_coordinator_metadata = get_default_coordinator_metadata();
+    CoordinatorInfo::new(signer_id, public_key, default_coordinator_metadata)
+}
+
+fn get_default_coordinator_metadata() -> CoordinatorMetadata {
```
jcnelson, line 1115: Nit: instead of having a function for this, can you just implement `Default` for `CoordinatorMetadata`?

stacks-signer/src/runloop.rs
```diff
@@ -937,21 +1069,98 @@ impl<C: Coordinator> SignerRunLoop<Vec<OperationResult>, RunLoopCommand> for Run
...
+    CoordinatorMetadata::new(ConsensusHash([0u8; 20]), 0)
+}
+
+/// Attempts to return the most up-to-date CoordinatorMetadata based on the current signer's view
+/// and the NACK messages received. If NACK messages indicate a majority view with a higher block height
+/// than the current signer's, the function updates to that view. Otherwise, it retains the current signer's view
```
jcnelson, line 1121: I don't think this is what we want to do.  The signer should be querying its Stacks node for the latest chain view, and _only_ its Stacks node.  The Stacks nodes take care of making sure they all have the same chain view as one another.  What we don't want is to allow other signers to compel the local signer to assume a chain view that its local node does not believe exists.  If this happened, then the data returned by `stacks_client.rs` would be inconsistent with the signer's chain view, and thus unreliable and useless for making progress on signing rounds and DKG.  For example, the signer would be unable to validate blocks since it would have two conflicting views of the Stacks chain tip.

8marz8: @jcnelson For 'The Stacks nodes take care of making sure they all have the same chain view as one another,' then when would the case arise where one signer might have a more outdated view? Are we suggesting that it's possible, though not a frequent event, since the nodes are already tasked with ensuring consistency? If the NACKs aren't necessary for determining the coordinator, could I remove that logic and simply use the NACKs for logging purposes only? If that's the case, I think I'll need to revise my NackMessage as it currently includes the signer_id of the one that is more updated and broadcasts its CoordinatorMetadata, which isn't directly aimed at the signers who are falling behind.

libsigner/src/messages.rs
8marz8: Sure, but `Packet` type holds a msg:Message so maybe that'll be repetitive too. 
I had it as `PacketInfo` before. Can rename to `PacketMessage` for now and we can alway refactor.

stacks-signer/src/runloop.rs
```diff
@@ -873,6 +1002,9 @@ impl From<&Config> for RunLoop<FireCoordinator<v2::Aggregator>> {
             signer_ids: config.signer_ids.clone(),
             signer_id: config.signer_id,
             signer_addresses: vec![],
+            nack_messages_received: HashMap::new(),
+            nack_messages_sent: HashSet::new(),
+            nack_threshold: get_nack_threshold(total_signers),
```
hstove, line 1007: Sorry to nit (again on this haha) but I think this should be a method of `RunLoop` so that it is always "tied" to the number of signers set elsewhere in the struct. It's just one less way of making a mistake in the future.

ie:

```rs
impl<C: Coordinator> RunLoop {
    // other code
    
    fn get_nack_threshold(&self) -> u32 {
        (&self.coordinator.get_config().num_signers + 1) / 2
    }
}
```

stacks-signer/src/runloop.rs
```diff
@@ -937,21 +1069,98 @@ impl<C: Coordinator> SignerRunLoop<Vec<OperationResult>, RunLoopCommand> for Run
...
+
+/// Clean up outdated coordinator_metadata NACKs that have a stale or equivalent chain view compared to the current signer
+fn cleanup_outdated_coordinator_nacks(
+    nack_messages_received: &mut HashMap<CoordinatorMetadata, HashSet<u32>>,
+    current_observed_height: u64,
+) {
+    let outdated_metadata_to_remove: Vec<CoordinatorMetadata> = nack_messages_received
+        .iter()
+        .filter(|(coordinator_metadata, _)| {
+            coordinator_metadata.stacks_block_height <= current_observed_height
```
hstove, line 1086: I'm not exactly sure how to think about this in Nakamoto (chime in anyone) but I think we should wait at least a few blocks until we remove old nacks.

Also, I just learned about [`HashMap::retain`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.retain) which is an easy "iterate and keep some items" method. Tiny impact in this case but it'll simplify the code too.

hstove (changes requested):
Looking good. I'd like to see just a few more unit tests that demonstrate how coordinator metadata is calculated with Nacks. Also some basic tests around ensuring the threshold is always >=50%. For example, check 4 nacks with 7 signers, 4 nacks with 8 signers, 3 with 7, etc. Just to make sure we don't catch off-by-one errors now or in the future

stacks-signer/src/runloop.rs
8marz8: This makes more sense!

stacks-signer/src/runloop.rs
8marz8: @jcnelson if you could confirm, I can proceed by deleting the nack_threshold part and how its used as the local signer should keep its coordinator calculation as is and only use NACKs for logging purposes and for that I will revise my NackMessage a bit too.

jcnelson: > when would the case arise where one signer might have a more outdated view?

It is possible that one signer's node is behind the chain tip.

> Are we suggesting that it's possible, though not a frequent event, since the nodes are already tasked with ensuring consistency?

Correct.

> If the NACKs aren't necessary for determining the coordinator, could I remove that logic and simply use the NACKs for logging purposes only? 

Yes, that is what I'm suggesting.  The local signer should not be tasked with assessing whether or not its view of the chain is fresh vis-a-vis what the remote signers say.  However, the local signer _may_ treat a threshold of NACKs as a _hint_ to back off of their current request for a time, in order to give their local node a chance to catch up (but since this is a hint, the signer may ignore it as a local policy decision or choose what the back-off delay will be).

8marz8: @jcnelson Thank you for clarifying. I pushed the fixes now by removing the NACK logic from coordinator calculation. I updated message types to include the signer_id of the sender and receiver for better logging. For the local policy decision aspect, I expanded the configs with default behavior to be ignoring the hints (can change this if needed). 
The back-off delay is a simple one-time thread sleep, but if needed, I could make this a repeated back-off up to the configured duration during which I query the Node to see if the view has been updated or not.
I'm working on adding more unit tests now, but happy to make other adjustments if you get a chance to look at my changes so far.

libsigner/src/messages.rs
```diff
@@ -1039,6 +1116,87 @@ impl From<BlockValidateReject> for SignerMessage {
...
+            packet_signer_id,
+            coordinator_metadata,
+        }
+    }
+}
+
+/// Calculated Coordinator Metadata per signer's view of Stacks chain
+#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq, Serialize, Deserialize)]
+pub struct CoordinatorMetadata {
+    /// The stacks_tip_consensus_hash from the signer's view
```
jcnelson, line 1149: You should be using `pox_consensus` here, since it's the value that advances with bitcoin blocks. `stacks_tip_consensus_hash` advances with Stacks blocks.  If there's a chain stall, you'll want to use a consensus hash value that advances without the Stacks nodes' help.

stacks-signer/src/runloop.rs
```diff
@@ -969,6 +1036,118 @@ impl<C: Coordinator> RunLoop<C> {
...
+    ) {
+        let current_signer_consensus_hash =
+            current_signer_coordinator_metadata.stacks_consensus_hash;
+        let current_signer_block_height = current_signer_coordinator_metadata.stacks_block_height;
+        let packet_sender_consensus_hash = packet_sender_coordinator_metadata.stacks_consensus_hash;
+        let packet_sender_block_height = packet_sender_coordinator_metadata.stacks_block_height;
+
+        if current_signer_consensus_hash != packet_sender_consensus_hash {
+            if current_signer_block_height < packet_sender_block_height {
+                warn!(
```
jcnelson, line 1055: Alternatively, the remote signer could be lying to us

stacks-signer/src/runloop.rs
```diff
@@ -969,6 +1036,118 @@ impl<C: Coordinator> RunLoop<C> {
...
+                }
+
+                // Insert or update the entry for the new CoordinatorMetadata
+                let nack_sent_to = self
+                    .nack_messages_sent
+                    .entry(*current_signer_coordinator_metadata)
+                    .or_default();
+                // Broadcast the NACK message since it's the first time for this signer ID with the current coordinator metadata
+                if nack_sent_to.insert(packet_sender_signer_id) {
+                    warn!(
```
jcnelson, line 1076: `info!`, not `warn!`.

stacks-signer/src/runloop.rs
```diff
@@ -969,6 +1036,118 @@ impl<C: Coordinator> RunLoop<C> {
...
+
+            // Apply back-off delay if nack_threshold is reached
+            if let Some(threshold) = self.nack_threshold {
+                // Get the count of NACKs for the current coordinator metadata
+                let nack_count = self
+                    .nack_messages_received
+                    .get(&remote_coordinator_metadata)
+                    .map_or(0, |signers| signers.len());
+
+                if nack_count >= threshold as usize {
```
jcnelson, line 1140: Isn't `threshold` a percentage, not an absolute quantity?

8marz8: I accept it as percentage from the local signer in their local policy, but then calculate what that quantity would be based on total number of signers.

stacks-signer/src/runloop.rs
```diff
@@ -969,6 +1036,118 @@ impl<C: Coordinator> RunLoop<C> {
...
+                    thread::sleep(Duration::from_millis(
```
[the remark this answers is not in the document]
jcnelson: This is something @jferrant may need to weigh in on.

8marz8:
@hstove  yep, I rebased once, I can do again once her PR merges...but based on the coordinator convo, chances are none of these would really be required.

PR #4319, opened 2024-01-31
feat: add signer CLI function to generate signature

### LINKED ISSUE #4302: Signer CLI command to generate Stacking transaction information
#4277 adds a new argument `signer-sig` to some PoX transactions. This signature needs to be made with the signer's `stacks_private_key`. 

We need to expose at least one way of generating this signature for signers to be able to make these PoX transactions. Because this private key already lives in the signer's configuration, an easy way to allow signers to make this signature is to add a CLI command to the `signer-signer` binary.

In the future there are other approaches we may want to explore, like using a wallet to generate this signature.

A hypothetical "spec" of the CLI command would be:

```
stacks-signer generate-stacking-signature {reward-address} {stacking-cycle}
```

This would output `signer-sig` and `signer-public-key` in hex format.

hstove:
This PR adds a new command to the `stacks-signer` CLI:

```
stacks-signer generate-stacking-signature --pox-address bc123... --reward-cycle 100 --config ./config.toml
```

This is built on top of #4277, which isn't merged yet, so I'm opening this as a draft until that PR is reviewed and merged.

- closes #4302


stacks-signer/src/cli.rs
```diff
@@ -170,11 +173,34 @@ pub struct GenerateFilesArgs {
...
+}
+
 /// Parse the contract ID
 fn parse_contract(contract: &str) -> Result<QualifiedContractIdentifier, String> {
     QualifiedContractIdentifier::parse(contract).map_err(|e| format!("Invalid contract: {}", e))
 }
 
+/// Parse a BTC address argument and return a `PoxAddress`
+pub fn parse_pox_addr(pox_address_literal: &str) -> Result<PoxAddress, String> {
+    if let Some(pox_address) = PoxAddress::from_b58(pox_address_literal) {
```
zone117x, line 197: @hstove the `pox_address` can be bech32/segwit right? https://github.com/stacks-network/stacks-core/blob/c659fca2f3f674309139dfc5b97379d2dacc5065/stackslib/src/chainstate/stacks/address.rs#L71-L85

hstove: Yep - added some tests

zone117x: Ah that function supports all addr types, `b58` name threw me off

jferrant (approved):
LGTM :D Thankss for doing this!

stacks-signer/src/cli.rs
```diff
@@ -170,11 +174,83 @@ pub struct GenerateFilesArgs {
...
+            "agg-commit" => Pox4SignatureTopic::AggregationCommit,
+            _ => return Err(format!("Invalid topic: {}", input)),
+        };
+        Ok(topic.into())
+    }
+}
+
+#[derive(Parser, Debug, Clone)]
+/// Arguments for the generate-stacking-signature command
+pub struct GenerateStackingSignatureArgs {
```
zone117x, line 221: It would be useful to have test vectors with hex strings, e.g. containing the serialized payload for this struct. That way other folks could leverage the same vectors to be more confident in in their implementation (like in a js one). 

Not super important though unless we actually envision needing to implement similar code in js any time soon.

zone117x (approved):
LGTM, thanks @hstove!

jcnelson (commented):
Thanks for this @hstove!  Is there a way to test this in an integration test, so as to verify that the output of this command can be used in `pox-4` for a stacking signature?

hstove:
> Is there a way to test this in an integration test, so as to verify that the output of this command can be used in pox-4 for a stacking signature?

Good call, added!

Issue #4315, opened 2024-01-30
Update StackerDB and .signers to handle 8000 StackerDB list entries. 


kantai:
The StackerDB config fetching and `.signers` calculations need to be updated to support up to 8000 entries in the principal/slots list.

1. Each signer should have the same number of slots (12?)

2. The signer list (in clarity) should be able to hold up to `2 * REWARD_SLOT` number of signers. `REWARD_SLOT` is currently 4000. The maximum clarity value size is smaller than an 8000 principal list, so the interface for StackerDB config fetching must change.

For (2), there's two processes that must change:

1. The Epoch-2.5 contracts related to `.signers` must use the new
   interface that supports storing multiple list sets.

2. The StackerDB config fetching code must use the new interface.

The new StackerDB interface would be:

```clarity
;; Get a page of up to 4096 (principal, num-slots) pairs.  `page` is the page number, and starts at 0.
;; Returns a non-empty list on success
;; Returns an error of `page` is too big (see `stackerdb-get-num-signer-slot-pages`)
(define-public (stackerdb-get-signer-slots (page uint)) (response (list 4096 { signer: principal, num-slots: uint }) uint))

;; Return the number of pages of (principal, num-slots) pairs for which `stackerdb-get-signer-slots` will have data.
;; Does not fail.
(define-public (stackerdb-get-num-signer-slot-pages) (response uint uint))
```

Issue #4307, opened 2024-01-30
Alternating StackerDB signer slot details and implementation


MarvinJanssen:
From my understanding, the consensus on StackerDB/slot usage is as follows:

- The slots of the signer StackerDB are divided among the current signer set and the previous signer set.
- The slots that contain the current signer set alternates.

If that is correct, then we still need to do the following from what I can see:

- [ ] Decide how to alternate the slots and at which offset. (I remember we just do even/odd check on the reward cycle.)
- [ ] Alternate which slots we write to in `handle_signer_stackerdb_update` and make sure we don't overwrite the last set.
- [ ] Determine which slots contain the current signer set in the voting contract.
- [ ] Only allow aggregate public key votes by members of the current signer set.
- _And potentially other things._

setzeus:
Just re-arranging assignee since it looks like #4323 is implementing this.

PR #4323, opened 2024-02-01
Paging in StackerDB and update to .signers

kantai:
### Description

This adds paging to StackerDB (it also converts a lot of `expects()` to guarded errors: I'm pretty sure those weren't reachable, but easy enough to be a little paranoid with those).

~~It also updates the `.signers` (and the downstream `.signer-voting`) contracts to use paging to store the alternate signers sets. As part of this, the `.signers` contract changed some: the StackerDB slots are stored separately from the signer voting weights: the prior implementation incorrectly conflated the two.~~

EDIT (@jcnelson): I'll be taking over this PR.  The system will implement multiple StackerDBs for signers -- one per message type -- and will support running two signer-sets at once.

This implements #4307 and #4315


stackslib/src/chainstate/nakamoto/signer_set.rs
```diff
@@ -244,7 +271,7 @@ impl NakamotoSigners {
                                 "signer".into(),
                                 Value::Principal(PrincipalData::from(signing_address)),
                             ),
-                            ("num-slots".into(), Value::UInt(signer.slots.into())),
+                            ("weight".into(), Value::UInt(signer.slots.into())),
```
friedger, line 274: I am confused about the naming here.
`signer.slots` are rewards slots (=weight), not signer slots (=num-slots). Correct?

stackslib/src/chainstate/nakamoto/signer_set.rs
```diff
@@ -261,22 +289,35 @@ impl NakamotoSigners {
             );
         }
 
-        let args = [
-            SymbolicExpression::atom_value(Value::cons_list_unsanitized(signers_list).expect(
+        let set_stackerdb_args = [
```
friedger, line 292: nit: naming should be `set_signer_slots_args`.

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
```diff
@@ -364,13 +361,14 @@ fn vote_for_aggregate_public_key_in_last_block() {
     let receipts = block.receipts.as_slice();
     assert_eq!(receipts.len(), 1);
 
-    // vote should succeed
+    // vote fails because the reward cycle has changed
+    //  and the signer set hasn't been set yet.
```
friedger, line 365: How can the reward cycle change during a tenure?

friedger (commented):
I just have some nits and questions. Lgtm

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
kantai: "Bobs block" in this test is the first block of the next reward cycle (it's after the prepare phase ended).

stackslib/src/chainstate/nakamoto/signer_set.rs
kantai: I think that `signer_slots` is part of the confusion here. `stackerdb_slots` and `signer_weights` aren't the same thing -- the function it is calling is setting the stackerdb config for `.signers`, the other function is setting the signer's weights.

stackslib/src/chainstate/nakamoto/signer_set.rs
kantai: Yes, that's pretty much correct.

signer.slots are (roughly) the reward slots associated with that signer. This is also their "weight" in the aggregate public key, and in the DKG voting.

num-slots is the number of stackerdb slots that the signer needs for communication -- this is a constant for every signer.

stacks-common/src/libcommon.rs
```diff
@@ -59,4 +59,8 @@ pub mod consts {
     pub const MINER_REWARD_MATURITY: u64 = 100;
 
     pub const STACKS_EPOCH_MAX: u64 = i64::MAX as u64;
+
+    /// The number of StackerDB slots each signing key needs
+    ///  to use to participate in DKG and block validation signing.
+    pub const SIGNER_SLOTS_PER_USER: u32 = 11;
```
jcnelson, line 65: Just want to confirm with @jferrant that this value is correct.

stackslib/src/net/stackerdb/config.rs
```diff
@@ -123,108 +138,250 @@ impl StackerDBConfig {
...
+        tip: &StacksBlockId,
+    ) -> Result<Vec<(StacksAddress, u32)>, NetError> {
+        let page_count = Self::eval_page_count(chainstate, burn_dbconn, contract_id, tip)?;
+        if page_count == 0 {
+            debug!("StackerDB contract {contract_id} specified zero pages");
+            return Ok(vec![]);
+        }
+        let mut return_set: Option<Vec<(StacksAddress, u32)>> = None;
+        let mut total_num_slots = 0u32;
+        for page in 0..page_count {
```
jcnelson, line 291: I think it might make sense to have a hard bound on the number of pages allowed, so this doesn't turn into a DoS vector in other StackerDB instances.  What about 4096?  Then, you can have ~2^16 StackerDB slots, and if each of those is 16 MB, that would allow the creation of a StackerDB of up to ~250 TB.

stackslib/src/net/stackerdb/mod.rs
```diff
@@ -142,7 +142,13 @@ use crate::util_lib::boot::boot_code_id;
 use crate::util_lib::db::{DBConn, DBTx, Error as db_error};
 
 /// maximum chunk inventory size
-pub const STACKERDB_INV_MAX: u32 = 4096;
+pub const STACKERDB_INV_MAX: u32 = STACKERDB_PAGE_MAX * 2;
```
jcnelson, line 145: Why is this `* 2`?

jcnelson (changes requested):
Thanks for this!  Just two blockers:

* I don't think the signer-set contract's StackerDB is correct
* I think we want a hard cap on the number of pages

stacks-common/src/libcommon.rs
kantai: It's just moved over from the existing constant -- it's easy enough to change in a follow up PR if necessary.

stackslib/src/net/stackerdb/mod.rs
kantai: I don't want to engage with what the theoretical limits of an inventory should be here -- just want it to be big enough to fit the needs of the `.signers` contract, which I realize this constant isn't.

It should be

```
2 * 4000 * SIGNER_SLOTS_PER_USER
```

stackslib/src/net/stackerdb/config.rs
kantai: Related to my comment above about the "clarity list size" versus the "page size" -- if we enforce the clarity list size constraint, I'd be inclined to just set the page count limit at `2` for now. If there's another use case for expanding that, it could always be addressed later.

stacks-common/src/libcommon.rs
jferrant: It will become 12 soon. But its okay to be 11 right now. I can update it as needed.

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
friedger: It is in the same tenure as Alice txs

kantai: 🤔

Intriguing. I added some print statements to see what's going on:

```
INFO [1706824307.469257] [clarity/./src/vm/functions/mod.rs:620] [chainstate::stacks::boot::signers_voting_tests::vote_for_aggregate_public_key_in_last_block] (tuple (burn-height u39) (cycle u8) (event "test-1") (signer-index u0))
INFO [1706824307.471293] [stackslib/src/chainstate/stacks/db/transactions.rs:1007] [chainstate::stacks::boot::signers_voting_tests::vote_for_aggregate_public_key_in_last_block] Contract-call successfully processed, contract_name: ST000000000000000000002AMW42H.signers-voting, function_name: vote-for-aggregate-public-key, function_args: [u0, 0x020000000000000000000000000000000000000000000000000000000000000000, u0], return_value: (err u10006), cost: ExecutionCost { write_length: 88, write_count: 1, read_length: 11240, read_count: 19, runtime: 72890 }
...
INFO [1706824307.487501] [clarity/./src/vm/functions/mod.rs:620] [chainstate::stacks::boot::signers_voting_tests::vote_for_aggregate_public_key_in_last_block] (tuple (burn-height u40) (cycle u9) (event "test-1") (signer-index u1))
INFO [1706824307.488056] [stackslib/src/chainstate/stacks/db/transactions.rs:1007] [chainstate::stacks::boot::signers_voting_tests::vote_for_aggregate_public_key_in_last_block] Contract-call successfully processed, contract_name: ST000000000000000000002AMW42H.signers-voting, function_name: vote-for-aggregate-public-key, function_args: [u1, 0x020000000000000000000000000000000000000000000000000000000000000000, u0], return_value: (err u2), cost: ExecutionCost { write_length: 0, write_count: 0, read_length: 8544, read_count: 10, runtime: 33662 }
```

In alice's transaction, burn-block-height is 39, in bob's its 40. So yeah, burn-block-height is going up.

Will need to see if this is a bug in `TestPeer` or in the actual handling of burn block heights.

kantai: I think this is a bug (or unexpected feature!) in the nakamoto burn state handling in the Clarity VM -- the tip it uses is the traditional one used in Stacks: the burn block of its parent block. This means that the first block of the tenure and the second block of the tenure will have different burn views. Will open an issue for this because its not really related to this PR.

stackslib/src/chainstate/stacks/boot/signers.clar
```diff
@@ -1,24 +1,43 @@
...
 
-(define-read-only (stackerdb-get-signer-by-index (signer-index uint))
-	(ok (element-at (var-get stackerdb-signer-slots) signer-index))
-)
+(define-read-only (get-signers (cycle uint))
+     (map-get? cycle-signer-set cycle))
+
+(define-read-only (stackerdb-get-page-count) (ok u2))
+
+(define-read-only (stackerdb-get-signer-slots (page uint))
```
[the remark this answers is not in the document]
jcnelson: The page size limit comes from Clarity itself -- a variable can have type `(list 5139 { signer: principal, num-slots: uint })`, but not `(list 5140 { signer: principal, num-slots: uint } })` because it would permit a value whose representation is too big.  So we need a list size limit.  We can have an unbounded number of pages though.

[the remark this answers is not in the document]
kantai: Right— that’s what my latest commit does. It enforces three limits: first on the size of the list, second on the size of the total number of slots across all pages, and third on the number of pages. The .signers contract is implemented with 2 pages.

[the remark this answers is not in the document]
jcnelson: The page limit needs to be a lot bigger, though.  Each signer is getting 12 slots, so at 4000 possible signers, that's 48,000 slots.  So, we need at least 24 pages -- 12 pages for the current signer set, and 12 for the upcoming signer set.

stackslib/src/net/stackerdb/mod.rs
```diff
@@ -142,7 +143,15 @@ use crate::util_lib::boot::boot_code_id;
 use crate::util_lib::db::{DBConn, DBTx, Error as db_error};
 
 /// maximum chunk inventory size
-pub const STACKERDB_INV_MAX: u32 = 4096;
+pub const STACKERDB_INV_MAX: u32 = 2 * 4000 * SIGNER_SLOTS_PER_USER;
+/// maximum length of an inventory page's Clarity list
+pub const STACKERDB_PAGE_LIST_MAX: u32 = 4096;
+/// maximum number of pages that can be used in a StackerDB contract
+pub const STACKERDB_MAX_PAGE_COUNT: u32 = 2;
```
jcnelson, line 150: Per the above, this needs to be a lot bigger, since we have 48,000 slots per signer set (and there are two signer sets).

jcnelson (changes requested):
We need the maximum page size to be 24, since each signer needs 12 slots and there are two signer sets.

stackslib/src/chainstate/stacks/boot/signers.clar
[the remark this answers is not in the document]
kantai: The only limits applied to pages are (1) the total inventory size limit, which is set at 2*NUM_SLOTS_PER_USER*4000, and (2) a limit on the _list_ length of 4096. So each page can hold 4000*12 slots just fine.

jcnelson (approved):
LGTM. Was confused about the way in which the signer list was translating into chunk inventories. What you have here is good.

jcnelson (changes requested):
Temporarily rejecting this because I think there's a simpler way to do this.

jcnelson:
Regarding large stacker DBs, I'm wondering if we can actually get away with a system design that doesn't need them.  What if, instead, we have 24 `.signers` stackerdbs -- two for each slot type needed (one for the current signers, and one for the next signers).  So, there'd be something like `.signers-X-Y`, where X is the signer batch (0 or 1) and Y is the slot type (between 0 and 11).

The networking burden of having 24 stackerdbs isn't that bad if we use the configuration system to its fullest potential.  First, the number of metadata bytes exchanged per peer per stackerdb -- one `StackerDBChunkInvData` -- isn't terrible by itself.  With 4096 slots (`STACKERDB_INV_MAX`), it's only 16KB; with 24 of them, the node would be sending about 394KB per full synchronization per peer.  For example, with 32 stackerdb neighbors (`MAX_DB_NEIGHBORS`), it's not great (about 12.6MB per sync), but it's manageable.  Second, stackerdbs have a `write-freq` configuration field, which is used as a hint to determine how often to run the state machine to go and synchronize with replicas.  We can use this to throttle down synchronization rates for stackerdbs that aren't written to often, and most of the time, that's all but one of them.
 
Now, DKG only happens once per reward cycle, so 22 out of the 24 stackerdbs would only really be written to every 2 weeks.  We could automatically set the `write-freq` field for these stackerdbs to something very large when the system is not in the prepare phase, so we wouldn't actually be needlessly sending `StackerDBChunkInvData` messages for them.  The same goes for the upcoming signers' signing-round stackerdb, so in practice there'd only be one stackerdb whose state machine needs to be aggressively executed -- the current signers' signing-round stackerdb.

When we enter the prepare phase, the node need to throttle up the 12 stackerdbs used for DKG by the new signers, since it would need to synchronize a lot more metadata while DKG is running.  But once DKG completes, and we have the new aggregate public key, the node would throttle down the DKG-related stackerdbs, but leave the new signers' signing round stackerdb active.  When the next reward phase begins, the old singers' stackerdb would be throttled down.  So at any given point in time, there would only be 13 throttled-up stackerdbs, which amounts to 212KB of metadata exchanged per full synchronization per peer (or 6.82MB in total for 32 neighbors).  This would only be needed while DKG was running and being voted on.  For the remainder of the reward cycle, synchronization metadata per peer would be a paltry 16 KB (524 KB for 32 neighbors).

Each of the stackerdbs would need to be controlled by a separate boot contract, but they could all just extract the relevant configuration and signer data via a `contract-call?` to a singleton contract that just serves as a repository for the reward set information (so the node only needs to write it once).

From an operational standpoint, the advantages of this approach are two-fold:

* StackerDB is already live in `master`, so hotfixes we make to it while testing it live can be immediately applied to `next` as long as `next` maintains the same StackerDB design
* We're about to kick off audits of StackerDB this coming Monday, so if we can keep the current design, we won't have to rush-implement a new StackerDB system in the next 48 hours.

What do you think?

kantai:
That approach could work, but it would push complication onto the signer and these additional boot contracts. It’s a much less flexible approach — if the signers needed more slots, any additional slot would require adding a new boot contract. If that’s worth avoiding changes in the stackerdb implementation, that’s fine, but I can’t really speak to that question.

It would also require all of these changes _before_ any end to end testing can be done on the signer.

jcnelson:
I think this approach is actually a better design, for a few reasons.

* It's an application of the end to end principle. We shouldn't be making
stackerdb more complex to service a particular application need if we don't
need to.

* The added complexity of the signer is less demanding in terms of code,
testing, maintenance, and cognitive overhead than the added complexity to
stackerdb. The principal changes to the signer are to listen to a set of
stackerdb messages instead of one, and it's obvious from the contract
address what the stackerdb chunk's signer message type will be (since
there's a one-to-one correspondence between the packet type and the
contract address).  This is essentially the same as it is today, except the
signer uses designated slot IDs within the same stackerdb.  By contrast,
we'd have to do a substantial amount of work and testing to stackerdb to
make it support an essentially unbound number of slots.

* This approach is actually more flexible. If we need more slots, or more
generally, we need an additional communication channel for signers, we can
just deploy additional stackerdb-compliant contracts incrementally and use
them. There's no requirement that they must be boot code contracts, as long
as they can obtain the necessary reward set information from a single
designated boot contract (which is what I was proposing).

* This approach is a better fit for the kinds of network QoS we will need
to handle in the signing lifecycle.  The fact that each stackerdb has its
own configuration means that we can directly tackle network performance for
different aspects of the signing lifecycle by giving each step of the
protocol its own stackerdb.  This way, we can prioritize and deprioritize
message delivery and replication for different kinds of messages based on
where we are in the cycle (e.g. by varying the `write-freq` value in each
DB's config, or even generating that value on-the-fly instead of reading it
from the contract). By contrast, we'd have to come up with some additional
way to do this QoS work if we were using a single large stackerdb.

* I don't think it changes the fact that we have more work to do before we
can do end to end testing of the signer -- we either have to make big
stackerdbs, or we have to find a way to use many of them. I believe the
latter will require less work (and less time), while also being easier to
test (since we already know the existing stackerdb works in the wild).

jcnelson:
To assess the feasibility, I spent a few cycles tonight trying to patch the signer to deal with 24 stackerdbs instead of one. It's really not that bad at all -- @jferrant has done an excellent job at keeping all the stackerdb-specific code isolated from the rest of the system, so only a handful of functions need to be modified.

jcnelson:
@kantai I'll take it from here, thanks

jcnelson:
@kantai I can't request since you opened the PR, but can you give this a look-over when you have a moment?

libsigner/src/events.rs
```diff
@@ -345,8 +346,10 @@ fn process_stackerdb_event(
             .filter_map(|chunk| read_next::<NakamotoBlock, _>(&mut &chunk.data[..]).ok())
             .collect();
         SignerEvent::ProposedBlocks(blocks)
-    } else if event.contract_id.name.to_string() == SIGNERS_NAME {
-        // TODO: fix this to be against boot_code_id(SIGNERS_NAME, is_mainnet) when .signers is deployed
+    } else if event.contract_id.name.to_string().starts_with(SIGNERS_NAME)
+        && event.contract_id.issuer.1 == [0u8; 20]
+    {
```
jferrant, line 351: I think we might need some sort of additional item passed through with the SignerMessage to indicate whether this came from an even or odd rewrd cycle number so the signer knows which state machine to feed it to. I would assume the state machine would just ignore it as it isn't signed correctly, but we could have the case where a signer is signing for both reward cycle N and N+1 and therefore produces a valid signature and the receiving signer feed them into both state machines will bork up one of them.

stacks-signer/src/client/stackerdb.rs
```diff
@@ -79,23 +121,50 @@ impl StackerDB {
...
     ) -> Result<StackerDBChunkAckData, ClientError> {
         let message_bytes = message.serialize_to_vec();
-        let slot_id = message.slot_id(self.signer_id);
-
+        let msg_id = message.msg_id();
+        let signer_set = self.signer_set;
+        let slot_id = self.signer_slot_id;
         loop {
-            let slot_version = *self.slot_versions.entry(slot_id).or_insert(0) + 1;
+            let slot_version =
```
jferrant, line 128: Could you not do 

```rust
let slot_version = {
    let versions = self.slot_versions.entry((signer_set, msg_id))
        .or_insert_with(|| HashMap::new());
    *versions.entry(slot_id).or_insert(0).saturating_add(1)
};
```
And remove the following NOTE this will always execute per the above?

stacks-signer/src/runloop.rs
```diff
@@ -138,21 +138,96 @@ pub struct RunLoop<C> {
...
+                return Ok((*signer_set, *signer_slot_id));
+            }
+            (_, _) => {}
+        }
+
+        let signer_set = if let Some(signer_set) = self.signer_set.as_ref() {
+            *signer_set
+        } else {
+            let rc = u32::try_from(self.stacks_client.get_current_reward_cycle()?)
+                .expect("FATAL: reward cycle exceeds u32::MAX")
```
jferrant, line 167: I thought the reward cycle max was u64? Why down cast it here? I would down cast the modulus component u32::try_from(rc %2) as this will always fit a u32

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -160,6 +162,10 @@ pub fn make_contract_id(addr: &StacksAddress, name: &str) -> QualifiedContractId
     )
 }
 
+pub fn make_signers_db_name(signer_set: u32, message_id: u32) -> String {
```
jferrant, line 165: I think it might be worthwhile having a helper function to parse out the message id and signer set from the signers_db_name

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -55,16 +55,14 @@
...
-            (ok (at-block
-                (unwrap! (get-block-info? id-header-hash height) err-invalid-burn-block-height)
-                    (get-current-signer-slots signer-index)))))
-
 (define-read-only (get-current-signer-slots (signer-index uint))
-    (let ((details (unwrap! (unwrap-panic (contract-call? .signers stackerdb-get-signer-by-index signer-index)) err-invalid-signer-index)))
+    (let ((cycle (+ u1 (burn-height-to-reward-cycle burn-block-height))))
+      (get-signer-slots signer-index cycle)))
+
+(define-read-only (get-signer-slots (signer-index uint) (reward-cycle uint))
```
jferrant, line 62: should this be get-signer-weight instead so it doesn't get confused with returning the num-slots ?

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -55,16 +55,14 @@
 (define-read-only (get-tally (reward-cycle uint) (round uint) (aggregate-public-key (buff 33)))
     (map-get? tally {reward-cycle: reward-cycle, round: round, aggregate-public-key: aggregate-public-key}))
 
-(define-read-only (get-signer-slots (signer-index uint) (reward-cycle uint))
-    (let ((height (reward-cycle-to-burn-height reward-cycle)))
-            (ok (at-block
-                (unwrap! (get-block-info? id-header-hash height) err-invalid-burn-block-height)
-                    (get-current-signer-slots signer-index)))))
-
 (define-read-only (get-current-signer-slots (signer-index uint))
```
jferrant, line 58: Should this get renamed to get-current-signer-weights? FOr the same reasoning as my prior comment.

stackslib/src/chainstate/stacks/boot/signers.clar
```diff
@@ -1,33 +1,61 @@
...
+(define-constant MAX_WRITES u4294967295)
 (define-constant CHUNK_SIZE (* u2 u1024 u1024))
+(define-constant ERR_NO_SUCH_PAGE u1)
+(define-constant ERR_CYCLE_NOT_SET u2)
+(define-map cycle-signer-set uint (list 4000 { signer: principal, weight: uint }))
 
+;; Called internally by the Stacks node.
+;; Stores the stackerdb signer slots for a given reward cycle.
+;; Since there is one stackerdb per signer message, the `num-slots` field will always be u1.
 (define-private (stackerdb-set-signer-slots
```
jferrant, line 14: I am confused now....set-signer-slots does seem to set the num-slots...but get-signer-slots seems to return the weight...Def need some clarification on naming here I think D:

stackslib/src/chainstate/stacks/boot/signers.clar
```diff
@@ -1,33 +1,61 @@
...
+                 (signers (list 4000 { signer: principal, weight: uint })))
+     (begin
+      (asserts! (is-eq (var-get last-set-cycle) reward-cycle) (err ERR_CYCLE_NOT_SET))
+      (ok (map-set cycle-signer-set reward-cycle signers))))
 
-(define-read-only (stackerdb-get-signer-by-index (signer-index uint))
-	(ok (element-at (var-get stackerdb-signer-slots) signer-index))
-)
+;; Get the list of signers and weights for a given reward cycle.
+(define-read-only (get-signers (cycle uint))
```
jferrant, line 35: Oh perfect! This is exactly what I need.

stackslib/src/net/stackerdb/config.rs
```diff
@@ -123,40 +130,87 @@ impl StackerDBConfig {
...
+        };
+
+        let Ok(ClarityValue::UInt(num_slots)) = slot_data.get("num-slots") else {
+            let reason = format!(
+                "StackerDB fn `{contract_id}.{STACKERDB_SLOTS_FUNCTION}` returned tuple without `num-slots` entry of type `uint`",
+            );
+            return Err(reason);
+        };
+
+        let num_slots = u32::try_from(*num_slots)
```
jferrant, line 199: I saw lots of earlier ""BUG: Failed to construct `{ signer: principal, num-slots: u64 }` tuple"," Is the num-slots meant to be a u32 or a u64? We do not seem to be consistent.

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -265,8 +264,17 @@ impl BlockMinerThread {
         let stackerdb_contracts = stackerdbs
             .get_stackerdb_contract_ids()
             .expect("FATAL: could not get the stacker DB contract ids");
-        // TODO: get this directly instead of this jankiness when .signers is a boot contract
-        let signers_contract_id = boot_code_id(SIGNERS_NAME, self.config.is_mainnet());
+
+        let reward_cycle = u32::try_from(
```
jferrant, line 268: Similar to my other comment. I think reward cycle should remain a u64. We should just do u32::try_from on (reward_cycle % 2)

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -82,28 +81,19 @@ struct SignerTest {
...
 
         // Build the stackerdb signers contract
-        // TODO: Remove this once it is a boot contract
-        let signers_stackerdb_contract =
-            build_stackerdb_contract(&signer_stacks_addresses, SIGNER_SLOTS_PER_USER);
-        let signers_stacker_db_contract_id = QualifiedContractIdentifier::new(
-            to_addr(&publisher_private_key).into(),
-            "signers".into(),
-        );
+        let signers_stacker_db_contract_id = boot_code_id("signers".into(), false);
```
jferrant, line 96: I think you can replace"signers" with SIGNERS_BOOT_NAME or something like that.

jferrant:
This will help me so much on my end XD have been struggling with making things work the way they were and seeing all these changes makes me realize why lol :P

libsigner/src/events.rs
jcnelson: Yeah, I was thinking about that when I wrote it, but I didn't know what the best way to proceed was

stackslib/src/chainstate/stacks/boot/signers-voting.clar
jcnelson: I'm just using the names that were already there.  I think the voting contract needs to be significantly refactored, but that's a separate PR I think

stackslib/src/chainstate/stacks/boot/signers-voting.clar
jcnelson: Same as above -- I'm just using the names that already exist

stackslib/src/chainstate/stacks/boot/signers.clar
jcnelson: I think it should be `stackerdb-set-signer-slots` and `set-signers`.  The latter is unrelated to the stackerdb subsystem.

stackslib/src/net/stackerdb/config.rs
jcnelson: It has to be a `u32` because the stackerdb system uses a `u32` to represent the total number of slots

stacks-signer/src/runloop.rs
jcnelson: Yeah, I'll do that

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -2511,7 +2511,7 @@ impl NakamotoChainState {
                 Self::calculate_matured_miner_rewards(
                     &mut clarity_tx,
                     sortition_dbconn.sqlite_conn(),
-                    coinbase_height,
+                    coinbase_height + 1,
```
jcnelson, line 2514: @kantai Can you explain why this change needed to be made?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -2511,6 +2511,7 @@ impl NakamotoChainState {
                 Self::calculate_matured_miner_rewards(
                     &mut clarity_tx,
                     sortition_dbconn.sqlite_conn(),
+                    // coinbase_height + 1,
```
jcnelson, line 2514: @kantai I had to comment this out because it was causing the chain to crash in some of the integration tests.  Can you please help me understand why this change was here in the first place?

jcnelson (approved):

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -265,8 +264,19 @@ impl BlockMinerThread {
...
+            .block_height_to_reward_cycle(self.burn_block.block_height)
+            .expect("FATAL: no reward cycle for burn block");
+
+        let signers_contract_id = boot_code_id(
+            &make_signers_db_name(
+                u32::try_from(reward_cycle % 2).expect("FATAL: infallible"),
+                BLOCK_MSG_ID,
+            ),
+            self.config.is_mainnet(),
+        );
```
kantai, line 279: This should be a method of `signer_set` like `get_signers_contract(is_mainnet, reward_cycle, msg_id)` -- that way, the miner thread doesn't have to encode logic about the reward cycle modulus.

stackslib/src/chainstate/nakamoto/mod.rs
kantai: There were some tests where the coordinator panicked without this change (I think they were signers tests that used mockamoto, or some configuration where nakamoto was active during very early Stacks block heights). This was because of some assumptions about the block height, maturation windows, and the size of the matured rewards. I don't think `+1` here is correct, but I do think that the height used for calculated miner rewards maturation in nakamoto needs another look.

The `coinbase_height` parameter of `NakamotoChainState::setup_block` includes the block being set up. The `chain_tip` parameter of `StacksChainState::setup_block`, however, is the parent tip, so that the stacks height used in this function is logically `-1`. This is why `StacksChainState` uses `+1` when invoking `handle_prepare_phase_start()`. This was discovered in trying to test that method, but I didn't try to completely test all of the behavior around these height parameters. It may be that the first block of a tenure has the parent's height passed in, but subsequent blocks in that tenure have the new coinbase height (where we could see something like https://github.com/stacks-network/stacks-core/issues/4333).

kantai (commented):
This LGTM, just some superficial comments. But also, a response to the coinbase_height argument for reward maturation probably deserves more attention.

friedger (approved):
LGTM

libsigner/src/events.rs
jferrant: I think I have a fix in my PR for this so no need to add :D

jferrant (approved):
LGTM! Has greatly improved my cast aggregate dkg vote PR so am excited to see this merged.

stackslib/src/net/stackerdb/config.rs
```diff
@@ -221,114 +219,64 @@ impl StackerDBConfig {
...
+                    "Contract {} stipulated more than maximum number of slots for one signer ({})",
+                    contract_id, STACKERDB_INV_MAX
+                );
+                warn!("{}", &reason);
+                return Err(NetError::InvalidStackerDBContract(
+                    contract_id.clone(),
+                    reason,
+                ));
+            }
+            let num_slots = num_slots_uint as u32;
```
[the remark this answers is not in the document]
jcnelson: Merge artifact; will fix

stackslib/src/chainstate/stacks/boot/mod.rs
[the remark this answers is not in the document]
jcnelson: Sure

stackslib/src/net/stackerdb/config.rs
```diff
@@ -221,114 +219,64 @@ impl StackerDBConfig {
             burn_dbconn,
             tip,
             contract_id,
-            "({STACKERDB_SLOTS_FUNCTION})",
+            "(stackerdb-get-signer-slots)",
```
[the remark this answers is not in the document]
jcnelson: For some reason, when I tried to run this code, the VM evaluated the literal string `({STACKERDB_SLOTS_FUNCTION})` and rightfully complained about the lack of a `:` in the tuple.  Not sure why, because it was imported, so I just replaced it with the function name.

I'll try again.

testnet/stacks-node/src/nakamoto_node/miner.rs
jcnelson: Done

PR #4231, opened 2024-01-04
PoX-4 Clarunit Setup & Initial Tests

setzeus:
### Description
This PR introduces the new `@stacks/clarunit` to allow for writing boot contract unit tests in Clarity. Clarunit is uses Clarinet SDK so all the goodness that comes with Clarinet is preserved.

Only some basic tests are introduced in this PR. Increasing code coverage will be done in follow-up PRs.

### Applicable issues
Issue wasn't created but this is a continuation of the work done in #4100.

PR #4324, opened 2024-02-01
Updated Vote Print Statement

setzeus:
As discussed during Un/block meeting, the simplest/fastest solution for #4231 is simply to update the print statement found in 'vote-for-aggregate-public-key.'

The node will then take this information & then check whether the consensus threshold was met / a new aggregate public key has been submitted. 

Since this is a simple print statement update there are no additional test to write.


jcnelson (approved):

obycode (approved):
LGTM!

kantai (approved):

PR #4327, opened 2024-02-01
Accept out-of-order votes

### LINKED ISSUE #4322: [clarity] ` .signer-voting` need to accept out of round votes
**Describe the bug**
Currently, vote txs fail if not on the last round. 

` .signer-voting` will need to accept out of round votes

friedger:
### Description
Currently, `.signer-voting` rejects votes that are not for the latest round.

This PR removes this restriction. Votes out of round are accepted.

### Applicable issues

- fixes #4322 

### Checklist

- [ ] Changelog is updated


stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
```diff
@@ -258,7 +259,8 @@ fn vote_for_aggregate_public_key_in_last_block() {
     );
 
     let cycle_id: u128 = current_reward_cycle;
-    let aggregated_public_key: Point = Point::new();
+    let aggregate_public_key: Point = Point::new();
+    let aggregate_public_key_1: Point = Point::G();
```
friedger, line 263: Is there a better way to create aggregate public keys for testing ? @MarvinJanssen

setzeus (approved):
Logic change that removes asserting that rounds-check / out-of-order-votes LGTM. 

Only have one NIT re: error messages. 

Note that we have an consistency in formatting. In [pox-4](https://github.com/stacks-network/stacks-core/blob/next/stackslib/src/chainstate/stacks/boot/pox-4.clar) the error constants are stored as an Int, in this contract, the error constants are stored as (err uint). 

No preference for either format but preferable we stick to a single syntax.

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
MarvinJanssen: You could do:

```rust
let private_key = StacksPrivateKey::from_seed(&[3]);
let public_key = StacksPublicKey::from_private(&private_key);
```

And change the seed input to different values.

If you want to generate actual aggregate public keys then take a look at `TestSigners`.

```rust
let test_signers = TestSigners::default();
let public_key = test_signers.aggregate_public_key.clone();
```

MarvinJanssen (commented):
Looks fine to me. The spec requires out-of-order votes?

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
friedger: Thank you. Code has been updated

friedger:
@MarvinJanssen Yes, as discussed with @kantai in https://nakamotoslack.slack.com/archives/C067BHSN1SL/p1706799771752509?thread_ts=1706690349.397339&cid=C067BHSN1SL 

Are you read to approve this PR?

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -1959,10 +1960,10 @@ pub mod test {
         key: &StacksPrivateKey,
         nonce: u64,
         signer_index: u128,
-        aggregate_public_key: &Point,
+        aggregate_public_key: &Secp256k1PublicKey,
```
jcnelson, line 1963: This is wrong.  The aggregate public key is a `Point`

jcnelson (changes requested):
Putting a hold on this for the time being; it's not ready to merge.  Will update with more details later.

stackslib/src/chainstate/stacks/boot/mod.rs
friedger: I reverted this change

MarvinJanssen: Sorry my mistake.

`test_signers.aggregate_public_key` is a `Point` too.

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
```diff
@@ -337,10 +346,10 @@ fn vote_for_aggregate_public_key_in_last_block() {
     // alice's block
     let block = &blocks[blocks.len() - 2].clone();
     let receipts = &block.receipts;
-    assert_eq!(receipts.len(), 4);
+    assert_eq!(receipts.len(), 5);
 
     // first vote should succeed
-    let tx1 = &receipts[receipts.len() - 2];
+    let tx1 = &receipts[receipts.len() - 3];
```
hstove, line 352: IMO it's better to use explicit nonces to avoid these kind of off-by-one errors

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
```diff
@@ -350,7 +359,7 @@ fn vote_for_aggregate_public_key_in_last_block() {
     );
 
     // second vote should fail with duplicate vote error
-    let tx2 = &receipts[receipts.len() - 1];
+    let tx2 = &receipts[receipts.len() - 2];
```
hstove, line 362: Also better to use explicit names for the tx instead of `tx2`. I'd have to read too much into the test just to find out what tx this is testing

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
```diff
@@ -359,15 +368,25 @@ fn vote_for_aggregate_public_key_in_last_block() {
...
     );
 
+    // third vote should succeed even though it is on an old round
+    let tx3 = &receipts[receipts.len() - 1];
+    assert_eq!(
+        tx3.result,
+        Value::Response(ResponseData {
+            committed: true,
+            data: Box::new(Value::Bool(true))
+        })
```
hstove, line 378:
```suggestion
        Value::okay_true()
```

hstove: Also it's not this PR's code but `Value::error(Value::UInt(10002)).unwrap()` for errors

hstove (changes requested):
Other than a few nits, I think this test should evaluate the contract state after the TX's are confirmed (via read-only calls)

Issue #4328, opened 2024-02-01
[ StackerDB ] The StackerDB config must be refreshed on `.signers` updates


kantai:
Currently, `StackerDB`'s configs are refreshed once per reward cycle, however, they really need to be updated when `.signers` receives updates from the `handle_prepare_phase()` method.

PR #4330, opened 2024-02-01
Feat: trigger stackerdb refresh when .signers is written

kantai:
### Description

This implements #4328 by adding a signaling bool to the coordinator channels structs. It's a bit cludgier than I would otherwise want, but I think it's the best we can do without majorly refactoring the coordinator's signatures (which is probably a good idea, but one best left to the future).

This is a draft for now, because it depends on `feat/larger-stackerdb`, but it could be cherry-picked/rebased to `next` instead.

### DIFF
--- a/testnet/stacks-node/src/nakamoto_node/peer.rs
+++ b/testnet/stacks-node/src/nakamoto_node/peer.rs
@@ -202,6 +202,21 @@ impl PeerThread {
         }
     }
 
+    fn check_stackerdb_reload(&mut self) {
+        if !self.globals.coord_comms.need_stackerdb_update() {
+            return;
+        }
+
+        if let Err(e) = self
+            .net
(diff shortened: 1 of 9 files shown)


testnet/stacks-node/src/nakamoto_node/peer.rs
```diff
@@ -202,6 +202,21 @@ impl PeerThread {
         }
     }
 
+    fn check_stackerdb_reload(&mut self) {
```
jcnelson, line 205: Minor nit: This is doing a check and then maybe doing some work based on that check.  Maybe this should be called `stackerdb_refresh()` or something?

jcnelson (approved):
LGTM; just one minor nit

jferrant (approved):

jferrant:
Thank you for doing this!

jferrant:
Having a really hard time rebasing on this XD (couldn't cherry pick it out)

PR #4332, opened 2024-02-01
fix: incorrect arg index in synthetic event

hstove:
In #4329 (just merged) one of the synthetic events used an incorrect arg index in the print for `stack-stx`.

I also noticed that `stack-aggregation-increase` and `stack-aggregation-commit` have diverged argument signatures - `-increase`'s third arg is `reward-cycle-index`, whereas `-commit`'s third arg is `signer-key`. I've made a separate `match` branch to reflect this change.


kantai (approved):

zone117x (approved):
Thanks @hstove!

PR #4331, opened 2024-02-01
fix: buids on various archs/platforms

zone117x:
fix building & cross compilation, including `stacks-signer`, tested building for the following:

* linux-glibc-x86-64
* linux-glibc-arm64
* linux-glibc-armv7
* linux-musl-x86_64
* linux-musl-arm64
* linux-musl-armv7
* windows-x86_64
* macos-x86_64
* macos-arm64

ty @wileyj and @hugocaillard

PR #4335, opened 2024-02-02
fix: builds on various archs/platforms re: PR 4331

wileyj:
ref:  https://github.com/stacks-network/stacks-core/pull/4331

these are the equivalent changes for `master` and `develop`. 

fix building & cross compilation, including stacks-signer, tested building for the following:

- linux-glibc-x86-64
- linux-glibc-arm64
- linux-glibc-armv7
- linux-musl-x86_64
- linux-musl-arm64
- linux-musl-armv7
- windows-x86_64
- macos-x86_64
- macos-arm64

*note*: target branch here is for develop since there is a workaround in place already in `master`. this change won't affect that workaround the next time develop is merged to master

cc: @zone117x @hugocaillard

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,12 @@ members = [
     "stacks-signer",
     "testnet/stacks-node"]
 
+# Dependencies we want to keep the same between workspace members
+[workspace.dependencies]  
+wsts = { version = "7.0", default-features = false }
+rand_core = "0.6"
+rand = "0.8"
+
 # Use a bit more than default optimization for
 #  dev builds to speed up test execution
 [profile.dev]
--- a/stacks-signer/Cargo.toml
+++ b/stacks-signer/Cargo.toml
@@ -27,9 +27,9 @@ clap = { version = "4.1.1", features = ["derive", "env"] }
 hashbrown = "0.14"
 libsigner = { path = "../libsigner" }
 libstackerdb = { path = "../libstackerdb" }
-p256k1 = "5.5"
+p256k1 = { version = "5.5", default-features = false }
 rand_core = "0.6"
-reqwest = { version = "0.11.22", features = ["blocking", "json"] }
+reqwest = { version = "0.11.22", default-features = false, features = ["blocking", "json", "rustls-tls"] }
 serde = "1"
 serde_derive = "1"
 serde_stacker = "0.1"
@@ -42,7 +42,7 @@ thiserror = "1.0"
 toml = "0.5.6"
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
-wsts = "4.0.0"
+wsts = { version = "4.0.0", default-features = false }
 
 [dependencies.serde_json]
 version = "1.0"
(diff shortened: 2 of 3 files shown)


zone117x (approved):

stacks-signer/Cargo.toml
```diff
@@ -27,9 +27,9 @@ clap = { version = "4.1.1", features = ["derive", "env"] }
 hashbrown = "0.14"
 libsigner = { path = "../libsigner" }
 libstackerdb = { path = "../libstackerdb" }
-p256k1 = "5.5"
+p256k1 = { version = "5.5", default-features = false }
 rand_core = "0.6"
-reqwest = { version = "0.11.22", features = ["blocking", "json"] }
+reqwest = { version = "0.11.22", default-features = false, features = ["blocking", "json", "rustls-tls"] }
```
jcnelson, line 32: Is there a way to simply build without TLS?  If not, then this is fine.

jcnelson (approved):

wileyj:
following discussion in #4331 , converting this to draft. 
there may be another option vs modifying the cargo.toml's here - eg using a native arm runner to build the binaries.

wileyj:
@xoloki can you confirm this is OK to do, or should be continue looking for build alternatives where we're not cross-compiling?

hstove (approved):

Cargo.toml
```diff
@@ -11,6 +11,12 @@ members = [
     "stacks-signer",
     "testnet/stacks-node"]
 
+# Dependencies we want to keep the same between workspace members
+[workspace.dependencies]  
+wsts = { version = "7.0", default-features = false }
```
xoloki, line 16: This is the correct way to include `wsts` to avoid including `bindgen` and friends.

stacks-signer/Cargo.toml
```diff
@@ -42,7 +42,7 @@ thiserror = "1.0"
 toml = "0.5.6"
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
-wsts = "4.0.0"
+wsts = { version = "4.0.0", default-features = false }
```
xoloki, line 45: Is there a reason why we have such an old version of `wsts` here in the `stacks-signer` `Cargo.toml`?  And why we have different versions here and in the root `Cargo.toml`?

xoloki (approved):
The general approach of using `default-features = false` is correct.

I'm a little concerned that we are using different versions of `wsts` in the root and `stacks-signer` `Cargo.toml` files, but that was present before this PR so not a blocker.

stacks-signer/Cargo.toml
wileyj: Can't say for certain, but best to leave as-is since this branch has been genesis-sync tested as working.

wileyj:
> The general approach of using `default-features = false` is correct.
> 
> I'm a little concerned that we are using different versions of `wsts` in the root and `stacks-signer` `Cargo.toml` files, but that was present before this PR so not a blocker.

any concerns with the similar PR https://github.com/stacks-network/stacks-core/pull/4331 ?

xoloki:
> > The general approach of using `default-features = false` is correct.
> > I'm a little concerned that we are using different versions of `wsts` in the root and `stacks-signer` `Cargo.toml` files, but that was present before this PR so not a blocker.
> 
> any concerns with the similar PR #4331 ?

#4331 is for `next` which uses a workspace dependency for `wsts` so no concerns there.

Issue #4316, opened 2024-01-31
Find Optimizations in the current Clarity VM via benchmarking current VM and Wasm


saralab:

jbencin:
## Summary of differences from `costs-3.clar` to `costs-4.clar`

### Legend

| Symbol | Meaning                 |
|--------|-------------------------|
| +      | Minor cost increase     |
| ++     | Cost increase           |
| +++    | Major cost increase     |
| -      | Minor cost decrease     |
| --     | Cost decrease           |
| ---    | Major cost decrease     |

### Costs

Only the functions that have different cost values are reported here:

| Function                              | `costs-3`                           | `costs-4`                             | Interpretation    |
|---------------------------------------|-------------------------------------|---------------------------------------|-------------------|
| `cost_analysis_type_annotate`         | `(runtime (linear n u1 u9))`        | `(runtime (linear n u1 u10))`         | +                 |
| `cost_analysis_type_check`            | `(runtime (linear n u113 u1))`      | `(runtime (linear n u115 u0))`        | ?                 |
| `cost_analysis_list_items_check`      | `(runtime (linear n u2 u4))`        | `(runtime (linear n u2 u8)))`         | +                 |
| `cost_analysis_check_tuple_cons`      | `(runtime (nlogn n u3 u5))`         | `(runtime (nlogn n u3 u142))`         | +++               |
| `cost_analysis_tuple_items_check`     | `(runtime (linear n u1 u28))`       | `(runtime (linear n u2 u52))`         | ++                |
| `cost_analysis_lookup_variable_const` | `(runtime u15)`                     | `(runtime u16)`                       | +                 |
| `cost_ast_cycle_detection`            | `(runtime (linear n u141 u72))`     | `(runtime (linear n u156 u23))`       | ?                 |
| `cost_analysis_fetch_contract_entry`  | `read_length: (linear n u1 u1)`     | `read_count: (linear n u1 u1)`        | Bug? `read_count` duplicated |
| `cost_lookup_variable_size`           | `(runtime (linear n u2 u1))`        | `(runtime (linear n u2 u18))`         | ++                | 
| `cost_user_function_application`      | `(runtime (linear n u26 u5))`       | `(runtime (linear n u27 u5))`         | +                 |
| `cost_tuple_merge`                    | `(runtime (linear n u4 u408))`      | `(runtime (linear n u5 u408))`        | +                 | 
[...]

PR #4336, opened 2024-02-02
chore: Fix lots of `clippy::perf` warnings

jbencin:
### Description

As part of the Clarity interpreter optimization, I've been using Clippy, with a focus on the `clippy::perf` lints, to find sub-optimal code that could be re-written to improve performance. Some of the changes in this PR include:

- Avoiding unnecessary string allocs in `.expect()` by using `.unwrap_or_else()`
- Replace `.drain(..)` with `.into_iter()`
- Replace unnecessary `Vec` use with slices
- Avoid allocating empty `Vec`s, then resizing or calling `push()` in a loop, triggering reallocs. Instead, use `Vec::with_capacity()` to pre-allocate all the space we will need.
- Remove unnecessary `.to_string()` calls, where a `&str` or `Display` type is acceptable
- Remove unnecessary `.clone()` calls

Some lints that haven't been addressed yet, because they will take more effort, but still want to get to:
- `clippy::result_large_err`
- `clippy::large_enum_variant`
- `clippy::or_fun_call`

### Applicable issues

- #4316

### Additional Info

Although I focused on Clarity, and the shared code in stackslib and stacks-common, some of the lints that could be safely auto-fixed were applied to the entire project

### Notes to reviewers

This PR is feature complete, I will put further optimizations in other PRs, although I may have to merge `next` again

A lot of these changes were made by `cargo clippy --fix`. In particular, there are a lot of `.expect(&format!(...))` -> `.unwrap_or_else(|| panic!(...))` changes (which enables lazy evaluation). Same with the `.clone()` removal. The code compiles, so it's pretty safe to assume these changes are okay. Try to focus on the manual changes I made with `Vec` allocation


jcnelson:
Thanks @jbencin!  How does this impact performance, versus without this patch?

jbencin:
@jcnelson Haven't tested yet, the benchmarks take 4 hours to run, and I have plans to do some more optimizations, so I'm waiting until I have more to test. This is fixing a lot of unnecessary allocs/reallocs and other issues, each of which has a small impact, and most of which aren't in a critical path, but the effects should add up to something noticeable

ASuciuX (approved):
Looked over the changes, ran the tests, LGTM.

clarity/src/vm/database/clarity_db.rs
```diff
@@ -1162,17 +1162,10 @@ impl<'a> ClarityDatabase<'a> {
         let value = Value::Tuple(
             TupleData::from_data(vec![
                 (
-                    ClarityName::try_from("reporter").map_err(|_| {
-                        InterpreterError::Expect("BUG: valid string representation".into())
-                    })?,
+                    ClarityName::from("reporter"),
```
obycode, line 1165: This is changing an error to a panic. That's probably okay in this case, since the input is a constant that we know will not error, but we should be careful with this kind of change. @kantai are you comfortable with leaving a panic in cases like this?

obycode: There are several cases of this in this PR. I won't comment on the others, but if we decide not to make this change, all of the others should be reverted as well.

obycode (commented):
I'm still going through these, but submitting a couple comments before breaking for dinner.

Could the clippy settings you used be included in CI?

I'm also very curious how these changes affect the benchmarks.

clarity/src/vm/database/clarity_db.rs
jbencin: Didn't realize there was a panic inside `ClarityName::from()`. If we're not even comfortable using this function on constant `&str`s, then what's the point of having it in the codebase? For test code only? In that case, we should make it explicit and add `#[cfg(test)]`

jbencin:
> Could the clippy settings you used be included in CI?

Absolutely, there's a lot of good stuff in Clippy, and it can catch lots of things `rustc` and testing can't. We should start introducing these rules gradually, and only include them in CI if:
- We've already cleaned up all the violations of the rule
- We are confident that this rule will pretty much never throw false positives

jbencin:
Also question for @kantai and @obycode: Should I close this PR, make these changes on `develop`, and merge that up to `next`? Are we planning more releases from `develop` and do we care about squeezing a little more performance from it (my guess is it doesn't matter much with 10 minute block times)?

clarity/src/vm/database/clarity_db.rs
obycode: I'm not opposed to that idea.

obycode:
> Also question for @kantai and @obycode: Should I close this PR, make these changes on `develop`, and merge that up to `next`? Are we planning more releases from `develop` and do we care about squeezing a little more performance from it (my guess is it doesn't matter much with 10 minute block times)?

I'm fine with keeping this on `next`.

obycode (approved):
This is great! Thanks @jbencin.

wileyj:
> > Could the clippy settings you used be included in CI?
> 
> Absolutely, there's a lot of good stuff in Clippy, and it can catch lots of things `rustc` and testing can't. We should start introducing these rules gradually, and only include them in CI if:
> 
> * We've already cleaned up all the violations of the rule
> * We are confident that this rule will pretty much never throw false positives

we did give clippy a test run a while back on a fork - it was quite verbose and would have required a lot of exceptions for it to be added to ci (while not making most of the recommended changes). 
if we can find a good baseline though, i'd be open to adding it back in.

jcnelson (changes requested):
Hi @jbencin, thanks for looking at this!

Before approving, can you do the following?

* Can you post a flame graph before and after these changes, so we can verify that this actually improves performance?

* Can you verify that the system processes each block and microblock stream correctly with these changes?  We need to do this with a genesis sync anyway when we boot `next`, but I'd rather catch this sooner than later.  We've been bitten in the recent past by a PR that was supposed to be _only_ Clippy changes, but instead introduced a consensus bug in `develop`'s Clarity VM that had to be ironed out in the current RC.

jbencin:
> Can you post a flame graph before and after these changes, so we can verify that this actually improves performance?

I'm not sure how a flame graph is going to show this. I can run the costs benchmarks again though. I can't confidently say this will produce a measurable improvement, because:

- Many of these code changes are not reachable from Clarity VM
- I don't know how well the Rust compiler was optimizing the previous code

I am confident that this won't _decrease_ performance though. I've intentionally excluded any Clippy lints with performance trade-offs or that potentially changed behavior from this PR.

> Can you verify that the system processes each block and microblock stream correctly with these changes?

@jcnelson how would I do this?

jbencin:
> Here, I think if you run a genesis sync with your chagnes, it would expose any issues

@wileyj was this supposed to be a link?

wileyj:
> > Here, I think if you run a genesis sync with your chagnes, it would expose any issues
> 
> @wileyj was this supposed to be a link?

no, sorry - i mean "here" as in a genesis sync would answer jude's question

saralab:
Running full Genesis Sync to validate this before merging.

jbencin:
I used `hyperfine` to benchmark `next` vs. this PR in processing the first 3000 blocks. By default, `hyperfine` benchmarks 10 runs of each command:

```console
❯ hyperfine -w 3 "stacks-core.next/target/release/stacks-inspect replay-block /home/jbencin/data/next/ --first 3000" "stacks-core/target/release/stacks-inspect replay-block /home/jbencin/data/next/ --first 3000"
Benchmark 1: stacks-core.next/target/release/stacks-inspect replay-block /home/jbencin/data/next/ --first 3000
  Time (mean ± σ):     31.000 s ±  0.059 s    [User: 27.014 s, System: 3.821 s]
  Range (min … max):   30.954 s … 31.140 s    10 runs
 
Benchmark 2: stacks-core/target/release/stacks-inspect replay-block /home/jbencin/data/next/ --first 3000
  Time (mean ± σ):     30.943 s ±  0.073 s    [User: 26.998 s, System: 3.787 s]
  Range (min … max):   30.852 s … 31.067 s    10 runs
 
Summary
  stacks-core/target/release/stacks-inspect replay-block /home/jbencin/data/next/ --first 3000 ran
    1.00 ± 0.00 times faster than stacks-core.next/target/release/stacks-inspect replay-block /home/jbencin/data/next/ --first 3000
```

The performance delta here is not significant, but this PR is very slightly faster. My guess is that there are a few reasons for this:
- Many of the changes in this PR are not in a path used by `stacks-inspect replay-block`
- I was never expecting major changes here, it's just getting rid of some unnecessary allocs
- I'm building `stacks-inspect` with the `--release` flag, which might already optimize away the inefficient code

jbencin:
Another run with 30 samples:

```console
❯ hyperfine -w 3 -r 30 "stacks-core.next/target/release/stacks-inspect replay-block /home/jbencin/data/next/ --first 3000" "stacks-core/target/release/stacks-inspect replay-block /home/jbencin/data/next/ --first 3000"
Benchmark 1: stacks-core.next/target/release/stacks-inspect replay-block /home/jbencin/data/next/ --first 3000
  Time (mean ± σ):     31.030 s ±  0.060 s    [User: 27.082 s, System: 3.788 s]
  Range (min … max):   30.911 s … 31.154 s    30 runs
 
Benchmark 2: stacks-core/target/release/stacks-inspect replay-block /home/jbencin/data/next/ --first 3000
  Time (mean ± σ):     30.934 s ±  0.077 s    [User: 27.018 s, System: 3.755 s]
  Range (min … max):   30.832 s … 31.179 s    30 runs
 
Summary
  stacks-core/target/release/stacks-inspect replay-block /home/jbencin/data/next/ --first 3000 ran
    1.00 ± 0.00 times faster than stacks-core.next/target/release/stacks-inspect replay-block /home/jbencin/data/next/ --first 3000
```

Again, I'm consistently seeing a *very* slight improvement. I will try validating the entire block history with `replay-block` now...

jbencin:
I used `stacks-inspect replay-block` to validate all blocks until [100492](https://explorer.hiro.so/block/0xe9f326faba66929a27a56a04fb0dac020f1078604541213a69b81dc52899731a?chain=mainnet), which fails in both this PR and on `next` (meaning the bug is not introduced by this PR)

jbencin:
I tried the benchmark again on a different range of blocks, which I expected to be more CPU-bound. I'm still seeing a consistent, but sub-1% performance improvement:

```console
❯ hyperfine -w 3 -r 10 "stacks-core/target/release/stacks-inspect.next replay-block /home/jbencin/data/next/ range 99990 100000" "stacks-core/target/release/stacks-inspect.clippy-fixes replay-block /home/jbencin/data/next/ range 99990 100000"
  Time (mean ± σ):     12.169 s ±  0.058 s    [User: 11.647 s, System: 0.472 s]
  Range (min … max):   12.113 s … 12.306 s    10 runs
 
Benchmark 2: stacks-core/target/release/stacks-inspect.clippy-fixes replay-block /home/jbencin/data/next/ range 99990 100000
  Time (mean ± σ):     12.124 s ±  0.040 s    [User: 11.599 s, System: 0.477 s]
  Range (min … max):   12.071 s … 12.220 s    10 runs
 
Summary
  stacks-core/target/release/stacks-inspect.clippy-fixes replay-block /home/jbencin/data/next/ range 99990 100000 ran
    1.00 ± 0.01 times faster than stacks-core/target/release/stacks-inspect.next replay-block /home/jbencin/data/next/ range 99990 100000
```

jbencin:
@jcnelson I've finished validating the entire chainstate as of 2024-02-06 (149739 blocks) with this branch and there were no errors

jcnelson:
Thanks @jbencin!

I'm going to ask now that you do something novel:  close this PR, re-run the clippy commands off of the current `next`, and re-open it.  **Do not attempt to merge `next` into this branch**.  I'm making this very specific request because we recently introduced a consensus bug into the RC this way -- the merger combined a PR that was only composed of automated clippy changes (like this one) with its target branch incorrectly.

Please link that PR here and I'll approve it.

jbencin:
@jcnelson that's not really feasible, most of these changes were made by Clippy automatically, but some were not, and would take significant time to make them again. And by the time I get that done, `next` will have changed again

What I've been doing with merge conflicts in this branch is rather than try to resolve the conflicts, just accept the incoming changes and run Clippy again. This is safer than trying to manually resolve a conflict.

Anyways, we do conflict resolution all the time when we merge and rebase, and that is always one of the biggest risks, whether it's a Clippy PR or not. The changes here are incredibly simple and far less likely than most merge conflicts to be resolved incorrectly

clarity/src/vm/database/clarity_db.rs
kantai: Yeah -- these change lines should be reverted, because they're actually changing semantics here.

I agree we should make `ClarityName::from` test code only, but let's do that in a separate PR.

clarity/src/vm/functions/assets.rs
```diff
@@ -239,21 +239,12 @@ pub fn special_stx_account(
...
-            "locked"
-                .try_into()
-                .map_err(|_| InterpreterError::Expect("Bad special tuple name".into()))?,
-            Value::UInt(stx_balance.amount_locked()),
-        ),
-        (
-            "unlock-height"
-                .try_into()
-                .map_err(|_| InterpreterError::Expect("Bad special tuple name".into()))?,
+            "unlock-height".into(),
```
kantai, line 247: Same here as above -- this should be reverted.

stacks-common/src/util/hash.rs
```diff
@@ -419,8 +419,8 @@ where
         loop {
             // next row
             let i = nodes.len() - 1;
-            let mut row_hashes = vec![];
-            row_hashes.reserve(nodes[i].len() / 2);
+            let capacity = (nodes[i].len() + 1) / 2;
```
kantai, line 422: Should be a saturating add

clarity/src/vm/database/clarity_db.rs
```diff
@@ -1183,12 +1183,7 @@ impl<'a> ClarityDatabase<'a> {
                     })?,
                     Value::Principal(PrincipalData::Standard(reporter.clone())),
                 ),
-                (
-                    ClarityName::try_from("sequence").map_err(|_| {
-                        InterpreterError::Expect("BUG: valid string representation".into())
-                    })?,
-                    Value::UInt(seq as u128),
-                ),
+                (ClarityName::from("sequence"), Value::UInt(seq as u128)),
```
kantai, line 1186: This should be reverted as well.

clarity/src/vm/database/clarity_db.rs
jbencin: Made an issue for this, #4398

kantai (approved):

stackslib/src/chainstate/burn/db/processing.rs
```diff
@@ -135,17 +135,13 @@ impl<'a> SortitionHandleTx<'a> {
         let total_burn = state_transition
             .accepted_ops
             .iter()
-            .fold(Some(0u64), |acc, op| {
```
jcnelson, line 138: Is this from clippy, or is this something you added?  Because I don't know if these are semantically equivalent.

stackslib/src/chainstate/burn/distribution.rs
```diff
@@ -390,12 +390,9 @@ impl BurnSamplePoint {
     pub fn get_total_burns(burn_dist: &[BurnSamplePoint]) -> Option<u64> {
         burn_dist
             .iter()
-            .fold(Some(0), |burns_so_far, sample_point| {
-                if let Some(burns_so_far) = burns_so_far {
-                    burns_so_far.checked_add(sample_point.burns.try_into().ok()?)
-                } else {
-                    None
-                }
+            .try_fold(0u64, |burns_so_far, sample_point| {
```
jcnelson, line 393: Again, not sure if these are semantically equivalent.

stackslib/src/net/prune.rs
```diff
@@ -71,10 +71,12 @@ impl PeerNetwork {
                         Some(peer) => {
                             let stats = convo.stats.clone();
                             let org = peer.org;
-                            if org_neighbor.contains_key(&org) {
-                                org_neighbor.get_mut(&org).unwrap().push((nk, stats));
+                            if let std::collections::hash_map::Entry::Vacant(e) =
```
jcnelson, line 74: I really don't like the `hash_map::Entry` API.  Can you instead factor this as follows:

```rust
if let Some(stats_list) = org_neighbor.get_mut(&org) {
   stats_list.push((nk, stats));
} else {
   org_neighbor.insert(org, vec![(nk, stats)]);
}
```

stackslib/src/net/prune.rs
```diff
@@ -326,8 +328,10 @@ impl PeerNetwork {
                 Some(ref convo) => {
                     if !convo.stats.outbound {
                         let stats = convo.stats.clone();
-                        if !ip_neighbor.contains_key(&nk.addrbytes) {
-                            ip_neighbor.insert(nk.addrbytes, vec![(*event_id, nk.clone(), stats)]);
+                        if let std::collections::hash_map::Entry::Vacant(e) =
```
jcnelson, line 331: Same here

jcnelson (commented):
I have a few questions that need to be addressed before merging.

stackslib/src/chainstate/burn/db/processing.rs
jbencin: Yes Clippy warned about this, but I had to make the change manually. These are not quite equivalent, `try_fold()` will quit when it finds `None`, whereas the `fold()` version will always run through the entire iterator. So `try_fold()` can be more efficient, but they always produce the same value

stackslib/src/net/prune.rs
jbencin: Sure

stackslib/src/chainstate/burn/db/processing.rs
jcnelson: Okay, since they aren't semantically equivalent, please revert this `try_fold` change and the one below.

jcnelson (approved):

PR #4338, opened 2024-02-02
Clarity-Wasm: fixes for the `principal` type

Acaccia:
This PR fixes two issues for Clarity-Wasm:

1. The representation in memory is only two Wasm values, but we tried to read 22 (+`name`),
2. Creating lists of qualified principal would make both the compilation and the evaluation crash, because those would be typed as `ListUnionType`, which does not have a defined representation in our ABI.

Those fixes are extensively tested with property testing in the Clarity-Wasm repo.

### DIFF
--- a/clarity/src/vm/analysis/type_checker/contexts.rs
+++ b/clarity/src/vm/analysis/type_checker/contexts.rs
@@ -77,7 +77,7 @@ impl TypeMap {
     /// [ListUnionType]: TypeSignature::ListUnionType
     pub fn concretize(&mut self) -> CheckResult<()> {
         for ty in self.map.values_mut() {
-            *ty = concretize_deep(ty.clone())?;
+            *ty = ty.clone().concretize_deep()?;
         }
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


clarity/src/vm/analysis/type_checker/contexts.rs
```diff
@@ -136,3 +152,46 @@ impl<'a> TypingContext<'a> {
         }
     }
 }
+
+/// Subroutine of [TypeMap::concretize]
+fn concretize_deep(ty: TypeSignature) -> CheckResult<TypeSignature> {
```
obycode, line 157: Do you think we should modify `TypeSignature::concretize` to do this? Seems like maybe it should already be doing this part for us, right?

obycode (approved):
Thanks @Acaccia. This looks good, I just had one question about the location of the implementation.

clarity/src/vm/analysis/type_checker/contexts.rs
Acaccia: You are very right, I should have thought a bit more about it. Will change that.

obycode: I was actually wondering if it should be inside of `concretize`, not adding a new method, `concretize_deep`. But any way, we haven't found any problems that require that change, so let's not do that for the time being.

Acaccia: Ah yes, I misread you first message. I do prefer to have another method then, to not change the current behavior of something already well established.

PR #4339, opened 2024-02-02
chore: Remove unused variables

jbencin:
### Description

I found some unused variables with `clippy::collection_is_never_read`. I don't know if these are bugs where we intended to use these variables and forgot, or if it's just old code that's no longer needed. Either way, they should be fixed


obycode (approved):
These are simple changes, but I am a bit concerned that these unused variables could be a sign of a bug. I dug into each, and I don't see any problems, but it might be a good idea for @jcnelson to take a quick look.

jcnelson (changes requested):
I fully support dropping user burn supports altogether, but this would be a much more expansive change than just this work here.

obycode:
> I fully support dropping user burn supports altogether, but this would be a much more expansive change than just this work here.

To clarify, this PR is not attempting to change any behavior, it is just removing unused variables. I flagged you @jcnelson to check that these unused variables were not a symptom of an accidental removal of anything important.

jcnelson (approved):

PR #4345, opened 2024-02-06
Hotfix/stackerdb stale view

jcnelson:
This fixes a race condition in the StackerDB system whereby two nodes that process a Stacks block with configuration updates at different times can accidentally declare each other as broken, and ban each other.  This can happen when, for example, the signer of a given slot changes and a node receives a chunk before it has processed the change.

The fix is to repurpose the `rc_consensus_hash` field to refer to the canonical Stacks tip consensus hash, instead of the reward-cycle starting consensus hash.  This not only captures the same information as `rc_consensus_hash`, but also acts as a "view ID" of the chainstate:  two nodes will agree on the same StackerDB replica configurations if they have processed the same Stacks tips.  Furthermore, nodes will NACK each other's requests for inventories and chunks early on if they disagree on the view; they won't ban each other.

### DIFF
--- a/stackslib/src/chainstate/burn/db/sortdb.rs
+++ b/stackslib/src/chainstate/burn/db/sortdb.rs
@@ -3883,26 +3883,13 @@ impl SortitionDB {
             .unwrap_or(&burnchain.first_block_hash)
             .clone();
 
-        let rc = burnchain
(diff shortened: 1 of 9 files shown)


wileyj (approved):

obycode (approved):
lgtm

PR #4287, opened 2024-01-25
Feat/signers cast dkg vote

jferrant:
### Description

This PR adds voting on the aggregate public key to the signer run loop.
AND...a whole lot of other things XD

I have changed the runloop to hold onto at most 2 stacks signers which are essentially wrappers around the fire coordinator and signing round logic that used to be in runloop. This is to enable having a signer that is registered for both the current and next reward cycles. 

There are a plethora of TODOs listed in this PR. I will wait until there is a bit of discussion around them before creating tickets for each of them. 

### Applicable issues

- closes:
#3981
#3971
#3973
#4243
#4283
#4273
#4299
- related to #3968

### Additional info

TODO: 
- #4386 
- #4387 
- #4388

PR #4348, opened 2024-02-07
feat: store signer state in sqlite

### LINKED ISSUE #4242: [Nakamoto] stacks-signer needs track processed blocks responses,  block requests, and block sign requests

hstove:
This PR adds a new mod at `stacks-signer/src/signerdb.rs` with functions to interact with a SQLite database for state management.

This is built on top of #4287 because that PR refactors a lot of the signer file structure.

At the moment, this replaces a `HashMap` that was used for storing block information.

- closes #4242


stacks-signer/src/client/stackerdb.rs
```diff
@@ -103,14 +104,21 @@ impl StackerDB {
...
             }
-            if let Some(reason) = chunk_ack.reason {
-                // TODO: fix this jankiness. Update stackerdb to use an error code mapping instead of just a string
-                // See: https://github.com/stacks-network/stacks-blockchain/issues/3917
-                if reason.contains("Data for this slot and version already exist") {
-                    warn!("Failed to send message to stackerdb due to wrong version number {}. Incrementing and retrying...", slot_version);
-                } else {
-                    warn!("Failed to send message to stackerdb: {}", reason);
-                    return Err(ClientError::PutChunkRejected(reason));
+            if let Some(code) = chunk_ack.code {
```
jferrant, line 107: Just out of curiousity, did it not work to do 
 `let code: StackerDBErrorCodes = serde_json::from_str(&chunk_ack.reason)?`
 
 I think I tried this myself (thought it would work because I think reason is serialized from a StackerDBErrorCode to begin with) and it didn't work but just want to confirm with you too.

stacks-signer/Cargo.toml
```diff
@@ -42,10 +42,16 @@ tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 wsts = { workspace = true }
 rand = { workspace = true }
+r2d2 = "0.8.9"
```
jcnelson, line 45: Please just use `rusqlite`

stacks-signer/src/signerdb.rs
```diff
@@ -0,0 +1,304 @@
...
+CREATE TABLE IF NOT EXISTS blocks (
+    signer_signature_hash TEXT PRIMARY KEY,
+    block_info TEXT NOT NULL
+)";
+
+impl SignerDb {
+    /// Create a new `SignerState` instance.
+    /// This will create a new SQLite database at the given path
+    /// if one doesn't exist.
+    pub fn new(db_path: &PathBuf) -> Result<SignerDb, DBError> {
```
jferrant, line 54: Just a nit. I would maybe make this an optional. If it is none, then open in mem :) instead of enforcing a specific string.

stacks-signer/src/runloop.rs
```diff
@@ -365,10 +402,15 @@ impl<C: Coordinator> RunLoop<C> {
...
-            self.blocks.insert(
-                block.header.signer_signature_hash(),
-                BlockInfo::new(block.clone()),
-            );
+            // self.blocks.insert(
+            //     block.header.signer_signature_hash(),
+            //     BlockInfo::new(block.clone()),
+            // );
+
+            println!("Inserting into DB in handle_proposed_blocks: {:?}", block);
```
jcnelson, line 410: Before converting to a full PR, please remove all `println!()`s and use the logging facilities.  Thanks!

stacks-signer/Cargo.toml
jcnelson: Now that I did a first-read of the PR, I think it's safe to say that you really do not need this at all.

hstove: That's how I started, but because `rusqlite::Connection` isn't thread safe and the signer's runloop is spawned in threads, I couldn't find any good way of managing connections other than opening and closing for every read/write. I'm 100% open to a better alternative - using `r2d2` was really just me finding a solution that worked

jcnelson: You'll need to instantiate the `rusqlite::Connection` within the thread, instead of putting yourself in a position of having to pass it between threads.

jcnelson: Like, the way we do this elsewhere in the code is we wrap the `rusqlite::Connection` struct within its own `struct`, and then instantiate _that_ struct within the thread that will use it.  These structs aren't singletons; we instead rely on sqlite for ACID guarantees in the event that multiple threads read and write to the DB via separate connections.

hstove: Thanks for the pointer! Makes sense, I'm working on it.

jcnelson: :pray:

saralab:
Reviews can progress , hold off merge until casting DKG vote PR is merged

stacks-signer/Cargo.toml
jferrant: ah neat! I was wondering exactly how we got around this in stacks block core code.

hstove:
@jcnelson FYI I've rebased off of Jacinta's PR and removed all of the `r2d2` stuff. I'll admit that I couldn't figure out maintaining a connection, so this PR just opens a new connection for every read/write. The issue stems from the `Signer` trait in `libsigner` requiring the `signer_loop` to have the `Sync` trait.

I follow your idea of instantiating a struct inside of a thread, but I haven't found a way of doing that without getting the same compiler error. I would still need some struct to have something like `signer_db: Optional<SignerDb>`, but that doesn't fix the fact that it breaks the `Sync` trait. It's not clear if I would need to use something like a Mutex or a channel. I apologize for hitting the boundaries of my current Rust knowledge - any more advice would be helpful!

stacks-signer/src/client/stackerdb.rs
hstove: That doesn't work because `StackerDBErrorCodes` doesn't implement `Deserialize` - this `reason` comes from the custom method `code.into_json()`

stacks-signer/src/signer.rs
```diff
@@ -594,25 +632,28 @@ impl Signer {
...
-                });
-            return false;
+        let mut block_info = match self
+            .signer_db
+            .block_lookup(&signer_signature_hash)
+            .expect("Failed to connect to signer DB")
+        {
+            Some(block_info) => block_info,
+            None => {
+                debug!("We have received a block sign request for a block we have not seen before. Cache the nonce request and submit the block for validation...");
```
jferrant, line 642: Should fix this debug log to include Signer #{} again.

stacks-signer/src/signer.rs
```diff
@@ -361,34 +379,48 @@ impl Signer {
...
             BlockValidateResponse::Reject(block_validate_reject) => {
                 let signer_signature_hash = block_validate_reject.signer_signature_hash;
-                let Some(block_info) = self.blocks.get_mut(&signer_signature_hash) else {
-                    // We have not seen this block before. Why are we getting a response for it?
-                    debug!("Signer #{}: Received a block validate response for a block we have not seen before. Ignoring...", self.signer_id);
-                    return;
+                let mut block_info = match self.signer_db.block_lookup(&signer_signature_hash) {
+                    Ok(Some(block_info)) => block_info,
+                    Ok(None) => {
+                        // We have not seen this block before. Why are we getting a response for it?
```
jferrant, line 416: I think actually all the logs accidentally remove "Signer #{}" from them. Make sure to add these back.

stacks-signer/src/tests/conf/signer-0.toml
```diff
@@ -2,3 +2,4 @@ stacks_private_key = "6a1fc1a3183018c6d79a4e11e154d2bdad2d89ac8bc1b0a021de8b4d28
 node_host = "127.0.0.1:20443"
 endpoint = "localhost:30000"
 network = "testnet"
+db_path = ":memory:"
```
jferrant, line 5: This should get deleted

netrome:
@hstove thanks for letting me dance with this PR. I have done a few things now which allows us to reuse a connection in the SignerDB.

### 1. Clean up trait bounds on `libsigner::Signer` struct
This is just a refactor. It is easy to put trait bounds on structs, but they are rarely needed and often just gets in the way. This allows us to clean up the `PhantomData` ugliness

### 2. Relax `Sync` requirement on the run loop in `libsigner::Signer::spawn`
I don't see why the run loop should need to be sync. I suspect it was just added together with `Send` as those two often goes together. Ideally I don't think the run loop should have to be send either, but since we currently instantiate it outside of the thread it would require a more substantial rewrite to remove this requirement - and just getting rid of `Sync` is good enough for now.

@jferrant, @jcnelson, @xoloki  or anyone else - let me know if we need run loops to be `Sync` (i.e. can be shared between threads) for some future requirement or if there is any thought behind it.

### 3. Make SignerDB hold a rusqlite connection
With this relaxed requirement, SignerDB can hold a connection without any problem. I also updated and simplified some parts of the code. For example, there's no need to explicitly begin and commit a transaction for single statements.

The current solution is quite simple as the struct just holds a connection. We might want to consider adding logic to reconnect if we encounter certain errors, but it seemed from previous comments that we wanted to start simple here for now - so I don't want to begin building a custom connection pool out of the blue. The signer should be designed for fault tolerance, and with persistence it should be safe to restart any failing signer if for example a connection is lost.

I will proceed to address unresolved comments on this PR

hstove:
Awesome! What a nice cleanup, and I appreciate the very clear commits + comments.

One side note - @jferrant mentioned above that it would be nice to have db_path be optional, and to default to memory in that case. I agree that it makes things like tests easier, but I worry that without validation it makes it too easy to use a memory DB when running the signer in prod, which we definitely don't want. Can I request that we either support memory with the string ":memory" or throw an error when spinning up the signer via `stacks-signer run`?

netrome:
> Awesome! What a nice cleanup, and I appreciate the very clear commits + comments.
> 
> One side note - @jferrant mentioned above that it would be nice to have db_path be optional, and to default to memory in that case. I agree that it makes things like tests easier, but I worry that without validation it makes it too easy to use a memory DB when running the signer in prod, which we definitely don't want. Can I request that we either support memory with the string ":memory" or throw an error when spinning up the signer via `stacks-signer run`?

Sure, that's a good point. I'd prefer then to also have the `SignerDB::in_memory()` constructor to make it explicit for users who don't know the ":memory:" [special file name](https://www.sqlite.org/inmemorydb.html). I just learned about it looking at your code. I'll do that unless you object.

netrome:
> Sure, that's a good point. I'd prefer then to also have the `SignerDB::in_memory()` constructor to make it explicit for users who don't know the ":memory:" [special file name](https://www.sqlite.org/inmemorydb.html). I just learned about it looking at your code. I'll do that unless you object.

Actually I won't since the tests are using temporary database files so the `in_memory()` constructor would just be dead code.

stacks-signer/src/tests/conf/signer-0.toml
netrome: We made it mandatory again (I presume), so I've kept this and added the same to `signer-1.toml`. Let me know if you disagree with having this parameter required in the config.

jferrant:
> > Sure, that's a good point. I'd prefer then to also have the `SignerDB::in_memory()` constructor to make it explicit for users who don't know the ":memory:" [special file name](https://www.sqlite.org/inmemorydb.html). I just learned about it looking at your code. I'll do that unless you object.
> 
> Actually I won't since the tests are using temporary database files so the `in_memory()` constructor would just be dead code.

Could also add some tests for in memory rather than using files :P

jferrant:
You need to update the construction of the signer file to include the database name

jferrant:
Whenever I try to run the integration tests I get complaints about too many open files.

netrome:
> You need to update the construction of the signer file to include the database name. i.e. add `db_path = ":memory:"`. See build_signer_config_tomls

Ah I see. I will update this. This feels like the type of error that shouldn't pass compilation or unit tests. I might refactor that function in the process if I spot any simple enhancements.

xoloki:
> Whenever I try to run the integration tests I get complaints about too many open files. I thought maybe you would need to reduce the number of signers, but I am surprised that having 10 open files is too much (it runs with 10 signers)
> 
> E.g. test that failed for me `BITCOIND_TEST=1 RUST_LOG=debug STACKS_LOG_DEBUG=1 cargo test --package stacks-node --bin stacks-node -- tests::signer::stackerdb_dkg --nocapture --ignored`

Remember that each signer has multiple connections (event queues, web server, rpc client, etc).  If we're adding sqlite to the mix we might be hitting some limits.

netrome:
> > Whenever I try to run the integration tests I get complaints about too many open files. I thought maybe you would need to reduce the number of signers, but I am surprised that having 10 open files is too much (it runs with 10 signers)
> > E.g. test that failed for me `BITCOIND_TEST=1 RUST_LOG=debug STACKS_LOG_DEBUG=1 cargo test --package stacks-node --bin stacks-node -- tests::signer::stackerdb_dkg --nocapture --ignored`
> 
> Interesting. I'll try running it and see if I encounter the same behavior.

I have tried multiple times now and I'm not managing to reproduce any error so far.

netrome:
> You need to update the construction of the signer file to include the database name. i.e. add `db_path = ":memory:"`. See build_signer_config_tomls

Done ✅

netrome:
> Could also add some tests for in memory rather than using files :P

Sure. I added another test now. It feels a bit outside our domain to do this since we have no specific code path for the in-memory database in our code base, but why not? Let's be defensive.

stackslib/src/net/api/poststackerdbchunk.rs
```diff
@@ -145,6 +145,14 @@ impl StackerDBErrorCodes {
...
         })
     }
+
+    pub fn from_code(code: u32) -> Option<Self> {
+        match code {
+            0 => Some(Self::DataAlreadyExists),
+            1 => Some(Self::NoSuchSlot),
+            _ => None,
+        }
+    }
```
hstove, line 155: @netrome this is what's needed to fix the output mutants check. You can find this by viewing [the "Summary" page on that check](https://github.com/stacks-network/stacks-core/actions/runs/8177496330?pr=4348)

```suggestion
    #[cfg_attr(test, mutants::skip)]
    pub fn from_code(code: u32) -> Option<Self> {
        match code {
            0 => Some(Self::DataAlreadyExists),
            1 => Some(Self::NoSuchSlot),
            _ => None,
        }
    }
```

netrome (approved):

jferrant (approved):

stackslib/src/net/api/poststackerdbchunk.rs
netrome: Oh, thanks. I missed that. :pray:

netrome (approved):

jferrant (approved):

PR #4350, opened 2024-02-07
chore: merge tag nakamoto-v0.4

obycode:


stackslib/src/chainstate/nakamoto/signer_set.rs
```diff
@@ -0,0 +1,387 @@
...
+impl RawRewardSetEntry {
+    pub fn from_pox_4_tuple(is_mainnet: bool, tuple: TupleData) -> Result<Self, ChainstateError> {
+        let mut tuple_data = tuple.data_map;
+
+        let pox_addr_tuple = tuple_data
+            .remove("pox-addr")
+            .expect("FATAL: no `pox-addr` in return value from (get-reward-set-pox-address)");
+
+        let reward_address = PoxAddress::try_from_pox_tuple(is_mainnet, &pox_addr_tuple)
+            .expect(&format!("FATAL: not a valid PoX address: {pox_addr_tuple}"));
```
krl, line 110: Is there a good reason not to treat this as a normal error?

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -750,7 +906,8 @@ impl StacksChainState {
                 POX_1_NAME,
                 &format!("(get-reward-set-size u{})", reward_cycle),
             )?
-            .expect_u128();
+            .expect_u128()
+            .expect("FATAL: unexpected PoX structure");
```
krl, line 910: Expect in function returning Result

krl (approved):
Woah this is a massive PR, I've been looking at only the .rs files, and ignoring the tests to try to limit the scope, and have left a few comments. 

A lot of new code, for example libsigner/src/messages.rs, is interspaced with mostly trivial error handling updates, I have had a look at them but in this scope it's not possible for me to comment on larger chunks of new code, also since i don't know my way around this codebase all too much.

In general, i wish i had a description of the PR to work from.

obycode:
Ok, sorry I should have specified. We really just want to verify that this merge doesn't affect our changes. Maybe it would be better if instead I rebase our changes on top?

cylewitruk:
There were a lot of conflicts in this PR with non-Wasm-related code, so closing this for a retry.

PR #4353, opened 2024-02-08
fix: fix #4351

### LINKED ISSUE #4351: /v2/info only updates on burn block
In running the managed testnets, we realized that the /v2/info endpoint does not stay up-to-date with the current Stacks tip height. Instead it is only updating when a new burn block is processed.

jcnelson:
This fixes #4351


kantai (approved):
LGTM

zone117x (approved):
Thanks @jcnelson!

PR #4360, opened 2024-02-09
Process successful .signers-voting results

### LINKED ISSUE #4322: [clarity] ` .signer-voting` need to accept out of round votes

### LINKED ISSUE #4354: Stacks node must process successful `.signers-voting` results
The `.signers-voting` contract emits an event every time a vote is cast, but the `stacks-node` needs to use this data for getting the actual aggregate public key.

This is the method that currently loads the aggregate public key:

https://github.com/stacks-network/stacks-core/blob/next/stackslib/src/chainstate/nakamoto/mod.rs#L1800

It would need to be updated to invoke something like

```clar
(get-approved-aggregate-key reward-cycle) -> (option (buff 33))
```

Where `get-approved-aggregate-key` is a read-only method of `.signers-voting`.

obycode:
### Description

Adds functionality into the .signers-voting contract to track when an aggregate public key has reached the 70% threshold to be accepted as the key for the next cycle. This key can be queried via the `get-approved-aggregate-key` read-only function.

### Applicable issues

- Builds off of #4323 (will rebase after that is merged)
- Fixes #4354
- Includes #4327 to fix #4322

### Checklist

- [ ] Changelog is updated


stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -77,35 +73,88 @@
...
+        )
+    )
+)
+
+;; Signer vote for the aggregate public key of the next reward cycle
+;;  The vote happens in the prepare phase of the current reward cycle but may be ran more than
+;;  once resulting in different 'rounds.' Each signer vote is based on the weight of stacked
+;;  stx tokens & fetched from the .signers contract. The vote is ran until the consensus 
+;;  threshold of 70% for a specific aggregate public key is reached. 
 (define-public (vote-for-aggregate-public-key (signer-index uint) (key (buff 33)) (round uint))
```
jferrant, line 108: This signer index...I am retreiving it with the following logic

stackerdb-get-signer-slots-page and then matching the offset in the returned list with the signer address to know the signer index...is this correct?

jferrant, line 108: I wonder if this needs to take the reward cycle as an option. 

For example Signer N could be registered for the current and the next reward cycle.

Assume Signer N gets assigned the same slot index in both page 0 and page 1 of the stacker db slot. 

For whatever reason, reward cycle page 0 requires a DKG vote. Simultaneously, reward cycle with page 1 is running a DKG vote. How can you distinguish these two votes if you do not include the reward cycle or page number? (where page number is reward cycle modulus 2 )

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -77,35 +73,88 @@
...
-        (asserts! (>= round (default-to u0 (map-get? rounds reward-cycle))) err-old-round)
-        (asserts! (is-eq (len key) u33) err-ill-formed-aggregate-public-key)
-        (asserts! (is-valid-aggregated-public-key key {reward-cycle: reward-cycle, round: round}) err-duplicate-aggregate-public-key)
-        (asserts! (map-insert votes {reward-cycle: reward-cycle, round: round, signer: tx-sender} {aggregate-public-key: key, reward-slots: num-slots}) err-duplicate-vote)
+            ;; vote by signer weight
+            (signer-weight (try! (get-current-signer-weight signer-index)))
+            (new-total (+ signer-weight (default-to u0 (map-get? tally tally-key))))
+            (total-weight (try! (get-total-weight reward-cycle))))
+        ;; Check we're in the prepare phase
+        (asserts! (is-in-voting-window burn-block-height reward-cycle) (err ERR_OUT_OF_VOTING_WINDOW))
```
jferrant, line 116: I am not sure we should have a limitation on voting window. what happens if the signers fail during the prepare phase to complete dkg...they should still be able ot perform dkg in the following reward cycle and the old signers continue to prop up the system until this actually happens. 

This makes me think I need to change my own signer logic to not immediately shut down a signer if it passes its own reward cycle...

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: I think this is resolved after our discussion, correct? Let me know if there is still an open question here.

obycode: @setzeus has now added this.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: @setzeus updated this check in the latest.

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -101,6 +102,23 @@ fn advance_to_nakamoto(
...
+                    make_signers_vote_for_aggregate_public_key(
+                        &test_stacker.signer_private_key,
+                        0,
+                        index as u128,
+                        &test_signers.aggregate_public_key,
+                        0,
+                        7,
+                    )
+                })
+                .collect()
```
obycode, line 121: This is not the correct way to get the signer indexes. We need to figure out how to get them here and sign once per signer with the correct index.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,32 +73,84 @@
...
+        (print {
+            event: "voted",
+            signer: tx-sender,
+            reward-cycle: reward-cycle,
+            round: round,
+            key: key,
+            new-total: new-total,
+        })
+        ;; Check if consensus has been reached
+        (and
```
setzeus, line 136: Personal preference but I'd split this up into a (begin ...) of sequential checks.

Currently if a 'false' is returned there's no way to tell what exactly threw it. I'd expect either an explicit error if key is already set.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,32 +73,84 @@
...
+            ;; If we already have consensus, skip this
+            (is-none (map-get? aggregate-public-keys reward-cycle))
+            ;; If the new total weight is greater than or equal to the threshold consensus
+            (>= (/ (* new-total u1000) total-weight) threshold-consensus)
+            ;; Save this approved aggregate public key for this reward cycle
+            (map-set aggregate-public-keys reward-cycle key)
```
setzeus, line 142: Why not use a "map-insert" here since it can only be set once & not overwritten?

setzeus (changes requested):
Mainly syntax NITs except for the sequential (and ...) near the end of the vote function.

Assuming these are fixed & includes everything @jferrant needs to continue tests it should be GTM.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,32 +73,84 @@
             )
         (get prepare-cycle-length pox-info)))
 
+;; get the aggregate public key for the given reward cycle (or none)
+(define-read-only (get-approved-aggregate-key (reward-cycle uint))
```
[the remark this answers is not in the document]
obycode: Ok, I'll do that to match, but it's not my favorite 😅.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: I don't think it should be an error if the key is already set, because then the last 30% to vote would always get an error.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: Good point, we could just remove the `(is-none (map-get? ...))` and just use the `map-insert` to break the `and` if it is already set.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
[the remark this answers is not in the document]
obycode: Changed in d8c8cfafd09cc3864eb2a61cd6f22f28bf3a51ce.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,32 +73,84 @@
...
-            (tally-key {reward-cycle: reward-cycle, round: round, aggregate-public-key: key})
-            ;; one slot, one vote
-            (num-slots (try! (get-current-signer-slots signer-index)))
-            (new-total (+ num-slots (default-to u0 (map-get? tally tally-key)))))
-        (asserts! (is-in-voting-window burn-block-height reward-cycle) err-out-of-voting-window)
-        (asserts! (>= round (default-to u0 (map-get? rounds reward-cycle))) err-old-round)
-        (asserts! (is-eq (len key) u33) err-ill-formed-aggregate-public-key)
-        (asserts! (is-valid-aggregated-public-key key {reward-cycle: reward-cycle, round: round}) err-duplicate-aggregate-public-key)
-        (asserts! (map-insert votes {reward-cycle: reward-cycle, round: round, signer: tx-sender} {aggregate-public-key: key, reward-slots: num-slots}) err-duplicate-vote)
+(define-private (sum-weights (signer { signer: principal, weight: uint }) (acc uint))
```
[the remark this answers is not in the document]
obycode: Changed in d8c8cfafd09cc3864eb2a61cd6f22f28bf3a51ce.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,32 +73,84 @@
...
+    (+ acc (get weight signer))
+)
+
+(define-private (get-total-weight (reward-cycle uint))
```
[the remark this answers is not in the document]
obycode: Changed in d8c8cfafd09cc3864eb2a61cd6f22f28bf3a51ce.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,32 +73,84 @@
...
+    (match (map-get? cycle-total-weight reward-cycle)
+        total (ok total)
+        (let (
```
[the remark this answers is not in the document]
obycode: Changed in d8c8cfafd09cc3864eb2a61cd6f22f28bf3a51ce.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -3,27 +3,38 @@
...
+(define-constant ERR_INVALID_BURN_BLOCK_HEIGHT u8)
+(define-constant ERR_FAILED_TO_RETRIEVE_SIGNERS u9)
 
 (define-constant pox-info
     (unwrap-panic (contract-call? .pox-4 get-pox-info)))
 
+;; Threshold consensus (in 3 digit %)
+(define-constant threshold-consensus u700)
+
 ;; maps reward-cycle ids to last round
```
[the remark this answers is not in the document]
obycode: Changed in d8c8cfafd09cc3864eb2a61cd6f22f28bf3a51ce.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: Changed in d8c8cfafd09cc3864eb2a61cd6f22f28bf3a51ce.

hstove (commented):
Looking good!

stackslib/src/chainstate/nakamoto/signer_set.rs
```diff
@@ -274,7 +274,7 @@ impl NakamotoSigners {
                                 "signer".into(),
                                 Value::Principal(PrincipalData::from(signing_address)),
                             ),
-                            ("weight".into(), Value::UInt(signer.slots.into())),
+                            ("weight".into(), Value::UInt(signer.stacked_amt.into())),
```
jferrant, line 277: I think stacked_amt needs to be normalized no? Or is this already? 

I.e. shoudl be some integer from 0 to 4000?

stackslib/src/chainstate/nakamoto/signer_set.rs
```diff
@@ -274,7 +274,7 @@ impl NakamotoSigners {
...
                         ])
                             .expect(
                                 "BUG: Failed to construct `{ signer: principal, num-slots: u64 }` tuple",
```
jferrant, line 280: This comment should be updated to reflect the "weight"

stackslib/src/chainstate/nakamoto/signer_set.rs
obycode: I was not aware that this is supposed to be normalized. I thought it was just supposed to be the amount of STX?

stackslib/src/chainstate/nakamoto/signer_set.rs
obycode: Fixed in c4be109ff.

stackslib/src/chainstate/nakamoto/signer_set.rs
jferrant: This "weight" is retrieved from the reward set entry which is a number of slots value and there are only 4000 of them. If signer A has 10 slots, they get key ids 1 through 10. The next signer in the list, Signer B, has 12 slots, they get key ids 11 through 22., etc. the max key id we can have is 4000. I do not believe simply putting the stacked amt is correct.

obycode: Ok, so I'm just going to revert this change and use the `slots` field, which seems to be the number of reward slots.

obycode: but I'm also going to rename it to `reward-slots` to avoid confusion with stacker DB slots, which is a term also used in this file.

jferrant: Sure though weight is also fine to keep (its what is used inside reward slot entry inside the rust codebase. Would prob try to keep it consistent)

obycode: Ok, then I'll rename the field in the struct to be `weight` then.

obycode: Done.

setzeus (approved):

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -39,21 +50,25 @@
...
+(define-read-only (get-current-signer-weight (signer-index uint))
     (let ((cycle (+ u1 (burn-height-to-reward-cycle burn-block-height))))
-      (get-signer-slots signer-index cycle)))
+      (get-signer-weight signer-index cycle)))
 
-(define-read-only (get-signer-slots (signer-index uint) (reward-cycle uint))
-    (let ((details (unwrap! (try! (contract-call? .signers get-signer-by-index reward-cycle signer-index)) err-invalid-signer-index)))
-        (asserts! (is-eq (get signer details) tx-sender) err-signer-index-mismatch)
+(define-read-only (get-signer-weight (signer-index uint) (reward-cycle uint))
+    (let ((details (unwrap! (try! (contract-call? .signers get-signer-by-index reward-cycle signer-index)) (err ERR_INVALID_SIGNER_INDEX))))
```
friedger, line 65: As a user of this function, I can receive `(err u2)` and don't know whether the error came from this contract or from `signers` contract.

The error codes should not overlap.

friedger (commented):
Just some comments

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -3,39 +3,37 @@
...
+(define-constant ERR_ILL_FORMED_AGGREGATE_PUBLIC_KEY u5)
+(define-constant ERR_DUPLICATE_AGGREGATE_PUBLIC_KEY u6)
+(define-constant ERR_DUPLICATE_VOTE u7)
+(define-constant ERR_INVALID_BURN_BLOCK_HEIGHT u8)
+(define-constant ERR_FAILED_TO_RETRIEVE_SIGNERS u9)
 
 (define-constant pox-info
     (unwrap-panic (contract-call? .pox-4 get-pox-info)))
 
+;; Threshold consensus (in 3 digit %)
```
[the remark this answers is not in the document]
obycode: Thanks for pointing that out -- I understand that confusion. I like the thought of using basis points, but is that well known enough by people who would be likely to read this code? "ppk" or "parts per thousand" seems a bit unusual to me too. Maybe we should just expand the comment to explain it more fully? How about:

```
;; Threshold consensus, expressed as parts-per-thousand to allow for integer
;; division with higher precision (e.g. 700 for 70%).

File: stackslib/src/chainstate/nakamoto/signer_set.rs (hunk 11 above)
Comment (obycode) [reply stackslib/src/chainstate/nakamoto/signer_set.rs]: It seems there were some misunderstandings and back and forth there, but what we landed on is that this should be the number of reward slots.

File: stackslib/src/chainstate/stacks/boot/signers-voting.clar (hunk 13 above)
Comment (obycode) [reply stackslib/src/chainstate/stacks/boot/signers-voting.clar:65]: Good point!

Comment (obycode) [reply stackslib/src/chainstate/stacks/boot/signers-voting.clar:65]: @setzeus, since you're already modifying these, can you update the error values in your PR to avoid overlap?

Comment (setzeus) [reply stackslib/src/chainstate/stacks/boot/signers-voting.clar:65]: @obycode @friedger thoughts on different prepended values per contract? IE:

signers: 100, 101, 102
signrs-voting: 200, 201,202

**Comment by obycode** [comment]
Looks like we still have these two tests failing
- `tests::signer::stackerdb_dkg_sign`
- `tests::signer::stackerdb_block_proposal`

@jferrant do these tests change with your PR, or should I fix them here?

**Comment by xoloki** [comment]
> Looks like we still have these two tests failing
> 
> * `tests::signer::stackerdb_dkg_sign`
> * `tests::signer::stackerdb_block_proposal`
> 
> @jferrant do these tests change with your PR, or should I fix them here?

dkg_sign changes significantly in Jacintas branch.

File: stackslib/src/burnchains/mod.rs [hunk 15]
```diff
@@ -595,6 +595,9 @@ impl PoxConstants {
 
             // NOTE: first block in reward cycle is mod 1, so mod 0 is the last block in the
             // prepare phase.
+            // TODO: I *think* the logic of `== 0` here requires some further digging.
+            //  `mod 0` may not have any rewards, but it does not behave like "prepare phase" blocks:
+            //  is it already a member of reward cycle "N" where N = block_height / reward_cycle_len
```
kantai, line 600: I added this comment here for some visibility -- this may impact block inventory logic, right, @jcnelson?

The "0" reward index isn't really a prepare phase block for cycle N: it's a member of reward cycle N, not N-1.

testnet/stacks-node/src/config.rs
```diff
@@ -506,13 +506,19 @@ lazy_static! {
...
         if !(self.burnchain.mode == "nakamoto-neon" || self.burnchain.mode == "mockamoto") {
             return None;
         }
         self.miner.self_signing_key.clone()
     }
 
+    #[cfg(not(any(test, feature = "testing")))]
+    pub fn self_signing(&self) -> Option<TestSigners> {
+        return None;
+    }
```
kantai, line 520: Should `self_signing` work outside of testing, or should it just be a test only mode for nakamoto?

I could be convinced either way I think. But if we do make self-signing test only, it may be better to just delete mockamoto and the self signing code paths. Then the existing nakamoto tests could just be updated to blindly send signatures to the stackerdb on responses. Thinking it through, I think I'm convincing myself that this would be better. I'll open an issue.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -3,27 +3,42 @@
...
-;; maps reward-cycle ids to last round
+;; Threshold consensus, expressed as parts-per-thousand to allow for integer
+;; division with higher precision (e.g. 700 for 70%).
+(define-constant threshold-consensus u700)
+
+;; Maps reward-cycle ids to last round
 (define-map rounds uint uint)
 
+;; Maps reward-cycle ids to aggregate public key.
+(define-map aggregate-public-keys uint (buff 33))
```
setzeus, line 35: Great job on importing the tests from #4392, going to close accordingly.

From writing the tests^ I believe everything LGTM, the *only* thing I can point out atm that we may need to change but isn't breaking anything is that we now have two maps that track cycle-id to aggregate public key (here & in PoX-4.clar).

The latter map currently isn't written into or fetched so probably best to just remove from PoX-4.clar & keep the one in here.

setzeus (dismissed):

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -39,21 +54,25 @@
...
+    total-weight: (map-get? cycle-total-weight reward-cycle)})
+
 (define-read-only (get-tally (reward-cycle uint) (round uint) (aggregate-public-key (buff 33)))
     (map-get? tally {reward-cycle: reward-cycle, round: round, aggregate-public-key: aggregate-public-key}))
 
-(define-read-only (get-current-signer-slots (signer-index uint))
+(define-read-only (get-current-signer-weight (signer-index uint))
     (let ((cycle (+ u1 (burn-height-to-reward-cycle burn-block-height))))
-      (get-signer-slots signer-index cycle)))
+      (get-signer-weight signer-index cycle)))
```
kantai, line 66: I'd suggest deleting `get-current-signer-weight`. Any caller can just use `get-signer-weight` and handle figuring out the current reward cycle themselves. I think the invocation in `vote-for-aggregate-public-key` is potentially dangerous as well: that function takes in a `reward-cycle` argument, but invokes `get-current-signer-weight`, so I think there could end up being a mismatch in the queried reward cycle.

kantai (commented):
This looks good to me, just had a comment in the contract about `get-current-signer-weight`

stackslib/src/burnchains/mod.rs
jcnelson: Changing this would impact far more than block inventory logic.  I'm afraid this is consensus-level logic.

Fundamentally, there's an unsoundness in the code for deciding whether or not a block height is in the prepare phase or reward phase.  In some places, like this one and others, a block whose height modulo the reward cycle length is 0 is in the prepare phase, since the reward phase begins when the block height modulo the reward cycle length is 1 or greater.  In other places, like `static_block_height_to_reward_cycle`, a block is in the reward phase if its height modulo the reward cycle length is 0.

Unfortunately, this code here (as well as the code paths that declare the start of a reward cycle to be at each block whose heights modulo the reward cycle length is 1) is used all over the place in the blockchain.  We can't change it.

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -101,21 +109,118 @@ fn advance_to_nakamoto(
...
+    let signer_vec = match signers_res.expect_optional().unwrap() {
+        Some(signer_vec) => signer_vec.expect_list().unwrap(),
+        None => {
+            debug!("No signers set for cycle {}", cycle_id);
+            return vec![];
+        }
+    };
+
+    let mut signers_to_index = HashMap::new();
+    for (index, value) in signer_vec.into_iter().enumerate() {
```
jcnelson, line 174: What if there are duplicate signers?  I think the code that calls `set-signers` guarantees this, but perhaps there should be an assertion here as well?

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
```diff
@@ -1673,6 +1868,41 @@ pub fn simple_nakamoto_coordinator_10_extended_tenures_10_sortitions() -> TestPe
 
         debug!("Next burnchain block: {}", &consensus_hash);
 
+        let block_height = peer
```
jcnelson, line 1871: Seems this could be DRY'ed up with the code at line 1140.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -2658,102 +2646,6 @@ impl NakamotoChainState {
         Ok(lockup_events)
     }
 
-    /// Set the aggregate public key for verifying stacker signatures.
-    /// TODO: rely on signer voting instead
-    /// DO NOT USE IN MAINNET
```
jcnelson, line 2663: :raised_hands:

stackslib/src/chainstate/nakamoto/test_signers.rs
```diff
@@ -0,0 +1,195 @@
+// Copyright (C) 2013-2020 Blockstack PBC, a public benefit corporation
```
jcnelson, line 1: If this is all test-only code, then please put it into `stackslib/src/chainstate/nakamoto/tests/test_signers.rs`.

stackslib/src/chainstate/nakamoto/tests/node.rs
```diff
@@ -1056,6 +983,22 @@ impl<'a> TestPeer<'a> {
         let mut stacks_node = self.stacks_node.take().unwrap();
         let sortdb = self.sortdb.take().unwrap();
 
+        let burn_height = self
```
jcnelson, line 986: Can this 6-line-long chain of references be turned into a method?

jcnelson: Actually, I think lines 986 through 1001 should be their own method, since you have it listed out twice.

testnet/stacks-node/src/config.rs
obycode: That sounds good to me. So we'll leave this as-is and do that in a separate PR?

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: Good call. Deleted in a92dbdb83.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: Yes, good catch! Fixed in f596c0bf5

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
obycode: Added in 51f0c9142.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,35 +81,96 @@
...
-            signer: tx-sender, 
-            reward-cycle: reward-cycle, 
-            round: round, 
-            key: key, 
-            new-total: new-total })
-        (ok true)))
+(define-private (sum-weights (signer { signer: principal, weight: uint }) (acc uint))
+    (+ acc (get weight signer)))
+
+(define-private (get-total-weight (reward-cycle uint))
```
jcnelson, line 92: Can you name this `get-or-set-total-weight` since it could write data?

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,35 +81,96 @@
...
+    (match (map-get? cycle-total-weight reward-cycle)
+        total (ok total)
+        (let ((signers (unwrap! (contract-call? .signers get-signers reward-cycle) (err ERR_FAILED_TO_RETRIEVE_SIGNERS)))
+                (total (fold sum-weights signers u0)))
+            (map-set cycle-total-weight reward-cycle total)
+            (ok total))))
 
 (define-private (update-last-round (reward-cycle uint) (round uint))
```
jcnelson, line 100: Can you add an API doc string here to indicate what a `true` or `false` return value should indicate?

Also, there's no documentation as to what `map-set` returns (but that's a separate issue!)

stackslib/src/chainstate/nakamoto/tests/node.rs
obycode: Good idea. Refactored in 0fdaf62fb.

stackslib/src/chainstate/nakamoto/coordinator/tests.rs
obycode: Good idea. Refactored in 0fdaf62fb.

stackslib/src/burnchains/mod.rs
jcnelson: This is also different from `is-in-prepare-phase` in the voting contract.

jcnelson: Also `reward-cycle-to-burn-height` in the voting contract is inconsistent with the code here.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,35 +81,96 @@
...
+;;   * key: the aggregate public key that this vote is in support of
+;;   * round: the voting round for which this vote is intended
+;;   * reward-cycle: the reward cycle for which this vote is intended
+;;  Returns:
+;;   * `(ok true)` if the vote was successful
+;;   * `(err <code>)` if the vote was not successful (see errors above)
+(define-public (vote-for-aggregate-public-key (signer-index uint) (key (buff 33)) (round uint) (reward-cycle uint))
+    (let ((tally-key {reward-cycle: reward-cycle, round: round, aggregate-public-key: key})
+            ;; vote by signer weight
+            (signer-weight (try! (get-current-signer-weight signer-index)))
```
jcnelson: I think there's potentially a logic bug here.  This function takes a `reward-cycle` as an argument, but it uses `get-current-signer-weight` instead of `get-signer-weight`.  This implicitly assumes `reward-cycle` is the same as the upcoming reward cycle, but this is never checked.

stackslib/src/chainstate/nakamoto/test_signers.rs
kantai: It isn't -- it's used by mockamoto and the self signing behavior of the stacks-node.

kantai (approved):

stackslib/src/chainstate/stacks/boot/signers-voting.clar
jcnelson: Generally speaking, I'd prefer it if Clarity code didn't use globals except in the top-level calling context (and even then, only sparingly).  The factoring I'd like to see is as follows:

```clarity
(define-public (vote-for-aggregate-public-key (signer-index uint) (key (buff 33)) (round uint) (reward-cycle uint))
   (begin
      (asserts! (is-eq reward-cycle (+ u1 (burn-height-to-reward-cycle burn-block-height)))
         (err ERR_OUT_OF_VOTING_WINDOW))
      (inner-vote-for-aggregate-public-key signer-index key round reward-cycle)))

(define-private (inner-vote-for-aggregate-public-key (signer-index uint) (key (buff 33)) (round uint) (reward-cycle uint))
   ;; etc. etc., but use `get-signer-weight` instead of `get-current-signer-weight`
```

Then, it's straightforward to unit-test and property-test `inner-vote-for-aggregate-public-key` without having to mock the global state.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,35 +81,96 @@
...
+            (new-total (+ signer-weight (default-to u0 (map-get? tally tally-key))))
+            (total-weight (try! (get-total-weight reward-cycle))))
+        ;; Check that the key has not yet been set for this reward cycle
+        (asserts! (is-none (map-get? aggregate-public-keys reward-cycle)) (err ERR_OUT_OF_VOTING_WINDOW))
+        ;; Check that the aggregate public key is the correct length
```
jcnelson, line 134: If this is a public key on secp256k1, we could further validate that it is at least a well-formed public key by running `principal-construct?` on it.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,35 +81,96 @@
...
+        (asserts! (is-eq (len key) u33) (err ERR_ILL_FORMED_AGGREGATE_PUBLIC_KEY))
+        ;; Check that aggregate public key has not been used in a previous reward cycle
+        (asserts! (is-valid-aggregate-public-key key reward-cycle) (err ERR_DUPLICATE_AGGREGATE_PUBLIC_KEY))
```
jcnelson, line 137: Suggestion -- use a more specific name for this function like `is-novel-aggregate-public-key`, since that's what's actually getting checked.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,35 +81,96 @@
...
+        (print {
+            event: "voted",
+            signer: tx-sender,
+            reward-cycle: reward-cycle,
+            round: round,
+            key: key,
+            new-total: new-total,
+        })
+        ;; Check if consensus has been reached
+        (and
```
jcnelson, line 155: I know `and` evaluates arguments left to right and will short-circuit, but can this be an `(if ...)` expression instead, just for better readability?

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,35 +81,96 @@
...
+            (ok total))))
 
 (define-private (update-last-round (reward-cycle uint) (round uint))
-    (match (map-get? rounds reward-cycle)
-        last-round (and (> round last-round) (map-set rounds reward-cycle round))
-        (map-set rounds reward-cycle round)))
+    (ok (match (map-get? rounds reward-cycle)
+        last-round (begin
+            (asserts! (<= round (+ last-round u1)) (err ERR_INVALID_ROUND))
+            (and (> round last-round) (map-set rounds reward-cycle round)))
```
jcnelson, line 104: Can this be an `(if ...)` expression instead?  I think it's more descriptive of what's actually happening.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,35 +81,96 @@
...
+        ;; Check that the key has not yet been set for this reward cycle
+        (asserts! (is-none (map-get? aggregate-public-keys reward-cycle)) (err ERR_OUT_OF_VOTING_WINDOW))
+        ;; Check that the aggregate public key is the correct length
+        (asserts! (is-eq (len key) u33) (err ERR_ILL_FORMED_AGGREGATE_PUBLIC_KEY))
+        ;; Check that aggregate public key has not been used in a previous reward cycle
+        (asserts! (is-valid-aggregate-public-key key reward-cycle) (err ERR_DUPLICATE_AGGREGATE_PUBLIC_KEY))
+        ;; Check that signer hasn't voted in this reward-cycle & round
+        (asserts! (map-insert votes {reward-cycle: reward-cycle, round: round, signer: tx-sender} {aggregate-public-key: key, signer-weight: signer-weight}) (err ERR_DUPLICATE_VOTE))
+        ;; Check that the round is incremented by at most 1
+        (try! (update-last-round reward-cycle round))
```
jcnelson, line 141: I'm not sure this behavior is what we want.  Suppose two sets of signers are participating in two distinct rounds A and B (A < B w.l.o.g.) in the same reward cycle.  Suppose we get the following votes in rounds A and B:

* A sequence of votes for key K_A that put it at 69% agreement.  The current round is A.
* A vote for key K_B that puts it at 1% agreement.  Now, the current round is B.
* A vote for key K_A that puts it over 70% agreement.  The vote will fail because A < B.

In other words, the code here allows a single malicious signer to deny service to voting by posting incrementally larger rounds over the prepare phase.

The property we want from this function is that the aggregate public key K is chosen as the aggregate public key for the given `reward-cycle` if it is the first key in which signers reach 70% agreement with a fixed round number.  Then, the scenario above doesn't happen; key K_A gets selected when the last vote in that list arrives.  It doesn't matter what the round number is, or what value it has relative to other ongoing rounds.  Also, round numbers can be reused across reward cycles.  All the round number is doing is identifying a DKG view that is shared by a bloc of signers.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,35 +81,96 @@
...
+            ;; Save this approved aggregate public key for this reward cycle.
+            ;; If there is already a key for this cycle, this will return false
+            ;; there will be no duplicate event.
+            (map-insert aggregate-public-keys reward-cycle key)
+            ;; Create an event for the approved aggregate public key
+            (begin
+                (print {
+                    event: "approved-aggregate-public-key",
+                    reward-cycle: reward-cycle,
+                    key: key,
```
jcnelson, line 167: We should also record the round number.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
```diff
@@ -62,35 +81,96 @@
...
+;;   * `(ok true)` if the vote was successful
+;;   * `(err <code>)` if the vote was not successful (see errors above)
+(define-public (vote-for-aggregate-public-key (signer-index uint) (key (buff 33)) (round uint) (reward-cycle uint))
+    (let ((tally-key {reward-cycle: reward-cycle, round: round, aggregate-public-key: key})
+            ;; vote by signer weight
+            (signer-weight (try! (get-current-signer-weight signer-index)))
+            (new-total (+ signer-weight (default-to u0 (map-get? tally tally-key))))
+            (total-weight (try! (get-total-weight reward-cycle))))
+        ;; Check that the key has not yet been set for this reward cycle
+        (asserts! (is-none (map-get? aggregate-public-keys reward-cycle)) (err ERR_OUT_OF_VOTING_WINDOW))
```
jcnelson, line 133: The error is not that we're out of the voting window; it's that the vote has closed since we've already chosen a key.

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
```diff
@@ -136,6 +137,420 @@ pub fn prepare_pox4_test<'a>(
...
+            .into()
+        );
+    } else {
+        panic!("Expected SmartContractEvent, got {:?}", approve_event);
+    }
+}
+
+/// In this test case, Alice & Bob both successfully vote for the same key
+/// Alice votes first, casting one correct vote & four incorrect votes
+/// that hit all tenure-agnostic errors. Bob votes once, successfully.
```
jcnelson, line 298: Comment bug: Bob is voting multiple times in this test.

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
```diff
@@ -136,6 +137,420 @@ pub fn prepare_pox4_test<'a>(
...
+        make_signers_vote_for_aggregate_public_key_value(
+            alice_key,
+            alice_nonce + 5,
+            alice_index,
+            aggregate_public_key.clone(),
+            0,
+            cycle_id + 1,
+        ),
+        // Bob casts a vote with the wrong round - should return an invalid round error
+        make_signers_vote_for_aggregate_public_key_value(
```
jcnelson, line 399: Per my above comment about round numbers, this shouldn't be a bug.  A round is simply a unique identifier for a bloc of signers.  It has no other semantic rules.

stackslib/src/clarity_vm/clarity.rs
```diff
@@ -1331,38 +1332,6 @@ impl<'a, 'b> ClarityBlockConnection<'a, 'b> {
             let pox_4_contract_tx =
                 StacksTransaction::new(tx_version.clone(), boot_code_auth.clone(), payload);
 
-            let initialized_agg_key = if !mainnet {
```
jcnelson, line 1334: :raised_hands:

stackslib/src/clarity_vm/clarity.rs
```diff
@@ -1394,47 +1363,6 @@ impl<'a, 'b> ClarityBlockConnection<'a, 'b> {
                     )
                     .expect("Failed to set burnchain parameters in PoX-3 contract");
 
-                // set the aggregate public key for all pre-pox-4 cycles, if in testnet, and can fetch a boot-setting
-                if !mainnet {
```
jcnelson, line 1398: :raised_hands:

stackslib/src/net/tests/mod.rs
```diff
@@ -588,6 +641,29 @@ impl NakamotoBootPlan {
...
 
                             let mut txs = vec![];
+                            // check if the stacker/signers need to vote for an aggregate key. if so, append those transactions
+                            //  to the end of the block.
+                            // NOTE: this will only work the block after .signers is updated, because `make_all_signers_vote...`
+                            //  checks the chainstate as of `tip` to obtain the signer vector. this means that some tests may
+                            //  need to produce an extra block in a tenure in order to get the signer votes in place.
+                            //  The alternative to doing this would be to either manually build the signer vector or to refactor
+                            //  the testpeer such that a callback is provided during the actual mining of the block with a
+                            //  `ClarityBlockConnection`.
```
jcnelson, line 651: This is good.  I'd rather the test framework faithfully simulate what signers ought to be doing (i.e. sending transactions) instead of directly filling data into a block.

stackslib/src/net/tests/mod.rs
```diff
@@ -811,7 +890,10 @@ fn test_boot_nakamoto_peer() {
         NakamotoBootTenure::NoSortition(vec![NakamotoBootStep::Block(vec![next_stx_transfer()])]),
         // prepare phase for 2
         NakamotoBootTenure::NoSortition(vec![NakamotoBootStep::Block(vec![next_stx_transfer()])]),
-        NakamotoBootTenure::Sortition(vec![NakamotoBootStep::Block(vec![next_stx_transfer()])]),
+        NakamotoBootTenure::Sortition(vec![
```
jcnelson, line 893: I'm curious why this was necessary?  It's not a problem; just making sure that having two transactions here isn't masking some underlying issue that arises when there's only one transaction.

testnet/stacks-node/src/config.rs
jcnelson: I'm all in favor of deleting self-signing.  I think if we don't delete it, it's going to be the new Helium node -- a vestigial piece of code we're going to be stuck maintaining until the next major architectural revision because someone somewhere might still be using it as part of their devnet setup.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -433,6 +476,61 @@ pub fn boot_to_epoch_3(
...
+            .unwrap();
+        let prepare_phase_start = btc_regtest_controller
+            .get_burnchain()
+            .pox_constants
+            .prepare_phase_start(
+                btc_regtest_controller.get_burnchain().first_block_height,
+                reward_cycle,
+            );
+
+        // TODO: Check if the vote has already happened
```
jcnelson, line 499: I think you should do this before merging. It's straightforward to do via the read-only-call API.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -433,6 +476,61 @@ pub fn boot_to_epoch_3(
...
+        if block_height >= prepare_phase_start {
+            // If we are self-signing, then we need to vote on the aggregate public key
```
jcnelson, line 501: Hooray! More code to delete if not self-signing

jcnelson (commented):
This PR is in pretty good shape overall, but I do have quite a few show-stopping thoughts about the voting contract that need some more discussion.

jcnelson (changes requested):
Putting a NACK on this temporarily to make sure that the voting contract bugs get fixed.

kantai:
The vote construction code relies on being able to query from the chain tip to read the signer set. So, it will submit a vote one block after the signer set is written. This doesn’t need to be a subsequent tenure, but it does need to be a subsequent block. Because of the way the Sortition/NoSortition tenures are structured in this test, there’s only one tenure during the prepare phase (aside from the 0 block which already needs the aggregate key to have been approved).On Feb 20, 2024, at 4:24 PM, Jude Nelson ***@***.***> wrote:﻿
@jcnelson commented on this pull request.



In stackslib/src/net/tests/mod.rs:
> @@ -811,7 +890,10 @@ fn test_boot_nakamoto_peer() {
         NakamotoBootTenure::NoSortition(vec![NakamotoBootStep::Block(vec![next_stx_transfer()])]),
         // prepare phase for 2
         NakamotoBootTenure::NoSortition(vec![NakamotoBootStep::Block(vec![next_stx_transfer()])]),
-        NakamotoBootTenure::Sortition(vec![NakamotoBootStep::Block(vec![next_stx_transfer()])]),
+        NakamotoBootTenure::Sortition(vec![

I'm curious why this was necessary?  It's not a problem; just making sure that having two transactions here isn't masking some underlying issue that arises when there's only one transaction.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
kantai: I think this is obscured by the `match/assert/and` logic in `update-last-round`.

If `round` is <= `last_round + 1`, then that function returns `(ok changed?)` where `changed?` is whether or not `last_round` got updated. The `try!` here only returns if there was an error, which only happens if the round is `> last_round + 1`. So, I think the desired behavior is here -- votes can accumulate for any round. Each signer can only vote once per round. The first key that crosses the threshold, regardless of round, is set as the key.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: `get-or-set` is a little confusing too. How about `get-and-cache-total-weight`?

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: I fixed that original issue in f596c0bf5b0c463859b2a25d0322f40a6efd8c0d.

We talked about the fact that we may at some point want to allow votes for cycle N to happen during cycle N, in which case, we won't want that new error check.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: `principal-construct?` takes in a `(buff 1)` and a `(buff 20)`, but we have a `(buff 33)`?

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: I slightly prefer the `and`, but it shouldn't be a huge difference in cost, and after writing the `if` version, I do see that it could be better for readability.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: Refactored in 252a98cc9

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: Refactored in 252a98cc9

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: Renamed in 252a98cc9

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: Added in 252a98cc9

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: Renamed in 252a98cc9

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: Added a test for this in 712c77e05

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: Sure. Added in ac80a889a

stackslib/src/chainstate/stacks/boot/signers-voting.clar
obycode: Yeah, I was hoping to simplify it by letting that error indicate both cases. If that's confusing, I can add a separate check for this case.

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
obycode: This comment has been updated.

stackslib/src/chainstate/stacks/boot/signers_voting_tests.rs
obycode: @jferrant and I discussed this and decided to make it an error, since it should be unexpected behavior for a signer to submit a vote in a round that is more than 1 above the highest round.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
obycode: Added in 91a1fd95e.

stackslib/src/chainstate/stacks/boot/signers-voting.clar
jcnelson: Ah, my mistake. I was thinking of something else.

stackslib/src/net/tests/mod.rs
jcnelson: > Because of the way the Sortition/NoSortition tenures are structured in this test, there’s only one tenure during the prepare phase (aside from the 0 block which already needs the aggregate key to have been approved).

I think this has less to do with the boot plan and more to do with the PoX parameters passed into `.with_pox_constants()`.

jcnelson (approved):
LGTM; just make sure CI passes after resolving any merge conflicts.  Thanks!

stackslib/src/net/tests/mod.rs
kantai: Yes -- it's both. If the prepare phase were longer, then there would be more sortitions in the prepare phase. But it's the combination of prepare phase length == 3 and the `NoSortition, NoSortition, Sortition` sequence that requires this behavior.

obycode:
Merged `next` and resolved conflicts in 055fad364.

kantai (dismissed):

obycode:
Merged again and waiting for test results.

jcnelson (approved):

PR #4364, opened 2024-02-10
chore: Drop support for old x86_64 CPUs

jbencin:
### Description

When building for x86_64 CPUs, use the `x86-64-v3` target. This allows the compiler to use newer instructions in order to better optimize the code, like AVX, AVX2, and FMA. This also means older (pre-Haswell or pre-Bulldozer) CPUs won't be supported by default. See [here](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) for full details.

This would mean than users with 10+ year old CPUs would not be able to use the binaries we distribute, and in order to run `stacks-node` they would have to modify the config and build it themselves

If we are not comfortable with this trade-off, we could build and distribute both `x86-64` and `x86-64-v3` binaries

This gives a small performance boost of about **1.4%** on my Laptop (Intel Core i7-1270P) when processing the first 3000 blocks:

```console
❯ hyperfine -w 3 -r 10 "stacks-core.next/target/release/stacks-inspect.x86-64 replay-block /home/jbencin/data/next/ first 3000" "stacks-core.next/target/release/stacks-inspect.x86-64-v3 replay-block /home/jbencin/data/next/ first 3000"
Benchmark 1: stacks-core.next/target/release/stacks-inspect.x86-64 replay-block /home/jbencin/data/next/ first 3000
  Time (mean ± σ):     31.088 s ±  0.041 s    [User: 27.082 s, System: 3.844 s]
  Range (min … max):   31.024 s … 31.152 s    10 runs
 
Benchmark 2: stacks-core.next/target/release/stacks-inspect.x86-64-v3 replay-block /home/jbencin/data/next/ first 3000
  Time (mean ± σ):     30.645 s ±  0.046 s    [User: 26.541 s, System: 3.946 s]
  Range (min … max):   30.568 s … 30.719 s    10 runs
 
Summary
  stacks-core.next/target/release/stacks-inspect.x86-64-v3 replay-block /home/jbencin/data/next/ first 3000 ran
    1.01 ± 0.00 times faster than stacks-core.next/target/release/stacks-inspect.x86-64 replay-block /home/jbencin/data/next/ first 3000

```

### Applicable issues

- #4316

### Checklist

- [ ] Changelog is updated

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -2,9 +2,11 @@
 stacks-node = "run --package stacks-node --"
 fmt-stacks = "fmt -- --config group_imports=StdExternalCrate,imports_granularity=Module"
 
-# Build for modern x86_64 CPUs to take advantage of newer instructions
+# For x86_64 CPUs, default to `native` and override in CI for release builds
+# This makes it slightly faster for users running locally built binaries
+# TODO: Same for other targets?
 [target.'cfg(all(target_arch = "x86_64"))']
-rustflags = ["-Ctarget-cpu=x86-64-v3"]
+rustflags = ["-Ctarget-cpu=native"]
 
 # Needed by perf to generate flamegraphs.
 #[target.x86_64-unknown-linux-gnu]
--- a/.cargo/config
+++ b/.cargo/config
@@ -2,7 +2,12 @@
 stacks-node = "run --package stacks-node --"
 fmt-stacks = "fmt -- --config group_imports=StdExternalCrate,imports_granularity=Module"
 
+# Build for modern x86_64 CPUs to take advantage of newer instructions
+[target.'cfg(all(target_arch = "x86_64"))']
+rustflags = ["-Ctarget-cpu=x86-64-v3"]
+
 # Needed by perf to generate flamegraphs.
 #[target.x86_64-unknown-linux-gnu]
 #linker = "/usr/bin/clang"
 #rustflags = ["-Clink-arg=-fuse-ld=lld", "-Clink-arg=-Wl,--no-rosegment"]
+
--- a/.github/workflows/create-source-binary-x64.yml
+++ b/.github/workflows/create-source-binary-x64.yml
@@ -1,6 +1,6 @@
 ## Github workflow to create multiarch binaries from source
 
-name: Create Binaries
+name: Create Binaries for x86_64
 
 on:
   workflow_call:
@@ -14,7 +14,13 @@ on:
         required: false
         type: string
         default: >-
-          ["linux-glibc-arm64", "linux-glibc-armv7", "linux-musl-arm64", "linux-musl-armv7"]
+          ["linux-glibc-x64", "linux-musl-x64", "macos-x64", "windows-x64"]
+      cpu:
+        description: "Stringified JSON object listing of target CPU matrix"
+        required: false
+        type: string
+        default: >-
+          ["x86-64", "x86-64-v3"]
 
 ## change the display name to the tag being built
 run-name: ${{ inputs.tag }}
@@ -39,13 +45,14 @@ jobs:
       max-parallel: 10
       matrix:
         platform: ${{ fromJson(inputs.arch) }}
+        cpu: ${{ fromJson(inputs.cpu) }}
     steps:
       ## Setup Docker for the builds
       - name: Docker setup
         uses: stacks-network/actions/docker@main
 
       ## Build the binaries using defined dockerfiles
-      - name: Build Binary (${{ matrix.platform }})
+      - name: Build Binary (${{ matrix.platform }}_${{ matrix.cpu }})
         id: build_binaries
         uses: docker/build-push-action@0565240e2d4ab88bba5387d719585280857ece09 # 5.0.0
         with:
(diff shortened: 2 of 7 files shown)


wileyj:
i'm not a huge fan of this: `they would have to modify the config and build it themselves`. 
it's simple for us to add that rustflag at binary build time (i do agree that we should only build v3 binaries) - do we gain anything by making the change here vs how we build the binary archive for release (which is then used in docker images)?

my concern is that we're adding a barrier to entry for someone wanting to compile on old hardware vs only providing a binary targeting newer chipset(s).

jbencin:
> it's simple for us to add that rustflag at binary build time

Yeah this is a better idea

> or hardcode the env var on line

This is a separate arg from the target triplet

wileyj:
> > it's simple for us to add that rustflag at binary build time
> 
> Yeah this is a better idea
> 
> > or hardcode the env var on line
> 
> This is a separate arg from the target triplet

yes, just highlighting where in the file it could be added as a new line. not suggesting we change the target triple

jbencin:
We would have to run `cargo` like this when building a release:

```sh
RUSTFLAGS='-C target-cpu=x86-64-v3' cargo build ...
```

Optionally, we can put this in Cargo.toml also (those settings are overriden by env variables)

```toml
rustflags = ["-Ctarget-cpu=native"]
```

So the default is to optimize for the user's machine. We could do this just for x86_64, or for all architectures, but then we'd have to override it in every release build

We could do this for dev builds also, so tests will run marginally faster

wileyj:
> We would have to run `cargo` like this when building a release:
> 
> ```shell
> RUSTFLAGS='-C target-cpu=x86-64-v3' cargo build ...
> ```
> 
> Optionally, we can put this in Cargo.toml also (those settings are overriden by env variables)
> 
> ```toml
> rustflags = ["-Ctarget-cpu=native"]
> ```
> 
> So the default is to optimize for the user's machine. We could do this just for x86_64, or for all architectures, but then we'd have to override it in every release build
> 
> We could do this for dev builds also, so tests will run marginally faster

interesting idea, i do like this method better because it still allows anyone to build on whatever old CPU they want to use and with minimal changes we can target the faster binaries with builds we need to do and provide. 

the changes to how we build are easy enough to implement, but i'm quite interested in the testing aspect. faster tests is always a good thing.

wileyj:
@jbencin in light of your last reply, i think you'd be interested in this PR (i've added you as a reviewer)
https://github.com/stacks-network/stacks-core/pull/4368

jbencin:
I found some high quality data on x86_64 instruction support from Valve [here](https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam) (click on "Other settings")

According to this, >99% of Steam users have support for all SSE instructions, but AVX is at 96.59% support, and AVX2 is at 92.33%. Since the `x86-64-v3` target allows AVX2 instructions, and there is still a single digit percentage of users who can't run these, there may be a small number of users who have to build `stacks-node` themselves

jbencin:
I tried this benchmark again on a block range which should be relatively more CPU-intensive, and it looks to be about a **2%** performance gain:

```console
❯ hyperfine -w 3 -r 10 "stacks-core/target/release/stacks-inspect.next replay-block /home/jbencin/data/next/ range 99990 100000" "stacks-core/target/release/stacks-inspect.x86-64-v3 replay-block /home/jbencin/data/next/ range 99990 100000"
  Time (mean ± σ):     12.158 s ±  0.026 s    [User: 11.645 s, System: 0.461 s]
  Range (min … max):   12.132 s … 12.205 s    10 runs
 
Benchmark 2: stacks-core/target/release/stacks-inspect.x86-64-v3 replay-block /home/jbencin/data/next/ range 99990 100000
  Time (mean ± σ):     11.898 s ±  0.026 s    [User: 11.373 s, System: 0.476 s]
  Range (min … max):   11.862 s … 11.935 s    10 runs
 
Summary
  stacks-core/target/release/stacks-inspect.x86-64-v3 replay-block /home/jbencin/data/next/ range 99990 100000 ran
    1.02 ± 0.00 times faster than stacks-core/target/release/stacks-inspect.next replay-block /home/jbencin/data/next/ range 99990 100000
```

jbencin:
@wileyj I am trying to make it so that we build for `--target-cpu=native` for local builds, and create release builds for both `x86-64` and `x86-64-v3`. I'm not that familiar Github Actions, can you let me know if I'm approaching this right?

wileyj:
looking at the ci changes, i think we can DRY this a bit more vs copy/pasting the same workflows. I'll take a closer look today. 
we also have that related PR to remove the build-script dockerfiles, so there is a potential conflict (just a heads up) that we'll have to decide on.

obycode (approved):
LGTM!

wileyj:
i'm not a fan of the workflows here, but since they're only used on a release build - we can target those changes in a new issue. i think we can DRY up your ci proposal a bit @jbencin , but that shouldn't prevent this from merging to `next`. 

https://github.com/stacks-network/stacks-core/issues/4383

wileyj (approved):
shipit

jbencin:
Hey @wileyj, I'm hesitant to merge this without testing the the "Create Release" step. Any idea how to do this without actually releasing. Also, how do I get the artifacts from this CI run to verify that they are built as expected?

wileyj:
> Hey @wileyj, I'm hesitant to merge this without testing the the "Create Release" step. Any idea how to do this without actually releasing. Also, how do I get the artifacts from this CI run to verify that they are built as expected?

not possible on this repo i'm afraid, because those workflows only trigger on a tagged release building the default branch. 
i can try pulling these changes into a fork and building

jbencin:
> i can try pulling these changes into a fork and building

@wileyj yeah if you could do that such that we don't accidentally release anything we don't want to DockerHub or elsewhere, that would be great. But if you think this looks good and we're not going to have issues with the release step, I'll merge it. I trust your judgement here, this is not my area of expertise.

wileyj:
> > i can try pulling these changes into a fork and building
> 
> @wileyj yeah if you could do that such that we don't accidentally release anything we don't want to DockerHub or elsewhere, that would be great. But if you think this looks good and we're not going to have issues with the release step, I'll merge it. I trust your judgement here, this is not my area of expertise.

👍 for `next`, there's very little concern - by the time it reaches `master` we'll have to scrutinize it a little more though. 
i can run tests on a fork in the meantime, pulling in these changes and see how the releases work - i'm not too worried about it, but it does raise a question we should answer of how to test a release (packages/binaries etc).

PR #4366, opened 2024-02-11
Nakamoto Inventory State Machine

### LINKED ISSUE #4062: [Nakamoto] Block inventory sync network state machine
There needs to be a Nakamoto-specific network state machine for synchronizing block inventories.  Per the SIP, the inventory bit-vectors exchanged are framed by reward cycle, and there exists a 1-bit in the ith slot in the vector if there was a tenure-change in that sortition, and a 0-bit if not.  As before, nodes need to synchronize the entirety of their neighbors' block inventory vectors on boot-up, and then only need to track the last two reward cycles' inventories when out of IBD.  Unlike before, there is no need to separately synchronize the PoX bit vector since Nakamoto cannot fork.

jcnelson:
Closes #4062.

This PR implements the Nakamoto tenure inventory state machine, and updates the top-level network work state machine to drive it.

The Stacks blockchain differs from most blockchains in that all peers already know the list of block hashes from the same source of truth -- the sortition history in the Bitcoin blockchain.  Instead of exchanging lists of block hashes, nodes exchange _inventory bitvectors_ to succinctly describe which blocks (or tenures) they have within a given reward cycle's worth of sortitions.  The current system performs this in two steps:  nodes first exchange PoX bitvectors describing which PoX anchor blocks they have, and from there, they exchange reward cycle inventories for each reward cycle in which they agree on the state of the PoX anchor block.

The Nakamoto inventory state machine is simpler, since we assume that all nodes always have all PoX anchor blocks.  Nodes simply request and reply tenure inventories for each reward cycle.

This PR contains some refactoring work in epoch 2.x code paths.  To facilitate reviews, I've annotated where this is.


stackslib/src/net/inv/epoch2x.rs
```diff
@@ -2623,6 +2623,216 @@ impl PeerNetwork {
             }
         }
     }
+
+    /// Update the state of our neighbors' epoch 2.x block inventories.
+    /// Return (finished?, throttled?)
+    fn do_network_inv_sync_epoch2x(&mut self, sortdb: &SortitionDB, ibd: bool) -> (bool, bool) {
```
jcnelson, line 2629: This used to live in `src/net/p2p.rs`.

stackslib/src/net/inv/epoch2x.rs
```diff
@@ -2623,6 +2623,216 @@ impl PeerNetwork {
...
+                finished_always_allowed_inv_sync = true;
+            }
+        }
+
+        finished_always_allowed_inv_sync
+    }
+
+    /// Do an inventory state machine pass for epoch 2.x.
+    /// Returns the new work state  
+    pub fn work_inv_sync_epoch2x(
```
jcnelson, line 2737: This used to be part of `do_network_work()` in `src/net/p2p.rs`.

stackslib/src/net/inv/epoch2x.rs
```diff
@@ -2623,6 +2623,216 @@ impl PeerNetwork {
...
+        // disconnect from dead connections
+        for dead in dead_neighbors.into_iter() {
+            self.deregister_neighbor(&dead);
+        }
+
+        (done, throttled)
+    }
+
+    /// Check to see if an always-allowed peer has performed an epoch 2.x inventory sync
+    fn check_always_allowed_peer_inv_sync_epoch2x(&self) -> bool {
```
jcnelson, line 2662: This used to be part of `do_network_work()` in `src/net/p2p.rs`.

stackslib/src/net/inv/nakamoto.rs
```diff
@@ -233,3 +242,569 @@ impl InvGenerator {
...
+    }
+
+    /// Highest reward cycle learned
+    pub fn highest_reward_cycle(&self) -> u64 {
+        let mut highest_rc = 0;
+        for (_, inv) in self.inventories.iter() {
+            highest_rc = inv.highest_reward_cycle().max(highest_rc);
+        }
+        highest_rc
+    }
```
kantai, line 521: Is this method used anywhere or is it only used in tests?

stackslib/src/net/inv/nakamoto.rs
```diff
@@ -233,3 +242,569 @@ impl InvGenerator {
...
+        let sn = SortitionDB::get_canonical_burn_chain_tip(sortdb.conn())?;
+        let tip_rc = sortdb
+            .pox_constants
+            .block_height_to_reward_cycle(sortdb.first_block_height, sn.block_height)
+            .expect("FATAL: snapshot occurred before system start");
+
+        for rc in highest_rc..=tip_rc {
+            if self.reward_cycle_consensus_hashes.contains_key(&rc) {
+                continue;
+            }
```
kantai, line 561:
```suggestion
        for rc in (highest_rc+1)..=tip_rc {
```

stackslib/src/net/inv/nakamoto.rs
```diff
@@ -233,3 +242,569 @@ impl InvGenerator {
...
+            let Some(ch) = Self::load_consensus_hash_for_reward_cycle(sortdb, rc)? else {
+                // NOTE: this should be unreachable, but don't panic
+                continue;
+            };
```
kantai, line 565: Should this error instead? This would make the inventory sync somewhat incoherent, right?

stackslib/src/net/inv/nakamoto.rs
```diff
@@ -233,3 +242,569 @@ impl InvGenerator {
...
+
+            // ask this neighbor for its inventory
+            if let Some(getnakamotoinv) = self.make_getnakamotoinv(inv_rc) {
+                if let Err(e) = self.comms.neighbor_send(network, &naddr, getnakamotoinv) {
+                    warn!("{:?}: failed to send GetNakamotoInv", network.get_local_peer();
+                          "peer" => ?naddr,
+                          "error" => ?e
+                    );
+                }
+            }
```
kantai, line 662:
```suggestion
            let Some(getnakamotoinv) = self.make_getnakamotoinv(inv_rc) else {
                warn!("{:?}: failed to make nakamoto inv, could not find consensus hash for reward cycle.";
                           "reward_cycle" => inv_rc);
                continue;
            };
            if let Err(e) = self.comms.neighbor_send(network, &naddr, getnakamotoinv) {
                    warn!("{:?}: failed to send GetNakamotoInv", network.get_local_peer();
                          "peer" => ?naddr,
                          "error" => ?e
                    );
            }
```

stackslib/src/net/p2p.rs
```diff
@@ -3874,6 +3865,145 @@ impl PeerNetwork {
...
+        let cur_epoch = self.get_current_epoch();
+        let prune = if cur_epoch.epoch_id >= StacksEpochId::Epoch30 {
+            debug!("{:?}: run Nakamoto work loop", self.get_local_peer());
+
+            // in Nakamoto epoch, so do Nakamoto things
+            let prune = self.do_network_work_nakamoto(sortdb, ibd);
+
+            // in Nakamoto epoch, but we might still be doing epoch 2.x things since Nakamoto does
+            // not begin on a reawrd cycle boundary.
+            if self.burnchain_tip.block_height <= cur_epoch.start_height
```
kantai, line 3878: Should this be explicitly the `Epoch30` boundary? Presumably, a hypothetical `Epoch31` wouldn't want to do Epoch 2.x work during `Epoch30`

stackslib/src/net/p2p.rs
```diff
@@ -3874,6 +3865,145 @@ impl PeerNetwork {
...
+            let learned = self.do_network_inv_sync_nakamoto(sortdb, ibd);
+            if learned {
+                debug!("{:?}: learned about new blocks!", self.get_local_peer());
+            }
+
+            debug!(
+                "{:?}: network work state is {:?}",
+                self.get_local_peer(),
+                &self.nakamoto_work_state
+            );
```
kantai, line 3936:
```suggestion
            debug!(
                "{:?}: network work state is {:?}",
                self.get_local_peer(),
                &self.nakamoto_work_state;
                "learned_new_blocks?" => learned
            );
```

kantai (commented):
This LGTM, have some superficial comments throughout, but a few questions that we should clear up before approval.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -9504,8 +9504,10 @@ fn test_problematic_blocks_are_not_relayed_or_stored() {
 
     let tip_info = get_chain_info(&conf);
 
-    // all blocks were processed
-    assert!(tip_info.stacks_tip_height >= old_tip_info.stacks_tip_height + 5);
+    // at least one block was mined (hard to say how many due to the raciness between the burnchain
+    // downloader and this thread).
+    assert!(tip_info.stacks_tip_height > old_tip_info.stacks_tip_height);
+
```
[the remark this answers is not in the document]
jcnelson: There's a race condition in `next` at least (and probably also `develop` and `master`) whereby the burnchain block downloader thread races the miner thread, such that a block-commit generated by the miner thread is not mined in the target block.  This, in turn, causes a sortition to be missed, which in turn causes this assertion to fail since the Stacks height would have advanced by 4 blocks instead of 5.  However, the nature of this test is to verify that a block with a problematic transaction gets mined (which happens below); it doesn't matter if it takes 4 or 5 attempts.  This check was needlessly strict.

I fixed it here because I earlier approved a PR to `next` where this test was failing for the same reason, and I told the author not to worry about it since it wasn't blocking the merge (and the code changes were unrelated).

stackslib/src/net/inv/nakamoto.rs
jbencin: Slightly clearer way to write this logic:

```suggestion
    pub fn highest_reward_cycle(&self) -> u64 {
        self.inventories.iter()
            .map(|(_, inv)| inv.highest_reward_cycle())
            .max()
            .unwrap_or(0)
    }
```

stackslib/src/net/connection.rs
```diff
@@ -411,13 +411,14 @@ pub struct ConnectionOptions {
     pub disable_inbound_handshakes: bool,
     pub disable_stackerdb_get_chunks: bool,
     pub force_disconnect_interval: Option<u64>,
+    pub force_nakamoto_epoch_transition: bool,
```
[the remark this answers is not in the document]
jbencin: +1, `rustdoc` comments on struct fields really help the newer developers like myself

stackslib/src/net/p2p.rs
```diff
@@ -1731,27 +1741,36 @@ impl PeerNetwork {
...
-        let mut nk_remove: Vec<NeighborKey> = vec![];
+        let mut nk_remove: Vec<(NeighborKey, Hash160)> = vec![];
         for (neighbor_key, ev_id) in self.events.iter() {
             if *ev_id == event_id {
-                nk_remove.push(neighbor_key.clone());
+                let pubkh = if let Some(convo) = self.get_p2p_convo(event_id) {
+                    convo.get_public_key_hash().unwrap_or(Hash160([0x00; 20]))
+                } else {
+                    Hash160([0x00; 20])
+                };
```
jbencin, line 1751:
```suggestion
                let pubkh = self.get_p2p_convo(event_id)
                    .and_then(|convo| convo.get_public_key_hash())
                    .unwrap_or(Hash160([0x00; 20]));
```

stackslib/src/net/p2p.rs
```diff
@@ -3874,6 +3865,145 @@ impl PeerNetwork {
...
+    ) -> bool {
+        let cur_epoch = self.get_current_epoch();
+        let prune = if cur_epoch.epoch_id >= StacksEpochId::Epoch30 {
+            debug!("{:?}: run Nakamoto work loop", self.get_local_peer());
+
+            // in Nakamoto epoch, so do Nakamoto things
+            let prune = self.do_network_work_nakamoto(sortdb, ibd);
+
+            // in Nakamoto epoch, but we might still be doing epoch 2.x things since Nakamoto does
+            // not begin on a reawrd cycle boundary.
```
jbencin, line 3877:
```suggestion
            // not begin on a reward cycle boundary.
```

stackslib/src/net/tests/inv/nakamoto.rs
```diff
@@ -434,6 +451,50 @@ fn check_inv_messages(
...
+            let msg_bit = if bit / 8 >= tenure_inv.len() {
+                // only allowed at the end
+                debug!(
+                    "bit = {}, tenure_rc = {}, tenure_inv = {:?}",
+                    bit, tenure_rc, &tenure_inv
+                );
+                assert_eq!(i, inv_state.tenures_inv.len() - 1);
+                false
+            } else {
+                tenure_inv[bit / 8] & (1 << (bit % 8)) != 0
```
jbencin, line 471: This looks dangerous. As Aaron mentioned earlier, can we use either `stacks_common::BitVec` or the `bitvec` crate (it's already in Cargo.lock)

jbencin (commented):
Things to address before I'd approve:
- Some comments need to be fixed
- The manual `BitVec` implementation in `check_inv_state()` looks risky. Can we use an existing and tested implementation?

Other than that, I left some style/readability suggestions, but I won't insist on making those changes

stackslib/src/net/inv/epoch2x.rs
```diff
@@ -2623,6 +2623,216 @@ impl PeerNetwork {
...
+                .collect();
+
+        // have we finished a full pass of the inventory state machine on an
+        // always-allowed peer?
+        let mut finished_always_allowed_inv_sync = false;
+
+        if always_allowed.len() == 0 {
+            // vacuously, we have done so
+            finished_always_allowed_inv_sync = true;
+        } else {
```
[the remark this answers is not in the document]
jcnelson: Sure; this was always in the epoch 2.x state machine, but was simply moved over to a new file.

stackslib/src/net/inv/epoch2x.rs
```diff
@@ -2623,6 +2623,216 @@ impl PeerNetwork {
...
+            // do we have an always-allowed peer that we have not fully synced
+            // with?
+            let mut have_unsynced = false;
+            if let Some(ref inv_state) = self.inv_state {
```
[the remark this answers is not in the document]
jcnelson: This was also just lifted directly from what was running in epoch 2.x

stackslib/src/net/inv/epoch2x.rs
```diff
@@ -2623,6 +2623,216 @@ impl PeerNetwork {
...
+        if inv_done {
+            if !download_backpressure {
+                // proceed to get blocks, if we're not backpressured
+                work_state = PeerNetworkWorkState::BlockDownload;
+            } else {
+                // skip downloads for now
+                work_state = PeerNetworkWorkState::Prune;
+            }
+
+            if !inv_throttled {
```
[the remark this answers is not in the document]
jcnelson: Sure

stackslib/src/net/inv/nakamoto.rs
```diff
@@ -233,3 +242,569 @@ impl InvGenerator {
...
+        let rc_height = sortition_height % self.reward_cycle_len;
+
+        let idx =
+            usize::try_from(rc_height / 8).expect("FATAL: reward cycle length exceeds host usize");
+        let bit = rc_height % 8;
+
+        rc_tenures
+            .get(idx)
+            .map(|bits| bits & (1 << bit) != 0)
+            .unwrap_or(false)
```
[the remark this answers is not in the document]
jcnelson: Your `BitVec` merged after I wrote this.

stackslib/src/net/inv/epoch2x.rs
```diff
@@ -2623,6 +2623,216 @@ impl PeerNetwork {
...
+        &mut self,
+        sortdb: &SortitionDB,
+        download_backpressure: bool,
+        ibd: bool,
+    ) -> PeerNetworkWorkState {
+        let mut work_state = PeerNetworkWorkState::BlockInvSync;
+
+        // synchronize epcoh 2.x peer block inventories
+        let (inv_done, inv_throttled) = self.do_network_inv_sync_epoch2x(sortdb, ibd);
+        if inv_done {
```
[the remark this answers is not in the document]
jcnelson: Sure; again, this is just what was already there.  Only mentioning it because I'm trying to save you and @jbencin the time of having to review soon-to-be-legacy code that's been in production for years already ;)

stackslib/src/net/inv/nakamoto.rs
```diff
@@ -233,3 +242,569 @@ impl InvGenerator {
...
+    /// Add in a newly-discovered inventory.
+    /// NOTE: inventories are supposed to be aligned to the reward cycle
+    /// Returns true if we learned about at least one new tenure-start block
+    /// Returns false if not.
+    pub fn merge_tenure_inv(
+        &mut self,
+        tenure_inv: Vec<u8>,
+        tenure_bitlen: u16,
+        reward_cycle: u64,
+    ) -> bool {
```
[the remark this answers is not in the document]
jcnelson: 1. Yes, this is buggy; will fix.  `self.highest_sortition` is the number of sortitions represented by this state machine.  It should be equal to (reward_cycle_len * highest_queried_rc + bitlen_of_highest_queried_rc)

2. The bitvectors themselves are not merged because `tenure_inv` represents the _latest_ knowledge of blocks for the given reward cycle.  If a bit in `tenure_inv` has flipped from 1 to 0 relative to `cur_inv`, or if `tenure_inv` is shorter than `cur_inv`, then that's still pertinent information, since the block downloader ultimately consumes this information to determine which tenures to fetch from this peer.  If a tenure somehow became unavailable on the remote peer, then the block downloader will need to know that so it won't go and request it.

stackslib/src/net/inv/nakamoto.rs
jcnelson: It will be needed by the block downloader.

stackslib/src/net/inv/nakamoto.rs
jcnelson: Sure

stackslib/src/net/inv/nakamoto.rs
jcnelson: Sure

stackslib/src/net/inv/nakamoto.rs
jcnelson: Sure

stackslib/src/net/p2p.rs
jcnelson: Yes; I'll constrain this check

stackslib/src/net/p2p.rs
jcnelson: Sure

stackslib/src/net/inv/nakamoto.rs
jcnelson: Actually, this is necessary -- tests fail with out it, since we need to populate the consensus hash for `highest_rc` on the first pass.

kantai: Is that because `highest_rc = 0` could mean either that the 0th reward cycle is the highest one processed or that no reward cycle has been processed?

stacks-common/src/bitvec.rs
```diff
@@ -162,6 +162,13 @@ impl<const MAX_SIZE: u16> BitVec<MAX_SIZE> {
         }
         Ok(())
     }
+
+    /// Set all bits to zero
+    pub fn clear(&mut self) {
```
jbencin, line 167: Shouldn't this function be:

```rust
self.data.clear();
self.len = 0;
```

stackslib/src/net/codec.rs
```diff
@@ -276,7 +277,17 @@ impl BlocksInvData {
     }
 
     pub fn compress_bools(bits: &Vec<bool>) -> Vec<u8> {
```
jbencin, line 279: In general it's preferable to use slices rather than `&Vec` for function parameters

```suggestion
    pub fn compress_bools(bits: &[bool]) -> Vec<u8> {
```

Also, this function looks like it's reimplementing `BitVec::try_from::<bool>()`

jbencin (approved):
Looks good, just one minor comment

stackslib/src/net/inv/nakamoto.rs
```diff
@@ -233,3 +243,559 @@ impl InvGenerator {
...
+        self.tenures_inv.insert(reward_cycle, tenure_inv);
+        self.last_updated_at = get_epoch_time_secs();
+        learned
+    }
+
+    /// Adjust the next reward cycle to query.
+    /// Returns the reward cycle to query.
+    pub fn next_reward_cycle(&mut self) -> u64 {
+        let query_rc = self.cur_reward_cycle;
+        self.cur_reward_cycle += 1;
```
jferrant, line 355: Just a nit: we are not particularly consistent with our +1 behaviour. Would we expect this to wrapping add? saturating add? I assume wrapping add. But I tend to avoid straight += 1

jferrant (approved):
Don't think my review is particularly helpful on this as I don't entirely understand what is happening throughout, but LGTM

stackslib/src/net/codec.rs
jcnelson: This is part of a legacy code path that's live in production, which I'm _very_ keen not to touch

stackslib/src/net/inv/nakamoto.rs
jcnelson: Thanks! I'll make this a saturating add

jbencin: Just FYI you can wrap any integer type with `std::num::Saturating` ([link](https://doc.rust-lang.org/std/num/struct.Saturating.html)) to force consistent use of saturating arithmetic operations

stackslib/src/net/inv/epoch2x.rs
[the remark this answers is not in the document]
kantai: I'm aware of that, but I see no reason not to apply some simple cleanup during the move.

stackslib/src/net/inv/epoch2x.rs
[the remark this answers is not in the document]
kantai: Yes, but I think it's a good opportunity to do simple cleanup as long as the behaviors are preserved.

kantai (approved):
LGTM!

PR #4372, opened 2024-02-13
Fix(Nakamoto): Block evaluation issues in `next`

kantai:
### Description

This fixes two issues in `next` which result in changes to the evaluation of stacks blocks prior to Epoch 2.5 (i.e., the first hardfork epoch of `next`).

The first issue was relatively straightforward and is solved by 999fe9bd29cde5a3b248e2ece1b2511bcfaa1731. The automatic unlocks for `pox-3` lockups in preparation for `pox-4` must be applied on `Epoch25`, _not_ `Epoch24`.

The second issue was more difficult to diagnose because its a little more subtle (and therefore scarier!) but it is solved by 
6f0936a69f55918bd93a87e132d743a36902ea95. PoX event synthesis, introduced for `pox-2` and `pox-3`, assesses a cost in the block limit. This means that any changes to those functions is consensus-critical. I don't think that this is a great position to be in with those synthesis functions, because one of the reasons they're handled specially to begin with is to try to avoid this. So to address this, I copied the current event synthesis code for 2.4 into a new module `events_24`, and epoch-switch in `events`. For Epochs `>= 2.5`, this function evaluates the synthesis code without assessing costs so that this behavior can be changed going forward.

### DIFF
--- a/clarity/src/vm/database/clarity_db.rs
+++ b/clarity/src/vm/database/clarity_db.rs
@@ -916,7 +916,7 @@ impl<'a> ClarityDatabase<'a> {
     /// Return the height for PoX v3 -> v4 auto unlocks
     ///   from the burn state db
     pub fn get_v3_unlock_height(&mut self) -> Result<u32> {
-        if self.get_clarity_epoch_version()? >= StacksEpochId::Epoch24 {
+        if self.get_clarity_epoch_version()? >= StacksEpochId::Epoch25 {
             Ok(self.burn_state_db.get_v3_unlock_height())
         } else {
             Ok(u32::MAX)
--- a/pox-locking/src/events_24.rs
+++ b/pox-locking/src/events_24.rs
@@ -16,30 +16,24 @@
 
 use clarity::vm::ast::ASTRules;
 use clarity::vm::contexts::GlobalContext;
-use clarity::vm::costs::LimitedCostTracker;
 use clarity::vm::errors::Error as ClarityError;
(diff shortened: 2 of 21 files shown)


pox-locking/src/events.rs
```diff
@@ -419,6 +423,51 @@ pub fn synthesize_pox_event_info(
     function_name: &str,
     args: &[Value],
     response: &ResponseData,
+) -> Result<Option<Value>, ClarityError> {
+    // the first thing we do is check the current epoch. In Epochs <= 2.4,
+    //  synthesizing PoX events was an assessed cost, so event generation
+    //  must remain identical.
```
jcnelson, line 429: Wow, what an oversight! The whole point of synthesizing events outside of the contract was so we could _change them_ without doing a hard fork.  Looks like that's not actually possible today?

jcnelson (approved):
Approving, but subject to a `replay-blocks` test that this does indeed remove all cost-overflow and state-root mismatch errors.  Thanks for fixing this @kantai!

EDIT: also subject to CI passing (obvs) ;)

pox-locking/src/events.rs
kantai: Agreed. Yeah, that's not possible today, but would be after 2.5 activates.

jbencin (approved):
I've tested these changes over a large number of blocks, and it seems to fix the consensus bugs without introducing new ones

Issue #3860, opened 2023-08-22
Nakamoto: New DB for Nakamoto Staging Blocks


kantai:
This should be a new SQLite db which contains a table
`StagingBlocks`. This will be used to store new StacksBlocks as they
arrive without needing to block on other block processing.

saralab:
@kantai : This was tackled in https://github.com/stacks-network/stacks-blockchain/pull/3992  iiuc, can we close this?

saralab:
Merged as part of https://github.com/stacks-network/stacks-blockchain/pull/3992

PR #4373, opened 2024-02-13
Separate Nakamoto DB for staging blocks

jcnelson:
This actually fixes https://github.com/stacks-network/stacks-core/issues/3860, which was closed prematurely since it was not addressed by PR #3992.  Doing this now because this is a blocker for the block sync system.

This PR creates a separate DB for holding Nakamoto blocks, separate from the DB that holds processed Nakamoto block headers.  The reason this is necessary is because we can expect writes to occur concurrently on both DBs -- the network stack will create transactions against the staging blocks DB to store blocks it receives, and both the miner and chains-coordinator will create transactions on the headers DB as they create / process blocks.  In Stacks today, these two data live in the same DB, and it regularly leads to DB contention since the chains-coordinator's and miner's block-processing regularly prevents the block downloader from storing downloaded blocks (which slows down IBD and block delivery).

Because latency is much more of a concern in Nakamoto, it's going to be important that the system can open both kinds of transactions concurrently.  This PR makes it possible by storing staging blocks separately from processed block headers.

The most substantial change in this PR is that the act of updating the `processed` flag on a block no longer executes within the same transaction that writes the block header.  However, this shouldn't lead to any concurrency issues because (1) only the chains coordinator thread performs these tasks, and (2) the code to query processed blocks has been updated to read from both the staging blocks DB and headers DB (this won't block ongoing writes, or be blocked by ongoing writes, since we use WAL mode in both DBs)

### DIFF
--- a/stackslib/src/chainstate/nakamoto/mod.rs
+++ b/stackslib/src/chainstate/nakamoto/mod.rs
@@ -80,6 +80,7 @@ use crate::chainstate::nakamoto::tenure::NAKAMOTO_TENURES_SCHEMA;
 use crate::chainstate::stacks::address::PoxAddress;
 use crate::chainstate::stacks::boot::{POX_4_NAME, SIGNERS_UPDATE_STATE};
 use crate::chainstate::stacks::db::{DBConfig as ChainstateConfig, StacksChainState};
+use crate::chainstate::stacks::index::marf::MarfConnection;
 use crate::chainstate::stacks::{
     TenureChangeCause, MINER_BLOCK_CONSENSUS_HASH, MINER_BLOCK_HEADER_HASH,
 };
@@ -1263,6 +1264,15 @@ impl StacksChainState {
         Ok((header_tx, NakamotoStagingBlocksTx(staging_tx)))
     }
 
+    /// Open a connection to the headers DB, and open a tx to the staging DB
+    pub fn headers_conn_and_staging_tx_begin<'a>(
+        &'a mut self,
+    ) -> Result<(&'a rusqlite::Connection, NakamotoStagingBlocksTx<'a>), ChainstateError> {
+        let header_conn = self.state_index.sqlite_conn();
+        let staging_tx = tx_begin_immediate(&mut self.nakamoto_staging_blocks_conn)?;
+        Ok((header_conn, NakamotoStagingBlocksTx(staging_tx)))
+    }
+
     /// Get a ref to the nakamoto staging blocks connection
     pub fn nakamoto_blocks_db(&self) -> NakamotoStagingBlocksConnRef {
         NakamotoStagingBlocksConnRef(&self.nakamoto_staging_blocks_conn)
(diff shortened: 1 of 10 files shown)


kantai (commented):
This LGTM -- just have some pretty superficial comments.

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1427,20 +1645,66 @@ impl NakamotoChainState {
...
+            );
+
+            // as a separate transaction, mark this block as processed and orphaned.
+            // This is done separately so that the staging blocks DB, which receives writes
+            // from the network to store blocks, will be available for writes while a block is
+            // being processed. Therefore, it's *very important* that block-processing happens
+            // within the same, single thread.  Also, it's *very important* that this update
+            // succeeds, since *we have already processed* the block.
+
+            loop {
```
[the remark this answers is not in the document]
jcnelson: Once we commit the block, we _must_ update the processed/orphaned flags in the staging DB.  Failure to do that is either a panic (which leaves the node in a corrupt state) or a loop (which at least gives the operator a chance to fix the underlying issue).

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -1410,7 +1323,17 @@ impl NakamotoChainState {
 
         // attach the block to the chain state and calculate the next chain tip.
         let pox_constants = sort_tx.context.pox_constants.clone();
-        let (receipt, clarity_commit) = match NakamotoChainState::append_block(
+
+        // NOTE: because block status is updated in a separate transaction, we need `chainstate_tx`
```
jferrant, line 1327: haha sneaky sneaky. Very helpful comment.

jferrant (approved):

stackslib/src/chainstate/nakamoto/staging_blocks.rs
```diff
@@ -0,0 +1,396 @@
...
+
+                     -- block data
+                     data BLOB NOT NULL,
+                    
+                     PRIMARY KEY(block_hash,consensus_hash)
+        );"#
+        .into(),
+        r#"CREATE INDEX by_index_block_hash ON nakamoto_staging_blocks(index_block_hash);"#.into()
+    ];
+}
```
kantai, line 77: `Vec::new` and `vec![]` are `const` functions -- meaning they can be invoked on `const` arguments and used in `const` definitions without `lazy_static!`. This hasn't always been a feature, which is partially why we do still have so many uses of `lazy_static!`, but when its easy to avoid, I think we should try to do so:


```suggestion
pub const NAKAMOTO_STAGING_DB_SCHEMA_1: Vec<&'static str> = vec![
      r#"
      -- Table for staging nakamoto blocks
      CREATE TABLE nakamoto_staging_blocks (
                     -- SHA512/256 hash of this block
                     block_hash TEXT NOT NULL,
                     -- the consensus hash of the burnchain block that selected this block's miner's block-commit
                     consensus_hash TEXT NOT NULL,
                     -- the parent index_block_hash
                     parent_block_id TEXT NOT NULL,

                     -- has the burnchain block with this block's `consensus_hash` been processed?
                     burn_attachable INT NOT NULL,
                     -- has this block been processed?
                     processed INT NOT NULL,
                     -- set to 1 if this block can never be attached
                     orphaned INT NOT NULL,

                     height INT NOT NULL,

                     -- used internally -- this is the StacksBlockId of this block's consensus hash and block hash
                     index_block_hash TEXT NOT NULL,
                     -- how long the block was in-flight
                     download_time INT NOT NULL,
                     -- when this block was stored
                     arrival_time INT NOT NULL,
                     -- when this block was processed
                     processed_time INT NOT NULL,

                     -- block data
                     data BLOB NOT NULL,
                    
                     PRIMARY KEY(block_hash,consensus_hash)
        );"#,
        r#"CREATE INDEX by_index_block_hash ON nakamoto_staging_blocks(index_block_hash);"#,
];
```

kantai (approved):
LGTM -- my only comment is that `lazy_static` should be avoided if possible.

stackslib/src/chainstate/nakamoto/staging_blocks.rs
jcnelson: Done

PR #4374, opened 2024-02-13
[CI] fix error in release step that generates checksums

wileyj:
During workflow testing a release in a downstream fork, an unexpected error kept surfacing:
```
Poetry could not find a pyproject.toml file in /github/workspace or its parents
```
https://github.com/wileyj/stacks-blockchain/actions/runs/7878083852/job/21495965471#step:4:11

updating this specific action's commit to the latest release v1.0.1 resolves the immediate issue. 
we'll be moving this workflow out of this repo (possibly refactoring it using bash or something, since it's a simple task that really shouldn't fail due to a python dependency).

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -59,7 +59,7 @@ jobs:
       ## Generate a checksums file to be added to the release page
       - name: Generate Checksums
         id: generate_checksum
-        uses: jmgilman/actions-generate-checksum@24a35957fba81c6cbaefeb1e3d59ee56e3db5077 # v1.0.0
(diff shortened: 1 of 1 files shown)


wileyj:
Note: ci will not run here without a code change, since it's only a small change to a `yml` file. 
that's a different discussion we should have about how and when  to trigger CI (currently, it's configured to not run on changes to markdown or yaml)

jbencin:
> we'll be moving this workflow out of this repo

I was going to suggest this, `sha512sum` available in pretty much every command line environment

jbencin (approved):

jcnelson (approved):

Issue #4272, opened 2024-01-22
[Nakamoto] Vote for aggregate public key on Bitcoin


jcnelson:
There needs to be an on-Bitcoin transaction that calls into voting for an aggregate public key.  This will ensure that even if the Stacks contract-call to vote isn't going through, then Stackers can compel the system to pick them up.

PR #4376, opened 2024-02-14
feat: burn op definition and storage for vote-for-agg-key

hstove:
Based on my notes in [this discussion](https://github.com/stacks-network/stacks-core/discussions/4371), this PR implements a burn op for `vote-for-aggregate-key`.

This PR is currently a work in progress. The code exists for parsing and storing this new burn op, but I've yet to actually process the burn op by making a Clarity tx.

This is built on top of #4360, because that PR has the updated function signature for `vote-for-aggregate-public-key`.

- closes https://github.com/stacks-network/stacks-core/issues/4272


stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -729,6 +763,23 @@ const SORTITION_DB_SCHEMA_8: &'static [&'static str] = &[
...
+        aggregate_key TEXT NOT NULL,
+        round INTEGER NOT NULL,
+        reward_cycle INTEGER NOT NULL,
+        signer_index INTEGER NOT NULL,
+        signer_key TEXT NOT NULL,
+
+        PRIMARY KEY(txid,burn_header_Hash)
+    );"#];
+
 const SORTITION_DB_INDEXES: &'static [&'static str] = &[
```
8marz8, line 783: Maybe create index here too?

stackslib/src/chainstate/burn/operations/vote_for_aggregate_key.rs
```diff
@@ -0,0 +1,384 @@
+// Copyright (C) 2013-2020 Blockstack PBC, a public benefit corporation
+// Copyright (C) 2020 Stacks Open Internet Foundation
```
obycode, line 2:
```suggestion
// Copyright (C) 2020-2024 Stacks Open Internet Foundation
```

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -2214,6 +2234,323 @@ fn stx_delegate_btc_integration_test() {
...
+                &mut spender_signer,
+                1
+            )
+            .is_some(),
+        "Vote for aggregate key operation should submit successfully"
+    );
+
+    info!("Submitted vote for aggregate key op at height {block_height}, mining a few blocks...");
+
+    // the second block should process the vote, after which the balaces should be unchanged
```
obycode, line 2514:
```suggestion
    // the second block should process the vote, after which the balances should be unchanged
```

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -2214,6 +2234,323 @@ fn stx_delegate_btc_integration_test() {
...
+                    .unwrap();
+                assert_eq!(agg_key, aggregate_key.to_hex());
+                let signer_key = vote_obj.get("signer_key").unwrap().as_str().unwrap();
+                assert_eq!(to_hex(&signer_pk_bytes), signer_key);
+
+                vote_for_aggregate_key_found = true;
+            }
+        }
+    }
+    assert!(
```
obycode, line 2545: Would it be possible to add an assertion here to verify that the vote was actually counted? This could be checking for the "voted" event, or it could be a call to the `get-candidate-info` read-only function in signers-voting.

obycode (commented):
This looks great! I just had a couple of comments.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -2214,6 +2234,323 @@ fn stx_delegate_btc_integration_test() {
...
+        reward_cycle.into(),
+        &Pox4SignatureTopic::StackStx,
+        CHAIN_ID_TESTNET,
+        12,
+    )
+    .unwrap();
+
+    let signer_pk_bytes = signer_pk.to_bytes_compressed();
+
+    let stacking_tx = make_contract_call(
```
8marz8, line 2416: I pretty much mirrored my integ tests off of yours and noticed there are some unused vars that can be removed to clean up…also I ran yours locally and saw it works without the stacking transaction…I [removed it ](https://github.com/stacks-network/stacks-core/pull/4412/commits/cd6a10a485935e97be1be5eef2b8f8a9e5e85aba)for mine but not sure if the vote op should pass or fail without it. If it doesn’t depend on it, then maybe you can remove it too for clarity.

hstove: Yeah, the test is really just verifying that the transaction was included, along with the proper params. I added the `stack-stx` call to try and ensure the tx was successful, but we'd need more checks to ensure that.

testnet/stacks-node/src/tests/neon_integrations.rs
hstove: Yep great call - I've added a helper function `get_key_for_cycle` that makes a read-only API call to get the key for a cycle, and I'm verifying that the agg key matches my payload in both `nakamoto_integrations` and `neon_integrations`.

testnet/stacks-node/src/tests/neon_integrations.rs
hstove: I've modified both integration tests to now ensure that the vote transaction was processed successfully. In yours, yeah, you can skip the `stack-stx` tx because that's what your burn op is doing. Now that I've added assertions around the key being set, I do need the `stack-stx` tx to make sure my sender is in the signer set

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -3081,6 +3133,7 @@ impl SortitionDB {
                     || version == "6"
                     || version == "7"
                     || version == "8"
+                    || version == "9"
```
obycode, line 3136: This can be removed now.

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -3097,6 +3150,7 @@ impl SortitionDB {
...
```
obycode, line 3153: This can be removed now.

obycode (dismissed):
lgtm! 🚀

hstove:
Ok, finally figured out why my Nakamoto integration test was failing. Ready for reviews now @obycode @8marz8 @setzeus

obycode (approved):
lgtm!

8marz8 (approved):

PR #4377, opened 2024-02-14
feat: allow signer-key authorizations instead of signature

### LINKED ISSUE #4379: [Nakamoto] Add signer key authorizations to negate requirement of signatures
We recently added a `signer-sig` parameter to `pox-4` transactions that include a `signer-key` as an argument. The purpose of that signature is to ensure that the signer has authorized the use of their signer key.

This new parameter is problematic to use in burn ops, because the total `OP_RETURN` payload is too long if the signer signature is included.

Instead, we should add the ability for signer key authorizations to happen out of band. If enabled, the `signer-sig` can be omitted from Stacking transactions.

hstove:
This PR implements the concept of signer key authorizations, which can be used to allow omitting the `signer-sig` parameter in pox-4 contract calls. This is a pre-requisite to updating the various stacking burn-ops, because this will allow us to handle burn ops without needing to include a signer key signature in the burn op.

Other than adding an authorization mechanism, this changes the `signer-sig` parameters to an `optional`. If the signature is omitted, the contract verifies that an authorization is present.

- closes #4379

### DIFF
--- a/stackslib/src/chainstate/stacks/boot/mod.rs
+++ b/stackslib/src/chainstate/stacks/boot/mod.rs
@@ -1997,17 +1997,6 @@ pub mod test {
         make_tx(key, nonce, 0, payload)
     }
 
-    pub fn make_pox_4_extend_delete_this(
-        key: &StacksPrivateKey,
-        nonce: u64,
-        addr: PoxAddress,
-        lock_period: u128,
-        signer_key: StacksPublicKey,
-        signature: Vec<u8>,
-    ) -> StacksTransaction {
-        make_pox_4_extend(key, nonce, addr, lock_period, signer_key, Some(signature))
-    }
-
     pub fn make_pox_4_extend(
         key: &StacksPrivateKey,
         nonce: u64,
--- a/stackslib/src/chainstate/stacks/boot/mod.rs
+++ b/stackslib/src/chainstate/stacks/boot/mod.rs
@@ -1844,13 +1844,17 @@ pub mod test {
         key: &StacksPrivateKey,
         nonce: u64,
         amount: u128,
-        addr: PoxAddress,
+        addr: &PoxAddress,
         lock_period: u128,
-        signer_key: StacksPublicKey,
+        signer_key: &StacksPublicKey,
         burn_ht: u64,
-        signature: Vec<u8>,
+        signature_opt: Option<Vec<u8>>,
     ) -> StacksTransaction {
         let addr_tuple = Value::Tuple(addr.as_clarity_tuple().unwrap());
+        let signature = match signature_opt {
+            Some(sig) => Value::some(Value::buff_from(sig).unwrap()).unwrap(),
+            None => Value::none(),
+        };
         let payload = TransactionPayload::new_contract_call(
             boot_code_test_addr(),
             "pox-4",
@@ -1860,7 +1864,7 @@ pub mod test {
                 addr_tuple,
                 Value::UInt(burn_ht as u128),
                 Value::UInt(lock_period),
-                Value::some(Value::buff_from(signature).unwrap()).unwrap(),
+                signature,
                 Value::buff_from(signer_key.to_bytes_compressed()).unwrap(),
             ],
         )
@@ -1993,7 +1997,7 @@ pub mod test {
         make_tx(key, nonce, 0, payload)
     }
 
-    pub fn make_pox_4_extend(
+    pub fn make_pox_4_extend_delete_this(
         key: &StacksPrivateKey,
         nonce: u64,
         addr: PoxAddress,
@@ -2001,7 +2005,22 @@ pub mod test {
         signer_key: StacksPublicKey,
         signature: Vec<u8>,
     ) -> StacksTransaction {
+        make_pox_4_extend(key, nonce, addr, lock_period, signer_key, Some(signature))
+    }
+
+    pub fn make_pox_4_extend(
+        key: &StacksPrivateKey,
+        nonce: u64,
+        addr: PoxAddress,
+        lock_period: u128,
+        signer_key: StacksPublicKey,
+        signature_opt: Option<Vec<u8>>,
+    ) -> StacksTransaction {
         let addr_tuple = Value::Tuple(addr.as_clarity_tuple().unwrap());
+        let signature = match signature_opt {
+            Some(sig) => Value::some(Value::buff_from(sig).unwrap()).unwrap(),
+            None => Value::none(),
+        };
         let payload = TransactionPayload::new_contract_call(
             boot_code_test_addr(),
             POX_4_NAME,
@@ -2009,7 +2028,7 @@ pub mod test {
             vec![
                 Value::UInt(lock_period),
                 addr_tuple,
-                Value::some(Value::buff_from(signature).unwrap()).unwrap(),
+                signature,
                 Value::buff_from(signer_key.to_bytes_compressed()).unwrap(),
             ],
         )
@@ -2103,10 +2122,14 @@ pub mod test {
         nonce: u64,
         pox_addr: &PoxAddress,
         reward_cycle: u128,
-        signature: Vec<u8>,
+        signature_opt: Option<Vec<u8>>,
         signer_key: &Secp256k1PublicKey,
     ) -> StacksTransaction {
         let addr_tuple = Value::Tuple(pox_addr.as_clarity_tuple().unwrap());
+        let signature = match signature_opt {
+            Some(sig) => Value::some(Value::buff_from(sig).unwrap()).unwrap(),
+            None => Value::none(),
+        };
         let payload = TransactionPayload::new_contract_call(
             boot_code_test_addr(),
             POX_4_NAME,
@@ -2114,7 +2137,7 @@ pub mod test {
             vec![
                 addr_tuple,
                 Value::UInt(reward_cycle),
-                Value::some(Value::buff_from(signature).unwrap()).unwrap(),
+                signature,
                 Value::buff_from(signer_key.to_bytes_compressed()).unwrap(),
             ],
         )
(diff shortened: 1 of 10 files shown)


stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -704,21 +717,32 @@
...
+;; and compared to `signer-key`.
+;; 
+;; If `signer-sig` is `none`, the function verifies that an authorization was previously
+;; added for this key.
 (define-read-only (verify-signer-key-sig (pox-addr { version: (buff 1), hashbytes: (buff 32) })
                                          (reward-cycle uint)
                                          (topic (string-ascii 12))
                                          (period uint)
-                                         (signer-sig (buff 65))
+                                         (signer-sig-opt (optional (buff 65)))
```
setzeus, line 730: I believe this & only this parameter was updated to include '-opt,' mistake?

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1300,6 +1324,26 @@
       (ok { stacker: stacker,
             unlock-burn-height: new-unlock-ht }))))
 
+;; Add an authorization for a signer key.
+;; When an authorization is added, the `signer-sig` argument is not required
+;; in the functions that use it as an argument.
+;; The `allowed` flag can be used to either enable or disable the authorization.
+;; Only the Stacks principal associated with `signer-key` can call this function.
+;; *New in Stacks 3.0*
+(define-public (set-signer-key-authorization (pox-addr { version: (buff 1), hashbytes: (buff 32)})
```
setzeus, line 1333: I'm not entirely sure but is there *any* situation in which a signer might revoke authorization? IE - do we need an "allowed" parameter or should this always set the 'signer-key-authorizations' to true?

Is there any unexpected behavior that'd happen as a result of a signer setting then revoking the signature?

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -211,6 +211,19 @@
 ;;   for the given reward cycle
 (define-map aggregate-public-keys uint (buff 33))
 
+;; State for setting allowances for signer keys to be used in
+;; certain stacking transactions
+(define-map signer-key-authorizations 
+    {
+        signer-key: (buff 33),
+        reward-cycle: uint,
+        period: uint,
```
setzeus, line 220: A bit unclear on what "period" means throughout, is this the same as "round"?

Was checking SIP-18 to see if it's related signing structure but don't believe it is...

setzeus: Nvm*** understood that period = # of cycles.

stackslib/src/chainstate/stacks/boot/pox-4.clar
hstove: Yeah you're right, but I did it so that I could use the variable `signer-sig` in the `match` branch. I can update this though to just use `sig` in the match and then we can keep consistent names.

stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -1993,23 +1997,38 @@ pub mod test {
         make_tx(key, nonce, 0, payload)
     }
 
-    pub fn make_pox_4_extend(
+    pub fn make_pox_4_extend_delete_this(
```
jcnelson, line 2000: ??

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -211,6 +211,19 @@
 ;;   for the given reward cycle
 (define-map aggregate-public-keys uint (buff 33))
 
+;; State for setting allowances for signer keys to be used in
+;; certain stacking transactions
```
jcnelson, line 215: Can you document these fields here?  What's the value of this data map?

stackslib/src/chainstate/stacks/boot/pox-4.clar
jcnelson: I personally prefer `-opt` as a suffix to indicate in the code body that it's an `optional`, but I don't feel it so strongly that it's a blocker.

stackslib/src/chainstate/stacks/boot/pox-4.clar
jcnelson: > I'm not entirely sure but is there any situation in which a signer might revoke authorization? IE - do we need an "allowed" parameter or should this always set the 'signer-key-authorizations' to true?

One use-case is that the signer can create and store a revocation transaction (akin to a key revocation certificate in PGP) that they can just have on-hand in case they believe their signing key (which might live somewhere hard to access) is compromised or rendered unusable.

> Is there any unexpected behavior that'd happen as a result of a signer setting then revoking the signature?

You mean setting and revoking the authorization?  A transaction that expects the authorization would subsequently fail.  If we're worried about signers abusing this by setting and revoking authorizations in rapid succession (e.g. within the same block), we can instead map the authorization to the Stacks block height at which it takes effect, and require that block to be in the future.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1300,6 +1324,26 @@
...
+(define-public (set-signer-key-authorization (pox-addr { version: (buff 1), hashbytes: (buff 32)})
+                                             (period uint)
+                                             (reward-cycle uint)
+                                             (topic (string-ascii 12))
+                                             (signer-key (buff 33))
+                                             (allowed bool))
+  (begin
+    ;; Validate that `tx-sender` has the same pubkey hash as `signer-key`
+    (asserts! (is-eq 
+      (unwrap! (principal-construct? (if is-in-mainnet 0x16 0x1a) (hash160 signer-key)) (err ERR_INVALID_SIGNER_KEY))
```
jcnelson, line 1342: Can `0x16` and `0x1a` be constants?

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1300,6 +1324,26 @@
...
+;; in the functions that use it as an argument.
+;; The `allowed` flag can be used to either enable or disable the authorization.
+;; Only the Stacks principal associated with `signer-key` can call this function.
+;; *New in Stacks 3.0*
+(define-public (set-signer-key-authorization (pox-addr { version: (buff 1), hashbytes: (buff 32)})
+                                             (period uint)
+                                             (reward-cycle uint)
+                                             (topic (string-ascii 12))
+                                             (signer-key (buff 33))
+                                             (allowed bool))
```
jcnelson, line 1338: An interesting consequence of this authorization table is that the signer has to perform this contract-call for each and every reward cycle they want to authorize.  This can get expensive (lots of contract-calls) and/or tedious (lots of transactions).  Can this code instead be written such that the caller can specify a _range_ of reward cycles?  Or perhaps an optional last reward cycle?  Then, they'd only need to update this map whenever they want to change the times for which this key is valid.

kantai: I think this contract can just punt that optimization to the caller -- they're perfectly capable of just calling this function multiple times in a single transaction through `contract-call`. That way this contract doesn't have to deal with either implementing the iteration or implementing a range authorization map.

stackslib/src/chainstate/stacks/boot/mod.rs
hstove: Ah crap - left over from some refactoring.

stackslib/src/chainstate/stacks/boot/pox-4.clar
friedger: Having a loop over pox contract calls is kind of expensive.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -2314,6 +2314,444 @@ fn stack_stx_signer_key() {
...
+        stacker_addr,
+        min_ustx,
+        pox_addr.clone(),
+        block_height.into(),
+        lock_period,
+    );
+
+    let topic = Pox4SignatureTopic::AggregationCommit;
+
+    // Stack agg failes without auth
```
jcnelson, line 2472: "failes"

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -704,21 +734,32 @@
...
+  (match signer-sig-opt
+    ;; `signer-sig` is present, verify the signature
+    signer-sig (ok (asserts! 
+      (is-eq 
+        (unwrap! (secp256k1-recover? 
+          (get-signer-key-message-hash pox-addr reward-cycle topic period) 
+          signer-sig) (err ERR_INVALID_SIGNATURE_RECOVER)) 
+        signer-key) 
+      (err ERR_INVALID_SIGNATURE_PUBKEY)))
+    ;; `signer-sig` is not present, verify that an authorization was previously added for this key
```
jcnelson, line 758: Based on the way this function is getting used, a couple things stand out to me here that should be documented.

First, **the reward cycle and duration are inflexible.**  The signer and stacker have to agree on _exactly_ the reward cycle _and_ period length before the stacker can stack, or this fails.  For example, if Alice wants signer Sally to be her signer, and Alice initially wanted to stack for 12 reward cycles but changes her mind and stacks for only 6 cycles, this function will fail unless Sally updates the signer record to indicate 6 cycles.  As another example, suppose Sally agrees to sign for Alice from cycle X through X+12, but Alice stacks in cycle X+1 but for only 11 reward cycles, then her stacking transaction fails unless Sally updates her signer key authorization for Alice.  This isn't a bug per se, but it might be surprising to users.

Second, **the caller must validate `reward-cycle` and `period`.**  By itself, this function would accept reward cycles in the past, or period lengths of 0 cycles or more than 12 cycles.  It may make sense to add defensive checks here, but I don't think it's a problem as long as all of the callers perform these checks.  However, I think this needs to be documented.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1300,6 +1341,26 @@
...
+                                             (period uint)
+                                             (reward-cycle uint)
+                                             (topic (string-ascii 12))
+                                             (signer-key (buff 33))
+                                             (allowed bool))
+  (begin
+    ;; Validate that `tx-sender` has the same pubkey hash as `signer-key`
+    (asserts! (is-eq 
+      (unwrap! (principal-construct? (if is-in-mainnet STACKS_ADDR_VERSION_MAINNET STACKS_ADDR_VERSION_TESTNET) (hash160 signer-key)) (err ERR_INVALID_SIGNER_KEY))
+      tx-sender) (err ERR_NOT_ALLOWED))
```
jcnelson, line 1360: I think there's a surprising behavior here that needs to be documented:  `reward-cycle` must be the reward cycle in which the stacker stacks, not the cycle in which stacking / signing begins.  This is accounted for by all the callers of this function, but it might surprise signers who set their authorization records.

jcnelson (approved):
This LGTM, but please address my two remaining comments regarding documentation.  After that I think you're good to go.  Thanks!

setzeus (approved):

PR #4378, opened 2024-02-14
Drop User Burn Support

### LINKED ISSUE #4071: [Nakamoto] Drop user burn support
This is an opcode we never implemented because there was never any economic incentive to use it.  Let's delete it.

8marz8:
### Description
Removing UserBurnSupportOp definition and usage throughout the codebase as it's not fully implemented and used.

### Applicable issues

- closes #4071


8marz8:
@jcnelson @kantai Should "StagingUserBurnSupport" also be deleted?

jcnelson (dismissed):
LGTM. Thanks for taking care of this!

jcnelson:
Also, yes -- feel free to remove `StagingUserBurnSupport`.  As long as CI passes, it should be fine.

kantai (commented):
These changes look fine to me, but I do have a question, that's probably more for @jcnelson than @8marz8:

`UserBurnSupport` transactions are currently not considered for reward distribution or miner commitment distributions, however, they are _parsed_ currently, correct? This would mean that they are included in the `OpsHash` and therefore `ConsensusHash`, meaning that any change to remove it from the `OpsHash` would need to be epoch-gated, right?

kantai (dismissed):
LGTM! I tested in a separate branch (because you can't test whether or not a UserBurn is parsed in this branch), and UserBurn ops are never parsed under 2.x consensus rules.

kantai (approved):

jcnelson (approved):

PR #4382, opened 2024-02-15
Chore: ed25519 dalek update in `next`

kantai:
### Description

As part of Brice working to unify the `TestSigners` and `SelfSigner` structs, etc., it became clear that having multiple versions of the various rand libraries was going to create headaches. This PR sets workspace versions for `rand`, `rand_core`, and `rand_chacha`, as well as updates `ed25519-dalek` to facilate some unification here.

The biggest interface changes were: `gen_range` accepting a _range_ argument rather than two low, high args, and changes to the `ed25519-dalek` interfaces. As part of this, I removed the `Deref` impls for `VRFPublicKey` and `VRFPrivateKey` -- those were only used to invoke the `as_bytes` and `to_bytes`, so I just added those explicitly. The problem with having `Deref` impls here is that it meant functions like `from_bytes` were overloaded, which I think can make it unclear which function is actually being invoked.

### DIFF
--- a/stacks-common/Cargo.toml
+++ b/stacks-common/Cargo.toml
@@ -18,7 +18,7 @@ name = "stacks_common"
 path = "./src/libcommon.rs"
 
 [dependencies]
-rand = "0.7.3"
+rand = { workspace = true }
 serde = "1"
 serde_derive = "1"
 serde_stacker = "0.1"
@@ -55,8 +55,7 @@ version = "=0.24.2"
 features = ["blob", "serde_json", "i128_blob", "bundled", "trace"]
 
 [dependencies.ed25519-dalek]
-version = "=1.0.0-pre.3"
-features = ["serde"]
+workspace = true
 
 [dependencies.curve25519-dalek]
 version = "=2.0.0"
@@ -70,7 +69,7 @@ features = ["std"]
 rstest = "0.11.0"
 rstest_reuse = "0.1.3"
 assert-json-diff = "1.0.0"
-rand_core = "0.6"
+rand_core = { workspace = true }
 
 [features]
 default = ["developer-mode"]
--- a/stacks-common/src/util/vrf.rs
+++ b/stacks-common/src/util/vrf.rs
@@ -31,7 +31,7 @@ use curve25519_dalek::constants::ED25519_BASEPOINT_POINT;
 use curve25519_dalek::edwards::{CompressedEdwardsY, EdwardsPoint};
 use curve25519_dalek::scalar::Scalar as ed25519_Scalar;
 use ed25519_dalek::{
-    Keypair as VRFKeypair, PublicKey as ed25519_PublicKey, SecretKey as ed25519_PrivateKey,
+    SecretKey as EdDalekSecretKeyBytes, SigningKey as EdPrivateKey, VerifyingKey as EdPublicKey,
 };
 use rand;
 use sha2::{Digest, Sha512};
@@ -39,9 +39,10 @@ use sha2::{Digest, Sha512};
 use crate::util::hash::{hex_bytes, to_hex};
 
 #[derive(Clone)]
-pub struct VRFPublicKey(pub ed25519_PublicKey);
+pub struct VRFPublicKey(pub ed25519_dalek::VerifyingKey);
 
-pub struct VRFPrivateKey(pub ed25519_PrivateKey);
+#[derive(Clone)]
+pub struct VRFPrivateKey(pub ed25519_dalek::SigningKey);
 
 impl serde::Serialize for VRFPublicKey {
     fn serialize<S: serde::Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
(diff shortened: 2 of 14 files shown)


jcnelson (approved):
LGTM as long as CI passes.  Maybe add @xoloki as a reviewer?

stacks-common/src/util/vrf.rs
```diff
@@ -149,71 +114,76 @@ impl Default for VRFPrivateKey {
...
-    pub fn from_private(pk: &VRFPrivateKey) -> VRFPublicKey {
-        VRFPublicKey(ed25519_PublicKey::from(&pk.0))
+    pub fn from_private(sk: &VRFPrivateKey) -> VRFPublicKey {
+        VRFPublicKey(sk.0.verifying_key())
     }
 
+    /// Verify that a given byte string is a well-formed EdDSA public
+    /// key (i.e. it's a compressed Edwards point that is valid), and return
+    /// a VRFPublicKey if so
     pub fn from_bytes(pubkey_bytes: &[u8]) -> Option<VRFPublicKey> {
```
jbencin, line 152: I know you didn't introduce this function in this PR, but this really should be `try_from_bytes()`, or even better, `impl TryFrom<&[u8]> for VRFPublicKey`, and return a `Result`

stacks-common/src/util/vrf.rs
```diff
@@ -149,71 +114,76 @@ impl Default for VRFPrivateKey {
...
+        //  that's what the docs say to do!
+
+        let checked_pubkey = CompressedEdwardsY(pubkey_slice);
+        if checked_pubkey.decompress().is_none() {
+            // invalid
+            return None;
         }
+
+        let key = ed25519_dalek::VerifyingKey::from_bytes(&pubkey_slice).ok()?;
+        Some(VRFPublicKey(key))
```
jbencin, line 168: Simpler way to write this:
```suggestion
        ed25519_dalek::VerifyingKey::from_bytes(&pubkey_slice).map(VRFPublicKey).ok()
```

stacks-common/src/util/vrf.rs
```diff
@@ -149,71 +114,76 @@ impl Default for VRFPrivateKey {
...
     }
 
     pub fn from_hex(h: &str) -> Option<VRFPublicKey> {
-        match hex_bytes(h) {
-            Ok(b) => VRF::check_public_key(&b),
-            Err(_) => None,
-        }
+        let bytes = hex_bytes(h).ok()?;
+        Self::from_bytes(bytes.as_slice())
```
jbencin, line 173: Slightly simpler:

```suggestion
        hex_bytes(h).ok().as_deref().and_then(Self::from_bytes)
```

jbencin (approved):
Looks good, a couple minor style comments. We should really be moving most, if not all, shared crates under `[workspace.dependencies]`

stacks-common/src/util/vrf.rs
kantai: I'm a little conflicted about this kind of thing -- it's actually not obvious to me which of the above is simpler. Sometimes the functional chaining in `option` and `result` can be hard to decipher (especially because it often means that the `Err` type survives to the end of the chain).

stacks-common/src/util/vrf.rs
kantai: I'm a little conflicted about this kind of thing -- it's actually not obvious to me which of the above is simpler. Sometimes the functional chaining in `option` and `result` can be hard to decipher (especially because it often means that the `None` type survives to the end of the chain).

stacks-common/src/util/vrf.rs
kantai: Yeah, I agree, but I don't want to do too much refactoring in this PR (and this would mean a pretty fair amount of refactoring).

stacks-common/src/util/vrf.rs
xoloki: I agree with @kantai , I find this kind of chaining to be less clear than the two line version.

stacks-common/Cargo.toml
```diff
@@ -55,8 +55,7 @@ version = "=0.24.2"
 features = ["blob", "serde_json", "i128_blob", "bundled", "trace"]
 
 [dependencies.ed25519-dalek]
-version = "=1.0.0-pre.3"
-features = ["serde"]
+workspace = true
 
 [dependencies.curve25519-dalek]
```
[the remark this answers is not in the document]
kantai: This PR doesn't change `curve25519-dalek`s dependency definitions, so I'd prefer to just leave this as is.

stacks-common/Cargo.toml
```diff
@@ -55,8 +55,7 @@ version = "=0.24.2"
 features = ["blob", "serde_json", "i128_blob", "bundled", "trace"]
 
 [dependencies.ed25519-dalek]
```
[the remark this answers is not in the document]
kantai: Sure

stacks-common/src/util/vrf.rs
```diff
@@ -31,17 +31,18 @@ use curve25519_dalek::constants::ED25519_BASEPOINT_POINT;
 use curve25519_dalek::edwards::{CompressedEdwardsY, EdwardsPoint};
 use curve25519_dalek::scalar::Scalar as ed25519_Scalar;
 use ed25519_dalek::{
-    Keypair as VRFKeypair, PublicKey as ed25519_PublicKey, SecretKey as ed25519_PrivateKey,
+    SecretKey as EdDalekSecretKeyBytes, SigningKey as EdPrivateKey, VerifyingKey as EdPublicKey,
```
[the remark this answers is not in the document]
kantai: Yeah, I'll remove

stackslib/Cargo.toml
```diff
@@ -81,8 +81,7 @@ version = "=0.24.2"
 features = ["blob", "serde_json", "i128_blob", "bundled", "trace"]
 
 [dependencies.ed25519-dalek]
```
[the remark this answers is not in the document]
kantai: Sure

xoloki (approved):

stacks-common/src/util/vrf.rs
jbencin: Am I the only one who finds chaining easier to read? It just seems like extra steps to take a value out of an `Option`/`Result` and then reconstruct another one, rather than using `map()`/`and_then()`

kantai: For me, it depends on the situation. Sometimes I think it is much clearer to chain, other times less so. When there's multiple fallible steps, I think I usually prefer the non-chained version, because with chaining it can be hard to reason about how the error cases are being handled. Conversely, iteration is often much clearer to me when chained rather than manually looped (in particular things like finding an element or folding a sum).

PR #4385, opened 2024-02-16
Mutants/add skip attribute to timeout functions

ASuciuX:
#### Why are mutations causing timeouts?
Mutations cause timeouts because mutants modify the return value of functions. If the function is essential for stopping a loop or waiting for an HTTP response, when it returns something else, it either doesn't stop waiting for the code or creates an infinite loop, leading to a timeout.

#### Attributes to skip mutants on functions that timeout. Packages covered:
- [x] stacks-node
- [x] stackslib


cylewitruk (approved):
I'm no expert in `mutants`, but I see no harm in these attributes. LGTM! :+1:

PR #4394, opened 2024-02-19
Feat/nakamoto block download API

jcnelson:
This PR adds three new endpoints:

* `/v3/tenures/info`:  this returns metadata about the highest-known tenure.  It will be used by the block-downloader state machine to identify peers with newer block data and fetch them.

* `/v3/tenures/{block-id}`:  this returns all blocks in a tenure ending with `block-id`, all the way back to either the first block in the tenure or the last 2 MB of block data (whichever is reached sooner).  This endpoint will be used to page through tenure data by the block downloader state machine.

* `/v3/blocks/{block-id}`: this returns a specific Nakamoto block.  This endpoint will be used by the block-downloader to fetch the tenure-start blocks, which will allow it to fetch and authenticate the tenures that span them.

In addition, this PR fixes a consensus bug of omission in `next` -- it ensures that a tenure-start Nakamoto block's block-commit commits to the parent tenure's tenure-start block.

### DIFF
--- a/docs/rpc-endpoints.md
+++ b/docs/rpc-endpoints.md
@@ -96,69 +96,24 @@ Callers who wish to download more headers will need to issue this query
 multiple times, with a `?tip=` query parameter set to the index block hash of
 the earliest header received.
 
-Returns a
-[SIP-003](https://github.com/stacksgov/sips/blob/main/sips/sip-003/sip-003-peer-network.md)-encoded
-vector with length up to [Count] that contains a list of the following SIP-003-encoded
-structures:
+Returns a JSON list containing the following:
 
-```rust
-struct ExtendedStacksHeader {
-    consensus_hash: ConsensusHash,
-    header: StacksBlockHeader,
-    parent_block_id: StacksBlockId,
-}
+```json
+[
+  {
+    "consensus_hash": "dff37af13badf99683228e61c71585bb7a82ac92",
+    "header":
+"0600000047ddfbee8c00000000000222c7ad9042e5a67a703ff3b06581e3fd8a2f1496a563dc41462ebf8e5b046b43e7085f20e828840f26fefbe93a048f6c390ce55b954b188a43781fa0db61c091dbb840717fda77f9fc16d8ac85f80bbf2d04a20d17328390e03b8f496986f6351def656fd12cc4b8fe5e2cfb8d3f2e67c3a700000000000000000000000000000000000000000000000000000000000000000000fb432fbe28fb60ab37c8f59eec2397a0d0bcaf679a34b39d02d338935c7e723e062d571e331fb5016d3000ab68da691baa02b4a5dde7befa2edceb219af959312544d306919a59ee4cfd616dc3cc44a6f01ac7c8",
+    "parent_block_id":
+"e0cb2be07552556f856503d2fbd855a27d49dc5a8c47fb2d9f0314eb6bad6861"
+  }
+]
 ```
 
-Where `ConsensusHash` is a 20-byte byte buffer.
-
-Where `StacksBlockId` is a 32-byte byte buffer.
-
-Where `StacksBlockHeader` is the following SIP-003-encoded structure:
-
-```rust
-struct StacksBlockHeader {
-    version: u8,
-    total_work: StacksWorkScore,
-    proof: VRFProof,
-    parent_block: BlockHeaderHash,
-    parent_microblock: BlockHeaderHash,
-    parent_microblock_sequence: u16,
-    tx_merkle_root: Sha512Trunc256Sum,
-    state_index_root: TrieHash,
-    microblock_pubkey_hash: Hash160,
-}
-```
-
-Where `BlockHeaderHash`, `Sha512Trunc256Sum`, and `TrieHash` are 32-byte byte
-buffers.
-
-Where `Hash160` is a 20-byte byte buffer.
-
-Where `StacksWorkScore` and `VRFProof` are the following SIP-003-encoded structures:
-
-```rust
-struct StacksWorkScore {
-    burn: u64,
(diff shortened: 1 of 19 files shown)


docs/rpc-endpoints.md
```diff
@@ -540,3 +495,41 @@ Error examples:
...
+  "tip_block_id": "317c0ee162d1ee02c67d5bca79003dafc59aa84579360387f43650c37491ac3b",
+  "tip_height": 116
+}
+```
+
+Here, `consensus_hash` identifies the highest-known tenure (which may not be the
+highest sortition), `reward_cycle` identifies the reward cycle number of this
+tenure, `tip_block_id` idenitifies the highest-known block in this tenure, and
+`tip_height` identifies that block's height.
+
```
kantai, line 535: Use `docs/openapi.yml` instead. `rpc-endpoints.md` should really just be removed.

jcnelson:
Well well well, looks like my consensus bugfix has caused some tests to correctly fail (they weren't supposed to be passing).  Will see if I can fix them.

stackslib/src/chainstate/nakamoto/staging_blocks.rs
```diff
@@ -188,6 +232,25 @@ impl<'a> NakamotoStagingBlocksConnRef<'a> {
...
+    ) -> Result<Option<u64>, ChainstateError> {
+        let qry = "SELECT length(data) FROM nakamoto_staging_blocks WHERE index_block_hash = ?1";
+        let args: &[&dyn ToSql] = &[index_block_hash];
+        let res: Option<i64> = query_row(self, qry, args)?;
+        let Some(size_i64) = res else {
+            return Ok(None);
+        };
+        Ok(Some(
+            u64::try_from(size_i64).expect("FATAL: block exceeds i64::MAX"),
+        ))
```
jbencin, line 251: Would be a bit simpler with a `.map()`

```suggestion
        let res = query_row(self, qry, args)?
            .map(|size: i64| u64::try_from(size).expect("FATAL: block exceeds i64::MAX"));
        Ok(res)
```

stackslib/src/net/api/getblock_v3.rs
```diff
@@ -0,0 +1,324 @@
...
+    ) -> Result<HttpRequestContents, Error> {
+        if preamble.get_content_length() != 0 {
+            return Err(Error::DecodeError(
+                "Invalid Http request: expected 0-length body".to_string(),
+            ));
+        }
+
+        let block_id_str = captures
+            .name("block_id")
+            .ok_or(Error::DecodeError(
```
jbencin, line 146: Lazy evaluation will avoid unnecessary string allocations:

```suggestion
            .ok_or_else(|| Error::DecodeError(
```

stackslib/src/net/api/gettenure.rs
```diff
@@ -0,0 +1,324 @@
...
+
+        // stop sending if the parent is an epoch2 block
+        let Some(parent_nakamoto_header) = parent_header.anchored_header.as_stacks_nakamoto()
+        else {
+            return Ok(false);
+        };
+
+        // stop sending if the parent is in a different tenure
+        if parent_nakamoto_header.consensus_hash != self.block_stream.consensus_hash {
+            return Ok(false);
```
jbencin, line 105: Should this and the above condition be errors? This function returns `Ok(false)` whether it's out of space (and the caller should try again), or the block is invalid

stackslib/src/net/api/gettenure.rs
```diff
@@ -0,0 +1,324 @@
...
+    ) -> Result<HttpRequestContents, Error> {
+        if preamble.get_content_length() != 0 {
+            return Err(Error::DecodeError(
+                "Invalid Http request: expected 0-length body".to_string(),
+            ));
+        }
+
+        let block_id_str = captures
+            .name("block_id")
+            .ok_or(Error::DecodeError(
```
jbencin, line 158: Lazy evaluation will avoid unnecessary string allocations:

```suggestion
            .ok_or_else(|| Error::DecodeError(
```

jbencin (approved):
Looks good! Left a few minor comments

jcnelson:
@kantai @obycode ping for re-review

docs/rpc-endpoints.md
jcnelson: Done

stackslib/src/net/api/gettenure.rs
kantai: @jcnelson to confirm, but I think that once `MAX_MESSAGE_LEN`, a new request must be made, so the caller actually shouldn't try again to get more chunks from the stream (i.e., the HTTP stream should close).

kantai (approved):
This LGTM, but just confirm my reasoning in the gettenure discussion.

stackslib/src/net/api/gettenure.rs
jcnelson: `Ok(false)` here means "end of stream", not an error.  @kantai that is correct -- the caller would resume downloading the tenure as of the last received block (whose `block_id` is the path parameter), and use the `stop=` query parameter to indicate to the server to stop sending blocks once a particular block is reached (such as, for example, if the caller is querying only new unconfirmed tenure blocks).  This is handled by the `NakamotoTenureDownloader` and `NakamotoUnconfirmedTenureDownloader` state machines in the downloader PR -- they'll both remain in their `GetTenureBlocks(last_block_id)` state until the full tenure has been fetched, and will update `last_block_id` if the blocks received did not include the tenure-start block.

The `stop=` query parameter isn't present in this PR; it's part of the downloader itself (which modified this API).

obycode (approved):

kantai (approved):

PR #4396, opened 2024-02-20
fix: include txid in more failure logs

janniks:
### Description

Adds `txid` to more failure logs for better interpretability of logs in production.

Can be used for better understanding what happened to a tx given a txid and logs.

### Checklist

- [ ] Changelog is updated
- [ ] ~~Required documentation changes (e.g., `docs/rpc/openapi.yaml` and `rpc-endpoints.md` for v2 endpoints, `event-dispatcher.md` for new events)~~
- [ ] ~~New clarity functions have corresponding PR in `clarity-benchmarking` repo~~
- [ ] ~~New integration test(s) added to `bitcoin-tests.yml`~~


jcnelson (changes requested):
Hey, we don't open PRs into `master` directly unless it's a hotfix or a release.  Can you instead open this against `develop`?

janniks:
Apologies, forgot to switch 👍🏻 Will update

jbencin (approved):

jcnelson (dismissed):

jcnelson:
Just make sure CI passes

janniks:
Right now tests are failing because some auto-generated contract names are too long. e.g. a contract name from setting up the test observer; `blockstack_lib-chainstate-stacks-boot-pox_3_tests-pox_3_getters`
Unfortunately, Txid serializes and panics on failure.

@jbencin any ideas on how to go about this, or point me to the correct code. e.g. if the exact string isn't important, we can slice it to the max contract length, or something similar.

jbencin:
@jcnelson @janniks I found the issue. `ContractName` is implemented using a macro, like this:

```rust
guarded_string!(
    ContractName,
    "ContractName",
    CONTRACT_NAME_REGEX,
    MAX_STRING_LEN,
    RuntimeErrorType,
    RuntimeErrorType::BadNameValue
);
```

`MAX_STRING_LEN` is 128, so it's possible to construct a `ContractName` with up to 128 characters. However, when we implement `StacksMessageCodec` for `ContractName`:

```rust
        if self.as_bytes().len() < CONTRACT_MIN_NAME_LENGTH as usize
            || self.as_bytes().len() > CONTRACT_MAX_NAME_LENGTH as usize
        {
            // Error!
        }
        // ...
    }
```

`CONTRACT_MAX_NAME_LENGTH` is 40, much smaller than `MAX_STRING_LEN`. This means that `ContractName::new()` allows us to create structs that can't be serialized

### Proposed solutions
- Change the macro to use `CONTRACT_MAX_NAME_LENGTH` instead of `MAX_STRING_LEN`. This will likely expose more invalid `ContractName`s that were never being serialized, but it will make them easy to find and fix
- Change `CONTRACT_MAX_NAME_LENGTH` to match `MAX_STRING_LEN`. Probably not what we want to do, but would fix the issue immediately

janniks:
Updated the test helpers to not allow invalid contract name generation. (Capping at `CONTRACT_MAX_NAME_LENGTH` and remove leading dashes). This way tests shouldn't fail due to the logging of txids, while in production no unserializable txs should reach this code.

stackslib/src/net/mod.rs
```diff
@@ -2403,7 +2404,19 @@ pub mod test {
...
+                                        .to_string()
+                                        .chars()
+                                        // ensure auto-generated contract names are not too long
+                                        .skip(
+                                            conf.test_name
+                                                .len()
+                                                .saturating_sub(CONTRACT_MAX_NAME_LENGTH),
+                                        )
+                                        .collect::<String>()
+                                        .trim_start_matches(|c: char| !c.is_alphabetic())
```
jbencin, line 2418: Why not do this before limiting the length?

jbencin:
@jcnelson Already made an issue, #4727

We are modifying the PR to avoid the issue for now

stackslib/src/chainstate/stacks/db/transactions.rs
```diff
@@ -1059,7 +1061,8 @@ impl StacksChainState {
                               "function_name" => %contract_call.function_name,
                               "function_args" => %VecDisplay(&contract_call.function_args),
                               "return_value" => %return_value,
-                              "cost" => ?total_cost);
+                              "cost" => ?total_cost,
+                              "txid" => %tx.txid());
```
obycode, line 1065: `txid` is already in this log, see line 1057. Similarly for the other logs in this file.

janniks: Thanks, removed this one again.

stackslib/src/chainstate/stacks/db/transactions.rs
```diff
@@ -6873,6 +6876,12 @@ pub mod test {
                 mode,
                 origin,
                 &ft_transfer_2,
+                Txid(
+                    "1232121232121232121232121232121232121232121232121232121232121232"
```
obycode, line 6880: I think this string is too long. Try 32 chars: `"12321212321212321212321212321212"`

obycode: Or even simpler, `Txid([0;32])`.

jcnelson (approved):

PR #4401, opened 2024-02-20
Check Multiple Tests' Success at Once

BowTiedDevOps:
Using a composite action to check that multiple jobs which are running tests are successful. 

The called action is succeeding only if all the jobs parsed through the `jobs` input (JSON string of `needs` field) have `success` as status, otherwise it fails.

The purpose of implementing this step is to add the possibility of only making the job that calls the composite required in a ruleset check, instead of having each individual test required (i.e. each job from a matrix strategy).

Reference runs:
- Successful (no test failures): https://github.com/BowTiedDevOps/stacks-core/actions/runs/7978701647?pr=3
- Failed (4 test failures - check jobs also failing): https://github.com/BowTiedDevOps/stacks-core/actions/runs/7979062950?pr=4
- In both of these runs, `Docker Image` is failing because of a push authorization - the workflow is unchanged and only works with the secrets from `stacks-network/stacks-core` repository

### DIFF
--- a/.github/workflows/atlas-tests.yml
+++ b/.github/workflows/atlas-tests.yml
@@ -54,3 +54,17 @@ jobs:
         uses: stacks-network/actions/codecov@main
         with:
           test-name: ${{ matrix.test-name }}
+
+  check-tests:
+    name: Check Tests
+    runs-on: ubuntu-latest
+    if: always()
+    needs:
+      - atlas-tests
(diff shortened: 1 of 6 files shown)


CharlieC3 (approved):
Great change. LGTM! :shipit:

jbencin (approved):

PR #4406, opened 2024-02-21
Broadcast `/new_pox_set` events to observers registered to `any` event

zone117x:
The `/new_pox_set` event was not broadcasted to observers who subscribed to the `*` any event.

### DIFF
--- a/testnet/stacks-node/src/event_dispatcher.rs
+++ b/testnet/stacks-node/src/event_dispatcher.rs
@@ -940,7 +940,7 @@ impl EventDispatcher {
         cycle_number: u64,
     ) {
         let interested_observers =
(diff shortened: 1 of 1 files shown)


zone117x:
Currently the `STACKS_EVENT_OBSERVER` env var is the easiest way to register an event observer in many deployments, for example in docker-compose. I don't think the `*` option should be removed unless the env var option is improved to be able to preserve the functionality. It's painful to depend on .toml template file manipulation to swap in host names.

I also don't think adding to the `*` events should be considered a breaking change.

zone117x:
Note that if https://github.com/stacks-network/stacks-core/issues/4407 is implemented, then we could table this discussion for now

PR #4410, opened 2024-02-22
Clarity-Wasm: add support for writing string-utf8 to Wasm memory

Acaccia:
Exactly what the title say, this adds support for writing of string-utf8 to Wasm memory.

This will be extensively tested in the clarity-wasm repo.


csgui (approved):
Code LGTM!

PR #4412, opened 2024-02-22
Update StackStxOp to include signer key

### LINKED ISSUE #4285: [Nakamoto] stack-stx burn op requires a signing key and signature
The Stacks-on-Bitcoin `stack-stx` transaction needs to encode a signing key and a signature as well as the existing arguments.  Right now, the OP_RETURN is as follows:

```
0      2  3                             19           20
|------|--|-----------------------------|------------|
magic  op         uSTX to lock (u128)     cycles (u8)
```

We can expand this to include the signing key, but not the signature:

```
0      2  3                             19           20                   53
|------|--|-----------------------------|------------|---------------------|
magic  op         uSTX to lock (u128)     cycles (u8)    signing key
```

What I was thinking is that in order to stack STX, you'd have to create a third output in its `PreStxOp` that is the hash160 of the signing key, and then spend it in the `StackStxOp`.  We already do this in order to prove ownership of the STX -- the `StackStxOp` spends the second output of the `PreStxOp` and thereby proves that the second output of the `PreStxOp` is `tx-sender`.  So, the `PreStxOp` for `StacksStxOp` would manifest as follows:

```
 PreStxOp                                                      StackStxOp

.-------------------.-------------------.                      .-------------------------------------.------------------.
|                   | OP_RETURN ...     |          .---------> | <tx-sender sig> <tx-sender pubkey>  | OP_RETURN ...    |
|                   |-------------------|   spend  |           |-------------------------------------|------------------|
| (funding inputs)  | tx-sender UTXO    |  --------*   .-----> | <signer sig> <signer pubkey>        | PoX reward addr  |
|                   |-------------------|   spend      |       |-------------------------------------|------------------|
|                   | signing key UTXO  |  ------------*       |        (funding inputs)             | (change outputs) |
*-------------------*-------------------*                      *-------------------------------------*------------------*
```

8marz8:
### Description
Following the signer-key authorization in #4377, this PR updates stack-stx burn op (StackStxOp) to include signer-key.

### Applicable issues

- closes #4285


stackslib/src/chainstate/stacks/db/blocks.rs
```diff
@@ -4239,25 +4239,38 @@ impl StacksChainState {
...
+                Value::UInt(u128::from(*num_cycles)),
+            ];
+            // Appending additional signer related arguments for pox-4
+            if POX_4_NAME == active_pox_contract {
+                // Passing None for signer-sig
+                args.push(Value::none());
+
+                let signer_key_value = signer_key
+                    .as_ref()
+                    .expect("signer_key is required for pox-4");
```
hstove, line 4264: I think we want to early exit the function here with a logged warning, instead of panicking.

hstove (changes requested):
Nice job! I think this is pretty close. You should mainly add an integration test (like in `nakamoto_integrations`)

stackslib/src/chainstate/burn/operations/stack_stx.rs
```diff
@@ -169,33 +177,58 @@ impl StackStxOp {
     fn parse_data(data: &Vec<u8>) -> Option<ParsedData> {
         /*
             Wire format:
-            0      2  3                             19        20
-            |------|--|-----------------------------|---------|
-             magic  op         uSTX to lock (u128)     cycles (u8)
+            0      2  3                             19           20                   54
+            |------|--|-----------------------------|------------|---------------------|
+            magic  op         uSTX to lock (u128)     cycles (u8)    signing key
```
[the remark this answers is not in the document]
8marz8: Yep, I have it below but will include it in wire format too.

hstove:
Ah, I just realized that https://github.com/stacks-network/stacks-core/pull/4461 will require an update to this burn op (two new `u128` fields)

stackslib/src/chainstate/burn/operations/stack_stx.rs
```diff
@@ -169,33 +177,59 @@ impl StackStxOp {
...
 
              parent-delta and parent-txoff will both be 0 if this block builds off of the genesis block.
+
+             "signer key" is encoded as follows: the first byte is an option marker
+               - if it is set to 1, the parse function attempts to parse the next 33 bytes as a StacksPublicKeyBuffer
+               - if it is set to 0, the value is interpreted as None
         */
 
-        if data.len() < 17 {
+        if data.len() < 18 {
```
hstove, line 196: I think we either need to:

- add an epoch check here
- default to `0` if there is no data past the `cycle` (IMO this is better)

Otherwise, in epoch <2.5, this would be a consensus-breaking change

stackslib/src/chainstate/burn/operations/stack_stx.rs
```diff
@@ -300,6 +334,7 @@ impl StackStxOp {
             reward_addr: reward_addr,
             stacked_ustx: data.stacked_ustx,
             num_cycles: data.num_cycles,
+            signer_key: data.signer_key, // QUESTION: is retrieving the signer_key correct in this way or should it get retrieved from tx?
```
hstove, line 337: Feel free to remove the comment :)

stackslib/src/chainstate/burn/operations/stack_stx.rs
```diff
@@ -322,16 +357,26 @@ impl StacksMessageCodec for PreStxOp {
...
+
+        if let Some(signer_key) = self.signer_key {
+            fd.write_all(&(1 as u8).to_be_bytes())
+                .map_err(|e| codec_error::WriteError(e))?;
+            fd.write_all(&signer_key.as_bytes()[..])
+                .map_err(codec_error::WriteError)?;
+        } else {
+            fd.write_all(&(0 as u8).to_be_bytes())
+                .map_err(|e| codec_error::WriteError(e))?;
+        }
```
hstove, line 379: Relating to my other comment, we need to make sure this is backwards compatible. Otherwise, the same exact op would serialize differently. Because of that, I wonder if it's better to just drop the option flag. If it weren't for backwards compatibility I think the option flag is good, but it might add too much complexity in this case.

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -357,6 +358,9 @@ impl FromRow<StackStxOp> for StackStxOp {
         let stacked_ustx = u128::from_str_radix(&stacked_ustx_str, 10)
             .expect("CORRUPTION: bad u128 written to sortdb");
         let num_cycles = row.get_unwrap("num_cycles");
+        let signer_key_str: String = row.get_unwrap("signer_key");
+        let signer_key = serde_json::from_str(&signer_key_str)
+            .expect("CORRUPTION: DB stored bad transition ops");
```
hstove, line 363: This works because we have `DEFAULT NULL` in the DB migration, right? Just want to double check

hstove (changes requested):
Everything looks great, I just have some comments around backwards compatibility with the existing burn op structure. Nice work!

stackslib/src/chainstate/burn/operations/stack_stx.rs
8marz8: TBH I didn't consider the consensus-breaking change and just mirrored what was done for delegate-stx for two of its optional fields...I fixed this now by going with your second suggestion since it didn't require an epoch check.

stackslib/src/chainstate/burn/db/sortdb.rs
8marz8: I think so...along with the default de/serialization. I had this doubt for [this](https://github.com/stacks-network/stacks-core/pull/4412/files#diff-f6160bafb8ea4919ebd27f68cd1a9b56e66c8e0be60c544cc379b6aa8e2c960cR5308) too but I included None for signer_key in unit tests and it doesn't cause panic.

jferrant (dismissed):

stackslib/src/chainstate/burn/db/sortdb.rs
jcnelson: If `NULL` is allowed, then shouldn't this be:

```
let signing_key_str_opt: Option<String> = row.get_unwrap("signer_key");
```

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -2213,6 +2217,262 @@ fn stx_delegate_btc_integration_test() {
...
+        "block_height" => block_height,
+        "reward_cycle" => reward_cycle,
+    );
+
+    let stack_stx_op = BlockstackOperationType::StackStx(StackStxOp {
+        sender: spender_stx_addr.clone(),
+        reward_addr: pox_addr,
+        stacked_ustx: 100000,
+        num_cycles: reward_cycle.try_into().unwrap(),
+        signer_key: Some(signer_key),
```
jcnelson, line 2418: Can you also test the case where `signer_key` is `None`?

stackslib/src/chainstate/burn/db/sortdb.rs
8marz8: Both work but I will use your suggestion with match statement since it'll be more clear/safer.
When I debugged mine with signer_key as `None`, it evaluated to:

`let signer_key_str: String = row.get_unwrap("signer_key");`
-> "null"
 `let signer_key = serde_json::from_str(&signer_key_str)
            .expect("CORRUPTION: DB stored bad transition ops");`
->  None

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -2213,6 +2217,319 @@ fn stx_delegate_btc_integration_test() {
...
+                let stack_stx_obj = burnchain_op.get("stack_stx").unwrap();
+                let signer_key_found = stack_stx_obj
+                    .get("signer_key")
+                    .expect("Expected signer_key in burn op")
+                    .as_str()
+                    .unwrap();
+                assert_eq!(signer_key_found, signer_key.to_hex());
+
+                stack_stx_found = true;
+            }
```
hstove, line 2524: Can you add some assertion that we only found 1 stack-stx op, and not two? Just so that we have some validation that the op without a signer key was not submitted?

I think in this case it could also be helpful to try and query `sortdb` to see that the op without a signer key was found (aka it was "valid") but still not executed.

hstove (changes requested):
This is looking great! IMO I'd like to see just one more assertion around the op without a signer key being found and not submitted. I know this is kinda nitty, but I just want the test to ensure it's not being rejected for some other reason.

testnet/stacks-node/src/tests/neon_integrations.rs
8marz8: I think `sortdb` call helped discover an issue directly or undirectly....because the query returned 0.
I came across this log while grepping and it seems since I am not including a signature in `StackStxOp` and passing `None` when submitting a clarity_tx, It's failing because an authorization wasn't added for that signer-key (I suppose that happens via `set-signer-key-authorization`.
I'll see if I can fix that up.

> DEBG [1709861612.139892] [stackslib/src/chainstate/stacks/db/blocks.rs:4202] [miner-block-http://127.0.0.1:33451] StackStx burn op rejected by PoX contract., txid: 6110c53d30f52cfca7498aef8cdc45c7932a7cddf9d7dd97034108449f99e597, burn_block: 31c8df6a265df548aa8f72ae56c625307067ef96c94e9ac57e271b3c144e032e, contract_call_ecode: 19

8marz8: My [last commit] (https://github.com/stacks-network/stacks-core/pull/4412/commits/a0f57913ab208569b7d9a76d2ef5fa4b1a18477e) takes care of this as verified by the added assertions and logs..._but_ I am not sure if this is what we want? Since it was a big thing to miss and I'm not sure if there was a diff idea behind making this work without a signature _and_ without an additional Clarity call.

Logs showing both Clarity transactions going through:

> DEBG [1709878156.643139] [stackslib/src/chainstate/stacks/db/blocks.rs:4668] [chains-coordinator-127.0.0.1:39177] Processed set-signer-key-authorization

>DEBG [1709878156.663462] [stackslib/src/chainstate/stacks/db/blocks.rs:4230] [chains-coordinator-
127.0.0.1:39177] Processed StackStx burnchain op, amount_ustx: 10000000000000, num_cycles: 6, burn_block_height: 210, sender: ST3WM51TCWMJYGZS1QFMC28DH5YP86782YGR113C1, reward_addr: {"Standard":[{"version":26,"bytes":"f942874ce525e87f21bbe8c121b12fac831d02f4"},"SerializeP2PKH"]}, txid: cc5d575bd9cd01553943495cddb0a05d1a071d7a757841dd252474b54b3683ef

8marz8: More updates on this: Integ tests were passing with this change (the burn op transaction is found in block events and from sortdb)
But, after I tried merging with next and resolved conflicts, the burn op transaction was no longer being found in block events (but it was still found in sortdb)
I didn't push that up since it broke the test and above approach wasn't confirmed. 

I found below warning that I think is indicative of the issue:

> WARN [1710112824.503133] [stackslib/src/chainstate/stacks/db/blocks.rs:5610] [chains-coordinator-127.0.0.1:43685] Block 6b225be385245404ed8ddc3ecefee607b67d763799fc697ff575d702b21a4ae9 state root mismatch: expected 32969e8edfa086f9f8c60782c33f1a5be0feded5609e8d1380cb108b44e0fe95, got cf9705f9acb1397c4ee6cf4214c4b906c32f188ddebac92b0f298db7988200a9
WARN [1710112824.503727] [stackslib/src/chainstate/stacks/db/blocks.rs:6348] [chains-coordinator-127.0.0.1:43685] Encountered invalid block: Block 6b225be385245404ed8ddc3ecefee607b67d763799fc697ff575d702b21a4ae9 state root mismatch: expected 32969e8edfa086f9f8c60782c33f1a5be0feded5609e8d1380cb108b44e0fe95, got cf9705f9acb1397c4ee6cf4214c4b906c32f188ddebac92b0f298db7988200a9

I suspect this is happening because the `set-signer-key-authorization` is not properly updating the `StacksTransactionReceipt` as in the current change I have, I didn't add a a new `BlockstackOperationType` and just called with without adding a `receipt`.
Not sure why this is passing without bringing in changes in `next`
But, if we are to make this additional `set-signer-key-authorization` clarity call for every `stack-stx` call, then we might need to define a new `BlockstackOperationType` for it.

hstove: I suspect this is failing because `max-amount` and `auth-id` were merged into next, so the clarity transaction is now failing. This updates both `stack-stx` and `set-signer-key-authorization`. 

The fix would be to append both of those params to those functions (the "vanilla" tx and the burn op). The burn op will need to be updated to include those two params in the data field.

Let me know if you need pointers about how to use those fields in tests, but it's pretty straightforward:

- You can use `u128::MAX` for `max-amount`
- Use any number for auth ID, they just have to match

Apologies for not catching this earlier, I should have thought about how the test would fail without the prior authorization.

Also, I wouldn't worry about making a burn op for `set-signer-key-authorization` in this PR

8marz8: This is done in my recent commit now and it helped the Neon integ test to pass locally. But the Nakamoto integ test fails, and it seems to be a the same issue with [vote burn op](https://github.com/stacks-network/stacks-core/actions/runs/8238927057/job/22531584884?pr=4376#step:5:2416) but I wasn't able to figure out why.

hstove:
So I pushed a small change to fix a test timeout issue, but now I'm seeing an issue where the BTC op isn't being "accepted" by the regtest controller. There isn't much of an error, but I think it might be that the total size of the OP_RETURN is too large. We already squeezed `auth_id` down to u64, and I'm not sure if downsizing `max_amount` is a good idea.

hstove:
It took me a while to figure out why the integration test was failing, but I finally narrowed it down to not having enough balance when building the stack-stx operation transaction.

I also modified the `auth_id` data to be `u32`, because it's not totally clear to me if using u64 would put us over the OP_RETURN max length when including all the other overhead. If this is not the right call, we can just revert my last commit.

stackslib/src/chainstate/burn/operations/stack_stx.rs
```diff
@@ -193,9 +208,26 @@ impl StackStxOp {
...
         let num_cycles = data[16];
 
+        let mut signer_key: Option<StacksPublicKeyBuffer> = None;
+        let mut max_amount: Option<u128> = None;
+        let mut auth_id: Option<u32> = None;
+
+        if data.len() >= 50 {
+            signer_key = Some(StacksPublicKeyBuffer::from(&data[17..50]));
+        }
+        if data.len() >= 66 {
```
jcnelson, line 218: This could be made slightly safer with the following:

```rust
if data.len() >= 66 {
   let Some(amt) = parse_u128_from_be(&data[50..66]) else {
      return None;
   };
   max_amount = Some(amt);
}
if data.len() >= 70 {
   let Some(id) = parse_u32_from_be(&data[66..70]) else {
      return None;
   };
   auth_id = Some(id);
}
```

stackslib/src/chainstate/burn/operations/stack_stx.rs
```diff
@@ -703,4 +833,43 @@ mod tests {
...
+            let mut bytes = ['T' as u8, '3' as u8].to_vec();
+            op.consensus_serialize(&mut bytes)
+                .expect("Expected to be able to serialize op into bytes");
+            bytes
+        };
+        let script = Builder::new()
+            .push_opcode(opcodes::All::OP_RETURN)
+            .push_slice(&op_bytes)
+            .into_script();
+        // assert_eq!(script.len(), 79);
```
jcnelson, line 871: Is there a reason we can't be precise in the length check?

stackslib/src/chainstate/nakamoto/mod.rs
```diff
@@ -2581,11 +2581,18 @@ impl NakamotoChainState {
 
         let active_pox_contract = pox_constants.active_pox_contract(burn_header_height.into());
 
+        let pox_reward_cycle = Burnchain::static_block_height_to_reward_cycle(
+            burn_header_height as u64,
+            sortition_dbconn.get_burn_start_height().into(),
+            sortition_dbconn.get_pox_reward_cycle_length().into(),
+        ).expect("FATAL: Unrecoverable chainstate corruption: Epoch 2.1 code evaluated before first burn block height");
```
jcnelson, line 2588: this would fail if the `burn_header_height` is less than 666050, which is when Stacks 2.0 started (not 2.1).

stackslib/src/chainstate/stacks/db/blocks.rs
```diff
@@ -4150,25 +4152,89 @@ impl StacksChainState {
...
+                // we parse the StackStxOp from a burnchain transaction
+                reward_addr
+                    .as_clarity_tuple()
+                    .expect("FATAL: stack-stx operation has no hash mode")
+                    .into(),
+                Value::UInt(u128::from(*block_height)),
+                Value::UInt(u128::from(*num_cycles)),
+            ];
+            // Appending additional signer related arguments for pox-4
+            if POX_4_NAME == active_pox_contract {
```
jcnelson, line 4173: Can this `if`-body be put into its own method?

stackslib/src/chainstate/stacks/db/blocks.rs
```diff
@@ -4689,6 +4755,68 @@ impl StacksChainState {
...
+                None,
+                &boot_code_id(active_pox_contract, mainnet),
+                "set-signer-key-authorization",
+                &signer_auth_args,
+                |_, _| false,
+            )
+        }) {
+            Ok((value, _, _)) => {
+                if let Value::Response(ref resp) = value {
+                    if !resp.committed {
```
jcnelson, line 4794: This shouldn't be an error.  If the transaction contract-call fails, we still want to capture it in a transaction receipt (just as we do today).  You can see an example of this in `process_stacking_ops` in how we handle the call to `stack-stx`.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -2427,3 +2427,371 @@ fn vote_for_aggregate_key_burn_op() {
...
+        "Both stack-stx ops with and without a signer_key should be considered valid."
+    );
+    assert!(
+        found_none,
+        "Expected one stacking_op to have a signer_key of None"
+    );
+    assert!(
+        found_some,
+        "Expected one stacking_op to have a signer_key of Some"
+    );
```
jcnelson, line 2788: Can you add a few extra tests for when the `StackStxOp` has invalid arguments, and verify that we get a transaction failure receipt?

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -2231,6 +2232,366 @@ fn stx_delegate_btc_integration_test() {
     channel.stop_chains_coordinator();
 }
 
+#[test]
+#[ignore]
+fn stack_stx_burn_op_test() {
```
jcnelson, line 2237: Can you help me understand the intended differences between this test and the `stack_stx_burn_op_integration_test()` function?

jcnelson (changes requested):
Hi @8marz8, thank you for the updates!  There seems to be at least one logic bug in this PR relating to how we handle a failed `StackStxOp`, but it's easily addressed.

testnet/stacks-node/src/tests/neon_integrations.rs
hstove: I'll respond because I know Marzi was somewhat following the test pattern I did for the vote-for-aggregate-key burn op PR.

Because the burn operations are handled in separate places in 2.x vs 3.0, it's important to test that it works in both scenarios. So that's why the two tests. In hindsight, it would probably be a lot more straightforward to just have one integration test that made these operations in both epochs.

stackslib/src/chainstate/burn/operations/stack_stx.rs
hstove: Added 👍🏼

hstove:
As I was updating this PR for one of @jcnelson's [review comments](https://github.com/stacks-network/stacks-core/pull/4412/files/584ca20659bd036495f2705ad1a30316d87666f1#diff-3b9efbb4d50c023573c1ab596f563031f94f18c8737ff9fd6f4c8e01b7a1b968), I noticed a logic bug: the function `process_stacking_ops` automatically calls `set-signer-key-authorization` to ensure that the internal Clarity transaction is successful. We missed this on previous reviews, but we definitely don't want this behavior: the whole point (for the most part) of `set-signer-key-authorization` is to support having `stack-stx` burn ops without needing the whole signature. However, we still want to ensure that the signer has provided this authorization beforehand. With the way this PR currently works, you could use this burn op to bypass the whole signer key authorization check.

This isn't a huge change - we'll need to strip out the mentioned code in `process_stacking_ops`, and then update the integration tests to add the `set-signer-key-authorization` check before making the burn ops.

testnet/stacks-node/src/tests/nakamoto_integrations.rs
hstove: In my opinion it makes more sense to do this in unit tests. I have added more checks to ensure that the parameters stored in the receipt are equivalent to what we submitted. As long as the contract call payload is setup correctly from the burn op, we can ensure that the receipt behavior matches the rest of our unit tests

stackslib/src/chainstate/stacks/db/blocks.rs
hstove: Resolving as this code was removed

stackslib/src/chainstate/stacks/db/blocks.rs
hstove: done 👍

hstove (dismissed):

stackslib/src/chainstate/stacks/db/blocks.rs
```diff
@@ -4152,23 +4153,37 @@ impl StacksChainState {
...
+                // we parse the StackStxOp from a burnchain transaction
+                reward_addr
+                    .as_clarity_tuple()
+                    .expect("FATAL: stack-stx operation has no hash mode")
+                    .into(),
+                Value::UInt(u128::from(*block_height)),
+                Value::UInt(u128::from(*num_cycles)),
+            ];
+            // Appending additional signer related arguments for pox-4
+            if POX_4_NAME == active_pox_contract {
```
kantai, line 4169: We should start using the New and Improved `PoxVersions` enum:

```suggestion
            if active_pox_contract == PoxVersions::Pox4.get_name() {
```

kantai (dismissed):
LGTM -- just one comment trying to spread the good news about the PoxVersions enum.

jferrant (approved):

kantai (approved):

jferrant (approved):

jcnelson (approved):

Issue #4413, opened 2024-02-22
Add `cycle-id` ranges to synthetic pox events


zone117x:
The signer data sent to event observers only contains a list of `signer_keys`, which doesn't contain information like the stacker principals associated with those keys. IIUC this is because there is a many-to-many relationship between stacker principals and signer keys that is difficult/fuzzy to calculate for the event data.

This requires event observers (like the Stacks Blockchain API) to match the signer key event data against historical synthetic pox events in order to determine the mapping of signers and associated stacking data like principals, delegation, etc.

Adding a `start-cycle-id` and `end-cycle-id` to the synthetic pox events would make logic in the event observer this much easier.

PR #4414, opened 2024-02-22
Add `cycle-id` ranges to synthetic pox events

zone117x:
Fixes https://github.com/stacks-network/stacks-core/issues/4413

WIP:
* [x] Add `start-cycle-id` to pox events
* [x] Add `end-cycle-id` to pox events
* [x] Tests


kantai:
Yes -- the existing `pox-4` tests would invoke these paths. Those tests should be expanded to check the event receipts though.

pox-locking/src/events.rs
```diff
@@ -195,11 +225,16 @@ fn create_event_info_data_code(
...
+                        end-cycle-id: (burn-height-to-reward-cycle unlock-height),
+                        ;; Get start cycle ID
+                        start-cycle-id: (+ (burn-height-to-reward-cycle unlock-height) prepare-offset),
                     }}
-                }}
+                }})
                 "#,
-                increase_by = &args[0]
+                increase_by = &args[0],
+                prepare_offset = prepare_offset.replace("%height%", "unlock-height"),
```
zone117x, line 237: With https://github.com/stacks-network/stacks-core/pull/4461 merging, we should also add signer-key to this `stacks-increase` event

zone117x: https://github.com/stacks-network/stacks-core/issues/4490

pox-locking/src/events.rs
```diff
@@ -378,7 +440,9 @@ fn create_event_info_data_code(
                         unlock-burn-height: {until_burn_height},
                         ;; optional PoX address tuple.
                         ;; equal to args[3].
-                        pox-addr: {pox_addr}
+                        pox-addr: {pox_addr},
+                        ;; Get end cycle ID
+                        end-cycle-id: (burn-height-to-reward-cycle {until_burn_height}),
```
8marz8, line 445: While testing I realized `burn-height-to-reward-cycle` doesn't expect an optional so when passing `until_burn_height` (which is an optional), it leads to a parsing issue and doesn't spit out an event for delegate-stx....just to prove this to myself, I used below (which is not proper at all given the default needs to be something else)
` until_burn_height_val = &args[2].clone().expect_optional().map_or(Value::UInt(0), |v| v.unwrap()),
`

So, if we want `end-cycle-id` to be present in all events, I'm not sure how else it can be calculated so we don't rely on an optional? 
Unless `until_burn_height` is optional for fork/update reasons but in current Stack we're always guaranteed to have Some value.

hstove: Great catch! I think we'll need to return an optional for `end-cycle-id` in this case. I haven't really tested this code, but I think this should work
```suggestion
                        end-cycle-id: (match {until_burn_height} 
                            height (some (burn-height-to-reward-cycle height))
                            none),
```

pox-locking/src/events.rs
```diff
@@ -111,18 +111,30 @@ fn create_event_info_data_code(
...
+            (mod (- %height% (var-get first-burnchain-block-height)) (var-get pox-reward-cycle-length))
+            (- (var-get pox-reward-cycle-length) (var-get pox-prepare-cycle-length))
+        ) u0 u1))
+    "#;
+
     match function_name {
         "stack-stx" => {
             format!(
                 r#"
+                (let (
```
8marz8, line 126: I think something might be off with these let bindings here...I'm getting below error for `stack-stx` and haven't been able to fix. (Possible it's from my testing too)

> ERRO [1709783156.742952] [pox-locking/src/pox_4.rs:352] [chainstate::stacks::boot::pox_4_tests::pox_4_cycle_id_events] Failed to synthesize PoX-4 event info: Unchecked(UndefinedVariable("current-pox-reward-cycle"))

zone117x: That's strange, it's a top-level function: https://github.com/stacks-network/stacks-core/blob/4e0222d61ad3f73a48e57c42acb21daf0dd15c43/stackslib/src/chainstate/stacks/boot/pox-4.clar#L250-L251

There's also some existing synthetic pox code using it: https://github.com/stacks-network/stacks-core/blob/4e0222d61ad3f73a48e57c42acb21daf0dd15c43/pox-locking/src/events_24.rs#L115

pox-locking/src/events.rs
```diff
@@ -136,28 +148,37 @@ fn create_event_info_data_code(
                         signer-sig: {signer_sig},
                         ;; equal to args[5]
                         signer-key: {signer_key},
+                        ;; Get end cycle ID
+                        end-cycle-id: (burn-height-to-reward-cycle unlock-burn-height),
+                        ;; Get start cycle ID
+                        start-cycle-id: (+ current-pox-reward-cycle prepare-offset),
```
8marz8, line 154: It was `current-pox-reward-cycle` missing () here

8marz8:
I took @hstove 's suggestion and chaned `end-cycle-id` to an optional since it was relying on an optional (`until_burn_height`) for calculation in `delegate-stx`.

The test ensures that the events are emitted without Clarity parsing issues BUT, I am not sure about the values I am asserting against at all - I basically filled those from what was being returned so @zone117x @hstove @kantai please check if those values actually make sense for what was intended here.

zone117x:
Thanks for the help with this @8marz8! I'm going to test this in an end-to-end environment because I'm also not totally sure if the cycle ids in the tests are accurate or not.

pox-locking/src/events.rs
```diff
@@ -210,19 +240,28 @@ fn create_event_info_data_code(
                         max-amount: {max_amount},
                         ;; equal to args[4]
                         auth-id: {auth_id},
+                        ;; Get end cycle ID
+                        end-cycle-id: (some (burn-height-to-reward-cycle unlock-height)),
+                        ;; Get start cycle ID
+                        start-cycle-id: (+ (burn-height-to-reward-cycle unlock-height) prepare-offset),
```
hstove, line 246: I think this is wrong - this is saying the start cycle is the cycle in which the stacker unlocks

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -1289,6 +1290,402 @@ fn pox_3_unlocks() {
...
+    let bob_aggregation_commit_tx_events = &bob_aggregation_commit_tx.unwrap().clone().events;
+    assert_eq!(bob_aggregation_commit_tx_events.len() as u64, 1);
+    let bob_aggregation_commit_tx_event = &bob_aggregation_commit_tx_events[0];
+    let bob_aggregation_commit_tx_op_data = HashMap::from([
+        ("start-cycle-id", Value::UInt(22)),
+        (
+            "end-cycle-id",
+            Optional(OptionalData {
+                data: Some(Box::from(Value::UInt(0))), //Is this supposed to be 0?!
+            }),
```
hstove, line 1672: There are helper functions for this to make it a bit easier. In this case I am a bit confused that this works because it uses Rust optionals along with Clarity values

```suggestion
            Value::some(Value::Uint(0)),
```

hstove: Secondly, this is 0 because when calling `stx-account`, `unlock-height` is zero if there is no fixed unlock height.

pox-locking/src/events.rs
```diff
@@ -281,6 +326,10 @@ fn create_event_info_data_code(
                         max-amount: {max_amount},
                         ;; equal to args[5]
                         auth-id: {auth_id},
+                        ;; Get end cycle ID
+                        end-cycle-id: (some (burn-height-to-reward-cycle new-unlock-ht)),
```
hstove, line 330: In this case you don't need the `some` because you'll always be returning a value

hstove (changes requested):
@8marz8 so for the [check against Steph's `stack-stx`](https://github.com/stacks-network/stacks-core/pull/4414/files#diff-1b7e3da4025244bdde5f977c36554b94ab89e844c05e4c3c0a106dd9da9c7636R1540-R1556), the assertion is that `start-cycle-id` is 22 and `end-cycle-id` is 24. Steph made the transaction during 22, with a lock period of 1. So she would start stacking in 23, and end at the start of 24. This might be ok if we said "the user starts stacking in the cycle after `start-cycle-id`, but we just need to be consistent with these events. This is also kinda subjective, I'm curious of @zone117x 's thoughts.

Then for the [`stack-increase` check](https://github.com/stacks-network/stacks-core/pull/4414/files#diff-1b7e3da4025244bdde5f977c36554b94ab89e844c05e4c3c0a106dd9da9c7636R1562-R1577), both the start and end are 24. When calling stack increase, you don't change your cycles at all, you're just adding more to the existing cycles. So in my opinion, this should match the stack-stx behavior, especially because the tx is made in the same cycle as the stack-stx tx.

This is very close though, nice job!

zone117x:
> the assertion is that start-cycle-id is 22 and end-cycle-id is 24. Steph made the transaction during 22, with a lock period of 1. So she would start stacking in 23, and end at the start of 24. This might be ok if we said "the user starts stacking in the cycle after start-cycle-id, but we just need to be consistent with these events. This is also kinda subjective, I'm curious of @zone117x 's thoughts.

In this case I think we'd want `start-cycle-id` to be 23, because we are using these IDs to match against reward set data. So IIUC we don't care about when stx are _locked_, we care about when stx are committed. Does that make sense, or is my understanding/terminology correct here?

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
zone117x: In this case, we'd expect `end-cycle-id` to be `none` right?

pox-locking/src/events.rs
zone117x: It's a bit nice to have type consistency here (e.g. `end-cycle-id` is always an `optional`), rather than a union type like `uint | Optional(Some(uint))`, but I don't feel strongly. The API code ingesting this data currently expects a uniform `Optional`.

janniks: The optionals are a bit cumbersome. Not really a big deal though. However, the `none` is only needed for "delegation" case, which doesn't impact stacking/locking until the actual `delegate-stack-...` calls. So this could be simplified -- I kept the `some` wrapping for now.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
janniks: Refactored to simpler definitions

pox-locking/src/events.rs
janniks: Agree +1
I updated these -- imo the prepare_offset can always only come from `burn_block_height`, and in some cases is not needed at all.

janniks:
✅ Updated this PR to match my understanding of what the range queries should achieve from a consumer perspective (be able to use event data for reward/voting-power calculations). I hope I haven't misinterpreted anything here.

- Update test to use dynamic cycle id and lock periods -- using the assumption: start-cycle-id == next_cycle by default, (if not in prepare phase case)
- Update `aggregation` functions to have start&end == {reward_cycle} (they don't span multiple cycles)
- Update prepare_offsets to always be based on current burn_height _(this could be hardcoded into the prepare snippet)_
- Removed let/prepare_offset blocks where not needed
- **Increase** methods will use the normal current burn_height for "start" and re-use the previous "end". (_**Careful**: this will generate [contained] overlapping ranges, either use the never one or modify the existing entry on ingest_)
- **Extend** methods are the same but use the new unlock-ht for "end". (_**Careful:**_ same warnings apply, but with an [protruding] overlap here)

**todo**: will add a few tests to cover more cases with the prepare-offset

pox-locking/src/events.rs
hstove: I'm good with consistency, let's keep the optional 👍

hstove (approved):
LGTM. Nice improvements and tests @janniks !

hstove (approved):

kantai (approved):

PR #4425, opened 2024-02-23
chore(stacks-common): Improvements to `Vec` allocation and operations

jbencin:
### Description

Similar to #4408, but for `./stacks-common`. This PR also removes unnecessary intermediate `Vec`s used by `Vec::append()` by using `Vec::extend_from_slice()` instead

This gives a significant improvement of around **5%**

```console
❯ hyperfine -w 3 -r 10 "stacks-core/target/release/stacks-inspect.next replay-block /home/jbencin/data/next/ range 99990 100000" "stacks-core/target/release/stacks-inspect.vec-allocation-stacks-common replay-block /home/jbencin/data/next/ range 99990 100000"
  Time (mean ± σ):      9.044 s ±  0.020 s    [User: 8.556 s, System: 0.442 s]
  Range (min … max):    9.022 s …  9.074 s    10 runs
 
Benchmark 2: stacks-core/target/release/stacks-inspect.vec-allocation-stacks-common replay-block /home/jbencin/data/next/ range 99990 100000
  Time (mean ± σ):      8.649 s ±  0.040 s    [User: 8.157 s, System: 0.441 s]
  Range (min … max):    8.597 s …  8.724 s    10 runs
 
Summary
  stacks-core/target/release/stacks-inspect.vec-allocation-stacks-common replay-block /home/jbencin/data/next/ range 99990 100000 ran
    1.05 ± 0.01 times faster than stacks-core/target/release/stacks-inspect.next replay-block /home/jbencin/data/next/ range 99990 100000
```

### Applicable issues

- #4316

Same as #4408

### Checklist

- [ ] Changelog is updated

### DIFF
--- a/stacks-common/src/address/c32.rs
+++ b/stacks-common/src/address/c32.rs
@@ -180,7 +180,9 @@ const C32_CHARACTERS_MAP: [Option<u8>; 128] = [
 ];
 
 fn c32_encode(input_bytes: &[u8]) -> String {
-    let mut result = vec![];
+    // c32-encoded size is 160%  that of ASCII
+    let size = input_bytes.len().saturating_mul(8).div_ceil(5);
+    let mut result = Vec::with_capacity(size);
     let mut carry = 0;
     let mut carry_bits = 0;
 
@@ -234,10 +236,6 @@ fn c32_decode(input_str: &str) -> Result<Vec<u8>, Error> {
 }
 
 fn c32_decode_ascii(input_str: &str) -> Result<Vec<u8>, Error> {
-    let mut result = vec![];
-    let mut carry: u16 = 0;
-    let mut carry_bits = 0; // can be up to 5
-
     let mut iter_c32_digits = Vec::<u8>::with_capacity(input_str.len());
 
     for x in input_str.as_bytes().iter().rev() {
@@ -250,6 +248,12 @@ fn c32_decode_ascii(input_str: &str) -> Result<Vec<u8>, Error> {
         // at least one char was None
         return Err(Error::InvalidCrockford32);
     }
+
+    // ASCII size is 62.5%  that of c32-encoded
+    let size = iter_c32_digits.len().saturating_mul(5).div_ceil(8);
+    let mut result = Vec::with_capacity(size);
+    let mut carry: u16 = 0;
+    let mut carry_bits = 0; // can be up to 5
 
     for current_5bit in &iter_c32_digits {
         carry += (*current_5bit as u16) << carry_bits;
--- a/stacks-common/src/address/c32.rs
+++ b/stacks-common/src/address/c32.rs
@@ -180,7 +180,8 @@ const C32_CHARACTERS_MAP: [Option<u8>; 128] = [
 ];
 
 fn c32_encode(input_bytes: &[u8]) -> String {
-    // c32-encoded size is 160%  that of ASCII
+    // ASCII characters are 8-bits and c32-encoding encodes 5-bits per
+    // character, so the c32-encoded size should be ceil((ascii size) * 8 / 5)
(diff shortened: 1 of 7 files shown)


stacks-common/src/address/c32.rs
```diff
@@ -180,7 +180,9 @@ const C32_CHARACTERS_MAP: [Option<u8>; 128] = [
 ];
 
 fn c32_encode(input_bytes: &[u8]) -> String {
-    let mut result = vec![];
+    // c32-encoded size is 160%  that of ASCII
```
obycode, line 183: Saying 160% here is not obvious. I think it would be more clear to say something like:
```rust
    // ASCII characters are 8-bits and c32-encoding encodes 5-bits per
    // character, so the c32-encoded size should be (ascii size) * 8 / 5.
```

stacks-common/src/address/c32.rs
```diff
@@ -251,6 +249,12 @@ fn c32_decode_ascii(input_str: &str) -> Result<Vec<u8>, Error> {
         return Err(Error::InvalidCrockford32);
     }
 
+    // ASCII size is 62.5%  that of c32-encoded
```
obycode, line 252: Similarly here, this could say:

```rust
    // c32-encoding encodes 5 bits into each character, while ASCII encodes
    // 8-bits into each character. So, the ASCII-encoded size should be
    // (c32 size) * 5 / 8.
```

stacks-common/src/deps_common/bitcoin/blockdata/script.rs
```diff
@@ -201,8 +202,8 @@ fn build_scriptint(n: i64) -> Vec<u8> {
 
     let neg = n < 0;
 
-    let mut abs = if neg { -n } else { n } as usize;
-    let mut v = vec![];
+    let mut abs = n.abs() as usize;
+    let mut v = Vec::with_capacity(size_of::<usize>() / 8);
```
obycode, line 206: The `size_of` is in bytes. I don't think the `/ 8` should be here.

stacks-common/src/util/hash.rs
```diff
@@ -451,9 +451,11 @@ where
...
     /// Get a non-leaf hash
     pub fn get_node_hash(left: &H, right: &H) -> H {
-        let mut buf = vec![];
-        buf.extend_from_slice(left.bits());
-        buf.extend_from_slice(right.bits());
+        let left_bits = left.bits();
+        let right_bits = right.bits();
+        let mut buf = Vec::with_capacity(left_bits.len() + right_bits.len());
+        buf.extend_from_slice(left_bits);
+        buf.extend_from_slice(right_bits);
```
obycode, line 458: How about combining it all at once?

```rust
        let buf = left.bits().iter().chain(right.bits().iter()).cloned().collect();
```

stacks-common/src/deps_common/bitcoin/blockdata/script.rs
```diff
@@ -201,8 +202,8 @@ fn build_scriptint(n: i64) -> Vec<u8> {
 
     let neg = n < 0;
 
-    let mut abs = if neg { -n } else { n } as usize;
-    let mut v = vec![];
+    let mut abs = n.abs() as usize;
+    let mut v = Vec::with_capacity(size_of::<usize>() + 2);
```
obycode, line 206: Oh, good call, I almost missed that additional push, but it should be just 1, right? The loop handles all but the last byte, then the last byte gets handled, along with an optional extra byte for negative value.

jbencin: Oh, I didn't notice that that loop always leaves one byte

obycode (approved):
lgtm!

jcnelson (approved):

PR #4432, opened 2024-02-27
Additional 'set-signer-key-authorization' Checks

setzeus:
### Description
While writing up the testing document for PoX-4/Boot contracts I believe I stumbled upon two checks that should be added to the 'set-signer-key-authorization' function recently added to PoX-4.clar.

The point of this authorization is for stackers to not have to provide the optional signature. However, the way it's currently written, it's possible for the signer to create malformed/invalid authorizations that would result in a seemingly-valid authorization but would later fail when stackers attempt to use it. 

Specifically, a 'period' of length 0 can be passed in; additionally, a previous reward-cycle can be submitted.

Still need input from @hstove before removing 'draft' status.

### Checklist
- [ ] Changelog is updated

### DIFF
--- a/stackslib/src/chainstate/stacks/boot/pox-4.clar
+++ b/stackslib/src/chainstate/stacks/boot/pox-4.clar
@@ -32,6 +32,7 @@
 (define-constant ERR_DELEGATION_ALREADY_REVOKED 34)
 (define-constant ERR_INVALID_SIGNATURE_PUBKEY 35)
 (define-constant ERR_INVALID_SIGNATURE_RECOVER 36)
+(define-constant ERR_INVALID_REWARD_CYCLE 37)
 
 ;; Valid values for burnchain address versions.
 ;; These first four correspond to address hash modes in Stacks 2.1,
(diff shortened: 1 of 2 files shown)


hstove (dismissed):
Thanks, these checks make sense and are good additions!

hstove (approved):
Re-approving after rebase

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -116,6 +117,9 @@
 ;;    (6) (reward-cycle-to-burn-height (+ lock-period first-reward-cycle)) == (get unlock-height (stx-account stacker))
 ;;  These invariants only hold while `cur-reward-cycle < (+ lock-period first-reward-cycle)`
 ;;
+(define-map protocols principal (list 10000 uint))
+(define-map protocol-positions {protocol: principal, position-id: uint} {status: (string-ascii 32), collateral: uint})
+
```
MarvinJanssen, line 122: Looks like this does not belong here?

MarvinJanssen (dismissed):
Looks good. Just noticed code that does not seem to belong in PoX 4.

stackslib/src/chainstate/stacks/boot/pox-4.clar
setzeus: Oh yikes. Yeah that's throwaway code, removing now, thanks.

setzeus: Removed.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -32,6 +32,7 @@
 (define-constant ERR_DELEGATION_ALREADY_REVOKED 34)
 (define-constant ERR_INVALID_SIGNATURE_PUBKEY 35)
 (define-constant ERR_INVALID_SIGNATURE_RECOVER 36)
+(define-constant ERR_INVALID_REWARD_CYCLE 37)
```
hstove, line 35: Can we re-use `ERR_STACKING_INVALID_LOCK_PERIOD` ?

hstove (changes requested):

hstove (approved):

MarvinJanssen (approved):

Issue #4387, opened 2024-02-16
[Nakamoto] Logic for transactions verification in both the signer and miner should be revamped


jferrant:
Currently the signer and miner do some minimal checks to prevent a single malicious signer spamming the block. This should be improved. In theory, a signer can only vote once per round and it should be only during DKG. Do not include any signer transactions when NOT in the prepare phase OR if the DKG for the reward cycle of that signer set has already been approved. Otherwise, only include valid cast aggregate key votes that are wellformed. 

Additionally, would be better if this logic was a util function if possible (might not be as the miner uses its own db's/stackerdb to get the info it needs while the signer uses the RPC endpoint plus stackerdb. Some of the verify payload logic though could be globally accessible)

Open Question: Should the miner and signers enforce only one transaction allowed in a single block for a signer address?

PR #4436, opened 2024-02-27
Move signer transaction to global location and utilize in both miner and signer

jferrant:
I simplified the transaction logic and made it the same between signers and miners. It simply filters invalidly formed transactions from the list of signer transactions, i.e. non vote for aggregate public key transactions, and invalidly formed vote for aggregate public key transactions. It no longer asserts that the actual values of the function args are as expected, leaving it up to the clarity contract to reject them. Additionally, it asserts that only one transaction is allowed per signer per block, sorting proposed transactions based first upon nonce, secondly upon txid if the nonce is the same. 

### Applicable issues

- Closes #https://github.com/stacks-network/stacks-core/issues/4387

### DIFF
--- a/stacks-signer/src/signer.rs
+++ b/stacks-signer/src/signer.rs
@@ -839,8 +839,8 @@ impl Signer {
                 OperationResult::SignTaproot(_) => {
                     debug!("Signer #{}: Received a signature result for a taproot signature. Nothing to broadcast as we currently sign blocks with a FROST signature.", self.signer_id);
                 }
-                OperationResult::Dkg(point) => {
-                    self.process_dkg(stacks_client, point);
+                OperationResult::Dkg(dkg_public_key) => {
+                    self.process_dkg(stacks_client, dkg_public_key);
                 }
                 OperationResult::SignError(e) => {
                     warn!("Signer #{}: Received a Sign error: {e:?}", self.signer_id);
@@ -855,7 +855,7 @@ impl Signer {
     }
 
     /// Process a dkg result by broadcasting a vote to the stacks node
-    fn process_dkg(&mut self, stacks_client: &StacksClient, point: &Point) {
+    fn process_dkg(&mut self, stacks_client: &StacksClient, dkg_public_key: &Point) {
         let epoch = retry_with_exponential_backoff(|| {
             stacks_client
                 .get_node_epoch()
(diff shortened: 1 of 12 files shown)


stacks-signer/src/signer.rs
```diff
@@ -982,12 +855,7 @@ impl Signer {
     }
 
     /// Process a dkg result by broadcasting a vote to the stacks node
-    fn process_dkg(
-        &mut self,
-        stacks_client: &StacksClient,
-        point: &Point,
-        current_reward_cycle: u64,
-    ) {
+    fn process_dkg(&mut self, stacks_client: &StacksClient, point: &Point) {
```
xoloki, line 858: Can you give `point` a better name?  I have no idea looking at this function what it might be.

xoloki (approved):
One minor parameter naming nit, otherwise LGTM!

stacks-signer/src/signer.rs
jferrant: Will do

xoloki (dismissed):
LGTM

stackslib/src/chainstate/nakamoto/signer_set.rs
```diff
@@ -487,4 +487,84 @@ impl NakamotoSigners {
...
+        if origin_nonce < *account_nonce {
+            debug!("valid_vote_transaction: Received a transaction with an outdated nonce ({account_nonce} < {origin_nonce}).");
+            return false;
+        }
+        Self::parse_vote_for_aggregate_public_key(transaction).is_some()
+    }
+
+    pub fn parse_vote_for_aggregate_public_key(
+        transaction: &StacksTransaction,
+    ) -> Option<(u64, Point, u64, u64)> {
```
kantai, line 518: In general, I think large tuple function signatures should be avoided because they can be prone to errors (the type checker won't help you differentiate between one u64 and another), and I think its better to use one-off structs if necessary:

```suggestion
    pub struct AggregateKeyVoteParams {
        pub signer_index: u64,
        pub aggregate_key: Point,
        pub voting_round: u64,
        pub reward_cycle: u64,
    }
    pub fn parse_vote_for_aggregate_public_key(
        transaction: &StacksTransaction,
    ) -> Option<AggregateKeyVoteParams> {```

kantai (approved):
This LGTM, just had one suggestion in the refactor.

stackslib/src/chainstate/nakamoto/signer_set.rs
jferrant: Done :)

xoloki (approved):

kantai (approved):

PR #4437, opened 2024-02-27
[Refactor] Refactor of Clarity storage (iteration 1)

cylewitruk:
❗❗❗ **This PR draft is currently being wrapped-up and is <u>NOT READY FOR CONSIDERATION**</u> ❗❗❗

**Remaining Work:**
- [ ] More mainnet sync tests
- [ ] Some additional unit tests are needed, particularly surrounding the Clarity cache
- [ ] Finish DB migration implementation and tests
- [ ] Performance benchmarking (w/hyperfine)
- [ ] Storage benchmarking (size-on-disk)
- [x] Create issues for changes

## 👋 Introduction
This work was born from a combination of the A/B-tester I have recently been working on and the open umbrella issue #4316. When testing the A/B-tester it became clear that replaying blocks up to more recent blocks in-between each bugfix takes way too much time, so I started investigating ways to optimize specifically the Clarity VM, with a focus on the `oracle-v1` contracts. Needless to say, there are a number of optimizations which can be done. While there have been other initiatives looking for low-hanging-fruits, I decided to try tackling some deeper optimizations.

The primary goal of this PR is to lay some groundwork for refactoring & optimizing the persistence layer of the Stacks node, with focus on the Clarity database. I decided to start with two rather large data-structures which are, today, being serialized using `serde_json` and written/read to SQLite as plain-text JSON. These structures are read + deserialized a number of times during contract execution.

As seems to be usual for my core refactoring attempts, this started out as a "small and isolated" change which quickly cascaded onto a much larger surface area. The PR in itself is quite large, but I hope that the changes are clear - I did make an effort on documentation :)

## ⛓ Consensus
While this PR does _potentially affect_ Clarity consensus-critical storage, no _behavior_ [should have] been changed (i.e. consensus keys, serialization, etc.). Therefore this PR should be seen as non-consensus-critical --  however **an extra eye on the consensus-critical code paths is warranted**, just in-case I missed something that tests aren't catching.

## ✅️ Issues Resolved in this PR

- #4444
- #4449
  - 📌 _This issue touches a lot of files, but mostly due to adding `speedy`'s `Readable` and `Writable` derives to types to be (de)serialized._
- #4450
- #4448
- [trivial] #4446
  - 📌 _This issue touches a lot of files, but mostly due to adding `fake`'s `Dummy` derive to types to be faked._
- [trivial] #4447
- [trivial] #4440
- [trivial] #4445
- [trivial] #4441
- [trivial] #4442
- [trivial] #4443

## 🚩 Additional Changes & Improvements
- The `AnalysisDatabase` type has been removed and baked into the `ClarityDatabase` struct instead. This is motivated by the fact that these are two different interfaces working against the same database and tables, and thus moving such logic together helps to reduce the perceived complexity in preparation for a larger refactor. It would have been another story IMO if there had been a significant number of analysis-related methods we were talking about -- but since it's only a few, the additional abstraction only adds complexity and confusion.
  - 📌 _This change touches a lot of files, but primarily due to the renaming of `AnalysisDatabase` to `ClarityDatabase` and `analysis_db` to `clarity_db`._

- This PR **makes it an error** to attempt to insert a contract multiple times into the backing store, even at different block hashes. Previously there was no check for this, allowing duplicate contracts to be deployed at different block heights, potentially without the caller realizing it. This situation needs to be explicitly handled now.

## 🧪 Testing
📌 _A lot of `/tests/*.rs` etc. files have been touched due to renames and test-tweaking._
- Unit tests have been written for new code.
  -  ℹ️ Some areas are still uncovered -- I'm working on more tests for them.
- Added new tests for existing code, particularly surrounding the `RollbackWrapper`.
- Existing tests have been adapted to work with the new paradigm, with a few "gotcha's":
  - The clarity cache (#4444) is <u>disabled by default</u> in `test` and `dev` modes. This is because too many tests fail otherwise which are relying on very specific control of what exists/doesn't exist in the underlying database.
  - A lot of tests use methods which now expect the contract and/or its analysis to exist in the underlying database, which now failed due to the use of `ClarityDatabase::test_insert_contract_hash()` -- this method only ensured that the contract hash existed in the MARF, so it has been changed to one of `test_insert_contract()` or `test_insert_contract_with_analysis` in said tests. These two methods have also been added to tests prior to their "need" of this stored data. Related to #4448.
  - A number of tests expect to be able to insert the same contract multiple times, which is now an error. I've adapted these tests as well.
  - ⚠️ Please put extra focus on reviewing all adapted tests -- it is possible that I misunderstood what the test was trying to accomplish, or simply made a mistake.
- All tests in the workspace are passing.
- A release-mode node has been partially synced to mainnet without issue.
  - ℹ️ I'm working on more testing here.

## ⏰ Performance
The following [benchmark](https://github.com/cylewitruk/stacks-node-db-bench/blob/main/crates/stacks-bench/benches/clarity.rs) shows a simulated single contract write / read in the current `next` vs. this branch. This benchmark was performed on a modern Intel CPU with an NVMe drive. Note that these benchmarks use disk-based databases and `optimized` includes both `speedy` serialization _and_ `lz4_flex` compression, where `next` uses only `serde_json` serialization. No caching is used.

|              | `next`                    | `optimized`                       |
|:-------------|:--------------------------|:--------------------------------- |
| **`insert`** | `672.04 us` (✅ **1.00x**) | `237.78 us` (🚀 **2.83x faster**)  |
| **`select`** | `295.95 us` (✅ **1.00x**) | `117.13 us` (🚀 **2.53x faster**)  |

ℹ️ _More to come with hyperfine + stacks-inspect replay benchmarking_

## 🚩 New Dependencies
- [fake](https://crates.io/crates/fake) - Used to help with filling structs with random (but format-correct) data in tests. This is where the `Dummy` attribute littered throughout this PR comes from. _Currently_ added as a _dependency_ and not _dev-dependency_ because I had issues with conditional compiles in test/dev mode -- but it should work to get this into `dev-dependencies` with a little more effort. All code using `fake` is conditional on `cfg!(test)`, so it should be optimized away during a build anyway.
- [lru](https://crates.io/crates/lru) - Used in global caching of Clarity data, including contracts and analyses. Note that `rusqlite` uses this crate, and specifically its `LruCache` struct, internally.
- [lz4_flex](https://crates.io/crates/lz4_flex) - A pure-Rust implementation of the LZ4 compression algorithm. Currently set to use safe-only code, but `default-features = false` enables its "performance mode" with unsafe code.
- [speedy](https://crates.io/crates/speedy) - A very fast binary serializer written in pure Rust. This is where the `Readable` and `Writable` attributes littered throughout this PR come from. This is only used for node-specific, non-consensus data (i.e. data which was previously stored in the Clarity `metadata_table` SQLite table):
  - Serialized `ContractContext`
  - Serialized `ContractAnalysis`

## 💡 Some Potential Next Steps
This PR sets the stage for further refactoring the Clarity (and other) database(s). Some ideas for next steps are things like (but not limited to):
- Optimizing how Clarity **maps** are stored. Maps are currently JSON-serialized in their entirety and written/read to/from the `metadata_table` database table. By normalizing this structure we could perform much more efficient, targeted reads and writes for specific map keys.
- Clarity **lists** are another contender with the same motivations as above. Lists could be normalized and operations like `index-of`, `element-at`, `slice` etc. could be delegated to the underlying database engine, which specializes in query logic, selecting only the items needed for a particular operation.
- Clarity **functions** can be normalized to allow fast lookup and validation of contract functions prior to loading the entire contract for a contract call.
- Ensuring all binary data is stored as binary instead of its hexadecimal string representation.
- Changing serialization of the remainder of non-consensus-critical node-local data to use `speedy` instead of `serde_json`.
- Employ compression on stored Clarity values which are large.
- Moving on to optimizing some of the other databases, with focus on migrating binary stored as hexidecimal strings to raw bytes.


clarity/src/vm/docs/mod.rs
```diff
@@ -2865,65 +2865,6 @@ mod test {
 
         let contract_id = QualifiedContractIdentifier::local("docs-test").unwrap();
 
-        {
```
cylewitruk, line 2868: This code was causing problems with contract existence/non-existence and it seems to be doing the same thing as the "more modern" code below - can anybody confirm that this removal is OK?

cylewitruk:
I realize that this PR may look intimidating. Now that I've taken the time to wrap my head around all of the changes and box-in the different moving parts it's easier for me (mentally) to break this into smaller PR's if preferred -- in which case I would break it up as follows:

1) The following, in no particular order:
- #4441
- #4442
- #4443
- #4445
- #4446
2) These issues **together**:
- #4448
- #4449
- #4450
3) Followed by these, in no particular order:
- #4444
- #4440

While #4447 could be incrementally applied depending on the code changes.

cylewitruk:
> I'm looking at the `speedy` docs and it looks like the purpose of this crate to read/write to a buffer with as little overhead as possible. Does it contain any sort of versioning info or other metadata? What happens if in the future we add/remove/reorder fields in a serializable struct?

Great question!

When it comes to `speedy`, you can add and remove fields, but:
- to be able to deserialize a `v1 { name }` into a `v2 { name, phone }` then `phone` (the added field) would need the `#[speedy(default_on_eof)]` attribute so say it's OK if it's empty.
- nothing is needed to deserialize from a `v2 { name, phone }` into a `v1 { name }`.
- however, ordering **must be maintained**.
Meaning that if there were any ordering changes then a migration would need to be written.

> **EDIT:** Have you considered `rkyv`? It's similar (fast, minimalistic binary protocol, not "self describing", lacks schema evolution), but much more widely used (several popular crates support it, such as `hashbrown`) and has some additional features, like a limited form of validation

Yeah I had a look at `rkyv` as well - it's a bit less straightforward, but I'd be more than happy to add a bench and test suite for adding/removing/reordering like I did with speedy :+1:

jbencin:
> I'd be more than happy to add a bench and test suite for adding/removing/reordering

There are some nice benchmarks of Rust serialization frameworks [here](https://github.com/djkoloski/rust_serialization_benchmark). Both are similar in performance, and either one would be much faster than JSON, so the choice should depend on other aspects

> it doesn't look like there was really any thought into forwards-/backwards-compatibility in the existing [Stacks] code, but then again field names are serialized

JSON is "self-describing" in that the field names, types (to some extent), as well as the overall structure is encoded into the output data. So even if you try to deserialize data into structures that don't quite match, it's easy to ignore unknown fields, use default values if a field is missing, or fail with a descriptive error

On the other hand, many of these fast binary formats get their speed by omitting the extra metadata. Instead, they make the assumption that the application knows exactly how the data should be structured. This limits portability between languages, CPU architectures, and even different versions of the application. Sometimes, the trade-off worth it, depending on the use-case. What I'm specifically concerned about here is:
- If something changes in the binary format...
  - Is there any chance of silent data corruption, like blindly `memcpy`ing binary data into the wrong variables?
  - In the event of an explicit failure, do we get meaningful error messages, like which struct/field it failed at?
- Would this be more brittle than the current JSON encoding? Would we have to rebuild the chainstate for changes that would be backwards compatible if we were using JSON?

There are also some fast binary serialization formats that are portable and support versioning, like Cap’n Proto or FlatBuffers, but they work by generating code from a custom schema language, so there's significant added complexity (much more work than adding a `#[derive(...)]`). Probably not something we want to invest in right now

cylewitruk:
> There are some nice benchmarks of Rust serialization frameworks [here](https://github.com/djkoloski/rust_serialization_benchmark). Both are similar in performance, and either one would be much faster than JSON, so the choice should depend on other aspects

Yeah, this is where I've looked as a source for encoders in Rust. 

I specifically ruled out frameworks which are schema-based -- while that would be great for consensus-critical messaging, it's way too much overhead for node-local data, especially given the polymorphic nature of `TypeSignature`s,  `SymbolicExpression`s and `Value`s, all which are heavily used and can recurse within eachother.

Realistically I would probably choose MessagePack over anything else for a good balance between speed and rigidity, largely because I have very positive experiences with it in a production setting. Another contender is [CBOR](https://cbor.io/), which is largely based on MessagePack, but the encoders I tried were closer to JSON speeds when benchmarked with the above types. BSON is another widely-used format, but it's basically just a binary form of JSON and is neither known for its speed nor compactness.



> > it doesn't look like there was really any thought into forwards-/backwards-compatibility in the existing [Stacks] code, but then again field names are serialized

By this I was referring more to the fact that business objects are being directly serialized instead of storage-specific models, which is dangerous for the exact reasons you've pointed out.

> On the other hand, many of these fast binary formats get their speed by omitting the extra metadata. Instead, they make the assumption that the application knows exactly how the data should be structured. This limits the validation you can do, and the portability between languages, CPU architectures, and even different versions of the application. Sometimes, the trade-off worth it, depending on the use-case.

For this particular case we're dealing with node-local data, i.e. data the node itself generates during the processing of blocks, so short of the Hiro Archive, portability isn't a large issue. I would personally rather add CLI commands to import/export chainstate in a well-defined format (where performance isn't an issue) which could work across all future node implementations -- i.e. a node written in C# or Java would most likely choose the encoding, backing database and data structures that works best within _their_ ecosystems for local storage.

> Is there any chance of silent data corruption, like blindly memcpying binary data into the wrong variables?

Speedy actually had a pretty OK safeguard solution for this with their `#[speedy(constant_prefix = $expr)]` attribute, where we could assign each field a short identifier, and deserialization will fail with an "expected constant xxx" error if the field it's trying to deserialize into's prefix doesn't match the value's prefix. But frankly, this is a potential risk in any sort of encoding, even hand-written, so the best thing we can do is ensure good test coverage+fuzzing regardless which solution we end up with.

> In the event of an explicit failure, do we get meaningful error messages, like which struct/field it failed at?

If using the above attribute, the "expected constant xxx" gives you a hint of which field has failed.

> Would this be more brittle than the current JSON encoding?

Absolutely -- but that will be the story with any performant binary encoding. Ensuring good test coverage of (de)serialization will be important to ensure no changes accidentally break the encoding/decoding (this is where `fake` is great as it makes it easy to do lightweight fuzzing in unit tests).

> Would we have to rebuild the chainstate for changes that would be backwards compatible if we were using JSON?

The risk of needing to re-encode (migrate) values is much higher, yes -- but that also implies changes to the above structures. The entire chainstate would not have to be re-built, i.e. the node doesn't need to re-process blocks; it would be a data migration of specific values in the database which should take minutes, not hours. This is the price we'll pay for moving to a more performant binary encoding -- changes to these types are extremely rare, however; the last change was during 2.1's hard fork iirc.

Could we move this discussion over to #4449 so that it stays with the serialization ticket? :smile:

jcnelson:
Closing due to inactivity

PR #4451, opened 2024-02-29
Fix signer id use in stackerdb

jferrant:
In my massive PR, I aaccidentally conflated signer id and signer slot id. this did not break tests because we happened to have everything line up (odds are we ALWAYS will have everything line up since the reward set is used to buidl stackerdb, but this will make sure we do not rely on this)

### DIFF
--- a/stacks-signer/src/client/mod.rs
+++ b/stacks-signer/src/client/mod.rs
@@ -154,6 +154,7 @@ pub(crate) mod tests {
 
     use super::*;
     use crate::config::{GlobalConfig, ParsedSignerEntries, SignerConfig};
+    use crate::signer::SignerSlotID;
 
     pub struct MockServerClient {
         pub server: TcpListener,
@@ -456,7 +457,7 @@ pub(crate) mod tests {
                     &StacksPublicKey::from_slice(ecdsa_public_key.to_bytes().as_slice())
                         .expect("Failed to create stacks public key"),
                 );
-                signer_slot_ids.push(signer_id); // Note in a real world situation, these would not always match
+                signer_slot_ids.push(SignerSlotID(signer_id));
                 signer_ids.insert(address, signer_id);
 
                 continue;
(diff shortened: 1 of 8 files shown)


stacks-signer/src/client/mod.rs
```diff
@@ -459,7 +456,7 @@ pub(crate) mod tests {
                     &StacksPublicKey::from_slice(ecdsa_public_key.to_bytes().as_slice())
                         .expect("Failed to create stacks public key"),
                 );
-                signer_slot_ids.insert(address, signer_id); // Note in a real world situation, these would not always match
+                signer_slot_ids.push(signer_id); // Note in a real world situation, these would not always match
```
xoloki, line 459: Can you update the comment?  Previously it looked like the comment meant that `address` and `signer_id` would not always match.

stacks-signer/src/client/mod.rs
```diff
@@ -486,7 +483,7 @@ pub(crate) mod tests {
                 &StacksPublicKey::from_slice(public_key.to_bytes().as_slice())
                     .expect("Failed to create stacks public key"),
             );
-            signer_slot_ids.insert(address, signer_id); // Note in a real world situation, these would not always match
+            signer_slot_ids.push(signer_id); // Note in a real world situation, these would not always match
```
xoloki, line 486: Same

xoloki (dismissed):
LGTM, one small question about a comment.

stacks-signer/src/client/mod.rs
jferrant: No this comment is accurate. It means that signer id and signer slot id would not always match

stacks-signer/src/client/mod.rs
jferrant: See other comment.

hstove (approved):

kantai (dismissed):

kantai (approved):

PR #4455, opened 2024-02-29
Round up thresholds 

jferrant:
@hstove found a bug when operating a single signer with a single key. Essentially the threshold became 0 which causes wsts to panic as it should due to an invalid threshold being provided. @xoloki subsequently pointed out that we actually had a bug in our threshold calculation due to integer arithmatic forcing us to always calculate a threshold "under" our desired amount whenever we had any remainder. This pr forces us to "round up" by adding 0.9 to the result. If any remainder exists, it will force our result over into the next integer amount (note that due to integer math and the nature of dividing by 10, the remainder will always be minimum 0.1 or more hence why adding 0.9 guarantees pushing it over to the next integer if remainder is present).

Due to Joey's suggestion, I have converted all integers involved to floating points instead to use the ceil function.

### DIFF
--- a/stacks-signer/src/signer.rs
+++ b/stacks-signer/src/signer.rs
@@ -156,9 +156,8 @@ impl From<SignerConfig> for Signer {
             .expect("FATAL: Too many registered signers to fit in a u32");
         let num_keys = u32::try_from(signer_config.registered_signers.public_keys.key_ids.len())
             .expect("FATAL: Too many key ids to fit in a u32");
-        // Always add +0.9 to force any remainder to round up to the next integer
-        let threshold = (num_keys * 7 + 9) / 10;
-        let dkg_threshold = (num_keys * 9 + 9) / 10;
+        let threshold = (num_keys as f64 * 7_f64 / 10_f64).ceil() as u32;
+        let dkg_threshold = (num_keys as f64 * 9_f64 / 10_f64).ceil() as u32;
 
         let coordinator_config = CoordinatorConfig {
             threshold,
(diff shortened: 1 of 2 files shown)


xoloki (approved):
While I'd prefer a floating point based calculation that explicitly calls `ceil` or some similar function, this is certainly an improvement on what we currently have.

hstove:
My main concern is how this interacts with the Clarity code for voting on the aggregate key: https://github.com/stacks-network/stacks-core/blob/db9616238dc40ba94ca380238a4b8da8fbc04fbf/stackslib/src/chainstate/stacks/boot/signers-voting.clar#L161-L162

If we round down in Clarity, and up in the signer, what happens?

jferrant:
> My main concern is how this interacts with the Clarity code for voting on the aggregate key:
> 
> https://github.com/stacks-network/stacks-core/blob/db9616238dc40ba94ca380238a4b8da8fbc04fbf/stackslib/src/chainstate/stacks/boot/signers-voting.clar#L161-L162
> 
> If we round down in Clarity, and up in the signer, what happens?

Nothing bad happens. It means clarity accepts a certain number of votes less than in the wsts calculation side. However, no one will get a result on the wsts side unless they reach the correct threshold. These are ultimately independent thresholds, though its not a bad idea to make them consistent, the clarity one is arbitrary.

kantai (approved):

hstove:
Ok, I think I follow. So basically, wsts and the signers are using this threshold to compute a specific threshold that relates to generating a valid signature for that public key, and then Clarity is using its own threshold to determine how many votes are needed to activate it as the aggregate key. That makes sense.

I can imagine some weird edge cases but nothing worth blocking this

hstove (approved):

xoloki (changes requested):
LGTM!

PR #4458, opened 2024-02-29
chore: Optimize `TupleData::from_data()`

jbencin:
### Description

Profiling showed that a considerable amount of time was spent in the function `TupleData::from_data()`. I took a look at the function and there was an easy optimization to make that would reduce the number of `BTreeMap` lookups by half. This results in a measurable performance increase when processing blocks of around `1%`:

```console
❯ hyperfine -w 3 -r 20 "target/release/stacks-inspect.next replay-block /home/jbencin/data/next/ range 99990 100000" "target/release/stacks-inspect.tuple-data replay-block /home/jbencin/data/next/ range 99990 100000"
Benchmark 1: target/release/stacks-inspect.next replay-block /home/jbencin/data/next/ range 99990 100000
  Time (mean ± σ):      8.716 s ±  0.014 s    [User: 8.234 s, System: 0.442 s]
  Range (min … max):    8.698 s …  8.747 s    20 runs
 
Benchmark 2: target/release/stacks-inspect.tuple-data replay-block /home/jbencin/data/next/ range 99990 100000
  Time (mean ± σ):      8.593 s ±  0.024 s    [User: 8.119 s, System: 0.432 s]
  Range (min … max):    8.563 s …  8.635 s    20 runs
 
Summary
  target/release/stacks-inspect.tuple-data replay-block /home/jbencin/data/next/ range 99990 100000 ran
    1.01 ± 0.00 times faster than target/release/stacks-inspect.next replay-block /home/jbencin/data/next/ range 99990 100000
```

### Applicable issues

- #4316


clarity/src/vm/types/mod.rs
```diff
@@ -1527,11 +1529,11 @@ impl TupleData {
         let mut data_map = BTreeMap::new();
         for (name, value) in data.into_iter() {
             let type_info = TypeSignature::type_of(&value)?;
-            if type_map.contains_key(&name) {
-                return Err(CheckErrors::NameAlreadyUsed(name.into()).into());
-            } else {
-                type_map.insert(name.clone(), type_info);
-            }
+            let entry = type_map.entry(name.clone());
```
cylewitruk, line 1532: Just a small style nit - you could simplify this slightly:
```rust
match type_map.entry(name.clone()) {
  ....
}
```

cylewitruk (approved):
One style nit but otherwise lgtm!

kantai (approved):

Issue #4422, opened 2024-02-23
Add validation for STX amount in signer key authorizations


hstove:
In pox-4 we have authorizations (via pre-set authorizations or via signatures) around usage of a signer key, but those authorizations do not cover the amount of STX used.

These authorizations should include a new field, `amount-ustx`, which indicates the **max** amount of STX that can be used by a particular reward set.

Also related is #4417, which adds `signer-key` to stack-increase. That ticket will need to also validate that the new amount of STX locked (after the increase) does not exceed the `amount-ustx` included in the authorization.

hstove:
Linking discussion: https://github.com/stacks-network/stacks-core/discussions/4424

Issue #4417, opened 2024-02-23
Add `signer-key` to `*-increase` pox events


zone117x:
The `signer-key` has effectively become the “primary key” for all signer data. So if I’m trying to lookup information about stackers based on `signer-key`, it gets pretty difficult when important events like `stack-increase` and associated events don’t include it.

Based on discussion with @kantai, it doesn't sound possible to add a `map` that could be used to lookup the `signer-key` during synthetic event generation for these functions. Another approach would be to just add the signer key and signature args to the functions. It sounds like this might be a good idea to do anyway -- from Aaron:
> functions like stack-increase do not take a signing key argument because they are updating an existing entry.
> maybe they should, because I could certainly see an argument that those functions actually also need signer-key authorization.
> like, imagine a signer authorizes a stacker to use their signing key to stack 100 STX, and then later on that stacker invokes stacks-increase to increase from 100 STX to 10000 STX.

hstove:
After looking at `stack-increase`, during part of the function it iterates over `reward-cycle-pox-address-list`, which includes `signer-key` in each entry. So it's possible to do the same thing when making synthetic events, but it would need to return a list of signer keys (one for each updated cycle).

hstove:
Linking discussion: https://github.com/stacks-network/stacks-core/discussions/4424

setzeus:
Discussing w/ ClarityWG it seems like the best / most practical solution here is to limit 'stack-increase' to users that are straddling two signer-keys between upcoming cycles (aka they did not 'stack-extend' recently). 

That way every check is straight-forward with every iteration just checking that the existing signer-key is equal to the signer-key in the param.

hstove:
Yep, I think that's the right approach and aligns with @kantai 's comments on expected behavior

setzeus:
Did work over the weekend on this that can be found here: https://github.com/stacks-network/stacks-core/compare/feat/signer-key-amount-auth...feat/signer-key-amount-auth-stack-increase

Reviewed with @hstove who'll 'git cherry-pick' commits into his main PR found here: https://github.com/stacks-network/stacks-core/pull/4461

I'll close issue once the PR linked above is merged-in.

PR #4461, opened 2024-02-29
feat: Validate amount with signer key authorizations

hstove:
This PR implements some of the changes discussed in https://github.com/stacks-network/stacks-core/discussions/4424.

- Closes https://github.com/stacks-network/stacks-core/issues/4422
- Closes https://github.com/stacks-network/stacks-core/issues/4417

The primary change is the addition of two fields to signer key authorization messages, which are validated either via a signature or a previously stored authorization. These fields are `max-amount` and `auth-id`. `max-amount` reflects the maximum amount of STX that can be used **in that specific stacking action**. The `auth-id` field is used to prevent replay attacks, which would allow a stacker to go beyond `max-amount` over multiple transactions.

Note that `stack-extend` doesn't verify the `amount` field in the authorization, because there are no additional STX added. In that function, the primary purpose of the authorization is around the number of reward cycles being extended.

`stack-increase` is also changed. It now includes the `signer-key`, along with the signature and related fields. Along with the same signature / authorization verifications that other stacking functions have, this function also checks to ensure that the `signer-key` parameter matches the `signer-key` for each updated cycle. **This means that you cannot call `stack-increase` if you've called `stack-extend` with a different signer key**.

In `pox-4` there is also a new map, `used-signer-key-authorizations`, which tracks whether a `(signer-key, pox-addr, reward-cycle, period, topic, max-amount, auth-id)` tuple has been used.


stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -231,13 +234,33 @@
         period: uint,
         ;; A string representing the function where this authorization is valid. Either
         ;; `stack-stx`, `stack-extend`, or `agg-commit`.
-        topic: (string-ascii 12),
+        topic: (string-ascii 14),
```
jferrant, line 237: Just out of curousity...why does the topic string need to increase in length from 12 to 14?

kantai (approved):

stackslib/src/chainstate/stacks/boot/pox-4.clar
hstove: It was only to allow for the whole string `stack-increase`, but definitely not strictly necessary. But this line points out that I need to update some of the documentation in this file (like this line).

hstove:
@kantai I just re-requested review after pushing some documentation improvements

stackslib/src/chainstate/stacks/boot/pox-4.clar
jferrant: Ah yeah this makes sense. With the comment updated, totally get it.

jferrant (dismissed):
Nice! :D extensive testing is great.

kantai (dismissed):

setzeus (approved):
This is some great work here @hstove, good job on the thorough tests in pox_4_tests. Couldn't see anything in pox-4 or related works that called my attention.

LGTM.

zone117x (commented):
If it's not too much of a hassle, could you add signer key to the pox4 `increase` synthetic event? If that's a pain we can do in a later PR

kantai (approved):

hstove:
@zone117x good call and I should have caught that earlier. Let's get this merged and I made a new ticket: https://github.com/stacks-network/stacks-core/issues/4490

jferrant (approved):

PR #4462, opened 2024-03-01
chore(clarity): Allocate `HashMap`s with capacity when possible

jbencin:
### Description

Like `Vec`, `HashMap` can also be initialized using `with_capacity()`. I made this change where it was possible in `./clarity` and saw around a **2%** performance gain:

```console
❯ hyperfine -w 3 -r 10 "target/release/stacks-inspect.next replay-block /home/jbencin/data/next/ range 99990 100000" "target/release/stacks-inspect.hashmap-alloc replay-block /home/jbencin/data/next/ range 99990 100000"
Benchmark 1: target/release/stacks-inspect.next replay-block /home/jbencin/data/next/ range 99990 100000
  Time (mean ± σ):      8.682 s ±  0.019 s    [User: 8.200 s, System: 0.435 s]
  Range (min … max):    8.651 s …  8.710 s    10 runs
 
Benchmark 2: target/release/stacks-inspect.hashmap-alloc replay-block /home/jbencin/data/next/ range 99990 100000
  Time (mean ± σ):      8.552 s ±  0.026 s    [User: 8.082 s, System: 0.425 s]
  Range (min … max):    8.510 s …  8.586 s    10 runs
 
Summary
  target/release/stacks-inspect.hashmap-alloc replay-block /home/jbencin/data/next/ range 99990 100000 ran
    1.02 ± 0.00 times faster than target/release/stacks-inspect.next replay-block /home/jbencin/data/next/ range 99990 100000
```

### Applicable issues

- #4316

### DIFF
--- a/clarity/src/vm/costs/mod.rs
+++ b/clarity/src/vm/costs/mod.rs
@@ -811,9 +811,10 @@ impl TrackerData {
 
         self.contract_call_circuits = contract_call_circuits;
 
-        let mut cost_contracts = HashMap::new();
-        let mut m = HashMap::new();
-        for f in ClarityCostFunction::ALL.iter() {
+        let iter = ClarityCostFunction::ALL.iter();
+        let mut cost_contracts = HashMap::with_capacity(iter.len());
+        let mut m = HashMap::with_capacity(iter.len());
+        for f in iter {
             let cost_function_ref = cost_function_references.remove(f).unwrap_or_else(|| {
                 ClarityCostFunctionReference::new(boot_costs_id.clone(), f.get_name())
             });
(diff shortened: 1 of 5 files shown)


clarity/src/vm/costs/mod.rs
```diff
@@ -811,9 +811,10 @@ impl TrackerData {
 
         self.contract_call_circuits = contract_call_circuits;
 
-        let mut cost_contracts = HashMap::new();
-        let mut m = HashMap::new();
-        for f in ClarityCostFunction::ALL.iter() {
+        let iter = ClarityCostFunction::ALL.iter();
```
cylewitruk, line 814: Probably better to put the `len()` in a variable instead of calling it twice - no guarantees the backing type has an efficient `len()` implementation.

cylewitruk (dismissed):
One comment but otherwise lgtm!

clarity/src/vm/costs/mod.rs
jbencin: Good point. This is also why Clippy complains when you use `var.len() == 0` instead of `var.is_empty()`

jcnelson (approved):

cylewitruk (approved):
lgtm!

PR #4465, opened 2024-03-01
Chore: Replace Hex Encoding/Decoding

ASuciuX:
### Description

I took a look at the functions that were doing the hex encoding and decoding in the `stacks-common` package and there was an easy replacement with an external crate (`faster-hex`) in order to improve the performance. The results show a performance increase of about `1%` when processing blocks:
```sh
$ hyperfine -w 3 -r 10 "./stacks-inspect-hex-modifications replay-block ~/chainstate/ range 99990 100000" "./stacks-inspect-9385b4236 replay-block ~/chainstate/ range 99990 100000"
Benchmark 1: ./stacks-inspect-hex-modifications replay-block ~/chainstate/ range 99990 100000
  Time (mean ± σ):      7.956 s ±  0.036 s    [User: 7.571 s, System: 0.352 s]
  Range (min … max):    7.929 s …  8.052 s    10 runs
 
Benchmark 2: ./stacks-inspect-9385b4236 replay-block ~/chainstate/ range 99990 100000
  Time (mean ± σ):      8.023 s ±  0.021 s    [User: 7.610 s, System: 0.379 s]
  Range (min … max):    7.992 s …  8.061 s    10 runs
 
Summary
  ./stacks-inspect-hex-modifications replay-block ~/chainstate/ range 99990 100000 ran
    1.01 ± 0.01 times faster than ./stacks-inspect-9385b4236 replay-block ~/chainstate/ range 99990 100000
```

### DIFF
--- a/stacks-common/src/util/hash.rs
+++ b/stacks-common/src/util/hash.rs
@@ -19,6 +19,7 @@ use std::convert::TryInto;
 use std::fmt::Write;
 use std::{fmt, mem};
 
+use faster_hex::{hex_decode, hex_string};
 use ripemd::Ripemd160;
 use serde::de::{Deserialize, Error as de_Error};
 use serde::ser::Error as ser_Error;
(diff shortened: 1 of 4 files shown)


stacks-common/src/util/hash.rs
```diff
@@ -588,26 +589,17 @@ where
...
-    // Check that there was no remainder
-    match iter.remainder() {
-        Some(_) => Err(HexError::BadLength(s.len())),
-        None => Ok(v),
-    }
+    let mut bytes = vec![0u8; s.len() / 2];
+    match hex_decode(s.as_bytes(), &mut bytes) {
+        Ok(()) => {},
+        Err(e) => match e {
+            faster_hex::Error::InvalidChar => return Err(HexError::BadCharacter(' ')),
```
jbencin, line 597: We should just propagate their error type rather than convert to our own

jbencin (dismissed):
Looks good. Smaller improvement than I expected. It's up to Aaron and Jude if they are okay introducing a dependency for this

zone117x:
Might also consider the [`hex-simd`](https://crates.io/crates/hex-simd) package which is ~2-3x faster than `faster-hex` https://nugine.github.io/simd-benches/

jcnelson:
Per the call today, I think we should defer further consideration of this until we remove hex encode/decode from the DB I/O paths (since that's where a lot of this happens, and in this case, we can just remove the hex codec altogether).

ASuciuX:
> Per the call today, I think we should defer further consideration of this until we remove hex encode/decode from the DB I/O paths (since that's where a lot of this happens, and in this case, we can just remove the hex codec altogether).

I'll convert this into `draft` till then.

jcnelson:
Closing due to inactivity

PR #4469, opened 2024-03-01
Remove self signer

obycode:
Note that for now, the blind-signer binary is completely untested and probably incomplete, but the main goal here is to remove the self-signer.


blind-signer/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+    let reward_cycle = burnchain
+        .block_height_to_reward_cycle(tip.block_height)
+        .unwrap();
+    let prepare_phase_start = burnchain
+        .pox_constants
+        .prepare_phase_start(burnchain.first_block_height, reward_cycle);
+
+    if tip.block_height >= prepare_phase_start {
+        // If the key is already set, do nothing.
+        if is_key_set_for_cycle(reward_cycle + 1, conf.is_mainnet(), &conf.node.rpc_bind)
```
jferrant, line 155: Nit: bit easy to forget this +1 throughout. Would assign to a var called "next_reward_cycle" just so we don't accidentally remove one/forget this +1 where appropriate.

jferrant:
I do wonder if this basically is just like running the current signer as a single signer especially since it still seems to use stackerdb/has the need to register a signer with the typical stack-stx function call unless I am missings omething?

obycode:
Ok, I reverted the refactoring so now this PR just has the removal of the self-signer with the blind signing in the tests.

jferrant (dismissed):
LGTM! :D

kantai (dismissed):

jferrant (dismissed):

zone117x (approved):

zone117x (approved):

jferrant (approved):

kantai (approved):

PR #4470, opened 2024-03-01
Add property tests for PoX-4 read-only functions

BowTiedRadone:
### Description
This PR adds property tests for PoX-4 read-only functions.


setzeus (changes requested):
Great job on these tests! They seem p straight-forward read-only checks since there's no way to mock stacking here. 

I only have one blocking comment atm.

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,434 @@
...
+            "check-pox-addr-version",
+            [Cl.bufferCV(Uint8Array.from([version]))],
+            caller
+          );
+
+          // Assert
+          assert(
+            isClarityType(
+              actual,
+              expected ? ClarityType.BoolTrue : ClarityType.BoolFalse
```
setzeus, line 281: I'm a bit more unfamiliar w/ Vitest than I'd like but this ternary throws me off? If we set "expected" in line 267 why is there a ternary in the asserts statement?

Couldn't that return a false positive / always succeed? Or am I misunderstanding the purpose here?

setzeus (changes requested):
Great job on these tests! They seem p straight-forward read-only checks since there's no way to mock stacking here.

I only have one blocking comment atm.

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
BowTiedRadone: That assert basically checks that the type of the `actual` result is the right one. If the randomly generated version is greater than 6, then the `actual`'s clarity type should be false. Else, it should be true. I'll split this into 2 separated tests to avoid the ternary inside the assertion.

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1527 @@
...
+const TESTNET_PREPARE_CYCLE_LENGTH = 50;
+const INITIAL_TOTAL_LIQ_SUPPLY = 1_000_000_000_000_000;
+const MIN_AMOUNT_USTX = 125_000_000_000n;
+// Clarity Constraints
+const MAX_CLAR_UINT = 340282366920938463463374607431768211455n;
+// Error Codes
+const ERR_STACKING_INVALID_LOCK_PERIOD = 2;
+const ERR_STACKING_THRESHOLD_NOT_MET = 11;
+const ERR_STACKING_INVALID_POX_ADDRESS = 13;
+const ERR_STACKING_INVALID_AMOUNT = 18;
```
moodmosaic, line 38: It's nice that all those are grouped (e.g. consts, error codes, etc.) and how about also sorting them alphabetically?

BowTiedRadone: I'll do it🫡

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+  );
+
+  const signer_key_message_hash = sha256(
+    Buffer.concat([structuredDataPrefix, domain_hash, message_hash])
+  );
+
+  return signer_key_message_hash;
+};
+
+describe("test pox-4 contract", () => {
```
moodmosaic, line 99: vitest reports the name of the file so this `describe` isn't needed. If we remove it we'll save 2 spaces of indentation.

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+        )
+      );
+    });
+
+    it("should return none get-stacker-info", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.constantFrom(...simnet.getAccounts().values()),
+          (stacker, caller) => {
```
moodmosaic, line 207: What happens is `stacker` and `caller` are the same account?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+          }
+        )
+      );
+    });
+
+    it("should return false check-pox-addr-version for version > 6", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 7, max: 255 }),
```
moodmosaic, line 376:
```suggestion
          fc.integer({ min: 7 }),
```

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+        )
+      );
+    });
+
+    it("should return true check-pox-lock-period for valid reward cycles number", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 1, max: 12 }),
+          (caller, reward_cycles) => {
```
moodmosaic, line 400:
```suggestion
          (caller, valid_reward_cycles) => {
```

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+            // Arrange
+            const expected = true;
+            // Act
+            const { result: actual } = simnet.callReadOnlyFn(
+              "pox-4",
+              "check-pox-lock-period",
+              [Cl.uint(reward_cycles)],
```
moodmosaic, line 407:
```suggestion
              [Cl.uint(valid_reward_cycles)],
```

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+        )
+      );
+    });
+
+    it("should return false check-pox-lock-period for reward cycles number > 12", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 13 }),
+          (caller, reward_cycles) => {
```
moodmosaic, line 423:
```suggestion
          (caller, invalid_reward_cycles) => {
```

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+            // Arrange
+            const expected = false;
+            // Act
+            const { result: actual } = simnet.callReadOnlyFn(
+              "pox-4",
+              "check-pox-lock-period",
+              [Cl.uint(reward_cycles)],
```
moodmosaic, line 430:
```suggestion
              [Cl.uint(invalid_reward_cycles)],
```

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+      );
+    });
+
+    it("should return false check-pox-lock-period for reward cycles number == 0", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          (caller) => {
+            // Arrange
+            const reward_cycles = 0;
```
moodmosaic, line 447:
```suggestion
            const invalid_reward_cycles= 0;
```

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+            const expected = false;
+            // Act
+            const { result: actual } = simnet.callReadOnlyFn(
+              "pox-4",
+              "check-pox-lock-period",
+              [Cl.uint(reward_cycles)],
```
moodmosaic, line 453:
```suggestion
              [Cl.uint(invalid_reward_cycles)],
```

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+    });
+
+    it("should return (ok true) can-stack-stx for versions 0-4 valid pox addresses, hashbytes, amount, cycles number", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 0, max: 4 }),
+          fc.array(fc.nat({ max: 255 }), {
+            minLength: 20,
+            maxLength: 20,
```
moodmosaic, line 471: Are the array bounds really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+    });
+
+    it("should return (ok true) can-stack-stx for versions 5/6 valid pox addresses, hashbytes, amount, cycles number", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 5, max: 6 }),
+          fc.array(fc.nat({ max: 255 }), {
+            minLength: 32,
+            maxLength: 32,
```
moodmosaic, line 528: Are the array bounds really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({
+            min: 7,
+            max: 255,
+          }),
+          fc.array(fc.nat({ max: 255 }), {
+            minLength: 32,
+            maxLength: 32,
```
moodmosaic, line 588: Are the array bounds really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+      );
+    });
+
+    it("should return (err 13) can-stack-stx for pox addresses having version > 6", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({
+            min: 7,
+            max: 255,
```
moodmosaic, line 584: Is `max: 255` needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+    });
+
+    it("should return (err 13) can-stack-stx for versions 0-4 pox addresses having hasbytes longer than 20", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 0, max: 4 }),
+          fc.array(fc.nat({ max: 255 }), {
+            minLength: 21,
+            maxLength: 32,
```
moodmosaic, line 645: Are the array bounds really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+      );
+    });
+
+    it("should return (err 13) can-stack-stx for versions 0-4 pox addresses having hasbytes shorter than 20", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 0, max: 4 }),
+          fc.array(fc.nat({ max: 255 }), {
+            maxLength: 19,
```
moodmosaic, line 701: Is the array bound really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+      );
+    });
+
+    it("should return (err 13) can-stack-stx for versions 5/6 pox addresses having hashbytes shorter than 32", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 0, max: 4 }),
+          fc.array(fc.nat({ max: 255 }), {
+            maxLength: 31,
```
moodmosaic, line 757: Is the array bound really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+    });
+
+    it("should return (ok true) minimal-can-stack-stx for versions 0-4 valid pox addresses, hashbytes, amount, cycles number", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 0, max: 4 }),
+          fc.array(fc.nat({ max: 255 }), {
+            minLength: 20,
+            maxLength: 20,
```
moodmosaic, line 922: Are the array bounds really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+    });
+
+    it("should return (ok true) minimal-can-stack-stx for versions 5/6 valid pox addresses, hashbytes, amount, cycles number", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 5, max: 6 }),
+          fc.array(fc.nat({ max: 255 }), {
+            minLength: 32,
+            maxLength: 32,
```
moodmosaic, line 979: Are the array bounds really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({
+            min: 7,
+            max: 255,
+          }),
+          fc.array(fc.nat({ max: 255 }), {
+            minLength: 32,
+            maxLength: 32,
```
moodmosaic, line 1039: Are the array bounds really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+    });
+
+    it("should return (err 13) minimal-can-stack-stx for versions 0-4 pox addresses having hasbytes longer than 20", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 0, max: 4 }),
+          fc.array(fc.nat({ max: 255 }), {
+            minLength: 21,
+            maxLength: 32,
```
moodmosaic, line 1096: Are the array bounds really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+      );
+    });
+
+    it("should return (err 13) minimal-can-stack-stx for versions 0-4 pox addresses having hasbytes shorter than 20", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 0, max: 4 }),
+          fc.array(fc.nat({ max: 255 }), {
+            maxLength: 19,
```
moodmosaic, line 1152: Is the array bound really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+      );
+    });
+
+    it("should return (err 13) minimal-can-stack-stx for versions 5/6 pox addresses having hashbytes shorter than 32", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 0, max: 4 }),
+          fc.array(fc.nat({ max: 255 }), {
+            maxLength: 31,
```
moodmosaic, line 1208: Is the array bound really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+
+    it("should return (err 2) minimal-can-stack-stx for lock period > 12", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 0, max: 6 }),
+          fc.array(fc.nat({ max: 255 })),
+          fc.bigInt({
+            min: MIN_AMOUNT_USTX,
+            max: MAX_CLAR_UINT,
```
moodmosaic, line 1266: Are the array bounds really needed?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+      );
+    });
+
+    it("should return none get-allowance-contract-caller", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.constantFrom(...simnet.getAccounts().values()),
+          (caller, sender, contract_caller) => {
```
moodmosaic, line 1443: What happens if any of the caller, sender, and contract_caller are the same?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+      );
+    });
+
+    it("should return some(until-burn-ht: none) get-allowance-contract-caller after allow-contract-caller", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.constantFrom(...simnet.getAccounts().values()),
+          (caller, sender, contract_caller) => {
```
moodmosaic, line 1465: What happens if any of the caller, sender, and contract_caller are the same?

moodmosaic (commented):
Great work. Left a couple of comments, mostly on the way some generators are configured.

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
BowTiedRadone: Nothing happens. The reason there are two generated values is to cover both the case when the `caller` and the `stacker` argument are the same and the one they are identical. Should I keep only the `caller` one?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
BowTiedRadone: Here, if I allow values greater than 255 (2^8-1), the UInt8Array will convert them to 8 bytes unsigned integer. E.g. when 266 is generated, it is converted to 0 and the function call returns `true`

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
BowTiedRadone: Yes, they are needed because otherwise the function will return err. 
This happens because of this assertion from the pox-4 SC:
```
(asserts! (check-pox-addr-version (get version pox-addr))(err ERR_STACKING_INVALID_POX_ADDRESS))
```

This is not inside `can-stack-stx`, but in `minimal-can-stack-stx` which is called by the tested method.

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
BowTiedRadone: Same as the previous comment.

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
BowTiedRadone: The lower bound is needed because it is the specific tested case. The upper bound is needed because the clarity argument is (buff 32), and if the 32 length is exceeded the contract call will fail.

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
BowTiedRadone: It is needed because of the test case.

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
BowTiedRadone: Sane as the previous comment

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
BowTiedRadone: Here, I will remove the lower bound.

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1605 @@
...
+        )
+      );
+    });
+
+    it("should return (err 18) minimal-can-stack-stx for amount == 0", () => {
+      fc.assert(
+        fc.property(
+          fc.constantFrom(...simnet.getAccounts().values()),
+          fc.integer({ min: 0, max: 6 }),
+          fc.array(fc.nat({ max: 255 }), { maxLength: 32 }),
```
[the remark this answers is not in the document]
BowTiedRadone: Yes, because of the type (buff 32) param

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
BowTiedRadone: Nothing, it just does not make sense that the `sender` and the `contract_caller` are the same in the context of pox-4 allowance. The `allow-contract-caller` method is called to allow a pool contract or another principal to stack on behalf of a principal.

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
BowTiedRadone: Same as the previous comment

BowTiedRadone:
@hugocaillard
Regarding `checks`, one of the test fails, even if it works locally. What to do in this situation?
https://github.com/stacks-network/stacks-core/actions/runs/8202455709/job/22433699088?pr=4470

It is [this test](https://github.com/BowTiedRadone/stacks-core/blob/feat/pox-4-property-tests/contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts#L1456-L1486) that first allows the contract caller and then checks that the result is some. Maybe the simnet does not work as expected when run there.

moodmosaic (approved):
LGTM 🙌

moodmosaic:
@setzeus can you do another round of review as well?

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1838 @@
...
+            caller
+          );
+          // Assert
+          assert(isClarityType(actual, ClarityType.OptionalNone));
+        }
+      )
+    );
+  });
+
+  it("should return some(until-burn-ht: none) get-allowance-contract-caller after allow-contract-caller", () => {
```
setzeus, line 1492: This is failing for me locally. Seeing a "TypeError" issue: Object.entries(...).toSorted is not a function.

setzeus (changes requested):
Overall this looks great! Just getting one test failing locally seen below, once that's updated/fixed this should be good to go.

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
moodmosaic: Perhaps Node.js/npm/TypeScript issue 🤔

BowTiedRadone: It is the same the CI is throwing. Not sure how to handle it or where it is coming from. No error for me locally.

BowTiedRadone: Just pushed an update. I have added another assertion before the failing expect. Hope this will solve the error.

moodmosaic: Thanks! However this still seem to fail locally. What's your environment? (Node.js version?)

moodmosaic: This looks like a bug in the assertion library, it's thrown here:

```
    1519|           expect(actual.value).toBeTuple({ "until-burn-ht": Cl.none() });
       |                                ^
```

@hugocaillard, perhaps you have an idea?

In case it helps, the dependencies used are:

```
    "@hirosystems/clarinet-sdk": "^1.1.0",
    "@stacks/transactions": "^6.12.0",
    "chokidar-cli": "^3.0.0",
    "@stacks/clarunit": "0.0.1",
    "fast-check": "^3.15.1",
    "typescript": "^5.2.2",
    "vite": "^4.4.9",
    "vitest": "^0.34.4",
    "vitest-environment-clarinet": "^1.0.0"
```

Here's the strack trace:

```
Got TypeError: Object.entries(...).toSorted is not a function
    at Object.toBeTuple (contrib/core-contract-tests/node_modules/@hirosystems/clarinet-sdk/vitest-helpers/src/clarityValuesMatchers.ts:232:54)
    at Proxy.expectWrapper (contrib/core-contract-tests/node_modules/@vitest/expect/dist/index.js:1368:40)
    at Proxy.<anonymous> (contrib/core-contract-tests/node_modules/@vitest/expect/dist/index.js:653:17)
    at Proxy.methodWrapper (contrib/core-contract-tests/node_modules/chai/lib/chai/utils/addMethod.js:57:25)
    at contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts:1519:32
    at Property.predicate (contrib/core-contract-tests/node_modules/fast-check/lib/esm/check/property/Property.js:14:54)
    at Property.run (contrib/core-contract-tests/node_modules/fast-check/lib/esm/check/property/Property.generic.js:46:33)
    at runIt (contrib/core-contract-tests/node_modules/fast-check/lib/esm/check/runner/Runner.js:18:30)
    at check (contrib/core-contract-tests/node_modules/fast-check/lib/esm/check/runner/Runner.js:62:11)
    at Module.assert (contrib/core-contract-tests/node_modules/fast-check/lib/esm/check/runner/Runner.js:65:17)
```

contrib/core-contract-tests/tests/pox-4/pox-4.prop.test.ts
```diff
@@ -0,0 +1,1839 @@
...
+            "pox-4",
+            "get-allowance-contract-callers",
+            [Cl.principal(sender), Cl.principal(contract_caller)],
+            caller
+          );
+          // Assert
+          assert(isClarityType(actual, ClarityType.OptionalSome));
+          assert(isClarityType(actual.value, ClarityType.Tuple));
+          expect(actual).toBeSome(Cl.tuple({ "until-burn-ht": Cl.none() }));
+          expect(actual.value).toBeTuple({ "until-burn-ht": Cl.none() });
```
moodmosaic, line 1519: This line causes the issue reported by @setzeus above. However, this line can be safely removed since the exact same assertion happens also in the line above.

(What's the reason for having the same assertion twice?)
```suggestion

```

@hugocaillard, this also resolves the question in https://github.com/stacks-network/stacks-core/pull/4470/files#r1519481753 so feel free to skip/ignore it.

moodmosaic (changes requested):

moodmosaic (approved):
LGTM

moodmosaic:
> Overall this looks great! Just getting one test failing locally seen below, once that's updated/fixed this should be good to go.

@setzeus failing test has been fixed.

setzeus (approved):
Can confirm no more errors, appreciate the hard work here @BowTiedRadone  - LGTM!

hugocaillard:
Looking at this now. I assume this comment https://github.com/stacks-network/stacks-core/pull/4470#issuecomment-1985660276 is no longer relevant right?

moodmosaic:
> Looking at this now. I assume this comment [#4470 (comment)](https://github.com/stacks-network/stacks-core/pull/4470#issuecomment-1985660276) is no longer relevant right?

It could be an issue with the test assertion, and not an actual bug. I would say don't spend time on it (unless you actually see/spot a potential bug looking at the stack trace).

hugocaillard:
In [this job](https://github.com/stacks-network/stacks-core/actions/runs/8202455709/job/22433699088?pr=4470), the error comes from `.toSorted() is not a function`. That's because a certain version of the sdk used `toSorted` that is only available in Node >= 20. It has been fixed, you need to use the latest version of @hirosystems/clarinet-sdk and @stacks/transactions to avoid that

PR #4471, opened 2024-03-01
Feat/block proposal authorization password for v2/block_proposal endpoint

jferrant:
Added config option to the signer and to the ConnectionOptions for adding a password. Just passing it around as a plaintext password in the authorization header.

### DIFF
--- a/stackslib/src/net/connection.rs
+++ b/stackslib/src/net/connection.rs
@@ -415,6 +415,8 @@ pub struct ConnectionOptions {
     /// the reward cycle in which Nakamoto activates, and thus needs to run both the epoch
     /// 2.x and Nakamoto state machines.
     pub force_nakamoto_epoch_transition: bool,
+    /// The authorization token to enable the block proposal RPC endpoint
+    pub block_proposal_token: Option<String>,
 }
 
 impl std::default::Default for ConnectionOptions {
@@ -508,6 +510,7 @@ impl std::default::Default for ConnectionOptions {
             disable_stackerdb_get_chunks: false,
             force_disconnect_interval: None,
             force_nakamoto_epoch_transition: false,
+            block_proposal_token: None,
         }
     }
 }
--- a/stacks-signer/src/cli.rs
+++ b/stacks-signer/src/cli.rs
@@ -173,6 +173,9 @@ pub struct GenerateFilesArgs {
     /// The number of milliseconds to wait when polling for events from the stacker-db instance.
     #[arg(long)]
     pub timeout: Option<u64>,
+    #[arg(long)]
+    /// The authorization password to use to connect to the validate block proposal node endpoint
+    pub password: String,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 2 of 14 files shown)


stackslib/src/net/connection.rs
```diff
@@ -415,6 +415,8 @@ pub struct ConnectionOptions {
     /// the reward cycle in which Nakamoto activates, and thus needs to run both the epoch
     /// 2.x and Nakamoto state machines.
     pub force_nakamoto_epoch_transition: bool,
+    /// The authorization password to enable the block proposal RPC endpoint
+    pub supports_block_proposals: Option<String>,
```
hstove, line 419: I think the var name could be a little better, maybe something like `block_proposal_token`? Or something more generic if we think it could be useful elsewhere?

I get that this field does two things - enable the endpoint, and then specify the token, so it is tricky.

stackslib/src/net/http/request.rs
```diff
@@ -34,6 +35,56 @@ use crate::net::http::{
...
+/// An HTTP bearer authorization struct
+pub struct HttpBearerAuthorization {
+    /// The bearer authorization token
+    token: String,
+}
+
+impl HttpBearerAuthorization {
+    /// Create a new bearer authorization
+    pub fn new(password: String) -> Self {
+        let token = to_hex(Sha256Sum::from_data(password.as_bytes()).as_bytes());
```
hstove, line 47: I think the main purpose of using hashed tokens is so that you can have the hashed password on disk (in something like an nginx config) without compromising it. On the other hand, we can't do that in the signer, because the signer needs it in plaintext. 

I'm totally fine with keeping this scheme, but the more that I think about it, the more I think it's easier (and the same security) to just check the plain, non-hashed string. This is common with things like API keys, and if think we want more security (I think we don't), we could do an HMAC or signature scheme.

hstove (approved):
Thanks for hustling on this!

obycode (approved):
🙌 lgtm!

stackslib/src/net/connection.rs
jcnelson: Seconded

stackslib/src/net/http/request.rs
jcnelson: Also, this isn't the right way to do HMAC.  You don't want to use SHA256 for it, because it's vulnerable to a length extension attack.

Can we just use a plain password?  Also, do we need a whole struct for this?  Can the block-proposal endpoint just check for the `Authorization` header directly?

jcnelson (changes requested):
This isn't the right way to do HMAC. Also, we don't need it here since we're not even supporting HTTPS.

Can you instead just have the block-proposal endpoint check for an authorization header and treat the value as a bare password?  Thanks!

jferrant:
> This isn't the right way to do HMAC. Also, we don't need it here since we're not even supporting HTTPS.
> 
> Can you instead just have the block-proposal endpoint check for an authorization header and treat the value as a bare password? Thanks!

Yep will do!

obycode (dismissed):
lgtm!

jcnelson (dismissed):

hstove (dismissed):
👏

hstove:
@jferrant as a sanity check, can you run the regtest env with this PR's latest commit and make sure it produces at least one Nakamoto block?

hstove:
Re-requesting reviews. I'm running `regtest-env` based on the last commit here and it's working as expected! I've added the needed config changes to https://github.com/hirosystems/stacks-regtest-env/pull/32

hstove (approved):
lgtm and have tested in regtest!

obycode (approved):
lgtm!

jcnelson (approved):

Issue #4466, opened 2024-03-01
Stacks-signer cannot resolve hostnames for `node_host` config


zone117x:
The `node_host` value in the stacks-signer toml config can only be an IP:port and fails to resolve if given a hostname.

This makes the signer difficult to use in many environments. For example, here's a work-around required for a very simple docker-compose setup:
```yaml
  stacks-signer-2:
    build:
      context: .
      dockerfile: Dockerfile.stacks-node
    volumes:
      - ./signer-0.toml:/root/config.toml.in
    environment:
      STACKS_NODE_HOST: stacks-node:20443
      STACKS_SIGNER_ENDPOINT: 0.0.0.0:30002
      STACKS_CORE_EVENT_PORT: 3700
      STACKS_SIGNER_LOG_LEVEL: debug
      SIGNER_PRIVATE_KEY: ce109fee08860bb16337c76647dcbc02df0c06b455dd69bcf30af74d4eedd19301
    entrypoint:
      - /bin/bash
      - -c
      - |
        set -e
        # signer cannot resolve hostnames..
        export STACKS_NODE_HOST=$$(getent hosts $${STACKS_NODE_HOST%%:*} | awk '{ print $1 }'):$${STACKS_NODE_HOST##*:}
        echo "Updated STACKS_NODE_HOST to $$STACKS_NODE_HOST"
        envsubst < config.toml.in > config.toml
        exec stacks-signer run --config config.toml --reward-cycle 1
```

zone117x:
Fixed by https://github.com/stacks-network/stacks-core/pull/4475

PR #4475, opened 2024-03-04
fix: resolve DNS/hostnames for signer node_host config

zone117x:
Fixes https://github.com/stacks-network/stacks-core/issues/4466

### DIFF
--- a/libsigner/src/session.rs
+++ b/libsigner/src/session.rs
@@ -31,7 +31,7 @@ pub trait SignerSession {
     /// connect to the replica
     fn connect(
         &mut self,
-        host: String,
+        host: &str,
         stackerdb_contract_id: QualifiedContractIdentifier,
     ) -> Result<(), RPCError>;
     /// query the replica for a list of chunks
@@ -75,12 +75,9 @@ pub struct StackerDBSession {
 
 impl StackerDBSession {
     /// instantiate but don't connect
-    pub fn new(
-        host: String,
-        stackerdb_contract_id: QualifiedContractIdentifier,
-    ) -> StackerDBSession {
+    pub fn new(host: &str, stackerdb_contract_id: QualifiedContractIdentifier) -> StackerDBSession {
         StackerDBSession {
-            host,
+            host: host.to_owned(),
             stackerdb_contract_id,
             sock: None,
         }
@@ -134,10 +131,10 @@ impl SignerSession for StackerDBSession {
     /// connect to the replica
     fn connect(
         &mut self,
-        host: String,
+        host: &str,
         stackerdb_contract_id: QualifiedContractIdentifier,
     ) -> Result<(), RPCError> {
-        self.host = host;
+        self.host = host.to_owned();
         self.stackerdb_contract_id = stackerdb_contract_id;
         self.connect_or_reconnect()
     }
--- a/libsigner/src/session.rs
+++ b/libsigner/src/session.rs
@@ -31,7 +31,7 @@ pub trait SignerSession {
     /// connect to the replica
     fn connect(
         &mut self,
-        host: &str,
+        host: String,
         stackerdb_contract_id: QualifiedContractIdentifier,
     ) -> Result<(), RPCError>;
     /// query the replica for a list of chunks
@@ -131,10 +131,10 @@ impl SignerSession for StackerDBSession {
     /// connect to the replica
     fn connect(
         &mut self,
-        host: &str,
+        host: String,
         stackerdb_contract_id: QualifiedContractIdentifier,
     ) -> Result<(), RPCError> {
-        self.host = host.to_owned();
+        self.host = host;
         self.stackerdb_contract_id = stackerdb_contract_id;
         self.connect_or_reconnect()
     }
--- a/libsigner/src/session.rs
+++ b/libsigner/src/session.rs
@@ -31,7 +31,7 @@ pub trait SignerSession {
     /// connect to the replica
     fn connect(
         &mut self,
-        host: SocketAddr,
+        host: String,
         stackerdb_contract_id: QualifiedContractIdentifier,
     ) -> Result<(), RPCError>;
     /// query the replica for a list of chunks
@@ -66,7 +66,7 @@ pub trait SignerSession {
 /// signer session for a stackerdb instance
 pub struct StackerDBSession {
     /// host we're talking to
-    pub host: SocketAddr,
+    pub host: String,
     /// contract we're talking to
     pub stackerdb_contract_id: QualifiedContractIdentifier,
     /// connection to the replica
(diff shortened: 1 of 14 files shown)


zone117x:
Changes work as expected in local e2e testing using hostnames
```
tacks-signer-1-1  | INFO [1709564711.725348] [stacks-signer/src/runloop.rs:202] [signer_runloop] Signer #1 (ST18MDW2PDTBSCR1ACXYRJP2JX70FWNM6YY2VX4SS) is registered for reward cycle 6.
stacks-signer-1-1  | INFO [1709564711.866834] [stacks-signer/src/signer.rs:1172] [signer_runloop] Signer #1 is the current coordinator for 6 and must trigger DKG. Queuing DKG command...
stacks-signer-1-1  | INFO [1709564711.866859] [stacks-signer/src/runloop.rs:310] [signer_runloop] Runloop successfully initialized!
stacks-signer-1-1  | INFO [1709564711.910008] [stacks-signer/src/signer.rs:263] [signer_runloop] Signer #1: Starting DKG vote, round: 1, cycle: 6
stacks-signer-1-1  | WARN [1709564711.986812] [stacks-signer/src/runloop.rs:268] [signer_runloop] Signer is not registered for reward cycle 5. Waiting for confirmed registration...
stacks-signer-1-1  | INFO [1709564712.033942] [stacks-signer/src/runloop.rs:310] [signer_runloop] Runloop successfully initialized!
```

kantai (approved):

jferrant (approved):

jcnelson (changes requested):
Not NACKing this outright since it might be workable pending some deeper digging, but a decisive advantage of keeping `SocketAddr` over `String` for node hosts is that a DNS resolution won't stall the calling thread.  Right now, DNS resolution in Rust is synchronous -- the implementation of `ToSocketAddrs` for `(&str, u16)` and friends does a blocking call to `gethostbyname(3)`.  This can lead to inexplicable delays every time you try to send data to a host named by a DNS hostname instead of an IP address.

Most systems that use DNS avoid this through caching, but I don't know what Rust does here.  Does the `ToSocketAddrs` implementation cache DNS names?  If so, for how long?  Can the policy be changed by the operator?  Where is the policy that enforces this -- for example, is it the host's `libc` config?  Should we cache DNS names? (the Stacks node does this directly by running `ToSocketAddrs` calls in a separate thread, for example).

zone117x:
@jcnelson this is fixing the issue with using DNS _at all_. Whatever dns caching or similar optimizations should be in a different issue.

netrome:
> > Most systems that use DNS avoid this through caching, but I don't know what Rust does here. Does the ToSocketAddrs implementation cache DNS names? If so, for how long? Can the policy be changed by the operator? Where is the policy that enforces this -- for example, is it the host's libc config? Should we cache DNS names? (the Stacks node does this directly by running ToSocketAddrs calls in a separate thread, for example).
> 
> Rust just uses getaddrinfo in libc -- this seems like actually the "right" thing. Caching is controlled by the OS in this case, and adding a cache on top of that removes that control.

Also, if the provided address can be parsed as an IPv4 or IPv6 address - no DNS resolution takes place

From https://doc.rust-lang.org/src/std/net/socket_addr.rs.html#251-268
```
#[stable(feature = "rust1", since = "1.0.0")]
impl ToSocketAddrs for (&str, u16) {
    type Iter = vec::IntoIter<SocketAddr>;
    fn to_socket_addrs(&self) -> io::Result<vec::IntoIter<SocketAddr>> {
        let (host, port) = *self;

        // try to parse the host as a regular IP address first
        if let Ok(addr) = host.parse::<Ipv4Addr>() {
            let addr = SocketAddrV4::new(addr, port);
            return Ok(vec![SocketAddr::V4(addr)].into_iter());
        }
        if let Ok(addr) = host.parse::<Ipv6Addr>() {
            let addr = SocketAddrV6::new(addr, port, 0, 0);
            return Ok(vec![SocketAddr::V6(addr)].into_iter());
        }

        resolve_socket_addr((host, port).try_into()?)
    }
}
```

libsigner/src/session.rs
```diff
@@ -31,7 +31,7 @@ pub trait SignerSession {
     /// connect to the replica
     fn connect(
         &mut self,
-        host: SocketAddr,
+        host: String,
```
netrome, line 34: Nit: Generally you shouldn't take owned wrapper types as input to function calls. It's better to take `&str` here.

```suggestion
        host: &str,
```

jcnelson:
> @kantai Rust just uses getaddrinfo in libc -- this seems like actually the "right" thing. Caching is controlled by the OS in this case, and adding a cache on top of that removes that control.

> @netrome  Also, if the provided address can be parsed as an IPv4 or IPv6 address - no DNS resolution takes place

Okay, these to facts are what I needed to be certain of.  Approving...

netrome (changes requested):

jcnelson (approved):

libsigner/src/session.rs
zone117x: Good call, I changed instances of `String` arg types to `&str` -- how's it look now?

stacks-signer/src/cli.rs
```diff
@@ -67,7 +67,7 @@ pub enum Command {
 pub struct StackerDBArgs {
     /// The Stacks node to connect to
     #[arg(long)]
-    pub host: SocketAddr,
+    pub host: String,
```
netrome, line 70: And here

libsigner/src/session.rs
netrome: Better, but you missed a few spots. Pointed them out.

libsigner/src/session.rs
```diff
@@ -66,7 +66,7 @@ pub trait SignerSession {
 /// signer session for a stackerdb instance
 pub struct StackerDBSession {
     /// host we're talking to
-    pub host: SocketAddr,
+    pub host: String,
```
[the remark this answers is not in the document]
zone117x: `&str` properties in structs start to require lifetime params. Are you sure those are necessary?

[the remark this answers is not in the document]
netrome: Oh sorry, my mistake

stacks-signer/src/cli.rs
netrome: Sorry, skimming error from my side.

netrome (approved):

libsigner/src/session.rs
```diff
@@ -134,10 +131,10 @@ impl SignerSession for StackerDBSession {
     /// connect to the replica
     fn connect(
         &mut self,
-        host: SocketAddr,
+        host: &str,
         stackerdb_contract_id: QualifiedContractIdentifier,
     ) -> Result<(), RPCError> {
-        self.host = host;
+        self.host = host.to_owned();
```
netrome, line 137: Since you're just storing the host parameter, this is actually a legitimate case to be specific and take an owned String as input. I missed this in my first review. My apologies for the extra round-trip. Feel free to revert the last change.

jferrant:
I am seeing a lot of build errors mostly due to clones. As soon as those are addressed, will approve.

netrome (approved):

Issue #4446, opened 2024-02-28
[Testing] Introduce the `fake` crate for helping to create faked data instead of repeating boilerplate initializations in tests


cylewitruk:
There are a number of situations where the need for initializing complex types with "faked", but still valid, exists -- primarily in the context of tests. Currently, most tests are using boilerplate code for this, or some tests simply just don't exist likely due to the complexity of creating these objects by hand.

The [fake](https://crates.io/crates/fake) crate makes this a breeze, providing a number of out-of-the-box generators but also allowing full customization of how types, or even specific fields of a type (for example `String`s), are generated. The idea here is that this should help to remove a lot of the mental "hinders" from writing good tests involving complex objects.

This ticket is a placeholder for introducing the crate into projects and defining generators for the most commonly used types.

You can see an example of the crate in-use in my PR [here](https://github.com/stacks-network/stacks-core/blob/9b5f9cd12431d10d9fd1eaa256289c1a477e20f2/clarity/src/vm/fakes.rs).

cylewitruk:
Closing this issue in favor of #4501.

PR #4477, opened 2024-03-04
[Testing] Introduce `proptest` property testing library and add type alises for `hashbrown` types

cylewitruk:
This PR introduces the [proptest](https://crates.io/crates/proptest) crate with some example property testing tests and wrapper types in `stacks-common` for [hashbrown](https://crates.io/crates/hashbrown)'s `HashMap` and `HashSet` types.

Note: This is a break-out from the larger draft PR #4437.

I have provided a number of `proptest` strategies for common Clarity types, including `Contract`, `TypeSignature`,  `SymbolicExpression` and `Value` (some courtesy of @Acaccia from `clarity-wasm`).

~~The `hashbrown` wrapper types are needed as we cannot implement external crates for external types and I have provided proptest strategies for these new types as well. To use the new types, simply replace `std::collections::HashMap/Set` or `hashbrown::HashMap/Set` with `clarity_common::types::StacksHashMap/Set as HashMap/Set`~~
This caused problems in the signerlib which uses a combination of hashbrown and std, so given that this isn't strictly needed for this PR I have removed these new concrete types in favor of type aliases.

**Note that I have gone through and replaced `use::hashbrown::xxx` with the above, hence the large number of touched files.**

I have also added additional tests for `ClarityDatabase` which 1) were missing, and 2) demonstrate the usage of `proptest`.

#### Resolves Issues
- ~~#4446~~
- #4501


clarity/src/vm/database/tests/clarity_db_tests.rs
```diff
@@ -0,0 +1,177 @@
+use fake::{Faker, Fake};
+use rusqlite::NO_PARAMS;
+use stacks_common::util::hash::Sha512Trunc256Sum;
+
+use crate::vm::{contracts::Contract, database::{clarity_store::ContractCommitment, ClarityBackingStore, ClarityDatabase, ClaritySerializable, MemoryBackingStore, NULL_BURN_STATE_DB, NULL_HEADER_DB}, fakes::raw::EnglishWord, Value};
+
+#[test]
+fn insert_contract() {
+    for _ in 0..1000 {
```
moodmosaic, line 16: Would it be possible to use `fake` through `proptest`? `proptest` automatically re-runs the test for you, so the `loop` can be removed. Depending on how you will define the strategy, you may have the option to shrink any errors found. Replaying a failed test is another important quality of `proptest`.

cylewitruk: I'm not so familiar with `proptest` so if there's a good way of doing this I'm all ears! The difference being that when you call `Faker.fake()` you instead call `Faker.fake_with_rng(&rng)` where you provide the `Rng` instance to use.

But yeah, the loop is there just to ensure that instances of varying levels of complexity are tested otherwise.

jcnelson (changes requested):
I'm not opposed to adding a way to create mocked data structures, but I think this is fundamentally the wrong way to do it.

First, this should not be a dependency at all. It's purely for testing; it should be only a dev-dependency.

Second, it doesn't look like `fake` is actually doing that much work for us.  You're still creating all these constructors for all these data structures, and populating them with Rng-synthesized data.  Why have `fake` at all then?

Third, every data type in Clarity has a bound size by design. But, none of these constructors exercise it.  For example, it would be very useful to be able to do things like construct data types of maximum size, or maximum depth, in order to e.g. benchmark the code or iron out resource usage overflows, but that doesn't appear to be something `fake` does for us.

I'm not convinced that this PR is necessary.

cylewitruk:
> I'm not opposed to adding a way to create mocked data structures, but I think this is fundamentally the wrong way to do it.

And the fundamentally correct way would be...?

> First, this should not be a dependency at all. It's purely for testing; it should be only a dev-dependency.

As I wrote in the description, this needs to be included as an _optional dependency_ because of cargo's limitation that it doesn't propagate the `test` profile to dependencies, even in the same workspace, so `#[cfg(test)]` in `stacks-common` will not be visible to `clarity` or `stackslib` etc. when built in the `test` profile. This is the recommended way according to [here](https://github.com/rust-lang/rust/issues/59168) and [here](https://users.rust-lang.org/t/conditional-compilation-with-cfg-test-on-dependent-crates-without-effect-on-when-testing-parent-crate/51030).

> Second, it doesn't look like `fake` is actually doing that much work for us. You're still creating all these constructors for all these data structures, and populating them with Rng-synthesized data. Why have `fake` at all then?

I've provided specific implementations to manage recursion and to handle types which require that have i.e. two separate `Vec`s which need to have the same `len()`. Without this, the recursion is unpredictable and can sometimes cause the test thread to go OOM. Fake provides a lot of boilerplate that frankly would be a waste of time to re-implement for for this project "just because" -- I encourage you to check out their [examples](https://github.com/cksac/fake-rs/tree/master/fake/examples).

> Third, every data type in Clarity has a bound size by design. But, none of these constructors exercise it. For example, it would be very useful to be able to do things like construct data types of maximum size, or maximum depth, in order to e.g. benchmark the code or iron out resource usage overflows, but that doesn't appear to be something `fake` does for us.

With a bit more effort this could easily be implemented, very similar to their built-in "locales" support. This isn't a use-case for the moment with basic testing that instances can be serialized, inserted in the database, read from the database, deserialized and be equal.

clarity/src/vm/database/tests/clarity_db_tests.rs
cylewitruk: @moodmosaic does `proptest` have any way to provide a generator-function? i.e. could I have a function `fn some_test(contract: Contract)` and specify how `Contract` should be created (and ideally with an `Rng` instance)?

jcnelson:
You deleted this, but I think it's a valid question:

> And the fundamentally correct way would be...?

If I were doing this, I would just create a trait called `Mocked`, similar to `Default`, which provided the function `fn mocked() -> Self`, which when called would instantiate the struct with random data from the domain of possible inputs.  Then, you could have things like `ClarityName::mocked()` which evaluate to a `ClarityName` instance with plausible but synthesized data.

 > I've provided specific implementations to manage recursion and to handle types which require that i.e. two separate Vecs which need to have the same len(). Without this, the recursion is unpredictable and can sometimes cause the test thread to go OOM.
 
 Right -- you're doing this with our without `fake`.
 
 > Fake provides a lot of boilerplate that frankly would be a waste of time to re-implement for for this project "just because"
 
It seems it provides a lot of boilerplate that this PR doesn't use?  Also, to contrast this with a hypothetical `Mocked` trait above, it's not like it's hard to implement `Mocked` for primitive types.  If we took that route, we'd likely just define a macro that would let us synthesize implementations for one-liners (kinda like how we do with the `impl_stacks_message_codec_for_int!` macro to auto-generate code to implement `StacksMessageCodec` for all the integer types we use).

> With a bit more effort this could easily be implemented, very similar to their built-in "locales" support. This isn't a use-case for the moment with basic testing that instances can be serialized, inserted in the database, read from the database, deserialized and be equal.

With a bit more effort we can also just avoid `fake` altogether ;)  More seriously, we're writing this code with or without `fake`, meaning `fake` isn't providing value here.

---

Circling back to your first (deleted) question, I want to explore the line of questioning from @moodmosaic more.  A value-add that `fake` or something like it could bring is easy integration with an existing property-testing framework (or more generally, other widely-used automated testing systems like fuzzers, model checkers, and so on).  An argument _against_ my hypothetical `Mocked` trait would be that we'd have to do all the integration work ourselves, whereas a more standardized system like `fake` might do this for us.  That would be a "fundamentally correct" way to do this IMO -- add the means of synthesizing data structures to the codebase such that we can trivially plug them into an existing automated testing system.

cylewitruk:
> You deleted this, but I think it's a valid question:

I deleted it simply because I thought it could come across as rude :)

> If I were doing this, I would just create a trait called Mocked, similar to Default, which provided the function fn mocked() -> Self, which when called would instantiate the struct with random data from the domain of possible inputs. Then, you could have things like ClarityName::mocked() which evaluate to a ClarityName instance with plausible but synthesized data.

Just to ensure correct terminology and expectations, this PR introduces **faking** and not [mocking](https://docs.rs/mockall/latest/mockall/).

There's nothing wrong with the approach you suggest - it was the first thing that came to mind and which I started with -- but very quickly remembered that there are libraries to do most of this for me. So I'm _personally_ of the opinion that it is a waste of energy to re-implement, both on development and maintenance thereafter, when there are crates which give us that boilerplate.  I would understand the scrutiny much more if this were a production crate, but this is a pretty popular crate which in probably 95% of cases _can_ be used as a `dev-dependency` as intended -- It's just unfortunate that `cargo` doesn't give us the option of propagating the current build profile to dependencies so that it would be a little bit cleaner instead of having to use the "feature"-workaround.

> With a bit more effort we can also just avoid fake altogether ;) More seriously, we're writing this code with or without fake, meaning fake isn't providing value here.

If the idea was that it would be limited to the small number of examples provided here, then I would totally agree. But I think there are a _lot_ of missing tests due to laziness of not wanting to manually fake such types, not at the least the example tests which I added in this PR. And on the other end of the spectrum, probably a lot of unnecessary initialization code which just adds to LoC and maybe could benefit from some randomization for fuzzing/prop-testing.

> Circling back to your first (deleted) question, I want to explore the line of questioning from @moodmosaic more. A value-add that fake or something like it could bring is easy integration with an existing property-testing framework (or more generally, other widely-used automated testing systems like fuzzers, model checkers, and so on). An argument against my hypothetical Mocked trait would be that we'd have to do all the integration work ourselves, whereas a more standardized system like fake might do this for us. That would be a "fundamentally correct" way to do this IMO -- add the means of synthesizing data structures to the codebase such that we can trivially plug them into an existing automated testing system.

@moodmosaic discussed this briefly prior to this PR -- it would be great if we could get them to integrate, absolutely. It's a giant plus that they both use `Rng` and that `fake` allows the `Rng` instance to be provided. I can't imagine that this would be difficult to achieve, and frankly would rather put effort on that than re-inventing wheels.

clarity/src/vm/database/tests/clarity_db_tests.rs
moodmosaic: >a generator-function

Proptest uses a bit different terminology than most other Haskell QuickCheck-clones (although, it's debatable whether Proptest is actually a Haskell QuickCheck-clone as it's heavily inspired by Python's Hypothesis and *Elrnag*'s QuickCheck).

The way most, if not all, property-based testing libraries work is be allowing you to create a pair of a generator and a shrinker. A generator knows how to produces values of type `a` and a shrinker knows how to reduce `a` into an list of of smaller `a`s (to simplify those values and attempt to identify minimal failing cases).

While the original Haskell QuickCheck (and clones) name such a pair Arbitrary, Proptest uses the name Strategy.

Given the context above, to answer the questions:

>could I have a function `fn some_test(contract: Contract)` and specify how `Contract` should be generated

Proptest allows for the creation of custom Strategies to generate (and shrink) any type, including `Contract`. Although the test runner may not automatically infer how to instantiate a function like `fn some_test(contract: Contract)`, you can define a Strategy for `Contract` and utilize it within a standard test using proptest.

>and ideally pass an existing `Rng` instance

You could theoretically [get the prng from proptest](https://altsysrq.github.io/rustdoc/proptest/0.9.1/proptest/test_runner/struct.TestRunner.html#method.rng) and [pass it in fake](https://docs.rs/fake/latest/fake/trait.Fake.html#method.fake_with_rng), however this approach has implications:
1. You're required to always implement a shrinker since Proptest won't inherently understand how to shrink data generated by fake.
2. This might affect the randomness distribution of your tests. Proptest aims to generate test cases that are uniformly distributed across the space of possible inputs defined by the strategies, which is crucial for the effectiveness of property-based testing. It's unclear if integrating fake in this manner maintains this distribution, a key factor that can be assessed using benchmarks like the [dieharder](https://linux.die.net/man/1/dieharder) tests.

At the end, the effort must be done as in [`fake::locales::Data`](https://docs.rs/fake/latest/fake/locales/trait.Data.html), and as @jcnelson mentions in https://github.com/stacks-network/stacks-core/pull/4477#issuecomment-1977076270, and even though feasible, might be best concentrated within a single tool to maintain consistency and the integrity of the testing process.

Just using Proptest alone for both generation and shrinking is advisable for simplicity and reliability...

cylewitruk: I'm exploring with Anthony's help moving all generation to `proptest` - it hasn't been trivial, but there was a lot of work already done in the `clarity-wasm` repo for Clarity types. Will get back on this PR once I've got a better grasp of the effort needed and how well it works for "regular" unit tests.

clarity/src/proptesting/callables.rs
```diff
@@ -0,0 +1,118 @@
...
+    .prop_map(|(args, returns)| 
+        FunctionSignature {
+            args,
+            returns
+        }
+    )
+}
+
+pub fn defined_function() -> impl Strategy<Value = DefinedFunction> {
+    let x: usize = (0..3).sample_single(&mut rand::thread_rng());
```
moodmosaic, line 92: Why the need for sampling here? Can't we just ask the generator instead? If you have to replay a failing test, this will break the playback.

Could something like this work?

```rust
    let length_strategy = prop::num::usize::ANY.within(0..3);

    length_strategy.prop_flat_map(|x| {
        (
            // ...

**Comment by cylewitruk** [comment]
This PR has been repurposed to introduce `proptest` instead of `fake` which achieves the same goal, but with the benefit of the type generators being usable for property testing  :rocket: 

Note that most touched files are simply replacing `hashbrown::HashMap/Set` with the new wrapper types `StacksHashMap/Set`, which makes it much easier to swap implementations/hashers in the future and allows us to implement external traits on them.

@moodmosaic @jcnelson please re-review.

File: clarity/src/proptesting/types.rs [hunk 3]
```diff
@@ -0,0 +1,78 @@
...
+            1 => Just(TypeSignature::NoType),
+            9 => inner.clone(),
+        ]
+        .prop_map(|t| TypeSignature::new_option(t).unwrap()),
+        // response type: 20% (NoType, any) + 20% (any, NoType) + 60% (any, any)
+        prop_oneof![
+            1 => inner.clone().prop_map(|ok_ty| TypeSignature::new_response(ok_ty, TypeSignature::NoType).unwrap()),
+            1 => inner.clone().prop_map(|err_ty| TypeSignature::new_response(TypeSignature::NoType, err_ty).unwrap()),
+            3 => (inner.clone(), inner.clone()).prop_map(|(ok_ty, err_ty)| TypeSignature::new_response(ok_ty, err_ty).unwrap()),
+        ],
```
Acaccia, line 67: All those probabilities to create `NoType` were designed to check on the limitations of the typechecker in clarity-wasm. I'm not sure this is needed for your use case.

clarity/src/proptesting/values.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+        TypeSignature::SequenceType(SequenceSubtype::ListType(list_type_data)) => {
+            list(list_type_data).boxed()
+        }
+        TypeSignature::TupleType(tuple_ty) => tuple(tuple_ty).boxed(),
+        TypeSignature::PrincipalType => {
+            prop_oneof![principal_standard(), principal_contract()].boxed()
+        }
+        // TODO
+        TypeSignature::ListUnionType(_) => todo!(),
```
Acaccia, line 38: This is not really a TODO anymore: `ListUnionType` are created by accident when a List of qualified principal is generated.

clarity/src/proptesting/values.rs
```diff
@@ -0,0 +1,255 @@
...
+    pub fn any_sequence(size: usize) -> impl Strategy<Value = Self> {
+        let any_list = type_signature()
+            .prop_ind_flat_map2(move |ty| prop::collection::vec(value(ty), size))
+            .prop_map(move |(ty, vec)| {
+                Value::Sequence(SequenceData::List(ListData {
+                    data: vec,
+                    type_signature: ListTypeData::new_list(ty, size as u32).unwrap(),
+                }))
+            });
+        // TODO: add string-utf8
```
Acaccia, line 235: That is easy to add, the `string_utf8` strategy exists now.

clarity/src/proptesting/types.rs
cylewitruk: Tbh I will probably need to add different strategies for different use-cases, like clarity1 vs clarity2, etc. since i.e. trait handling is different between the versions, and other nuances.

clarity/src/proptesting/values.rs
cylewitruk: I'm not sure that I want to do anything here just yet as it's not so clear how this will evolve within `clarity` as more property tests are introduced, and it's nothing specifically required by this PR. OK to resolve?

clarity/src/proptesting/types.rs
cylewitruk: OK to resolve this with the same reasons as the other comment? Not sure how this functionality will evolve as more property tests are added to `clarity`.

Acaccia: Sure. Personally I don't mind. I was just not sure how useful it was for you, and if it was worth it to handle the `NoType` in your code.

cylewitruk:
@jcnelson would love a re-review of this so we can clear your requested changes, when you have time.

clarity/src/proptesting/callables.rs
```diff
@@ -0,0 +1,110 @@
+use proptest::prelude::*;
```
jcnelson, line 1: Missing copyright and licensing comment block

jcnelson:
Closing due to inactivity

PR #4480, opened 2024-03-04
Fix signer nonces

obycode:
Use the latest tip when retrieving the nonces for the singers to filter transactions from the stackerdb.


xoloki (approved):
LGTM

jferrant (approved):

jcnelson (approved):

PR #4481, opened 2024-03-04
Feat: Miner acts as signer set coordinator during block signing

kantai:
This PR alters the stacks-node miner thread to act as the signing set coordinator for gettings its own blocks signed. There's a couple hacks here that I know are hacks. I added a note about it below (it has to do with figuring out the current miner's key).

Otherwise, this PR does a handful of things in support of the miner-as-coordinator:

* **Replaces msg_id u32 with an enum for message identification**. Straight-forward refactor. 
* **Adds an additional slot for miner messages**. Miners need to send signer messages, not just block proposals. This adds a new slot for that.
* **Adds a sync channel for listening to StackerDB events**. The miner needs to receive StackerDB events. It could either poll the database (seems like a bad idea) or listen for events. Registering for RPC callbacks seems bad. Instead, this uses a singleton sync channel.
* **Adds a StackerDBs method for pushing a chunk locally and emitting event**. The node shouldn't make RPC calls to itself. This could trigger a deadlock if we aren't careful, it could lead to some pretty weird network patterns (depending on the deployment), and its less overhead this way anyways.  
* **A new signer message type to store DKG results**. This is read by miners to instantiate FIRE coordinator state. It is checked against the chainstate's aggregate key, and if even just one signer produces this message, the miner should be able to retrieve it. I understand this data can be cached on the miner for the duration of a reward cycle. I do not think that should happen in this PR, though, because I want to first focus on correctness and deal with optimizations separately.
* **Test signing channel for nakamoto integration tests.** Because the miner now doesn't listen for block approvals, but instead actively acts as a coordinator, the blind signer could no longer work as it did. It would need to mimic the entire FIRE protocol, which I think would be brittle, a waste of time, and not really the point of the non-signer nakamoto integration tests. So instead, there's a `[cfg(test)]` code block which checks if the blind signing channel is initialized and uses it for the test if so.
* Currently builds with a branch of wsts (this is reflected in the Cargo.toml, and should allow builds to work!) Once https://github.com/stacks-network/wsts/pull/3 merges and a release is built, the version in this repo will have to be bumped rather than building off the git branch.

### Hack: Determining the coordinator public key as a signer

If the active miner is acting as the signer, its public key needs to be used as the coordinator public key. Right now, the way this works is that the signer instance checks if its reward cycle is active, and if so, it uses the most recent miner public key it has received.

It fetches this from the stackerdb chunks messages -- using the most recent received event to get the public key. This is _mostly safe_: the .miners stackerdb really will only allow the current miner and the prior miner to send events. However, this _doesn't_ deal with issues related to the current miner being either non-responsive or malicious. To deal with that, the signer binary needs to be actually tracking some additional state, like the current burn chain view of the node. That way, the signer can have a designated "current trusted miner", that it only moves forward when a miner is selected. This is true regardless of the changes in this PR, though. The right way to handle this is to subscribe to burn block events, and expand that event's information. I don't necessarily think that needs to be done at the same time as this miner-as-coordinator work, though.

### Former Hacks!

These all _used to be_ necessary or hacky things in this branch, but are all no longer necessary or have otherwise been fixed.

#### Reusing the FIRE coordinator library in the miner

This is the correct thing to do, and the FIRE coordinator is instantiated with sufficient state to be able to detect which signers returned bad data so that it can restart signing with malicious signers removed.

~~As you could tell by inspecting the new `sign_coordinator` module, the wsts fire coordinator is reused by the miner. I don't think this is necessarily so hack-y (in fact, code reuse is A Good Thing), but it's something to think about. The miner isn't actually involved in, say, DKG, and so the fire coordinator may be a over-parameterized from its perspective, as well as theoretically capable of reaching a bunch of bad internal states (if the miner thread's coordinator was ever in a DKG state, e.g.). You can also see that I needed to make some internal state of the coordinator public, as well as implementing an Aggregator that took the aggregate public key as a given.~~

~~It's worth thinking about whether or not the miner should actually have its own state machine here.~~

#### Signers continue processing messages on their local coordinator

This was necessary to get the signers integration tests passing (as well as setting the DKG ID in the miner) without changes to them. This had more to do with the tests than anything else, and I updated the tests. I'm still working through fixing all of the signer tests though.

#### Removing the signature assembled checks from the signers tests

This isn't really a hack, but instead, the change from signers being the coordinator to the miner being a coordinator breaks some of the assumptions made in the signers tests setup. Namely, they no longer have a channel to a coordinator that assembles the final signature. There's definitely ways to fix these tests so that they still can assert the signatures are in place (like, we can add a variable to the tracking globals used by the miner in other tests), so I can just do that.

My fix was to just use the block production events from the event observer, and assert that the signature on the confirmed block matches the expected aggregate key.


Cargo.toml
```diff
@@ -20,7 +20,8 @@ rand_core = "0.6"
 rand = "0.8"
 rand_chacha = "0.3.1"
 tikv-jemallocator = "0.5.4"
-wsts = { version = "8.1", default-features = false }
+# wsts = { version = "8.1", default-features = false }
+wsts = { path = "/Users/aaron/devel/wsts" }
```
kantai, line 24: This was necessary to get access to some private variables in the coordinator. If we want to continue reusing the fire coordinator implementation in the miner's coordinator, I can open a PR with the necessary changes.

libsigner/src/events.rs
```diff
@@ -334,47 +342,86 @@ fn process_stackerdb_event(
...
-    }
+    ack_dispatcher(request);
 
     Ok(signer_event)
 }
 
+impl TryFrom<StackerDBChunksEvent> for SignerEvent {
+    type Error = EventError;
+
+    fn try_from(event: StackerDBChunksEvent) -> Result<Self, Self::Error> {
```
kantai, line 368: I refactored this into a `TryFrom` implementation because the miner's coordinator doesn't use the RPC callback interface. Instead, it receives the StackerDB events directly over a synchronous channel.

stacks-signer/src/runloop.rs
```diff
@@ -272,10 +272,17 @@ impl RunLoop {
...
+        current_reward_cycle: u64,
+        in_prepare_phase: bool,
+    ) -> Result<(), ClientError> {
         let next_reward_cycle = current_reward_cycle.saturating_add(1);
         self.refresh_signer_config(current_reward_cycle);
-        self.refresh_signer_config(next_reward_cycle);
+        // don't try to refresh the next reward cycle's signer state if there's no state for that cycle yet.
+        if in_prepare_phase {
+            self.refresh_signer_config(next_reward_cycle);
+        }
```
kantai, line 285: I don't think its _wrong_ to try to refresh the next cycle's state before the prepare phase, but it definitely results in lots of unnecessary RPC calls and log warnings.

stacks-signer/src/runloop.rs
```diff
@@ -306,8 +313,11 @@ impl RunLoop {
             self.state = State::Uninitialized;
             return Err(ClientError::NotRegistered);
         }
+        if self.state != State::Initialized {
+            // only info log if the state wasn't already initialized before.
+            info!("Runloop successfully initialized!");
+        }
```
kantai, line 319: Same here -- this is a very chatty log line otherwise, and it seems like this isn't really very informative unless the state was actually uninitialized before this call.

libsigner/src/events.rs
```diff
@@ -334,47 +342,86 @@ fn process_stackerdb_event(
...
+                        continue;
+                    };
+                    messages.push(msg);
+                } else {
+                    return Err(EventError::UnrecognizedEvent(
+                        "Unrecognized slot_id for miners contract".into(),
+                    ));
+                };
+            }
+            SignerEvent::ProposedBlocks(blocks, messages, miner_pk)
```
kantai, line 402: The big change here is that the miners contract now submits two kinds of information: block proposals and fire coordinator messages.

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -74,6 +77,21 @@ pub const PATH_ATTACHMENT_PROCESSED: &str = "attachments/new";
 pub const PATH_PROPOSAL_RESPONSE: &str = "proposal_response";
 pub const PATH_POX_ANCHOR: &str = "new_pox_set";
 
+lazy_static! {
+    pub static ref STACKER_DB_CHANNEL: StackerDBChannel = StackerDBChannel::new();
```
kantai, line 81: This channel is how the miner thread receives StackerDB events. The channel only gets events if the channel is actually held by a thread.

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -105,6 +123,40 @@ pub struct MinedNakamotoBlockEvent {
...
+            .take()
+    }
+
+    /// Is there a thread holding the receiver?
+    pub fn is_active(&self) -> bool {
+        // if the receiver field is empty (i.e., None), then a thread must have taken it.
+        self.receiver
+            .lock()
+            .expect("FATAL: poisoned StackerDBChannel lock")
+            .is_none()
```
kantai, line 156: There's a race-condition in how this gets used, but the solution is pretty easy. Right now, the miner sends its block proposal to the StackerDB before taking the receiver channel. This means that if the signer set was doing its own coordination, they might start the protocol, sending messages to the miner coordinator before the miner coordinator took the channel, meaning those events would be lost the miner. Instead `take_receiver` should be invoked first, then the proposal broadcast. Right now, there's no actual race, though, because the miner coordinator is what initiates the signing protocol. But if the protocol changed such that signers individually sent back NACKs on reading the proposal before the miner submitted NonceRequests, those NACKs could end up being missed.

stacks-signer/src/signer.rs
```diff
@@ -154,6 +154,42 @@ pub struct Signer {
...
+    pub burn_block_height: u32,
+    /// The active miner
+    pub active_miner: PublicKey,
+    /// The reward cycle
+    pub reward_cycle: u64,
+}
+
+impl Signer {
+    /// Return the current coordinator. If in the active reward cycle, this is the miner,
+    ///  so the first element of the tuple will be None (because the miner does not have a signer index).
```
jferrant, line 175: Does this mean DKG is still coordinated by the signers?

kantai: Yeah, exactly. DKG will still need to be coordinated among signers: miners don't have the same relationship to DKG that they do to block signing (that is: DKG can progress without the current miner. block production _cannot_ progress without the miner).

jferrant:
Is it not possible to just have a fire coordinator within the miner that triggers a sign round exactly as the signers do now? The only thing that would be different is the miner would not include its own key as part of the CoordinatorConfig I think... Do we actually need an aggregator?

xoloki:
> > Is it not possible to just have a fire coordinator within the miner that triggers a sign round exactly as the signers do now? The only thing that would be different is the miner would not include its own key as part of the CoordinatorConfig/if it did, it would be assigned zero key ids. Do we actually need an aggregator?
> 
> That's basically what `sign_coordinator` does. It instantiates a fire coordinator and triggers a sign round basically the same way the signers do now. The reason that a new Aggregator implementation is necessary is because the miner doesn't participate in DKG, which means it doesn't read the DKG results and therefore can't be initialized the same way that the default aggregator is (see https://github.com/Trust-Machines/wsts/blob/main/src/v2.rs#L383-L406). This is okay for the miner, though, because it doesn't need those commitments, it just takes the aggregate public key as given.

The miner needs those commitments if it wants to be able to check signature shares when aggregation fails.  If you can't check signature shares then you can't identify malicious actors.  Which means you can't keep signing blocks.

jcnelson:
@kantai Is this ready for review now?

kantai:
> @kantai Is this ready for review now?

No, not yet. I'm working on getting the existing nakamoto integrations tests to work with my changes (in branch `feat/miner-coordinator+dream-fixes`). The issue is that the previous auto-signer used in the integration tests depended on the stacks miner not really being involved in the signing communications, so there needs to be a workaround for that.

libsigner/src/events.rs
```diff
@@ -64,8 +65,12 @@ pub struct BlockProposalSigners {
 /// Event enum for newly-arrived signer subscribed events
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub enum SignerEvent {
-    /// The miner proposed blocks for signers to observe and sign
-    ProposedBlocks(Vec<BlockProposalSigners>),
+    /// The miner sent proposed blocks or messages for signers to observe and sign
+    ProposedBlocks(
+        Vec<BlockProposalSigners>,
```
jcnelson, line 70: Can you comment on what each of these fields are?

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
```diff
@@ -0,0 +1,517 @@
...
+            let Ok(signer_event) = SignerEvent::try_from(event).map_err(|e| {
+                warn!("Failure parsing StackerDB event into signer event. Ignoring message."; "err" => ?e);
+            }) else {
+                continue;
+            };
+            let SignerEvent::SignerMessages(signer_set, messages) = signer_event else {
+                debug!("Received signer event other than a signer message. Ignoring.");
+                continue;
+            };
+            if signer_set != u32::try_from(reward_cycle_id % 2).unwrap() {
```
obycode, line 420: In a case like this, I feel like it's cleaner to use an `as u32` instead of a `try_from` + `unwrap`.

obycode (commented):
This looks great!

libsigner/src/events.rs
```diff
@@ -371,47 +379,86 @@ fn process_stackerdb_event(
...
+                    };
+                    blocks.push(block);
+                } else if chunk.slot_id % 2 == 1 {
+                    // message
+                    let Ok(msg) = SignerMessage::consensus_deserialize(&mut chunk.data.as_slice())
+                    else {
+                        continue;
+                    };
+                    messages.push(msg);
+                } else {
```
jcnelson, line 433: How is this path reachable?

libsigner/src/messages.rs
```diff
@@ -168,35 +219,126 @@ impl From<&RejectCode> for RejectCodeTypePrefix {
...
+}
+
+impl SignerMessage {
+    /// Provide an interface for consensus serializing a DkgResults message
+    ///  without constructing the DkgResults struct (this eliminates a clone)
+    pub fn serialize_dkg_result<'a, W: Write, I>(
+        fd: &mut W,
+        aggregate_key: &Point,
+        party_polynomials: I,
+        write_prefix: bool,
```
jcnelson, line 292: Minor nit: can you instead create separate methods for when `write_prefix` is either true or false?  It's confusing for me the reader to remember that the fourth argument to `serialize_dkg_result()` influences what the first byte of the message is going to be, since the `bool` value does not convey intent the way a method name would.

libsigner/src/messages.rs
```diff
@@ -233,6 +386,46 @@ impl StacksMessageCodec for SignerMessage {
...
+                    for _ in 0..commit_len {
+                        let poly = Self::deserialize_point(fd)?;
+                        polynomial_poly.push(poly);
+                    }
+                    let polynomial_id = ID {
+                        id: polynomial_id_id,
+                        kG: polynomial_id_kg,
+                        kca: polynomial_id_kca,
+                    };
+                    let polynomial = PolyCommitment {
```
jcnelson, line 418: If we're maintaining a fork of `wsts` anyway, should we instead create a `StacksMessageCodec` implementation for `PolyCommitment` over there instead?

stacks-common/src/util/mod.rs
```diff
@@ -69,6 +70,48 @@ impl fmt::Display for HexError {
...
+        for key in keys.into_iter() {
+            let Some(value) = self.0.get(key) else {
+                continue;
+            };
+            write!(f, "{key}: {value:?}")?;
+        }
+        write!(f, "}}")
+    }
+}
+
```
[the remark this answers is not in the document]
jcnelson: Yeah, was about to ask how this is any different from the `Debug` implementation for `HashMap` today.

stacks-signer/src/client/stacks_client.rs
```diff
@@ -375,6 +375,22 @@ impl StacksClient {
         Ok(blocks_mined / reward_cycle_length)
     }
 
+    /// Get the current reward cycle and whether the prepare phase has started for the next cycle
+    pub fn get_current_reward_cycle_and_prepare_status(&self) -> Result<(u64, bool), ClientError> {
```
jcnelson, line 379: This makes me wonder if perhaps `PoxConstants` should get moved to `stacks_common` at some point.

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -3468,6 +3468,96 @@ impl SortitionDB {
...
+            SortitionDB::get_block_snapshot(self.conn(), &from_tip)?.ok_or_else(|| {
+                error!(
+                    "Could not find snapshot for sortition";
+                    "sortition_id" => %from_tip,
+                );
+                db_error::NotFoundError
+            })?;
+        let mut cursor = from_tip;
+        let mut last = None;
+        while self
```
jcnelson, line 3485: I'm not sure why you need this loop here?   Given a sortition at a given height, you can calculate what reward cycle it's in, and from there work out the height of the first prepare phase block within that reward cycle (`SortitionDB` (`self`) has a `PoxConstants` instance and a `first_burn_height` value so you can easily compute it).

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -3468,6 +3468,96 @@ impl SortitionDB {
...
+            .block_height_to_reward_cycle(self.first_block_height, tip_sn.block_height)
+            .expect("FATAL: stored snapshot with block height < first_block_height");
+
+        let prepare_phase_end = self
+            .pox_constants
+            .reward_cycle_to_block_height(self.first_block_height, reward_cycle_id)
+            .saturating_sub(1);
+
+        // find the sortition at height
+        let prepare_phase_end =
```
jcnelson, line 3527: Why even have this intermediate step?  Why doesn't this work?

```rust
let prepare_phase_start = self
   .pox_constants
   .reward_cycle_to_block_height(self.first_block_height, reward_cycle_id)
   .saturating_sub(self.pox_constants.prepare_phase_length);

let first_sortition = get_ancestor_sort_id(&self.index_conn(), prepare_phase_start, &tip_sn.sortition_id)?
   .ok_or_else(|| { /* elided */ )?
   .sortition_id;
```

stackslib/src/chainstate/nakamoto/miner.rs
```diff
@@ -485,16 +485,6 @@ impl NakamotoBlockBuilder {
 
         let ts_end = get_epoch_time_ms();
 
-        if let Some(observer) = event_observer {
```
jcnelson, line 488: Curious why this is getting dropped?

stackslib/src/net/rpc.rs
```diff
@@ -553,7 +553,7 @@ impl ConversationHttp {
                         self.handle_request(req, node)
                     })?;
 
-                    info!("Handled StacksHTTPRequest";
+                    debug!("Handled StacksHTTPRequest";
```
jcnelson, line 556: Can you please revert this?  We use the `info!` level at the Foundation so we can assess the node's HTTP performance behavior.

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -102,10 +124,59 @@ pub struct MinedNakamotoBlockEvent {
...
+            sender,
+        }
+    }
+
+    /// Return the receiver to the StackerDBChannel. This must be done before
+    ///  another interested thread can subscribe to events.
+    ///
+    /// The StackerDBChnnel's receiver is guarded with a Mutex, so that ownership can
+    /// be taken by different threads without unsafety.
+    pub fn replace_receiver(&self, receiver: Receiver<StackerDBChunksEvent>) {
```
jcnelson, line 147: This feels brittle.  I realize that there should never be two miner threads, but this API requires the miner implementation to _perfectly_ take and replace the receiver on each code path it could possibly take during instantiation and tear-down, or risk leaving the system in a broken state.  Is there a way to instead have the miner _subscribe itself_ to the channel by registering its own singleton receiver (e.g. paired to its thread ID by this code), and have the subscription logic garbage-collect stale receivers?  Then, all the miner thread has to do is subscribe for events, without any thought to what happens afterwards.

testnet/stacks-node/src/event_dispatcher.rs
jcnelson: If it's easy to fix, then can you just do it in this PR?

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -1065,19 +1159,30 @@ impl EventDispatcher {
...
-            modified_slots: new_chunks,
-        })
-        .expect("FATAL: failed to serialize StackerDBChunksEvent to JSON");
+            modified_slots,
+        };
+        let payload = serde_json::to_value(&event)
+            .expect("FATAL: failed to serialize StackerDBChunksEvent to JSON");
+
+        if interested_receiver {
+            if let Err(send_err) = STACKER_DB_CHANNEL.sender.send(event) {
```
jcnelson, line 1179: Suggestion: implement `StackerDBChannel::send(..)` directly, instead of exposing the inner `sender`.

libsigner/src/events.rs
kantai: It isn't, but really these `2`s should be replaced with `MINER_SLOT_COUNT`, in which case, it may be in the future.

libsigner/src/messages.rs
kantai: Then it would need to depend on `stacks-common`-- I think it's better to just have a serialization specific to the DKG results message.

stacks-common/src/util/mod.rs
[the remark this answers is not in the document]
kantai: Sorting, and using Display for values, but this isn't actually used, so it should be reverted anyways.

stacks-signer/src/client/stacks_client.rs
kantai: I think the bigger issue for stacks-signer is that it needs to construct the PoxConstants struct in the first place. That could be part of the deserialization of `/v2/pox_info`, but it isn't today.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
```diff
@@ -0,0 +1,517 @@
...
+            let key_ids: HashSet<u32> = (weight_start..weight_end).collect();
+            for key_id in key_ids.iter() {
+                wsts_key_ids.insert(*key_id, ecdsa_pk.clone());
+            }
+            signer_key_ids.insert(signer_id, key_ids);
+            wsts_signers.insert(signer_id, ecdsa_pk);
+        }
+
+        let num_keys = weight_end - 1;
+        let threshold = (num_keys * 70) / 100;
```
jcnelson, line 115: Isn't this supposed to be `(num_keys * 700) / 1000` now?

stackslib/src/chainstate/burn/db/sortdb.rs
kantai: Looking back at this code, I'm honestly not sure 🤷‍♂️ , will switch to invoking `get_ancestor_sort_id` directly for the prepare phase start.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
```diff
@@ -0,0 +1,517 @@
...
+        miners_db_config: StackerDBConfig,
+        config: &Config,
+    ) -> Result<Self, ChainstateError> {
+        let Some(ref reward_set_signers) = reward_set.signers else {
+            error!("Could not initialize WSTS coordinator for reward set without signer");
+            return Err(ChainstateError::NoRegisteredSigners(0));
+        };
+
+        let Some(receiver) = STACKER_DB_CHANNEL
+            .receiver
```
jcnelson, line 221: Isn't there a method for this?

stackslib/src/chainstate/nakamoto/miner.rs
kantai: It's invoked by the caller so that it can include the miner's signature.

stackslib/src/net/rpc.rs
kantai: Sure

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
```diff
@@ -0,0 +1,517 @@
...
+    }
+
+    fn get_sign_id(burn_block_height: u64, burnchain: &Burnchain) -> u64 {
+        burnchain
+            .pox_constants
+            .reward_cycle_index(burnchain.first_block_height, burn_block_height)
+            .expect("FATAL: tried to initialize WSTS coordinator before first burn block height")
+    }
+
+    fn send_signers_message(
```
jcnelson, line 292: Is this meant to broadcast the message to other nodes?  Because, this will be pretty slow -- it requires waiting for the StackerDB state machine to synchronize DB inventories with its neighbors, notice that it has a newer slot version than everybody else, and then broadcast the new chunk to them.  Instead, this function could directly send the StackerDB chunk to connected peers that replicate this chunk.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
```diff
@@ -0,0 +1,517 @@
...
+
+        let Some(ref mut receiver) = self.receiver else {
+            return Err(NakamotoNodeError::SigningCoordinatorFailure(
+                "Failed to obtain the StackerDB event receiver".into(),
+            ));
+        };
+
+        let start_ts = Instant::now();
+        while start_ts.elapsed() <= self.signing_round_timeout {
+            let event = match receiver.recv_timeout(Duration::from_millis(50)) {
```
jcnelson, line 393: why 50ms?

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
```diff
@@ -0,0 +1,517 @@
...
+                Ok(event) => event,
+                Err(std::sync::mpsc::RecvTimeoutError::Timeout) => {
+                    continue;
+                }
+                Err(std::sync::mpsc::RecvTimeoutError::Disconnected) => {
+                    return Err(NakamotoNodeError::SigningCoordinatorFailure(
```
jcnelson, line 399: Is this recoverable? Or even reachable?

testnet/stacks-node/src/event_dispatcher.rs
kantai: The SignerCoordinator, which obtains the receiver on instantiation, defines `Drop`. I think that the `take()` should be called closer to the end of the instantiation, so that there's no fallible code between it and the constructor.

The problem with having the signer coordinator push its receiver is that sender threads would need to get new send channels ever time a new signer coordinator was instantiated. The way its implemented here, once a thread has ownership of a sender, it doesn't need to refresh that view. It's certainly possible to do, but would involve a mutex on the event dispatching side (i.e., the sender).

testnet/stacks-node/src/event_dispatcher.rs
kantai: But there's actually not a race for the reason mentioned.

kantai: The "fix" would be sending the block proposal after instantiating the signer coordinator and obtaining the receiver channel -- this is easy enough to move around, but it would delay the message being sent slightly.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
kantai: Yep

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -1363,13 +1239,8 @@ fn stackerdb_filter_bad_transactions() {
...
     info!("------------------------- Verify Nakamoto Block Mined -------------------------");
-    let mined_block_event = signer_test.mine_nakamoto_block(timeout);
-    let hash = signer_test.wait_for_validate_ok_response(timeout);
-    let signatures = signer_test.wait_for_frost_signatures(timeout);
-    // Verify the signers accepted the proposed block and are using the previously determined dkg to sign it
-    for signature in &signatures {
-        assert!(signature.verify(&current_signers_dkg, hash.0.as_slice()));
-    }
+    let mined_block_event =
+        signer_test.mine_and_verify_confirmed_naka_block(&current_signers_dkg, timeout);
```
jcnelson, line 1243: A moment of truth for this PR is that once we get my downloader PR merged, we can spawn two miner runloops and verify that they can each produce blocks that the other builds atop.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
jcnelson: This might take a bit of code, but it's not much. You'd want to give `SignerCoordinator` a `NetworkHandle` instance connected to the `PeerNetwork` instance (we already have one in the `Relayer`).  From there, you can queue messages to broadcast on the p2p network via `::send_message()`.  We'd want to extend the logic to specifically accommodate StackerDB messages so we only send the message to a peer that replicates the same StackerDB, but that wouldn't be too bad.  I can add this on top of this PR (or as a follow-on one) if you don't want to add it here.

testnet/stacks-node/src/tests/signer.rs
kantai: A moment of truth for this PR, or any PR that has ever consumed StackerDB events?

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
kantai: It's not recoverable, and I don't think its reachable.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
kantai: That sounds like it definitely needs to be a separate PR.

testnet/stacks-node/src/tests/signer.rs
jcnelson: Well, considering that this PR introduces a new way of sending and receiving StackerDB events, I guess both? ;)

All I was trying to say is that I wish it was possible right now to have an integration test that spanwed multiple miner runloops to verify that they can use a shared StackerDB to build blocks, but this is currently blocked for want of a downloader -- the two runloops would have no way of synchronizing their blocks in the first place.  But, I do believe it's a worthwhile integration test because it would prove that the new miner-as-signer system here works end-to-end.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
kantai: Yes, that sounds right. I think that this code should be deduped with the signer binary somehow. I'll try to see what I can do in that respect.

libsigner/src/messages.rs
kantai: Sure -- refactored so that the only public method serializes the full payload, and the internal method serializes the components.

libsigner/src/events.rs
kantai: Refactored to use `MINER_SLOT_COUNT` in the modulus, so now this `else` serves as a failsafe if that constant is bumped up.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
kantai: I kind of agree, but would like to see some lints activated before switching from defaulting to `try_from().unwrap()`:

```
#![warn(
    clippy::cast_possible_wrap, // unsigned -> signed
    clippy::cast_sign_loss, // signed -> unsigned
    clippy::cast_lossless,
    clippy::cast_possible_truncation,
)]
```

That way, at least CI would tell me if my reasoning about modulus arithmetic was wrong.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
kantai: The signer actually converts to floats and then rounded -- I refactored this so that the signer and this method both share this calculation via libsigner.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
kantai: It would now be reachable if a miner thread didn't clean up before another miner thread started. The prior miner thread could reach this error at which point it should exit.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
kantai: Switched to using the RPC interface, which avoids this issue for now.

obycode (dismissed):
lgtm

libsigner/src/messages.rs
```diff
@@ -54,28 +65,67 @@ use wsts::state_machine::{signer, SignError};
...
-const DKG_PRIVATE_SHARES_MSG_ID: u32 = 5;
-const NONCE_REQUEST_MSG_ID: u32 = 6;
-const NONCE_RESPONSE_MSG_ID: u32 = 7;
-const SIGNATURE_SHARE_REQUEST_MSG_ID: u32 = 8;
-const SIGNATURE_SHARE_RESPONSE_MSG_ID: u32 = 9;
-/// The slot ID for the block response for miners to observe
-pub const BLOCK_MSG_ID: u32 = 10;
-/// The slot ID for the transactions list for miners and signers to observe
-pub const TRANSACTIONS_MSG_ID: u32 = 11;
+define_u8_enum!(
```
jferrant, line 68: Noice. So much cleaner.

libsigner/src/messages.rs
```diff
@@ -168,35 +219,135 @@ impl From<&RejectCode> for RejectCodeTypePrefix {
...
+    }
+
+    fn deserialize_point<R: Read>(fd: &mut R) -> Result<Point, CodecError> {
+        let mut bytes = [0; 33];
+        fd.read_exact(&mut bytes).map_err(CodecError::ReadError)?;
+        Point::try_from(&Compressed::from(bytes))
+            .map_err(|e| CodecError::DeserializeError(e.to_string()))
+    }
+
+    fn deserialize_scalar<R: Read>(fd: &mut R) -> Result<Scalar, CodecError> {
```
jferrant, line 347: We already have an implimentation of this no?

StacksMessageCodecExtensions is implemented for Scalar and for Point?

libsigner/src/messages.rs
```diff
@@ -233,6 +394,46 @@ impl StacksMessageCodec for SignerMessage {
...
             }
+            SignerMessageTypePrefix::DkgResults => {
+                let aggregate_key = Self::deserialize_point(fd)?;
+                let party_polynomial_len = u32::consensus_deserialize(fd)?;
+                let mut party_polynomials = Vec::with_capacity(
+                    party_polynomial_len
+                        .try_into()
+                        .expect("FATAL: u32 could not fit in usize"),
+                );
+                for _ in 0..party_polynomial_len {
```
jferrant, line 405: Could you implement StacksMessageCodecExtensions for the party polynomial?

stacks-common/src/libcommon.rs
```diff
@@ -62,5 +62,5 @@ pub mod consts {
 
     /// The number of StackerDB slots each signing key needs
     ///  to use to participate in DKG and block validation signing.
-    pub const SIGNER_SLOTS_PER_USER: u32 = 12;
+    pub const SIGNER_SLOTS_PER_USER: u32 = 13;
```
jferrant, line 65: Is this required anymore? Can we use MessageSlotID::ALL? Or would that cause a circular reference here?

stacks-signer/src/client/stacks_client.rs
```diff
@@ -383,6 +383,22 @@ impl StacksClient {
         })
     }
 
+    /// Get the current reward cycle and whether the prepare phase has started for the next cycle
+    pub fn get_current_reward_cycle_and_prepare_status(&self) -> Result<(u64, bool), ClientError> {
```
jferrant, line 387: This doesn't seem to be used anywhere?

stacks-signer/src/runloop.rs
```diff
@@ -189,10 +136,9 @@ impl RunLoop {
             warn!("No registered signers found for reward cycle {reward_cycle}.");
             return Ok(None);
         }
-        Ok(Some(Self::parse_nakamoto_signer_entries(
-            &signers,
-            self.config.network.is_mainnet(),
-        )))
+        let entries =
+            ParsedSignerEntries::parse(self.config.network.is_mainnet(), &signers).unwrap();
```
jferrant, line 140: Just a nit: seems a bit odd to call something "ParsedSignerEntries::parse" feel like it should be "SignerEntries::parse" or "ParsedSignerEntries::new" or "ParsedSignerEntries::from_raw_signer_entries" but feel free to ignore.

stacks-common/src/libcommon.rs
kantai: Yeah, that would cause a circular reference.

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -105,10 +133,93 @@ pub struct MinedNakamotoBlockEvent {
...
+        stackerdb: &QualifiedContractIdentifier,
+    ) -> Option<Sender<StackerDBChunksEvent>> {
+        // if the receiver field is empty (i.e., None), then there is no listening thread, return None
+        let guard = self
+            .sender_info
+            .lock()
+            .expect("FATAL: poisoned StackerDBChannel lock");
+        let sender_info = guard.as_ref()?;
+        if sender_info.interested_in_signers
+            && stackerdb.issuer.1 == [0; 20]
```
jferrant, line 211: Nit: This check is done often enough. Would it be possible to ad a helper function like "is_boot_contract" or something like that to the QualifiedContractIdentifier?

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -59,6 +60,7 @@ use crate::{neon_node, ChainTip};
 const ABORT_TRY_AGAIN_MS: u64 = 200;
 /// If the signers have not responded to a block proposal, how long should
 ///  the miner thread sleep before trying again?
+#[allow(unused)]
 const WAIT_FOR_SIGNERS_MS: u64 = 200;
```
jferrant, line 64: Should we just delete this then?

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -273,24 +239,221 @@ impl BlockMinerThread {
...
+                "Failed to obtain the active aggregate public key. Cannot mine!".into(),
+            ));
+        };
+
+        #[cfg(test)]
+        {
+            // In test mode, short-circuit spinning up the SignCoordinator if the TEST_SIGNING
+            //  channel has been created. This allows integration tests for the stacks-node
+            //  independent of the stacks-signer.
+            let mut signer = crate::tests::nakamoto_integrations::TEST_SIGNING
```
jferrant, line 313: Nit: might be nice to have this in a sep function.

libsigner/src/signer_set.rs
```diff
@@ -0,0 +1,137 @@
...
+/// A reward set parsed into the structures required by WSTS party members and coordinators.
+#[derive(Debug, Clone)]
+pub struct ParsedSignerEntries {
+    /// The signer addresses mapped to signer id
+    pub signer_ids: HashMap<StacksAddress, u32>,
+    /// The signer ids mapped to public key and key ids mapped to public keys
+    pub public_keys: PublicKeys,
+    /// The signer ids mapped to key ids
+    pub signer_key_ids: HashMap<u32, Vec<u32>>,
+    /// The signer ids mapped to wsts public keys
```
jcnelson, line 32: Wouldn't this be key IDs mapped to WSTS public keys?  Or am I missing something?

libsigner/src/signer_set.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub fn parse(is_mainnet: bool, reward_set: &[NakamotoSignerEntry]) -> Result<Self, Error> {
+        let mut weight_end = 1;
+        let mut signer_key_ids = HashMap::with_capacity(reward_set.len());
+        let mut signer_public_keys = HashMap::with_capacity(reward_set.len());
+        let mut coordinator_key_ids = HashMap::with_capacity(4000);
+        let mut signer_ids = HashMap::with_capacity(reward_set.len());
+        let mut wsts_signers = HashMap::new();
+        let mut wsts_key_ids = HashMap::new();
+        for (i, entry) in reward_set.iter().enumerate() {
+            let signer_id = u32::try_from(i).map_err(|_| Error::SignerCountOverflow)?;
```
jcnelson, line 61: Shouldn't we just error out of `reward_set.len()` is bigger than u32::MAX? Or even something more realistic, like 4000?

libsigner/src/signer_set.rs
```diff
@@ -0,0 +1,137 @@
...
+            .signers
+            .len()
+            .try_into()
+            .map_err(|_| Error::SignerCountOverflow)
+    }
+
+    /// Return the number of Key IDs required to sign a message with the WSTS group signature
+    pub fn get_signing_threshold(&self) -> Result<u32, Error> {
+        let num_keys = self.count_keys()?;
+        Ok((num_keys as f64 * 7_f64 / 10_f64).ceil() as u32)
```
jcnelson, line 129: Should we use consistent math here?  As in, `num_keys * 700 / 1000`?

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -133,50 +139,157 @@ pub struct MinedNakamotoBlockEvent {
     pub signer_bitvec: String,
 }
 
+impl InnerStackerDBChannel {
+    pub fn new_miner_receiver() -> (Receiver<StackerDBChunksEvent>, Self) {
+        let (sender, recv) = channel();
```
jcnelson, line 144: This channel can grow to unbounded size.  Is that something we should worry about?

jcnelson (approved):

jferrant (commented):
Pretty much just miner nits/cleanup! Awesome job @kantai

libsigner/src/signer_set.rs
kantai: Are floats not consistent here? Either way, these are the existing calculations from the signer side.

testnet/stacks-node/src/event_dispatcher.rs
kantai: I don't think so -- the channel is garbage collected after every block production. It only contains the signer event messages, so if the total number of those for a single block production is creating memory pressure, there's other issues (and there wouldn't be an easy fix either: we wouldn't want the sender to stall on a full buffer, so we'd need to switch to polling the db for updates).

libsigner/src/signer_set.rs
kantai: This does error out in exactly that case. We could error earlier by doing something like:

```
let _total_len = u32::try_from(reward_set.len()).map_err(|_| Error::SignerCountOverflow)?;
for (i, entry) in reward_set.iter().enumerate() {
    let signer_id = u32::try_from(i).unwrap(); // safe because we checked above
```

But I'd prefer avoiding the `unwrap()` even if we know its safe.

jcnelson (dismissed):

testnet/stacks-node/src/nakamoto_node/miner.rs
kantai: Yep, added in [c50a930](https://github.com/stacks-network/stacks-core/pull/4481/commits/c50a93088d7c0261f1dbe31ab24b95028a038447)

testnet/stacks-node/src/nakamoto_node/miner.rs
kantai: Yep, done in [c50a930](https://github.com/stacks-network/stacks-core/pull/4481/commits/c50a93088d7c0261f1dbe31ab24b95028a038447)

stacks-signer/src/runloop.rs
kantai: Renamed to SignerEntries in [c50a930](https://github.com/stacks-network/stacks-core/pull/4481/commits/c50a93088d7c0261f1dbe31ab24b95028a038447)

stacks-signer/src/client/stacks_client.rs
kantai: Yep, removed in [c50a930](https://github.com/stacks-network/stacks-core/pull/4481/commits/c50a93088d7c0261f1dbe31ab24b95028a038447)

libsigner/src/messages.rs
kantai: Yep, added in [c50a930](https://github.com/stacks-network/stacks-core/pull/4481/commits/c50a93088d7c0261f1dbe31ab24b95028a038447)

libsigner/src/signer_set.rs
obycode: If we want to be really sure of consistency, there is a readonly function, `get-threshold-weight`,  in the signers-voting clarity contract that you can call.

libsigner/src/events.rs
```diff
@@ -258,7 +267,7 @@ impl EventReceiver for SignerEventReceiver {
     /// Errors are recoverable -- the caller should call this method again even if it returns an
     /// error.
     fn next_event(&mut self) -> Result<SignerEvent, EventError> {
-        self.with_server(|event_receiver, http_server, is_mainnet| {
+        self.with_server(|event_receiver, http_server, _is_mainnet| {
```
jferrant, line 270: This is a nit and doesn't need to be done here if you just want to merge, but we should change with_server fcn sig since it no longer needs is_mainnet

libsigner/src/messages.rs
jferrant: Can we delete deserialize_scalar and deserialize_point then if they are no longer used?

obycode (dismissed):
👍

jferrant (changes requested):
Just delete the defunct deserialize_scalar and deserialize_point fcns (unless I am missing something) and I will approve :D

obycode (dismissed):

libsigner/src/signer_set.rs
xoloki: The reason we're using floats here is that we need to round these calculations up, and integer math will silently round down.  We need to round up because if say 70% of the number of keys contains a fraction, then rounding down means that our threshold is less than 70%.

I know floats are not particularly consistent in general but for this calculation they should return a consistent answer across all platforms and environments.

obycode (approved):

jferrant (approved):

obycode (approved):

jcnelson (approved):

obycode (approved):

jferrant (approved):
LGTM! :D

PR #4482, opened 2024-03-04
chore: quiet some noisy logs

obycode:
This is just changing some `info` logs to `debug` because they are a bit too frequent and probably not necessary to be always printing.

### DIFF
--- a/stackslib/src/net/rpc.rs
+++ b/stackslib/src/net/rpc.rs
@@ -553,7 +553,7 @@ impl ConversationHttp {
                         self.handle_request(req, node)
                     })?;
 
-                    info!("Handled StacksHTTPRequest";
+                    debug!("Handled StacksHTTPRequest";
                            "verb" => %verb,
                            "path" => %request_path,
                            "processing_time_ms" => start_time.elapsed().as_millis(),
--- a/testnet/stacks-node/src/neon_node.rs
+++ b/testnet/stacks-node/src/neon_node.rs
@@ -1251,7 +1251,7 @@ impl BlockMinerThread {
         chain_state: &mut StacksChainState,
         at_stacks_height: Option<u64>,
     ) -> Option<TipCandidate> {
-        info!("Picking best Stacks tip");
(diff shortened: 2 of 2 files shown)


stackslib/src/net/rpc.rs
```diff
@@ -553,7 +553,7 @@ impl ConversationHttp {
                         self.handle_request(req, node)
                     })?;
 
-                    info!("Handled StacksHTTPRequest";
+                    debug!("Handled StacksHTTPRequest";
```
jcnelson, line 556: Please keep this as info.  We use this to log HTTP traffic on our nodes without resorting to the debug firehose.

testnet/stacks-node/src/neon_node.rs
```diff
@@ -1251,7 +1251,7 @@ impl BlockMinerThread {
         chain_state: &mut StacksChainState,
         at_stacks_height: Option<u64>,
     ) -> Option<TipCandidate> {
-        info!("Picking best Stacks tip");
+        debug!("Picking best Stacks tip");
```
jcnelson, line 1254: The reason these are all `info` is that this is pertinent information for miners who need to debug their chain-tip selection without turning on the debug firehose.

obycode: The trouble is that this is printing every 100ms.

obycode: So it becomes a firehose on its own.

jcnelson: Why are you mining every 100ms?

jcnelson: That's still buggy behavior (or buggy configuration).  The node shouldn't be trying to build blocks so quickly.

obycode: Yeah, you're right. Will try to debug that issue separately. Glad to hear that this is not expected to constantly do this! I'll close this PR.

PR #4483, opened 2024-03-05
Chore: Replace Contains Key with Entry

ASuciuX:
### Benchmark Results
```bash
hyperfine -w 3 -r 10 "./stacks-inspect-383d586a7-4-mar replay-block ~/chainstate range 99990 100000" "./stacks-inspect-with-entry-4-mar replay-block ~/chainstate range 99990 100000" "./stacks-inspect-immutable-entries replay-block ~/chainstate range 99990 100000"
Benchmark 1: ./stacks-inspect-383d586a7-4-mar replay-block ~/chainstate range 99990 100000
  Time (mean ± σ):      7.977 s ±  0.060 s    [User: 7.584 s, System: 0.365 s]
  Range (min … max):    7.934 s …  8.105 s    10 runs
 
Benchmark 2: ./stacks-inspect-with-entry-4-mar replay-block ~/chainstate range 99990 100000
  Time (mean ± σ):      7.918 s ±  0.024 s    [User: 7.524 s, System: 0.364 s]
  Range (min … max):    7.891 s …  7.965 s    10 runs
 
Benchmark 3: ./stacks-inspect-immutable-entries replay-block ~/chainstate range 99990 100000
  Time (mean ± σ):      7.946 s ±  0.012 s    [User: 7.554 s, System: 0.362 s]
  Range (min … max):    7.925 s …  7.966 s    10 runs
 
Summary
  ./stacks-inspect-with-entry-4-mar replay-block ~/chainstate range 99990 100000 ran
    1.00 ± 0.00 times faster than ./stacks-inspect-immutable-entries replay-block ~/chainstate range 99990 100000
    1.01 ± 0.01 times faster than ./stacks-inspect-383d586a7-4-mar replay-block ~/chainstate range 99990 100000
```

The benchmark above showcase the performance of data for the following:
1. default run
2. only update mutable occurrences
3. update mutable and immutable as well

The best benchmark is the 2nd case, with updated mutable occurrences and default immutable ones.

### Applicable issues
- https://github.com/stacks-network/stacks-core/issues/4316

### DIFF
--- a/clarity/src/vm/ast/traits_resolver/mod.rs
+++ b/clarity/src/vm/ast/traits_resolver/mod.rs
@@ -14,8 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use hashbrown::hash_map::Entry;
-use hashbrown::{HashMap, HashSet};
+use hashbrown::hash_map::{Entry as EntryHashMap, OccupiedEntry as OccupiedEntryHashMap};
+use hashbrown::HashMap;
+use std::collections::btree_map::{Entry as EntryBTreeMap, OccupiedEntry as OccupiedEntryBTreeMap};
+use std::hash::{BuildHasher, Hash};
 
 use crate::vm::analysis::AnalysisDatabase;
 use crate::vm::ast::errors::{ParseError, ParseErrors, ParseResult};
@@ -38,6 +40,109 @@ impl BuildASTPass for TraitsResolver {
         let mut command = TraitsResolver::new();
         command.run(contract_ast)?;
         Ok(())
+    }
+}
+
+pub trait EntryHashMapExt<'a, K: 'a, V: 'a, S: 'a>
+where
+    S: BuildHasher,
+    Self: Sized,
+{
+    fn occupied_and_then<T, F, E>(self, f: F) -> Result<Self, E>
+    where
+        F: FnOnce(&OccupiedEntryHashMap<'a, K, V, S>) -> Result<T, E>;
+
+    fn or_try_insert<F, E>(self, default: F) -> Result<&'a mut V, E>
+    where
+        F: FnOnce() -> Result<V, E>;
+
+    fn vacant_or<E>(self, err: E) -> Result<Self, E>;
+}
+
+pub trait EntryBTreeMapExt<'a, K: 'a, V: 'a>
+where
+    // S: BuildHasher,
+    Self: Sized,
+{
+    fn occupied_and_then<T, F, E>(self, f: F) -> Result<Self, E>
+    where
+        F: FnOnce(&OccupiedEntryBTreeMap<'a, K, V>) -> Result<T, E>;
+
+    fn or_try_insert<F, E>(self, default: F) -> Result<&'a mut V, E>
+    where
+        F: FnOnce() -> Result<V, E>;
+
+    fn vacant_or<E>(self, err: E) -> Result<Self, E>;
+}
+
+impl<'a, K: 'a, V: 'a, S: 'a> EntryHashMapExt<'a, K, V, S> for EntryHashMap<'a, K, V, S>
+where
+    S: BuildHasher,
+    K: Hash,
+{
+    fn occupied_and_then<T, F, E>(self, f: F) -> Result<Self, E>
+    where
+        F: FnOnce(&OccupiedEntryHashMap<'a, K, V, S>) -> Result<T, E>,
+    {
+        match self {
+            EntryHashMap::Occupied(ref e) => {
+                f(e)?;
+                Ok(self)
+            }
+            EntryHashMap::Vacant(_) => Ok(self),
+        }
+    }
+
+    fn or_try_insert<F, E>(self, default: F) -> Result<&'a mut V, E>
+    where
+        F: FnOnce() -> Result<V, E>,
+    {
+        match self {
+            EntryHashMap::Occupied(e) => Ok(e.into_mut()),
+            EntryHashMap::Vacant(e) => Ok(e.insert(default()?)),
+        }
+    }
+
+    fn vacant_or<E>(self, err: E) -> Result<Self, E> {
+        match self {
+            EntryHashMap::Occupied(_) => Err(err),
+            EntryHashMap::Vacant(_) => Ok(self),
+        }
+    }
+}
+
+impl<'a, K: 'a, V: 'a> EntryBTreeMapExt<'a, K, V> for EntryBTreeMap<'a, K, V>
+where
+    K: Hash + Ord,
+{
+    fn occupied_and_then<T, F, E>(self, f: F) -> Result<Self, E>
+    where
+        F: FnOnce(&OccupiedEntryBTreeMap<'a, K, V>) -> Result<T, E>,
+    {
+        match self {
+            EntryBTreeMap::Occupied(ref e) => {
+                f(e)?;
+                Ok(self)
+            }
+            EntryBTreeMap::Vacant(_) => Ok(self),
+        }
+    }
+
+    fn or_try_insert<F, E>(self, default: F) -> Result<&'a mut V, E>
+    where
+        F: FnOnce() -> Result<V, E>,
+    {
+        match self {
+            EntryBTreeMap::Occupied(e) => Ok(e.into_mut()),
+            EntryBTreeMap::Vacant(e) => Ok(e.insert(default()?)),
+        }
+    }
+
+    fn vacant_or<E>(self, err: E) -> Result<Self, E> {
+        match self {
+            EntryBTreeMap::Occupied(_) => Err(err),
+            EntryBTreeMap::Vacant(_) => Ok(self),
+        }
     }
 }
 
@@ -64,21 +169,17 @@ impl TraitsResolver {
 
                     match (&args[0].pre_expr, &args[1].pre_expr) {
                         (Atom(trait_name), List(trait_definition)) => {
-                            // Check for collisions
-                            match contract_ast.referenced_traits.entry(trait_name.to_owned()) {
-                                Entry::Occupied(_) => {
-                                    return Err(ParseErrors::NameAlreadyUsed(
-                                        trait_name.to_string(),
-                                    )
-                                    .into());
-                                }
-                                Entry::Vacant(e) => {
-                                    // Traverse and probe for generics nested in the trait definition
+                            contract_ast
+                                .referenced_traits
+                                .entry(trait_name.to_owned())
+                                .vacant_or(ParseErrors::NameAlreadyUsed(trait_name.to_string()))?
+                                .or_try_insert(|| {
                                     self.probe_for_generics(
                                         trait_definition.iter(),
                                         &mut referenced_traits,
                                         true,
-                                    )?;
+                                    )
+                                    .map_err(|e| e.err)?;
 
                                     let trait_id = TraitIdentifier {
                                         name: trait_name.clone(),
@@ -86,9 +187,8 @@ impl TraitsResolver {
                                             .contract_identifier
                                             .clone(),
                                     };
-                                    e.insert(TraitDefinition::Defined(trait_id));
-                                }
-                            }
+                                    Ok::<_, ParseErrors>(TraitDefinition::Defined(trait_id))
+                                })?;
                         }
                         _ => return Err(ParseErrors::DefineTraitBadSignature.into()),
                     }
@@ -99,14 +199,11 @@ impl TraitsResolver {
                     }
 
                     if let Some(trait_name) = args[0].match_atom() {
-                        // Check for collisions
-                        match contract_ast.referenced_traits.entry(trait_name.to_owned()) {
-                            Entry::Occupied(_) => {
-                                return Err(
-                                    ParseErrors::NameAlreadyUsed(trait_name.to_string()).into()
-                                );
-                            }
-                            Entry::Vacant(e) => {
+                        contract_ast
+                            .referenced_traits
+                            .entry(trait_name.to_owned())
+                            .vacant_or(ParseErrors::NameAlreadyUsed(trait_name.to_string()))?
+                            .or_try_insert(|| {
                                 let trait_id = match &args[1].pre_expr {
                                     SugaredFieldIdentifier(contract_name, name) => {
                                         let contract_identifier = QualifiedContractIdentifier::new(
@@ -121,9 +218,8 @@ impl TraitsResolver {
                                     FieldIdentifier(trait_identifier) => trait_identifier.clone(),
                                     _ => return Err(ParseErrors::ImportTraitBadSignature.into()),
                                 };
-                                e.insert(TraitDefinition::Imported(trait_id));
-                            }
-                        }
+                                Ok::<_, ParseErrors>(TraitDefinition::Defined(trait_id))
+                            })?;
                     } else {
                         return Err(ParseErrors::ImportTraitBadSignature.into());
                     }
--- a/clarity/src/vm/costs/mod.rs
+++ b/clarity/src/vm/costs/mod.rs
@@ -14,10 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use std::collections::BTreeMap;
 use std::{cmp, fmt};
 
-use hashbrown::hash_map::Entry;
 use hashbrown::HashMap;
 use lazy_static::lazy_static;
 use rusqlite::types::{FromSql, FromSqlResult, ToSql, ToSqlOutput, ValueRef};
@@ -39,6 +37,7 @@ use crate::vm::types::{
     TypeSignature, NONE,
 };
 use crate::vm::{ast, eval_all, ClarityName, SymbolicExpression, Value};
+use crate::vm::ast::traits_resolver::EntryHashMapExt;
 
 pub mod constants;
 pub mod cost_functions;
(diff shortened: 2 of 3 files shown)


clarity/src/vm/ast/traits_resolver/mod.rs
```diff
@@ -63,7 +64,9 @@ impl TraitsResolver {
                     match (&args[0].pre_expr, &args[1].pre_expr) {
                         (Atom(trait_name), List(trait_definition)) => {
                             // Check for collisions
-                            if contract_ast.referenced_traits.contains_key(trait_name) {
+                            if let Entry::Occupied(_) =
```
jbencin, line 67: Using `Entry` will help performance if it's replacing both a check and an insert, because it avoids needing to hash the key twice. If you aren't using it for the insert, you might as well just use `contains_key()`. In order to make sense here, you need use the vacant entry later for the insert a few lines later

clarity/src/vm/ast/traits_resolver/mod.rs
```diff
@@ -94,7 +97,9 @@ impl TraitsResolver {
 
                     if let Some(trait_name) = args[0].match_atom() {
                         // Check for collisions
-                        if contract_ast.referenced_traits.contains_key(trait_name) {
+                        if let Entry::Occupied(_) =
+                            contract_ast.referenced_traits.entry(trait_name.to_owned())
+                        {
```
jbencin, line 102: Again, need to use this entry for `insert()`

jbencin (changes requested):

ASuciuX:
I've run the benchmark again after the latest commit ([a8597f1](https://github.com/stacks-network/stacks-core/pull/4483/commits/a8597f14c0a9b719c9eb77138df4099647fe66f7)), compared to the previous one ([885b6e5](https://github.com/stacks-network/stacks-core/pull/4483/commits/885b6e5f4d5d2223c01619d5b6231d593691b0cd)). The results show a decrease of `2%` in block processing time, above the initial `1%`:
```console
Benchmark 1: ./stacks-inspect-entry-no-inserts replay-block ~/chainstate range 99990 100000
  Time (mean ± σ):      6.694 s ±  0.011 s    [User: 6.310 s, System: 0.350 s]
  Range (min … max):    6.680 s …  6.711 s    10 runs
 
Benchmark 2: ./stacks-inspect-entry-inserts replay-block ~/chainstate range 99990 100000
  Time (mean ± σ):      6.587 s ±  0.009 s    [User: 6.227 s, System: 0.325 s]
  Range (min … max):    6.573 s …  6.606 s    10 runs
 
Summary
  ./stacks-inspect-entry-inserts replay-block ~/chainstate range 99990 100000 ran
    1.02 ± 0.00 times faster than ./stacks-inspect-entry-no-inserts replay-block ~/chainstate range 99990 100000
  ```
  In this case, `Benchmark 1` is `885b6e5`, and `Benchmark 2` is `a8597f1`.

File: stackslib/src/net/dns.rs [hunk 3]
```diff
@@ -292,18 +293,15 @@ impl DNSClient {
         loop {
             match self.requests_rx.try_recv() {
                 Ok(resp) => {
-                    if self.requests.contains_key(&resp.request) {
+                    if let Entry::Occupied(mut e) = self.requests.entry(resp.request.to_owned()) {
```
jbencin, line 296: I'd recommend `get_mut()` here, it's cleaner if you're modifying occupied entries:

```suggestion
                    if let Some(e) = self.requests.get_mut(resp.request) {
```

stackslib/src/net/dns.rs
```diff
@@ -324,30 +322,24 @@ impl DNSClient {
 
     pub fn poll_lookup(&mut self, host: &str, port: u16) -> Result<Option<DNSResponse>, net_error> {
         let req = DNSRequest::new(host.to_string(), port, 0);
-        if !self.requests.contains_key(&req) {
-            return Err(net_error::LookupError(format!(
+        match self.requests.entry(req.to_owned()) {
```
jbencin, line 325: Since you're not inserting on a vacant entry, use `get_mut()`

testnet/stacks-node/src/chain_data.rs
```diff
@@ -362,10 +363,14 @@ impl MinerStats {
 
         for (_, commit) in active_miners_and_commits.iter() {
             let addr = commit.apparent_sender.to_string();
-            if dist.contains_key(&addr) {
-                continue;
+            match dist.entry(addr.to_owned()) {
+                Entry::Occupied(_) => {
+                    continue;
```
jbencin, line 368: `continue` isn't necessary because we're at the end of the loop, how about changing this block to an `if let Entry::Vacant(e)...`

stackslib/src/net/inv/nakamoto.rs
```diff
@@ -546,14 +547,18 @@ impl<NC: NeighborComms> NakamotoInvStateMachine<NC> {
...
+            match self.reward_cycle_consensus_hashes.entry(rc) {
+                Entry::Vacant(e) => {
+                    let Some(ch) = Self::load_consensus_hash_for_reward_cycle(sortdb, rc)? else {
+                        // NOTE: this should be unreachable, but don't panic
+                        return Err(DBError::NotFoundError.into());
+                    };
+                    e.insert(ch);
+                }
+                Entry::Occupied(_) => {
+                    continue;
```
jbencin, line 559: `continue` isn't necessary because we're at the end of the loop, how about changing this block to an `if let Entry::Vacant(e)...`

stackslib/src/net/dns.rs
```diff
@@ -423,16 +416,18 @@ mod test {
...
-                    continue;
-                }
-                match client.poll_lookup(name, 80).unwrap() {
-                    Some(addrs) => {
-                        test_debug!("name {} addrs: {:?}", name, &addrs);
-                        resolved_addrs.insert(name.to_string(), addrs);
-                        break;
+                match resolved_addrs.entry(name.to_string()) {
+                    Entry::Occupied(_) => {
+                        continue;
```
jbencin, line 421: `continue` isn't necessary because we're at the end of the loop. How about changing this block to an `if let Entry::Vacant(e)...`

jbencin (commented):
Looks good, just some minor changes left to make and I'll approve

clarity/src/vm/database/key_value_wrapper.rs
```diff
@@ -541,6 +541,7 @@ impl<'a> RollbackWrapper<'a> {
                 "ERROR: Clarity VM attempted GET on non-nested context.".into(),
             )
         })?;
+
```
jbencin, line 544: Can you remote this newline?

stackslib/src/net/dns.rs
```diff
@@ -324,30 +322,27 @@ impl DNSClient {
...
-                host, port
-            )));
-        }
+        if let Some(e) = self.requests.get_mut(&req) {
+            let _ = match e {
+                None => {
+                    return Ok(None);
+                }
+                Some(resp) => resp,
+            };
```
jbencin, line 331: This can be replaced by
```suggestion
            if e.is_none() {
                return Ok(None);
            };
```

stackslib/src/net/dns.rs
```diff
@@ -324,30 +322,27 @@ impl DNSClient {
...
-                return Ok(None);
-            }
-            Some(Some(resp)) => resp,
-            None => {
-                unreachable!();
-            }
-        };
+            let resp = self
+                .requests
+                .remove(&req)
```
jbencin, line 335: I was wrong here, this was better using `Entry` because of the `remove()`

jbencin (dismissed):

jcnelson (changes requested):
See earlier comment

jbencin:
>  It takes several (sometimes 10x) lines of code to express that which can be done in a single line with contains_key()

This PR is currently at `+275, −247`. So there is an increase in line count, but it's pretty small

> Moreover, to use this API, you needed to manually transform lots of code on consensus-critical code paths, which makes merging this PR super risky

It's a simple and straightforward transformation, similar to transforming `if` statements into `match` statements. It also let's us remove some ugly `.expect()` calls like this:

```rust
let resp = self
    .requests
    .remove(&req)
    .expect("BUG: had key but then didn't")
    .expect("BUG: had response but then didn't");
```

IMO this makes the entry API safer to use than indexing a `HashMap` multiple times with other methods

The unit and integration tests are currently passing, but we can also do some extra testing here

> Finally, the speed improvements are basically nonexistent (a questionable 1% speedup?), which means the wins from the PR are basically nonexistent

The numbers in the description are incorrect, it's about a 2% speedup in block processing

jcnelson:
> It's a simple and straightforward transformation, similar to transforming if statements into match statements. It also let's us remove some ugly .expect() calls like this:

I'm really not seeing how this:

```rust
match foo.get(&bar).entry() {
   Entry::Occupied(value) => {
      /* do something */
   }
   Entry::Vacant(e) => {
      /* do something else */
   }
}
```

is a legibility improvement over this:

```rust
if foo.contains_key(&bar) {
   /* do something */
}
else {
   /* do something else */
}
```

The former has two levels of indentation, whereas the latter has one. Also, the former has 6 lines of boilerplate, whereas the latter has 4.  We've been trying to _reduce_ gratuitous indentations for the past few years because it aids in comprehensibility, but this PR works against it.

Do we know for sure that using the `Entires` API is required to gain that 2% speed improvement? Because, if the transformation is really that straightforward, then the compiler really should be doing that internally (or have an option to do so).

jbencin:
> I'm really not seeing how this... is a legibility improvement over this

I don't know about that. Sure the second example is less text, but the first is more explicit because both cases are clearly labeled, rather than using an `else` for one case. This example also misses that when using `Entry`, the code inside these blocks can be much more readable because there's no need to access the `HashMap` again. Consider this example from the PR:

```rust
if !self.prune_inbound_counts.contains_key(prune) {
    self.prune_inbound_counts.insert(prune.clone(), 1);
} else {
    let c = self.prune_inbound_counts.get(prune).unwrap().to_owned();
    self.prune_inbound_counts.insert(prune.clone(), c + 1);
}
```

vs.

```rust
match self.prune_inbound_counts.entry(prune.to_owned()) {
    Entry::Occupied(mut e) => {
        *e.get_mut() += 1;
    }
    Entry::Vacant(e) => {
        e.insert(1);
    }
}
```

Line count went up, but the second is fewer characters and **much** more readable

> Do we know for sure that using the Entires API is required to gain that 2% speed improvement?

The `hyperfine stacks-inspect replay-block` benchmark we use is very consistent between runs, so I'm confident this represents a performance gain. This is not surprising, it's halving the number of hashes needed in some places.

The perf gains are probably entirely from the few changes in clarity. We should at least merge those. I'm not sure the other changes in stackslib are even used in the `replay-block` benchmark

> Because, if the transformation is really that straightforward, then the compiler really should be doing that internally (or have an option to do so)

You'd think, but during my testing I've been surprised by what the compiler doesn't optimize away at `-O3`

jcnelson:
> The perf gains are probably entirely from the few changes in clarity. We should at least merge those. I'm not sure the other changes in stackslib are even used in the replay-block benchmark

Yup, 100% onboard with this.  Now that you mention it, it was surprising that these changes impacted code outside of `stacks_common` and `clarity`.

> Line count went up, but the second is fewer characters and much more readable. In the first, you have to read it and think about what it's doing, whereas it's obvious as soon as you look at the second that you're incrementing a counter. Also, there's no unwrap() needed when using the Entry

The example you're citing here could also be refactored as follows, with less boilerplate and the same number of clones:

```rust
if let Some(c) = self.prune_inbound_counts.get_mut(prune) {
   *c += 1;
} else {
   self.prune_inbound_counts.insert(prune.clone(), 1);
}
```

Granted, there's a lot of really old code in the `stackslib/src/net` directory that was written before I had a good handle on Rust, and it could really use a spit-shine.

jbencin:
Also, I think that using `if let` with `.entry()` is also the cleanest way to insert a key iff it doesn't exist already. Another example from the PR:

```rust
if !lowest_reward_cycle_with_missing_block.contains_key(nk) {
    lowest_reward_cycle_with_missing_block.insert(nk.clone(), reward_cycle);
}
```

vs.

```rust
if let Entry::Vacant(e) = lowest_reward_cycle_with_missing_block.entry(nk) {
    e.insert(reward_cycle);
}
```

Aside from the Clarity perf gains, there are some clear readability improvements and `.unwrap()` or `.expect()` eliminations in this PR that I think are worth keeping as well

jcnelson:
I disagree -- I think it requires more cognitive overhead.  Instead of reading "okay, insert this thing into the hash table if it doesn't have it", I'm reading "okay, get this opaque pointer-that-is-not-a-pointer thing ("entry") via a let-binding, and then call this deref-and-store-that-is-not-a-deref-and-store function on it ("insert") to put this thing into the hash table".  This is why I avoid the `Entry` API in my code -- I think it creates a needless layer of indirection that can be avoided with the containers' APIs.

That all said, I'm not really keen to pursue the bike-shedding further since I don't think we're going to ever come to agreement.  Also, it's off-topic.

The purpose of the workstream to which this PR belongs is to improve the speed of the Clarity VM.  A few comments ago, you mentioned that the speedup you see does not appear to come from the introduction of the `Entries` API (or if it does, there's no need to be touching files outside of `clarity` and perhaps `stacks_common`).  When viewed only through this lens, without regards to the `Entries` API, this PR is not achieving this -- it's making changes that do not improve the Clarity VM, and it's making changes in things outside of the Clarity VM.  As such, I'm leaving my review as-is until these two problems are addressed.

ASuciuX:
@jcnelson I've removed all the modifications outside of `clarity` package. Can you please review it?

cc: @jbencin

ASuciuX:
@jcnelson @jbencin I've removed the `.collect()` as it was failing the build, now it should work. Can you please re-approve this?

jbencin (approved):
I assume this was failing to compile because of a merge conflict?

ASuciuX:
> I assume this was failing to compile because of a merge conflict?

There were no previous merge conflicts, what I've seen was that the `.collect()` was previously implemented at the time when i created this PR, but on the latest `next` it was removed - that's why the compile was failing. I manually solved it today before updating the branch, so I didn't see any conflicts when re-syncing.

jbencin:
It looks like it's because I changed `probe_for_generics()` to accept `impl Iterator` in #4500

cylewitruk (changes requested):
I've requested (via Slack) that we try to retain the readability of `contains_key` via a new trait or method, since these changes make the code not as obvious as to what it's doing. This is probably something that's done in several other places throughout stacks-core, so having such a util method would likely be beneficial.

jbencin:
> I've requested (via Slack) that we try to retain the readability of `contains_key` via a new trait or method

Do you have a suggestion for how to do that?

cylewitruk:
> > I've requested (via Slack) that we try to retain the readability of `contains_key` via a new trait or method
> 
> Do you have a suggestion for how to do that?

Yeah I gave a few to @ASuciuX on Slack -- but if it proves difficult I'm happy to take it.

cylewitruk (changes requested):

clarity/src/vm/ast/traits_resolver/mod.rs
```diff
@@ -63,26 +64,30 @@ impl TraitsResolver {
...
+
+                                    let trait_id = TraitIdentifier {
+                                        name: trait_name.clone(),
+                                        contract_identifier: contract_ast
+                                            .contract_identifier
+                                            .clone(),
+                                    };
+                                    e.insert(TraitDefinition::Defined(trait_id));
+                                }
                             }
```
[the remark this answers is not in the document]
jbencin: One minor nitpick here: We should be consistent with Rust naming conventions and use
 - `map` for infallible conversions
 - `and_then` for fallible conversions which return a `Result`

Otherwise this looks cool and would be a nice addition to the `StacksHashMap` type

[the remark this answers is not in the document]
ASuciuX: I'll finish the stacking testing and do into this afterwards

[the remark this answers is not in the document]
cylewitruk: @jbencin recommend `occupied_and_then()` + `vacant_and_then()` in that case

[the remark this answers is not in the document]
cylewitruk: @ASuciuX here comes an update for the `move` issue you ran into:

**Trait + Impl**
```rust
pub trait EntryExt<'a, K: 'a, V: 'a, S: 'a>
where
    S: BuildHasher,
    Self: Sized,
{
    fn occupied_and_then<T, F, E>(self, f: F) -> Result<Self, E>
    where
        F: FnOnce(&OccupiedEntry<'a, K, V, S>) -> Result<T, E>;

    fn or_insert_fallible<F, E>(self, default: F) -> Result<&'a mut V, E>
    where
        F: FnOnce() -> Result<V, E>;

    fn vacant_or<E>(self, err: E) -> Result<Self, E>;
}

impl<'a, K: 'a, V: 'a, S: 'a> EntryExt<'a, K, V, S> for Entry<'a, K, V, S>
where
    S: BuildHasher,
    K: Hash
{    
    fn occupied_and_then<T, F, E>(self, f: F) -> Result<Self, E>
    where 
        F: FnOnce(&OccupiedEntry<'a, K, V, S>) -> Result<T, E>
    {
        match self {
            Entry::Occupied(ref e) => {
                f(e)?;
                Ok(self)
            },
            Entry::Vacant(_) => Ok(self)
        }
    }

    fn or_insert_fallible<F, E>(self, default: F) -> Result<&'a mut V, E>
    where
        F: FnOnce() -> Result<V, E>
    {
        match self {
            Entry::Occupied(e) => Ok(e.into_mut()),
            Entry::Vacant(e) => Ok(e.insert(default()?)),
        }
    }
    
    fn vacant_or<E>(self, err: E) -> Result<Self, E> {
        match self {
            Entry::Occupied(_) => Err(err),
            Entry::Vacant(_) => Ok(self)
        }
    }
}
```

**Usage**
```rust
match (&args[0].pre_expr, &args[1].pre_expr) {
    (Atom(trait_name), List(trait_definition)) => {
        contract_ast
            .referenced_traits
            .entry(trait_name.to_owned())
            .vacant_or(ParseErrors::NameAlreadyUsed(trait_name.to_string()))?
            .or_insert_fallible(|| {
                self.probe_for_generics(
                    trait_definition.iter(),
                    &mut referenced_traits,
                    true,
                ).map_err(|e| e.err)?;

                let trait_id = TraitIdentifier {
                    name: trait_name.clone(),
                    contract_identifier: contract_ast
                        .contract_identifier
                        .clone(),
                };
                Ok::<_, ParseErrors>(TraitDefinition::Defined(trait_id))
            })?;
    }
    _ => return Err(ParseErrors::DefineTraitBadSignature.into()),
}
```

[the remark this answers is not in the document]
cylewitruk: @jbencin also updated naming in the above. couldn't think of anything awesome for `or_insert_fallible`... `or_insert` etc. are already taken by hashbrown itself...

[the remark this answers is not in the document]
cylewitruk: `or_try_insert` perhaps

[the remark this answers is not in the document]
cylewitruk: @jcnelson I invite you to give your input on the above as well as you had comments on the original impl.

ASuciuX:
If there is a more appropriate place for the traits and implements please let me know and I will move them.

jbencin:
> If there is a more appropriate place for the traits and implements please let me know and I will move them.

If @cylewitruk can merge his `StacksHashMap` changes, that would be the best place to put it. Or if not, this should go in it's own file, maybe in stacks-common/src/util or stacks-common/src/types

cylewitruk:
> If @cylewitruk can merge his `StacksHashMap` changes, that would be the best place to put it. Or if not, this should go in it's own file, maybe in stacks-common/src/util or stacks-common/src/types

I might actually suggest starting a new dir stacks_common/src/ext where all "extension" traits can be placed. This pattern can be applied to a lot of repetitive code..

jcnelson:
Closing due to inactivity

PR #4488, opened 2024-03-05
fix: expected reward cycle in integration test for pox info

hstove:
The `neon_integrations::pox_integration_test` has been failing CI pretty consistently (but not 100%?), such as in this build: https://github.com/stacks-network/stacks-core/actions/runs/8161203003/job/22313964190?pr=4461

This changes the test to not use hard-coded reward cycles that are expected, and instead to calculate the reward cycle from the current burnchain height.


kantai (approved):
LGTM -- thanks for this -- my PR that changed pox info RPC handling must have started tripping this up

obycode (approved):
lgtm! 👍

PR #4460, opened 2024-02-29
[CI] Dry x86-64-v3 builds/default images to use glibc 

wileyj:
ref:
https://github.com/stacks-network/stacks-core/pull/4364
https://github.com/stacks-network/stacks-core/issues/4383

This adjusts the changes introduced by @jbencin that were merged in 4364 for jemalloc, and adding an additional cpu arch `x86-64-v3`. 

- removes the duplicated workflow to build x86-64-v3 binaries (current workflow was DRY'ed to build them via a matrix with some specific excludes)
- fixes the dockerfiles currently in `./build-scripts` to set the RUSTFLAGS env var if TARGET_CPU build-arg is provided 
- adjusts the conditional compilation flags in several files to exclude the target_os (required when cross-compiling - target_env is only application when building on a specific target triple. for github actions, the excude of `target_env = msvc` will not match to true since default is ubuntu vm). 
- changes the default release docker image to use debian glibc as opposed to alpine (alpine is still built, but is tagged as `<release version>-alpine`
- builds the docker source image (which is triggered on any PR) with `TARGET_CPU` provided as `x86-64-v3`
-  release docker images are changed to support the following cpu types: x86-64-v3, arm64, armv7. *if desired, i **think** we can add support for x86-64 alongside x86-64-v3*

sample actions runs from my fork (testing these chagnes on a release):
https://github.com/wileyj/stacks-blockchain/actions/runs/8087234345/job/22100013368
release: https://github.com/wileyj/stacks-blockchain/releases/tag/3.0.0.0-rc1
images:
[debian release image](https://hub.docker.com/layers/wileyj/stacks-core/3.0.0.0-rc1/images/sha256-507894b983b96fd73df6740c665b83c465ee14bf8de5fe4589d57771642725c3?context=explore)
[alpine release image](https://hub.docker.com/layers/wileyj/stacks-core/3.0.0.0-rc1-alpine/images/sha256-0ed64449d3f504c191c0a7511018e46b95bb33b19fe46bd3a21f09d1d161e8ce?context=explore)

sample action for the source docker image:
https://github.com/wileyj/stacks-blockchain/actions/runs/8101666085/job/22142408056
resulting image:
[debian x86-64-v3 image from source](https://hub.docker.com/layers/wileyj/stacks-core/master/images/sha256-7480669cbaaf7ee85f9b065d2cbcdb16db3adf61758875f67241c6ab16a0d9c2?context=explore)

PR #4489, opened 2024-03-05
feat: gate `target-cpu=native` with `portable` feature

obycode:
We believe this flag can cause some problems with docker images.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -3,9 +3,11 @@ stacks-node = "run --package stacks-node --"
 fmt-stacks = "fmt -- --config group_imports=StdExternalCrate,imports_granularity=Module"
 
 # For x86_64 CPUs, default to `native` and override in CI for release builds
-# This makes it slightly faster for users running locally built binaries
+# This makes it slightly faster for users running locally built binaries.
+# This can cause trouble when building "portable" binaries, such as for docker,
+# so disable it with the "portable" feature.
 # TODO: Same for other targets?
-[target.'cfg(all(target_arch = "x86_64"))']
+[target.'cfg(all(target_arch = "x86_64", not(feature = portable))']
 rustflags = ["-Ctarget-cpu=native"]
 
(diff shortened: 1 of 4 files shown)


CharlieC3 (approved):
Related PR: https://github.com/stacks-network/stacks-core/pull/4460

The PR linked above explicitly defines the CPU target when an image is being built via GHA. But when this Docker image is being built locally (which I've had to do numerous times) intended for cloud deployments, it'll be helpful for have this feature flag in place to ensure the image is being built the same way it's always been, which has been compatible with most cloud infra.

I would suggest getting @wileyj's approval in this before merging as well.

obycode:
So maybe it would be good to have that "portable" feature enabled conditionally, with an environment variable for example?

CharlieC3:
> So maybe it would be good to have that "portable" feature enabled conditionally, with an environment variable for example?

@obycode I actually like the way you have it right now. In the scenario I painted above when I need to build the image locally, I use the `.github/actions/dockerfiles/Dockerfile.debian-source` Dockerfile. So this PR effectively helps my specific workflow while leaving the others untouched.
So if a dev wants to build from source, unless they're using the `.github/actions/dockerfiles/Dockerfile.debian-source`, it would still build using the `native` CPU target. If there's an ENV var required, I feel it's another manual step I may forget.

obycode:
> @obycode I actually like the way you have it right now. In the scenario I painted above when I need to build the image locally, I use the `.github/actions/dockerfiles/Dockerfile.debian-source` Dockerfile. So this PR effectively helps my specific workflow while leaving the others untouched. So if a dev wants to build from source, unless they're using the `.github/actions/dockerfiles/Dockerfile.debian-source`, it would still build using the `native` CPU target. If there's an ENV var required, I feel it's another manual step I may forget.

Ah, got it. Thanks for clarifying!

zone117x (approved):

Issue #4484, opened 2024-03-05
Miner is repeatedly hitting "Have only attempted one block; unconditionally trying again"


obycode:
The miner is repeatedly hitting "Have only attempted one block; unconditionally trying again" and retrying, every 5-10 ms in my environment.

Sample logging:
```
FO [1709594218.842938] [testnet/stacks-node/src/neon_node.rs:1254] [miner-block-http://127.0.0.1:20443] Picking best Stacks tip
INFO [1709594218.843593] [testnet/stacks-node/src/neon_node.rs:1420] [miner-block-http://127.0.0.1:20443] Tip #0 a567f18f56a283874acca24324ba7b72c985db39/5c2d18470498c2a37bb47efe03ad51397541b9d950ec5da187ffeb7937466d19 at 111:9 has score 0 (0 (earlier sibs) + 0 (earlier sibs) + 0 (earlier sibs))
INFO [1709594218.843604] [testnet/stacks-node/src/neon_node.rs:1451] [miner-block-http://127.0.0.1:20443] Best tip is #0 a567f18f56a283874acca24324ba7b72c985db39/5c2d18470498c2a37bb47efe03ad51397541b9d950ec5da187ffeb7937466d19
[...]

PR #4495, opened 2024-03-06
Bugfix/do not unconditionally attempt block post submit failure

jferrant:
This is to quiet the logs which were unconditionally reattempting block submissions with the exact same info post failure due to an attempt to handle the case where the first attempted block was much smaller than subsequent attempts would be. @kantai gave me a diff...
Closes https://github.com/stacks-network/stacks-core/issues/4484

### DIFF
--- a/testnet/stacks-node/src/neon_node.rs
+++ b/testnet/stacks-node/src/neon_node.rs
@@ -516,6 +516,8 @@ pub(crate) struct BlockMinerThread {
     burn_block: BlockSnapshot,
     /// Handle to the node's event dispatcher
     event_dispatcher: EventDispatcher,
+    /// Failed to submit last attempted block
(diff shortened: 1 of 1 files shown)


hstove (approved):
This seems good to me - is this the issue where the logs go in an infinite loop of submitting the same mining tx?

jferrant:
> This seems good to me - is this the issue where the logs go in an infinite loop of submitting the same mining tx?

Yes. It still will infinitely loop as this does not affect retry logic much, but it will at least return the appropriate error and abort earlier than before.

obycode:
I worry that this makes the code more confusing just to remove this one log message, while the rest of the logs, coming from the block building process, are still there. Is this definitely worth changing?

jcnelson (approved):

PR #4497, opened 2024-03-06
Move the Creation of Source Binaries to Composite

BowTiedDevOps:
### Description
I have moved the creation of the source binaries for release (along with the `build-scripts` directory) to the actions repository (https://github.com/stacks-network/actions/tree/main/stacks-core/create-source-binary/action.yml).
The change is useful because every time a change has to be made to the workflow (or the dockerfiles behind it), there's only a branch that needs to be changed instead of more, and any future changes are made into the `actions` repository instead of `stacks-core`.

This PR precedes https://github.com/stacks-network/stacks-core/pull/4368. Shout-out to @wileyj for fixing the dockerfiles and helping with the composite failures.

Successful run (Only the `Create Release` job is affected by this PR): https://github.com/BowTiedDevOps/stacks-core/actions/runs/8177217259

**Note**: These changes are dependent on `ci/dry_x64_builds` branch, this PR can only be merged after https://github.com/stacks-network/stacks-core/pull/4460 is.


.github/workflows/create-source-binary.yml
```diff
@@ -30,41 +24,41 @@ jobs:
...
-        id: compress_artifact
-        run: zip --junk-paths ${{ matrix.platform }} ./release/${{ matrix.platform }}/*
-
-      ## Upload the binary artifact to the github action (used in `github-release.yml` to create a release)
-      - name: Upload artifact
-        id: upload_artifact
-        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+      - name: Build Binary (${{ matrix.arch }}_${{ matrix.cpu }})
+        id: build_binary
+        timeout-minutes: ${{ fromJSON(env.TEST_TIMEOUT) }}
```
CharlieC3, line 59: Why not just use `${{ env.TEST_TIMEOUT }}` here?

CharlieC3 (dismissed):
LGTM! Although it creates a fair amount of misdirection, it's far better to decouple these kinds of actions from the release schedule of stacks core.

.github/workflows/create-source-binary.yml
wileyj: @BowTiedDevOps  I would argue this env var shouldn't even be here. 
if a timeout is desired, it should be present in the build step of the composite action. 

i've never seen a timeout from a build though (only on tests where it's stuck in a loop)  so i'm inclined to say this should be removed.

wileyj:
ref: #4498

wileyj:
ping for pr review

CharlieC3 (approved):

obycode (approved):
lgtm 👍

wileyj:
@BowTiedDevOps check on the conflicts here please

obycode (approved):

PR #4363, opened 2024-02-10
docs: mutation-testing

ASuciuX:
Added mutation testing-related info:
- testing docs file
  - how to run mutation testing locally
- mutation testing check on the PR template
- contributing section for how mutation testing should be taken into account when contributing

PR #4498, opened 2024-03-06
fix: `target-cpu=native` flag working again

jbencin:
### Description

Pass the flag `target-cpu=native` to `rustc` for local builds and tests for improved performance

**This PR also require stacks-network/actions#31 to be approved and merged**

### Applicable issues

- #4316 
- #4364: Original commit enabling this flag
- #4489: Broke flag being passed to `rustc` (not sure why)

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -2,12 +2,11 @@
 stacks-node = "run --package stacks-node --"
 fmt-stacks = "fmt -- --config group_imports=StdExternalCrate,imports_granularity=Module"
 
-# For x86_64 CPUs, default to `native` and override in CI for release builds
-# This makes it slightly faster for users running locally built binaries.
+# Default to `native`
+# This makes it slightly faster for running tests and locally built binaries.
 # This can cause trouble when building "portable" binaries, such as for docker,
 # so disable it with the "portable" feature.
-# TODO: Same for other targets?
-[target.'cfg(all(target_arch = "x86_64", not(feature = portable))']
+[target.'cfg(not(feature = "portable"))']
 rustflags = ["-Ctarget-cpu=native"]
 
 # Needed by perf to generate flamegraphs.
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN apk add --no-cache musl-dev
 
 RUN mkdir /out
 
-RUN cd testnet/stacks-node && cargo build --features monitoring_prom,slog_json --release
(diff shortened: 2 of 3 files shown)


wileyj:
this changes quite a lot when considering ti's for 
> Pass the flag target-cpu=native to rustc for local builds and tests for improved performance

what it is goal of this PR?

jbencin:
> this changes quite a lot when considering ti's for

It touches 12 of files, but it's only adding a single word in 11 of them

> what it is goal of this PR?

It's a few percent perf improvements on local builds. See the discussion on the Slack `#performance` channel. I added this in #4363, but it was broken in #4489

wileyj:
> > this changes quite a lot when considering ti's for
> 
> It touches 12 of files, but it's only adding a single word in 11 of them
> 
> > what it is goal of this PR?
> 
> It's a few percent perf improvements on local builds. See the discussion on the Slack `#performance` channel. I added this in #4363, but it was broken in #4489

i guess that's my question - you've mentioned "local builds" twice now, but there are only 2 Dockerfiles that *should* ever be used to build locally: `Dockerfile` and `Dockerfle.debian`. the rest are quite specific to PR and release builds. 
 you could make the case to add the change here as well: `./.github/actions/dockerfiles/Dockerfile.debian-source`.

wileyj:
if #4489 added a regression, wouldn't it be simpler to revert that change vs adding more changes?

wileyj (changes requested):
requesting changes as a placeholder, i think a call is in order

wileyj:
@jbencin can PR the `./build-scripts` changes here as well? there is an open PR to start using these files that hasn't merged into `next` yet. 
https://github.com/stacks-network/actions/tree/main/stacks-core/create-source-binary/build-scripts

we can discuss concerns around removing these from the stacks-core repo elsewhere, my main concern is merging this PR, then not having these chagnes in teh `stacks-core/build-scripts` is removed in another open PR

wileyj (dismissed):
shipit

wileyj:
ref: https://github.com/stacks-network/stacks-core/pull/4497

obycode (dismissed):
ok, I think this seems fine. 👍

obycode:
Looks like there are still conflicts here @jbencin

wileyj:
> Looks like there are still conflicts here @jbencin

it's related to #4497

jcnelson (approved):

obycode (approved):

wileyj (approved):
shipit

PR #4499, opened 2024-03-06
fix: use testnet consts when generating xenon config

hstove:
I was playing around with running a xenon node off of the `next` branch (which I know won't work), and I caught a startup error that originates from validation of burnchain epochs. Looking in to the code, I caught that we were using mainnet consts instead of testnet consts.

I know these consts will definitely change, but I'd still like to get this merged, because I'm trying to help signers get their infra setup. Even if their node doesn't sync any Stacks blocks, it would be helpful to not have a startup crash.


8marz8 (approved):

obycode (approved):
👍

PR #4505, opened 2024-03-07
update signer logging to include reward cycle to differentiate curren…

jferrant:
added reward cycle to signer logs to differentiate between current and next reward cycle signer state machines


hstove (dismissed):
Approving because I think it's better to merge now, but I think the size of the diff here demonstrates that it would be helpful to have some kind of `log_label(signer)` helper. For example, it might also be prudent to log the pubkey or address

But that's a nit, this LGTM

xoloki (dismissed):
LGTM

obycode (dismissed):

jferrant:
> Approving because I think it's better to merge now, but I think the size of the diff here demonstrates that it would be helpful to have some kind of `log_label(signer)` helper. For example, it might also be prudent to log the pubkey or address
> 
> But that's a nit, this LGTM

Yes I considered this but assumed could be done later.

xoloki (approved):
LGTM

hstove (approved):
LGTM!

Issue #4490, opened 2024-03-06
[Nakamoto] Include signer-key (and other args) in `stack-increase` events


hstove:
In https://github.com/stacks-network/stacks-core/pull/4461, we added `signer-key` to `stack-increse`, along with a few other new arguments. In that PR, we aren't emitting these new arguments via synthetic PoX events. We should, so that the API can index them properly.

To summarize:

- Update `stack-increase` event with signer key, signature, max amount, and auth ID
- Update `stack-stx`, `stack-extend`, and `stack-aggregation-commit` with max amount and auth ID

PR #4508, opened 2024-03-08
feat: add signer-key to stack-increase event

hstove:
This does two things:

- Adds `max-amount` and `auth-id` to synthetic events for `stack-stx`, `stack-extend`, and `stack-aggregation-commit`
- Adds all the signer key and related args to the synthetic event for `stack-increase`

- Closes https://github.com/stacks-network/stacks-core/issues/4490


hstove:
This will conflict with #4414. I'm happy to add this to that PR or the opposite, just lmk @8marz8 @zone117x

8marz8 (approved):

PR #4510, opened 2024-03-08
Have miners abort signing a block if the burn tip changes

jferrant:
Miners should abort waiting for a signer signature if the burnchain tip has changed as the block will be invalidated.


obycode (approved):
🚀

hstove (approved):

PR #4511, opened 2024-03-09
Feat: use block proposal struct for miner -> signers communication

kantai:
This uses a struct for communicating a block proposal from the miner to signers.

The signer also checks the claimed reward cycle of the block before attempting to validate the proposal.


obycode (approved):
LGTM!

PR #4513, opened 2024-03-09
Bugfix/cleanup outdated signers

jferrant:
Signers need to remove outdated reward cycle signer states from their map

### DIFF
--- a/stacks-signer/src/runloop.rs
+++ b/stacks-signer/src/runloop.rs
@@ -280,12 +280,12 @@ impl RunLoop {
         // TODO: do not use an empty consensus hash
         let pox_consensus_hash = ConsensusHash::empty();
         let mut to_delete = Vec::new();
-        for signer in self.stacks_signers.values_mut() {
+        for (idx, signer) in &mut self.stacks_signers {
             if signer.reward_cycle < current_reward_cycle {
                 debug!("{signer}: Signer's tenure has completed.");
                 // We don't really need this state, but it's useful for debugging
                 signer.state = SignerState::TenureCompleted;
-                to_delete.push(signer.reward_cycle % 2);
+                to_delete.push(*idx);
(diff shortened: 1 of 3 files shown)


stacks-signer/src/runloop.rs
```diff
@@ -279,7 +279,14 @@ impl RunLoop {
         self.refresh_signer_config(next_reward_cycle);
         // TODO: do not use an empty consensus hash
         let pox_consensus_hash = ConsensusHash::empty();
+        let mut to_delete = Vec::new();
         for signer in self.stacks_signers.values_mut() {
+            if signer.reward_cycle < current_reward_cycle {
+                debug!("{signer}: Signer's tenure has completed.");
+                // We don't really need this state, but it's useful for debugging
+                signer.state = SignerState::TenureCompleted;
+                to_delete.push(signer.reward_cycle % 2);
```
obycode, line 288: I'm still going through this logic to understand what's happening, but my current question is how do we ensure that the signer for say cycle N-2 does not cause the signer for cycle N to get deleted?

xoloki: Yeah it seems a little strange that this HashMap is indexed by whether the reward cycle is even or odd.  I get that we only care about either last/current, or current/next cycles.  But it seems like there should be a cleaner way to do this.

obycode: Looking at this again, I see why you're doing this. Really I'd like to just do the following before the loop, but then we don't get a log for each one.

```rust
        // Filter out signers for cycles which have now passed.
        self.stacks_signers
            .retain(|_, signer| signer.reward_cycle >= current_reward_cycle);
```

obycode: I just pushed a commit to make it a little more clear.

PR #4514, opened 2024-03-09
fix: try not deleting accepted blocks from signerDB

hstove:
In looking at testnet logs, I've seen that we hit scenarios where a block is saved as "accepted", which deletes the block from SignerDB, and then shortly later the signer tries to process the block, but the signer has no about the block (because it's deleted).

image

### DIFF
--- a/stacks-signer/src/signer.rs
+++ b/stacks-signer/src/signer.rs
@@ -73,7 +73,7 @@ pub struct BlockInfo {
     /// The associated packet nonce request if we have one
     nonce_request: Option<NonceRequest>,
     /// Whether this block is already being signed over
-    signed_over: bool,
+    pub signed_over: bool,
 }
 
 impl BlockInfo {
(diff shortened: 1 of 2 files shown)


stacks-signer/src/signer.rs
```diff
@@ -992,10 +992,12 @@ impl Signer {
...
-        // TODO: proper garbage collection...This is currently our only cleanup of blocks
-        self.signer_db
-            .remove_block(&block_vote.signer_signature_hash)
-            .expect(&format!("{self}: Failed to remove block from to signer DB"));
+        // WIP: try not deleting a block from signerDB until we have a better garbage collection strategy.
+        // This causes issues when we have to reprocess a block and we have already deleted it from the signerDB
+        // // TODO: proper garbage collection...This is currently our only cleanup of blocks
+        // self.signer_db
+        //     .remove_block(&block_vote.signer_signature_hash)
+        //     .expect(&format!("{self}: Failed to remove block from to signer DB"));
```
kantai, line 1000: Yep -- I'd advocate just dropping this line permanently (no need to comment out). Now that this information is stored in a file, I don't think that there's a strong need to clean this state up.

kantai (approved):

PR #4516, opened 2024-03-11
Minor changes / clean up

obycode:


stacks-signer/src/runloop.rs
```diff
@@ -266,7 +266,11 @@ impl RunLoop {
                     .insert(reward_index, Signer::from(new_signer_config));
                 debug!("Reward cycle #{reward_cycle} Signer #{signer_id} initialized.");
             } else {
-                warn!("Signer is not registered for reward cycle {reward_cycle}. Waiting for confirmed registration...");
+                if current {
+                    warn!("Signer is not registered for the current reward cycle ({reward_cycle}). Waiting for confirmed registration...");
```
obycode, line 270: Maybe this should actually be an error log?

hstove: Yeah, except when booting up the first time when setting up your signer, you'd always see that. Maybe I'm wrong on that assumption if `refresh_signer_config` isn't called until a new cycle starts

hstove (approved):

stacks-signer/src/runloop.rs
obycode: I'll check this.

PR #4517, opened 2024-03-11
feat: add const to determine whether to rotate coordinator

hstove:
Added logic to use a const to determine whether to rotate coordinators when `update_coordinator` is called


kantai (approved):

obycode (approved):
lgtm!

PR #4520, opened 2024-03-11
feat: return any value type from `execute_contract_allow_private`

hugocaillard:
The helper `execute_contract_allow_private` does not allow the private function to return any type and can throw `PublicFunctionMustReturnResponse` if the private function return a non-Response value.

On Clarinet, we get recurrent requests from the community and internal teams to add the ability to call private function in unit tests 

### Checklist

- [ ] Changelog is updated
- N/A Required documentation changes (e.g., `docs/rpc/openapi.yaml` and `rpc-endpoints.md` for v2 endpoints, `event-dispatcher.md` for new events)
- N/A New clarity functions have corresponding PR in `clarity-benchmarking` repo
- N/A New integration test(s) added to `bitcoin-tests.yml`

### DIFF
--- a/clarity/Cargo.toml
+++ b/clarity/Cargo.toml
@@ -55,4 +55,4 @@ default = []
 developer-mode = []
 slog_json = ["stacks_common/slog_json"]
 testing = []
-
+devtools = []
--- a/clarity/src/vm/callables.rs
+++ b/clarity/src/vm/callables.rs
@@ -340,8 +340,8 @@ impl DefinedFunction {
     pub fn apply(&self, args: &[Value], env: &mut Environment) -> Result<Value> {
         match self.define_type {
             DefineType::Private => self.execute_apply(args, env),
-            DefineType::Public => env.execute_function_as_transaction(self, args, None),
-            DefineType::ReadOnly => env.execute_function_as_transaction(self, args, None),
+            DefineType::Public => env.execute_function_as_transaction(self, args, None, false),
+            DefineType::ReadOnly => env.execute_function_as_transaction(self, args, None, false),
         }
     }
 
--- a/clarity/src/vm/contexts.rs
+++ b/clarity/src/vm/contexts.rs
@@ -1739,7 +1739,7 @@ impl<'a, 'hooks> GlobalContext<'a, 'hooks> {
                     self.roll_back()?;
                 }
                 Ok(Value::Response(data))
-            } else if allow_private {
+            } else if allow_private && cfg!(feature = "developer-mode") {
                 self.commit()?;
                 Ok(result)
             } else {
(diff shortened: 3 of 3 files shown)


jcnelson (changes requested):
I think this should be epoch-gated so we don't accidentally use it prior to epoch 2.5

hugocaillard:
@jcnelson @kantai 
Following your remarks (here and on slack), I pushed a new commit that only have effect is developer-mode (which clarinet uses)
```rs
            else if allow_private && cfg!(feature = "developer-mode") {
                self.commit()?;
                Ok(result)
            }
```

@jcnelson 
> I think this should be epoch-gated so we don't accidentally use it prior to epoch 2.5

Ideally, I'd like to be able to use it for any epoch in simnet

hugocaillard:
@kantai 

>  as long as its controlled with a compile flag

Right now i'm using the `cfg!` macro, which, as I understand is slightly different from a compile flag (`#[cgf(...)]`). I could use a compile flag instead but it would make the code slightly less readable I think
```rs
    pub fn handle_tx_result(
        &mut self,
        result: Result<Value>,
        allow_private: bool,
    ) -> Result<Value> {
        if let Ok(result) = result {
            if let Value::Response(data) = result {
                if data.committed {
                    self.commit()?;
                } else {
                    self.roll_back()?;
                }
                Ok(Value::Response(data))
            } else {
                #[cfg(feature = "developer-mode")]
                if allow_private {
                    self.commit()?;
                    return Ok(result);
                }
                Err(
                    CheckErrors::PublicFunctionMustReturnResponse(TypeSignature::type_of(&result)?)
                        .into(),
                )
            }
        } else {
            self.roll_back()?;
            result
        }
    }
```

---
    
>  something specifically like allow-private

I'm starting to think that just "allow-private" is confusing. Because it should really be `allow-private-to-return-any-type` (or something like that)

Should I just introduce a flag like "devtools" or "clarinet"? I'm sure I can find tons of usage for such a flag in the future

kantai:
> Right now i'm using the cfg! macro, which, as I understand is slightly different from a compile flag (#[cgf(...)]). I could use a compile flag instead but it would make the code slightly less readable I think

Ah, yes, the `cfg!` macro is fine here. I agree it results in more readable code.

> Should I just introduce a flag like "devtools" or "clarinet"? I'm sure I can find tons of usage for such a flag in the future

Yes -- either works.

hugocaillard:
Done @kantai, I went for "devtools"

kantai (approved):
LGTM!

jcnelson (dismissed):
LGTM, but before merging, please add a comment to `handle_tx_result()` to indicate that `allow_private` is ignored unless the node is compiled with the `devtools` feature (the comment should include the line that they'd have to add to their `Cargo.toml` or the build flags to pass to `cargo`).

hugocaillard:
Done @jcnelson

kantai (approved):
LGTM!

jcnelson (approved):

PR #4522, opened 2024-03-12
fix: `stacked_amt: u128` should serialize as string

zone117x:
The `stacked_amt: u128` reward set property in `/new_block` event was being serialized as a json number which is unsafe for the majority of applications which treat these as 64-bit floating point. This PR fixes the issue by serializing the large integer as a string like we do in other places. 

Note: most other json serialization for u128 values in the codebase are manually constructed rather than going directly from a rust struct, so a new serialize/deserialize helper was added for this case.


hstove (approved):
LGTM

obycode (approved):
LGTM!

PR #4523, opened 2024-03-12
Fixes for clarity-wasm map-type functions

Acaccia:
The clarity type `Map` had a mistake in all its related functions: the key/values read from memory in the linked functions should be able to dereference them or not. This PR adds this capability.

Tests on those fixes also showed a bug while writing the values into the Wasm memory. It seems we had a few mistakes in the computation of the type sizes for lists and buffers. This PR also adds this fix.

As usual for PRs related to clarity-wasm, the tests are in the clarity-wasm repo.


obycode (approved):
Thanks @Acaccia. LGTM! 🚀

PR #4524, opened 2024-03-12
feat: add `pox_stx_threshold` amount to `/new_block` event data

zone117x:
Event indexers (e.g. the Stacks API) analyze historical pox-4 events to generate rich signer data. One of the challenges is determining which past stacking transactions should be considered for a given pox cycle. In this case, we want the ability to know if a historical pox-4 stacking event was discarded because of not meeting the minimum stx threshold (i.e. not given any slots/weight).

This PR adds the `(stx) threshold` value used for a given cycle's reward calculations to the `/new_block` event payload.


stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -843,6 +873,7 @@ impl StacksChainState {
                 missed_reward_slots: missed_slots,
             },
             signers: signer_set,
+            pox_stx_threshold: Some(threshold),
```
obycode, line 876: Maybe this should be `pox_ustx_threshold`?

zone117x: Done!

obycode (dismissed):
lgtm!

hstove (dismissed):

kantai (approved):

hstove (approved):

PR #4525, opened 2024-03-12
fix: add reward cycle to signerdb

obycode:
The blocks should always be associated with a specific reward cycle so that the signers do not take actions on blocks from the previous cycle. This should prevent a halt like the one that we saw this morning.

### DIFF
--- a/stacks-signer/src/signer.rs
+++ b/stacks-signer/src/signer.rs
@@ -317,7 +317,7 @@ impl Signer {
                 let signer_signature_hash = block.header.signer_signature_hash();
                 let mut block_info = self
                     .signer_db
-                    .block_lookup(&signer_signature_hash)
+                    .block_lookup(self.reward_cycle, &signer_signature_hash)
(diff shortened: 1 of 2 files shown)


hstove:
The code looks good to me, but it's odd that `signer_signature_hash` would be the same across different burn blocks. Looking at [the implementation](https://github.com/stacks-network/stacks-core/blob/e3a76f7c3c2327c22ea2d5de39bb33238931fd01/stackslib/src/chainstate/nakamoto/mod.rs#L445), wouldn't `consensus_hash` be different? And/or `miner_signature`?

If not, then we could add `burn_block_height` to `BlockInfo` and modify `signer_signature_hash` to include it.

obycode:
> The code looks good to me, but it's odd that `signer_signature_hash` would be the same across different burn blocks. 

What happened in the last stall was that the block validation approval for a block in cycle N-1 arrived during cycle N, and the signer did not know that it was intended for the previous cycle's signer.

xoloki (approved):
LGTM

hstove (approved):
LGTM

PR #4526, opened 2024-03-12
feat: include signer bitvec in nakamoto block event

### LINKED ISSUE #4257: [Nakamoto] Emit signer performance data
Related to signer support in the API & Explorer https://github.com/hirosystems/explorer/issues/1248

The API and Explorer would like to show signer performance, e.g. a record of uptime and slot success rate.

It sounds like some changes in addition to the event emitter may be required:
> I think this requires an extension to the NakamotoBlock event: inclusion of which stackers signed off on the block. The FROST protocol doesn't actually make this recoverable, but stacks-core needs this in place anyways to punish non-responsive signers, so it should be easy to add once that's in place.

Note: this is probably a nice-to-have and lower priority compared to the other signer related events.

hstove:
This includes the `signer_bitvec` in the `/mined_nakamoto_block` event.

- Closes #4257 

This property is formatted as a hex string, using the [`consensus_serialize` implementation of BitVec](https://github.com/stacks-network/stacks-core/blob/25fba11b99006cd2ef792f909148f2bcb41d68b0/stacks-common/src/bitvec.rs#L55). The first bytes are the length of the bitvec, encoded as a `u16`. Then the rest of the string is the bitvec data.

### DIFF
--- a/testnet/stacks-node/src/event_dispatcher.rs
+++ b/testnet/stacks-node/src/event_dispatcher.rs
@@ -102,6 +102,7 @@ pub struct MinedNakamotoBlockEvent {
     pub block_size: u64,
     pub cost: ExecutionCost,
     pub tx_events: Vec<TransactionEvent>,
+    pub signer_bitvec: String,
 }
 
 impl EventObserver {
@@ -1023,6 +1024,8 @@ impl EventDispatcher {
             return;
         }
 
+        let signer_bitvec = bytes_to_hex(block.header.signer_bitvec.serialize_to_vec().as_slice());
+
         let payload = serde_json::to_value(MinedNakamotoBlockEvent {
             target_burn_height,
             block_hash: block.header.block_hash().to_string(),
@@ -1031,6 +1034,7 @@ impl EventDispatcher {
             block_size: block_size_bytes,
             cost: consumed.clone(),
             tx_events,
+            signer_bitvec,
         })
         .unwrap();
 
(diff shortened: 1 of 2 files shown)


zone117x (approved):
TIL about the `/mined_nakamoto_block`. Could you please add this to the `/new_block` event? This one isn't consumed anywhere AFAIK (and first time I've seen it).

zone117x (changes requested):
TIL about the `/mined_nakamoto_block`. This one isn't consumed anywhere AFAIK (and first time I've seen it). Could you please add this to the `/new_block` event?

obycode (dismissed):
lgtm!

kantai (approved):

hstove:
@zone117x I've added the bitvec string to `/new_block`, mind taking another look?

jcnelson (approved):

testnet/stacks-node/src/event_dispatcher.rs
```diff
@@ -433,6 +436,13 @@ impl EventObserver {
             "pox_v3_unlock_height": pox_constants.v3_unlock_height,
         });
 
+        if let Some(signer_bitvec) = signer_bitvec_opt {
+            payload.as_object_mut().unwrap().insert(
+                "signer_bitvec".to_string(),
+                serde_json::to_value(signer_bitvec).unwrap_or_default(),
```
zone117x, line 442: I can't tell what the serialized JSON is for this. Is it a hex string?

hstove: Yeah, check out the PR description (last sentence). It's a length-prefixed hex string

zone117x: Oh okay, thanks!

zone117x (approved):

PR #4527, opened 2024-03-12
Fixes discovered while bringing up the pre-launch testnet

obycode:
### Description

This includes fixes to all of the issues discovered during the process of bringing up and stabilizing the pre-launch testnet, as well as various improvements to logging that we found to be helpful in finding and understanding problems.

### Applicable issues

- fixes #

### Checklist

- [ ] Changelog is updated


stacks-signer/src/signer.rs
```diff
@@ -967,22 +1029,27 @@ impl Signer {
...
-        // TODO: proper garbage collection...This is currently our only cleanup of blocks
-        self.signer_db
-            .remove_block(&block_vote.signer_signature_hash)
-            .expect(&format!("{self}: Failed to remove block from to signer DB"));
+        // WIP: try not deleting a block from signerDB until we have a better garbage collection strategy.
+        // This causes issues when we have to reprocess a block and we have already deleted it from the signerDB
+        // // TODO: proper garbage collection...This is currently our only cleanup of blocks
+        // self.signer_db
+        //     .remove_block(&block_vote.signer_signature_hash)
+        //     .expect(&format!("{self}: Failed to remove block from to signer DB"));
```
obycode, line 1037: Let's go ahead and delete this now.

zone117x (dismissed):
Great PR folks, maybe the greatest

libsigner/src/events.rs
```diff
@@ -50,11 +50,22 @@ use wsts::state_machine::signer;
...
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+/// BlockProposal sent to signers
+pub struct BlockProposalSigners {
+    /// The block itself
+    pub block: NakamotoBlock,
+    /// The burn height the block is mined during
+    pub burn_height: u64,
+    /// The reward cycle the block is mined during
+    pub reward_cycle: u64,
```
jcnelson, line 61: I don't think this field is necessary, since it's a derived value from the `burn_height`.  I don't think it should be possible to ever construct one of these structs with inconsistent `burn_height` and `reward_cycle` values.

stacks-signer/src/coordinator.rs
```diff
@@ -69,6 +69,9 @@ impl From<PublicKeys> for CoordinatorSelector {
     }
 }
 
+/// Whether or not to rotate to new coordinators in `update_coordinator`
+const ROTATE_COORDINATORS: bool = false;
```
jcnelson, line 73: Shouldn't this be a configurable value?

stacks-signer/src/runloop.rs
```diff
@@ -266,7 +266,11 @@ impl RunLoop {
                     .insert(reward_index, Signer::from(new_signer_config));
                 debug!("Reward cycle #{reward_cycle} Signer #{signer_id} initialized.");
             } else {
-                warn!("Signer is not registered for reward cycle {reward_cycle}. Waiting for confirmed registration...");
+                if current {
```
jcnelson, line 269: So is `current` only controlling the contents of the warning message here?  If so, then can we omit `current` as an argument, and instead compute whether or not `reward_cycle` is the last reward cycle based on our last-seen burnchain block?

stacks-signer/src/signer.rs
```diff
@@ -497,20 +513,59 @@ impl Signer {
...
+                    continue;
+                }
+                Ok(None) => {
+                    // Store the block in our cache
+                    self.signer_db
+                        .insert_block(self.reward_cycle, &BlockInfo::new(proposal.block.clone()))
+                        .unwrap_or_else(|e| {
+                            error!("{self}: Failed to insert block in DB: {e:?}");
+                        });
+                    // Submit the block for validation
```
jcnelson, line 555: Wait, if we fail to store the block, then why continue to try and make progress on it?

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -396,6 +409,10 @@ impl BlockMinerThread {
         let now = Instant::now();
         debug!("Miner: waiting for block response from reward cycle {reward_cycle } signers...");
         while now.elapsed() < self.config.miner.wait_on_signers {
+            if self.check_burn_tip_changed(&sortdb).is_err() {
```
jcnelson, line 412: This check isn't sufficient.  While waiting for a block, the Bitcoin chain can "flap" between tips A, B, and back to A again.  Then, this check could pass (i.e. return `Ok(..)`) if the flap happens during one pass of the loop, but the signers could notice the flap and produce two conflicting Stacks histories (one for each Bitcoin fork). However, it could be the case that tips A and B have no sortition, but they share a common ancestor that does (so, the conflicting histories' tenure-change transactions would be consistent, but their in-flight tenures would diverge).  To address this, I think this code should additionally check that the Stacks tip hasn't changed, except to add the miner's in-flight block.

stacks-signer/src/runloop.rs
obycode: We are not currently tracking our last-seen burnchain block, so that would require a call to our stacks-node (which the caller of this function has already done).

libsigner/src/events.rs
obycode: The signer does not currently have the information needed to map burn height to reward cycle, so it would otherwise require a call to the stacks-node to compute.

testnet/stacks-node/src/nakamoto_node/miner.rs
kantai: I agree, but I think the miner should just continue trying to get its block accepted up to the configured timeout with signers. If the signer set has moved onto another miner, it will be realized in a timeout, and I don't think that the logic needs to be more complicated than that.

stacks-signer/src/signer.rs
kantai: In theory, the database failure may just be intermittent, and the signer will just try to store the block again when it gets the validation response back from the node.

stacks-signer/src/signer.rs
obycode: deleted in 49eb61a705e7f64be204ecdf15fcca3bd623dabc

stacks-signer/src/signerdb.rs
```diff
@@ -85,14 +93,30 @@ impl SignerDb {
...
+        let signed_over = &block_info.signed_over;
+        debug!(
+            "Inserting block_info: reward_cycle = {reward_cycle}, sighash = {hash}, block_id = {block_id}, signed = {signed_over} vote = {:?}",
+            block_info.vote.as_ref().map(|v| {
+                if v.rejected {
+                    "REJECT"
+                } else {
+                    "ACCEPT"
+                }
+            })
```
netrome, line 114: Nit: I think it would be cleaner to have this logic as a `Display` implementation on the `BlockInfo` type. Any reason not to do that?

stacks-signer/src/signer.rs
```diff
@@ -497,20 +513,59 @@ impl Signer {
...
+                        "signer_sighash" => %sig_hash,
+                        "proposal_burn_height" => proposal.burn_height,
+                        "vote" => ?block.vote.as_ref().map(|v| {
+                            if v.rejected {
+                                "REJECT"
+                            } else {
+                                "ACCEPT"
+                            }
+                        }),
+                        "signed_over" => block.signed_over,
```
netrome, line 544: Same as my previous nit: If we implemented `Display` for the BlockInfo type we could reuse the implementation here and avoid some duplication.

netrome (commented):
Just a few minor nits from my side but nothing major. I'll approve once Jude's comments are addressed.

jferrant (dismissed):
Really nice to see all these changes!

zone117x (approved):

stacks-signer/src/runloop.rs
jcnelson: Okay, can you add a TODO item to remove this once the signer starts tracking the last-seen burnchain block?  I think that's something you're going to want to be doing anyway?

libsigner/src/events.rs
jcnelson: This is fine for now, but PoX constants are well-known and unchanging so it's safe to rely on them in the signer (no RPC required).

jcnelson (approved):

netrome (approved):

stacks-signer/src/runloop.rs
kantai: Added TODO in 6e4d9cce7

zone117x (approved):

jferrant (approved):

stacks-signer/src/signerdb.rs
obycode: This is a good idea. I'll add this in a separate PR.

PR #4528, opened 2024-03-13
test: upgrade core-contract-tests dependencies

hugocaillard:
Upgrade npm dependencies of core-contract-tests
  - vite and vitests to v1, no big impact impacts vs the previous version. Require slight changes in vitest.config file
  - clarinet-sdk upgraded to the latest beta, it now uses deployment-plans, hence the new file

### DIFF
--- a/contrib/core-contract-tests/deployments/default.simnet-plan.yaml
+++ b/contrib/core-contract-tests/deployments/default.simnet-plan.yaml
@@ -1,0 +1,87 @@
+---
+id: 0
+name: "Simulated deployment, used as a default for `clarinet console`, `clarinet test` and `clarinet check`"
+network: simnet
+genesis:
+  wallets:
+    - name: deployer
+      address: ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM
+      balance: "100000000000000"
+    - name: wallet_1
+      address: ST1SJ3DTE5DN7X54YDH5D64R3BCB6A2AG2ZQ8YPD5
+      balance: "100000000000000"
+    - name: wallet_2
+      address: ST2CY5V39NHDPWSXMW9QDT3HC3GD6Q6XX4CFRK9AG
+      balance: "100000000000000"
+    - name: wallet_3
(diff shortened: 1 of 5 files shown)


contrib/core-contract-tests/deployments/default.simnet-plan.yaml
```diff
@@ -0,0 +1,87 @@
...
+            contract-name: signers
+            emulated-sender: ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM
+            path: "../../stackslib/src/chainstate/stacks/boot/signers.clar"
+            clarity-version: 2
+        - emulated-contract-publish:
+            contract-name: signers-voting
+            emulated-sender: ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM
+            path: "../../stackslib/src/chainstate/stacks/boot/signers-voting.clar"
+            clarity-version: 2
+      epoch: "2.4"
```
moodmosaic, line 87: I guess we need the this and switch to `3.0` on demand in the tests, right?

moodmosaic (dismissed):
LGTM 👍

contrib/core-contract-tests/deployments/default.simnet-plan.yaml
hugocaillard: @moodmosaic There a multiple things to consider here;
1. the deployment plan file is just saved with the recent version of the SDK, but in terms of contract deployment, it reflects the clarinet.toml, if the epoch has to be changed, it should probably happen in an other PR to keep this one just a dependency update PR
2. There are a few other things to changes in the clarinet.toml, especially regarding epochs (ie: bns wasn't deployed in 2.4)
3. if the purpose really is to test boot contracts as such, maybe the tests should call the boot contracts deployed by clarinet (as in this example) https://github.com/hugocaillard/pox-testing-sample. Especially for the pox contracts, since calling a custom pox contract won't actually lock the stx

moodmosaic (approved):

friedger (approved):
LGTM.

obycode (approved):

PR #4530, opened 2024-03-13
Chore: Remove `Box`ing of `SymbolicExpression`s

### LINKED ISSUE #4445: [Chore] Remove `Box`ing of `SymbolicExpression`s in Clarity
`SymbolicExpression`s are unnecessarily `Box`'d in a few places in the `clarity` crate when used in `Vec`s. As `Vec`s are already on the heap, a `Box` is not necessary.

ASuciuX:
### Description
`SymbolicExpression`s and `PreSymbolicExpression`s are unnecessarily `Box`ed in a few places in the `clarity` crate when used in `Vec`s. As `Vec`s are already on the heap, a `Box` is not necessary.

Closes #4445

### DIFF
--- a/clarity/src/vm/ast/parser/v1.rs
+++ b/clarity/src/vm/ast/parser/v1.rs
@@ -525,7 +525,7 @@ pub fn parse_lexed(input: Vec<(LexItem, u32, u32)>) -> ParseResult<Vec<PreSymbol
                 if let Some((list, start_line, start_column, parse_context)) = parse_stack.pop() {
                     match parse_context {
                         ParseContext::CollectList => {
-                            let checked_list: ParseResult<Box<[PreSymbolicExpression]>> = list
+                            let checked_list: ParseResult<Vec<PreSymbolicExpression>> = list
                                 .into_iter()
                                 .map(|i| match i {
                                     ParseStackItem::Expression(e) => Ok(e),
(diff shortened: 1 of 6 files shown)


cylewitruk (approved):
Lgtm :) This suggestion stemmed from playing with different binary serialization formats, which aren't as forgiving as JSON.

jcnelson:
@kantai Is there a reason we were `Box`-ing these before?

kantai:
Classic case of early optimization, I think. These structs never need to be resized, so boxed slices were maybe a closer match to their usage than vecs. No real reason that they need to be boxed slices instead of vecs, though, and I doubt that there's a performance benefit to keeping them as boxed slices (I wouldn't be surprised if it was the opposite).

jcnelson (approved):
If it's fine with @kantai it's fine with me :+1:

kantai (approved):

jbencin:
> I doubt that there's a performance benefit to keeping them as boxed slices (I wouldn't be surprised if it was the opposite).

They compile to pretty much the same thing. A `Box`ed slice is represented on the stack with 2 elements: `ptr` and `len`. `Vec` has an additional element: `capacity`. So the cost of using a `Vec` is just an extra `usize`, which is trivial

cylewitruk:
Boxed slices also had issues with several binary encodings I tried for these types, where Vecs were handled otob -- that's essentially the birth place of the issue, preparing for that kind of change.

Issue #4533, opened 2024-03-13
[BUG] VoteForAggregateKeyOp incorrectly validates aggregate public key


jcnelson:
The `VoteForAggregateKeyOp::check()` method needs to use `wsts`'s `Point` struct, since that's the container struct that will be used for the key.  Otherwise, if there's ever any discrepancy between `Secp256k1PublicKey` and `Point`, this method will incorrectly reject a valid operation or accept an invalid operation.

PR #4536, opened 2024-03-13
fix: add wsts::Point validation to `vote-for-aggregate-key` burn op

hstove:
As noted in https://github.com/stacks-network/stacks-core/issues/4533, the `vote-for-aggregate-key`'s `check` method doesn't validate the aggregate key used in the payload against `wsts::Point`, which can lead to issues if there are any discrepancies between `Secp256k1PublicKey` and `Point`. This PR adds validation against `Point`, along with the existing `Secp256k1PublicKey` check.

- Closes https://github.com/stacks-network/stacks-core/issues/4533


jcnelson (approved):
Thanks @hstove for the quick turn-around on this!

xoloki (approved):
LGTM

PR #4539, opened 2024-03-14
Fix/remove needless asserts

jcnelson:
This synchronizes `develop` to `master` and removes a few needless `assert!` statements that could occasionally cause problems in stress-testing.


obycode (approved):
👍

kantai (approved):

PR #4540, opened 2024-03-14
Clarity-wasm: fix computation of in-memory size

Acaccia:
Here is another fix for `get_type_in_memory_size`.

At first, it did not account for the actual value of an in-memory type, causing issues when returning complex types from linked functions.
Then, it accounted for it all the time, requiring too much space when sometimes not needed and breaking the function `append`.

Here is a fix that takes the best of both worlds.

This passes 500 iterations of each proptests in clarity-wasm, where the tests are.


obycode (approved):
Thanks for cleaning up all of my messes! 🙏 🙌

PR #4542, opened 2024-03-14
fix: remove `--reward-cycle` from `stacks-signer run`

hstove:
Right now, when running `stacks-signer run ...`, you need to include a `--reward-cycle` argument. This is actually now longer used, but it's still required, so all of our docs/infra just use `--reward-cycle 1`.

This PR separates the args for `run` and `dkg`, so that you no longer need to pass `--reward-cycle` to `run`.

I've also added the `short` macro to all `--config` args, so you can optionally use `-c`.

NB: this will break existing infra as well as some of our docs, so we need to be mindful of updating them once this is merged.

### DIFF
--- a/stacks-signer/src/cli.rs
+++ b/stacks-signer/src/cli.rs
@@ -55,7 +55,7 @@ pub enum Command {
     /// Run a DKG round through the stacker-db instance
     Dkg(RunDkgArgs),
     /// Run the signer, waiting for events from the stacker-db instance
-    Run(RunDkgArgs),
+    Run(RunSignerArgs),
     /// Generate necessary files for running a collection of signers
     GenerateFiles(GenerateFilesArgs),
     /// Generate a signature for Stacking transactions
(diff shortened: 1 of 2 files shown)


obycode (approved):
lgtm

stacks-signer/src/cli.rs
```diff
@@ -138,16 +138,24 @@ pub struct SignArgs {
...
+/// Arguments for the Run command
+pub struct RunSignerArgs {
+    /// Path to config file
+    #[arg(long, short, value_name = "FILE")]
+    pub config: PathBuf,
+}
+
 #[derive(Parser, Debug, Clone)]
 /// Arguments for the generate-files command
 pub struct GenerateFilesArgs {
```
jferrant, line 161: Actually do you think you could clean up the GenerateFilesArgs generate files command? It should not need the reward cycle anymore. Actually might not need some other stuff either (we no longer generate stackerdb. so mind quickly going over as part of this PR? )

hstove: yep np!

jferrant: LOL I legit missed the little arrows below this and thought I saw "reward_cycle" as part of generate files args...Ignore me.

jferrant (approved):
LGTM

Issue #4504, opened 2024-03-07
[Nakamoto] stacks-signers should listen to burn block events to determine if in prepare phase


jferrant:
Currently stacks-signers poll the stacks node on every pass of the run loop to determine if they are registered for the current or next reward cycles. This ends up pinging the node rather aggressively and also results in excessive logging about not being registered for the current or upcoming reward cycle. 

Update libsigner to subscribe to new burn block events and pass these through to the signer. The signers hould use these burn block events to determine if it is in the prepare phase or not. Note that on first spin up, it will have to query the node, but on subsequent passes of the run one pass, it should not have to.

PR #4549, opened 2024-03-16
Feat/signers subscribe to burn blocks

jferrant:
- Subscribes to burn block events
- Store current reward cycle on the side. Only refresh reward cycle when a new burn block event arrives (prevents spamming the node needlessly)
- Only attempt to initialize/refresh the current reward cycle signer if initializing the runloop
- Only attempt to refresh a signer config for the NEXT reward cycle if in the prepare phase during runloop initialization OR have received a new burn block in the prepare phase and have still not been registered
- Removes unused ClientErrors
- This uncovered a bug in poststackerdbchunk error codes. Cleaned up as part of this PR

Closes: https://github.com/stacks-network/stacks-core/issues/4504

### DIFF
--- a/libsigner/src/events.rs
+++ b/libsigner/src/events.rs
@@ -73,6 +73,8 @@ pub enum SignerEvent {
     BlockValidationResponse(BlockValidateResponse),
     /// Status endpoint request
     StatusCheck,
+    /// A new burn block event was received
+    NewBurnBlock,
 }
 
 impl StacksMessageCodec for BlockProposalSigners {
@@ -281,6 +283,8 @@ impl EventReceiver for SignerEventReceiver {
                 process_stackerdb_event(event_receiver.local_addr, request, is_mainnet)
             } else if request.url() == "/proposal_response" {
                 process_proposal_response(request)
+            } else if request.url() == "/new_burn_block" {
+                process_new_burn_block_event(request)
             } else {
                 let url = request.url().to_string();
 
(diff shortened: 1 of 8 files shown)


libsigner/src/events.rs
```diff
@@ -73,6 +74,18 @@ pub enum SignerEvent {
     BlockValidationResponse(BlockValidateResponse),
     /// Status endpoint request
     StatusCheck,
+    /// A new burn block event was received with the given burnchain block height
+    NewBurnBlock(u64),
+}
+
+/// A struct to aid in deserializing the new burn block event
+#[derive(Debug, Deserialize)]
+struct TempBurnBlockEvent {
```
jcnelson, line 83: If this is only used in `process_new_burn_block_event()`, you could potentially move it to within the function body so it doesn't look like this struct is needed in the module itself.

jcnelson (approved):
LGTM; just minor nits

libsigner/src/events.rs
jferrant: Will do!

jcnelson (approved):

stacks-signer/src/signer.rs
```diff
@@ -1123,6 +1139,7 @@ impl Signer {
     /// Update the DKG for the provided signer info, triggering it if required
     pub fn update_dkg(&mut self, stacks_client: &StacksClient) -> Result<(), ClientError> {
         let reward_cycle = self.reward_cycle;
+        let old_dkg = self.approved_aggregate_public_key;
```
xoloki, line 1142: Is it possible for `stacks_client` to return `None` because of transient errors?  If so should we check before overwriting the `self` key?  We only seem to be updating the `self.coordinator` key if `stacks_client` returns `Some`

xoloki (approved):
LGTM just a tiny question about caching the approved agg key at various layers

PR #4550, opened 2024-03-16
[pox-4] Pool Stacking stateful property-based tests

moodmosaic:
### Description

This is an integral part of our commitment to adopting a *stateful* property-based testing strategy for PoX-4.

### Applicable issues

- https://github.com/stacks-network/stacks-core/issues/4548

### Additional info (benefits, drawbacks, caveats)

The stateful property-based tests can run via `npm test` and also directly via 

```
npx vitest watch tests/pox-4/pox-4.stateful-prop.test.ts --reporter=dot
```

### Sample output

```
✓ wallet_4        allow-contract-caller  wallet_3               until                  179942593
✓ wallet_8        delegate-stx           amount                 33784629952454         delegated to           wallet_2               until                  1875298417
✓ wallet_4        allow-contract-caller  wallet_4               until                  none
✓ wallet_9        delegate-stx           amount                 36970141191493         delegated to           wallet_9               until                  1405718799
✓ wallet_4        allow-contract-caller  wallet_7               until                  1739629828
✓ wallet_5        stx-account            lock-amount            625000000000           unlocked-amount        99375000000000         unlocked-height        11550
✓ wallet_9        revoke-delegate-stx
✓ wallet_5        get-stacking-minimum   pox-4                  125000000000
✓ wallet_9        allow-contract-caller  wallet_4               until                  858124873
✓ wallet_2        stx-account            lock-amount            1525922319735          unlocked-amount        98474077680265         unlocked-height        12600
✓ wallet_7        allow-contract-caller  wallet_1               until                  1307633494
✓ wallet_9        stx-account            lock-amount            625000000000           unlocked-amount        99375000000000         unlocked-height        13650
✓ wallet_7        revoke-delegate-stx
✓ wallet_9        delegate-stx           amount                 11433545341619         delegated to           wallet_2               until                  1653742613
✓ wallet_2        allow-contract-caller  wallet_8               until                  1376025588
✓ wallet_5        get-stacking-minimum   pox-4                  125000000000
✓ wallet_9        get-stacking-minimum   pox-4                  125000000000
✓ wallet_2        stx-account            lock-amount            1525922319735          unlocked-amount        98474077680265         unlocked-height        12600
✓ wallet_1        allow-contract-caller  wallet_7               until                  89399923
✓ wallet_6        delegate-stx           amount                 34970137606759         delegated to           wallet_2               until                  288872382
✓ wallet_9        get-stacking-minimum   pox-4                  125000000000
```

Issue #4548, opened 2024-03-15
[pox-4] Stacking Stateful Tests


setzeus:
With the latest progress from @BowTiedRadone & @moodmosaic, it's now possible to test the core stacking functions in PoX-4 in a stateful manner. Starting with the core sequence for a solo stacker, a "complete" instance of stateful testing would cover *all* of the following state changes in a single run, across multiple reward cycles:

**Generating Authorization**
- [x] Signer explicitly creates an authorization

**Solo Stacking**
- [x] Solo stacking with a signature (stacks-stx)
- [x] Solo stacking with an authorization (stacks-stx)
- [x] Solo stacking extension with a signature (stack-extend)
- [x] Solo stacking extension with an authorization (stack-extend)
- [x] Solo stacking increase (stack-increase)

**Pool Stacking**
- [x] Pool delegating (delegate-stx)
- [x] Pool revoking (revoke-delegate-stx)
- [x] Pool stacking (delegate-stack-stx)
- [x] Pool stacking extended (delegate-stack-extend)
- [x] Pool stacking increased (delegate-stack-increase)
- [x] Allow contract caller (allow-contract-caller)
- [x] Disallow contract caller (disallow-contract-caller)
- [x] Pool aggregate commit with a signature (stack-aggregation-commit)
- [x] Pool aggregate commit with an authorization (stack-aggregation-commit)
- [x] Pool aggregate commit indexed with a signature (stack-aggregation-commit)
- [x] Pool aggregate commit indexed with an authorization (stack-aggregation-commit)
- [x] Pool aggregate increase (stack-aggregation-increase)

See also https://github.com/stacks-network/stacks-core/discussions/4433.

moodmosaic:
Resolved via #4550 and #4725.

PR #4552, opened 2024-03-18
Add delegate-stx command

BowTiedRadone:
This PR adds the `delegate-stx` command to the stateful property testing environment.


moodmosaic:
This PR is part of #4548 and can be merged as-is since it targets `feat/pox-4-stateful-property-testing` (https://github.com/stacks-network/stacks-core/pull/4550).

Issue #4479, opened 2024-03-04
Epoch 2.5 should disable microblock production


kantai:
Epoch 2.5 should disable microblock production. This ensures that reward sets during the Epoch 2.5 -> 3.0 transition are identical when calculated with the stackerdb Clarity updates and the Epoch 2.x reward set calculations.

kantai:
Closed with #4561

PR #4561, opened 2024-03-19
Disable Microblocks in Epoch 2.5

kantai:
### Description

This PR disables microblock production and confirmation in Epoch 2.5 and will reject any blocks that attempt to confirm microblocks in Epoch 2.5 and beyond.

This is tested with a new integration test that first mines some blocks in Epoch 2.5 with a microblock only tx in the mempool, asserting that the transaction does not get confirmed and no new microblocks are produced. It then sets a test only flag which tells the miner to mine microblocks in 2.5 anyways, and asserts that the microblocks are produced but never confirmed. Finally, it sets another flag which tells the miner to confirm microblocks in 2.5 anyways, and asserts that the miner's blocks are no longer accepted.

### Applicable issues

#4479


testnet/stacks-node/src/tests/epoch_25.rs
```diff
@@ -0,0 +1,296 @@
...
+    // second block will be the first mined Stacks block
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    // push us to block 205
+    next_block_and_wait(&mut btc_regtest_controller, &blocks_processed);
+
+    let tx = make_stacks_transfer_mblock_only(&spender_1_sk, 0, 500, &spender_2_addr, 500);
+    submit_tx(&http_origin, &tx);
+
+    // wait until just before epoch 2.1
```
jcnelson, line 171: copypasta comment?

jcnelson (dismissed):
The code overall LGTM, so I'll approve for now.  However, I think we'll also want to update the mempool admission logic to reject microblock-only transactions at epoch 2.5.  That can be a separate PR if you want

kantai:
> The code overall LGTM, so I'll approve for now. However, I think we'll also want to update the mempool admission logic to reject microblock-only transactions at epoch 2.5. That can be a separate PR if you want

Yep -- we would probably also want to update the microblock downloading logic to reject microblocks after Epoch 2.5, but like the mempool changes, since those are not consensus critical, I'd prefer to handle them in a follow on PR.

obycode (approved):
LGTM. My only thought was about the mempool rejection for transactions after 2.5 with the wrong anchor mode, but I see that was already discussed.

jcnelson (approved):

PR #4562, opened 2024-03-19
Add delegate-stack-stx command

BowTiedRadone:
This PR adds the `delegate-stack-stx` command to the stateful property testing environment.


moodmosaic:
This PR can be closed. We replaced it with https://github.com/stacks-network/stacks-core/pull/4564/.

PR #4563, opened 2024-03-20
fix: handle burn chain flapping

obycode:
This test (from Jude) can reproduce the problematic behavior when the burnchain flaps between two branches.

- We get blocks at height 211 - 213
- Then we get a fork, with different blocks at 211-213, as well as 214 and 215.
- We then flap back to the original fork, so it goes back to the common ancestor, 210, and tries to download 211-215
- When it gets block 211, it is already in the database, so it fails, cancelling the download of the rest of the blocks, leaving 214 and 215 in this branch not stored
- Then we try to store 216, but its parent 215 is not stored yet, so we cannot continue.

The fix for this is to ignore attempts to store duplicate blocks.

### Applicable issues

- fixes #

### Checklist

- [ ] Changelog is updated

### DIFF
--- a/stackslib/src/burnchains/bitcoin/indexer.rs
+++ b/stackslib/src/burnchains/bitcoin/indexer.rs
@@ -837,9 +837,10 @@ impl BitcoinIndexer {
                 }
             } else {
                 // ignore the reorg
-                debug!(
+                test_debug!(
                     "Reorg chain does not overtake original Bitcoin chain ({} >= {})",
-                    orig_total_work, reorg_total_work
+                    orig_total_work,
+                    reorg_total_work
                 );
                 new_tip = orig_spv_client.get_headers_height()?;
             }
--- a/stackslib/src/burnchains/bitcoin/indexer.rs
+++ b/stackslib/src/burnchains/bitcoin/indexer.rs
@@ -837,7 +837,11 @@ impl BitcoinIndexer {
                 }
             } else {
                 // ignore the reorg
-                test_debug!("Reorg chain does not overtake original Bitcoin chain");
+                test_debug!(
+                    "Reorg chain does not overtake original Bitcoin chain ({} >= {})",
+                    orig_total_work,
+                    reorg_total_work
+                );
                 new_tip = orig_spv_client.get_headers_height()?;
             }
         }
--- a/stackslib/src/burnchains/bitcoin/indexer.rs
+++ b/stackslib/src/burnchains/bitcoin/indexer.rs
@@ -837,10 +837,9 @@ impl BitcoinIndexer {
                 }
             } else {
                 // ignore the reorg
-                test_debug!(
+                debug!(
                     "Reorg chain does not overtake original Bitcoin chain ({} >= {})",
-                    orig_total_work,
-                    reorg_total_work
+                    orig_total_work, reorg_total_work
                 );
                 new_tip = orig_spv_client.get_headers_height()?;
             }
--- a/stackslib/src/burnchains/db.rs
+++ b/stackslib/src/burnchains/db.rs
@@ -312,7 +312,7 @@ impl<'a> BurnchainDBTransaction<'a> {
     fn store_burnchain_db_entry(
         &self,
         header: &BurnchainBlockHeader,
-    ) -> Result<i64, BurnchainError> {
+    ) -> Result<(), BurnchainError> {
         let sql = "INSERT OR IGNORE INTO burnchain_db_block_headers
                    (block_height, block_hash, parent_block_hash, num_txs, timestamp)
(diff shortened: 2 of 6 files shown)


obycode:
@jcnelson I couldn't come up with a good reason why we couldn't just ignore the duplicate entry. Let me know if you see any potential problems from this.

stackslib/src/burnchains/db.rs
```diff
@@ -323,9 +323,17 @@ impl<'a> BurnchainDBTransaction<'a> {
...
+        if affected_rows == 0 {
+            // This means a duplicate entry was found and the insert operation was ignored
+            debug!(
+                "Duplicate entry for block_hash: {}, insert operation ignored.",
+                header.block_hash
+            );
+            Ok(-1)
+        } else {
+            // A new row was inserted successfully
+            Ok(self.sql_tx.last_insert_rowid())
```
jcnelson, line 336: I think you can actually remove the returned rowid, since I don't think it's used anywhere.  This can just return `Ok(())`.

testnet/stacks-node/src/tests/neon_integrations.rs
```diff
@@ -11174,3 +11178,133 @@ fn filter_txs_by_origin() {
...
+        .map_err(|_e| ())
+        .expect("Failed starting bitcoind");
+
+    // carry out the flap back to fork A
+    for _i in 0..7 {
+        btc_regtest_controller.build_next_block(1);
+        thread::sleep(Duration::from_secs(5));
+    }
+
+    assert_eq!(channel.get_sortitions_processed(), 225);
```
jcnelson, line 11307: Just want to confirm that this `assert_eq!` will fail if the fix to `stackslib/src/burnchains/db.rs` is _not_ applied?

jcnelson:
Can you actually base this on `develop`? We'd want to test this prior to shipping it in a release. We'll also want this merged into `next` ASAP because we'll definitely hit it in 2.5 testnet.

testnet/stacks-node/src/tests/neon_integrations.rs
obycode: Yes, this fails without the fix.

stackslib/src/burnchains/db.rs
obycode: Agreed, I thought about changing the return type too. Will make this change.

obycode: Do you think the log about the duplicate entry is useful to keep, or should we just simplify it to:
```rust
self.sql_tx.execute(sql, args).map(|_| Ok(()))?
```

obycode:
Heads up: I will reapply these changes on `develop` and force push in a few minutes.

stackslib/src/burnchains/bitcoin/indexer.rs
```diff
@@ -837,7 +837,11 @@ impl BitcoinIndexer {
                 }
             } else {
                 // ignore the reorg
-                test_debug!("Reorg chain does not overtake original Bitcoin chain");
+                test_debug!(
+                    "Reorg chain does not overtake original Bitcoin chain ({} >= {})",
+                    orig_total_work,
+                    reorg_total_work
+                );
```
kantai, line 844: Might be useful to upgrade this to `debug!`

stackslib/src/burnchains/db.rs
kantai: I think keeping the log is a good idea.

kantai (dismissed):
LGTM -- just a few comments

stackslib/src/burnchains/bitcoin/indexer.rs
jcnelson: This is actually pretty chatty.  It'll get printed each time the node tries to sync with Bitcoin, because it looks for reorgs each time.

jcnelson (dismissed):

stackslib/src/burnchains/bitcoin/indexer.rs
obycode: ok, reverting.

jcnelson (approved):
LGTM. Just make sure CI passes :pray:

obycode:
Hmm, the new tests failed with `Failed starting bitcoind: ()`. Is there something else that needs to be added for the CI setup? It works locally.

kantai:
Not that I know of, but we might get a better error from CI with this map_err removed (since you added Debug to the error anyways):

```rust
    btcd_controller
        .start_bitcoind()
        .map_err(|_e| ())
```

jcnelson (approved):
Approving for now, but we need to get that CI test working

kantai (approved):

PR #4570, opened 2024-03-20
feat: Nakamoto signer persisting its party shares to enable restart

### LINKED ISSUE #4388: [Nakamoto] stacks-signer should store its party shares on the side to enable restart
Currently if a signer crashes, it is not enough to simply retrieve the DKG key from the contract
Use the wsts save/load functionality to persist state needed to continue signing/dkg between crashes.

Depends on https://github.com/stacks-network/wsts/issues/2

netrome:
### Description
This PR extends `SignerDB` to support persisting private polynomials and keys for signers as a signer state (`wsts::traits::SignerState`). This is the necessary state for a signer to be able to participate in signing rounds after a restart.

### Applicable issues

- fixes #4388


jferrant (commented):
Looking good :)

stacks-signer/src/signer.rs
```diff
@@ -46,6 +46,7 @@ use wsts::state_machine::coordinator::{
 };
 use wsts::state_machine::signer::Signer as WSTSSigner;
```
xoloki, line 47: Can you rename this `SignerStateMachine` or something else descriptive?  `WSTSSigner` is really ambiguous.

stacks-signer/src/signer.rs
```diff
@@ -263,6 +255,28 @@ impl From<SignerConfig> for Signer {
         );
         let signer_db =
             SignerDb::new(&signer_config.db_path).expect("Failed to connect to signer Db");
+
+        let mut signing_round = WSTSSigner::new(
```
xoloki, line 259: Likewise, can you think of a better name than `signing_round` for this?  Maybe just `state_machine`?  We don't make a new state machine for each signing round.  And this name token is being reused elsewhere in the PR so it's not just local.

xoloki (dismissed):
Looks good so far, just a couple of naming nits.

stacks-signer/src/signer.rs
netrome: Yeah I found that name confusing as well. Will change it with pleasure. Thanks for the suggestion :pray:

netrome: Updated :heavy_check_mark:

stacks-signer/src/signer.rs
netrome: Done. Also renamed the method `save_signing_round` to `save_signer_state`. :heavy_check_mark:

xoloki:
@netrome we got a preliminary audit report today, and one of the things they mentioned was that we weren't deleting the private polynomials after DKG.  I'm gonna make a small `wsts` change to do this for `9.0.0` before releasing the crate; I'll make sure that this PR still works, but please check and review as well.

xoloki:
I updated your `stacks-wsts` github rev to the latest PR with the clearing changes.

Cargo.toml
```diff
@@ -20,8 +20,7 @@ rand_core = "0.6"
 rand = "0.8"
 rand_chacha = "0.3.1"
 tikv-jemallocator = "0.5.4"
-# wsts = { version = "8.1", default-features = false }
-wsts = { git = "https://github.com/stacks-network/wsts.git", branch = "feat/public-sign-ids" }
+wsts = { git = "https://github.com/stacks-network/wsts.git", default-features = false, rev = "cf395701cb1332e4d66e70d8baf60b1658c168b5" }
```
netrome, line 23: This will be updated once https://github.com/stacks-network/wsts/pull/7 is merged and a new `wsts` version is released.

xoloki: Commit 1b9589df9c1ee299423bffdf8f74298d9f4464f2 is the latest on the PR branch if you want to test the Party API changes

xoloki (dismissed):
LGTM

Cargo.toml
netrome: Thanks, I've updated this now. It seemed to work well running the integration test locally.

stacks-signer/src/config.rs
```diff
@@ -338,19 +338,23 @@ pub fn build_signer_config_tomls(
...
-        let endpoint = format!("localhost:{}", port);
-        port += 1;
+    for stacks_private_key in stacks_private_keys {
+        let endpoint = format!("localhost:{}", port_start);
+        port_start += 1;
+
+        let stacks_public_key = StacksPublicKey::from_private(stacks_private_key).to_hex();
+        let db_dir = std::env::temp_dir().join(format!(
+            "stacks-node-tests/integrations-signers/signer_{stacks_public_key}"
+        ));
```
kantai, line 352: I'd prefer specifically using `/tmp/stacks-node/tests` rather than `temp_dir()` -- its what the rest of the integration tests do, and it makes it much easier to have them always in the same place as the rest of the test directories.

This would also mean that the run stamp should be added back in.

stacks-signer/src/signerdb.rs
```diff
@@ -70,6 +81,41 @@ impl SignerDb {
         )
     }
 
+    /// Get the signer state for the provided reward cycle if it exists in the database
+    pub fn get_signer_state(&self, reward_cycle: u64) -> Result<Option<SignerState>, DBError> {
+        let result: Option<String> = query_row(
+            &self.db,
+            "SELECT state FROM signer_states WHERE reward_cycle = ?",
+            &[reward_cycle.to_string()],
```
kantai, line 89: I think it's bad to `to_string()` a u64 to an integer. This leads to sqlite's conversion behavior, and I don't really know what it would do in the case of trying to pass `(i64::MAX as u64) + 1` here. We use `u64_to_sql()` for this.

stacks-signer/Cargo.toml
```diff
@@ -46,6 +46,8 @@ url = "2.1.0"
 
 [dev-dependencies]
 clarity = { path = "../clarity", features = ["testing"] }
+polynomial = "0.2.6"
+num-traits = "0.2.18"
```
kantai, line 50: Where do these get used?

Cargo.toml
```diff
@@ -20,8 +20,7 @@ rand_core = "0.6"
 rand = "0.8"
 rand_chacha = "0.3.1"
 tikv-jemallocator = "0.5.4"
-# wsts = { version = "8.1", default-features = false }
-wsts = { git = "https://github.com/stacks-network/wsts.git", branch = "feat/public-sign-ids" }
+wsts = { git = "https://github.com/stacks-network/wsts.git", default-features = false, rev = "1b9589df9c1ee299423bffdf8f74298d9f4464f2" }
```
kantai, line 23: Why `default-features = false`?

kantai (commented):
This LGTM, just had a few comments.

stacks-signer/src/config.rs
netrome: Sure thing, I'd prefer to use `temp_dir()` instead of hard-coding `/tmp` but I see it's hard-coded in more places so that's a separate refactor.

Why should the run stamp be added back in? It was randomly generated within the function so it does not seem to correspond to any other directory. Also, I can't have this function be non-deterministic because I need a restarted signer to be able to read it's signerdb - which is why I'm using the public key to differentiate them here instead of a random run stamp.

stacks-signer/Cargo.toml
netrome: In the SignerDB tests (line 212 and 213 on the current commit).

Cargo.toml
xoloki: This is how we support musl. It simplifies the p256k1 build.

xoloki: No bindgen so no need for libclang etc

netrome: I have two answers here, depending on how to interpret the question:

1. **Why would we like to have `default-features=false` here?**
The wsts library uses bindings to the underlying `libsecp256k1` C library. It defaults to rebuilding these bindings, but Jude has added a set of precompiled bindings to avoid recompiling these. Setting `default-features=false` makes use of the precompiled `libsecp256k1` bindings instead of recompiling these.

2. **Why did I change this?**
I didn't. You did. When I began this change set, `default-features = false` was already set for the WSTS dependency - which you removed in `207cb690fe`. It did not seem intentional since that change had commented out the original dependency and pointed to a branch, so I chose to keep `default-features = false` while rebasing. If you prefer the other way around I can change this. I have no strong preference here.

stacks-signer/src/signerdb.rs
netrome: Done :heavy_check_mark:

stacks-signer/src/config.rs
netrome: Done :heavy_check_mark:

Cargo.toml
jcnelson: Can we avoid having a `rev` here?  Let's just publish a stable version of `wsts` with the features you need before merging this.

I don't care how you want to deal with `default-features = false`.  All I care about is that we can still build both glibc and musl Linux images.

stacks-signer/src/config.rs
```diff
@@ -338,19 +338,20 @@ pub fn build_signer_config_tomls(
...
-        port += 1;
+    for stacks_private_key in stacks_private_keys {
+        let endpoint = format!("localhost:{}", port_start);
+        port_start += 1;
+
+        let stacks_public_key = StacksPublicKey::from_private(stacks_private_key).to_hex();
+        let db_dir =
+            format!("/tmp/stacks-node-tests/integrations-signers/signer_{stacks_public_key}");
+        let db_path = format!("{db_dir}/signerdb.sqlite");
+        fs::create_dir_all(&db_dir).unwrap();
```
jcnelson, line 353: You should remove the directory before creating it, in case the same private keys get reused across invocations.  Otherwise you'd have some surprising test behavior on the second and subsequent runs.

jcnelson (commented):
Honestly, I find myself questioning why this PR is necessary at all.  We already treat the signer as having soft state; all of the state for the signer's continued operation ought to be stored in the relevant StackerDB, so a signer can boot from a cold VM.  Shouldn't we be storing the serialized, encrypted signer state in a StackerDB slot?  Then, the signer would be able to come back online from any VM anywhere as long as the operator had the decryption key handy.

Cargo.toml
xoloki: Yeah this isn't getting merged until I publish the new wsts crate.  But I can't do that until @jcnelson  or @kantai  approves wsts#7

jcnelson (changes requested):
We already have a more robust means of storing signing state that is resilient to signer failure -- you encrypt the state and store it into a StackerDB, so each and every stacker has a copy they an serve you in the event your signer VM crashes.  This approach is more robust because it would allow the signer to come back online even if the hardware or host OS corrupts local persisted state, which something that we're likely to encounter in a heterogeneous signer environment like the one we're building towards.

I'm having a very hard time envisioning a world where the signer just stops and restarts as part of its normal operation, except for software upgrades.  It seems to me that the signer ought to be online 24/7 otherwise, and any unplanned cessation is due to an exceptional circumstance which includes the host environment becoming unusuable.  Let's craft a persistence solution that accounts for this failure mode, since we're already 90% there.  This PR could simply instead write the serialized state to a StackerDB slot on save, and load it from the same slot on restore, and use an encryption key given in the config to keep the state confidential.

Cargo.toml
jcnelson: Ah, please feel free to be aggressive at pinging me when you're waiting for one of our reviews! :D

Approved :pray:

xoloki:
> Shouldn't we be storing the serialized, encrypted signer state in a StackerDB slot? Then, the signer would be able to come back online from any VM anywhere as long as the operator had the decryption key handy.

I like that idea.  Operators should all be able to store a single private key as a matter of art, or they shouldn't be in crypto.

I would worry though that StackerDB wouldn't be able to hold all of the state machine data.  So we might not be able to come back in the middle of a signing round.  But at least we would never lose the critical private data.

Maybe doing both is a good idea?  Writing everything locally but keeping the critical data in StackerDB?

jcnelson:
> I like that idea. Operators should all be able to store a single private key as a matter of art, or they shouldn't be in crypto.

> I would worry though that StackerDB wouldn't be able to hold all of the state machine data. So we might not be able to come back in the middle of a signing round. But at least we would never lose the critical private data.

> Maybe doing both is a good idea? Writing everything locally but keeping the critical data in StackerDB?

I think that if a StackerDB-based system could be made to work, then there'd be no need for local storage.  If you can't reach your StackerDB-hosted persisted state, you can't perform signing rounds and DKG anyway.

I think the question of the extent to which we should rely on local storage comes down to three things:  (1) your second point about whether or not a StackerDB instance could hold all of the state machine data, (2) how long-lived the signer's persisted data will be, and (3) how often we need to store the machine data.  If we can completely store it in StackerDB, and we don't need to store copies very often, then local storage would be unnecessary.  But if instead the state couldn't fit into a StackerDB slot, or if we had to write a lot of state frequently, then there's a design space to be explored whereby some state could be stored to StackerDB, and other state stored locally, depending on what kinds of signer failures (including host failures) we'd like to be able to recover from.  Also, if losing the local state isn't consequential to the signer (i.e. the data is transient as it is), then it's less important that the data be persisted to StackerDB since a re-instantiated signer won't need it anyway.  For example, local state might be necessary to recover from failure in the middle of a signing round, since singing rounds happen frequently, but it's not the end of the world if that data gets lost due to a signer host failure because the signer can skip some fraction of signing rounds without a penalty.

Do you know the maximum size of the serialized state's serialized and encrypted representation?  Do you know how often we'll need to persist it, and what the consequences of losing part or all of it would be?  If we can get good answers to these questions, then this PR could be reworked to make the right trade-offs.

stacks-signer/src/config.rs
kantai: Right -- this is why I suggested using a run stamp. The way that this is setup now, test runs will step on each other and use the same db_path. The run stamp prevents that from happening, and also lets you access the prior run's state for debugging.

For the concern that you actually need this to be deterministic, I'm not sure that I agree -- if you just generate the configs at the start of the test, when you restart a signer, you can just pass it the same config it started with and it will have the same db.

kantai:
I disagree with trying to use the StackerDB to persist state. Fundamentally, the signer binary needs to have its own view of the world, somewhat independent of its stacks-node. The stacks-node cannot tell the signer binary when a tenure extension is necessary, whether a miner's first proposed block is building off an appropriate tip, or when no more blocks in a miner's tenure should be approved. Using StackerDB as a persistence layer just adds even more complexity to the signer's view: access to signerdb slots in `.signers-x-y` depends on whether or not the signer activated during a cycle _and_ which other signers activated during that cycle.

Getting up to date information from the stackerdb further requires that the stacks-node be at the chain tip and that the stackerdb information be propagated not as a best-effort, but eventually consistent, like block storage.

The most important reason for signers to maintain state is so that they _never_ present inconsistent information to the world. A signer is a single centralized party, so they never need to exhibit split-brain behavior: if a signer signs off on Block B, they should never sign off on Block B', unless Block B failed to clear the signing threshold or propagate. Stacks nodes _are not_ like this and they cannot be like this. There are many ways in which a signer could present inconsistent information after restarts if their state is weakly stored (which would be the case in the stackerdb). If the signer's infrastructure crashes, the stacks-node's state is not guaranteed to be resilient: `stacks-node` chainstates are only guaranteed to be resilient to safe stops. So, the infrastructure manager needs to restart from a chainstate snapshot or genesis sync (or otherwise) -- losing their stackerdb state. This state is only replicated in a best-effort manner, so its unclear if the signer binary on restart would be able to recover anything, let alone the latest operation that they performed.

Instead, if signers are responsible for storing their own state, people who operate signers are given complete control over the management of that state. This matches the responsibility for correct storage (the signer) with the component responsible for correct operation (the signer). While DKG state is less critical on restarts than things like outstanding sign-offs, the simplest approach is to just store all signer specific state in signer managed storage.

jcnelson:
> Fundamentally, the signer binary needs to have its own view of the world, somewhat independent of its stacks-node. The stacks-node cannot tell the signer binary when a tenure extension is necessary, whether a miner's first proposed block is building off an appropriate tip, or when no more blocks in a miner's tenure should be approved.

To be clear, I'm not advocating for dropping local storage altogether.  It's vital that signers maintain a history of the block proposals they've signed so they don't accidentally sign forks, as you pointed out.  What I'm saying in this specific PR is that saving party shares in StackerDB is more resilient than saving it locally.

> Using StackerDB as a persistence layer just adds even more complexity to the signer's view: access to signerdb slots in .signers-x-y depends on whether or not the signer activated during a cycle and which other signers activated during that cycle.

> Getting up to date information from the stackerdb further requires that the stacks-node be at the chain tip and that the stackerdb information be propagated not as a best-effort, but eventually consistent, like block storage.

Technically speaking, StackerDB is read-your-writes locally with monotonic reads globally.  It's at least as resilient as local storage for a signer -- whatever data you write to your slots is immediately readable (just like a local DB), since the Stacks node immediately stores the data to local storage on the RPC to store a chunk.  Everyone else may see stale data, but a remote reader never reads a slot with an older timestamp than the last read (and eventually, all nodes everywhere see your chunk).

It sounds like you're concerned about the signer acting on stale party shares?  If so, then neither storage solution helps or hurts-- they both offer the same consistency guarantee to local clients.  Problems relating to stale party shares arise when the Stacks node gets behind the chain tip.

> The most important reason for signers to maintain state is so that they never present inconsistent information to the world. A signer is a single centralized party, so they never need to exhibit split-brain behavior: if a signer signs off on Block B, they should never sign off on Block B', unless Block B failed to clear the signing threshold or propagate.

The Stacks node doesn't meddle with the contents of a StackerDB unless its controlling smart contract tells it to do so, and in this case, that only happens at a reward cycle boundary (and recall that during a prepare phase, both the current and upcoming reward cycles' StackerDBs will be able to persist data, so the only way you'd be left in in an irrecoverable state is if your node is so far behind that you miss participating in DKG -- something that would happen with local or StackerDB storage).

So, it should be safe to use the StackerDB to store party share data even if the node is a little bit behind the tip.

> If the signer's infrastructure crashes, the stacks-node's state is not guaranteed to be resilient: stacks-node chainstates are only guaranteed to be resilient to safe stops. So, the infrastructure manager needs to restart from a chainstate snapshot or genesis sync (or otherwise) -- losing their stackerdb state. This state is only replicated in a best-effort manner, so its unclear if the signer binary on restart would be able to recover anything, let alone the latest operation that they performed.

If your node has corrupted itself to the point that a restoration from a chainstate snapshot or a genesis sync is necessary, you won't be using that node for signing blocks for quite some time anyway.

Furthermore, the StackerDB state in the node would be preserved during an unsafe shutdown since it's not part of the chainstate.  A recovery from a snapshot could still happen without losing the StackerDB replica, and due to the read-your-writes guarantee from StackerDB, your local StackerDB replica would represent your last successful operation.

If you use StackerDB to store this information, you're getting all the same benefits and consistency guarantees as local storage, plus a measure of (admittedly best-effort) durability.  However, the signer can _enhance_ the durability of their signing state by posting their party shares to multiple nodes, thereby ensuring that as long as one of those nodes is still reachable when the signer recovers, resuming with fresh party shares is doable.  Due to the read-your-writes nature of StackerDB, each of those nodes will have stored the written chunk before acknowledging.

Cargo.toml
netrome: Set to 9.0.0 :heavy_check_mark:

stacks-signer/src/config.rs
netrome: Sure, allowing private keys to get reused across invocations was my intention here. But I understand that other integration tests might hard-code private keys or use seeded random numbers so I will re-introduce the run stamp. This will require us to store the signer config in the `SignerTest` - but that shouldn't be any problem.

netrome:
> To be clear, I'm not advocating for dropping local storage altogether. It's vital that signers maintain a history of the block proposals they've signed so they don't accidentally sign forks, as you pointed out. What I'm saying in this specific PR is that saving party shares in StackerDB is more resilient than saving it locally.

Replicating private DKG shares over the p2p-network feels like an unnecessary security risk if the signers have a private storage solution available.

>  It seems to me that the signer ought to be online 24/7 otherwise, and any unplanned cessation is due to an exceptional circumstance which includes the host environment becoming unusuable.

I wouldn't be surprised if a node operator would run a Signer instance in Kubernetes, in which pod restarts are very common while persisted volumes are rarely compromised. Power outages are also non-exceptional circumstances which typically doesn't compromise persisted storage.

I'm less worried about exceptional circumstances since that would by definition be very unlikely to compromise over 30% of the signing power during a reward cycle.

stacks-signer/src/config.rs
netrome: Done :heavy_check_mark:

jcnelson (changes requested):
Still a NACK until I can get some further clarity on how durable party shares need to be.

xoloki:
@netrome rebase before reviews since the rebase will dismiss reviews 😭

netrome:
> @netrome rebase before reviews since the rebase will dismiss reviews 😭

Right, my apologies. I'll rebase once we've sorted out the durability concerns regarding using SignerDB for storing DKG shares.

netrome:
As we agreed today, I have created a follow-up ticket to address the durability concerns by using StackerDB to store the DKG shares: https://github.com/stacks-network/stacks-core/issues/4595

Therefore, with a few approvals we'll be ready to merge this one and begin working on the encryption and SignerDB implementation.

xoloki (approved):
LGTM thanks!

jferrant (dismissed):

kantai (dismissed):

netrome:
@jcnelson just awaiting your approval now.

stacks-signer/src/signerdb.rs
```diff
@@ -134,15 +175,29 @@ impl SignerDb {
         hash: &Sha512Trunc256Sum,
     ) -> Result<(), DBError> {
         debug!("Deleting block_info: sighash = {hash}");
+
         self.db.execute(
             "DELETE FROM blocks WHERE reward_cycle = ? AND signer_signature_hash = ?",
-            &[reward_cycle.to_string(), format!("{}", hash)],
+            &[
+                &u64_to_sql(reward_cycle)? as &dyn ToSql,
+                &format!("{}", hash),
```
jcnelson, line 183: `hash.to_string()`

stacks-signer/src/signerdb.rs
```diff
@@ -80,15 +128,13 @@ impl SignerDb {
         let result: Option<String> = query_row(
             &self.db,
             "SELECT block_info FROM blocks WHERE reward_cycle = ? AND signer_signature_hash = ?",
-            &[&reward_cycle.to_string(), &format!("{}", hash)],
+            &[
```
jcnelson, line 131: You can use `rusqlite::params!` to avoid the cast to `&dyn ToSql`

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -1266,3 +1322,72 @@ fn stackerdb_filter_bad_transactions() {
     }
     signer_test.shutdown();
 }
+
+#[test]
+#[ignore]
```
jcnelson, line 1327: Why is this `#[ignore]`?

jcnelson: Ah nm this is an integration test

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -1266,3 +1322,72 @@ fn stackerdb_filter_bad_transactions() {
...
+/// to Epoch 3.0 boundary to allow block signing.
+///
+/// Test Execution:
+/// The signers sign one block as usual.
+/// Then, one of the signers is restarted.
+/// Finally, the signers sign another block with the restarted signer.
+///
+/// Test Assertion:
+/// The signers are able to produce a valid signature after one of them is restarted.
+fn stackerdb_sign_after_signer_reboot() {
```
jcnelson, line 1342: Please add this to the `bitcoin-tests.yaml` file so it runs in CI

stacks-signer/src/signerdb.rs
```diff
@@ -70,6 +83,41 @@ impl SignerDb {
...
 
+    /// Get the signer state for the provided reward cycle if it exists in the database
+    pub fn get_signer_state(&self, reward_cycle: u64) -> Result<Option<SignerState>, DBError> {
+        let result: Option<String> = query_row(
+            &self.db,
+            "SELECT state FROM signer_states WHERE reward_cycle = ?",
+            &[u64_to_sql(reward_cycle)?],
+        )?;
+
+        try_deserialize(result)
```
jcnelson, line 94: Given that the callers of this function panic if it fails to load the signer state, should this function simply panic if it fails to load or deserialize state?  When does it make sense to _not_ panic?

jcnelson (commented):
I have a couple questions that I'd like answers to before approving.

stacks-signer/src/signerdb.rs
netrome: This is unrelated to the current PR. Changing this to `.to_string()` would render this function useless since the hashes are inserted using `format!("{}", hash` in the `insert_block` function.

stacks-signer/src/signerdb.rs
netrome: `rusqlite::params!` just does `&dyn ToSql` under the hood https://docs.rs/rusqlite/latest/src/rusqlite/lib.rs.html#176-183. Given that it's not much more condense or easy to read, I would rather skip having a macro invocation here.

testnet/stacks-node/src/tests/signer.rs
netrome: Thanks for noting. Will do.

stacks-signer/src/signerdb.rs
```diff
@@ -70,6 +83,41 @@ impl SignerDb {
...
+        let serialized_state = serde_json::to_string(signer_state)?;
+        self.db.execute(
+            "INSERT OR REPLACE INTO signer_states (reward_cycle, state) VALUES (?1, ?2)",
+            &[&u64_to_sql(reward_cycle)? as &dyn ToSql, &serialized_state],
+        )?;
+        Ok(())
+    }
+
+    /// Delete the signer state for the provided reward cycle and signer ID
+    pub fn delete_signer_state(&self, reward_cycle: u64) -> Result<(), DBError> {
```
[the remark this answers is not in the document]
netrome: It's unused. I added it to just for the sake of completeness of the interface, and to be consistent with the `insert_block`, `remove_block` functions. `remove_block` is also unused. If you prefer to call YAGNI here, I can happily remove them. Let me know what you think.

stacks-signer/src/signerdb.rs
netrome: All functions in this module currently return errors which are panicked on by their callers. One argument for this would be that it makes testing easier, because you can assert that specific errors occur.

If we want to change this, it would be cleaner to do it as a follow-up so that it can be reviewed separately. I'd be happy to do that if you think this makes more sense.

stacks-signer/src/signerdb.rs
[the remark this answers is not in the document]
netrome: Actually, since there are other considerations that are more relating to this module than the current PR - I'd rather agree on what we want to do and create a separate PR for that refactor. Topics for this would be:

1. Return results or do eager panics?
2. `rusqlite::params` vs `&dyn ToSql`
3. Prune unused removal functions?

testnet/stacks-node/src/tests/signer.rs
netrome: Done :heavy_check_mark:

netrome:
> I have a couple questions that I'd like answers to before approving.

I've responded to your questions and updated `bitcoin-tests.yml` now.

stacks-signer/src/signerdb.rs
jcnelson: So change it there too?

stacks-signer/src/signerdb.rs
jcnelson: I'm aware of what `rusqlite::params!` does. That's why I'm asking you to use it. We're in the process of switching the codebase over to it.

stacks-signer/src/signerdb.rs
jcnelson: That answers the question -- it should return a `Result` because we can then test for specific errors, which is desirable.

stacks-signer/src/signerdb.rs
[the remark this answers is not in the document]
jcnelson: 1. Return results
2. `rusqlite::params`
3. YAGNI
4. Do it all in this PR.  We're talking less than 10 lines of refactoring.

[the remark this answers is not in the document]
netrome: Done.

stacks-signer/src/signerdb.rs
netrome: Did not know we were in such a process. Changed now.

stacks-signer/src/signerdb.rs
netrome: Done.

jcnelson (approved):

xoloki (approved):
LGTM

kantai (approved):

PR #4572, opened 2024-03-21
Chore/fix clippy nursery warning

ASuciuX:
Fix nursery warnings.
Left Warnings: 
- [ ] clippy::option_if_let_else
- [ ] clippy::cognitive_complexity

As this is mainly for perf and correctness lints, not so much about the style ones, the left ones aren't a concern.


jferrant (dismissed):

jferrant:
@wileyj do you think we could add cargo clippy to stacks-signer? I used to have it clippy compliant and have since let it fall to the wayside and ti would be nice to enforce it so we don't have massive clippy changes in the future.

stacks-signer/src/cli.rs
```diff
@@ -262,11 +262,10 @@ fn parse_contract(contract: &str) -> Result<QualifiedContractIdentifier, String>
 
 /// Parse a BTC address argument and return a `PoxAddress`
 pub fn parse_pox_addr(pox_address_literal: &str) -> Result<PoxAddress, String> {
-    if let Some(pox_address) = PoxAddress::from_b58(pox_address_literal) {
-        Ok(pox_address)
-    } else {
-        Err(format!("Invalid pox address: {}", pox_address_literal))
-    }
+    PoxAddress::from_b58(pox_address_literal).map_or_else(
+        || Err(format!("Invalid pox address: {}", pox_address_literal)),
```
jbencin, line 266: If we're changing this line, might as well inline the variable too:

```suggestion
        || Err(format!("Invalid pox address: {pox_address_literal}")),
```

jbencin (approved):
Looks good, might even get a tiny perf gain from the `.clone()` deletions and lazy evaluation

jbencin:
> do you think we could add cargo clippy to stacks-signer?

We should do this for Clarity and `stacks-common` too, as they are also mostly Clippy compliant right now

wileyj:
> @wileyj do you think we could add cargo clippy to stacks-signer? I used to have it clippy compliant and have since let it fall to the wayside and ti would be nice to enforce it so we don't have massive clippy changes in the future.

possibly - we could specifically target areas of the code base to enforce with clippy. open a backlog issue for me, happy to look into it (or maybe @BowTiedDevOps ).

ASuciuX:
I had to solve a conflict with next. @jferrant @jbencin can you please re-approve this?

jbencin (approved):

PR #4576, opened 2024-03-21
2.5: PoX missed-slot updates

kantai:
### Description

This PR removes the unlock behavior of PoX-4 when solo stackers miss a reward slot. This behavior was broken `next`: missed reward slot unlocks have consequences for signer set calculation, which has to be addressed. This PR addresses it by removing the unlock behavior.

This PR also adds a test that covers the case of a 2.5 reward cycle having participation but no slots. Such a cycle should still allow the system to make forward progress in 2.5.

### DIFF
--- a/stacks-common/src/types/mod.rs
+++ b/stacks-common/src/types/mod.rs
@@ -92,6 +92,16 @@ impl StacksEpochId {
             StacksEpochId::Epoch24 | StacksEpochId::Epoch25 | StacksEpochId::Epoch30 => true,
         }
     }
+
+    /// Does this epoch support unlocking PoX contributors that miss a slot?
+    ///
+    /// Epoch 2.0 - 2.05 didn't support this feature, but they weren't epoch-guarded on it. Instead,
+    ///  the behavior never activates in those epochs because the Pox1 contract does not provide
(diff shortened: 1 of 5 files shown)


jcnelson (approved):
LGTM as long as CI passes.  I notice that this PR doesn't remove the ancillary PoX state that is required for auto-unlock.  Is that scheduled for a follow-on PR, or should it be done here?

kantai:
>  Is that scheduled for a follow-on PR, or should it be done here?

That's a great question -- it impacts 2.5 activation in testnet, because the PoX contract would necessarily change. I will try to see how easy of a change that is.

kantai:
> LGTM as long as CI passes. I notice that this PR doesn't remove the ancillary PoX state that is required for auto-unlock. Is that scheduled for a follow-on PR, or should it be done here?

Quick follow-up!

The reward set indexes data is used both for the (removed) `handle-unlocks` method and for `stack-increase` (its how it figures out which items to increase), so we can't remove that tracking state. However, that tracking state is much less volatile without `handle-unlocks` (because you never have to perform removals in the `stacking-state`).

obycode (approved):
🎉

PR #4578, opened 2024-03-22
fix: allow `nakamoto-neon` mode

obycode:
PR #4577 removed `nakamoto-neon` mode, in order to make nakamoto the default, but did not remove `nakamoto-neon` everywhere. This change re-adds support for "nakamoto-neon" mode, but also leaves nakamoto mode as the default for xenon and mainnet modes. 

Some integration tests failed in #4577, but the PR was merged anyway because it seems that the "Bitcoin Tests / Check Tests" check is not required. @wileyj can we make that check required now?

### DIFF
--- a/stackslib/src/burnchains/mod.rs
+++ b/stackslib/src/burnchains/mod.rs
@@ -469,8 +469,8 @@ impl PoxConstants {
     pub fn regtest_default() -> PoxConstants {
         PoxConstants::new(
             5,
-            1,
-            1,
+            3,
+            2,
             3333333333333333,
             1,
             BITCOIN_REGTEST_FIRST_BLOCK_HEIGHT + POX_SUNSET_START,
--- a/stackslib/src/chainstate/burn/operations/leader_block_commit.rs
+++ b/stackslib/src/chainstate/burn/operations/leader_block_commit.rs
@@ -1299,6 +1299,8 @@ mod tests {
         });
 
         let mut burnchain = Burnchain::regtest("nope");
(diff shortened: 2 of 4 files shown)


jferrant (dismissed):

hstove (approved):

wileyj:
> PR https://github.com/stacks-network/stacks-core/pull/4577 removed nakamoto-neon mode, in order to make nakamoto the default, but did not remove nakamoto-neon everywhere. This change re-adds support for "nakamoto-neon" mode, but also leaves nakamoto mode as the default for xenon and mainnet modes.

> Some integration tests failed in https://github.com/stacks-network/stacks-core/pull/4577, but the PR was merged anyway because it seems that the "Bitcoin Tests / Check Tests" check is not required. @wileyj can we make that check required now?

that's the end goal, yes - but until we can resolve that failing test `tests::neon_integrations::bitcoin_reorg_flap`, requiring the `check tests` for PRs will mean all PR's will not have successful required checks

obycode:
I removed that test from the yml file for now.

wileyj:
> I removed that test from the yml file for now.

confirmed https://github.com/stacks-network/stacks-core/blob/next/.github/workflows/bitcoin-tests.yml#L90

in that case, yes we can take a look to ensure all tests taht should be passsing **are** passing, and then require that check tests workflow (which is the whole reason we added it!)

kantai (approved):

hstove (approved):

kantai (approved):

PR #4580, opened 2024-03-22
feat: add flag to output signer key signature in JSON

hstove:
This PR adds a new (optional) flag, `--json`, to the `generate-stacking-signature` command in `stacks-signer`. If specified, the CLI will output all the relevant information in JSON form.

The utility of this JSON is that Stacking apps can add a simple "paste JSON" field, which then auto-fills the relevant fields. Because there is a lot of information that must be shared between the signer and stacker, providing this "standard" can help reduce mistakes.

This "paste JSON" functionality is also already implemented in a beta version of Lockstacks:

image

### DIFF
--- a/stacks-signer/src/main.rs
+++ b/stacks-signer/src/main.rs
@@ -320,12 +320,27 @@ fn handle_generate_stacking_signature(
     )
     .expect("Failed to generate signature");
 
-    if do_print {
-        println!(
-            "\nSigner Public Key: 0x{}\nSigner Key Signature: 0x{}\n\n",
+    let output_str = if args.json {
+        serde_json::to_string_pretty(&serde_json::json!({
+            "signerKey": to_hex(&public_key.to_bytes_compressed()),
+            "signerSignature": to_hex(signature.to_rsv().as_slice()),
+            "authId": format!("{}", args.auth_id),
+            "rewardCycle": args.reward_cycle,
+            "maxAmount": format!("{}", args.max_amount),
+            "period": args.period,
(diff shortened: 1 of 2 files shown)


8marz8 (dismissed):

jferrant (dismissed):

kantai (approved):

stacks-signer/src/main.rs
```diff
@@ -320,12 +320,28 @@ fn handle_generate_stacking_signature(
     )
     .expect("Failed to generate signature");
 
-    if do_print {
-        println!(
-            "\nSigner Public Key: 0x{}\nSigner Key Signature: 0x{}\n\n",
+    let output_str = if args.json {
+        serde_json::to_string_pretty(&serde_json::json!({
```
jcnelson, line 324: If it's going to be for copy/paste, why not just print the JSON directly?  If I wanted it pretty-printed, I'd pipe to `jq`

hstove: Do you feel like pretty-printed JSON is harder to copy/paste and/or share via DM? My thought is that they're both easy to handle, while pretty-print by default allows for more easily debugging the output for someone who might not know about `jq`.

xoloki: Pretty printed json is significantly more difficult to copy paste due to the expansion.

hstove:
I've updated the log line to not pretty-print the JSON. @xoloki @jcnelson @jferrant would you mind giving this a re-review? 🙏

xoloki (approved):

jferrant (approved):

PR #4582, opened 2024-03-22
Remove unnecessary retries throughout code

jferrant:
Was having a hard time following the retry logic throughout the stacks-signer. Have made it more explicit. Removing any outer retry calls unless trying to initialize/refresh the signer as a signer may be started before a node boots.


jferrant:
Replaced by https://github.com/stacks-network/stacks-core/pull/4583

Issue #4543, opened 2024-03-14
Use fee estimation API when creating `vote-for-aggregate-key` transactions


hstove:
During epoch 2.5, signers are required to pay a transaction fee when broadcasting their `vote-for-aggregate-key` transactions. At the moment, this fee is decided from the field `tx_fee_ustx` in the signer's config file.

A more robust approach would be to call the node's fee estimation API and use that as the fee.

Some questions:

- Which fee estimate should be used (low/medium/high)? medium
- Should this be configurable (aka opt-in or opt-out)? opt-out
- Should there be a "max fee" option? yes

PR #4583, opened 2024-03-22
Feat/stacks signer fee estimate

jferrant:
Closes https://github.com/stacks-network/stacks-core/issues/4543

### DIFF
--- a/stacks-signer/src/client/stacks_client.rs
+++ b/stacks-signer/src/client/stacks_client.rs
@@ -459,13 +459,12 @@ impl StacksClient {
     }
 
     /// Helper function to create a stacks transaction for a modifying contract call
-    pub fn build_vote_for_aggregate_public_key(
+    pub fn build_unsigned_vote_for_aggregate_public_key(
         &self,
         signer_index: u32,
         round: u64,
         dkg_public_key: Point,
         reward_cycle: u64,
-        tx_fee: Option<u64>,
         nonce: u64,
     ) -> Result<StacksTransaction, ClientError> {
         debug!("Building {SIGNERS_VOTING_FUNCTION_NAME} transaction...");
@@ -478,9 +477,8 @@ impl StacksClient {
             ClarityValue::UInt(round as u128),
             ClarityValue::UInt(reward_cycle as u128),
         ];
-        let tx_fee = tx_fee.unwrap_or(0);
 
-        Self::build_signed_contract_call_transaction(
+        let unsigned_tx = Self::build_unsigned_contract_call_transaction(
             &contract_address,
             contract_name,
             function_name,
@@ -489,8 +487,8 @@ impl StacksClient {
             self.tx_version,
             self.chain_id,
             nonce,
-            tx_fee,
-        )
+        )?;
+        Ok(unsigned_tx)
     }
 
     /// Helper function to submit a transaction to the Stacks mempool
@@ -602,7 +600,7 @@ impl StacksClient {
 
     /// Helper function to create a stacks transaction for a modifying contract call
     #[allow(clippy::too_many_arguments)]
-    pub fn build_signed_contract_call_transaction(
+    pub fn build_unsigned_contract_call_transaction(
         contract_addr: &StacksAddress,
         contract_name: ContractName,
         function_name: ClarityName,
@@ -611,7 +609,6 @@ impl StacksClient {
         tx_version: TransactionVersion,
         chain_id: u32,
         nonce: u64,
-        tx_fee: u64,
     ) -> Result<StacksTransaction, ClientError> {
         let tx_payload = TransactionPayload::ContractCall(TransactionContractCall {
             address: *contract_addr,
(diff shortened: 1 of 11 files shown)


stacks-signer/src/signer.rs
```diff
@@ -517,31 +517,29 @@ impl Signer {
                 sig: vec![],
             };
             self.handle_packets(stacks_client, res, &[packet], current_reward_cycle);
+        } else if block_info.valid.unwrap_or(false)
```
jcnelson, line 520: I'm curious why this code path is preserved?  I thought the Stacks node miner itself would always be the signing round coordinator for a block?

jcnelson (dismissed):
Overall LGTM, just had a couple of questions that need to be answered prior to merging

obycode (dismissed):

stacks-signer/src/signer.rs
jferrant: Good catch. This was just me fixing a clippy warning without looking into what the branch was. Will remove.

stacks-signer/src/signer.rs
```diff
@@ -1054,6 +1032,44 @@ impl Signer {
...
+            self.reward_cycle,
+            nonce,
+        )?;
+        let tx_fee = if epoch < &StacksEpochId::Epoch30 {
+            debug!("{self}: in pre Epoch 3.0 cycles, must set a transaction fee for the DKG vote.");
+            let fee = if let Some(max_fee) = self.max_tx_fee_ustx {
+                let estimated_fee = stacks_client
+                    .get_medium_estimated_fee_ustx(&unsigned_tx)
+                    .map_err(|e| {
+                        debug!("{self}: unable to estimate fee for DKG vote transaction: {e:?}");
```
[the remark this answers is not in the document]
jferrant: Done :)

stacks-signer/src/signer.rs
jferrant: @kantai could you confirm that me removing this is okay? I assumed it was just overlooked, but want to make sure we are not anticipating that signers operate as acoordinator if for some reason they never set the self.miner key?

kantai: Removing this should be okay, yes.

jcnelson (approved):

stacks-signer/src/signer.rs
jferrant: Just realized in this thought exercise...I am not properly handling DKG triggered in the middle of a current reward cycle still. A signer might try to trigger it as it should. But all other signers will ignore it as they treat the message as not coming from the miner coordinator.

hstove (changes requested):
Everything looks good except the logging of the whole config at startup - I really think we shouldn't have that logged as-is.

stacks-signer/src/client/stacks_client.rs
```diff
@@ -196,9 +198,43 @@ impl StacksClient {
...
+        };
+        let send_request = || {
+            self.stacks_node_client
+                .post(self.fees_transaction_path())
+                .header("Content-Type", "application/json")
+                .json(&request)
+                .send()
+                .map_err(backoff::Error::transient)
+        };
+        let response = retry_with_exponential_backoff(send_request)?;
```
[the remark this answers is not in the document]
jferrant: A 400 does not trigger a retry. Only errors return a retry (like reqwest errors)

stacks-signer/src/config.rs
```diff
@@ -394,7 +411,8 @@ mod tests {
         let network = Network::Testnet;
         let password = "melon";
 
-        let config_tomls = build_signer_config_tomls(&[pk], node_host, None, &network, password);
+        let config_tomls =
+            build_signer_config_tomls(&[pk], node_host, None, &network, password, None);
```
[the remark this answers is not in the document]
jferrant: It will not be set. Its an optional.

hstove (dismissed):
LGTM!

kantai (dismissed):

jcnelson (dismissed):

stacks-signer/src/signer.rs
```diff
@@ -1071,6 +1025,44 @@ impl Signer {
...
+    ) -> Result<StacksTransaction, ClientError> {
+        let mut unsigned_tx = stacks_client.build_unsigned_vote_for_aggregate_public_key(
+            self.stackerdb.get_signer_slot_id().0,
+            self.coordinator.current_dkg_id,
+            dkg_public_key,
+            self.reward_cycle,
+            nonce,
+        )?;
+        let tx_fee = if epoch < &StacksEpochId::Epoch30 {
+            debug!("{self}: in pre Epoch 3.0 cycles, must set a transaction fee for the DKG vote.");
```
jcnelson, line 1044: Might want to make this `INFO` so it's more discoverable by operators

jcnelson (dismissed):

kantai (dismissed):

kantai (approved):

jcnelson (approved):

obycode:
Should we go ahead and merge this, or re-target it to develop?

PR #4588, opened 2024-03-25
Feat: Set activation height for 2.5 in xenon (mainline) testnet

kantai:
### Description

This sets the activation height for 2.5 in mainline testnet to block `2,583,893`. This is 57 blocks before a prepare phase. It is 449 blocks from now (as of opening the PR), which sets an activation height of 3PM eastern time, Thursday.


wileyj (approved):

CharlieC3 (approved):
:shipit:

PR #4591, opened 2024-03-25
Do not attempt to treat miner as a coordinator when updating DKG

jferrant:
This only really matters for epoch 2.5. If this was to happen in epoch 3.0 the chain would be halted.

### DIFF
--- a/stacks-signer/src/coordinator.rs
+++ b/stacks-signer/src/coordinator.rs
@@ -91,17 +91,15 @@ impl CoordinatorSelector {
                 }
             }
             new_index
-        } else {
-            if ROTATE_COORDINATORS {
-                let mut new_index = self.coordinator_index.saturating_add(1);
-                if new_index == self.coordinator_ids.len() {
-                    // We have exhausted all potential coordinators. Go back to the start
-                    new_index = 0;
-                }
-                new_index
-            } else {
-                self.coordinator_index
+        } else if ROTATE_COORDINATORS {
+            let mut new_index = self.coordinator_index.saturating_add(1);
+            if new_index == self.coordinator_ids.len() {
+                // We have exhausted all potential coordinators. Go back to the start
+                new_index = 0;
             }
+            new_index
+        } else {
+            self.coordinator_index
         };
         self.coordinator_id = *self
             .coordinator_ids
--- a/stacks-signer/src/signer.rs
+++ b/stacks-signer/src/signer.rs
@@ -185,10 +185,12 @@ impl std::fmt::Display for Signer {
 }
 
 impl Signer {
-    /// Return the current coordinator. If in the active reward cycle, this is the miner,
-    ///  so the first element of the tuple will be None (because the miner does not have a signer index).
-    fn get_coordinator(&self, current_reward_cycle: u64) -> (Option<u32>, PublicKey) {
-        if self.reward_cycle == current_reward_cycle {
+    /// Return the current coordinator.
+    /// If the current reward cycle is the active reward cycle, this is the miner,
+    /// so the first element of the tuple will be None (because the miner does not have a signer index).
+    /// Otherwise, the coordinator is the signer with the index returned by the coordinator selector.
+    fn get_coordinator(&self, current_reward_cycle: Option<u64>) -> (Option<u32>, PublicKey) {
+        if Some(self.reward_cycle) == current_reward_cycle {
             let Some(ref cur_miner) = self.miner_key else {
                 error!(
                     "Signer #{}: Could not lookup current miner while in active reward cycle",
@@ -415,7 +417,7 @@ impl Signer {
         stacks_client: &StacksClient,
         current_reward_cycle: u64,
     ) {
-        let coordinator_id = self.get_coordinator(current_reward_cycle).0;
+        let coordinator_id = self.get_coordinator(Some(current_reward_cycle)).0;
         match &self.state {
             State::Idle => {
                 if coordinator_id != Some(self.signer_id) {
@@ -445,7 +447,7 @@ impl Signer {
         res: Sender<Vec<OperationResult>>,
         current_reward_cycle: u64,
     ) {
-        let coordinator_id = self.get_coordinator(current_reward_cycle).0;
+        let coordinator_id = self.get_coordinator(Some(current_reward_cycle)).0;
         let mut block_info = match block_validate_response {
             BlockValidateResponse::Ok(block_validate_ok) => {
                 let signer_signature_hash = block_validate_ok.signer_signature_hash;
@@ -556,7 +558,7 @@ impl Signer {
         messages: &[SignerMessage],
         current_reward_cycle: u64,
     ) {
-        let coordinator_pubkey = self.get_coordinator(current_reward_cycle).1;
+        let coordinator_pubkey = self.get_coordinator(Some(current_reward_cycle)).1;
         let packets: Vec<Packet> = messages
             .iter()
             .filter_map(|msg| match msg {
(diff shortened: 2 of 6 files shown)


jcnelson (approved):

stacks-signer/src/signer.rs
```diff
@@ -1230,8 +1226,8 @@ impl Signer {
             }
             return Ok(());
         };
-        let coordinator_id = self.get_coordinator(current_reward_cycle).0;
-        if Some(self.signer_id) == coordinator_id && self.state == State::Idle {
+        let coordinator_id = self.coordinator_selector.get_coordinator().0;
+        if self.signer_id == coordinator_id && self.state == State::Idle {
```
kantai, line 1230: I think it'd be better to update the `self.get_coordinator()` method to take an argument like `consider_miner: bool`. It's not obvious from just looking at the code that using the `coordinator_selector` field directly does this and `self.get_coordinator()` doesn't. Also, coordinator selection may be a place where the signer is making more complex decisions in the future, so I think it makes sense to just abstract the signers' invocation of `coordinator_selector` entirely within `self.get_coordinator()`

kantai (commented):
This looks okay, but I think that this should be encapsulated in the `Signer::get_coordinator()` function.

stacks-signer/src/signer.rs
jferrant: I think if you want to use the self.get_coordinator function I would just make current_reward_cycle an optional though am not sure that makes it much clearer...

jcnelson (approved):
LGTM

stacks-signer/src/signer.rs
jferrant: See https://github.com/stacks-network/stacks-core/pull/4583#discussion_r1541541907

I am starting to question whether I should close this ticket without merging.  It is not gonna work as is. A signer can trigger the DKG round, but when we receive incoming packets we always get the miner id when in the middle of an active reward cycle. this introduces unnecessary complexity around unwrapping the packet and checking to see if its a DKG packet or not before either trying to treat the miner as the coordinator or the signer. I can certainly do this, but its only for the case where in epoch 2.5 we fail to run dkg before the next reward cycle which if this happened in epoch 3.0 would cause a whole host of other issues. 


@jcnelson @kantai @hstove any opinions?

jferrant: See https://github.com/stacks-network/stacks-core/pull/4598 for my proposed alternative.

kantai: I think that the thing to do here is to match the packet type as like `MinerCoordinated` or `SignerCoordinated`, then use the miner for one and the coordinator selection from the signer set for the other. I would try not to worry _too_ much about the fact that aggregate key voting isn't possible after the reward cycle started, because follow on work may try to address that in Epoch 3.0 by allowing late aggregate key votes to go through.

hstove: > I think maybe a better solution is to fix the error to be "is DKG not set and we are in the active reward cycle? Just immediately return as we cannot fix this"

My understanding was that we updated signers-voting to allow voting on an aggregate key even if the cycle has already started? If so, this PR seems like the right fix, because it'll execute DKG for the correct reward cycle?

jferrant: Alright, I have enabled DKG in the active reward cycle. Additionally, i have maintained the functionality that a signer could trigger a sign round (mostly for testing purposes) if no miner pubkey is known (will also help for when we make signers sign things other than blocks).

obycode (approved):
lgtm

kantai (dismissed):

stacks-signer/src/coordinator.rs
```diff
@@ -91,17 +91,15 @@ impl CoordinatorSelector {
...
-                let mut new_index = self.coordinator_index.saturating_add(1);
-                if new_index == self.coordinator_ids.len() {
-                    // We have exhausted all potential coordinators. Go back to the start
-                    new_index = 0;
-                }
-                new_index
-            } else {
-                self.coordinator_index
+        } else if ROTATE_COORDINATORS {
+            let mut new_index = self.coordinator_index.saturating_add(1);
```
jcnelson, line 95: Not required, but you could simplify this to

```rust
let new_index = self.coordinator_index.saturating_add(1) % self.coordinator_ids.len();
```

and remove the `if` and `mut`.

stacks-signer/src/signer.rs
```diff
@@ -415,24 +423,34 @@ impl Signer {
         stacks_client: &StacksClient,
         current_reward_cycle: u64,
     ) {
-        let coordinator_id = self.get_coordinator(current_reward_cycle).0;
         match &self.state {
             State::Idle => {
+                let Some(command) = self.commands.pop_front() else {
```
jcnelson, line 428: Not required, but one way this could be made a bit more robust IMHO is to avoid doing the `pop_front()` until you've actually carried out the command.  You're just peeking the command for most of this `State::Idle` case; you can defer popping the command until the call to `execute_command` at the end.  Then, if this code body gets bigger, the future authors don't have to worry about "putting back" the command that got popped at the start of this block.  Just a thought :)

stacks-signer/src/signerdb.rs
```diff
@@ -116,13 +116,13 @@ impl SignerDb {
         self.db
             .execute(
                 "INSERT OR REPLACE INTO blocks (reward_cycle, signer_signature_hash, block_info) VALUES (?1, ?2, ?3)",
-                &[reward_cycle.to_string(), format!("{}", hash), block_json],
+                [reward_cycle.to_string(), format!("{}", hash), block_json],
```
jcnelson, line 119: Nit: Can you use `hash.to_string()` here instead?

stacks-signer/src/signerdb.rs
```diff
@@ -136,7 +136,7 @@ impl SignerDb {
         debug!("Deleting block_info: sighash = {hash}");
         self.db.execute(
             "DELETE FROM blocks WHERE reward_cycle = ? AND signer_signature_hash = ?",
-            &[reward_cycle.to_string(), format!("{}", hash)],
+            [reward_cycle.to_string(), format!("{}", hash)],
```
jcnelson, line 139: Same here with `hash.to_string()`

jcnelson (dismissed):
LGTM; just minor things

stacks-signer/src/signer.rs
jferrant: I purposefully did it this way because we cannot borrow commands while simultaneously borrowing self in the execute command call.

jcnelson (dismissed):

stacks-signer/src/signer.rs
```diff
@@ -1268,14 +1270,17 @@ impl Signer {
...
-            // Only get the account nonce of THIS signer as we only care about our own votes, not other signer votes
-            let signer_address = stacks_client.get_signer_address();
-            let account_nonces = self.get_account_nonces(stacks_client, &[*signer_address]);
-            let old_transactions = self.get_signer_transactions(&account_nonces).map_err(|e| {
+        if self.state != State::Idle || self.signer_id != self.get_coordinator_dkg().0 {
+            // We are not the coordinator or we are in the middle of an operation. Do not attempt to queue DKG
+            return Ok(());
+        }
+        debug!("{self}: Checking if old DKG vote transaction exists in StackerDB...");
+        debug!("{self}: Checking if old DKG vote transaction exists in StackerDB...");
```
obycode, line 1278: duplicate logging

obycode (approved):
LGTM just one accidental repeated log line.

kantai (dismissed):

stacks-signer/src/signer.rs
jferrant: argh! Done.

kantai (approved):

jcnelson (approved):

kantai (approved):

jcnelson (approved):

PR #4592, opened 2024-03-26
adding missing flag to cargo.toml

wileyj:
#4331 added the ability to build for multiple arches and OSes. 

this was changed in https://github.com/stacks-network/stacks-core/commit/207cb690fe8cb25cba6734ec4492f077108cc30b resulting in only glibc based sytems building as part of the ci workflows. 

this change brings back the flag added in #4331

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ rand = "0.8"
 rand_chacha = "0.3.1"
 tikv-jemallocator = "0.5.4"
 # wsts = { version = "8.1", default-features = false }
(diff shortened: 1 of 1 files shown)


wileyj:
confirmed that adding this back builds on linux-musl (will check the rest of the binary builds later)

xoloki (approved):
Good catch.

I published wsts 9.0 crate today, so we could also point to that instead.

kantai (approved):
Thanks for the catch. Sorry for missing this.

PR #4601, opened 2024-03-28
Attempt to fix bitcoin reorg flap test

obycode:
Forcefully kill the bitcoind in case it fails to stop.


obycode:
It looks like some change in between the original fix to this problem and the current next has caused the test to fail, so that needs to get fixed. I'll look into it tomorrow.

wileyj:
this PR still needs more work. brice and I have discussed some alternatives but it may not be ready for 2.5. 

i would recommend we move this to draft/backlog for now

obycode:
Yes, this was fixed by #5207.

Issue #4587, opened 2024-03-25
[Mutations Testing] Highlighting Functions not fully covered by their tests


ASuciuX:
**Is your feature request related to a problem? Please describe.**
Some functions don't have all their cases covered.

**Describe the solution you'd like**
This issue aims to keep track of them, update the CI to pass for the moment with these functions skipped.
It also intends to add TODO comments as references for these functions so the other developers are aware of which functions should have their tests updated.

**Additional context**
This is an example of a workflow summary that will be updated in the code
https://github.com/stacks-network/stacks-core/actions/runs/8361354334
The missed will have specified that in the todo, the unviable as well.

Link the solution to the documentation https://github.com/stacks-network/stacks-core/blob/next/docs/ci-release.md#mutation-testing

**PRs covered**
- [ ] #4412
- [ ] #4414 
- [ ] #4527
- [ ] #4530
- [ ] #4556 
- [ ] #4560 
- [ ] #4561 
- [ ] #4575 
- [ ] #4576 
- [ ] #4578

PR #4560, opened 2024-03-19
Add affirmation overrides config option and add default xenon ones ne…

jferrant:
Closes: https://github.com/stacks-network/stacks-core/issues/4493

PR #4604, opened 2024-03-28
Add skips and comments for certain mutants 

ASuciuX:
Related issue #4587 

List of PRs addressed:
- [x] #4412
- [x] #4414 
- [x] #4527
- [x] #4530
- [x] #4556 
- [x] #4560 
- [x] #4561 
- [x] #4575 
- [x] #4576 
- [x] #4578


ASuciuX:
This follows the same structure as https://github.com/stacks-network/stacks-core/pull/4593 which was previously merged. It includes the mutants for multiple PRs.

jbencin (commented):
I think the associated comments should be regular comments, not doc comments. `cargo doc` is great for figuring out how to use an API, and these comments don't help anyone actually *use* the functions

ASuciuX:
> I think the associated comments should be regular comments, not doc comments. `cargo doc` is great for figuring out how to use an API, and these comments don't help anyone actually _use_ the functions

Those comments act as placeholders for missing tests, they should be reference for who wants to write the test cases that are in the comments and aren't treated by the functions. There should be something on the code to point to the specific missing cases so that a developer could know which test cases to add. Would it be enough to change them to simple comments instead of doc comments? Or do you think we should move them somewhere externally but also keep a short reference so that the cases can be seen when going to that external source. Also curious what that external source should be and how should I give access to the others to modify it if some of the tests are added.

jbencin:
> Would it be enough to change them to simple comments instead of doc comments?

This is fine by me. Although you should add a second reviewer on here for another opinion

ASuciuX:
> > Would it be enough to change them to simple comments instead of doc comments?
> 
> This is fine by me. Although you should add a second reviewer on here for another opinion

I've added the blockchain team, but it treated your review in that behalf, so I'll add it again.

ASuciuX:
@jbencin I've updated this to have regular comments

ASuciuX:
As this is outdated and for `next` branch, I will move it to `develop`.

ASuciuX:
Migrated the PR here https://github.com/stacks-network/stacks-core/pull/4790.

PR #4610, opened 2024-04-01
Add delegate-stack-extend command

BowTiedRadone:
This PR adds the `delegate-stack-extend` command to the stateful property testing environment. It is part of #4548 and targets `feat/pox-4-stateful-property-testing` (#4550).

### DIFF
--- a/contrib/core-contract-tests/tests/pox-4/pox_DelegateStackExtendCommand.ts
+++ b/contrib/core-contract-tests/tests/pox-4/pox_DelegateStackExtendCommand.ts
@@ -1,0 +1,160 @@
+import {
+  logCommand,
+  PoxCommand,
+  Real,
+  Stub,
+  Wallet,
+} from "./pox_CommandModel.ts";
+import { poxAddressToTuple } from "@stacks/stacking";
+import { assert, expect } from "vitest";
+import {
+  Cl,
+  ClarityType,
+  isClarityType,
+} from "@stacks/transactions";
+import {
+  FIRST_BURNCHAIN_BLOCK_HEIGHT,
+  REWARD_CYCLE_LENGTH,
+} from "./pox_Commands.ts";
+
+/**
+ * The `DelegateStackExtendCommand` allows a pool operator to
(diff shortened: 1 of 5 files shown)


moodmosaic (commented):
LGTM; it follows precisely the pattern we use for all the existing commands. I'd like @friedger to look into the [`check`](https://github.com/stacks-network/stacks-core/pull/4610/files#diff-ae5601b1d16d77f2722b63c2cbd0ace3f8de89a3380c20d430a31b96ca1dcdeeR60) method, and then we can merge.

moodmosaic:
Merging into `feat/pox-4-stateful-property-testing`.

contrib/core-contract-tests/tests/pox-4/pox_DelegateStackExtendCommand.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    const firstRewardCycle =
+      this.currentCycle > this.stacker.firstLockedRewardCycle
+        ? this.currentCycle
+        : this.stacker.firstLockedRewardCycle;
+    const firstExtendCycle = Math.floor(
+      (this.stacker.unlockHeight - FIRST_BURNCHAIN_BLOCK_HEIGHT) /
+        REWARD_CYCLE_LENGTH,
+    );
+    const lastExtendCycle = firstExtendCycle + this.extendCount - 1;
```
friedger, line 78: Why do you need this value? You are substracting one and adding one afterwards.

friedger (commented):
check function looks good

contrib/core-contract-tests/tests/pox-4/pox_DelegateStackExtendCommand.ts
BowTiedRadone: @friedger I am doing this to simulate how the lock period is [calculated](https://github.com/stacks-network/stacks-core/blob/next/stackslib/src/chainstate/stacks/boot/pox-4.clar#L1328-L1330) inside PoX-4 contract

PR #4612, opened 2024-04-02
Fix invalid nonce log in valid_vote_transaction and flush stackerdb transactions that are outdated

jferrant:
This is not the most efficient thing ever as we will be polling the node quite frequently. might be better to do this every x number of stacks blocks for example. however, currently we never refresh stackerdb which spams the node with lots of "outdated nonce" logs as confirmed vote transactions are never flushed.


jferrant:
Actually I don't think this PR is worth it. It will needlessly spam the node just to prevent a log message from hitting for a case that is already properly handled by signers and miners.

Issue #4615, opened 2024-04-02
[stacks-signer] Signers should broadcast the same DKG when their broadcasted transaction is no longer valid


jferrant:
Currently signers only best effort construct a DKG vote-for-aggregate transaction. If the same account builds a vote at the same time as another transaction submission, nonces could be wrong. The signer will not think to broadcast an updated transaction if the nonce is bad and may even trigger an entirely new dkg round. Should not do a new DKG round, but should attempt to rebroadcast their vote if it was never registered in the contract and their nonce is oudated.

PR #4621, opened 2024-04-02
Consider nonce when checking for existing vote transactions in stacke…

jferrant:
This PR addresses the issue that can occur if a signer account is used to generate stacks transactions external to the signer binary operation. This is not normally an issue unless the created transaction occurs in the same block. Signers should attempt to resubmit a vote if it is found to be unconfirmed (i.e. they have not voted yet) and the tx is still sitting in stackerdb transaction slot with an outdated nonce.

Closes https://github.com/stacks-network/stacks-core/issues/4615


hstove (approved):
LGTM!

kantai (approved):

Issue #4558, opened 2024-03-19
Construct `signer_bitvec` based on who signed each block


hstove:
We currently include a `signer_bitvec` in Nakamoto blocks and associated emitted events, but this BitVec is not actually constructed with "real" values based on which signers submitted signatures. We need to change this behavior so that each block's bitvec is updated with real information.

A suggestion was to build this off of #4481 , because the miner can readily access this information when acting as a coordinator.

PR #4627, opened 2024-04-02
feat: start of ability for miner to keep track of active signers bitvec

hstove:
This PR adds two things:

- When a new Nakamoto block is formed, it starts with a BitVec of 1's with the length of the signer set.
- In the miner's coordinator thread, a `next_signer_bitvec` property is added. After a miner has collected all block responses from the signers, it updates this BitVec to match the signers who gave a response.

Note that, from a functional perspective, all this changes is that the length of the BitVec will match the signer's length. However, the BitVec will be all 1's, even if not all signers respond. Future work is required to construct a new block in the case that a block is rejected, and in that work we can use this `next_signer_bitvec` in that new block.

- Closes https://github.com/stacks-network/stacks-core/issues/4558

### DIFF
--- a/stackslib/src/chainstate/burn/db/sortdb.rs
+++ b/stackslib/src/chainstate/burn/db/sortdb.rs
@@ -3515,15 +3515,14 @@ impl SortitionDB {
         let Some(reward_set) = reward_info.known_selected_anchor_block() else {
             return None;
         };
-        Some(
-            reward_set
-                .signers
-                .clone()
-                .map(|x| x.len())
-                .unwrap_or(0)
-                .try_into()
-                .expect("FATAL: size of reward set is larger than u16"),
-        )
+
+        reward_set
+            .signers
+            .clone()
+            .map(|x| x.len())
+            .unwrap_or(0)
+            .try_into()
+            .ok()
     }
 }
 
--- a/stackslib/src/chainstate/burn/db/sortdb.rs
+++ b/stackslib/src/chainstate/burn/db/sortdb.rs
@@ -3507,6 +3507,24 @@ impl SortitionDB {
 
         Ok(rc_info)
     }
+
+    pub fn get_preprocessed_reward_set_size(&self, tip: &SortitionId) -> Option<u16> {
+        let Ok(Some(reward_info)) = &self.get_preprocessed_reward_set_of(&tip) else {
+            return None;
+        };
+        let Some(reward_set) = reward_info.known_selected_anchor_block() else {
+            return None;
+        };
+        Some(
+            reward_set
+                .signers
+                .clone()
+                .map(|x| x.len())
+                .unwrap_or(0)
+                .try_into()
+                .expect("FATAL: size of reward set is larger than u16"),
+        )
+    }
 }
 
 impl<'a> SortitionDBTx<'a> {
--- a/testnet/stacks-node/src/nakamoto_node/miner.rs
+++ b/testnet/stacks-node/src/nakamoto_node/miner.rs
@@ -830,7 +830,7 @@ impl BlockMinerThread {
             //  correct signer_sighash for `process_mined_nakamoto_block_event`
             Some(&self.event_dispatcher),
             signer_transactions,
-            signer_bitvec_len.unwrap_or(1),
+            signer_bitvec_len.unwrap_or(0),
         )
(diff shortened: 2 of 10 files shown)


kantai (approved):

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
```diff
@@ -385,6 +397,20 @@ impl SignCoordinator {
                 debug!("Ignoring StackerDB event for non-signer contract"; "contract" => %event.contract_id);
                 continue;
             }
+            let modified_slots = &event.modified_slots;
+
+            // Update `next_signers_bitvec` with the slots that were modified in the event
+            modified_slots.iter().for_each(|chunk| {
+                if let Ok(slot_id) = chunk.slot_id.try_into() {
```
jcnelson, line 404: What's the alternative if this conversion fails?

jcnelson (approved):

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
hstove: Hm, I'd love your input here. If we're at this point in the code, and we already know this is from a signer stackerDB contract, is `slot_id` ever expected to be > u16? My understanding is that slot_id should never be larger than the max reward set size, so I guess my thought is that this is a "if we get here, we have other problems" thing. So, maybe a panic or warn?

jcnelson: Probably an `error!`, but yeah it really ought to be unreachable

kantai (approved):

jcnelson (dismissed):

jferrant (dismissed):

hstove:
Updated to add an error log based on a comment from Jude in [d26e5a6](https://github.com/stacks-network/stacks-core/pull/4627/commits/d26e5a6c473eedb784751f2e207315235251dd2a). Requesting re-reviews

jferrant (dismissed):

jcnelson (dismissed):

hstove:
FYI - I fixed an issue with how the block was formed by the miner, which was breaking integration tests. I've also opened #4669 , which is the same code but off of develop. I'm happy to close this, but wanted to double check if that's the right move.

stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -3341,6 +3341,24 @@ impl SortitionDB {
...
+            return None;
+        };
+        Some(
+            reward_set
+                .signers
+                .clone()
+                .map(|x| x.len())
+                .unwrap_or(0)
+                .try_into()
+                .expect("FATAL: size of reward set is larger than u16"),
```
jcnelson, line 3359: Maybe this should just be `.ok()`?  It would be unreachable, but it would at least result in a predictable failure mode -- it would appear that no signers have signed.

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -827,6 +830,7 @@ impl BlockMinerThread {
             //  correct signer_sighash for `process_mined_nakamoto_block_event`
             Some(&self.event_dispatcher),
             signer_transactions,
+            signer_bitvec_len.unwrap_or(1),
```
jcnelson, line 833: Why 1 and not 0?

hstove: I don't have a good argument - mainly `NakamotoBlock::empty()` and `NakamotoBlock::genesis()` use 1, so I went with that.

stackslib/src/chainstate/burn/db/sortdb.rs
hstove: Added 👍🏼

hstove:
This PR is now based off of develop. Would love re-reviews please 😄

jferrant (approved):

kantai (approved):

testnet/stacks-node/src/nakamoto_node/miner.rs
hstove: Changed this to be zero

PR #4629, opened 2024-04-03
Reward cycle length was being incorrectly stored as reward phase length

jferrant:
Sometimes regtest would stall on boot with the following messages:
	
2024-04-02, 05:32:20.104 PM	
[signer_runloop] Received a new burnchain block height (117) in the prepare phase of the next reward cycle (6). Checking for signer registration...
2024-04-02, 05:32:20.057 PM	
[signer_runloop] Received a new burnchain block height (117) in the prepare phase of the next reward cycle (5). Checking for signer registration.

This made me realize something was up with the reward cycle info or prepare phase calculation. Was accidentally storing the full reward cycle length but calling it the reward phase length and then adding and subtracting the prepare phase length again. This should fix the problem. 

Fixes https://github.com/stacks-network/stacks-core/issues/4630

### DIFF
--- a/stacks-signer/src/client/stacks_client.rs
+++ b/stacks-signer/src/client/stacks_client.rs
@@ -413,13 +413,13 @@ impl StacksClient {
         let blocks_mined = pox_data
             .current_burnchain_block_height
             .saturating_sub(pox_data.first_burnchain_block_height);
-        let reward_phase_block_length = pox_data
+        let reward_cycle_length = pox_data
             .reward_phase_block_length
             .saturating_add(pox_data.prepare_phase_block_length);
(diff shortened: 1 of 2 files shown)


obycode (changes requested):
Can we add some unit tests that would have caught this issue?

kantai (commented):
Yep -- this is definitely a bug and a corresponding fix!

Can you add an assertion to one of the end to end signers tests that would have caught this?

kantai (approved):

jcnelson (approved):

obycode (approved):

obycode:
Sorry, didn't realize this was waiting on me.

PR #4640, opened 2024-04-04
Bugfix: verify block transactions only when the upcoming signers appr…

jferrant:
I just noticed this while explaining an issue to Alin. Signers should be checking the UPCOMING signers aggregate key not their own when determining if they need to include a transaction set. 

Closes https://github.com/stacks-network/stacks-core/issues/4641

### DIFF
--- a/stacks-signer/src/signer.rs
+++ b/stacks-signer/src/signer.rs
@@ -810,10 +810,14 @@ impl Signer {
         stacks_client: &StacksClient,
         block: &NakamotoBlock,
     ) -> bool {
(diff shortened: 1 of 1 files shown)


kantai (approved):

hstove (approved):
This makes sense. I think an argument could be made that this check should be done for the current cycle, too, in 2.5. But I think we should add this as-is

jferrant:
> This makes sense. I think an argument could be made that this check should be done for the current cycle, too, in 2.5. But I think we should add this as-is

There is no point. You will never sign blocks in 2.5

PR #4642, opened 2024-04-04
 Add stack-aggregation-commit-indexed command signature and authorization

BowTiedRadone:
This PR adds the `stack-aggregation-commit-indexed` command, using both signature and authorization, to the stateful property testing environment. It is part of #4548 and targets `feat/pox-4-stateful-property-testing` (#4550).

### DIFF
--- a/contrib/core-contract-tests/tests/pox-4/pox_CommandModel.ts
+++ b/contrib/core-contract-tests/tests/pox-4/pox_CommandModel.ts
@@ -30,9 +30,9 @@ export class Stub {
   }
 
   reportCommandRuns() {
-    process.stdout.write("Command run method execution counts:");
+    console.log("Command run method execution counts:");
     this.statistics.forEach((count, commandName) => {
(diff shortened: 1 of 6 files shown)


moodmosaic (approved):
LGTM

friedger (approved):
LGTM

moodmosaic:
Merging into `feat/pox-4-stateful-property-testing`.

moodmosaic:
> There is code duplication here. However, if you want to keep each command separate, then LGTM.

Yep, it's fine as long as it's not duplicated in more than two<sup>*</sup> commands for now.

---

<sup>*</sup> [Rule of three](https://wou.edu/las/cs/csclasses/cs161/Lectures/rulesofthumb.html), [explicit is better than implicit](https://peps.python.org/pep-0020/), et cetera, but most importantly, each command should be autonomous for now that we're still in progress. Once we have the complete picture (happy and unhappy paths), we may extract any duplicates.

Issue #4643, opened 2024-04-04
[pox-4] Missing Update Aggregation Increase Check


setzeus:
**Describe the bug**
The current version of pox-4 is missing a check required in 'stack-aggregation-increase.' A pool operator submits a signature | approval 'max-amount' when first calling 'stack-aggregation-commit-indexed.' Accordingly, if/when a pool operator _increases_ the aggregate amount stacked, there must still be a check that the new total is lower than the original 'max-amount' provided.

**Steps To Solve**
Update the existing 'stack-aggregation-increase' function to pass in & consume a new signature | authorization, including a new 'max-amount.'

PR #4586, opened 2024-03-25
Integrated/Scenario Tests In Rust

setzeus:
### Description

Out of draft & ready for review, this PR implements & addresses all five scenarios laid out in issue #4554. The scenarios test multiple users & span for multiple reward cycles with the goal of the tests cross through all three boot Clarity contracts [pox-4/signers/signers-voting].

### Applicable issues

- Addresses #4554 

### Additional info (benefits, drawbacks, caveats)
Also slightly updates an error message found in pox-4 to make signer authorizations fail more clearly. Can likely refactor & extract out a few common setup functions but logic is ready for review.

PR #4644, opened 2024-04-04
[pox-4]  Stack Aggregation Increase Update

setzeus:
### Description
Initial draft of a solution to address #4643 found by @janniks. Keeping the logic in parallel with solo 'stack-increase', we opted to adding a required signature | set approval to the existing 'stack-aggregation-increase' function. 

This PR also includes minimal testing setup seen in #4586 as well (which will likely be rebased once this is first merged in). Also includes some comment clean up.

### Applicable issues
#4643

### Checklist

- [ ] Changelog is updated

### DIFF
--- a/stackslib/src/chainstate/stacks/boot/mod.rs
+++ b/stackslib/src/chainstate/stacks/boot/mod.rs
@@ -2185,6 +2185,40 @@ pub mod test {
         make_tx(key, nonce, 0, payload)
     }
 
+    pub fn make_pox_4_aggregation_increase(
+        key: &StacksPrivateKey,
+        nonce: u64,
+        pox_addr: &PoxAddress,
+        reward_cycle: u128,
+        reward_cycle_index: u128,
+        signature_opt: Option<Vec<u8>>,
+        signer_key: &Secp256k1PublicKey,
+        max_amount: u128,
+        auth_id: u128,
+    ) -> StacksTransaction {
+        let addr_tuple = Value::Tuple(pox_addr.as_clarity_tuple().unwrap());
+        let signature = signature_opt
+            .map(|sig| Value::some(Value::buff_from(sig).unwrap()).unwrap())
+            .unwrap_or_else(|| Value::none());
+        let payload = TransactionPayload::new_contract_call(
+            boot_code_test_addr(),
+            POX_4_NAME,
+            "stack-aggregation-increase",
+            vec![
+                addr_tuple,
+                Value::UInt(reward_cycle),
+                Value::UInt(reward_cycle_index),
+                signature,
+                Value::buff_from(signer_key.to_bytes_compressed()).unwrap(),
+                Value::UInt(max_amount),
+                Value::UInt(auth_id),
+            ],
+        )
+        .unwrap();
+
+        make_tx(key, nonce, 0, payload)
+    }
+
     pub fn make_pox_4_stack_increase(
         key: &StacksPrivateKey,
         nonce: u64,
--- a/stackslib/src/chainstate/stacks/boot/pox-4.clar
+++ b/stackslib/src/chainstate/stacks/boot/pox-4.clar
@@ -894,21 +894,21 @@
     (asserts! (> reward-cycle (current-pox-reward-cycle))
               (err ERR_STACKING_INVALID_LOCK_PERIOD))
 
-    (let ((amount-ustx (get stacked-amount partial-stacked))
+    (let ((partial-amount-ustx (get stacked-amount partial-stacked))
           ;; reward-cycle must point to an existing record in reward-cycle-total-stacked
           ;; infallible; getting something from partial-stacked-by-cycle succeeded so this must succeed
-          (existing-total (unwrap-panic (map-get? reward-cycle-total-stacked { reward-cycle: reward-cycle })))
+          (existing-cycle-total (unwrap-panic (map-get? reward-cycle-total-stacked { reward-cycle: reward-cycle })))
           ;; reward-cycle and reward-cycle-index must point to an existing record in reward-cycle-pox-address-list
-          (existing-entry (unwrap! (map-get? reward-cycle-pox-address-list { reward-cycle: reward-cycle, index: reward-cycle-index })
+          (existing-entry-total (unwrap! (map-get? reward-cycle-pox-address-list { reward-cycle: reward-cycle, index: reward-cycle-index })
                           (err ERR_DELEGATION_NO_REWARD_SLOT)))
-          (increased-ustx (+ (get total-ustx existing-entry) amount-ustx))
-          (total-ustx (+ (get total-ustx existing-total) amount-ustx)))
+          (increased-entry-total (+ (get total-ustx existing-entry-total) partial-amount-ustx))
+          (increased-cycle-total (+ (get total-ustx existing-cycle-total) partial-amount-ustx)))
 
           ;; must be stackable
-          (try! (minimal-can-stack-stx pox-addr increased-ustx reward-cycle u1))
+          (try! (minimal-can-stack-stx pox-addr increased-entry-total reward-cycle u1))
 
           ;; new total must exceed the stacking minimum
-          (asserts! (<= (get-stacking-minimum) increased-ustx)
+          (asserts! (<= (get-stacking-minimum) increased-entry-total)
                     (err ERR_STACKING_THRESHOLD_NOT_MET))
 
           ;; there must *not* be a stacker entry (since this is a delegator)
@@ -920,24 +920,24 @@
                     (err ERR_DELEGATION_WRONG_REWARD_SLOT))
 
           ;; Validate that amount is less than or equal to `max-amount`
-          (asserts! (>= max-amount increased-ustx) (err ERR_SIGNER_AUTH_AMOUNT_TOO_HIGH))
+          (asserts! (>= max-amount increased-entry-total) (err ERR_SIGNER_AUTH_AMOUNT_TOO_HIGH))
 
           ;; Verify signature from delegate that allows this sender for this cycle
           ;; 'lock-period' param set to one period, same as aggregation-commit-indexed
-          (try! (consume-signer-key-authorization pox-addr reward-cycle "agg-increase" u1 signer-sig signer-key increased-ustx max-amount auth-id))
+          (try! (consume-signer-key-authorization pox-addr reward-cycle "agg-increase" u1 signer-sig signer-key increased-entry-total max-amount auth-id))
 
           ;; update the pox-address list -- bump the total-ustx
           (map-set reward-cycle-pox-address-list
                    { reward-cycle: reward-cycle, index: reward-cycle-index }
                    { pox-addr: pox-addr,
-                     total-ustx: increased-ustx,
+                     total-ustx: increased-entry-total,
                      stacker: none,
                      signer: signer-key })
 
           ;; update the total ustx in this cycle
           (map-set reward-cycle-total-stacked
                    { reward-cycle: reward-cycle }
-                   { total-ustx: total-ustx })
+                   { total-ustx: increased-cycle-total })
 
           ;; don't update the stacking-state map,
           ;;  because it _already has_ this stacker's state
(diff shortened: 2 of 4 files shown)


stackslib/src/chainstate/stacks/boot/mod.rs
```diff
@@ -2185,6 +2185,40 @@ pub mod test {
...
+                Value::buff_from(signer_key.to_bytes_compressed()).unwrap(),
+                Value::UInt(max_amount),
+                Value::UInt(auth_id),
+            ],
+        )
+        .unwrap();
+
+        make_tx(key, nonce, 0, payload)
+    }
+
```
kantai, line 2221: Since this is only used in `pox_4_tests`, can you put this function in there instead?

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -912,14 +915,19 @@
           (asserts! (is-eq pox-addr (get pox-addr existing-entry))
                     (err ERR_DELEGATION_WRONG_REWARD_SLOT))
 
+          ;; Validate that amount is less than or equal to `max-amount`
```
kantai, line 918: Commenting on lines 904 to 908 -- this is not actually an issue, because stack increase could only ever be called on something that already passed those checks, but those checks use the wrong let binding:

```clar
          ;; must be stackable
          (try! (minimal-can-stack-stx pox-addr total-ustx reward-cycle u1))
          ;; new total must exceed the stacking minimum
          (asserts! (<= (get-stacking-minimum) total-ustx)
                    (err ERR_STACKING_THRESHOLD_NOT_MET))
```

Should actually be:

```clar
          ;; must be stackable
          (try! (minimal-can-stack-stx pox-addr increased-ustx reward-cycle u1))
          ;; new total must exceed the stacking minimum
          (asserts! (<= (get-stacking-minimum) increased-ustx)
                    (err ERR_STACKING_THRESHOLD_NOT_MET))
```

Can you update that in this PR?

kantai (commented):
LGTM -- just caught a different issue in the function that should be handled as well.

stackslib/src/chainstate/stacks/boot/mod.rs
setzeus: We currently define all of the 'make_pox_4_[x]' helper functions in mod.rs though, I think it makes sense to keep them together?

stackslib/src/chainstate/stacks/boot/pox-4.clar
setzeus: Updated accordingly.

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -912,14 +915,19 @@
...
+          (asserts! (>= max-amount increased-ustx) (err ERR_SIGNER_AUTH_AMOUNT_TOO_HIGH))
+
+          ;; Verify signature from delegate that allows this sender for this cycle
+          (try! (consume-signer-key-authorization pox-addr reward-cycle "agg-increase" u1 signer-sig signer-key increased-ustx max-amount auth-id))
```
friedger, line 922: Gut be helpful to explain the u1

friedger (commented):
Looks good

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -912,14 +915,19 @@
...
+
           ;; update the pox-address list -- bump the total-ustx
           (map-set reward-cycle-pox-address-list
                    { reward-cycle: reward-cycle, index: reward-cycle-index }
                    { pox-addr: pox-addr,
                      total-ustx: increased-ustx,
```
obycode, line 928: Not a problem with your change, but this naming scares me and wanted to point it out here. It would be very easy to make the same mistake from pox-2 here because of `total-ustx` and `increased-ustx`.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -3282,6 +3283,323 @@ fn stack_agg_commit_verify_sig() {
...
+    let reward_cycle_len = peer.config.burnchain.pox_constants.reward_cycle_length;
+    let prepare_phase_len = peer.config.burnchain.pox_constants.prepare_length;
+
+    // Advance into pox4
+    let mut target_height = peer.config.burnchain.pox_constants.pox_4_activation_height;
+    let mut latest_block = None;
+    // Produce blocks until the first reward phase that everyone should be in
+    while peer.get_burn_block_height() < u64::from(target_height) {
+        latest_block = Some(peer.tenure_with_txs(&[], &mut peer_nonce));
+        observer.get_blocks();
```
obycode, line 3410: I don't think this call is useful.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -3282,6 +3283,323 @@ fn stack_agg_commit_verify_sig() {
...
+        &AddressHashMode::SerializeP2PKH,
+        1,
+        &vec![StacksPublicKey::from_private(&private_key)],
+    )
+    .unwrap();
+
+    // reward cycles are 5 blocks long
+    // first 25 blocks are boot-up
+    // reward cycle 6 instantiates pox-3
+    // we stack in reward cycle 7 so pox-3 is evaluated to find reward set participation
```
obycode, line 3380: Are these copied from somewhere? I don't think they are correct here.

setzeus: Yeap, done in #4586 with the original setup coming from 'test_scenario_five.'

setzeus:
Confirmed change: update confusing (let ...) variable names.

stackslib/src/chainstate/stacks/boot/pox-4.clar
setzeus: Changed variables accordingly

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
setzeus: Tried removing it but looks like it's necessary for get_stacking_minimum.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
kantai: I would just delete those four lines of comments (they were reasoning about a calculation that's no longer done here).

kantai (dismissed):
This LGTM

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
obycode: I don't think so. It passes for me with that removed.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
obycode: Agreed.

stackslib/src/chainstate/stacks/boot/pox-4.clar
obycode: Much better, and clearly those names were a problem! Thanks!

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -665,6 +664,12 @@
                 (err ERR_STACKING_INVALID_POX_ADDRESS))
          true)
 
+      (match pox-addr
```
jcnelson, line 667: Great catch!

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -887,21 +896,21 @@
...
+    (let ((partial-amount-ustx (get stacked-amount partial-stacked))
           ;; reward-cycle must point to an existing record in reward-cycle-total-stacked
           ;; infallible; getting something from partial-stacked-by-cycle succeeded so this must succeed
-          (existing-total (unwrap-panic (map-get? reward-cycle-total-stacked { reward-cycle: reward-cycle })))
+          (existing-cycle (unwrap-panic (map-get? reward-cycle-total-stacked { reward-cycle: reward-cycle })))
           ;; reward-cycle and reward-cycle-index must point to an existing record in reward-cycle-pox-address-list
           (existing-entry (unwrap! (map-get? reward-cycle-pox-address-list { reward-cycle: reward-cycle, index: reward-cycle-index })
                           (err ERR_DELEGATION_NO_REWARD_SLOT)))
-          (increased-ustx (+ (get total-ustx existing-entry) amount-ustx))
-          (total-ustx (+ (get total-ustx existing-total) amount-ustx)))
```
jcnelson, line 898: Thanks for changing these names

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -912,19 +921,25 @@
...
           ;; update the pox-address list -- bump the total-ustx
           (map-set reward-cycle-pox-address-list
                    { reward-cycle: reward-cycle, index: reward-cycle-index }
                    { pox-addr: pox-addr,
-                     total-ustx: increased-ustx,
+                     total-ustx: increased-entry-total,
                      stacker: none,
-                     ;; TODO: this must be authorized with a signature, or tx-sender allowance!
-                     signer: (get signer existing-entry) })
+                     signer: signer-key })
```
jcnelson, line 937: You need to check that this is the same signer key in `existing-entry`, no?  `stack-increase` does this by way of `increase-reward-cycle-entry`.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
```diff
@@ -3282,6 +3283,323 @@ fn stack_agg_commit_verify_sig() {
...
+        &bob.private_key,
+        bob.nonce,
+        &bob.pox_address,
+        next_reward_cycle,
+        bob_aggregate_commit_reward_index_actual
+            .clone()
+            .expect_u128()
+            .unwrap(),
+        Some(alice_signature_increase),
+        &alice.public_key,
```
jcnelson, line 3570: You should add a check here to verify that passing a different signing key other than what was given in the initial `stack-aggregation-commit` will fail.

jcnelson (changes requested):
I think you still have one more bug.

stackslib/src/chainstate/stacks/boot/pox-4.clar
kantai: Ah-- good catch. I thought allowing this to change was okay, but actually, yes, we must enforce equality here, because increasing an existing entry isn't otherwise auth-checked: anyone can choose to increase a given entry. But they shouldn't be allowed to change the signer key.

setzeus: Copy. Changing this accordingly.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
setzeus: Test now included.

stackslib/src/chainstate/stacks/boot/pox-4.clar
setzeus: Changed.

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
setzeus: Pinging to pair to review this real quick when you have a few mins, all other issues addressed.

kantai (approved):

stackslib/src/chainstate/stacks/boot/pox_4_tests.rs
obycode: It's not a big deal, so it doesn't have to hold up this PR, I just don't think it's doing anything useful, but I could be missing something. I'm available to discuss.

setzeus: Removed accordingly 👍

obycode (dismissed):
🎉

obycode (approved):

kantai (approved):

stackslib/src/chainstate/stacks/boot/pox-4.clar
```diff
@@ -1433,6 +1437,9 @@
                                              (max-amount uint)
                                              (auth-id uint))
   (begin
+    ;; must be called directly by the tx-sender or by an allowed contract-caller
+    (asserts! (check-caller-allowed)
```
jcnelson, line 1441: Great catch!

jcnelson (approved):

Issue #4638, opened 2024-04-04
cargo target features not being applied correctly when cross-compiling


wileyj:
when trying to build for arm64 on an x64 cpu using something like this (this is the build for macos-arm64, but is applicable for any arm64 build):
```
STACKS_NODE_VERSION="No Version Info"
GIT_BRANCH='No Branch Info'
GIT_COMMIT='No Commit Info'
BUILD_DIR=/build
TARGET=aarch64-unknown-linux-gnu

apt-get update && apt-get install -y git gcc-aarch64-linux-gnu libclang-dev llvm
git clone --single-branch https://github.com/stacks-network/stacks-core -b next ${BUILD_DIR}
cd ${BUILD_DIR}
rustup target add ${TARGET}

echo "RUSTFLAGS: ${RUSTFLAGS}"

CC=aarch64-linux-gnu-gcc \
CC_aarch64_unknown_linux_gnu=aarch64-linux-gnu-gcc \
CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc \
cargo build --features monitoring_prom,slog_json,portable --release --workspace --target ${TARGET}
```

this is the resulting output:
```
...
'+fxsr' is not a recognized feature for this target (ignoring feature)
'-avx512dq' is not a recognized feature for this target (ignoring feature)
'-sse4a' is not a recognized feature for this target (ignoring feature)
'goldmont' is not a recognized processor for this target (ignoring processor)
^C  Building [=>                        ] 54/480: memchr, libc, ahash, futures-channel, rustc_version, syn, aho-corasick, syn, rustix(build.rs), quote, memchr, cc, standback(build), proc-macro2, ryu, tracing-core, typenum, hashbrown, regex-syntax, libc           
```

in .cargo/config, there was a change that allowed this work:
https://github.com/stacks-network/stacks-core/pull/4489/files#diff-f6009bd0d260464389ace37ab2f89adae993e1fa4a47f779e4c9859937005cedR10

but this was changed in https://github.com/stacks-network/stacks-core/commit/8ff05eb946f82b220aa6b78eb1c654e22295e353

the current directive *should* work, but it's clearly not being applied. confirmed that revertnig the change suppresses the warnings. 

becuase of this, the output of some builds is incredibly noisy - but the only build that will fail outright (for stacks-signer) is the cross-compiled macos-arm64 build

zone117x:
From the slack chat:

[...]

PR #4646, opened 2024-04-05
remove portable flag

wileyj:
ref: #4638 

the options in .cargo/config are not being applied, this PR removes them entirely and adds some text in the README on how to enable building for `native-cpu`. 

removing the portable flag will have some downstream effects to the release build process that we'll need to address once merged (basically removing the portable feature flag, and re-enabling macos-arm64 stacks-signer binary builds).

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -2,6 +2,11 @@
 stacks-node = "run --package stacks-node --"
 fmt-stacks = "fmt -- --config group_imports=StdExternalCrate,imports_granularity=Module"
 
+# Uncomment to improve performance slightly, at the cost of portability
+#   * Note that native binaries may not run on CPUs that are different from the build machine
+# [build]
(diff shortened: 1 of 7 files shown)


obycode (approved):
This is strange that this doesn't work, but the reversal looks good to me.

wileyj:
> This is strange that this doesn't work, but the reversal looks good to me.

it's mentioned in the ref ticket #4638  - it was silently being ignored and the docs aren't very up front about it.  
apparently setting rustflags via that method is not supported in cargo (thanks to @zone117x pointing that out)

kantai (approved):
This LGTM, but do you want this in `develop` rather than `next`?

wileyj:
> This LGTM, but do you want this in `develop` rather than `next`?

good catch - i think it should be fine, since this is mostly needed for release builds (and should make it's way to develop soon anyways)

PR #4651, opened 2024-04-06
Add stack-aggregation-increase command

BowTiedRadone:
This PR adds the `stack-aggregation-commit` command, using both signature and authorization, to the stateful property testing environment. It is part of #4548 and targets `feat/pox-4-stateful-property-testing` (#4550).


BowTiedRadone:
@friedger @moodmosaic @setzeus adding the `stack-aggregation-increase` brought up some intermittent test failures. I have saved the error traces (that I'll continue investigating after the command is merged).  

In case the command is correct, there may be some **PoX-4 unexpected behavior**.

Thank you in advance!

BowTiedRadone:
Closed this PR in favor of #4652

Issue #4648, opened 2024-04-06
[stacks-signer] Audit of the code has revealed that signer does not limit its read size when retrieving from stackerb


jferrant:
This is simply an extra precaution. Though stackerdb is configured through the node itself and places a limit on who and what amount can be written to a stackerdb slot, we should have signers also employ sensible limits on the amount of data it reads from stackerdb to ensure we do not have a single point of failure.

PR #4656, opened 2024-04-08
feat: limit length of chunks when signer reads from stackerdb

hstove:
- Closes https://github.com/stacks-network/stacks-core/issues/4648

This uses the `CHUNK_SIZE` constant, which is set in `signers.clar` and used as part of its StackerDB config, when reading chunks from the `stacks-signer`. This maximum chunk length is already enforced on the node, but this PR adds a some redundancy on that maximum length check.

### DIFF
--- a/libsigner/src/session.rs
+++ b/libsigner/src/session.rs
@@ -218,7 +218,8 @@ impl SignerSession for StackerDBSession {
         let limit = if self.stackerdb_contract_id.name.starts_with("signer") {
             SIGNERS_STACKERDB_CHUNK_SIZE
         } else {
-            STACKERDB_MAX_CHUNK_SIZE as usize
+            usize::try_from(STACKERDB_MAX_CHUNK_SIZE)
+                .expect("infallible: StackerDB chunk size exceeds usize::MAX")
         };
         for slot_id in slot_ids.iter() {
             let path = stackerdb_get_chunk_path(self.stackerdb_contract_id.clone(), *slot_id, None);
--- a/libsigner/src/session.rs
+++ b/libsigner/src/session.rs
@@ -17,11 +17,11 @@
 use std::net::{SocketAddr, TcpStream};
 use std::str;
 
-use blockstack_lib::net::stackerdb::SIGNERS_STACKERDB_CHUNK_SIZE;
 use clarity::vm::types::QualifiedContractIdentifier;
 use libstackerdb::{
     stackerdb_get_chunk_path, stackerdb_get_metadata_path, stackerdb_post_chunk_path, SlotMetadata,
-    StackerDBChunkAckData, StackerDBChunkData,
+    StackerDBChunkAckData, StackerDBChunkData, SIGNERS_STACKERDB_CHUNK_SIZE,
+    STACKERDB_MAX_CHUNK_SIZE,
 };
 use stacks_common::codec::StacksMessageCodec;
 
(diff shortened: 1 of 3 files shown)


jferrant (approved):

stackslib/src/net/stackerdb/mod.rs
```diff
@@ -148,6 +148,8 @@ pub const STACKERDB_INV_MAX: u32 = 4096;
 pub const STACKERDB_PAGE_LIST_MAX: u32 = 4096;
 /// maximum number of pages that can be used in a StackerDB contract
 pub const STACKERDB_MAX_PAGE_COUNT: u32 = 2;
+/// CHUNK_SIZE constant for signers StackerDBs
+pub const SIGNERS_STACKERDB_CHUNK_SIZE: u32 = 2 * 1024 * 1024; // 2MB
```
jcnelson, line 152: This belongs in `libsigner`.  `libstackerdb` has a 16MB max

libsigner/src/session.rs
```diff
@@ -217,7 +218,16 @@ impl SignerSession for StackerDBSession {
         for slot_id in slot_ids.iter() {
             let path = stackerdb_get_chunk_path(self.stackerdb_contract_id.clone(), *slot_id, None);
             let chunk = match self.rpc_request("GET", &path, None, &[]) {
-                Ok(body_bytes) => Some(body_bytes),
+                Ok(body_bytes) => {
+                    // Verify that the chunk is not too large
+                    if body_bytes.len() > u32::MAX as usize {
+                        None
+                    } else if body_bytes.len() as u32 > SIGNERS_STACKERDB_CHUNK_SIZE {
```
jcnelson, line 225: `SIGNERS_STACKERDB_CHUNK_SIZE` is vastly smaller than `u32::MAX`, so let's just use that for comparison

stackslib/src/net/stackerdb/mod.rs
jcnelson: Also, you can (and should) make this a `usize` so you can avoid a risky cast to `usize`.

jferrant:
This will have to get reopened against develop . see https://nakamotoslack.slack.com/archives/C067BHSN1SL/p1712679342282899

hstove:
I've rebased my original commit off of develop and changed the base branch. I've also addressed some review comments.

stackslib/src/net/stackerdb/mod.rs
hstove: I've moved this specific max chunk size to `libsigner`. 

I haven't changed the existing constant in that file, because this const is really for limiting reads of chunk sizes for the signers contract, which has a lower max chunk size than other StackerDBs

libsigner/src/session.rs
hstove: I'll admit that I don't fully follow the "let's just use that for comparison" phrase, i.e. what do you mean by _that_ exactly? I'm not familiar with some of the potential gotchas relating to these casting/comparison functions in Rust, so that may be what I'm missing.

jcnelson: Use the smaller of the two values here for checking that `body_bytes.len()` small enough.  I.e., only compare `body_bytes.len()` to `SIGNERS_STACKERDB_CHUNK_SIZE`.  You ought to be able to safely cast `SIGNERS_STACKERDB_CHUNK_SIZE` to `usize` since it's pretty small compared to what `usize::MAX` will be on all supported platforms (which is the maximum size of a memory address IIRC).

libsigner/src/session.rs
```diff
@@ -214,10 +215,22 @@ impl SignerSession for StackerDBSession {
     /// query the replica for zero or more latest chunks
     fn get_latest_chunks(&mut self, slot_ids: &[u32]) -> Result<Vec<Option<Vec<u8>>>, RPCError> {
         let mut payloads = vec![];
+        let limit = if self.stackerdb_contract_id.name.starts_with("signer") {
+            SIGNERS_STACKERDB_CHUNK_SIZE
+        } else {
+            STACKERDB_MAX_CHUNK_SIZE as usize
```
jcnelson, line 221: minor nit: use `usize::try_from(STACKERDB_MAX_CHUNK_SIZE).expect("infallible: StackerDB chunk size exceeds usize::MAX");`

jcnelson (dismissed):
Just a couple of nits to address before merging but otherwise LGTM

jferrant (approved):

hstove:
@jferrant @jcnelson sorry to request re-review. Just made on small change based on Jude's last comment

libsigner/src/session.rs
```diff
@@ -214,10 +215,23 @@ impl SignerSession for StackerDBSession {
     /// query the replica for zero or more latest chunks
     fn get_latest_chunks(&mut self, slot_ids: &[u32]) -> Result<Vec<Option<Vec<u8>>>, RPCError> {
         let mut payloads = vec![];
+        let limit = if self.stackerdb_contract_id.name.starts_with("signer") {
+            SIGNERS_STACKERDB_CHUNK_SIZE
```
jferrant, line 219: Ah I just noticed this. We should be putting a limit on the miner contract ids as well. I think there is already a global defined max for the miner messages and we should use that when reading miner messages as well.

jferrant (approved):

kantai (approved):

Issue #4636, opened 2024-04-03
[stacks-signer] Update signers to query stackerdb for DKG messages to prevent DKG stalls due to timing issues


jferrant:
Suppose we have three signers, signer-0, signer-1, and signer-2.

Signer-0 and signer-1  receive burn block X which happens to be in the prepare phase of the next reward cycle, but it happens to be right in the middle of their nodes calculating the registration set so they do not see themselves as registered.
Signer-2 gets burn block X, but the milliseconds later is all that is needed for the node to finish the registration calculations. so it finds itself as registered AND that it is the coordinator. So it triggers DKGPublicGather sendings its public shares to other signers.
Signer-0 and Signer-1 receive this public share but because they ONLY check for a refresh once a burn block arrives, it ignores this event until burn block event X+1 shows up and THEN it queries for the registered signers. However, it never responds to the initial coordiantor message as it was ignored when it first arrived since they did not think themselves to be reigstered.

Fix this.

Issue #4654, opened 2024-04-08
Nakamoto Signer[3.0] -  A signer must be able to save multiple DKG aggregate key round states and load it where appropriate


jferrant:
See: https://github.com/stacks-network/stacks-core/blob/5d2cbdd1ab6bd6cfde8f4528eb5fae5777037af5/stacks-signer/src/signer.rs#L1372

If signers see that a threshold has been reached and no approved key set, they may trigger a new DKG round. However, the key could have been approved but not yet set in the contract and this new round should be aborted once they read that it is actually set in the contract. However, they must also load the appropriate DKG shares associated with this successful round to successfully participate.

As part of this ticket, store DKG private shares per round. 
If DKG is set in the contract, abort any ongoing rounds and update state.
- Overwrite local shares with the ones corresponding to the winning round (also be sure to update current_dkg_id to match)
- If the signer has no shares for that DKG round, print a warning that it cannot participate in signing asit did not participate in the winning DKG round and ignore any incoming sign requests

netrome:
Implementation note: We need to ensure the structure we store in StackerDB has a bounded size as discussed here https://github.com/stacks-network/stacks-core/pull/4659#discussion_r1560353035

netrome:
> If signers see that a threshold has been reached and no approved key set, they may trigger a new DKG round. However, the key could have been approved but not yet set in the contract.

So I'm working on an integration test for this, and I suspect that this scenario is not attainable in the current implementation. If the threshold has been reached, the vote has been concluded and the approved key must be set in the contract. This is happening in the same transaction, so there shouldn't exist an intermediate inconsistent state as I see it. According to the docs for `vote-for-aggregate-public-key`.

Has this scenario been observed in practice? Am I missing something here? Ping @jferrant 

That being said, I still think the feature of storing a few previous aggregate keys makes sense. There could potentially be other scenarios where a redundant DKG round is triggered. I'll see if I can figure out any good scenario to test.

netrome:
Closing this ticket given that the premise has been invalidated, and recent discussions have lead to a different direction in the signer development.

saralab:
We deprioritized this for a later time, is this still needed for when we kick off the WSTS work?

PR #4658, opened 2024-04-08
Feat/signers read stackerdb

jferrant:
Closes https://github.com/stacks-network/stacks-core/issues/4636

Attempted to do for sign as well, but this would require substantial changes to the way a miner stores its messages as we require every single sign message to replicate state and miners overwrite their own messages as they use the same slot for everything. Sign rounds are also quite quick and a miner already has timeout logic in place for triggering a new sign request, so this is less of an issue anyway.

### DIFF
--- a/.github/workflows/bitcoin-tests.yml
+++ b/.github/workflows/bitcoin-tests.yml
@@ -82,12 +82,14 @@ jobs:
           - tests::nakamoto_integrations::vote_for_aggregate_key_burn_op
           - tests::nakamoto_integrations::follower_bootup
           - tests::signer::stackerdb_dkg
-          - tests::signer::stackerdb_sign
+          - tests::signer::stackerdb_sign_request_rejected
           - tests::signer::stackerdb_block_proposal
           - tests::signer::stackerdb_filter_bad_transactions
           - tests::signer::stackerdb_mine_2_nakamoto_reward_cycles
           - tests::signer::stackerdb_sign_after_signer_reboot
           - tests::nakamoto_integrations::stack_stx_burn_op_integration_test
+          - tests::signer::stackerdb_delayed_dkg
+          - tests::signer::stackerdb_delayed_sign
           # Do not run this one until we figure out why it fails in CI
           # - tests::neon_integrations::bitcoin_reorg_flap
     steps:
--- a/.github/workflows/bitcoin-tests.yml
+++ b/.github/workflows/bitcoin-tests.yml
@@ -81,6 +81,7 @@ jobs:
           - tests::nakamoto_integrations::correct_burn_outs
           - tests::nakamoto_integrations::vote_for_aggregate_key_burn_op
           - tests::nakamoto_integrations::follower_bootup
+          - tests::nakamoto_integrations::forked_tenure_is_ignored
           - tests::signer::stackerdb_dkg
           - tests::signer::stackerdb_sign_request_rejected
           - tests::signer::stackerdb_block_proposal
--- a/.github/workflows/bitcoin-tests.yml
+++ b/.github/workflows/bitcoin-tests.yml
@@ -89,7 +89,6 @@ jobs:
           - tests::signer::stackerdb_sign_after_signer_reboot
           - tests::nakamoto_integrations::stack_stx_burn_op_integration_test
           - tests::signer::stackerdb_delayed_dkg
-          - tests::signer::stackerdb_delayed_sign
           # Do not run this one until we figure out why it fails in CI
           # - tests::neon_integrations::bitcoin_reorg_flap
     steps:
--- a/stacks-signer/src/signer.rs
+++ b/stacks-signer/src/signer.rs
@@ -126,6 +126,8 @@ pub enum Command {
 /// The Signer state
 #[derive(PartialEq, Eq, Debug, Clone)]
 pub enum State {
+    /// The signer is uninitialized and should read stackerdb to restore state
+    Uninitialized,
     /// The signer is idle, waiting for messages and commands
     Idle,
     /// The signer is executing a DKG or Sign round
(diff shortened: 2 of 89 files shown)


.github/workflows/bitcoin-tests.yml
```diff
@@ -82,12 +82,13 @@ jobs:
           - tests::nakamoto_integrations::vote_for_aggregate_key_burn_op
           - tests::nakamoto_integrations::follower_bootup
           - tests::signer::stackerdb_dkg
-          - tests::signer::stackerdb_sign
+          - tests::signer::stackerdb_sign_request_rejected
```
jferrant, line 85: Note: this is just a rename as the test does not actually perform sign anymore (changed in a different PR)

kantai (approved):
This code LGTM! I'm curious about the slow test, though. It doesn't seem like restarting the signer should require re-syncing the stacks-node.

jferrant:
> This code LGTM! I'm curious about the slow test, though. It doesn't seem like restarting the signer should require re-syncing the stacks-node.

Yeah, I am not 100% sure why this is the case. I actually can't merge this until https://github.com/stacks-network/stacks-core/issues/4654 is done. So will keep this as is for now and will look into it further when that PR is completed.

stacks-signer/src/signer.rs
```diff
@@ -1380,8 +1455,27 @@ impl Signer {
...
                 );
             }
+            if matches!(self.state, State::OperationInProgress(Operation::Dkg)) {
+                // We already have DKG, abort operation and reset state.
+                debug!(
+                    "{self}: DKG has already been set. Aborting DKG round {}.",
+                    self.coordinator.current_dkg_id
+                );
+                self.finish_operation();
+            }
```
hstove, line 1465: Can you explain this code a little bit? I have no reason to doubt it, I'm just not totally following this code path

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -1098,6 +1110,107 @@ fn stackerdb_sign() {
...
+    );
+    // Wait a bit so DKG is actually triggered and signers are not available to respond
+    std::thread::sleep(Duration::from_secs(5));
+
+    // Make sure DKG did not get set
+    assert!(signer_test
+        .stacks_client
+        .get_approved_aggregate_key(reward_cycle)
+        .expect("Failed to get approved aggregate key")
+        .is_none());
```
hstove, line 1184: Won't this never happen, even if you didn't stop a signer, because you haven't advanced the Stacks tip? So the `vote-for-aggregate-key` votes wouldn't be confirmed here?

Also because, if I'm reading this right, we're still in 2.5, so there wouldn't be Nakamoto blocks.

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -1098,6 +1110,107 @@ fn stackerdb_sign() {
     info!("Sign Time Elapsed: {:.2?}", sign_elapsed);
 }
 
+#[test]
+#[ignore]
+/// Test that a signer can be offline when a DKG round has commenced and
+/// can rejoin the DKG round after it has restarted
+fn stackerdb_delayed_dkg() {
```
hstove, line 1117: This is a really good test!

hstove (approved):
I know it's in draft, but I had already started a review, so I figured I'd just finish it.

stacks-signer/src/signer.rs
jferrant: Yeah this is because I was working on a different PR simultaneously (see https://github.com/stacks-network/stacks-core/pull/4660) It is one of the reasons I converted it to a draft so I can rebase against it once merged :) 

Basically this is to handle the fact that we could trigger a DKG round due to timing issues where we might have an outdated view (for example, exceeded the threshold check but still not gotten the approved key though the other PR will mitigate this). If this happens and then we later discover that DKG was triggered and it shouldn't have been, we should abort our attempt. This is incomplete however until we can then load the appropriate DKG shares from storage for the winning round. Hence the link to Marten's issue. So I have put this ticket a bit on pause until I can build off these two other PRs.

testnet/stacks-node/src/tests/signer.rs
jferrant: Oh good point. I should mine one block before to prove that DKG did not proceed. Will add. And yes this is 2.5. I don't make any references to Nakamoto blocks in this test for that reason.

hstove (dismissed):

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -1108,6 +1121,167 @@ fn stackerdb_sign() {
     info!("Sign Time Elapsed: {:.2?}", sign_elapsed);
 }
 
+#[test]
+#[ignore]
+/// Test that a signer can be offline when a DKG round has commenced and
```
xoloki, line 1126: It's so exciting that this works now!!!

xoloki (approved):
Looks great! So exciting to get this in, we've talked about it forever.

stacks-signer/src/signer.rs
```diff
@@ -1316,42 +1361,78 @@ impl Signer {
...
         self.approved_aggregate_public_key =
             stacks_client.get_approved_aggregate_key(self.reward_cycle)?;
         if self.approved_aggregate_public_key.is_some() {
             // TODO: this will never work as is. We need to have stored our party shares on the side etc for this particular aggregate key.
             // Need to update state to store the necessary info, check against it to see if we have participated in the winning round and
             // then overwrite our value accordingly. Otherwise, we will be locked out of the round and should not participate.
+            let internal_dkg = self.coordinator.aggregate_public_key;
+            if internal_dkg != self.approved_aggregate_public_key {
+                warn!("{self}: we do not support changing the internal DKG key yet. Expected {internal_dkg:?} got {:?}", self.approved_aggregate_public_key);
+            }
```
hstove, line 1412: Is this TODO still relevant? From my read it seems like it is?

In that case, does the later `finish_operation` command essentially prevent the signer from voting on that key?

hstove: Also, if my read is right, I think the `warn` when updating DKG should be an `error` with something more explicit (like "didn't participate in DKG, unable to sign during this reward cycle")

hstove (approved):
Nice job. A little nit regarding comments / logs, and that may be to me misunderstanding things. It shouldn't be blocking a merge, but a few more comments around why this state happened in refresh/update DKG would be help.

Also, I was going to comment "nice job on that test" but I already did 😂

stacks-signer/src/signer.rs
```diff
@@ -234,6 +236,43 @@ impl Signer {
...
+        let ordered_packets = self
+            .stackerdb
+            .get_dkg_packets(&self.signer_slot_ids)?
+            .iter()
+            .filter_map(|packet| {
+                let coordinator_pubkey = if Self::is_dkg_message(&packet.msg) {
+                    self.get_coordinator_dkg().1
+                } else {
+                    debug!(
+                        "{self}: Received a non-DKG message in the DKG message queue. Ignoring it."
```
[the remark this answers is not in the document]
jferrant: It shouldn't unless we have a malicious or buggy signer.

stacks-signer/src/signer.rs
```diff
@@ -1316,42 +1361,78 @@ impl Signer {
...
-        } else if should_queue {
-            if self.commands.front() != Some(&Command::Dkg) {
-                info!("{self} is the current coordinator and must trigger DKG. Queuing DKG command...");
-                self.commands.push_front(Command::Dkg);
-            } else {
-                debug!("{self}: DKG command already queued...");
+            match self.state {
+                State::OperationInProgress(Operation::Dkg) => {
+                    debug!(
+                        "{self}: DKG has already been set. Aborting DKG operation {}.",
```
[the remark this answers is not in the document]
jferrant: No this message is not stale. You can set the DKG key from the contract while in the middle of a round and in which case you should abort that round. (this happens if a coordinator thinks DKG round has failed and a new one gets triggered)

jcnelson (approved):

PR #4659, opened 2024-04-08
feat: Signer using StackerDB for persisting DKG shares

### LINKED ISSUE #4595: [Nakamoto] Stacks signer should store its DKG shares in StackerDB to enable disaster recovery
In #4570, we introduced persistence for the signer to enable it to recover from restarts by storing their DKG shares in the signers own SignerDB. However, concerns were raised that in catastrophic scenarios the Signers may lose the contents of their SignerDB. Since DKG shares are connected to the liveliness of the chain, we want to take every measure possible to ensure the durability of these.

Therefore, it was proposed to have the signer encrypt their DKG shares and utilize StackerDB as a replicated durable storage for these.

### Definition of Done
A running signer can recover from a catastrophic restart (losing their SignerDB data in addition to all in-memory state), and continue signing blocks in the same signing round.

netrome:
### Description
This PR follows up https://github.com/stacks-network/stacks-core/pull/4570 with the discussed change to use StackerDB instead of SignerDB to persist DKG shares for signers.

### Applicable issues

- closes #4595


xoloki (dismissed):
LGTM

jferrant (approved):
This def replaces existing functionality so I approve. But something that is missing that will need to be added (should do in a follow on PR for https://github.com/stacks-network/stacks-core/issues/4654) is the need to store more than one encrypted signer state. It should probably a hashmap of round id to encrypted signer state. The reason for this is that multiple DKG rounds can get triggered (generally due to timeout/differing views of the system). The signer should be able to load their party shares for the winning round which could be an old round rather than a newly triggered one.

jferrant:
This will have to get reopened against develop . See https://nakamotoslack.slack.com/archives/C067BHSN1SL/p1712679342282899

netrome:
> This will have to get reopened against develop . See https://nakamotoslack.slack.com/archives/C067BHSN1SL/p1712679342282899

Right, thanks for the pointer. I'll retarget once next has been merged in https://github.com/stacks-network/stacks-core/pull/4666

libsigner/Cargo.toml
```diff
@@ -20,6 +20,7 @@ clarity = { path = "../clarity" }
 hashbrown = { workspace = true }
 libc = "0.2"
 libstackerdb = { path = "../libstackerdb" }
+mutants = "0.0.3"
```
jcnelson, line 23: This should be a dev-dependency

libsigner/src/messages.rs
```diff
@@ -383,6 +402,10 @@ impl StacksMessageCodec for SignerMessage {
                     party_polynomials,
                 }
             }
+            SignerMessageTypePrefix::EncryptedSignerState => {
+                let encrypted_state = read_next::<_, _>(fd)?;
```
jcnelson, line 406: There needs to be a `BoundReader` wrapping `fd` so as to impose a maximum amount of bytes that will be read from it.  Do we know how big this blob of data can be?  Can it fit in a StackerDB slot?

stacks-signer/Cargo.toml
```diff
@@ -31,6 +31,7 @@ reqwest = { version = "0.11.22", default-features = false, features = ["blocking
 serde = "1"
 serde_derive = "1"
 serde_stacker = "0.1"
+sha2 = "0.10"
```
jcnelson, line 34: Can you please instead use the hashing algorithms in `stacks_common`?  They pull in `sha2` as a transitive dependency as it is.

stacks-signer/src/client/stacks_client.rs
```diff
@@ -1040,7 +1040,8 @@ mod tests {
...
+            "0x070b000000050c00000002096e756d2d736c6f7473010000000000000000000000000000000e067369676e6572051a8195196a9a7cf9c37cb13e1ed69a7bc047a84e050c00000002096e756d2d736c6f7473010000000000000000000000000000000e067369676e6572051a6505471146dcf722f0580911183f28bef30a8a890c00000002096e756d2d736c6f7473010000000000000000000000000000000e067369676e6572051a1d7f8e3936e5da5f32982cc47f31d7df9fb1b38a0c00000002096e756d2d736c6f7473010000000000000000000000000000000e067369676e6572051a126d1a814313c952e34c7840acec9211e1727fb80c00000002096e756d2d736c6f7473010000000000000000000000000000000e067369676e6572051a7374ea6bb39f2e8d3d334d62b9f302a977de339a";
```
jcnelson, line 1043: What's the delta here?

stacks-signer/src/signer.rs
```diff
@@ -28,7 +28,9 @@ use libsigner::{
     BlockProposalSigners, BlockRejection, BlockResponse, MessageSlotID, RejectCode, SignerEvent,
     SignerMessage,
 };
+use rand_core::OsRng;
```
jcnelson, line 31: Why `OsRng`?  What's wrong with `rand::RngCore` etc.?

stacks-signer/src/signer.rs
```diff
@@ -1232,12 +1244,19 @@ impl Signer {
     /// DKG and participate in signing rounds accross crashes
     ///
     /// # Panics
-    /// Panics if the insertion fails
-    fn save_signer_state(&self) {
+    /// Panics if the encryption fails or if the insertion into stackerDB fails
```
jcnelson, line 1247: Neither of these should be panicking failures

stacks-signer/src/signer.rs
```diff
@@ -1453,3 +1472,41 @@ impl Signer {
...
+        .expect("Failed to encrypt message")
+}
+
+fn decrypt(private_key: &Scalar, encrypted_msg: &[u8]) -> Vec<u8> {
+    wsts::util::decrypt(&derive_encryption_key(private_key), encrypted_msg)
+        .expect("Failed to decrypt message")
+}
+
+fn derive_encryption_key(private_key: &Scalar) -> [u8; 32] {
+    let mut hasher = Sha256::new();
```
jcnelson, line 1487: You can safely use `Sha512Trunc256Sum` here from `stacks_common`.

stacks-signer/src/signerdb.rs
```diff
@@ -83,31 +72,6 @@ impl SignerDb {
         )
     }
 
-    /// Get the signer state for the provided reward cycle if it exists in the database
-    pub fn get_signer_state(&self, reward_cycle: u64) -> Result<Option<SignerState>, DBError> {
```
jcnelson, line 87: I think it's fine to keep a variation of this -- you'd just store your encrypted state to the DB instead of the cleartext state.  However, it should be something performed _in addition to_ saving to StackerDB.

jcnelson: To be clear, I stand by my initial claim that storing locally is redundant if you're storing to StackerDB.  However, a lack of signer state availability is such a catastrophic failure condition that keeping a copy around locally isn't the worst idea, especially for operators who run their Stacks nodes in different hosts (or even different networks), which can go offline.  Also, if there's a network-wide Stacks node crash, then keeping the signer state locally can at least provide one additional avenue for the signers to boot back up.

stacks-signer/src/signer.rs
```diff
@@ -1453,3 +1472,41 @@ impl Signer {
         Ok(())
     }
 }
+
+fn encrypt(private_key: &Scalar, msg: &[u8]) -> Vec<u8> {
+    wsts::util::encrypt(&derive_encryption_key(private_key), msg, &mut OsRng)
```
jcnelson, line 1477: I'm not sure that either this method nor `decrypt()` should panic?  I think they should instead return a `Result`.  Can you describe the conditions under which they'd panic?

libsigner/src/messages.rs
xoloki: Polynomial size is threshold*32 bytes, key shares are |shares|*32 bytes.  So worst case scenario is like 32MB, for 4k total reward slots and 250 owned key_ids.

Is there some hard max size for a slot, or a performed cliff?

stacks-signer/src/signer.rs
jcnelson: Also, I see that you're prepending a topic string to the cleartext.  Perhaps that topic string should be exposed in the `encrypt()` and `decrypt()` interfaces?  Then, `decrypt()` can detect bad decryption if the topic string is absent from the decrypted cleartext and return an appropriate "bad key" error.

libsigner/src/messages.rs
jcnelson: Yikes, that's too big.  Network messages max out at 2MB.

jcnelson: Hmm, this may require some re-thinking.  Unfortunately, we can't just increase the maximum network message size, because that's the only limit stopping blocks from getting too big (meaning, it'd be highly disruptive to change this without some kind of gating).  Splitting the state into multiple chunks isn't too appealing either, because then there needs to be extra logic for that (and there'd need to be 16 more StackerDBs just for holding each signer's state).

xoloki: LoL sorry that should be `(2800+250)*32` bytes, which is much more reasonable.

jcnelson: Oh great, that's only ~180k

netrome: Great. Note that this will be extended in https://github.com/stacks-network/stacks-core/issues/4654 though, but we can probably limit it to a factor two so we should still have a good margin. I'll make a note in that ticket.

I'll update to use a `BoundReader`.

stacks-signer/Cargo.toml
netrome: Sure thing, I'll update.

stacks-signer/src/client/stacks_client.rs
netrome: Changing a hard-coded 13 to a 14 for `num-slots` in a type signature. Though I don't like that we use a hard-coded hex here because it obfuscates what's going on, and hex deserialization is not the purpose of this test.

I'll expand this into Rust code so that

1. We can use the appropriate constant instead of hard-coding a value and
2. Future deltas becomes easier to read.

stacks-signer/src/signer.rs
netrome: Just to be cautious. I'm not an expert in which random number generators are cryptographically secure or not, and `OsRng` is the one used in the existing `wsts` and Signer code in `dkg_private_begin` among other places. So I figured that it's best to be consistent and not introduce any other rng.

I can definitely make the encryption logic generic with a `rand::RngCore` trait bound, or use any other rng if you have any concerns about `OsRng` in particular.

Let me know what you think.

stacks-signer/src/signer.rs
netrome: Nice point, I'll do that. I'm also happy to return a result rather than panicking. I'll improve the error handling here.

stacks-signer/src/signer.rs
netrome: Thanks, will do.

stacks-signer/src/signer.rs
xoloki: Anywhere `wsts` is hard coding a specific `RNG` is a bug, please file a ticket

stacks-signer/src/signer.rs
netrome: Agreed. These can only fail if the underlying aes-gcm implementation fails. This seems to only fail in the current implementation if the encrypted message is larger than 1 << 36 = 68719476736. So in our case this is infallible, though it's probably safer to return an error in case the library gets updated and the underlying implementation somehow would turn fallible.

I'll change to return an error if these fail.

stacks-signer/src/signer.rs
netrome: Okay, will do. This will entail changes in the public interfaces though.

netrome: https://github.com/Trust-Machines/wsts/issues/63

libsigner/src/messages.rs
netrome: Done. I added a factor 4 to the bound to have margin for the other fields in the saved state beyond the dkg shares. Let me know if you prefer to have the bound be more exact or if you want another margin.

So the bound right now is (2800 + 2800) * 32 * 4.

libsigner/Cargo.toml
netrome: Thanks for spotting, updated.

stacks-signer/Cargo.toml
netrome: Done.

stacks-signer/src/signer.rs
netrome: Updated to return a result instead of panicking. I previously misunderstood the topic string suggestion. I'm only prefixing the private key when hashing to derive the encryption key. The message is encrypted as-is without any topic string. I can add one if you want, but that would make the persisted state in StackerDB larger.

stacks-signer/src/signer.rs
netrome: Done.

stacks-signer/src/signerdb.rs
netrome: Updated. The logic is now:

Persistence:
1. Attempt to persist in both StackerDB and in the local sqlite database.
2. Return both results to the caller, which will now panic if either operation fails (this was panicking failure before as well, but I'd be open to explore if we'd like the signer to continue if one of the storage mechanisms work)

Loading:
Attempt to load the state from stackerdb first, write a warning if this fails and secondly attempt to load the state from the local database. Panic if both fails.

Let me know if you want the logic to be different or have any thoughts on how it is implemented.

stacks-signer/src/signer.rs
netrome: Updated to return results.

xoloki (approved):
LGTM

libsigner/src/messages.rs
```diff
@@ -403,7 +404,12 @@ impl StacksMessageCodec for SignerMessage {
                 }
             }
             SignerMessageTypePrefix::EncryptedSignerState => {
-                let encrypted_state = read_next::<_, _>(fd)?;
+                // Typically the size of the signer state is much smaller, but in the fully degenerate case the size of dkg shares is
+                // (2800 + 2800) * 32 (threshold * shares * 32).
+                // The signer state holds some additional information, so we're adding a factor 4 to have some margin
```
jcnelson, line 409: Is `* 4` enough margin?

stacks-signer/src/signer.rs
```diff
@@ -279,19 +279,13 @@ impl From<SignerConfig> for Signer {
...
-                "Reward cycle #{} Signer #{}: Loading signer",
-                signer_config.reward_cycle, signer_config.signer_id
-            );
-            state_machine.signer = v2::Signer::load(&state);
-        }
+        if let Some(state) = load_encrypted_signer_state(
+            &mut stackerdb,
+            signer_config.signer_slot_id.into(),
+            &state_machine.network_private_key,
+        ).or_else(|err| {warn!("Failed to load encrypted signer state from StackerDB, falling back to SignerDB: {err}"); load_encrypted_signer_state(&signer_db, signer_config.reward_cycle, &state_machine.network_private_key)}).expect("Failed to load encrypted signer state from both StackerDB and SignerDB") {
```
jcnelson, line 286: nit: formatting

stacks-signer/src/client/stackerdb.rs
```diff
@@ -243,6 +243,51 @@ impl StackerDB {
         Self::get_transactions(&mut self.next_transaction_session, signer_ids)
     }
 
+    /// Get the encrypted state for the given signer
+    pub fn get_encrypted_signer_state(
+        &mut self,
+        signer_id: SignerSlotID,
+    ) -> Result<Option<Vec<u8>>, ClientError> {
+        debug!("Getting the persisted encrypted state for signer {signer_id}");
+        let Some(state_session) = self
```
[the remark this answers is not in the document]
netrome: Yeah it's quite handy at times :)

libsigner/src/messages.rs
netrome: Yes, but looking at the saved state again it seems to be a bit more narrow than intended. As already established we have a polynomial of $t*32$ where $t=2800$ is the threshold, and key shares that gives another $t*32$ in the worst case. However, in addition to the shares we store their ID as a u32 twice (once with the shares in `PartyState` and once alone in `SignerState`) giving an additional term of $t*32*2$. This means that our state is of size $t*32*4 + C$ for some small constant $C$ (accounting for some additional fields in the `SignerState`, length prefixes etc.). This is slightly more than half the current margin.

While it is very unrealistic that we'd in practice get even close to this margin, I'd personally feel more comfortable with a larger margin.

What's the main purpose of using a bound reader here? Is it a security measure to prevent from certain dos attacks? In that case I suppose we could extend this margin to be much larger.

stacks-signer/src/signer.rs
netrome: Oh, I wonder why this didn't get formatted automatically :thinking: `cargo fmt-stacks` failed to get to this for some reason.

I've manually formatted this now. Thanks for spotting :pray:

stacks-signer/src/client/stacks_client.rs
netrome: Done.

stacks-signer/src/signer.rs
```diff
@@ -1453,3 +1506,116 @@ impl Signer {
...
+}
+
+trait SignerStateStorage {
+    type IdType;
+
+    fn get_encrypted_signer_state(
+        self,
+        signer_config: Self::IdType,
+    ) -> Result<Option<Vec<u8>>, PersistenceError>;
+}
```
hstove, line 1532: I like the way you designed and use this trait 👍

stacks-signer/src/signer.rs
```diff
@@ -1228,16 +1236,61 @@ impl Signer {
...
-    /// DKG and participate in signing rounds accross crashes
-    ///
-    /// # Panics
-    /// Panics if the insertion fails
-    fn save_signer_state(&self) {
+    /// Persist signer state in both SignerDB and StackerDB
+    fn save_signer_state(&mut self) -> [Result<(), PersistenceError>; 2] {
+        let rng = &mut OsRng;
+        let signerdb_result = self.save_signer_state_in_signerdb(rng);
+        let stackerdb_result = self.save_signer_state_in_stackerdb(rng);
```
hstove, line 1243: I'm not sure if the overhead is worth caring about, but both of these functions call `encrypt` along with `state_machine.signer.save()`. Maybe add `save_signer_state(encrypted)` to `SignerStateStorage`?

stacks-signer/src/signer.rs
```diff
@@ -1453,3 +1506,116 @@ impl Signer {
...
+    wsts::util::decrypt(derive_encryption_key(private_key).as_bytes(), encrypted_msg)
+        .map_err(|_| EncryptionError::Decrypt)
+}
+
+fn derive_encryption_key(private_key: &Scalar) -> Sha512Trunc256Sum {
+    let mut prefixed_key = "SIGNER_STATE_ENCRYPTION_KEY/".as_bytes().to_vec();
+    prefixed_key.extend_from_slice(&private_key.to_bytes());
+
+    Sha512Trunc256Sum::from_data(&prefixed_key)
+}
```
hstove, line 1574: This is the first time I've seen derived keys for encryption. I'm not knocking it at all, but what's the purpose of this? Some sort of attack where you can get the private key with a malicious payload?

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
```diff
@@ -407,6 +408,7 @@ impl SignCoordinator {
                 .filter_map(|msg| match msg {
                     SignerMessage::DkgResults { .. }
                     | SignerMessage::BlockResponse(_)
+                    | SignerMessage::EncryptedSignerState { .. }
```
hstove, line 411: Just curious, why not `SignerMessage::EncryptedSignerState(_)`? Just a preference thing?

hstove (approved):
LGTM! Just small questions and comments.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
netrome: Not a preference, just an artifact that I missed from when I developed this because I started it out with this as a struct variant. Will update, thanks for noting :pray:

stacks-signer/src/signer.rs
netrome: Yeah this is just for extra caution. While we could use the private key directly for encryption, this adds an extra layer in case someone manages to break the encryption as they won't be able to get the private key from this hash, since the private key is also used for other things like signing network packets.

stacks-signer/src/signer.rs
netrome: Good point, the computational overhead is probably negligible but eliminate code duplication makes maintenance easier so I'll factor this out.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
netrome: Updated.

stacks-signer/src/signer.rs
netrome: Changed.

stacks-signer/src/signer.rs
netrome: Thanks :blush:

libsigner/src/messages.rs
netrome: Extended the margin to $16\cdot32t$.

netrome:
Targeting develop for this PR now. I've addressed all comments so-far, so if everything looks good we should be ready to merge.

libsigner/src/messages.rs
netrome: Resolving this conversation. Any suggestions on alternative sizes of the margin should be in this thread: https://github.com/stacks-network/stacks-core/pull/4659#discussion_r1563121809

stacks-signer/src/signer.rs
jferrant: Seconded.

stacks-signer/src/signer.rs
```diff
@@ -1453,3 +1506,116 @@ impl Signer {
...
+    Decrypt,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn encrypted_messages_should_be_possible_to_decrypt() {
+        let msg = "Nobody's gonna know".as_bytes();
```
jferrant, line 1610: XD

jferrant (approved):
Awesome :)

hstove (approved):
ship it!

netrome:
Merged now. I'll continue working on https://github.com/stacks-network/stacks-core/issues/4654 so if there are any late realizations or proposals, I'm happy to accommodate them in the context of this next issue.

PR #4660, opened 2024-04-08
Debug/dkg results

jferrant:
We needlessly were triggering DKG rounds due to a timing issue between our first check to see if a key was approved and the later threshold check being exceeded. 
Now if we have not locally stored an approved key, we will always attempt the DKG queue checks. It will always follow this check by a final loading of whatever is set in the contract. Not bothering to queue a command if it finds an approved key. This should mitigate needless triggering of DKG for the obvious case. 

However, we still need the later update to retrieve DKG shares based on round ID in https://github.com/stacks-network/stacks-core/issues/4654 as we can never be sure that our view of the contract is correct at the time of triggering a DKG round. (we could be milliseconds off when checking the approved key is set still).

### DIFF
--- a/stacks-signer/src/signer.rs
+++ b/stacks-signer/src/signer.rs
@@ -996,6 +996,10 @@ impl Signer {
     /// Process a dkg result by broadcasting a vote to the stacks node
     fn process_dkg(&mut self, stacks_client: &StacksClient, dkg_public_key: &Point) {
         let mut dkg_results_bytes = vec![];
+        debug!(
+            "{self}: Received DKG result. Broadcasting vote to the stacks node...";
+            "dkg_public_key" => %dkg_public_key
+        );
         if let Err(e) = SignerMessage::serialize_dkg_result(
             &mut dkg_results_bytes,
             dkg_public_key,
--- a/stacks-signer/src/signer.rs
+++ b/stacks-signer/src/signer.rs
@@ -124,13 +124,22 @@ pub enum Command {
     },
 }
 
+/// The specific operations that a signer can perform
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub enum Operation {
+    /// A DKG operation
+    Dkg,
+    /// A Sign operation
+    Sign,
+}
+
 /// The Signer state
 #[derive(PartialEq, Eq, Debug, Clone)]
 pub enum State {
     /// The signer is idle, waiting for messages and commands
     Idle,
     /// The signer is executing a DKG or Sign round
-    OperationInProgress,
+    OperationInProgress(Operation),
 }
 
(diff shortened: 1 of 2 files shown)


xoloki (approved):
LGTM

stacks-signer/src/signer.rs
```diff
@@ -1235,35 +1269,56 @@ impl Signer {
...
             if old_dkg != self.approved_aggregate_public_key {
-                debug!(
+                warn!(
                     "{self}: updated DKG value to {:?}.",
                     self.approved_aggregate_public_key
                 );
             }
-            return Ok(());
-        };
+            if matches!(self.state, State::OperationInProgress(Operation::Dkg)) {
```
xoloki, line 1295: Why use `if matches!` instead of `if let`?  I thought clippy hated single branch matching, though maybe he can't tell if it's a macro?

xoloki (approved):
LGTM, trivial question about `if matches!()`

stacks-signer/src/signer.rs
```diff
@@ -1235,35 +1269,56 @@ impl Signer {
...
+                debug!(
+                    "{self}: DKG has already been set. Aborting DKG operation {}.",
```
hstove, line 1297: Is this log message correct? To me the code reads as "DKG already in progress - finish the operation"

hstove (approved):
LGTM - small comment about a log message that might just be me misreading something

stacks-signer/src/signer.rs
```diff
@@ -1235,35 +1269,56 @@ impl Signer {
...
+                    self.coordinator.current_dkg_id
+                );
+                self.finish_operation();
+            }
+        } else if should_queue {
+            info!("{self} is the current coordinator and must trigger DKG. Queuing DKG command...");
```
jcnelson, line 1303: Nit: duplicate log entry?

jcnelson (dismissed):
Small nits but LGTM

stacks-signer/src/signer.rs
obycode: The outer `if`:
```
```
indicates that DKG has already completed, then the inner `if`:
```
if let State::OperationInProgress(Operation::Dkg) = self.state {
```
indicates that it was triggered again, so it should be aborted.

kantai (dismissed):

hstove (approved):

obycode:
@jferrant can you double check my conflict resolution in the last commit here?

obycode (approved):

jferrant (commented):
Can't approve, but looks good to me :D

xoloki (approved):
LGTM

PR #4661, opened 2024-04-08
Hex Encode signer CLI outputs from StackerDB chunks

### LINKED ISSUE #4650: [stacks-signer] the cli commands for outputting data can be manipulated
Via audit: "The Stacks-signer commands get-chunk, get-latest-chunk and list-chunks retrieve chunks from StackerDB and print them to the system console. But the system console can be manipulated by binary control sequences, causing output data to be hidden or modified...Filter escape sequences, or preferentially, escape all binary data from the output to the terminal"

8marz8:
### Description
Hex encoding StackerDB chunks before writing to terminal to address audit security issue. 

### Applicable issues

- closes #4650

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3456,7 +3456,6 @@ dependencies = [
  "polynomial",
  "rand 0.8.5",
  "rand_core 0.6.4",
- "regex",
  "reqwest",
  "rusqlite",
  "secp256k1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3456,6 +3456,7 @@ dependencies = [
  "polynomial",
  "rand 0.8.5",
  "rand_core 0.6.4",
+ "regex",
  "reqwest",
  "rusqlite",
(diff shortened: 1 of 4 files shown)


jferrant:
Side note, this will have to be reopened against develop as soon as Brice finishes merging next into develop . See https://nakamotoslack.slack.com/archives/C067BHSN1SL/p1712679342282899

jferrant:
> @jferrant @jcnelson When I remove the ANSI filtering and only encode to hex, if I pipe the output to `xxd`, it will still hide the sequence provided in the audit example. Would it be the user's responsibility to safely read the output? Just want to ensure that on our end, it's sufficient that we only encode it.

I think it can be the user's responsibility to escape the output :)

8marz8:
@jferrant  Updated the base branch for the PR

jferrant (approved):

hstove (approved):

PR #4665, opened 2024-04-09
fix: fix delegate stack increase events

janniks:
- missing `)` in injected code was causing an event to not emit

> not consensus critical, but very important for the api


kantai (approved):
LGTM, assuming the CI passes.

wileyj (approved):
copypaste of @kantai comment

zone117x (approved):
Good find!

janniks:
Thanks to @rafaelcr 🕵

PR #4667, opened 2024-04-09
Feat: add affirmation map overrides to xenon default config for epoch 2.5

kantai:
This adds default overrides for the affirmation maps in Epoch 2.5 on the Xenon testnet. This is necessary because the 2.5 release version contains a different pox-4 contract than the one instantiated in Xenon testnet initially. This is similar to the reorg required during Epoch 2.4 on Xenon Testnet (see: https://github.com/stacks-network/stacks-core/pull/4560).

### DIFF
--- a/testnet/stacks-node/src/config.rs
+++ b/testnet/stacks-node/src/config.rs
@@ -13,7 +13,7 @@ use rand::RngCore;
 use serde::Deserialize;
 use stacks::burnchains::affirmation::AffirmationMap;
 use stacks::burnchains::bitcoin::BitcoinNetworkType;
-use stacks::burnchains::{Burnchain, MagicBytes, BLOCKSTACK_MAGIC_MAINNET};
+use stacks::burnchains::{Burnchain, MagicBytes, PoxConstants, BLOCKSTACK_MAGIC_MAINNET};
 use stacks::chainstate::nakamoto::signer_set::NakamotoSigners;
 use stacks::chainstate::stacks::boot::MINERS_NAME;
 use stacks::chainstate::stacks::index::marf::MARFOpenOpts;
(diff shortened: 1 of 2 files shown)


testnet/stacks-node/src/config.rs
```diff
@@ -211,7 +213,7 @@ mod tests {
             ConfigFile::from_str(&format!(
                 r#"
                     [[burnchain.affirmation_overrides]]
-                    reward_cycle = 1
```
wileyj, line 214: is this meant to only apply ovverrides for the known cycles (starting at 413), vs the existing method from cycle 1?

testnet/stacks-node/src/config.rs
```diff
@@ -222,7 +224,7 @@ mod tests {
         assert_eq!(config.burnchain.affirmation_overrides.len(), 1);
         assert_eq!(config.burnchain.affirmation_overrides.get(&0), None);
         assert_eq!(
-            config.burnchain.affirmation_overrides.get(&1),
```
wileyj, line 225: same question as above re: changing 1->413

testnet/stacks-node/src/config.rs
```diff
@@ -267,7 +269,7 @@ mod tests {
         )
         .expect("Expected to be able to parse affirmation map from file");
         // Should default add xenon affirmation overrides
-        assert_eq!(config.burnchain.affirmation_overrides.len(), 3);
+        assert_eq!(config.burnchain.affirmation_overrides.len(), 5);
```
wileyj, line 272: if we're adding 1 more override, should this be `4`?

testnet/stacks-node/src/config.rs
kantai: Yep, like above -- basically, the existing test supplied an illegal affirmation map / reward cycle number combination.

PR #4669, opened 2024-04-10
feat: initialize BitVec with correct length and track next BitVec

hstove:
This PR extends off of #4627, but is setup against the `develop` branch. My understanding is that this should be a develop-based PR now, because it's not consensus critical.

This PR adds two things:

- When a new Nakamoto block is formed by a miner, it starts with a BitVec of 1's with the length of the signer set.
- In the miner's coordinator thread, a `next_signer_bitvec` property is added. After a miner has collected all block responses from the signers, it updates this BitVec to match the signers who gave a response.

Note that, from a functional perspective, all this changes is that the length of the BitVec will match the signer's length. However, the BitVec will be all 1's, even if not all signers respond. Future work is required to construct a new block in the case that a block is rejected, and in that work we can use this `next_signer_bitvec` in that new block.

- Closes https://github.com/stacks-network/stacks-core/issues/4558

### DIFF
--- a/stacks-common/src/bitvec.rs
+++ b/stacks-common/src/bitvec.rs
@@ -1,3 +1,19 @@
+// Copyright (C) 2013-2020 Blockstack PBC, a public benefit corporation
+// Copyright (C) 2020-2024 Stacks Open Internet Foundation
+//
+// This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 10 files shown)


hstove:
Closing this - https://github.com/stacks-network/stacks-core/pull/4627 is now off of develop

PR #4670, opened 2024-04-10
Feat(pox-4): return an err on improper agg-incr invocations

kantai:
This PR alters the behavior of `stacks-aggregation-increase` in an improper invocation scenario. If the delegate invokes stacks-aggregation-increase before they (or anyone else) has performed a _commit_ (or solo stacking operation), the transaction will abort rather than returning a caller-catchable error. This is a slight bug: the difference is just whether or not the error is catchable by a contract that invokes the function.

This change would require tagging an rc-3, so the tradeoff is whether or not a nicer error handling here is worth a new release candidate for 2.5.

### DIFF
--- a/stackslib/src/chainstate/stacks/boot/pox-4.clar
+++ b/stackslib/src/chainstate/stacks/boot/pox-4.clar
@@ -897,12 +897,12 @@
               (err ERR_STACKING_INVALID_LOCK_PERIOD))
 
     (let ((partial-amount-ustx (get stacked-amount partial-stacked))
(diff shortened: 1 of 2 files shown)


kantai:
I think @wileyj can also weigh in on the tradeoff with an `2.5.0.0.0-rc3` here.

obycode (approved):
LGTM!

friedger (approved):
LGTM

jcnelson (approved):

Issue #4605, opened 2024-03-28
Nakamoto: miner needs rules for selecting parent chain tip


kantai:
The nakamoto miner must always build off of a chain tip that is the highest of:

1. The highest block in the miner's current tenure
2. The highest block in the current tenure's parent tenure

Where the current tenure's parent tenure is _the tenure start block committed to in the current tenure's associated block commit_.

This requires an update from the current logic:

https://github.com/stacks-network/stacks-core/blob/next/testnet/stacks-node/src/nakamoto_node/miner.rs#L663-L665

Which just queries the current canonical tip.

jcnelson:
The miner has to build off of the highest processed tip in all cases, right?  But, wouldn't that always be the canonical chain tip?  In what circumstances is it not?

kantai:
The issue is when the tenure immediately preceding the current miner doesn't produce a block until _after_ the current miner's block commit is submitted.

The current miner's block commit must be the first block of the parent tenure -- if they didn't receive a block from the preceding tenure, then they must assume the parent tenure is the preceding one to that, and that is where they must build.

Tenure A: Produces blocks `A_1...A_x`
Tenure B: Doesn't produce any blocks until Miner C has already produced and submitted block commit.
Tenure C: First block must descend from `A_x`, not `B_1`.

PR #4674, opened 2024-04-10
Do not use canonical tip when getting parent info as a fork may have happened

jferrant:
This was a mostly a fix by @kantai with a lot of help from @obycode so we prob should get @jcnelson review here. 

Closes: https://github.com/stacks-network/stacks-core/issues/4605

### DIFF
--- a/stackslib/src/chainstate/nakamoto/tenure.rs
+++ b/stackslib/src/chainstate/nakamoto/tenure.rs
@@ -528,6 +528,17 @@ impl NakamotoChainState {
         }
     }
 
+    /// Get the nakamoto tenure by id
+    pub fn get_nakamoto_tenure_change_by_tenure_id(
+        headers_conn: &Connection,
+        tenure_consensus_hash: &ConsensusHash,
+    ) -> Result<Option<NakamotoTenure>, ChainstateError> {
+        let sql = "SELECT * FROM nakamoto_tenures WHERE tenure_id_consensus_hash = ?1 ORDER BY tenure_index DESC LIMIT 1";
+        let args: &[&dyn ToSql] = &[&tenure_consensus_hash];
+        let tenure_opt: Option<NakamotoTenure> = query_row(headers_conn, sql, args)?;
+        Ok(tenure_opt)
+    }
+
     /// Get a nakamoto tenure-change by its tenure ID consensus hash.
     /// Get the highest such record.  It will be the last-processed BlockFound tenure
     /// for the given sortition consensus hash.
@@ -544,7 +555,7 @@ impl NakamotoChainState {
         Ok(tenure_opt)
     }
 
-    /// Get the highest processed tenure on the canonical sortition history.
+    /// Get the highest non-empty processed tenure on the canonical sortition history.
     pub fn get_highest_nakamoto_tenure(
         headers_conn: &Connection,
         sortdb_conn: &Connection,
@@ -555,10 +566,7 @@ impl NakamotoChainState {
             // no chain tip, so no tenure
             return Ok(None);
         }
-        let sql = "SELECT * FROM nakamoto_tenures WHERE tenure_id_consensus_hash = ?1 ORDER BY tenure_index DESC LIMIT 1";
-        let args: &[&dyn ToSql] = &[&tip_ch];
-        let tenure_opt: Option<NakamotoTenure> = query_row(headers_conn, sql, args)?;
-        Ok(tenure_opt)
+        Self::get_nakamoto_tenure_change_by_tenure_id(headers_conn, &tip_ch)
     }
 
     /// Verify that a tenure change tx is a valid first-ever tenure change.  It must connect to an
@@ -655,7 +663,7 @@ impl NakamotoChainState {
     /// * previous_tenure_blocks
     /// * cause
     ///
-    /// Returns Ok(Some(highest-processed-tenure)) on success
+    /// Returns Ok(Some(processed-tenure)) on success
     /// Returns Ok(None) if the tenure change is invalid
     /// Returns Err(..) on DB error
     pub(crate) fn check_nakamoto_tenure<SH: SortitionHandle>(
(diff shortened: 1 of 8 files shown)


hstove:
This is the approach that I thought made sense given how I read #4605, but I'm curious what you think @kantai

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -660,8 +660,18 @@ impl BlockMinerThread {
...
-        let Some(stacks_tip) =
-            NakamotoChainState::get_canonical_block_header(chain_state.db(), burn_db)
+        // The nakamoto miner must always build off of a chain tip that is the highest of:
+        // 1. The highest block in the miner's current tenure
+        // 2. The highest block in the current tenure's parent tenure
+        // Where the current tenure's parent tenure is the tenure start block committed to in the current tenure's associated block commit.
+        let stacks_block_id = if let Some(block) = self.mined_blocks.last() {
+            block.block_id()
+        } else {
+            self.parent_tenure_id
```
jferrant, line 670: This is wrong. I need to find the TIP of the parent tenure. I.e. the last block of the parent tenure list of blocks.

obycode (dismissed):
🎉 LGTM!

testnet/stacks-node/src/nakamoto_node/miner.rs
```diff
@@ -683,6 +714,19 @@ impl BlockMinerThread {
...
+            if let Some(last_tenure_finish_block_header) =
+                NakamotoChainState::get_nakamoto_tenure_finish_block_header(
+                    chain_state.db(),
+                    &stacks_tip_header.consensus_hash,
+                )
+                .expect("FATAL: could not query parent tenure finish block")
+            {
+                stacks_tip_header = last_tenure_finish_block_header;
+            }
+        }
```
kantai, line 729: I think it makes it more clear if the logic of like 689-729 is combined so that first we get the tenure consensus hash of the parent we're building on, and then we always fetch the finish block header.

```rust
let tenure_of_parent = if let Some(block) = self.mined_blocks.last() {
    block.header.consensus_hash.clone()
} else {
    let Some(parent_tenure_block) = NakamotoChainState::get_block_header(chain_state.db(), &self.parent_tenure_id) else {
        debug!("No Stacks chain tip known, will return a genesis block");
        ...
    };
    parent_tenure_block.consensus_hash
};

let Some(parent_block_header) = NakamotoChainState::get_nakamoto_tenure_finish_block_header(..).expect("FATAL: ...") else {
    warn!("Failed to load the last block of the parent tenure");
    return Err(NakamotoNodeError::...)
};
```

Note: I also suggest renaming `stacks_tip_header` to `parent_block_header`. The parent block is no longer _necessarily_ the stacks tip (even though we usually expect it to be so).

kantai (commented):
This looks great to me, I just had a comment on some refactoring of the logic in miner.

testnet/stacks-node/src/nakamoto_node/miner.rs
jferrant: I can prob rework it, but pretty sure one of the reasons I did it this way is that it could be the first nakamoto tenure ever and therefore its parent tenure tip check will fail as it has no nakamoto parent. This enforces that it uses the genesis block in that case. Otherwise I think we will error on the second block in the first nakamoto tenure? Will give it a try though.

jferrant: K I can't get this to work. It seems that there is no nakamoto tenure stored for the case of the first nakamoto tenure. i.e. the tenure is not empty, It is not the first block. it has a valid prior block so it doens't use a genesis block...BUT it is part of the first tenure still. therefore its not stored in the last completed list of nakamoto tenures it seems and it therefore hits "Failed to load the last block of the parent tenure".

testnet/stacks-node/src/tests/nakamoto_integrations.rs
```diff
@@ -3170,17 +3170,19 @@ fn forked_tenure_is_ignored() {
...
     );
     let stacker_sk = setup_stacker(&mut naka_conf);
+    let http_origin = format!("http://{}", &naka_conf.node.rpc_bind);
 
     test_observer::spawn();
     let observer_port = test_observer::EVENT_OBSERVER_PORT;
     naka_conf.events_observers.insert(EventObserverConfig {
         endpoint: format!("localhost:{observer_port}"),
-        events_keys: vec![EventKeyType::AnyEvent],
+        events_keys: vec![EventKeyType::AnyEvent, EventKeyType::MinedBlocks],
```
obycode, line 3185: Isn't this redundant?

obycode (commented):
This looks good to me, but the observer still doesn't work for me locally. I'm not sure why. If it works for you and on CI, then let's merge.

stackslib/src/chainstate/nakamoto/tenure.rs
```diff
@@ -764,84 +775,34 @@ impl NakamotoChainState {
                     );
                     return Ok(None);
                 }
-                if tenure_payload.burn_view_consensus_hash
```
jcnelson, line 767: I'm beginning to think that a lot of the assumptions that went into this module are wrong.  In particular, this module assumes that the tenure history does not fork unless the Bitcoin history also forks, but it seems that this is now a design requirement?  This assumption made sense while we were also assuming that the signer would never permit a Stacks fork, but it appears that this is actually unavoidable.

I think some substantial refactoring is in order.

stackslib/src/chainstate/nakamoto/tenure.rs
```diff
@@ -528,6 +528,17 @@ impl NakamotoChainState {
         }
     }
 
+    /// Get the nakamoto tenure by id
+    pub fn get_nakamoto_tenure_change_by_tenure_id(
```
jcnelson, line 532: N.B. this doesn't deal with tenure forks.  There could be multiple Nakamoto tenures with the same `tenure_consensus_hash`

kantai: I don't think that there could be multiple tenures with the same consensus hash. The only permitted forking scenarios would be triggered by block commits (i.e., a tenure's first block is a sibling of another tenure's first block).

testnet/stacks-node/src/tests/nakamoto_integrations.rs
jferrant: No it is not. I know its confusing AF, but EventKeyType::AnyEvent does not subscribe to all events. You must subscribe to them explicitly.

obycode: Oh, that's fun. Thanks!

stackslib/src/chainstate/nakamoto/tenure.rs
jcnelson: (This doesn't have to be done in this PR; I'll create a separate ticket once I can spec out the work)

stackslib/src/chainstate/nakamoto/tenure.rs
jcnelson: Yes; that's what I'm referring to.

kantai (approved):

obycode (approved):

jcnelson (approved):

PR #4677, opened 2024-04-11
fix: default 2.5 xenon affirmation overrides, allow config setting to…

kantai:
### Description

This changes the 2.5 xenon overrides to `an` instead of `aa`, and also allows config settings to override the default values.


jcnelson (approved):

wileyj (approved):
shipt

Issue #4602, opened 2024-03-28
[stacks-signer] add clean logging to the signer for a signer that has been successfully configured


jferrant:
It helps with debugging/knowing if you have properly configured a signer the way you expect if you print a human readable version of the loaded config

PR #4678, opened 2024-04-11
feat: add startup logging with config to signer

hstove:
- Closes https://github.com/stacks-network/stacks-core/issues/4602

This adds a simple log when the signer is started with the signer's configuration information. The goal is just to help signers ensure their signer is configured as expected. It also provides the signer's STX address and public key.

I've also added a `check-config` command which validates a config file and outputs the same startup logging information.

### DIFF
--- a/stacks-signer/src/cli.rs
+++ b/stacks-signer/src/cli.rs
@@ -64,6 +64,8 @@ pub enum Command {
     GenerateFiles(GenerateFilesArgs),
     /// Generate a signature for Stacking transactions
     GenerateStackingSignature(GenerateStackingSignatureArgs),
+    /// Check a configuration file and output config information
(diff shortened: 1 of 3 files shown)


xoloki (dismissed):
LGTM

wileyj:
@hstove if you want this in the next rc, can you rebase and PR against `release/2.5.0.0.0`?

hstove:
@wileyj I've made https://github.com/stacks-network/stacks-core/pull/4679 which targets the release branch. I figured we should keep both PRs to prevent mismatch between develop and release. If that's not a problem we can close the PR.

jferrant (approved):

kantai (approved):

PR #4679, opened 2024-04-12
feat: add signer startup logging

hstove:
This is the same code as #4678 , but targeted to the release branch, in hopes that we can get this included in the next RC. I think it's best to keep both PRs open, otherwise we'll have to pull from the release branch into develop (instead of the other way around as usual)

- Closes https://github.com/stacks-network/stacks-core/issues/4602

This adds a simple log when the signer is started with the signer's configuration information. The goal is just to help signers ensure their signer is configured as expected. It also provides the signer's STX address and public key.

I've also added a `check-config` command which validates a config file and outputs the same startup logging information.


wileyj (approved):
shipit

obycode (approved):
lgtm

PR #4681, opened 2024-04-13
fix: Use finite set of strings for HTTP metrics

### LINKED ISSUE #4574: High cardinality for stacks_node_rpc_call_latencies_histogram metrics
With the latest release ([2.4.0.1.0](https://github.com/stacks-network/stacks-core/releases/tag/2.4.0.1.0)), the `stacks_node_rpc_call_latencies_histogram.*` prometheus metrics seemed to have changed and are now creating large amounts of cardinality due to the unique `path` labels being applied to them.
This can cause prometheus to consume more memory and become unusable pretty quickly. A short-term fix for this would be to drop the metric series from being scraped like so:
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: my-service-mon
spec:
  endpoints:
  - path: /metrics
    port: my-metrics-port
    relabelings:
    - action: drop
      regex: stacks_node_rpc_call_latencies_histogram.*
      sourceLabels:
      - __name__
...
```

However, a long-term fix would require making the `path` label less unique. The Stacks Blockchain API does something like this by globbing similar paths like so:
`/v2/attachments/012c8195f4381d0b0eadc2daee6e2c5b8243f4ff` would become `/v2/attachments/:attachment`

Very small snippet of the problematic metrics series (there are hundreds of thousands of these):
```
stacks_node_rpc_call_latencies_histogram_count{path="/v2/attachments/012c8195f4381d0b0eadc2daee6e2c5b8243f4ff"} 2
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a",le="0.005"} 1
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a",le="0.01"} 1
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a",le="0.025"} 1
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a",le="0.05"} 1
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a",le="0.1"} 1
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a",le="0.25"} 1
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a",le="0.5"} 1
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a",le="1"} 1
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a",le="2.5"} 1
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a",le="5"} 1
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a",le="10"} 1
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a",le="+Inf"} 1
stacks_node_rpc_call_latencies_histogram_sum{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a"} 0.000895564
stacks_node_rpc_call_latencies_histogram_count{path="/v2/attachments/01586bc1be60143421cc3d7828e50bb26ded049a"} 1
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/016a3cb605d341b16c214dbb78b4bb90618bee80",le="0.005"} 2
stacks_node_rpc_call_latencies_histogram_bucket{path="/v2/attachments/016a3cb605d341b16c214dbb78b4bb90618bee80",le="0.01"} 2
```

jbencin:
### Description

When sending metrics data to Prometheus, map URLs to a finite set of strings. This will reduce the number of unique strings the server has to keep track of. As an example, it might rewrite:

```
/v2/attachments/012c8195f4381d0b0eadc2daee6e2c5b8243f4ff
```

to

```
/v2/attachments/:hash
```

### Applicable issues

- fixes #4574

**NOTE:** The test `net::stackerdb::tests::sync::test_stackerdb_replica_2_neighbors_1_chunk` is currently failing, but also fails in `master`, meaning it was not broken by this PR

- [ ] Changelog is updated

### DIFF
--- a/stackslib/src/net/httpcore.rs
+++ b/stackslib/src/net/httpcore.rs
@@ -438,6 +438,8 @@ pub struct StacksHttpRequest {
     preamble: HttpRequestPreamble,
     contents: HttpRequestContents,
     start_time: u128,
+    /// Cache result of `StacksHttp::find_response_handler` so we don't have to do the regex matching twice
+    response_handler_index: Option<usize>,
 }
 
 impl StacksHttpRequest {
@@ -446,6 +448,7 @@ impl StacksHttpRequest {
             preamble,
             contents,
             start_time: get_epoch_time_ms(),
+            response_handler_index: None,
         }
     }
 
@@ -481,6 +484,7 @@ impl StacksHttpRequest {
             preamble,
             contents,
             start_time: get_epoch_time_ms(),
+            response_handler_index: None,
         })
     }
 
@@ -912,9 +916,7 @@ impl StacksHttp {
             if request_verb != verb {
                 continue;
             }
-            let _captures = if let Some(caps) = regex.captures(request_path) {
-                caps
-            } else {
+            let Some(_captures) = regex.captures(request_path) else {
                 continue;
             };
 
@@ -985,9 +987,7 @@ impl StacksHttp {
             if &preamble.verb != verb {
                 continue;
             }
-            let captures = if let Some(caps) = regex.captures(&decoded_path) {
-                caps
-            } else {
+            let Some(captures) = regex.captures(&decoded_path) else {
                 continue;
             };
 
@@ -1085,21 +1085,18 @@ impl StacksHttp {
         node: &mut StacksNodeState,
     ) -> Result<(HttpResponsePreamble, HttpResponseContents), NetError> {
         let (decoded_path, _) = decode_request_path(&request.preamble().path_and_query_str)?;
-        let response_handler_index =
-            if let Some(i) = self.find_response_handler(&request.preamble().verb, &decoded_path) {
-                i
-            } else {
-                // method not found
-                return StacksHttpResponse::new_error(
-                    &request.preamble,
-                    &HttpNotFound::new(format!(
-                        "No such API endpoint '{} {}'",
-                        &request.preamble().verb,
-                        &decoded_path
-                    )),
-                )
-                .try_into_contents();
-            };
+        let Some(response_handler_index) = request.response_handler_index.or_else(|| self.find_response_handler(&request.preamble().verb, &decoded_path)) else {
+            // method not found
+            return StacksHttpResponse::new_error(
+                &request.preamble,
+                &HttpNotFound::new(format!(
+                    "No such API endpoint '{} {}'",
+                    &request.preamble().verb,
+                    &decoded_path
+                )),
+            )
+            .try_into_contents();
+        };
 
         let (_, _, request_handler) = self
             .request_handlers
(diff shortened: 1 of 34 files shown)


stackslib/src/net/httpcore.rs
```diff
@@ -1242,6 +1240,29 @@ impl StacksHttp {
...
+    /// Get a unique `&str` identifier for each request type
+    /// This can only return a finite set of identifiers, which makes it safer to use for Prometheus metrics
+    /// For details see https://github.com/stacks-network/stacks-core/issues/4574
+    pub fn metrics_identifier(&self, req: &mut StacksHttpRequest) -> &str {
+        let Ok((decoded_path, _)) = decode_request_path(&req.request_path()) else {
+            return "<err-url-decode>";
+        };
+
+        let Some(response_handler_index) = req.response_handler_index
+            .or_else(|| self.find_response_handler(&req.preamble().verb, &decoded_path))
```
jcnelson, line 1252: Is this `.or_else()` closure ever reachable?  Presumably, we only emit a metric _after_ we match the path?

jbencin: It's reachable if we build without the `monitoring_prom` feature flag and the metrics code is never run

jbencin:
Added unit test for `metrics_identifier()`. This is now fully ready for review.

Also note that `net::stackerdb::tests::sync::test_stackerdb_replica_2_neighbors_1_chunk` is failing (at least locally for me), but also fails in the current `master`

stackslib/src/net/httpcore.rs
jbencin: Ignore my last comment, I thought you were referring to the `or_else()` in `try_handle_request()`. That one will only be executed if metrics is disabled, this one should always be executed, because `response_handler_index` should not be set before this point

CharlieC3:
I built and deployed this branch to one of our dev seed nodes. I can't comment on all endpoints addressed in this PR because it takes some time for all endpoints to eventually be hit, but it certainly seems like this will fix the issue!

Screenshot 2024-04-19 at 11 44 22 AM

CharlieC3 (approved):
Very thorough! Thanks for much for taking the time to make these changes, it'll really help with our alerting and reducing false positives. Left a couple minor suggestions but otherwise LGTM :shipit:

CharlieC3 (approved):

CharlieC3:
@wileyj What would it take to get a new version cut with this PR? It's not in 2.5.0.0.3. If a new version is coming soon, am fine with waiting.

wileyj:
@CharlieC3 nothing planned immediately, but we can start prepping something - i know there are some improvements to the signer that may be useful to release as well. 
i would bring it up during eng. call and we can see what other PR's (if any) should be merged into an rc

PR #4682, opened 2024-04-13
Add missing stacks-signer binary to release docker images

wileyj:
cc @friedger 

Reported that the stacks-signer binary was missing in the rc4 docker release image. 
this changes the current dockerfile to also add `stacks-signer` along with `stacks-node` to the release docker image. 

More binaries can be added at the expense of image size, but i think these 2 are the bare minimum.


friedger (approved):
Looks good.

Alternatively, a different image would need to be published for the signer.

wileyj:
> Looks good.
> 
> Alternatively, a different image would need to be published for the signer.

yea, i like that idea. will look into it for next release

cylewitruk (approved):
lgtm!

Issue #4649, opened 2024-04-06
[stacks-signer] enable some sort of garbage collection in stacks signer


jferrant:
Currently we have no garbage collection for stacks-signer signerDB. We should cleanup blocks after some number of bitcoin or stacks blocks have passed. A simple cleanup would be to delete any data belonging to a past reward cycle, but depending on the size of each block being stored, a reward cycle could grow quite large and require additional cleanup.

jferrant:
Closing, won't do. We want to keep blocks around indefinitely and if signers complain about lack of auto deletion, we will reopen and explore config options related to when to cleanup blocks.

brice-stacks:
Time to reopen this issue. The db is growing without limit and it is causing problems for signers.

PR #4685, opened 2024-04-15
Chore/stacks signer garbage collection

jferrant:
Implements garbage collection based on burn block height and reward cycle. If a block is more than 2100 burn blocks, delete it. If a signer state is older than the previous reward cycle, delete it. 

As part of this change, I also removed the block proposal message type. We originally proposed a block and then signers coordinated a signing round based on their assessment of it. Now, a miner will always immediately trigger a sign round following a block proposal. Instead, this PR makes a miner's nocne request be across a block proposal effectively removing the need for this sep block proposal message type. (signers already uupdate their nonce response according to their vote so was an easy change)

**NOTE** This does not really account for a malicious miner trying to bring down a signer by spamming the signer with multiple block proposals though we do have some mitigations already in place (stackerdb does not allow too frequent writing, nor does it allow excessively large stackerdb message sizes. However, should there be some further check on signers to prevent their block db from getting filled? I had assumed that the miner has motivation enough to not be a dick since they want blocks to be signed to earn fees, but let me know if that is not good enough :P

Closes https://github.com/stacks-network/stacks-core/issues/4649

### DIFF
--- a/stacks-signer/src/runloop.rs
+++ b/stacks-signer/src/runloop.rs
@@ -296,7 +296,9 @@ impl RunLoop {
             }
         }
         self.cleanup_stale_signers(current_reward_cycle);
-        if let Err(e) = self.cleanup_stale_data(current_reward_cycle) {
+        // Cleanup stale signer state (reward cycle states that are older than the prev reward cycle
+        // and blocks that are older than a full reward cycle length)
+        if let Err(e) = self.cleanup_stale_data(current_reward_cycle, current_burn_block_height) {
             error!("Failed to cleanup stale signer data: {e}");
         }
         if self.stacks_signers.is_empty() {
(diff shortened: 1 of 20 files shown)


stacks-signer/src/runloop.rs
```diff
@@ -287,7 +291,12 @@ impl RunLoop {
             *reward_cycle_info = new_reward_cycle_info;
         }
         let current_reward_cycle = reward_cycle_info.reward_cycle;
-        // We should only attempt to refresh the signer if we are not configured for the next reward cycle yet and we received a new burn block for its prepare phase
+        let current_burn_block_height = reward_cycle_info.last_burnchain_block_height;
```
hstove, line 294: Why overwrite the `current_burn_block_height` passed in as a function arg?

jferrant: beacuse our view may have changed. Most likely not. But could happen. and we already get themost current burn block info by calling get reward cycle info again, so figured, make it consistent. It most liekly is the same value, but we cannot guarantee it.

jferrant: Correct solution is to use the burn block height of the Block proposal of the signer (they do store the burn block height at which it was created) but I haven't done that yet because I want confirmation that this is the approach to go with as its slightly more involved change.

jferrant: Ended up just doing it.

xoloki (changes requested):
LGTM.  Ideally I'd like to have the miner changes in a separate PR, but it's really not worth your time to split them out now.

xoloki (approved):

jferrant:
> LGTM. Ideally I'd like to have the miner changes in a separate PR, but it's really not worth your time to split them out now.

It could definitely be a sep PR and was debating doing this... but without it, I don't have the block info needed to actually delete the block in 2100 sortitions and I didn't want another PR built off another PR since I have been having a really hard time getting these things merged XD

jferrant:
Closing won't do. We want to keep block data around indefinitely. If signers take issue with this, we can reopen at a later time/explore config options.

PR #4690, opened 2024-04-16
Fix: validate epochs must special case 2.5

kantai:
The 2.5 node defines Epoch 3.0, but it does not use its network version byte, which means validate_epochs must special case 2.5 and 3.0.


obycode (approved):
LGTM!

saralab (approved):

jcnelson (approved):

PR #4692, opened 2024-04-16
Update SortitionDB epochs on connect

jcnelson:
See title. Fixes existing 2.5 chainstate.

### DIFF
--- a/stackslib/src/chainstate/burn/db/sortdb.rs
+++ b/stackslib/src/chainstate/burn/db/sortdb.rs
@@ -2848,6 +2848,35 @@ impl SortitionDB {
         Ok(())
     }
 
+    /// Validates given StacksEpochs (will runtime panic if there is any invalid StacksEpoch structuring) and
+    /// replaces them into the SortitionDB's epochs table
+    fn validate_and_replace_epochs(
+        db_tx: &Transaction,
+        epochs: &[StacksEpoch],
+    ) -> Result<(), db_error> {
+        let epochs = StacksEpoch::validate_epochs(epochs);
+        let existing_epochs = Self::get_stacks_epochs(db_tx)?;
+        if existing_epochs == epochs {
+            return Ok(());
+        }
+        info!("Replace existing epochs with new epochs");
+        db_tx.execute("DELETE FROM epochs;", NO_PARAMS)?;
+        for epoch in epochs.into_iter() {
+            let args: &[&dyn ToSql] = &[
+                &(epoch.epoch_id as u32),
+                &u64_to_sql(epoch.start_height)?,
+                &u64_to_sql(epoch.end_height)?,
+                &epoch.block_limit,
+                &epoch.network_epoch,
+            ];
+            db_tx.execute(
+                "INSERT INTO epochs (epoch_id,start_block_height,end_block_height,block_limit,network_epoch) VALUES (?1,?2,?3,?4,?5)",
+                args
+            )?;
+        }
+        Ok(())
+    }
+
     /// Get a block commit by its content-addressed location in a specific sortition.
     pub fn get_block_commit(
         conn: &Connection,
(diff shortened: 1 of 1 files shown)


stackslib/src/chainstate/burn/db/sortdb.rs
```diff
@@ -2848,6 +2848,35 @@ impl SortitionDB {
...
+    /// replaces them into the SortitionDB's epochs table
+    fn validate_and_replace_epochs(
+        db_tx: &Transaction,
+        epochs: &[StacksEpoch],
+    ) -> Result<(), db_error> {
+        let epochs = StacksEpoch::validate_epochs(epochs);
+        let existing_epochs = Self::get_stacks_epochs(db_tx)?;
+        if existing_epochs == epochs {
+            return Ok(());
+        }
```
kantai, line 2861: I'd prefer to see some kind of early erroring here in the event that `diff(existing_epochs, epochs).start_height` has been reached already by the sortition db.

wileyj (approved):
looks good to me, pending @kantai conversation resolution

stackslib/src/chainstate/burn/db/sortdb.rs
jcnelson: Sure, one sec

saralab (approved):
LFG 🚀

jcnelson:
@kantai does the current state address your feedback?

kantai:
> @kantai does the current state address your feedback?

 Not exactly.

I was imagining something along the lines of this commit: https://github.com/stacks-network/stacks-core/commit/124c052fd319664ac9644dd4f91db183d1cdeb17

Where the first changed epoch's block height is checked against the Sortition DB's current height.

I understand that something like that isn't without its own risks -- its definitely more complicated logic, so its easier for something to go wrong there, but it would provide an early exit in systems where the chainstate isn't recoverable.

I don't feel super strongly about this, because really, the release process should prevent this kind of thing from occurring. Also happy to put a pin in this and revisit in a subsequent release, because the updating logic in this PR works for now.

wileyj:
> > @kantai does the current state address your feedback?
> 
> Not exactly.
> 
> I was imagining something along the lines of this commit: [124c052](https://github.com/stacks-network/stacks-core/commit/124c052fd319664ac9644dd4f91db183d1cdeb17)
> 
> Where the first changed epoch's block height is checked against the Sortition DB's current height.
> 
> I understand that something like that isn't without its own risks -- its definitely more complicated logic, so its easier for something to go wrong there, but it would provide an early exit in systems where the chainstate isn't recoverable.
> 
> I don't feel super strongly about this, because really, the release process should prevent this kind of thing from occurring. Also happy to put a pin in this and revisit in a subsequent release, because the updating logic in this PR works for now.



> > @kantai does the current state address your feedback?
> 
> Not exactly.
> 
> I was imagining something along the lines of this commit: [124c052](https://github.com/stacks-network/stacks-core/commit/124c052fd319664ac9644dd4f91db183d1cdeb17)
> 
> Where the first changed epoch's block height is checked against the Sortition DB's current height.
> 
> I understand that something like that isn't without its own risks -- its definitely more complicated logic, so its easier for something to go wrong there, but it would provide an early exit in systems where the chainstate isn't recoverable.
> 
> I don't feel super strongly about this, because really, the release process should prevent this kind of thing from occurring. Also happy to put a pin in this and revisit in a subsequent release, because the updating logic in this PR works for now.

if that's the case, can you resolve the above conversation and we'll merge this (while creating a new issue to resolve this in a subsequent release)?

jcnelson:
> Not exactly.

Oh lol I forgot to push my latest when I asked you about that, @kantai.  What do you think now?

wileyj:
> > Not exactly.
> 
> Oh lol I forgot to push my latest when I asked you about that, @kantai. What do you think now?

rust fmt?

kantai (approved):

wileyj (approved):
shipit

PR #4695, opened 2024-04-18
Remove redundant block proposal message and fix tests

jferrant:
This is a cleanup of the miner block proposal process. Since the miner immediately follows up a block proposal with a sign request over the same block, it would be better to make the sign request also the block proposal message. Removes some extra logic and the extra block slot.


xoloki (dismissed):
LGTM

hstove (dismissed):
:+1: and always like a net code removal 😀

jcnelson (dismissed):

kantai (dismissed):

kantai (approved):

xoloki (approved):
LGTM

PR #4698, opened 2024-04-19
PoX-4 unit testing

obycode:
Add Clarity unit testing for the pox-4 contract.


friedger:
Why not using the pox contract from stacks lib?

obycode:
I was actually curious about that too. I think there was some reason related to coverage that @hugocaillard did that. I'll let him answer.

friedger:
Coverage reports are broken. How do we ensure that the two pox files are always in sync? Maybe we can copy it over? Or make it a requirement?

hugocaillard:
Today I moved these unit tests to their own directory, so that it doesn't conflicts with the property tests. I also added a readme that quickly explain how the tests work.

@friedger 
> Why not using the pox contract from stacks lib?

The contract is only used to render the diff, it's not contract that is called to perform the tests. So i guess it's not critical.
It need to be in sync with the pox-4 contract that is embedded in clarinet, which is the contract that is actually tested here.

It's not ideal, in the future we could expect that Clarinet can take custom code as boot-contracts (so that we could enforce to use the contracts as they exists in this repo)

hugocaillard (dismissed):
LGTM  🙌

kantai (dismissed):
LGTM!

kantai (approved):

hugocaillard (approved):
LGTM

PR #4704, opened 2024-04-22
feat: stacks signer able to save multiple dkg shares and load it where appropriate

### LINKED ISSUE #4654: Nakamoto Signer[3.0] -  A signer must be able to save multiple DKG aggregate key round states and load it where appropriate

netrome:
### Description
This PR extends the signer persistence to allow for storing shares for multiple concurrent DKG rounds, identified by the aggregate public key for that round. On startup, the signer looks up the approved aggregate key in the contract and loads the corresponding shares.

To ensure the Nakamoto sign coordinator is able to pick up loaded DKG shares, the signers broadcast `DkgResults` messages when loaded with their polynomial commitments. The sign coordinator, which previously assumed that any `DkgResults` message contained the complete set of polynomial commitments, has been extended to also allow individual polynomial commitments in these messages - which it will then aggregate for all signers to form the complete set of commitments.

### Applicable issues

- closes #4654 

### Additional info

To ensure a bounded size stored in StackerDB, we limit the implementation to only hold two sets of shares concurrently to handle the scenario described in #4654.


stacks-signer/src/runloop.rs
```diff
@@ -235,7 +236,22 @@ impl RunLoop {
...
+            let mut new_signer = Signer::from(new_signer_config);
+            let dkg_id = self
+                .stacks_client
+                .get_last_round(reward_cycle)
+                .expect("Failed to get last DKG round")
+                .unwrap_or(0);
+            let approved_aggregate_key = self
+                .stacks_client
+                .get_approved_aggregate_key(reward_cycle)
+                .expect("Failed to get approved aggregate key");
```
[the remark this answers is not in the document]
netrome: Fair enough.  Setting it to `None` feels quite brittle here, because the signer wouldn't be able to load the appropriate DKG shares - so we'd essentially start a broken signer unable to participate in signing rounds. But I can go the retry route like you suggested. I'll  wrap this in a retry loop, and return an error if it fails - failing the call to `self.initialize_runloop()` so that it gets retried again next pass.

stacks-signer/src/signer.rs
```diff
@@ -1403,9 +1503,6 @@ impl Signer {
         self.approved_aggregate_public_key =
             stacks_client.get_approved_aggregate_key(self.reward_cycle)?;
         if self.approved_aggregate_public_key.is_some() {
-            // TODO: this will never work as is. We need to have stored our party shares on the side etc for this particular aggregate key.
-            // Need to update state to store the necessary info, check against it to see if we have participated in the winning round and
-            // then overwrite our value accordingly. Otherwise, we will be locked out of the round and should not participate.
```
hstove, line 1408: Nice 😀

stacks-signer/src/signer.rs
```diff
@@ -1609,15 +1706,19 @@ fn load_encrypted_signer_state<S: SignerStateStorage>(
...
+) -> Result<StoredSignerStates, PersistenceError> {
+    let Some(encrypted_state) = storage.get_encrypted_signer_state(id)? else {
+        return Ok(VecDeque::new());
+    };
+    let serialized_state = decrypt(private_key, &encrypted_state)?;
+
+    Ok(serde_json::from_slice(&serialized_state)?)
+}
+
+fn get_signer(loaded_signers: StoredSignerStates, aggregate_key: Point) -> Option<SignerState> {
```
hstove, line 1718: Nit
```suggestion
fn get_signer_state(loaded_signers: StoredSignerStates, aggregate_key: Point) -> Option<SignerState> {
```

stacks-signer/src/signer.rs
```diff
@@ -1280,8 +1290,30 @@ impl Signer {
     fn save_signer_state(&mut self) -> Result<(), PersistenceError> {
```
hstove, line 1290: nit
```suggestion
    fn save_signer_states(&mut self) -> Result<(), PersistenceError> {
```

hstove (dismissed):
Everything LGTM! There are a few instances where it would be helpful to pluralize function names that now return states instead of one state.

stacks-signer/src/runloop.rs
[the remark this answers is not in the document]
netrome: Done ✅

stacks-signer/src/signer.rs
netrome: Thanks for spotting, updated :pray:

stacks-signer/src/signer.rs
netrome: This function still just stores the current signer state, so I'd prefer not to pluralize it.

netrome:
> Everything LGTM! There are a few instances where it would be helpful to pluralize function names that now return states instead of one state.

Good point. I've updated `get_signer` as you noted, and also `load_encrypted_signer_state` which should be pluralized.

stacks-signer/src/runloop.rs
```diff
@@ -235,12 +236,33 @@ impl RunLoop {
...
+            })?
+            .unwrap_or(0);
+
+            let approved_aggregate_key = retry_with_exponential_backoff(|| {
+                self.stacks_client
+                    .get_approved_aggregate_key(reward_cycle)
+                    .map_err(backoff::Error::transient)
+            })?;
+
+            new_signer.state_machine.reset(dkg_id, &mut OsRng);
```
kantai, line 254: Can you add an inline-comment here that OsRng is the same source of randomness used in `wsts::state_machine::signer::Signer::new()`?

stacks-signer/src/signer.rs
```diff
@@ -1328,6 +1360,74 @@ impl Signer {
...
+                &aggregate_key,
+                party_polynomials.into_iter(),
+            ) {
+                error!(
+                    "{self}: Failed to serialize DKGResults message after loading saved state: {e}"
+                );
+            } else if let Err(e) = self
+                .stackerdb
+                .send_message_bytes_with_retry(&MessageSlotID::DkgResults, dkg_results_bytes)
+            {
```
kantai, line 1391: Why does this send a DkgResults message when it finishes loading saved state?

stacks-signer/src/signer.rs
```diff
@@ -1609,15 +1706,22 @@ fn load_encrypted_signer_state<S: SignerStateStorage>(
...
+    Ok(serde_json::from_slice(&serialized_state)?)
+}
+
+fn get_signer_state(
+    loaded_signers: StoredSignerStates,
+    aggregate_key: Point,
+) -> Option<SignerState> {
+    loaded_signers
+        .into_iter()
+        .find(|state| state.group_key == aggregate_key)
```
kantai, line 1724: It seems like this method is only ever invoked once, and it's a pretty small function -- I think this function obscures more than it simplifies the code -- my suggestion is to just remove this function and move the code to `load_saved_state_for_aggregate_key`

stacks-signer/src/signer.rs
```diff
@@ -1328,6 +1360,74 @@ impl Signer {
...
+            &mut self.stackerdb,
+            self.signer_slot_id.into(),
+            &self.state_machine.network_private_key,
+        ).or_else(|err| {
+                warn!("Failed to load encrypted signer state from StackerDB, falling back to SignerDB: {err}");
+                load_encrypted_signer_state(
+                    &self.signer_db,
+                    self.reward_cycle,
+                    &self.state_machine.network_private_key)
+        })?;
```
kantai, line 1426: I'm curious about the ordering here. Loading the state from signer_db doesn't require RPC calls while loading from stackerdb does. Even in a perfect world, RPCs require a inter-process call at least, but the stacks-node RPC interface isn't perfect and could definitely stall while doing things like processing blocks, etc. It seems preferable to query the signer_db first -- is there a good reason not to?

stacks-signer/src/signer.rs
```diff
@@ -1609,15 +1706,22 @@ fn load_encrypted_signer_state<S: SignerStateStorage>(
...
-) -> Result<Option<v2::Signer>, PersistenceError> {
-    if let Some(encrypted_state) = storage.get_encrypted_signer_state(id)? {
-        let serialized_state = decrypt(private_key, &encrypted_state)?;
-        let state = serde_json::from_slice(&serialized_state)
-            .expect("Failed to deserialize decryoted state");
-        Ok(Some(v2::Signer::load(&state)))
-    } else {
-        Ok(None)
-    }
+) -> Result<StoredSignerStates, PersistenceError> {
```
kantai, line 1709: I am aware that this isn't really new to this PR, but the `signer.rs` file is getting pretty bloated, and these storage routines seem like perfect candidates for separating out into a separate module. How much of a refactor would that require?

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
```diff
@@ -177,6 +183,16 @@ fn get_signer_commitments(
...
     }
+
+    let computed_key = all_party_polynomials
+        .iter()
+        .fold(Point::default(), |s, (_, comm)| s + comm.poly[0]);
+
+    if &computed_key == expected_aggregate_key {
+        debug!("Aggregate key computed from combined DKG results match expected. Using this one");
+        return Ok(all_party_polynomials.into_iter().collect());
+    }
```
kantai, line 194: What case is this capturing? My understanding is that honest signers produce a DkgResults message that has all of the party polynomials required to compute the aggregate key. So, the only case in which DkgResults aren't recoverable via a single signer's DkgResults is if all of the signer's DkgResults messages are modified specifically to make the aggregate key unrecoverable by signers. This would be dishonest behavior on the part of all the signers, so why would they include any such information to recover the aggregate key at all (even aggregated across all of the signers)?

kantai (commented):
Just have a few questions

stacks-signer/src/runloop.rs
netrome: I can do that. However I think the plan is to remove the hard-coded `OsRng` in `wsts`. I created an issue for it https://github.com/Trust-Machines/wsts/issues/63, after Joey commented in my previous PR https://github.com/stacks-network/stacks-core/pull/4659#discussion_r1560665096

With that in mind, do you have any other preferences on how we manage RNGs here?

stacks-signer/src/signer.rs
netrome: This is to share the freshly loaded DKG shares with the Nakamoto sign coordinator. Without this the call to `get_signer_commitments` will load DKG shares from the last round which might not correspond to the one that was loaded by the signers.

I am not super happy with this solution, so if you have any improvement suggestions for how we can more properly communicate the appropriate shares to the sign coordinator, I'm all ears. The options I've considered were to add another message type for this message to more specifically convey that the signer has loaded a set of old shares, alternatively to implement persistence for old shares in the sign coordinator as well (which I'm not sure how complex it would be).

stacks-signer/src/runloop.rs
kantai: Using `OsRng` seems fine to me, but maybe a safer strategy would be to add something like:

```
/// This is the RNG implementation that the signer uses when randomness is required for cryptographic operations. Currently, this is OsRng, which is also the RNG used by WSTS when initializing signer state.
pub type SignerRng = OsRng;
```

And use that throughout. Then, whenever that WSTS issue is cleared, it'll be easy just set the RNG during initialization.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
netrome: So this is for the extra DkgResults message that is sent by signers when they are loading their persisted shares. In that scenario, the signer only has their own party polynomial available so the sign coordinator has to aggregate all of the DkgResults messages to get all of the party polynomials.

Again, I'm not super happy about that solution and open to suggestions - but any alternative I have considered so far would be a much more extensive change.

stacks-signer/src/signer.rs
netrome: Yes the miner-as-coordinator (`testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs`). It loads the DkgResults from stackerdb as you pointed out. The purpose of sending this message is to ensure that the latest DkgResults in stackerdb is the same as the ones the signer has just loaded on startup.

To clarify things consider the following scenario:
1. The signers run a successful DKG round and vote for the aggregate key for the current reward cycle.
2. For any reason, another DKG round was triggered and successfully run except that the voting in `signers-voting.clar` fails since an aggregate key is already set for the current reward cycle.

In this scenario, the latest DkgResults in stackerdb does not align with the approved aggregate key. Therefore, having the signers send new DkgResults messages whenever they load an old state ensures that the miner-as-coordinator can keep relying on fetching the latest DkgResults from stackerdb and still coordinate signing rounds.

stacks-signer/src/runloop.rs
netrome: > Then, whenever that WSTS issue is cleared, it'll be easy just set the RNG during initialization.

I'm not sure I follow what you mean here. The `Signer::reset` method won't change in the WSTS issue since it is already generic. The problem is that WSTS hard-codes the rng in some other places that calls this function which it should not do as a library.

I don't see what benefit a type alias would give us here. If we want to refactor the code so that we can easily change RNG in the future a simple type alias doesn't cut it, because most RNGs would have an internal state so you'd like to pass it around in the code instead of creating it on the fly which we can do with OsRng.

A middle-ground would be to create a method to return the rng as

```
/// This is the RNG implementation that the signer uses when randomness is required for cryptographic operations. Currently, this is OsRng, which is also the RNG used by WSTS when initializing signer state.
const fn signer_rng() -> impl CryptoRng { OsRng }
```

This would make it easier to swap between OsRng and ThreadRng, and possibly some other types of Rngs.

stacks-signer/src/signer.rs
kantai: It seems like the issue then is step (2) of that scenario.

If its really not possible to make sure that signers don't try to assemble DkgResults messages for reward cycles that have already selected their aggregate keys, it may make more sense to use a new message for "DkgResult for approved aggregate key".

stacks-signer/src/runloop.rs
kantai: It's so that whenever `SignerState::new()` is changed to `SignerState::new(rng: R)`, it's very clear that it should be invoked in the signer binary using `SignerState::new(&mut SignerRng)`.

netrome: Updated to use a single function as discussed yesterday ✅

stacks-signer/src/signer.rs
netrome: Done.

stacks-signer/src/signer.rs
netrome: Factored out to a `storage` module. Let me know if you have any opinions on how this module looks.

stacks-signer/src/signer.rs
netrome: Changed the ordering.

stacks-signer/src/signer.rs
netrome: Alright, based on your comments and our conversation yesterday I have updated the behavior so that the Signers will hold on to the DKG results message and send it only when an aggregate key is approved. This eliminates the need to also send this message when loading an old signer so I have removed that code.

testnet/stacks-node/src/nakamoto_node/sign_coordinator.rs
netrome: Removed this case now since I updated the signer to only send the messages after an aggregate key has been approved - so the extra DkgResults message doesn't exist any more and we don't need this case.

stacks-signer/src/runloop.rs
```diff
@@ -235,12 +236,35 @@ impl RunLoop {
...
+                    .get_approved_aggregate_key(reward_cycle)
+                    .map_err(backoff::Error::transient)
+            })?;
+
+            new_signer
+                .state_machine
+                .reset(dkg_id, &mut storage::crypto_rng());
+            new_signer.approved_aggregate_public_key = approved_aggregate_key;
+            new_signer
+                .load_saved_state()
```
jcnelson, line 259: I think you'll want to check that `approved_aggregate_key` is consistent with the private state you load here?

stacks-signer/src/signer.rs
```diff
@@ -334,20 +339,6 @@ impl From<SignerConfig> for Signer {
             signer_config.signer_entries.public_keys,
         );
 
-        if let Some(state) = load_encrypted_signer_state(
```
jcnelson, line 337: Why are we no longer instantiating the signer's state here?  Does this leave open the possibility of creating an invalid `Signer` object whose state is does not reflect the persisted signer state?

stacks-signer/src/signer.rs
```diff
@@ -1059,25 +1052,26 @@ impl Signer {
...
     /// Process a dkg result by broadcasting a vote to the stacks node
     fn process_dkg(&mut self, stacks_client: &StacksClient, dkg_public_key: &Point) {
-        let mut dkg_results_bytes = vec![];
+        // Abort if an aggregate key has already been set for the current rewards cycle
+        if let Some(aggregate_key) = stacks_client
+            .get_approved_aggregate_key(self.reward_cycle)
+            .map_err(|e| error!("{self}: Unable to get approved aggregate key: {e}"))
+            .unwrap_or_default()
+        {
+            self.approved_aggregate_public_key = Some(aggregate_key);
```
jcnelson, line 1061: Again, I think you'll want to check that the aggregate key is consistent with the saved state?

stacks-signer/src/signer.rs
```diff
@@ -1320,14 +1336,61 @@ impl Signer {
...
         let message = SignerMessage::EncryptedSignerState(encrypted_state);
 
         self.stackerdb.send_message_with_retry(message)?;
 
         Ok(())
     }
 
+    /// Load the saved signer state for the current dkg round
+    pub fn load_saved_state(&mut self) -> Result<(), storage::PersistenceError> {
+        let Some(aggregate_key) = self.approved_aggregate_public_key else {
```
jcnelson, line 1349: This check leads to surprising behavior.  Should this method, as named, _unconditionally_ load the state?

stacks-signer/src/signer.rs
```diff
@@ -1417,6 +1477,10 @@ impl Signer {
                     "{self}: updated DKG value from {old_dkg:?} to {:?}.",
                     self.approved_aggregate_public_key
                 );
+                if let Err(_) = self.load_saved_state_for_aggregate_key(approved_aggregate_key) {
+                    warn!("{self}: Failed to load saved state for key {approved_aggregate_key}");
+                }
+                self.send_dkg_results(&approved_aggregate_key);
```
jcnelson, line 1483: Should you be doing this if you failed to load your saved state?

jcnelson (changes requested):
The structure of this code is giving me a lot of pause.  It seems that there are several places where the signer state can just be willy-nilly reset from saved state, but without any consideration to how it will affect the execution of the signer at that point in the reload.  I fear this will lead to hard-to-reproduce bugs and state corruption, because it is hard to see in advance what the downstream consequences can be when changing this state mid-way through a main loop pass.

Instead, can this code be factored to make a state reload happen _only_ at the top-level of the main loop?  A reload should cancel any in-progress activities and force the signer to restart whatever it was doing anew.

stacks-signer/src/signer.rs
netrome: I've changed this now because the signer needs to know the approved aggregate key to load the right shares. This is not known in the config, and you need a stacks client to fetch it which the signer doesn't have access to.

I appreciate you highlighting this. I am not happy about the current code structure either. I'll add a proper constructor for the Signer so we don't have to be limited to the fields in `SignerConfig` for the initialization logic, and remove the `From<SignerConfig>` implementation which imposes unnecessary restrictions on how signers can be constructed without providing any value anywhere as far as I can see.

stacks-signer/src/signer.rs
netrome: Perhaps not, especially since it's not critical that all signers send these DKG results. So I can definitely condition the sending on these on the success of the load.

netrome:
> The structure of this code is giving me a lot of pause. It seems that there are several places where the signer state can just be willy-nilly reset from saved state, but without any consideration to how it will affect the execution of the signer at that point in the reload. I fear this will lead to hard-to-reproduce bugs and state corruption, because it is hard to see in advance what the downstream consequences can be when changing this state mid-way through a main loop pass.
> 
> Instead, can this code be factored to make a state reload happen _only_ at the top-level of the main loop? A reload should cancel any in-progress activities and force the signer to restart whatever it was doing anew.

I'm not too happy about the current structure either. While I have attempted to be careful with how and when we load state from the signer, I do share your concern. We've already seen plenty of hard-to-reproduce bugs in the signer so far, and I definitely don't want to add to it, especially since the intention of this change is to make the signer more robust towards bugs caused by network delays and timing issues.

I don't fully understand right now how exactly we'd be able to make the state reloads only happen at the top-level of the main loop (assuming you mean within the`SignerRunLoop::run_one_pass` function and not the top-level provided `SignerRunLoop::main_loop` function), but I will look into it with fresh eyes on Monday and see what I can do.

stacks-signer/src/signer.rs
netrome: Good point. Perhaps it's better to take the aggregate key as input, thus moving the condition to the calling code instead (unless the caller has the aggregate key already and not as an option). I'll change that.

stacks-signer/src/runloop.rs
netrome: I'll make the `load_saved_state()` function take the aggregate key as input instead so the caller don't have to worry about this.

stacks-signer/src/signer.rs
netrome: Yeah, I'll change `load_saved_state` to take the aggregate key as input to make the behavior more explicit here.

netrome:
I've updated now so that state reloads only happen in `signer::update_approved_aggregate_key` which is only called in `signer::refresh_dkg`, which in turn is only called from the main loop if no approved aggregate key is set. This eliminates some of the unnecessary boilerplate and allows us to instantiate the signer without doing yet another state reload after the signer has been constructed.

This feels much cleaner than the previous solution. Thank you for bringing this up @jcnelson. Let me know if you have further opinions on how to structure this. For example, if you want state reloads to be explicit in the main loop I can do some further refactors to achieve this - but I think the solution we have now fits better with the current structure of the signer.

stacks-signer/src/signer.rs
netrome: Updated so it's conditioned on a successful state load.

stacks-signer/src/signer.rs
netrome: Changed to only use the `load_saved_state_for_aggregate_key` method.

stacks-signer/src/signer.rs
netrome: I've purged this now, in favor of only doing the state reload from the main loop if an aggregate key is not set in the signer - which will always be the case for a new signer.

stacks-signer/src/runloop.rs
netrome: Removed this code, and changed to use only the `load_saved_state_for_aggregate_key` method in the only place we load state now.

stacks-signer/src/signer.rs
```diff
@@ -1514,6 +1453,12 @@ impl Signer {
                     "{self}: updated DKG value from {old_dkg:?} to {:?}.",
                     self.approved_aggregate_public_key
                 );
+                match self.load_saved_state_for_aggregate_key(approved_aggregate_key) {
+                    Ok(()) => self.send_dkg_results(&approved_aggregate_key),
+                    Err(e) => warn!(
+                        "{self}: Failed to load saved state for key {approved_aggregate_key}: {e}"
+                    ),
+                }
```
kantai, line 1461: I'm still a little confused by the logic here. `update_approved_aggregate_key` is invoked to refresh the signer's view of the approved aggregate key. Why should that trigger a load from saved state? Shouldn't the signer's state already be loaded into memory at this point? Why doesn't the signer _just_ load state into memory during process startup (as I think was the general idea behind Jude's suggestion)?

The general logic seems like it should be:

```
On boot:
    self.refresh_approved_aggregate_key();
    if self.approved_aggregate_public_key.is_some():
         self.signer_state = self.load_saved_states()[self.aggregate_key];
         if self.signer_state.pending_dkg_results.is_some():
            send_dkg_results();
            self.signer_state.pending_dkg_results.clear();
            self.overwrite_saved_states()

On refresh aggregate public key:
     if self.approved_aggregate_public_key is changed:
         if self.signer_state.pending_dkg_results.is_some():
            send_dkg_results();
            self.signer_state.pending_dkg_results.clear();
            self.overwrite_saved_states()
```

That way, the signer sends a DkgResults message after the aggregate key is approved if they're running at the time, or if they wakeup after the aggregate key has been approved, they send the message then.

kantai: This avoids excessively reloading the signer states _and_ it avoids resending the DkgResults message potentially every time a state reload occurs.

netrome: Hmm yeah I think we could do it that way. Two thoughts on the logic:

1. On boot the signer won't have any pending DKG results so we don't have to send DKG results at that point.
2. State writes happen now as soon as DKG is finished (i.e. a `DkgEnd(_)` message is received), which I think makes sense because you want to persist the shares as soon as you have confirmation that you have been able to create an aggregate key.

With these I think we can simplify the logic to

```
On boot:
    self.refresh_approved_aggregate_key();
    if self.approved_aggregate_public_key.is_some():
         self.signer_state = self.load_saved_states()[self.aggregate_key];

On refresh aggregate public key:
     if self.approved_aggregate_public_key is changed:
         if self.signer_state.pending_dkg_results.is_some():
            send_dkg_results();
            self.signer_state.pending_dkg_results.clear();

On DkgEnd received:
      self.save_signer_state()
```

What do you think about this?

netrome: > Why should that trigger a load from saved state? Shouldn't the signer's state already be loaded into memory at this point?

It probably should be, but the rationale behind the current logic is to ensure that the saved state is always consistent with the approved aggregate key. If we for any reason observe a change in the approved aggregate key, a natural reaction would be to reload the state observed with the changed aggregate key.

kantai: I think its possible to have a pending dkg results on boot, because the signer may have gone offline between `DkgEnd` and the key being approved by the network. I think we should still be persisting these states after the pending_dkg_results is cleared as well, so I think my suggested logic is about as simplified as it can get. I do agree that we should be persisting on `DkgEnd` as well.

netrome: Actually, a problem with the proposed logic is if you boot the signer and an approved aggregate key is not yet set you would never load any persisted state. If you are unlucky with the timing, your signer might boot just before the last vote for an aggregate key arrives and sets that key in the contract.

From that perspective, I think it is safer to load the saved state when you observe a change in the aggregate key.

To prevent excessive state reloads I could add a condition to only load the saved state in case the approved aggregate key is different from the one already present in the state.

netrome: So I think this logic would make sense:

```python
On boot:
     Start without any approved_aggregate_key or loaded signer state

On refresh aggregate public key:
     if self.approved_aggregate_public_key is changed:
         if the new approved_aggregate_public_key is not the same as the one in our signer state:
             self.load_saved_state(approved_aggregate_key)
         if self.signer_state.pending_dkg_results.is_some():
            send_dkg_results();
            self.signer_state.pending_dkg_results.clear();

On DkgEnd received:
      self.save_signer_state()
```

stacks-signer/src/storage.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/// This is the RNG implementation that the signer uses when randomness is required for cryptographic operations. Currently, this is OsRng, which is also the RNG used by WSTS when initializing signer state.
+pub(crate) const fn crypto_rng() -> impl CryptoRng + RngCore {
+    OsRng
+}
+
+/// Error stemming from a persistence operation
+#[derive(Debug, thiserror::Error)]
+pub enum PersistenceError {
```
jcnelson, line 93: So, the convention we follow in the codebase is to have a top-level `Error` structure for the module, and have _that_ encapsulate any specific sub-types of errors via an `enum`.  For example:

```
pub enum Error {
   Persistence(PersistenceError)
   Crypto(EncryptoinError)
}
```

And, we'd have `From` implementations for `Error` for each of its sub-types to make conversions easier.  Then, you don't have to force the callers in other modules to know what specific flavor of error needs to be returned; you just use the catch-all `Error` type from the module.  If the caller wants to deconstruct the `Error` type, they can do with a `match` statement or the like.

stacks-signer/src/signer.rs
netrome: This would avoid excessive state reloads, but ensure that the signer doesn't miss loading the state for a particular aggregate key.

stacks-signer/src/storage.rs
netrome: Alright, I'll update this.

stacks-signer/src/signer.rs
```diff
@@ -1438,6 +1480,42 @@ impl Signer {
...
+            "{self}: DKG public key changed. Sending DKG results";
+            "dkg_public_key" => %dkg_public_key
+        );
+        if let Err(e) = SignerMessage::serialize_dkg_result(
+            &mut dkg_results_bytes,
+            &dkg_public_key,
+            self.coordinator.party_polynomials.iter(),
+        ) {
+            error!("{}: Failed to serialize DKGResults message for StackerDB, will continue operating.", self.signer_id;
+                   "error" => %e);
```
jcnelson, line 1512: Shouldn't this return an error instead of soldering on and trying to insert the (presumably undefined?) `dkg_results_bytes` into `pending_dkg_results`?

stacks-signer/src/signer.rs
```diff
@@ -1438,6 +1480,42 @@ impl Signer {
...
+            warn!("{self}: No DKG results found for key: {dkg_public_key}");
+            return;
+        };
+
+        if let Err(e) = self
+            .stackerdb
+            .send_message_bytes_with_retry(&MessageSlotID::DkgResults, dkg_results_bytes)
+        {
+            error!("{}: Failed to send DKGResults message to StackerDB, will continue operating.", self.signer_id;
+                       "error" => %e);
```
jcnelson, line 1495: Shouldn't this return an error?

stacks-signer/src/signer.rs
```diff
@@ -1438,6 +1480,42 @@ impl Signer {
         Ok(())
     }
 
+    /// Send DKG results to the miner coordinator so that it can coordinate signing rounds
+    fn send_dkg_results(&mut self, dkg_public_key: &Point) {
+        let Some(dkg_results_bytes) = self.pending_dkg_results.remove(dkg_public_key) else {
+            warn!("{self}: No DKG results found for key: {dkg_public_key}");
+            return;
```
jcnelson, line 1487: Shouldn't this return an error?

stacks-signer/src/storage.rs
```diff
@@ -0,0 +1,117 @@
+use std::collections::VecDeque;
```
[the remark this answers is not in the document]
netrome: Thanks for spotting, will add :+1:

stacks-signer/src/signer.rs
netrome: Right, this was just inherited from the code I broke out. Returning an error is better. I'll update.

stacks-signer/src/signer.rs
netrome: Yeah, I'll add error handling to this function and the `process_dkg` function which calls it.

stacks-signer/src/signer.rs
netrome: Yes it should. This code was basically just moved out as-is from `process_dkg`. I'll improve the error handling.

stacks-signer/src/signer.rs
jcnelson: The way we've addressed this in the past (albeit in very different contexts) is to have two structs: a "preload" struct and a "loaded" struct which is instantiated from a "preload" struct and some additional state.  You'd have an `UninitializedSigner` struct, which represents a `Signer` that does not yet have its state loaded.  Its impl would have a conversion method like `into_signer(state)`, which fully instantiated the `Signer` from the loaded state once it became available.

Then, you'd make it impossible to accidentally have a partially-instantiated or corrupted `Signer` used somewhere.  Given the importance of the signer to chain liveness, I think it's worth the effort to make invalid Signer states unrepresentable here.

testnet/stacks-node/src/tests/signer.rs
```diff
@@ -714,6 +714,38 @@ impl SignerTest {
             .collect()
     }
 
+    /// Constructs and signs a transaction for each signer to vote for the provided key
+    fn build_vote_transactions(
+        &self,
+        key: Point,
+        reward_cycle: u64,
+        dkg_round: u64,
+    ) -> Vec<StacksTransaction> {
```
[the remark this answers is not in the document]
jcnelson: Oh wait nm this is in a test battery lol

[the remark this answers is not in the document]
netrome: Even in test code?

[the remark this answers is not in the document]
netrome: Right didn't get your other comment :sweat_smile:

stacks-signer/src/signer.rs
netrome: I agree that it makes a lot of sense to differentiate between an `UninitializedSigner` and a `Signer` instead of having the current `State` field within the signer. It's not clear to me what the relation is between an initialized/uninitialized signer and which DKG shares the signer should have loaded. Afaik, a Signer may still be initialized but without having a known approved aggregate key or any DKG shares.

stacks-signer/src/storage.rs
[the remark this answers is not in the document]
netrome: Added ✅

stacks-signer/src/signer.rs
netrome: Updated.

stacks-signer/src/signer.rs
netrome: Done. This does change the logic as you noted since the above function would continue on errors previously. I haven't observed any test failures due to this yet, but I don't know the signer well enough to guarantee that this is 100% safe - even though it feels like it should be better to return errors eagerly.

stacks-signer/src/signer.rs
netrome: Updated. I also changed the calling function to check for pending dkg results before calling this function to make the behavior more explicit and lead to fewer surprises.

stacks-signer/src/storage.rs
netrome: Done.

netrome:
Closing this PR due to us going in a different direction with the signer.

